PR #963, opened 2020-03-18
PBFT member client test

### LINKED ISSUE #952: member_clients_test makes merkle roots between replicas un-matched
Potentially has to do with how we sign in PBFT

`member_clients_test` should be enabled when this issue is resolved

olgavrou:
Resolves #952 

Enables `member_client_test` for pbft.

When pre-prepare is committed to the kv we now also store the hash of the pre-prepare signature in the signatures table (i.e. root). The pre-prepare signature (which contains the root of the merkle tree) will be stored in the ledger via the pre-prepare map.

### DIFF
--- a/src/consensus/pbft/pbfttypes.h
+++ b/src/consensus/pbft/pbfttypes.h
@@ -50,9 +50,10 @@ namespace pbft
     virtual kv::Version commit_pre_prepare(
       const pbft::PrePrepare& pp,
       pbft::PrePreparesMap& pbft_pre_prepares_map,
+      CBuffer root,
       ccf::Signatures& signatures) = 0;
     virtual kv::Version commit_tx(
-      ccf::Store::Tx& tx, CBuffer pp_sig, ccf::Signatures& signatures) = 0;
+      ccf::Store::Tx& tx, CBuffer root, ccf::Signatures& signatures) = 0;
     virtual std::shared_ptr<kv::AbstractTxEncryptor> get_encryptor() = 0;
   };
 
@@ -81,6 +82,7 @@ namespace pbft
     kv::Version commit_pre_prepare(
       const pbft::PrePrepare& pp,
       pbft::PrePreparesMap& pbft_pre_prepares_map,
+      CBuffer root,
       ccf::Signatures& signatures)
     {
       while (true)
@@ -97,12 +99,7 @@ namespace pbft
               auto pp_view = tx.get_view(pbft_pre_prepares_map);
               pp_view->put(0, pp);
               auto sig_view = tx.get_view(signatures);
-              crypto::Sha256Hash sig_hash(
-                {CBuffer{pp.digest_sig.data(), pp.digest_sig.size()}});
-              // we only need the hash of the pre prepare as the pre-prepare
-              // which contains the root of the merkle tree will be stored in
-              // the pbft_pre_prepares map
-              ccf::Signature sig_value(sig_hash);
+              ccf::Signature sig_value({root});
               sig_view->put(0, sig_value);
               return tx.commit_reserved();
             },
@@ -116,7 +113,7 @@ namespace pbft
     }
 
     kv::Version commit_tx(
-      ccf::Store::Tx& tx, CBuffer pp_sig, ccf::Signatures& signatures)
+      ccf::Store::Tx& tx, CBuffer root, ccf::Signatures& signatures)
     {
       while (true)
       {
@@ -124,11 +121,7 @@ namespace pbft
         if (p)
         {
           auto sig_view = tx.get_view(signatures);
-          crypto::Sha256Hash sig_hash({pp_sig});
-          // we only need the hash of the pre prepare as the pre-prepare
-          // which contains the root of the merkle tree will be stored in
-          // the pbft_pre_prepares map
-          ccf::Signature sig_value(sig_hash);
+          ccf::Signature sig_value({root});
           sig_view->put(0, sig_value);
           auto success = tx.commit();
           if (success == kv::CommitSuccess::OK)
--- a/src/consensus/pbft/pbfttypes.h
+++ b/src/consensus/pbft/pbfttypes.h
@@ -5,6 +5,7 @@
 #include "consensus/pbft/pbftpreprepares.h"
 #include "ds/ringbuffer_types.h"
 #include "kv/kvtypes.h"
+#include "node/signatures.h"
 
 namespace pbft
 {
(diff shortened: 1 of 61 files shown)


cimetrics:
pbft_member_client_test@[6080 aka 20200318.12](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6080&view=results) vs master ewma over 30 builds from [5800](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5800&view=results) to [6077](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6077&view=results)
images

src/consensus/pbft/pbfttypes.h
```diff
@@ -91,6 +96,14 @@ namespace pbft
...
               auto pp_view = tx.get_view(pbft_pre_prepares_map);
               pp_view->put(0, pp);
+              auto sig_view = tx.get_view(signatures);
+              crypto::Sha256Hash sig_hash(
+                {CBuffer{pp.digest_sig.data(), pp.digest_sig.size()}});
+              // we only need the hash of the pre prepare as the pre-prepare
+              // which contains the root of the merkle tree will be stored in
+              // the pbft_pre_prepares map
+              ccf::Signature sig_value(sig_hash);
+              sig_view->put(0, sig_value);
```
ashamis: I am confused by the logic here. Why not just include the signature what property does the hash give us? Especially since we need to pay for doing the extra hashing?

src/consensus/pbft/pbfttypes.h
```diff
@@ -102,13 +115,21 @@ namespace pbft
...
       {
         auto p = x.lock();
         if (p)
         {
+          auto sig_view = tx.get_view(signatures);
+          crypto::Sha256Hash sig_hash({pp_sig});
+          // we only need the hash of the pre prepare as the pre-prepare
+          // which contains the root of the merkle tree will be stored in
+          // the pbft_pre_prepares map
+          ccf::Signature sig_value(sig_hash);
```
ashamis: Same as above

src/crypto/hash.h
```diff
@@ -18,7 +18,7 @@ namespace crypto
     Sha256Hash();
     Sha256Hash(std::initializer_list<CBuffer> il);
 
-    std::array<uint8_t, SIZE> h;
+    std::array<uint8_t, SIZE> h = {};
```
ashamis: Why do we need to initialize this array every time we create a hash? Can we not only pay the cost when we need to initialize it?

src/node/signatures.h
```diff
@@ -32,6 +32,8 @@ namespace ccf
       commit(0)
     {}
 
+    Signature(crypto::Sha256Hash root_) : root(root_) {}
```
achamayou: What's the use case for this? Saving signatures with uninitialised term, index etc does not sound at all.

src/consensus/pbft/pbfttypes.h
olgavrou: The reason that is is getting added is for the members to be able to ack over the root of the latest signature. In pbft since the signature table is not being persisted to the ledger, we don't really need to store the actual signature or the other information at this point (view, seqno) since it is (1) stored in the pre-prepares table and (2) not used anywhere

src/crypto/hash.h
olgavrou: The array needs to be zeroe'd out for the genesis transaction, which queries the signature table for the root of the latest signature, but in genesis the signature table is empty, causing the default constructor to be called.

src/node/signatures.h
olgavrou: For PBFT all of this information is stored in the pre-prepare. It is the signature hash that we need to store in the signatures table for the members to use when they `ack`. I don't think we use the other signature information anywhere else, and since the signature table is not persisted to the ledger and the pre-prepare that contains all of that info is, I didn't think we should duplicate the data

src/consensus/pbft/pbfttypes.h
ashamis: discussed offline - I think this is fine

src/crypto/hash.h
ashamis: discussed offline - I think this is fine

src/consensus/pbft/pbfttypes.h
olgavrou: Replaced the hash over the pre-prepare signature with the merkle root of the pre-prepare

src/node/nodestate.h
```diff
@@ -1651,6 +1651,7 @@ namespace ccf
         *network.tables->get<pbft::RequestsMap>(pbft::Tables::PBFT_REQUESTS),
         *network.tables->get<pbft::PrePreparesMap>(
           pbft::Tables::PBFT_PRE_PREPARES),
+        network.signatures,
```
jumaffre: Any reason why this map is passed in like this while the request and pre-prepare tables are passed in with `*network.tables->get<>()`?

src/node/rpc/memberfrontend.h
```diff
@@ -681,10 +681,16 @@ namespace ccf
             HTTP_STATUS_BAD_REQUEST, "Submitted state digest is not valid");
         }
 
-        ma_view->put(
-          args.caller_id,
-          MemberAck(sig_view->get(0)->root, signed_request.value()));
-
+        const auto s = sig_view->get(0);
+        if (!s)
```
jumaffre: Just to confirm that my understanding is right: this would happen if a member acks before the first signature/pre-prepare has been emitted?

src/node/rpc/memberfrontend.h
```diff
@@ -713,8 +719,13 @@ namespace ccf
               fmt::format("No ACK record exists for caller {}", caller_id));
           }
 
-          auto root = sig_view->get(0)->root;
-          ma->state_digest = std::vector<uint8_t>(root.h.begin(), root.h.end());
+          auto s = sig_view->get(0);
+          if (s)
```
jumaffre: If `!s`, I guess there's no point `put`ting the same `ma` value in the `ma_view`?

src/node/rpc/test/membervoting_test.cpp
```diff
@@ -534,6 +534,15 @@ DOCTEST_TEST_CASE("Add new members until there are 7 then reject")
 
   DOCTEST_SUBCASE("ACK from newly added members")
   {
+    {
+      // make sure that there is a signature in the signatures table since ack's
+      // depend on that
```
jumaffre: Is this still true?

src/node/rpc/memberfrontend.h
olgavrou: Yes

src/node/rpc/memberfrontend.h
olgavrou: true :)

src/node/nodestate.h
olgavrou: No just left over code, I update it

src/node/signatures.h
achamayou: Can we still zero it though?

olgavrou: Zero'd

PR #965, opened 2020-03-19
Fast curve for users for performance tests with signatures

jumaffre:
The performance tests now use the fast bitcoin curve when generating users' identities (and members).  Note that there is no need to recompile CCF for this. Also renamed `args.default_curve` to `args.participants_curve`.


cimetrics:
fast_curve_performance@[6109 aka 20200319.12](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6109&view=results) vs master ewma over 30 builds from [5820](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5820&view=results) to [6095](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6095&view=results)
images

PR #966, opened 2020-03-19
Some RPCs are GET-only

### LINKED ISSUE #633: Modify read-only built-in endpoints to GET
All CCF built-in endpoints are POST at the moment for backwards compatibility with JSON-RPC. We should switch relevant endpoints such as getCommit etc to be GET methods.

eddyashton:
Resolves #633. 

This PR:
- Adds support for restricting which verbs are supported for a specific RPC (`install(...).set_http_get_only()`, or more generally `.restrict_allowed_verbs({HTTP_DELETE, HTTP_GET,...})`
- Sets some built-in RPCs to be GET-only (`getCommit`, `listMethods`, etc). Mostly things that are `Read` _and_ have simple params types
- Modifies the Python infra to take a specific verb, in particular using `c.get` rather than `c.rpc`
- Adds a url-decoder/unescaper. For GETs we need to put params in the query, not the body. Some of these queries need escaping (whenever we have a json-string _value_, its surrounding quotes need to be escaped). I spent a long time trying to convince Python-requests to _not_ escape the query, but all the workarounds are just arounds (they don't work), so I bit the bullet and wrote a quick escaper on the C++ side
- Related tidying (removing the unneeded `client.do` in Python, removing empty params so we have actually empty bodies rather than JSON nulls)

NB: I'm using "verb" to describe GET, POST etc. These are generally referred to as _methods_, but we already have a lot of code using `method` to describe the RPC-identifier. This is a bit clumsy - maybe we should do a renaming pass to make the old `method` a _resource_, and then we can use `method` instead of `verb`? Opinions welcome.


samples/apps/txregulator/clients/loader.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, network, name, country):
         network.consortium.add_users(primary, [self.name])
 
         with primary.user_client(user_id=self.name) as client:
-            self.ccf_id = client.rpc("whoAmI", {}).result["caller_id"]
+            self.ccf_id = client.get("whoAmI").result["caller_id"]
```
achamayou: We can probably scrap the result/error fields too.

samples/apps/txregulator/tests/txregulatorclient.py
```diff
@@ -99,20 +99,20 @@ def run(args):
...
             )
             check(
-                c.rpc("BK_register", {}), error=check_status(http.HTTPStatus.FORBIDDEN),
+                c.rpc("BK_register"), error=check_status(http.HTTPStatus.FORBIDDEN),
             )
 
         with primary.user_client(user_id=banks[0].name) as c:
             check(
-                c.rpc("REG_register", {}),
+                c.rpc("REG_register"),
```
achamayou: That probably wants to become “regulator/register” now.

src/node/rpc/handlerregistry.h
```diff
@@ -119,6 +122,43 @@ namespace ccf
...
         return *this;
       }
+
+      Handler& restrict_allowed_verbs(std::set<http_method>&& allowed_verbs)
+      {
+        func = [prev_func = func,
+                verbs = std::move(allowed_verbs)](ccf::RequestArgs& args) {
+          const auto verb = (http_method)args.rpc_ctx->get_request_verb();
+          if (verbs.find(verb) == verbs.end())
+          {
```
achamayou: Almost certainly pointless but I wonder if this would be faster as a string lookup in a string made of all allowed verbs concatenated.

jumaffre:
Looks like there are still some issues with the CI but the changes look good!

samples/apps/txregulator/tests/txregulatorclient.py
eddyashton: Agreed: https://github.com/microsoft/CCF/issues/968

samples/apps/txregulator/clients/loader.py
eddyashton: We could, but it may be useful to keep the distinction as they generally have different types (`result` is a dict-built-from-JSON, `error` is a string). I think ideally this response would look like (or even be?) a `requests.Response` object, so the caller could access the status/headers/contents in a standard way.

src/node/rpc/handlerregistry.h
eddyashton: Maybe, although we've actually got enum values rather than strings here (we can _get_ strings, but that extra step almost certainly makes it slower). Actually faster, and perhaps better, would be to build a bitmask of accepted verbs (34 verbs, comfortably in a uint64_t). With that approach we could avoid re-wrapping the func, which I'm a little uncomfortable with (it's already a deeply nested lambda), and put the "is this the correct verb" logic somewhere more obvious. I think I've talked myself into this while writing this comment, will look at doing it that way instead.

samples/apps/txregulator/clients/loader.py
achamayou: yes, that's what I meant. For the curl client, we might have to construct the Response slightly, but that shouldn't be too bad. I'll take a look at that.

src/node/rpc/handlerregistry.h
achamayou: But don't you get the enum from a map lookup/switch/equivalent on the verb string that comes in? With the string lookup you'd save that step?

eddyashton: `http-parser` does this conversion for us while its parsing the request.

eddyashton: The allowed verbs are now packed into a single `uint64_t`. A nice thing about this approach is that the set of verbs can be completely reset or altered by future calls to `set_allowed_verbs`, the same as any other `Handler.set_` operation (in contrast to the original lambda-wrapping approach, which could only add additional restrictions, and where repeated calls would actually produce an inaccurate `ALLOW` header response).

achamayou: Ah ok, then yeah, we might as well use a bitfield!

src/node/rpc/test/frontend_test.cpp
```diff
@@ -807,6 +832,81 @@ TEST_CASE("MinimalHandleFunction")
   }
 }
 
+TEST_CASE("Restricted verbs")
```
olgavrou: I am generally against unit tests that are in a loop because when one of them breaks it sometimes less straightforward to debug. Do you think the test would be too long if it wasn't done in a while loop?

cimetrics:
verb_selection@[6153 aka 20200320.10](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6153&view=results) vs master ewma over 30 builds from [5855](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5855&view=results) to [6149](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6149&view=results)
images

src/node/rpc/test/frontend_test.cpp
eddyashton: We could test a few specific cases rather than the full set, but it feels sensible to be systematic and get full coverage? In this case there are 34 verbs to test, which doesn't seem sensible to write out explicitly.

I hope the `INFO` line makes it obvious which loop iteration has failed, so a failure says `logged: PROBLEMATIC', and its trivial to locally comment out the loop and replace it with `verb = PROBLEMATIC`? And looping over all verbs has caught something in my refactor that would have been missed otherwise (my initial bit-masking only handled the first 32 due to a casting error, so this test failed on `UNLINK` after logging `UNLINK` as the first failing case).

PR #967, opened 2020-03-19
snmalloc for virtual builds + some fixes

ashamis:
- Use snmalloc for virtual builds (not SAN)
- fix a bug in the ringbuffer chunking code where we attempt to write to the end of a full buffer
- send + receive node to node channel synchronously when there is only 1 thread


cimetrics:
snmalloc_2@[6132 aka 20200319.24](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6132&view=results) vs master ewma over 30 builds from [5838](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5838&view=results) to [6121](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6121&view=results)
images

PR #970, opened 2020-03-20
Flatten msgpack-c

eddyashton:
Using CCF (outside of the `add_ccf_app` exported CMake function, for instance in a standalone test of your app frontend) requires setting various include directories. Maybe we'll export these in the install, but ideally they'd be obvious (independently discoverable).

Currently we need to add both `3rdparty/` and `3rdparty/msgpack-c/` to the include paths, because some of _our_ code talks about `msgpack-c/msgpack.hpp`, while the msgpack includes themselves use `"msgpack/..."`. This is annoying.

This PR flattens the msgpack directory - moving `3rdparty/msgpack-c/msgpack/` to `3rdparty/msgpack/`, and moving the single-include `.hpp` down into there. This is small amount of extra work next time we upgrade, but I think it makes the consumption story much simpler.

Related - CCF itself is also awkward. The install puts the headers under `install_dir/include/ccf/`, so ideally consumers would be able to refer to these headers as eg `ccf/ds/logger.h`. However the CCF files themselves refer to each other without the `ccf/` prefix, so we still need to add this dir explicitly. Fixing this should be part of the public include sanitisation.


cmake/common.cmake
```diff
@@ -114,7 +114,7 @@ include_directories(${CCF_DIR}/src)
 
 include_directories(
   SYSTEM ${CCF_DIR}/3rdparty ${CCF_DIR}/3rdparty/hacl-star
-  ${CCF_DIR}/3rdparty/msgpack-c ${CCF_DIR}/3rdparty/flatbuffers/include
```
eddyashton: A classic of the "Here's what I was actually trying to do" genre.

cimetrics:
flatten_msgpack@[6157 aka 20200320.12](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6157&view=results) vs master ewma over 30 builds from [5888](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5888&view=results) to [6155](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6155&view=results)
images

PR #962, opened 2020-03-18
Unauthenticated handlers

jumaffre:
Building block for #926 

This PR introduces a new `caller_auth_disabled` handler flag. If the flag is set (default to `false`), the handler does not go through the usual caller validation process as anyone can execute the corresponding endpoint. 

As well as unit tests (see `frontend_test.cpp`), I've added a new endpoint to the logging app that lets anonymous users (and only anonymous users) register "anonymous" messages. 

Most of the changes are in `frontend.h`:
- A lot of the code was moved from `process()` to `process_command()` as the handler now influences the authorisation policy. 
- Similarly, some of the forwarding logic was moved down to `process_command()`.

PR #971, opened 2020-03-20
PBFT and forwarding client certificates

ashamis:
- Do not forward the request signatures unless required - with PBFT
- Forward the signatures for the members frontend


src/node/rpc/memberfrontend.h
```diff
@@ -920,13 +920,6 @@ namespace ccf
       members(&network.members)
     {}
 
-    std::vector<uint8_t> get_cert_to_forward(
```
jumaffre: This should stay?

jumaffre:
As discussed, there's a more general bug here since https://github.com/microsoft/CCF/pull/962. We should no longer forward the caller certificate uniquely based on which frontend we are executing on but based on whether the handler required client identity too. I'll create an issue for this but it shouldn't block this PR.

cimetrics:
less_msg@[6162 aka 20200320.15](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6162&view=results) vs master ewma over 30 builds from [5890](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5890&view=results) to [6160](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6160&view=results)
images

src/node/rpc/memberfrontend.h
ashamis: yes you are right. The very first request needs to be signed, but it is signed by the primary node so that signature does not have much value. What we are doing in Raft is we are skipping verifying said signature and we should do the same thing when running PBFT

Issue #926, opened 2020-03-05
Self-registered Users


jumaffre:
Currently, new users are only permitted to issue transactions to applications after being voted in by members through the governance process. This is expensive as it requires member intervention and does not necessarily scale well. 

Instead, one option could be to let users self-register as long as they meet certain criteria. For example, assuming that users run in an enclave, they could provide their quote (containing the hash of their public certificate in the report data) via a new _unauthenticated_ `selfRegisterWithQuote` RPC. The CCF endpoint would verify the quote, check that the public certificate from the TLS connection matches the hash of the report data and that the client enclave runs a well-known `MRENCLAVE` (registered in advance by others members or users). 

-- Implementation details below -- 

This new `selfRegisterWithQuote` could be added as a new endpoint in the user frontend. However, the granularity of our current AuthZ scheme is at the `Frontend` level. What we would need here is a specific AuthZ scheme for that new endpoint, while the other endpoints would still be protected by the current `valid_caller()` function. 

One way to do this is would be allow apps to register several `UserRpcFrontEnd`s. That way, one frontend would define most of the endpoints guarded by `valid_caller()` while a second frontend would define the `selfRegisterWithQuote` handler and bypass `valid_caller()` (by overriding it). This would cause some changes in the way `libccf` registers the `UserRpcFrontend`(s) defined by the app. More generally, splitting an application in different classes/files would also be a nice way for users to build more modular applications.

jazzybluesea:
Thank you @jumaffre for opening this issue.  Regarding your statement "registered in advance by others members or users", could you please add some more detail on which rpc should be used pre-register the users?  Is it through the current member's mechanism to add a new user?

jumaffre:
@jazzybluesea "registered in advance by others members or users" in the issue description refers to the `MRENCLAVE` (i.e. the version of the code and enclave configuration) of the client enclave. This could be registered by members via the propose/vote RPCs (see https://microsoft.github.io/CCF/members/proposals.html#submitting-a-new-proposal - the member RPC can be pure Lua in this case and would not require changes to the CCF binaries).

jumaffre:
As discussed with @achamayou today, let's go for a simpler solution for now as the need for several frontends is not yet well justified. Instead, the new `selfRegisterWithQuote` RPC can be marked as "non-authorised" and the call to `valid_caller` can be skipped for it. 

[...]

PR #972, opened 2020-03-20
Users' quote verification

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/926

Building on https://github.com/microsoft/CCF/pull/962, this PR allows application endpoints to verify a quote supplied by an unknown (i.e. not registered in the `ccf.users` table) client against 1) the client certificate supplied by the TLS connection and 2) a list of trusted enclave measurements, stored in the new `ccf.users.code_ids` table. 

Assuming that clients are running in enclaves, this lets users self-register without going through a full round of governance to add a new user, as long as new users can provide evidence that they are running in a trusted enclave.

Since the solution is very much like the join protocol, I have refactored `quoteverification.h` (now `quote.h` as I moved quote generation code from `nodestate.h` in there too).

The aim is not to test this end-to-end (this would require setting up client enclaves, etc.) but provide the building blocks for app developers to do that.


src/apps/logging/logging.cpp
```diff
@@ -196,21 +199,143 @@ namespace ccfapp
...
-          auto view = tx.get_view(records);
-          view->put(in.id, log_line);
-          return make_success(true);
-        };
+      auto log_record_anonymous = [this](
+                                    RequestArgs& args,
+                                    nlohmann::json&& params) {
+        const auto in = params.get<LoggingRecord::In>();
+
+        std::vector<uint8_t> raw_quote = tls::raw_from_b64(
```
jumaffre: Obviously, all of this code will go before I merge this. This is a quote I self-generated on another node and extracted the code id from, just to prove myself that it works.

jumaffre: * In practice, the quote would be passed as as argument to the RPC.

cimetrics:
self_register_users@[6168 aka 20200320.18](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6168&view=results) vs master ewma over 30 builds from [5890](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5890&view=results) to [6160](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6160&view=results)
images

src/apps/logging/logging.cpp
```diff
@@ -196,21 +199,144 @@ namespace ccfapp
...
+          tls::cert_der_to_pem(args.rpc_ctx->session->caller_cert));
+        if (rc != QuoteVerificationResult::VERIFIED)
+        {
+          const auto [code, message] =
+            QuoteVerifier::quote_verification_error(rc);
+          return make_error(code, message);
+        }
+#endif
+
+        if (in.msg.empty())
```
olgavrou: could be nice to first check if the msg can be recorded and then check the quote?

jumaffre: It's generally good practice to first check authZ and then enforce application-specific checks. This code will go before I merge this PR anyway.

PR #973, opened 2020-03-23
Change assert to have more useful message

ashamis:
When the merkle roots do not match provide a more useful message in the error log


cimetrics:
change_assert@[6179 aka 20200323.2](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6179&view=results) vs master ewma over 30 builds from [5893](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5893&view=results) to [6177](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6177&view=results)
images

PR #974, opened 2020-03-23
Do not verify quotes for PBFT

ashamis:

PR #976, opened 2020-03-23
Log service state during service state query

eddyashton:
We have an [occasional test failure](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=6191&view=logs&j=480d76f6-ea78-52b2-3a24-b637e62df4cb&s=d654deb9-056d-50a2-1717-90c08683d50a&t=1c70dd88-c0d2-5741-cd2d-376784170dee&l=6026) on the CI, where it looks like PBFT is getting a different result on some nodes for this `/members/query` RPC. We have no more information on this crash, though it may be caused by the slow attestation that #974 avoids.

This PR adds logging of the service state during the execution of this RPC, so we can at least confirm how the service state differs on each node. This actually enables logging during member lua RPCs (previously only in the luageneric app for user commands), which should be useful for general debugging.


cimetrics:
logging_in_service_state_query@[6233 aka 20200323.30](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6233&view=results) vs master ewma over 30 builds from [5918](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5918&view=results) to [6227](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6227&view=results)
images

PR #977, opened 2020-03-23
Add sample handler and docs for non-JSON body

eddyashton:
We've supported non-JSON bodies in the framework for a while, but we didn't have a sample. This adds a sample in the C++ logging app, adds some docs, and tweaks the Python infra to allow these payloads though unhindered.


cimetrics:
non_json_body_sample@[6267 aka 20200323.47](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6267&view=results) vs master ewma over 30 builds from [5953](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=5953&view=results) to [6261](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6261&view=results)
images

sphinx/source/developers/logging_cpp.rst
```diff
@@ -57,17 +57,29 @@ The logging app defines :cpp:class:`ccfapp::LoggerHandlers`, which creates and i
     :end-before: SNIPPET_END: get
```
jumaffre: I wonder if it's worth adding a new page/section that describes the different types and options on handlers without mentioning the logging app specifically? I think we'll need one sooner or later to describe the different types/flags on handlers. We could always link to the logging app as examples (with back links too). Thinking about it, we should doxygen `handlerregistry.h` and get most of this for free. What do you think?

eddyashton: I agree! This is where we should aim, ideally with most of this coming 'for free' from doxygen. I think the next step is to split the thing apps see from the thing the framework sees (ie - they get an opaque `Handle` with the methods, we see the raw struct state), then the app side gets doxygen'd.

Naming is awkward - what's a 'handle' for a 'handler'? Maybe we should go more HTTP-heavy and talk about installing _resources_ or _endpoints_ rather than RPC handlers?

Anyway I think this is all future work.

sphinx/source/developers/logging_cpp.rst
```diff
@@ -57,17 +57,29 @@ The logging app defines :cpp:class:`ccfapp::LoggerHandlers`, which creates and i
...
+Each function is installed as the handler for a specific RPC ``method``, the name of the HTTP resource at which your handler will be invoked:
 
 .. literalinclude:: ../../../src/apps/logging/logging.cpp
     :language: cpp
     :start-after: SNIPPET_START: install_get
     :end-before: SNIPPET_END: install_get
     :dedent: 6
 
-These handlers use the simple signature provided by the ``handler_adapter`` wrapper function, which pre-parses a JSON params object from the HTTP request body.
+The return value from ``install`` is a ``Handler&`` object which can be used to alter how the handler is executed. For example, the handler for ``LOG_get`` shown above sets a `schema` for the handler, which will be used in calls to the ``/getSchema`` endpoint. It also marks the handler as `GET-only`, so the framework will return a ``405 Method Not Allowed`` for any requests which are not HTTP ``GET``.
```
eddyashton: This is wrong - I didn't actually mark `LOG_get` as GET-only yet.

sphinx/source/developers/logging_cpp.rst
achamayou: doxygen + links is absolutely the way to go! For naming, I think we should pick a popular framework, align, and reference explicitly.

PR #979, opened 2020-03-23
[IGNORE] Trying out the scale set pool for CI

achamayou:


achamayou:
Not quite ready for us to use yet, filed https://github.com/microsoft/azure-pipelines-agent/issues/2866

Issue #830, opened 2020-02-12
Installed CMake files should export more


eddyashton:
We have several CMake files and only some are installed for use by external repos, but these installed files refer to variables set in the non-installed files - so all external consumers need to reproduce those definitions.

- The installed CMake files expect the `TARGET` variable to have been set, so they should set it. `set(TARGET ....)` should move from `common.cmake` to `ccf_app.cmake`.

- These files also rely on `find_package(OpenEnclave ...)`, so this should also be done in the installed files.

- We may need to expose the `SERVICE_IDENTITY_CURVE` option, but we should check whether its safe for this to differ from the version in the installed `libccf`.

We may simply move these definitions, but it may be possible (and neater) to do this via `install(EXPORTS ...)` - needs investigation.

eddyashton:
Resolved by #986.

PR #986, opened 2020-03-24
Improve install with exported options

eddyashton:
Resolve items listed in #830.

- I've renamed our `TARGET` option to `COMPILE_TARGETS`, to distinguish it from all the other uses of `TARGET` in cmake
- I've used `add_compile_definitions` rather than `add_definitions`, for no precise benefit
- I've removed `SERVICE_IDENTITY_CURVE` definitions. If a user wants CCF itself to use a faster curve, they can change one line of code. But the installed headers as seen by the libccf build and the app build should match, without requiring too many obscure preprocessor defines


samples/apps/smallbank/smallbank.cmake
```diff
@@ -14,17 +14,10 @@ sign_app_library(
...
-      ${CMAKE_CURRENT_LIST_DIR}/tests/verify_small_bank_50k.json
-  )
-  set(SMALL_BANK_SIGNED_ITERATIONS 50000)
-else()
-  set(SMALL_BANK_SIGNED_VERIFICATION_FILE
-      ${CMAKE_CURRENT_LIST_DIR}/tests/verify_small_bank_2k.json
-  )
-  set(SMALL_BANK_SIGNED_ITERATIONS 2000)
-endif()
+set(SMALL_BANK_SIGNED_VERIFICATION_FILE
```
jumaffre: We now use the fast curve for all users' identities in the client signature perf tests. Should we default to 50,000 rather than 2,000?

cmake/common.cmake
```diff
@@ -25,29 +25,6 @@ find_package(Threads REQUIRED)
 
 set(PYTHON unbuffer python3)
 
-set(SERVICE_IDENTITY_CURVE_CHOICE
```
jumaffre: Is `DEFAULT_PARTICIPANTS_CURVE` still used in `function(add_e2e_test)`?

sphinx/source/quickstart/build.rst
```diff
@@ -44,11 +44,9 @@ The full list of build switches can be obtained by running:
 * **BUILD_TESTS**: Boolean. Build all tests for CCF. Default to ON.
 * **BUILD_SMALLBANK**: Boolean. Build SmallBank performance benchmark. Default to OFF.
 * **CLIENT_MBEDTLS_PREFIX**: Path. Prefix to mbedtls install to be used by test clients. Default to ``/usr/local``.
-* **SERVICE_IDENTITY_CURVE_CHOICE**: String, one of ``secp384r1``, ``secp256k1_mbedtls``, ``secp256k1_bitcoin``. Elliptic curve to use for CCF network and node identities. Defaults to ``secp384r1``.
```
jumaffre: Worth updating https://microsoft.github.io/CCF/developers/cryptography.html to mention this?

samples/apps/smallbank/smallbank.cmake
eddyashton: Maybe - not sure how reasonable this is for PBFT. I've made the change, will see what the CI looks like on this PR.

In a related change I've added a function to construct the `_VERIFICATION_FILE` from the `ITERATIONS`, so we don't have to repeat ourselves so often.

cmake/common.cmake
eddyashton: It was, but I don't think it needed to be - have removed it (if we change the default value in `curve.h`, it may not match the default argument to `--participants-curve`. I think this is fine).

sphinx/source/quickstart/build.rst
eddyashton: I don't see any reference to this option there, not sure what should be changed in the docs?

achamayou: I guess that bit:

```
They can be configured to use one of the following implementations:
```

jumaffre: I meant mentioning network and node identities in our cryptography page (which is rather generic atm).

samples/apps/smallbank/smallbank.cmake
jumaffre: I'd think longer is better/more accurate, as long as it doesn't take more than 1 minute to run the test in the CI?

eddyashton: `50003 transactions took 15747ms`

15s sounds reasonable, so I think 50k for all the signed tests is good.

sphinx/source/quickstart/build.rst
eddyashton: Have added a more up-to-date comment on service-curves vs user-curves, and how service-curve is selected.

sphinx/source/developers/cryptography.rst
```diff
@@ -44,8 +44,11 @@ Algorithms and Curves
...
+Public-key certificates, signatures, and ephemeral Diffie-Hellman key exchanges all rely on elliptic curves. The supported curves are listed in `tls/curve.h`:
 
- * secp384r1 from `mbedTLS <https://tls.mbed.org/>`_.
- * secp256k1 from `bitcoin core <https://github.com/bitcoin-core/secp256k1>`_.
+    .. literalinclude:: ../../../src/tls/curve.h
+        :language: cpp
+        :start-after: SNIPPET_START: supported_curves
+        :end-before: SNIPPET_END: supported_curves
+
+The ``service_identity_curve_choice`` determines the curve used by CCF itself, and can be changed to another value if a different curve is preferred. User and member certificates do not need to match this, and can be created on any supported curve.
```
jumaffre: It's a little unclear how the CCF curve can be changed (one would have to recompile libccf). Should we not mention it at all?

Also, "used by CCF itself" -> used by CCF for nodes and service identities perhaps?

eddyashton: Fair, have updated this.

cimetrics:
install_export_more@[6375 aka 20200325.2](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6375&view=results) vs master ewma over 30 builds from [6018](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6018&view=results) to [6373](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6373&view=results)
images

PR #989, opened 2020-03-24
Rename files for consistency

### LINKED ISSUE #707: File names are inconsistent.h
`memberfrontend.h` vs `start_type.h`. We primarily avoid underscores, but this produces some unreadable filenames. I believe we should add underscores everywhere we would otherwise have a space.

This should be done in a single PR, if at all.

eddyashton:
Resolves #707.

We have no fixed naming scheme for our file names. This PR adds one - we `snake_case.h` the basenames (~dirs are still `namespacelikeconcatenated`~ dirs are also now `./snake_case/snake_case.h`).

Opening as a draft so people have a chance to veto if they wish, but I'd like to clear this tonight/tomorrow morning so I don't spend too much time resolving conflicts.

Do we like this style?
Do we prefer oneword for some of our core concepts? eg this PR says `ring_buffer.h`, `key_pair.h`, etc, does this go too far?
~I've ignored `libbyz/` so far - should I lower-case those files in the same PR?~ This now also renames libbyz files to `lower_snake_case`. This reordered some includes in breaking ways, so there are additional include statements to correct this.


achamayou:
+1 for this convention. Can we have a quick script that checks all source files match that convention? [a-z_]+?

olgavrou:
I like this style, I prefer `ring_buffer` vs `ringbuffer` just so that it is consistent throughout. If you want to also do `libbyz` then go for it, but it can comfortably go into a different PR (might be quicker to merge this without libbyz at this point)

achamayou:
@eddyashton I suggest we align directories also, at least things like: /src/apps/jsgeneric/js_generic.cpp, as well as public headers, and probably binary artifacts too (js_generic.enclave.so.signed)?

src/consensus/raft/test/logging_stub.h
```diff
@@ -3,7 +3,7 @@
 #pragma once
 
 #include "consensus/raft/raft.h"
-#include "consensus/raft/rafttypes.h"
```
achamayou: That seems like it could become types.h

src/crypto/test/crypto.cpp
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the Apache 2.0 License.
 #define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
-#include "../cryptobox.h"
+#include "../crypto_box.h"
 #include "../hash.h"
-#include "../symmkey.h"
-#include "tls/base64.h"
```
achamayou: I think of base64 as one word, but maybe that's just me.

src/luainterp/tx_script_runner.h
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the Apache 2.0 License.
 #pragma once
-#include "luainterp/luainterp.h"
-#include "luainterp/luakv.h"
```
achamayou: does all the stuff under luainterp need lua_ prefixing?

jumaffre:
LGTM.

eddyashton:
`check-format.sh` now checks that filenames are lower_snake_cased. It can't check you've put underscores in sensible places, we'll have to enforce that manually. I've also converted libbyz, rather than adding an exception to it for this check.

src/crypto/test/crypto.cpp
eddyashton: Sure, I'll change this back.

src/consensus/raft/test/logging_stub.h
eddyashton: I prefer the duplication of `raft/raft_types.h` and `lua/lua_kv.h` to several files called `types.h` and `kv.h`, slightly easier to navigate the code? And ideally these generic names shrink and get removed eventually. Will discuss tomorrow.

achamayou:
@eddyashton thank you for taking the time to do this, the repo looks much cleaner now!

src/http/ws_upgrade.h
```diff
@@ -4,7 +4,7 @@
 
 #include "http_parser.h"
 #include "tls/base64.h"
-#include "tls/keypair.h"
+#include "tls/key_pair.h"
```
jumaffre: Since `key_pair.h` was split, this should actually be `tls/hash.h`.

Issue #51, opened 2019-05-13
Catastophic recovery using key shares


achamayou:
Use https://github.com/dsprenkels/sss adapted to run in an enclave to implement Shamir's Secret Sharing.

- [x] 1. Link sss to CCF enclave (#810)
- [x] 2. Add `ksk_m` to `ccf.members` table and generate in test infra/docs.
- [x] 3. Generate `k_z` on start-up, split it into "some" shares, encrypt the shares and keep encrypted shares in node's memory for now.
- [x] 4. Add new `ccf.shares` table and populate it with `(ledger secrets)k_z` and `[(shares_m)ksk_m]` on start-up.
- [x] 5. Serve members' shares on `getEncryptedRecoveryShare` RPC and allow members to easily decrypt them.
- [x] 6. Pass share to new `submitRecoveryShare` RPC (after vote).
- [x] 6'. Re-construct `k_z` when sufficient shares (i.e. secret sharing threshold) have been submitted, decrypt ledger secrets from `ccf.shares` table and initiate private ledger recovery.
- [x] 6''. Support for multiple recoveries.
**[When all the steps above are implemented, this should be enough to have a functioning, yet incomplete, recovery scheme]**
- [x] 7. Generate ~_endorsed_~ network encryption key `nk` and pass it via the join protocol. The public key should be output by the starting node.
- [x] ~8. Members should encrypt their shares with `nk_pub` before passing it to their vote RPC. CCF should decrypt the encrypted share on proposal completion.~
- [ ] 9. Resilience to elections: submitted shares should be stored in the KV + end-to-end test for election during recovery (will be addressed in https://github.com/microsoft/CCF/issues/1051). 
- [x] 10. Issue new member shares on consortium update.
- [x] 11. Issue new member shares on re-keying (to be addressed in https://github.com/microsoft/CCF/issues/1049).
- [ ] 12. Remove sealing and update docs.
- [x] 13. ~Always re-key the ledger when issuing new shares.~ Re-key ledger when retiring active member.
- [x] 14. Threshold should be defined in a new configuration table and updatable via governance.

**Misc*

- [ ] `ServiceStatus` should use a state machine, in a similar way to `NodeState`.

jumaffre:
[...]

PR #992, opened 2020-03-25
Multiple recoveries with key shares

jumaffre:
Resolves item 6'' in https://github.com/microsoft/CCF/issues/51.

Adds support for multiple recoveries when using key shares.

- We now store the (encrypted) msgpack-serialised ledger secrets in the KV (instead of the initial raw ledger secret). This means that CCF now stores the ledger secrets, along with the version at which they're applicable from. 
- CCF now issues new shares after the recovery completes (in the same way that we seal all secrets). 
- Some refactoring (especially in `sharemanager.h`). 

We now have a recovery scheme using key shares that is equivalent to our existing scheme using sealing keys. It still suffers from the following limitations which will be addressed next:
- Shares are not updated when a new member is added/removed, not when the ledger is rekeyed.
- We are still not resilient to elections during recovery (as submitted shares are not yet stored in the KV).


cimetrics:
multiple_recoveries_key_shares@[6381 aka 20200325.5](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6381&view=results) vs master ewma over 30 builds from [6018](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6018&view=results) to [6373](https://msrc-ccf.visualstudio.com/CCF/_build/results?buildId=6373&view=results)
images

Issue #953, opened 2020-03-13
Enable governance history test with PBFT


olgavrou:
governance maps are derived with pbft enabled and so they don't get written to the ledger

`governance_history` test should be enabled when this issue is resolved

jumaffre:
Looking at the `governance_history` test in more detail, it seems to me that the only thing we need to do is write a parser for the PBFT ledger that we can use from our e2e infra. Since requests get written to the ledger, we can then directly read the signed governance requests from there. Ideally the ledger format for Raft and PBFT should be unified.

ashamis:
moving to AFT

PR #994, opened 2020-03-25
In pbft keep tracks of the last pre-prepare that performed a governance request

ashamis:
related to #953

This change creates a "chain" of pre-prepares that contain requests that performed some kind of governance operation.  What this means that when a user gets their receipt (which covers a pre-prepare) they can make a determination if they want to audit the log and if they do they do not need to look into every pre-prepare.

### DIFF
--- a/src/node/rpc/frontend.h
+++ b/src/node/rpc/frontend.h
@@ -653,5 +653,10 @@ namespace ccf
     {
       return true;
     }
+
+    virtual bool is_members_frontend() override
+    {
+      return false;
+    }
   };
 }
--- a/src/consensus/pbft/libbyz/governance_update_tracking.h
+++ b/src/consensus/pbft/libbyz/governance_update_tracking.h
@@ -1,0 +1,60 @@
+#pragma once
+
+#include "ds/dl_list.h"
+#include "types.h"
+
+class GovernanceRequestTracking
+{
+public:
+  bool was_updated_in(Seqno seqno)
+  {
+    SeqnoWithMemberReq* current = seqnoWithMemberReqs.get_head();
+    while (current != nullptr && current->seqno >= seqno)
+    {
+      if (current->seqno == seqno)
+      {
+        return true;
+      }
+      current = current->next;
+    }
+    return false;
+  }
+
+  void rollback(Seqno seqno)
+  {
+    mark_stable(seqno);
+  }
+
+  void update(Seqno seqno)
+  {
+    seqnoWithMemberReqs.insert(new SeqnoWithMemberReq{seqno, nullptr, nullptr});
+  }
+
+  void mark_stable(Seqno seqno)
+  {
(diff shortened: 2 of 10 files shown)


src/consensus/pbft/libbyz/Replica.h
```diff
@@ -497,6 +499,10 @@ class Replica : public Node, public IMessageReceiveBase
   // used to register a callback for a client proxy to collect replies sent to
   // this replica
 
+  GovernanceRequestTracking gov_req_track;
+  // used to track the last series of pre-prepares that had requests that
+  // affected the goverenance of the service
```
jumaffre: governance

src/consensus/pbft/libbyz/Pre_prepare.cpp
```diff
@@ -34,6 +34,8 @@ Pre_prepare::Pre_prepare(
   rep().view = v;
```
jumaffre: Why is this change necessary? Ideally, the consensus should be agnostic to the type of transaction being executed?

src/consensus/pbft/libbyz/Replica.h
ashamis: thanks! my spell checking adding is really failing me lately. Any chance you have a good one?

src/consensus/pbft/libbyz/Pre_prepare.cpp
achamayou: I'm also curious.

src/consensus/pbft/libbyz/Pre_prepare.h
```diff
@@ -35,6 +35,8 @@ struct Pre_prepare_rep : public Message_rep
   View view;
   Seqno seqno;
   std::array<uint8_t, MERKLE_ROOT_SIZE> replicated_state_merkle_root;
+  uint64_t contains_gov_req;
```
achamayou: Why not bool?

src/node/rpc/frontend.h
```diff
@@ -653,5 +653,10 @@ namespace ccf
     {
       return true;
     }
+
+    virtual bool is_governance_front_end() override
```
achamayou: Members frontend?

src/consensus/pbft/libbyz/Pre_prepare.cpp
ashamis: spoke off-line and updated the comment

src/consensus/pbft/libbyz/Pre_prepare.h
ashamis: for alignment purposes :(

achamayou: Maybe good to have a comment then ;)

Issue #832, opened 2020-02-12
Setup daily build


jumaffre:
We should setup a CI nightly build to run tasks that a) take a relatively big amount of time and 2) may not necessarily need to be tested for each build. So far, these tasks include:
- Release build + full suite of tests (SGX only, with both CFT and BFT)
- An extended random test suite (after https://github.com/microsoft/CCF/issues/712). 
- TLS server test, as per https://github.com/microsoft/CCF/issues/626.

achamayou:
I suggest we also move the coverage build there for the following reasons:

1. it slows down the build and test
2. the codecov output on PRs is more distracting than valuable, from experience, and is more suitable for a periodic review of the project than a systematic per-PR examination.

achamayou:
I will also move the SAN build there, because post-PBFT merge, ninja w/ SAN is 5 minutes all by itself on a D16s_v3. It's good to have per build, but not so good that want slow builds.

achamayou:
Still to be done:

1. Release build and tests
2. TLS compliance test
3. Randomised test "suite", aka end to end fuzzing (probably over the SAN build)

PR #996, opened 2020-03-25
CI efficiency improvements

achamayou:
This is trying to collapse our CI into something more compact, so that it:

1. runs faster (ie. less than 5 minutes)
2. is easier to make changes to (ie. single common.yml with switches)
3. is cheaper (some stuff doesn't need custom VMs)

Some of that is achieved by pushing expensive tasks off to #832 (SAN, Coverage).


.azure-pipelines-templates/matrix.yml
```diff
@@ -1,188 +1,83 @@
...
-        ctest_args: '-LE "benchmark|perf|raft"'
-      CFT:
-        ctest_args: '-LE "benchmark|perf|pbft"'
+      ctest_args: '-LE "benchmark|perf" -j2'
     Release:
-      BFT:
-        ctest_args: '-LE "benchmark|perf|raft"'
-      CFT:
-        ctest_args: '-LE "benchmark|perf|pbft"'
+      ctest_args: '-LE "benchmark|perf" -j2'
```
eddyashton: If we're doing a release, should we just run the full test suite?

.azure-pipelines-templates/matrix.yml
```diff
@@ -1,188 +1,83 @@
...
-      dependsOn: ['Test_SGX_CFT_Release', 'Test_SGX_BFT_Release']
     Perf:
       pool: CCF-Perf
 
   build:
     common:
       cmake_args: '-DCMAKE_C_COMPILER_LAUNCHER="ccache" -DCMAKE_CXX_COMPILER_LAUNCHER="ccache"'
     NoSGX:
-      cmake_args: '-DCOMPILE_TARGETS=virtual -DCOVERAGE=ON'
+      cmake_args: '-DTARGET=virtual'
```
eddyashton: Looks like a bad merge - this should still be `COMPILE_TARGETS`

.azure-pipelines-templates/matrix.yml
achamayou: Sure, but unless it's also Perf, we don't record the values. Seems fair though.

.azure-pipelines-templates/matrix.yml
achamayou: Yep, it was. Also the SGX build was really both (oops!).

achamayou:
So as far as the tests go, there seem to be a number of prime candidates:

SGX:
image

No SGX:
image

.azure-pipelines-templates/matrix.yml
```diff
@@ -1,188 +1,83 @@
...
+      cmake_args: '-DCOMPILE_TARGETS=virtual'
     SGX:
-      cmake_args: ''
+      cmake_args: '-DCOMPILE_TARGETS=sgx'
     debug:
-      cmake_args: '-DCMAKE_BUILD_TYPE=Debug -DVERBOSE_LOGGING=ON -DBUILD_SMALLBANK=OFF'
+      cmake_args: '-DCMAKE_BUILD_TYPE=Debug -DBUILD_SMALLBANK=OFF'
     install:
       install_prefix: '/tmp/ccf-install'
       cmake_args: '-DCMAKE_INSTALL_PREFIX=/tmp/ccf-install'
```
olgavrou: SAN is moving to the nightly build?

achamayou: Yep #832

.daily.yml
```diff
@@ -0,0 +1,20 @@
+schedules:
```
achamayou: Can't set up the pipeline from a fork, annoyingly, so it'll have to get merged first.

achamayou:
Added files for Daily build, but can't set up from a fork, so I will set up the pipeline once this merges.

src/node/rpc/test/member_voting_test.cpp
```diff
@@ -821,7 +821,7 @@ DOCTEST_TEST_CASE("Propose raw writes")
   logger::config::level() = logger::INFO;
   DOCTEST_SUBCASE("insensitive tables")
   {
-    const auto n_members = 10;
+    const auto n_members = 3;
```
achamayou: Incremental store setup is kinda expensive, ~10 seconds in all for this test. Down to 3 members, it's consistently <1s, and I don't think it changes what we're testing functionally here.

Once this is done, ctest -L unit is consistently <3s, and there's no point using -j anymore. Small win, but still a win.

cmake/common.cmake
```diff
@@ -405,11 +405,10 @@ function(add_e2e_test)
       TEST ${PARSED_ARGS_NAME} APPEND
       PROPERTY ENVIRONMENT "PYTHONPATH=${CCF_DIR}/tests:$ENV{PYTHONPATH}"
     )
-    if(${PARSED_ARGS_IS_SUITE})
-      set_property(TEST ${PARSED_ARGS_NAME} APPEND PROPERTY LABELS suite)
-    else()
-      set_property(TEST ${PARSED_ARGS_NAME} APPEND PROPERTY LABELS end_to_end)
-    endif()
+    set_property(TEST ${PARSED_ARGS_NAME} APPEND PROPERTY LABELS end_to_end)
```
achamayou: This is to give us the freedom to split up stuff that doesn't parallelise well, like tests that do a lot of quote verification.

achamayou:
Current builds are about 20 minutes, Pipelines doesn't have a good way to show that, this is the best I can do:

image

After this change we are in the same ballpark:

image

But we run much fewer VMs, and they're less than fully busy, so we should be able to get more gains from running some tests concurrently without paying minutes for packing/unpacking. ccache should also be warmer in general. This also shaves about 20% of our YAML off, even after adding the Daily build.

To be followed by more test optimisations...

.azure-pipelines-templates/daily-matrix.yml
```diff
@@ -0,0 +1,21 @@
...
+      cmake_args: '-DCOMPILE_TARGETS=virtual'
+    daily:
+      cmake_args: '-DCMAKE_BUILD_TYPE=Debug -DSAN=ON -DCOVERAGE=ON -DBUILD_SMALLBANK=OFF'
+
+jobs:
+  - template: common.yml
+    parameters:
+      target: NoSGX
+      env: parameters.env.NoSGX
+      cmake_args: '${{ parameters.build.common.daily }} ${{ parameters.build.NoSGX.cmake_args }}'
```
eddyashton: Typo, first should be `parameters.build.daily.cmake_args`?

achamayou: Absolutely, thank you!

PR #998, opened 2020-03-26
Daily build

achamayou:
I think we want:

1. SAN
2. Coverage
3. Release build and tests
4. TLS compliance test
5. Randomised test "suite", aka end to end fuzzing (probably over the SAN build)

This only adds the first two, will do the rest after the release.

### DIFF
--- a/.daily.yml
+++ b/.daily.yml
@@ -1,7 +1,7 @@
 pr:
   branches:
     exclude:
-      - *
+      - '*'
 
 schedules:
 - cron: "0 3 * * Mon-Fri"
--- a/.daily.yml
+++ b/.daily.yml
@@ -2,16 +2,14 @@ pr:
   branches:
     include:
       - master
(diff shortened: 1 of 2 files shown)


.daily.yml
```diff
@@ -1,3 +1,8 @@
+pr:
+  branches:
+    include:
+      - master
```
eddyashton: This pipeline is currently running for every PR. I think we want `pr: none`.

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#pr-triggers

achamayou: Yes I know, I've put that back on purpose so I can make changes on this PR and see it run.

achamayou: (I did try to add a branch filter, but I couldn't get that to work because it comes from a fork).

PR #999, opened 2020-03-26
Enable 2 round trip receipts

ashamis:
Each pre-prepare and prepare are given a hash of a nonce and then in the reply we send the plain text nonce to the client-proxy.

To get two-round-trips latency, replicas would return a reply to the client-proxy after getting a pre-prepare and 2f matching prepares. The client-proxy will get its reply and receipt before the request is committed in the ledger, but we guarantee that the request is eventually committed to the ledger or we can prove that some replicas are malicious.

### DIFF
--- a/src/consensus/pbft/libbyz/replica.cpp
+++ b/src/consensus/pbft/libbyz/replica.cpp
@@ -73,6 +73,7 @@ Replica::Replica(
 #endif
   rep_cb(nullptr),
   global_commit_cb(nullptr),
+  entropy(tls::create_entropy()),
   state(
     this,
     mem,
@@ -477,7 +478,7 @@ void Replica::playback_request(ccf::Store::Tx& tx)
   vec_exec_cmds[0] = std::move(execute_tentative_request(
     *req, playback_max_local_commit_value, true, &tx, true));
 
-  exec_command(vec_exec_cmds, playback_byz_info, 1);
+  exec_command(vec_exec_cmds, playback_byz_info, 1, 0);
   did_exec_gov_req = did_exec_gov_req || playback_byz_info.did_exec_gov_req;
   brt.add_request(req.release());
 }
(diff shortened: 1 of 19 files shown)


src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -2001,8 +2006,8 @@ void Replica::process_new_view(Seqno min, Digest d, Seqno max, Seqno ms)
     if (primary() == id())
     {
       ByzInfo info;
```
olgavrou: why do we set the nonce to `0` when we add the pre-prepare to the `pc` but not when we add the prepare to the `pc`?

src/consensus/pbft/libbyz/prepared_cert.h
```diff
@@ -173,25 +177,30 @@ inline bool Prepared_cert::add(Prepare* m)
   return result;
 }
 
-inline bool Prepared_cert::add_mine(Prepare* m)
+inline bool Prepared_cert::add_mine(Prepare* m, uint64_t nonce)
```
olgavrou: regarding `add_mine` for both `Pre_prepare` and `Prepare`, since both prepare and pre prepare contain the nonce, can't we use the messages (`m`) to get the nonce and set it here instead of passing it in to the `add_mine` method(s)?

src/consensus/pbft/libbyz/replica.cpp
olgavrou: Is the nonce set correctly here on the `pp` message? Since now the nonce is included in pre-prepare and will go onto the ledger, I am just concerned about getting misaligned merkle roots here.

ashamis: I put the response to the question as a comment in the code

src/consensus/pbft/libbyz/prepared_cert.h
ashamis: if you have a look at the pre-prepare and prepare they only have the hashed nonce not the unhashed nonce.
This is passing the unhashed nonce

ashamis: I ended up rewriting a bit this code is different and the comment no longer applies

src/consensus/pbft/libbyz/replica.cpp
ashamis: I rewrote a bunch and now this comment no longer applies

PR #1001, opened 2020-03-27
Late joiners plus view changes

### LINKED ISSUE #837: Pre prepares need to be stored with the same view in the kv
Follow up from #825 

If at least one replica has prepared on a seqno n and the pre-prepare for that seqno is signed, that replica will have globally committed on seqno n, and the pre-prepare will be stored in the kv for that replica in view v.

If there is a view change in the mean time and the other replicas haven't prepared seqno n yet, they will have to rollback before n, and during the view change they will re-execute the request for seqno n, but the pre-prepare will be stored in the kv with view v' which will cause a merkle root mis-match between the replicas.

To solve this, while executing prepared requests in a view change, we need to store the pre-prepares into the kv with the view v that the seqno n was prepared in.

### LINKED ISSUE #859: PBFT should globally commit on commit not on prepare
Right now while in pbft, each replica calls the globally committed callback (which calls `compact` on the kv) when it has prepared and before that replica sends out its own commit message.

Sequence numbers that have been prepared can be rolled back, so it is important to be able to roll back before those sequence numbers. Therefore the globally committed callback should be called when we have committed and not on prepare.

olgavrou:
Resolves #837 
Resolves #859 

- if a request has been executed on view `v` for a majority of replicas but but is executed during the view change from `v` to `v'` on a different replica, then the request is seen as executed on view `v`. This means it will be encrypted with view `v` set in the IV and the pre-prepare will be stored in the ledger with view `v` for this replica
- various small bug fixes
- global commit is called on commit and not on prepare (#859) except for when `f == 0` where we need to globally commit earlier
- remove `test_UDP_with_delays` and related files

late joiners test changed slightly to avoid adding nodes after the network has been opened  as we don't currently support reconfiguration with pbft. It now starts with 3 active replicas, and the late joiner is the fourth replica


src/consensus/pbft/libbyz/pre_prepare.cpp
```diff
@@ -474,4 +474,24 @@ const std::array<uint8_t, MERKLE_ROOT_SIZE>& Pre_prepare::
...
+{
+  if (rep().num_prev_pp_sig == 0)
+  {
+    return {};
+  }
+  std::vector<int> principals_with_proof;
+  // signatures start at
+  uint8_t* sigs = (uint8_t*)contents() + sizeof(Pre_prepare_rep) +
+    rep().rset_size + rep().n_big_reqs * sizeof(Digest);
+  auto s = sigs;
```
achamayou: Why create s? sigs doesn't seem to be otherwise reused?

src/consensus/pbft/pbft.h
```diff
@@ -434,6 +434,7 @@ namespace pbft
         {
           gb_info->view_change_list->emplace_back(view, version);
         }
```
achamayou: That's not right, the KV manages the history, including compacts and rollbacks.

tests/late_joiners.py
```diff
@@ -53,46 +53,46 @@ def find_primary(network):
...
-        t.start()
-    else:
-        node.resume()
-
-
-def assert_node_up_to_date(check, node, final_msg, final_msg_id):
+def assert_node_up_to_date(check, node, final_msg, final_msg_id, timeout=5):
     with node.user_client() as c:
-        for x in range(0, 5):
+        while timeout > 0:
```
achamayou: We'll switch to a library but for now let's do timeouts consistently with the code now that Eddy has updated it and use elapsed time.

tests/late_joiners.py
```diff
@@ -153,19 +153,19 @@ def run(args):
 
             # check that a new node can catch up after all the requests
             LOG.info("Adding a very late joiner")
-            last_joiner_node = network.create_and_trust_node(
+            late_joiner = network.create_and_trust_node(
                 lib_name=args.package, host="localhost", args=args,
             )
```
achamayou: create_and_trust_node raises on error, it does not return None, the assert is not useful here.

tests/late_joiners.py
```diff
@@ -186,7 +186,7 @@ def run(args):
                     if i % 2 != 0 and args.consensus == "pbft":
                         continue
                     LOG.success(f"Will suspend node with id {node.node_id}")
-                    t = random.uniform(1, 10)
+                    t = random.uniform(1, 2)
```
achamayou: Why is the value 2 here? This seems like it should be calculated on the basis of the pbft timeout?

src/consensus/pbft/pbft.h
ashamis: I think what you are trying to do is ensure that the history is applied when we compact something?
If that is the case then execute pending should be called by compact itself as it would make sense to do that.
Additionally, we do not want to apply everything just the versions that are <= the version you are compacting at.

src/consensus/pbft/pbft.h
```diff
@@ -446,6 +447,7 @@ namespace pbft
         global_commit_cb, &register_global_commit_ctx);
 
       auto rollback_cb = [](kv::Version version, RollbackInfo* rollback_info) {
+        rollback_info->store->get_history()->discard_pending();
```
ashamis: this is similar to the above, discard should be part of rollback, and only discarding the versions that are >= to the target version of the rollback call

src/consensus/pbft/libbyz/pre_prepare.cpp
olgavrou: That's just leftover from something else, thanks for spotting it

src/consensus/pbft/pbft.h
olgavrou: discard is called from `history::rollback`, but I think the correct thing to do is to discard any versions that are `>` than the target version of the rollback, instead of `>=`

src/consensus/pbft/pbft.h
olgavrou: moved to be handled by history

tests/late_joiners.py
olgavrou: The first timeout indicates how many seconds to wait until we first suspend the node(s). The second timeout, which is calculated based on the pbft timeout, is how many seconds we let the node(s) stay suspended before resuming, causing view changes. If we wait too long before we suspend the nodes in the first place then the test might finish without view changes actually happening.

src/node/history.h
```diff
@@ -383,6 +381,35 @@ namespace ccf
     std::optional<ResultCallbackHandler> on_result;
     std::optional<ResponseCallbackHandler> on_response;
 
+    void execute_pending(kv::Version v)
```
achamayou: So that flushes already executed entries in the pending_inserts queue into the history unless I am missing something. execute_pending() doesn't seem like the best name for it.?

src/node/history.h
```diff
@@ -383,6 +381,35 @@ namespace ccf
...
+        if (p->version <= v)
+        {
+          add_result(p->id, p->version, *p->replicated);
+          pending_inserts.remove(p);
+        }
+        p = next;
+      }
+    }
+
+    void discard_pending(kv::Version v)
```
achamayou: Considering those methods are identical except for one line, I think we should consider a single flush_pending(kv::Version v, bool apply).

src/consensus/pbft/libbyz/pre_prepare.cpp
```diff
@@ -474,4 +474,23 @@ const std::array<uint8_t, MERKLE_ROOT_SIZE>& Pre_prepare::
...
   return rep().ctx;
+}
+
+std::vector<int> Pre_prepare::get_prev_pp_valid_principals()
+{
+  if (rep().num_prev_pp_sig == 0)
+  {
+    return {};
+  }
+  std::vector<int> principals_with_proof;
```
ashamis: you can pre-allocate this vector

src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -479,6 +479,57 @@ void Replica::playback_request(ccf::Store::Tx& tx)
...
+
+void Replica::populate_certificates(Pre_prepare* pp, bool add_mine)
+{
+  if (pp->seqno() <= 0)
+  {
+    // first pre prepare will not contain proofs for a previous pre prepare
+    return;
+  }
+
+  auto principals_with_proofs = pp->get_prev_pp_valid_principals();
```
ashamis: you can remove this function call & vector allocation by just combining it with line 506

src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -479,6 +479,57 @@ void Replica::playback_request(ccf::Store::Tx& tx)
...
+  auto prev_seqno = pp->seqno() - 1;
+  if (!plog.within_range(prev_seqno))
+  {
+    LOG_DEBUG_FMT(
+      "seqno {} is out of range, can not add prepare proofs to plog",
+      prev_seqno);
+    return;
+  }
+  auto& prev_prepared_cert = plog.fetch(prev_seqno);
```
ashamis: move this inside the scope of  "if (prev_pp != 0)"

tests/late_joiners.py
```diff
@@ -53,46 +53,48 @@ def find_primary(network):
...
-        for x in range(0, 5):
+        end_time = time.time() + timeout
+        while time.time() < end_time:
             try:
                 check(
                     c.get("LOG_get", {"id": final_msg_id}), result={"msg": final_msg},
                 )
                 return
-            except TimeoutError:
+            except (TimeoutError, requests.exceptions.ReadTimeout,) as e:
```
achamayou: Why do we need to specifically catch read timeout here? Is it because the enclave blocks completely during catch up?

src/consensus/pbft/libbyz/pre_prepare.cpp
olgavrou: I wouldn't worry about this on this PR as it will change on the next one where we will also be verifying the proofs

src/consensus/pbft/libbyz/replica.cpp
olgavrou: but we use `prev_prepared_cert` to get `prev_pp` and then check if it is `0`  or not

tests/late_joiners.py
olgavrou: no, but there might be an early view change triggered and in progress, in which case we will not get responses to `LOG_get` requests right away

src/node/history.h
olgavrou: renamed `execute_pending` to `flush_pending`

src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -518,14 +569,19 @@ void Replica::playback_pre_prepare(ccf::Store::Tx& tx)
 
     last_te_version = ledger_writer->write_pre_prepare(tx, executable_pp.get());
 
+    global_commit(executable_pp.get());
```
ashamis: why does this need to occur before last_executed++?

src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -2006,20 +2063,23 @@ void Replica::process_new_view(Seqno min, Digest d, Seqno max, Seqno ms)
...
       ByzInfo info;
       pc.add_old(pp);
-
       if (execute_tentative(pp, info))
       {
         if (ledger_writer && req_in_pp > 0)
         {
+          pp->set_view(prev_view);
           last_te_version = ledger_writer->write_pre_prepare(pp);
+          pp->set_view(v);
```
ashamis: it seems like the view should be pass to write_pre_prepare and it can be responsible for setting the view on the pre-prepare?

src/consensus/pbft/libbyz/replica.cpp
olgavrou: because we open the network on global commit, and we do `seqno_at_last_f_change = last_executed + 1` there, so we should not update `last_executed` before that

src/consensus/pbft/libbyz/prepare.cpp
```diff
@@ -41,7 +47,7 @@ Prepare::Prepare(View v, Seqno s, Digest& d, Principal* dst, bool is_signed) :
       signature(Digest d_, NodeId id_) : d(d_), id(id_) {}
     };
 
-    signature s(d, pbft::GlobalState::get_node().id());
+    signature s(d, rep().id);
```
ashamis: why this change? we want this node's id.

src/consensus/pbft/libbyz/prepare.cpp
```diff
@@ -27,6 +28,11 @@ Prepare::Prepare(View v, Seqno s, Digest& d, Principal* dst, bool is_signed) :
   rep().seqno = s;
   rep().digest = d;
 
+  if (id < 0)
+  {
+    rep().id = pbft::GlobalState::get_node().id();
+  }
+
```
ashamis: In what circumstance can this happened? can you please add a comment?

src/consensus/pbft/libbyz/prepare.cpp
olgavrou: Because we create `Prepare`s from the pre-prepare proofs in playback and we need to set the correct id otherwise the digest will be wrong

src/consensus/pbft/libbyz/prepare.h
```diff
@@ -53,7 +53,12 @@ class Prepare : public Message
   Prepare(uint32_t msg_size = 0) : Message(msg_size) {}
 
   Prepare(
-    View v, Seqno s, Digest& d, Principal* dst = 0, bool is_signed = false);
+    View v,
+    Seqno s,
+    Digest& d,
+    Principal* dst = 0,
+    bool is_signed = false,
+    int id = -1);
```
ashamis: please add a comment for what id is?

src/consensus/pbft/libbyz/prepare.cpp
olgavrou: This is to indicate that the replica id is not set (see above for when it will/should be set), so that means that the the replica that is creating the Prepare is the same replica issuing the Prepare, so we can just set the node id

src/consensus/pbft/libbyz/prepare.cpp
```diff
@@ -56,7 +62,6 @@ Prepare::Prepare(View v, Seqno s, Digest& d, Principal* dst, bool is_signed) :
   rep().prepare_sig_size = 0;
 #endif
 
-  rep().id = pbft::GlobalState::get_node().id();
```
ashamis: without this line is rep().id not set if if >=0?

olgavrou: id has nothing to do with `f`, please see my comments above

src/node/history.h
```diff
@@ -381,6 +381,20 @@ namespace ccf
...
 
+    void discard_pending(kv::Version v)
+    {
+      auto* p = pending_inserts.get_head();
+      while (p != nullptr)
+      {
+        auto* next = p->next;
+        if (p->version > v)
+        {
+          pending_inserts.remove(p);
```
ashamis: this is a memory leak, p needs to be deleted

src/node/history.h
```diff
@@ -381,6 +381,20 @@ namespace ccf
     std::optional<ResultCallbackHandler> on_result;
     std::optional<ResponseCallbackHandler> on_response;
 
+    void discard_pending(kv::Version v)
+    {
+      auto* p = pending_inserts.get_head();
```
ashamis: pending_inserts is not thread safe it can only be accessed under a lock
please have a look at execute_pending.

Also while you are here could you please rename execute pending to flush pending. I did not give it a great name :S

src/consensus/pbft/libbyz/replica.cpp
olgavrou: I think it should be clear that we are setting the view during view change here instead of having `write_pre_prepare` hide that detail? Also I think that `write_pre_prepare` should just write things to the ledger and not implement any more logic than it needs to.

src/consensus/pbft/libbyz/nv_info.cpp
```diff
@@ -869,21 +869,22 @@ bool NV_info::check_new_view()
...
-Pre_prepare* NV_info::fetch_request(Seqno n, Digest& d)
+Pre_prepare* NV_info::fetch_request(Seqno n, Digest& d, View& prev_view)
 {
   PBFT_ASSERT(is_complete, "Invalid state");
   PBFT_ASSERT(n > nv->min() && n < nv->max(), "Invalid arguments");
 
   Pre_prepare* pp = 0;
   Digest null;
   View pv = vcs[nv->which_picked(n)].vc->req(n, d);
+  prev_view = v;
```
ashamis: set this in the else statement?

src/consensus/pbft/libbyz/replica.cpp
ashamis: The issue here is that this logic is repeated twice and is not clear what it does by just glancing at it.

And alternative is create a wrapper function.

olgavrou: fair enough :)

tests/late_joiners.py
ashamis: It seems like we would want to increase the timeout then just swallowing the error

src/consensus/pbft/libbyz/pre_prepare.cpp
ashamis: This function is used in one place and can just be combined with said function

src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -479,6 +479,56 @@ void Replica::playback_request(ccf::Store::Tx& tx)
...
+  if (!plog.within_range(prev_seqno))
+  {
+    LOG_DEBUG_FMT(
+      "seqno {} is out of range, can not add prepare proofs to plog",
+      prev_seqno);
+    return;
+  }
+  auto& prev_prepared_cert = plog.fetch(prev_seqno);
+  auto prev_pp = prev_prepared_cert.pre_prepare();
+  if (prev_pp != 0)
```
ashamis: prev_pp != nullptr

src/consensus/pbft/libbyz/replica.cpp
olgavrou: Done

src/consensus/pbft/libbyz/nv_info.cpp
olgavrou: Done

tests/late_joiners.py
```diff
@@ -200,8 +201,8 @@ def run(args):
 
                     # if pbft suspend the primary more than the other suspended nodes
                     if node.node_id == cur_primary_id and args.consensus == "pbft":
-                        et += et * 0.5
-                    tm = Timer(t, timeout, args=[node, True, et])
+                        et += et
```
ashamis: I am confused what this does so instead of multipling the timeout by 1.5 it is now multiplied by 2? What do either of these numbers mean? Can you please add a comment

olgavrou: Sure I will add comments in all places that have raised uncertainty in this test

olgavrou: For clarity, this here wasn't really been used up till now (we were never testing view changes), so this value wasn't really assessed. The reason for the increase comes from the need to make sure that view changes will be triggered, and so making the primary node stay suspended for double the view change timeout (which defaults to `5000`) should always cause a view change to be triggered.

tests/late_joiners.py
olgavrou: Added comment here, pasting here for clarity:

        # Wait until final_msg_id is available in the node.
        # We need to catch timeout and assertion errors as
        # the node (e.g. late joiner) might not be up to date yet
        # or we might be in the middle of a view change so the node
        # will not be responsive. This timeout should be reduced when
        # checkpoints are implemented making catchup take less time

src/consensus/pbft/libbyz/nv_info.h
```diff
@@ -98,11 +98,13 @@ class NV_info
...
   // view-change's) or deletes vca (otherwise) and returns true.
 
-  Pre_prepare* fetch_request(Seqno n, Digest& d);
+  Pre_prepare* fetch_request(Seqno n, Digest& d, View& prev_view);
   // Requires: "complete" and "n" is the sequence number of a request
   // in new-view message.
   // Effects: Sets "d" to the digest of the request with sequence
-  // number "n". If enough information to make a pre-prepare is available, it
+  // number "n". Sets "prev_view" to the view that this request was executed in
+  // if it is has completed, otherwise sets it to the current view.
```
achamayou: Extra is?

PR #1008, opened 2020-03-27
Build and test release in Daily

achamayou:
Continue #832


.azure-pipelines-templates/daily-matrix.yml
```diff
@@ -7,15 +7,30 @@ parameters:
...
+
+jobs:
+  - template: common.yml
+    parameters:
+      target: SGX
+      env: '${{ parameters.env.SGX }}'
+      cmake_args: '${{ parameters.build.install.cmake_args }} ${{ parameters.build.SGX.cmake_args }}'
+      suffix: 'Release'
+      artifact_name: 'SGX_Release'
+      ctest_filter: '-LE "benchmark|perf"'
```
eddyashton: This is label _exclude_, I think we want `-L` here (and `-LE` on the `NoSGX Debug` job?).

eddyashton: Actually I guess we don't want the perf tests during Daily, unless we're aiming to run this job in the perf pool?

achamayou: No we don't, that's why I excluded them?

The perf stuff is run on the Perf pool by the once a day CI already, so I suggest it should stay out of the Daily altogether.

I'm less sure about the NoSGX Debug job, where we have SAN. That seems like it might usefully want to run the perf stuff, not for the numbers, but in case it's triggering safety issues that aren't otherwise visible in our slower paced integration tests.

So I'm going to go with "this is what we want here"?

eddyashton: I'm convinced

PR #1010, opened 2020-03-30
Configure PR user and EWMA span for CI Metrics

achamayou:
Pick up fixes from https://github.com/jumaffre/cimetrics/releases/tag/v0.2.10

### DIFF
--- a/metrics.yml
+++ b/metrics.yml
@@ -2,4 +2,6 @@ db: 'ccf'
 collection: 'perf'
 view: 'trend'
 columns: 3
-span: 30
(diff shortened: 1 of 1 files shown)


achamayou:
5 seems a little too short, also this will test the comment update.

achamayou:
Picture upload has failed on an apparently mundane get, perhaps because of capacity issues.

achamayou:
Ok, update works. I personally think 7 is a good ewma window, but we can revisit that later if needed.

PR #880, opened 2020-02-24
Cache quote inside enclave

jumaffre:
Further to https://github.com/microsoft/CCF/pull/875 and the intermittent node startup failures. 

Instead of retrieving the quote for every single join request, cache the quote inside the enclave (retrieved once, at startup) and use that instead.

Draft PR for now as I was not able to confirm locally whether this fix resolved the issue.

PR #680, opened 2020-01-08
Fix reconfiguration end-to-end test

jumaffre:
The `reconfiguration` end-to-end test had become very flaky recently and this PR fixes the instability. 

The issue was that for the reconfiguration test, a node with a untrusted code version was trying to join the network, triggering its quote verification by the leader. Because the quote verification currently takes ~2 secs and the join timer was around the same period (see `nodestate.h`), the leader's node ring buffer was getting filled with expensive join requests, either not processing other join requests or client connections.

The join timer has now been relaxed in our tests to 4 seconds (configurable via `cchost`, default to 1 second) and our python client waits for a joining node to show up in the leader's store for twice that duration.

PR #1011, opened 2020-03-30
Reduce timeouts in slow CI tests

eddyashton:
This started as a minor change, and got into a rabbit hole of debugging election failures. Our timeouts appear to have some relationships I don't understand, and setting the election timeout too low breaks some tests in ways I haven't been able to fix.

I'll document the changes inline.


tests/infra/ccf.py
```diff
@@ -305,7 +303,7 @@ def create_and_add_pending_node(self, lib_name, host, args, target_node=None):
             self.consortium.wait_for_node_to_exist_in_store(
                 primary,
                 new_node.node_id,
-                timeout=JOIN_TIMEOUT,
```
eddyashton: This was the original goal of this PR. When we're testing adding an _untrusted_ node, we don't need to wait so long - we can give the node a few seconds for a single join attempt, then we check its log for errors. Ideally this would be replaced by asking the new node directly for its status (+ response to its last join request), but this should shave some time off for now.

tests/infra/ccf.py
```diff
@@ -318,9 +316,13 @@ def create_and_add_pending_node(self, lib_name, host, args, target_node=None):
             LOG.error(f"New pending node {new_node.node_id} failed to join the network")
             errors = new_node.stop()
             if errors:
+                # Add error lines from file as custom field on this error
+                err.joining_errors = []
                 for error in errors:
                     if "An error occurred while joining the network" in error:
-                        err.message = f"TimeoutError: {error.split('|', 1)[1]}"
```
eddyashton: Perhaps unnecessary, but hopefully now clearer what's happening here. `err.message` looks like a standard field that we're overwriting, but its not, so we give it a more clearly custom name. And we save all the joining errors, rather than just the last.

tests/infra/e2e_args.py
```diff
@@ -138,9 +138,9 @@ def cli_args(add=lambda x: None, parser=None, accept_unknown=False):
     )
     parser.add_argument(
         "--join-timer",
-        help="Timer period when tyring to join an existing network (ms)",
+        help="Timer period when trying to join an existing network (ms)",
         type=int,
-        default=4000,
+        default=1000,
```
eddyashton: The default in `host/main.cpp` is 1000, so that should be good enough for Python too? A separate concern is that we're still duplicating these args and defaults in both places, but I don't see an easy way to avoid that when so many are actually used in the Python infra.

src/consensus/raft/raft.h
```diff
@@ -969,8 +971,8 @@ namespace raft
 
     void restart_election_timeout()
     {
-      // Randomise timeout_elapsed to get an election timeout that is
-      // effectively between zero and double the configured election timeout.
+      // Randomise timeout_elapsed to get a random election timeout
+      // between 0.5x and 1x the configured election timeout.
```
eddyashton: I think the old behaviour was odd and this is more intuitive. If I say the election timeout is 10s, I want an election to happen after 10s of inactivity (maybe sooner for random distribution, sure!), I shouldn't need to know this implementation detail to understand my 20s of downtime. In particular, this means that "wait for 2 full election timeouts" is `2 * election_timeout` rather than `f * election_timeout`.

src/host/node_connections.h
```diff
@@ -143,10 +145,33 @@ namespace asynchost
 
       void reconnect()
       {
-        auto s = parent.find(node);
+        // Use uv_timer_t to reconnect after a short delay
```
eddyashton: Previously, if we tried to talk to another node and got a failure, this got very spammy - we looped back into `reconnect` directly and retried immediately, indefinitely. The most common case for this failure is that the other node really is dead (eg - at the end of a test), so retrying is useless. Now we at least have a delay between retries - shouldn't cause a noticeable slowdown if the retry succeeds, and makes the logs useable even with a persistent failure.

tests/code_update.py
```diff
@@ -66,22 +75,18 @@ def run(args):
...
             LOG.debug(f"Stopping old node {node.node_id}")
             node.stop()
 
-        LOG.info("Waiting for a new primary to be elected...")
         sleep_time = (
-            args.pbft_view_change_timeout * 6 / 1000
+            args.pbft_view_change_timeout * 2 / 1000
             if args.consensus == "pbft"
-            else args.raft_election_timeout * 6 / 1000
+            else args.raft_election_timeout * 2 / 1000
```
eddyashton: Its possible this should still be `3` rather than `2`, I'm not sure. We really want to wait for a successful election + everybody has heard about it, but if we're still seeing election failures this needs to be `2 * timeout + gossip_time` rather than `1 * timeout + gossip_time`

tests/election.py
```diff
@@ -96,7 +96,7 @@ def run(args):
             LOG.debug("Stopping primary")
             primary.stop()
 
-            LOG.debug("Waiting for a new primary to be elected...")
+            LOG.debug(f"Waiting {max_election_duration} for a new primary to be elected...")
             time.sleep(max_election_duration)
```
ashamis: we should skip this sleep for pbft

eddyashton: I've tried removing this locally and it seems to introduce a new failure case (ie - beyond the view change instability that should be fixed by #1001). Going to leave the sleep in for now, and we can take another look if this is still unnecessarily slow.

tests/code_update.py
```diff
@@ -43,14 +43,23 @@ def run(args):
 
         LOG.info(f"Adding a node with unsupported code id {new_code_id}")
         try:
-            network.create_and_trust_node(args.patched_file_name, "localhost", args)
+            network.create_and_add_pending_node(
+                args.patched_file_name, "localhost", args, timeout=3
```
achamayou: So how does 3 work here? We used to wait the default amount before, does 3 work because we/OE check mrenclave first and fail fast if it matches, whereas a full quote verification can take longer?

tests/code_update.py
```diff
@@ -43,14 +43,23 @@ def run(args):
...
             assert (
                 False
             ), f"Adding a node with unsupported code id {new_code_id} should fail"
         except TimeoutError as err:
-            assert "CODE_ID_NOT_FOUND" in err.message, err.message
+            joining_errors = getattr(err, "joining_errors", None)
+            if joining_errors is not None:
+                assert "CODE_ID_NOT_FOUND" in joining_errors, joining_errors
+            else:
+                raise
```
achamayou: This seems fairly awkward, can we instead throw a JoiningError that subclasses TimeoutError and catch just that?

tests/code_update.py
```diff
@@ -43,14 +43,23 @@ def run(args):
...
+
+        # Slow quote verification means that any attempt to add a node may cause an election, so confirm primary after adding node
+        primary, others = network.find_primary()
 
-        network.consortium.add_new_code(1, primary, new_code_id)
+        network.consortium.add_new_code(0, primary, new_code_id)
```
achamayou: Why did that go down to 0? What is it?

tests/code_update.py
achamayou: gossip time is probably sig-max-ms effectively.

tests/e2e_logging.py
```diff
@@ -96,7 +96,7 @@ def test_anonymous_caller(network, args):
             assert r.result is not None
             assert msg in r.result["msg"]
     else:
-        LOG.warning("Skipping test_cert_prefix as application is not C++")
+        LOG.warning("Skipping test_anonymous_caller as application is not C++")
```
achamayou: inspect.currentframe() if we copy paste the skipping stuff often enough

tests/infra/ccf.py
achamayou: As suggested at the call site, let’s throw a specific exception here, it will be clearer.

tests/infra/e2e_args.py
achamayou: One way would be to call cchost once with —help and collect the defaults from there. Another would be to have both Python and CLI11 fetch them from a config somewhere...

src/consensus/raft/raft.h
```diff
@@ -153,7 +153,7 @@ namespace raft
       ledger(std::move(ledger_)),
       channels(channels_),
 
-      distrib(-(int)election_timeout_.count(), 0),
+      distrib(0, (int)election_timeout_.count() / 2),
```
olgavrou: does this make a difference anywhere or is it just easier to reason with like this?

tests/code_update.py
olgavrou: Another way to go here, instead of trying to guess the sleep time for each consensus etc, would be to loop a couple of times and wait for the `getPrimary` rpc to respond. That way the test time will depend on how quickly the primary was elected, and the test might also be more stable for the edge cases when the primary took a bit longer to get elected.

src/consensus/raft/raft.h
eddyashton: This halves the election times for our existing tests (when the arg is `12s`, an election happens between 6s and 12s rather than 12s and 24s). I expected to need to go and double all of these values, but the tests seem happy so I haven't.

tests/infra/e2e_args.py
jumaffre: This was originally set to 4 secs in https://github.com/microsoft/CCF/pull/680, because the primary would get overloaded with join requests. 1 second should be fine for `cchost` (users of CCF in the wild) but I think we still want to 4 seconds in our tests? Unless something has changed since #680?

tests/code_update.py
eddyashton: The node-to-node join request has always failed quickly (I believe much quicker than a full quote verification). But this code doesn't look for that error (in the sending node's log) until the timeout has elapsed. So in the old case we waited for ~60s, while the node repeatedly polled the network and was constantly told "bad code ID", then finally checked the logs. Now we leave it for only a few seconds before failing and checking the logs.

jumaffre: I agree. This will work for the wrong reason I think if the quote verification takes too long. We first verify the quote and then check MRENCLAVE, mostly because `oe_verify_report()` returns the parsed report.

tests/code_update.py
eddyashton: This is the ID of the member that we use to propose the new change. We went from 1-indexed to 0-indexed a while back but haven't updated all of these. I think the actual fix should be to remove the arg in most places (default to proposing as member 0), but its an unnecessary change regardless so I've removed it from this PR.

tests/code_update.py
achamayou: Why do we even retry if the code id doesn't check out? Should we just terminate, with an error code, and then we can check for that rather than the log?

achamayou: @jumaffre I don't think it will work for the wrong reasons because we check the log for explicitly this, but if this does verification first, then I don't see why shrinking the timeout is justified.

tests/code_update.py
```diff
@@ -42,13 +42,18 @@ def run(args):
...
-            ), f"Adding a node with unsupported code id {new_code_id} should fail"
-        except TimeoutError as err:
-            assert "CODE_ID_NOT_FOUND" in err.message, err.message
+            network.create_and_add_pending_node(
+                args.patched_file_name, "localhost", args, timeout=3
+            )
+        except infra.ccf.CodeIdNotFound as err:
+            code_not_found_exception = err
+
+        assert code_not_found_exception is not None, f"Adding a node with unsupported code id {new_code_id} should fail"
```
achamayou: Why the assert? That can never be None, it's a CodeIdNotFound.

tests/code_update.py
jumaffre: FYI, we've still got the issue of the first election not completing as the node-to-node connection between the then-follower-now-candidate nodes has not yet been established (vote requests are replaced with n2n connection establishment messages). This is because we do not want to have n^2 n2n connections open at any given times but only n (to the current leader to the followers). The consequence of this is that it takes 2 election rounds (i.e. two timeouts) for a new leader to be elected. 

It might be worth creating an issue for this, at least for documentation purposes?

tests/code_update.py
jumaffre: Good point. I think we should just `join_timer.reset();` in `initiate_join()` if an error occurred?

eddyashton: I think there's a few angles on this, written up as #1014.

If this fails because of a slow quote retrieval, it will fail with a different error type (a `TimeoutError` which will now propagate past this code and be seen elsewhere). I'd like to leave this timeout reduced - if we see slow quote retrieval failures we can either prioritise #1014 or increase this timeout. But waiting a long time _after_ failure to detect the failure, multiple times per CI run, is hopefully something we can avoid.

tests/code_update.py
eddyashton: Thanks for the explanation! I've added #1015 to track this.

tests/code_update.py
jumaffre: @eddyashton Do you mean slow quote verification on the primary side? Since https://github.com/microsoft/CCF/pull/880, quote retrieval shouldn't be an issue on the backup side.

I'm happy to leave this as it is for now and we can fix the test and prioritise #1014 as you said when necessary.

tests/infra/e2e_args.py
eddyashton: I've restored the 4s on this and added a comment explaining why.

I think parsing the defaults from `cchost` makes sense and loading from a separate config seems cumbersome (its a third source of the available argument names, in a custom format, with a necessary binding to the C++ expected names at least?). CLI11 allows custom help formatters so we could make one that's more reliably/easily parseable. We could also go further and try to write a formatter that produced JSON rather than string (ie - machine-parseable help formatting), but that's a lot of work for a minor inconvenience.

PR #1018, opened 2020-03-31
Fix forward caller unauthenticated handler

### LINKED ISSUE #987: Forwarding caller certificate for unauthenticated handlers
Since https://github.com/microsoft/CCF/pull/962, it is now possible to install handlers that do not require a valid caller (i.e. no need to have the client registered in the corresponding caller table). We also currently only forward a caller certificate if it's not possible to look it up on the forwardee node (i.e. primary) from the always-forwarded caller ID. 

These two features clash. The right behaviour should be to forward the caller certificate if the executing frontend has no caller table _or_ if the executing handler is unauthenticated (i.e. `set_require_client_identity(false)` is set).

jumaffre:
Resolves #987

When forwarding a request to the primary, the forwarder node no longer forwards the caller certificate only based on whether the executing frontend has a `certs` table (i.e. the forwardee node can lookup the caller certificate from the forwarded caller id) but **also if the executing handler requires a known caller identity.**

Also:
- Made `process_command` private in the `RpcFrontend` class (hence the awkward diff).
- `liblibbyz` -> `libbyz`.


src/node/rpc/frontend.h
```diff
@@ -73,12 +73,47 @@ namespace ccf
       handlers.set_history(history);
     }
 
+    std::vector<uint8_t> get_cert_to_forward(
```
jumaffre: This is the main change in this PR. We no longer override this function in each frontend but base it on `handlers` and the `handler`'s `require_client_identity` flag.

src/node/rpc/frontend.h
```diff
@@ -613,6 +431,194 @@ namespace ccf
...
+    /** Process a serialised command with the associated RPC context
+     *
+     * If an RPC that requires writing to the kv store is processed on a
+     * backup, the serialised RPC is forwarded to the current network primary.
+     *
+     * @param ctx Context for this RPC
+     * @returns nullopt if the result is pending (to be forwarded, or still
+     * to-be-executed by consensus), else the response (may contain error)
+     */
+    std::optional<std::vector<uint8_t>> process(
```
jumaffre: Note: in the Raft case, `process()` is now pretty much a wrapper for `process_command()`. 
As we move more and more logic down to `process_command()` depending on which handler is neing executed, `process_command()` might be renamed to `process()` and the balance of the universe will be restored.

src/node/rpc/test/frontend_test.cpp
```diff
@@ -1005,6 +995,60 @@ TEST_CASE("Forwarding" * doctest::test_suite("forwarding"))
...
+      auto [fwd_ctx, node_id] =
+        backup_forwarder
+          ->recv_forwarded_command(forwarded_msg.data(), forwarded_msg.size())
+          .value();
+
+      auto response =
+        parse_response(user_frontend_primary.process_forwarded(fwd_ctx));
+      CHECK(response.status == HTTP_STATUS_OK);
+
+      CHECK(user_frontend_primary.last_caller_cert == user_caller_der);
```
jumaffre: Before this PR, `user_frontend_primary.last_caller_cert.size() == 0`.

src/node/rpc/test/frontend_test.cpp
```diff
@@ -1005,6 +995,60 @@ TEST_CASE("Forwarding" * doctest::test_suite("forwarding"))
...
+      auto [fwd_ctx, node_id] =
+        backup_forwarder
+          ->recv_forwarded_command(forwarded_msg.data(), forwarded_msg.size())
+          .value();
+
+      auto response =
+        parse_response(user_frontend_primary.process_forwarded(fwd_ctx));
+      CHECK(response.status == HTTP_STATUS_OK);
+
+      CHECK(user_frontend_primary.last_caller_cert == invalid_caller_der);
```
jumaffre: Before this PR, `user_frontend_primary.last_caller_cert.size() == 0`.

PR #1020, opened 2020-04-01
Late joiner test is not stable yet

achamayou:
#1001 doesn't seem to be stable yet, it's failed twice on master.


olgavrou:
You don't need to disable the whole test, just pass the additional python argument `--skip-suspension` which will disable view changes

PR #1023, opened 2020-04-02
Maintain ordering of enum CLI options

eddyashton:
`std::map`s are ordered by key, so if we pass these to CLI11 our enum options get lexicographically sorted. I think its more readable if we maintain the enum-order.

Before:
`-l,--host-log-level ENUM:value in {debug->1,fail->3,fatal->4,info->2,trace->0} OR {1,3,4,2,0}=2`

After:
`-l,--host-log-level ENUM:value in {trace->0,debug->1,info->2,fail->3,fatal->4} OR {0,1,2,3,4}=2`

### DIFF
--- a/src/host/main.cpp
+++ b/src/host/main.cpp
@@ -52,7 +52,7 @@ int main(int argc, char** argv)
     VIRTUAL
   };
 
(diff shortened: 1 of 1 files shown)


achamayou:
Nice!

PR #1024, opened 2020-04-02
Some PBFT cleanup

ashamis:
- remove unused stats info
- increase batch size
- when we have less than 10 pending tx to run in a batch always write to the Merkle tree instead of waiting.

### DIFF
--- a/src/consensus/pbft/pbft_config.h
+++ b/src/consensus/pbft/pbft_config.h
@@ -86,8 +86,9 @@ namespace pbft
 
       if (
         info.pending_cmd_callbacks %
-          PbftConfigCcf::max_update_merkle_tree_interval ==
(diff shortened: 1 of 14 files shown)


src/consensus/pbft/pbft_config.h
```diff
@@ -86,8 +87,10 @@ namespace pbft
 
       if (
         info.pending_cmd_callbacks %
-          PbftConfigCcf::max_update_merkle_tree_interval ==
-        0)
+            PbftConfigCcf::max_update_merkle_tree_interval ==
+          0 ||
+        info.pending_cmd_callbacks <
```
jumaffre, line 92: What are the effects of this change?

ashamis: functionally nothing, but performance wise ~5% for smallbank

PR #1025, opened 2020-04-02
Add version check to keep CMake version up-to-date

eddyashton:
The project version in CMake (used for default install dir) has lagged behind our tags.

This PR adds a check in the CI that this CMake version is at least as high as our git tags, so we won't forget to increment this before releases in future (the failure will be on a git tag CI build if we haven't pre-incremented, but hopefully this reminds us to pre-increment).


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ include(${CCF_DIR}/cmake/preproject.cmake)
 
 project(
   ccf
-  VERSION 0.8.2
+  VERSION 0.8.2 # CCF_VERSION < This is a search marker for checks.yml
```
achamayou, line 10: Should be 0.9.1?

jumaffre, line 10: for `check-versions.sh`?

check-versions.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
jumaffre, line 1: Would `check-cmake-with-tag.sh` be slightly more explicit? Also, we are starting to have quite a few scripts at the top level directory of our repo. I wonder if it's worth creating a new `scripts/` directory so that people don't have to scroll too much to get to our `README.md`?

eddyashton:
[Here's](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=6942&view=logs&j=435ca956-9126-505f-566f-fff31072e2ba&t=ffc59b44-7438-5626-aa7a-09fa4141874b&l=15) the error we'll see when the version hasn't been updated:

```
Git tag (0.9.1) is later than CMake version (0.8.2) - please update CMake version!
```

CMakeLists.txt
eddyashton: Yes, was just getting a failing version first to confirm this works.

check-versions.sh
eddyashton: It may be, but I'd like to leave this for another PR with some more thought (these are probably `scripts/checks`, there are some in `tests/` that could be moved out, does `start_test_network.sh` get to stay at the top-level for visibility?)

CMakeLists.txt
eddyashton: Updated comment, removed reference to file. This is just a "don't delete this line" comment, and if we want to find where its used we search for "CCF_VERSION".

PR #1027, opened 2020-04-02
The rest of the daily build

### LINKED ISSUE #832: Setup daily build

achamayou:
fix #832 

1. Run testssl, and archive HTML report
2. shuffle e2e test suite


achamayou:
@jumaffre no, there’s already a morning run of the CI every day that does this, I did not see any point duplicating that.

What we may do is more/more detailed perf testing in the Daily than we do in CI, but that will need support in CI metrics to not be picked up by the regression view. I don’t think that’s going to be a priority any time soon.

Does the sync label actually sync or did you press the button?

jumaffre:
> @jumaffre no, there’s already a morning run of the CI every day that does this, I did not see any point duplicating that.
> 
> What we may do is more/more detailed perf testing in the Daily than we do in CI, but that will need support in CI metrics to not be picked up by the regression view. I don’t think that’s going to be a priority any time soon.
> 
> Does the sync label actually sync or did you press the button?

I did press the button when I merged my branch. I removed `automerge` yesterday as it doesn't work with forks.

tests/e2e_suite.py
```diff
@@ -22,6 +24,11 @@ class TestStatus(Enum):
 
 def run(args):
 
+    if os.getenv("SHUFFLE_SUITE"):
```
jumaffre, line 27: Could we add an argument to the suite to accept a given seed (so that it's easy to re-run the same suite in case of errors)?

achamayou: Sure, I'll make it an env var also, because the sheer amount of stuff to change in pipelines yaml, cmake and then python to expose generic args is too much for such a niche switch

Issue #806, opened 2020-02-05
Restrict members access to critical tables


jumaffre:
Follow up from: https://github.com/microsoft/CCF/pull/797#discussion_r375140010 

Flexible governance allows members to perform a virtually-unlimited set of operations on the white-listed tables in the KV. And even if we provide a sensible default white-list, the constitution may allow this white-list to be updated (e.g. via unanimity of members). 

This causes issues in terms of:
- Security. For example, it might be possible that members write a new node to the KV, bypassing the attestation and code version verification performed during the join protocol.
- Usability. Members may decide to add a new user but only update the `ccf.users` table. However, the `ccf.user_certs` and `ccf.values` tables should also be updated. (This particular point may be addressed by providing a richer KV API.)

This should be addressed by (at least) documenting these restrictions.

achamayou:
I think we should look at this from the perspective of preserving invariants, beyond Lua's type system abilities and beyond the KV's interface

jumaffre:
Now that we issue new shares on consortium updates, etc., we should definitely make sure that members cannot `raw_puts` to the `ccf.members` table directly. Some unit and e2e tests still rely on this so I've left this table in the whitelist for now.

achamayou:
@jumaffre yes, this is otherwise a serious issue!

achamayou:
With governance proposals being restricted to pre-defined constitution functions, and a safe set of default actions, I don't think this is a problem anymore, since:

1. There is no raw_puts in our default actions
2. Multi-table actions have been simplified (thanks to new ids), and the default actions are a safe starting point for customisation if desired.
3. Unsafe actions would need to be voted in as constitution amendments before they can be used. This is equivalent in governance complexity to a code update, so I think we've effectively made as much of an improvement as possible.

PR #1032, opened 2020-04-03
Key shares and consortium update

jumaffre:
Resolves items 10 and 14 in #51 

Apologies for the large PR: in hindsight, some of the following items could have been committed independently. 

This PR:
- Introduces a new configuration table, which for now only contains the `recovery_threshold` value. This value indicates how many recovery shares are required to re-store the ledger secrets on recovery. The `recovery_threshold` can be set when the first node starts (`--recovery-threshold` parameter to `cchost` - defaults to the number of original members) or updated via the new `set_recovery_threshold` governance command. 
- Updates the recovery key shares when consortium membership changes happen: a) when a new member is activated (`ack` RPC), they are allocated a recovery share (+ all the shares are reshuffled) and b) when a active member is retired, all the shares are reshuffled. 
- You will see that they are a number of checks in places to make sure that we never end up in a non-recoverable situation, i.e. `recovery_threshold` > number of active members. 
- I had to refactor a bunch of Python infra code to make sure that we can easily test membership changes from our e2e infra. In particular, the `Consortium` class was split (see new `Member` class) and a new `Proposal` class was created. `Member`s can `propose()` (returns a `Proposal`), `vote()`, `ack()`, `getEncryptedRecoveryShare()`, `submitRecoveryShare()`, etc. while the `Consortium` orchestrates members, keeping track of who is active and can `vote_using_majority()`.
- I have added new unit tests at the top of `memberclient.py` to test for membership and recovery threshold updates.

Still to do:
- Re-keying when a member is retired as it turns out that re-keying and re-sharing at the same time is awkward.
- Chaining the e2e test blocks created in this PR and adding them to our test suite (i.e. actually test that we can recover a network when the threshold has been updated). 
- Remove sealing altogether.


src/node/rpc/member_frontend.h
```diff
@@ -138,6 +138,55 @@ namespace ccf
            auto new_member_id =
              g.add_member(parsed.cert, parsed.keyshare, MemberStatus::ACCEPTED);
 
+           return true;
+         }},
+        // retire an existing member
+        {"retire_member",
```
jumaffre, line 144: Added comment here: https://github.com/microsoft/CCF/issues/806#issuecomment-608516853

src/host/main.cpp
```diff
@@ -358,26 +368,55 @@ int main(int argc, char** argv)
     public_rpc_address = rpc_address;
   }
 
-  if (domain.empty() && !ds::is_valid_ip(rpc_address.hostname.c_str()))
+  uint32_t oe_flags = 0;
+  try
```
jumaffre, line 372: Making validation errors abort in a nicer way here.

src/runtime_config/default_whitelists.h
```diff
@@ -25,17 +25,19 @@ namespace ccf
...
+      Tables::CONFIGURATION}},
 
     {MEMBER_CAN_PROPOSE,
-     {Tables::MEMBERS,
-      Tables::MEMBER_CERTS,
-      Tables::USERS,
+     {Tables::USERS,
       Tables::USER_CERTS,
+      Tables::MEMBERS,
+      Tables::MEMBER_CERTS,
```
jumaffre, line 35: These will go soon: https://github.com/microsoft/CCF/issues/806#issuecomment-608516853

jumaffre:
Looks like I shouldn't have deleted one of the `pbft_open` parameter somewhere in the e2e infra when opening a network. This is the cause of the CI failures.

tests/infra/consortium.py
```diff
@@ -213,7 +211,9 @@ def open_network(self, remote_node, pbft_open=False):
         return Calls:call("open_network")
         """
         proposal = self.get_any_active_member().propose(remote_node, script)
-        self.vote_using_majority(remote_node, proposal)
+        self.vote_using_majority(
+            remote_node, proposal, wait_for_global_commit=(not pbft_open)
```
achamayou, line 215: That doesn't seem like it should change because of the re-keying? I am missing something?

src/host/main.cpp
```diff
@@ -358,26 +368,55 @@ int main(int argc, char** argv)
     public_rpc_address = rpc_address;
   }
 
-  if (domain.empty() && !ds::is_valid_ip(rpc_address.hostname.c_str()))
+  uint32_t oe_flags = 0;
```
achamayou, line 371: Is there an OE type for this?

tests/infra/consortium.py
jumaffre: I've reverted the change to the `pbft_open` flag. I was (wrongly) suspecting that we didn't need this flag anymore but we still do.

src/node/genesis_gen.h
```diff
@@ -349,5 +379,23 @@ namespace ccf
...
+    }
+
+    size_t get_recovery_threshold()
+    {
+      auto config_view = tx.get_view(tables.config);
+      auto config = config_view->get(0);
+      if (!config.has_value())
+      {
+        throw std::logic_error(
+          "Failed to set recovery threshold: No active configuration found");
```
achamayou, line 396: Failed to get?

src/node/rpc/member_frontend.h
```diff
@@ -719,8 +797,16 @@ namespace ccf
         if (member->status == MemberStatus::ACCEPTED)
         {
           member->status = MemberStatus::ACTIVE;
+          members->put(args.caller_id, *member);
+
+          // New active members are allocated a new recovery share
+          if (!node.split_ledger_secrets(args.tx))
```
achamayou, line 803: I don't think we ought to transition the member to active if we cannot give them a share. They effectively aren't, for the purposes of the constitution.

tests/infra/consortium.py
achamayou: I was missing something, ignore this.

tests/infra/consortium.py
```diff
@@ -309,53 +291,57 @@ def store_current_network_encryption_key(self):
...
-                ctx = infra.crypto.CryptoBoxCtx(
-                    os.path.join(self.common_dir, f"member{m}_kshare_priv.pem"),
-                    os.path.join(self.common_dir, f"network_enc_pubk_orig.pem"),
-                )
-
-                nonce_bytes = bytes(r.result["nonce"])
-                encrypted_share_bytes = bytes(r.result["encrypted_share"])
-                decrypted_share = ctx.decrypt(encrypted_share_bytes, nonce_bytes,)
+    def recover_with_shares(self, remote_node):
+        # TODO: Do something better here!
```
achamayou, line 295: A better comment would describe what's wrong with the current code :)

src/node/rpc/member_frontend.h
jumaffre: Even if the member is transitioned to `ACTIVE` before we issue the share, the `Tx` will be aborted aborted altogether if `split_ledger_secrets()` fails and the member will not actually be recorded as `ACTIVE`?

tests/infra/consortium.py
jumaffre: Well spotted! Something I wanted to fix before PR-ing this but didn't. I will fix it now.

src/host/main.cpp
jumaffre: It doesn't look like it: https://github.com/openenclave/openenclave/search?p=1&q=OE_ENCLAVE_FLAG_DEBUG&unscoped_q=OE_ENCLAVE_FLAG_DEBUG

src/node/rpc/member_frontend.h
achamayou: Yes, especially so I would think. The member would be active on paper and potentially in recovery quotas but unable to provide a share. That does not sound good.

Is there a scenario where we think there could be a transient failure to share that can be fixed with a re-share? Even if there is, transitioning the member without them having a share seems risky...

jumaffre: As discussed, this is fine as the handler is transactional.

PR #1034, opened 2020-04-05
encrypt PBFT request and response messages

ashamis:
- Encryption/Decryption should be done across multiple threads
- If USE_NULL_ENCRYPTOR=ON do not encrypt the messages

### DIFF
--- a/src/consensus/pbft/pbft.h
+++ b/src/consensus/pbft/pbft.h
@@ -614,7 +614,9 @@ namespace pbft
     struct RecvAuthenticatedMsg
     {
       RecvAuthenticatedMsg(
-        bool should_decrypt_, OArray&& d_, Pbft<LedgerProxy, ChannelProxy>* self_) :
+        bool should_decrypt_,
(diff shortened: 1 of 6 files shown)


src/consensus/pbft/pbft.h
```diff
@@ -594,21 +632,40 @@ namespace pbft
     static void recv_authenticated_msg_cb(
       std::unique_ptr<enclave::Tmsg<RecvAuthenticatedMsg>> msg)
     {
-      try
+      if (msg->data.should_decrypt)
```
olgavrou, line 635: why are we using `msg->data.should_decrypt` to determine whether a message is encrypted or not, instead of utilizing the `encrypted_pbft_message` message type?

ashamis: I am, that is how this property I set. I might be misunderstanding your comment.

olgavrou: ah yes I didn't notice the threading callback in the PR diff

PR #1035, opened 2020-04-06
Keep PR up-to-date and merge automatically

jumaffre:
It turns out, I think GitHub Actions are the solution to what we've been discussing for a while now. This PR introduces our first GitHub Action that will take care of (based on https://github.com/pascalgn/automerge-action): 
- Auto-merging a PR if the `automerge` label has been set on a PR (and if the other CI/reviewers requirements are met). Note that the `wip` label would explicitly disable this and no label has no effect.
- Auto-updating all PRs if the `sync` label has been set on a PR.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -30,5 +30,5 @@ jobs:
           MERGE_METHOD: "squash"
           MERGE_REMOVE_LABELS: "automerge,wip"
           MERGE_FORKS: true
-          UPDATE_LABELS: "" # All branches are automatically kept up-to-date
+          UPDATE_LABELS: "sync,!wip"
           UPDATE_METHOD: "merge"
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: automerge
+        uses: "pascalgn/automerge-action@135f0bdb927d9807b5446f7ca9ecc2c51de03c4a"
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          # https://github.com/pascalgn/automerge-action#configuration
+          MERGE_LABELS: "automerge,!wip"
+          MERGE_METHOD: "squash"
+          MERGE_REMOVE_LABELS: "automerge,wip"
+          MERGE_FORKS: true
+          UPDATE_LABELS: "" # All branches are automatically kept up-to-date
```
eddyashton, line 33: Can we add labels to enable/disable auto-updating as well? I have my remotes setup so that I can manually merge from origin/master, but its extra hoops if Github has modified my fork. And I think we should disable it for `wip` as well.

jumaffre: Good point. I've added a new `sync` label for this (and `wip` prevents updating). I think it's better to keep the `automerge` and auto-update (`sync`) flags different for now.

Issue #1006, opened 2020-03-27
Performance tests don't wait for global commit during setup


eddyashton:
Our performance tests sometimes create new connections (ie - for setup). The results of this connection may not be visible to other connections (especially in the forwarded test!). We should either use a single connection, or ensure that the things we expect to read are globally committed.

eddyashton:
Fixed by #1039.

PR #1039, opened 2020-04-06
Perf test session consistency - wait for global commit

eddyashton:
Fix for #1006.

The key fix is that we use a single session in the perf clients, for both creation transactions and the timed workload, so that even with forwarding we will see the results of any initialisation as expected. Additionally, we wait for these initial transactions to be globally committed before proceeding, so we should get a clearer error if they turn out to be uncommittable. And I've changed _how_ we wait for global commit (wait for target `term.index` to be globally committed, rather than waiting for global commit to stop changing), which hopefully gives slightly more consistent results, and could be a step towards running perf clients in parallel.

Large diff as this includes a refresh of `perf_client`, using the logging macros for consistency, and using a separate `Options` struct for option parsing so we can later construct the `timing` object.


samples/apps/smallbank/clients/small_bank_client.cpp
```diff
@@ -5,7 +5,18 @@
...
 
-using Base = client::PerfBase;
+struct SmallBankClientOptions : public client::PerfOptions
+{
+  size_t total_accounts = 10;
+
+  SmallBankClientOptions(CLI::App& app) :
+    client::PerfOptions("Small_Bank_ClientCpp", app)
+  {
+    app.add_option("--accounts", total_accounts);
```
achamayou, line 15:
```suggestion
    app.add_option("--accounts", total_accounts)->capture_default_str();
```

samples/apps/smallbank/clients/small_bank_client.cpp
```diff
@@ -146,23 +158,23 @@ class SmallBankClient : public Base
...
     {
-      cout << "Creating " << total_accounts << " accounts..." << endl;
+      LOG_INFO_FMT("Creating {} accounts", options.total_accounts);
     }
   }
 
   void post_creation_hook() override
   {
-    if (verbosity >= 2)
+    if (options.verbosity >= 2)
```
achamayou, line 169: Convert to LOG_DEBUG or TRACE perhaps?

samples/perf_client/perf_client.h
```diff
@@ -28,27 +28,133 @@ namespace client
...
+    PerfOptions(const std::string& default_label, CLI::App& app) :
+      label(default_label)
+    {
+      // Enable config from file
+      app.set_config("--config");
+
+      app.add_option(
+        "--label",
+        label,
+        "Identifier for this client, written to " + std::string(perf_summary));
```
achamayou, line 84: I think we want capture_default_str() everywhere we have a default value?

achamayou: ie. everywhere that's not required?

samples/perf_client/perf_client.h
```diff
@@ -79,43 +185,48 @@ namespace client
...
+          throw std::logic_error(fmt::format(
+            "Term went backwards (expected {}, saw {})!",
+            last_response_commit.term,
+            commits.term));
+        }
+        else if (
+          commits.term > last_response_commit.term &&
+          commits.local <= last_response_commit.index)
+        {
+          throw std::logic_error(fmt::format(
```
achamayou, line 214: I guess in the future, if we did want to test across elections as Peng Peng suggested, we would have replay logic here. For now this is an improvement over the current situation already.

eddyashton:
> Can you please hold off on commit this. This break the small bank perf tests i am running for the paper

@ashamis This should be fixed by the [previous commit](https://github.com/microsoft/CCF/pull/1039/commits/183c783dd28f8c0e18f69a5782b7de3552068fbe), please let me know if you're still having problems.

samples/perf_client/timing.h
```diff
@@ -209,35 +212,88 @@ namespace timing
...
-          std::cout << timestamp() << "Waiting for " << target << ", at "
-                    << global << std::endl;
-          last_printed = current;
+          record_send(get_commit, response.id, false);
+        }
+
+        const auto body = net_client->unpack_body(response);
+        if (response.status != HTTP_STATUS_OK)
+        {
+          throw runtime_error(
```
jumaffre, line 242: Minor: maybe report the HTTP status here?

PR #1041, opened 2020-04-06
Remove member tables from whitelist

jumaffre:
Relates to https://github.com/microsoft/CCF/issues/806 

Now that membership changes (addition or deletion of nodes) trigger issuing of new shares, it should not be possible for existing members to changes the members tables from a `raw_puts()`.

We were still relying on this behaviour from one unit test and I've updated the proposal there accordingly.


achamayou:
Thanks for fixing that!

PR #1043, opened 2020-04-07
Forwarding performance test with multiple clients

jumaffre:
Previously, we only spawned one local client for all performance tests. This is not ideal for the forwarding test with client signatures which is supposed to show the gain in verifying signatures on backup nodes. 

This PR adds a new boolean `--one-client-per-backup` option to our `perfclient.py` which will spawn one client per backup.


tests/infra/remote.py
```diff
@@ -260,7 +262,8 @@ def start(self):
         cmd = self._cmd()
         LOG.info("[{}] {}".format(self.hostname, cmd))
         self.client.exec_command(cmd, get_pty=True)
-        self.pid()
+        if self.is_ccf_node:
```
jumaffre, line 265: Our remotes used to work fine for both nodes and client until the recent fact that we rely on the node outputing its PID in a file.

achamayou: FSVO fine, we were grepping ps to decide what to suspend :)

tests/infra/remote.py
```diff
@@ -278,13 +281,19 @@ def pid(self):
         return self._pid
 
     def suspend(self):
+        if not self.is_ccf_node:
```
achamayou, line 284: I'm not convinced it's worth adding all this extra stuff and the flag just in testing infra on the off chance someone tries to suspend a client.

Can we just get the call to pid() out of the ctor if the command doesn't start with "./cchost"?

tests/infra/runner.py
```diff
@@ -44,6 +44,8 @@ def filter_nodes(primary, backups, filter_type):
     if filter_type == "primary":
         return [primary]
     elif filter_type == "backups":
+        if len(backups) == 0:
```
achamayou, line 47:
```suggestion
        if not backups:
```

https://www.python.org/dev/peps/pep-0008/#programming-recommendations

tests/infra/runner.py
```diff
@@ -93,20 +95,28 @@ def run(get_command, args):
...
         command_args = get_command_args(args, get_command)
 
-        nodes = filter_nodes(primary, backups, args.send_tx_to)
+        nodes_to_send_to = filter_nodes(primary, backups, args.send_tx_to)
         clients = []
         client_hosts = []
-        if args.num_localhost_clients:
-            client_hosts = ["localhost"] * int(args.num_localhost_clients)
+        if args.one_client_per_backup:
+            if len(backups) == 0:
```
achamayou, line 102:
```suggestion
            if not backups:
```

tests/perfclient.py
```diff
@@ -19,12 +19,18 @@ def cli_args(add=lambda x: None, accept_unknown=False):
...
+    client_args_group = parser.add_mutually_exclusive_group()
+    client_args_group.add_argument(
         "-cn",
         "--client-nodes",
         help="List of hostnames for spawning client(s). If empty, one client is spawned locally",
         action="append",
     )
+    client_args_group.add_argument(
+        "--one-client-per-backup",
+        help="If set, allocates one client per backup",
```
eddyashton, line 31: We already have several ways of specifying clients, and its not clear how they all interact. If we're disallowing clients on remote machines, then `--client-nodes` can be removed/simplified (we want to specify the node a client talks _to_, not the node it runs _on_, but we want to specify as "backup 0" rather than by IP?)

achamayou: I agree about simplifying, but allowing remote clients seems like a good feature to preserve (assuming it does work now, which we don't know because the CI certainly doesn't do that)?

tests/infra/runner.py
```diff
@@ -93,20 +95,27 @@ def run(get_command, args):
 
         command_args = get_command_args(args, get_command)
 
-        nodes = filter_nodes(primary, backups, args.send_tx_to)
+        nodes_to_send_to = filter_nodes(primary, backups, args.send_tx_to)
         clients = []
         client_hosts = []
-        if args.num_localhost_clients:
-            client_hosts = ["localhost"] * int(args.num_localhost_clients)
+        if args.one_client_per_backup:
```
eddyashton, line 101: I think these options should all be additive, since we don't document their dependencies. So `--one-client-per-backup --num-localhost-clients 3 --client-nodes 1.2.3.4` should spawn 3 localhost + 1 on `1.2.3.4` + 1-per-backup.

tests/infra/remote.py
```diff
@@ -153,6 +153,7 @@ def __init__(
         self.out = os.path.join(self.root, "out")
         self.err = os.path.join(self.root, "err")
         self.suspension_proc = None
+        self.pid_file = "cchost.pid" if "./cchost" in self.cmd else "perf_client.pid"
```
eddyashton, line 156: Can we make this more general? Something like `self.pid_file = f"{os.path.basename(self.cmd[0])}.pid}"` here, matched by `files::dump(fmt::format("{}.pid", argv[0]), ::getpid())` in both instances in C++? We can keep the `--pid-file` argument for overrides, by constructing its default from `argv[0]`.

tests/infra/runner.py
eddyashton: Ah I've just seen that some of these are currently mutually exclusive, so the code almost matches the arg spec, but we `--num-localhost-clients` is also exclusive with `--one-client-per-backup` in this code. Don't think they need to be or should be exclusive.

tests/infra/remote.py
jumaffre: Good point. I've implemented this now.

tests/infra/runner.py
jumaffre: The two options (`--one-client-per-backup` and `--num-localhost-clients`) are now non mutually exclusive.

tests/perfclient.py
jumaffre: I was also thinking of deleting `--client-nodes` too but as @achamayou mentioned, it may be useful to keep this for more interesting perf tests. 

I had to fix a couple of things but remote clients work now.

jumaffre:
The `SB_sig_SGX_CFT` metrics is missing on the latest run of `master`. Indeed, the sig perf tests _silently_ failed most of the time before the fix in this PR. This will be fixed on the next PR.

PR #1050, opened 2020-04-08
smallbank with flatbuffers

olgavrou:


samples/apps/smallbank/app/accounts.fbs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
ashamis, line 1: Can all the fbs tx types files be combined?

samples/apps/smallbank/app/flatbuffer_wrapper.h
```diff
@@ -0,0 +1,178 @@
...
+
+public:
+  std::unique_ptr<flatbuffers::DetachedBuffer> get_detached_buffer()
+  {
+    return std::make_unique<flatbuffers::DetachedBuffer>(builder.Release());
+  }
+
+  CBuffer get_buffer()
+  {
+    return {builder.GetBufferPointer(), builder.GetSize()};
```
ashamis, line 25: Do we need to create a new vector here? can pointers and size be used?

olgavrou: It's not a vector, it is a CBuffer, i.e. pointers and vectors _are_ used

samples/apps/smallbank/app/accounts.fbs
olgavrou: Yes I believe they can be combined, it is a matter of how we want to create our flatbuffer files per app. The combination would result in one file with all the potential fields, only some being populated per RPC, and one serializer/deserializer. I think it is a bit cleaner to have them separate, but I don't have a very strong opinion :)

ashamis:
Maybe what I am saying is wrong but the speedup in the sign bft small bank test is concerning. Since the sig test now has the same throughput as the non signing variant, is signing broken?

olgavrou:
@Alex signing was broken but it looks like it is fixed now, no?

PR #1054, opened 2020-04-09
Allow handlers to decide whether a transaction gets committed

### LINKED ISSUE #1037: Commitability should not be determined by HTTP response status
Currently we decide whether to call `kv::commit()` (ie - whether to try and keep a transaction's KV changes) by looking at the HTTP response status. We keep `200 OK`s, and drop everything else (returning the response at this point, without injecting commit headers).

This has 2 big failings:
- Unnecessarily restrictive on accepted statuses. All `2xx` statuses should be considered successes (and our empty responses should become `204 No Content`?)
- Imposes restrictions on supported apps. If an app wants to log data from requests that return errors, it should be able to

`RpcContext::response_is_error` should be extended to handle these cases.

eddyashton:
Resolves #1037.

Previously we applied any transaction that returned `200 OK`, and reverted all others. Now we _by default_ apply any with successful `2xx` statuses, but also allow the handler to say "apply" or "don't apply" for each transaction.

I've added some brief docs on this. I think it belongs in the "how to interact with the KV" docs rather than the "how to write an app" docs. But maybe we should split the latter, to separate "here's the simple logging app and the minimal handler you can write" from "here's examples of all the things you can set on/in/for a transaction handler".

Also, I'm saying "apply" even though the function we're calling is `Tx::commit`, to distinguish "changes that may be visible to a future transaction" from "changes that are globally committed and will never be lost".

### DIFF
--- a/sphinx/source/developers/kv/kv_how_to.rst
+++ b/sphinx/source/developers/kv/kv_how_to.rst
@@ -180,3 +180,21 @@ A ``View`` offers a :cpp:class:`kv::Map::TxView::foreach` member function to ite
 
         }
     });
+
+Applying and reverting writes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Changes to the ``Store`` are made by atomic transactions. For a given :cpp:class:`kv::Tx`, either all of the writes it makes are applied, or none are. Only applied writes are replicated and may be globally committed. Transactions may be abandoned without applying their writes - their changes will never be seen by other transactions.
+
+By default CCF decides which transactions are successful, so should be applied to the persistent store, by looking at the status code contained in the response: all transactions producing ``2xx`` status codes will be applied, while any other status code will be treated as an error and will `not` be applied to the persistent store. If this behaviour is not desired, for instance if an app which wants to log incoming requests even when they produce an error, then it can be dynamically overridden by explicitly telling CCF whether it should apply a given transaction:
+
+.. code-block:: cpp
+
+    args.rpc_ctx->set_response_status(HTTP_STATUS_FORBIDDEN);
+    auto forbidden_requests_view = tx.get_view(forbidden_requests);
+
+    // Log details of forbidden request
+    forbidden_requests_view->put(...);
+
+     // Apply this, even though it has an error response
+    args.rpc_ctx->set_apply_writes(true);
--- a/sphinx/source/developers/kv/kv_how_to.rst
+++ b/sphinx/source/developers/kv/kv_how_to.rst
@@ -186,7 +186,7 @@ Applying and reverting writes
 
 Changes to the ``Store`` are made by atomic transactions. For a given :cpp:class:`kv::Tx`, either all of its writes are applied, or none are. Only applied writes are replicated and may be globally committed. Transactions may be abandoned without applying their writes - their changes will never be seen by other transactions.
 
-By default CCF decides which transactions are successful (so should be applied to the persistent store) by looking at the status code contained in the response: all transactions producing ``2xx`` status codes will be applied, while any other status code will be treated as an error and will `not` be applied to the persistent store. If this behaviour is not desired, for instance if an app which wants to log incoming requests even when they produce an error, then it can be dynamically overridden by explicitly telling CCF whether it should apply a given transaction:
+By default CCF decides which transactions are successful (so should be applied to the persistent store) by looking at the status code contained in the response: all transactions producing ``2xx`` status codes will be applied, while any other status code will be treated as an error and will `not` be applied to the persistent store. If this behaviour is not desired, for instance when an app wants to log incoming requests even though they produce an error, then it can be dynamically overridden by explicitly telling CCF whether it should apply a given transaction:
 
 .. code-block:: cpp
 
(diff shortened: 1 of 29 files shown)


src/http/http_rpc_context.h
```diff
@@ -261,9 +263,19 @@ namespace http
       response_headers[std::string(name)] = value;
     }
 
-    virtual bool response_is_error() const override
+    virtual void set_apply_writes(bool apply) override
```
jumaffre, line 266: Do you reckon we could have a flag on the handler as well? So that app developers can enforce this statically too (e.g. "I want my handler always apply the transaction, even on failure").

eddyashton: We could do this, but I think duplicating ways to set these things is just adding a complication for now. Arguably several of the options we set per-handler (or in this case, per-request) should actually have adjustable per-frontend defaults. But I think we should build a generic solution for all of them rather than adding multiple ways to set _some_.

eddyashton: And in this case, its definitely something that could be done within the app itself (add a decorator like our `json_adapter` that always calls `set_apply_writes(true)`), so maybe not something we need to add on our side.

sphinx/source/developers/kv/kv_how_to.rst
```diff
@@ -180,3 +180,21 @@ A ``View`` offers a :cpp:class:`kv::Map::TxView::foreach` member function to ite
 
         }
     });
+
+Applying and reverting writes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Changes to the ``Store`` are made by atomic transactions. For a given :cpp:class:`kv::Tx`, either all of its writes are applied, or none are. Only applied writes are replicated and may be globally committed. Transactions may be abandoned without applying their writes - their changes will never be seen by other transactions.
+
+By default CCF decides which transactions are successful (so should be applied to the persistent store) by looking at the status code contained in the response: all transactions producing ``2xx`` status codes will be applied, while any other status code will be treated as an error and will `not` be applied to the persistent store. If this behaviour is not desired, for instance if an app which wants to log incoming requests even when they produce an error, then it can be dynamically overridden by explicitly telling CCF whether it should apply a given transaction:
```
olgavrou, line 189: if an app which wants to -> if an app wants to (remove which)

src/http/http_rpc_context.h
```diff
@@ -261,9 +263,19 @@ namespace http
...
+
+    virtual bool get_apply_writes() const override
     {
-      return response_status != HTTP_STATUS_OK;
+      if (explicit_apply_writes.has_value())
+      {
+        return explicit_apply_writes.value();
+      }
+
+      return (response_status / 100) == 2;
```
olgavrou, line 278: what does `response_status / 100 == 2` do? maybe add a comment?

src/node/rpc/frontend.h
```diff
@@ -347,7 +347,7 @@ namespace ccf
         {
           func(args);
 
-          if (ctx->response_is_error())
+          if (!ctx->get_apply_writes())
```
olgavrou, line 350: since  `get_apply_writes` returns a boolean maybe rename to `writes_applied` or `check_writes_applied` something similar

src/http/http_rpc_context.h
eddyashton: This checks whether the response_status is in the 200 range - have rewritten to be clearer and added a comment.

src/node/rpc/frontend.h
eddyashton: This is asking whether we _should_ apply writes, not whether we have applied them in the past. I considered `get_applyable`, but I think `set_apply_writes` is clearer for app-writers and then this is symmetric.

src/http/http_rpc_context.h
achamayou: HTTP error codes in [200, 300) indicate success, this checks for that. I agree a comment is warranted :)

src/node/rpc/frontend.h
olgavrou: then maybe `should_apply_writes`?

src/enclave/rpc_context.h
```diff
@@ -103,7 +103,8 @@ namespace enclave
       set_response_header(name, fmt::format("{}", n));
     }
 
-    virtual bool response_is_error() const = 0;
+    virtual void set_apply_writes(bool apply) = 0;
```
achamayou, line 106: I wonder if an enum would be more descriptive:

```
enum Apply {
  ON_SUCCESS,
  ALWAYS
}
...
args.rpc_ctx->set_writes(Apply::ALWAYS);
```

But also, this is a perfectly good bike shed, and I will approve and shut up now.

src/node/rpc/frontend.h
eddyashton: I prefer `get`/`set`, unless this is significantly less readable?

olgavrou: Well get implies that we are `getting` some internal state of the context, where as `should/check` etc implies making a decision based on something. Without knowing the context it looks like you are `getting` what you are about to write

eddyashton: Fair, I hadn't twigged this looks like _writes_ as a noun, have changed it to `should_`

PR #1055, opened 2020-04-09
SmallBank with Flatbuffers

olgavrou:


samples/apps/smallbank/app/amalgamate.fbs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.	
+// Licensed under the Apache 2.0 License.
+
+table Amalgamate
+{
+    name_src:uint;
```
eddyashton, line 6: Should these be `uint64`? The other ID fields are.

olgavrou: Probably, I was just mimicking the way there were defined in the `small_bank_client` and `smallbank.cpp` files

samples/apps/smallbank/app/smallbank.cpp
```diff
@@ -23,6 +24,8 @@ namespace ccfapp
     static constexpr auto SMALL_BANKING_WRITE_CHECK = "SmallBank_write_check";
   };
 
+  static constexpr auto expected = http::headervalues::contenttype::TEXT;
```
achamayou, line 27: application/octet-stream would be more accurate, since this is arbitrary binary content rather than text. Not that it will have any functional or performance impact.

samples/apps/smallbank/app/smallbank.cpp
```diff
@@ -41,6 +44,45 @@ namespace ccfapp
   private:
     SmallBankTables tables;
```
achamayou, line 46: Checking the content type on every message is nice for debugging/users, but arguably not necessary in a benchmark.

samples/apps/smallbank/app/smallbank.cpp
```diff
@@ -106,235 +169,338 @@ namespace ccfapp
...
-        std::string name = params["name"];
-        auto account_view = tx.get_view(tables.accounts);
+        if (headers_unmatched(args))
+        {
+          set_unmatched_header_status(args);
+          return;
+        }
+
+        BankDeserializer bd(args.rpc_ctx->get_request_body().data());
+        auto name = to_string(bd.name());
```
achamayou, line 218: Can’t bd.name() be a string? Having to copy values out of the FB rather defeats the point of using it.

Edit: the names are integers, why is the table keyed by string? We should align.

src/http/http_builder.h
```diff
@@ -76,25 +77,31 @@ namespace http
...
     {
-      if (body == nullptr)
-      {
-        return nullptr;
-      }
-      else
+      return body;
+    }
+
+    void set_body(const std::vector<uint8_t>* b)
```
achamayou, line 89: That seems like it should just call the other set_body?

samples/apps/smallbank/app/smallbank.cpp
olgavrou: I don't know I was also wondering that. I can change the table to be keyed by int

samples/apps/smallbank/app/smallbank.cpp
olgavrou: Sure, I can leave the methods around for debugging purposes and remove the check from the test

samples/apps/smallbank/app/smallbank.cpp
```diff
@@ -41,6 +44,45 @@ namespace ccfapp
...
+    {
+      args.rpc_ctx->set_response_status(status);
+      args.rpc_ctx->set_response_header(
+        http::headers::CONTENT_TYPE, http::headervalues::contenttype::TEXT);
+      args.rpc_ctx->set_response_body(std::move(message));
+    }
+
+    void set_ok_status(RequestArgs& args)
+    {
+      args.rpc_ctx->set_response_status(HTTP_STATUS_OK);
```
eddyashton, line 81: Since there's no body here, we should (since #1054) be able to set `HTTP_STATUS_NO_CONTENT` and no content-type header.

samples/apps/smallbank/app/smallbank.cpp
eddyashton: Can we check this with @ashamis? I think the strings were used to match the original SmallBank spec, where (IIRC) we lookup from `name:string` to `id:int`? Still, if that's true, then the account names we pass in everywhere should be `string` not `int`

src/http/http_parser.h
```diff
@@ -77,6 +77,11 @@ namespace http
     s.resize(dst - s.data());
   }
 
+  static bool status_ok(http_status status)
+  {
+    return status >= 200 && status < 300;
+  }
```
olgavrou, line 83: not sure this should be in this file and not sure it should be called `status_ok`, @eddyashton ?

samples/apps/smallbank/app/smallbank.cpp
ashamis: Yes Eddy is right. Smallbank schema is defined as 
Account(Name, CustomerID),
Saving(CustomerID, Balance), Checking(CustomerID, Balance). 

so name should be a string

olgavrou: Ok I'll change the flatbuffer shcema to take a string for the account names instead of using integers

achamayou:
It's a little surprising that this doesn't seem to have a detectable performance impact. It may be interesting to have a picobench running encoding/decoding of one of the smallbank messages and the equivalent json and msgpack.

src/http/http_parser.h
eddyashton: This file's fine, but I think `status_success` rather than `status_ok` (`OK` is a single status, 200, and the general term for the `2xx` family seems to be 'Success').

samples/apps/smallbank/app/smallbank.cpp
olgavrou: Done

olgavrou:
smallbank payload serialize:   
----------------------------------------------------------------------------------
        flatbuffers_ser * |      10 |     0.006 |     640 |      - |  1562500.0
         json_msgpack_ser |      10 |     0.041 |    4050 |  6.328 |   246913.6
        flatbuffers_ser * |     100 |     0.064 |     637 |      - |  1569858.7
         json_msgpack_ser |     100 |     0.397 |    3971 |  6.234 |   251824.5
        flatbuffers_ser * |    1000 |     0.640 |     639 |      - |  1562981.1
         json_msgpack_ser |    1000 |     3.947 |    3947 |  6.169 |   253355.8

smallbank payload deserialize:
----------------------------------------------------------------------------------
        flatbuffers_des * |      10 |     0.001 |     120 |      - |  8333333.3
         json_msgpack_des |      10 |     0.058 |    5820 | 48.500 |   171821.3
        flatbuffers_des * |     100 |     0.011 |     110 |      - |  9090909.1
         json_msgpack_des |     100 |     0.558 |    5575 | 50.682 |   179371.6
        flatbuffers_des * |    1000 |     0.109 |     109 |      - |  9157425.3
         json_msgpack_des |    1000 |     5.582 |    5581 | 51.114 |   179156.0

smallbank payload merkle tree bench:
----------------------------------------------------------------------------------
    flatbuffers_mt_append * |      10 |     0.026 |    2560 |      - |   390625.0
     json_msgpack_mt_append |      10 |     0.026 |    2600 |  1.016 |   384615.4
    flatbuffers_mt_append * |     100 |     0.265 |    2653 |      - |   376930.4
     json_msgpack_mt_append |     100 |     0.267 |    2671 |  1.007 |   374390.2
    flatbuffers_mt_append * |    1000 |     2.665 |    2664 |      - |   375289.0
     json_msgpack_mt_append |    1000 |     2.680 |    2679 |  1.006 |   373160.4

large payload merkle tree bench:
----------------------------------------------------------------------------------
     fb_large_payload<10> * |     100 |     0.265 |    2649 |      - |   377498.1
       jm_large_payload<10> |     100 |     0.268 |    2676 |  1.010 |   373689.3
      fb_large_payload<100> |     100 |     0.426 |    4260 |  1.608 |   234740.7
      jm_large_payload<100> |     100 |     0.342 |    3420 |  1.291 |   292396.8
     fb_large_payload<1000> |     100 |     1.503 |   15031 |  5.674 |    66528.8
     jm_large_payload<1000> |     100 |     2.036 |   20363 |  7.687 |    49108.4
    fb_large_payload<10000> |     100 |    12.449 |  124489 | 46.995 |     8032.8
    jm_large_payload<10000> |     100 |    18.775 |  187749 | 70.875 |     5326.2

PR #1059, opened 2020-04-09
Add app docker image

achamayou:
Again, this is a change to support application CI & testing.


docker/app_ci
```diff
@@ -0,0 +1,11 @@
+FROM ccfciteam/ccf-ci-18.04-oe-8fe3697:latest
+
+RUN git clone https://github.com/microsoft/CCF.git
+RUN cd CCF \
```
jumaffre, line 4: Do we need to build from source? Could we not download the latest release instead?

achamayou: In general, yes, but right now there's a number of unreleased improvement over 0.9.1 we need in there, and there'll probably be a few more still.

But yes, in general, that's what we should use here, absolutely.

PR #1064, opened 2020-04-13
Fix how pbft deal with tx conflicts + other stuff

ashamis:
- Move adding of the request tx into the retry loop (otherwise we would be dropping the request on conflict)
- introduce and MIT licensed robin-hood hash that has the same api as std::unordered_map. This is used to replace some of PBFT's unordered_maps and we should start using this in other places.
- remove the un-used only once semantics files (This code path was broken anyway :())
- reorder the request in the pp so they match execution order, and in some circumstances make the backup execute said requests in single threaded order.
- other misc changes

More changes on how to make the secondary deal with conflicts in the near future.


src/node/rpc/frontend.h
```diff
@@ -213,7 +213,11 @@ namespace ccf
     std::optional<std::vector<uint8_t>> process_command(
       std::shared_ptr<enclave::RpcContext> ctx,
       Store::Tx& tx,
-      CallerId caller_id)
+      CallerId caller_id,
+      std::function<
+        void(Store::Tx& tx, enclave::RpcContext& ctx, RpcFrontend& frontend)>
```
eddyashton, line 218: Can we please name this type with a `using`, rather than repeating it? And I believe this default lambda can be replaced with `= {}`.

src/node/rpc/frontend.h
```diff
@@ -345,6 +349,7 @@ namespace ccf
       {
         try
         {
+          pre_exec(tx, *ctx.get(), *this);
```
eddyashton, line 352: So this pre-exec pattern may be needed (the "drop views on conflict" change is a bit dodgy, and we know it loses signatures too). But we believed it was safe for the PBFT case because we wouldn't get conflicts in the PBFT execution path - the order is pre-calculated/pre-executed to be non-conflicting? Can you explain why that assumption was wrong?

eddyashton:
Why do we need the robin-hood hash? If we think its faster, can we add a picobench simulating the workloads we think we're running, to confirm that it really is? And ideally a comparison of the memory footprint of each implementation, though we don't have an easy drop-in framework for this.

samples/apps/smallbank/app/smallbank.cpp
```diff
@@ -167,26 +173,33 @@ namespace ccfapp
...
+        {
           return make_error(
             HTTP_STATUS_BAD_REQUEST, "Savings account does not exist");
+        }
 
         if (savings_r.value() + value < 0)
         {
-          return make_error(
-            HTTP_STATUS_BAD_REQUEST, "Not enough money in savings account");
+          // We do not want to abort the entire tx, just not complete
```
achamayou, line 200: So there's an explicit flag to preserve writes on error now added by @eddyashton (set_apply_writes, #1054), pretty much exactly for this exact purpose: the transaction returns an error but should still be committed.

ashamis:
> Why do we need the robin-hood hash? If we think its faster, can we add a picobench simulating the workloads we think we're running, to confirm that it really is? And ideally a comparison of the memory footprint of each implementation, though we don't have an easy drop-in framework for this.

So the person who wrote the robin-hood implementation did this
https://martin.ankerl.com/2019/04/01/hashmap-benchmarks-01-overview/

src/node/rpc/frontend.h
ashamis: On the primary we are producing the order so there are conflicts. This should not be a problem on the backup. Since there are conflicts we are dropping the part of the tx that writes the request.

achamayou:
@ashamis it's grest that the author has their own benchmarks, but what @eddyashton was asking for is some benchmarks in the context where we use this showing it makes a difference to our use case.

The metrics for this PR show a performance hit for both end to end PBFT benchmarks, and no change elsewhere, so it's not immediately obvious that this is beneficial.

src/consensus/pbft/libbyz/pre_prepare.h
```diff
@@ -130,6 +138,14 @@ class Pre_prepare : public Message
 
   int64_t get_ctx() const;
 
+  void set_request_digest(uint32_t at, Digest& d);
+
+  bool should_reorder() const;
+
+  void record_tx_execution_conflict();
```
olgavrou, line 145: could you add comments here explaining what these methods do?

PR #1066, opened 2020-04-14
Update release and docker app instructions

achamayou:
Preparation for 0.9.2, for which the intended release notes are:

**Added**

- Handlers can decide if transaction writes are applied independently from error status (#1054)
- Scenario Perf Client is now part of the CCF install to facilitate performance tests (#1058)

**Changed**

- Handle writes when host is reconnecting (#1038)
- Member tables are no longer whitelisted for raw_puts (#1041)
- Projects including CCF's CMake files now use the same build type default (#1057)


docker/app_ci
```diff
@@ -1,11 +1,16 @@
 FROM ccfciteam/ccf-ci-18.04-oe-8fe3697:latest
 
-RUN git clone https://github.com/microsoft/CCF.git
-RUN cd CCF \
+ARG VERSION=0.9.2
+
+RUN apt install wget \
+    && wget https://github.com/microsoft/CCF/archive/v$VERSION.tar.gz \
+    && tar xf v$VERSION.tar.gz \
+    && rm v$VERSION.tar.gz
```
eddyashton, line 8: Minor: If this is always installing a release, can we download the prebuilt release tarball rather than building from the archived release sources?

achamayou: Doh, yes.

Issue #803, opened 2020-02-05
Investigate unnecessary copies


achamayou:
- [x] Sending pbft message
- [x] Data.cpp:17 (Data ctor)

achamayou:
@ashamis since you're looking at performance in PBFT, this is probably a good thing to have a look at.

ashamis:
yes I have been dealing with this issue. I have this mostly done.

PR #1067, opened 2020-04-14
Remove copy when sending messages in PBFT

ashamis:
Add a counter to the PBFT message so that we do not need to create a copy every time we attempt to send a message

related to #803

### DIFF
--- a/src/consensus/pbft/libbyz/message.cpp
+++ b/src/consensus/pbft/libbyz/message.cpp
@@ -10,19 +10,19 @@
 
 #include <stdlib.h>
 
-Message::Message(unsigned sz) : msg(0), max_size(ALIGNED_SIZE(sz))
+Message::Message(unsigned sz)
 {
   if (sz != 0)
   {
-    should_delete = true;
-
-    msg = (Message_rep*)malloc(max_size);
-    if (msg != nullptr)
+    int max_size = (ALIGNED_SIZE(sz));
+    msg_buf = (Message_rep*)malloc(max_size);
+    msg = std::move(make_Ref<MsgBufCounter>(msg_buf, max_size, true));
+    if (msg_buf != nullptr)
     {
-      PBFT_ASSERT(ALIGNED(msg), "Improperly aligned pointer");
-      msg->tag = -1;
-      msg->size = 0;
-      msg->extra = 0;
+      PBFT_ASSERT(ALIGNED(msg_buf), "Improperly aligned pointer");
+      msg_buf->tag = -1;
+      msg_buf->size = 0;
+      msg_buf->extra = 0;
     }
   }
   auth_type = Auth_type::unknown;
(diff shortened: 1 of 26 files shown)


src/consensus/pbft/libbyz/message.cpp
```diff
@@ -50,50 +50,43 @@ Message::Message(int t, unsigned sz)
...
   {
-    ((char*)msg)[i] = 0;
+    ((char*)msg_buf)[i] = 0;
   }
-  msg->size = aligned;
+  msg_buf->size = aligned;
 }
 
-bool Message::convert(char* src, unsigned len, int t, int sz, Message& m)
+bool Message::convert(char* src, unsigned len, int t, int sz, Message& msg_buf)
```
eddyashton, line 89: Minor - I think this should still be `m`. It looks like it shadows the member variable `msg_buf`, it doesn't actually because this is static, but that's non-obvious. But everywhere else, `msg_buf` refers to the `Message_rep*` member variable, so this `Message&` argument should be given a distinct name.

src/ds/ref.h
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include <utility>
+
+template <typename T>
+class Ref final
```
eddyashton, line 8: Why do we need a custom reference-counted container - can this be replaced by `std::shared_ptr`?

src/consensus/pbft/libbyz/message.h
```diff
@@ -118,6 +119,36 @@ class Message
...
+        free(msg);
+      }
+    }
+
+    Message_rep* msg; // Pointer to the contents of the message.
+    int max_size; // Maximum number of bytes that can be stored in "msg"
+                  // or "-1" if this instance is not responsible for
+                  // deallocating the storage in msg.
+    int32_t counter;
+    bool should_delete = false;
```
eddyashton, line 144: I don't understand the ownership model of this object. We have a `should_delete` bool, but also a comment that `max_size == -1` may affect deallocation, and on top of that a reference-count wrapper.

src/ds/ref.h
ashamis: My first attempt at this was with shared_ptr. The use of the atomics used in shared_ptr introduced a really high CPU overhead and make this not worthwhile

src/consensus/pbft/libbyz/message.h
ashamis: I will add this as a comment in the code:

There are 2 types of messages (in regards to memory ownership).
1 - the message owns its own buffer (This is the normal case)
2 - A projection on top of a message on an array

The comment about -1 is no longer correct and I am deleting it

src/ds/ref.h
olgavrou: Could `Ref` usage be replaced by a `unique_ptr` and have it passed between threads to avoid the copying? If this is shared between threads how do we guarantee that the internal state is not being modified by multiple threads?

achamayou:
@ashamis as discuss, let's target this and other changes to the pbft branch for now.

ashamis:
> @ashamis as discuss, let's target this and other changes to the pbft branch for now.

Yes - this has been pushed to the pbft branch => closing this PR

PR #1071, opened 2020-04-15
Add transaction rate setting to perfclient

lynshi:
The `--transaction-rate` flag allows one to specify the transaction rate (tx/s) of the client. If a transaction rate is specified, the client computes the interval between writes needed to maintain this rate, and waits this amount between writes.

### DIFF
--- a/samples/perf_client/perf_client.h
+++ b/samples/perf_client/perf_client.h
@@ -64,7 +64,7 @@ namespace client
     size_t max_writes_ahead = 0;
     size_t latency_rounds = 1;
     size_t generator_seed = 42u;
-    size_t ttransactions_per_s = 0;
+    size_t transactions_per_s = 0;
 
     bool sign = false;
     bool no_create = false;
@@ -127,7 +127,7 @@ namespace client
 
       app.add_option(
         "--transaction-rate", 
-        ttransactions_per_s,
+        transactions_per_s,
         "The number of transactions per second to send"
       );
 
@@ -391,8 +391,8 @@ namespace client
       size_t read;
       size_t written;
 
-      if (options.ttransactions_per_s > 0) {
-        write_delay_us = 1000000 / options.ttransactions_per_s;
+      if (options.transactions_per_s > 0) {
+        write_delay_us = 1000000 / options.transactions_per_s;
       }
 
       last_write_time = std::chrono::high_resolution_clock::now();
--- a/samples/perf_client/perf_client.h
+++ b/samples/perf_client/perf_client.h
@@ -13,8 +13,8 @@
 #include "ds/logger.h"
 
 // STL/3rdparty
-#include <chrono>
 #include <CLI11/CLI11.hpp>
+#include <chrono>
 #include <fstream>
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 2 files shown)


samples/perf_client/perf_client.h
```diff
@@ -63,6 +64,7 @@ namespace client
     size_t max_writes_ahead = 0;
     size_t latency_rounds = 1;
     size_t generator_seed = 42u;
+    size_t transactions_per_s = 0;
```
ashamis, line 67: Can this be made to be ms. Second seems like a very high granularity?

achamayou: It's just used to set the interval, it's purely a UX concern. We mostly report tx/s rates, so this seems consistent with our existing practice/numbers?

achamayou:
@lynshi you need to format your code, that can be done with:

```
./check-format -f src samples
```

samples/perf_client/perf_client.h
```diff
@@ -379,6 +391,12 @@ namespace client
       size_t read;
       size_t written;
 
+      if (options.transactions_per_s > 0) {
+        write_delay_us = 1000000 / options.transactions_per_s;
+        connection->set_tcp_nodelay(true);
```
eddyashton, line 396: We should set `TCP_NODELAY` in all cases, outside of this check. Even without a target transaction rate, this should make the performance tests more predictable.

samples/perf_client/perf_client.h
```diff
@@ -431,13 +449,31 @@ namespace client
       size_t& written,
       const std::shared_ptr<RpcTlsClient>& connection)
     {
+      if (write_delay_us > 0) {
```
eddyashton, line 452: If I'm reading this correctly, we can remove `time_elapsed_between_writes_ns` entirely:

```
while (std::chrono::high_resolution_clock::now() - last_write_time < write_delay)
{
  continue;
}
```

Then we can remove the `if (write_delay > 0)` checks here and below - this while loop will always break on the first iteration, and writing `last_write_time` all the time is harmless.

samples/perf_client/perf_client.h
```diff
@@ -289,6 +297,10 @@ namespace client
     timing::ResponseTimes response_times;
     timing::CommitPoint last_response_commit = {0, 0};
 
+    std::chrono::high_resolution_clock::time_point last_write_time;
+    int64_t time_elapsed_between_writes_us;
+    int64_t write_delay_us = 0;
```
eddyashton, line 302: This could be more strongly typed: `std::chrono::nanoseconds write_delay = 0;`

(I'd prefer nanoseconds rather than microseconds so we keep some precision as we approach 100k tx/s)

samples/perf_client/perf_client.h
achamayou: In throughput tests, I don't expect it will produce a measurable effect, we are writing heavily.

lynshi:
Thanks for the suggestions @eddyashton; I've implemented them. However, I still only set `TCP_NODELAY` when a transaction rate is specified, since I'm not sure if setting it for your tests is what you want.

PR #1074, opened 2020-04-16
Python linter + cleanup

eddyashton:
I intended to make a few minor changes to the Python infra, including cleaning up unused imports. The linter found a few other issues which I've also fixed.

Interesting changes (non-linter):
- `e2e_suite` no longer prints errors in success-green
- `perfclient`'s `--nodes` argument has an accurate comment, and we aim for the _minimum_ number of nodes by default rather than a maybe-interesting-if-there-are-2-threads number

### DIFF
--- a/tests/code_update.py
+++ b/tests/code_update.py
@@ -4,8 +4,6 @@ import infra.e2e_args
 import infra.ccf
 import infra.path
 import infra.proc
-import json
-import logging
 import os
 import subprocess
 import sys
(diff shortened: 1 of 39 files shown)


eddyashton:
@achamayou: This is pylint, only because that was the first linter that VSCode suggested. There are some additional failures that I think are safe, but cause the linter to fail. I'll take a closer look at them, add disables where appropriate, and add this to the CI in a separate PR.

achamayou:
@eddyashton pylint has a comment syntax for disables, which we can use where appropriate.

PR #1078, opened 2020-04-17
[DRAFT] Hosted jobs don't need a container

eddyashton:
I _think_ that the jobs we mark as `Hosted` don't actually need to run in our `nosgx` Docker container? Opening as a draft to check this. If true, this should knock the ~1m Docker spinup time off the Checks job.


eddyashton:
@achamayou You're right. Abandoning this for now.

PR #1080, opened 2020-04-20
Upgrade quickjs to 2020-04-12

achamayou:


achamayou:
Throughput for the logging scenario test seems to be substantially more volatile than for SmallBank, I don't have a good guess as to why, I think it's something that's probably worth looking into to.

Issue #839, opened 2020-02-13
Request only executed by primary is lost during view changes


olgavrou:
If the primary has executed a request r at seqno n, but no other replica has, and there is a view change then we see the following:

- the pre-prepare of that request is added to the view change as incomplete (`View_info::add_incomplete()`)
- when the view change message is created, the new view keeps the requests of seqnos that have prepared but not of seqnos that have only pre-prepared (`View_info::view_change()`)
- when all replicas process the new view (`Replica::process_new_view()`) they will try to execute the request for seqno n, `execute_tentative` will make `last_tentative_execute` advance, the replicas will send out prepares for that seqno
- the replicas will have advanced their `last_tentative_execute` index but they can never commit for seqno n because the prepared certificate will never be complete. The prepared certificate will never be complete because the pre-prepare info will never be complete, and that is because the request is missing.

I believe this was not causing issues before in `test_UDP_with_delays` because (a) this doesn't happen very often and (b) certificates were being marked as compete because we were not updating f correctly, and therefore these requests were able to commit since the prepared certificate was coming back as complete (see `Prepared_cert::is_complete()`

PR #1085, opened 2020-04-20
Fix no-op pre prepare

olgavrou:
Resolves problem described in #839 

The problem here is that the no-op pre-prepare digests were not being calculated consistently during view changes (no-op pre-prepares are created per-replica vs created by one replica and broadcasted to other replicas as is done in normal execution) and so the no-op pre-prepares would never complete.

This fixes that and also puts the no-op pre-prepares into the ledger during view changes. In order for the no-op pp's to go into the ledger, we need them to not be signed, as each no-op pre-prepare will be signed by the replica that created it, leading eventually to an inconsistent merkle tree. 

These pre-prepares will be signed _eventually_ as part of the next pre-prepare.

### DIFF
--- a/src/consensus/pbft/libbyz/pre_prepare.cpp
+++ b/src/consensus/pbft/libbyz/pre_prepare.cpp
@@ -104,12 +104,9 @@ Pre_prepare::Pre_prepare(
   if (prepared_cert == nullptr)
   {
     // if there are no pre-prepare proofs (e.g. no-op pre-prepare)
-    // then set zero out the rest of the allocated message
+    // then zero out the rest of the allocated message
     rep().num_prev_pp_sig = 0;
-    size_t s = sizeof(Pre_prepare_rep) + rep().rset_size +
-      rep().n_big_reqs * sizeof(Digest);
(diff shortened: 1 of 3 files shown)


src/consensus/pbft/libbyz/pre_prepare.cpp
```diff
@@ -102,7 +103,10 @@ Pre_prepare::Pre_prepare(
 
   if (prepared_cert == nullptr)
   {
+    // if there are no pre-prepare proofs (e.g. no-op pre-prepare)
+    // then zero out the rest of the allocated message
     rep().num_prev_pp_sig = 0;
+    std::memset(proofs(), 0, size() - size_to_proofs());
```
ashamis, line 109: std::fill

ashamis, line 109: please use std::fill

olgavrou:
@achamayou that is not a bad idea, I'll see how easy/hard it is to add that. In any case when the view change test is enabled this code path should be crossed in every view change.

olgavrou:
Adding unit test in separate PR

Issue #1049, opened 2020-04-08
Concise and version-less ledger secrets in KV


jumaffre:
As described by the TR and in the current implementation of recovery with key shares, the schema of the `ccf.shares` table is as follows (assuming three ledger secrets, at `1`, `10` and `20`):

```
ccf.shares (k -> v): 0 -> [ ( {1, ls1}, {10, ls10}, {20, ls20} )k_z ; [(share_0)ksk_0; (share_1)ksk_1; ... ; (share_m)ksk_m; ] ]
```
_Note: (x)y indicates that x is encrypted by y_

There are two issues with this scheme:
- The first one is pretty obvious: each re-key and recovery will generate a new ledger secret that has to be stored (encrypted) in the `ccf.shares` table. The size of the value of this table will grow linearly with the number of ledger secrets created since the creation of the service.
- On re-key, CCF needs to issue new shares. However, it is not trivial to know at which version the rekey transaction will be executed before this one is actually executed. We briefly discussed only issuing the shares on the next signature or using the `commit_reserved()` KV mechanism for this but all schemes suffer from some limitations/added complexity. 

The following design should address both issues.

First, the version of each ledger secret should not be included in the `ccf.shares` table. Instead, it is when a node replays the public ledger that a new hook on the `ccf.shares` table detects when this table is modified (i.e. new ledger secrets have been issued). At that point, the hook knows which version the new ledger secret is applicable from and can update the encryptor accordingly.

Second, the latest ledger secret should encrypt the previous ledger secret (and so on) so that we only need to keep track of the (encrypted with `k_z`) very latest ledger secret in the `ccf.shares` table. If, when a node replays the public ledger, we keep track of the encrypted ledger secrets, then once we re-store `k_z` (from member shares), we can decrypt the latest ledger secret, which in turn can decrypt the previous ledger secret and so on. 

The schema for the `ccf.shares` table then becomes:

```
ccf.shares (k -> v): 0 -> [ (latest_ls)k_z ; (previous_ls)latest_ls; [(share_0)ksk_0; (share_1)ksk_1; ... ; (share_m)ksk_m; ] ]
```

Taking the previous example of 3 ledger secrets, the public ledger would be replayed as follows:

```
0 -> [ (ls1)k_z1;     None;     [encrypted shares at 1]  ] // replayed at version 0
0 -> [ (ls10)k_z10; (ls0)ls10;  [encrypted shares at 10] ] // replayed at version 10
[...]

PR #1088, opened 2020-04-21
Shares table is now of constant size

jumaffre:
Addresses most of https://github.com/microsoft/CCF/issues/1049.

**Before:** 
- The value of (the only key in) the `ccf.shares` table would get larger after each re-key/recovery as the whole set of ledger secrets since the beginning of time were serialised and wrapped there.

**Now:**
- As per the design outlined in #1049, the `ccf.shares` table only keeps track of the latest ledger secret wrapped with `k_z` and the penultimate ledger secret encrypted with the latest ledger secret.
- On recovery of the public ledger, a new hook (see `setup_recovery_hook()` in `node_state.h`) constructs a list of recovered encrypted secrets and when the members' recovery shares are combined to `k_z`, the latest ledger secret is unwrapped and then used to decrypt the penultimate one, etc. 

Note:
The implementation is not quite as straightforward as in #1049 as the version at which the ledger secret is written in the `ccf.shares` table is not necessarily the version at which they are applicable from. This is the case on recovery where new ledger secrets are created to protect the integrity of the governance-only transactions (after having read the public ledger) but only written to the ledger later, once the historical ledger secrets have been recovered.

Also:
- Refactored a bunch of things around the `LedgerSecrets` class, mainly due to the fact that the class was originally written pre-light-genesis/with sealing in mind. Its API should be a lot more straightforward (e.g. no need to promote a `LedgerSecret` anymore) and the internal implementation now uses a `std::list`. 
- Renaming in places for clarity (e.g. `KeyShareInfo` -> `RecoverySharesInfo`).

--

Next:
- Get re-keying working with this scheme
- Remove sealing


src/node/ledger_secrets.h
```diff
@@ -49,154 +46,107 @@ namespace ccf
...
-            fmt::format("Ledger secret could not be unsealed: {}", v));
-        }
-
-        LOG_DEBUG_FMT(
-          "Ledger secret successfully unsealed at version {}", it.key());
-
-        add_secret(v, LedgerSecret(s.value()), false);
-
-        restored_versions.push_back(v);
+        return {};
```
eddyashton, line 101: Minor: Please `return std::nullopt;` here. I know this does the correct thing, but I'm wary of the obscure bugs when the return type changes and we miss that this does something totally different.

src/node/ledger_secrets.h
```diff
@@ -206,38 +156,42 @@ namespace ccf
...
-        return {};
+        if (s.version == v)
+        {
+          return s.secret;
+        }
       }
-
-      return search->second;
+      LOG_FAIL_FMT("Ledger secret at version {} does not exist", v);
+      return {};
```
eddyashton, line 194: As above.

src/node/ledger_secrets.h
jumaffre: Since this is minor, I will fix this in the follow-up PR which will come later today.

PR #1089, opened 2020-04-21
[WIP] WebSockets support

achamayou:
This is very bare bones support for Web Sockets, currently with the following caveats:

Blockers for running SmallBank:

- [x] No input CCF frame, method is hardcoded
- [x] No output CCF frame, no way to get commit seqno/view/rc
- [x] Commit info not set
- [ ] Perf client has no ws support
- [ ] Does not work with forwarding/pbft yet

Nice to have later:
- [ ] Does not close connections politely, does not implement heartbeats
- [ ] Can't pylint very well
- [ ] No support for masked requests
- [ ] No support for fragmented requests
- [ ] Can't compute size of parsed FB in Python!

A few things that came up along the way:

1. Flatbuffers are quite large because of alignment constraints. For example an InHeader with a 16 bytes path results in 44 bytes (!) of header.
2. Nesting flatbuffers isn't straightforward, there are specific APIs to not forget to call.
3. Using flatbuffers as headers works in C++ because there's a FLATBUFFERS_TRACK_VERIFIER_BUFFER_SIZE define that makes the verifier track the size of the buffer. This (and size prefixed flatbuffers) is largely absent from bindings written in other languages. So parsing fb-prefixed output in Python is impossible without additional, manual framing.
4. The C++ API is passable, the Python API is so poor that I am struggling to imagine how it could be any worse.

So I think we should probably have a discussion about:

1. Are Flatbuffers the right choice for our WebSocket frames?
2. Are Flatbuffers the right choice in general, for our user-facing serialisation and in particular for the KV?


src/http/http_parser.h
```diff
@@ -432,3 +434,128 @@ namespace http
     }
   };
 }
+
+namespace ws
```
eddyashton, line 438: Can these move to their own files?

src/http/http_rpc_context.h
```diff
@@ -339,4 +335,204 @@ namespace enclave
       packed,
```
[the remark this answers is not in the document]
eddyashton: `make_rpc_context` is mostly used as a convenience function for the tests, creating a working context without going through a TLS connection. Maybe we should split this useage into a separate function (always returning a `HttpRpcContext`), separate from the `pbft_config`, `forwarder`, and `node_state` uses, which want to be closer to `HttpRpcEndpoint::handle_request`.

src/http/http_parser.h
achamayou: Sure, there's some slight awkwardness with HTTP* things being either HTTP as opposed to WS or HTTP the superset of HTTP and WS, including the upgrade path. I do not have a clever naming idea to resolve this though.

src/http/http_rpc_context.h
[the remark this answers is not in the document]
achamayou: Yes, for forwarding I think we must pass a flag telling the remote which it is, because it won't have seen the upgrade handshake. I don't see how things could work otherwise.

Technically, if we stick to not supporting fragmentation and/or de-fragment before we send, we could probably tell them apart because the MSB will always be set on byte 0 of WS frames, whereas none of the first letters of HTTP verbs will have it on...

But just because we can, doesn't mean we should!

src/http/ws_parser.h
```diff
@@ -0,0 +1,141 @@
...
+
+          size = be16toh(*(uint16_t*)data.data());
+          state = READ_BODY;
+          return size;
+        }
+        case READ_LLEN:
+        {
+          assert(data.size() == 8);
+
+          size = be64toh(*(uint64_t*)data.data());
```
olgavrou, line 118: does it matter that size is `uint16_t` but here we are using `uint64_t`?

achamayou: Yes, thank you for catching that!

tests/infra/clients.py
```diff
@@ -415,12 +418,36 @@ def __init__(
...
+        InHeader.InHeaderStart(builder)
+        InHeader.InHeaderAddPath(builder, path)
+        header = InHeader.InHeaderEnd(builder)
+        builder.Finish(header)
+        h = builder.Output()
+        # FIN, no RSV, BIN, UNMASKED every time, because it's all we support right now
+        frame = websocket.ABNF(1, 0, 0, 0, websocket.ABNF.OPCODE_BINARY, 0, h + payload)
+        self.ws.send_frame(frame)
+        out = self.ws.recv_frame().data
+        ohs = 20  # Calculate, somehow
```
achamayou, line 447: In C++, the Verifier can do that, so long as the right define is set. No such luck in Python!

Slightly awkward, but not impossible to monkey patch.

achamayou: Actually this seems to be genuinely difficult...

achamayou: Going to skip that for now in favour of trying to get smallbank running asap.

PR #1090, opened 2020-04-22
Update to OE-0.9.0-rc1

eddyashton:


achamayou:
The perf kit will need to be re-playbook'ed, since it's not using docker. I suggest tearing it down completely and making it fresh even.

PR #1092, opened 2020-04-22
enable view changes test for pbft

olgavrou:
- client proxy clears reply certificates when there is a view change (all other certificates are cleared in replica during view change)
- setting the correct view on the encryptor during view changes
- handling prepares correctly during playback
- added `getCommitLocal` rpc for test to monitor the late joiners commit status without going through the protocol


CMakeLists.txt
```diff
@@ -540,7 +540,6 @@ if(BUILD_TESTS)
     NAME late_joiners
```
jumaffre, line 540: Could we rename this test to `late_joiners_pbft` as it is the only e2e test that is PBFT specific?

src/consensus/pbft/libbyz/client_proxy.h
```diff
@@ -346,6 +348,15 @@ void ClientProxy<T, C>::recv_reply(Reply* reply)
             << " reply->full: " << (reply->full() ? "true" : "false")
             << " reps.cvalue: " << (void*)ctx->t_reps.cvalue() << std::endl;
 
+  if (current_view < reply->view())
+  {
+    LOG_INFO_FMT(
```
jumaffre, line 353: `LOG_DEBUG_FMT`?

src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -1165,6 +1166,7 @@ void Replica::send_prepare(Seqno seqno, std::optional<ByzInfo> byz_info)
 
 void Replica::send_commit(Seqno s, bool send_only_to_self)
 {
+  LOG_DEBUG_FMT("Sending commit for seqno: {}", s);
```
jumaffre, line 1169: Is this going to be very verbose?

tests/late_joiners.py
```diff
@@ -198,11 +198,10 @@ def run(args):
                         if args.consensus == "pbft"
                         else args.raft_election_timeout / 1000
                     )
-
                     if node.node_id == cur_primary_id and args.consensus == "pbft":
-                        # if pbft suspend the primary for twice the elecetion timeout
+                        # if pbft suspend the primary for > 2 the elecetion timeout
```
jumaffre, line 202: Typo: election

tests/late_joiners.py
```diff
@@ -198,11 +198,10 @@ def run(args):
...
                         # in order to make sure view changes will be triggered
-                        et += et
+                        et += 1.5 * et
```
eddyashton, line 204: Minor: Took me a while to parse this. Can we keep the comment in English "suspend the primary for more than twice the election timeout", and make this line `et = et * 2.5` so it more obviously matches the comment? And maybe rename the variable from `et` to `suspend_time` (we modify the suspend time in this case, the election timeout is a constant).

src/consensus/pbft/libbyz/replica.cpp
olgavrou: I keep adding this every time I debug pbft. It is a good indicator of what state the replica is in if it does send a commit message but then it doesn't add the commit message to it's certificate, so then you don't see the next debug message in `Replica::send_commit` (https://github.com/microsoft/CCF/blob/master/src/consensus/pbft/libbyz/replica.cpp#L1188)

achamayou: Things that are per-transaction or more often should be TRACE, this is presumably per-batch so DEBUG or TRACE should be fine.

tests/late_joiners.py
achamayou, line 202:
```suggestion
                        # if pbft suspend the primary for > 2 the election timeout
```

achamayou: For typos, I suggest we use the suggestion feature in GitHub as much as possible. It's often quicker than typing "Typo:" :)

CMakeLists.txt
ashamis: From my understanding this should not be PBFT specific but this catches the known RAFT bug?

olgavrou: yes exactly

src/consensus/pbft/libbyz/replica.cpp
ashamis: Please use trace

CMakeLists.txt
olgavrou: I'll change it to `late_joiners_pbft` for now and we can add one for raft later

PR #1093, opened 2020-04-23
If no default, do not set an empty var to allow Python default logic …

achamayou:
…to kick in


cmake/common.cmake
```diff
@@ -398,10 +398,12 @@ function(add_e2e_test)
       TEST ${PARSED_ARGS_NAME} APPEND PROPERTY LABELS ${PARSED_ARGS_CONSENSUS}
     )
 
-    set_property(
-      TEST ${PARSED_ARGS_NAME} APPEND
-      PROPERTY ENVIRONMENT "DEFAULT_ENCLAVE_TYPE=${DEFAULT_ENCLAVE_TYPE}"
-    )
+    if(${DEFAULT_ENCLAVE_TYPE})
```
eddyashton, line 401: This doesn't seem to work - both of the debug builds are still using `--enclave-type release`. [CMake docs](https://cmake.org/cmake/help/latest/command/if.html#condition-syntax) are as convoluted as ever. Maybe `if(DEFINED DEFAULT_ENCLAVE_TYPE)`?

achamayou: CTest delenda est.

PR #1094, opened 2020-04-23
Rekey with recovery shares

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1049.

This PR adds support for re-keying and recovery with key shares. Most changes are in `share_manager.h` (cleaning up the API for this class) and in the recovery hook in `node_state.h`. 

I have also updated the `e2e_suite` to test for these in combination and tidied up some things in `memberclient.py`.


src/node/node_state.h
```diff
@@ -1481,15 +1483,37 @@ namespace ccf
...
+                // from the hook plus one. Otherwise (recovery), use the
+                // version specified.
+                ledger_secret_version =
+                  v.value.wrapped_latest_ledger_secret.version ==
+                    kv::NoVersion ?
+                  (version + 1) :
+                  v.value.wrapped_latest_ledger_secret.version;
+              }
+
+              if (
```
jumaffre, line 1505: This check is necessary for the following reason. On a pure re-share (i.e. no new ledger secret created, for example when a new member is added), the `ccf.shares` table is now updated as follows:

```
Before:
ccf.shares: 0 -> [ (Latest_LS)k_z0; (Previous_LS)Latest_LS; (shares)k_z0 ]

After:
ccf.shares: 0 -> [ (Latest_LS)k_z1; None                  ; (shares)k_z1 ]
```

When replaying the ledger, the second value is not added to the `recovery_ledger_secrets` list as the ledger secrets were not updated in this case. When the recovery is finished and the members' shares have been combined, `Latest_LS` can be retrieved from the latest state and decrypted and the historical secrets decrypted in turn.

tests/memberclient.py
```diff
@@ -53,33 +47,59 @@ def test_add_member(network, args):
...
+
+
+@reqs.description("Issue new recovery shares (without re-key)")
+def test_update_recovery_shares(network, args):
+    primary, _ = network.find_primary()
+    network.consortium.update_recovery_shares(primary)
+    return network
+
+
+def assert_recovery_shares_update(func, network, args, **kwargs):
```
jumaffre, line 74: Wrapping the "extended" part (i.e. checking that the recovery shares actually change) of the add/retire member and set recovery threshold tests so that they are only called from `memberclient.py`. When the suite is run, the "extended" part is skipped for conciseness/speed.

tests/recovery.py
```diff
@@ -10,7 +10,7 @@
 
 @reqs.description("Recovering a network")
 @reqs.recover(number_txs=2)
-def test(network, args, use_shares=False):
+def test(network, args, use_shares=True):
```
jumaffre, line 13: We now test recovery with key shares in the e2e suite runs (instead of sealing key).

tests/suite/test_requirements.py
```diff
@@ -71,6 +71,21 @@ def check(network, args, *nargs, **kwargs):
     return ensure_reqs(check)
 
 
+def sufficient_member_count():
```
jumaffre, line 74: In practice, when we randomise the test suite (i.e. Daily build), the tests marked with this (`test_retire_member`) will often be skipped as the recovery threshold is set to the initial number of members (3) and there is only one test in the suite that adds a new member. 

I'm OK with that for now as we run the test non-randomised in the CI anyway. But if we want to extend this, we could make sure we run the `test_add_member` when this condition fails.

tests/memberclient.py
```diff
@@ -216,49 +236,55 @@ def run(args):
...
-            primary, network.consortium.get_member_by_id(0)
-        )
-
-        LOG.debug("Retired member cannot make a new proposal")
-        try:
-            response = network.consortium.get_member_by_id(0).propose(
-                primary, script, 0
+        # Membership changes trigger re-sharing and re-keying and are
+        # only supported with Raft
+        if args.consensus == "raft":
```
jumaffre, line 241: https://github.com/microsoft/CCF/issues/1099

src/node/node_state.h
achamayou: I think this is worth a comment inline.

PR #1100, opened 2020-04-24
Get time in enclave from host memory, not OCall

eddyashton:
Opening as a draft to see how this affects performance.

We have worker threads inside the enclave, and sometimes they want to idle because there is no work today. Currently they busy-wait with `__mm_pause`, which may do nothing, and wastes processor time. To actually idle, they need some concept of time, and a way to sleep. This PR adds that concept of time (with a uv job on the host writing time to a known memory location, which the enclave reads), and a sketch of backoff behaviour (busy-wait for a while, then start sleeping for chunks of time). If this works, I'll look at a more precise idle notification system (OCalling to wait on a `condition_variable` in host-space).

### DIFF
--- a/src/enclave/enclave.h
+++ b/src/enclave/enclave.h
@@ -5,8 +5,8 @@
 #include "crypto/hash.h"
 #include "ds/logger.h"
 #include "ds/oversized.h"
-#include "interface.h"
 #include "enclave_time.h"
+#include "interface.h"
 #include "node/entities.h"
 #include "node/network_state.h"
 #include "node/node_state.h"
@@ -248,7 +248,7 @@ namespace enclave
         }
 
         bp.run(circuit->read_from_outside(), [](size_t consecutive_idles) {
-          static uint64_t idling_start_time;
+          static std::chrono::microseconds idling_start_time;
           const auto time_now = enclave::get_enclave_time();
 
           if (consecutive_idles == 0)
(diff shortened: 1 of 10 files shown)


achamayou:
This may be relevant: https://github.com/openenclave/openenclave/pull/2929

src/enclave/enclave.h
```diff
@@ -245,7 +246,28 @@ namespace enclave
...
+          static std::chrono::microseconds idling_start_time;
+          const auto time_now = enclave::get_enclave_time();
+
+          if (consecutive_idles == 0)
+          {
+            idling_start_time = time_now;
+          }
+
+          // Handle initial idles by pausing, eventually sleep (in host)
+          constexpr std::chrono::milliseconds timeout(5);
```
olgavrou, line 260: do we want this to be configurable?

src/enclave/enclave_time.h
```diff
@@ -0,0 +1,27 @@
...
+  extern std::atomic<std::chrono::microseconds>* host_time;
+  extern std::chrono::microseconds last_value;
+
+  static std::chrono::microseconds get_enclave_time()
+  {
+    // Update cached value if possible, but never move backwards
+    if (host_time != nullptr)
+    {
+      const auto current_time = host_time->load();
+      if (current_time > last_value)
```
achamayou, line 19: We probably want to log this for now, make the count available to the operator later.

src/enclave/enclave.h
eddyashton: In future yes, but we also want the sleep to do something smarter - I'll plumb in configurable parameters in a future PR.

PR #1103, opened 2020-04-27
Use 0.9 release of OpenEnclave

eddyashton:
I've rebuilt the perf CI machines on OE 0.9 as well.


achamayou:
Could you also re-build and update the app container for TPAL?

eddyashton:
> Could you also re-build and update the app container for TPAL?

Done.

Issue #851, opened 2020-02-17
It should be possible to have more than 255 consortium members


jumaffre:
The support for member key shares introduces a hard cap on the number of members active at a given time in the consortium: 255, as per the limitations of the `sss` library.

It should be possible for CCF to have more than 255 active members at a given time. If we stick to `sss`, we may decide to change the 1-to-1 mapping between members and recovery key shares, i.e. not all members are given a recovery key share.

Relates to https://github.com/microsoft/CCF/issues/51

achamayou:
@jumaffre can we make this limitation explicit in the documentation? We have separately agreed that not all members should have a share, which changes this slightly to "it should be possible to have more than 255 members with a share".

In the absence of a likely use case in the near future, I think documenting the limitation is sufficient, keeping an open-ended issue does not seem necessary.

Issue #1021, opened 2020-04-01
Support for share-less members


jumaffre:
#51 introduced recovery shares for each member. However, there may be cases where only _some_ members are privileged enough to be able to recover a crashed service. In this case, only these members should be handed recovery shares. 

Notes:
- When adding a new member, only members that have a public encryption key should be handed recovery shares. Other members should not be counted in the total number of shares generated.
- It should be possible to determine from the Lua constitution whether a member has been given a recoverable key share. 
- If a share-less member is added or removed, there is no need for CCF to issue new share, not re-key the ledger.

achamayou:
It should also be possible to determine in a vote if a member that is proposed for addition (and probably for removal as well) will receive/has a share.

achamayou:
There is no immediate requirement for this, let's re-open if/when needed.

achamayou:
This is necessary to support the operating member governance model, particularly in models where this only a single high-privileged member in charge of recovery (share threshold is effectively one).

PR #1107, opened 2020-04-28
Number of active members is limited to 255

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/851

Following #51, the number of active members at any given time should be <= 255 (limitation of the underlying secret sharing library). We believe this is OK for now (until we decide that not all members should be handed a recovery share, see https://github.com/microsoft/CCF/issues/1021).

CCF will now return a meaningful error message if the number of active members > 255 (either if the service is `OPENING` when a member is added or when the service is `OPEN` and a member transitions from `ACCEPTED` to `ACTIVE`).


src/node/genesis_gen.h
```diff
@@ -88,8 +88,7 @@ namespace ccf
 
     auto add_member(
       const std::vector<uint8_t>& member_cert_pem,
-      const std::vector<uint8_t>& member_keyshare_pub,
-      MemberStatus member_status = MemberStatus::ACTIVE)
+      const std::vector<uint8_t>& member_keyshare_pub)
```
jumaffre, line 91: IIRC, this was written before the `ccf.service` table existed. The initial status of the member can be directly deduced from the state of the service. 


This is why many unit tests now require the `g.create_service({})` call.

src/node/service.h
```diff
@@ -9,10 +9,6 @@
 
 namespace ccf
 {
-  // ServiceId is used at the key of the SERVICE table. As there is only one
-  // service active at a given time, this key is always 0.
-  using ServiceId = uint64_t;
```
jumaffre, line 14: This was unused.

src/apps/lua_generic/test/lua_generic_test.cpp
```diff
@@ -318,6 +319,7 @@ TEST_CASE("simple bank")
   Store::Tx gen_tx;
   GenesisGenerator gen(network, gen_tx);
   gen.init_values();
+  gen.create_service({});
```
achamayou, line 322: Can that be rolled into init_values()?

src/node/genesis_gen.h
```diff
@@ -104,15 +103,33 @@ namespace ccf
...
       }
 
+      MemberStatus member_status = MemberStatus::ACCEPTED;
+      auto service_status = get_service_status();
+      if (!service_status.has_value())
+      {
+        throw std::logic_error("Failed to get active service");
+      }
+
+      // If the service is opening, members are added as ACTIVE
```
achamayou, line 121: Is that right? Do we then reset their status after opening?

It wouldn't be right for a member to be added who has neither voted on opening nor acked to be counted as active.

src/apps/lua_generic/test/lua_generic_test.cpp
jumaffre: We should do that. However, the whole `GenesisGenerator` class needs splitting and I didn't want to start combining some functions before doing that. Ideally, there will eventually be a `init_service(cert)` that initialises the state of the service.

src/node/genesis_gen.h
jumaffre: That's a very good point. We don't reset their status after opening the service indeed. 

I see two options:
- All members are added as `ACCEPTED` and `ACCEPTED` members can propose/vote when the service is opening. The members who vote for the service to open automatically become `ACTIVE` when the service is open - other members stay `ACCEPTED`. 
- All members are added as `ACTIVE` when the service is opening. When the service is opening, the members who haven't voted for the proposal are reverted to `ACCEPTED`.

achamayou: I am starting to wonder if a single value is a good way to capture this.

Members can:

1. Have endorsed the ledger up to a point (vote in opening, ack)
2. Vote
3. Count towards a vote count
4. Use their keyshare to execute a recovery

achamayou: Having read the relevant parts of the TR and thought about this a bit, I don't think the service status should have a bearing on this.

```
ACCEPTED - Ack (no exceptions) -> ACTIVE - Removal proposal passes -> RETIRED
```

This should be the state machine at all times. Members can never be automatically ACTIVE, because it would wrongly imply their participation, and it would count towards quorum, potentially stalling votes. That's true all the same when OPENING.

So I suggest a third option, which is that members have to Ack before they can vote/be counted.

If we want to avoid unnecessary acking, I think the right solution is to allow ACCEPTED members to always vote, and count the first vote as an ack. This is problematic though, for them to vote, they must be counted, and counting identities that never have shown activity or signed a single thing with their private key seems overly optimistic.

achamayou: @jumaffre as discussed, let's stick with the TR approach for now, which has the merit of being simple to explain and to implement. Only ACTIVE members can vote. Members become ACTIVE on first ack.

jumaffre: Raised https://github.com/microsoft/CCF/issues/1113. Merging this PR now as it does not change what will be addressed in #1113.

PR #1108, opened 2020-04-28
sgxinfo should run on driverless machines

eddyashton:
Failing to run one of these steps shouldn't halt the entire script. Changes output on driverless boxes from

```
$ ./tests/sgxinfo.sh 
DRIVER INFO:
libkmod: ERROR ../libkmod/libkmod.c:586 kmod_search_moddep: could not open moddep file '/lib/modules/4.4.0-18362-Microsoft/modules.dep.bin'
modinfo: ERROR: Module alias intel_sgx not found.
```

to

```
$ ./tests/sgxinfo.sh 
DRIVER INFO:
libkmod: ERROR ../libkmod/libkmod.c:586 kmod_search_moddep: could not open moddep file '/lib/modules/4.4.0-18362-Microsoft/modules.dep.bin'
modinfo: ERROR: Module alias intel_sgx not found.

PSW INFO:
...

DCAP CLIENT INFO:
...

SGX INFO:
./tests/sgxinfo.sh: line 23: /opt/openenclave/bin/oesgx: No such file or directory
```

### DIFF
--- a/tests/infra/remote.py
+++ b/tests/infra/remote.py
@@ -637,8 +637,7 @@ class CCFRemote(object):
                 f"Unexpected CCFRemote start type {start_type}. Should be start, join or recover"
             )
 
-        # Necessary for the az-dcap-client >=1.1 (https://github.com/microsoft/Azure-DCAP-Client/issues/84)
-        env = {"HOME": os.environ["HOME"]}
+        env = {}
         self.profraw = None
         if enclave_type == "virtual":
             env["UBSAN_OPTIONS"] = "print_stacktrace=1"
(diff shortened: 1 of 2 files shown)


achamayou:
Actually I think I want to tweak this some more, because the modinfo lookup may fail even though the driver is loaded, and that's even typically the case inside a container.

When that happens, the output should be "SGX driver loaded, version could not be determined" or something along those lines. "Not found" makes it sound like it's not there at all.

eddyashton:
> Actually I think I want to tweak this some more, because the modinfo lookup may fail even though the driver is loaded, and that's even typically the case inside a container.
> 
> When that happens, the output should be "SGX driver loaded, version could not be determined" or something along those lines. "Not found" makes it sound like it's not there at all.

I could add `lsmod | grep intel_sgx || echo "Not loaded"` to detect if the module is loaded, which seems to work even when `modinfo` doesn't. Unless you can recommend a better diagnostic?

achamayou:
@eddyashton sounds good!

Issue #968, opened 2020-03-20
Endpoints should look like HTTP resources


eddyashton:
Our method names inherited from JSON-RPC are looking increasingly out-of-place. We should do a renaming pass: `/users/LOG_get_public` to `GET /users/log/public` or similar.

We're also considering changing the frontend prefixes, from `users` to `app`, and from `members` to `gov`. It would make sense to do this in the same PR.

achamayou:
Not going to be in time for 0.9, will go in 0.9.1.

achamayou:
We should aim for compliance with https://github.com/Microsoft/api-guidelines/blob/vNext/Guidelines.md as much as possible.

PR #1111, opened 2020-04-28
Add getTxStatus RPC

eddyashton:
This adds a `getTxStatus` RPC, replacing the `getCommit(N)` variant. `getCommit` now has a single purpose of fetching the current kv version, and the "is this committed" logic is handled in a single place, by the service.

I'm unsure about naming:
- I've used `term` and `commit` for consistency with existing RPCs/header names/externally facing things, are we planning to convert all of these to `view` and `index`? FWIW, I'm still fond of formalising `version = view.index`, so I've sprinkled that into the docs.
- What do we call a rolled back transactions? I've gone with `LOST`, but am open to suggestions.
- Should we add an `UNKNOWN` (I've never heard of this transaction) state distinct from `PENDING` (I know it, but I don't know if it is globally committed)? A long time in the `UNKNOWN` state may indicate that we've had an election but haven't yet reached your index, but this is only a heuristic (not true for partitioned nodes), and introduces possibility of odd response chains (`UNKNOWN` -> `PENDING` -> `UNKNOWN` -> `PENDING` -> ... -> [`COMMITTED`|`LOST`]).

I believe in the current implementation, if you ask a single node, the response will always transition exactly once from `PENDING` to either `COMMITTED` or `LOST`, and will be that way forever. And the transition will be consistent across all nodes - once you've heard `COMMITTED`, another node may still say `PENDING`, but can never say `LOST` (and vice versa). These are things I don't think we've clearly described in the docs, so I've added a note - let me know if it makes sense.

I've also removed the `SigHttpRpcTlsClient` - its functionality was almost entirely (and now entirely) handled by the base `HttpRpcTlsClient`.

EDIT - leaving the original description for posterity, but here's a more accurate current state:

- Endpoint is simply `/tx` (ie `/users/tx`, `/members/tx`), as an exemplar of where we want to go with HTTP naming style
- Similarly the transaction ID is given `view.seqno`. Although we currently use `term` and `commit` in response headers/other RPCs, we intend to rename soon.
- There are 4 states: `UNKNOWN`, `PENDING`, `COMMITTED`, and `INVALID`. Unknown and Pending can be usefully distinguished, and we can report Invalid in some cases I'd missed initially. There's a non-systematic unit test for how we derive this status.


achamayou:
I suggest we start with #968 with new endpoints, to minimise future disruption to users.

Not sure about LOST, I think ROLLED_BACK is more precise. I believe we'd agreed a `commit id` was composed of a `view` and `seqno`, which is PBFT terminology?

eddyashton:
> I suggest we start with #968 with new endpoints, to minimise future disruption to users.
>
> Not sure about LOST, I think ROLLED_BACK is more precise. I believe we'd agreed a `commit id` was composed of a `view` and `seqno`, which is PBFT terminology?

I was thinking ROLLED_BACK is too Raft-specific, its not really true for PBFT or a general consensus? We don't know if your transaction was partially accepted then rolled back (and the fact we do this in the kv seems like an implementation detail - in standard Raft terminology these 'local commits' are just stored in a pending lost, not applied then rolled back). The general externally visible behaviour is just "this didn't happen".

And we've settled on `view`/`seqno` within the code, but all of the external references are still to `term`/`commit`. I think we fix that with #968, in a single sweeping change. Unsure if its preferable for this RPC to be consistent with others, or in its final form already? Same point applies to removing the `get` prefix - we could make this inconsistent early, if we choose `txStatus` vs `tx_status` - but I think its easier to describe if it stays consistent?

achamayou:
@eddyashton there's a tradeoff between consistency and user-pain here. If we start making the new endpoints from this one onwards correct, users who rely on those won't have to update the code that deals with it. If we go for consistency, we're making more work for ourselves and for users, in return for slightly less surprise across RPCs.

I would prefer to go straight for the end state here, and to add an OpenAPI schema for this, but I'd happy to hear everyone's opinion about this too.

Lost to me sounds like Rolled back but worse, it's an acknowledgement it existed, but then it was misplaced. I do agree that rolled back isn't quite right either.

achamayou:
How about:

```
[ REPLICATING | UNKNOWN ] -> [ COMMITTED | WONT_COMMIT | UNKNOWN ]
```

eddyashton:
> How about:
> 
> ```
> [ REPLICATING | UNKNOWN ] -> [ COMMITTED | WONT_COMMIT | UNKNOWN ]
> ```

`UNKNOWN` is definitely growing on me. Don't think its safe/worth it to distinguish `REPLICATING` and `UNKNOWN`. `WONT_COMMIT` sounds a little too much like a _choice_. Maybe `UNKNOWN -> [ COMMITTED | NOT_COMMITTED ]`? (This may look like a bikeshed, but it houses the emergency generator)

sphinx/source/users/issue_commands.rst
```diff
@@ -52,29 +52,23 @@ Checking for Commit
...
 
-In this example, the ``result`` field indicates that the request was executed at ``23`` (``commit``), and in term ``2``, the same term that the ``LOG_record`` executed in. Moreover, the ``global_commit`` (``33``) is now greater than the ``commit`` version. The ``LOG_record`` request issued earlier was successfully committed to the ledger.
+Note that once a request has been assigned a version number, this version number will never be associated with a different transaction. In normal operation, the next requests will be given versions 2.19, then 2.20, and so on, and after a short delay 2.18 will be globally committed. If requests are submitted in parallel, they will be applied in a consistent order indicated by their assigned versions. If the network is unable to reach consensus, it will trigger a leadership election which increments the term. In this case the user's next request may be given a version 3.16, followed by 3.17, then 3.18. The index is reused, but in a different term; the service knows that 2.18 can never be assigned, so it has been lost.
```
jumaffre, line 71: "version **number**" is a little confusing here. Is `2.18` a number of a string? I would lean towards using "version" instead of "version number" wherever possible.

src/node/rpc/common_handler_registry.h
```diff
@@ -68,6 +61,67 @@ namespace ccf
...
+            // pending. This means they may be pending indefinitely
+            // (requesting 1.8, the service is stable with no new transactions
+            // at 2.5)!
+            out.status = GetTxStatus::TxStatus::Pending;
+          }
+          else if (actual_term < in.term)
+          {
+            // The requested transaction was committed in an _earlier_ view?
+            // Should be impossible for queries about real assigned versions,
+            // but essentially the same as the case below ("this version does
```
jumaffre, line 92: Minor: "case below" would no longer apply if the order of this if/else gets changed. Better to refer to actual case explicitly?

sphinx/source/users/issue_commands.rst
```diff
@@ -52,29 +52,23 @@ Checking for Commit
 
 Because of the decentralised nature of CCF, a request is committed to the ledger only once a number of nodes have agreed on that request.
 
-To guarantee that their request is successfully committed to the ledger, a user needs to issue a ``getCommit`` request, specifying the ``commit`` version received in the response. If CCF returns a ``global-commit`` greater than the ``commit`` version at which the ``LOG_record`` request was issued `and` that result ``commit`` is in the same ``term``, then the request was committed to the ledger.
+To guarantee that their request is successfully committed to the ledger, a user should issue a ``getTxStatus`` request, specifying the transaction version received in the response. The response may say the initial transaction is still pending global commit, has been globally committed, or has been lost due to a consensus leadership change (in which case the request should be resubmitted).
```
jumaffre, line 55: We should explain how users got the "transaction version" from the response, probably by giving an example.

jumaffre, line 55: Should we also replace most occurrences of "global commit" with "commit" on this page? With this new status RPC, a transaction is either `PENDING` or `COMMITTED`.

olgavrou:
`NOT_COMMITTED` may imply that the transaction is not committed yet, but it might get committed in the future. Another option is `REJECTED` which clearly states the transaction state.

sphinx/source/users/issue_commands.rst
```diff
@@ -52,29 +52,23 @@ Checking for Commit
...
```
olgavrou, line 56: nit: I think the phrase `has been lost due to a consensus leadership change` is not entirely correct. We don't loose transactions because the leadership changed, but the leadership changes as a result of not being able to process transactions.

sphinx/source/users/issue_commands.rst
eddyashton: I've used "transaction ID" as the general term we want to use now.

src/node/rpc/common_handler_registry.h
eddyashton: I've rewritten this logic, now in `tx_status.h`, hopefully with clearer decision path.

sphinx/source/users/issue_commands.rst
eddyashton: Agreed, please let me know if the current draft makes sense.

eddyashton: Yes we should, but I've probably left a few 'global's for emphasis in this version. Its tricky to be precise while we still have mixed terminology, let's aim to replace all `term.commit` and remove local commit, then we can cleanly reduce global commit to commit.

sphinx/source/users/issue_commands.rst
eddyashton: I've reworded this, let me know what you think now.

sphinx/source/users/issue_commands.rst
```diff
@@ -52,29 +52,41 @@ Checking for Commit
...
 
-    {
-      "commit": 23,
-      "term": 2
-    }
+    {"status":"COMMITTED"}
+
+This example queries the status of transaction ID 2.18 (constructed from view 2 and sequence number 18). The response indicates this was successfully committed. The headers also show that the service has since made progress with other requests, and global commit index has continued to increase.
+
+The possible statuses are:
```
olgavrou, line 72: I know we are not removing all the mentions to `globally committed` just yet, but I think that in the description of the statuses it is a bit confusing. I would say globally replicated, replicated to enough nodes to make the transaction durable, or something similar

sphinx/source/users/issue_commands.rst
```diff
@@ -52,29 +52,41 @@ Checking for Commit
...
+
+Note that transaction IDs are uniquely assigned by the service - once a request has been assigned an ID, this ID will never be associated with a different write transaction. In normal operation, the next requests will be given versions 2.19, then 2.20, and so on, and after a short delay 2.18 will be globally committed. If requests are submitted in parallel, they will be applied in a consistent order indicated by their assigned versions. If the network is unable to reach consensus, it will trigger a leadership election which increments the view. In this case the user's next request may be given a version 3.16, followed by 3.17, then 3.18. The sequence number is reused, but in a different view; the service knows that 2.18 can never be assigned, so it can report this as an invalid ID. Read-only transactions are an exception - they do not get a unique transaction ID but instead return the ID of the last write transaction whose state they may have read.
```
olgavrou, line 89: I'd suggest to remove `globally committed` from this paragraph too and just leave it as `committed`

Issue #459, opened 2019-10-17
Ledger Replay needs to take into account prepare and view change messages


olgavrou:
Right now `pre-prepares` and `prepare proof`s get written to the ledger. Ledger replay work so far has only been taking into account the `pre-prepares` with their requests. This needs to evolve to include `view changes` and `prepare proof`s.

olgavrou:
- [ ] validate prepare proofs during playback
- [ ] view change messages written to the ledger and processed during playback

olgavrou:
After discussion with @ashamis  we think that what needs to go into the ledger is the `New_view` message that will be altered to include the view change message signatures coming from the backups (new view messages are sent by the primary, view change messages are sent by all replicas).

The new view messages' view change proofs will need to be verified (a) by the other backups when they receive the new view and (b) by late joiners if they receive the new view message during playback.

PR #1116, opened 2020-04-29
Validate prepare proof

olgavrou:
#459 Resolves the `prepare proof` section of this issue.

Prepare proof is validated on playback
Added unit test


src/consensus/pbft/libbyz/certificate.h
```diff
@@ -374,6 +374,7 @@ bool Certificate<T>::add(T* msg)
     bmap.set(id);
     Message_val& val = vals[0];
     val.count++;
+    cur_size = 1;
```
olgavrou, line 377: `cur_size` holds the `current number of elements in vals`. Needed here so that if the `Val_iter` is called [here](https://github.com/microsoft/CCF/blob/master/src/consensus/pbft/libbyz/certificate.h#L284) then the iterator will return this message, since it _is_ stored in the certificate

Issue #512, opened 2019-11-04
Add sample JSON to operator RPC API


letmaik:
https://microsoft.github.io/CCF/operators/operator_rpc_api.html

This includes a weirdly formatted table of the schema, but mostly people just want to see an example JSON. For the schema itself, I would probably be happy with the JSON schema dumped/referenced somewhere.

achamayou:
What we have now is somewhat temporary until HTTP support gets better, at which point we'll use OpenAPI (https://swagger.io/docs/specification/about/)

achamayou:
#583

eddyashton:
Added these in #1119.

PR #1119, opened 2020-04-29
Operator API samples

eddyashton:
Adds samples of operator API responses, resolving #512.

Also changes the raw field in getQuote[s] to be hex-encoded.

I've changed the docs to display the schema as a raw file, rather than using the `jsonschema` extension. I think the table format doesn't actually make [this](https://microsoft.github.io/CCF/operators/operator_rpc_api.html) any more readable? And its more clear we have a json schema (that could be copied/pasted directly) if we just display the files? Open to discussion, should do either `literalinclude` or `jsonschema` for all these schemas. In any case we'll hopefully get an OpenAPI doc before 0.11 and we can display that instead.

### DIFF
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -976,7 +976,7 @@ namespace ccf
           {
             GetQuotes::Quote q;
             q.node_id = nid;
-            q.raw = ni.quote;
+            q.raw = fmt::format("{:02x}", fmt::join(ni.quote, ""));
 
 #ifdef GET_QUOTE
             if (this->network.consensus_type != ConsensusType::PBFT)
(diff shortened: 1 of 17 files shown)


src/node/node_state.h
```diff
@@ -976,7 +976,7 @@ namespace ccf
           {
             GetQuotes::Quote q;
             q.node_id = nid;
-            q.raw = ni.quote;
+            q.raw = fmt::format("{:02x}", fmt::join(ni.quote, ""));
```
jumaffre, line 979: Curious: what motivated hex over base 64 here?

eddyashton: Consistency with the nearby mrenclave which is already hex, and doing the minimal useful thing. base64 may be better (it has an official OpenAPI format, which hex doesn't), but requires a fatter parser.

PR #1124, opened 2020-04-30
Doc versioning

### LINKED ISSUE #770: Versioned documentation
We should make sure that our documentation is versioned so that it stays in sync with our releases as well as `master`. 

https://sphinxcontrib-versioning.readthedocs.io/en/latest/index.html seems to be the way to go. However, it does not seem to be actively maintained and I had to apply the patches described [here ](https://github.com/sphinx-contrib/sphinxcontrib-versioning/issues/70) to get it working locally (see for changes required to CCF: https://github.com/microsoft/CCF/compare/master...jumaffre:versioned_docs?expand=1). 

The (experimental) command I then ran was: `$ sphinx-versioning build -r versioned_docs -w versioned_docs -W sometag sphinx/source ./build/html`, which built the default docs (`versioned_docs`) and the one pointed by `sometag`.

```
$ ll ./build/html
_images        design         index.html         operators       sometag
_sources       developers     introduction.html  quickstart      users
_static        genindex.html  members            search.html     versioned_docs
concepts.html  glossary.html  objects.inv        searchindex.js
```

image

Let's keep an eye on the `sphinxcontrib-versioning` repo until these issues are fixed and we should hopefully be able to set this up for `v0.8`.

achamayou:
Fix #770 

- [x] Fix doxygen/breathe


sphinx/requirements.txt
```diff
@@ -5,3 +5,9 @@ pygments-style-solarized
 bottle
 sphinx-autobuild
 sphinxcontrib-mermaid
+<<<<<<< HEAD
+sphinx-jsonschema
+sphinx-multiversion
+pydata-sphinx-theme
+=======
+>>>>>>> master
```
eddyashton, line 13: This is unmerged.

achamayou: Now it's merged.

sphinx/source/conf.py
```diff
@@ -179,10 +180,44 @@
...
     "github_version": "master/sphinx/source/",
 }
+"""
+
+# Set up multiversion extension
+
+smv_tag_whitelist = r"^v.*$"
+smv_branch_whitelist = r"^master$"
+smv_remote_whitelist = None
+smv_released_pattern = r"^tags/.*$"
```
eddyashton, line 198: Is this the only thing choosing which versions we build? If so, we may want to make this a semver regex, so we don't render temporary/debugging tags.

achamayou: No, this is just some weird categorisation I don't think we need: https://holzhaus.github.io/sphinx-multiversion/master/templates.html#version-listings

What chooses what we build is the tag whitelist, and the branch whitelist. Then we can split them visually, but that didn't seem all that necessary. I guess I might as well remove this line to avoid confusion.

eddyashton: Ah right, cool!

achamayou:
@jumaffre yeah it's building a lot more stuff, so we have to decide how many versions we want to cover I guess. Right now it's anything released, but we could change that to the last X for example.

achamayou:
This looks super promising for many of our warnings: https://github.com/michaeljones/breathe/pull/512

achamayou:
This is unfortunately going to need to merge with a red Pages build, because there are literally no branches it can check out with the config in the right path...

achamayou:
Amusingly, our tests do in fact depend on the documentation (schema_test)!

Issue #1120, opened 2020-04-30
No Open Enclave version check in CCF releases


achamayou:
CCF binary releases do not clearly state which version of Open Enclave they require, nor enforce it. We ought to at the very least document it, and ideally automate it.

Since we require effectively a custom build of Open Enclave at the moment, which users typically produce with ansible playbooks, we should perhaps consider including that in the release and pointing to it in the cmake files we ship.

The release would be larger, but also easier to use.

achamayou:
Fixed by shipping OE build with CCF for now. Once we're packaged and we can use binary releases of OE, this is fixed by being explicit about dependency versions.

PR #1125, opened 2020-04-30
Copy OE into install dir

eddyashton:
2-part fix for #1120:
- When building install folder in ADO, include the current OE installed dir
- When trying to find the OE package in CMake, prefer a local instance if found

Sample release available here until I delete it: https://github.com/microsoft/CCF/releases/tag/untagged-bfa184e16a4d52090092

This only increases the size of our install package slightly - from 55.4MB in 0.9.3 to 63.5MB on current master with OE.

### DIFF
--- a/check-cmake-version-vs-tag.sh
+++ b/check-cmake-version-vs-tag.sh
@@ -4,11 +4,17 @@
 
 git_tag=$(git describe --tags --abbrev=0)
 git_tag=${git_tag#v}
-cmake_version=$(grep CCF_VERSION CMakeLists.txt | grep -Po "(\d)+(\.(\d)+)*")
-# Check git tag is <= CMake version. Use sort --version-sort to handle semver (0.9 < 0.10)
-if echo "$git_tag $cmake_version" | tr " " "\n" | sort --version-sort -c ; then
-  echo "Git tag ($git_tag) is safely <= CMake version ($cmake_version)"
+
+# Check if git tag looks like a semver value - ignore if not
+if [[ ${git_tag} =~ ^([[:digit:]])+(\.([[:digit:]])+)*(-.*)?$ ]]; then
+  cmake_version=$(grep CCF_VERSION CMakeLists.txt | grep -Po "(\d)+(\.(\d)+)*")
+  # Check git tag is <= CMake version. Use sort --version-sort to handle semver (0.9 < 0.10)
+  if echo "$git_tag $cmake_version" | tr " " "\n" | sort --version-sort -c ; then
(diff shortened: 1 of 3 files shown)


check-cmake-version-vs-tag.sh
```diff
@@ -4,11 +4,17 @@
...
+  if echo "$git_tag $cmake_version" | tr " " "\n" | sort --version-sort -c ; then
+    echo "Git tag ($git_tag) is safely <= CMake version ($cmake_version)"
+  else
+    echo "Git tag ($git_tag) is later than CMake version ($cmake_version) - please update CMake version!"
+    exit 1
+  fi
 else
-  echo "Git tag ($git_tag) is later than CMake version ($cmake_version) - please update CMake version!"
-  exit 1
+  echo "Skipping check - ${git_tag} doesn't look like semver"
```
olgavrou, line 19: semver? semantic versioning?

check-cmake-version-vs-tag.sh
```diff
@@ -4,11 +4,17 @@
...
-# Check git tag is <= CMake version. Use sort --version-sort to handle semver (0.9 < 0.10)
-if echo "$git_tag $cmake_version" | tr " " "\n" | sort --version-sort -c ; then
-  echo "Git tag ($git_tag) is safely <= CMake version ($cmake_version)"
+
+# Check if git tag looks like a semver value - ignore if not
+if [[ ${git_tag} =~ ^([[:digit:]])+(\.([[:digit:]])+)*(-.*)?$ ]]; then
+  cmake_version=$(grep CCF_VERSION CMakeLists.txt | grep -Po "(\d)+(\.(\d)+)*")
+  # Check git tag is <= CMake version. Use sort --version-sort to handle semver (0.9 < 0.10)
+  if echo "$git_tag $cmake_version" | tr " " "\n" | sort --version-sort -c ; then
+    echo "Git tag ($git_tag) is safely <= CMake version ($cmake_version)"
```
achamayou, line 13: Is < ever a good thing? It seems like we would want strict equality?

cmake/ccf_app.cmake
```diff
@@ -29,7 +29,13 @@ if((NOT ${IS_VALID_TARGET}))
   )
 endif()
 
-find_package(OpenEnclave 0.8 CONFIG REQUIRED)
+# Find OpenEnclave package, preferring local version if found (in the install
+# case)
+find_package(
+  OpenEnclave 0.9 CONFIG PATHS ${CMAKE_CURRENT_LIST_DIR}/../openenclave
```
achamayou, line 35: Hoist out version to a variable?

Issue #589, opened 2019-11-26
Raft persistence bug


achamayou:
Our raft implementation can lose transactions that have been advertised as committed.

1. Primary issues signature at i
2. Primary gather f + 1 acks for signature at i
3. Primary commits i, advertises all transactions up to i to callers as committed
4. Primary fails
5. Secondary gets elected, rolls back to last commit, earlier than i because primary did not have time to advertise i as committed
6. i is lost

This was spotted before as #519 / #520, but the fix is incorrect.

Instead of using commit_idx as their log_idx/value to roll back to, secondaries must use the last verified index (ie. the index of the last signature transaction they verified), which is the logical equivalent of log_idx with added verifiability.
Using commit_idx instead is unnecessarily strong and creates a potential loss window.

Fixing this will also mitigate #521, because the suffix committed on the primary, verified but not yet committed on the followers will be on a majority of followers and therefore part of the future ledger post-election.

achamayou:
Switching to last verified index isn't quite enough, because if the failure at 4. occurs after a majority of secondaries have acked the last verified index but before they have been notified that the commit index has moved to it, they will not be able to confirm to the user exactly what has been committed on failover, even though the transactions won't be lost.

Instead they will need to re-establish consensus over the verified suffix, which is effectively commited, since they cannot tell it is.

We therefore need to modify raft further to keep track of the secondaries' views on what the commit index is, and only confirm to the user that transactions are committed once a majority of followers has seen that commit index.

On election, voting must also happen by comparing the last commit index first, and the last verified index second.

achamayou:
This is also insufficient, because a node post-election would not be able to use its commit index as the new commit index (it could be further than the majority). I think an additional field and round trip is necessary.

eddyashton:
Quick notes from today's long discussions around this:

[...]

PR #1126, opened 2020-04-30
Update late joiners

### LINKED ISSUE #769: Refactor late_joiners.py test to be part of the existing end-to-end suite

olgavrou:
Resolves #769

Updated late joiners to use the `LoggingTxs` class from `logging_app.py` which has the positive side effect that we can remove of a lot of custom code that `late_joiners.py` test had.

Modified the `LoggingTxs` class to tolerate a suspended network.


tests/infra/logging_app.py
```diff
@@ -2,60 +2,117 @@
...
+        self.next_pub_index = tx_index_start
+        self.next_priv_index = tx_index_start
         self.notifications_queue = notifications_queue
+        self.can_fail = can_fail
+        self.timeout = timeout
+        self.wait_for_sync = wait_for_sync
+        self.tx_index_start = tx_index_start
 
     def issue(self, network, number_txs, consensus, on_backup=False):
         LOG.success(f"Applying {number_txs} logging txs")
```
jumaffre, line 31: This log message should be removed

tests/infra/logging_app.py
```diff
@@ -2,60 +2,117 @@
...
-                    check_commit_n(rep_priv, result=True)
-                    check_commit(rep_pub, result=True)
-
-                    self.priv[self.next_priv_index] = priv_msg
-                    self.pub[self.next_pub_index] = pub_msg
-                    self.next_priv_index += 1
-                    self.next_pub_index += 1
-
-        network.wait_for_node_commit_sync(consensus)
+                for _ in range(self.tx_index_start, self.tx_index_start + number_txs):
```
jumaffre, line 44: `for _ in range (number_txs)` still works here?

tests/infra/logging_app.py
```diff
@@ -2,60 +2,117 @@
...
         self.priv = {}
-        self.next_pub_index = 0
-        self.next_priv_index = 0
+        self.next_pub_index = tx_index_start
+        self.next_priv_index = tx_index_start
         self.notifications_queue = notifications_queue
+        self.can_fail = can_fail
+        self.timeout = timeout
+        self.wait_for_sync = wait_for_sync
+        self.tx_index_start = tx_index_start
```
jumaffre, line 28: I think this is not necessary.

tests/infra/logging_app.py
```diff
@@ -2,60 +2,117 @@
...
+                            )
+                            check_commit_n(rep_priv, result=True)
+                            check_commit(rep_pub, result=True)
+
+                            self.priv[self.next_priv_index] = priv_msg
+                            self.pub[self.next_pub_index] = pub_msg
+                            self.next_priv_index += 1
+                            self.next_pub_index += 1
+                            break
+                        except (TimeoutError, requests.exceptions.ReadTimeout,) as e:
```
jumaffre, line 68: `requests.exceptions.ReadTimeout` implies that this wouldn't work with the `CurlClient`?

tests/infra/logging_app.py
```diff
@@ -2,60 +2,117 @@
...
+        timeout=3,
+    ):
         self.pub = {}
         self.priv = {}
-        self.next_pub_index = 0
-        self.next_priv_index = 0
+        self.next_pub_index = tx_index_start
+        self.next_priv_index = tx_index_start
         self.notifications_queue = notifications_queue
+        self.can_fail = can_fail
```
jumaffre, line 25: Should the `can_fail` be passed to `issue()` instead of being an attribute of the whole class?

tests/infra/logging_app.py
```diff
@@ -2,60 +2,117 @@
...
+                            check_commit_n(rep_priv, result=True)
+                            check_commit(rep_pub, result=True)
+
+                            self.priv[self.next_priv_index] = priv_msg
+                            self.pub[self.next_pub_index] = pub_msg
+                            self.next_priv_index += 1
+                            self.next_pub_index += 1
+                            break
+                        except (TimeoutError, requests.exceptions.ReadTimeout,) as e:
+                            LOG.info("Network is unavailable")
```
jumaffre, line 69: `LOG.warning`?

tests/late_joiners.py
```diff
@@ -36,117 +35,92 @@ def timeout_handler(node, suspend, election_timeout):
...
-    return term_info
 
 
-def get_node_local_commit(node):
-    with node.node_client() as c:
-        r = c.get("getLocalCommit")
-        return r.commit, r.global_commit
+@reqs.description("Running transactions against logging app")
+@reqs.supports_methods("LOG_record", "LOG_record_pub", "LOG_get", "LOG_get_pub")
+@reqs.at_least_n_nodes(3)
```
jumaffre, line 48: I think this test would work with any number of nodes? Also, since this test is generic, perhaps it could be moved to `logging_app.py`?

tests/late_joiners.py
```diff
@@ -36,117 +35,92 @@ def timeout_handler(node, suspend, election_timeout):
...
+@reqs.description("Suspend nodes")
+@reqs.at_least_n_nodes(3)
+def test_suspend_nodes(network, args, nodes_to_keep):
+    cur_primary, _ = network.find_primary()
+
+    # first timer determines after how many seconds each node will be suspended
+    timeouts = []
+    for i, node in enumerate(nodes_to_keep):
+        # if pbft suspend half of them including the primary
+        if i % 2 != 0 and args.consensus == "pbft":
```
jumaffre, line 104: I guess the reason this doesn't work with Raft is https://github.com/microsoft/CCF/issues/589?

tests/late_joiners.py
```diff
@@ -36,117 +35,92 @@ def timeout_handler(node, suspend, election_timeout):
...
+
+@reqs.description("Adding a very late joiner")
+def test_add_late_joiner(network, args, nodes_to_keep):
+    new_node = network.create_and_trust_node(args.package, "localhost", args)
+    nodes_to_keep.append(new_node)
+    return network
+
+
+@reqs.description("Suspend nodes")
+@reqs.at_least_n_nodes(3)
```
jumaffre, line 96: Same here?

tests/late_joiners.py
```diff
@@ -156,111 +130,79 @@ def run(args):
...
+        update_term_info(network, term_info)
+
+        test_run_txs(network=network, args=args, num_txs=TOTAL_REQUESTS)
+        update_term_info(network, term_info)
+
+        nodes_to_kill = [network.find_any_backup()]
+        nodes_to_keep = [n for n in all_nodes if n not in nodes_to_kill]
+
+        # check that a new node can catch up after all the requests
+        test_add_late_joiner(network, args, nodes_to_keep)
```
jumaffre, line 144: I believe that we should prefer `network.create_and_trust_node()` here and remove `test_add_late_joiner`. `nodes_to_keep` can then be set to the difference between `network.get_joined_nodes` and `nodes_to_kill`.

tests/late_joiners.py
```diff
@@ -36,117 +35,92 @@ def timeout_handler(node, suspend, election_timeout):
...
 
-def find_primary(network):
-    # track if there is a new primary
-    term_info = {}
+def update_term_info(network, term_info):
     try:
         cur_primary, cur_term = network.find_primary()
         term_info[cur_term] = cur_primary.node_id
     except TimeoutError:
         LOG.info("Trying to access a suspended network")
```
jumaffre, line 43: `LOG.warning`?

tests/late_joiners.py
```diff
@@ -156,111 +130,79 @@ def run(args):
...
+        # check that a new node can catch up after all the requests
+        test_add_late_joiner(network, args, nodes_to_keep)
+
+        # some requests to be processed while the late joiner catches up
+        # (no strict checking that these requests are actually being processed simultaneously with the node catchup)
+        test_run_txs(
+            network=network,
+            args=args,
+            num_txs=int(TOTAL_REQUESTS / 2),
+            start_idx=1000,
```
jumaffre, line 152: `start_idx` is not required here. If you keep the same `LoggingTxs` object throughout the whole test, the indices will never conflict.

tests/late_joiners.py
```diff
@@ -156,111 +130,79 @@ def run(args):
...
+        # kill the old node(s) and ensure we are still making progress with the new one(s)
+        for backup_to_retire in nodes_to_kill:
+            LOG.success(f"Stopping node {backup_to_retire.node_id}")
+            backup_to_retire.stop()
+
+        # check nodes are ok after we killed one off
+        test_run_txs(
+            network=network,
+            args=args,
+            nodes=nodes_to_keep,
```
jumaffre, line 167: The logging app should know not to send transactions to a node that is dead?

tests/late_joiners.py
```diff
@@ -156,111 +130,79 @@ def run(args):
...
+            num_txs=len(nodes_to_keep),
```
jumaffre, line 168: Not clear why sending as many transactions as there are nodes?

tests/late_joiners.py
```diff
@@ -156,111 +130,79 @@ def run(args):
...
+            start_idx=4000,
+            can_fail=True,
+        )
+
+        update_term_info(network, term_info)
 
-    def add(parser):
-        parser.add_argument(
-            "--skip-suspension",
-            help="Don't suspend any nodes (i.e. just do late join)",
```
jumaffre, line 259: Should the name of this test be changed now that this test is more about node suspension than late joiner?

tests/late_joiners.py
```diff
@@ -36,117 +35,92 @@ def timeout_handler(node, suspend, election_timeout):
...
+@reqs.description("Adding a very late joiner")
+def test_add_late_joiner(network, args, nodes_to_keep):
+    new_node = network.create_and_trust_node(args.package, "localhost", args)
+    nodes_to_keep.append(new_node)
+    return network
+
+
+@reqs.description("Suspend nodes")
+@reqs.at_least_n_nodes(3)
+def test_suspend_nodes(network, args, nodes_to_keep):
```
jumaffre, line 97: It would be awesome to add this to our existing test suite. We could check for elections without having to kill one node (i.e. just suspend the primary for long enough).

tests/late_joiners.py
olgavrou: yes

tests/late_joiners.py
olgavrou: If it is retired and not killed then yes? I think if you just kill a node without retiring it the network won't know about it

tests/late_joiners.py
olgavrou: one for each node

tests/infra/logging_app.py
```diff
@@ -2,60 +2,117 @@
...
+    def node_commit_sync(self, network, consensus):
+        end_time = time.time() + self.timeout
+        while time.time() < end_time:
+            try:
+                network.wait_for_node_commit_sync(consensus)
+                break
+            except (TimeoutError, infra.clients.CCFConnectionException) as e:
+                LOG.error("Timeout error while waiting for nodes to sync")
+                if not self.can_fail:
+                    raise RuntimeError(e)
```
achamayou, line 85: Just raise will re-raise, no need to wrap this

olgavrou:
parking this until the CI is stable again

tests/suspend_nodes.py
```diff
@@ -0,0 +1,195 @@
...
+
+from loguru import logger as LOG
+
+# pbft will store up to 34 of each message type (pre-prepare/prepare/commit) and retransmit these messages to replicas that are behind, enabling catch up.
+# If a replica is too far behind then we need to send entries from the ledger, which is one of the things we want to test here.
+# By sending 18 RPC requests and a getCommit for each of them (what raft consideres as a read pbft will process as a write),
+# we are sure that we will have to go via the ledger to help late joiners catch up (total 36 reqs > 34)
+TOTAL_REQUESTS = 9  # x2 is 18 since LoggingTxs app sends a private and a public request for each tx index
+
+s = random.randint(1, 10)
```
achamayou, line 22: I would be a tiny bit nicer to make the seed an extra CLI arg, and set the default to the random seed.

tests/infra/logging_app.py
```diff
@@ -2,61 +2,173 @@
...
+):
+    if nodes is None:
+        nodes = network.get_joined_nodes()
+    num_nodes = len(nodes)
+
+    for tx in range(num_txs):
+        logging_app.issue_on_node(
+            network=network,
+            remote_node=nodes[tx % num_nodes],
+            number_txs=1,
```
jumaffre, line 32: `number_txs=num_txs`?

tests/infra/logging_app.py
```diff
@@ -2,61 +2,173 @@
...
 
 from loguru import logger as LOG
 
 
+@reqs.description("Running transactions against logging app")
+@reqs.supports_methods("LOG_record", "LOG_record_pub", "LOG_get", "LOG_get_pub")
+def test_run_txs(
+    network,
+    args,
+    logging_app,
```
jumaffre, line 16: I believe that this won't work when plugged in the test suite. All arguments except `network` and `args` should be defaulted. We've come across this problem before and the solution is that you can attach a `txs` object to the network (see example in `recovery.py`).

tests/suspend_nodes.py
```diff
@@ -0,0 +1,195 @@
...
+            nodes=original_nodes,  # doesn't contain late joiner
+            verify=False,  # will try to verify for late joiner and it might not be ready yet
+            logging_app=txs,
+        )
+
+        wait_for_late_joiner(original_nodes[0], late_joiner)
+
+        # kill the old node(s) and ensure we are still making progress
+        for backup_to_retire in nodes_to_kill:
+            LOG.success(f"Stopping node {backup_to_retire.node_id}")
```
jumaffre, line 144: Minor: I don't think we need this log as it gets printed by the node on shutdown anyway?

tests/infra/logging_app.py
olgavrou: this sends `num_txs` transactions in total, it just sends 1 to each node in a round robin fashion

tests/suspend_nodes.py
olgavrou: The hostname is printed by the node shutdown, it's just a bit easier to see in the logs this way which node was actually stopped

tests/infra/logging_app.py
olgavrou: thanks, done

tests/suspend_nodes.py
olgavrou: done

PR #1130, opened 2020-05-02
Update devcontainer.json to use oe-0.9.0 container

sangeethkumarp:
Making this change as OpenEnclave 0.9 release is being installed in new setups due to #1103


achamayou:
Hi @sangeethkumarp, thank you for updating the dev container config!

PR #1131, opened 2020-05-02
Switching to CMAKE_CURRENT_SOURCE_DIR in pbft.cmake

sangeethkumarp:
Replacing all `CMAKE_SOURCE_DIR` to `CMAKE_CURRENT_SOURCE_DIR` in `pbft.cmake` as cmake breaks on using CCF as a subfolder in other repos.

### DIFF
--- a/cmake/pbft.cmake
+++ b/cmake/pbft.cmake
@@ -10,45 +10,45 @@ if(SAN)
 endif()
 
 set(PBFT_SRC
-    ${CMAKE_SOURCE_DIR}/src/consensus/pbft/libbyz/global_state.cpp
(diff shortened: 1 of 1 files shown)


achamayou:
@sangeethkumarp thank you for fixing that, I would suggest you use the pre-built release tarballs rather than build CCF as a submodule/subdirectory if possible.

They can be downloaded from the release page: https://github.com/microsoft/CCF/releases, or from any "Daily" build run if you need the very latest, for example https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=7887&view=artifacts&type=publishedArtifacts

PR #1132, opened 2020-05-04
Add wrapper for cmake format

achamayou:
Checking and formatting cmake is annoyingly manual at the moment, this is a wrapper that implements the same interface as check-format.sh and make it easier to check/format in a single command.

I wonder if we shouldn't just have a single entry point for all languages, but I'll leave that for another time.


.azure-pipelines-templates/checks.yml
```diff
@@ -40,10 +40,7 @@ jobs:
       displayName: 'Check Python code format'
       condition: succeededOrFailed()
 
-    - script: |
-        source env/bin/activate
-        pip install cmake_format
-        cmake-format --check CMakeLists.txt $(find ./cmake ./samples ./src ./tests -name "*.cmake" | tr "\n" " ")
+    - script: ./check-format.sh cmake samples src tests CMakeLists
```
jumaffre, line 43: `check-cmake-format.sh`?

achamayou: Indeed!

PR #1133, opened 2020-05-04
Update README.md

jumaffre:
I noticed the image got broken since the versioned docs.


jumaffre:
Merging this as this is `README.md` only and the end-to-end tests issues have been identified before and are definitely not related to this.

PR #1136, opened 2020-05-05
Remove incorrect error on tx status RPC

eddyashton:
The assumption in the removed tests is incorrect: after we win an election, there is a period where we can serve RPCs (producing valid new `Pending` Tx IDs) but the highest committed Tx ID is still in the previous view. So this case shouldn't throw an error, it should fall through to `views_match => Pending`.

This RPC should also not be `execute_locally` (this was my misunderstanding - the common case of `getCommit` still requires consensus, so `tx` does as well).

These should resolve some of our transient CI failures.


eddyashton:
Currently at 5/5 green builds. Running a few more, and marking as ready for review.

Issue #1042, opened 2020-04-07
Multi-threaded CCF nodes kill the SGX driver


eddyashton:
Opening this thread to track information on this bug.

Occasionally, when running CCF tests with multi-threaded nodes, we put the SGX driver in an unrecoverable state. The current identifying diagnostic is this trace from dmesg:

```
[ 9950.700321] BUG: unable to handle kernel paging request at 00000008ca5ae000
[ 9950.702518] #PF error: [normal kernel read fault]
[ 9950.703988] PGD 800000084d8c1067 P4D 800000084d8c1067 PUD 0 
[ 9950.704311] Oops: 0000 [#1] SMP PTI
[ 9950.704311] CPU: 6 PID: 29464 Comm: cchost Tainted: G           OE     5.0.0-1035-azure #37-Ubuntu
[ 9950.704311] Hardware name: Microsoft Corporation Virtual Machine/Virtual Machine, BIOS Hyper-V UEFI Release v4.0 03/12/2019
[ 9950.711609] RIP: 0010:sgx_eldu+0xc2/0x420 [intel_sgx]
[ 9950.711609] Code: 48 c7 85 68 ff ff ff 00 00 00 00 0f 84 2e 02 00 00 48 8b 7d 80 48 89 95 70 ff ff ff e8 97 fe ff ff 48 89 45 80 49 8b 44 24 08 <48> 8b 38 e8 86 fe ff ff 49 89 c2 65 48 8b 0c 25 40 4d 01 00 83 81
[ 9950.711609] RSP: 0000:ffffb3004c4c3c60 EFLAGS: 00010286
[ 9950.711609] RAX: 00000008ca5ae000 RBX: 0000000000000580 RCX: 0000000000000000
[ 9950.711609] RDX: 0000000000000000 RSI: 0000000000000000 RDI: 00000008c2761000
[ 9950.711609] RBP: ffffb3004c4c3d10 R08: ffffb3004c4c3af0 R09: 0000000000000000
[ 9950.711609] R10: ffffb3004c4c3d28 R11: 00000008c2761000 R12: ffff928b0e6535c0
[ 9950.711609] R13: ffff928b0113fa00 R14: fffff065604a0300 R15: fffff0656032fec0
[ 9950.711609] FS:  00007f1b7cdfc700(0000) GS:ffff928b5fb80000(0000) knlGS:0000000000000000
[ 9950.711609] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[...]

PR #1137, opened 2020-05-05
Manually install 1.33 driver

achamayou:
Related to #1042


getting_started/setup_vm/roles/intel/tasks/sgx-driver.yml
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) Open Enclave SDK contributors.
```
eddyashton, line 1: Should we remove this? Or add a CCF copyright header to all of our ansible `.yml`s?

achamayou: I don't think we should remove it, that's where we got it from.

My view is that our ansible files are ancillary configuration, not code, and so they do not need copyright headers.

Issue #1139, opened 2020-05-05
UBSan issue in libbyz/digest.h


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=7987&view=ms.vss-test-web.build-test-results-tab&runId=42056&resultId=100000&paneView=debug

```
2020-05-05T12:54:59.700433Z        100 [info ] /consensus/pbft/libbyz/digest.cpp:67 | digest=[10312665180570818544,7596743427861467266,8316672188112260173,12166594345395003768]
../src/consensus/pbft/libbyz/digest.h:62:18: runtime error: member call on misaligned address 0x612000000c7c for type 'Digest *', which requires 8 byte alignment
0x612000000c7c: note: pointer points here
  00 00 00 00 be be be be  be be be be be be be be  be be be be be be be be  be be be be be be be be
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/consensus/pbft/libbyz/digest.h:62:18 in 
```

PR #1142, opened 2020-05-05
Objects must be 8 byte aligned

achamayou:
fix for #1139


eddyashton:
What is this fixing? There's a `pragma pack(1)` on this struct, should it be `pragma pack(8)` (or removed entirely)? Every earlier field in this struct is already 8-bytes or a multiple-of, so what does this change?

achamayou:
@eddyashton yes, hashed_nonce was the source of the alignment problem.

I agree the layout should probably be cleaned up somehow, but this is a targeted fix, not a drive-by refactoring attempt :)

PR #1146, opened 2020-05-06
Steps towards an all-inclusive suite test

eddyashton:
This adds a list of all tests we currently have which could be run in our e2e suite, and a (currently disabled) ctest which runs these. There's a few outstanding issues in some shuffles so I haven't enabled the test yet. The existing suite test remains, and has been renamed to `recovery_test_suite` to more accurately describe what it is testing.

This includes a new `test_view_history` which has revealed the term history is not fully reconstructed after a recovery - transaction 1's view is unknown (https://github.com/microsoft/CCF/issues/1150).


CMakeLists.txt
```diff
@@ -460,13 +460,47 @@ if(BUILD_TESTS)
   )
 
   add_e2e_test(
-    NAME test_suite
+    NAME rekey_tests
```
jumaffre, line 463: rekey_test**~s~**?

tests/rekey.py
```diff
@@ -14,31 +15,17 @@ def test(network, args):
...
     with infra.ccf.network(
         hosts, args.binary_dir, args.debug_nodes, args.perf_nodes, pdb=args.pdb,
     ) as network:
         network.start_and_join(args)
-        primary, _ = network.find_primary()
 
-        record_transactions(primary)
-        test(network, args)
-        record_transactions(primary)
+        network = e2e_logging.test(network, args)
```
jumaffre, line 26: I wonder whether test "blocks" should include other test blocks or whether only the test suite should be allowed to include blocks. I'm leaning towards the latter and I think the `logging_app` should be used here instead?

CMakeLists.txt
eddyashton: Sure, seems we've been a bit inconsistent on this. Are we happy with `recovery_test`, `rekey_test`, `governance_test`, `schema_test_...`, `election_test_...`, or does anyone think these are a deliberate plurality?

tests/rekey.py
eddyashton: Yeah I think you're right. I thought this one was fine since its in the `__main__`/`run()` of this test, but it should still be using `logging_app` rather than depending on another suite-block.

tests/e2e_suite.py
```diff
@@ -139,6 +150,13 @@ def add(parser):
         parser.add_argument(
             "--test-duration", help="Duration of full suite (s)", type=int
         )
+        parser.add_argument(
```
jumaffre, line 153: As we are extending the suite, I wonder if we should create a new class that records the different sub-suites. Something like:
```
s = Suites()
s.add_suite(suite_rekey_recovery, "description")
s.add_suite(suite_membership_recovery, "description")
```

`choices` here would be the list of all the suites that have been registered. And we could filter based on the selected `--test-suite`. Then we wouldn't have to add a new `elif` at the top of this file every time a new suite is added.

CMakeLists.txt
```diff
@@ -460,13 +460,47 @@ if(BUILD_TESTS)
...
     LABEL suite
-    ADDITIONAL_ARGS --test-duration 150 --enforce-reqs
+    ADDITIONAL_ARGS
+      --test-duration
+      150
+      --enforce-reqs
+      --test-suite
+      rekey_recovery
+      --test-suite
+      rekey_recovery
```
jumaffre, line 480: Should this be `membership_recovery`?

eddyashton: Yes it should, good catch.

tests/suite/test_suite.py
```diff
@@ -32,10 +34,36 @@
     recovery.test,
 ]
 
-
-full_suite = []
-full_suite.extend(suite_rekey_recovery)
-full_suite.extend(suite_membership_recovery)
+all_tests_suite = [
```
jumaffre, line 37: What about we use `dir()` on all the local imports and grep for all the methods that start with `test_`? We could even add a `@skip` decorator if we don't want a certain test not to be included.

tests/e2e_suite.py
eddyashton: We could do something like this, though we still need a name mapping (ie `s.add_suite("rekey_recovery", suite_rekey_recovery, ...)` that can be entered from the command line. I agree this should be ideally be in `test_suite.py` rather than `e2e_suite.py`, but I think it can just be a dict rather than a class?

CMakeLists.txt
achamayou: I'm not sure the _test bit is even useful?

tests/suite/test_suite.py
eddyashton: We could do this - we could even make our `@reqs.description` decorator an opt-in, rather than adding an `@skip` opt-out.

However I'm not sure its worth it. I think its valuable to have an explicit list of all the tests we think we're running. Sure we can remove some user error in failing to register new tests here, but on the other hand modifying the list (eg - for debugging I want a suite of `[recovery, recovery, recovery, view_history]`) becomes more difficult as the available/present test names aren't visible.

jumaffre: Running the test suite with a new `--list-tests` could list all the available tests and we could even pass the name of test cases from the command line, e.g. `--test-cases recovery,recovery,recovery,view_history` to run specific tests. Running the test suite with no `--test-cases` or `--test-suites` filters would mean running all the `test_` included from `test_suite`. 

We could still define custom sub-suites in code though.

jumaffre: ^ Not in the scope of this PR. Just discussing what I'd been thinking about the evolution of our test suite.

tests/e2e_logging.py
```diff
@@ -264,6 +265,53 @@ def test_update_lua(network, args):
...
+    with primary.user_client() as c:
+        r = c.get("getCommit")
+        check(c)
+
+        commit_view = r.term
+        commit_seqno = r.global_commit
+
+        # Retrieve status for all possible Tx IDs
+        seqno_to_views = dict()
+        end_seqno = commit_seqno + 1
```
jumaffre, line 283: Minor: no need for `end_seqno` temporary variable?

tests/e2e_logging.py
```diff
@@ -264,6 +265,53 @@ def test_update_lua(network, args):
     return network
 
 
+@reqs.description("Check for commit of every prior transaction")
+@reqs.supports_methods("getCommit", "tx")
+def test_view_history(network, args):
```
jumaffre, line 270: Should this be run against all nodes of the network?

tests/e2e_logging.py
```diff
@@ -264,6 +265,53 @@ def test_update_lua(network, args):
...
+
+    with primary.user_client() as c:
+        r = c.get("getCommit")
+        check(c)
+
+        commit_view = r.term
+        commit_seqno = r.global_commit
+
+        # Retrieve status for all possible Tx IDs
+        seqno_to_views = dict()
```
achamayou, line 282:
```suggestion
        seqno_to_views = {}
```

tests/suite/test_suite.py
achamayou: Pretty soon that's going to be reimplementing all of the pytest runner :)

CMakeLists.txt
eddyashton: I've turned `_tests` into `_test` for now. I'm also unconvinced by the `_test` suffix the more I think about it. Maybe its useful to distinguish labels, build targets, other concepts, from ctest tests? Let's rename broadly in a focussed PR if we're going to.

tests/suite/test_suite.py
eddyashton: Yeah you're right - with inspected suite names and test case names, we can build precise tests on the command-line. That's cool, but going to leave it for future PRs.

tests/e2e_logging.py
eddyashton: Good idea, it is now.

PR #1148, opened 2020-05-06
Update fmtlib

achamayou:


src/node/channels.h
```diff
@@ -102,7 +102,7 @@ namespace ccf
         LOG_FAIL_FMT(
           "Invalid nonce, possible replay attack, received:{}, last_seen:{}, "
           "recv_nonce.tid:{}",
-          recv_nonce.nonce,
+          reinterpret_cast<uint64_t>(recv_nonce.nonce),
```
eddyashton, line 105: Why is this needed?

achamayou: Because taking non-const refs to bitfields is illegal. I'm guessing that fmtlib's type resolution has changed in a way that's now stricter, and this was presumably previously coerced somehow.

eddyashton: Ah cool - I'd missed that this was a bitfield.

PR #1153, opened 2020-05-06
Update governance req tracker during view changes

### LINKED ISSUE #1152: election test pbft is failing
Election test for pbft is failing on the CI with merkle roots unmatched or with latest governance seqno mismatch between replicas.

What I have seen is:

- node 3 is left behind (seqno 23 executed last)
- node 0, 1, 2 make progress until seqno 30
- node 0 dies before it manages to retransmit all of the requests that node 3 needs so node 3 can't process pre-prepares 24 through 30
- view change rota starts with the first successful view change happening when node 3 becomes primary because node 3 can't process everyone else's view change messages but everyone else can process 3's
- merkle roots match, the thing that is off is the governance seqno

I am able to reproduce this reliably and I suspect the issue is with how we update the governance seqno.

olgavrou:
Resolves #1152 

Update gov req tracker during view changes and a bit of refactoring


src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -2050,37 +2047,35 @@ void Replica::process_new_view(Seqno min, Digest d, Seqno max, Seqno ms)
...
-        }
-      }
-
+      did_execute = execute_tentative(pp, info, nonce);
       Prepare* p = new Prepare(v, i, d, nonce, nullptr, pp->is_signed());
       pc.add_mine(p);
       send(p, All_replicas);
     }
 
+    if (did_execute && ledger_writer)
```
achamayou, line 2068: I think this would be clearer with a top level check on did_execute and two checks for ledger_writer and req_in_pp nested.

olgavrou: true, thanks

Issue #1127, opened 2020-05-01
Late Joiners PBFT Test Failures


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=7906&view=ms.vss-test-web.build-test-results-tab&runId=41744&resultId=100028&paneView=attachments

```
Traceback (most recent call last):
  File "/mnt/build/1/s/tests/late_joiners.py", line 270, in <module>
    run(args)
  File "/mnt/build/1/s/tests/late_joiners.py", line 238, in run
    cant_fail=False,
  File "/mnt/build/1/s/tests/late_joiners.py", line 141, in run_requests
    result=True,
  File "/mnt/build/1/s/tests/infra/checker.py", line 74, in __call__
    self.node_client, rpc_result.commit, rpc_result.term
  File "/mnt/build/1/s/tests/infra/checker.py", line 39, in wait_for_global_commit
    f"Transaction ID {term}.{commit_index} is marked invalid and will never be committed"
RuntimeError: Transaction ID 3.176 is marked invalid and will never be committed
```

achamayou:
Another instance this morning: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=7921&view=ms.vss-test-web.build-test-results-tab

achamayou:
Fixed by #1136

achamayou:
Happened again, after #1136 

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8005&view=ms.vss-test-web.build-test-results-tab&runId=42108&resultId=100028&paneView=attachments

```
Traceback (most recent call last):
  File "/mnt/build/1/s/tests/late_joiners.py", line 270, in <module>
    run(args)
  File "/mnt/build/1/s/tests/late_joiners.py", line 238, in run
    cant_fail=False,
  File "/mnt/build/1/s/tests/late_joiners.py", line 141, in run_requests
    result=True,
  File "/mnt/build/1/s/tests/infra/checker.py", line 74, in __call__
    self.node_client, rpc_result.commit, rpc_result.term
  File "/mnt/build/1/s/tests/infra/checker.py", line 39, in wait_for_global_commit
    f"Transaction ID {term}.{commit_index} is marked invalid and will never be committed"
RuntimeError: Transaction ID 3.254 is marked invalid and will never be committed
```

achamayou:
[...]

PR #1157, opened 2020-05-07
Return the view that's relevant to the transaction, not the latest view!

achamayou:


eddyashton:
In case anyone else is staring at this and pondering the implications: I initially thought this could be replaced with an error - if these views don't match then there's been a view change during execution and the transaction is about to be rolled back. But I believe its actually possible for a transaction to complete, be replicated, signed and globally committed, and then for a legitimate view change to occur, before the executing node reaches this point in constructing the response (the executing thread _could_ be stalled arbitrarily long, anywhere). This change is necessary in that situation, so the response headers correctly identify the executed transaction.

achamayou:
@eddyashton yes, exactly, that'd have to happen pretty fast, but it could.

So this breaks some tests that incorrectly interpret the transaction term as being the current term when calling getPrimaryInfo (more often than not, around election time, it's not!). So I'm adding the current_term (as known by the node) to fix that.

Next up, the purge of global_commit from end to end tests and Python infra.

achamayou:
Also to come, The Great Big Renaming of the HTTP headers, variable names and everything.

achamayou:
@olgavrou interestingly, this fix seems to make #1127 happen reliably.

olgavrou:
@achamayou view changes seem to be happening fine and there is no merkle root (or other) mismatch, so I am guessing it has to do with the commit/view assumptions

Issue #1134, opened 2020-05-04
Raft election test failure


olgavrou:
here: 

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=7924&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=efdbc34d-c6b6-5dc6-e05c-1751d4c47c20

and here:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=7918&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=efdbc34d-c6b6-5dc6-e05c-1751d4c47c20

olgavrou:
> Waiting for transaction to be committed by all nodes
> 50: 2020-05-04 07:27:59.964 | INFO     | infra.clients:log_request:147 - [127.78.54.96:45609] GET /nodes/tx {'view': 4, 'seqno': 20} (node 1 (node))
**> 50: 2020-05-04 07:27:59.975 | DEBUG    | infra.clients:log_response:160 - {'status': 500, 'result': None, 'error': 'Should not believe 4 occurred in view 4, ahead of the current committed view 2', 'commit': None, 'term': None, 'global_commit': None}**
> 50: 2020-05-04 07:27:59.975 | INFO     | infra.remote:stop:455 - [127.78.54.96] closing
> 50: 2020-05-04 07:28:00.007 | INFO     | infra.remote:stop:455 - [127.83.120.69] closing
> 50: 2020-05-04 07:28:00.071 | INFO     | infra.ccf:stop_all_nodes:294 - All remotes stopped...
> 50: Traceback (most recent call last):
> 50:   File "/mnt/build/2/s/tests/election.py", line 133, in <module>
> 50:     run(args)
> 50:   File "/mnt/build/2/s/tests/election.py", line 101, in run
> 50:     commit_index, current_term, network.get_joined_nodes()
> 50:   File "/mnt/build/2/s/tests/election.py", line 30, in **wait_for_index_globally_committed**
> 50:     ), f"tx request returned HTTP status {r.status}"
> 50: AssertionError: tx request returned HTTP status 500
> 41/51 Test #50: election_tests_raft ..............***Failed   16.86 sec
>

olgavrou:
[...]

PR #1158, opened 2020-05-07
Fix get commit test

jumaffre:
On startup, when checking that all nodes have caught up with the leader before issuing further transactions, we incorrectly used the primary's current seqNo **and current view** instead of the primary's current seqNo **and the view at that seqNo.**

For example, assume that the the last global commit was `3` in term `2` and that there was an election during the startup of the nodes (e.g. slow quote verification). After the election, the global commit was still at `3` but the term was now `3`. This fix makes sure that we use the term `2` and not `3`.

Note: https://github.com/microsoft/CCF/pull/1157 fixes the same issue but the test should use the result from `getCommit`.


jumaffre:
I believe this fixes https://github.com/microsoft/CCF/issues/1134#issuecomment-624151415

tests/infra/ccf.py
```diff
@@ -462,22 +462,27 @@ def find_primary_and_any_backup(self, timeout=3):
...
         Wait for all nodes to have joined the network and globally replicated
-        all transactions executed on the primary (including the transactions
+        all transactions globally executed on the primary (including transactions
         which added the nodes).
         """
-        with primary.node_client() as c:
-            res = c.get("getCommit")
-            local_commit_leader = res.commit
-            term_leader = res.term
+        for _ in range(timeout):
```
eddyashton, line 468: We've swapped out these `for range(timeout)` loops with `while`s so that the timeouts specify time rather than iterations (if `getCommit` takes 5 seconds, this loop may take ~18 seconds, which seems unrelated to the timeout of 3). Can we do the same here? The total timeout should probably be shared with the loop below (ie - the `end_time` is calculated once at the start of the function and used by both loops)

tests/infra/ccf.py
```diff
@@ -462,22 +462,27 @@ def find_primary_and_any_backup(self, timeout=3):
     def wait_for_all_nodes_to_catch_up(self, primary, timeout=3):
         """
         Wait for all nodes to have joined the network and globally replicated
-        all transactions executed on the primary (including the transactions
+        all transactions globally executed on the primary (including transactions
```
eddyashton, line 465: I don't think this comment is correct. The body of `getCommit` returns `tables->commit_version`, which is the node's local commit version (the _global_ commit is retrieved through `consensus->get_commit_seqno`). I think the logic is still fine (the nodes can 'catch up' with a target index whether its a global commit point or not), but the comment is misleading.

I think this is a strong argument for renaming some of the kv concepts - we already have duplicated consensus terms, but if we're talking about transactions being _committed_ at a _seqno_ in the _consensus_, we shouldn't mistake this with the _version_ where they're _applied_ to the _kv_.

jumaffre: `tables->commit_version()` returns the `compacted` version which is the global commit version? I got caught by this too originally, hence the retry logic around this as the start node may not have anything globally committed when it first starts up.

eddyashton: Ah ha, you're absolutely right (I was looking at `TxView::commit_version()`, but we're calling `Store::commit_version()`). So this _is_ a global commit point.

PR #1159, opened 2020-05-07
Fix initial term history during recovery

### LINKED ISSUE #1150: Commit status of first transaction is lost after recovery
After recovery, if we call `/tx?view=2&seqno=1`, we will get an `INVALID` response. Before recovery this transaction is correctly recorded as `COMMITTED`. The issue is presumably with the reconstructed `term_history` in Raft. I believe this is only ever happening with the first (`create`) transaction, which is also special in other ways.

eddyashton:
Fixes #1150.

- Added unit tests of raft's term history, so we can check that its visible behaviour is stable or else deliberately changed
- Changed this visible behaviour: don't store a redundant term 0, return 0 for all too-low indices (never `UINT_MAX`), clear during `initialise`, validate that the given indices are ordered
- Fixed the core bug in `nodestate`, with an explanatory comment
- Added a more targeted tx-status checking test in the e2e_suites, which I _thought_ would reveal another issue but actually everything's fine*

(* The `TermHistory` only stores lower bounds, and reports that all future indices are in the current term. So if the last thing it has heard is that term 3 started at index 10, and you ask about index 1,000, it will say term 3. This should mean that `/tx` RPCs get `Pending` instead of `Unknown` for future transactions, right? That would be a bug against our claimed behaviour - either `Unknown` never happens, or maybe we can transition from `Pending` _to_ `Unknown`. So I added an e2e test that checks for valid transitions, but it showed that future (never executed) seqnos are correctly seen as `Unknown`. Stumped me for a while - the secret is `Raft::get_term_internal`, which is already adding the _upper_-bound check that I thought would be needed in `get_tx_status`. All is right with the world)


src/consensus/raft/raft.h
```diff
@@ -20,28 +20,50 @@ namespace raft
 {
   class TermHistory
   {
+    // Entry i stores the first index in term i+1
```
eddyashton, line 23: The old behaviour was `Entry i stores the first index in term i`. Since term 0 doesn't really exist, we always had a duplicate first entry (`[2, 2, 5, 10, ...]`). This removes that unused first entry, and shifts the off-by-ones in the code.

src/node/node_state.h
```diff
@@ -563,9 +563,18 @@ namespace ccf
...
           LOG_DEBUG_FMT(
             "Read signature at {} for term {}", ledger_idx, last_sig->term);
-          for (auto i = term_history.size(); i <= last_sig->term; ++i)
+          // Initial transactions, before the first signature, must have
+          // happened in the first signature's term (eg - if the first
+          // signature is at idx 20 in term 4, then transactions 1->19 must
+          // also have been in term 4). The brief justification is that while
+          // the first node may start in an arbitrarily high term (it does not
+          // necessarily start in term 1), it cannot _change_ term before a
+          // valid signature.
```
achamayou, line 572: Another way to look at this is that the first term change can only happen in a network of more than one node. For the network to have more than one node, there must have been at least one successful config change, which means at least one commit, which means at least one signature.

tests/e2e_logging.py
```diff
@@ -337,6 +339,88 @@ def test_view_history(network, args):
...
 
+class SentTxs:
+    # view -> seqno -> status
+    txs = defaultdict(lambda: defaultdict(lambda: TxStatus.Unknown))
+
+    @staticmethod
+    def update_status(view, seqno, status=None):
+        current_status = SentTxs.txs[view][seqno]
+        if status is None:
+            # If you don't know the current status, just ensure it is present but don't change it
```
achamayou, line 350: What does "ensure it is present" mean here?

tests/e2e_logging.py
```diff
@@ -337,6 +339,88 @@ def test_view_history(network, args):
...
+        return [
+            (view, seqno)
+            for view, view_txs in SentTxs.txs.items()
+            for seqno, status in view_txs.items()
+        ]
+
+
+@reqs.description("Build a list of Tx IDs, check they transition states as expected")
+@reqs.supports_methods("LOG_record", "tx")
+def test_tx_statuses(network, args):
```
achamayou, line 381: That's an excellent test.

tests/e2e_logging.py
eddyashton: I've reworded the comment to hopefully be clearer. Essentially if you pass none, then the only action is accessing this `view.seqno` in the defaultdict (potentially creating it with the default `Unknown` status). I convinced myself that this was maybe necessary in case a previous call has already probed this Tx ID's status? Maybe we never know that it is currently 'Unknown'? It may be unneeded in the current test form.

PR #1161, opened 2020-05-07
secp256k1 node identity works with request client

jumaffre:
When using `secp256k1` as the service identity curve (for both network and node certificate), our python client used to fail to handshake successfully with CCF.

I've looked at a way to get this working with `curl` as well but failed to find any solution for it as `curl` does not seem to call the [appropriate openssl backend method](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_tmp_ecdh.html).

### DIFF
--- a/tests/infra/ccf.py
+++ b/tests/infra/ccf.py
@@ -465,19 +465,19 @@ class Network:
         all transactions globally executed on the primary (including transactions
         which added the nodes).
         """
-        for _ in range(timeout):
+        end_time = time.time() + timeout
+        while time.time() < end_time:
             with primary.node_client() as c:
                 resp = c.get("getCommit")
                 commit_leader = resp.result["commit"]
                 term_leader = resp.result["term"]
                 if commit_leader != 0:
                     break
-            time.sleep(1)
+            time.sleep(0.1)
         assert (
             commit_leader != 0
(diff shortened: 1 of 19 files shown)


tests/infra/ccf.py
```diff
@@ -465,19 +465,19 @@ def wait_for_all_nodes_to_catch_up(self, primary, timeout=3):
         all transactions globally executed on the primary (including transactions
         which added the nodes).
         """
-        for _ in range(timeout):
+        end_time = time.time() + timeout
```
jumaffre, line 468: Leftover from #1158

ashamis:
Did you try running the logging test with this change and using secp256k1 for node certs. I was seeing errors for these tests.

jumaffre:
> Did you try running the logging test with this change and using secp256k1 for node certs. I was seeing errors for these tests.

@ashamis Good point. I have pushed a commit (to-be-reverted before merging this) on this branch to enable `secp256k1` by default so we can see if all tests pass. I would expect the logging test to fail as it uses the curl client by default.

jumaffre:
> What happens if we try this with curl? Can we provide an obvious error by loading the cert in the `CurlClient` and looking at the `curve.name`?

@eddyashton `curl` fails with an handshake error. You're right though, I will had a check so that we fail early if the server is configured with the `secp256k1` curve and the `CurlClient` is used.

PR #1167, opened 2020-05-11
Store new views into the ledger

olgavrou:
Regarding issue #459 

This PR adds a kv map for new view messages and stores the new view message into the ledger.

To follow
- adding the view change message proofs into the new view message
- verifying the proofs
- processing new view messages during playback


src/consensus/pbft/libbyz/replica.h
```diff
@@ -93,7 +94,7 @@ class Replica : public Node, public IMessageReceiveBase
   // of a propagated request +1; and "ms" is the maximum sequence
   // number known to be stable.
 
-  void write_view_change_to_ledger();
+  void write_new_view_to_ledger();
```
jumaffre, line 97: Minor: comment here?

src/consensus/pbft/libbyz/test/replica_unit_tests.cpp
```diff
@@ -203,6 +205,7 @@ struct PbftState
   pbft::RequestsMap& pbft_requests_map;
   ccf::Signatures& signatures;
   pbft::PrePreparesMap& pbft_pre_prepares_map;
+  pbft::NewViewsMap& pbft_new_views_map;
```
jumaffre, line 208: Curious: why `NewViewsMap` and not `ViewsMap`?

olgavrou: It is to indicate that is is a `New_view` message that is getting stored here. I think it just makes it less ambiguous since there are more than one message types that participate in view changes

src/consensus/pbft/pbft_new_views.h
```diff
@@ -0,0 +1,30 @@
...
+#include <array>
+#include <msgpack/msgpack.hpp>
+#include <vector>
+
+namespace pbft
+{
+  struct NewView
+  {
+    int64_t view;
+    int node_id;
```
jumaffre, line 17: `NodeId node_id`

src/consensus/pbft/pbft_new_views.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <array>
+#include <msgpack/msgpack.hpp>
+#include <vector>
+
+namespace pbft
+{
+  struct NewView
+  {
+    int64_t view;
```
jumaffre, line 16: `View view`? `View` seems to be defined in `libbyz/types.h`.

src/consensus/pbft/libbyz/replica.cpp
```diff
@@ -1896,41 +1898,19 @@ void Replica::send_view_change()
...
-    LOG_TRACE_FMT(
-      "Writing view for {} with digest {} to ledger",
-      vc->view(),
-      vc->digest().hash());
-
-    ledger_writer->write_view_change(vc);
-  }
+  auto nv = vi.new_view();
+  PBFT_ASSERT(nv != nullptr, "Invalid state");
+  LOG_TRACE_FMT("Writing view for {} from {} to ledger", nv->view(), nv->id());
```
jumaffre, line 1912: `Writing new view <view> from node <node_id>` would be slightly clearer I think?

src/consensus/pbft/pbft_new_views.h
olgavrou: See the other files `pbft_pre_prepares.h` and `pbft_requests.h`. `libbyz/types.h` includes `libbyz/pbft_types.h` which in turn includes `pbft_new_views.h` causing a circular dependency if we do this. This needs to be cleaned up and dealt with, but is outside the scope if this PR.

src/consensus/pbft/pbft_new_views.h
olgavrou: same here

src/consensus/pbft/libbyz/replica.cpp
olgavrou: Sure

src/consensus/pbft/libbyz/replica.h
olgavrou: done

Issue #1162, opened 2020-05-08
Possible leak in ds_test


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8206&view=ms.vss-test-web.build-test-results-tab&runId=42690&resultId=100005&paneView=debug

```
==1435==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x502cf2 in operator new(unsigned long) (/mnt/build/1/s/build/ds_test+0x502cf2)
    #1 0x5cfaa4 in std::__1::__libcpp_allocate(unsigned long, unsigned long) /usr/lib/llvm-8/bin/../include/c++/v1/new:239:10
    #2 0x9069d8 in std::__1::allocator<std::__1::__tree_node<std::__1::__value_type<std::__1::__thread_id, unsigned short>, void*> >::allocate(unsigned long, void const*) /usr/lib/llvm-8/bin/../include/c++/v1/memory:1814:37
    #3 0x905fd4 in std::__1::allocator_traits<std::__1::allocator<std::__1::__tree_node<std::__1::__value_type<std::__1::__thread_id, unsigned short>, void*> > >::allocate(std::__1::allocator<std::__1::__tree_node<std::__1::__value_type<std::__1::__thread_id, unsigned short>, void*> >&, unsigned long) /usr/lib/llvm-8/bin/../include/c++/v1/memory:1547:21
[...]

PR #1168, opened 2020-05-11
Safe read access to thread_ids

eddyashton:
Fix for #1162. I believe this leak could only happen in the unit test, so this _removes_ the unrelated thread-messaging from this unit and adds some checks to make sure we don't trigger a similar issue in real nodes.

There's a few aspects to this fix:

- Rather than accessing the `thread_ids` map directly by `operator []`, call `get_current_thread_id` which uses `find` and throws an error if the thread hasn't been registered yet
- Move thread-message handling into the `IdleHandler`, so it doesn't need to be initialised in the messaging unit tests or benchmarks

Separately I've moved these from `enclave::` to `threading::`. There's still an awkward case around startup where we may access the thread ID (eg - for logging) before we've initialised the map (in the `enclave_run` ECALL). Here I match the old behaviour where the initial thread is given ID 0 - this should be safe since all the code in `enclave_create_node` is guarded by the `create_lock`. We could handle this more neatly by initialising this thread's ID earlier, but I think we then strongly assume the same thread returns to `run` later (true in the current code, but not enforced). We should be able to simplify this initialisation logic once we flatten to a single ECALL.

### DIFF
--- a/src/ds/thread_ids.h
+++ b/src/ds/thread_ids.h
@@ -11,6 +11,11 @@ namespace threading
 
   static inline uint16_t get_current_thread_id()
   {
+    if (thread_ids.empty())
+    {
+      return 0;
+    }
+
     const auto tid = std::this_thread::get_id();
     const auto it = thread_ids.find(tid);
     if (it == thread_ids.end())
--- a/src/ds/thread_ids.h
+++ b/src/ds/thread_ids.h
@@ -1,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include <map>
+#include <thread>
+
+namespace threading
+{
+  extern std::map<std::thread::id, uint16_t> thread_ids;
+
+  static inline uint16_t get_current_thread_id()
+  {
+    return thread_ids[std::this_thread::get_id()];
+  }
+}
--- a/src/ds/thread_ids.h
+++ b/src/ds/thread_ids.h
@@ -2,6 +2,9 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+#define FMT_HEADER_ONLY
+#include <fmt/format.h>
+#include <fmt/ostream.h>
 #include <map>
 #include <thread>
 
@@ -21,7 +24,7 @@ namespace threading
     if (it == thread_ids.end())
     {
       throw std::runtime_error(
-        "Accessed uninitialised thread_ids - ID unknown");
+        fmt::format("Accessed uninitialised thread_ids - ID {} unknown", tid));
     }
 
     return it->second;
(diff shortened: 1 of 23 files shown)


src/enclave/main.cpp
```diff
@@ -104,10 +104,10 @@ extern "C"
       {
         std::lock_guard<SpinLock> guard(create_lock);
 
-        tid = enclave::ThreadMessaging::thread_count.fetch_add(1);
-        num_pending_threads.fetch_sub(1);
-        thread_ids.emplace(std::pair<std::thread::id, uint16_t>(
+        tid = threading::ThreadMessaging::thread_count.fetch_add(1);
+        threading::thread_ids.emplace(std::pair<std::thread::id, uint16_t>(
           std::this_thread::get_id(), tid));
+        num_pending_threads.fetch_sub(1);
```
eddyashton, line 110: This may just be paranoia, but I think this ordering is safer.

If we decrement `num_pending_threads` _first_, then we allow other threads to advance past the busy-wait below before this thread has been added to the map. When we emplace, we're potentially moving the entries other threads are trying to read.

By putting ourselves in the map and _then_ decrementing `num_pending_threads`, we provide a strict gating: the map is fully populated before any thread can proceed.

src/ds/thread_ids.h
```diff
@@ -0,0 +1,29 @@
...
+    if (thread_ids.empty())
+    {
+      return 0;
+    }
+
+    const auto tid = std::this_thread::get_id();
+    const auto it = thread_ids.find(tid);
+    if (it == thread_ids.end())
+    {
+      throw std::runtime_error(
```
jumaffre, line 23: Any value in printing the thread ID here?

src/enclave/enclave.h
```diff
@@ -256,6 +256,22 @@ namespace enclave
             idling_start_time = time_now;
           }
 
+          // If we have pending thread messages, handle them now (and don't
```
jumaffre, line 259: I'm not too familiar with this code: why did this get moved from `messaging.h`?

eddyashton: This removes the dependency of `messaging.h` on `thread_messaging.h`, so we can test the "read messages from the ringbuffer" logic without any thread state initialisation.

This _does_ change the semantics slightly. Thread 0 inside the enclave may now do more reading-from-the-ringbuffer jobs before it handles any thread tasks. But the difference is minimal; its now "read everything from the ringbuffer, try again and see its empty, try a thread task" rather than "read everything from the ringbuffer, try a thread task".

src/ds/thread_ids.h
eddyashton: I think very little - we're either fully initialised or we're not, and we deliberately never use the raw `thread_id` so we have nothing to associate this with.

eddyashton: On the other hand, more information can't be a _bad_ thing, so I'll add it.

PR #1169, opened 2020-05-11
Add documentation of logging macros

eddyashton:


src/ds/logger.h
```diff
@@ -27,7 +27,7 @@ namespace logger
   enum Level
   {
     TRACE = 0,
-    DBG, // events useful for debugging
+    DEBUG, // events useful for debugging
```
eddyashton, line 30: Slightly unrelated, but consistency should help people find the macros (`INFO` => (`LOG_INFO_FMT`, `--log-level="info"`, `[info ]`), `DBG` => ???)

PR #1171, opened 2020-05-12
Add and verify view change signatures in new view message

olgavrou:
Regarding #459 

- Add `View_change` message digest signatures in `New_view` 
- Verify those signatures in `New_view::pre_verify()` and check that we have at least `2 * f + 1` of them, and if not verification fails
- Remove redundant signing over the contents of `View_change` messages, since we already sign over the digest of the contents
- Added a helper method to copy signatures when needed

Next: process new view messages during playback

### DIFF
--- a/src/consensus/pbft/libbyz/new_view.cpp
+++ b/src/consensus/pbft/libbyz/new_view.cpp
@@ -25,13 +25,16 @@ New_view::New_view(View v) : Message(New_view_tag, Max_message_size)
   }
 }
 
-void New_view::add_view_change(int id, Digest& d)
+void New_view::add_view_change(
+  int id, Digest& d, PbftSignature& sig, size_t sig_size)
 {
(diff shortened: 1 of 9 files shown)


src/consensus/pbft/libbyz/new_view.cpp
```diff
@@ -106,5 +109,32 @@ bool New_view::pre_verify()
...
+          vc_info()[i].sig.data(),
+          vc_info()[i].sig_size,
+          true /* allow self*/))
+    {
+      proof_count++;
+    }
+  }
+  LOG_TRACE_FMT("new view has verified {} view change messages", proof_count);
+
+  return proof_count >= (2 * pbft::GlobalState::get_node().f() + 1);
```
ashamis, line 139: node has a threshold variable that does the same thing

olgavrou: thanks!

PR #1172, opened 2020-05-12
Handler documentation

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1170 

- Adds documentation for `Handler` and `HandlerRegistry`
- Artificially limit the documented KV API using Sphinx breathe `:members:` directive

### DIFF
--- a/doc/developers/kv/api.rst
+++ b/doc/developers/kv/api.rst
@@ -1,12 +1,16 @@
 Key-Value Store API
 ===================
 
+This page presents the API that a CCF application must use to read and mutate the replicated key-value store.
+
+A CCF application should define one or multiple public or private :cpp:class:`kv::Map`. Then, each :cpp:class:`ccf::HandlerRegistry::Handler` should use the :cpp:class:`kv::Tx` transaction object to read and write to specific :cpp:class:`kv::Map`.
+
 Store
 -----
 
 .. doxygenclass:: kv::Store
    :project: CCF
-   :members:
+   :members: create, get
 
 Map
 ---
@@ -19,18 +23,18 @@ Map
 
 .. doxygenclass:: kv::Map
    :project: CCF
-   :members:
+   :members: get_name, set_global_hook
 
 Transaction
 -----------
 
 .. doxygenclass:: kv::Tx
    :project: CCF
-   :members:
+   :members: get_view,
 
 Transaction View
 ----------------
 
 .. doxygenclass:: kv::Map::TxView
    :project: CCF
-   :members:
+   :members: get, put, remove, foreach
--- a/src/node/rpc/handler_registry.h
+++ b/src/node/rpc/handler_registry.h
@@ -15,11 +15,6 @@
 
 namespace ccf
 {
-  /** TODO:
-   *
-   *
-   *
-   */
   struct RequestArgs
   {
     std::shared_ptr<enclave::RpcContext> rpc_ctx;
@@ -34,19 +29,12 @@ namespace ccf
 
   using HandleFunction = std::function<void(RequestArgs& args)>;
 
-  /** TODO:
-   *
-   *
-   *
-   *
+  /** The HandlerRegistry records the user-defined Handlers for a given
+   * CCF application.
    */
   class HandlerRegistry
   {
   public:
-    /** TODO:
-     *
-     *
-     */
     enum ReadWrite
     {
       Read,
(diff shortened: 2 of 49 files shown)


doc/developers/kv/api.rst
```diff
@@ -1,12 +1,16 @@
...
+
+A CCF application should define one or multiple public or private :cpp:class:`kv::Map`. Then, each :cpp:class:`ccf::HandlerRegistry::Handler` should use the :cpp:class:`kv::Tx` transaction object to read and write to specific :cpp:class:`kv::Map`.
+
 Store
 -----
 
 .. doxygenclass:: kv::Store
    :project: CCF
-   :members:
+   :members: create, get
```
jumaffre, line 13: Only `create()` and `get()` gets documented. Obviously not the long-term solution but might prevent some users from calling other methods.

src/node/rpc/handler_registry.h
```diff
@@ -56,12 +75,30 @@ namespace ccf
...
       }
 
+      /** Sets the schema that the request parameters and response must comply
+       * with based on JSON-serialisable data structures.
+       *
+       * \verbatim embed:rst:leading-asterisk
+       * .. note::
+       *  See ``DECLARE_JSON_`` serialisation macros for serialising
+       *  user-defined data structures.
+       * \endverbatim
```
jumaffre, line 96: I think this is cool as it can make notes/warnings really stand out, e.g.:
image

src/node/rpc/handler_registry.h
```diff
@@ -132,6 +217,7 @@ namespace ccf
       // Default is that all verbs are allowed
       uint64_t allowed_verbs_mask = ~0;
 
+      // https://github.com/microsoft/CCF/issues/1102
       Handler& set_allowed_verbs(std::set<http_method>&& allowed_verbs)
```
jumaffre, line 221: I haven't documented this not the `ReadWrite` enum as they will most likely be removed in the near future.

src/node/rpc/handler_registry.h
```diff
@@ -39,6 +42,17 @@ namespace ccf
       MayWrite
     };
 
+    /** A Handler represents a user-defined endpoint that can be invoked by
+    * authorised users via HTTP requests, over TLS. A Handler is accessible at a
+    * specific verb and URI, e.g. POST /app/accounts or GET /app/records.
+    *
+    * Handlers can read from and mutate the state of the replicated key-value
+    store. The effects of Handler execution are handled by the underlying
```
olgavrou, line 50: `The effects of Handler execution ....` maybe tmi here?

src/node/rpc/handler_registry.h
```diff
@@ -86,24 +123,57 @@ namespace ccf
...
       bool require_client_identity = true;
 
+      /** Requires that the HTTPS request is emitted by a user whose public
+       * identity has been registered in advance by consortium members.
+       *
+       * By default, a known client identity is required.
+       *
+       * \verbatim embed:rst:leading-asterisk
+       * .. warning::
+       *  It is left to the application developer to implement the
```
eddyashton, line 170: `If set to false then it is left to the application developer...`?

src/node/rpc/handler_registry.h
```diff
@@ -119,9 +189,24 @@ namespace ccf
         return *this;
       }
 
-      // If true, request is executed without consensus (PBFT only)
       bool execute_locally = false;
 
+      /** Indicates that the execution of the handler does not require
+       * consensus from other nodes in the network. This has effect only if PBFT
+       * is used as the underlying consensus algorithm.
```
eddyashton, line 196: I don't think this _should_ be PBFT-only (not sure if it currently _is_). A node-status RPC should be local-only regardless of consensus choice. And I think the core detail is that it only _reads_ node-local state, not things which require consensus. Devs could provide as many of these as they like for per-node debugging, but in general they shouldn't need to. This also breaks our session consistency claims from elsewhere since it will never be forwarded (this is useful for some RPCs!). I think they shouldn't even read from the KV? But perhaps this is unavoidable.

src/node/rpc/handler_registry.h
```diff
@@ -205,15 +299,15 @@ namespace ccf
...
+     * @return The installed Handler for further modification
      */
     Handler& set_default(HandleFunction f, ReadWrite read_write)
     {
       default_handler = {"", f, read_write, this};
       return default_handler.value();
     }
 
-    /** Populate out with all supported methods
+    /*** Populate out with all supported methods
```
eddyashton, line 310: Minor: extra `*`?

src/node/rpc/handler_registry.h
eddyashton: Nit: Is it possible to change the formatting of this function title/type? It bothers me that the spacing is different in `Handler &foo()` (vs `Handler& foo()` in the actual code).

src/node/rpc/handler_registry.h
jumaffre: You're absolutely right. At the moment, the flag is only used to circumvent the broadcast and consensus on requests issued with PBFT, for the `getMetrics` and `getLocalCommit` endpoints. I agree that they shouldn't even read from the KV. I will had a note on that.

src/node/rpc/handler_registry.h
jumaffre: Yeah, I agree 👍

src/node/rpc/handler_registry.h
jumaffre: As discussed, this does not seem possible with Doxygen/Breathe.

PR #1175, opened 2020-05-13
Add support for floating point types, correct integer type

### LINKED ISSUE #1174: Struct serialisation fails for double/float types
JSON schema mapping (json_schema:.h:119) is missing entries for those types.

achamayou:
fix #1174

### DIFF
--- a/src/ds/json_schema.h
+++ b/src/ds/json_schema.h
@@ -43,7 +43,7 @@ namespace ds
     template <typename T>
     inline void fill_number_schema(nlohmann::json& schema)
     {
-      schema["type"] = "number";
+      schema["type"] = "integer";
       schema["minimum"] = std::numeric_limits<T>::min();
       schema["maximum"] = std::numeric_limits<T>::max();
     }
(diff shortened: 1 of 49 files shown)


src/ds/test/json_schema.cpp
```diff
@@ -179,10 +179,25 @@ TEST_CASE("schema types")
...
   REQUIRE(schema["items"]["items"].is_array());
   REQUIRE(schema["items"]["items"].size() == 2);
-  REQUIRE(schema["items"]["items"][0]["type"] == "number");
+  REQUIRE(schema["items"]["items"][0]["type"] == "integer");
   REQUIRE(schema["items"]["items"][1]["type"] == "string");
 }
 
+TEST_CASE("schema types II")
+{
+  std::map<size_t, float> m;
```
eddyashton, line 188: Minor: Would be good to test double support as well? I may have syntax wrong, but I believe this will work:

```suggestion
TEST_CASE_TEMPLATE("schema types II", T, float, double)
{
  std::map<size_t, T> m;
```

src/ds/json_schema.h
```diff
@@ -43,7 +43,7 @@ namespace ds
     template <typename T>
     inline void fill_number_schema(nlohmann::json& schema)
     {
-      schema["type"] = "number";
+      schema["type"] = "integer";
```
achamayou, line 46: Not sure why this was done that way to begin with, it's not incorrect, but slightly less precise than it could be. https://cswr.github.io/JsonSchema/spec/basic_types/

src/ds/test/json_schema.cpp
achamayou: Good point, and the syntax is right. I've tweaked the integer test as well while I was at it.

Issue #1163, opened 2020-05-08
Incorrect transaction id reporting when using multiple threads


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8222&view=ms.vss-test-web.build-test-results-tab&runId=42682&resultId=100010&paneView=attachments

```
2020-05-08 14:19:34.546 | INFO     | infra.remote:log_errors:74 - 2020-05-08T13:19:29.576005Z        100 [info ] amples/perf_client/perf_client.h:530 | Triggered signature at 2.20
2020-05-08 14:19:34.547 | INFO     | infra.remote:log_errors:74 - 2020-05-08T13:19:29.576026Z        100 [info ] ../samples/perf_client/timing.h:261  | Waiting for transaction ID 2.20
2020-05-08 14:19:34.547 | INFO     | infra.remote:log_errors:74 - 2020-05-08T13:19:30.111115Z        100 [info ] ../samples/perf_client/timing.h:299  | Found global commit 2.20
2020-05-08 14:19:34.547 | INFO     | infra.remote:log_errors:74 - 2020-05-08T13:19:30.111134Z        100 [info ] ../samples/perf_client/timing.h:304  |  (headers term: 2, local: 21, global: 21)
2020-05-08 14:19:34.547 | INFO     | infra.remote:log_errors:74 - 2020-05-08T13:19:32.023628Z        100 [info ] amples/perf_client/perf_client.h:439 | About to begin timing
2020-05-08 14:19:34.547 | INFO     | infra.remote:log_errors:74 - 2020-05-08T13:19:32.023642Z        100 [info ] amples/perf_client/perf_client.h:441 | Began timing
2020-05-08 14:19:34.547 | INFO     | infra.remote:log_errors:74 - 2020-05-08T13:19:33.231753Z        100 [fail ] amples/perf_client/perf_client.h:682 | Transaction exception: Term went backwards (expected 2, saw 0)!
2020-05-08 14:19:34.547 | ERROR    | infra.remote:log_errors:83 - Contents of /mnt/build/1/s/build/workspace/SB_sig_SGX_CFT^_client_0/err:
terminate called after throwing an instance of 'std::exception'
  what():  std::exception
```

achamayou:
Another instance in https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8301&view=ms.vss-test-web.build-test-results-tab&runId=42962&resultId=100014&paneView=attachments

[...]

PR #1179, opened 2020-05-14
Split kv.h

eddyashton:
`kv.h` is unwieldy. We have longer term plans to formalise and simplify its public API, but in the interim it remains difficult to make small alterations due to the size of the file, repetition of names, and dense connections between its components. In particular I was struggling to implement, much less reason about, the fixes for #1163.

This PR chops it very roughly into multiple headers (`kv/map.h`, `kv/store.h`, `kv/tx.h`, `kv/views.h`), and then tries to break some of the dependencies between the separated classes.

Summary of interesting changes:
- `Tx` is no longer a friend of `Map`, so `Tx` doesn't need to be forward-declared. The only thing that `Tx` access on `Map` (seen as a templated `M`) is the name, so this is exposed on a public function
- The static `Tx::commit()` which did the real work is now a standalone function `apply_views`, so that `Store` can use it without knowing about `Tx`. As a result, the `Store::Tx` typedef is removed. This was convenient - we used the type `ccf::Store::Tx` in several places so that we didn't have to think about templates, this is replaced by `ccf::StoreTx` (this forms the majority of the changed files)

Draft for now so any objections can be raised, and while I see if there are any other productive incisions worth making.


src/kv/test/kv_bench.cpp
```diff
@@ -39,7 +39,7 @@ static void serialise(picobench::state& s)
 
   auto& map0 = kv_store.create<std::string, std::string>("map0", SD);
   auto& map1 = kv_store.create<std::string, std::string>("map1", SD);
-  Store::Tx tx;
+  StoreTx tx;
```
eddyashton, line 42: `StoreTx` looks strange. These are really `ccf::StoreTx`, and should perhaps instead be `ccf::Tx`? But I'm wary of just using `Tx` everywhere (which is possible because all of our tests do `using namespace ccf;`). Maybe its fine. This is a tomorrow problem.

eddyashton: I've decided to call this `ccf::Tx`, but including the namespace everywhere this is used for clarity, even where its redundant. So (hopefully), `Tx` is always `kv::Tx<S, D>`, and we can distinguish it from `ccf::Tx` aka `kv::Tx<CCF's serialiser, CCF's deserialiser>`. 2 related points:

1) When we go to everything-in-the-kv-is-a-byte-vector we can probably remove the `<S, D>` in most places?
2) I think we're using the `ccf` namespace in too many places. I've moved the logging app into its own `logging` namespace as a test, we should do this with the other apps. And the unit tests should, where possible, avoid `using namespace ccf` (eg - the kv unit tests should really only test the `kv::` namespace, not the specialisation choices that CCF makes).

src/apps/logging/logging.cpp
```diff
@@ -15,9 +15,8 @@
 
 using namespace std;
 using namespace nlohmann;
-using namespace ccf;
 
-namespace ccfapp
+namespace logging
```
eddyashton, line 19: API design point that we've been a bit lax on: Only `get_rpc_handler` needs to be in namespace ccfapp, the rest of the app can (and probably should) be in its own, clearly-not-CCF namespace. I've moved logging since its our main sample, and removed the `using namespace ccf` so we're clear on the framework types/dependencies.

src/kv/test/kv_test.cpp
```diff
@@ -632,55 +630,56 @@ TEST_CASE("private recovery map swap")
...
   }
-  s2.compact(s2.current_version());
   {
-    Store::Tx tx;
+    ccf::Tx tx;
     auto v = tx.get_view(priv2);
     v->put(13, 13);
     tx.commit();
   }
+  s2.compact(s2.current_version());
```
eddyashton, line 675: This is a real change!

I've removed `set_read_committed` entirely since it was unused apart from in this unit test, and it should be replaceable with `get_globally_committed`. However, that broke this unit test. When we swap private maps, the `set_read_committed` method (creating views at only globally-committed read_versions) did what we wanted, but calling `get_globally_committed` returned a different value! Specifically, key 13 was _not_ present in s1's committed roll! There's a comment on `swap_private_maps` that describes exactly this behaviour, pasting verbatim:

```
     * It's also important to note that swapping in private state
     * may result in previously uncompacted writes becoming effectively
     * compacted, from a user's perspective (they would not be internally
     * compacted until the next compact() however). So it is important to
     * make sure that the private state being swapped in is fully compacted
     * before the swap.
```

So I've followed the guidance in the final sentence - the private state is _fully_ compacted before swapping. If anyone understands this better than I do and sees a problem, please shout!

src/apps/logging/logging.cpp
achamayou: Can we make this loggingapp or app::logging or something less likely to evoke logging.h?

src/kv/test/kv_bench.cpp
achamayou: 1. it depends if Map is the wrapper type or not. In practice we will, but users will probably still be looking at UserMap<K, V> where K and V implement serialise()/deserialise() or something along those lines.
2. Agreed, but logging is too similar to one of the internal namespaces.

src/apps/logging/logging.cpp
eddyashton: Good point - I've gone for `loggingapp` for now.

achamayou:
Let's make sure we have a good descriptive entry in the release notes because this is a breaking change for users. Also, the doxygen pointers in sphinx need updating.

eddyashton:
@achamayou Agree on the release notes, and will try to consolidate KV API changes with this release. I believe the doxygen pointers are fine - the few that were `kv::Store::Tx` are now `kv::Tx`, and since they're all looked up by typenames rather than files they still work? Unless I'm missing something...

PR #1180, opened 2020-05-14
Create codeql-analysis.yml

achamayou:


achamayou:
The setup playbook is either failing or timing out, it's difficult to work out which because the log isn't available.

A sensible next step would be to add a variant of our playbooks that uses the binary OE release, and try it with it (since this isn't performance sensitive).

PR #1188, opened 2020-05-15
Update signature documentation

### LINKED ISSUE #1182: Sig max tx and ms aren't documented
These flags and their implications ought to be documented.

achamayou:
Fix #1182

### DIFF
--- a/doc/operators/start_network.rst
+++ b/doc/operators/start_network.rst
@@ -9,18 +9,20 @@ Starting a New Network
 Starting the First Node
 -----------------------
 
-To create a new CCF network, the first node of the network should be started with the ``start`` option:
+To create a new CCF network, the first node of the network should be invoked with the ``start`` option:
 
 .. code-block:: bash
 
     $ cchost
     --enclave-file /path/to/enclave_library
-    --node-address node_ip:node_port
-    --rpc-address <ccf-node-address>
+    --node-address <ccf-node-address>
+    --rpc-address <ccf-node-rpc-address>
     --public-rpc-address <ccf-node-public-address>
     [--domain domain]
     --ledger-file /path/to/ledger
     --node-cert-file /path/to/node_certificate
+    [--sig-max-tx number_of_transactions]
+    [--sig-max-ms number_of_milliseconds]
     start
     --network-cert-file /path/to/network_certificate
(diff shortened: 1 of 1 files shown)


doc/operators/start_network.rst
```diff
@@ -9,41 +9,63 @@ Starting a New Network
...
+
+Signature Interval
+~~~~~~~~~~~~~~~~~~
+
+Transaction commit latency in a CCF network is primarily a function of signature frequency. A network emitting signatures more frequently will be able to commit transactions faster,
+but will spend a larger proportion of its execution resources creating and verifying signatures. Setting signature frequency is a trade-off between transaction
+latency and throughput.
 
-.. note:: Once a CCF network is started, :ref:`members can add other members and users via governance <members/open_network:Opening a Network>`.
-.. note:: If starting up the network with PBFT enabled as the consensus protocol, be sure to add the ``--consensus pbft`` CLI argument  when starting up the node. For more info on the provided consensus protocols please see :ref:`here <developers/consensus:Consensus Protocols>`
```
achamayou, line 46: That was a lot of notes. As a rule of thumb, I would suggest that a section that contains more than, say, two notes, probably ought to be broken into sub-sections.

doc/operators/start_network.rst
```diff
@@ -9,41 +9,63 @@ Starting a New Network
...
+Two options are provided to that end, the first ```--sig-max-tx``` is the maximum number of transactions that should be allowed to occur between two
```
eddyashton, line 67: I think this would be easier to read as a list/bullet points, rather than a single sentence.

achamayou: Absolutely!

doc/operators/start_network.rst
```diff
@@ -9,41 +9,65 @@ Starting a New Network
...
+To create a new CCF network, the first node of the network should be invoked with the ``start`` option:
 
 .. code-block:: bash
 
     $ cchost
     --enclave-file /path/to/enclave_library
-    --node-address node_ip:node_port
-    --rpc-address <ccf-node-address>
+    --node-address <ccf-node-address>
+    --rpc-address <ccf-node-rpc-address>
```
jumaffre, line 19: The RPC address "<ccf-node-address>" is referred to everywhere else in our docs. I think it will need updating too.

PR #1191, opened 2020-05-18
Remove S and D templates from kv types

eddyashton:
`kv::Store` and `kv::Tx` are no longer templates. We use msg-pack to serialise both Ks and Vs to byte-vectors, and write-sets (containing those byte-vectors) to ledger entries. The nljson serialiser can still be used by a compile-time switch, but it currently doesn't compile (and hasn't for a while as new types haven't been given serialisers - this is low-priority, and the requirement should be removed entirely in the near future).

As a result, `ccf::Store` and `ccf::Tx` are no longer needed (they were just typedefs for `kv::Store<MsgPackSerialiser, MsgPackDeserialiser>`). Most of the changes are renaming uses of these types, and related include changes (`entities.h` no longer includes `kv/`, so the things which actually define framework tables don't get `kv::Map` transitively and must do so explicitly).

I've also moved some of the stub implementations of kv types into `kv/`, for instance `kv/tests/null_encryptor.h`. The most involved of these is the refactoring of `kv::TxEncryptor`, which doesn't know anything about `LedgerSecrets` types - the actual implementations come from `src/node/encryptor.h`. The goal is for `kv/` to depend only on `ds/` and `crypto/`; I believe the only violations of this are now under `kv/test/`.

`kv.h` is removed entirely (RIP), and each includer now picks which types it actually wants (`map.h` to declare tables, `store.h` for the per-node/per-test collection of all tables, `tx.h` to safely manipulate maps).


src/kv/encryptor.h
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
eddyashton, line 1: This contains most of what was `node/encryptor.h`. There are 2 notable changes:
- it knows nothing about `LedgerSecrets` types. It is instead constructed from a `KeyInfo` list, and calls a virtual `record_compacted_keys` for the reverse communication
- the different ways of setting iv_id (`set_view` in PBFT, node ID at construction in Raft) are combined into a single `set_iv_id` method on this base class. So Raft should always to remember call `set_iv_id` afterwards, once

src/kv/map.h
```diff
@@ -110,8 +108,25 @@ namespace kv
...
+      }
+      else
+      {
+        c->~LocalCommit();
+        new (c) LocalCommit(std::forward<Args>(args)...);
+      }
+      return c;
+    }
+
+  public:
```
eddyashton, line 127: Big diff here is actually just "make constructor and clone public", so they can be called by the non-friend `Store`.

src/kv/msg_pack_serialise.h
```diff
@@ -2,10 +2,9 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
-#include "../ds/msgpack_adaptor_nlohmann.h"
-#include "../ds/serialized.h"
+#include "ds/msgpack_adaptor_nlohmann.h"
```
eddyashton, line 5: It seems all of our relative includes are unnecessary - I may do a pass removing them for consistency and clarity. They should all be relative to `src/`, and eventually they should get a `ccf/` prefix.

src/kv/test/kv_serialisation.cpp
```diff
@@ -324,12 +325,18 @@ TEST_CASE("Integrity" * doctest::test_suite("serialisation"))
...
-    secrets->init();
-    auto encryptor = std::make_shared<ccf::RaftTxEncryptor>(1, secrets);
+    std::list<kv::TxEncryptor::KeyInfo> keys;
+    std::vector<uint8_t> raw_key(crypto::GCM_SIZE_KEY);
+    for (size_t i = 0; i < raw_key.size(); ++i)
+    {
+      raw_key[i] = i;
+    }
+    keys.push_back({kv::Version(0), raw_key});
+    auto encryptor = std::make_shared<kv::TxEncryptor>(keys);
```
eddyashton, line 335: Side effect of the stripping of `Encryptor` dependencies. The tests don't need to create a 'real' `LedgerSecrets`, just a "here's the key to use from version 0" list.

src/kv/test/null_encryptor.h
```diff
@@ -0,0 +1,51 @@
...
+  public:
+    void encrypt(
+      const std::vector<uint8_t>& plain,
+      const std::vector<uint8_t>& additional_data,
+      std::vector<uint8_t>& serialised_header,
+      std::vector<uint8_t>& cipher,
+      kv::Version version) override
+    {
+      crypto::GcmHeader<crypto::GCM_SIZE_IV> gcm_hdr = {};
+      serialised_header = std::move(gcm_hdr.serialise());
```
eddyashton, line 22: Minor: I'm surprised that the null encryptor needs a correctly-sized-but-invalid header, but haven't tried removing it yet to see what breaks.

src/kv/msg_pack_serialise.h
jumaffre: I don't know if you're referring to those too but I quite like the files in `test/` to use relative includes for the class they test. Maybe it's just me...

jumaffre, line 5: Nitpick: `msgpack_` for this include but `msg_pack` for the name of this file.

src/kv/test/null_encryptor.h
jumaffre: Good point. I can't remember either. It looks like it's been there since the beginning of time so it might not break anything.

src/kv/msg_pack_serialise.h
eddyashton: Good catch, I've renamed this to `msgpack_serialise` (to match both `msgpack_adaptor` and the 3rd party `msgpack`).

eddyashton: I don't have a strong opinion, but it seems cleaner to use the same name for a file everywhere we can? So we can replace `"ds/foo.h"` and not worry about missing `"../../../../foo.h"`.

This also makes it easier to move the test files around, or build them externally, where they're testing the external API that 'real' things use. The exception might be internal or test-only headers that aren't visible externally, but I think these would just be under `_private` or `test` rather than relative?

src/kv/test/null_encryptor.h
eddyashton: Looks like this is safe to remove, so I've removed it.

src/kv/msg_pack_serialise.h
jumaffre: Yeah, consistency certainly is more important. In any case, I'm not too fussed either way.

src/consensus/pbft/libbyz/replica.h
```diff
@@ -367,7 +367,7 @@ class Replica : public Node, public IMessageReceiveBase
     Request& request,
     int64_t& max_local_commit_value,
     bool include_markle_roots,
```
achamayou, line 369: Unrelated but merkle, not markle.

achamayou, line 369:
```suggestion
    bool include_merkle_roots,
```

src/kv/encryptor.h
```diff
@@ -0,0 +1,214 @@
...
+namespace kv
+{
+  class TxEncryptor : public kv::AbstractTxEncryptor
+  {
+  public:
+    struct KeyInfo
+    {
+      kv::Version version;
+
+      // This is unfortunate. Because the encryptor updates the ledger secrets
```
achamayou, line 18: @jumaffre Why is this unfortunate? Is it because we keep another copy of the ledger secrets somewhere else? If so, where?

eddyashton: @achamayou AIUI, we also have a _useable_ representation of the key in-memory:

```
    struct EncryptionKey : KeyInfo
    {
      crypto::KeyAesGcm key;
    };
```

So its unfortunate that we need to store both a transferable raw representation, and the struct that operates with it.

PR #1193, opened 2020-05-19
Websockets support

achamayou:
Replaces #1089 

The main differences are:

1. This uses a custom frame, which is essentially similar to what the In/Out flatbuffers contained, but is a lot more compact. Only very slight extensions to serialized.h are made for that purpose, the direction of travel is that this will likely end up being done with Everparse (but the binary layout will likely remain the same).
2. Forwarding and PBFT are enabled.

Changes to the perf client will be done in a separate PR.

I have removed signatures from the in frame, I'm not convinced there's a use case for them on what's essentially our efficient transport option.

One reason why this is tested by WS scaffold rather than the regular logging test is that although the current code makes some effort to be compatible in its current form, it deliberately omits content type and method.


src/http/http_rpc_context.h
```diff
@@ -339,4 +381,92 @@ namespace enclave
...
+    std::shared_ptr<enclave::SessionContext> s,
+    const std::vector<uint8_t>& packed,
+    enclave::FrameFormat frame_format,
+    const std::vector<uint8_t>& raw_pbft = {})
+  {
+    http::SimpleRequestProcessor processor;
+
+    LOG_DEBUG_FMT("PARSING FORWARDED REQUEST");
+    LOG_DEBUG_FMT("{:02x}", fmt::join(packed, " "));
+    LOG_DEBUG_FMT("{}", std::string(packed.begin(), packed.end()));
```
eddyashton, line 395: Is this staying? And if so, should it be `LOG_TRACE_FMT`?

src/http/ws_builder.h
```diff
@@ -0,0 +1,292 @@
...
+    std::vector<uint8_t> build_request(bool header_only = false) const
+    {
+      const auto uri = fmt::format("{}{}", path, get_formatted_query());
+
+      const auto body_view = (header_only || body == nullptr) ?
+        std::string_view() :
+        std::string_view((char const*)body, body_size);
+
+      const auto request_string = fmt::format(
+        "{} {} HTTP/1.1\r\n"
```
eddyashton, line 184: A lot of this file seems to be HTTP-specific. What does WebSockets use it for?

src/http/http_rpc_context.h
achamayou: No, that's not useful anymore.

src/http/ws_builder.h
achamayou: That's unused now, I've removed it.

src/ds/serialized.h
```diff
@@ -60,6 +60,19 @@ namespace serialized
     return v;
   }
 
+  inline std::string_view read_lpsv(const uint8_t*& data, size_t& size)
```
jumaffre, line 63: `lpsv`?

achamayou: Length-prefixed string view.

src/http/http_endpoint.h
```diff
@@ -273,9 +282,18 @@ namespace http
       }
       catch (const std::exception& e)
       {
-        send_response(
-          fmt::format("Exception:\n{}\n", e.what()),
-          HTTP_STATUS_INTERNAL_SERVER_ERROR);
+        if (is_websocket)
```
jumaffre, line 285: Why is there a special case for ws and http here while there isn't on line 263 for example?

tests/infra/clients.py
```diff
@@ -366,33 +367,50 @@ def request(self, request, is_signed=False):
...
-        # pylint: disable=unused-variable
-        ws = create_connection(
-            f"wss://{self.host}:{self.port}",
-            sslopt={"certfile": self.cert, "keyfile": self.key, "ca_certs": self.ca},
+    def request(self, request, is_signed=False):
+        assert not is_signed
+
+        if not self.ws:
+            LOG.info("Creating WSS connection")
+            self.ws = websocket.create_connection(
```
jumaffre, line 385: We should raise a `CCFConnectionException` exception here is the connection fails.

src/http/http_endpoint.h
achamayou: At 263 and most places we can, we delegate error serialisation to the context (rpc_ctx->serialise_error), but if we fail to create the context (:207) or if the context is gone (:285), we can't do that.

This is a particularly happy situation, if you have a better idea, I'm very happy to explore it!

tests/infra/clients.py
```diff
@@ -366,33 +367,50 @@ def request(self, request, is_signed=False):
...
+        header = struct.pack("<h", len(path)) + path
+        # FIN, no RSV, BIN, UNMASKED every time, because it's all we support right now
+        frame = websocket.ABNF(
+            1, 0, 0, 0, websocket.ABNF.OPCODE_BINARY, 0, header + payload
         )
-
-    def signed_request(self, request):
-        raise NotImplementedError("Signed requests not yet implemented over WebSockets")
+        self.ws.send_frame(frame)
+        out = self.ws.recv_frame().data
```
jumaffre, line 401: And a `TimeoutError` here if we don't receive anything.

src/http/ws_parser.h
```diff
@@ -0,0 +1,247 @@
...
+#include <algorithm>
+#include <cctype>
+#include <endian.h>
+#include <string>
+
+namespace ws
+{
+  static constexpr size_t INITIAL_READ = 2;
+  static constexpr size_t OUT_CCF_HEADER_SIZE =
+    sizeof(uint16_t) + sizeof(size_t) * 3;
```
jumaffre, line 17: Is it `* 3` here because of commit, term and global commit? Might be worth documenting this more clearly.

tests/infra/clients.py
achamayou: My understanding is that this is done at the level above by CCFClient, but I'll check.

tests/infra/clients.py
achamayou: We get that from websocket, but we should set the timeout!

src/http/ws_parser.h
achamayou: Yes, will do.

src/http/ws_parser.h
```diff
@@ -0,0 +1,247 @@
...
+        case READ_BODY:
+        {
+          assert(data.size() == size);
+
+          const uint8_t* buf = data.data();
+          size_t s = data.size();
+          auto path = serialized::read_lpsv(buf, s);
+          std::vector<uint8_t> body(buf, buf + s);
+
+          proc.handle_request(
```
jumaffre, line 231: I guess this means JSON-only payloads with hardcoded `HTTP_POST` only for now?

src/http/ws_rpc_context.h
```diff
@@ -0,0 +1,271 @@
...
+        break;
+      }
+      case 8:
+      {
+        msg[1] = 0x7f;
+        *((uint64_t*)&msg[2]) = htobe64(frame_size);
+        break;
+      }
+      default:
+        throw std::logic_error("Unreachable");
```
jumaffre, line 49: Error message is rather vague here. Perhaps mention the `sz_size` here too?

src/http/ws_parser.h
achamayou: Sort of, in practice the payload can be anything if the handler ignores the content-type, this is just convenient to reuse the logging test right now. Once websocket has it own handler signature, both the verb and the content type will go.

src/http/ws_rpc_context.h
achamayou: sz_size is 0 | 2 | 8 though (see line 20), this really is unreachable.

tests/infra/clients.py
jumaffre: The underlying client implementation has to raise a `CCFConnectionException` is the initial connection fails.

src/ds/serialized.h
achamayou: Added a description.

tests/infra/clients.py
achamayou: Ok, I see, I'll add that and make sure that's chained.

PR #1197, opened 2020-05-19
Extract TxView

eddyashton:
Another step in the kv cleanup:

- Removes some more unused fields and functions. `start_order` and `end_order` are removed, hooks don't need to be passed to table constructor.
- Pulls `TxView` out of `Map` (into `tx_view.h`), and separates the "app-visible" `TxView` from the `Map`-specific `ConcreteTxView` (implementing serialisation and commitability, but can be private to the kv).

There's still some work for the 2 features we want out of this:
1) Having kv-tables store byte-vectors, and wrapping these in app-defined serialisers. I think we can write new `Map` and `TxView` implementations that do this, and they should be storable in the same `Store` and `Tx` types (which _store_ `Abstract` types, and only deal with specific implementation types in templated methods). We want to reuse as much of the existing `Map` code as possible, essentially just creating a different `View` type, so I'll look for the minimum extension point. As a transition step, I'm planning to define these as `kv::experimental::TxView` or similar, so that we can modify a handful of instances by replacing `store.create<kv::Map>` with `store.create<kv::experimental::Map>`. Then once we're happy with the API, we can do a rename where the _old_ `kv::Map` becomes `kv::untyped::Map` (or similar), and we remove `experimental` from the new one - it _should_ be a drop-in replacement for existing code, modulo perhaps some declaration of the serialisation scheme.
2) Grabbing the current _view_ wherever we set a `Version`, so we can report the correct `view.version` for the transaction. I think the safe way to do this is to make the slow slightly more functional, explicitly producing `WriteSets` or `ChangeSet` from certain methods rather than adding to the internal state tracking.


src/kv/tx_view.h
```diff
@@ -0,0 +1,239 @@
...
+
+        if ((write == w.end()) && !is_deleted(v.version))
+          return f(k, v.value);
+        return true;
+      });
+
+      for (auto write = writes.begin(); write != writes.end(); ++write)
+      {
+        if (!is_deleted(write->second.version))
+          if (!f(write->first, write->second.value))
```
achamayou, line 233: Bit strange not to &&, but not wrong as such.

src/kv/tx.h
```diff
@@ -254,11 +255,11 @@ namespace kv
 
       for (auto domain_it : grouped_maps)
       {
-        for (auto curr_map : domain_it.second)
+        for (auto curr_view : domain_it.second)
```
jumaffre, line 258: As `get_maps_grouped_by_domain` returns a `std::map<kv::SecurityDomain, std::vector<AbstractMap*>>`, I would have thought that `curr_map` was correct?

eddyashton: This is, if I'm reading it right, `std::map<kv::SecurityDomain, std::vector<AbstractTxView*>>`. So its a collection of views, not maps. There are 2 functions called `get_maps_grouped_by_domain`, and only one is called? I'll rename this (and related variales) to `get_VIEWs_...`, and remove the unused one.

src/kv/view_containers.h
```diff
@@ -3,13 +3,21 @@
...
+    // Weak pointer to source map
+    AbstractMap* map;
+
+    // Owning pointer of TxView over that map
+    std::unique_ptr<AbstractTxView> view;
+  };
+
   // When a collection of Maps are locked, the locks must be acquired in a
   // stable order to avoid deadlocks. This ordered map will claim in name-order
   using OrderedViews = std::map<std::string, MapView>;
```
jumaffre, line 23: Is `get_maps_grouped_by_domain` defined twice (in `store.h` as well).

eddyashton: Yes it was, with subtly different types. I've removed it, now only `get_views_grouped_by_domain` in `tx_view.h`

eddyashton:
There's a small semantic change in this PR that I'd forgotten about but should be mentioned explicitly, since its not obvious due to the other code shuffling:

The methods in `TxView` (`get`, `put`, etc) used to all return false/null early if the view had already been committed. This has been removed, as this base `TxView` has no concept of being committed (this is handled in the derived `ConcreteTxView`). I think this is sensible as we try to separate the app-visible behaviour from the framework-view; we should be able to ensure that these methods aren't called post-commit with other means, ideally at compile-time or with errors. This now means some of these bool return values are redundant, always returning `true`. They were entirely ignored anyway, and keeping them reduces the visible API changes.

Issue #1187, opened 2020-05-15
Use consistent view.seqno terminology


eddyashton:
We're aiming to use consistent terminology in our external API, where transactions execute in a _view_ at a _seqno_. We use this in a few places, but many places still use the old _term_ and _commit_.

- [x] Rename response headers (`x-ccf-term` to `x-ccf-view`)
- [x] Remove `x-ccf-global-commit` header (potentially a separate issue, but for clarity it would be nice to announce at the same time)
- [x] Update documentation to use new terminology. In particular, to talk about `Transaction IDs` composed of `view.seqno`
- [x] Update all external/sample code to use new terminology

achamayou:
@eddyashton I think we want `x-ccf-tx-view` and `x-ccf-tx-seqno`, with the documentation describing those two values as Transaction ID. The status of that transaction can then be queried through the tx endpoint.

achamayou:
- [ ] Update naming in Python infra for clarity, since it is visible through the start_test_network and recovery scripts

achamayou:
So I think #1127 wraps this up effectively. It does not remove global_commit for now, because we'd need to update the perf tests substantially, but it's removed from the documentation.

I think there's follow up work on improving the documentation around the user RPC/transaction flow, but I'll do that separately.

PR #1206, opened 2020-05-22
View seqno rename

achamayou:
Towards #1187, starting with the user-visible parts.


tests/infra/ccf.py
```diff
@@ -538,20 +538,20 @@ def wait_for_all_nodes_to_catch_up(self, primary, timeout=3):
         while time.time() < end_time:
             with primary.node_client() as c:
                 resp = c.get("getCommit")
```
eddyashton, line 540: Is it worth renaming `getCommit`, and/or its response fields `commit` and `term` in this PR? Or do we think this is going to be replaced by a generic `status` RPC soon, so there's no point breaking the names of it now?

doc/users/issue_commands.rst
```diff
@@ -19,17 +19,15 @@ For example, to record a message at a specific id with the :ref:`developers/exam
...
     content-type: application/json
-    x-ccf-commit: 23
-    x-ccf-global-commit: 22
-    x-ccf-term: 2
+    x-ccf-tx-seqno: 23
+    x-ccf-tx-view: 2
 
     true
 
 The HTTP response some CCF commit information in the headers:
```
eddyashton, line 27: Unchanged, but spotted here: Should say "response **contains** some"

tests/infra/ccf.py
achamayou: I thought about it but it seemed best to do it separately because I'd like that to become the node status.

doc/users/issue_commands.rst
achamayou: Yes, thank you!

PR #1209, opened 2020-05-22
Websocket perf tests

achamayou:
Enable perf test for websockets.


eddyashton:
Can you resubmit this PR so we can get a clean perf graph comment from `ccf-bot`? Would like to confirm that we have a plausible baseline.

achamayou:
@eddyashton yes I'll do that, I thought I could delete it, but that won't kill the branch history,

PR #1210, opened 2020-05-22
Process new view

### LINKED ISSUE #459: Ledger Replay needs to take into account prepare and view change messages

olgavrou:
Resolves #459 

- consume new view messages during playback
- added end to end test that first suspends nodes to cause a view change, and then adds a late joiner to process the new view via ledger
- refactored some methods from `suspend_node.py` into `suspension.py` to be re-used in `replay_new_view.py`


ashamis:
There is a know bug I want to fix before we commit this. Let's hold off for a couple of days

achamayou:
@ashamis no problem, I'm holding off until you've approved it in any case.

Issue #1051, opened 2020-04-08
Resilience of recovery with shares


jumaffre:
Follow up from #51 (does not block the removal of sealing).

Now that we do not rely on the specific CPU to decrypt the ledger secrets, we should be fine with elections happening between the moment the first recovery share is submitted and the last one is (i.e. when the recovery of the private ledger starts). 

#51 was implemented in a way that is not resilient to elections as the submitted shares are only stored locally, on the primary node at-the-time. Instead, we should cache the submitted shares in a public table (since we are still in public mode when recovering) but encrypt the shares with a key known by all nodes in the network. That way, any (primary) node that receives the last recovery share will be able to retrieve and decrypt all the encrypted shares from the KV and initiate the recovery of the private ledger.

We should also ensure that the scheme handles the case when a member submits a bogus share as CCF will only find out after all the shares have been submitted. Do _all_ the members have to submit shares again? ~Perhaps we can do something clever-er where CCF re-encrypts the submit share with the submitting member's public key and check that the share is valid (against the content of the `ccf.shares` table).~ **It is not possible to re-encrypt the share on the new network as the network encryption private key of the defunct network is gone**

jumaffre:
The election part of this issue was implemented by #1212. However, if one or more submitted recovery shares were bogus, the share combination will fail when the last share is submitted and the submitted shares table will be cleared. Members can then resubmit their recovery shares.

In practice, implementing https://github.com/microsoft/CCF/issues/1096 will prevent members to submit bogus shares as a) the format of the submitted share will be consistently handled by the script and b) the `cryptobox` decryption will fail if any of the parameters is wrong.

If we want further resilience/convenience, we can always store member shares and try to perform k-of-n share combinations when the recovery threshold is reached.

jumaffre:
Closing this as it is very unlikely that members will submit a corrupted recovery share since we now ship the `submit_recovery_share.sh` script that makes use of authenticated decryption.

PR #1212, opened 2020-05-22
Resilient recovery

jumaffre:
Addresses most of https://github.com/microsoft/CCF/issues/1051

This PR adds support for elections during the recovery process. This was a current limitation of the current recovery share scheme whereby recovery shares were stored on the primary-at-the-time's node memory. 

In particular, this PR introduces:
- A new `ccf.submitted_shares` table to store the submitted recovery shares. This table has to be _public_ since the network is in public-only mode when the shares are being submitted. To preserve confidentiality, the submitted recovery shares are "manually" encrypted with the current ledger secret.
- The commit index at the of the public recovery (as decided by the primary-at-the-time) needs to be communicated to all nodes (via the join protocol) so that the ledger secrets can be initialised at the correct version at the end of the recovery protocol. 
- A new variant of the recovery test where the primary is killed just before the last recovery share is submitted.

I also refactored some code between the `member_frontend`, `node_state` and `share_manager` so that the member frontend can directly call the share_manager. This means that we can test the full flow of share submission from the `member_voting_test` unit test.


src/node/genesis_gen.h
```diff
@@ -398,10 +398,41 @@ namespace ccf
...
+      {
+        LOG_FAIL_FMT(
+          "Failed to set recovery threshold: service is currently waiting for "
+          "recovery shares");
+        return false;
+      }
+
+      if (threshold == 0)
+      {
+        LOG_FAIL_FMT("Recovery threshold cannot be set to 0");
```
eddyashton, line 420: Why?

src/node/genesis_gen.h
```diff
@@ -398,10 +398,41 @@ namespace ccf
...
+        return false;
+      }
+
+      auto active_members_count = get_active_members_count();
+      if (threshold > active_members_count)
+      {
+        LOG_FAIL_FMT(
+          "Recovery threshold cannot be set to {} as it is greater than "
+          "the number of active members ({})",
```
eddyashton, line 429: Minor: Is it worth enforcing this at this level? AFAICT, we don't do the same test in the opposite direction (ie - retiring members can move you below the recovery threshold). Can we reliably prevent members from putting the network in an unrecoverable state? If we can't, should we be suggesting (with errors like this) that we _are_ generally enforcing this?

CMakeLists.txt
```diff
@@ -461,7 +461,7 @@ if(BUILD_TESTS)
     NAME recovery_test
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/recovery.py
     CONSENSUS raft
-    ADDITIONAL_ARGS --recovery 2
+    ADDITIONAL_ARGS --recovery 2 --raft-election-timeout 4000
```
achamayou, line 464: Why do we need to make this adjustment now?

src/node/genesis_gen.h
```diff
@@ -398,10 +398,41 @@ namespace ccf
...
+
+      // While waiting for recovery shares, the recovery threshold cannot be
+      // modified. Otherwise, the threshold could be passed without triggering
+      // the end of recovery procedure
+      if (
+        get_service_status().value() ==
+        ServiceStatus::WAITING_FOR_RECOVERY_SHARES)
+      {
+        LOG_FAIL_FMT(
+          "Failed to set recovery threshold: service is currently waiting for "
```
achamayou, line 413: I think "Cannot adjust recovery threshold while waiting for recovery shares" would make it clearer that is a restriction/validation error rather than something that didn't work.

src/node/genesis_gen.h
achamayou: Re-assembling a key from 0 shares is an open research problem ;)

I think the consequence of the threshold being set to zero would be that the ledger secret is effectively stored in clear in the ledger. Or stored encrypted with a key that's never been shared with anyone and the network is effectively irrecoverable.

The latter seems like a possible, if uncommon use case?

src/node/genesis_gen.h
achamayou: I think we ought to try to validation on retiring members as well. Putting the network in an irrecoverable state is such a far-reaching decision that it should be a clearly explicit "make_network_irrecoverable" proposal if it's allowed at all, not something you can sleepwalk into.

src/node/node_state.h
```diff
@@ -966,6 +969,20 @@ namespace ccf
...
+      // Because submitted recovery shares are encrypted with the latest ledger
+      // secret, it is not possible to rekey the ledger if the service is in
+      // that state.
+      GenesisGenerator g(network, tx);
+      if (
+        g.get_service_status().value() ==
+        ServiceStatus::WAITING_FOR_RECOVERY_SHARES)
+      {
+        LOG_FAIL_FMT(
+          "Ledger could not be rekeyed while the service is waiting for "
```
achamayou, line 981: Likewise I would prefer "Cannot ...", which implies pre-condition validation rather than "could not" which sounds like an after-the-fact error. Maybe I'm being pedantic, deferring to @eddyashton.

src/node/genesis_gen.h
jumaffre: This is a design "constraint" of the underlying secret sharing library: it is not possible to share a secret in `N` shares with a recovery threshold of `0`.

src/node/genesis_gen.h
jumaffre: Good spot! I moved this code from `member_frontend.h` but did not move the opposite code to here. Just to clarify: we already do enforce that we never end up in a non-recoverable situation by checking that the number of active members is always `>=` than the recovery threshold.

tests/infra/consortium.py
```diff
@@ -315,22 +315,24 @@ def accept_recovery(self, remote_node):
...
+                decrypted_share = m.get_and_decrypt_recovery_share(
+                    remote_node, defunct_network_enc_pubk
+                )
+                r = m.submit_recovery_share(remote_node, decrypted_share)
+                submitted_shares_count += 1
+                check_commit(
+                    r,
+                    result=True
+                    if submitted_shares_count >= self.recovery_threshold
+                    else False,
```
achamayou:
```suggestion
                    result=(submitted_shares_count >= self.recovery_threshold)
```

CMakeLists.txt
jumaffre: Because a version of the recovery test now kills a node and waits for a new primary to be elected. The default for the raft election timeout is `100000`.

src/node/genesis_gen.h
eddyashton: I think a recovery threshold of 0 is trivially recoverable, but this is a plausible use case? The network offers no confidentiality, and it can be forked by any member (anyone who can start a new node and send a "try to recover" proposal, which will succeed without any shares). This is similar to a standard blockchain (you can fork whenever you like), but you get stronger binding of the _identity of the network_ which produced your favourite chain.

Of course you can almost do this with a threshold == 1, but then you're still restricted to 255 shares and members need to remember their share-decrypting keys; with 0 you could make everyone a member?

This is a blue-sky digression, feel free to ignore :)

CMakeLists.txt
achamayou, line 464:
```suggestion
    # Shorten Raft election timeout to speed up test when it kills a node on purpose
    # to check that a recovery network is robust to a view change.
    ADDITIONAL_ARGS --recovery 2 --raft-election-timeout 4000
```

src/node/genesis_gen.h
eddyashton: After some more discussion I'm convinced this trivially-forkable network is useless. The opposite interpretation (an unrecoverable network) may be applicable in some scenarios, but would require breaking more things (eg - code ID changes and governance changes) than just the recovery threshold.

Issue #1208, opened 2020-05-22
The write-set passed to KV commit-hooks doesn't clearly distinguish deletions


eddyashton:
Commit hooks take a `Write` argument, which is something like `std::map<K, VersionV<V>>`. The `VersionV` contains a version, and the written value, but the interpretation of these is not always clear.

It appears that this always contains either `{0, v}` for actual writes, or `{NoVersion, V{}}` for deletions. I'm not sure that all of our hooks are correctly checking whether the entry is a deletion, and this remains error-prone for future hooks. Additionally, we could tell the hook _what_ value was deleted, instead of just the key and a default-constructed value.

I think we should either:
- replace `VersionV<V>` with `std::optional<V>` (since the _version_ in this struct is meaningless) with `std::nullopt` indicating a deletion
- pass a list of deleted keys, or map of deleted {key, values}, as a separate argument from the `Write`

achamayou:
The first change sounds like an obvious win. The second is less clearly beneficial today because we presumably don't need that feature now and it potentially extends the lifetime of the value.

PR #1216, opened 2020-05-26
Add experimental KV type wrapper

eddyashton:
Opening as a draft while I clean up some TODOs.

Continues refactor of KV, adding `kv::experimental::` alternatives to `kv::` for `Map` and `TxView` types. The `experimental::` variants wrap untyped (ie - `std::vector<uint8_t> -> std::vector<uint8_t>`) instances of the old types, exposing the same interface (not just implementing the same `Abstract` types, but also matching template-discovered public typedefs) but implementing serialisation inside `get`/`put` etc.

The eventual goal is to use these maps as standard, so `kv::Map` will become `kv::untyped::Map`, and  `kv::experimental::Map` will become `kv::Map`, with no visible API changes. For now they are experimental while I test the compatibility, and next I'll do a rollout testing them in existing `Maps`.


src/kv/map.h
```diff
@@ -205,101 +199,40 @@ namespace kv
...
+    void set_commit_version(Version v)
     {
-      return map.is_replicated();
+      commit_version = v;
     }
   };
 
+  template <typename K, typename V, typename H>
+  struct ConcreteTxView : public TxViewCommitter<K, V, H>,
+                          public TxView<K, V, H>
```
eddyashton, line 221: May rename these for clarity.
- A `ChangeSet` is pure state: it holds some reads, writes, and versions, but has no logic for operating on them.
- A `TxView` lets us `get` and `put` changes, to interactively _modify_ a `ChangeSet`.
- A `TxViewCommitter` implements the `AbstractTxView` interface and deals with applying a `ChangeSet` to a `Map`.

This `ConcreteTxView` is both a `TxViewCommitter` (and transitively, `AbstractTxView`), and `TxView`: it can be used for either. But the _state_ it modifies is distinct, currently owned by `TxViewCommitter` and _referred_ to by `TxView` - it can also be modified by other things that look like a `TxView`.

src/kv/map.h
```diff
@@ -205,101 +199,40 @@ namespace kv
       }
     }
 
-    // Serialisation-related methods
-    void serialise(KvStoreSerialiser& s, bool include_reads) override
```
eddyashton, line 209: `serialise` and `deserialised` moved from `TxView` to `Map`. They deal only with a `ChangeSet` (which could be owned and populated by anyone) and `Map` state. There are a few important caveats to this, where we move checks that were inside these methods to the caller (roughly speaking, the caller knows and decides "does this contain interesting things I want to serialise", and potentially doesn't call, rather than the method having an early-out).

src/kv/map.h
```diff
@@ -391,6 +327,125 @@ namespace kv
...
+          s.serialise_remove(it->first);
+        }
+      }
+    }
+
+    AbstractTxView* deserialise(
+      KvStoreDeserialiser& d, Version version) override
+    {
+      // Create a new change set, and deserialise d's contents into it.
+      // TODO: Would really like this to be _only_ a TxViewCommitter, since we
```
eddyashton, line 408: This would have been a neat separation of concerns (when deserialising, you build the `ChangeSet` and then the only thing you want to do with the resulting 'view' is commit it, so you shouldn't need a modifiable view?), but is not possible in the current code - PBFT's approach needs to access a real, deserialised-but-not-committed `TxView` from the same `Tx`. I've added some errors if this type-expectation invariant is broken in `tx.h`, and will remove this comment.

src/kv/map.h
```diff
@@ -674,5 +689,42 @@ namespace kv
       std::swap(rollback_counter, map->rollback_counter);
       std::swap(roll, map->roll);
     }
+
+    template <typename TView>
+    TView* create_view(Version version)
```
eddyashton, line 694: `create_view` is now a template (so that `experimental::Map` can `create_view<experimental::TxView>`), and doesn't pretend to return an abstract type. The only caller is `tx.h`, which knows exactly what kind of view it expects - it casts from `AbstractTxView` immediately after calling. So we return a concrete (templated) type, and then confirm that's castable to the `AbstractTxView` we want to store.

src/kv/test/kv_serialisation.cpp
```diff
@@ -13,70 +14,42 @@
...
-DECLARE_JSON_TYPE(CustomClass)
-DECLARE_JSON_REQUIRED_FIELDS(CustomClass, m_i)
+  using StringString = kv::experimental::Map<std::string, std::string>;
+  using NumNum = kv::experimental::Map<size_t, size_t>;
+  using NumString = kv::experimental::Map<size_t, std::string>;
+  using StringNum = kv::experimental::Map<std::string, size_t>;
+};
 
-TEST_CASE(
+TEST_CASE_TEMPLATE(
```
eddyashton, line 33: Most of the test cases are now templated, so we test both `kv::Map` and `kv::experimental::Map` through the exact same code. When we rename (and perhaps `kv::Map` becomes untemplated), this will go away.

src/kv/test/kv_serialisation.cpp
```diff
@@ -88,30 +61,87 @@ TEST_CASE(
...
+    view0->put("pubk1", "pubv1");
+    auto view1 = tx.get_view(map1);
+    view1->put(0xab, 0xef);
+    REQUIRE(tx.commit() == kv::CommitSuccess::OK);
+    exp_data = consensus->get_latest_data().first;
+  }
+
+  // The new form is slightly longer but this is inevitable, and acceptable?
+  // TODO: Remove this test
+  // CHECK(raw_data == exp_data);
```
eddyashton, line 119: I was interested if the 2 implementations produce identical serialisations, but (at least for msgpack) they don't quite. Essentially we create an extra layer of nesting, since the 'raw' serialisation is now contained in a vector, and we include 'type' information in that vector. The difference is just a few bytes, so I think this is fine for now. In future, we can remove this by blitting the serialised Ks/Vs directly, and not calling `serialise_value` etc for them.

For comparison, here's the output of this test, highlighting the additional nested headers:

Old:
01 02 a4 6d 61 70 30 d3 80 00 00 00 00 00 00 00 00 01 a5 70 75 62 6b 31 a5 70 75 62 76 31 00 02 a4 6d 61 70 31 d3 80 00 00 00 00 00 00 00 00 01 cc ab cc ef 00
New:
01 02 a4 6d 61 70 30 d3 80 00 00 00 00 00 00 00 00 01 **c4 06** a5 70 75 62 6b 31 **c4 06** a5 70 75 62 76 31 00 02 a4 6d 61 70 31 d3 80 00 00 00 00 00 00 00 00 01 **c4 02** cc ab **c4 02** cc ef 00

src/kv/test/kv_serialisation.cpp
```diff
@@ -291,6 +382,137 @@ TEST_CASE(
   }
 }
 
+struct CustomClass2
```
eddyashton, line 385: Here's the real purpose of this PR:

`CustomClass2` has far fewer requirements than `CustomClass`. You don't need to `DECLARE_JSON_TYPE` and `MSGPACK_DEFINE` (you needed both in case `USE_NLJSON_SERIALISER` got turned on), it doesn't need to be hashable (just less-comparable). You may _want_ to define one of these things so you can use our standard serialisers, but you can also plug in your own.

src/kv/test/kv_serialisation.cpp
```diff
@@ -291,6 +382,137 @@ TEST_CASE(
...
+  static constexpr auto prefix = "This is a key:";
+};
+
+struct VPrefix
+{
+  static constexpr auto prefix = "Here follows a value:";
+};
+
+template <typename T>
+struct CustomVerboseDumbSerialiser
```
eddyashton, line 425: This is a bit over-complicated for test scaffolding. It is supposed to show both "you can serialise however you want" and "you can serialise your Ks and Vs differently, even when they're the same type", but it also shows "you can serialise in really stupid formats if you want".

src/kv/test/kv_test.cpp
```diff
@@ -131,26 +153,154 @@ TEST_CASE("Reads/writes and deletions")
     auto vc = view3->get(k);
     REQUIRE(!vc.has_value());
   }
+}
+
+TEST_CASE_TEMPLATE("foreach", MapImpl, RawMapTypes, ExperimentalMapTypes)
```
eddyashton, line 158: We were missing a lot of interesting cases of `foreach`, so I've added them here.

src/kv/tx.h
```diff
@@ -241,37 +248,34 @@ namespace kv
...
       auto e = map->get_store()->get_encryptor();
 
       KvStoreSerialiser replicated_serialiser(e, version);
-      // flags that indicate if we have actually written any data in the
-      // serializers
-      auto grouped_views = get_views_grouped_by_domain(view_list);
 
-      for (auto domain_it : grouped_views)
+      // Process in security domain order
+      for (auto domain : {SecurityDomain::PUBLIC, SecurityDomain::PRIVATE})
```
eddyashton, line 268: Replaced `get_views_grouped_by_domain` with this loop, which I think is equivalent. We iterate multiple times (here, and for `has_changes` above), and this could be flattened into a single pass building an ordered `view_list` of serialise-worth views, but we don't need to.

src/node/scripts.h
```diff
@@ -2,12 +2,16 @@
...
 #pragma once
 
-#include "kv/map.h"
+#include "kv/experimental.h"
 #include "script.h"
 
 namespace ccf
 {
-  using Scripts = kv::Map<std::string, Script>;
+  using Scripts = kv::experimental::Map<
```
eddyashton, line 10: Here's one example of using this in a real map, where we really want governance state in trivially-readable JSON rather than msgpack. In a future PR I'll do a pass like this on more governance tables, and remove the now unused msgpack serialisation machinery.

src/kv/experimental.h
```diff
@@ -0,0 +1,341 @@
...
+{
+  namespace experimental
+  {
+    using SerialisedRep = std::vector<uint8_t>;
+
+    // TODO: I don't think this needs to be customisable? If the map is storing
+    // _your_ types as keys, you might need to tell it how to compare them. But
+    // we know how to compare byte-vectors, and this is an internal detail so
+    // why would you change it?
+    using RepHasher = std::hash<SerialisedRep>;
```
achamayou, line 22: I'm not sure this is right, because for T t1, t2 where t1 == t2, it definitely seems possible that serialised(t1) != serialised(t2) (eg. "10.0" and "1e1"), as well as the opposite (NaN).

src/node/node_state.h
```diff
@@ -1507,45 +1502,49 @@ namespace ccf
...
+        [this](kv::Version version, const Nodes::Write& w) {
           auto configure = false;
-          std::unordered_set<NodeId> configuration;
+          std::unordered_set<NodeId> configuration =
+            consensus->get_latest_configuration();
 
           for (auto& [node_id, ni] : w)
           {
-            switch (ni.value.status)
+            if (kv::is_deleted(ni.version))
```
jumaffre, line 1512: Would nodes be ever deleted from the KV?

eddyashton: I think not currently, but every hook should still be checking for deletions. Maybe this should be an exception instead?

I'd like to do a pass making sure all hooks handle deletions correctly, but I'll do this with #1208.

jumaffre:
LGTM!

src/kv/test/kv_serialisation.cpp
achamayou: We'll want to replace this with our own/everparse framing anyway.

src/kv/experimental.h
eddyashton: For posterity, a brief summary of what's been discussed offline:

We agree the _hashing_ is an irrelevant detail we can hide from users, but we need to be precise with the equality implications of pluggable serialisers. Essentially we'd like injective serialisers, and this injectivity should be maintained across versions (ie - if you previously emit `{"foo": 1, "bar": 2}`, you shouldn't in future emit `{"bar": 2, "foo": 1}`).

For our serialisers, we'd like to eventually use verified code with a formal proof that it meets this property. For user-defined serialisers, we need to be clear about this in documentation: equality and storage deal only over the serialised forms, and these typed wrappers are just a convenience.

Issue #1147, opened 2020-05-06
code_update_test fails


jumaffre:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8042&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=efdbc34d-c6b6-5dc6-e05c-1751d4c47c20&l=3121

```
38: 2020-05-06 10:24:49.715 | INFO     | __main__:run:87 - Waiting 40.0s for a new primary to be elected...
38: 2020-05-06 10:25:29.752 | INFO     | infra.clients:log_request:147 - [127.210.253.201:56400] GET /nodes/getPrimaryInfo (node 3 (node))
38: 2020-05-06 10:25:29.764 | DEBUG    | infra.clients:log_response:160 - {'status': 200, 'result': {'primary_host': '127.109.220.177', 'primary_id': 6, 'primary_port': '42250'}, 'error': None, 'commit': 45, 'term': 4, 'global_commit': 45}
38: 2020-05-06 10:25:29.764 | INFO     | __main__:run:91 - Waited, new_primary is 6
38: 2020-05-06 10:25:29.764 | INFO     | __main__:run:93 - Adding another node to the network
38: 2020-05-06 10:25:29.765 | INFO     | infra.clients:log_request:147 - [127.210.253.201:56400] GET /nodes/getPrimaryInfo (node 3 (node))
38: 2020-05-06 10:25:29.776 | DEBUG    | infra.clients:log_response:160 - {'status': 200, 'result': {'primary_host': '127.109.220.177', 'primary_id': 6, 'primary_port': '42250'}, 'error': None, 'commit': 45, 'term': 4, 'global_commit': 45}
38: 2020-05-06 10:25:29.776 | INFO     | infra.remote:_rc:395 - [127.56.96.88] rm -rf /mnt/build/2/s/build/workspace/code_update_test_7
38: 2020-05-06 10:25:29.780 | INFO     | infra.remote:_rc:395 - [127.56.96.88] mkdir -p /mnt/build/2/s/build/workspace/code_update_test_7
38: 2020-05-06 10:25:29.785 | INFO     | infra.remote:_rc:395 - [127.56.96.88] ln -s /mnt/build/2/s/build/cchost /mnt/build/2/s/build/workspace/code_update_test_7/cchost
[...]

PR #1218, opened 2020-05-26
Switch to new playbook name

achamayou:
Also renamed old container to ccfciteam/ccf-ci-18.04-oe-0.9.0:az-dcap-1.4, built and pushed ccfciteam/ccf-ci-18.04-oe-0.9.0:az-dcap-1.5 which latest now points to as well.

There should be a nicer, automated way to do this. It's not totally obvious what that is.

### DIFF
--- a/docker/ccf_ci
+++ b/docker/ccf_ci
@@ -12,7 +12,7 @@ COPY getting_started/setup_vm/ /setup_vm/
 RUN apt update \
     && apt install -y ansible software-properties-common bsdmainutils dnsutils \
     && cd setup_vm \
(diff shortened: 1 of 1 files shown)


achamayou:
Renamed old container to ccfciteam/ccf-ci-18.04-oe-0.9.0:az-dcap-1.4, built and pushed ccfciteam/ccf-ci-18.04-oe-0.9.0:az-dcap-1.5 which latest now points to as well.

There should be a nicer, automated way to do this. It's not totally obvious what that is.

achamayou:
Build break is unrelated and tracked under #1147

PR #1221, opened 2020-05-27
Make it harder to mistake deletions for legitimate writes in commit hooks

### LINKED ISSUE #1208: The write-set passed to KV commit-hooks doesn't clearly distinguish deletions

eddyashton:
Fixes #1208.

The value type in a `Write` is now `std::optional<V>` instead of `VersionV<V>`. The version was supposed to be `NoVersion` to indicate a deletion, but all of our own hooks were ignoring `version` and accessing `value` directly. assuming it was a real written value. This is a breaking API change - even if you want to ignore the possibility of deletions, you now need to access the actual values by `.value()` (preferably gated by a check to `.has_value()`,) not `.value`.

Our hooks don't expect deletions and don't have anything sensible to do if they do occur, so they all simply throw an exception. It would be nice to have this expectation statically enforced (some kind of `NoRemovals` modifier type for a `kv::Map`, which returned `TxView`s with no `remove` method, and whose `Write` type could store values directly), but this is an unnecessary extra bit of machinery for now.

### DIFF
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -1317,16 +1317,31 @@ namespace ccf
       // inform the host of any nodes that no longer need to be tracked.
       network.nodes.set_global_hook(
         [this](kv::Version version, const Nodes::Write& w) {
-          for (auto& [node_id, ni] : w)
+          for (const auto& [node_id, opt_ni] : w)
           {
-            if (ni.value.status == NodeStatus::RETIRED)
+            if (!opt_ni.has_value())
+            {
+              throw std::logic_error("Unexpected: removal from nodes table");
+            }
+
+            const auto& ni = opt_ni.value();
+            if (ni.status == NodeStatus::RETIRED)
+            {
               remove_node(node_id);
+            }
           }
         });
 
       network.service.set_global_hook(
         [this](kv::Version version, const Service::Write& w) {
-          if (w.at(0).value.status == ServiceStatus::OPEN)
+          const auto it = w.find(0);
+          if (it == w.end() || !it->second.has_value())
+          {
+            throw std::logic_error(
+              "Unexpected: write to service table does not include key 0");
+          }
+
+          if (it->second.value().status == ServiceStatus::OPEN)
           {
             this->consensus->set_f(1);
             open_user_frontend();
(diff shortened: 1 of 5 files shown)


src/node/node_state.h
```diff
@@ -1507,38 +1535,37 @@ namespace ccf
...
             consensus->get_latest_configuration();
 
-          for (auto& [node_id, ni] : w)
+          for (const auto& [node_id, opt_ni] : w)
           {
-            if (kv::is_deleted(ni.version))
+            if (!opt_ni.has_value())
             {
-              configuration.erase(node_id);
+              throw std::logic_error("Unexpected: removal from nodes table");
```
achamayou, line 1542: Not likely this'll happen, but if it does there's a small chance we may be happy to have the node id.

src/node/node_state.h
```diff
@@ -1317,96 +1317,118 @@ namespace ccf
       // inform the host of any nodes that no longer need to be tracked.
       network.nodes.set_global_hook(
         [this](kv::Version version, const Nodes::Write& w) {
-          for (auto& [node_id, ni] : w)
+          for (const auto& [node_id, opt_ni] : w)
           {
-            if (ni.value.status == NodeStatus::RETIRED)
+            if (!opt_ni.has_value())
+            {
+              throw std::logic_error("Unexpected: removal from nodes table");
```
achamayou, line 1324: Not likely this'll happen, but if it does there's a small chance we may be happy to have the node id.

src/node/node_state.h
```diff
@@ -1317,96 +1317,118 @@ namespace ccf
...
+        bool has_secrets = false;
+        std::list<LedgerSecrets::VersionedLedgerSecret> restored_secrets;
 
-          for (auto& [v, secret_set] : w)
+        for (const auto& [v, opt_secret_set] : w)
+        {
+          if (!opt_secret_set.has_value())
           {
-            for (auto& encrypted_secret_for_node : secret_set.value.secrets)
+            throw std::logic_error("Unexpected: removal from secrets table");
```
achamayou, line 1362: Not likely this'll happen, but if it does there's a small chance we may be happy to have the version.

src/node/node_state.h
```diff
@@ -1669,14 +1696,18 @@ namespace ccf
       // map the node id to a hostname and service and inform pbft
       network.nodes.set_local_hook(
         [this](kv::Version version, const Nodes::Write& w) {
-          for (auto& [node_id, ni] : w)
+          for (const auto& [node_id, opt_ni] : w)
           {
-            add_node(node_id, ni.value.nodehost, ni.value.nodeport);
+            if (!opt_ni.has_value())
+            {
+              throw std::logic_error("Unexpected: removal from nodes table");
```
achamayou, line 1703: Not likely this'll happen, but if it does there's a small chance we may be happy to have the node id.

PR #1222, opened 2020-05-27
Make the foreach early-exit semantics consistent

eddyashton:
Noticed as part of #1221 (and one need manually merging when the other goes in, due to similar changes in `tx_view.h`):

`TxView::foreach` takes a visitor lambda, and returning false is supposed to mean "I'm done, don't call me again". We need to run this twice, for both the original state and the current write set. The early-out-on-false behaviour only applied to the writes - if you decided you were done after looking at something in the state, you still get called with something in the write set. This fixes that - return false means you won't get called again at all.

On a related point, `foreach` itself returned a bool with no meaning. It could return "the lambda returned false at some point", or "you saw everything", but none of these seem particularly clear or useful. For clarity, I've just removed the return type entirely.

++unit tests

### DIFF
--- a/src/kv/test/kv_test.cpp
+++ b/src/kv/test/kv_test.cpp
@@ -281,18 +281,55 @@ TEST_CASE_TEMPLATE("foreach", MapImpl, RawMapTypes, ExperimentalMapTypes)
 
   SUBCASE("Early termination")
   {
-    kv::Tx tx;
-    auto view = tx.get_view(map);
-    view->put("key1", "value1");
-    view->put("key2", "value2");
-    view->put("key3", "value3");
-    size_t ctr = 0;
-    view->foreach([&ctr](const auto& key, const auto& value) {
-      ++ctr;
-      return ctr <= 1; // Continue after the first, but not the second (so never
-                       // see the third)
-    });
-    REQUIRE(ctr == 2);
+    {
+      kv::Tx tx;
+      auto view = tx.get_view(map);
+      view->put("key1", "value1");
+      view->put("key2", "value2");
+      view->put("key3", "value3");
+      size_t ctr = 0;
+      view->foreach([&ctr](const auto& key, const auto& value) {
+        ++ctr;
+        return ctr < 2; // Continue after the first, but not the second (so
+                        // never see the third)
+      });
+      REQUIRE(ctr == 2);
+      REQUIRE(tx.commit() == kv::CommitSuccess::OK);
+    }
+
+    {
+      kv::Tx tx;
+      auto view = tx.get_view(map);
+      view->put("key4", "value4");
+      view->put("key5", "value5");
+
+      {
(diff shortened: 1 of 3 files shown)


src/kv/test/kv_test.cpp
```diff
@@ -281,18 +281,55 @@ TEST_CASE_TEMPLATE("foreach", MapImpl, RawMapTypes, ExperimentalMapTypes)
...
+          return ctr < 2; //< See only committed state
+        });
+        REQUIRE(ctr == 2);
+      }
+
+      {
+        size_t ctr = 0;
+        view->foreach([&ctr](const auto&, const auto&) {
+          ++ctr;
+          return ctr < 4; //< See mix of old and state
```
jumaffre, line 319: word missing?

src/kv/test/kv_test.cpp
```diff
@@ -281,18 +281,55 @@ TEST_CASE_TEMPLATE("foreach", MapImpl, RawMapTypes, ExperimentalMapTypes)
...
+      kv::Tx tx;
+      auto view = tx.get_view(map);
+      view->put("key4", "value4");
+      view->put("key5", "value5");
+
+      {
+        size_t ctr = 0;
+        view->foreach([&ctr](const auto&, const auto&) {
+          ++ctr;
+          return ctr < 2; //< See only committed state
```
jumaffre, line 310: I am not sure I understand this test. From looking at the next `foreach()`, it would continue and loop through all 5 keys?

src/kv/tx_view.h
```diff
@@ -229,29 +229,41 @@ namespace kv
...
       {
-        if (!is_deleted(write->second.version))
-          if (!f(write->first, write->second.value))
-            return false;
+        for (auto write = tx_changes.writes.begin();
+             write != tx_changes.writes.end();
+             ++write)
+        {
+          if (!is_deleted(write->second.version))
+            if (!f(write->first, write->second.value))
```
jumaffre, line 258: Minor: extra `{}` here maybe?

eddyashton: Yes: This is the awkward merge with #1221, just reverted to the original for now but I'll merge that to master first and then safely wrap this when integrating this PR.

eddyashton: Lol no this was a broken revert completely, hence the test being broken. Oops.

src/kv/test/kv_test.cpp
jumaffre: As discussed, this makes perfect sense now!

eddyashton: Discussed offline, hopefully clarified here for anyone reading in future: This test would fail under the previous code. The lambda would be called twice for the committed state (returning false the second time), and then _again for the first write_. So the counter would be 3, and our lambda was called after it returned false.

PR #1223, opened 2020-05-27
Use libc++ on cchost as well

jumaffre:
Also replaces OE's deprecated `oehostapp` with `oehost`.

### DIFF
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -168,15 +168,15 @@ include(${CCF_DIR}/cmake/sss.cmake)
 
 find_package(CURL REQUIRED)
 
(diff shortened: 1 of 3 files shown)


cmake/common.cmake
```diff
@@ -206,7 +207,11 @@ if("sgx" IN_LIST COMPILE_TARGETS)
             ${CRYPTO_LIBRARY}
             ${CMAKE_DL_LIBS}
             ${CMAKE_THREAD_LIBS_INIT}
-            openenclave::oehostapp
+            openenclave::oehost
+            -lc++
```
achamayou, line 211: We probably ought to have a ${USE_LIBCXX} that contains all of this.

jumaffre: Good point. Done now.

PR #1225, opened 2020-05-28
Use early-serialising KV implementation by default

eddyashton:
Following #1216, this:

- De-templates `kv::Map` (and `kv::TxView`), making them operate _only_ on `byte vector -> byte vector`
- Renames `kv::Map` (and `kv::TxView`) to `kv::untyped::` (and the files renamed to `kv/untyped_`)
- Renames `kv::experimental::Map` to `kv::Map` (and deletes now unused `experimental.h`)
- Makes all the governance tables use Json to write their Ks and Vs, rather than MsgPack

The resulting semantic change is that _all_ of our maps do early-serialisation/late-deserialisation (at the point of `put`/`get`, not at communication with the ledger). Additionally, our governance tables are more easily readable in the ledger, since they're stored as Json (this also means they're a bit bigger, with byte-vectors particularly enlarged - we should look at storing these as text where appropriate).

The resulting diff is awkward. Its _mostly_ copying and pasting between files, but complicated by some additional shuffling, and because `kv/map.h` has both been swapped out-of and swapped in-to. I'll try to clarify the changes inline.


src/apps/js_generic/js_generic.cpp
```diff
@@ -13,7 +14,7 @@ namespace ccfapp
   using namespace kv;
   using namespace ccf;
 
-  using Table = kv::Map<std::vector<uint8_t>, std::vector<uint8_t>>;
+  using Table = kv::untyped::Map;
```
eddyashton, line 17: This is the only instance of this, but its something we probably want: the JS app doesn't need typed wrappers at the C++ level, its tables are bytes-to-bytes. Serialisation can be handled within the JS code.

src/kv/change_set.h
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
eddyashton, line 1: This was the top of `tx_view.h`, now it has its own file.

src/kv/kv_types.h
```diff
@@ -31,6 +36,8 @@ namespace kv
   using BatchVector = std::vector<
     std::tuple<kv::Version, std::shared_ptr<std::vector<uint8_t>>, bool>>;
 
+  using SerialisedEntry = std::vector<uint8_t>;
```
eddyashton, line 39: This was `kv::experimental::SerialisedRep`. I decided `SerialisedEntry` was a clearer name, and it gets promoted to `kv::` as `experimental::` dies.

src/kv/store.h
```diff
@@ -147,12 +147,12 @@ namespace kv
...
      */
-    template <class K, class V, class H = std::hash<K>>
-    Map<K, V, H>& create(
+    template <class K, class V>
+    Map<K, V>& create(
       std::string name,
       SecurityDomain security_domain = kv::SecurityDomain::PRIVATE)
     {
-      return create<Map<K, V, H>>(name, security_domain);
+      return create<Map<K, V>>(name, security_domain);
```
eddyashton, line 155: Although this looks almost identical, it is creating a new type of `Map`. The third argument to `kv::Map` is now `KSerialiser`, not `Hasher`, so `H` is removed here.

src/kv/test/kv_test.cpp
```diff
@@ -15,73 +14,62 @@
...
-struct ExperimentalMapTypes
-{
-  using StringString = kv::experimental::Map<std::string, std::string>;
-  using NumNum = kv::experimental::Map<size_t, size_t>;
-  using NumString = kv::experimental::Map<size_t, std::string>;
-  using StringNum = kv::experimental::Map<std::string, size_t>;
-};
-
-TEST_CASE_TEMPLATE("Map creation", MapImpl, RawMapTypes, ExperimentalMapTypes)
+TEST_CASE("Map creation")
```
eddyashton, line 25: This file is just a regex replace: `TEST_CASE_TEMPLATE(XXX, ...)` becomes `TEST_CASE(XXX)`.

src/kv/tx_view.h
```diff
@@ -2,266 +2,83 @@
...
-      committed(committed_state),
-      start_version(current_version)
-    {}
-
-    ChangeSet(ChangeSet&) = delete;
-  };
-
-  template <typename K, typename V, typename H = std::hash<K>>
-  class TxView
+  template <typename K, typename V, typename KSerialiser, typename VSerialiser>
```
eddyashton, line 11: Useless diff. The left side is now `src/kv/untyped_tx_view.h`, minus the definitions that were pulled into `src/kv/change_set.h`. The right side was in `experimental.h`.

src/kv/untyped_tx_view.h
```diff
@@ -0,0 +1,220 @@
...
+  using VersionV = kv::VersionV<SerialisedEntry>;
+  using State =
+    kv::State<SerialisedEntry, SerialisedEntry, SerialisedKeyHasher>;
+  using Read = kv::Read<SerialisedEntry, SerialisedKeyHasher>;
+  using Write =
+    kv::Write<SerialisedEntry, SerialisedEntry, SerialisedKeyHasher>;
+  using ChangeSet =
+    kv::ChangeSet<SerialisedEntry, SerialisedEntry, SerialisedKeyHasher>;
+
+  class TxView
```
eddyashton, line 22: This was in `tx_view.h`, and now its here with no template.

src/node/certs.h
```diff
@@ -5,5 +5,5 @@
 
 namespace ccf
 {
-  using Certs = kv::Map<Cert, ObjectId>;
+  using Certs = kv::JsonSerialisedMap<Cert, ObjectId>;
```
eddyashton, line 8: This is what all the changes look like for our governance tables. If they were `kv::Map`, they'd write their Ks and Vs as msgpack, but we want Json so the type of the Map (and only the Map) changes.

We could now go through and remove a bunch of `MSGPACK_DEFINE`s - these types only need to be Json-able, not MsgPack-able. But maybe we'll change our minds, so the `MSGPACK` transformations can stay for now.

src/kv/untyped_map.h
```diff
@@ -0,0 +1,734 @@
...
+  };
+  using LocalCommits = snmalloc::DLList<LocalCommit, std::nullptr_t, true>;
+
+  struct Roll
+  {
+    std::unique_ptr<LocalCommits> commits;
+    size_t rollback_counter;
+  };
+
+  struct RollContainer
```
eddyashton, line 79: Template dependent lookup let us get away with some circular dependencies between `kv::Map<K, V>` and `kv::TxViewCommitter<K, V>`. This doesn't work now they're non-templated, `kv::Map` and `kv::untyped::TxView>`. To break the loop we create this little interface, which is how a `TxViewCommitter` sees its parent `Map`. No more friendship, just a vague `RollContainer`.

src/kv/untyped_map.h
```diff
@@ -0,0 +1,734 @@
...
+      Version v) :
+      TxViewCommitter(m, rollbacks, current_state, committed_state, v),
+      TxView(TxViewCommitter::change_set)
+    {}
+  };
+
+  class Map : public AbstractMap, public RollContainer
+  {
+  public:
+    using K = SerialisedEntry;
```
eddyashton, line 264: Although the Key and Value are the same type, I've retained the K and V names so its clear which we _intend_ to use within this class.

src/kv/test/kv_serialisation.cpp
```diff
@@ -236,112 +212,19 @@ TEST_CASE_TEMPLATE(
 }
 
 struct CustomClass
-{
```
eddyashton, line 239: This file is mostly a find and replace, plus deletions like this for things that only tested the old implementation.

src/kv/map.h
```diff
@@ -2,723 +2,202 @@
...
-    }
-  }
-
-  template <class K, class V, class H = std::hash<K>>
-  class Map;
-
-  template <class K, class V, class H>
-  class TxViewCommitter : public AbstractTxView
+  template <typename K, typename V, typename KSerialiser, typename VSerialiser>
+  class TypedMap : public AbstractMap
```
eddyashton, line 13: This was `kv::experimental::Map`, now its `kv::TypedMap`, providing the `kv::Map` replacement via the typedefs at the bottom of this file.

src/kv/map.h
```diff
@@ -2,723 +2,202 @@
...
+  using MsgPackSerialisedMap = TypedMap<
+    K,
+    V,
+    kv::serialisers::MsgPackSerialiser<K>,
+    kv::serialisers::MsgPackSerialiser<V>>;
+
+  // The default kv::Map will use msgpack serialisers. Custom types are
+  // supported through the MSGPACK_DEFINE macro
+  template <typename K, typename V>
+  using Map = MsgPackSerialisedMap<K, V>;
```
eddyashton, line 202: Specifying serialisers is slightly verbose, so we name some sensible defaults for you. And finally we provide `kv::Map<K, V>` again here. Note that this is a slight breaking change - anyone who had a custom hasher (ie `kv::Map<K, V, H>`) will get a compile error, and they'll have to remove it. We had no instances of this, and user custom types should no longer need it.

src/apps/js_generic/js_generic.cpp
achamayou: Yes, that's very much the intention!

src/node/certs.h
achamayou: Let's have a look at the cost of decoding the Cert from JSON on every lookup.

eddyashton:
@achamayou Agreed, this performance hit is unexpected. The certs may account for the e2e test, but clearly something's much slower in the standalone benchmark as well. Converting to Draft while I investigate this.

achamayou:
@eddyashton yeah no, it's not just the certs, there's something else for sure.

src/kv/serialise_entry_json.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <nlohmann/json.hpp>
+#include <vector>
+
+namespace kv::serialisers
+{
+  template <typename T>
+  struct JsonSerialiser
+  {
+    using Bytes = std::vector<uint8_t>;
```
achamayou, line 13: If we're going to have Bytes, we should probably have it somewhere common to at least the serialisers.

eddyashton:
Closing this PR and re-opening a clean replacement.

PR #1227, opened 2020-05-29
Finish the view.seqno renaming

### LINKED ISSUE #1187: Use consistent view.seqno terminology

achamayou:
fix #1187

Scrub the public APIs and RPCs to make them all aligned to the new transaction ID scheme.


src/node/history.h
```diff
@@ -526,21 +526,26 @@ namespace ccf
       {
         auto version = store.next_version();
         auto view = consensus->get_view();
-        auto commit = consensus->get_commit_seqno();
+        auto commit_txid = consensus->get_committed_txid();
         LOG_DEBUG_FMT("Issuing signature at {}", version);
         LOG_DEBUG_FMT(
-          "Signed at {} view: {} commit: {}", version, view, commit);
+          "Signed at {} in view: {} commit was: {}.{}",
```
achamayou, line 532: Note that `view` != `commit_view`, across view changes. `view` is the view as of the signature, and is useful to the view history replication across nodes. `commit_view` is the view of the last commit at the time of that signature (`commit_seqno`).

src/node/rpc/common_handler_registry.h
```diff
@@ -146,7 +145,7 @@ namespace ccf
         if ((nodes != nullptr) && (consensus != nullptr))
         {
           NodeId primary_id = consensus->primary();
-          auto current_term = consensus->get_view();
+          auto [current_term, _] = consensus->get_committed_txid();
```
achamayou, line 148: I'll revert this back to get_view(), I briefly thought we didn't need it anymore, but we do for signatures it turns out.

src/node/rpc/common_handler_registry.h
```diff
@@ -264,11 +263,8 @@ namespace ccf
       };
 
       install(GeneralProcs::GET_COMMIT, json_adapter(get_commit), Read)
+        .set_execute_locally(true)
```
achamayou, line 266: That's always what we want for commit.

tests/infra/ccf.py
```diff
@@ -596,12 +596,12 @@ def wait_for_node_commit_sync(self, consensus, timeout=3):
...
             for node in self.get_joined_nodes():
                 with node.node_client() as c:
-                    r = c.get("getCommit")
+                    r = c.get("commit")
                     commits.append(r.seqno)
             if [commits[0]] * len(commits) == commits:
                 break
             time.sleep(0.1)
-        # in pbft getCommit increments the commit version, so commits will not be the same
+        # in pbft the commit rpc increments the commit version, so commits will not be the same
```
eddyashton, line 604: This isn't true now that we're calling `set_execute_locally` for `commit`. If the PBFT tests work, I guess we just remove this comment?

achamayou: I think we can even make the check the same as for Raft.

achamayou: Yep.

Issue #691, opened 2020-01-13
PBFT code base cleanup


olgavrou:
Renamed issue to general pbft code base cleanup. We can add bullet points for things that come up.

- [x] Remove un necessary definitions from `parameters.h` and clean up `ifdefs` from all files under `libbyz`
- [x] `PBFT_ASSERT` to be `CCF_ASSERT`
- [x] Make all the `LOG_INFO <<`, `LOG_DEBUG <<`s into `LOG_INFO_FMT`, `LOG_DEBUG_FMT`, etc
- [x] Proactive recovery code from ePBFT
- [x] File names under libbyz should comply with the rest of the code base

olgavrou:
Renamed issue to general pbft code base cleanup. We can add bullet points for things that come up.

- [ ] Remove un necessary definitions from `parameters.h` and clean up `ifdefs` from all files under `libbyz`
- [ ] `PBFT_ASSERT` to be `CCF_ASSERT`
- [ ] make message tags and enum instead of `const shorts`
- [ ] put method comments before the method calls
- [ ] Make all the `LOG_INFO <<`, `LOG_DEBUG <<`s into `LOG_INFO_FMT`, `LOG_DEBUG_FMT`, etc

PR #1243, opened 2020-06-04
Clean up logging

### LINKED ISSUE #1239: Audit Logs
Audit all logs that are log level info or above to ensure they are not leaking information

Look at:
- [ ] Fatal logs
- [ ] Fail logs
- [ ] Info logs

ashamis:
- remove any data that should not be printed on a prod service that has a log level of Info+
- change to use LOG_*_FMT for fatal, fail, and info.

resolves #1239 
part of #691

### DIFF
--- a/src/consensus/pbft/libbyz/client_proxy.h
+++ b/src/consensus/pbft/libbyz/client_proxy.h
@@ -211,7 +211,7 @@ bool ClientProxy<T, C>::send_request(
   if (current_outstanding.fetch_add(1) >= Max_outstanding)
   {
     current_outstanding.fetch_sub(1);
-    LOG_FAIL_FMT("Too many outstanding requests, rejecting!");
+    LOG_FAIL_FMT("Too many outstanding requests, rejecting");
     return false;
   }
 
--- a/src/consensus/pbft/libbyz/client_proxy.h
+++ b/src/consensus/pbft/libbyz/client_proxy.h
@@ -211,7 +211,7 @@ bool ClientProxy<T, C>::send_request(
   if (current_outstanding.fetch_add(1) >= Max_outstanding)
   {
     current_outstanding.fetch_sub(1);
-    LOG_FAIL << "Too many outstanding requests, rejecting!" << std::endl;
+    LOG_FAIL_FMT("Too many outstanding requests, rejecting!");
     return false;
   }
 
@@ -242,7 +242,8 @@ bool ClientProxy<T, C>::send_request(
   }
   catch (const std::exception& e)
   {
-    LOG_FAIL_FMT("Failed to parse arguments, e.what:", e.what());
+    LOG_FAIL_FMT("Failed to parse arguments");
+    LOG_DEBUG_FMT("Failed to parse arguments, e.what:", e.what());
     return false;
   }
 
--- a/src/consensus/pbft/libbyz/message.cpp
+++ b/src/consensus/pbft/libbyz/message.cpp
@@ -80,9 +80,11 @@ void Message::set_size(int size)
   CCF_ASSERT(msg && ALIGNED(msg), "Invalid state");
   if (!(max_size < 0 || ALIGNED_SIZE(size) <= max_size))
   {
-    LOG_INFO << "Error - size:" << size
-             << ", aligned_size:" << ALIGNED_SIZE(size)
-             << ", max_size:" << max_size << std::endl;
+    LOG_INFO_FMT(
+      "Error - size:{}, aligned_size:{}, max_size:{}",
+      size,
+      ALIGNED_SIZE(size),
+      max_size);
   }
   CCF_ASSERT(max_size < 0 || ALIGNED_SIZE(size) <= max_size, "Invalid state");
   int aligned = ALIGNED_SIZE(size);
--- a/src/consensus/pbft/libbyz/node.cpp
+++ b/src/consensus/pbft/libbyz/node.cpp
@@ -71,8 +71,8 @@ Node::Node(const NodeInfo& node_info_) : node_info(node_info_)
     threshold = num_replicas - max_faulty;
   }
 
-  LOG_INFO << " max faulty (f): " << max_faulty
-           << " num replicas: " << num_replicas << std::endl;
+  LOG_INFO_FMT(
+    " max faulty (f): {} num replicas: {}", max_faulty, num_replicas);
 
   // Read authentication timeout
   int at = node_info.general_info.auth_timeout;
@@ -87,7 +87,7 @@ Node::Node(const NodeInfo& node_info_) : node_info(node_info_)
     PBFT_FAIL("Could not find my principal");
   }
 
-  LOG_INFO << "My id is " << node_id << std::endl;
+  LOG_INFO_FMT("My id is {}", node_id);
 
   // Initialize current view number and primary.
   v = 0;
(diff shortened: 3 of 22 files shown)


src/consensus/pbft/libbyz/client_proxy.h
```diff
@@ -211,7 +211,7 @@ bool ClientProxy<T, C>::send_request(
   if (current_outstanding.fetch_add(1) >= Max_outstanding)
   {
     current_outstanding.fetch_sub(1);
-    LOG_FAIL << "Too many outstanding requests, rejecting!" << std::endl;
+    LOG_FAIL_FMT("Too many outstanding requests, rejecting!");
```
jumaffre, line 214: Better to avoid `!`?

src/consensus/pbft/libbyz/node.cpp
```diff
@@ -88,7 +87,7 @@ Node::Node(const NodeInfo& node_info_) : node_info(node_info_)
     PBFT_FAIL("Could not find my principal");
   }
 
-  LOG_INFO << "My id is " << node_id << std::endl;
+  LOG_INFO_FMT("My id is {}", node_id);
```
jumaffre, line 90: The node ID gets written by CCF nodes on startup.

src/consensus/pbft/libbyz/pre_prepare.cpp
```diff
@@ -347,8 +347,7 @@ bool Pre_prepare::pre_verify()
             get_digest_sig().data(),
             rep().sig_size))
       {
-        LOG_INFO << "failed to verify signature on the digest, seqno:"
-                 << rep().seqno << std::endl;
+        LOG_INFO_FMT("failed to verify signature on the digest, seqno:{}", rep().seqno);
```
jumaffre, line 350: `LOG_FAIL_FMT()`?

src/consensus/pbft/libbyz/state.cpp
```diff
@@ -197,12 +197,11 @@ void Checkpoint_rec::dump_state(std::ostream& os)
...
 void Checkpoint_rec::print()
 {
-  LOG_INFO << "Checkpoint record: " << parts.size() << " blocks" << std::endl;
+  LOG_INFO_FMT("Checkpoint record: {} blocks", parts.size());
   for (auto const& p : parts)
   {
-    LOG_INFO << "Block: level= " << p.first.level << " index= " << p.first.index
-             << std::endl;
-    LOG_INFO << "last mod= " << p.second->lm << std::endl;
+    LOG_INFO_FMT("Block: level={}, index={}", p.first.level, p.first.index);
```
jumaffre, line 203: Combine in one line?

src/host/main.cpp
```diff
@@ -453,7 +453,8 @@ int main(int argc, char** argv)
   }
   catch (const std::logic_error& e)
   {
-    LOG_FATAL_FMT("{}. Exiting.", e.what());
+    LOG_DEBUG_FMT("{}. Exiting.", e.what());
```
eddyashton, line 456: I think we can keep this one, since its purely a host error?

src/node/rpc/frontend.h
```diff
@@ -424,7 +424,8 @@ namespace ccf
           // If serialising the committed transaction fails, there is no way
           // to recover safely (https://github.com/microsoft/CCF/issues/338).
           // Better to abort.
-          LOG_FATAL_FMT("Failed to serialise: {}", e.what());
+          LOG_DEBUG_FMT("Failed to serialise: {}", e.what());
+          LOG_FATAL_FMT("Failed to serialise");
```
eddyashton, line 428: Minor: Can we re-order these 2 be consistent with the others (`FAIL` then `DEBUG`).

src/ds/ccf_assert.h
```diff
@@ -56,7 +56,7 @@
 #  define PBFT_FAIL(msg) \
     do \
     { \
-      LOG_INFO << " FATAL_ERROR: " << (msg) << "\n"; \
+      LOG_INFO_FMT(" FATAL_ERROR: {}",(msg)); \
```
eddyashton, line 59: Nit: Should these be at `FAIL` or `FATAL` level, rather than `INFO`?

src/consensus/pbft/libbyz/client_proxy.h
achamayou: For what it's worth, these are short and reasonable: https://docs.microsoft.com/en-us/windows/win32/debug/error-message-guidelines?redirectedfrom=MSDN#style-considerations

https://docs.microsoft.com/en-us/windows/win32/uxguide/mess-error?redirectedfrom=MSDN is also insightful, even if a lot of it is only really relevant to pop up boxes.

In particular:

```
In contrast to the previous bad examples, good error messages have:

A problem. States that a problem occurred.
A cause. Explains why the problem occurred.
A solution. Provides a solution so that users can fix the problem.
Additionally, good error messages are presented in a way that is:

Relevant. The message presents a problem that users care about.
Actionable. Users should either perform an action or change their behavior as the result of the message.
User-centered. The message describes the problem in terms of target user actions or goals, not in terms of what the code is unhappy with.
Brief. The message is as short as possible, but no shorter.
Clear. The message uses plain language so that the target users can easily understand problem and solution.
Specific. The message describes the problem using specific language, giving specific names, locations, and values of the objects involved.
Courteous. Users shouldn't be blamed or made to feel stupid.
Rare. Displayed infrequently. Frequently displayed error messages are a sign of bad design.
```

src/consensus/pbft/libbyz/message.cpp
```diff
@@ -80,9 +80,11 @@ void Message::set_size(int size)
   CCF_ASSERT(msg && ALIGNED(msg), "Invalid state");
   if (!(max_size < 0 || ALIGNED_SIZE(size) <= max_size))
   {
-    LOG_INFO << "Error - size:" << size
-             << ", aligned_size:" << ALIGNED_SIZE(size)
-             << ", max_size:" << max_size << std::endl;
+    LOG_INFO_FMT(
```
achamayou, line 83: I'm a bit confused by this, should that be part of the assert message on the next line instead?

src/consensus/pbft/libbyz/pre_prepare.cpp
achamayou: Arguable in the case of PBFT...

src/ds/ccf_assert.h
achamayou: We don't then need the FATAL_ERROR bit.

src/consensus/pbft/libbyz/message.cpp
ashamis: ccf_assert currently cannot take a format string. #1246 tracks this.

src/kv/store.h
```diff
@@ -325,14 +325,16 @@ namespace kv
         auto search = maps.find(map_name);
         if (search == maps.end())
         {
-          LOG_FAIL_FMT("No such map {} at version {}", map_name, v);
+          LOG_FAIL_FMT("failed to deserialize");
```
achamayou, line 328: I don't mind either way, but I think we should either capitalise or not capitalise error messages.

src/consensus/pbft/libbyz/message.cpp
achamayou: Ah fair enough. It could still take a `fmt::format("something something")`, but building it in is probably better anyway.

Issue #1230, opened 2020-05-30
Lua end to end PBFT test failure


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8814&view=ms.vss-test-web.build-test-results-tab&runId=44556&resultId=100045&paneView=attachments

Overwhelmingly likely caused or revealed by https://github.com/microsoft/CCF/pull/1227, which switched to using a "local" commit RPC and therefore tightened the "sync check" assertion.

eddyashton:
Here's another repro, with the more detailed assert:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8911&view=ms.vss-test-web.build-test-results-tab&runId=44874&resultId=100044&paneView=attachments

eddyashton:
And again, with identical versions:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8958&view=ms.vss-test-web.build-test-results-tab&runId=44970&resultId=100044&paneView=attachments

achamayou:
@ashamis I think this is an existing bug revealed by the stricter check, rather than a problem with the check itself. One of the nodes is consistently failing to catch up and is stuck at 2.46 when other nodes all reach 2.62.

PR #1245, opened 2020-06-04
Support custom serialisation formats when parsing ledger

eddyashton:
Quick followup to #1234:

That PR added all the machinery for serialising Ks and Vs in user-defined formats, but in practice everything was msgpack and so we could parse a ledger as (mostly) a stream of msgpack objects. This removes that restriction by length-prefixing Ks and Vs in the ledger entries. If you know how the table was serialised you can parse it, otherwise you can skip it/store it as bytes.


jumaffre:
We've still got [this rogue documentation page](https://microsoft.github.io/CCF/master/developers/kv/kv_serialisation.html) that describes the format of the ledger. I guess it needs updating?

tests/infra/ledger.py
```diff
@@ -1,9 +1,11 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache 2.0 License.
 import io
-import msgpack
+import msgpack.fallback as msgpack  # Default implementation has buggy interaction between read_bytes and tell, so use fallback
```
eddyashton, line 4: https://github.com/msgpack/msgpack-python/issues/426

src/kv/msgpack_serialise.h
```diff
@@ -102,14 +93,34 @@ namespace kv
 
     kv::serialisers::SerialisedEntry read_next_pre_serialised()
     {
-#if MSGPACK_DONT_REPACK
+      // Check there are sufficient bytes for a size-prefix
+      constexpr auto size_size = sizeof(size_t);
```
achamayou, line 97: serialized::read() does this nicely.

src/kv/msgpack_serialise.h
```diff
@@ -102,14 +93,34 @@ namespace kv
...
+      if (remainder < size_size)
+      {
+        throw std::runtime_error(fmt::format(
+          "Expected {} bytes holding entry size, found only {}",
+          size_size,
+          remainder));
+      }
+
+      // Read the size-prefix
+      const auto entry_size =
```
achamayou, line 108: Same as above.

src/kv/msgpack_serialise.h
eddyashton: Sure, I've replaced the size-prefix parsing code with `serialized::read`. We need a little wrapping code around this, since this `MsgPackReader` is storing an incrementing `data_offset`, which has to be adjusted to match the decrementing `remainder` we pass to `serialized::read`. Hope the distinction is clear.

eddyashton:
CI failure looks unrelated, known issue #1230. Re-running to confirm.

PR #1252, opened 2020-06-05
Term in kv

### LINKED ISSUE #1163: Incorrect transaction id reporting when using multiple threads

achamayou:
fix #1163

In a nutshell, this change means that the KV is always made aware of what `view` (*) transactions are executing in. This enables accurate Transaction ID reporting in a multi-threaded context.

(*) although it really is the `view`, and that's how it's communicated externally to users, the code deliberately refers to it as `term` internally, because there's already a concept called a `view` in the KV, which is really a `TxTableView`. Naming is hard. I'm open to suggestions here.


tests/infra/remote.py
```diff
@@ -607,9 +606,6 @@ def __init__(
                 f"--notify-server-address={notify_server_host}:{notify_server_port[0]}"
             ]
 
-        if domain:
```
jumaffre, line 610: Looks like a merge issue undid some changes here?

tests/e2e_scenarios.py
```diff
@@ -38,7 +38,7 @@ def run(args):
             check = infra.checker.Checker()
             check_commit = infra.checker.Checker(mc)
             with primary.user_client() as uc:
-                check_commit(uc.rpc("mkSign"), result=True)
+                uc.rpc("mkSign")
```
eddyashton, line 41: Why are we no longer waiting for commit here?

tests/infra/ccf.py
```diff
@@ -72,7 +72,6 @@ class Network:
         "gov_script",
         "join_timer",
         "worker_threads",
-        "domain",
```
eddyashton, line 75: This reverts #1229 - is that deliberate?

samples/perf_client/perf_client.h
```diff
@@ -524,23 +527,28 @@ namespace client
...
 
       return response;
     }
 
+    RpcTlsClient::Response get_tx_status(
+      const std::shared_ptr<RpcTlsClient>& connection,
+      size_t view,
+      size_t seqno)
+    {
+      // Send a mkSign RPC to trigger next global commit
```
eddyashton, line 545: This comment is inaccurate

src/consensus/raft/raft_types.h
```diff
@@ -65,6 +66,13 @@ namespace raft
       if (p)
         p->rollback(v);
     }
+
+    void set_term(Term t)
+    {
+      auto p = x.lock();
```
eddyashton, line 72: Why does the PBFT equivalent loop until this succeeds, while this version only tries once?

src/kv/tx.h
```diff
@@ -54,7 +55,9 @@ namespace kv
       if (read_version == NoVersion)
       {
         // Grab opacity version that all Maps should be queried at.
-        read_version = m.get_store()->current_version();
+        auto [t, rv] = m.get_store()->current_term_and_version();
+        term = t;
```
eddyashton, line 59: Just trying to make sure I understand this properly: we're setting up a long-range invariant here, that the term when we create our first `TxView` will match the term when our transaction is committed. We don't explicitly check/enforce this. In most situations I think this holds (a term change will cause a transaction to fail on commit, or to be discarded from `pending_txs`).

The potential hole I see is if we've put this serialised transaction into a batch, and we're just about to call `replicate` with it, but there's an election and the term changes. If replicate can succeed, we'll report to the user that this transaction was `2.5`, but we'll replicate it to other nodes in term 3. I think this is still fine - usually the node is no longer the `Leader` so replicate fails, and even if it remains `Leader` it has probably rolled back so will not accept this batch as the next index. But is this narrow case possible, where we have a locally committed write about to be replicated, there's an election but we become `Leader`, no other writes happen, and so we replicate this transaction in the wrong term?

src/kv/tx.h
```diff
@@ -239,6 +238,21 @@ namespace kv
       return version;
     }
 
+    /** Commit term if committed
+     *
+     * @return Commit term
+     */
+    Version commit_term()
```
jumaffre, line 245: Minor: `get_commit_term()`?

tests/e2e_logging.py
```diff
@@ -209,18 +209,18 @@ def test_raw_text(network, args):
...
     primary, backup = network.find_primary_and_any_backup()
 
     with primary.node_client() as nc:
         check_commit = infra.checker.Checker(nc)
-        with backup.node_client() as c:
-            check_commit(c.rpc("mkSign"), result=True)
         with backup.member_client() as c:
-            check_commit(c.rpc("mkSign"), result=True)
+            check_commit(
+                c.rpc("read", {"table": "ccf.service", "key": 0}),
```
jumaffre, line 221: Does this test forwarding if the RPC is installed as `Read`?

src/kv/tx.h
jumaffre: > We don't explicitly check/enforce this.

I assume that this will be enforced once the transaction term is passed to `store->commit()` (see `TODO` on line `209`).

tests/infra/remote.py
achamayou: This hasn't been rebased on a recent master, yes.

samples/perf_client/perf_client.h
achamayou: Yes, copy paste, will clear, thanks.

src/consensus/raft/raft_types.h
achamayou: I don't know, I've just kept it consistent with the rest of the adaptor, @ashamis?

tests/e2e_scenarios.py
achamayou: I'm not sure why we even trigger a signature here, and as you know, we cannot usefully wait for this particular commit. What we could do is sample commit, issue a sig, and wait for it to go up. Unless someone remembers why we issue a sig here, I'm just going to remove it.

tests/infra/ccf.py
achamayou: Old master, will rebase.

src/kv/tx.h
achamayou: @eddyashton @jumaffre I agree this is a hole, and the validation I am thinking of is that we should pass the term to store->commit(), and that under the version lock, we check the term of inbound commits versus current term. If they don't match, there's been a term change and we drop the tx (and return NO_REPLICATE).

Thoughts?

jumaffre: > ... versus current term.

Is that retrieved from the consensus `r` directly? If so, this was my understanding too.

eddyashton: That explicit validation sounds good, but I'm not sure it catches the final hole I described. We hold the version lock while building the batch, but deliberately release it before calling `replicate()`. So we can get a term match while building the batch, but that doesn't mean the term is still 'correct' when we call `replicate()`.

src/kv/tx.h
achamayou: We have commit_version() above, so commit_term() I think.

tests/e2e_logging.py
achamayou: Good point. I think that should be updateAck then.

src/kv/tx.h
achamayou: Yes, although as discussed we don't need to store the term per-batch because we can:

1. reject Store::commit()s for a term that does not match current
2. send the current term to replicate(), which can in turn do the same, and yied NO_REPLICATE

src/consensus/raft/raft_types.h
achamayou: poke @ashamis

ashamis: Sorry just saw this :(

Why does the raft version not spin? x.lock() can fail so the term would then not be set and this happens silently?

achamayou: I'll fix that, even the pbft adapter is doing this inconsistently (see calls up and down).

achamayou: https://github.com/microsoft/CCF/pull/1283

achamayou:
Lives on as #1288 plus a bunch of test-updates-not-to-use-non-rpc-tx-seqno spinoff PRs.

PR #1255, opened 2020-06-05
Upgrade snmalloc to 0.4.1

### LINKED ISSUE #1253: Upgrade snmalloc

ashamis:
resolves #1253


src/ds/dl_list.h
```diff
@@ -6,6 +6,8 @@
 #include <cstdint>
 
 #define SNMALLOC_ASSERT assert
```
achamayou, line 8: I don't think we should have those, instead we should include the right PAL/AAL and get the right includes.

PR #1259, opened 2020-06-05
Add option to change console to be json

### LINKED ISSUE #1258: Add option to push json logging to console

ashamis:
resolves #1258


src/host/main.cpp
```diff
@@ -142,6 +142,8 @@ int main(int argc, char** argv)
     json_log_path,
     "Path to file where the json logs will be written");
 
+  bool json_log_std_out = false;
+  app.add_flag("--json-std-out", json_log_std_out, "send json to std out");
```
jumaffre, line 146: Nitpick: Better to move the option next to `--json-log-path`.  And also perhaps rename it to `--log-format-json` and description to "Set node stdout log format to JSON".

src/ds/logger.h
```diff
@@ -129,12 +132,21 @@ namespace logger
...
       dump(log_line);
     }
   };
 
+  class JsonConsoleLogger : public JsonLogger
+  {
+  public:
+    void write(const std::string& log_line) override
+    {
+      std::cout << log_line << std::flush;
```
achamayou, line 146: Do we want to force a flush for every line? That's usually done for stderr but not stdout.

src/ds/logger.h
```diff
@@ -129,12 +132,21 @@ namespace logger
...
+    virtual void write(const std::string& log_line) override
     {
       dump(log_line);
     }
   };
 
+  class JsonConsoleLogger : public JsonLogger
+  {
+  public:
+    void write(const std::string& log_line) override
```
eddyashton, line 144: I think we just need to set the `ostream f` member to `std::cout`? So rather than overriding write, we could set `f` in a constructor for `JsonConsoleLogger`.

src/host/main.cpp
```diff
@@ -142,6 +142,11 @@ int main(int argc, char** argv)
     json_log_path,
     "Path to file where the json logs will be written");
 
+  bool log_format_stdout_json = false;
```
eddyashton, line 145: It seems odd to have separate options for "format as JSON in std out", and "format as JSON to this file". Can we remove the file option, and if someone wants this they can redirect stdout? And if we have a use case for multiple loggers, we can make these fully configurable from the CLI (ie - give a file _and_ a format type, and maybe a log level, and...)

PR #1261, opened 2020-06-06
Backtrace and other cleanup

### LINKED ISSUE #1246: Debugging improvements
- [ ] ccf_assert should be able to take a format string
- [ ] create a new exception that prints a callstack as part of its what() function

ashamis:
- Add backward-cpp to replace code that we wrote that does a subset of the same thing
- Add ccf_logic_error an exception that captures a backtrace when it is created

resolves #1246
part of #691


src/ds/ccf_exception.h
```diff
@@ -0,0 +1,47 @@
...
+namespace ccf
+{
+  class ccf_logic_error : public std::exception
+  {
+  public:
+    ccf_logic_error(const std::string& what_arg)
+    {
+      if (!what_arg.empty())
+      {
+        result.append(what_arg.c_str());
```
achamayou, line 22: fmt::format(“{}\n”, what_arg)

PR #1264, opened 2020-06-08
Hash bench

### LINKED ISSUE #1249: hash_bench with SAN timing out
2020-06-04T20:37:36.2561523Z test 34
2020-06-04T20:37:36.2561786Z       Start 34: hash_bench
2020-06-04T20:37:36.2561989Z 
2020-06-04T20:37:36.2563488Z 34: Test command: /bin/bash "-c" "/mnt/build/2/s/build/hash_bench --samples=1000 --out-fmt=csv --output=hash_bench.csv && cat hash_bench.csv"
2020-06-04T20:37:36.2564457Z 34: Test timeout computed to be: 300
2020-06-04T20:42:36.3509403Z 34/60 Test #34: hash_bench .......................***Timeout 300.10 sec

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=8944&view=logs&j=2f246a49-43ac-5250-d63c-293adb2ef928&t=2bd49aea-a0ec-5c63-8430-09adb0b47aca

ashamis:
resolves #1249

We should run less iterations with hash_bench when using the SAN


src/ds/test/hash_bench.cpp
```diff
@@ -26,7 +26,11 @@ static void hash(picobench::state& s)
   }
 }
 
-const std::vector<int> hash_sizes = {1, 8, 64, 1024, 16536};
+#ifdef SAN
```
jumaffre, line 29: There probably is something I'm missing but what is the value in running micro-benchmarks with SAN on?

eddyashton:
Why are we running the benchmarks in a SAN build?

We're already excluding the perf tests, I believe we should exclude the benchmark label as well. This is combined with our coverage so _may_ drop our coverage numbers slightly, but those 'should' be covered by unit tests.

And I don't think these picobenchmarks will uncover any interesting SAN issues, since they're not multi-threaded and take a static execution path every time.

PR #1266, opened 2020-06-08
Correctly handle oversized compound messages in ringbuffer

eddyashton:
@jumaffre spotted there was an issue here.

In practice, currently all of our oversized messages (which get split into multiple fragments) consist of a single byte-vector. There was a bug if you tried to write something like `<size_t, bool, std::vector<uint8_t>>`, where we wouldn't track that part of the first fragment was used, so we'd never finalise the resulting message. This fixes that, and adds a unit test to check it.


src/ds/serializer.h
```diff
@@ -311,8 +311,19 @@ namespace serializer
...
+      using StrippedT = details::remove_cvref_t<T>;
+
+      if constexpr (
+        std::is_same_v<StrippedT, std::vector<uint8_t>> ||
+        std::is_same_v<StrippedT, ByteRange>)
+      {
+        static_assert(
+          sizeof...(Ts) == 0,
+          "Byte vectors must be the final element in message");
+      }
```
eddyashton, line 323: This is a vaguely related static assert to enforce a restriction that was a) unenforced, and b) I'd completely forgotten about. We don't size prefix byte ranges, we assume they take up all the space left in the message. We could waste some space by size-prefixing them all, or do a complicated check where only mid-sequence ones are size-prefixed, but since we don't actually have any messages that need this yet here's the easy solution - disallow it with a static_assert.

Issue #1207, opened 2020-05-22
Support for historical queries


jumaffre:
It should be possible for users to query the service to get the value of a specific key at a given version.

_More details to follow..._

eddyashton:
Summarising some early design discussions on this:

- A node can store old ledger entries and materialised instances of the KV in-memory, which do not interact with the 'current' KV used for consensus.
- To verify the integrity of a KV reconstructed from ledger entries read from the host, we need to replay from the last signature _before_ this version to the first signature _after_ it. We must check the validity of the latter signature, since the entries now come from an untrusted host (normally, they'd come over an integrity-protected channel directly from a trusted node).
- Our current method of fetching ledger entries uses the ringbuffer and is asynchronous. We cannot do this within an RPC. Either we need to add an ECALL to read these, or we need to pre-fetch targeted historical entries. The latter seems preferable, making these requests 2-stage and adding a clear "Not fetched yet" state while entries are slowly fetched/replayed.
- The simplest approach is to make this an entirely node-local system. Eventually we may look at coordination across nodes, with each responsible for some range of historical versions, to quickly service historical read requests.
- We should establish whether the common use case is querying a _single_ version or a _range_ of versions, and make sure we efficiently maintain the latter if needed.

jumaffre:
With regards to "we need to replay from the last signature before this version", this will involve either fetching the signature closing the previous chunk or keeping a flag indicating whether each transaction in the ledger is a signature.

achamayou:
I think we're going to want the flag for many reasons. We could also decide to duplicate transactions across chunks, and always have the last Tx in a chunk be copied at the beginning of the next. It'd make chunks verifiable/auditable on their own, it's slightly wasteful but practical in other ways. To be discussed.

eddyashton:
Historical queries are supported since #1267.

Future improvements are tracked in other issues:
 - #1237 will allow us to request the desired signature directly, rather than searching linearly for it.
 - #1269 will reduce memory consumption by only cloning required tables.
 - [Snapshotting](https://github.com/microsoft/CCF/milestone/12) may allow us to serve a more complete historical state, of all writes _visible at_ rather than merely _produced by_ the target version.

PR #1267, opened 2020-06-09
Add a system for retrieving, verifying, and caching historical state in each node

eddyashton:
Initial work for #1207.

Opening as a draft for early feedback - still has several `TODO`s that I need to clean up.

This is the first stage of historical query support. It adds a `ccf::historical::StateCache`, which can be asked to create a `kv::Store` at a specific version. There's a basic unit test (the majority of the test body is mocking out all of the necessary surrounding infrastructure, but the tail checks that something 'works'), and a sample RPC in the logging app that shows how to access it. The access is still very manual and custom at the moment, will need some thought to how we can hide/generalise this - not an immediate priority.


samples/apps/smallbank/app/smallbank.cpp
```diff
@@ -486,7 +486,7 @@ namespace ccfapp
   };
 
   std::shared_ptr<ccf::UserRpcFrontend> get_rpc_handler(
-    NetworkTables& nwt, AbstractNotifier& notifier)
+    NetworkTables& nwt, ccfapp::AbstractNodeContext& context)
```
eddyashton, line 489: Making the historical state cache available is a breaking change for all apps. Rather than adding an argument every time we add a system like this (changing the signature all apps need to implement), I've wrapped these in a single `context`. New services are exposed as new functions on this context, but apps that don't use them can upgrade with no changes.

src/kv/store.h
```diff
@@ -377,31 +410,35 @@ namespace kv
...
+          {
             if (!h->verify(term))
             {
               LOG_FAIL_FMT("Failed to deserialize");
               LOG_DEBUG_FMT("Signature in transaction {} failed to verify", v);
               return DeserialiseSuccess::FAILED;
             }
-            success = DeserialiseSuccess::PASS_SIGNATURE;
           }
+          success = DeserialiseSuccess::PASS_SIGNATURE;
```
eddyashton, line 437: To clarify the changes here - it is now possible to return `PASS_SIGNATURE` when there is no `history`. If there _is_ a history we will use it to verify the signature, but in unit tests where there isn't we can accept stub signatures.

src/apps/logging/logging.cpp
```diff
@@ -257,6 +259,81 @@ namespace loggingapp
...
+        kv::Tx historical_tx;
+        auto view = historical_tx.get_view(*historical_map);
+        const auto v = view->get(in.id);
+
+        if (v.has_value())
+        {
+          args.rpc_ctx->set_response_status(HTTP_STATUS_OK);
+          args.rpc_ctx->set_response_header(
+            http::headers::CONTENT_TYPE, http::headervalues::contenttype::TEXT);
+          auto s = v.value();
```
ashamis, line 328: should this be auto& s, to avoid a copy here?

src/apps/logging/logging.cpp
```diff
@@ -257,6 +259,81 @@ namespace loggingapp
       };
       // SNIPPET_END: log_record_text
 
+      auto& historical_state = context.get_historical_state();
```
ashamis, line 262: This changes the pattern of how the handlers are created, instead of taking this it now takes this and historical_state.

Why not make historical_state part of this? I think it has the same life span

src/apps/logging/logging.cpp
```diff
@@ -257,6 +259,81 @@ namespace loggingapp
...
+      auto get_historical = [this, &historical_state](ccf::RequestArgs& args) {
+        const auto params =
+          nlohmann::json::parse(args.rpc_ctx->get_request_body());
+        const auto in = params.get<LoggingGetHistorical::In>();
+
+        // Check that the requested transaction ID is committed
+        {
```
ashamis, line 269: This seems like boilier plate that will be written for every handler that implements a historical query. Can this be moved into a helper method?

src/apps/logging/logging.cpp
```diff
@@ -257,6 +259,81 @@ namespace loggingapp
...
+              in.view,
+              in.seqno,
+              ccf::tx_status_to_str(tx_status)));
+            return;
+          }
+        }
+
+        auto historical_store = historical_state.get_store_at(in.seqno);
+        if (historical_store == nullptr)
+        {
```
ashamis, line 294: same boiler plate comment

src/node/historical_queries.h
```diff
@@ -0,0 +1,341 @@
...
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "consensus/ledger_enclave_types.h"
+#include "kv/store.h"
+#include "node/historical_queries_interface.h"
+#include "node/history.h"
+#include "node/rpc/node_interface.h"
+
+#include <deque>
```
ashamis, line 11: can this be removed?

src/node/historical_queries.h
```diff
@@ -0,0 +1,341 @@
...
+      pending_fetches.insert(idx);
+    }
+
+    std::optional<ccf::Signature> get_signature(const StorePtr& sig_store)
+    {
+      kv::Tx tx;
+      auto sig_table = sig_store->get<ccf::Signatures>(ccf::Tables::SIGNATURES);
+      if (sig_table == nullptr)
+      {
+        throw std::logic_error(
```
ashamis, line 87: ccf_logic_error will also capture the callstack

src/node/historical_queries.h
```diff
@@ -0,0 +1,341 @@
...
+
+      auto verifier = tls::make_verifier(node_info->cert);
+      const auto verified = verifier->verify_hash(
+        real_root.h.data(),
+        real_root.h.size(),
+        sig->sig.data(),
+        sig->sig.size());
+      if (!verified)
+      {
+        throw std::logic_error(
```
ashamis, line 151: Do we want to crash the service if there is an issue with the data that we read from outside the enclave?

src/node/historical_queries.h
```diff
@@ -0,0 +1,341 @@
...
+    void deserialise_ledger_entry(
+      consensus::Index idx, const LedgerEntry& entry)
+    {
+      StorePtr store = std::make_shared<kv::Store>();
+
+      store->set_encryptor(source_store.get_encryptor());
+
+      // TODO: Add a lazy clone option?
+      store->clone_schema(source_store);
+      store->set_strict_versions(false);
```
ashamis, line 218: Since this is set once at construction time and never reset can this flag be passed to the ctor?

src/node/historical_queries.h
```diff
@@ -0,0 +1,341 @@
...
+
+      const auto deserialise_result = store->deserialise_views(entry);
+
+      switch (deserialise_result)
+      {
+        case kv::DeserialiseSuccess::FAILED:
+        {
+          // TODO: Host gave us junk? Do we fail silently?
```
ashamis, line 226: I vote against this. 
This seems to be on the same level of crashing if we get a malformed message

src/node/historical_queries.h
```diff
@@ -0,0 +1,341 @@
...
+      const auto request_it = requests.find(idx);
+      if (request_it != requests.end())
+      {
+        if (request_it->second.current_stage == RequestStage::Fetching)
+        {
+          requests.erase(request_it);
+        }
+      }
+
+      const auto it =
```
ashamis, line 330: can this be replaced with pending_fetches.erase(idx)?

src/apps/logging/logging.cpp
```diff
@@ -257,6 +259,81 @@ namespace loggingapp
...
+          args.rpc_ctx->set_response_body(fmt::format(
+            "Unable to get historical table '{}'.", records.get_name()));
+          return;
+        }
+
+        kv::Tx historical_tx;
+        auto view = historical_tx.get_view(*historical_map);
+        const auto v = view->get(in.id);
+
+        if (v.has_value())
```
ashamis, line 323: Just a question. At this point the historical query has been answered, show it be deleted from memory to save EPC?

src/node/historical_queries.h
eddyashton: We can't avoid throwing if the host gives us malformed data - `deserialise_views` will throw on many kinds of malformed data, the merkle tree can only report failures by throwing, so its no more dangerous to throw here as well. But I agree this shouldn't propagate up to crashing the node, so I've added a catch for these failures at the top-level `handle_ledger_entry`.

src/node/historical_queries.h
```diff
@@ -0,0 +1,341 @@
...
+    using LedgerEntry = std::vector<uint8_t>;
+
+    struct Request
+    {
+      RequestStage current_stage = RequestStage::Fetching;
+      crypto::Sha256Hash entry_hash = {};
+      StorePtr store = nullptr;
+    };
+
+    // TODO: Make this an LRU, evict old entries
```
ashamis: Question: Is LRU the right thing to do here? Consider doing something time based, it might be better to be slow rather than fail to return results if (MAX_ACTIVE_REQUESTS+1) clients are constantly trying perform historical queries.

src/apps/logging/logging.cpp
eddyashton: In part: If we pass the target view and seqno in headers rather than the body they can be parsed by the framework/helper method. But I'd like to get the structure in in this PR, and work on how cleanly we expose it to apps later.

src/apps/logging/logging.cpp
```diff
@@ -257,6 +259,81 @@ namespace loggingapp
...
+        {
+          const auto tx_view = consensus->get_view(in.seqno);
+          const auto committed_seqno = consensus->get_committed_seqno();
+          const auto committed_view = consensus->get_view(committed_seqno);
+
+          const auto tx_status = ccf::get_tx_status(
+            in.view, in.seqno, tx_view, committed_view, committed_seqno);
+          if (tx_status != ccf::TxStatus::Committed)
+          {
+            args.rpc_ctx->set_response_status(HTTP_STATUS_BAD_REQUEST);
```
achamayou, line 278: Debatable, but I would expect "404: No such committed transaction". 400 is generally some sort of request validation failure (invalid syntax, request too large etc).

src/apps/logging/logging.cpp
eddyashton: Good point - we could store `historical_state` as a member since its the same lifespan as the tables. Alternatively it might make sense to pass this directly to the lambda (since historical queries likely need to have a distinct signature anyway), ie:

```
install("get_historical", [this](ccf::RequestArgs& args, AbstractStateCache& historical_state){ ... });
```

As below, I'd like to leave this API cleanup for a separate PR.

src/apps/logging/logging.cpp
eddyashton: This was a deliberate copy so we can move into `set_response_body`, but of course we can just move directly from `v.value()`.

src/apps/logging/logging.cpp
eddyashton: Still an open question. If we delete here, we make some things unnecessarily more expensive (2 users asking for the same version, or the same user asking twice - we have to do a full refetch!). And when we support range queries, its not so obvious we're immediately done with this one.

My gut is that the model should be "supporting historical queries means reserving N bytes of EPC". We'll cull old states so we don't store too many, but we'll keep states around in case they're still useful.

src/node/historical_queries.h
eddyashton: Thanks. We don't need the callstack here - there's a single caller, and we're going to catch and flatten the exception a short way up the stack. But I'll keep it in mind for future.

src/apps/logging/logging.cpp
eddyashton: We should probably do a pass on our error codes - we're using 400 as a fairly general "your request didn't meet our expectations". We can change some of these to 404s if we've _parsed_ the request, but don't have an answer. I was thinking a 404 that varies with the request body seems odd, but I guess the response body makes it clear what kind of 404 this is.

src/node/historical_queries.h
eddyashton: This is a good option. I think the LRU makes it easier to reason about the memory useage (to enforce a cap, we cull old requests), but we could instead say "remove entries after a day". Might be more general.

Our current use case intends to limit historical queries anyway (eg - an ACL so only some users can call it, assume coordination between users). Since they offer control over node memory useage/performance, I think they should be restricted in general. But that doesn't imply global coordination, so maybe the timeout is preferable. Will give this some more thought!

src/apps/logging/logging.cpp
```diff
@@ -257,6 +259,81 @@ namespace loggingapp
...
+              ccf::tx_status_to_str(tx_status)));
+            return;
+          }
+        }
+
+        auto historical_store = historical_state.get_store_at(in.seqno);
+        if (historical_store == nullptr)
+        {
+          args.rpc_ctx->set_response_status(HTTP_STATUS_ACCEPTED);
+          static constexpr size_t retry_after_seconds = 3;
```
jumaffre, line 296: I wonder what this number should be. Perhaps a rolling average of how long it has taken to retrieve historical queries so far? Definitely not in the scope of this PR though.

src/node/historical_queries.h
```diff
@@ -0,0 +1,358 @@
...
+            else
+            {
+              LOG_DEBUG_FMT(
+                "Not fetching ledger entry {}: already have it in stage {}",
+                request_it->first,
+                request.current_stage);
+            }
+          }
+
+          if (deserialise_result == kv::DeserialiseSuccess::PASS_SIGNATURE)
```
jumaffre, line 264: This will always be true in this part of the switch?

src/node/historical_queries.h
```diff
@@ -0,0 +1,358 @@
...
+          {
+            receipt = tree.get_receipt(untrusted_idx);
+            LOG_DEBUG_FMT(
+              "From signature at {}, constructed a receipt for {}",
+              sig_idx,
+              untrusted_idx);
+          }
+          catch (const std::exception& e)
+          {
+            // This signature doesn't cover this untrusted idx, try the next
```
jumaffre, line 190: Nitpick: "the next **request**" (as opposed to the next signature)

PR #1273, opened 2020-06-11
Extend Merkle Tree API

eddyashton:
Adds a few utility functions to `ccf::MerkleTreeHistory`:

- `begin_index()`
- `end_index()`
- `get_hash(index)`

These involve poking at the `merkle_tree`'s internals so may break across version upgrades - eventually we should implement these in the source F* rather than assumptions on the extracted code.

Also adds a unit test of the merkle tree, and some additional size checks (of serialised tree and receipts) in `merkle_mem`.


src/node/history.h
```diff
@@ -518,10 +558,7 @@ namespace ccf
     void compact(kv::Version v) override
     {
       flush_pending();
-      if (v > MAX_HISTORY_LEN)
-      {
-        replicated_state_tree.flush(v - MAX_HISTORY_LEN);
-      }
+      replicated_state_tree.flush(v);
```
eddyashton, line 561: This change is probably unnecessary - in unit tests we can flush the history directly to a specific index, but in the node's real history we were keeping some prior hashes from before the compaction point. I guess this means the signature minitrees include more historical transactions, not just the block to the previous signature?

src/node/history.h
```diff
@@ -374,6 +372,48 @@ namespace ccf
...
+    {
+      return tree->offset + tree->j - 1;
+    }
+
+    bool in_range(uint64_t index)
+    {
+      return index >= begin_index() && index <= end_index();
+    }
+
+    crypto::Sha256Hash get_hash(uint64_t index)
```
achamayou, line 391: I think get_leaf() would be marginally more precise, but this is also fine.

src/node/history.h
achamayou: Yes, we should stop doing that, it's not particularly useful.

Issue #1135, opened 2020-05-05
Split ledger into multiple files


jumaffre:
Currently, the CCF ledger on one node is a unique file that would grow infinitely as the service processes more transactions. This suffers from the memory growth described in https://github.com/microsoft/CCF/issues/1112.

Instead, the ledger could be split into multiple files so that the node only writes to the latest ledger file. Other ledger files are read-only (i.e. when a new node joins the network and needs to catch up from the beginning of time). 

A ledger file can safely be "archived" once a certain size has been reached (set by the operator at node startup - it can be different for every node) and the latest signature transaction has been globally committed.

jumaffre:
Implementation details:
- [ ] 1. New ringbuffer message (enclave -> host) for the consensus to notify that a transaction index (at a signature) has been committed.
- [ ] 2. On the new commit message, check whether the number of transactions in the ledger is greater than a threshold. If it's the case, split the ledger. Otherwise, no effect.
- [ ] 3. Splitting the ledger consists of (hand-wavy for now, optimisations possible after first implementation):
 - Writing the current transactions offsets in a file
 - Writing the committed prefix to disk
 - Writing the uncommitted suffix to disk
- [ ] 4. Name the archived ledger files so that it is possible to easily/quickly query transactions in archived ledger files based on the transaction index.

PR #1274, opened 2020-06-11
Ledger chunking

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1135

Adds support for ledger chunking. With this PR, a new ledger chunk will be created every time the size of the current ledger chunk passes a operator-defined threshold (specified as `ledger-chunk-threshold` to `cchost`). 

Already-completed ledger chunks also get finalised when the last index they contain is globally committed. In that scenario, their name changes from `ledger_<start_idx>` to `ledger_<start_idx>-<end_idx>.committed`. These files can then be backed up easily as they will no longer be written to (only read to for historical queries and node catch-up).

Implementation details:
- We now pass a `committable` flag on each `ledger_append` ring buffer message so that the host knows which entries are signature. It is only when a `committable` entry is written and the threshold has been passed that a new chunk is created. This guarantees that each chunk ends on a signature transaction.
- Each completed (i.e. whenever the threshold has been passed) ledger file contains the `positions` table at the end of the file so that any index in that file can be randomly accessed when reading the file. As such, each file starts with a 8-byte offset which points to that table.
- Only so many files are kept open at any given time (so that `cchost` should never run out of file descriptors). In particular, all chunks that are not compacted are kept open (as they can be written to or truncated). All compacted chunks are closed and opened when reading them. There is also a read cache so that we do not have to open a file for every historical read query.


src/consensus/raft/raft.h
```diff
@@ -673,27 +658,31 @@ namespace raft
...
           send_append_entries_response(r.from_node, false);
           return;
         }
 
         Term sig_term = 0;
         auto deserialise_success =
-          store->deserialise(ret.first, public_only, &sig_term);
+          store->deserialise(entry, public_only, &sig_term);
+
+        ledger->put_entry(
```
jumaffre, line 684: Changed the order of things a little bit here: we only write to the ledger after the entry has been deserialised because it is only then that we know that it was a signature transaction.

src/host/ledger.h
```diff
@@ -196,7 +196,10 @@ namespace asynchost
       DISPATCHER_SET_MESSAGE_HANDLER(
```
jumaffre, line 196: Keeping this file around for now to have a clean diff for `multiple_ledger.h`. I will remove this one and rename `multiple_ledger.h` to `ledger.h` before merging this PR.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: This is the new implementation for the ledger, as unit-tested in `ledger.cpp`.

src/host/node_connections.h
```diff
@@ -247,18 +247,22 @@ namespace asynchost
               serialized::overlay<consensus::AppendEntriesIndex>(p, psize);
             // Find the total frame size, and write it along with the header.
             auto count = ae.idx - ae.prev_idx;
-            uint32_t frame = (uint32_t)(
-              size_to_send +
-              ledger.framed_entries_size(ae.prev_idx + 1, ae.idx));
+
+            auto framed_entries =
```
jumaffre, line 251: Removed the `framed_entries_size()` ledger API as the frame size can be deduced from the vector returned by `read_framed_entries()`.

src/host/test/ledger.cpp
```diff
@@ -1,61 +1,685 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: These unit tests should test all edge cases that our end-to-end tests do not. 

They use the `TestEntrySubmitter` class to write to/truncate the ledger. This class allows us to trivially verify that something was written to the ledger the value of the transaction index is written in the ledger itself. On read, it is then easy to verify that when asking for all entries between `from` and `to` that the values retrieves are `from`, `from+1`, `from+2`, ..., `to`.

tests/infra/remote.py
```diff
@@ -11,6 +11,8 @@
 import ctypes
```
jumaffre, line 11: Some changes here so that we can copy directories to/from `LocalRemote` and `SSHRemote`.

tests/governance_history.py
```diff
@@ -61,7 +61,11 @@ def run(args):
         network.start_and_join(args)
         primary, _ = network.find_primary()
 
-        ledger_filename = network.find_primary()[0].remote.ledger_path()
+        ledger_directory = network.find_primary()[0].remote.ledger_path()
+        # For now, this test only works with one ledger file
+        for l in os.listdir(ledger_directory):
```
jumaffre, line 66: Hack for now. Our `ledger.py` infra should be able to work on multiple chunks. I'll ticket that.

src/consensus/ledger_enclave.h
```diff
@@ -28,70 +28,58 @@ namespace consensus
...
      *
      * @param entry Serialised entry
+     * @param globally_committable True is entry is signature transaction
      */
-    void put_entry(const std::vector<uint8_t>& entry)
+    void put_entry(const std::vector<uint8_t>& entry, bool globally_committable)
     {
       // write the message
-      RINGBUFFER_WRITE_MESSAGE(consensus::ledger_append, to_host, entry);
+      RINGBUFFER_WRITE_MESSAGE(
```
eddyashton, line 36: Minor: This could be `put_entry(entry.data(), entry.size(), globally_committable)`.

src/consensus/ledger_enclave_types.h
```diff
@@ -30,6 +30,7 @@ namespace consensus
     /// Modify the local ledger. Enclave -> Host
     DEFINE_RINGBUFFER_MSG_TYPE(ledger_append),
     DEFINE_RINGBUFFER_MSG_TYPE(ledger_truncate),
+    DEFINE_RINGBUFFER_MSG_TYPE(ledger_compact),
```
eddyashton, line 33: Minor: I wonder if this should be `ledger_committed`? The enclave's responsibility ends at telling the host that a specific committ-_able_ entry has now been committ-_ed_, and it doesn't care what the host does with this information. We currently use "compact" as the general term here but it means different things for different systems. In particular, every where within the enclave it means "forget about data from before this index", which is very much not what the host should be doing.

jumaffre:
Quick observation on our performance test (`SB_SGX_CFT`) and a ledger chunking threshold of ~8.4MB. We end-up with 3 ledger files, 2 of them are committed:
```
$ ll -h ./workspace/SB_SGX_CFT^_0/0.ledger/
total 25M
drwxrwxr-x 2 jumaffre jumaffre 4.0K Jun 11 12:18 ./
drwxrwxr-x 3 jumaffre jumaffre 4.0K Jun 11 12:18 ../
-rw-rw-r-- 1 jumaffre jumaffre 9.0M Jun 11 12:18 ledger_1-52501.committed
-rw-rw-r-- 1 jumaffre jumaffre 6.9M Jun 11 12:18 ledger_102502
-rw-rw-r-- 1 jumaffre jumaffre 8.6M Jun 11 12:18 ledger_52502-102501.committed
```

The size of the closed chunks is sensibly close to the threshold. The difference is due to the signatures which happen every 5000 tx/1000 ms and the size of the `positions` table which we don't account for when deciding when to chunk.

tests/governance_history.py
jumaffre: https://github.com/microsoft/CCF/issues/1275

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+
+  public:
+    LedgerFile(const std::string& dir, size_t start_idx) :
+      dir(dir),
+      file(NULL),
+      start_idx(start_idx)
+    {
+      file = fopen(
+        (fs::path(dir) /
+         fs::path(fmt::format("{}_{}", file_name_prefix, start_idx)))
```
eddyashton, line 78: Can we split this for readability? `file_name = ...; file_dir = ...; file = fopen(file_dir);`

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+          .c_str(),
+        "w+b");
```
eddyashton, line 80:
```suggestion
      const auto filename = fs::path(fmt::format("{}_{}", file_name_prefix, start_idx));
      const auto file_path = fs::path(dir) / filename;
      file = fopen(file_path.c_str(), "w+b");
```

CMakeLists.txt
```diff
@@ -478,6 +478,10 @@ if(BUILD_TESTS)
       # purpose to check that a recovery network is robust to a view change.
       --raft-election-timeout
       4000
+      # Smaller ledger chunks than default to make sure that a ledger made of
+      # multiple chunks can be recovered
+      --ledger-chunk-threshold
```
achamayou, line 483: Can we make this --chunk-max-bytes, that way we're consistent with the --sig-max-* flags, and that leaves the ability to have --chunk-max-ms later.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+      compacted = is_ledger_file_compacted(file_name);
+      start_idx = get_start_idx_from_file_name(file_name);
+
+      // First, get full size of file
+      fseeko(file, 0, SEEK_END);
+      size_t total_file_size = ftello(file);
+
+      if (total_file_size == 0)
+      {
+        // If the file is empty, initialise it as if it were new
```
eddyashton, line 109: I don't understand this case. If this is during recovery, why is it empty?

doc/users/deploy_app.rst
```diff
@@ -41,9 +41,9 @@ The ``start_test_network.sh`` script can also be used to automatically recover a
 .. code-block:: bash
 
     $ cd CCF/build
-    $ cp ./workspace/test_network_0/0.ledger .
+    $ cp -r ./workspace/test_network_0/0.ledger .
     $ cp ./workspace/test_network_0/network_enc_pubk.pem .
-    $ ../start_test_network.sh -p liblogging.enclave.so.signed --recover --ledger 0.ledger --network-enc-pubk network_enc_pubk.pem --common-dir ./workspace/test_network_common/
+    $ ../start_test_network.sh -p liblogging.enclave.so.signed --recover --ledger-dir 0.ledger --network-enc-pubk network_enc_pubk.pem --common-dir ./workspace/test_network_common/
```
achamayou, line 46:
```suggestion
    $ ../start_test_network.sh -p liblogging.enclave.so.signed --recover --ledger-dir ./ledger --network-enc-pubk network_enc_pubk.pem --common-dir ./workspace/test_network_common/
```

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+      {
+        throw std::logic_error(fmt::format("Failed to flush ledger file"));
+      }
+
+      fs::rename(
+        fs::path(dir) / fs::path(get_file_name()),
+        fs::path(dir) /
+          fs::path(fmt::format(
+            "{}-{}{}",
+            get_file_name(),
```
eddyashton, line 420: Just to confirm - `get_file_name()` includes the start-index, right? So this produces something like `ledger_100-200.committed`?

src/consensus/ledger_enclave_types.h
achamayou: Agreed, `ledger_committed`, or maybe `ledger_freeze` or `ledger_make_read_only` if we want to only express the file consequences and leaved out the consensus implications (technically what matter is, this is now frozen, we won't go back on it. It's because it's committed, but that's the reason, not the action we're telling the host to take).

src/host/main.cpp
```diff
@@ -117,10 +117,20 @@ int main(int argc, char** argv)
     "Address to advertise publicly to clients (defaults to same as "
     "--rpc-address)");
 
-  std::string ledger_file("ccf.ledger");
-  app.add_option("--ledger-file", ledger_file, "Ledger file")
+  std::string ledger_dir("ledger");
+  app.add_option("--ledger-dir", ledger_dir, "Ledger directory")
     ->capture_default_str();
 
+  size_t ledger_chunk_threshold = 23; // ~8MB
```
achamayou, line 124: CLI11 has a very nice facility for this:

```
CLI::AsSizeValue(...): Convert inputs like 100b, 42 KB, 101 Mb, 11 Mib to absolute values. KB can be configured to be interpreted as 10^3 or 2^10.
```

I think it would be much more user friendly that having to specify a power of two (even if we do end up rounding up to that for some reason).

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+          max_chunk_threshold_size));
+      }
+
+      if (fs::is_directory(ledger_dir))
+      {
+        // If the ledger directory exists, recover ledger files from it
+        for (auto const& f : fs::directory_iterator(ledger_dir))
+        {
+          files.push_back(
+            std::make_shared<LedgerFile>(ledger_dir, f.path().filename()));
```
eddyashton, line 579: Looks like `LedgerFile` always reconstructs the full path, so we don't need to split these? (We can construct `LedgerFile` from just `f.path()`?)

src/host/multiple_ledger.h
achamayou: Yes, that's the idea. The start index isn't useful when reading, only the end index is used, but it's nice to have for an operator wanting to eyeball some files.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+      require_new_file = true;
+
+      auto f_from = get_it_contains_idx(idx + 1);
+      auto f_to = get_it_contains_idx(last_idx);
+      auto f_end = std::next(f_to);
+
+      for (auto it = f_from; it != f_end;)
+      {
+        // Truncate the first file to the truncation index while the more recent
+        // files are deleted entirely
```
eddyashton, line 712: Is this right? Isn't it possible to truncate to the middle of a later file, if we don't get a commit for a long time?

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+          std::make_move_iterator(v.end()));
+        idx = to_ + 1;
+      }
+
+      return entries;
+    }
+
+    size_t write_entry(const uint8_t* data, size_t size, bool committable)
+    {
+      if (require_new_file)
```
eddyashton, line 673: Haven't looked at this in detail, but it seems risky to be juggling this bool between functions. Can we instead create a new file at any point where we'd be setting this? So we create new files eagerly, rather than lazily.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+          throw std::logic_error(fmt::format(
+            "Error: Could not create ledger directory: {}", ledger_dir));
+        }
+        require_new_file = true;
+      }
+    }
+
+    MultipleLedger(const MultipleLedger& that) = delete;
+
+    const std::vector<uint8_t> read_entry(size_t idx)
```
eddyashton, line 629: Nitpick: I'd prefer `std::optional<std:vector<uint8_t>>` here, so we can distinguish missing entries from present-but-empty entries (sure the latter is probably invalid, but that's not for us to decide). Also there's little point in returning const values.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+      // directory
+      std::map<size_t, std::string> all_files;
+      for (auto const& f : fs::directory_iterator(ledger_dir))
+      {
+        all_files.emplace(
+          get_start_idx_from_file_name(f.path().filename()),
+          f.path().filename());
+      }
+
+      auto f_ = std::upper_bound(
```
eddyashton, line 496: Do we need to build this map of all files? This map will include our unfinished write files, so there's a risk we add them to our read cache (looks like that's currently avoided because the only calling point checks the write-files first, but seems the intent of this function could be narrowed to ensure it only talks about finished, readable files). While iterating through the committed files, we can use the first one whose range includes our index?

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,815 @@
...
+      return get_file_from_cache(idx);
+    }
+
+    std::shared_ptr<LedgerFile> get_latest_file() const
+    {
+      if (files.empty())
+      {
+        return nullptr;
+      }
+      return *(files.rbegin()++);
```
eddyashton, line 552: This looks like it returns the last-but-one file? `files.rbegin()` should be the last element in files (its not the same as `files.end()`)

src/consensus/ledger_enclave_types.h
jumaffre: I think `ledger_committed` is the best option. This message does not necessarily mean that the chunk that contains this index will be frozen/read-only from then on as that chunk might not yet be complete.

CMakeLists.txt
jumaffre: Went with `--ledger-chunk-max-bytes`.

src/host/multiple_ledger.h
jumaffre: Good point. Committed your suggestion 👍

src/host/main.cpp
jumaffre: I didn't know about this - that's excellent! We are now using it for the ledger chunk size.

src/host/multiple_ledger.h
jumaffre: Good spot! It is indeed an outdated edge case that I forgot to remove.

doc/users/deploy_app.rst
jumaffre: This is actually the ledger of the first node that gets copied locally (line `44`) before being used for recovery. I think it is better to keep the original ledger folder name here to demonstrate that we are recovering the ledger of a specific node.

src/host/multiple_ledger.h
jumaffre: Yup. It's actually the `start_idx` that we care about when we're reading files from disk, the `last_idx` is there for the reason @achamayou  mentioned.

src/host/multiple_ledger.h
jumaffre: I think this makes the two constructors of `LedgerFile` cleaner: 
- `LedgerFile(dir, start_idx)` that create a fresh new ledger file from `start_idx`
- `LedgerFile(dir, file_name)` that restores an existing ledger file 

We need to get the `directory` in `LedgerFile` so we'd have to split it in there anyway.

src/consensus/ledger_enclave_types.h
achamayou: But the message applies to an index, not a chunk? I does mean that all entries in the chunk(s) up to that point are frozen/read-only. Anyway, committed is good too.

doc/users/deploy_app.rst
achamayou: Ah, I hadn't realised 0.ledger was a directory, since the name matched the old files.

src/host/multiple_ledger.h
achamayou: Agreed, although in practice it's hard to imagine a valid use case for empty entries.

src/consensus/ledger_enclave.h
```diff
@@ -103,5 +89,15 @@ namespace consensus
     {
       RINGBUFFER_WRITE_MESSAGE(consensus::ledger_truncate, to_host, idx);
     }
+
+    /**
+     * Compact the ledger at a given index.
+     *
+     * @param idx Index to compact at
+     */
+    void compact(Index idx)
```
achamayou, line 98: Calling this compact isn't right. We compact the store, but we `freeze` the ledger, or we `make_immutable`, or `advance_read_only_watermark`, possibly `finalise()`.

At worst, `commit()`.

src/consensus/ledger_enclave_types.h
```diff
@@ -46,6 +47,7 @@ DECLARE_RINGBUFFER_MESSAGE_PAYLOAD(
   consensus::Index,
   consensus::LedgerRequestPurpose);
 DECLARE_RINGBUFFER_MESSAGE_PAYLOAD(
-  consensus::ledger_append, std::vector<uint8_t>);
+  consensus::ledger_append, bool, std::vector<uint8_t>);
```
achamayou, line 50:
```suggestion
  consensus::ledger_append, bool /* committable */, std::vector<uint8_t>);
```

src/host/main.cpp
```diff
@@ -117,10 +117,19 @@ int main(int argc, char** argv)
...
     ->capture_default_str();
 
+  size_t ledger_chunk_threshold = 5'000'000;
+  app
+    .add_option(
+      "--ledger-chunk-max-bytes",
+      ledger_chunk_threshold,
+      "Minimum size (bytes) at which a new ledger chunk is created.")
+    ->capture_default_str()
+    ->transform(CLI::AsSizeValue(true)); // 1000 is kb
```
achamayou, line 131: The comment is slightly mysterious, do we need it?

src/host/multiple_ledger.h
jumaffre: Good point. Done now.

src/host/main.cpp
jumaffre: The comment refers to the `true` which indicates that `k` is 1000 and not 1024.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,805 @@
...
+#include <unistd.h>
+#include <vector>
+
+namespace fs = std::filesystem;
+
+namespace asynchost
+{
+  static constexpr size_t max_chunk_threshold_size =
+    std::numeric_limits<uint32_t>::max(); // 4GB
+  static constexpr size_t max_read_cache_size_default = 5;
```
achamayou, line 26: I presume this is entries, as opposed to bytes? If so, I suggest max_read_cache_entries.

jumaffre: This is actually the number of files open. I'll clarify.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,805 @@
...
+    LedgerFile(const std::string& dir, const std::string& file_name) :
+      dir(dir),
+      file(NULL)
+    {
+      auto full_path = (fs::path(dir) / fs::path(file_name));
+      file = fopen(full_path.c_str(), "r+b");
+      if (!file)
+      {
+        throw std::logic_error(
+          fmt::format("Unable to open ledger file {}", full_path));
```
achamayou, line 97: I suggest we throw in `strerror(errno)`.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,805 @@
...
+          pos += (entry_size + frame_header_size);
+        }
+        completed = false;
+      }
+    }
+
+    ~LedgerFile()
+    {
+      fflush(file);
+      fclose(file);
```
achamayou, line 176:
```suggestion
      fclose(file);
```

Explicit flushing is unnecessary, fclose already does that: https://linux.die.net/man/3/fclose

src/host/multiple_ledger.h
jumaffre: This somehow returns the last element in the list. But `files.back()` should be preferred.

src/host/multiple_ledger.h
```diff
@@ -0,0 +1,805 @@
...
+    }
+
+    std::string get_file_name() const
+    {
+      int fd = fileno(file);
+      auto path = fmt::format("/proc/self/fd/{}", fd);
+      char result[1024];
```
achamayou, line 183: Should be `PATH_MAX` from `linux/limits.h`, which is 4096.

src/host/multiple_ledger.h
achamayou: Unless I'm missing something, f_from is queried as "the first file that contains idx+1", so this does seem right.

tests/infra/e2e_args.py
```diff
@@ -151,6 +151,11 @@ def cli_args(add=lambda x: None, parser=None, accept_unknown=False):
         type=int,
         default=10,
     )
+    parser.add_argument(
+        "--ledger-chunk-max-bytes",
+        help="Minimum size (bytes) at which a new ledger chunk is created.",
+        default= "100MB",  # 100MB chunks by default
```
achamayou, line 157: I think the code is now self-explanatory :)

src/host/multiple_ledger.h
eddyashton: Ah cool, I'd missed this was only operating on a suffix range.

src/host/main.cpp
achamayou: Ah, got it, sorry!

src/host/multiple_ledger.h
achamayou: That's what I meant by files, but you're right that files is better.

src/host/multiple_ledger.h
jumaffre: As discussed, creating a new file whenever the threshold is passed would complicate the truncation logic which I think we should avoid. Leaving the bool in for now but happy to revisit in the future (e.g., when implementing #1276).

src/consensus/ledger_enclave.h
jumaffre: I went for `commit()` to stay consistent with the rest of the interface. It doesn't necessarily mean that the ledger is frozen at that point though, it just notifies the ledger that the consensus has agreed on an index. What it guarantees though is that all indices included in _complete_ chunks that are `<=` the commit idx will be committed and forever closed for writing. I hope this is clearer now but let me know if it isn't.

src/host/main.cpp
jumaffre: I was very confused by it too at first and had to look into `CLI11.hpp` directory to find this out.

tests/infra/e2e_args.py
jumaffre: Yes indeed!

src/host/multiple_ledger.h
jumaffre: The case for empty entries was actually used when sending raft heartbeats. It was calling `read_framed_entries(from = N+1, to = 6)` which would return an empty vector. I've adapted the code slightly in `node_connections.h`.

src/host/multiple_ledger.h
jumaffre: You're right. We can simply loop through all the files and only consider the ones that are committed (as you rightly pointed, the non-committed ones will be in the write cache). I've implemented this.

src/host/multiple_ledger.h
jumaffre: Done here and for all `f...()` file operations that touch `errno`.

src/consensus/ledger_enclave.h
achamayou: I think it logically means that entries in the ledger up to that point cannot ever be rolled back or modified, even if they are part of a file that still has a mutable suffix. I'm fine with commit(), but compact() was inaccurate.

jumaffre:
Still some double free issues raised on the CI that I have failed to repro locally. Investigation this now.

PR #1277, opened 2020-06-11
Build common helper/adapter for historical queries

eddyashton:
Follow up to #1267.

Pulls some common functionality that will be required for many historical queries (find out what transaction it is asking for, find out whether the transaction is valid, find out whether we have that transaction cached) into `ccf::historical::adapter`. I think this is a useful simplification; the body of `get_historical` in `logging.cpp` is now far more readable.

Remaining TODOs, currently aimed at future PRs:
- Finish verification of historical stores (#1278)
- Handle range queries

### DIFF
--- a/src/apps/logging/logging.cpp
+++ b/src/apps/logging/logging.cpp
@@ -261,8 +261,8 @@ namespace loggingapp
 
       auto& historical_state = context.get_historical_state();
       auto get_historical = [this, &historical_state](ccf::RequestArgs& args) {
-        const auto params =
-          nlohmann::json::parse(args.rpc_ctx->get_request_body());
+        const auto [pack, params] =
+          ccf::jsonhandler::get_json_params(args.rpc_ctx);
         const auto in = params.get<LoggingGetHistorical::In>();
 
         // Check that the requested transaction ID is committed
@@ -318,14 +318,14 @@ namespace loggingapp
 
         kv::Tx historical_tx;
         auto view = historical_tx.get_view(*historical_map);
-        auto v = view->get(in.id);
+        const auto v = view->get(in.id);
 
         if (v.has_value())
         {
-          args.rpc_ctx->set_response_status(HTTP_STATUS_OK);
-          args.rpc_ctx->set_response_header(
-            http::headers::CONTENT_TYPE, http::headervalues::contenttype::TEXT);
-          args.rpc_ctx->set_response_body(std::move(v.value()));
+          LoggingGetHistorical::Out out;
+          out.msg = v.value();
+          nlohmann::json j = out;
+          ccf::jsonhandler::set_response(std::move(j), args.rpc_ctx, pack);
         }
         else
         {
--- a/src/apps/logging/logging.cpp
+++ b/src/apps/logging/logging.cpp
@@ -259,51 +259,13 @@ namespace loggingapp
       };
       // SNIPPET_END: log_record_text
 
-      auto& historical_state = context.get_historical_state();
-      auto get_historical = [this, &historical_state](ccf::RequestArgs& args) {
+      auto get_historical = [this](
+                              ccf::RequestArgs& args,
+                              ccf::historical::StorePtr historical_store,
+                              kv::Consensus::View historical_view,
+                              kv::Consensus::SeqNo historical_seqno) {
         const auto [pack, params] =
           ccf::jsonhandler::get_json_params(args.rpc_ctx);
-        const auto in = params.get<LoggingGetHistorical::In>();
-
-        // Check that the requested transaction ID is committed
-        {
-          const auto tx_view = consensus->get_view(in.seqno);
-          const auto committed_seqno = consensus->get_committed_seqno();
-          const auto committed_view = consensus->get_view(committed_seqno);
-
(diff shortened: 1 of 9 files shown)


tests/e2e_logging.py
```diff
@@ -510,24 +514,24 @@ def run(args):
...
-                network, args, verify=args.package is not "libjs_generic"
-            )
-            network = test_large_messages(network, args)
-            network = test_remove(network, args)
-            network = test_forwarding_frontends(network, args)
-            network = test_update_lua(network, args)
-            network = test_cert_prefix(network, args)
-            network = test_anonymous_caller(network, args)
-            network = test_raw_text(network, args)
+            # network = test(
```
achamayou, line 517: Presumably to be reverted?

src/apps/logging/logging.cpp
```diff
@@ -312,27 +274,62 @@ namespace loggingapp
...
+          nlohmann::json j = out;
+          ccf::jsonhandler::set_response(std::move(j), args.rpc_ctx, pack);
         }
         else
         {
           args.rpc_ctx->set_response_status(HTTP_STATUS_NO_CONTENT);
         }
       };
 
+      auto is_tx_committed = [this](
```
jumaffre, line 303: Why is this defined by the app? I understand that this might be nice for the app to override this check but the default seems very much generic to all apps.

src/node/rpc/handler_registry.h
```diff
@@ -29,6 +29,20 @@ namespace ccf
     return 1ul << verb;
   }
 
+  static bool is_committed(
```
jumaffre, line 32: Is this used anywhere?

src/node/historical_queries_interface.h
```diff
@@ -27,4 +28,108 @@ namespace ccf::historical
...
+    AbstractStateCache& state_cache,
+    const CheckAvailability& available)
+  {
+    return [f, &state_cache, available](RequestArgs& args) {
+      // Extract the requested transaction ID
+      kv::Consensus::View target_view;
+      kv::Consensus::SeqNo target_seqno;
+
+      {
+        const auto target_view_opt =
```
jumaffre, line 54: Why are we using HTTP headers instead of query string here? Is that because the view and seqno are already returned as response headers when a transaction is committed. In other words, the view and seqno are related CCF whereas the everything app-related uses query strings?

src/apps/logging/logging.cpp
eddyashton: Yeah this is a bit of a bodge, I'll clean it up. The problem is that we need to capture the `consensus`, at the point of execution, and check it before we request a historical store. Although every historical query handler may want to use this, it still needs to be 'functionalised' to pass it to the functional adapters. But we can pass a bound member function, doesn't need to be a lambda.

src/node/historical_queries_interface.h
eddyashton: I'm fine with putting view and seqno in the query string/body in other places - that's what we do with `getReceipt`, getTxStatus` etc.

But in this case I'm trying to make a distinction between the things the framework/wrapper does, and what the app does. We could parse a few things from the query string and leave the rest for the app but this seems ugly - we don't currently split the query for easy consumption _in the framework_ at all, we're deliberately leaving that for the app to decide what types it expects. I think its cleaner if this framework-defined wrapper uses framework-defined headers, and still leaves the query/body parsing to the app.

Also just a note, not sure if this PR makes it clear, this adapter is entirely optional - its still possible to write a historical query entirely in the app that use your own headers, or the query string, or whatever you like.

src/node/rpc/handler_registry.h
eddyashton: No, good catch.

src/node/historical_queries_interface.h
jumaffre: I agree with you - it is much better/cleaner this way. The only minor drawback I can think of is that it's _slightly_ more fiddly to set a HTTP request header than a query string with `curl`.

src/apps/logging/logging.cpp
jumaffre: If I understand this correctly, we could get away without this if we had access to the `consensus` from the historical query adapter - I don't think we need to capture it "at the point of execution" as `consensus` is a long-lived object? I reckon we could pass `consensus` down the the adapter from `frontend.h`?

eddyashton: `consensus` is quite long-lived, but not the full node lifetime - we can submit RPCs before we have a consensus (I believe, since we currently have a lot of RPC handlers that expect it might not exist yet). We could pass it to every handler from `frontend.h` in the `Args`, that's probably nicer!

eddyashton: So I'd like to do the last idea, passing `consensus` to the handlers that want it rather than having it as a member of the registry. Its a slightly involved change though, and dovetails with making `non-Tx` handlers, so I'd like to do it in a separate PR. Are we happy with the ugly lambda as a temporary solution?

jumaffre: Absolutely 👍

PR #1279, opened 2020-06-12
Restore historical receipts

eddyashton:
In #1278 I changed our compaction behaviour, to compact all the way to the requested index rather than maintaining `MAX_HISTORY_LEN` old entries. I've worked out why that was a mistake - we can only generate receipts for the in-memory, non-flushed tree. We could look at using the historical state cache to serve arbitrarily old receipts instead, but for now I'm restoring the old behaviour.

### DIFF
--- a/src/node/history.h
+++ b/src/node/history.h
@@ -558,7 +558,12 @@ namespace ccf
     void compact(kv::Version v) override
     {
       flush_pending();
-      replicated_state_tree.flush(v);
+      // Receipts can only be retrieved to the flushed index. Keep a range of
+      // history so that a range of receipts are available.
+      if (v > MAX_HISTORY_LEN)
+      {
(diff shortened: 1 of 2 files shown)


src/node/history.h
```diff
@@ -559,7 +571,12 @@ namespace ccf
     void compact(kv::Version v) override
     {
       flush_pending();
-      replicated_state_tree.flush(v);
+      // Receipts can only be retrieved to the flushed index. Keep a range of
+      // history so that a range of receipts are available.
+      if (v > MAX_HISTORY_LEN)
+      {
+        replicated_state_tree.flush(v - MAX_HISTORY_LEN);
+      }
```
ashamis, line 579: I think this is going to run into issues with PBFT. Should be ok for Raft however.

PR #1280, opened 2020-06-12
Fix PBFT suspension timeout

achamayou:
Fix for failures such as https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=9181&view=ms.vss-test-web.build-test-results-tab

This contains two changes:

1. A timeout extension, because we're currently cutting it really close on the replay_new_view test, hence the failure
2. A fix to use the actual value returned by commit, as opposed to the incidental version header, which is meaningless since it is a "No KV" transaction

### DIFF
--- a/tests/infra/suspension.py
+++ b/tests/infra/suspension.py
@@ -42,10 +42,10 @@ def update_view_info(network, view_info):
 def get_node_local_commit(node):
     with node.node_client() as c:
         r = c.get("commit")
-        return r.seqno, r.global_commit
+        return r.result["seqno"], r.global_commit
 
 
-def wait_for_late_joiner(old_node, late_joiner, strict=False, timeout=30):
+def wait_for_late_joiner(old_node, late_joiner, strict=False, timeout=60):
     old_node_lc, old_node_gc = get_node_local_commit(old_node)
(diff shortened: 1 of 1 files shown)


tests/infra/suspension.py
```diff
@@ -42,10 +42,10 @@ def update_view_info(network, view_info):
 def get_node_local_commit(node):
     with node.node_client() as c:
         r = c.get("commit")
-        return r.seqno, r.global_commit
```
achamayou, line 45: global_commit is still there, and was never tied to the transaction (it's just whatever the consensus thinks it is right now). Its removal is a separate matter.

tests/infra/suspension.py
```diff
@@ -42,10 +42,10 @@ def update_view_info(network, view_info):
...
+        return r.result["seqno"], r.global_commit
 
 
-def wait_for_late_joiner(old_node, late_joiner, strict=False, timeout=30):
+def wait_for_late_joiner(old_node, late_joiner, strict=False, timeout=60):
```
achamayou, line 48: Replay relies on ledger reads, and disk I/O on the build nodes is surprisingly chaotic.

PR #1283, opened 2020-06-12
Fix adapter locking

achamayou:


eddyashton:
I think these should go the other way - removing the loop in all cases. The `lock()` is on a `weak_ptr`, to support the consensus potentially being destroyed, its not a mutex for concurrent access. So this adapter is deliberately happy to fail. (In practice, we never switch consensus, so this is never used).

achamayou:
@ashamis no?

https://en.cppreference.com/w/cpp/memory/weak_ptr/lock

It will block, until it can acquire the pointer, unless the pointer is expired. We don't expire the KV, nor do we have plan to do that?

PR #1285, opened 2020-06-15
Put version check back in after 0.11-rc tests

achamayou:


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ include(${CCF_DIR}/cmake/preproject.cmake)
 
 project(
   ccf
-  VERSION 0.11 # CCF_VERSION < This is a search marker for CI version check
+  VERSION 0.10 # CCF_VERSION < This is a search marker for CI version check
```
eddyashton, line 10: I think it's ok to bump this version to 0.11 now. The script lets master claim a version ahead of the tag, just not the other way around. We're about to release 0.11, so master can claim 0.11

```suggestion
  VERSION 0.11 # CCF_VERSION < This is a search marker for CI version check
```

PR #1288, opened 2020-06-15
Snapshot Term/whole TxID in KV so as to report status correctly in multi-threaded configuration

achamayou:
Replaces #1252 

In a nutshell, this change means that the KV is always made aware of what `view` (*) transactions are executing in. This enables accurate Transaction ID reporting in a multi-threaded context.

(*) although it really is the `view`, and that's how it's communicated externally to users, the code deliberately refers to it as `term` internally, because there's already a concept called a `view` in the KV, which is really a `TxTableView`. Naming is hard. I'm open to suggestions here.


achamayou:
Depends on #1286

src/consensus/pbft/pbft_types.h
```diff
@@ -198,6 +201,31 @@ namespace pbft
       return kv::NoVersion;
     }
 
+    kv::TxID next_txid()
+    {
+      while (true)
```
eddyashton, line 206: These still have the `while (true)` loop, which has been removed from the other methods.

src/node/history.h
```diff
@@ -591,26 +591,25 @@ namespace ccf
 
       if (consensus->type() == ConsensusType::RAFT)
       {
-        auto version = store.next_version();
-        auto view = consensus->get_view();
+        auto txid = store.next_txid();
         auto commit_txid = consensus->get_committed_txid();
-        LOG_DEBUG_FMT("Issuing signature at {}", version);
+        LOG_DEBUG_FMT("Issuing signature at {}", txid.version);
```
eddyashton, line 596: This log line can be removed - it repeats what the next line says.

src/node/node_state.h
```diff
@@ -614,9 +611,14 @@ namespace ccf
 
       network.ledger_secrets->init(last_recovered_commit_idx + 1);
       setup_encryptor(network.consensus_type);
+      // KV term must be set before the first Tx is committed
+      LOG_INFO_FMT(
+        "Setint term on public recovery KV to {}", term_history.size() + 2);
```
eddyashton, line 616:
```suggestion
        "Setting term on public recovery KV to {}", term_history.size() + 2);
```

src/consensus/pbft/pbft_types.h
achamayou: Sorry, thanks for spotting that.

src/kv/store.h
```diff
@@ -251,18 +254,28 @@ namespace kv
...
       // at the specified version.
       // No transactions can be prepared or committed during rollback.
       std::lock_guard<SpinLock> mguard(maps_lock);
 
-      if (v >= current_version())
-        return;
+      {
+        std::lock_guard<SpinLock> vguard(version_lock);
+        // The term should always be updated on rollback() when passed
+        // regardless of whether version needs to updated or not
```
jumaffre, line 267: Minor: "needs to **be** updated"

PR #1289, opened 2020-06-15
Do not stop uv loop from signal handler

jumaffre:
Locally solves the CI issue experienced on #1274 

We were calling `uv_stop()`, which is not [async-signal-safe](https://github.com/libuv/libuv/issues/2173#issuecomment-460566756), from the signal handler for `SIGTERM`. This PR adds another `enclave -> host` ring buffer message so that the uv loop is only closed once the enclave has acknowledged to the host that it has stopped, leaving the `SIGTERM` handler very short.

It is not 100% clear to me whether the double free issue was caused by this but it seems like a necessary change anyway. Fixing #1274 will hopefully be confirmed once this PR is merged.

### DIFF
--- a/src/enclave/enclave.h
+++ b/src/enclave/enclave.h
@@ -204,6 +204,9 @@ namespace enclave
           bp, AdminMessage::stop, [&bp, this](const uint8_t*, size_t) {
             bp.set_finished();
             threading::ThreadMessaging::thread_messaging.set_finished();
+            auto w = writer_factory.create_writer_to_outside();
+            LOG_INFO_FMT("Enclave stopped successfully. Stopping host...");
+            RINGBUFFER_WRITE_MESSAGE(AdminMessage::stopped, w);
           });
 
         DISPATCHER_SET_MESSAGE_HANDLER(
--- a/src/host/handle_ring_buffer.h
+++ b/src/host/handle_ring_buffer.h
@@ -55,6 +55,12 @@ namespace asynchost
           std::cerr << msg << std::endl << std::flush;
           throw std::logic_error(msg);
         });
+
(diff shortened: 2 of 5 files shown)


src/host/handle_ring_buffer.h
```diff
@@ -55,6 +55,12 @@ namespace asynchost
           std::cerr << msg << std::endl << std::flush;
           throw std::logic_error(msg);
         });
+
+      DISPATCHER_SET_MESSAGE_HANDLER(
+        bp, AdminMessage::stopped, [](const uint8_t* data, size_t size) {
+          uv_stop(uv_default_loop());
```
jumaffre, line 61: I believe that doing this as late as here gives the host sufficient time to print out all the enclave logs on shutdown.

src/enclave/enclave.h
```diff
@@ -204,6 +204,9 @@ namespace enclave
           bp, AdminMessage::stop, [&bp, this](const uint8_t*, size_t) {
             bp.set_finished();
             threading::ThreadMessaging::thread_messaging.set_finished();
+            auto w = writer_factory.create_writer_to_outside();
```
eddyashton, line 207: I think this should go at the end of `run_main` in `enclave.h` - anything after `bp.run()` will only execute once this `set_finished` has been processed, so this moves the notification slightly closer to the point where we actually exit.

src/host/handle_ring_buffer.h
eddyashton: It is possible that worker threads in the enclave are still executing and writing outbound messages - we may drop those. But this should still fix the safety issue, since this `uv_stop` should always be called from the "host thread", whereas the signal could be handled on any.

Issue #1102, opened 2020-04-27
Handlers should be associated with a single HTTP method


eddyashton:
Currently each installed handler can handle multiple HTTP methods, and by default they accept _every_ method. ie

```
install(foo_lambda, "foo");
```

will result in `foo_lambda` being called for `GET /foo`, `POST /foo`, `DELETE /foo` etc. We can restrict the set of accepted methods. But for many HTTP servers we want `GET /foo` and `POST /foo` to do totally different things, needing different handlers with different schemas. This is currently not possible in CCF (the handler may do different things depending on the method, but the schema and other options are fixed).

I think handling multiple methods was just a quirk of the transition from framed-TLS to HTTP, accepting everything for ease. Instead we should associate each handler with its intended method, and dispatch by `(path, method)` rather than just `(path)`. This will also make it simpler to generate OpenAPI specs, where each [path item](http://spec.openapis.org/oas/v3.0.3#paths-object) is keyed by method.

If we have a handler that _does_ wish to accept multiple verbs, this will still be possible by simply installing the handler twice (thus allowing _other_ options, like read-only or the schema, to be method-specific):

```
install(foo_lambda, "foo", Write).set_method(HTTP_POST).set_auto_schema<FooPost>();
install(foo_lambda, "foo", Read).set_method(HTTP_GET).set_auto_schema<FooGet>();
```

We'll need to look at how this interacts with WebSockets, either including the method in the InHeader or considering all WS RPCs to be a separate method, having their own handlers (`install(foo_lambda, "foo").set_setwebsockets(true);`).

achamayou:
Websockets doesn't have methods, nor is there a plan to have one, so it can either be stubbed, or omitted, or like you said, made separate altogether.

PR #1290, opened 2020-06-15
Rename standard RPCs to look more RESTful.

### LINKED ISSUE #968: Endpoints should look like HTTP resources

eddyashton:
Resolves #968.

Open to discussion on these names. This only touches the common framework endpoints - for the logging app we'd really like #1102 (so `POST /users/logging/public` and `GET /users/logging/public` are different things).


doc/operators/operator_rpc_api.rst
```diff
@@ -26,12 +26,12 @@ Example
         ]
     }
 
-getQuotes
+quotes
 ---------
```
jumaffre, line 30: Minor: `----` is too long

PR #1294, opened 2020-06-16
Associate each handler with a single HTTP verb

### LINKED ISSUE #1102: Handlers should be associated with a single HTTP method

eddyashton:
Fixes #1102 

Previously, each installed handler could accept a set of verbs, and dispatch was done purely by URL path (so `GET /foo` and `POST /foo` are handled by the same bit of code, and the same read-write hint etc).

Now each install gets a single verb, and this is used in dispatch (so `GET /foo` and `POST /foo` can go to different lambdas, or the same lambda installed with different args).


src/node/rpc/handler_registry.h
```diff
@@ -213,47 +208,46 @@ namespace ccf
...
-      uint64_t allowed_verbs_mask = ~0;
+      http_method verb = HTTP_POST;
 
-      // https://github.com/microsoft/CCF/issues/1102
-      Handler& set_allowed_verbs(std::set<http_method>&& allowed_verbs)
+      /** Indicates which HTTP verb the handler should respond to.
+       *
+       * @return The installed Handler for further modification
+       */
+      [[deprecated("HTTP Verb should not be changed after installation: pass verb to install()")]] Handler&
```
eddyashton, line 217: For discussion: Is it worth listing these as deprecated for a release? It means the API change is just a warning rather than a compile error, and points clearly to the intended resolution. On the other hand, it ties us to supporting old APIs for longer, and we need to remember to go delete our deprecations at some point.

I'm thinking this instance is a useful test run, and we should get in the habit of deprecating-then-deleting moving forward, as we'll need to post-1.0.

src/node/rpc/handler_registry.h
```diff
@@ -213,47 +208,55 @@ namespace ccf
...
+      http_method verb = HTTP_POST;
 
-      // https://github.com/microsoft/CCF/issues/1102
-      Handler& set_allowed_verbs(std::set<http_method>&& allowed_verbs)
+      /** Indicates which HTTP verb the handler should respond to.
+       *
+       * @return The installed Handler for further modification
+       */
+      // clang-format off
+      [[deprecated("HTTP Verb should not be changed after installation: pass verb to install()")]]
```
achamayou, line 218: Very nice.

src/node/rpc/handler_registry.h
```diff
@@ -213,47 +208,55 @@ namespace ccf
...
       }
     };
 
   protected:
     std::optional<Handler> default_handler;
-    std::unordered_map<std::string, Handler> handlers;
+    // Handler lookup uses method and HTTP verb. We assume there is usually a
+    // single verb-per-method, so this map looks up only by method and then we
+    // iterate.
+    std::unordered_map<std::string, std::vector<Handler>> handlers;
```
achamayou, line 259: Why do that and not have (path, verb) be the key instead?

src/node/rpc/common_handler_registry.h
```diff
@@ -242,39 +253,32 @@ namespace ccf
           HTTP_STATUS_INTERNAL_SERVER_ERROR, "Unable to verify receipt");
       };
 
-      install(GeneralProcs::GET_COMMIT, json_adapter(get_commit), Read)
+      install_get(GeneralProcs::GET_COMMIT, json_adapter(get_commit), Read)
```
jumaffre, line 256: As discussed before, my understand is that the `Read/Write` verb passed as the last argument to `install()` is now redundant and we could use the HTTP verb to forward or not (we can of course build a `.set...()` exception for `GET` handlers that end up writing to the KV).

In other words, it feels like there should only be one API for installing: `install(URI, lambda, HTTP_VERB)` so I am not too keen on having a specific `install_get()` and `install_post()`.

eddyashton:
Converting this to a draft for now while I play with the API changes.

eddyashton:
This is superceded by #1310, which includes a broader set of API changes.

PR #1295, opened 2020-06-16
Recovery share CLI

### LINKED ISSUE #1096: Member CLI to handle encrypted recovery share
When a member retrieves their encrypted recovery shares, they need to decrypt the share (using their private encryption key and the network public encryption key) before submitting the decrypted share to CCF. In the e2e Python infra, we currently perform the decryption using `pynacl`. 

However, we should provide a tool for members to do this for the command line. 

Several options:
- Write a small Python script which would use `pynacl`. However, members would have to install the dependencies/create a venv before running it. 
- Use an OTS tool (e.g. https://github.com/danderson/gobox). However, this would involve some manipulation on the member side: extract the encrypted share from the RPC response and write it to file, etc. 
- Write our own binary that handles the decryption (à la now-defunct `memberclient`). This means another binary to build/ship/maintain but we could simplify the whole recovery share submission process for members: the CLI would retrieve the encrypted share, decrypt it, and submit it for recovery in one step.

jumaffre:
Resolves #1096 

This PR introduces a new `submit_recovery_shares.sh` CLI to let members retrieve, decrypt and submit their recovery shares in one step. This script has a dependency on the [step CLI](https://smallstep.com/docs/cli/crypto/nacl/) to decrypt the recovery share.

The `GET recovery_share` and `POST recovery_share/submit` endpoints now work with base64 instead of byte array for convenience.


doc/members/accept_recovery.rst
```diff
@@ -42,28 +42,31 @@ To restore private transactions and complete the recovery procedure, members sho
...
-    $ curl https://<ccf-node-address>/members/recovery_share/submit -X POST --data-binary @submit_recovery_share.json --cacert network_cert --key member2_privk --cert member2_cert -H "content-type: application/json"
-    true
-
-When the recovery threshold is reached, the ``recovery_share/submit`` RPC returns ``true``. At this point, the private recovery procedure is started and the private ledger is being recovered.
+    $ ./submit_recovery_share.sh --rpc-address <ccf-node-address> --member-enc-privk member0_enc_privk.pem --network-enc-pubk network_enc_pubk --cert member0_cert
+    --key member0_privk --cacert network_cert
+    HTTP/1.1 200 OK
+    content-length: 46
+    content-type: application/json
+    x-ccf-global-commit: 27
```
achamayou, line 54: Please remove the global commit header, it's deprecated, we don't want it in the documentation.

doc/members/accept_recovery.rst
```diff
@@ -42,28 +42,31 @@ To restore private transactions and complete the recovery procedure, members sho
...
+    x-ccf-tx-seqno: 28
+    x-ccf-tx-view: 4
+    "1/2 recovery shares successfully submitted."
+
+    $ ./submit_recovery_share.sh --rpc-address <ccf-node-address> --member-enc-privk member1_enc_privk.pem --network-enc-pubk network_enc_pubk --cert member1_cert
+    --key member1_privk --cacert network_cert
+    HTTP/1.1 200 OK
+    content-length: 83
+    content-type: application/json
+    x-ccf-global-commit: 29
```
achamayou, line 64: Ditto. I think the content-length can also be elided.

tests/submit_recovery_share.sh
```diff
@@ -0,0 +1,85 @@
...
+openssl asn1parse -in "${network_enc_pubk}" -i -strparse "${der_header_pubk_len}" -out network_enc_pubk.raw -noout
+
+# step CLI base64 standard to URL encoding for input
+encrypted_share_b64_url=$(echo "${encrypted_share}" | sed 's/+/-/g; s/\//_/g')
+
+# Decrypt encrypted share with nonce, member private key and defunct network public key
+decrypted_share=$(echo "${encrypted_share_b64_url}" | step crypto nacl box open base64:"${nonce}" network_enc_pubk.raw member_enc_privk.raw | openssl base64 -A)
+
+# Finally, submit decrypted share
+curl -i -sS --fail https://"${node_rpc_address}"/members/recovery_share/submit "${@}" -H "Content-Type: application/json" -d '{"recovery_share": "'"${decrypted_share}"'"}'
```
achamayou, line 85: The decrypted share would be better placed in a file created by mkstemp, than passed inline where it can be very widely read by other users.

jumaffre: This is now done. For the decrypted share in particular, the `recovery_share/submit` endpoint now takes the raw shares as input so we can pipe the decryption command directly to `curl`.

PR #1297, opened 2020-06-16
Run test suite in CI

jumaffre:
It seems that this was removed. Let's see if this works.


eddyashton:
Can we keep this as a draft until we've had a few green runs? I'm happy with the change if the CI is, but I don't think we should merge as soon as we have an approving review + 1 green build.

jumaffre:
This PR adds the `recovery_test_suite` (~100 secs) and `full_test_suite` (~67 secs) to the CI. These tests are also run randomised in the daily build (roughly the same duration).

PR #1309, opened 2020-06-17
Use backward-cpp to produce stacktrace on sigfaults

eddyashton:
In virtual enclaves and host apps, we install signal handlers via `backward-cpp` which should produce nice stacktraces for us.


src/ds/stacktrace_utils.h
```diff
@@ -31,20 +31,4 @@ namespace logger
     print_stacktrace();
     exit(1);
   }
-
-  static inline void Init(const char* file_name_extention)
-  {
-    signal(SIGSEGV, handler); // install our handler
```
jumaffre, line 37: Should `handler()` go too?

eddyashton: Sure, unless this is still used for debugging. @ashamis?

ashamis: This should be replaced by 

#ifndef INSIDE_ENCLAVE
"backward::SignalHandling sh;"
#endif

from https://github.com/bombela/backward-cpp

Issue #1296, opened 2020-06-16
Discuss handler installation API


eddyashton:
NB: All terminology in this issue is in-flux. I'm just trying to describe a distinction, not mandate that we consolidate on "transaction" vs "command".

We've established that there are several different types of RPC that CCF provides:
- Read-write transactions. These use a `kv::Tx`, and (sometimes) call `TxView::put`.
- Read-only transactions. These use a `kv::Tx`, but never call `TxView::put` (only `TxView::get`).
- Non-KV commands. These don't have any impact on the KV. They read node-local state like the current commit point, metrics, etc.
- Historical queries.

I think we should distinguish these types of RPC when they are installed, so we can call them with different signatures and potentially take different execution paths when handling them. This issue is for sketching and discussing possible C++ API designs.

Currently:
```
install("foo", [this](ccf::RequestArgs& args){}, ReadWrite::Write)
  .set_http_get_only();
```

We have some adapters for the lambda, so most 'real' handlers look like:
```
install(
  "foo",
  json_adapter([this](kv::Tx& tx, json&& params){}),
  ReadWrite::Write)
    .set_http_get_only();
```

This proposes that we have different `install` functions taking different lambda types:
```
install_transaction(
  "foo"
  [this](kv::Tx& tx, json&& params){} ...);

install_read_only_transaction(
  "foo"
  [this](kv::ReadOnlyTx& tx, json&& params){} ...); // ReadOnlyTx is a proposed new type, with no put or remove

install_command(
  "foo"
  [this](json&& params){} ...); // Takes no Tx! Hopefully we didn't even construct one
```

2 related points:
1) I think we should make  (_method_, _verb_, _lambda_) the most-critical, set-at-construction fields, and do a _build-then-install_ pattern instead of _install-then-modify_:
```
make_handler("foo", HTTP_GET, [this](ccf::RequestArgs& args) {}).
  .install();
```
This lets us do validation of the _final_ handler, and gives us flexibility in dispatch (removes the ability to modify how we select the handler, after its been installed)

2) We'd like to support some kind of wildcarding in path dispatch, related to the OpenAPI doc. Very strawman, but the idea is to give something like:
```
install(
  "foo/{userId}/bar",
  [this](..., const ParsedArgs& ?from_path?) { ... }
  ... );
```

eddyashton:
[...]

PR #1310, opened 2020-06-17
Update endpoint installation API

eddyashton:
This is a monster PR, implementing the suggestions in #1296 and some related cleanup.

Opening as a draft for further discussion, happy to break this into incremental PRs for easier consumption, though each will still touch many files.

I've collapsed the latter stages into individual commits so they can be viewed independently. I thought these were neat single features, but on closer inspection they still combine some renames/code shuffling with the intended changes. In short:

- `install("foo", FUNCTION, ReadWrite);` became `make_handler("foo", VERB, FUNCTION).install();`. The VERB must always be supplied, `ReadWrite` is inferred (but _can_ be changed with `set_read_write`), and the handler must be configured _before_ installation.
- In [this commit](https://github.com/microsoft/CCF/commit/88b89067769396e7a43849e6ced3be84242055c1), I removed `consts.h`. We were hardly using any of these definitions, and I think its much easier to read/search for `make_handler("foo"...` than `make_handler(SomeConstexprSomewhere...`. Also naming is hard, and this stops us having to find names for the namespaces for the names of URIs. This also moves installation straight after the handler definition, I don't know why we separated them?
- In [this commit](https://github.com/microsoft/CCF/commit/d1b181a8bf956063e7dc97d2a04697c2d74722ac), I removed `MayWrite`. It was unused, now we don't pretend to support it.
- In [this commit](https://github.com/microsoft/CCF/commit/ac9f1f139fa76136fea46029aba709cda9e03e73), I renamed `RequestArgs` to `HandlerArgs` and created a distinction between "commands" which don't work with the KV, and normal transactions. We have `make_command_handler` to install them, and their lambdas have a different signature. Note that currently we basically just wrap a lambda and don't pass the `kv::Tx`. This sets up the app-facing API, and in future we may be able to take a different execution path for these commands behind the scenes.
- In [this commit](https://github.com/microsoft/CCF/commit/23712efc3fe65f1b5ecd936ff11d0d28cf713508) I added read-only TxViews, Txs, and handlers. Annoyingly its hard to make a lot of our code generic to take either a `Tx` or a `ReadOnlyTx`, and we don't know what handler we're dispatching to until a long time after we constructed the `Tx` (which may have written a signature!), so everything's writeable behind the scenes. But again, I think this is the wall we need on the app-facing API, and we can work out what optimisations this lets us do in the framework later.

TODOs:
- [X] `handler` is a horrible word. I decided `RequestArgs` was wrong, but `HandlerArgs` is uselessly generic. Maybe this is `make_endpoint`, with a lambda taking `EndpointContext&`? **I've gone for `Endpoint`, with a broad rename touching a great many things. I've kept the old typedefs, but deprecated them, so apps _should_ still compile - still need to test/demo this**
- [X] Perhaps `make_handler` should return a smart pointer? Currently we can easily copy the object, so switching between copies and refs is fine, but we may add something uncopyable in future so better to change now? Also exposing an abstract type will give us more freedom to fiddle with the implementation, separately from the public API. **Haven't found a good way to handle ownership of the returned abstract type if we can't just copy it. Going to leave this one undone**
- [X] We should change `set_read_write` to `set_always_forward`. That's what it means, and we now have a better definition of `read-only`. **Done [here](https://github.com/microsoft/CCF/pull/1310/commits/01e9630c1dadb6dac31239cfde677f840423c401).**
- [X] Documentation. **Primarily `logging_cpp.rst` updates, to describe the new process and try to clarify some of the terminology.**
- [X] We currently call `get_view` all the time, for read-only and writeable views. Perhaps this should be explicitly `get_read_only_view`, available on both read-only and writeable Txs? Need to stare at this a bit, think its neater. **I've now done this, and I think its a much better approach. Handler code needs to change slightly more to become read-only, but it was already opt-in. And now we can build a much cleaner idea of a transaction which has _never asked to create something writeable_, so we know its read-only easily _after_ execution, even if we can't beforehand*


achamayou:
Oatpp uses "endpoint": https://oatpp.io/examples/crud/, I agree it's a better name.

src/ds/ccf_deprecated.h
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#define CCF_DEPRECATED(reason) [[deprecated(reason)]]
```
achamayou, line 5: Why do we wrap this? That's standard as of 14, the wrapper isn't shorter.

achamayou:
This looks great!

src/ds/ccf_deprecated.h
eddyashton: The main reason is that `clang-format` seemed very confused by this attribute. Either the formatter splits it across multiple lines, or we add extra lines with `//clang-format off/on` - the macro keeps this as a single line.

As minor side advantages we can easily grep for exactly these lines, we have a single place to disable them if we ever need to, we can add versioned macros (`CCF_DEPRECATED_1_0_1`) to find the things removed in specific versions.

src/node/rpc/handler_registry.h
```diff
@@ -143,6 +159,25 @@ namespace ccf
         return set_auto_schema<typename T::In, typename T::Out>();
       }
 
+      ForwardingRequired forwarding_required = ForwardingRequired::Always;
+
+      /** Override whether a handler is always forwarded, or whether it is safe
+       * to sometimes execute on followers
```
jumaffre, line 165: Minor: document arg and return time too.

src/ds/ccf_deprecated.h
achamayou: That seems like the kind of thing upgrading to a newer clang-format and/or tweaking the config might solve?

I do like the versioned deprecated, especially if we had an automated check for it (ie. cutting X.Y fails until all deprecated X.Y are removed).

Issue #1213, opened 2020-05-26
Consensus does not tick while recovering private ledger


jumaffre:
The state of the node guards whether the consensus `periodic()` function is called. Currently, it is only when the node is in `partOfNetwork` and `partOfPublicNetwork` that the consensus's view of time is updated.

However, during private recovery, the node transitions from `partOfPublicNetwork` to `readingPrivateLedger` while recovering the private ledger. In this state, the consensus is no longer updated which means that primary changes can no longer occur. It also means that the last submitted recovery share cannot be checked for global commit.

I have tried to fix it by also calling `periodic()` on `readingPrivateLedger` as well but this caused spurious primary changes while recovering the private ledger. 

I will investigate this when the current recovery issues get closed.

PR #1308, opened 2020-06-17
PBFT test failures

jumaffre:
New `long_test` label to exclude from the CI and only run in the daily. 

Also minor changes to the PBFT tests adding timeouts.

Issue #1275, opened 2020-06-11
Ledger Python infra should be able to work on ledger chunks


jumaffre:
Follow-up from https://github.com/microsoft/CCF/issues/1135.

Now that the ledger is split into multiple chunks, the Python `ledger.py` infra class should be able to deal with multiple chunks. For now, the chunking threshold is high enough that the `governance_history` test is guaranteed to only create one chunk.

jumaffre:
Closing this as this got implemented in https://github.com/microsoft/CCF/pull/1408

PR #1312, opened 2020-06-17
Tick consensus while recovering private ledger

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1213

We now tick the consensus while reading the private ledger. This wasn't the case before, which meant that the last submitted recovery share was not committed until the private ledger was fully recovered.

Also removed the raft suspension mechanism which was used on recovery to make sure that no private transaction was committed while the private ledger was recovered. This is now redundant as the user frontend (the only one that writes private transactions) is closed until the recovery is complete. The `users` and `user_certs` tables are also now public so that all governance tables are public - there was no point keeping them private as the members' proposals to add new users are public anyway!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -469,6 +469,8 @@ if(BUILD_TESTS)
     NAME governance_history_test
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/governance_history.py
     CONSENSUS raft
+    ADDITIONAL_ARGS # https://github.com/microsoft/CCF/issues/1275
+                    --ledger-chunk-max-bytes 20MB
   )
 
   add_e2e_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -556,12 +556,16 @@ if(BUILD_TESTS)
     NAME suspend_nodes_pbft
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/suspend_nodes.py
     CONSENSUS pbft
+    LABEL long_test
+    ADDITIONAL_ARGS --seed 42
   )
 
   add_e2e_test(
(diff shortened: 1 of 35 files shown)


CMakeLists.txt
```diff
@@ -482,10 +482,6 @@ if(BUILD_TESTS)
       # purpose to check that a recovery network is robust to a view change.
       --raft-election-timeout
       4000
-      # Smaller ledger chunks than default to make sure that a ledger made of
-      # multiple chunks can be recovered
-      --ledger-chunk-max-bytes
-      20KB
```
jumaffre, line 488: All end-to-end tests use this value now, so we test chunking for all tests.

src/node/node_state.h
```diff
@@ -1347,7 +1355,9 @@ namespace ccf
               "Unexpected: write to service table does not include key 0");
           }
 
-          if (it->second.value().status == ServiceStatus::OPEN)
+          if (
+            it->second.value().status == ServiceStatus::OPEN &&
+            this->is_part_of_network())
```
jumaffre, line 1360: On recovery, we used to open the user frontend way too early when we were replaying the previous service opening.

achamayou:
@jumaffre on the governance_history test, you'll want:

```
    ADDITIONAL_ARGS
      --ledger-chunk-max-bytes
      100Mb
```

Otherwise it chunks and the bit of the logic (circa :66) that assumes it doesn't will fail.

I'd have done it with a suggestion inline, but GitHub doesn't allow it more than a few lines away from an existing change in the PR.

jumaffre:
@achamayou Yup! Done now. It is linked to https://github.com/microsoft/CCF/issues/1275.

achamayou:
@jumaffre it probably makes sense to merge #1308 to unlock the build on this.

Issue #1299, opened 2020-06-16
Support for KV store snapshot


jumaffre:
It should be possible to take a snapshot of the KV store at a given version (or at the last globally committed version) and successfully apply it to another KV store.

Note that for public recovery, the state of the public tables in the snapshot should only be integrity protected.

jumaffre:
As discussed with @ashamis, let's go for a snapshot serialisation scheme that is similar to the KV write sets for now as it means more code paths that we can re-use. We can then build a benchmark and iterate on the design later but this should unblock other snapshot-related tickets quicker.

jumaffre:
I believe this is done as we can serialise/deserialise snaphots between store. As per my earlier comment, we can still optimise performance on this but I think raising a separate issue for it is preferable. @ashamis, feel free to close this ticket if you think this is done and I'll create a new ticket for performance improvement.

PR #1313, opened 2020-06-17
Serialize the Maps

ashamis:
- serialize all the maps
- only lock the maps when taking a copy of the required champ map
- add key and value serializers to enable serializing complex types

part of #1299

### DIFF
--- a/src/ds/champ_map.h
+++ b/src/ds/champ_map.h
@@ -412,17 +412,19 @@ namespace champ
       K* k;
       Hash h_k;
       V* v;
-      uint32_t key_size;
-      uint32_t value_size;
 
-      pair(K* k_, Hash h_k_, V* v_, uint32_t k_size, uint32_t v_size) : k(k_), h_k(h_k_), v(v_), key_size(k_size), value_size(v_size) {}
+      pair(K* k_, Hash h_k_, V* v_) : k(k_), h_k(h_k_), v(v_) {}
     };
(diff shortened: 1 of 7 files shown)


src/ds/champ_map.h
```diff
@@ -415,37 +415,57 @@ namespace champ
...
-        serialized_state.push_back(pair(k, static_cast<Hash>(H()(key)), v));
+        uint32_t key_size = k_size(key) + get_padding(k_size(key));
+        uint32_t value_size = v_size(value) + get_padding(v_size(value));
+
+        size += (key_size + value_size);
 
-        size += (sizeof(uint64_t) * 2); // headers
-        size += k_size(key) + get_padding(k_size(key));
-        size += v_size(value) + get_padding(k_size(value));
+        serialized_state.push_back(pair(k, static_cast<Hash>(H()(key)), v));
```
achamayou, line 468: Might as well emplace_back()?

src/kv/store.h
```diff
@@ -215,6 +215,35 @@ namespace kv
       return *result;
     }
 
+    std::unique_ptr<Snapshot> snapshot(Version v) override
+    {
+      std::lock_guard<SpinLock> mguard(maps_lock);
+
+      if (v > current_version())
+      {
+        return nullptr;
```
achamayou, line 224: May as well throw right away?

Issue #959, opened 2020-03-16
Sample app buildable standalone


jumaffre:
It is not currently easy to test that an app can be built against a CCF install before a new version of CCF is released. This means that it is possible that we release a version of CCF that cannot be used out-of-the-box to build a CCF app (like it was the case for all versions until v0.8.2). 

We could fix this by making one of our sample apps (probably `logging` as it is the one we document) buildable standalone. It would then be easy to, from our release pipeline, install CCF, checkout the _same_ repo and build the app pointing at the freshly-installed CCF.

achamayou:
Dupe #1028

PR #1316, opened 2020-06-18
Clean up and simplify quickstart

### LINKED ISSUE #1186: Quickstart split
Quickstart should be split for the following use cases:

- [ ] Build a CCF App
- [ ] Contribute to CCF
- [ ] Deploy a CCF App

Since we have different playbooks and containers for these scenarios now. The README ought to direct users at the relevant quickstart.

achamayou:
fix #1186 

Can't quite go all the way without #959, but I think the key bits here are highlighting the availability of containers, and making clear they are the easier/preferred route for users trying to build an app, as opposed to setting up an ACC VM.


doc/developers/build_app.rst
```diff
@@ -73,3 +73,18 @@ Additionally, the `cchost` binary must be told that the enclave type is debug:
...
 
     $ cchost --enclave-file liblua_generic.enclave.so.debuggable --enclave-type debug [args]
+
+Build Container
+---------------
+
+With every release of CCF, a base build container is provided to facilitate reproducible builds and continuous integration.
+It contains everything needed to build and test CCF applications.
+
+.. literalinclude:: ../../docker/app_ci
```
jumaffre, line 83: Is there much value including the `Dockerfile` here?

doc/operators/container.rst
```diff
@@ -0,0 +1,23 @@
+Runtime Container
+=================
+
+With every release, a sample runtime base container is provided. It contains the ``cchost`` executable and its dependencies,
```
ashamis, line 4: we should probably say for "minor release", i.e. we might provide a containers for release candidates or for some minor release (depending on need)

doc/developers/build_app.rst
achamayou: I think it’s useful to have a reference to it if you want to tweak it, I’ll see if I can change it to a link somehow.

doc/operators/container.rst
achamayou: I’m not sure I understand your point. I would think that we provide containers for anything that is on the Release page of GitHub, regardless of how minor it is?

doc/developers/build_app.rst
achamayou: Can't find a nice way to do that. It's easy for Python code, but not for arbitrary files.

doc/quickstart/index.rst
```diff
@@ -1,31 +1,113 @@
...
+Container
+---------
 
-The install directory contains (under ``getting_started/``) the Azure and Ansible scripts required to setup the CCF environment. See :ref:`how to setup the CCF environment here <quickstart/requirements:Azure Confidential Compute>`.
+The quickest way to get started building CCF applications is to use the :ref:`developers/build_app:Build Container`.
 
-.. note::
+.. code-block:: bash
+
+    sudo docker run -it ccfciteam/ccf-app-ci:0.11
```
eddyashton, line 11: Should we say `:latest` for docker versions, rather than keeping these docs in-sync manually?

achamayou: We don't have a latest (yet), but that is an excellent point.

doc/operators/container.rst
achamayou: @ashamis ?

doc/quickstart/index.rst
achamayou: Done.

doc/operators/container.rst
ashamis: sure, just wanted to make sure this was part of the plan

Issue #1113, opened 2020-04-29
All members should ACK before transitioning to ACTIVE


jumaffre:
Follow up from https://github.com/microsoft/CCF/pull/1107#discussion_r416559996

All members added pre network opening are currently added as `ACTIVE` and can propose/vote without first issuing a `ack`. This is not ideal for usability/auditability as bogus member certificates could be added in that phase (either when the first node is started or later, via the `new_member` proposal). Instead, all members should first be added as `ACCEPTED` and only transition to `ACTIVE` once they have issued their first `ack`. 

One of the issues with this is the recovery threshold `t` set in the `ccf.config` table when the first node starts up (`--recovery-threshold` option). Implementing the changes described above imply that the only sensible initial value for the recovery threshold is `1` (note: `N` is the number of active members): when the first members acks, `t=1/N=1` recovery shares are created. If the recovery threshold was higher, the secret sharing scheme would fail as `t` > number of active members. 

Instead, I propose that recovery shares are only issued when sufficient members have become active. The corollary of this is that the service can only be recovered once `t` members have become active - which I believe is acceptable. 

We also need to make sure that when the network is open, some recovery shares have been issued (i.e. more than `t` members have become active). Otherwise, the `open_network` proposal should fail. For example, if `t = 3` and there are `N=2` active members, the `open_network` proposal could be voted in by these 2 members with no recovery shares issued.

PR #1318, opened 2020-06-18
All members should ack before becoming active...

jumaffre:
... even when the service is `OPENING`.

The service checks that on opening the number of active members is `>=` than the recovery threshold (otherwise, we would be opening a service that is not recoverable). Consequently, recovery shares are now only issued when the service is opened.

Resolves https://github.com/microsoft/CCF/issues/1113


src/node/genesis_gen.h
```diff
@@ -148,6 +131,29 @@ namespace ccf
...
+        throw std::logic_error(fmt::format(
+          "Member {} cannot be activated as they do not exist", member_id));
+      }
+
+      if (member->status == MemberStatus::ACCEPTED)
+      {
+        member->status = MemberStatus::ACTIVE;
+        if (get_active_members_count() >= max_active_members_count)
+        {
+          throw std::logic_error(fmt::format(
```
ashamis, line 149: consider using cc_logic_error, this will also produce a stack trace when running in virtual

achamayou: @ashamis should we just do a pass and replace'em all?

src/node/rpc/test/member_voting_test.cpp
```diff
@@ -1777,78 +1797,113 @@ DOCTEST_TEST_CASE("Maximum number of active members")
...
+        crypto::BoxKey::public_from_private(private_encryption_key));
+      auto id = gen.add_member(cert, public_encryption_key.raw());
+      members[id] = {cert, private_encryption_key};
     }
+    gen.set_recovery_threshold(recovery_threshold);
     gen.finalize();
+    frontend.open();
+  }
+
+  DOCTEST_INFO("Opens fails as recovery threshold is too high");
```
achamayou, line 1880:
```suggestion
  DOCTEST_INFO("Open fails as recovery threshold is too high");
```

src/node/genesis_gen.h
jumaffre: Raised https://github.com/microsoft/CCF/issues/1324 as it is something we should do consistently in our codebase.

PR #1319, opened 2020-06-18
RPC to retrieve IDs assigned by CCF to a given node's RPC address

### LINKED ISSUE #1247: Retrieve CCF-assigned Node ID By RPC Address
When a CCF node crashes after the network is open for user transactions and is brought back up with the same network identity (i.e. RPC address), two things need to happen:
1. The old node needs to be retired.
2. The new node needs to be trusted.

The proposals for retiring and trusting nodes must include the ID assigned to that node by CCF. While ID assignment has a known behavior (increment from 0), there should be a method to retrieve a node's ID instead of requiring the network operator to independently keep track of the ID.

An RPC address is unique within the CCF network and provided to each `cchost` process at startup by the operator, so it provides a way of uniquely identifying a node and retrieving its ID using information already known by the operator. Thus a method for querying the node ID by RPC address is requested; please view the [API Spec on Swagger](https://app.swaggerhub.com/apis/lynshi/CCF_GetNodeIDByRPC/1.0).

lynshi:
Added the handler to common handlers and updated documentation (resolves #1247).

The IDs are retrieved by looping over the ccf.nodes table as is done by the getNetworkInfo RPC. If faster querying is required, a reverse map from RPC addresses to node IDs needs to be created and maintained.


src/node/rpc/common_endpoint_registry.h
```diff
@@ -195,6 +195,29 @@ namespace ccf
         .set_auto_schema<void, GetNetworkInfo::Out>()
         .install();
 
+      auto get_nodes_by_rpc_address = [this](auto& args, nlohmann::json&& params) {
```
eddyashton, line 198:
```suggestion
      auto get_nodes_by_rpc_address = [this](
                                        auto& args, nlohmann::json&& params) {
```
Minor formatting fix.

PR #1320, opened 2020-06-19
Deserialize the KV

ashamis:
Part of #1299

### DIFF
--- a/src/ds/champ_map.h
+++ b/src/ds/champ_map.h
@@ -345,13 +345,13 @@ namespace champ
       while (size != 0)
       {
         // Deserialize the key
-        uint64_t key_size = size;
+        size_t key_size = size;
         K key = make_k(data, size);
         key_size -= size;
         serialized::skip(data, size, get_padding(key_size));
 
         // Deserialize the value
-        uint64_t value_size = size;
+        size_t value_size = size;
         V value = make_v(data, size);
         value_size -= size;
         serialized::skip(data, size, get_padding(value_size));
--- a/src/ds/champ_map.h
+++ b/src/ds/champ_map.h
@@ -344,11 +344,13 @@ namespace champ
 
       while (size != 0)
       {
+        // Deserialize the key
         uint64_t key_size = size;
         K key = make_k(data, size);
         key_size -= size;
         serialized::skip(data, size, get_padding(key_size));
 
+        // Deserialize the value
         uint64_t value_size = size;
         V value = make_v(data, size);
         value_size -= size;
--- a/src/ds/champ_map.h
+++ b/src/ds/champ_map.h
@@ -335,8 +335,8 @@ namespace champ
 
     Map<K, V, H> static deserialize_map(
       const std::vector<uint8_t>& serialized_state,
-      std::function<const K&(const uint8_t*, uint32_t)> make_k,
-      std::function<const V&(const uint8_t*, uint32_t)> make_v)
+      std::function<K(const uint8_t*&, size_t&)> make_k,
+      std::function<V(const uint8_t*&, size_t&)> make_v)
     {
       Map<K, V, H> map;
       const uint8_t* data = serialized_state.data();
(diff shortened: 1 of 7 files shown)


src/ds/champ_map.h
```diff
@@ -345,16 +345,17 @@ namespace champ
       while (size != 0)
       {
         // Deserialize the key
-        auto key_size = serialized::read<uint64_t>(data, size);
-        const uint8_t* key = data;
-        serialized::skip(data, size, key_size + get_padding(key_size));
+        uint64_t key_size = size;
```
jumaffre, line 348: `size_t`

jumaffre: Also, this seems inconsistent with the type of `key_size` in `get_buffer()`.

src/kv/store.h
```diff
@@ -219,35 +219,76 @@ namespace kv
...
+    }
 
+    std::unique_ptr<Snapshot> snapshot(Version v) override
+    {
+      auto snapshot = std::make_unique<Snapshot>(v);
+
+      {
+        std::lock_guard<SpinLock> mguard(maps_lock);
+
+        if (v > current_version())
```
jumaffre, line 268: We also need to check that `v >= commit_version()`.

src/kv/store.h
```diff
@@ -219,35 +219,76 @@ namespace kv
...
     {
-      std::lock_guard<SpinLock> mguard(maps_lock);
-
-      if (v > current_version())
       {
-        throw ccf::ccf_logic_error(fmt::format(
-          "Attempting to snapshot at invalid version v:{}, current_version:{}",
-          v,
-          current_version()));
+        std::lock_guard<SpinLock> vguard(version_lock);
```
jumaffre, line 225: It looks like these changes should be applied _after_ the snapshot has been applied.

src/kv/untyped_map.h
```diff
@@ -682,7 +690,41 @@ namespace kv::untyped
...
+
+      std::function<SerialisedEntry(const uint8_t*& key, size_t&)> make_k =
+        [](const uint8_t*& data, size_t& size) -> SerialisedEntry {
+        uint64_t key_size = serialized::read<uint64_t>(data, size);
+        SerialisedEntry ret;
+        ret.assign(key_size, *data);
+        serialized::skip(data, size, key_size);
+        return std::move(ret);
+      };
+      std::function<kv::VersionV<SerialisedEntry>(const uint8_t*& key, size_t&)>
```
jumaffre, line 712: `kv::VersionV<SerialisedEntry>` ->`VersionV`

jumaffre: Same in `snapshot()`.

src/kv/untyped_map.h
```diff
@@ -682,7 +690,41 @@ namespace kv::untyped
...
+        roll.commits->get_head() == roll.commits->get_tail(),
+        "We are apply a snapshot and there are pending commits");
+
+      std::function<SerialisedEntry(const uint8_t*& key, size_t&)> make_k =
+        [](const uint8_t*& data, size_t& size) -> SerialisedEntry {
+        uint64_t key_size = serialized::read<uint64_t>(data, size);
+        SerialisedEntry ret;
+        ret.assign(key_size, *data);
+        serialized::skip(data, size, key_size);
+        return std::move(ret);
```
jumaffre, line 710: No need for `std::move()` here?

src/kv/store.h
ashamis: I am simplifying this but the order should not matter. We are skipping forward an arbitrary number of version so there should not be any other transactions in progress.

src/kv/store.h
ashamis: why >= ? Should we not be able to create a snapshot of the current version?

jumaffre: `commit_version()` refers to the version at which the map was compacted (i.e. the first element in the roll). We should not be able to snapshot before this version.

PR #1326, opened 2020-06-23
Support make as CMake Generator

eddyashton:
Trying to build with `make` rather than `ninja` would produce an error, saying that the enclave `.so` files weren't generated. This fixes that. `make` is _much_ slower than `ninja`, so I don't think we need to maintain this, but its nice if it works.

### DIFF
--- a/.azure-pipelines-templates/daily-matrix.yml
+++ b/.azure-pipelines-templates/daily-matrix.yml
@@ -19,7 +19,8 @@ parameters:
       cmake_args: '-DCMAKE_INSTALL_PREFIX=/tmp/ccf-install'
 
 jobs:
-  - job: 'Make Generator'
+  - job: 'Make'
+    displayName: 'Make generator'
     ${{ insert }}: '${{ parameters.env.NoSGX }}'
     variables:
       skipComponentGovernanceDetection: true
--- a/.azure-pipelines-templates/daily-matrix.yml
+++ b/.azure-pipelines-templates/daily-matrix.yml
@@ -33,24 +33,24 @@ jobs:
           cmake -L .. ${{ parameters.build.NoSGX.cmake_args }}
         displayName: CMake
 
-  # - template: common.yml
-  #   parameters:
-  #     target: NoSGX
-  #     env: '${{ parameters.env.NoSGX }}'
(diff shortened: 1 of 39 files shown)


achamayou:
If we're going to support it, let's add it to the Daily.

achamayou:
```
mkdir make_build
cd make_build
cmake ..
```

.azure-pipelines-templates/daily-matrix.yml
```diff
@@ -19,24 +19,37 @@ parameters:
...
+  #   parameters:
+  #     target: SGX
+  #     env: '${{ parameters.env.SGX }}'
+  #     cmake_args: '${{ parameters.build.install.cmake_args }} ${{ parameters.build.SGX.cmake_args }} -DTLS_TEST=ON -DSHUFFLE_SUITE=ON'
+  #     suffix: 'Release'
+  #     artifact_name: 'SGX_Release'
+  #     ctest_filter: '-LE "benchmark|perf"'
+  #     ctest_timeout: '240'
+  #     install_prefix: '${{ parameters.build.install.install_prefix }}'
+  #     tls_report: true
```
achamayou, line 56:
```suggestion
  - template: common.yml
    parameters:
      target: NoSGX
      env: '${{ parameters.env.NoSGX }}'
      cmake_args: '${{ parameters.build.debug.cmake_args }} ${{ parameters.build.NoSGX.cmake_args }}'
      suffix: 'Coverage'
      artifact_name: 'NoSGX_Coverage'
      ctest_filter: '-LE "benchmark|perf|long_test"'
      ctest_timeout: '300'

  - template: common.yml
    parameters:
      target: SGX
      env: '${{ parameters.env.SGX }}'
      cmake_args: '${{ parameters.build.install.cmake_args }} ${{ parameters.build.SGX.cmake_args }} -DTLS_TEST=ON -DSHUFFLE_SUITE=ON'
      suffix: 'Release'
      artifact_name: 'SGX_Release'
      ctest_filter: '-LE "benchmark|perf"'
      ctest_timeout: '240'
      install_prefix: '${{ parameters.build.install.install_prefix }}'
      tls_report: true
```

eddyashton: Done - just had these disabled temporarily to test the new job in isolation.

PR #1328, opened 2020-06-23
Allow passing in a buffer when serializing the champ_maps

ashamis:
+ other cleanup

part of #1299

### DIFF
--- a/src/ds/champ_map.h
+++ b/src/ds/champ_map.h
@@ -73,7 +73,7 @@ namespace champ
     }
   };
 
-  template <class K, class V, class H>
+  template <class K, class V, class H, class k_size, class v_size>
   struct SubNodes;
 
   template <class K, class V>
@@ -103,10 +103,18 @@ namespace champ
     return padding;
   }
 
+  template<class K, class V, class k_size, class v_size>
+  uint32_t static get_size_with_padding(const K& k, const V& v)
+  {
+    uint32_t size_k = k_size()(k);
+    uint32_t size_v = v_size()(v);
+    return size_k + get_padding(size_k) + size_v + get_padding(size_v);
+  }
+
   template <class K, class V, class H>
   using Node = std::shared_ptr<void>;
 
-  template <class K, class V, class H>
+  template <class K, class V, class H, class k_size, class v_size>
   struct Collisions
   {
     std::array<std::vector<std::shared_ptr<Entry<K, V>>>, collision_bins> bins;
@@ -123,7 +131,8 @@ namespace champ
       return nullptr;
     }
 
-    bool put_mut(Hash hash, const K& k, const V& v)
+    // TODO: we can return the size here
+    size_t put_mut(Hash hash, const K& k, const V& v)
     {
       const auto idx = mask(hash, collision_depth);
       auto& bin = bins[idx];
@@ -133,11 +142,11 @@ namespace champ
         if (k == entry->key)
         {
           bin[i] = std::make_shared<Entry<K, V>>(k, v);
-          return false;
+          return k_size()(k) + v_size()(v);
         }
       }
       bin.push_back(std::make_shared<Entry<K, V>>(k, v));
-      return true;
+      return 0;
     }
 
     template <class F>
@@ -153,7 +162,7 @@ namespace champ
     }
   };
 
-  template <class K, class V, class H>
+  template <class K, class V, class H, class k_size, class v_size>
   struct SubNodes
   {
     std::vector<Node<K, V, H>> nodes;
@@ -194,12 +203,12 @@ namespace champ
         return node_as<Entry<K, V>>(c_idx)->getp(k);
 
       if (depth == (collision_depth - 1))
-        return node_as<Collisions<K, V, H>>(c_idx)->getp(hash, k);
+        return node_as<Collisions<K, V, H, k_size, v_size>>(c_idx)->getp(hash, k);
 
-      return node_as<SubNodes<K, V, H>>(c_idx)->getp(depth + 1, hash, k);
+      return node_as<SubNodes<K, V, H, k_size, v_size>>(c_idx)->getp(depth + 1, hash, k);
     }
 
-    bool put_mut(SmallIndex depth, Hash hash, const K& k, const V& v)
+    size_t put_mut(SmallIndex depth, Hash hash, const K& k, const V& v)
     {
       const auto idx = mask(hash, depth);
       auto c_idx = compressed_idx(idx);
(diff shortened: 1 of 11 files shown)


src/kv/untyped_tx_view.h
```diff
@@ -12,12 +12,37 @@ namespace kv::untyped
...
+    {
+      return sizeof(uint64_t) + sizeof(value.version) + value.value.size();
+    };
+  };
+
+  using State = kv::State<
+    SerialisedEntry,
+    SerialisedEntry,
+    SerialisedKeyHasher,
+    k_size,
```
eddyashton, line 36: It's unfortunate if we have to add these templates to all the kv types, and not just `Snapshot`. Why is this necessary?

src/kv/untyped_tx_view.h
```diff
@@ -12,12 +12,37 @@ namespace kv::untyped
...
 
   using VersionV = kv::VersionV<SerialisedEntry>;
-  using State =
-    kv::State<SerialisedEntry, SerialisedEntry, SerialisedKeyHasher>;
+
+  struct k_size
+  {
+    uint32_t operator()(const SerialisedEntry& key)
+    {
+      return sizeof(uint64_t) + key.size();
```
eddyashton, line 20: This fixes the _size_ calculation logic at type-definition, but the actual serialisation is still an arbitrary function - there's a stronger connection between these 2 functors, so they should be kept together. If all we're really doing is calling `serialized::write<T>`, we could add `serialized::size<T>` to return the required size?

And what's the motivation for making this snapshot-serialisation pluggable - can we just call `serialized::` directly in `champ_map.h` rather than a templated or member functor?

src/kv/kv_types.h
```diff
@@ -363,11 +363,13 @@ namespace kv
...
     public:
       virtual ~Snapshot() = default;
-      virtual std::vector<uint8_t> get_buffer() = 0;
+      virtual void serialize(uint8_t* data) = 0;
+      virtual size_t get_serialized_size() = 0;
       virtual std::string& get_name() = 0;
       virtual SecurityDomain get_security_domain() = 0;
       virtual bool get_is_replicated() = 0;
       virtual kv::Version get_version() = 0;
+      virtual CBuffer& get_serialized_buffer() = 0;
```
eddyashton, line 372: Minor: I think this should return `const CBuffer&`.

src/kv/untyped_tx_view.h
ashamis: discussed offline. We came to a different solution. All the serialization, deserialization, size, ... are now in a separate file. We should look at this again before pushing the PR.

src/ds/champ_map.h
```diff
@@ -318,41 +327,54 @@ namespace champ
...
+  template <class T>
+  struct default_size
+  {
+    uint32_t operator()(T type) const
+    {
+      return champ::get_size<T>(type);
+    }
+  };
+
+  static std::atomic<uint32_t> foobar = 0;
```
jumaffre, line 339: To delete?

src/ds/champ_map_serializers.h
```diff
@@ -0,0 +1,139 @@
...
+  {
+    return sizeof(uint64_t) + sizeof(data.version) + data.value.size();
+  }
+
+  template <class T>
+  inline size_t serialize(const T& t, uint8_t*& data, size_t& size)
+  {
+    uint64_t data_size = sizeof(T);
+    serialized::write(
+      data, size, reinterpret_cast<const uint8_t*>(&data_size), sizeof(T));
```
jumaffre, line 62: Not that it matters here but the the last parameter should be `sizeof(uint64_t)`?

src/ds/test/map_test.cpp
```diff
@@ -251,13 +217,13 @@ TEST_CASE("serialize map")
...
 
-    champ::Snapshot<K, V, H> snapshot_2(
-      map, fn_size_k, fn_serialize_k, fn_size_v, fn_serialize_v);
-    const std::vector<uint8_t>& s_2 = snapshot_2.get_buffer();
+    champ::Snapshot<K, V, H> snapshot_2(map);
+    std::vector<uint8_t> s_2(map.get_serialized_size());
+    snapshot_2.serialize(s_2.data());
 
     REQUIRE_EQ(s_1.size(), s_2.size());
     for (uint32_t i = 0; i < s_1.size(); ++i)
```
jumaffre, line 229: This `for` could just be replaced with `REQUIRE_EQ(s_1, s_2)`?

src/ds/champ_map.h
```diff
@@ -271,10 +280,10 @@ namespace champ
           nodes.begin() + c_idx,
           std::make_shared<Collisions<K, V, H>>(std::move(sub_node)));
       }
-      return true;
+      return 0;
     }
 
-    std::pair<std::shared_ptr<SubNodes<K, V, H>>, bool> put(
+    std::pair<std::shared_ptr<SubNodes<K, V, H>>, uint32_t> put(
```
jumaffre, line 286: `size_t`?

src/kv/kv_types.h
```diff
@@ -417,6 +422,23 @@ namespace kv
         return snapshots;
       }
 
+      std::vector<uint8_t>& get_buffer()
```
jumaffre, line 425: This is introducing quite a lot of implementation in `kv_types.h`. Could we create a new `kv/snapshot.h` file instead?

jumaffre, line 425: `get_buffer()` seems to only be called from `serialize()` below.

src/kv/kv_types.h
```diff
@@ -417,6 +422,23 @@ namespace kv
...
+      {
+        buffer.resize(serialized_size);
+        return buffer;
+      }
+
+      void serialize()
+      {
+        uint8_t* buffer = get_buffer().data();
+        uint32_t position = 0;
```
jumaffre, line 434: `position` is now unused?

src/kv/store.h
```diff
@@ -290,6 +290,7 @@ namespace kv
           map.second->unlock();
         }
       }
+      snapshot->serialize();
       return std::move(snapshot);
```
jumaffre, line 294: No need for `std::move()` here.

src/ds/champ_map.h
ashamis: thank you!

src/ds/champ_map_serializers.h
```diff
@@ -0,0 +1,139 @@
...
+    return serialized::read<T>(data, size);
+  }
+
+  template <>
+  inline champ::untyped::SerialisedEntry deserialize<
+    champ::untyped::SerialisedEntry>(const uint8_t*& data, size_t& size)
+  {
+    uint64_t data_size = serialized::read<uint64_t>(data, size);
+    champ::untyped::SerialisedEntry ret;
+    ret.assign(data_size, *data);
```
eddyashton, line 121: Mentioned in call, think this should be something like
```suggestion
    champ::untyped::SerialisedEntry ret(data_size);
    ret.assign(data, data + data_size);
```

src/ds/champ_map.h
```diff
@@ -103,6 +104,14 @@ namespace champ
     return padding;
   }
 
+  template <class K, class V>
+  uint32_t static get_size_with_padding(const K& k, const V& v)
```
eddyashton, line 108: We primarily use `size_t` for the sizes, but `uint32_t` here - what do we actually write?

Minor: For safety, we could add a `using SerializedSizeT = size_t;` in `champ_map_serializers`, and then we can distinguish uses in this file from _other_ `size_t`s (and change to `uint32_t` at a single point), but this is almost certainly overkill!

src/ds/champ_map.h
```diff
@@ -318,41 +327,52 @@ namespace champ
     }
   };
 
+  template <class T>
+  struct default_size
```
eddyashton, line 331: Is this struct still used anywhere?

PR #1325, opened 2020-06-22
Rename URI prefixes

eddyashton:
Fixes #1322.

Renames:
- `/users` => `/app`
- `/members` => `/gov`
- `/nodes` => `/node`

If the old names are used, returns a `308 Permanent Redirect` response pointing to the new prefix. We only do this check on the first node, which I think is safe? Perhaps a mismatch (between "what prefix did they give me" and "what prefix would I like to see") in the pbft or forwarded cases should be an error?

PR #1329, opened 2020-06-23
Support multiple HTTP verbs in schema.py

eddyashton:
Opening as a standalone PR for clarity.

The response to `api/schema` used to look like:
```
{
  "params": { ... },
  "result": { ... }
}
```

Recently, with single-verb-per-handler support, it changed to:
```
{
  "get":
  {
    "params": { ... },
    "result": { ... }
  },
  "post": { ... }
}
```

This updates `schema.py` to parse the new format, dumping the potentially distinct schemas into appropriately named files. So all of our schema files have gained a `VERB` in their name. This is all a temporary situation - when we get to a single OpenAPI document the per-handler files will probably go away.


doc/developers/logging_cpp.rst
```diff
@@ -69,7 +69,7 @@ Each function is installed as the handler for a specific HTTP resource, defined
...
 
-The return value from ``make_endpoint`` is an ``Endpoint&`` object which can be used to alter how the handler is executed. For example, the handler for ``LOG_record`` shown above sets a `schema` for the handler, declaring the types of its request and response bodies. These will be used in calls to the ``/api/schema`` endpoint to generate JSON documents describing the API. Since this is the only handler installed for ``"LOG_record"`` only HTTP ``POST``s will be accepted for this URI - the framework will return a ``405 Method Not Allowed`` for requests with any other verb.
+The return value from ``make_endpoint`` is an ``Endpoint&`` object which can be used to alter how the handler is executed. For example, the handler for ``LOG_record`` shown above sets a `schema` for the handler, declaring the types of its request and response bodies. These will be used in calls to the ``/api/schema`` endpoint to generate JSON documents describing the API. Since this is the only handler installed for ``"LOG_record"`` only HTTP ``POST``\s will be accepted for this URI - the framework will return a ``405 Method Not Allowed`` error for requests with any other verb.
```
eddyashton, line 72: reST doesn't allow arbitrary characters directly after inline markup. Github's inline diff actually shows this - the "s will be accepted for this URI..." is part of the marked up section, not after it. The solution seems to be using
```
``POST``\s
```
instead of 
```
``POST``s
```
although this is [discouraged](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#character-level-inline-markup).

doc/developers/logging_cpp.rst
```diff
@@ -99,7 +99,7 @@ If a handler makes no writes to the KV, it may be installed as read-only:
     :end-before: SNIPPET_END: install_get
     :dedent: 6
 
-This offers some additional type safety (accidental `put`s or `remove`s will be caught at compile-time) and also enables performance scaling since read-only operations can be executed on any receiving node, whereas writes must always be executed on the primary node.
+This offers some additional type safety (accidental `put`\s or `remove`\s will be caught at compile-time) and also enables performance scaling since read-only operations can be executed on any receiving node, whereas writes must always be executed on the primary node.
```
achamayou, line 102: Does something like :cpp:func:`kv::TxView::put()` work here?

eddyashton: No, this also gives a Sphinx error that I saw [here](https://github.com/microsoft/CCF/pull/1325/files#diff-27c722e5d73616262ec810f6ccb8ca8d) (and worked around in a different way).

Issue #629, opened 2019-12-11
Support for WebSockets


jumaffre:
As of today, CCF only supports framed JSON-RPC and HTTP. The former is fast but not ideal for client compatibility while the latter causes a large per-request overhead (i.e. HTTP header). 

CCF should also support WebSocket endpoints to maximise requests throughput. 

Tasks:
- [x] Add a WebSocket client to `clients.py`. 
- [x] Support for WebSocket handshake.
- [x] Support for deserialisation of WebSocket frames on CCF side.
- [x] Support for specific user defined handlers (rather than forward to http definition when possible)
- [ ] Protocol-level close and heartbeats
- [ ] Support for masked traffic
- [ ] Stricter negotiation check (*)
- [ ] Client documentation

(*) ws_upgrade.h:53

Our logic does not check for the Connection header, and does not check what protocol the client is requesting an upgrade to: https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#Upgrading_to_a_WebSocket_connection

achamayou:
Looking at https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers, I think there's a number of restrictions we can place on WebSockets support, at least initially.

1. Client masking. This is clearly a complete waste of time and space on a TLS connection. While it's nice to add for compatibility with existing clients, it isn't necessary at first.
2. WebSockets-level fragmentation, ie. supporting multiple frames with FIN 0 followed by a FIN 1. Same as above, there's no clear use case for that in the context of CCF clients (none of the popular client libraries I've looked at even implement it), so it does not need to be there in the initial implementation.
3. binary/text opcode distinction, we can restrict ourselves to binary.
4. The sub-protocol feature may be a nice way to expose the supported FlatBuffer schemas, we don't need to use it at first, but it's potentially something to look into.

achamayou:
Most of this went in with #1193 and #1211, but there are still a few bit and pieces that need sorting out before it's really complete (listed as tickboxes in the OP).

achamayou:
In the absence of user interest for WebSockets, I am shelving this. We will discuss removing support altogether, to reduce TCB and build/test times.

PR #1333, opened 2020-06-24
Explicit websocket handlers

achamayou:
Extend http_method to RESTVerb (alternative naming suggestions welcome), so it's possible to explicitly register WS handlers. Context stays the same for now, we can fork it without user changes in the future if we decide it's useful to do so, but I'm not convinced right now.

Next step for #629

### DIFF
--- a/src/node/rpc/common_endpoint_registry.h
+++ b/src/node/rpc/common_endpoint_registry.h
@@ -3,6 +3,8 @@
 #pragma once
 
 #include "endpoint_registry.h"
+#include "http/http_consts.h"
+#include "http/ws_consts.h"
 #include "json_handler.h"
 #include "metrics.h"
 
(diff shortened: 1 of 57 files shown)


src/node/rpc/common_endpoint_registry.h
```diff
@@ -247,7 +249,7 @@ namespace ccf
 
         for (auto& [verb, endpoint] : it->second)
         {
-          std::string verb_name = http_method_str(verb);
+          std::string verb_name = verb.c_str();
```
ashamis, line 252: is this right? I am assuming that verb is a std::string?

ashamis:
This might be unrelated (or noise) but the websocket perf tests look like they may have regressed. Maybe run the pref tests a couple of times?

src/node/rpc/common_endpoint_registry.h
achamayou: Verb is an RPCVerb now.

achamayou: Here we allocate a copy because we lowercase it for the schema.

achamayou:
@ashamis I think that was just noise, there doesn't seem to be a noticeable difference.

PR #1335, opened 2020-06-25
Update version tag format

eddyashton:
For consistency, we're moving to a format for our release versions. Tags will be prefixed `ccf-` rather than `v`, to match dir name and other (docker, cmake) versions.


eddyashton:
Note to self: Once #1323 goes in, this PR should also update `version.cmake` (specifically, `tr -d ccf-` rather than `tr -d v`).

PR #1337, opened 2020-06-25
Parametrise container playbooks

achamayou:
I think this needs to be merged before we can really try the workflow.


.github/workflows/containers.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: "Build Containers"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build App CI container
+        run: docker build -f docker/app_ci . --build-arg="extra_vars=--extra-vars ccf_ver=$GITHUB_REF"
+        
+      - name: Build App Run container
+        run: docker build -f docker/app_run . --build-arg="extra_vars=--extra-vars ccf_ver=$GITHUB_REF"
```
eddyashton, line 18:
```suggestion
        run: docker build -f docker/app_run . --build-arg="extra_vars=--extra-vars ccf_ver=$GITHUB_REF" -t  $GITHUB_REF
```

.github/workflows/containers.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  build:
+    name: "Build Containers"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build App CI container
+        run: docker build -f docker/app_ci . --build-arg="extra_vars=--extra-vars ccf_ver=$GITHUB_REF"
```
eddyashton, line 15: We should tag these as both the current version and latest. I think that's something like this?

```suggestion
        run: docker build -f docker/app_ci . --build-arg="extra_vars=--extra-vars ccf_ver=$GITHUB_REF" -t  ccfciteam/ccf-app-ci:$GITHUB_REF -t ccfciteam/ccf-app-ci
```

.github/workflows/containers.yml
achamayou: Yes!

.github/workflows/containers.yml
achamayou: For latest, it's not so clear, unless we always release the latest. Let's do it separately when this works.

PR #1340, opened 2020-06-26
Update container dependencies

achamayou:
Recent run containers are missing libc++ and libc++abi, this is adding them explicitly.

Also adding clang-format as a convenience in ccf_build/ccf-app-ci, not strictly required but quite useful when building native apps, and consistent with our inclusion of tools serving similar purposes such as doxygen.

```
root@c4a0ab881a02:/# /usr/bin/cchost -v
CCF host: ccf-0.11.1
root@c4a0ab881a02:/# ldd /usr/bin/cchost | grep libc++
        libc++.so.1 => /usr/lib/x86_64-linux-gnu/libc++.so.1 (0x00007fa01ba8c000)
        libc++abi.so.1 => /usr/lib/x86_64-linux-gnu/libc++abi.so.1 (0x00007fa01b85c000)
```

Edit: I intend to cut 0.11.2, to avoid leaving the app_run container broken.

### DIFF
--- a/getting_started/setup_vm/ccf-dev.yml
+++ b/getting_started/setup_vm/ccf-dev.yml
@@ -13,6 +13,9 @@
         name: openenclave
         tasks_from: install.yml
     - import_role:
+        name: llvm_repo
+        tasks_from: install.yml
+    - import_role:
         name: ccf_run
         tasks_from: install.yml
     - import_role:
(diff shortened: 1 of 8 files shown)


getting_started/setup_vm/ccf-dev.yml
```diff
@@ -12,6 +12,9 @@
     - import_role:
         name: openenclave
         tasks_from: install.yml
+    - import_role:
+        name: llvm_repo
```
jumaffre, line 16: Should this folder be checked in the PR?

achamayou: Yes! Thank you!

PR #1341, opened 2020-06-26
Don't double-prefix the install directory

eddyashton:
Before:
```
$ cmake .. -UCMAKE_INSTALL_PREFIX
-- CCF version=ccf-0.11.1
-- CCF release version=0.11.1
-- CMAKE_INSTALL_PREFIX is '/opt/ccf/ccf-ccf-0.11.1'
```

After:
```
$ cmake .. -UCMAKE_INSTALL_PREFIX
-- CCF version=ccf-0.11.1-3-g99e8c55a
-- CCF release version=0.11.1
-- CMAKE_INSTALL_PREFIX is '/opt/ccf/ccf-0.11.1'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ message(STATUS "CCF release version=${CCF_RELEASE_VERSION}")
 #
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX
-      "/opt/ccf/ccf-${CCF_RELEASE_VERSION}"
+      "/opt/ccf/${CCF_VERSION}"
       CACHE PATH "Default install prefix" FORCE
   )
 endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ message(STATUS "CCF release version=${CCF_RELEASE_VERSION}")
 #
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX
-      "/opt/ccf/ccf-${CCF_VERSION}"
+      "/opt/ccf/ccf-${CCF_RELEASE_VERSION}"
```
jumaffre, line 27: In the case of a "between releases" commit, I think it would still be nice to have a suffix that doesn't match the previous release. E.g. if we are currently 5 commits after 0.11, it would be nice if the install path was something like `ccf-0.11-5-g<sha>` rather than `ccf-0.11`, which would override the latest release.

eddyashton: Sure

PR #1343, opened 2020-06-26
Additional store snapshot tests and conflict handling

jumaffre:
Part of #1299 

- Move the snapshot tests from `kv_test.cpp` to a new `kv_snapshot.cpp` for clarity.
- Add more tests to check the content of snapshots once deserialised into a new store.
- Support for transaction commit conflict when a snapshot is applied between the time a `Tx` calls `get_view()` and `commit()`. This is handled using the `rollback_counter` which gets incremented when a snapshot is applied.


src/kv/test/kv_snapshot.cpp
```diff
@@ -0,0 +1,169 @@
...
+  {
+    kv::Tx tx1, tx2;
+    auto view_1 = tx1.get_view(string_map);
+    auto view_2 = tx2.get_view(num_map);
+
+    view_1->put("foo", "bar");
+    view_2->put(42, 123);
+
+    REQUIRE(tx1.commit() == kv::CommitSuccess::OK); // Committed at 1
+    REQUIRE(tx2.commit() == kv::CommitSuccess::OK); // Committed at 2
```
eddyashton, line 35: This looks unnecessarily like conflicting transactions. because we interleave the operations. Since these are really applied in series, and we're not trying to test interleaved operations, can we write these in sequence? ie

```
kv::Tx tx1;
auto view_1 = ...;
view_1->put(...);
REQUIRE(tx1...);

kv::Tx tx2;
auto view2 = ...;
...
```

src/kv/test/kv_snapshot.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    kv::Tx tx3;
+    auto view_3 = tx1.get_view(string_map);
+    view_3->put("key", "not committed");
+    // Do not commit tx3
+  }
+
+  auto s_1 = store.snapshot(1);
```
eddyashton, line 43: Can we use a variable rather than 1? Something like
```
kv::Version first_commit;
...
  tx1.commit();
  first_commit = tx1.current_version()
...
store.snapshot(first_commit);
```

src/kv/test/kv_snapshot.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  INFO("Verify content of snapshot");
+  {
+    auto& vec_s = s_1->get_snapshots();
```
eddyashton, line 47: Hadn't seen this before. `Snapshot::get_snapshots` is odd, can we rename the type to `AbstractSnapshots` if it is a collection? Or maybe the function is `get_map_snapshots` to distinguish _Store_ Snapshots from _Map_ Snapshots?

eddyashton: And it looks like the return type could be const

src/kv/test/kv_test.cpp
```diff
@@ -1080,68 +1080,4 @@ TEST_CASE("Conflict resolution")
...
-    auto view = tx.get_view(map);
-
-    // Numroduce read-dependency
-    view->get("foo");
-    view->put("foo", s);
-
-    view->put(s, s);
-  };
-
-  INFO("Simulate parallel execution by interleaving tx steps");
```
eddyashton, line 1103: This test case looks like it is still useful, was it removed by accident?

src/kv/test/kv_snapshot.cpp
```diff
@@ -0,0 +1,169 @@
...
+  INFO("Apply snapshot while committing a transaction");
+  {
+    kv::Store new_store;
+    new_store.clone_schema(store);
+
+    auto new_string_map = new_store.get<MapTypes::StringString>("string_map");
+    kv::Tx tx1, tx2;
+    auto view = tx1.get_view(*new_string_map);
+    view->put("baz", "baz");
+    REQUIRE(tx1.commit() == kv::CommitSuccess::OK); // Committed at 1
```
eddyashton, line 153: This seems like an extra complication we're not fully testing - we're deserialising a snapshot onto a kv which has a _different prefix_, it has had some totally different transactions executed on it before. It looks like this works (we drop the roll when applying a snapshot?), but we should test exactly this in isolation (ie - _this_ tx1 is totally lost after deserialising the snapshot)

achamayou:
Being able to handle tx on a store which receives a snapshot sounds nice, but I doubt it'll be useful in practice. The only use case I can think of is a node that's fallen so far behind that it requests a snapshot rather than catch up through regular consensus updates. The usefulness of continuing to serve very stale reads while the snapshot is applied is unclear.

PR #1349, opened 2020-06-26
Snapshot testing

jumaffre:
Duplicate from https://github.com/microsoft/CCF/pull/1343 which got accidentally merged (and [reverted](https://github.com/microsoft/CCF/commit/3ed1a655ac0fa15a93a5c889727e5b1974deace0)).


src/kv/test/kv_snapshot.cpp
```diff
@@ -0,0 +1,172 @@
...
+  kv::Version first_snapshot_version = kv::NoVersion;
+  kv::Version second_snapshot_version = kv::NoVersion;
+
+  INFO("Apply transactions to original store");
+  {
+    kv::Tx tx1;
+    auto view_1 = tx1.get_view(string_map);
+    view_1->put("foo", "bar");
+    REQUIRE(tx1.commit() == kv::CommitSuccess::OK);
+    first_snapshot_version = tx1.commit_version();
```
eddyashton, line 34: This code is fine, but just adding a note because it confused me for a while: `Tx::commit_version()` is completely different from `Store::commit_version()`. The latter returns `compacted`, due to our usual confusion around local/global commit. We should do some kind of renaming, either make the latter `compact_version`, or change all of the 'local' `commit`s to `apply` or similar.

jumaffre: Absolutely. I have raised https://github.com/microsoft/CCF/issues/1355 to do this work separately.

PR #1351, opened 2020-06-29
Code versions RPC

### LINKED ISSUE #239: Add an RPC to view supported code hashes
There should be a way to verify whether code can be executed according to its signature.
An RPC that retrieves the data from the codeid table should be added.

achamayou:
fix #239


src/node/rpc/common_endpoint_registry.h
```diff
@@ -197,6 +199,24 @@ namespace ccf
...
+          [&out](const ccf::CodeDigest& cd, const ccf::CodeStatus& cs) {
+            auto digest = fmt::format("{:02x}", fmt::join(cd, ""));
+            out.versions.push_back({digest, cs});
+            return true;
+          });
+
+        return make_success(out);
+      };
+      make_read_only_endpoint(
+        "code", HTTP_GET, json_read_only_adapter(get_code))
```
jumaffre, line 216: "nodes/code"?

achamayou: But then that means "node/nodes/code"?

jumaffre: Does it? I was comparing it with the `get_nodes_by_rpc_address` endpoint below. 

Also, from the e2e test, it seems that we are calling this endpoint from the `user_client`, thus calling `app/code` which is weird? There's probably something I'm missing here!

achamayou: It's one of those common handlers we have, which is exposed on every frontend (ie. node/app/gov), like commit, tx and so on. Arguably, they ought to be top-level, but our current frontend-based organisation makes this awkward. I think there are two discussion here:

1. What to do with common/global endpoints?
2. Where does code/getCode belong?

Since code essentially exposes a public fact about the network (the digest of the code it runs, stored in a public table), I think it ought to be a common/global handler, like commit, tx, etc. That only answers 2.

As far as 1. is concerned, we've had some discussions about this already, but none of them were conclusive. I suggest the slightly unsatisfying meta-answer that we should solve all of those at the same time, and that this instance isn't special.

eddyashton: For reference, #1353 now tracks abolishing the frontends entirely. Under the proposed scheme, this RPC would have a single unauthed endpoint at `/codes` or `/ccf/codes`.

PR #1356, opened 2020-06-29
Report errors to clients for invalid HTTP signatures

### LINKED ISSUE #1348: Signed command parsing failures do not return a HTTP error
There are 2 related bugs here:

- Our `scurl.sh` wrapper script doesn't properly handle files which end in a newline. We use `echo -n` which appears to _strip any existing newline_, so we produce a different signature than the data we eventually send.
- Signature validation errors throw on a path that doesn't get converted to a HTTP error. The client sees `curl: (52) Empty reply from server`, while the node log has `Request body does not match digest header` and `Error parsing request`. We should wrap more of the request parsing code to send a HTTP response body on exceptions.

eddyashton:
Fix #1348.

HTTP signature validation failures throw an exception that is caught in `http_endpoint.h:307`. We call `send_raw` to try and tell the user about the error, and then close the connection. Previously, `send_raw` was sent asynchronously (queue the task onto the session thread's task queue) while `close` was executed inline, so we actually `close`d first and the client saw an empty response/unexpectedly closed connection. This fixes that by making more operations on a `TLSEndpoint` async, so they are executed in the order they are called.

I think there should be a better way to enforce this in the single-threaded case, perhaps creating a fake threadID for the async work so we can distinguish the work the main thread is _always_ doing from the work it does in the single-threaded case, when it is also pretending to be a worker thread? Haven't found a neat way to do this though, will drop it unless someone has a clean idea.

This also fixes another issue described in #1348, by passing files directly to `openssl dgst` rather than trying to `cat` them and potentially stripping trailing newlines.


eddyashton:
This PR originally made `TLSEndpoint::flush()` async in the same way it does with `TLSEndpoint::close()`. This seems to cause an infinite loop in the perf tests, though I've not yet worked out why. I've ignored this for now, and made only `close()` async. For safe threading support, we should take another look at the functions which delegate to an async task and make sure we use them consistently across an object.

PR #1359, opened 2020-06-30
Gather CI checks in a single entry point

### LINKED ISSUE #1350: Single Format/Lint entry point
There ought to be a single run-ci-checks (or perhaps scripts/run-ci-checks?) that takes care of all of https://github.com/microsoft/CCF/blob/master/.azure-pipelines-templates/checks.yml, to make life easier for contributors trying to submit PRs.

Spotted by @letmaik

achamayou:
Fix #1350


scripts/ci-checks.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# Shell scripts
```
eddyashton, line 7: Can we replace these comments with echos? Would be helpful to explicitly list the stages we're running, to help distinguish what has failed/needs fixing.

scripts/notice-check.py
```diff
@@ -31,7 +31,7 @@
 
 
 def has_notice(path, prefixes):
-    print(f"Reading {path}")
+    # print(f"Reading {path}")
```
ashamis, line 34: delete this line?

PR #1360, opened 2020-06-30
Run scan build in Daily

achamayou:
While I was looking at the checks script I noticed I never hooked up scan build to the Daily. Here goes.


.azure-pipelines-templates/common.yml
```diff
@@ -31,6 +31,9 @@ jobs:
     - ${{ if eq(parameters.suffix, 'Coverage') }}:
       - template: coverage.yml
 
+    - ${{ if eq(parameters.suffix, 'Coverage') }}:
```
jumaffre, line 34: Might be worth renaming this flag as it no longer only guards the coverage job?

achamayou: Do you have a good name in mind?

PR #1361, opened 2020-06-30
Respect relative path in LocalRemote.get

eddyashton:
For some time our perf runs have been failing to retrieve `perf_summary.csv`. This fixes that.

### DIFF
--- a/tests/infra/remote_client.py
+++ b/tests/infra/remote_client.py
@@ -72,24 +72,18 @@ class CCFRemoteClient(object):
         return self.remote.debug_node_cmd()
 
     def stop(self):
-        try:
-            self.remote.stop()
-            remote_files = self.remote.list_files()
-            remote_csvs = [f for f in remote_files if f.endswith(".csv")]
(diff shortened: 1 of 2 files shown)


tests/infra/remote_client.py
```diff
@@ -72,24 +72,18 @@ def debug_node_cmd(self):
...
+        remote_csvs = [f for f in remote_files if f.endswith(".csv")]
+
+        for csv in remote_csvs:
+            remote_file_dst = f"{self.name}_{csv}"
+            self.remote.get(csv, self.common_dir, 1, remote_file_dst)
+            if csv == "perf_summary.csv":
+                with open("perf_summary.csv", "a") as l:
+                    with open(os.path.join(self.common_dir, remote_file_dst), "r") as r:
+                        for line in r.readlines():
+                            l.write(line)
```
achamayou, line 86:
```suggestion
                        l.write(r.read())
```

Strange to do it line by line? [I know this isn't the change, but while we're here]

Issue #1362, opened 2020-06-30
Update evercrypt code


ashamis:
Update the evercrypt do when https://github.com/FStarLang/FStar/pull/2073 has been committed.
This fixes the merkle tree serialization + deserialization + append memory issue

wintersteiger:
https://github.com/FStarLang/FStar/pull/2073 has been merged and the new snapshot should be available shortly. project-everest/hacl-star#328 will be extended with more/better proofs about serialization, but we don't have to wait for that.

achamayou:
Done in #1576, thank you @wintersteiger!

PR #1363, opened 2020-06-30
HotFix Merkle tree deserialization bug

ashamis:
This is a hotfix that will be properly resolved by #1362 

This fixes the issue where we see:
1. Serialize the Merkle tree
2. deserialize the Merkle tree
3. append to the Merkle tree
==> At this point we would see a memory violation.

Thanks to @wintersteiger for the hotfix


achamayou:
Thank you @wintersteiger!

PR #1365, opened 2020-07-01
Update Contribution section

achamayou:
That change seems to have been lost from the ci-checks PR somehow.


jumaffre:
There's also https://github.com/microsoft/CCF/blob/master/.github/CONTRIBUTING.md that needs updating

achamayou:
@jumaffre hah, that's been updated, I thought I'd updated something! We shouldn't have two of those, we should have a link, or an include, but not a duplicate.

achamayou:
@jumaffre couldn't include, but linked.

Issue #1353, opened 2020-06-29
Explore removal of frontends


eddyashton:
We currently have an unnecessarily strict connection between the prefix of a URL path and the auth needed to access that path. A CCF service is divided into 3 frontends maintaining this connection, with member-authed resources on `/gov`, user-authed (app-extensible) resources on `/app`, and unauthorised 'public' resources on `/node`. We're adding more and more 'common' endpoints which we currently make available on _every_ frontend, but which should really be a single, unauthed instance.

I believe it should be possible to remove this concept of frontends entirely. We will have a _single_ frontend, with a large dispatch map of _all_ registered endpoints. The auth-scheme will be determined after endpoint-lookup, and initially that will be one of the 3 current schemes ("don't care", "is cert in members table", "is cert in users table").

A related thing we may change sooner than this is to use full URIs in more places, in both client and app code, to remove assumptions about prefixes.

An additional feature of this approach is that apps could trivially extend the member frontend - we don't have a clear use case for this currently but its likely to be requested at some point.

A code sketch:
`enclave.h`:
```
add_common_endpoints(...); // Calls make_anonymous_endpoint() for the things currently in node_frontend.h,
                           // and many of the things in common_endpoint_registry.h

add_member_endpoints(...); // Calls make_member_endpoint() for the things currently in member_frontend.h

add_app_endpoints(...); // Calls make_endpoint() (aka make_app_endpoint()?), and maybe make_anonymous_endpoint()
                        // or make_member_endpoint() where appropriate
```

`endpoint_registry.h`
```
Endpoint make_member_endpoint(path, verb, fn)
{
  return make_endpoint_impl(path, verb, [fn](args)
    {
      if (member_certs.contains(args.caller_id))
      {
        fn(args);
      }
      args.set_response(AUTH_ERROR);
    }
  );
}
```

Of course we don't need to do this with wrapping lambdas, it may make more sense to store an enum with each installed `Endpoint` and take the explicit auth path in `frontend.h` (or whatever replaces it).

achamayou:
@eddyashton on a somewhat related note, I think we need to make an effort to clean up the samples to be more consistent and intuitive. For example:

`record()` takes a `tx` and `params`, but `get()` takes `args` and `params`. They should both take the context instead.

[...]

PR #1367, opened 2020-07-01
Client prefix drop

achamayou:
This removes our hardcoded assumption in the tests that URL path prefix must coincide with access permissions. They don't fundamentally, this forces us to have a series of common handlers that we install on every prefix (commit, tx etc) instead of being able to scope them correctly (eg. service/ or some other appropriate name).

Some handlers are public, some may be actors only (ie. users and members), others may be just one or the other, or more restricted still. This is best decided per-handler, the notion of frontend grouping prefix and hardcoded policy isn't helpful.

This is an enabling step towards #1353


tests/e2e_logging.py
```diff
@@ -464,13 +464,13 @@ def get_all_tx_ids():
 
 
 @reqs.description("Build a list of Tx IDs, check they transition states as expected")
-@reqs.supports_methods("log/private", "tx")
+@reqs.supports_methods("log/private", "node/tx")
```
eddyashton, line 467:
```suggestion
@reqs.supports_methods("app/log/private", "node/tx")
```

achamayou: I think should be node/tx -> tx instead, but I'll double check.

tests/infra/node.py
```diff
@@ -244,43 +244,19 @@ def wait_for_node_to_join(self, timeout=3):
     def get_ledger(self):
         return self.remote.get_ledger()
 
-    def user_client(self, user_id=0, **kwargs):
+    def client(self, identity=None, **kwargs):
+        akwargs = {
+            "cert": os.path.join(self.common_dir, f"{identity}_cert.pem") if identity else None,
+            "key": os.path.join(self.common_dir, f"{identity}_privk.pem") if identity else None,
+            "ca": os.path.join(self.common_dir, "networkcert.pem"),
+            "description": f"node {self.node_id} as {identity}",
```
eddyashton, line 252: Minor, just a suggestion, open to a better term here (I think we've used "anonymous" elsewhere?)

```suggestion
            "description": f"node {self.node_id} as {identity or "anyone"}",
```

tests/perf_logging_scenario_100txs.json
```diff
@@ -1,700 +1,700 @@
 {
   "transactions": [
     {
-      "method": "log/private",
+      "method": "aapp/log/private",
```
eddyashton, line 4:
```suggestion
      "method": "app/log/private",
```

eddyashton:
If we're putting the full URI path, I'd prefer if they were `/`-prefixed (ie - `"/app/logging/public"` rather than `"app/logging/public"`. Although we'd be repeating the leading slash everywhere, I think its useful to make it clear this _is_ a path. Also this string will then match in more places (in particular, it'll be the exact key in the `paths` object of the OpenAPI doc). We don't do currently do this on the C++ side for slightly unfortunate reasons, since we're regularly constructing and destructing the actor split as `/{}/{}`, but in removing the frontend split I intend to change this so that the C++ registrations are explicitly `"/app/logging/public"`

tests/infra/node.py
achamayou: That'll show "None" right now. I think I'd mildly prefer "unauthenticated" rather than "anyone" (there _is_ a caller, so it feels more like "someone" but we don't know who?).

tests/perf_logging_scenario_100txs.json
achamayou: Thanks!

achamayou:
@eddyashton I don't feel all that strongly about that, I think there's still the open question of the versioning, so these could be version-relative. Happy to change it though.

achamayou:
@eddyashton done now.

tests/e2e_logging.py
achamayou: Yes, that's only looking methods up under the app frontend. It needs to either stay unprefixed or be extended to support all prefixes.

Issue #1352, opened 2020-06-29
Improve proposal and voting samples for new users


eddyashton:
The membership proposals are generally simple, and the most complicated step for new users isn't the scripting logic but the formatting of the proposal object in JSON. We should add Python scripts to generate these proposals. They essentially exist in `consortium.py` already, but we should extract them to dump proposals to file so they can be used more generally (while still being tested by all of our infra.

As a minor extension these generators should also produce sample accepting votes, which include validation that the proposal's params match the proposed value (ie - when generating a proposal that says "add Bob as a new user", create a ballot that returns true iff the proposal has a single `new_user` call with Bob's key).

Additionally, the proposal endpoint could do some schema validation of proposal params, so that (most) type errors are caught at proposal time rather than after a successful vote.

- [x] Factor generation of proposal and vote objects out of `consortium.py`
- [x] Add a CLI frontend to this proposal generation
- [x] Build more interesting sample votes which validate, don't simply `return true`
- [x] Use these sample votes in the infra
- [x] In the infra, dump proposals and votes to file to test more of the CLI flow and to leave a a permanent record of proposals
- [x] Point at the generator CLI in the docs
- [x] Accept strings rather than byte-arrays for Pem values in proposals (related to #278)
- [x] Put proposal IDs into path rather than params, so we generate a complete vote sample which is independent of the proposal ID (ie - `POST /gov/vote/5 {"ballot": ...}` rather than the current `POST /gov/vote {"id": 5, "ballot": ...}`) (sort of #583, but perhaps we should track this feature independently)

eddyashton:
I believe we should document this for 0.12. Schema validation can wait until later.

achamayou:
Agreed, particularly since we intend to migrate the governance to Javascript.

PR #1370, opened 2020-07-01
Add generator for membership proposals

eddyashton:
This completes some of the tasks in #1352.

This pulls the generation of lua proposal scripts and votes out of `consortium.py` and scattered elsewhere and places everything in `proposal_generator.py`. This generates both proposals and votes in a standard form, with the votes providing a sample of extensive verification of the targeted proposal.

This script has a CLI, so it can be used to generate proposals from a terminal. This CLI isn't tested directly, but the hope is that as much of the code as possible is reused by `consortium.py` and stressed in our end to end tests so it stays correct.

The key tasks for future PRs are (plus some related points listed in the issue):
- Dump proposals and votes to file and use them from file in the infra. This the dual benefits of testing more of the CLI flow, and producing sample request bodies in the workspace every time an end-to-end test is run.
- Use the sample votes. Currently we're _generating_ detailed validating votes but we don't actually use them. There's an awkwardness from the fact that the proposal_id must be included in the `vote` object. So we may need to generate a _partial_ vote object (ie - just the script, and insert the `proposal_id` once it is known), which doesn't play nicely with the "dump a sample vote to file" above. Ideally we get to the final bullet point of #1352 - put the `proposal_id` into the URI rather than the body, so I may explore this next.
- Pointing at this CLI tool in the docs, and perhaps including generated proposals so they automatically stay up to date.


tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,336 @@
...
+
+    LOG.trace(f"Made {proposed_call} proposal:\n{json.dumps(proposal, indent=2)}")
+    LOG.trace(f"Accompanying vote:\n{json.dumps(vote, indent=2)}")
+
+    return proposal, vote
+
+
+class Proposals:
+    @staticmethod
+    def new_member_proposal(member_cert_path, member_keyshare_encryptor_path):
```
eddyashton, line 133: The other proposals delegate most of their work to `build_proposal`. This could also do that, but I've left it with an explicit manual implementation for readability. I guess this can go, since we really only care about the proposal and vote objects themselves and not how they're produced.

samples/apps/txregulator/clients/loader.py
```diff
@@ -46,23 +47,15 @@ def run(args):
...
-                        user_data = {{
-                            privileges = {{
-                                REGISTER_REGULATORS = true,
-                                REGISTER_BANKS = true,
-                            }}
-                        }}
-                    }}
-                )
-                """,
+            proposal_body, _ = Proposals.set_user_data_proposal(
```
eddyashton, line 50: This appears to be untested by the CI so has likely rotted. I've tried to update it appropriately here, but eventually we may need either an automated test or to remove this.

achamayou: I think it's probably time to let it go and replace it with the price comparison sample suggested by Brent, written in Javascript.

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,334 @@
...
+        "id": PROPOSAL_ID_PLACEHOLDER,
+    }
+
+    LOG.trace(f"Made {proposed_call} proposal:\n{json.dumps(proposal, indent=2)}")
+    LOG.trace(f"Accompanying vote:\n{json.dumps(vote, indent=2)}")
+
+    return proposal, vote
+
+
+class Proposals:
```
achamayou, line 131: Why a class with static methods? infra.proposals with functions seems clearer?

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,334 @@
...
+from loguru import logger as LOG
+
+
+def dump_proposal(output_path, proposal, dump_args):
+    LOG.info(f"Writing proposal to {output_path}")
+    with open(output_path, "w") as f:
+        json.dump(proposal, f, **dump_args)
+
+
+def dump_vote(output_path, vote, dump_args):
```
achamayou, line 18: dump_proposal() and dump_vote() are identical, so... dump()?

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,334 @@
...
+  end
+end
+"""
+
+
+PROPOSAL_ID_PLACEHOLDER = "<replace with desired proposal_id>"
+
+
+def add_arg_construction(lines, arg, arg_name="args"):
+    if isinstance(arg, list):
```
achamayou, line 60: In general it's a bit nicer to take collections.abc.Sequence, Mapping (with a special case for basestring, which is a Sequence but needs different handling). It's a more precise requirement, and it means passing generators, custom iterators etc would work.

tests/infra/proposal_generator.py
eddyashton: We generate the CLI commands/arguments using `inspect`, so I thought it was cleaner to have a single class holding the "functions to expose" to introspect rather than the entire module? They could instead get a decorator, or maybe there's some other Pythonic way of doing this? Or maybe I shouldn't be so lazy and we should populate the CLI explicitly...

tests/infra/member.py
```diff
@@ -63,25 +63,17 @@ def set_active(self):
         # Use this with caution (i.e. only when the network is opening)
         self.status = MemberStatus.ACTIVE
 
-    def propose(self, remote_node, script=None, params=None, vote_for=True):
+    def propose(self, remote_node, proposal):
```
jumaffre, line 66: It seems that we used to be able to pass `vote_for=False` here and now we always pass a vote with the proposal? I'm not sure we even tested this before but perhaps it is something that would be nice to keep?

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,334 @@
...
+        json.dump(proposal, f, **dump_args)
+
+
+def dump_vote(output_path, vote, dump_args):
+    LOG.info(f"Writing vote to {output_path}")
+    with open(output_path, "w") as f:
+        json.dump(vote, f, **dump_args)
+
+
+def file_to_byte_array(path):
```
jumaffre, line 24: If we're working on usability/friendliness of the governance API, is it time that we move those to base64 strings?

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,334 @@
...
+
+    LOG.trace(f"Made {proposed_call} proposal:\n{json.dumps(proposal, indent=2)}")
+    LOG.trace(f"Accompanying vote:\n{json.dumps(vote, indent=2)}")
+
+    return proposal, vote
+
+
+class Proposals:
+    @staticmethod
+    def new_member_proposal(member_cert_path, member_keyshare_encryptor_path):
```
jumaffre, line 133: Minor: The bad naming of this member key in our infra is definitely on me. But I think we should rename `member_keyshare_encryptor` to `member_enc_pubk` wherever possible?

I'll take care of renaming the very elusive `"keyshare"` in the add member proposal in a subsequent PR.

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,334 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+
+import argparse
+import inspect
+import json
+import os
+
+from loguru import logger as LOG
```
jumaffre, line 9: Do we need a very limited `virtualenv` setup so that members can run this without pulling all the CCF infra?

jumaffre: Also, I haven't run the CLI locally but I believe that the default for `loguru` is pretty verbose (i.e. prints the file name and line number on which the log function is called) which might be a little overwhelming for members running this CLI.
We could perhaps do something like in `start_network.py` instead so that the output is more concise?

tests/infra/proposal_generator.py
eddyashton: Absolutely - looking at this in a separate PR

tests/infra/proposal_generator.py
achamayou: A registration decorator would be nice, works irrespective of the scope. A module is more accurate than a class, because there's no state, and it's just as easy to use inspect.getmembers() on (or .__dict__.items() really...).

```
def proposal(fun):
  cli.register(fun, description=fun.__doc__)

@proposal
def thingie(...):
  """
  Some doc about the proposal
  """
```

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,336 @@
...
+            add_arg_construction(lines, v, arg_name=f"{arg_name}.{k}")
+    elif isinstance(arg, str):
+        lines.append(f'{arg_name} = "{arg}"')
+    else:
+        lines.append(f"{arg_name} = {arg}")
+
+
+def add_arg_checks(lines, arg, arg_name="args"):
+    lines.append(f"if {arg_name} == nil then return false")
+    if isinstance(arg, list):
```
achamayou, line 74: Same as before, using concrete types is a little unfortunate.

tests/infra/proposal_generator.py
```diff
@@ -0,0 +1,336 @@
...
+    )
+
+    # Auto-generate CLI args based on the inspected generator signatures
+    proposal_generators = inspect.getmembers(Proposals, predicate=inspect.isfunction)
+    subparsers = parser.add_subparsers(
+        title="Possible proposals", dest="proposal_type", required=True
+    )
+
+    for func_name, func in proposal_generators:
+        suffix = "_proposal"
```
achamayou, line 299: We can avoid that kludge if we explicitly decorate them (and shorten the names!).

tests/infra/proposal_generator.py
eddyashton: Replaced this class with a decorator, works great

tests/infra/member.py
eddyashton: I had a big waffley answer here but it didn't make sense. In short:

It wasn't tested before and I'm not sure the remaining code around it is all correct. I want to take another look in a future PR, with a test for "initial-no-votes", but I'd rather plumb that through separately.

tests/infra/proposal_generator.py
eddyashton: `loguru`: Great idea - I've added a simpler logging format in `__main__`.
`virtualenv`: Probably. I think this falls under future Python packaging - this is one of the bits of infra we want to expose outside, and that probably involves a Python package with narrower set of requirements.

tests/infra/member.py
jumaffre: Sounds good. This behaviour seems to be tested by some unit tests in `member_voting_test.cpp` so I'm happy to remove this here.

tests/infra/proposal_generator.py
eddyashton: For posterity, that's #1374.

PR #1371, opened 2020-07-02
Improve handling of node-to-node channels and host connections

### LINKED ISSUE #1342: Couple host node connections to enclave channel manager
The node-to-node channels created in the enclave and the node-to-node TCP connections on the host are currently handled independently:
- A new TCP connection is created from the local hook on the `ccf.nodes` table when a new node is added as `TRUSTED` while the TCP connection is closed on the global hook on the same table when a node is `RETIRED`.
- A new node-to-node channel is created when the first message is sent on that channel (most likely a Raft Append Entries message). n2n channels are never closed. 

Instead, the two modules should be coupled so that the channel manager is in charge of creating/deleting enclave n2n channels and notify the host to create/delete channel. In particular, Raft should also notify the channel manager to close a n2n channel (and the associated TCP connection) whenever a node is removed from its list of nodes (i.e. rollback of non-committed tx or node deletion gets globally committed).

This is related to https://github.com/microsoft/CCF/issues/486#issuecomment-649581017

jumaffre:
Resolves #1342 

**Before**

- Node outgoing connections (host) are created from the `ccf.nodes` local hook, which means that _all_ nodes in the network have sockets to open to _all_ nodes (minus self) at any given time. 
- Node outgoing connections (host) are closed from the `ccf.nodes` global hook, when a node is retired. 
- Node outgoing connections (host) and node-to-node channels (enclave) are not in sync.

**After**

- At any given time (except during elections), there are only `n-1` outgoing channels between the primary and the backups.
- Node outgoing connections (host) are managed by the node-to-node channels, i.e. when an outgoing channel is created, a new host outgoing connection is created and when an outgoing channel is closed, the host outgoing connection is closed.
- The process is creating channels is delegated to the Raft reconfiguration "module" (which may be pulled out of Raft at some point?)

Notes:
- The whole process of creating channels is very much asymmetrical. On the leader, (outgoing) channels are created when a node is `TRUSTED` from the store. On a backup, the (incoming) channel is created way before the peer node is known from the store. 
- The configuration code now needs to know about the peer's host/port so that it can create the channel appropriately when the node is added. 
- On election, when a node becomes candidate, it creates an outgoing connection to all nodes it knows in its configuration. If the election is lost (i.e. the node becomes backup), all of these outgoing channels are closed (i.e. there should only be one incoming connection from the new leader).
- Also added a new test suite to test for elections when adding/removing nodes (+ associated test requirement).


src/consensus/pbft/pbft.h
```diff
@@ -580,29 +583,31 @@ namespace pbft
     }
 
     void add_configuration(
-      SeqNo seqno,
-      const std::unordered_set<kv::NodeId>& config,
-      const NodeConf& node_conf) override
+      SeqNo seqno, const Configuration::Nodes& config) override
     {
-      if (node_conf.node_id == local_id)
+      auto new_node_id = config.begin()->first;
```
achamayou, line 588: config is never empty?

src/kv/kv_types.h
```diff
@@ -188,6 +188,7 @@ namespace kv
     // primary changes
     using View = uint64_t;
 
+    // TODO: Delete this whenever possible
```
achamayou, line 191: `^v^`

src/kv/store.h
```diff
@@ -702,6 +702,8 @@ namespace kv
 
           LOG_DEBUG_FMT(
             "Batching {} ({})", last_replicated + offset, data_shared->size());
+
+          // TODO: Encryptor
```
achamayou, line 706: That's where we want to notify the encryptor of new secrets, right?

jumaffre: Absolutely. I clearly wrote it there as a note to my future self and forgot to remove it.

src/consensus/pbft/pbft.h
jumaffre: PBFT reconfiguration is currently not fully handled and the hook that calls into this only passes one node. I've added a check there and we can revisit when PBFT reconfiguration is addressed.

src/consensus/raft/raft.h
```diff
@@ -1184,58 +1206,80 @@ namespace raft
...
           // A new node is sent only future entries initially. If it does not
           // have prior data, it will communicate that back to the leader.
           auto index = last_idx + 1;
-          nodes[node_id] = {0, index};
+          nodes.try_emplace(node_info.first, 0, index, node_info.second);
 
           if (state == Leader)
-            send_append_entries(node_id, index);
+          {
+            channels->create_channel(
```
jumaffre, line 1269: When a new `TRUSTED` node is added, only the leader should create a channel to the backups.

src/host/node_connections.h
```diff
@@ -305,11 +309,10 @@ namespace asynchost
       for (const auto node : local_queue)
       {
         LOG_DEBUG_FMT("reconnecting node {}", node);
-        auto s = find(node);
-
-        if (s)
+        auto s = outgoing.find(node);
```
jumaffre, line 312: Only looking at the `outgoing` channels here because `find()` displays an undesirable log message if the channel is not found (and it probably won't be there as we are reconnecting).

src/node/node_state.h
```diff
@@ -1320,26 +1320,6 @@ namespace ccf
 
     void setup_basic_hooks()
     {
-      // When a transaction that changes the configuration commits globally,
-      // inform the host of any nodes that no longer need to be tracked.
-      network.nodes.set_global_hook(
```
jumaffre, line 1325: This global hook is now gone. The channel (and associated host outgoing connection) will get closed by Raft directly on global commit of the configuration in which the node is deleted.

PR #1380, opened 2020-07-03
Start packaging Python infra

jumaffre:
Part of #1382 

Packages the `clients.py`, `ledger.py`, `checker.py` and `proposal_generator.py` files in a Python package to be uploaded to PyPi for easy consumption by users/members/operators. There is now a new `python` directory under the root of this repo and which contains the new `ccf` package (creating `ccf/` directly under the root of this repo was awkward): 
```bash
$ cd CCF/
$ ls python
MANIFEST.in  README.md  ccf  setup.py
$ ls python/ccf
__init__.py  checker.py  clients.py  ledger.py  proposal_generator.py  tx_status.py
```

- The `MANIFEST.in` file is used to include the `scurl.sh`, `keygenerator.sh` and `submit_recovery_share.sh` scripts in the install (we could rewrite them in Python, but it does not seem required for now).
- Once the package is installed (e.g. under a `venv`), all above utilities + `proposal_generator.sh` are available in the `PATH`. 
- The _files_ in the package also gets installed in the CCF release under `bin/ccf`. Once the package is released on PyPi, we will just point to that in the `requirements.txt` instead. 
- Automated released to `PyPi` will come as soon as the code is cleaned up and documented.


tests/setup.py
```diff
@@ -0,0 +1,34 @@
...
+# Licensed under the Apache 2.0 License.
+
+from setuptools import setup
+
+with open("README.md", "r") as fh:
+    long_description = fh.read()
+
+setup(
+    name="ccftools",
+    version="0.11.9",
```
jumaffre, line 11: The version of the PyPi package is an interesting one. Ideally, we want it in sync with the CCF release (and we can do so by reading from the metadata of the repo/directory, just like for cmake).

However, it would mean that we cannot release an update of this package _without_ having to release a new version of CCF.

tests/README.md
```diff
@@ -0,0 +1,3 @@
+# ccftools
```
eddyashton, line 1: Should this be `tests/ccftools/README.md`?

tests/setup.py
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+
+from setuptools import setup
+
+with open("README.md", "r") as fh:
+    long_description = fh.read()
+
+setup(
+    name="ccftools",
```
jumaffre, line 10: Name is very much tentative for now. I don't think that `ccf` is ideal (users of the framework may think they can run a network, etc. from Python); `ccfclients` might be too restrictive (we will probably end up release more utilities in there).

tests/README.md
jumaffre: From the different examples I've seen online, the `README.md` is often at the top directory. I agree with you though and I've moved it in there 👍

tests/setup.py
```diff
@@ -0,0 +1,39 @@
...
+setup(
+    name=PACKAGE_NAME,
+    version="0.11.10",
+    description="Set of tools and utilities for the Confidential Consortium Framework (CCF)",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/microsoft/CCF",
+    license="Apache License 2.0",
+    author="CCF Team",
+    author_email="ccfeng@microsoft.com",
```
achamayou, line 23: Is that set up?

tests/setup.py
achamayou: I think they should stay on the same release cycle at first, doing a release where just the tools changed doesn't sound too bad.

tests/README.md
achamayou: I think we probably ought to move ccftools out of tests?

tests/setup.py
achamayou: I think `ccf` would be alright, it's understood that the Python package is going to be supporting tooling/clients, see k8s, zookeeper, etcd.

python/setup.py
```diff
@@ -4,16 +4,14 @@
...
-package_dir = path.join(path.abspath(path.dirname(__file__)), PACKAGE_NAME)
-print(package_dir)
-with open(path.join(package_dir, "README.md"), encoding="utf-8") as f:
+with open(path.join(path.abspath(path.dirname(__file__)), "README.md"), encoding="utf-8") as f:
     long_description = f.read()
 
 setup(
     name=PACKAGE_NAME,
-    version="0.11.10",
+    version="0.11.7",
```
achamayou, line 16: I remember trying and failing to get rid of this in cimetrics. I'll try to work it out this time, because I don't expect we'll be successful at not forgetting to update the version in the Python package.

python/setup.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+
+from setuptools import setup
+from os import path
+
+PACKAGE_NAME = "ccf"
+
+path_here = path.abspath(path.dirname(__file__))
+utilities_path = "tests"
```
eddyashton, line 10: Minor: Why `PACKAGE_NAME` vs `utilities_path`?

python/setup.py
```diff
@@ -0,0 +1,49 @@
...
+) as f:
+    long_description = f.read()
+
+setup(
+    name=PACKAGE_NAME,
+    version="0.11.7",
+    description="Set of tools and utilities for the Confidential Consortium Framework (CCF)",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/microsoft/CCF",
```
eddyashton, line 23: Nit: Should this point to the `CCF/python` folder rather than the repo homepage? I don't know if we have any other examples of "Python tools as a subfolder" to base this on.

tests/infra/ccf.py
```diff
@@ -5,12 +5,12 @@
 import logging
 from contextlib import contextmanager
 from enum import Enum, IntEnum
-import infra.clients
+from ccf.clients import CCFConnectionException
```
eddyashton, line 8: Nit: To avoid confusion between `ccf` and `infra.ccf`, perhaps this should be `infra/network.py`? This just means more renames, and I think we moved from `network.py` to `ccf.py` already...

tests/tests.sh
```diff
@@ -10,6 +10,7 @@ if [ ! -f "env/bin/activate" ]
 fi
 
 source env/bin/activate
+pip install -q -U -e ../python/
```
eddyashton, line 13: Dev question: If I've done `source env/bin/activate` manually and I'm running `ctest ...` rather than `./tests.sh ...`, do I need to rerun this command each time I make changes under `python/ccf`? Or are changes picked up automagically since its a local project?

eddyashton:
I can't work out how to comment on a binary file - do we need `tests/networkcert.pem` to be checked in?

CMakeLists.txt
```diff
@@ -179,6 +179,15 @@ install(
...
+# Install CCF Python tools
+install(
+  DIRECTORY python/ccf
+  DESTINATION bin/
+  FILES_MATCHING
+  PATTERN "*.py"
+  PATTERN "*/__pycache__*" EXCLUDE
+)
+
 # Install CCF Python infrastructure
```
eddyashton, line 191: We're still installing `infra/`, and the helper scripts, and `start_network.py`. We should work out which of these can be removed, though probably in a future PR once this is on PyPI.

tests/README.md
jumaffre: Done. This is now under `python/ccf`.

tests/setup.py
jumaffre: `ccf` it is

tests/setup.py
jumaffre: I have not setup an account on PyPi with this email address yet. As I understand it, this email address can be different from the one of the account we actually use to upload to PyPi. This will be addressed in a future PR once we automate the release.

python/setup.py
jumaffre: Yep. IIRC, the issue was that automated version detection based on the git tag would break outside of git. I think we could base it off cmake (see `version.cmake`) which handles both git and non-git repos.

python/setup.py
jumaffre: Good point. Fixed now.

CMakeLists.txt
jumaffre: Absolutely. Added a task for this under #1382 for when automated releases are done.

jumaffre:
> I can't work out how to comment on a binary file - do we need `tests/networkcert.pem` to be checked in?

Good spot! Removed now.

tests/infra/ccf.py
jumaffre: I did actually mean to do that at some point but never got around to doing. I guess I can't postpone this anymore... ;)

tests/tests.sh
jumaffre: Just tested it to confirm: changes get picked up automagically and there's no other command to run.

tests/infra/ccf.py
jumaffre: Done.

PR #1384, opened 2020-07-07
Support endpoints at templated URI paths

eddyashton:
This adds basic support for using what OpenAPI calls [path templating](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#pathTemplating). So if you install an endpoint at `/users/foo/{id}`, it will be called for HTTP requests to `/users/foo/5`. We could make each endpoint-lambda re-extract the params, but since we need to do most of the matching work to do this dispatch we also expose the parsed param as `rpc_ctx->get_request_path_params()["id"]`.

The OpenAPI spec is a bit fuzzy about collisions/ambiguities. The current behaviour is that we'll prefer a fully-qualified path, and then we iterate through templated paths to find a regex match, so if you have `/foo/{id}` and `/{id}/bar`, then `/foo/bar` could go to either. I'd like to detect these collisions and make it an error at installation time, but this is awkward so ignored for now. I believe the OpenAPI spec also restricts to single elements within a path (ie - you can't write `/files/{path}` to match `/files/home/eda/photos`), so I've done the same (the regex we produce matches to the next `/`).

I've chosen to have 2 maps in `EndpointRegistry` - one for fully qualified paths and one for paths containing templates. This makes dispatch fast (try precise lookup in the fully qualified map, if that fails iterate through only the templated paths comparing regexes), but means we need to iterate through both maps in a bunch of places (when looking for supported verbs, building API doc, etc). I think this tradeoff is fine.

Only using this on a single sample endpoint for now (`/logging/private/raw_text/{id}`), I'll do a pass to use this wherever we can in a future PR.

Related: I've used "templated path" to name some of these concepts, but we're still calling this a "method" for historical reasons. We now have enough HTTP-support that we should probably get rid of method/verb/RPC terminology entirely.


achamayou:
One thing we may want to use later is https://github.com/c9s/r3, but this is a very good first step.

src/enclave/rpc_context.h
```diff
@@ -103,6 +103,8 @@ namespace enclave
     {}
   };
 
+  using PathParams = std::map<std::string, std::string>;
```
achamayou, line 106: Are params already URL-decoded?

eddyashton: No. Currently we only URL-decode (with a function called `url_unescape` because I uhhhhh, brain stumbled) the query. So `/users/Bob%20Bobson` will match `/users/{id}` with `id = "Bob%20Bobson"`. I'll add a decode of the whole path, and then they'll get `id = "Bob Bobson"`

eddyashton: After further discussion, we've decided not to decode the whole path. This may introduce awkward null bytes in the things we're still processing as strings. Instead we pass around URL-encoded strings for as long as possible, and its up to the handler to URL-decode things that may be encoded if and when they care about it.

PR #1387, opened 2020-07-08
Node state RPC

### LINKED ISSUE #1052: Node status endpoint should cover recovery status
The more general point raised by #1045 is that there should be a node status rpc, which reveals its state, public recovery progress and private recovery progress. The tests can then make use of it to become more robust to timeouts, and it's a useful thing for operators to have.

achamayou:
fix #1052

Remove `node/signed_index` and replace it with `node/state`, which is always available regardless of whether the node is recovering or not.


tests/infra/network.py
```diff
@@ -488,7 +488,8 @@ def wait_for_state(self, node, state, timeout=3):
         while time.time() < end_time:
             try:
                 with node.client(connection_timeout=timeout) as c:
-                    r = c.get("/node/signed_index")
+                    r = c.get("/node/state")
+                    LOG.success(r.result)
```
jumaffre, line 492: To remove?

src/node/rpc/serialization.h
```diff
@@ -9,13 +9,16 @@
...
   DECLARE_JSON_ENUM(
-    GetSignedIndex::State,
-    {{GetSignedIndex::State::ReadingPublicLedger, "readingPublicLedger"},
-     {GetSignedIndex::State::ReadingPrivateLedger, "readingPrivateLedger"},
-     {GetSignedIndex::State::PartOfNetwork, "partOfNetwork"},
-     {GetSignedIndex::State::PartOfPublicNetwork, "partOfPublicNetwork"}})
-  DECLARE_JSON_TYPE(GetSignedIndex::Out)
-  DECLARE_JSON_REQUIRED_FIELDS(GetSignedIndex::Out, state, signed_index)
+    ccf::State,
+    {{ccf::State::uninitialized, "uninitialized"},
```
jumaffre, line 13: This serialisation is now defined twice, also in `node_state.h` to format the `enum` when printing it with `fmt::format()`. Could we only define it once?

src/node/rpc/node_frontend.h
```diff
@@ -255,42 +255,22 @@ namespace ccf
...
+          result.last_signed_index = 0;
         else
-          result.signed_index = sig.value().seqno;
+          result.last_signed_index = sig.value().seqno;
 
-        return make_success(result);
+        return result;
       };
       make_read_only_endpoint(
-        "signed_index", HTTP_GET, json_read_only_adapter(get_signed_index))
```
jumaffre, line 292: `signed_index` is still mentioned in a few places in our docs, e.g. https://microsoft.github.io/CCF/master/operators/recovery.html?highlight=signed_index

src/node/rpc/node_frontend.h
```diff
@@ -255,42 +255,22 @@ namespace ccf
...
-        .set_auto_schema<GetSignedIndex>()
+        "state", HTTP_GET, json_read_only_adapter(get_state))
```
jumaffre, line 272: #1052 also mentions that we should expose the private recovery progress as well, which we should get in a similar way with the recovery store created in `node_state.h`.

tests/infra/network.py
achamayou: Yep, sorry.

src/node/rpc/serialization.h
achamayou: Let me see.

src/node/rpc/node_frontend.h
achamayou: Thanks, will fix!

src/node/rpc/node_frontend.h
achamayou: Yes!

doc/users/rpc_api.rst
```diff
@@ -7,15 +7,15 @@ The API can also be retrieved from a running service using the `api`_ and `api/s
 
 .. code-block:: bash
 
-    $ curl https://<ccf-node-address>/app/api --cacert networkcert.pem --key user0_privk.pem --cert user0_cert.pem
+    $ curl https://<ccf-node-address>/node/api --cacert networkcert.pem --key user0_privk.pem --cert user0_cert.pem
```
eddyashton, line 10: We don't need the user0 key/cert if we're calling the `node` frontend. But this is in the `users` docs, so I think we should either stick to user-auth or explicitly call out "you (or anyone else) can access the `/node` endpoints without auth.

achamayou: Sure, but the list was wrong, because getSignedIndex was always a node thing and not a common handler. So I'll go back to app and remove it then,

src/node/rpc/serialization.h
```diff
@@ -9,13 +9,16 @@
...
+     {ccf::State::initialized, "initialized"},
+     {ccf::State::pending, "pending"},
+     {ccf::State::partOfPublicNetwork, "partOfPublicNetwork"},
+     {ccf::State::partOfNetwork, "partOfNetwork"},
+     {ccf::State::readingPublicLedger, "readingPublicLedger"},
+     {ccf::State::readingPrivateLedger, "readingPrivateLedger"}})
+  DECLARE_JSON_TYPE(GetState::Out)
+  DECLARE_JSON_REQUIRED_FIELDS(GetState::Out, state, last_signed_index)
```
jumaffre, line 21: Other recovery seqnos should also be serialised here?

achamayou: Doh, yes.

doc/schemas/state_GET_result.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "properties": {
+    "last_recovered_seqno": {
+      "maximum": 9223372036854775807,
+      "minimum": -9223372036854775808,
+      "type": "integer"
+    },
+    "last_signed_index": {
```
jumaffre, line 9: Should this be `last_signed_seqno`?

achamayou: Yup

PR #1388, opened 2020-07-08
Remove jsonrpc leftovers

achamayou:
Remove what we have left of jsonrpc references in the code, and in particular the awkward dichotomy between result and error in the Python client.


eddyashton:
Can we also replace all of the `.rpc()` calls on `CCFClient`s in Python with `.post()`? And rename the actual implementation `def rpc` with `def send` or `def call` or something.

achamayou:
@eddyashton done

PR #1392, opened 2020-07-09
First election no longer fails

### LINKED ISSUE #1015: First election fails
In our Raft configuration, the first attempt by each node to trigger an election will fail. This is because the node has not established connections to the other nodes, and in the current code it now tries to establish a connection but doesn't actually send election requests. When the node's election timer expires for the second time, a successful election should occur.

This is handled `node_to_node.h`, where the choice to drop initial messages is documented. We should explore queuing messages rather than dropping them (with a pending size limit?), or at least indicating to the caller that the connection is unestablished so they can retry more promptly.

jumaffre:
Resolves #1015

**Before**

All messages sent on a node-to-node channel before the channel is established (i.e. ECDH key exchange round-trip not yet complete between two nodes) were dropped. So when a node became candidate and tried to send request votes to other (backup) nodes, the request vote would be dropped. Because of this, elections would normally take 2 election timeouts to complete.

**Now**

No messages on a channel are dropped as each channel now has a queue of authenticated/unencrypted messages if the channel is not established. When the channel is established, all messages are authenticated/encrypted and sent to the peer. 

Notes:
- Some refactoring of the node-to-node channels code as channels are now responsible for sending messages over the ring buffer. This makes the code much clearer but the unit tests are slightly more complex now. 
- I have reduced the time nodes wait for a new leader to be elected when the primary dies in our end-to-end infra. Let's see if it works reliably.


src/node/channels.h
```diff
@@ -46,9 +52,57 @@ namespace ccf
...
-    // Used for key exchange
-    tls::KeyExchangeContext ctx;
-    ChannelStatus status = INITIATED;
+    class OutgoingQueue
+    {
+    public:
+      struct Msg
+      {
+        NodeMsgType type;
+        std::vector<uint8_t> raw_plain; // To be intergrity-protected
```
achamayou, line 61:
```suggestion
        std::vector<uint8_t> raw_plain; // To be integrity-protected
```

src/node/channels.h
```diff
@@ -46,9 +52,57 @@ namespace ccf
...
+      void queue_msg(Msg&& msg)
+      {
+        if (outgoing_msgs.size() < max_queued_msgs)
+        {
+          outgoing_msgs.emplace(std::move(msg));
+        }
+        else
+        {
+          LOG_DEBUG_FMT(
+            "Cannot queue msg for node as channel is full (max size: {})",
```
achamayou, line 88: Why do we have a max?

achamayou: Can we use the same MSPCQ we use elsewhere?

eddyashton: I think we don't need the MPSCQ here, since the work should already have been queued onto a single thread?

The max looks like a sensible safeguard. In normal operation we'll only have a handful of queued messages, and if we have more its because establishing a channel is taking a long time (longer than multiple raft heartbeats). In the latter case we're just matching the old behaviour - saying these messages were lost in flight and dropping them - rather than queuing indefinitely.

jumaffre: As discussed offline, it's actually OK to cap the size of the queue to 1 (there's no actual need for a queue in that case). There's only one message we don't want to lose (the `RequestVote` when a node becomes candidate) and other messages are resent shortly (e.g. `AppendEntries`).

jumaffre: Done now.

.azure-pipelines-templates/test.yml
```diff
@@ -5,7 +5,7 @@ parameters:
 
 steps:
 - script: |
-    ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
+    ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --repeat-until-fail 20 -R full_test_suite
```
jumaffre, line 8: To be reverted before merging this PR.

src/consensus/raft/raft.h
```diff
@@ -579,8 +563,13 @@ namespace raft
         return;
       }
 
+      const auto prev_term = get_term_internal(r.prev_idx);
```
jumaffre, line 566: Interesting one here. It seems that this PR has uncovered a behaviour that was hidden by the fact that we never sent the first few append entries message.

On late join (this appeared on the full test suite after a recovery), it is possible that the primary sends several AE messages with the very last index `V` in its log. The late joiner will reply with a `false` AER message as it does not have the latest message `V - 1` in its log, to which the primary will respond with many AE messages, from 0 to `V`, which will be happily deserialised by the late joiner. *All is good so far.* 

However, it is possible that one of the `false` AER was in flight and only read by the primary after it has sent the AEs from 0 to `V`. In that case, the primary will send all these entries again, which is wasteful (but that's not the main issue here, and snapshots should somehow address that). 

All the new AEs will not be deserialised by the late joiner (it has already deserialised those entries) and based on the code on the left-hand side of this diff, the election timeout will not be restarted (it is guarded by the `if (r.prev_idx < commit_idx)` check on line `545`), and which will eventually cause an election. 

I've shuffled the orders of the checks around here so that the election timeout is re-started even if the AE index is behind our commit index. However, if the term is stale, we don't want to reset our election timeout. This seems to be what [the Raft reference implementation does too](https://github.com/logcabin/logcabin/blob/master/Server/RaftConsensus.cc#L1307).

achamayou: That’s a very good catch!

PR #1393, opened 2020-07-09
Use templated paths for common and proposal endpoints

eddyashton:
Trying to make a more RESTful API:
```
POST /gov/proposals
GET /gov/proposals/{proposal_id}
POST /gov/proposals/{proposal_id}/votes
GET /gov/proposals/{proposal_id}/votes/{member_id}
POST /gov/proposals/{proposal_id}/withdraw
```

The last one could be `DELETE`, but we don't actually delete it, so I think `POST` fits better.

And since I was renaming a bunch of entries in the docs, I've done a pass updating the sample governance commands with current responses and tried to clarify some of the wording around the proposal flow.

Let me know if there's any common endpoints that should get the templated-path treatment. I'd like to add `GET /user/{id}` and `GET /member/{id}`, and potentially anything else we're currently using `/read` for, but these don't exist currently.

Opinions on REST style welcome. ~~Should we be prefer `/proposals` over `/proposal`? Should we prefer `{proposal_id}` over `{id}`, especially if these are all soon to become opaque hashes (eventually there'll be some regex limiting the accepted template substitutions)?~~ I've opted for YES on both of these questions.


src/http/http_rpc_context.h
```diff
@@ -100,21 +100,6 @@ namespace http
           auto& signed_headers = parsed_sign_params->signed_headers;
           signed_headers.emplace_back(http::headers::AUTHORIZATION);
 
-          for (const auto& required_header :
```
eddyashton, line 103: This is during canonicalisation. We no longer mandate an odd set of required signed headers here. We will canonicalise anything, keeping the signed headers. Later, when we verify the signature in `http_sig,h`, we check for some required headers. These are currently static, in `http_consts.h`. There may be a use case for making this choice endpoint-specific, but we haven't seen it yet.

src/node/rpc/member_frontend.h
```diff
@@ -647,8 +672,41 @@ namespace ccf
...
+            return make_error(
+              HTTP_STATUS_BAD_REQUEST,
+              fmt::format("Proposal {} does not exist", proposal_id));
+          }
+
+          return make_success(proposal.value());
+        };
+      make_endpoint("proposal/{id}", HTTP_GET, json_adapter(get_proposal))
+        .set_auto_schema<void, Proposal>()
+        //.set_require_client_signature(true)
```
eddyashton, line 709: It seems odd to require a signature for a `GET` (turning it into a write!), but on the other hand it may be simpler to require signatures for _everything on the member frontend_, or at least _every operation on proposals_? I'm leaning towards saying this is unsigned and truly read-only.

jumaffre: Leaning towards unsigned and truly read-only too as the proposal could be read from the ledger directly anyway.

python/ccf/checker.py
```diff
@@ -27,7 +27,7 @@ def wait_for_global_commit(client, seqno, view, mksign=False, timeout=3):
 
     end_time = time.time() + timeout
     while time.time() < end_time:
-        r = client.get("/node/tx", {"view": view, "seqno": seqno})
+        r = client.get(f"/node/tx/{view}/{seqno}")
```
jumaffre, line 30: I could be swayed otherwise but I find this syntax a little odd. It makes it look as if `view` was a resource and `seqno` another resource under `view`. In practice, it's kind of true (a committed `seqno` belongs to one `view`).

tests/e2e_logging.py
```diff
@@ -302,7 +302,7 @@ def test_historical_query(network, args):
 
 
 @reqs.description("Testing forwarding on member and node frontends")
-@reqs.supports_methods("/node/tx")
+@reqs.supports_methods("/node/tx/{view}/{seqno}")
```
jumaffre, line 305: Should we assume that this endpoint is offered by all CCF services?

tests/memberclient.py
```diff
@@ -74,6 +74,31 @@ def test_update_recovery_shares(network, args):
     return network
 
 
+@reqs.description("Send an unsigned request where signature is required")
+def test_missing_signature(network, args):
```
jumaffre, line 78: This is kind of already tested in `governance_history.py` when submitting unsigned votes... but this is muh better! Could we remove the test case in `governance_history.py` and also remove the `force_unsigned` altogether?

python/ccf/checker.py
eddyashton: A fair observation. I don't have strong opinions either way, I'm happy to switch this back to query params if that's what people prefer. I'm just thinking about what the `/{view}` resource could/would hold, the leader in that term and the begin/end seqnos in that term?

jumaffre: Yeah, I'm not too strongly for one rather than the other either. `/view` should probably return the begin/end seqnos as you said. An interesting case to consider is when `seqno` does not belong to `view`: should we return HTTP `404`? I believe we currently return `UNKNOWN` or `COMMITTED` in this case and should continue to do so.

src/node/rpc/member_frontend.h
achamayou: I was going to suggest that a PBFT world, where all requests including reads are persisted, it's nice to have the same auditability on governance reads that we have on writes? On the other hand, governance state is public in the ledger, so it can absolutely be read in a non-auditable way. So I agree with pure read-only without signature (another nail in the frontend coffin).

python/ccf/checker.py
achamayou: Since this isn't really a clear-cut improvement over the current state, I suggest we stay compatible?

tests/e2e_logging.py
achamayou: It is!

doc/members/accept_recovery.rst
```diff
@@ -21,17 +21,26 @@ A member proposes to recover the network and other members can vote on the propo
         }
     }
 
-    $ ./scurl.sh https://<ccf-node-address>/gov/propose --cacert network_cert --key member1_privk --cert member1_cert --data-binary @accept_recovery.json -H "content-type: application/json"
+    $ ./scurl.sh https://<ccf-node-address>/gov/proposal --cacert network_cert --key member1_privk --cert member1_cert --data-binary @accept_recovery.json -H "content-type: application/json"
     {
-        "completed": false,
-        "id": 1
+        "proposal_id": 1,
+        "proposer_id": 0,
```
achamayou, line 27: Why do we return proposer_id?

doc/members/accept_recovery.rst
```diff
@@ -21,17 +21,26 @@ A member proposes to recover the network and other members can vote on the propo
...
+        "state": "OPEN"
     }
 
-    $ ./scurl.sh https://<ccf-node-address>/gov/vote --cacert network_cert --key member2_privk --cert member2_cert --data-binary @vote_accept_1.json -H "content-type: application/json"
-    false
+    $ ./scurl.sh https://<ccf-node-address>/gov/proposal/1/vote --cacert network_cert --key member2_privk --cert member2_cert --data-binary @vote_accept.json -H "content-type: application/json"
+    {
+        "proposal_id": 1,
```
achamayou, line 33: Why return proposal_id here?

doc/members/accept_recovery.rst
```diff
@@ -21,17 +21,26 @@ A member proposes to recover the network and other members can vote on the propo
...
+        "proposer_id": 0,
+        "state": "OPEN"
+    }
 
-    $ ./scurl.sh https://<ccf-node-address>/gov/vote --cacert network_cert --key member3_privk --cert member3_cert --data-binary @vote_accept_1.json -H "content-type: application/json"
-    true
+    $ ./scurl.sh https://<ccf-node-address>/gov/proposal/1/vote --cacert network_cert --key member3_privk --cert member3_cert --data-binary @vote_accept.json -H "content-type: application/json"
+    {
+        "proposal_id": 1,
```
achamayou, line 40: Why return anything other than the state here?

eddyashton: We want to make it clear that the response is about the proposal, not the vote. So `"ACCEPTED"` is ambiguous, though perhaps `{ "proposal_state": "ACCEPTED" }` is fine. Its maybe useful to return consistent metadata about the proposal? But maybe not - the full state can be retrieved from `GET /proposals/{id}`, and the proposal_id itself is already obvious from the URL path.

achamayou: I agree `proposal_state` is an improvement, I suspect we can remove the rest.

Alternatively, we could take a minimalist view and return application/text "PROPOSAL_PASSED".

src/node/rpc/member_frontend.h
eddyashton: I've made this truly read-only.

python/ccf/checker.py
eddyashton: Agreed, have reverted this to pass in query rather than path.

tests/memberclient.py
eddyashton: Sure, I've removed the similar bit from `governance_history.py`.

doc/members/accept_recovery.rst
eddyashton: I think we should stick to JSON objects for now. Once we have general OpenAPI support for other return formats, we can do a pass and go minimal, but for now I think the generality of JSON objects makes the API easier to use. On that line, I don't think its worth trimming fields out at the moment, if they're doing no harm?

Issue #66, opened 2019-05-13
Formally define public headers/API


achamayou:

olgavrou:
Regarding the discussion here: https://github.com/microsoft/CCF/pull/781#discussion_r374558463

Right now we make specific checks on `deserialise` depending on the consensus. We should abstract this away, for example having each consensus pass a lambda to `deserialise` or a define a deserialise lambda at table creation, which lambda will be called after deserialise has happened (or something else entirely). 

That way the consensus assumptions will be separated from the function call.

PR #1394, opened 2020-07-09
Give the C++ e2e logging test a uniquely regexable name

eddyashton:
One of our most thorough tests is `end_to_end_logging_`. This often picks things up in the CI that other tests miss, as it tests all the obscure variants of our endpoints.

However, it is also one of the most difficult to run in isolation, because `ctest` is useless. The usual way to run a single `ctest` is with a regex, but:

```
$ ctest -N -R end_to_end_logging_raft
Test project /data/src/2.CCF/build
  Test #58: js_end_to_end_logging_raft
  Test #61: lua_end_to_end_logging_raft
  Test #66: end_to_end_logging_raft

Total Tests: 3
```

I don't want the `js_` and `lua_` variants which do a subset of the work. So previously I needed to run some `-N` that would include this good test, then the devil-summoning `ctest -VV -I 66,66`. (NB: You can't add `-I 66,66` to the previous command, because even though it still says `Test #66` its now `-I 3,3`. ctest!)

Now the good test has a unique name, so we can easily run it by itself.

I'm tempted to go with `end_to_end_logging_cpp_raft`, `end_to_end_logging_js_pbft` etc, so the unique bit for each test is shorter, but we already have a _bunch_ with `lua_` as a prefix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -554,7 +554,7 @@ if(BUILD_TESTS)
   )
 
   add_e2e_test(
-    NAME lua_end_to_end_batched
+    NAME lua_e2e_batched
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/e2e_batched.py
     CONSENSUS raft
     ADDITIONAL_ARGS --app-script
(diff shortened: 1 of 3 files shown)


achamayou:
If we want shorter we could s/end_to_end/e2e/g

PR #1395, opened 2020-07-10
Fix update ack schema

achamayou:
Also spotted by Soren.


src/node/rpc/member_frontend.h
```diff
@@ -869,7 +869,7 @@ namespace ccf
         };
       make_endpoint(
         "ack/update_state_digest", HTTP_POST, json_adapter(update_state_digest))
-        .set_auto_schema<void, StateDigest>()
+        .set_auto_schema<void, MemberAck>()
```
jumaffre, line 872: I don't understand this change. Why would we want to return the whole `MemberAck` here (i.e. the new root to-be-acked + the Signed Request)? The member only needs the new root and then submit it to `gov/ack` (with `scurl.sh`) to become active (see https://microsoft.github.io/CCF/master/members/adding_member.html#activating-a-new-member). 

I can see that the endpoint code is a little misleading though as we return the whole `MemberAck`, but only the `StateDigest` in it ends up being serialised in the response. Perhaps we could `return make_success(ma->state_digest)` instead?

achamayou: No, we return and serialise the whole MemberAck at the moment, this change just fixes the schema to be compliant with the reality of what we do.

Whether we should do that or not is a different matter, and I agree that we perhaps ought to change that.

jumaffre: Ah! Good point. I hadn't noticed that, and it must have changed at some point since the initial implementation. We should only return the `StateDigest` here - the whole `MemberAck` is just for auditing purposes.

achamayou: That's being discussed in #1393, so I will merge this fix for now and we can change it there or elsewhere.

PR #1398, opened 2020-07-10
Remove mkSign

eddyashton:
UPDATE: After further discussion, we've decided now is a good time to kill off `mkSign` entirely, so that's what this PR does. It was only ever intended as a hint, and this hint is hard to justify in the RESTful API. We can get the same effect in our end-to-end tests by simply reducing the `--sig-max-ms` threshold so that we _always_ sign regularly, rather than signing on demand.

Original description below:
Noticed while debugging something else: If we `mkSign` on a backup it will fail quietly - 
```
2020-07-10T10:45:12.126076Z -0.113 0   [trace] ../src/http/http_parser.h:148        | Received url to parse: /node/mkSign
2020-07-10T10:45:12.126153Z -0.113 0   [trace] ../src/http/http_endpoint.h:185      | Processing msg(POST, /node/mkSign, , [0 bytes])
...
2020-07-10T10:45:12.131839Z -0.119 0   [fail ] ../src/consensus/raft/raft.h:313     | Failed to replicate 1 items: not leader
```

`mkSign` should be forwarded since it tries to write to the KV. It looks a bit like a Tx-free `command` because it doesn't use the normal `Tx`, it creates its own.

Draft while I add a test for this, and work out what `mkSign` should now be called.


achamayou:
mkSign only serves a dubious purpose because of a quirk in our perf tests, it's confusing at best the rest of the time. Let's get rid of it altogether.

Functional tests can lower signing interval to get the same effect as they do now.

tests/e2e_logging.py
```diff
@@ -538,15 +538,15 @@ def run(args):
             network = test_illegal(
                 network, args, verify=args.package is not "libjs_generic"
             )
-            network = test_large_messages(network, args)
-            network = test_remove(network, args)
+            # network = test_large_messages(network, args)
```
jumaffre, line 541: Un-comment?

PR #1403, opened 2020-07-14
KV snapshot serialisation/deserialisation

jumaffre:
Related to https://github.com/microsoft/CCF/issues/1299 

This PR adds support for KV store snapshot serialisation. Until now, the interface returned a snapshot object which could be passed between stores but unpractical for passing snapshots between nodes. Snapshots are now serialised so that they can be generated on one node (`store.serialise_snapshot(v)`) and deserialised in another (`store2.deserialise_snapshot(data)`).

Notes: 
- As discussed on https://github.com/microsoft/CCF/issues/1299#issuecomment-657498475, we're going for a less efficient to create snapshots by re-using most of the serialisation code for write sets. This involves more copies when serialising to `msgpack` but makes the _first_ implementation much quicker.
- I've cleanup code that we are not using anymore and that was going to be used for the more efficient snapshot generation approach. The champ still keeps track of its size though.
- The `Store` `deserialise_snapshot()` is very similar to `deserialise()` and should/will eventually be unified. I'm keeping both separate for now and will try to unify once the protocol changes have been implemented.


src/kv/generic_serialise_wrapper.h
```diff
@@ -272,16 +281,24 @@ namespace kv
       return KvOperationType::KOT_NOT_SUPPORTED;
     }
 
+    // Should only be called once, once the GCM header and length of public
+    // domain have been read
+    void read_public_header()
+    {
+      is_snapshot = public_reader.template read_next<bool>();
```
jumaffre, line 288: There is now a new snapshot flag in each serialised transaction (including write sets, where this flag is `false`).

src/kv/untyped_map.h
```diff
@@ -626,34 +605,27 @@ namespace kv::untyped
 
     std::unique_ptr<AbstractMap::Snapshot> snapshot(Version v) override
     {
+      // This takes a snapshot of the state of the map at the last entry
+      // committed at or before this version. The Map expects to be locked while
+      // taking the snapshot.
       auto r = roll.commits->get_head();
-      while (r != nullptr)
+
+      for (auto current = roll.commits->get_tail(); current != nullptr;
```
jumaffre, line 613: Simplifying this loop here which had the same logic as the one in `create_view()` (i.e. find the commit with the largest value `<=` the target `v`).

src/ds/champ_map.h
```diff
@@ -428,13 +428,13 @@ namespace champ
     Map<K, V, H> map;
     CBuffer serialized_buffer;
 
-    struct pair
+    struct KVTuple
```
achamayou, line 431: That's still a pair of data elements, essentially, despite the presence of the hashing function?

src/kv/generic_serialise_wrapper.h
```diff
@@ -380,8 +394,16 @@ namespace kv
...
-      return {current_reader->read_next_pre_serialised(),
-              current_reader->read_next_pre_serialised()};
+      return {
+        current_reader->template read_next_pre_serialised<SerialisedKey>(),
+        current_reader->template read_next_pre_serialised<SerialisedValue>()};
+    }
+
+    std::vector<uint8_t> deserialise_raw()
+    {
+      // return current_reader->read_next_raw();
```
achamayou, line 404: To be deleted?

src/kv/store.h
```diff
@@ -220,80 +220,112 @@ namespace kv
...
-      }
-
-      for (auto& map : maps)
+      if (v < commit_version())
       {
-        map.second->unlock();
+        throw ccf::ccf_logic_error(fmt::format(
+          "Attempting to snapshot at version {} which is earlier than "
+          "committed version {} ",
+          v,
```
achamayou, line 230:
```suggestion
          "Cannot snapshot at uncommitted version {}: earlier than "
          "committed version {} ",
```

Is more accurate in my opinion. "Attempting" isn't clearly an error, it may be confused for an in-progress operation, I think "Failed to", "Cannot" or something similarly definite would be preferable.

src/kv/store.h
```diff
@@ -220,80 +220,112 @@ namespace kv
...
       }
 
+      if (v > current_version())
       {
-        std::lock_guard<SpinLock> vguard(version_lock);
-        version = snapshot->get_version();
-        last_replicated = snapshot->get_version();
-        last_committable = snapshot->get_version();
+        throw ccf ::ccf_logic_error(fmt::format(
+          "Attempting to snapshot at version {} which is later "
```
achamayou, line 237: Same as above.

src/kv/store.h
```diff
@@ -220,80 +220,112 @@ namespace kv
...
+      }
+
+      for (auto r = d.start_map(); r.has_value(); r = d.start_map())
+      {
+        const auto map_name = r.value();
+
+        auto search = maps.find(map_name);
+        if (search == maps.end())
+        {
+          LOG_FAIL_FMT("Failed to deserialize snapshot");
```
achamayou, line 298: I think including the version information in the fail message would be useful and doesn't leak anything useful. The map name on the other hand is rightly debug only.

src/kv/store.h
```diff
@@ -220,80 +220,112 @@ namespace kv
...
+
+        auto map_version = d.deserialise_entry_version();
+        auto map_snapshot = d.deserialise_raw();
+
+        search->second->apply_snapshot(map_version, map_snapshot);
+      }
+
+      if (!d.end())
+      {
+        LOG_FAIL_FMT("Failed to deserialize snapshot");
```
achamayou, line 311: I think rolling both statements in the fail message is ok here.

src/kv/untyped_map.h
```diff
@@ -626,34 +605,27 @@ namespace kv::untyped
...
-          snapshot(r->state);
-
       return std::make_unique<Snapshot>(
-        name, security_domain, replicated, r->version, std::move(snapshot));
+        name, security_domain, replicated, r->version, StateSnapshot(r->state));
     }
 
-    void apply(const std::unique_ptr<AbstractMap::Snapshot>& s) override
+    void apply_snapshot(
+      Version version, const std::vector<uint8_t>& snapshot) override
```
achamayou, line 628: We haven't been very good with this, but in places where the API is largely thread-safe except in places, it would be good to prefix with unsafe_ those calls that aren't.

src/ds/champ_map.h
jumaffre: Yeah, not too sure about the name here. I think `KVTuple` is generic enough and it's only used (so far) by one function anyway. Happy to change it though.

src/kv/store.h
jumaffre: Good point. I have re-phrased to `"Cannot snapshot at version {} which is earlier than committed version {}"` (slightly different from what you proposed as earlier version was probably committed).

src/kv/untyped_map.h
jumaffre: True. I think there's a comment at the top of each function that requires a lock on the `Map` before being called. Since this is an internal call, I would think it's OK to leave it as is for now but we can ticket that separately.

PR #1404, opened 2020-07-14
Snapshot benchmark

jumaffre:
Follow up from #1403 (to be merged after it!)

~**Only look at diff in `kv_bench.cpp` and `upload_pico_metrics.py`**~

Adds benchmarks for serialising and deserialising KV snapshots. Should hopefully show up in the cimetrics report 🤞


src/kv/generic_serialise_wrapper.h
```diff
@@ -380,8 +394,16 @@ namespace kv
...
-      return {current_reader->read_next_pre_serialised(),
-              current_reader->read_next_pre_serialised()};
+      return {
+        current_reader->template read_next_pre_serialised<SerialisedKey>(),
+        current_reader->template read_next_pre_serialised<SerialisedValue>()};
+    }
+
+    std::vector<uint8_t> deserialise_raw()
+    {
+      // return current_reader->read_next_raw();
```
ashamis, line 404: delete this?

src/kv/store.h
```diff
@@ -220,80 +220,112 @@ namespace kv
...
-        std::lock_guard<SpinLock> vguard(version_lock);
-        version = snapshot->get_version();
-        last_replicated = snapshot->get_version();
-        last_committable = snapshot->get_version();
+        throw ccf ::ccf_logic_error(fmt::format(
+          "Attempting to snapshot at version {} which is later "
+          "than current version {}",
+          v,
+          current_version()));
       }
```
ashamis, line 241: consider using CCF_ASSERT_FMT

jumaffre: Good point. Done now.

src/kv/store.h
```diff
@@ -297,7 +294,8 @@ namespace kv
         {
           LOG_FAIL_FMT("Failed to deserialise snapshot at version {}", v);
           LOG_DEBUG_FMT("No such map in store {}", map_name);
-          return DeserialiseSuccess::FAILED;
+          success = false;
```
jumaffre, line 297: Fixing early exist issue here where the maps would not be unlocked.

src/kv/store.h
```diff
@@ -306,15 +304,20 @@ namespace kv
         search->second->apply_snapshot(map_version, map_snapshot);
       }
 
+      for (auto& map : maps)
+      {
+        map.second->unlock();
+      }
+
       if (!d.end())
```
ashamis, line 312: consider:  if(success && !d.end())

jumaffre:
> It might be interesting to add a different configuration. Where we have a set small number of transactions but change the number of maps that are serialized and deserialized.

@ashamis You're absolutely right. I've updated the benchmark accordingly, testing with different number of maps (20 and 100) and with different number of keys in each map. 

Interestingly, the number of maps does not seem to affect the number of serialisation/deserialisation performance. See output below (`Dim` is number of maps, and template value if number of keys in the map).

```
serialise_snapshot:
          ser_snap<100> * |      20 |     0.396 |   19820 |      - |    50454.0
           ser_snap<1000> |      20 |     5.341 |  267040 | 13.473 |     3744.7
          ser_snap<100> * |     100 |     2.053 |   20527 |      - |    48716.2
           ser_snap<1000> |     100 |    28.875 |  288751 | 14.067 |     3463.2 << Displayed in cimetrics (snapshot size: 4MB)
deserialise_snapshot:
          des_snap<100> * |      20 |     1.039 |   51970 |      - |    19241.8
           des_snap<1000> |      20 |    15.591 |  779542 | 15.000 |     1282.8
          des_snap<100> * |     100 |     5.157 |   51567 |      - |    19392.2
           des_snap<1000> |     100 |    77.852 |  778519 | 15.097 |     1284.5 << Displayed in cimetrics  (snapshot size: 4MB)



```

achamayou:
@jumaffre 
```
src/kv/test/kv_bench.cpp:239:  LOG_INFO_FMT("Size of snapshot: {}", snapshot.size()); // TODO: Delete
```

PR #1406, opened 2020-07-15
Pbft tests in daily only

achamayou:


CMakeLists.txt
```diff
@@ -42,7 +42,11 @@ configure_file(
   ${CCF_DIR}/src/host/version.h.in ${CCF_GENERATED_DIR}/version.h @ONLY
 )
 
-set(CONSENSUSES raft pbft)
+if(PBFT_TESTS)
+  set(CONSENSUSES raft pbft)
```
eddyashton, line 46:
```suggestion
  set(CONSENSUSES "raft;pbft")
```

If we make this a list, we can use `IN_LIST` on it.

eddyashton: Ah, it seems this already creates a list, cool.

PR #1408, opened 2020-07-16
File chunk chaining and getting transactions

Bamuir3:
I have added a parameter to the initializer for the ledger so that you are able to iterate through multiple file chunks of a ledger, along with a few changes to the iterator and next function to allow that. The parameter is default initialized to false. Considering verifying that all of the files are sequential, for example, ledger1-20 must be followed by ledger21-.... and so forth, but not sure if that will be better for the ledger verification tool instead. 

I have also added a get_public_tx to retrieve the bytes of the transaction.


python/ccf/ledger.py
```diff
@@ -169,13 +180,42 @@ def __next__(self):
...
 class Ledger:
 
-    _filename = None
+    _filenames = []
+    _directory = None
+    _fileindex = 0
 
-    def __init__(self, filename):
-        self._filename = filename
+    def __init__(self, name:str, is_dir:bool == False):
```
jumaffre, line 189: I think it's OK for `name` to always be a directory here. The directory contains one or multiple ledgers that can be parsed with this class.

python/ccf/ledger.py
```diff
@@ -169,13 +180,42 @@ def __next__(self):
...
+        if is_dir:
+            contents = os.listdir(name)
+            # Sorts the list based off the first number after ledger_ so that the ledger is verified in sequence 
+            sort = sorted(contents, key=lambda x: int(
+                x.replace(".committed", "").replace("ledger_", "").split("-")[0]))
+
+            for chunk in sort:
+                # Add only the .committed ledgers to be verified 
+                if os.path.isfile(os.path.join(name, chunk)):
+                    if chunk.find(".committed") != -1:
```
jumaffre, line 199: Could you use `chunk.endswith(".committed")` here?

python/ccf/ledger.py
```diff
@@ -169,13 +180,42 @@ def __next__(self):
...
-        self._filename = filename
+    def __init__(self, name:str, is_dir:bool == False):
+        if is_dir:
+            contents = os.listdir(name)
+            # Sorts the list based off the first number after ledger_ so that the ledger is verified in sequence 
+            sort = sorted(contents, key=lambda x: int(
+                x.replace(".committed", "").replace("ledger_", "").split("-")[0]))
+
+            for chunk in sort:
+                # Add only the .committed ledgers to be verified
```
jumaffre, line 197: It would be great to print out the names of the uncommitted files that are not included here, possibly with `LOG.warning()`.

python/ccf/ledger.py
jumaffre, line 189: `= False` with only one equal sign.

Bamuir3: Removed the parameter itself, just checking now if os.path.isdir(name)

Bamuir3: removed param

python/ccf/ledger.py
```diff
@@ -173,10 +192,46 @@ def __next__(self):
 
 class Ledger:
 
-    _filename = None
+    _filenames = []
+    _fileindex = 0
+
+    def __init__(self, name: str):
```
jumaffre, line 198: The `governance_test` on the CI (see [here](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=10713&view=logs&j=ee9484ec-b476-577b-aa0e-d5ddfcbcd149&t=e2c1a4cf-16d0-5038-f9f6-343f9806d38b&l=700)) is still failing because of an edge case. In the test, there is only one chunk at this point which is not yet committed. Hence, `len(self._filenames)` is `0`.

Would you mind updating the constructor of `Ledger` slightly so that it works with this edge case too? I also think the `if os.path.isdir(name)` check is no longer required as we always expect `name` to be a directory.

achamayou:
@Bamuir3 changes look good, thank you for contributing these improvements to the framework!

python/ccf/ledger.py
Bamuir3: Updated now so that even files that are not .committed can be used, but logging a warning just in case.

python/ccf/ledger.py
```diff
@@ -1,8 +1,12 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache 2.0 License.
 import io
-import msgpack.fallback as msgpack  # Default implementation has buggy interaction between read_bytes and tell, so use fallback
+from os import sendfile
```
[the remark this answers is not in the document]
Bamuir3: No I will remove, missed that

python/ccf/ledger.py
```diff
@@ -44,12 +47,16 @@ def __init__(self, buffer):
         self._buffer = buffer
         self._buffer_size = buffer.getbuffer().nbytes
         self._unpacker = msgpack.Unpacker(self._buffer, **UNPACK_ARGS)
-        self._is_snapshot = self._read_next()
```
jumaffre, line 47: This seems to have been deleted.

PR #1410, opened 2020-07-16
Do not serialise if length of vector is 0

jumaffre:
Fixes daily SAN `kv_test` failure: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=10660&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=bbcdfa66-260b-5619-96df-9537b476c35e&l=195

We should not serialise a raw entry if its size is 0 - which happens in the case of snapshots when a map is empty.

Daily build run for this PR (in progress): https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=10670&view=results

### DIFF
--- a/src/kv/msgpack_serialise.h
+++ b/src/kv/msgpack_serialise.h
@@ -41,7 +41,10 @@ namespace kv
     {
       const uint64_t size = entry.size();
       sb.write(reinterpret_cast<char const*>(&size), sizeof(size));
-      sb.write(reinterpret_cast<char const*>(entry.data()), entry.size());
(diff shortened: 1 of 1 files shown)


eddyashton:
Tempted to just blacklist the msgpack files. If we were calling `memcpy` directly then sure, we should avoid the undefined behaviour (as we do in `serialized.h`) ourselves. We should avoid adding too many workarounds for edge-case UB in 3rdparty libraries, but I think this case is fine.

PR #1411, opened 2020-07-16
Don't segfault with unquoted values in auth header

eddyashton:
The HTTP Signatures spec is vague, so we accept either quoted or unquoted values when parsing the `Authorization` header.

eg:
`Authorization: Signature algorithm="ecdsa-sha256", headers="(request-target) host date", ...`
or
`Authorization: Signature algorithm=ecdsa-sha256, headers=(request-target) host date, ...`
or any mixed combination


src/http/http_sig.h
```diff
@@ -285,9 +279,11 @@ namespace http
           auto k = p.substr(0, eq_pos);
           auto v = p.substr(eq_pos + 1);
 
-          // Remove inverted commas around value
-          v.remove_prefix(v.find_first_of("\"") + 1);
-          v.remove_suffix(v.size() - v.find_last_of("\""));
+          // Remove quotes around value, if present
+          if (v.size() >= 2 && v[0] == '"' && v[v.size() - 1] == '"')
+          {
+            v = v.substr(1, v.size() - 2);
```
eddyashton, line 285: This is the actual fix: Only try to strip if quotes are present. Also avoid unnecessary `remove_XXXfix`, which have UB on invalid ranges, whereas substr will only throw.

src/http/test/http_test.cpp
```diff
@@ -307,4 +309,161 @@ DOCTEST_TEST_CASE("Escaping")
...
+  // Produce signed requests with some formatting variations, ensure we can
+  // parse them
+  auto kp = tls::make_key_pair();
+
+  http::Request request("/foo", HTTP_POST);
+  request.set_query_param("param", "value");
+  request.set_query_param("pet", "dog");
+  request.set_header("Host", "example.com");
+  request.set_header("Date", "Sun, 05 Jan 2014 21:31:40 GMT");
+  request.set_header("Content-Type", "application/json");
```
eddyashton, line 367: This is based on the test vectors from the draft RFC. I was going to use them directly, but our minimal crypto doesn't support the curve used with those test values, so instead we just test that we can parse our own produced values.

src/http/http_consts.h
```diff
@@ -11,7 +11,9 @@ namespace http
     static constexpr auto AUTHORIZATION = "authorization";
     static constexpr auto CONTENT_LENGTH = "content-length";
     static constexpr auto CONTENT_TYPE = "content-type";
+    static constexpr auto DATE = "date";
```
achamayou, line 14: Do we still want the date in here? I thought the latest spec didn't require it and so we wanted to remove it.

src/http/http_sig.h
```diff
@@ -285,9 +279,11 @@ namespace http
           auto k = p.substr(0, eq_pos);
           auto v = p.substr(eq_pos + 1);
 
-          // Remove inverted commas around value
-          v.remove_prefix(v.find_first_of("\"") + 1);
-          v.remove_suffix(v.size() - v.find_last_of("\""));
+          // Remove quotes around value, if present
+          if (v.size() >= 2 && v[0] == '"' && v[v.size() - 1] == '"')
```
achamayou, line 283: font()/back()? this is correct too of course.

src/http/http_sig.h
achamayou: Should we error early on unmatched quotes?

src/http/http_consts.h
eddyashton: We're not enforcing that this is signed, just naming the standard header for use in the test, and potentially for future use when building HTTP requests.

src/http/http_sig.h
eddyashton: Yeah this is probably worth doing. It blocks something like `separator="`, but I think that's an edge case we don't need to support (and it remains _possible_ as `separator="""`), and should produce a more useful error when we misparse `foo="bar,baz"` as `foo="bar`, `baz"`.

eddyashton: We now return an error for unbalanced quotes, and test this in `http_test`.

PR #1415, opened 2020-07-17
Update demo doc

eddyashton:
Replace docs references to `e2e_scenarios.py` with current quickstart tutorial - use `start_test_network.sh` and curl manually.

Also updates `member_rpc_api.rst` with links to correct current schema, and brief descriptions of what each gov endpoint actually does.


doc/developers/demo.rst
```diff
@@ -1,76 +1,127 @@
...
+- verifying that each node has successfully joined the new service
+- proposing and passing governance votes using the generated member identities
 
-    $ cd build
-    $ ctest -VV -R e2e_scenario -N
+The following command can be run from your ``CCF/build`` directory to create and run a test network:
+
+.. code-block:: bash
 
+    $ ../start_test_network.sh --package ./liblogging.enclave.so.signed
```
jumaffre, line 25: Minor: `$ cd CCF/build` just in case readers only read the code snippets?

doc/developers/demo.rst
```diff
@@ -1,76 +1,127 @@
...
+
+When establishing a TLS connection with a CCF service both the service and client must prove their identity.
+
+The service identity is created at startup. The initial node generates a fresh private key which exists solely within the service's enclaves. A certificate of the corresponding public key is emitted (``networkcert.pem``) and used by all subsequent connections to confirm they are communicating with the intended service.
+
+Each member and user is identified by the cert with which they were registered with the service, either at genesis or in a subsequent ``new_member`` or ``new_user`` governance proposal. Access to the corresponding private key allows a client to submit commands as this member or user. For this test network these are all freshly generated and stored in the same common workspace for easy access. In a real deployment only the certificates would be shared; the private keys would be distributed and remain confidential.
+
+When using curl the server's identity is provided by ``--cacert`` and the client identity by ``--cert`` and ``--key``. Resources under the ``/gov`` path require member identities, while those under ``/app`` require user identities.
```
jumaffre, line 50: Nitpick: "while those under `/app` _typically_ require user identities" (as we can install handlers without authz, and will make use of those more and more in the future).

eddyashton: I hummed and hawed about this a bit, because I expect it to change soon (with frontend abolition, the auth requirement is entirely per-endpoint, and apps can probably add user, unauth'd, and member endpoints?). So I don't want to be too specific about the current options. But I think your wording is accurate for now, so I'll add it.

doc/members/member_rpc_api.rst
```diff
@@ -3,76 +3,106 @@ Member RPC API
 
 As well as the following methods, :ref:`users/rpc_api:Common Methods` are also available to members.
 
-ack
----
+POST /gov/ack/update_state_digest
+---------------------------------
+
+Get the current state digest which should be signed to indicate member liveness.
```
ashamis, line 9: should be? can this be replace with either: "is" or "optionally"

doc/developers/demo.rst
```diff
@@ -1,76 +1,129 @@
 End-to-end demo
 ===============
 
-This document explains how the Python testing infrastructure is used to run a complete end-to-end test of CCF.
+This document explains how to spin up a test CCF network and submit simple commands to it using `curl`_. We use curl here because it is basic, general, and broadly available - you should be able to get the same results with any HTTP client, provided you can configure the appropriate caller and CA identities. Once you have built your own app, you should be able to test it in the same way - simply replace ``liblogging`` with the name of your app binary, and call the endpoints defined by your app.
```
ashamis, line 4: I would drop the word basic ==> consider: "we use curl, an industry stand tool, " ...

PR #1416, opened 2020-07-21
Compile warnings enforcing and fixes

jumaffre:
We now check for warnings when compiling CCF (specifically,`cchost`, `libccf` and all enclave applications built via `add_ccf_app()`, both in normal and virtual modes).

Nothing major to report. The changes are mostly related to unused variables (happy to pass `-Wno-unused-parameter` if we don't want to bother with this), incorrect order of member variables initialisation in constructors, undesirable `std::move`s, signed vs. unsigned comparisons, etc. I'll point out interesting things inline with the PR.


src/ds/ring_buffer_types.h
```diff
@@ -137,6 +137,12 @@ namespace ringbuffer
       "No payload specialization for this Message");
   };
 
+#define DECLARE_RINGBUFFER_MESSAGE_NO_PAYLOAD(MTYPE) \
```
jumaffre, line 140: Some messages don't have a payload (e.g. `stop` or `stopped`) and clang complains about the _empty_ variables args to `DECLARE_RINGBUFFER_MESSAGE_PAYLOAD()`.

src/node/rpc/endpoint_registry.h
```diff
@@ -269,6 +269,7 @@ namespace ccf
       Endpoint& set_allowed_verb(RESTVerb v)
       {
         const auto previous_verb = verb;
+        verb = v;
```
jumaffre, line 272: I think this is the only functional change (for a deprecated feature). @eddyashton can you confirm if this is correct?

src/node/rpc/member_frontend.h
```diff
@@ -847,58 +847,57 @@ namespace ccf
         .set_require_client_signature(true)
         .install();
 
-      auto get_vote =
-        [this](ReadOnlyEndpointContext& args, nlohmann::json&& params) {
-          if (!check_member_active(args.tx, args.caller_id))
-          {
-            return make_error(HTTP_STATUS_FORBIDDEN, "Member is not active");
-          }
+      auto get_vote = [this](ReadOnlyEndpointContext& args, nlohmann::json&&) {
```
eddyashton, line 850: Nit: Not a task for this PR, but if we ever make a change to our clang-format I'd really like to fix this behaviour. Minor reflowing shouldn't the indentation of so many lines.

src/node/rpc/node_call_types.h
```diff
@@ -31,8 +31,10 @@ namespace ccf
     {
       ccf::State state;
       kv::Version last_signed_seqno;
-      std::optional<kv::Version> recovery_target_seqno;
-      std::optional<kv::Version> last_recovered_seqno;
+
+      // Only on recovery
+      kv::Version recovery_target_seqno;
```
eddyashton, line 36: Why have these lost their `std::optional`s?

src/node/script.h
```diff
@@ -43,6 +43,15 @@ namespace ccf
   };
 
   DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(Script);
+
+  // Current limitation of the JSON macros: It is necessary to defined
+  // DECLARE_JSON_REQUIRED_FIELDS for Script even though there are no required
+  // fields. This raises some compiler warnings that are disabled locally.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+#pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
```
eddyashton, line 52: Should these be in `json.h`? We already have some pragma ignores there.

src/ds/ring_buffer_types.h
eddyashton: Minor: If we can disable that warning at the macro definition point, it'd be nicer than introducing a new macro name to the API surface.

eddyashton:
I'm not a huge fan of `(void)foo` peppered throughout the code. Since these are mostly unused params to match an expected signature, I think we should prefer removing the parameter name or use `-Wno-unused-parameter`.

If we're sticking with an inline warning-disabler, I'd prefer `CCF_UNUSED(foo)` to `(void)foo`, as the _intent_ of the invocation makes it far more readable. In fact since we're already reliant on modern compilers, perhaps we should be using `[[maybe_unused]]` attribute? We can put this into the function arguments (`void foo([[maybe_unused]]int i)` rather than `void foo(int i) { \n UNUSED(i);`)

src/node/rpc/node_call_types.h
jumaffre: Good spot! My bad.

CMakeLists.txt
```diff
@@ -122,6 +123,7 @@ if("virtual" IN_LIST COMPILE_TARGETS)
   )
 
   target_compile_options(ccf.virtual PUBLIC -stdlib=libc++)
+  add_warning_checks(ccf.virtual)
```
achamayou, line 126: As discussed, it would have been nicer to have this be applied automatically by directory than something we need to manually opt into, but I guess it wasn't possible.

src/apps/js_generic/js_generic.cpp
```diff
@@ -3,7 +3,12 @@
 #include "enclave/app_interface.h"
 #include "kv/untyped_map.h"
 #include "node/rpc/user_frontend.h"
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+#pragma clang diagnostic ignored "-Wc99-extensions"
```
achamayou, line 9: That should not be necessary if we correctly -isystem 3rdparty, could you double check?

src/apps/js_generic/js_generic.cpp
```diff
@@ -16,9 +21,13 @@ namespace ccfapp
 
   using Table = kv::Map<std::vector<uint8_t>, std::vector<uint8_t>>;
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wc99-extensions"
+
```
achamayou, line 26: Likewise.

src/apps/logging/logging.cpp
```diff
@@ -228,6 +228,15 @@ namespace loggingapp
         const auto& cert_data = args.rpc_ctx->session->caller_cert;
         const auto ret =
           mbedtls_x509_crt_parse(&cert, cert_data.data(), cert_data.size());
+        if (ret != 0)
```
achamayou, line 231: Good catch.

src/ds/ccf_exception.h
```diff
@@ -3,7 +3,10 @@
 #pragma once
 
 #ifdef VIRTUAL_ENCLAVE
-#  include "../3rdparty/backward-cpp/backward.hpp"
+#  pragma clang diagnostic push
+#  pragma clang diagnostic ignored "-Wundef"
+#  include <backward-cpp/backward.hpp>
```
achamayou, line 8: Same as before with -isystem?

src/node/rpc/member_frontend.h
achamayou: Things may or may not get better once we go clang*-10.

CMakeLists.txt
jumaffre: Agreed. I think this bigger cmake refactoring should be done in a different PR though.

src/apps/js_generic/js_generic.cpp
jumaffre: You're right. The include needed updating to `#include <quickjs/quickjs.h>` 👍

src/apps/js_generic/js_generic.cpp
jumaffre: This one seems necessary because of the `JS_NULL` macro which we get from QuickJS.

src/ds/ring_buffer_types.h
jumaffre: As discussed, left it as in for now as it gives better compile-checks for payload-less messages.

src/apps/js_generic/js_generic.cpp
achamayou: Ah yes!

PR #1417, opened 2020-07-21
Snapshot size calculation fix

jumaffre:
Noticed this when plugging the new snapshot mechanism into the rest of CCF. The size of the pre-insertion serialised K,V pair returned by `put_mut()` was actually the size of the post-insertion serialised K,V. This worked so far in our unit tests because we use same keys and values but didn't in the case of the `ccf.member_acks` table which changes in size as members ack.

### DIFF
--- a/src/ds/champ_map.h
+++ b/src/ds/champ_map.h
@@ -243,7 +243,8 @@ namespace champ
       const auto& entry0 = node_as<Entry<K, V>>(c_idx);
       if (k == entry0->key)
       {
-        auto current_size = get_size_with_padding<K, V>(entry0->key, entry0->value);
+        auto current_size =
+          get_size_with_padding<K, V>(entry0->key, entry0->value);
         nodes[c_idx] = std::make_shared<Entry<K, V>>(k, v);
         return current_size;
       }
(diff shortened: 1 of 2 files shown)


src/ds/champ_map.h
```diff
@@ -498,7 +500,7 @@ namespace champ
         "size:{}, map->size:{} ==> count:{}, vect:{}",
```
jumaffre, line 500: This assert has certainly saved hours of debugging!

eddyashton:
> This worked so far in our unit tests because we use same keys and values

Can we add a unit test that uses different key/value types to stress this?

jumaffre:
> > This worked so far in our unit tests because we use same keys and values
> 
> Can we add a unit test that uses different key/value types to stress this?

@eddyashton Good point. I have added a unit test for this.

src/ds/test/map_test.cpp
```diff
@@ -140,12 +140,9 @@ TEST_CASE("persistent map operations")
 static const champ::Map<K, V, H> gen_map(size_t size)
 {
   champ::Map<K, V, H> map;
-  for (uint32_t j = 0; j < 2; ++j)
+  for (uint64_t i = 0; i < size; ++i)
```
achamayou, line 143: size_t

PR #1421, opened 2020-07-21
Fix --gov-script requirements

eddyashton:
Fix daily build failure, caused by the code changes in [#1415](https://github.com/microsoft/CCF/pull/1415/files#diff-797fc5ba90a3fcd2d165899c4b735d24). We don't need a normalised abs path (what the previous code did), just an abs path. Additionally, we can't make it required because its actually _not_ required for the `--recover` case. ~~This sadly means it is not required for anything apart from `start_network.py`, but if its ever missing we'll get ugly errors later...~~ Instead, we assert its set just before use, at least getting a clear error.

### DIFF
--- a/tests/infra/e2e_args.py
+++ b/tests/infra/e2e_args.py
@@ -8,12 +8,11 @@ import sys
 
 
 def absolute_path_to_existing_file(arg):
-    abs_arg = os.path.abspath(arg)
-    if arg != abs_arg:
+    if not os.path.isabs(arg):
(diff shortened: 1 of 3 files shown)


eddyashton:
[Here](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=10844&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=fe98d859-2537-5570-75af-de67d4290f3a) is a successful run of the Daily CI's "Test installed CCF" job on this branch.

Issue #1183, opened 2020-05-15
Metrics and state endpoints


achamayou:
Expose a metrics endpoint (based on https://microsoft.github.io/CCF/master/users/rpc_api.html?highlight=status#getmetrics) with throughput and latency information, for service monitoring purposes. We also probably want to expose things like that conflict rates/counts.

achamayou:
After discussion, we probably want multiple different metric endpoints to cover different aspects of the service.

Consensus:

1. `/commit`, which exposes committed `seqno` and `view` on the node
2. `/status`, which exposes the status of the node (also see #1052)
2. `/topology`, which exposes the node's knowledge of the service topology (ie. primary in current config, active configs if primary etc)

Resource usage:

1. `/memory` which should expose the total vs used amount of memory

Endpoint-level metrics:

For each business endpoint, we want to expose:

1. number of requests
2. number of successful responses (2xx. 3xx)
3. number of error responses (4xx)
4. number of failure responses (5xx)

We will probably want to add more detail to this over time (request/response size and latency distribution over a rolling period), but this should be enough to start with.

lynshi:
Will a user or member identity be required to communicate with these endpoints, or can the client be unauthenticated? In particular, `/status` will have to be open because a joining node has no access to the identity tables if it's not part of the network yet. In general, it's more convenient to not have to provide certificates for these as well as long as they're read-only.

achamayou:
@lynshi I don’t think we want to require authorisation, they don’t give access to any confidential information. We can change that easily if need be.

achamayou:
Update: `/memory` is blocked on the necessary API being available in Open Enclave.

`/topology` or perhaps `consensus/config` is queued behind current work on raft.

achamayou:
`/config` is there now, but `/memory` is pending the next Open Enclave release (changes are in master now).

PR #1422, opened 2020-07-22
Initial endpoint metrics implementation

achamayou:
#1183

Add a basic `/endpoint_metrics`:

```
{'metrics':
  [
    {'calls': 0, 'errors': 0, 'failures': 0, 'path': 'api', 'verb': 'GET'},
    {'calls': 0, 'errors': 0, 'failures': 0, 'path': 'api/schema', 'verb': 'GET'},
    ...
  ]}
```

Calls are all user requests that were successfully dispatched to an endpoint, errors are 4xx and failures are 5xx.

Forwarded calls are counted as calls, and as failures if they fail to forward. Failures in actual logic are counted on the executing node rather than the forwarder.


src/node/rpc/frontend.h
```diff
@@ -252,6 +253,8 @@ namespace ccf
         }
       }
 
+      endpoint->metrics.calls++;
```
eddyashton, line 256: Is there a reason that the `404`s and `405`s above aren't counted in any of the metrics?

src/apps/js_generic/js_generic.cpp
```diff
@@ -290,7 +290,7 @@ namespace ccfapp
 
       auto scripts = tx.get_view(this->network.app_scripts);
       scripts->foreach([&out](const auto& key, const auto&) {
-        out.methods.push_back(key);
+        out.endpoints.push_back({"POST", key});
```
eddyashton, line 293: This isn't correct. We currently have keys like `GET log/private`, and use these for dispatch (see ~line 171 in `js_generic.cpp`). For backwards compatibility, if we don't a `verb` prefix then we'll actually match anything (ie - if we install a handler at just `foo/bar`, it'll accept `GET foo/bar`, `DELETE foo/bar`, etc).

This backwards compatibility is probably pointless since our js/lua support is still barebones, so we should probably instead require that we have a verb-prefix when we add things to the scripts table. Then here we can split on `" "` (or a better delimiter) to find the supported verb and path.

src/node/rpc/frontend.h
achamayou: Sorry, I don't understand what you mean by that.

achamayou: If you mean the failure to dispatch, I don't know what endpoint we would attribute those to. Perhaps they ought to be their own category?

eddyashton: Ah of course, any errors before this aren't attributable to a specific endpoint. Fair enough! I'd be tempted to thrown in a global "dispatch failures" counter, but ideally that'd be incremented from everywhere that returns a pre-dispatch error, which is awkward. So I think ignoring them for now is fine!

src/apps/js_generic/js_generic.cpp
achamayou: Thanks for the heads up!

achamayou:
@eddyashton what do you think of re-structuring the output for this and for /api along those lines:

```
{'paths':
  [
    'api': {'GET': {'calls': 0, 'errors': 0, 'failures': 0}},
    'foo': {'GET': {'calls': 0, 'errors': 0, 'failures': 0},
              'PUT': {'calls': 0, 'errors': 0, 'failures': 0},
    },
    ...
  ]}
```

That's closer to the OpenAPI structure and it seems easier to use.

eddyashton:
@achamayou I think that's a sensible structure for `/metrics`.

For `/api`, the goal is to return a real OpenAPI document (I have a branch that's 80% of the way there, just need to find time to finish it). I don't mind if we tweak the structure a little, but I don't think its worth investing much effort/thought when its going to be replaced 'soon'. Its only really used by the `schema_test` and for quick manual validation of test networks, so I don't know if there's value in making its temporary form more structured?

achamayou:
Ok, I'll change /metrics and leave /api as is. They are extremely similar now, and I think there's a good argument for keeping them similarly structured/keyed in the future too, but if you've got the OpenAPI changes coming soon, that's fine.

src/apps/js_generic/js_generic.cpp
```diff
@@ -290,8 +290,14 @@ namespace ccfapp
...
-        out.methods.push_back(key);
-        return true;
+        size_t s = key.find(' ');
+        if (s != std::string::npos)
+        {
+          out.endpoints.push_back(
+            {key.substr(0, s), key.substr(s + 1, key.size() - (s + 1))});
+          return true;
+        }
+        return false;
```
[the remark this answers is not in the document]
achamayou: I was doing that while testing, but it didn't catch any, so yes, defaulting to POST is ok. It's fairly temporary anyway, I think I want the keys to be structured rather than space separated, and we can do that when we make the handlers independent.

tests/e2e_logging.py
```diff
@@ -220,6 +220,31 @@ def test_raw_text(network, args):
...
 
+@reqs.description("Read metrics")
+@reqs.supports_methods("endpoint_metrics")
+def test_metrics(network, args):
+    primary, _ = network.find_primary()
+
+    with primary.client("user0") as c:
+        r = c.get("/app/endpoint_metrics")
+        metrics = {m["path"]: m for m in r.body["metrics"]}
+        assert metrics["endpoint_metrics"]["calls"] == 1
```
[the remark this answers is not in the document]
achamayou: Yes, good point.

PR #1427, opened 2020-07-23
Build CCF with LVI mitigations

### LINKED ISSUE #1315: CCF should build and use a build of OpenEnclave using LVI mitigations
Instructions for OpenEnclave are here: https://github.com/openenclave/openenclave/blob/1b67e0e1332e9c4890dbeb449cbec9c91f9f14e5/docs/GettingStartedDocs/Contributors/AdvancedBuildInfo.md#lvi-mitigation

Unfortunately it looks like only clang-7 is supported at the moment.

eddyashton:
Fixes #1315

- Adds a `LVI_MITIGATIONS` option to CCF's CMake
- If enabled, this links against the `-lvi-cfg` variants of OE libraries, and adds various `-mlfence` options when compiling
- Adds some CMake configuration (based closely on OE's) to enable LVI mitigations for a target + use custom toolchain
- Installs the LVI toolchain in ansible, based on OE's `install_lvi_mitigation_bindir` script, currently targetting only clang-8

Opening as a draft PR so we can run it for a while to get perf impact, and to discuss intended integration path. I've built a CI docker image with these changes, and also run the updated playbooks on the perf machines. Currently this adds `-DLVI_MITIGATIONS=ON` to _all CI builds_. If the results are acceptable, I'll instead make `ON` the default value for this setting. Another option is to add a separate LVI-mitigated CI job, but we'd prefer not to expand the test matrix further.


cmake/common.cmake
```diff
@@ -29,6 +29,35 @@ if(VERBOSE_LOGGING)
   set(TEST_HOST_LOGGING_LEVEL "debug")
 endif()
 
+option(LVI_MITIGATIONS "Enable LVI mitigations" OFF)
```
achamayou, line 32: We should have this default to ON.

getting_started/setup_vm/roles/openenclave/tasks/binary_install.yml
```diff
@@ -4,4 +4,82 @@
...
+
+- name: Copy invoke_compiler to OE LVI dir
+  copy:
+    remote_src: yes
+    src: "{{ oe_lvi_scripts_dir }}/invoke_compiler"
+    dest: "{{ oe_lvi_bin_dir }}/invoke_compiler"
+    mode: "0555"
+  become: yes
+
+- name: Create OE LVI symlink to clang-8
```
achamayou, line 23: I suggest lifting 8 to a clang_version under vars/, to make future upgrades easier.

.azure-pipelines-templates/ninja.yml
```diff
@@ -1,4 +1,10 @@
 steps:
+
+# TODO: Temporary debugging feature - show the commands for an enclave so we can confirm LVI mitigations
```
achamayou, line 3: Have we checked the binaries built with the mitigations do contain a number of additional lfences as expected?

cmake/common.cmake
```diff
@@ -29,6 +29,35 @@ if(VERBOSE_LOGGING)
   set(TEST_HOST_LOGGING_LEVEL "debug")
 endif()
 
+option(LVI_MITIGATIONS "Enable LVI mitigations" ON)
+if(LVI_MITIGATIONS)
+  set(OE_TARGET_LIBC openenclave::oelibc-lvi-cfg)
+  set(OE_TARGET_ENCLAVE_AND_STD
```
achamayou, line 35: It would have been nice if the list was the same and the LVI path just decorated the paths, but cmake is probably exceptionally awkward for this.

.azure-pipelines-templates/ninja.yml
eddyashton: I hadn't, and looking at them now the `-lvi-cfg` variants are actually consistently _smaller_ than the unmitigated versions, which doesn't sound right.

```
-rw-r--r-- 1 root root  1057412 Jul 21 08:57 liboeenclave-lvi-cfg.a
-rw-r--r-- 1 root root  1363660 Jul 21 08:59 liboeenclave.a
```

Also `liboesnmalloc.a` and `liboesnmalloc-lvi-cfg.a` are exactly the same file. I'll go look at the OE build steps again and see what's going on here.

Do you have a link to that assembly-instruction-counting suggestion?

eddyashton: I can compare the lfence counts in our enclave binaries:

```
edashton@edashton2:/data/src/2.CCF$ objdump -d ../CCF/build/libsmallbank.enclave.so | grep lfence | wc -l
22
edashton@edashton2:/data/src/2.CCF$ objdump -d build.lvi/libsmallbank.enclave.so | grep lfence | wc -l
7424
```

But objdump has a bunch of errors trying to parse the `-lvi-cfg.a` OE libs. I've raised this and the size discrepancy with OE.

cmake/common.cmake
eddyashton: Since the targets are already such general abstractions, I was really hoping that `openenclave::oelibc` could just point at `liboelibc.a` or `liboelibc-lvi-cfg.a` depending on some other setting, but that's not how those targets are set up. I thought this was more readable (and searchable) that appending (sometimes-empty) suffixes everywhere.

.azure-pipelines-templates/ninja.yml
achamayou: I was thinking of something along those lines:

```
$ objdump -d liblogging.virtual.so | grep -c cpuid
129
```

achamayou: For reference: https://reviews.llvm.org/D75939

python/ccf/clients.py
```diff
@@ -54,9 +54,13 @@ class Request:
     headers: dict
 
     def __str__(self):
-        return f"{self.http_verb} {self.path} {self.headers}" + (
-            truncate(f" {self.params}") if self.params is not None else ""
-        )
+        string = f"{self.http_verb} {self.path}"
+        if self.headers:
```
achamayou: Not sure omitting the headers entirely when they're empty is a good thing since we don't name those fields.

python/ccf/clients.py
```diff
@@ -467,9 +471,7 @@ def _direct_call(
     ):
         description = ""
         if self.description:
-            description = f"({self.description})" + (" [signed]" if is_signed else "")
-        LOG.info(f"{self.name} {r} ({description})")
-        return self._response(self.client_impl.request(r, is_signed))
+            description = f"({self.description})" + (" [signed]" if signed else "")
```
achamayou:
```suggestion
            description = f"({self.description})" + " [signed]" * signed
```

python/ccf/clients.py
eddyashton: These formatting changes were in #1429. Think they showed up here briefly due to a merge conflict? Gone now.

Issue #1382, opened 2020-07-06
Package Python infra/utilities


jumaffre:
We should export some of our Python infra for easy consumption of some of our tooling. More specifically:
- **Users** should be able to connect to a CCF service and issue requests (HTTP or websocket) and verify transaction receipts.
- **Members/Operators** should be able to do what **users** can do _as well as_ issue and vote for proposals and parse/verify the ledger.

As such, the following files should be cleaned up and packaged to be release on PyPi: 
- `clients.py` - we should only ship the `request` and `websocket` clients and clean up a few things (e.g. the `result` response field). 
- `ledger.py` - which should support verifying multiple chunks (see https://github.com/microsoft/CCF/issues/1275). 
- `checker.py` to let clients verify that their transactions has been committed. 
- `proposal_generator.py` to let members/operators generate proposals conveniently.

The `ccf` python package should be published on PyPi with, for now at least, the same release cycle as the main C++ framework. 

---

Tasks: 
- [x] Package relevant Python classes. See #1380 
- [x] Cleanup package
- [x] Document package and automatically include in Sphinx
- [x] Automated releases to PyPi

achamayou:
@lynshi your feedback would be very welcome.

lynshi:
These enhancements are great! I think it addresses our issues quite well:
1. `clients.py` reduces code duplication for common procedures, such as signing requests and providing identities.
2. `ledger.py` supporting multiple chunks will make it easier to load files and parse entire ledgers for offline verification.
3. `checker.py` improves the end-to-end story for the user since they won't have to figure out how to verify commits for themselves.
4. `proposal_generator.py` makes it easier for members/operators to submit proposals without learning Lua or reading through the existing test infra. It will also give users a stable API (such as when the scripting language changes).

I'm happy to review associated PRs as well.

PR #1429, opened 2020-07-23
Python clients cleanup and docs

jumaffre:
Relates to #1382

- Cleanup of `ccf.clients`: 
  - `*args` and `**kwargs` removed whenever possible for cleaner interface
  - `binary_dir` argument gone (it was only used for `scurl.sh` and now that we ship `scurl.sh` and put it in the path, we no longer need it.
- Docstring all the things
- Automated docs in Sphinx using `autodoc`
- There is also a new tutorial section that explains how to use `ccf.clients` step by step. It is guaranteed to always be in sync (see new `tutorial.py` which runs as part of every install). 
- Move `checker` back into `tests/infra` as it was overly complicated for what we want. Instead, the `CCFClient` can be directly used to verify that a transaction is committed. 

Next:
- Ledger docs
- Proposal generator docs
- Automated releases to PyPi


test_install.sh
```diff
@@ -0,0 +1,40 @@
...
+timeout --signal=SIGINT --kill-after=30s --preserve-status 30s \
+python $INSTALL_PREFIX/bin/start_network.py \
+    -p ../../build/liblogging \
+    -b $INSTALL_PREFIX/bin \
+    -g $(pwd)/../../src/runtime_config/gov.lua \
+    -v &
+
+# Issue tutorial transactions to ephemeral network
+sleep 20
+python ../../python/tutorial.py client_info.txt
```
eddyashton: Minor: This is only run in the daily CI. I guess we run it for every release, so the released docs are definitely good. Is it worth adding a `ctest` instance that runs this in the normal CI though, for easier repro/validation?

doc/users/python_tuto.rst
```diff
@@ -0,0 +1,83 @@
+Python Client Tutorial
```
eddyashton, line 1: Minor: Why is this `tuto` rather than `python_tutorial.rst`?

jumaffre: Some French-ism for you. Renamed to `python_tutorial.rst` 👍

test_install.sh
jumaffre: I'm tempted to say let's not run it in `ctest` as I don't really want to have to think about/see this script during my daily CCF dev experience? Also, what it tests is definitely already tested by `e2e_logging`.

python/ccf/clients.py
```diff
@@ -36,31 +38,44 @@ def truncate(string, max_len=256):
 
 DEFAULT_CONNECTION_TIMEOUT_SEC = 3
 DEFAULT_REQUEST_TIMEOUT_SEC = 3
+DEFAULT_COMMIT_TIMEOUT_SEC = 3
+
+
+def build_query_string(params):
```
achamayou, line 44: urllib.parse.urlencode() https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode

python/ccf/clients.py
```diff
@@ -369,26 +367,29 @@ def request(self, request, is_signed=False):
...
-        cert=None,
-        key=None,
-        ca=None,
-        request_timeout=DEFAULT_REQUEST_TIMEOUT_SEC,
-        **kwargs,
-    ):
+    """
+    CCF WebSocket client implementation.
+
+    Warning: Does not handle client signatures.
```
achamayou, line 373: Client signatures over WS aren't supported in CCF, so I think this should either be clarified or removed, but as it stands it sounds like it's a limitation of the client.

tests/infra/checker.py
```diff
@@ -2,51 +2,37 @@
...
         self.client = client
         self.notification_queue = notification_queue
         self.notified_commit = 0
 
     # TODO: that API's not right!
     def __call__(self, rpc_result, result=None, error=None, timeout=2):
+        """
+        Lalala.
+
+        :param ccf.clients.Response rpc_result: Hey there.
```
achamayou, line 28: Will fix, sorry :)

python/ccf/clients.py
jumaffre: Good point. I've re-phrased the now-note.

test_install.sh
eddyashton: Fair!

python/ccf/clients.py
```diff
@@ -36,31 +38,37 @@ def truncate(string, max_len=256):
...
         self.path = path
         self.params = params
         self.http_verb = http_verb
         self.headers = headers
-        self.params_in_query = params_in_query
+        self.params_in_query = http_verb == "GET" or http_verb == "DELETE"
+
+    def get_params(self):
+        if self.params_in_query and self.params is not None:
+            return f"?{urllib.parse.urlencode(self.params)}"
```
eddyashton, line 57: We try to parse the values as JSON, so we need `foo="bar"` rather than `foo=bar` (this is why we had `build_query_string`, calling `urlencode` directly won't work). This is a hopefully temporary restriction, but probably not worth tackling in this PR.

tests/test_install.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+set -e
+
+# Setup env
+INSTALL_PREFIX=$(cat /tmp/install_prefix)
```
eddyashton, line 7: Can we pass this as the first CLI arg, rather than reading from file again? Makes this script more useful if we can use to test install directories in arbitrary places, without following the CI's restricted setup step.

doc/users/python_tutorial.rst
```diff
@@ -0,0 +1,83 @@
+Python Client Tutorial
+======================
+
+.. note:: See :ref:`Python Client API <users/python_api:Python Client API>` for complete full API specification.
```
eddyashton, line 4:
```suggestion
.. note:: See :ref:`Python Client API <users/python_api:Python Client API>` for complete API specification.
```

doc/users/python_tutorial.rst
```diff
@@ -0,0 +1,83 @@
...
+
+First, install the CCF Python package:
```
eddyashton, line 6: Meta: We should add a preface to this tutorial. It describes how to use this Python package to build a client, how to talk to a live CCF service with it. Need to make it clear this is _only_ a client. From there, I might still call this `the CCF Python tools package` or similar (even though the package is _named_ CCF), but that's a nit.

doc/users/python_tutorial.rst
```diff
@@ -0,0 +1,83 @@
...
+    :lines: 1
+
+In fact, even an anonymous client can be used to verify that a transaction is committed:
+
+.. literalinclude:: ../../python/tutorial.py
+    :language: py
+    :start-after: SNIPPET: any_client_can_wait
+    :lines: 1
+
+Finally, the authenticated client can used to issue ``GET`` requests and verify that the previous messages have successfully been recorded.
```
eddyashton, line 78:
```suggestion
Finally, the authenticated client can used to issue ``GET`` requests and verify that the previous messages have successfully been recorded:
```

doc/users/python_tutorial.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Set the following CCF node variables:
+
+.. code-block:: python
+
+    host = "<node-rpc-host>"        # Node RPC address or domain
+    port = <node-rpc-port>          # Node RPC port
+    ca = "<path/to/network/cert>"   # Network certificate path
+
+.. note:: :ref:`When deploying a test network <users/deploy_app:Deploying an Application>`, use any node's IP address and port number. All certificates and keys can be found in the associated ``common_dir`` folder.
```
eddyashton, line 27: I think we should specify that host is a string and port is a number. And we should drop the `RPC` terminology if we can. As far as a _client_ is concerned, this is just the node's `IP address or domain`. Unrelated - since we're trying so hard to look like a HTTPS server, we can probably default to port 443 (/no port) in the client?

python/ccf/clients.py
```diff
@@ -179,28 +183,16 @@ def get_curve(ca_file):
 
 class CurlClient:
     """
-    We keep this around in a limited fashion still, because
-    the resulting logs nicely illustrate manual usage in a way using the requests API doesn't
+    Curl client.
+    Note: The resulting logs nicely illustrate manual usage in a way that using other client implementations don't.
```
eddyashton, line 187:
```suggestion
    This client uses Curl to send HTTP requests to CCF, and logs all Curl commands it runs. These commands could also be run manually, or used by another client tool.
```

python/ccf/clients.py
```diff
@@ -480,42 +532,56 @@ def call(self, *args, **kwargs):
                 time.sleep(0.1)
 
     def get(self, *args, **kwargs):
+        """
+        Issue ``GET`` request.
+        Wrapper for :py:meth:`ccf.clients.CCFClient.call` with ``http_verb`` set to ``GET``.
```
eddyashton, line 537: I'd be tempted to just say "Issue GET request. See :py:meth:`ccf.clients.CCFClient.call`"

python/ccf/clients.py
```diff
@@ -480,42 +532,56 @@ def call(self, *args, **kwargs):
...
+        Wrapper for :py:meth:`ccf.clients.CCFClient.call` with ``http_verb`` set to ``DELETE``.
+
+        :return: :py:class:`ccf.clients.Response`
+        """
         return self.call(*args, http_verb="DELETE", **kwargs)
 
+    def wait_for_commit(self, response, timeout=DEFAULT_COMMIT_TIMEOUT_SEC):
+        """
+        Given a :py:class:`ccf.clients.Response`, this functions waits
+        for the associated sequence number and view be committed by the CCF network.
```
eddyashton, line 573:
```suggestion
        for the associated sequence number and view to be committed by the CCF network.
```

python/ccf/clients.py
```diff
@@ -480,42 +532,56 @@ def call(self, *args, **kwargs):
...
-        ca=ca,
-        description=description,
-        binary_dir=binary_dir,
-        connection_timeout=connection_timeout,
-        request_timeout=request_timeout,
-        ws=ws,
-    )
+        The client will poll the ``/node/tx`` endpoint until ``COMMITTED`` is returned.
+
+        :param ccf.clients.Response response: Response returned by :py:class:`ccf.clients.CCFClient.
```
eddyashton, line 577:
```suggestion
        :param ccf.clients.Response response: Response returned by :py:class:`ccf.clients.CCFClient.call`
```

tests/governance.py
```diff
@@ -74,7 +74,7 @@ def test_user(network, args, notifications_queue=None, verify=True):
     network.consortium.remove_user(primary, new_user_id)
     with primary.client(f"user{new_user_id}") as c:
         r = c.get("/app/log/private")
-        assert r.status == 403
+        assert r.status_code == 403
```
eddyashton, line 77:
```suggestion
        assert r.status_code == http.HTTPStatus.FORBIDDEN.value
```

tests/infra/consortium.py
```diff
@@ -348,10 +348,7 @@ def check_for_service(self, remote_node, status, pbft_open=False):
         """
         # When opening the service in PBFT, the first transaction to be
```
eddyashton, line 349: If we're removing the PBFT-specific timeout, I guess this comment should go too?

eddyashton: Oh my bad, we still have it on the `.post` call below! Cool.

tests/infra/consortium.py
```diff
@@ -391,7 +389,7 @@ def _check_node_exists(self, remote_node, node_id, node_status=None):
         with remote_node.client(f"member{self.get_any_active_member().member_id}") as c:
             r = c.post("/gov/read", {"table": "ccf.nodes", "key": node_id})
 
-            if r.status != 200 or (
+            if r.status_code != 200 or (
```
eddyashton, line 392:
```suggestion
            if r.status_code != http.HTTPStatus.OK.value or (
```

tests/infra/logging_app.py
```diff
@@ -162,11 +162,11 @@ def _verify_tx(self, node, idx, priv=True, timeout=5):
         while time.time() < end_time:
             with node.client(self.user) as uc:
                 rep = uc.get(cmd, {"id": idx})
-                if rep.status == 404:
+                if rep.status_code == 404:
```
eddyashton, line 165:
```suggestion
                if rep.status_code == http.HTTPStatus.NOT_FOUND.value:
```

tests/start_network.py
```diff
@@ -91,6 +107,11 @@ def add(parser):
             action="store_true",
             default=False,
         )
+        parser.add_argument(
+            "--client-info-file",
+            help="Path to output file where client information will be dumped to (useful for scripting)",
+            default="client_info.txt",
```
eddyashton, line 113: Minor: I think client-info is a bad name, this is really `node_address` or something? Or maybe `node_info` or `service_info` (given "foo_info", I think an implicit "info about Foo" is more obvious than a "info _for_ Foo").

Anyway this can be default=None, check it looks like a path if provided, and we only write the file if it was specified?

doc/users/python_tutorial.rst
jumaffre: Good point! Done.

doc/users/python_tutorial.rst
jumaffre: Using 443 for the RPC port makes perfect sense. Let me do that in a separate PR as I think `cchost` will need updating as well.

tests/start_network.py
jumaffre: Good point. I have updated the name of the file and changed the default behaviour.

python/tutorial.py
```diff
@@ -35,6 +35,7 @@
 # SNIPPET_START: anonymous_requests
 r = anonymous_client.get("/node/state")
 assert r.status_code == http.HTTPStatus.OK
+# r.body is HTTP response body
```
achamayou, line 38: If we have the right Sphinx, I think the text surround the snippet should instead link to the api doc for the Response type.

src/http/http_endpoint.h
```diff
@@ -320,7 +320,7 @@ namespace http
         }
 
         // On any exception, close the connection.
-        LOG_TRACE_FMT("Closing connection due to exception: {}", e.what());
+        LOG_FAIL_FMT("Closing connection due to exception: {}", e.what());
```
achamayou, line 323: This is good for debugging, but probably not something we want in the final build because it may impact confidentiality.

jumaffre: Done now.

jumaffre:
Merging this PR as the only test failure is unrelated and the fix is well-known (I will push it it a subsequent PR today).

PR #1435, opened 2020-07-23
Python Ledger docs and tutorial

jumaffre:
Related to #1382 (to go in after #1429). 

Adding basic documentation and a short tutorial for our `ledger` Python module.


doc/users/python_tutorial.rst
```diff
@@ -0,0 +1,83 @@
...
+    :lines: 1
+
+The authenticated client can then be used to issue ``POST`` requests, e.g. registering new public and private messages to the default logging application:
+
+.. literalinclude:: ../../python/tutorial.py
+    :language: py
+    :start-after: SNIPPET_START: authenticated_post_requests
+    :end-before: SNIPPET_END: authenticated_post_requests
+
+It is posible to use the same :py:class:`ccf.clients.CCFClient` instance to wait for the transaction to be committed by the network:
```
achamayou, line 64:
```suggestion
It is possible to use the same :py:class:`ccf.clients.CCFClient` instance to wait for the transaction to be committed by the network:
```

doc/users/python_tutorial.rst
```diff
@@ -0,0 +1,83 @@
...
+    :lines: 1
+
+In fact, even an anonymous client can be used to verify that a transaction is committed:
+
+.. literalinclude:: ../../python/tutorial.py
+    :language: py
+    :start-after: SNIPPET: any_client_can_wait
+    :lines: 1
+
+Finally, the authenticated client can used to issue ``GET`` requests and verify that the previous messages have successfully been recorded.
```
achamayou, line 78:
```suggestion
Finally, the authenticated client can be used to issue ``GET`` requests and verify that the previous messages have successfully been recorded.
```

python/ccf/clients.py
```diff
@@ -36,31 +38,37 @@ def truncate(string, max_len=256):
...
-        # TODO: remove
-        if params_in_query is None:
-            params_in_query = http_verb == "GET"
-
         self.path = path
         self.params = params
         self.http_verb = http_verb
         self.headers = headers
-        self.params_in_query = params_in_query
+        self.params_in_query = http_verb == "GET" or http_verb == "DELETE"
```
achamayou, line 53:
```suggestion
        self.params_in_query = http_verb in {"GET", "DELETE"}
```

python/ccf/clients.py
```diff
@@ -36,31 +38,37 @@ def truncate(string, max_len=256):
...
+
+    def get_params(self):
```
achamayou, line 55: I think we should remove all mention of params, and just have an optional body. Users can add params at the end of the path, there's no need for CCFClient to be aware of the details of the URL string.

python/ccf/clients.py
```diff
@@ -76,27 +84,28 @@ def makefile(self, *args, **kwargs):
 
 
 class Response:
-    def __init__(self, status, body, seqno, view, global_commit, headers):
-        self.status = status
+    """
+    Response to request sent via :py:class:`ccf.clients.CCFClient`
+    """
+
+    def __init__(self, status_code, body, seqno, view, global_commit, headers):
```
achamayou, line 91: Response would probably make a nice dataclass

python/ccf/clients.py
```diff
@@ -179,28 +183,16 @@ def get_curve(ca_file):
 
 class CurlClient:
     """
-    We keep this around in a limited fashion still, because
-    the resulting logs nicely illustrate manual usage in a way using the requests API doesn't
+    Curl client.
```
achamayou, line 186: Having a short description repeating the name of the class isn't necessary.

python/ccf/clients.py
eddyashton: This is a good approach. We still have this [awkward encoding requirement](https://github.com/microsoft/CCF/pull/1429#discussion_r459456063), but that only really makes a difference where we have string query params (which I think is only `/api`?), so not onerous to do at the calling point.

achamayou: Yes, but that's a quirk we're hoping to iron out eventually, not something we want the CCFClient interface to reflect or know about.

python/ccf/clients.py
jumaffre: Addressed in #1429

python/ccf/clients.py
jumaffre: Addressed in #1429

python/ccf/clients.py
jumaffre: Addressed in #1429

python/ccf/clients.py
jumaffre: Addressed in #1429

doc/users/python_tutorial.rst
jumaffre: Addressed in #1429

doc/users/python_tutorial.rst
jumaffre: Addressed in #1429

PR #1436, opened 2020-07-24
Try to use cTest XML directly and skip JUnit conversion step

achamayou:
@letmaik kindly pointed out that Azure Pipelines has support for ctest XML now.


achamayou:
This is nearly as good as the existing stuff, and it does let us get rid of the xslt. The only downside is that the bash color codes are being escaped and there's now no hook to get rid of that and make sure the per-test view is readable in the Pipelines UI.

eddyashton:
We can stop installing `xsltproc` as well (`getting_started/setup_vm/roles/ccf_build/vars/common.yml`).

.azure-pipelines-templates/test.yml
```diff
@@ -27,13 +27,10 @@ steps:
   condition: or(failed(), canceled())
   displayName: 'Upload logs (only on failure)'
 
-- script: 'xsltproc --stringparam suiteName "$(Agent.MachineName) ${{ parameters.suite_name_suffix }}" ../tests/infra/ctest_to_junit.xslt Testing/*/Test.xml | sed -r "s/\[NON-XML-CHAR-0x1B\]\[([0-9]*)m//g"> JUnit.xml'
```
eddyashton, line 30: I'd be tempted to keep a preprocessing step where we run sed to strip out the non-xml-chars, so we can still read this in the ADO web view? Non-critical though.

achamayou: Yes, I was hoping to find something in ctest, but I think this is the best solution we have.

PR #1437, opened 2020-07-24
Remove local copy of OE

### LINKED ISSUE #1419: Remove Open Enclave from CCF release package
Now that CCF uses a binary release of Open Enclave, there is no need and no utility to including a build of it in the install package.

eddyashton:
Fixes #1419


achamayou:
We need to modify the app-dev and app-run playbooks, or at least not forget to do it when we we release, to add the openenclave bin_install role, otherwise we'll be missing Open Enclave altogether in our release containers.

eddyashton:
Ah, I'd missed that these weren't already, but that makes sense. When doing this, we could/should also split the LVI-mitigations out of `binary_install.yml`, since `app-run.yml` doesn't need them.

achamayou:
@eddyashton sorry yes, I should have done in the upgrade, but it slipped my mind. Good point about the LVI toolchain, there's no need for that in app-run, I think we can either make it part of ccf_build (which is where we pull in the rest of our toolchain) or give it its own role.

PR #1438, opened 2020-07-24
Debug GH Pages pipeline

achamayou:


achamayou:
Waiting for something else to merge to master, so we get a merge commit picked up via Update etc, since that's possibly what's causing trouble for GH Pages.

achamayou:
Also filed https://github.com/jumaffre/cimetrics/issues/102, because this is quite misleading metrics behaviour.

PR #1439, opened 2020-07-24
Primary endpoint

achamayou:
As requested by @syhamza, for load balancing purposes, this adds a `/node/primary` which will either return 200 if the node is primary, or 308 and attempt to redirect to the primary if possible.

### DIFF
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -341,7 +341,7 @@ class RequestClient:
             "url": f"https://{self.host}:{self.port}{request.path}",
             "auth": auth_value,
             "headers": extra_headers,
-            "allow_redirects": False
+            "allow_redirects": False,
         }
 
         if request.params is not None:
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -341,6 +341,7 @@ class RequestClient:
             "url": f"https://{self.host}:{self.port}{request.path}",
             "auth": auth_value,
             "headers": extra_headers,
+            "allow_redirects": False
         }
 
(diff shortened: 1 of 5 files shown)


syhamza:
> As requested by @syhamza, for load balancing purposes, this adds a `/node/primary` which will either return 200 if the node is primary, or 308 and attempt to redirect to the primary if possible.

Wondering if it should be a temporary redirect or permanent redirect if it's not the primary.

achamayou:
@syhamza yes, I did wonder too, and I do not know. It's only as permanent as until the next election, but it's not temporary in the sense that the node answering may never have been primary, or become primary. There's no expectation of reverting to a permanent state either.

I'm happy to switch to 307, or any other code we think is more appropriate of course.

eddyashton:
FWIW, we currently use `307` (Temporary Redirect) if forwarding fails. I think this is similar, and we should call it Temporary rather than Permanent.

achamayou:
@eddyashton will switch. The lua and js tests failed on this change, interestingly, because of a combination of our session consistency property, the fact that they aggressively mark rpcs as writes (because they're a single handler at the moment, and they can't tell until it's too late), and requests connection pooling. This leads to the primary command being forwarded, with predictably disastrous results!

The right fix is to have a NeverForward variant for the handlers, which ought to be the default for pure commands (no tx), and must be selectable for at least read_only handlers such as primary.

achamayou:
Amusingly, the forwarding fix isn't quite enough either, because requests is helpful and retries client-side unless we tell it not to!

PR #1442, opened 2020-07-27
Fix message leak

achamayou:
Our current thread_messaging queue leaks the Tmsgs that aren't popped, which is a not an issue in normal operations, but does show up in ASAN in tests, when for example a parsing error causes a shutdown with messages still queued.

This is a minimal fix for that, along with a somethat unsatisfying testcase that only serves to make the issue very obvious when ASAN is on:

```

=================================================================
==26692==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x503eb2 in operator new(unsigned long) (/data/src/CCF/build/ds_test+0x503eb2)
    #1 0xbf3607 in std::__1::__unique_if<threading::Tmsg<Nothing> >::__unique_single std::__1::make_unique<threading::Tmsg<Nothing>, void (*)(std::__1::unique_ptr<threading::Tmsg<Nothing>, std::__1::default_delete<threading::Tmsg<Nothing> > >)>(void (*&&)(std::__1::unique_ptr<threading::Tmsg<Nothing>, std::__1::default_delete<threading::Tmsg<Nothing> > >)) /usr/lib/llvm-8/bin/../include/c++/v1/memory:3132:28
    #2 0xbf287a in _DOCTEST_ANON_FUNC_2() /data/src/CCF/build/../src/ds/test/thread_messaging.cpp:31:19
    #3 0x53cd83 in doctest::Context::run() /data/src/CCF/build/../3rdparty/doctest/doctest.h:5863:21
    #4 0x5437e4 in main /data/src/CCF/build/../3rdparty/doctest/doctest.h:5947:71
    #5 0x7fec4babab96 in __libc_start_main /build/glibc-2ORdQG/glibc-2.27/csu/../csu/libc-start.c:310

SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
```

If someone has a good idea on how to do this better, I'm interested. Edit: the test checks we don't leak Payloads now, but still relies on ASAN to check we don't leak Tmsg.


CMakeLists.txt
```diff
@@ -283,6 +283,7 @@ if(BUILD_TESTS)
       ${CMAKE_CURRENT_SOURCE_DIR}/src/ds/test/typed_messages.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/src/ds/test/serializer.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/src/ds/test/hash.cpp
+      ${CMAKE_CURRENT_SOURCE_DIR}/src/ds/test/thread_messaging.cpp
```
eddyashton, line 286: Looks like this is missing from this PR?

achamayou: Uh sorry, yes.

src/ds/test/thread_messaging.cpp
```diff
@@ -0,0 +1,40 @@
...
+  happened = true;
+}
+
+static void never(std::unique_ptr<threading::Tmsg<Nothing>> msg)
+{
+  CHECK(false);
+}
+
+// Note: this only works with ASAN turned on, which catches m2 not being
+// freed.
```
eddyashton, line 23: To test this without ASAN, I think you can create a struct that changes some static state in its destructor, then you can confirm that its destructor is/isn't called? Something like:

```
struct Foo {
  static bool destructed = false;
  ~Foo() { CHECK(!destructed); destructed = true; }
};

{
  threading::ThreadMessaging tm;
  auto m1 = std::make_unique<threading::Tmsg<Foo>>(&always);
  CHECK(!Foo::destructed);
  tm.run_one(task);
  // Running a task destroys the attached payload
  CHECK(Foo::destructed);
}

Foo::destructed = false; // RESET

{
   threading::ThreadMessaging tm;
   // add task but don't run
   CHECK(!Foo::destructed);
}

// Destroying ThreadMessaging destroys contained messages
CHECK(Foo::destructed);
```

src/ds/thread_messaging.h
```diff
@@ -65,6 +65,41 @@ namespace threading
 #endif
     }
 
+    ~Task()
+    {
+#ifdef USE_MPSCQ
+      while (!queue.is_emtpy())
```
eddyashton, line 71:
```suggestion
      while (!queue.is_empty())
```

Worth building this with `USE_MPSCQ` to confirm it still compiles.

src/ds/thread_messaging.h
```diff
@@ -65,6 +65,41 @@ namespace threading
...
+        bool result;
+        std::tie(current, result) = queue.dequeue();
+        if (result)
+        {
+          delete current;
+        }
+      }
+#else
+      bool done = false;
+      while (!done)
```
jumaffre, line 83: `done` never changes value?

src/ds/test/thread_messaging.cpp
achamayou: Yeah that checks that Payload is destructed, but it isn't quite the same as checking TMsg is. Still good though.

src/ds/thread_messaging.h
achamayou: Yes, thank you.

src/ds/thread_messaging.h
achamayou: Yes, I'll have a look.

PR #1443, opened 2020-07-27
Build, publish and use (in containers) .deb

### LINKED ISSUE #698: Install directory should be packaged as .deb
Initially, the install will require cloning the repo and building manually, then running something like `ninja install`. For future releases, we should package the install directory as a .deb.

achamayou:
fix #698

Conversion from .tgz to single .deb:

- Make a .deb rather than a .tar.gz
- Test install that .deb
- Use .deb to make build an run containers, rather than .tgz

Still to do: split ccf from ccf-dev

- ccf:
  - cchost
  - no scripts? everything in ccf pypi package?
  - depends on intel psw, libc++/abi and libuv
- ccf-dev
 - full install, .a and headers
 - still no scripts? everything in ccf pypi package?
 - depends on open-enclave, and many other -dev packages
 - ccf_build role continues to exist with useful utilities that aren't required to build a ccf app but are useful to run end to end tests etc.


cmake/cpack_settings.cmake
```diff
@@ -0,0 +1,18 @@
...
+set(CPACK_PACKAGE_NAME "ccf")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Confidential Consortium Framework")
+set(CPACK_PACKAGE_CONTACT "ccfeng@microsoft.com")
+set(CPACK_RESOURCE_FILE_LICENSE "${CCF_DIR}/LICENSE")
+set(CPACK_PACKAGE_VERSION ${CCF_RELEASE_VERSION})
+set(CPACK_PACKAGING_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
+
+# CPack variables for Debian packages
+set(CPACK_DEBIAN_PACKAGE_DEPENDS
+    "open-enclave (>=0.10.0)"
```
achamayou, line 14: I think we also want those, not sure which base versions yet:

  - libc++abi1
  - libc++1
  - libuv1
  - libcurl4

cmake/cpack_settings.cmake
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+
+include(InstallRequiredSystemLibraries)
+set(CPACK_PACKAGE_NAME "ccf")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Confidential Consortium Framework")
+set(CPACK_PACKAGE_CONTACT "ccfeng@microsoft.com")
```
jumaffre, line 7: This email address is not reachable from outside from Microsoft.

cmake/cpack_settings.cmake
```diff
@@ -0,0 +1,18 @@
...
+set(CPACK_RESOURCE_FILE_LICENSE "${CCF_DIR}/LICENSE")
+set(CPACK_PACKAGE_VERSION ${CCF_RELEASE_VERSION})
+set(CPACK_PACKAGING_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
+
+# CPack variables for Debian packages
+set(CPACK_DEBIAN_PACKAGE_DEPENDS
+    "open-enclave (>=0.10.0), libuv1 (>= 1.18.0), libc++1-8, libc++abi1-8"
```
jumaffre, line 14: As discussed, add `libcurl` here too.

cmake/cpack_settings.cmake
achamayou: Replaced with github

cmake/cpack_settings.cmake
achamayou: done

PR #1444, opened 2020-07-27
Integrate with VS Code debugging

letmaik:
grafik

So instead of copy/pasting a command to the terminal and debugging there, this launches the VS Code debugger directly.

Questions/TODOs:
- It may make sense to replace the fixed list of node numbers with a free-form text input. It's not possible to populate the list dynamically unfortunately.
- I don't know much about the test infra code and just hacked it until it worked. I'm sure there's room for improvement.


jumaffre:
> It may make sense to replace the fixed list of node numbers with a free-form text input. It's not possible to populate the list dynamically unfortunately.

I think that free-form text input is fine too. 

It seems that the format of `network.py` is not quite right. Could you run `black` on it and commit the reformatting? The CI should be all happy then!

PR #1445, opened 2020-07-27
Double app opening

### LINKED ISSUE #1405: On backups, app frontend is opened twice
On a backup, when joining late after a recovery, the app frontend is open twice: 

```
2020-07-15T09:32:09.286188Z -0.008 0   [info ] ../src/node/node_state.h:454         | Node 10 is waiting for votes of members to be trusted
2020-07-15T09:32:13.462034Z -0.202 0   [info ] ../src/node/node_state.h:1081        | Network TLS connections now accepted
2020-07-15T09:32:13.462131Z -0.202 0   [info ] ../src/node/node_state.h:448         | Node has now joined the network as node 10: all domains
2020-07-15T09:32:13.511735Z -0.224 0   [info ] ../src/node/channels.h:375           | node channel with 8 is now established
2020-07-15T09:32:13.534892Z -0.180 0   [info ] ../src/consensus/raft/raft.h:1057    | Becoming follower 10: 5

... << Catching up

2020-07-15T09:32:13.695405Z -0.331 0   [info ] ../src/node/node_state.h:1330        | Network is OPEN, now accepting user transactions

... << More catching up, including going through recovery

2020-07-15T09:32:14.844219Z -1.309 0   [info ] ../src/node/node_state.h:1330        | Network is OPEN, now accepting user transactions
```

This is because the original `open_network` proposal is replayed, followed by the re-opening of the network on recovery.

The opening f the app frontend is already guarded by a check on the node state so that the primary does not open the app frontend twice while recovering the ledger but this is not the case on post-recovery late joiners.

jumaffre:
Resolves #1405 

The application frontend used to be opened twice: on recovery/late join after recovery, the replay of the old service's opening transaction was not sufficiently guarded, which meant the app frontend was opened way too early. 

This should fix the current `full_test_suite` CI issue.


src/node/genesis_gen.h
```diff
@@ -270,11 +270,10 @@ namespace ccf
     }
 
     // Service status should use a state machine, very much like NodeState.
-    void create_service(const tls::Pem& network_cert, kv::Version version = 1)
+    void create_service(const tls::Pem& network_cert)
```
jumaffre, line 273: Removed the unused `version` in the `ccf.service` table.

src/node/node_state.h
```diff
@@ -1327,9 +1326,11 @@ namespace ccf
               "Unexpected: write to service table does not include key 0");
           }
 
+          // Only open service to users if the current service is opened (i.e.
+          // past services openings are ignored)
           if (
             it->second.value().status == ServiceStatus::OPEN &&
-            !this->is_part_of_public_network())
+            this->network.identity->cert == it->second->cert)
```
jumaffre, line 1333: This is the fix.

PR #1446, opened 2020-07-28
Fix full test suite failure

jumaffre:
Fixes `full_test_suite` CI issue where the first request sent to the primary node while a new joiner is catching up hangs for the time it takes for the primary to send all append entries messages. The long-term fix is: https://github.com/microsoft/CCF/issues/1451


tests/infra/network.py
```diff
@@ -509,7 +513,7 @@ def wait_for_state(self, node, state, timeout=3):
     def _get_node_by_id(self, node_id):
         return next((node for node in self.nodes if node.node_id == node_id), None)
 
-    def find_primary(self, timeout=3, request_timeout=3):
+    def find_primary(self, timeout=3):
```
eddyashton, line 516: We're still passing this arg in `election.py`, line 89.

jumaffre: Good spot! This is now using `timeout`.

tests/infra/network.py
```diff
@@ -522,14 +526,18 @@ def find_primary(self, timeout=3, request_timeout=3):
...
                             view = res.body["current_view"]
                             break
                         else:
                             assert "Primary unknown" in res.body, res
+                            LOG.warning("Primary unknown. Retrying...")
                     except CCFConnectionException:
+                        LOG.warning(
+                            f"Could not successful connect to node {node.node_id}. Retrying..."
+                        )
                         pass
```
achamayou, line 541:
```suggestion
```

PR #1447, opened 2020-07-28
Workaround for lifetime mismatch between thread messaging and ringbuffers/enclave

achamayou:
Follow up to #1442, which revealed in some Daily build runs that ownership of Endpoints by the Tasks can cause them to be destroyed after the ringbuffers (which is problematic because the TLSEndpoint destructor uses the ringbuffer to send a session close).

The right long term solution is probably to change the lifetime of thread_messaging so it isn't static and so it's scoped to be shut down before the ringbuffers/while the enclave/node is still effectively in scope. This is a bigger change than we want to squeeze in just before a release though.

### DIFF
--- a/src/enclave/tls_endpoint.h
+++ b/src/enclave/tls_endpoint.h
@@ -220,13 +220,14 @@ namespace enclave
     struct SendRecvMsg
     {
       std::vector<uint8_t> data;
-      std::shared_ptr<Endpoint> self;
+      std::weak_ptr<Endpoint> self;
     };
 
     static void send_raw_cb(std::unique_ptr<threading::Tmsg<SendRecvMsg>> msg)
     {
-      reinterpret_cast<TLSEndpoint*>(msg->data.self.get())
-        ->send_raw_thread(msg->data.data);
+      auto ptr = msg->data.self.lock();
+      reinterpret_cast<TLSEndpoint*>(ptr.get())->send_raw_thread(
+        msg->data.data);
     }
 
     void send_raw(const std::vector<uint8_t>& data)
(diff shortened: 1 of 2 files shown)


src/enclave/tls_endpoint.h
```diff
@@ -220,13 +220,14 @@ namespace enclave
...
+      std::weak_ptr<Endpoint> self;
     };
 
     static void send_raw_cb(std::unique_ptr<threading::Tmsg<SendRecvMsg>> msg)
     {
-      reinterpret_cast<TLSEndpoint*>(msg->data.self.get())
-        ->send_raw_thread(msg->data.data);
+      auto ptr = msg->data.self.lock();
+      reinterpret_cast<TLSEndpoint*>(ptr.get())->send_raw_thread(
+        msg->data.data);
```
eddyashton, line 230:
```suggestion
      auto ptr = msg->data.self.lock();
      if (ptr == nullptr)
      {
        throw std::logic_error("Executing SendRecvMsg on a dead Endpoint");
      }
      reinterpret_cast<TLSEndpoint*>(ptr.get())->send_raw_thread(
        msg->data.data);
```

achamayou: I literally have that locally :)

For some reason though, tls sessions seem to die early, trying to work out why.

eddyashton: Also while we're here - we can remove the `reintepret_cast` if we store `weak_ptr<TLSEndpoint>`. `TLSEndpoint` (and `HTTPEndpoint`) should inherit directly from `std::enable_shared_from_this` (rather than via `Endpoint`).

So
`class TLSEndpoint : public Endpoint, public std::enable_shared_from_this<TLSEndpoint>`
`weak_ptr<TLSEndpoint> self;`
`ptr->send_raw_thread(...`

Issue #1402, opened 2020-07-14
Maintain list of Root CAs and demonstrate their usage from JS Code


achamayou:
The end-user scenario we aim to support is the use of MAA tokens as a mechanism for application authentication and/or authorisation. This ticket specifically covers the provision and maintenance of a root cert authority, which can be used to validate MAA tokens by the application code before it examines the claims they contain.

That MAA root CA is understood to contain an SGX quote. The validity of the root CA is to be decided by validating that quote, and examining its claims. Specifically, comparing MRSIGNER to a stable value, although that logic may change in future revisions.

- [x] obtain sample MAA root cert and confirm that it contains quote
  - The first cert in https://shareduks.uks.attest.azure.net/certs contains a quote under X509 extension OID 1.2.840.113556.10.1.1.
- [ ] confirm whether the collateral is included in the cert.
  - unclear, need to check with OE
- [x] confirm whether the MRSIGNER method is appropriate
  - all claims are exposed, the exact method can be chosen by the governance script.
- [x] confirm the format of the quote
  - The cert is generated with OE's `oe_generate_attestation_certificate` which contains an OE-style report.

The governance mechanism is a straightforward choice to provision and maintain this root CA. It's an existing flow, part of the total order of transactions, auditable to all actors, and offers a choice of mechanisms to validate new root CAs when submitted. In particular, the constitution is an appealing location for the validation to take place, since it isn't expected that a meaningful multi-member vote should take place on root CA renewal.

Several functional elements are missing from CCF to allow the constitution to handle this workflow:

- [x] a root CA table (for now, hard-coded in CCF, later dynamically created)
- [x] a procedure to add/update entries in that table. A Lua function similar to raw_puts, inlined in the constitution is probably the easiest choice (in that case, it would not be part of CCF but of the sample).
- [x] a call that takes a root CA certificate, extracts the quote, validates it and returns a table of claims (including, but not necessarily limited to MRSIGNER) to its Lua caller
- [x] replicate oe_verify_attestation_certificate for MAA's custom format as an intermediate solution

When those are added, an end-to-end sample, run as an automated test will be used to lock in the functionality, and document its usage for applications wishing to make use of it. Using MAA certs may not be practical, in which case we should strive to find a way to create logically equivalent certs embedding quotes, for the purpose of the automated test.

achamayou:
In time, what we really want is #1202, but this is a tactical change that allows the same result for application needs at the moment.

yentsanglee:
[...]

PR #1448, opened 2020-07-28
Add support for proposing attested CA certs

letmaik:
This implements part of https://github.com/microsoft/CCF/issues/1402.

Some notes:
- In the sample `src/runtime_config/gov_root_ca_certs.lua` constitution script I hard-coded an expected mrsigner corresponding to the current MAA cert.
- Since constitution scripts cannot be JS yet, there is also no JS sample as requested in the linked issue.


CMakeLists.txt
```diff
@@ -147,6 +147,8 @@ if("virtual" IN_LIST COMPILE_TARGETS)
            secp256k1.host
            sss.host
            byz.host
+           #openenclave::oehostverify
```
achamayou, line 150: ?

cmake/common.cmake
```diff
@@ -177,7 +177,7 @@ function(add_unit_test name)
   target_compile_options(${name} PRIVATE -stdlib=libc++)
   target_include_directories(${name} PRIVATE src ${CCFCRYPTO_INC})
   enable_coverage(${name})
-  target_link_libraries(${name} PRIVATE ${LINK_LIBCXX} ccfcrypto.host)
+  target_link_libraries(${name} PRIVATE ${LINK_LIBCXX} ccfcrypto.host openenclave::oehost) #openenclave::oehostverify
```
achamayou, line 180: ?

CMakeLists.txt
letmaik: See PR description.

achamayou, line 150:
```suggestion
           #TODO: replace with openenclave::oehostverify once OE 0.11 is released (OE#3312)
```

cmake/common.cmake
achamayou, line 180:
```suggestion
  target_link_libraries(${name} PRIVATE ${LINK_LIBCXX} ccfcrypto.host openenclave::oehost) #TODO: replace with openenclave::oehostverify once OE 0.11 is released (OE#3312)
```

tests/governance_root_ca_certs.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def add(parser):
+        pass
+
+    # temporary hack...
+    for i, arg in enumerate(sys.argv):
+        if 'gov.lua' in arg:
+            sys.argv[i] = arg.replace('gov.lua', 'gov_root_ca_certs.lua')
+
+    args = infra.e2e_args.cli_args(add=add)
```
eddyashton, line 100:
```suggestion
    args = infra.e2e_args.cli_args(add=add)
    
    # temporary hack...
    args.gov_script = "gov_root_ca_certs.lua"
```

To support this properly, I think we should change the `add_e2e_test` function in CMake; add a `CONSTITUTION` arg, remove `-g ${CCF_DIR}/src/runtime_config/gov.lua` from `CCF_NETWORK_TEST_ARGS`, when constructing `add_test` include either `-g ${PARSED_ARGS}_CONSTITUTION}` or default to `gov.lua`.

python/ccf/proposal_generator.py
```diff
@@ -299,6 +301,30 @@ def update_recovery_shares(**kwargs):
 def set_recovery_threshold(threshold, **kwargs):
     return build_proposal("set_recovery_threshold", threshold, **kwargs)
 
+@cli_proposal
+def update_root_ca_cert(cert_name, cert_path, skip_checks=False, **kwargs):
+    cert_pem = open(cert_path).read()
```
achamayou, line 306: I prefer to see fds scoped explicitly.

python/ccf/proposal_generator.py
```diff
@@ -299,6 +301,30 @@ def update_recovery_shares(**kwargs):
...
+
+    if not skip_checks:
+        try:
+            cert = x509.load_pem_x509_certificate(cert_pem.encode(), crypto_backends.default_backend())
+        except:
```
achamayou, line 311:
```suggestion
        except Exception:
```

src/runtime_config/gov_root_ca_certs.lua
```diff
@@ -0,0 +1,126 @@
...
+      end
+    end
+  end
+
+  -- validate update_root_ca_cert calls
+  for _, call in pairs(calls) do
+    if call.func == "update_root_ca_cert" then
+      cert_der = pem_to_der(call.args.cert)
+      claims = verify_cert_and_get_claims(cert_der)
+      expected_mrsigner = "ca9ad7331448980aa28890ce73e433638377f179ab4456b2fe237193193a8d0a"
```
eddyashton, line 55: Could acceptable mrsigners also be stored in a KV table, modified by governance proposals?

src/node/rpc/member_frontend.h
```diff
@@ -20,8 +22,147 @@
...
+      oe_verifier_initialize();
+      oe_result_t res = oe_verify_attestation_certificate_with_evidence(
+        cert_der.data(),
+        cert_der.size(),
+        oe_verify_attestation_certificate_with_evidence_cb,
+        &claims);
+
+      if (res != OE_OK)
+      {
+        // TODO should this raise an exception?
```
achamayou, line 94: As discussed, the right way to handle this is to support proposal validation earlier on (#1458), and this only needs to return false.

src/node/rpc/member_frontend.h
```diff
@@ -20,8 +22,147 @@
...
+        throw std::runtime_error("certificate not valid");
+      }
+
+      lua_newtable(l);
+      const int table_idx = -2;
+
+      for (auto const& item : claims)
+      {
+        LOG_INFO_FMT("claim: {}", item.first);
+        // TODO not all claims should be hex-encoded
```
achamayou, line 104: How do we decide which should and which shouldn't?

src/node/rpc/member_frontend.h
```diff
@@ -20,8 +22,147 @@
...
+      lua_setfield(l, table_idx, "mrsigner");
+
+      std::string mrenclave =
+        fmt::format("{:02x}", fmt::join(identity.unique_id, ""));
+      lua::push_raw(l, mrenclave);
+      lua_setfield(l, table_idx, "mrenclave");
+
+      lua::push_raw(
+        l, static_cast<bool>(identity.attributes & OE_REPORT_ATTRIBUTES_DEBUG));
+      lua_setfield(l, table_idx, "is_debuggable");
```
achamayou, line 156: Is there a way we can generically extract and expose attributes?

src/runtime_config/gov_root_ca_certs.lua
achamayou, line 55: Move up as an all-caps, "static" value, along with an explanatory comment?

tests/governance_root_ca_certs.py
```diff
@@ -0,0 +1,110 @@
...
+import suite.test_requirements as reqs
+import infra.logging_app as app
+import ccf.proposal_generator
+
+from loguru import logger as LOG
+
+this_dir = os.path.dirname(__file__)
+
+@reqs.description("Add certificate with quote, query, update")
+#@reqs.supports_methods("log/private")
```
achamayou, line 26:
```suggestion
```

src/node/rpc/member_frontend.h
letmaik: This is the old API, so no. In the new one (the one you commented on as well) it's already generic.

src/node/rpc/member_frontend.h
letmaik: I don't think there's a way to do it automatically. Probably the right solution is to expose conversion functions like `buf_to_hex(v)`, `buf_to_int(v)`, `buf_to_str(v)` to the gov script and let the developer do it.

src/node/rpc/member_frontend.h
letmaik: I'll change the comment and reference the PR.

src/runtime_config/gov_root_ca_certs.lua
letmaik: I guess so, not sure if there's any advantage to it. @achamayou ?

tests/governance_root_ca_certs.py
letmaik: Should this be a separate PR?

python/ccf/proposal_generator.py
letmaik: Any advantage with this?

python/ccf/proposal_generator.py
letmaik: Alright, just copied from other functions in the same file.

python/ccf/proposal_generator.py
achamayou: Yes, you only want to catch user-defined exceptions, otherwise you will swallow things like SyntaxError or KeyboardInterrupt, which you generally want to let bubble up.

src/runtime_config/gov_root_ca_certs.lua
achamayou: They could be, the only benefit is you could do a more limited proposal to change them rather than change the whole constitution if they need updating. I think it's overkill for now, but we can do it easily enough if/when necessary.

python/ccf/proposal_generator.py
letmaik: SyntaxError is actually derived from Exception, but yeah, KeyboardInterrupt is not. Makes sense.

achamayou: Yes, you're right about SyntaxError, my bad. I think the point still stands.

achamayou:
Let's merge this now but update the parent issue to not forget the TODOs for when Intel SDK quotes support comes to OE.

tests/governance_root_ca_certs.py
```diff
@@ -0,0 +1,107 @@
...
+        cert_ref = x509.load_pem_x509_certificate(cert_pem_str.encode(), crypto_backends.default_backend())
+        cert_kv = x509.load_der_x509_certificate(bytes(r.body), crypto_backends.default_backend())
+        assert cert_ref == cert_kv, f"stored cert not equal to input cert: {cert_ref} != {cert_kv}"
+
+    return network
+
+def run(args):
+    hosts = ["localhost"] * (3 if args.consensus == "pbft" else 2)
+
+    with infra.notification.notification_server(args.notify_server) as notifications:
```
eddyashton, line 73: I think we we can remove all references to `notifications`, `notify_server` etc in this (see `governance_history.py` as a prototype, rather than `governance.py`, which is the only test case that uses/tests this extra stuff.

src/node/rpc/member_frontend.h
letmaik: I think it makes more sense to wait with this after we switch to JS. This disabled code path uses the new OE functions of 0.11, so it will take a while anyway until this gets in and probably requires some iteration.

tests/governance_root_ca_certs.py
```diff
@@ -0,0 +1,86 @@
...
+import suite.test_requirements as reqs
+import ccf.proposal_generator
+
+from loguru import logger as LOG
+
+this_dir = os.path.dirname(__file__)
+
+@reqs.description("Add certificate with quote, query, update")
+def test_cert_store(network, args):
+    # TODO: propose a cert update with valid cert but mismatching mrsigner
```
jumaffre, line 23: ?

letmaik: Yep, it's a TODO :) I wasn't sure how to test this easily. I think I would have to create yet another copy of the gov script with some mismatching MRSIGNER value.

letmaik: Done

tests/governance_root_ca_cert_mismatch.py
```diff
@@ -0,0 +1,53 @@
...
+import suite.test_requirements as reqs
+import ccf.proposal_generator
+
+from loguru import logger as LOG
+
+this_dir = os.path.dirname(__file__)
+
+
+@reqs.description("Add certificate with mismatching MRSIGNER")
+def test_cert_mrsigner_mismatch(network, args):
```
eddyashton, line 23: Can we put this test case in `governance_root_ca_cert.py` rather than its own file? Just add a call to `network = test_cert_mrsigner_mismatch` at the end of `run()`,

src/runtime_config/gov_root_ca_cert_mismatch.lua
```diff
@@ -0,0 +1,126 @@
...
+  -- interface definitions
+  PASSED = 1
+  PENDING = 0
+  REJECTED = -1
+  STATE_ACTIVE = "ACTIVE"
+
+  -- Root CA validation definitions
+  -- This MRSIGNER is intentially random and does not match the actual certificate.
+  -- See the governance_root_ca_cert_mismatch.py test.
+  EXPECTED_MRSIGNER = "abcd"
```
eddyashton, line 18: Its a shame to duplicate this entire file for this change. An ugly but smaller solution would be to grep for this line and replace it in the original script. I think I'd prefer that, but willing to be overruled.

achamayou: Or maybe this is the reason we were looking for to put it in a KV table?

tests/governance_root_ca_cert_mismatch.py
letmaik: It needs to be its own file because it uses a different gov script.

src/runtime_config/gov_root_ca_cert_mismatch.lua
letmaik: I'm happy to put it in a kv table once they can be created dynamically. I don't think it makes sense to hard-code a certain scenario now because the validation logic will depend on the exact claim checks that are done. Here we only compare MRSIGNER but this will likely change.

achamayou: So that's a fair point, and I'm not even sure dynamic tables will be enough because the constitution (or at least a constitution) must be there from the first transaction, before anyone has the opportunity to create any tables.

I think the template + replace (%%%MRSIGNER%%% or so) looks attractive. If the code does the instantiation, that helps merge the test scripts too (each test instantiates it with a different value)?

letmaik:
This PR now works for MAA certs, see the updated description.
@achamayou @eddyashton Could you please leave more comments on the open discussions or mark them resolved?

Note that the tests will currently fail as the MAA cert uses a deprecated RSA key size which will be fixed soon in MAA. I tested locally by patching mbedtls to allow the deprecated size. The PR should only be merged once the MAA cert gets updated.

letmaik:
We should rename this to something more general than Root CA certs since it should just be arbitrary certs with quotes in them and in fact MAA just uses self-signed certs to sign tokens.

achamayou (approved):

achamayou:
@letmaik well they're Token CAs, whether they're root, self-signed or not isn't particularly important in this case. We validate them based on the quote and its claims.

letmaik:
ca_certs?

CMakeLists.txt
```diff
@@ -143,6 +143,9 @@ if("virtual" IN_LIST COMPILE_TARGETS)
            lua.host
            secp256k1.host
            sss.host
+           # TODO: replace with openenclave::oehostverify once OE 0.11 is
```
achamayou, line 146: Could you please open an issue for this? There's a TODO template in issues for this particular type of item.

python/ccf/proposal_generator.py
```diff
@@ -336,6 +338,31 @@ def set_recovery_threshold(threshold: int, **kwargs):
     return build_proposal("set_recovery_threshold", threshold, **kwargs)
 
 
+@cli_proposal
+def update_root_ca_cert(cert_name, cert_path, skip_checks=False, **kwargs):
```
achamayou, line 342: update_ca_cert then?

python/ccf/proposal_generator.py
```diff
@@ -336,6 +338,31 @@ def set_recovery_threshold(threshold: int, **kwargs):
...
+    if not skip_checks:
+        try:
+            cert = x509.load_pem_x509_certificate(
+                cert_pem.encode(), crypto_backends.default_backend()
+            )
+        except Exception as exc:
+            raise ValueError("Cannot parse PEM certificate") from exc
+
+        try:
+            oid = x509.ObjectIdentifier("1.2.840.113556.10.1.1")
```
achamayou, line 355: Like most non-trivial literals, this would benefit from being a named, package-scope variable. The name would document it, and setting it at the package scope means users can overrides if they wish to without having to modify the code.

src/node/maa.h
```diff
@@ -0,0 +1,417 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
achamayou, line 1: Can we have a ticket documenting the temporary existence of this file and the condition that needs to be met for it to be removed (ideally with links to the relevant OE issue(s))?

src/node/rpc/member_frontend.h
```diff
@@ -2,26 +2,112 @@
...
+        cert_der.data(),
+        cert_der.size(),
+        oe_verify_attestation_certificate_with_evidence_cb,
+        &claims);
+
+      if (res != OE_OK)
+      {
+        // Validation should happen before the proposal is registered.
+        // See https://github.com/microsoft/CCF/issues/1458.
+        throw std::runtime_error("certificate not valid");
```
achamayou, line 90:
```suggestion
        throw std::runtime_error(fmt::format("Invalid certificate, verify_maa_root_ca_certificate() returned {}", res));
```

tests/governance_root_ca_cert_mismatch.py
achamayou: args are passed explicitly to network.start_and_join(args), this could be in the same file and run against a network passed different args.

Another possibility would be to reuse the network and update the constitution to the "mismatch" variant before proceeding with the test, in which case what Eddy describes could be done.

achamayou (approved):

CMakeLists.txt
letmaik: Obsolete, as we're not depending on `oe_verify_attestation_certificate_with_evidence` anymore and the issue was fixed in OE master.

src/node/maa.h
letmaik: https://github.com/microsoft/CCF/issues/1546

letmaik:
This PR is "done", though CI will fail until the MAA cert is replaced by one with 2048 bit key (as mbedtls will reject that). Hopefully the re-deployment should only take a few days.

letmaik:
This PR is done now, again, but still blocked due to the 1024 bit issue.

src/node/rpc/member_frontend.h
```diff
@@ -10,18 +11,176 @@
...
+    static int lua_pem_to_der(lua_State* l)
+    {
+      std::string pem = get_var_string_from_args(l);
+      std::vector<uint8_t> der = tls::make_verifier(pem)->der_cert_data();
+      nlohmann::json json = der;
+      lua::push_raw(l, json);
+      return 1;
+    }
+
+#if 0
```
achamayou, line 56:
```suggestion
#if USING_OE_11
```

would  be a little more greppable and descriptive :)

PR #1449, opened 2020-07-28
Python type annotations

### LINKED ISSUE #1433: Use type annotations in Python code
See https://docs.python.org/3/library/typing.html.

This would make it easier for new developers/contributors to navigate the Python tests, and for new users to use the ccf Python package.

jumaffre:
Resolves #1433 

We also check for type correctness in `python/` with `mypy` from `ci-checks.sh`.

### DIFF
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -543,7 +543,7 @@ class CCFClient:
         :return: :py:class:`ccf.clients.Response`
         """
         if self.is_connected:
-            self._direct_call(path, body, http_verb, headers, signed, timeout)
+            return self._direct_call(path, body, http_verb, headers, signed, timeout)
 
         end_time = time.time() + self.connection_timeout
         while True:
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -18,7 +18,7 @@ import base64
 from typing import Union, Optional
 
 import requests
-from loguru import logger as LOG
+from loguru import logger as LOG  # type: ignore
 from requests_http_signature import HTTPSignatureAuth  # type: ignore
 import websocket  # type: ignore
 
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -343,6 +343,7 @@ class RequestClient:
             "allow_redirects": False,
         }
 
+        request_body = None
         if request.body is not None:
             if isinstance(request.body, str) and request.body.startswith("@"):
                 # Request is a file path - read contents, assume json
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass
 from http.client import HTTPResponse
 from io import BytesIO
 from requests.adapters import HTTPAdapter
-from urllib3.util.ssl_ import create_urllib3_context
+from urllib3.util.ssl_ import create_urllib3_context  # type: ignore
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
 import struct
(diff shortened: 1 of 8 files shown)


python/ccf/ledger.py
```diff
@@ -112,7 +124,7 @@ def _read(self):
                 f"Found {read_count} reads, {write_count} writes, and {remove_count} removes"
             )
 
-    def get_tables(self):
+    def get_tables(self) -> dict:
         """
         Returns a dictionnary of all public tables (with their content) in a :py:class:`ccf.ledger.Transaction`.
```
achamayou, line 129:
```suggestion
        Returns a dictionary of all public tables (with their content) in a :py:class:`ccf.ledger.Transaction`.
```

python/ccf/clients.py
```diff
@@ -294,23 +294,39 @@ class RequestClient:
...
+        request: Request,
+        signed: bool = False,
+        timeout: int = DEFAULT_REQUEST_TIMEOUT_SEC,
+    ):
         extra_headers = {}
         extra_headers.update(request.headers)
 
         auth_value = None
         if signed:
+            if self.key is None:
```
achamayou, line 327: We ought to remove the signed flag and pass the key explicitly, and sign with that key when it's provided. That could be different from self.key, legitimately.

python/ccf/clients.py
```diff
@@ -389,6 +426,9 @@ def request(self, request, signed=False, timeout=DEFAULT_REQUEST_TIMEOUT_SEC):
                 )
             except Exception as exc:
                 raise CCFConnectionException from exc
+
+        assert self.ws is not None
```
achamayou, line 430: Is this necessary? Can create_connection() fail but not raise?

python/ccf/commit.py
```diff
@@ -8,21 +8,25 @@
...
+
+    A TimeoutError exception is raised if the commit index is not committed within the given timeout.
     """
     end_time = time.time() + timeout
     while time.time() < end_time:
         r = client.get(f"/node/tx?view={view}&seqno={seqno}")
         assert (
             r.status_code == http.HTTPStatus.OK
         ), f"tx request returned HTTP status {r.status_code}"
+        assert isinstance(r.body, dict), "/node/tx should return a JSON object"
```
achamayou, line 29: I'm not sure that's necessary, if the rc checks out, we should either trust the answer, or do a schema validation, but checking it's a dict seems like a strange validation point.

python/ccf/clients.py
jumaffre: I don't think that CCF supports that - the HTTP signature is verified using the caller's TLS certificate.

python/ccf/clients.py
jumaffre: This seems necessary for `mypy` to be happy. I guess that's because `websocket` doesn't have type annotations. Without it: 
```
$ mypy ../python/
../python/ccf/clients.py:439: error: "None" has no attribute "send_frame"
../python/ccf/clients.py:440: error: "None" has no attribute "recv_frame"
```

python/ccf/commit.py
jumaffre: Again, this is necessary for `mypy` to be happy. As `r.body` can either be a string or a dict, we have to tell it that we're expecting a dict here. Not great indeed but I haven't found a better way unfortunately.

python/ccf/clients.py
achamayou: Looking at the code now, no, but it was definitely the idea that proxying/sending requests on behalf of would work that way.

python/ccf/clients.py
achamayou: That's unfortunate...

python/ccf/commit.py
achamayou: I think we want to be careful how much stuff we do to keep mypy happy vs what we get out of it...

python/ccf/clients.py
eddyashton: We discussed a solution for this, but haven't implemented it. The problem is key identification, finding the key the signer claims was used, that we should verify against. The simplest way is to use the caller's. But we can use the `keyId` in the HTTP signature scheme, set to something like `user0` or `member1`, and then verify proxy'd signed requests.

python/ccf/commit.py
eddyashton: We should maybe prefer `typing.cast` to `assert`s, as I think its clearer that these are _just_ a statement for the type checker?

PR #1456, opened 2020-07-29
PyPi releases

jumaffre:
Automated PyPi releases (this time, to the real PyPi!)

The version number is hardcoded to `0.12` for now. A further PR will address that so that it is derived from the git tag/cmake version.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,7 +1,11 @@
 wheel
 paramiko
+msgpack
 loguru
 coincurve
 psutil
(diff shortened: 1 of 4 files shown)


tests/requirements.txt
```diff
@@ -1,11 +1,7 @@
 wheel
 paramiko
-msgpack
 loguru
 coincurve
 psutil
 cimetrics>=0.2.1
```
achamayou, line 7: There's no actual dependency on cimetrics, is there?

PR #1457, opened 2020-07-29
Require / prefix on URL paths in Python

eddyashton:
While trying to repro a failure in `full_test_suite`, I noticed we've been silently skipping one of the tests for a while:

```
2020-07-29 10:14:06.413 | WARNING  | __main__:run:141 - Test #25:
{
    "name": "election.test_kill_primary",
    "status": "skipped",
    "elapsed (s)": 0.07,
    "reason": "Could not check if test requirements were met: Request client failed with unexpected error"
}
```

The error is trivial, a missing `/` prefix when constructing a requirement-checking request. Now we enforce that all request paths include this prefix. Open to suggestions if anyone sees a neater way to handle requirement failures to ensure these skips are louder in future.

### DIFF
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -491,7 +491,7 @@ class CCFClient:
         """
         Issues one request, synchronously, and returns the response.
 
-        :param str path: URI of the targeted resource.
+        :param str path: URI of the targeted resource. Must begin with '/'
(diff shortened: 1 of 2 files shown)


achamayou:
Ideally the checks would be able to distinguish between exceptional failures and conditions being false.

PR #1467, opened 2020-07-30
Support LVI when building from installed CCF

### LINKED ISSUE #1466: ccf_app.cmake depends on add_lvi_mitigations, but that isn't exposed to apps
ccf_app.cmake makes use of add_lvi_mitigations in wrappers we expose for apps to use, but that's defined in common.cmake, which we do not expect applications to pick up.

eddyashton:
Fixed #1466 

Mostly moving LVI CMake machinery from `common.cmake` to `ccf_app.cmake`. Also installing the LVI cmake files.

While testing I noticed something else:
`oesign` now throws an error if it sees multiple `Debug=X` entries in an `oe_sign.conf` (previously it happily used the latter). This means our old approach of appending `Debug=1` or `Debug=0` for `.debuggable`/`.signed` builds no longer works if apps have an explicit `Debug=X` in their `.conf`. So now the CMake trickery says "sed replace if its already there, else append". Getting this ternary command through CMake's command mangling was surprisingly awkward, so the resulting (formatted) CMake is _ugly_.

### DIFF
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -66,10 +66,17 @@ function(add_lvi_mitigations name)
 endfunction()
 
 if(LVI_MITIGATIONS)
-  install(FILES ${CMAKE_CURRENT_LIST_DIR}/lvi/lvi_mitigation_config.cmake DESTINATION cmake/lvi)
-  install(FILES ${CMAKE_CURRENT_LIST_DIR}/lvi/configure_lvi_mitigation_build.cmake DESTINATION cmake/lvi)
-  install(FILES ${CMAKE_CURRENT_LIST_DIR}/lvi/apply_lvi_mitigation.cmake DESTINATION cmake/lvi)
-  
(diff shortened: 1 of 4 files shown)


cmake/ccf_app.cmake
```diff
@@ -78,8 +127,17 @@ function(sign_app_library name app_oe_conf_path enclave_sign_key_path)
...
     add_custom_command(
       OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/lib${name}.so.signed
-      COMMAND cp ${app_oe_conf_path} ${SIGNED_CONF_NAME}
-      COMMAND echo "Debug=0" >> ${SIGNED_CONF_NAME}
+      COMMAND
+        cp ${app_oe_conf_path} ${SIGNED_CONF_NAME} && (grep
+                                                       "Debug\=.*"
+                                                       ${SIGNED_CONF_NAME}
+                                                       &&
+                                                       (sed -i
```
achamayou, line 135: Yikes!

PR #1474, opened 2020-07-31
Python install cleanup

jumaffre:
Python packaging improvements:
- No longer copy the `python/` folder to the install. Instead, get it from `pip install ccf`.
- Automated versioning number, in sync with main format (using cmake's `configure_file`)
- Pin requirements before pushing to PyPi
- Build a wheel


.azure-pipelines-templates/release.yml
```diff
@@ -31,8 +31,10 @@ jobs:
     - script: |
         python3.7 -m venv env
         source ./env/bin/activate
-        python setup.py sdist
-        pip install twine
+        pip install -r requirements.txt
+        pip freeze > requirements.txt
```
achamayou, line 35: I'm not convinced this is a good idea. For one thing, someone who tries to use ccf in a script that requires a slightly newer psutil is going to have a very unpleasant time.

Can we instead set minimum versions manually?

python/requirements.txt
```diff
@@ -0,0 +1,6 @@
+msgpack==1.0.0
```
achamayou, line 1: Can we make those >=?

Issue #1462, opened 2020-07-29
Host-side fairness in libuv processing


eddyashton:
The jobs in our host-side uv loop are currently greedy, executing all available work before yielding. This leads to some resource hogging and queuing behaviour, which can delay the sending of tick messages.

We should instead have caps on all of these tasks, limiting the amount of work we do each loop and thus bounding the time between ticks. This issue lists the systems where we should introduce this cap, and an initial plan for how the cap should be measured:

- [x] `HandleRingbuffer` shouldn't run a `while` loop, it should only run `max_messages` a single time.
- [x] TCP connections. We should cap the amount we read from each connection, by reducing the allocated size in `on_alloc` in `tcp.h`.

achamayou:
Ideally we'd prefer max_bytes than max_messages, because it's a better proxy for work/memory consumption in enclave. We may not be able to be that precise (could be more like the smallest number of messages that exceeds max_bytes), but it should still be the idea, I think.

PR #1480, opened 2020-08-04
Host fairness with load monitor

eddyashton:
See #1462.

This adds caps to the TCP socket reads and ringbuffer message reads that the host thread executes in each UV loop iteration. Those changes are relatively minor. This also adds some code for monitoring the load on each node, currently by counting the bytes and messages read from the ringbuffer. There's a crude script to get some quick plots of this currently, so we could use it to gain insight on exceptional failures in the CI. If it proves useful we can build more metrics and automated analysis.


src/host/tcp.h
```diff
@@ -594,6 +612,12 @@ namespace asynchost
         return;
       }
 
+      if (sz == UV_ENOBUFS)
+      {
+        LOG_DEBUG_FMT("TCP on_read reached allocation quota");
```
jumaffre, line 617: Minor: `LOG_TRACE_FMT()`?

tests/plot_node_load.py
```diff
@@ -0,0 +1,119 @@
...
+import argparse
+import datetime
+import json
+import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
+import os
+
+# For consistency between plots we want a function from label (name) to colour.
+# We could do this programatically from the hashes to handle general values, but
+# this is sufficient and makes it simple to group similar messages with similar
```
jumaffre, line 12: with similar colours?

src/host/load_monitor.h
```diff
@@ -0,0 +1,100 @@
...
+      enclave_output_file.open("enclave_load.log", std::fstream::out);
+      enclave_counts = nlohmann::json::object();
+
+      // Register message handler for work_stats message from enclave
+      DISPATCHER_SET_MESSAGE_HANDLER(
+        bp, AdminMessage::work_stats, [this](const uint8_t* data, size_t size) {
+          auto [dumped_json] =
+            ringbuffer::read_message<AdminMessage::work_stats>(data, size);
+
+          const auto j = nlohmann::json::parse(dumped_json);
```
jumaffre, line 41: Should we check that this doesn't throw an exception here?

tests/plot_node_load.py
```diff
@@ -0,0 +1,119 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: Could you attach a sample output of this script to this PR?

src/host/tcp.h
eddyashton: I think this should be rare enough that `DEBUG` is fine. There's already a `TRACE` in `on_alloc` with more info. I think if you're in a `DEBUG` but not `TRACE` run, you want to recognise any unusual paths which are taken, which this is.

eddyashton:
## Sample plots

### Enclave
enclave_load_bytes
enclave_load_count

### Host
host_load_bytes
host_load_count

tests/plot_node_load.py
eddyashton: [Done](https://github.com/microsoft/CCF/pull/1480#issuecomment-672850824)

src/host/handle_ring_buffer.h
```diff
@@ -20,7 +20,9 @@ namespace asynchost
   class HandleRingbufferImpl
   {
   private:
-    static constexpr size_t max_messages = 128;
+    // Maximum number of outbound ringbuffer messages which will be processed in
+    // a single iteration
+    static constexpr size_t max_messages = 256;
```
achamayou, line 25: @eddyashton I think we would prefer sizes everywhere.

src/host/main.cpp
```diff
@@ -512,9 +515,15 @@ int main(int argc, char** argv)
...
   });
 
+  // reset the inbound-TCP processing quota each iteration
+  asynchost::ResetTCPReadQuota reset_tcp_quota;
+
   // regularly update the time given to the enclave
   asynchost::TimeUpdater time_updater(1);
 
+  // regularly record some load statistics
+  asynchost::LoadMonitor load_monitor(500, bp);
```
achamayou, line 525: 500ms ideally.

Issue #1478, opened 2020-08-03
Relative import paths in JS modules


letmaik:
QuickJS seems to support relative import paths already. A test case for this should be added, and module names should be restricted to be paths only, currently they can be arbitrary strings. This will enable storing existing module libraries (without external imports) as-is in CCF, like [lodash](https://github.com/lodash/lodash/tree/es).

PR #1481, opened 2020-08-04
Support relative paths in js modules

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1478.

### DIFF
--- a/python/ccf/proposal_generator.py
+++ b/python/ccf/proposal_generator.py
@@ -8,6 +8,7 @@ import json
 import os
 import sys
 import functools
+from pathlib import PurePosixPath
 from typing import Union, Optional, Any
 
 from loguru import logger as LOG  # type: ignore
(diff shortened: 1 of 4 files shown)


python/ccf/proposal_generator.py
```diff
@@ -268,7 +269,12 @@ def set_js_app(app_script_path: str, **kwargs):
 
 @cli_proposal
 def set_module(module_name, module_path, **kwargs):
-    if module_name.endswith(".js"):
+    module_name_ = PurePosixPath(module_name)
+    if not module_name_.is_absolute():
+        raise ValueError("module name must be an absolute path")
+    if any(folder in [".", ".."] for folder in module_name_.parents):
+        raise ValueError("module name must not contain . or .. components")
```
eddyashton, line 276: I may be missing something - doesn't the `is_absolute()` check above forbid this already?

letmaik: Nope

letmaik: `/foo/../bar.js` is still an absolute path.

eddyashton: Right - for some reason I assumed the path construction would normalise.

PR #1482, opened 2020-08-05
Ensure connections_test actually creates and exhausts fds

eddyashton:
Our `connections_test` e2e test is intended to test the behaviour of a CCF node when the file descriptor cap is reached. It does this by setting a low cap, and then creating many connections to the CCF node. However, creating our Python client connection doesn't necessarily claim a fd - this is only done lazily, at the first message on that connection.

This PR updates the test to send logging messages from the create connections, and asserts that once the cap is reached any future connections will receive errors.

### DIFF
--- a/tests/connections.py
+++ b/tests/connections.py
@@ -25,58 +25,77 @@ def run(args):
 
         primary_pid = primary.remote.remote.proc.pid
         num_fds = psutil.Process(primary_pid).num_fds()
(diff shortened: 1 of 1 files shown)


achamayou:
Thanks for catching that, in the days of the FTCP client, we would create a connection at client construction, but this hasn't been true since we switched to requests/curl of course.

PR #1484, opened 2020-08-05
node/npm app test

letmaik:


tests/modules.py
```diff
@@ -108,6 +133,49 @@ def test_module_import(network, args):
...
+    LOG.info("Calling npm app endpoints")
+    with primary.client("user0") as c:
+        body = [1, 2, 3, 4]
+        r = c.post("/app/npm/partition", body)
+        assert r.status_code == 200, r.status_code
+        assert json.loads(r.body) == [[1, 3], [2, 4]], r.body
+
+        r = c.post("/app/npm/pb", body)
+        assert r.status_code == 200, r.status_code
+        assert len(r.body) > 0, r.body
```
eddyashton, line 171: What's this call actually doing? This seems like a very weak assertion, can we add something more precise?

tests/npm-app/src/endpoints.js
```diff
@@ -0,0 +1,31 @@
...
+import protobuf from 'protobufjs/dist/protobuf.js'
+import * as rs  from 'jsrsasign';
+
+export function partition() {
+    // Example from https://lodash.com.
+    let arr = JSON.parse(body);
+    return JSON.stringify(_.partition(arr, n => n % 2));
+}
+
+export function pb() {
```
eddyashton, line 11: Can we call this `protobuf` rather than `pb`? NPM's `pb` is a real and very different thing: https://npm.runkit.com/pb

tests/modules.py
```diff
@@ -108,6 +133,49 @@ def test_module_import(network, args):
     return network
 
 
+@reqs.description("Test Node.js/npm app")
+def test_npm_app(network, args):
+    primary, _ = network.find_nodes()
+
+    LOG.info("Building npm app")
+    app_dir = os.path.join(THIS_DIR, "npm-app")
```
eddyashton, line 141: Could we do this under `build` (or whatever the current directory is when the test is run), rather than in `tests` with additional `.gitignore`? As far as I can see, that should be fine for the `modules` test, and only makes the demo HTML page a little more awkward?

letmaik: npm doesn't support out-of-tree installs, so we would have to copy the project to the build folder first. Should I do that?

tests/modules.py
letmaik: I changed it to be more strong. This endpoint is wrapping the input data in a simple protobuf message with an ad-hoc created schema.

tests/modules.py
eddyashton: After further discussion, we're happy to leave this under `tests/` for now, as this is how npm packages are usually built.

tests/npm-app/src/endpoints.js
```diff
@@ -0,0 +1,35 @@
...
+
+    // CCF doesn't support binary responses yet, so we'll convert the Uint8Array into a hex string.
+    let hex = [...buffer].map(b => ('00' + b.toString(16)).slice(-2)).join("");
+
+    return hex;
+}
+
+export function crypto() {
+    let response;
+    if (rs.KEYUTIL.generateKeypair) {
```
eddyashton, line 29: Are we not calling this because its so slow? If so, please add a comment noting that.

PR #1490, opened 2020-08-12
Add note explaining where to find keygenerator.sh

eddyashton:
Part of our docs were still pointing to a dead script file. I've added a note explaining where these scripts now live (under `python/utils/` rather than `tests/`). Every other reference to these (`scurl.sh`, `keygenerator.sh`, `submit_recovery_keyshare.sh`) uses relative paths so I think they're fine, but maybe this note should be somewhere more central/obvious since it applies everywhere?


doc/users/index.rst
```diff
@@ -5,12 +5,12 @@ To generate the certificate and private key of trusted users should be generated
...
 .. code-block:: bash
 
-    $ CCF/tests/keygenerator.sh --name user1
+    $ keygenerator.sh --name user1
     -- Generating identity private key and certificate for participant "user1"...
     Identity private key generated at:   user1_privk.pem
     Identity certificate generated at:   user1_cert.pem (to be registered in CCF)
 
-.. note:: See :ref:`developers/cryptography:Algorithms and Curves` for the list of supported cryptographic curves.
+.. note:: The ``keygenerator.sh`` script can be found in the :ref:`CCF Python package <users/python_tutorial:Python Client Tutorial>` or in the CCF repo under ``python/utils``. This is a simple wrapper around openssl - CCF will accept key may generated by other tools, as long as they are in the correct format and on a supported cryptographic curve. See :ref:`developers/cryptography:Algorithms and Curves` for the list of supported cryptographic curves.
```
jumaffre, line 13: "can be found in" -> "is installed as part of" maybe, since `keygenerator.sh` & co get installed in the path when `pip install ccf` is run?

PR #1492, opened 2020-08-12
Host memory validation

eddyashton:
We've been missing validation of ECall memory arguments for a long time. This restores those checks.

All arguments which point to a memory range should be checked to ensure they don't overlap with the enclave memory. There's an additional similar threat surface every time we read from the ringbuffer in enclave. I believe we either copy these into the enclave or parse simple types in-place, but haven't done a thorough audit.


douglasmaciver:
After the memory validation, the code must call `oe_lfence()` to protect against speculative threats.

PR #1493, opened 2020-08-12
Add example for proposal generation in Python

lynshi:
The existing example shows `ccf.proposal_generator` used as a command-line tool, and CCF tests use the `ccf.proposal_generator.ProposalGenerator` class which wraps proposal generation functions to write proposals and votes to disks as JSON (as expected by the `CurlClient` client implementation).

To generate proposals to be used with the `RequestClient`, proposal generation functions need to be called directly to get proposals and votes that can be submitted as the `body` argument of the client. This pull request adds such an example to the documentation.


doc/members/proposals.rst
```diff
@@ -67,6 +67,24 @@ Some of these subcommands require additional arguments, such as the node ID or u
...
+
+    proposal, vote = ccf.proposal_generator.open_network()
+
+    # Suppose `ccf_client` is an instance of `CCFClient` with 
+    # the default `RequestClient` client implementation.
+    response = ccf_client.post(
+        "/gov/proposals", body=proposal, signed=True,
+    )
+
+Avoid using the ``ccf.proposal_generator.ProposalGenerator`` class, which wraps the proposal generation functions to write proposals and votes to disk as JSON instead of directly returning dictionaries.
```
eddyashton, line 86: This isn't quite right. The `ProposalGenerator` class writes to file, but still returns the useable proposal. If it writes `proposal_foo.json`, it returns the string `"@proposal_foo.json"`, which both the `RequestClient` and `CurlClient` will treat curl-style (ie - use the contents of the path). This is how our tests generate proposals. The advantage is that the proposals can be inspected after the fact. So if the caller wants to inspect or modify the proposal as a dict, they can use `ccf.proposal_generator`, but in most cases they should also be able to use a `ccf.proposal_generator.ProposalGenerator` instance.

lynshi: That's my fault - I didn't read through `RequestClient` exactly and assumed it wouldn't work with the `@<filename>` input since `requests` calls take a `dict`. I thought the `ProposalGenerator` was just for the `CurlClient`.

lynshi:
@jumaffre I've moved the Python code to `tutorial.py` and also just verified that `tutorial.py` runs against a test network.

jumaffre:
@lynshi LGTM - thanks for updating `tutorial.py`. I have run our daily CI (the one that runs `tutorial.py` against a test network) against your branch and it is all happy: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=11660&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=fe98d859-2537-5570-75af-de67d4290f3a. 

To fix the formatting error, could you run `black` on `tutorial.py`?

lynshi:
@jumaffre ran `black` and updated.

achamayou:
@lynshi mypy flags:

```
python/tutorial.py:109: error: "ProposalGenerator" has no attribute "open_network
python/tutorial.py:118: error: "ProposalGenerator" has no attribute "open_network"
```

lynshi:
@achamayou proposal generation methods on the `ProposalGenerator` class are auto-generated; at runtime the file is inspected for functions annotated with `cli_proposal`, and these functions are wrapped to write to disk and then attached to the class. I think the static type checker is unable to understand this.

How can we resolve this?

achamayou:
@lynshi one possibility may be https://mypy.readthedocs.io/en/stable/extending_mypy.html#current-list-of-plugin-hooks
> get_customize_class_mro_hook() 

But I haven't found good examples for it. The only alternative I can think of is to make the wrapper explicit:

```
write_output_to_disk(open_network(...))
```

eddyashton:
I think we can follow the suggestion at the end of [this](https://mypy.readthedocs.io/en/stable/common_issues.html#spurious-errors-and-locally-silencing-the-checker), ie add

```
    def __getattr__(self, a: str) -> Any:
        return getattr(self, a)
```

to `class ProposalGenerator`?

python/tutorial.py
```diff
@@ -86,3 +91,34 @@
...
+# SNIPPET_END: json_proposal
+
+# SNIPPET_START: json_proposal_with_file
+proposal, vote = generator.open_network(
+    proposal_output_path_="member0_open_network_proposal.json",
+    vote_output_path_="member0_open_network_vote_for.json",
+)
+# >>> proposal
+# '@./member0_open_network_proposal.json'
+# SNIPPET_END: json_proposal_with_file
```
eddyashton, line 124:
```suggestion
# SNIPPET_START: json_proposal_with_file
with open("my_open_network_proposal.json", "w") as f:
    f.write(json.dumps(proposal, indent=2))

# This is still valid though `proposal` is not a `dict`.
response = member_client.post("/gov/proposals", body="@my_open_network_proposal.json", signed=True,)
# SNIPPET_END: json_proposal_with_file
```

doc/members/proposals.rst
```diff
@@ -67,6 +67,37 @@ Some of these subcommands require additional arguments, such as the node ID or u
...
+    :language: py
+    :start-after: SNIPPET_START: json_proposal
+    :end-before: SNIPPET_END: json_proposal
+
+Proposal generation calls from a ``ProposalGenerator`` can optionally include file names specifying where to write output to, which allows you to save different proposals separately.
+
+.. literalinclude:: ../../python/tutorial.py
+    :language: py
+    :start-after: SNIPPET_START: json_proposal_with_file
+    :end-before: SNIPPET_END: json_proposal_with_file
```
eddyashton, line 99:
```suggestion
You may wish to write these proposals to files so they can examined or modified further. These proposal files can be submitted directly - ``CCFClient`` will treat string request bodies beginning with an ``@`` as file paths in the same way that ``curl`` does, and use the content of the file when sending.

.. literalinclude:: ../../python/tutorial.py
    :language: py
    :start-after: SNIPPET_START: json_proposal_with_file
    :end-before: SNIPPET_END: json_proposal_with_file
```

eddyashton (changes requested):
@lynshi After further discussion we've realised `ProposalGenerator` isn't a useful enough wrapper to expose - there's nothing special about the paths it generates, so clients can choose to write their proposals (or any request body) to file themselves. We've removed `ProposalGenerator` in #1513, and suggest the following changes in this PR - we don't call out the `"@file"` support elsewhere, and this still seems a sensible place to do so!

Sorry for the confusion, and thanks again for the PR. Happy to merge after #1513 with these changes included.

lynshi:
@eddyashton no problem! I've updated the files accordingly.

eddyashton (approved):

PR #1494, opened 2020-08-13
lfence after memory validation

eddyashton:
When our `oe_is_within_enclave` checks fail, it is possible for speculative execution to reveal enclave data past the execution point. While we run these checks at initialisation and are unlikely to have any interesting secrets at this time, its best to be sure. So I'm adding lfences to prevent speculative execution over these boundaries.

### DIFF
--- a/src/enclave/main.cpp
+++ b/src/enclave/main.cpp
@@ -71,7 +71,7 @@ extern "C"
     }
 
     // Check that where we expect arguments to be in host-memory, they really
-    // are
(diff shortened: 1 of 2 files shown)


achamayou:
@eddyashton that seems redundant with the Spectre mitigations we get from LLVM. Am I missing something?

I'd rather we switched to pass the struct by deepcopy (which the EDL now supports I believe), and got rid of the usercheck if possible.

Issue #1491, opened 2020-08-12
KV threading issues


jumaffre:
This ticket lists known issues with the key-value store that may hinder our implementation of multi-threading inside the enclave. 

For conciseness, I'm opening this unique ticket with a list of known issues rather than individual tickets for each - we can create such tickets to discuss/fix specific issues when required.

- [x] 1. Map `roll`'s tail is read when serialising the write set of a transaction _without taking a lock on the map_. https://github.com/microsoft/CCF/blob/05e0805db3a8accb63d6467a5a6be33e523e32ac/src/kv/untyped_map.h#L397

- [ ] 2. When deserialising append entries messages, a new entry is appended to the history _non-atomically_ with `apply_views`. In other words, another transaction could write to the history between the state of the KV being updated and the history being updated. Note that this is unlikely to happen in practice (even in a multi-threaded scenario) as transactions are deserialised by the consensus one-by-one. https://github.com/microsoft/CCF/blob/05e0805db3a8accb63d6467a5a6be33e523e32ac/src/kv/store.h#L579

achamayou:
I don't think I understand what the impact of 2. would be, considering that the history is never accessed by read-only transactions (the only kind that could execute on backup). What's the possible user impact?

achamayou:
@jumaffre I don't believe #2 is a problem, because the consensus would not allow a write during a deserialise. So it would only be a read, which wouldn't have a consistent view of the KV and the history. But then that's not a guarantee we make, nor one any read relies on, so I don't think there's an issue here?

jumaffre:
@achamayou Only reading this now. I agree. Closing this issue.

Issue #1300, opened 2020-06-16
Snapshots should restore Merkle Tree


jumaffre:
Snapshots should not only restore the state of the KV store but also the history Merkle Tree.

Issue #1495, opened 2020-08-13
Support KV hooks with snapshots


jumaffre:
To be able to setup its state correctly, a follower that has caught up from a snapshot should be able to execute the relevant KV hooks from that snapshot only. This is especially important for the consensus configuration (which depends on the `ccf.nodes` table) and opening the app frontend (`ccf.service` table).

PR #1496, opened 2020-08-13
Restore Merkle history from KV snapshot

jumaffre:
Relates to #1300 (resolves it functionally but not closing this ticket until the atomicity issue is fixes - see note below).

We now restore the Merkle history from the snapshot when applying it to the target (i.e. backup) store. To do so, we apply to the target history two things: 1) the mini-tree in the signature at which the snapshot was taken (*) and 2) the hash of the write set of the signature at which the snapshot was taken.

The mini-tree (1) can be read by a transaction (see note below) after the snapshot has been applied to the store by reading the `ccf.signatures` table. The hash of the write set of the signature (2) is attached to the snapshot by the source (i.e. primary) store by retrieving it from the history (see new `get_raw_leaf(v)`).

Also:
- Updated the snapshot deserialisation logic so that it resembles the write set deserialisation logic: deserialise the snapshot into a list of `OrderedViews` and apply those atomically via a new `SnapshotTxView` view and associated `SnapshotChangeSet` change set.

Note: The history is not restored atomically with the snapshot being applied to the store, i.e. another signature transaction could be emitted between the time the snapshot is applied and the history is restored, which would break the history restoration process. This requires a bit more fiddling with the interface between the store and the history so I will leave this to a separate PR. Also, this non-atomicity issue is also true when deserialising write sets (see item 2. in https://github.com/microsoft/CCF/issues/1491) but not an actual issue in practice as backups deserialise entries sequentially.

(*) This now means that _if a store has a history_, the snapshot has to be taken at version at which a signature was emitted. If the store has no history, we can still take a snapshot at any version (see unit tests in `kv_snapshot.cpp`).


src/kv/kv_types.h
```diff
@@ -136,6 +136,13 @@ namespace kv
     virtual void append(const uint8_t* replicated, size_t replicated_size) = 0;
     virtual bool verify(Term* term = nullptr) = 0;
     virtual void emit_signature() = 0;
+    virtual crypto::Sha256Hash get_replicated_state_root() = 0;
```
jumaffre, line 139: Unrelated note: we will need to get rid of the dependency on `crypto` at some point.

src/kv/store.h
```diff
@@ -509,27 +537,17 @@ namespace kv
           return DeserialiseSuccess::FAILED;
         }
 
-        // if we are not committing now then use NoVersion to deserialise
+        // If we are not committing now then use NoVersion to deserialise
         // otherwise the view will be considered as having a committed
         // version
         auto deserialise_version = (commit ? v : NoVersion);
-        auto deserialised_write_set =
+        auto deserialised_view =
```
jumaffre, line 544: Small changes here: AIUI, `deserialise()` returns a view and not a write set (rename) and which can never be `nullptr` (removing error path below).

src/kv/store.h
```diff
@@ -694,12 +712,15 @@ namespace kv
 
       {
         std::lock_guard<SpinLock> vguard(version_lock);
-        // This can happen when a transaction started before a view change,
-        // but tries to commit after the view change is complete.
-        LOG_DEBUG_FMT(
-          "Want to commit for term {}, term is {}", txid.term, term);
         if (txid.term != term)
+        {
+          // This can happen when a transaction started before a view change,
```
jumaffre, line 717: This was always displayed. I think we only want to display this message if a transaction fails because the term has changed between the time the opacity version was taken and the transaction is committed (i.e. here).

src/kv/untyped_map.h
```diff
@@ -423,6 +423,70 @@ namespace kv::untyped
...
+
+        auto r = map.roll.commits->get_head();
+
+        r->state = change_set.state;
+        r->version = change_set.version;
+      }
+
+      void post_commit() override
+      {
+        // For now, local hooks with snapshots are not supported
```
jumaffre, line 472: See https://github.com/microsoft/CCF/issues/1495

src/node/node_state.h
```diff
@@ -1521,8 +1521,7 @@ namespace ccf
 
       notifier.set_consensus(consensus);
 
-      // When a node is added, even locally, inform the host so that it can
-      // map the node id to a hostname and service and inform raft so that it
+      // When a node is added, even locally, inform raft so that it
```
jumaffre, line 1524: Left-over from #1371

src/kv/store.h
```diff
@@ -294,30 +302,40 @@ namespace kv
...
+      if (!d.end())
       {
+        LOG_FAIL_FMT("Unexpected content in snapshot at version {}", v);
         return DeserialiseSuccess::FAILED;
       }
 
-      if (!d.end())
+      // Each map is committed at a different version, independently of the
+      // snapshot version
+      auto c = apply_views(views, []() { return NoVersion; });
```
eddyashton, line 334: This seems weird. Why does it work? Aren't we stamping the wrong version into the KV with this?

jumaffre: For normal transactions, this lambda returns a version which is applied to the roll of each map affected by the transaction (typically, the current version + 1 for transactions on the source store, or the specific version of the serialised write set when deserialising transactions on target stores). This version is returned by `apply_views()` and then stamped into the KV. 

For snapshots, the version of the (only) element in the roll after the snapshot has been applied differs for each map (some maps in the source store may have been at different versions). So we get that version from the serialised snapshot for each map rather than a common value, hence ignoring the value returned by this lambda entirely. We could write `[]() { return v; }` here and use the return value `c` to set the value of the store but because we don't need to pass `v` to the commit logic, it seems wasteful/could cause confusion? OTOH, if we merge `deserialise_views()` and `deserialise_snapshot()`, we may end up doing that. 

If you think there's a better way to doing this, I'm happy to change this logic.

src/kv/store.h
```diff
@@ -327,6 +346,17 @@ namespace kv
         last_committable = v;
       }
 
+      if (h)
+      {
+        // For now, this is not atomic as a new transaction is created in
```
achamayou, line 351: I think this is fine for the foreseeable future, there's no clear requirement to make snapshot application atomic at this point.

src/kv/store.h
achamayou: It's a debug/diag, it's not really useful in normal operation.

src/kv/store.h
jumaffre: I've removed the comment.

src/kv/store.h
jumaffre: As discussed, this approach is OK. I've updated the comment above to make it clearer where the version applied to each map comes from.

Issue #1301, opened 2020-06-16
Snapshots should be generated at regular interval


jumaffre:
Just like signature transactions, snapshots should be generated at regular interval (either elapsed time or number of signatures).

The hash of the snapshot should be written to the (new) `ccf.snapshots` table so that the snapshot becomes endorsed by the leader node that has generated it.

Once generated, snapshots should be stored on disk and retrievable for when a new node joins the network.

- [x] 1. Generate snapshots on primary node at regular interval (based on a `cchost` CLI parameter)
- [ ] 2. Store snapshot evidence in new `ccf.snapshot_hash` table 
- [ ] 3. Make snapshot generation asynchronous (snapshot generation takes time and it shouldn't block the thread on which it is started on)

jumaffre:
I wonder how this intersect with the ledger chunking feature. On recovery, we will probably want to only pass the latest snapshot + the ledger suffix (i.e. ledger after the snapshot) to the recovery node. Perhaps a new ledger chunk could be created when a new snapshot is generated?

PR #1500, opened 2020-08-14
Generate and store snapshots on primary node

jumaffre:
Resolves item 1. in #1301 

The primary node now generates snapshots at regular interval, as specified by the `--snapshot-min-tx` `cchost` CLI parameter. The snapshots are stored on disk (in the `--snapshot-dir` folder specific to each node). 

By default, snapshots are never generated (hence the new temporary `reconfiguration_snapshot_test` which sets it to a low value on purpose). We don't do anything with the snapshot just yet, but it will get passed to new joiners by Raft.


src/consensus/ledger_enclave_types.h
```diff
@@ -6,7 +6,7 @@
 
 namespace consensus
 {
-  using Index = uint64_t;
+  using Index = int64_t;
```
achamayou, line 9: Why is that signed now?

src/consensus/raft/test/driver.h
```diff
@@ -16,7 +16,8 @@
 #include "logging_stub.h"
 
 using ms = std::chrono::milliseconds;
-using TRaft = raft::Raft<raft::LedgerStubProxy, raft::ChannelStubProxy>;
+using TRaft = raft::
+  Raft<raft::LedgerStubProxy, raft::ChannelStubProxy, raft::StubSnashotter>;
```
achamayou, line 20:
```suggestion
  Raft<raft::LedgerStubProxy, raft::ChannelStubProxy, raft::StubSnapshotter>;
```

src/consensus/raft/test/driver.h
```diff
@@ -46,6 +47,7 @@ class RaftDriver
         std::make_unique<Adaptor>(kv),
         std::make_unique<raft::LedgerStubProxy>(node_id),
         std::make_shared<raft::ChannelStubProxy>(),
+        std::make_shared<raft::StubSnashotter>(),
```
achamayou, line 50:
```suggestion
        std::make_shared<raft::StubSnapshotter>(),
```

src/consensus/raft/test/logging_stub.h
```diff
@@ -184,4 +189,14 @@ namespace raft
       return kv::DeserialiseSuccess::PASS_SIGNATURE;
     }
   };
+
+  class StubSnashotter
```
achamayou, line 193:
```suggestion
  class StubSnapshotter
```

src/consensus/raft/test/driver.h
jumaffre: D'oh.

src/consensus/raft/raft.h
```diff
@@ -507,12 +515,16 @@ namespace raft
 
       auto& node = nodes.at(to);
 
-      // Record the most recent index we have sent to this node.
-      node.sent_idx = end_idx;
-
       // The host will append log entries to this message when it is
       // sent to the destination node.
-      channels->send_authenticated(ccf::NodeMsgType::consensus_msg, to, ae);
+      if (!channels->send_authenticated(
```
jumaffre, line 520: Small change here: as we can detect whether a message was successfully sent over the channel (i.e. `send_authenticated()` returns `false` if the channel is not yet established), let's not pretend we've sent the idx to the node if the channel is not yet established.

src/consensus/ledger_enclave_types.h
jumaffre: Good point. This wasn't meant to change so I've reverted this and the snapshot now use the unsigned `consensus::Index` instead of the signed `kv::Version`.

src/consensus/raft/raft.h
```diff
@@ -115,6 +116,9 @@ namespace raft
     Index commit_idx;
     TermHistory term_history;
 
+    // Snapshot
```
achamayou, line 119: ?

src/consensus/raft/test/logging_stub.h
putyoutoosleep00: Fuck you

PR #1507, opened 2020-08-18
Initial support for creating KV tables within a transaction

eddyashton:
Opening as a draft for now.

I had a more integrated version, but hit some thorny issues. For now this consists of a few new code paths, tested by `kv_dynamic_tables` unit test, that shouldn't affect the flow of existing transactions/real nodes.

The goal is to allow KV tables to be created after initialisation, during the execution of a transaction. This means that a `Tx`, which was previously just a collection of `TxView`s (each operating over a single `Map`), now operates over an entire `Store`. Rather than

```
Map& map = store.create<Map>("my_map");
...
auto view = tx.get_view(map);
```

we can now do

```
auto view = tx.get_view<Map>("my_map");
```

If `"my_map"` already exists (it was created by a previous transaction, or via the old-style `store.create`) this behaves like it used to. But if this map doesn't create, the `Tx` will instead create the underlying map at this point. This map is owned by the `Tx` while the transaction executes. During commit, if there are no conflicts, ownership of this new map is transferred to the `Store`, after which it (and its state) are visible to other transactions. If the `Store` is rolled back, we will drop (delete) any dynamic `Map`s which were created at a later commit version.

There are a few remaining TODOs for basic functionality:
- [x] Handling map creation on backups when deserialising entries/snapshots
- [x] Changing some of `Map*` to owning pointers, so they don't do dangerous things if their target map is rolled back mid-execution

Please use this PR for discussion. Does the basic approach look sound? Are there any cases you think the test cases are missing?

If we're happy with this, there are a few tasks for future PRs:
- Replace `Tx`s elsewhere in the code base with the 'new style' (ie - constructed with reference to a target `Store`). This will allow these `Tx`s to create `Map`s. I'd like to do this _everywhere_, and get rid of the 'old style' `Tx` (ie - the empty constructor), but we still use this in _a lot_ of places and its not clear they can all be replaced without deeper KV refactoring. If we're convinced these edge cases are read-only and will never created `Map`s, this is probably fine.
- Add some samples/documentation of creating `Map`s in the new style.
- Plumb this view retrieval into `js_generic`, so that `tables.foo` in a JS handler is (potentially) creating a table named "foo".

And maybe longer term work, which may simplify things but don't have an immediate benefit:
- Removing all use of `store.create`, and creating _all_ tables like this. In practice this means most of CCF's service tables will be constructed by the initial `create` RPC, with some additional governance tables seeded by additional governance requests.
- Removing the deprecated KV APIs. If we go with this approach everywhere, we don't need `Store::create`, or `Store::get` (nobody but `Tx` needs to hold references to `Map`s?), `Map`s don't need to know which `Store` they belong to (there's only one, its determined by the `Tx`).


src/kv/tx.h
```diff
@@ -74,13 +80,101 @@ namespace kv
...
+      {
+        // Store doesn't know this map yet - create it dynamically
+        {
+          const auto map_it = created_maps.find(map_name);
+          if (map_it != created_maps.end())
+          {
+            throw std::logic_error("Created map without creating view over it");
+          }
+        }
+        // TODO: Currently assuming all dynamic maps are replicated and private
```
eddyashton, line 131: This assumption is a big one, and I'm not sure how to deal with it. It seems ugly to have to specify these properties every time you ask for a map, but if we're potentially creating the map we need to know them somehow. Perhaps these properties should be rolled into the map type? We assume in C++ you somewhere have a definition `using MyMap = kv::Map<Foo, Bar>`, and that could easily be `using MyMap = kv::Map<Foo, Bar, PRIVATE, REPLICATED>`. That's really just sugar so you can write `auto view = tx.get<MyMap>("my_map")` rather than `auto view = tx.get<MyMap>("my_map", PRIVATE, REPLICATED)` every time. And we need to make this decision again in `js_generic` - does it create a single type of map, or allow these fields to be set.

achamayou:
How does `get_view()` specify whether the table should be public or not?

src/kv/test/kv_dynamic_tables.cpp
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  const auto version_before = kv_store.current_version();
+
+  {
+    INFO("Multiple dynamic tables can be created in a single tx");
+    auto tx = kv_store.create_tx();
+
+    auto [v1, v2] =
+      tx.get_view2<MapTypes::StringString, MapTypes::StringNum>("1", "2");
```
jumaffre, line 86: Nitpick: Could the tables be named "mapX" or something similar as "1" and "2" look more like keys and values.

src/kv/test/kv_dynamic_tables.cpp
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  SUBCASE("Map creation blocked by standard conflict")
+  {
+    constexpr auto key = "foo";
+    auto tx1 = kv_store.create_tx();
+    {
+      auto view1 = tx1.get_view(static_map);
+      const auto v = view1->get(key);
+      if (!v.has_value())
```
jumaffre, line 261: Why this check here? It seems that "foo" has never been `put` in the store so far?

src/kv/test/kv_dynamic_tables.cpp
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    auto tx2 = kv_store.create_tx();
+    {
+      auto view2 = tx2.get_view(static_map);
+      const auto v = view2->get(key);
+      if (!v.has_value())
+      {
+        view2->put(key, "bar");
+        auto dynamic_view = tx2.get_view2<MapTypes::StringNum>(dynamic_map_b);
```
jumaffre, line 276: Asking this here but it is a general question that I will probably answer myself while I go through the rest of the PR: if we create a new dynamic map through `get_view2` but do not `put` anything in that view, does the map still end up being created? If it does, perhaps a good name for `get_view2` would be `create_map` (which also returns a view on the newly created map).

src/kv/tx.h
```diff
@@ -74,13 +80,101 @@ namespace kv
...
+      {
+        // Grab opacity version that all Maps should be queried at.
+        auto txid = store->current_txid();
+        term = txid.term;
+        read_version = txid.version;
+      }
+
+      M* typed_map = nullptr;
+
+      auto type_erased_map = store->get_map(read_version, map_name);
```
jumaffre, line 120: I don't understand why this variable is called `type_erased_map`?

src/kv/tx.h
```diff
@@ -74,13 +80,101 @@ namespace kv
...
+      if (type_erased_map == nullptr)
+      {
+        // Store doesn't know this map yet - create it dynamically
+        {
+          const auto map_it = created_maps.find(map_name);
+          if (map_it != created_maps.end())
+          {
+            throw std::logic_error("Created map without creating view over it");
```
jumaffre, line 128: Does this make `tx.get_view2()` non-idempotent? My understanding is that it would be preferable for it to be idempotent?

src/kv/tx.h
jumaffre: What about we have `tx.create_view<type>(name, domain, replicated)` (basically the existing `get_view2()`, also returning a view over the new map) and a separate `tx.get_view(name)` which only returns a view? Not saying it's a great idea as this would break the "a map is created the first time a view on it is committed" assumption.

src/kv/tx.h
```diff
@@ -74,13 +80,101 @@ namespace kv
...
+          }
+        }
+        // TODO: Currently assuming all dynamic maps are replicated and private
+        const bool replicated = true;
+
+        auto new_map = std::make_shared<M>(
```
jumaffre, line 134: I am sure there's a very good reason but why not `make_unique`?

src/kv/test/kv_dynamic_tables.cpp
jumaffre: A (edge-case) corollary is: if within the same transaction, we create a map, put a key and remove that key, will the table be created on commit?

src/kv/store.h
```diff
@@ -409,6 +472,23 @@ namespace kv
       for (auto& map : maps)
         map.second->rollback(v);
 
+      // TODO: Lock dynamic maps? Or perhaps flatten both into a single
+      // container?
```
jumaffre, line 476: `map->rollback(v)` expects the map to be locked beforehand so I would say yes, definitely.

src/kv/tx.h
jumaffre: Ah! But what about if the type of the map is different between the two calls to `get_view2()`? This should definitely fail in that case.

src/kv/test/kv_dynamic_tables.cpp
jumaffre: `view_containers.h`:
```
// Transfer ownership of these new maps to their target stores, iff we
// have writes to them
```
Answers all my questions!

src/kv/view_containers.h
```diff
@@ -59,20 +73,49 @@ namespace kv
         ok = false;
         break;
       }
+
+      for (const auto& [map_name, map_ptr] : new_maps)
```
jumaffre, line 77: Minor: It feels like this should be done before we look at the conflicts (`prepare()`) between keys within maps?

src/kv/tx.h
eddyashton: So its actually an `AbstractMap`. Trying to distinguish the 3 levels of typing; `AbstractMap`, `kv::untyped::Map`, and `kv::Map<K, V>`. We need one of the latter 2 to call `create_view` on. Anyway I've renamed this in a later change, so we now have variables `abstract_map`, `untyped_map`, `typed_map`.

src/kv/tx.h
eddyashton: Will discuss this in depth later, but the reason I'm wary of adding an explicit `create_view` is that it complicates the programming model. My endpoint implementation just wants to get a view and write to the table - if creation is explicit I either need a _separate_ endpoint that creates the table first (and error handling in my endpoint when the table doesn't exist), or I need a creation path in my endpoint.

As a sketch, I want to do:
```
add_widget = fn(tx, request){
  auto widget_view = tx.get_view<WidgetID, Widget>("widgets");
  widget_view->put(request.widget_id, request.widget);
};
```

and ideally at this level I don't need to know about/care about the table metaproperties (domain, access rights, replicated, etc).

I think separate creation requires something like:
```
create_table = fn(tx, request){
  auto view = tx.create_view(request.table_name, request.table_properties);
  // NB: This creates an empty table?
};

add_widget = fn(tx, request){
  auto widget_view = tx.get_view<WidgetID, Widget>("widgets");
  if (widget_view == nullptr)
    return Error("widgets table does not exist");
  widget_view->put(request.widget_id, request.widget);
};
```

or

```
add_widget = fn(tx, request){
  auto widget_view = tx.get_view<WidgetID, Widget>("widgets");
  if (widget_view == nullptr)
  {
    // In this case, I guess widget_table_properties are specified by the script?
    widget_view = tx.create_view("widgets", widget_table_properties);
  }
  widget_view->put(request.widget_id, request.widget);
};
```

For completeness, another thing we've been considering is something like `"widgets"` is a public unencrypted table, and `"_widgets"` is a private encrypted table. This is neat for now, but doesn't necessarily scale to future table properties. It has the big advantage that its not possible to disagree on these properties in one location from another - they're all included in they key.

src/kv/tx.h
eddyashton: We need to transfer ownership of these maps from this `Tx` to the `Store`, on commit - for that they could be `unique_ptr`. But there's an additional case we need to handle (that we don't currently) - if a transaction is executing over a dynamic map, the map could be rolled back (and deleted) while a `View` over it still exists. Currently this is just a standard pointer - _maybe_ that's fine because we check the rollback counter before accessing the pointer? But to be sure I'd like that `View` to hold a `weak_ptr` or `shared_ptr` to this map, so this map would need to be owned by a `shared_ptr`.

src/kv/view_containers.h
eddyashton: I think we can do these in either order, I don't see a reason to prefer doing one first? Though looking at it in the diff I've just noticed we're doing this in the loop, once-per-view, which is wrong.

src/kv/tx.h
eddyashton: I don't understand the idempotence concern. This check is really just dumb and should never be true, but if it is (as with the error for non-`AbstractTxView` below) we want a loud failure to tell us some expectation is not true.

jumaffre: As discussed, this is fine because if we call `get_view2` twice with the same `"map"` on the same `tx`, the `map` would be in the `view_list` on the second call and we would return early.

src/kv/store.h
```diff
@@ -199,25 +216,68 @@ namespace kv
...
-      auto replicated = true;
-      if (replicate_type == kv::ReplicateType::NONE)
+
+      auto result = std::make_shared<M>(
+        this, name, security_domain, is_map_replicated(name));
+      maps[name] = std::make_pair(NoVersion, result);
+      return *result;
+    }
+
+    std::shared_ptr<AbstractMap> get_map(
```
achamayou, line 228: Can we explicitly make this experimental with a comment (and keep it out of doxygen or mark it as such there too?).

src/kv/store.h
```diff
@@ -241,14 +301,16 @@ namespace kv
       {
         std::lock_guard<SpinLock> mguard(maps_lock);
 
-        for (auto& map : maps)
+        for (auto& it : maps)
```
achamayou, line 304: `for (auto& [_, map] : maps)` should work I would have thought?

src/kv/test/kv_dynamic_tables.cpp
```diff
@@ -0,0 +1,501 @@
...
+    REQUIRE(public_view != private_view);
+
+    const auto pub_v = public_view->get("foo");
+    REQUIRE(pub_v.has_value());
+    REQUIRE(pub_v.value() == "bar");
+
+    const auto priv_v = private_view->get("hello");
+    REQUIRE(priv_v.has_value());
+    REQUIRE(priv_v.value() == "world");
+  }
```
achamayou, line 500: I may be useful to additionally check that dynamic maps happily survive being swapped from one store to another (which happens during recovery, in the public -> private phase).

achamayou (approved):

jumaffre (approved):

src/kv/store.h
eddyashton: No, the value here is a pair, so the `map` is actually `it.second.second`. Ideally we'd have nested structured bindings so I could write something like `auto& [_, [v, map]] : maps`, but this isn't supported yet so we're stuck with 2 lines.

achamayou (approved):

src/kv/store.h
achamayou: Ah right

PR #1510, opened 2020-08-19
Async snapshot generation and evidence

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1301

We now generate snapshots asynchronously on the primary, so that the consensus thread is not immediately blocked when a snapshot needs to be generated (note that with 0 worker thread, the snapshot is still generated/serialised on the consensus thread). The evidence of the snapshot (i.e. hash of serialised snapshot) is now also stored in a new `ccf.snapshot_evidences` table. 

- I've also separated the snapshot generation (_cheap_) from its serialisation (_expensive_) in the KV store. The snapshot is taken by the consensus thread at the time we snapshot but the serialisation only happens later, possibly on a different thread.
- If there are multiple worker threads, we round-robin the generation of snapshots.


src/consensus/pbft/libbyz/client_proxy.h
```diff
@@ -281,7 +281,7 @@ bool ClientProxy<T, C>::send_request(
 
   if (threading::ThreadMessaging::thread_count > 1)
   {
-    threading::ThreadMessaging::thread_messaging.add_task<ExecuteRequestMsg>(
+    threading::ThreadMessaging::thread_messaging.add_task(
```
jumaffre, line 284: Here and throughout the PR, removing this template as the type is deduced from `msg`.

src/consensus/raft/raft.h
```diff
@@ -1170,11 +1170,7 @@ namespace raft
       LOG_DEBUG_FMT("Compacting...");
       if (state == Leader)
       {
-        auto snapshot_idx = snapshotter->snapshot(idx);
-        if (snapshot_idx.has_value())
-        {
-          last_snapshot_idx = snapshot_idx.value();
-        }
+        snapshotter->snapshot(idx);
```
jumaffre, line 1173: Raft doesn't need to know about the last snapshot idx. Instead, it will ask the snapshotter for the latest snapshot idx when a new node is added.

src/kv/store.h
```diff
@@ -263,10 +263,14 @@ namespace kv
...
-      auto e = get_encryptor();
-      KvStoreSerialiser serialiser(e, v, true);
+      return snapshot;
+    }
 
-      return snapshot.serialise(serialiser);
+    std::vector<uint8_t> serialise_snapshot(
+      std::unique_ptr<AbstractSnapshot> snapshot) override
+    {
+      auto e = get_encryptor();
```
achamayou, line 272: Let's check + throw if nullptr.

src/node/entities.h
```diff
@@ -85,6 +85,7 @@ namespace ccf
     static constexpr auto USER_CODE_IDS = "ccf.users.code_ids";
     static constexpr auto CONFIGURATION = "ccf.config";
     static constexpr auto SUBMITTED_SHARES = "ccf.submitted_shares";
+    static constexpr auto SNAPSHOT_EVIDENCES = "ccf.snapshot_evidences";
```
achamayou, line 88: I believe evidence is nearly always uncountable, and so this should be SNAPSHOT_EVIDENCE.

src/kv/store.h
jumaffre: As for append entries, the serialiser still works if the encryptor is `nullptr`, i.e. if there are only public tables and the encryptor is `nullptr`, the serialisation will work. I believe this is useful to avoid boilerplate in many of our unit tests and to keep the KV and encryptor separate.

src/node/entities.h
jumaffre: You're right - I have singular-ised the table name.

src/kv/store.h
achamayou: Fair enough!

achamayou (approved):

PR #1511, opened 2020-08-19
DRAFT: one worker thread

jumaffre:
Let's keep this ticking in the background and see if/when we get a stable build.


achamayou:
This isn't going to produce meaningful results until we have fixed the issue with the hooks.

PR #1514, opened 2020-08-20
AFT scaffolding 

ashamis:
This change allows us to open an AFT network, however most of the work here is just scaffolding that we will use later.

This is currently a draft as I have some open questions here


src/node/rpc/test/tx_status_test.cpp
```diff
@@ -65,12 +65,12 @@ TEST_CASE("edge cases")
...
-    CHECK_THROWS(get_tx_status(3, 10, 0, 3, 10));
-    CHECK_THROWS(get_tx_status(3, 10, 0, 3, 11));
-    CHECK_THROWS(get_tx_status(3, 10, 0, 4, 10));
-    CHECK_THROWS(get_tx_status(3, 10, 0, 4, 11));
+    CHECK_THROWS(get_tx_status(3, 10, -1, 1, 10));
+    CHECK_THROWS(get_tx_status(3, 10, -1, 1, 11));
+    CHECK_THROWS(get_tx_status(3, 10, -1, 3, 10));
+    CHECK_THROWS(get_tx_status(3, 10, -1, 3, 11));
+    CHECK_THROWS(get_tx_status(3, 10, -1, 4, 10));
+    CHECK_THROWS(get_tx_status(3, 10, -1, 4, 11));
```
ashamis, line 73: I am not sure what the correct solution is here, will discuss and fix before pushing

PR #1515, opened 2020-08-20
Make -1 the invalid view

ashamis:


eddyashton:
This definition is just trying to describe what the existing consensus does. In `raft.h` `get_term_internal`, we return 0 for an unknown index.

ashamis:
> This definition is just trying to describe what the existing consensus does. In `raft.h` `get_term_internal`, we return 0 for an unknown index.

I updated the PR to make this dependancy explicit.

src/node/rpc/tx_status.h
```diff
@@ -48,7 +48,7 @@ namespace ccf
      {TxStatus::Committed, tx_status_to_str(TxStatus::Committed)},
      {TxStatus::Invalid, tx_status_to_str(TxStatus::Invalid)}});
 
-  constexpr size_t VIEW_UNKNOWN = 0;
+  constexpr size_t VIEW_UNKNOWN = -1;
```
achamayou, line 51: Let's have the View be signed everywhere including in the tx rpc then.

eddyashton:
I'm all for using the named value in more places, but I don't see why we're changing the invalid value from 0 to -1?

achamayou (approved):

ashamis:
> I'm all for using the named value in more places, but I don't see why we're changing the invalid value from 0 to -1?

We are freeing up 0 to allow that to be a valid view. This is useful when we determine the primary using the view.

PR #1528, opened 2020-08-24
Expose dynamic table creation in js_generic

eddyashton:
Follow up to #1507, exposing dynamic tables to JS apps.


src/apps/js_generic/js_generic.cpp
```diff
@@ -86,16 +91,23 @@ namespace ccfapp
     JS_FreeCString(ctx, k);
 
     if (v.has_value())
+    {
       return JS_NewStringLen(
         ctx, (const char*)v.value().data(), v.value().size());
+    }
     else
+    {
+      // TODO: Should this just return null?
```
achamayou, line 100: Definitely not `null`, if it returns something it should be `undefined`. I think we should stick to an error though, because that's what happens these days when you access a non-existent property in strict mode: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

eddyashton: This isn't accessing a property though, this is an explicit call to `get`, ie - `v = foo.get("key")`. Since we can detect whether a key is defined in C++, it seems we should also be able to in JS? Just replacing `== std::nullopt` with `== undefined` (or `null`, whatever the JS equivalent is).

achamayou: I know, but if we decide to treat a map as closely to an `Object` as possible, which is what's you'd generally use for a map in Javascript, then it probably makes sense to follow the same style.

achamayou: Actually, we should probably consider implementing https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

eddyashton: Created https://github.com/microsoft/CCF/issues/1530 to track this. Renaming to match `Map`'s methods will be a breaking change - I think we should aim to do this in 0.13 alongside the breaking change to how endpoints are defined (`export default function`), so I'll include the renames in this PR.

eddyashton: Actually even in the simplest case we have a fundamental incompatibility - `Map.get` returns a modifiable reference, which we can't emulate. To me, this suggests that trying to mimic a standard JS type is going to be more work for less compatibility than expected, so I'd like to punt on it in this PR - we can discuss in #1530 and revisit if we find a suitable target.

achamayou (approved):

PR #1532, opened 2020-08-25
Join from an existing snapshot

jumaffre:
The PR that puts all the previous snapshot PRs together. 

It is now possible for a node to join a network from an existing snapshot, making the catch-up phase much quicker (basically, only replaying the entries from the last snapshot to the last index).

To do so, the operator(s) of the CCF network are responsible for copying the snapshot file(s) produced by one node (under the `snapshots/` folder) to the directory of the new joiner. The new joiner will automatically try to join from the latest (based on the `snapshot.idx` file name) snapshot, deserialising the snapshot in its store when joining (*). 

Raft will then be forced to become follower at a specific commit index. The snapshotter and the ledger will also have to be initialised at that index so that the ledger and snapshots on the joiner node are the same than those on other nodes in the network.

I've also added a new `snapshot` test suite to test this more thoroughly + all the necessary changes to the Python infra to get the snapshots from one node copied over to the new node.

(*) The snapshot can only be deserialised _after_ the ledger secrets have been passed in by the network so that the snapshot can be decrypted. 

Next:
- More thorough assertions on the content of ledger/snapshot files during end-to-end testing
- Joiner only becomes part of network once it has seen the signed snapshot evidence


CMakeLists.txt
```diff
@@ -571,13 +588,14 @@ if(BUILD_TESTS)
       NAME reconfiguration_test
       PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
       CONSENSUS raft
+      ADDITIONAL_ARGS --raft-election-timeout 4000
```
jumaffre, line 591: Lowering the raft election timeout here (default is 100s) as the reconfiguration test now also waits for a new election to complete.

src/node/node_state.h
```diff
@@ -426,9 +425,44 @@ namespace ccf
...
+              if (!sig.has_value())
+              {
+                throw std::logic_error(
+                  fmt::format("No signatures found after applying snapshot"));
+              }
+
+              auto seqno = network.tables->current_version();
+              consensus->force_become_backup(seqno, sig->view);
+
+              reset_data(config.joining.snapshot);
```
jumaffre, line 455: Once the snapshot has been applied, we don't want to keep it around in enclave memory.

src/consensus/raft/raft.h
```diff
@@ -28,7 +28,7 @@ namespace raft
     std::vector<Index> terms;
 
   public:
-    static constexpr Term InvalidTerm = 0;
+    static constexpr Term InvalidTerm = ccf::VIEW_UNKNOWN;
```
achamayou, line 31: As discussed in #1533, can we use NoVersion everywhere?

src/consensus/raft/raft.h
```diff
@@ -271,6 +275,23 @@ namespace raft
       become_leader();
     }
 
+    void force_become_follower(Index index, Term term)
```
achamayou, line 278:
```suggestion
    void init_as_follower(Index index, Term term)
```

We have too many things called force_become, which imply that we start out as something else and perform an arbitrary transition. Here, we know we initialise from a snapshot, there's nothing forced about that.

src/host/main.cpp
```diff
@@ -634,6 +639,20 @@ int main(int argc, char** argv)
...
+    {
+      ccf_config.joining.snapshot = files::slurp(snapshot_file.value());
+      LOG_INFO_FMT(
+        "Found latest snapshot file: {} (size: {})",
+        snapshot_file.value(),
+        ccf_config.joining.snapshot.size());
+    }
+    else
+    {
+      LOG_INFO_FMT("No snapshot found, node will join from genesis");
```
achamayou, line 654:
```suggestion
      LOG_INFO_FMT("No snapshot found, node will request transactions from the beginning");
```

I suggest we generally drop the term `genesis` everywhere we can. We've made a number of changes to have the first transaction be less special, and that's our direction of travel.

src/host/snapshot.h
```diff
@@ -58,6 +59,33 @@ namespace asynchost
...
+      std::optional<std::string> snapshot_file = std::nullopt;
+      size_t latest_idx = 0;
+
+      for (auto& f : fs::directory_iterator(snapshot_dir))
+      {
+        auto file_name = f.path().filename().string();
+        auto pos = file_name.find(fmt::format("{}.", snapshot_file_prefix));
+        if (pos == std::string::npos)
+        {
+          LOG_FAIL_FMT("File {} does not appear to be a snapshot", file_name);
```
achamayou, line 73:
```suggestion
          LOG_FAIL_FMT("Ignoring \"{}\" because it does not start with {}", file_name, snapshot_file_prefix);
```

Is more explicit and actionable.

src/node/node_state.h
```diff
@@ -426,9 +425,44 @@ namespace ccf
...
+              LOG_INFO_FMT(
+                "Joiner successfully resumed from snapshot at seqno {} and "
```
achamayou, line 457: This sounds a slightly more final than it really is: we've started resuming from this snapshot, but until we see the evidence from the primary, we don't know if the snapshot is valid.

src/node/snapshotter.h
```diff
@@ -108,6 +108,17 @@ namespace ccf
       next_snapshot_indices.push_back(last_snapshot_idx);
     }
 
+    void set_last_snapshot_idx(consensus::Index idx)
+    {
+      // Warning: Should only be called once, after a snapshot has been applied
```
achamayou, line 113: Can we guard on last_snapshot_idx being set then?

tests/infra/network.py
```diff
@@ -173,12 +181,23 @@ def _add_node(self, node, lib_name, args, target_node=None, recovery=False):
                 timeout=args.ledger_recovery_timeout if recovery else 3
             )
 
+        snapshot_dir = None
+        if from_snapshot:
+            LOG.info("Joining from snapshot")
+            snapshot_dir = target_node.get_snapshots()
+            # For now, we must have a snapshot to resume from when attempting
```
achamayou, line 188: That's for now and for the foreseeable future.

achamayou (approved):
LGTM with some minor comments.

Zenix27 (approved):

tests/infra/network.py
jumaffre: I think that eventually, joining from a snapshot will become the default when joining a new node in our Python infra. If the `target_node` has no snapshot, then we'll join from append entries since the beginning of time.

src/node/node_state.h
jumaffre: You're right. That's the next step for this feature so I will address that in the next PR.

src/consensus/raft/raft.h
jumaffre: Let's discuss this on #1533

PR #1533, opened 2020-08-26
Updated websocket client invalid view

ashamis:


python/ccf/clients.py
```diff
@@ -174,6 +174,13 @@ def get_curve(ca_file):
     )
 
 
+def unpack_seqno_or_view(data):
+    (value,) = struct.unpack("<q", data)
+    if value == -1:
```
jumaffre, line 179: `seqno` default is `kv::NoVersion` (`std::numeric_limits<Version>::min()`) while `view` default is `-1`. I think each of those should have a different path for these edge-case values.

achamayou: Why not make both -1?

python/ccf/clients.py
```diff
@@ -174,6 +174,13 @@ def get_curve(ca_file):
     )
 
 
+def unpack_seqno_or_view(data):
+    (value,) = struct.unpack("<q", data)
```
achamayou, line 178:
```suggestion
    value, = struct.unpack("<q", data)
```

tests/infra/checker.py
```diff
@@ -34,7 +34,7 @@ def __call__(self, rpc_result, result=None, error=None, timeout=2):
                     result, rpc_result.body
                 )
 
-            assert rpc_result.seqno and rpc_result.view, rpc_result
+            assert rpc_result.seqno and rpc_result.view >= 0, rpc_result
```
achamayou, line 37:
```suggestion
            assert rpc_result.seqno >= 0 and rpc_result.view >= 0, rpc_result
```

python/ccf/clients.py
ashamis: The braces are enforced by black

python/ccf/clients.py
ashamis: I changed kv::NoVersion to be -1

python/ccf/clients.py
achamayou: Ah, yes!

achamayou (approved):

python/ccf/clients.py
achamayou: As pointed by @jumaffre separately, we can't do that because we do use negative versions internally, so we should use `std::numeric_limits<Version>::min()` for both.

src/node/rpc/test/tx_status_test.cpp
```diff
@@ -65,12 +65,18 @@ TEST_CASE("edge cases")
...
     // Impossible: view for all global txs must be known
     // get_tx_status(a, N, 0, b, >=N)
-    CHECK_THROWS(get_tx_status(3, 10, -1, 1, 10));
-    CHECK_THROWS(get_tx_status(3, 10, -1, 1, 11));
-    CHECK_THROWS(get_tx_status(3, 10, -1, 3, 10));
-    CHECK_THROWS(get_tx_status(3, 10, -1, 3, 11));
-    CHECK_THROWS(get_tx_status(3, 10, -1, 4, 10));
-    CHECK_THROWS(get_tx_status(3, 10, -1, 4, 11));
+    CHECK_THROWS(
+      get_tx_status(3, 10, std::numeric_limits<int64_t>::min(), 1, 10));
```
jumaffre, line 69: Minor: Use `VIEW_UNKNOWN` instead of `std::numeric_limits<int64_t>::min()`?

PR #1535, opened 2020-08-26
Update black

jumaffre:
A new version of `black` [was just released](https://github.com/psf/black/releases).


achamayou:
@jumaffre thanks for taking care of this!

Issue #1497, opened 2020-08-13
IV reuse when encrypting snapshot


jumaffre:
When a snapshot is generated, the integrity/encryption of the snapshot is done with the encryptor, using the same code path than for write sets. The IV used to encrypt the snapshot depends on the node ID and the unique version number at which the snapshot is taken. However, if we generated and encrypted a write set a version `V`, a snapshot generated at `V` will use the same IV (same node ID and same version number). 

We could use one bit of the IV (we use 4 bytes for the node ID/view number, which is more than enough) to indicate whether the content to encrypt is a snapshot or not.

PR #1536, opened 2020-08-27
IV re-use with snapshots

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1497

I lied. The IV would not get re-used when encrypting a snapshot as we use a ever-growing sequence number when encrypting with the encryptor (so that we never re-use the same IV across rollbacks). I've still added a fix for it as the plain `kv` encryptor (which was used for PBFT) would indeed re-use the same IV when generating a snapshot.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -577,7 +577,7 @@ if(BUILD_TESTS)
       NAME reconfiguration_snapshot_test
       PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
       CONSENSUS raft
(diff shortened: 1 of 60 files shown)


achamayou:
This is fine, but I'd rather see the encryptor disappear altogether if we don't use it, it is effectively dead code now.

jumaffre:
> This is fine, but I'd rather see the encryptor disappear altogether if we don't use it, it is effectively dead code now.

Agreed. I'll ticket this separately as removing the mixin logic for both Raft and PBFT encryptors is not trivial.

achamayou (approved):

PR #1537, opened 2020-08-27
Allow setting Subject Name and Subject Alternative Name

### LINKED ISSUE #1257: Record both domain and IP in node certificate SAN
We currently only support either the node's RPC IP address or domain (if `--domain` is passed to `cchost`) in the Subject Alternative Name of the node's x509 certificate. 

It might be useful to support both IP and DNS in there so that an operator can connect to a node _securely_ without relying on the DNS/load balancer to to be up and running.

achamayou:
fix #1257

### DIFF
--- a/src/ds/cli_helper.h
+++ b/src/ds/cli_helper.h
@@ -2,7 +2,7 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
-#include "../tls/san.h"
+#include "tls/san.h"
 
 #include <CLI11/CLI11.hpp>
 
--- a/src/ds/cli_helper.h
+++ b/src/ds/cli_helper.h
@@ -2,7 +2,12 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+#include "../tls/san.h"
+
 #include <CLI11/CLI11.hpp>
+
+#define FMT_HEADER_ONLY
+#include <fmt/format.h>
 
 namespace cli
 {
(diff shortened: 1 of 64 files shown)


src/host/main.cpp
```diff
@@ -297,6 +297,16 @@ int main(int argc, char** argv)
   app.add_option(
     "--domain", domain, "DNS to use for TLS certificate validation");
 
+  std::string subject_name("CN=CCF Node");
+  app.add_option("--sn", subject_name, "Subject Name in node certificate, eg. CN=CCF Node")
+    ->capture_default_str();
+
+  std::vector<std::string> subject_alternative_names;
+  app.add_option(
```
jumaffre, line 305: Should we use a special parser here, similarly to the address options, to check that the values are in the expected format?

achamayou: I hadn't thought of that, but absolutely, yes!

src/ds/cli_helper.h
```diff
@@ -2,8 +2,13 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+#include "../tls/san.h"
```
jumaffre, line 5: Minor: `#include "tls/san.h`

src/node/node_state.h
```diff
@@ -67,6 +67,9 @@ namespace ccf
   using RaftType =
     raft::Raft<consensus::LedgerEnclave, NodeToNode, Snapshotter>;
 
+  static const std::string IP_ADDRESS_PREFIX("iPAddress:");
```
jumaffre, line 70: Duplicated in `cli_helper.h`.

achamayou: Yes, forgot to remove it when it moved, thanks for spotting it.

src/ds/cli_helper.h
achamayou: Thank you.

jumaffre (approved):

PR #1542, opened 2020-08-28
Fix ci-checks scripts

letmaik:
`-f` was broken for clang-format.
This also adds `-U` to `pip install` to avoid using old dependency versions and potentially formatting/linting files differently than in CI.

### DIFF
--- a/scripts/check-format.sh
+++ b/scripts/check-format.sh
@@ -34,11 +34,11 @@ badly_named_files=""
 for file in $(find "$@" -name "*.h" -or -name "*.hpp" -or -name "*.cpp" -or -name "*.c"); do
   # Workaround for https://bugs.llvm.org/show_bug.cgi?id=39216
   d=$(cat "$file" | clang-format-8 -style=file --assume-filename "${file%.*}".cpp | diff "$file" -)
-  if $fix ; then
-    cat "$file" | clang-format-8 -style=file --assume-filename "${file%.*}".cpp > "$file".tmp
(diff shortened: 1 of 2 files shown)


achamayou (approved):

achamayou:
Thank you @letmaik!

scripts/check-format.sh
```diff
@@ -34,11 +34,11 @@ badly_named_files=""
 for file in $(find "$@" -name "*.h" -or -name "*.hpp" -or -name "*.cpp" -or -name "*.c"); do
   # Workaround for https://bugs.llvm.org/show_bug.cgi?id=39216
   d=$(cat "$file" | clang-format-8 -style=file --assume-filename "${file%.*}".cpp | diff "$file" -)
-  if $fix ; then
-    cat "$file" | clang-format-8 -style=file --assume-filename "${file%.*}".cpp > "$file".tmp
-    mv "$file".tmp "$file"
-  fi
   if [ "$d" != "" ]; then
+    if $fix ; then
+      cat "$file" | clang-format-8 -style=file --assume-filename "${file%.*}".cpp > "$file".tmp
```
achamayou, line 39: The bug entry on line 35 seems like it has been fixed in llvm 9, we should give that a try and see if we can't simplify this code quite a bit as a result.

letmaik: Another PR?

achamayou: Yes, absolutely.

Issue #1303, opened 2020-06-16
Snapshots should be able to be used on recovery


jumaffre:
It should be possible to recover from a given snapshots and ledger suffix on recovery.

jumaffre:
Specifically, we leave it to the operator(s) to make sure that a joiner joining a network from a node that started/joined from a snapshot also starts from a snapshot which is more recent than the first entry that the joinee knows about. This is because a node that started/joined from a snapshot doesn't know about the ledger entries prior to that snapshot and thus can't serve these entries to any new joiner.

PR #1548, opened 2020-08-28
Raft receive append entries tweaks

jumaffre:
Small tweaks to the Raft's receive append entries logic after trying to implement https://github.com/microsoft/CCF/issues/1303 

- Check that `r.prev_idx` is not in the future
- Move leader assignment earlier, when we restart the election timeout
- Do not throw if deserialision fails
- Comments


src/consensus/raft/raft.h
```diff
@@ -635,8 +637,17 @@ namespace raft
         return;
       }
 
+      // If the terms match up, it is sufficient to convince us that the sender
+      // (leader) is legit in our term
```
achamayou, line 641:
```suggestion
      // (leader) is leader in our term
```

achamayou (approved):

src/consensus/raft/raft.h
```diff
@@ -624,7 +626,7 @@ namespace raft
         {
           LOG_DEBUG_FMT(
             "Recv append entries to {} from {} but our log at {} has the wrong "
-            "term (ours: {}, theirs: {})",
+            "previous term (ours: {}, theirs: {})",
```
ashamis, line 629: nit: view

src/consensus/raft/raft.h
```diff
@@ -635,8 +637,17 @@ namespace raft
         return;
       }
 
+      // If the terms match up, it is sufficient to convince us that the sender
```
ashamis, line 640: term => view

src/consensus/raft/raft.h
jumaffre: `raft.h` still uses the Raft-specific term-inology (no pun intended) for both comments and logs. I think we should address this separately if we want to? The `raft_consensus.h` adapter provides the translation between both.

PR #1552, opened 2020-09-01
Extend asn1_san to support multiple entries

achamayou:
`x509write_crt_set_subject_alt_name()` does not support adding multiple subject alternative names to a cert, so this adds a `x509write_crt_set_subject_alt_names()` which does.

### DIFF
--- a/src/tls/key_pair.h
+++ b/src/tls/key_pair.h
@@ -670,15 +670,12 @@ namespace tls
       // Because mbedtls does not support parsing x509v3 extensions from a
       // CSR (https://github.com/ARMmbed/mbedtls/issues/2912), the CA sets the
       // SAN directly instead of reading it from the CSR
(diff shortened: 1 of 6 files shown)


src/tls/key_pair.h
```diff
@@ -670,15 +670,12 @@ namespace tls
       // Because mbedtls does not support parsing x509v3 extensions from a
       // CSR (https://github.com/ARMmbed/mbedtls/issues/2912), the CA sets the
       // SAN directly instead of reading it from the CSR
-      for (auto& subject_alt_name : subject_alt_names)
+      auto rc =
+        x509write_crt_set_subject_alt_names(&sign.crt, subject_alt_names);
```
ashamis, line 674: put a try/catch around this? We should not crash if we are given a malformed cert

ashamis (approved):

PR #1557, opened 2020-09-02
Update collections of JS modules in one proposal

### LINKED ISSUE #1479: Update collections of JS modules in one proposal
Currently, the Python client supports `set_module` and `remove_module` proposals which set and remove a single module, respectively. This is fine for simple cases but users may want to update one or more folders/prefixes in one go, for example libraries made up of multiple files. The Python client should be extended to support building such proposals based on the existing `set_module` and `remove_module` calls.

This new proposal could look something like
```lua
tables = ...
calls = []
# remove existing modules
for module_name, _ in tables["ccf.modules"]:
  if module_name.startswith("foo/") or module_name.startswith("bar/") then
    calls.append(("remove_module", module_name))
  end
end
# add new modules
calls.append(("set_module", "foo/a.js", "..."))
calls.append(("set_module", "foo/b.js", "..."))
calls.append(("set_module", "bar/a.js", "..."))
return calls
```

Maybe this could be called `update_modules`, accepting a set of folder-to-prefix mappings on the command line:
```sh
# update a single external library
python -m ccf.proposal_generator update_modules ./app1/libs/cryptlib=/app1/libs/cryptlib
# update the whole application
python -m ccf.proposal_generator update_modules ./app1=/app1
```

The proposal generator would then scan the given folders and include all *.js files.

In the future it may make sense to treat this as a lower-level operation where the high-level operation is something like `update_app` which also deals with updating the endpoint metadata.

@eddyashton @achamayou @ad-l

letmaik:
Fixes #1479.

### DIFF
--- a/python/ccf/proposal_generator.py
+++ b/python/ccf/proposal_generator.py
@@ -313,7 +313,7 @@ def update_modules(module_name_prefix: str, modules_path: Optional[str], **kwarg
     if modules_path:
         for path in glob.glob(f"{modules_path}/**/*.js", recursive=True):
             rel_module_name = os.path.relpath(path, modules_path)
-            rel_module_name = rel_module_name.replace('\\', '/') # Windows support
+            rel_module_name = rel_module_name.replace("\\", "/")  # Windows support
             with open(path) as f:
                 js = f.read()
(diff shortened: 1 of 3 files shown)


achamayou (approved):
Thank you @letmaik!

PR #1562, opened 2020-09-03
Remove --no-undefined in SAN builds

eddyashton:
This option doesn't play nicely with the SAN builds. I got a lucky combination on a local build that gave a huge amount of errors:

```
/data/src/CCF/build.san/../src/enclave/thread_local.cpp:(.text.startup+0x2d): undefined reference to `__asan_before_dynamic_init'
/data/src/CCF/build.san/../src/enclave/thread_local.cpp:(.text.startup+0x34): undefined reference to `__asan_option_detect_stack_use_after_return'
/data/src/CCF/build.san/../src/enclave/thread_local.cpp:(.text.startup+0x43): undefined reference to `__asan_stack_malloc_0'
/data/src/CCF/build.san/../src/enclave/thread_local.cpp:(.text.startup+0x10e): undefined reference to `__asan_after_dynamic_init'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

It seems to mostly cause builds to deadlock/time-out, as it has been in our Daily CI. No idea why.

### DIFF
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -257,8 +257,6 @@ function(add_ccf_app name)
       ${virt_name} PRIVATE ${PARSED_ARGS_LINK_LIBS_VIRTUAL} ccf.virtual
     )
 
-    target_link_options(${virt_name} PRIVATE LINKER:--no-undefined)
-
     set_property(TARGET ${virt_name} PROPERTY POSITION_INDEPENDENT_CODE ON)
 
     enable_coverage(${virt_name})
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -257,6 +257,10 @@ function(add_ccf_app name)
       ${virt_name} PRIVATE ${PARSED_ARGS_LINK_LIBS_VIRTUAL} ccf.virtual
     )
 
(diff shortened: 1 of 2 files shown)


cmake/ccf_app.cmake
```diff
@@ -257,8 +257,6 @@ function(add_ccf_app name)
       ${virt_name} PRIVATE ${PARSED_ARGS_LINK_LIBS_VIRTUAL} ccf.virtual
     )
 
-    target_link_options(${virt_name} PRIVATE LINKER:--no-undefined)
```
eddyashton, line 260: @letmaik Do you know why this was added? I guess to catch the versions of the OE libraries that did/didn't include the new symbol?

ashamis (approved):

achamayou (approved):

cmake/ccf_app.cmake
letmaik: I added this to be consistent in how OE links enclave images when linking against oeenclave. And yes, in this case during testing without it an unresolved symbol only lead to an error at runtime. So I think you should keep it for non-SAN builds.

PR #1563, opened 2020-09-03
Support custom authorization header in client requests

### LINKED ISSUE #1203: "Authorization" header support for endpoints
Currently when a Bearer token is in to an HTML request to a CCF node the server will nack with the following error:
```
Error parsing request: Unable to parse signature params from: Bearer eyJ0eXAiOiJKV...
```
In order to support access tokens from AAD. The server would have to change to allow such Bearer tokens in the HTML header.

Ideally, there could be some functions for decoding, validating such tokens.

letmaik:
Fixes #1203.

This also exposes headers to JS endpoints.

Note that the change in src/http/http_rpc_context.h is mostly indentation, but GitHub is bad at displaying that.


achamayou (approved):

tests/custom_authorization.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+import tempfile
+import http
+import os
```
achamayou, line 5:
```suggestion
```

PR #1566, opened 2020-09-04
Store less data when re-executing requests

ashamis:
When backups re-execute requests we need to store the actual client request so it can be executed. We were storing the request twice in different formats. This change means we only store it once.

* remove pbft::Request 
* in aft::Request remove pbft_raw field


src/consensus/aft/request.h
```diff
@@ -14,19 +14,18 @@ namespace aft
...
-    MSGPACK_DEFINE(caller_id, caller_cert, raw, pbft_raw, frame_format);
+    MSGPACK_DEFINE(caller_id, rid, caller_cert, raw, frame_format);
 
     std::vector<uint8_t> serialise()
     {
       bool include_caller = false;
-      size_t size = sizeof(caller_id) + sizeof(bool) + sizeof(size_t) +
-        raw.size() + sizeof(size_t) + sizeof(enclave::FrameFormat) +
-        pbft_raw.size();
+      size_t size = sizeof(caller_id) + sizeof(rid) + sizeof(bool) +
```
achamayou, line 27:
```suggestion
      size_t size = sizeof(caller_id) + sizeof(rid) + sizeof(include_caller) +
```

achamayou (approved):

PR #1575, opened 2020-09-04
Support more content types

### LINKED ISSUE #1554: Unit test helpers should allow sending raw body
(from https://github.com/microsoft/CCF/issues/1549)

The Python test helpers convert all POST payloads to JSON. There should be a `raw=True` kw or similar to avoid this in order to test arbitrary payload formats.

letmaik:
Fixes #1554.

Inside a JS endpoint:
- Request body is a `Body` object which is similar to https://developer.mozilla.org/en-US/docs/Web/API/Body, except that it has a synchronous API and only offers `.text()`, `.json()`, and `.arrayBuffer()`.
- Response body can be a string, a JSON-serializable object, a typed array object, or an ArrayBuffer object. The type decides the response content type. In the future, an API should be added so that the endpoint can override the default content type. Note that previously strings were serialized as JSON -- this doesn't happen anymore as it's generally not what you want.

Using the Python client (and hence in unit tests):
- Request body can now be `str`, `bytes`, and all other types are assumed to be JSON. The type determines the default content type (`text/plain`, `application/octet-stream`, `application/json`). If the body is a `str` starting with `@` then it is only assumed to be JSON if the filename ends with `.json` -- no JSON parsing is done here, this is just for determining the default content type.
- Response body is `str` (for `text/plain`), a parsed JSON object (for `application/json`), or else `bytes`.


python/ccf/clients.py
```diff
@@ -452,7 +474,14 @@ def request(
         global_commit = unpack_seqno_or_view(out[18:26])
         payload = out[26:]
         if status_code == 200:
-            body = json.loads(payload) if payload else None
+            if payload:
+                # Ideally, the content type should be sent like for HTTP.
```
achamayou, line 478: The view here is that websockets is the efficient option, and it is the efficient option because it does away with the large variable header. Endpoints are expected to expose their schema out of band rather than send (or expect) self-describing traffic.

An application that wishes to replicate the content-type concept across their websockets endpoints would do so in the outer frame of their schema.

src/apps/js_generic/js_generic.cpp
```diff
@@ -365,9 +372,23 @@ namespace ccfapp
         JS_SetPropertyStr(ctx, global_obj, "query", query_str);
 
         const auto& request_body = args.rpc_ctx->get_request_body();
-        auto body_str = JS_NewStringLen(
-          ctx, (const char*)request_body.data(), request_body.size());
-        JS_SetPropertyStr(ctx, global_obj, "body", body_str);
+        auto content_type = args.rpc_ctx->get_request_header("content-type");
```
achamayou, line 375: Let's talk about this, but I think we'd like to pass the ArrayBuffer every time and let the JS code decide how it wants to deal with wrt content-type, rather than manage it on the C++ side, unless there's a good reason to do it otherwise.

achamayou (approved):

python/ccf/clients.py
letmaik: So, is it fine to keep the try-except? If not, then maybe for WebSocket it should always return `bytes`?

src/apps/js_generic/js_generic.cpp
letmaik: I'll try to make body something like a [`Body`](https://developer.mozilla.org/en-US/docs/Web/API/Body).

letmaik: Done, can you have another look?

letmaik:
While changing the npm app to use the new Body class I stumbled upon an issue when returning ArrayBuffer: protobufjs returns an Uint8Array with byteLength 14 and the underlying ArrayBuffer has a byteLength of 8192, which means we should probably also accept typed arrays as return values to avoid potential issues where a slice was taken and the lengths would mismatch. I'll see if quickjs has the necessary APIs to deal with typed arrays.

src/apps/js_generic/js_generic.cpp
```diff
@@ -198,6 +199,54 @@ namespace ccfapp
...
+      return JS_ThrowTypeError(
+        ctx, "Passed %d arguments, but expected none", argc);
+    
+    auto body = static_cast<const std::vector<uint8_t>*>(JS_GetOpaque(this_val, body_class_id));
+    auto body_ = JS_NewArrayBufferCopy(
+          ctx, body->data(), body->size());
+    return body_;
+  }
+
+  // Partially replicates https://developer.mozilla.org/en-US/docs/Web/API/Body
```
achamayou, line 251: Ah, what an excellent find :D

src/apps/js_generic/js_generic.cpp
achamayou: The new scheme looks excellent :)

python/ccf/clients.py
achamayou: We probably should, but then we have to adjust the logging test accordingly. Let's keep it as is and do that clean up in a separate PR?

achamayou (approved):

python/ccf/clients.py
letmaik: https://github.com/microsoft/CCF/issues/1584

PR #1576, opened 2020-09-04
Upgrade Evercrypt code

wintersteiger:
This is the latest Evercrypt code. I made a few minor changes; this corresponds to https://github.com/project-everest/hacl-star/pull/345.


ashamis:
Can you please run the daily tests. It would be good to have a run with the sanitizer.

achamayou:
@wintersteiger to trigger the daily build, the easiest thing to do is to do a small edit to .daily.yml (like add a line return).

Was anything under 3rdparty manually edited, or did it all come from the script and only other paths needed updates?

achamayou:
@wintersteiger Could you also update the sha in cgmanifest.json for https://github.com/project-everest/hacl-star please?

cmake/crypto.cmake
```diff
@@ -24,7 +24,7 @@ if("sgx" IN_LIST COMPILE_TARGETS)
   )
   target_compile_definitions(
     evercrypt.enclave PRIVATE INSIDE_ENCLAVE KRML_HOST_PRINTF=oe_printf
-                              KRML_HOST_EXIT=oe_abort
+                              KRML_HOST_EXIT=oe_abort LINUX_NO_EXPLICIT_BZERO
```
achamayou, line 27: What does that flag do?

src/node/history.h
```diff
@@ -208,20 +208,21 @@ namespace ccf
   private:
     uint64_t index;
     uint32_t max_index;
+    static const uint32_t hash_size = 32U;
```
achamayou, line 211: I don't think we need that, that's just Sha256Hash::SIZE.

achamayou (approved):

cmake/crypto.cmake
wintersteiger: This tells the build that there's no [explicit_bzero](https://www.freebsd.org/cgi/man.cgi?query=explicit_bzero&sektion=3). I don't know what exactly that's used for though.

src/node/history.h
wintersteiger: That's right, done.

wintersteiger:
@achamayou no modifications, all of those have gone into hacl-star before updating the files. `cgmanifest` updated.

cmake/crypto.cmake
achamayou: Huh. There doesn't seem to be any calls to `bzero()` in the evercrypt code. musl as provided in openenclave does expose both `bzero()` and `explicit_bzero()` though, so should we have this flag at all?

wintersteiger: I see this:

    root@45ddd2ed032b:/CCF/3rdparty/hacl-star/evercrypt# grep bzero *
    Lib_Memzero0.c:    explicit_bzero(dst, len_);

If it's properly supported by musl we may want to update the macro-checks in `Lib_Memzero0.c` (the build explicitly asks for pull requests...)

wintersteiger:
Done! This is ready to merge once the CI finishes.

wintersteiger:
Hmm:

    /opt/oe_lvi/ld: libevercrypt.enclave.a(Lib_Memzero0.c.o): in function `Lib_Memzero0_memzero':
    /CCF/3rdparty/hacl-star/evercrypt/Lib_Memzero0.c:40: undefined reference to `explicit_bzero'

Is this special to the LVI mitigations and do we have to compile them with `_GNU_SOURCE` too to get access to `explicit_bzero`?

wintersteiger:
It looks like OE's musl doesn't have `explicit_bzero`, I can't find it in the .a. I'm falling back to `LINUX_NO_EXPLICIT_BZERO` so we can merge this while I search for a substitute.

PR #1581, opened 2020-09-07
Propagate errors from foreach functors called in Lua

eddyashton:
Spotted while debugging something else. When calling `tx:foreach( foo )` in Lua, any errors in foo would be silently ignored (they're invoked from a separate `lua_pcall`, and we were ignoring the return code from this). Errors in these functors are now raised as exceptions, the same as we do with any other lua error.


achamayou (approved):
Good catch.

numba1stunna80 (approved):

PR #1589, opened 2020-09-08
Upgrade quickjs from 2020-04-12 to 2020-09-06

achamayou:


src/apps/js_generic/js_generic.cpp
```diff
@@ -467,8 +467,8 @@ namespace ccfapp
         const std::string path = "/__endpoint__.js";
         JSValue module = JS_Eval(
           ctx,
-          code.c_str(),
-          code.size() + 1,
+          code.data(),
```
letmaik, line 470: Does `data()` guarantee that you get a null byte at the end? The size() fix looks fine.

letmaik: Although it looks like quickjs actually doesn't read past the end anyway, probably the docs are slightly incorrect.

achamayou: No, it doesn’t, but if you use c_str/size+1, then the lexer chokes on the null byte, which it doesn’t expect.

This isn’t what the doc suggests, but it’s what the code does.

achamayou: I've reported this to the quickjs-devel list.

eddyashton (approved):

src/apps/js_generic/js_generic.cpp
achamayou: I've also reverted back to c_str() for consistency with the other calls, even though it doesn't matter here (we pass size()).

letmaik (approved):

PR #1590, opened 2020-09-08
Non-recursive callstacks

eddyashton:
The callstacks we get through backward-cpp on signals have a bad habit of recursing infinitely. This removes that behaviour by manually installing/uninstalling the signal handlers. This is based on the code in `backward::SignalHandling`, but explicitly resetting the signal handler to default inside the handler before re-raising it.

Note that this signal handler is not [async-signal-safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html) (and was not previously). So we're optimistically gathering and printing stacks during certain kinds of signals, but in particularly unlucky signals we may corrupt the program state further before exiting.


src/ds/stacktrace_utils.h
```diff
@@ -27,12 +27,78 @@ namespace stacktrace
...
+
   static inline void init_sig_handlers()
   {
 #ifndef INSIDE_ENCLAVE
-    // This registers signal handlers on construction, and these remain after
-    // destruction.
-    backward::SignalHandling sh;
+    // This is based on the constructor of backward::SignalHandling, but avoids
+    // infinitely recursing stacktraces
+    constexpr size_t stack_size = 1024 * 1024 * 8;
```
achamayou, line 48: We typically run with 1024 stack pages, so unless I'm missing something, this will turn out to be too much in most of our SGX tests?

ashamis (approved):

PR #1592, opened 2020-09-09
python 3.7 to 3.8

achamayou:
Update Python to 3.8


letmaik:
Shouldn't CI test the minimum supported Python version? If this PR is raising the minimum version then it should also be raised in setup.py.

achamayou:
@letmaik it does run in CI, but thanks for pointing out the setup.py upgrade.

eddyashton (approved):

PR #1594, opened 2020-09-09
Return an error when the node quote cannot be found

### LINKED ISSUE #1571: /node/quote return code is incorrect before node is trusted
The `/node/quote` rpc returns node_id: 0 before the node is trusted. We do not need to provide a node_id there, /node/state has a node id with necessary context, we should either default to null or remove it altogether.

achamayou:
Fix #1571


eddyashton (approved):

tests/governance.py
```diff
@@ -82,6 +82,15 @@ def test_user(network, args, notifications_queue=None, verify=True):
     return network
 
 
+@reqs.description("Add untrusted node, check no quote is returned")
+def test_no_quote(network, args, notifications_queue=None, verify=True):
+    primary, _ = network.find_nodes()
+    untrusted_node = network.create_and_add_pending_node(args.package, "localhost", args)
+    with untrusted_node.client(ca=os.path.join(untrusted_node.common_dir, f"{untrusted_node.node_id}.pem")) as uc:
```
achamayou, line 89: client() otherwise defaults to the network identity as the root CA.

Issue #1601, opened 2020-09-10
Fix re-entrant lock issues BFT has with the AFT class


ashamis:
When replicas execute on the backup they sometimes run into the issue that they need to take locks that the same thread has already acquired. This caused a deadlock! The current solution is to skip locking but this is a short term hack and a real solution needs to be put in.

Functions with this issue:
- [ ] aft::get_commit_idx (#2065)
- [ ] aft::get_term (#2065)
- [x] get_commit_term_and_idx (#2062)
- [x] aft::get_term(idx) (#2062)
- [x] aft::replicate (#1887)
- [x] aft::periodic (#1887)

Issue #1600, opened 2020-09-10
BFT cleanup


ashamis:
Remove PBFT files that are no longer needed

- [x] Cleanup and merge pbft_tables.h with entities.h
- [ ] Remove pbft_config.h
- [ ] Remove parameters.h
- [ ] Remove pbft_deps.h
- [ ] Remove pbft_globals.h
- [ ] Remove pbft.h

Issue #1577, opened 2020-09-05
AFT: Support execution on all replicas


ashamis:
We will need the following features to be able to execute all consensus request on the primary and backups

- [x] Execution on the backups
- [x] Ensure that backups and primary maintain the same Merkle roots while executing
- [x] Backups can be passed the signature from the primary correctly (currently just passing the tx not the actual signature)

PR #1602, opened 2020-09-10
Execution on backup with signatures

ashamis:
Part of #1577

- Replicate the signatures table when executing on the backup.
- Do not forward requests that are meant to be executed locally
- set tx/commit to execute locally
- when forwarding request with backup execution write the forwarded request to the requests table
- When opening a network make sure that we wait for all the nodes to be ready

NOTE:
 - There is an issue with how we deal with locks in the AFT class, it is outside the scope of this PR and tracked in #1601
 - Several cleanup cleanup items tracked #1600

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -745,13 +745,12 @@ namespace aft
             if (sig_term)
             {
               LOG_INFO_FMT("AAAAAA");
-              state->view_history.update(state->commit_idx + 1, sig_term);
-
               if (consensus_type == ConsensusType::BFT)
               {
                 state->last_idx = i;
               }
               commit_if_possible(r.leader_commit_idx);
+              state->view_history.update(state->commit_idx + 1, sig_term);
             }
             break;
           }
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -210,7 +210,6 @@ namespace aft
       state->current_view = term;
       state->last_idx = index;
       state->commit_idx = commit_idx_;
-        LOG_INFO_FMT("AAAAAA");
       state->view_history.update(index, term);
       state->current_view += 2;
       become_leader();
@@ -232,7 +231,6 @@ namespace aft
       state->last_idx = index;
       state->commit_idx = commit_idx_;
       state->view_history.initialise(terms);
-        LOG_INFO_FMT("AAAAAA");
       state->view_history.update(index, term);
       state->current_view += 2;
       become_leader();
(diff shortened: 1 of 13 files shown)


src/consensus/aft/raft.h
```diff
@@ -280,6 +287,10 @@ namespace aft
 
     std::pair<Term, Index> get_commit_term_and_idx()
     {
+      if (consensus_type == ConsensusType::BFT && is_follower())
```
achamayou, line 290: That's what #1601 is about, right?

src/consensus/pbft/pbft_tables.h
```diff
@@ -17,5 +17,8 @@ namespace pbft
 
   static constexpr auto replicate_type_pbft = kv::ReplicateType::SOME;
   static const std::unordered_set<std::string> replicated_tables_pbft = {
-    Tables::PBFT_REQUESTS, Tables::PBFT_PRE_PREPARES, Tables::PBFT_NEW_VIEWS};
+    Tables::PBFT_REQUESTS,
```
achamayou, line 20: Let's try to merge the requests table when we can, and rename the rest?

src/kv/store.h
```diff
@@ -746,6 +746,17 @@ namespace kv
         {
           success = DeserialiseSuccess::PASS_NEW_VIEW;
         }
+        else if (views.find("ccf.signatures") != views.end())
+        {
```
achamayou, line 750: Just for my understanding, is it right that this the branch we take when re-execute on a secondary? What we're doing here is taking the primary's signature, and applying it without re-execution?

src/node/rpc/common_endpoint_registry.h
```diff
@@ -75,6 +75,7 @@ namespace ccf
       };
       make_command_endpoint("tx", HTTP_GET, json_command_adapter(get_tx_status))
         .set_auto_schema<GetTxStatus>()
+        .set_execute_locally(true)
```
achamayou, line 78: I don't think this is right, because the session could otherwise be forwarded and the tx status will be out of sync with the results of execution.

achamayou (approved):

src/consensus/aft/raft.h
ashamis: yes

src/consensus/pbft/pbft_tables.h
ashamis: This is part of  #1600

src/node/rpc/common_endpoint_registry.h
ashamis: As discussed offline - I added a new endpoint called local_tx which is the same as tx but sets "set_execute_locally"

PR #1599, opened 2020-09-10
Support optional arguments when creating proposals via the CLI

eddyashton:
When we made it possible to specify the user data at creation (ie - in the `new_user` proposal), we broke the CLI as this `user_data` argument was required (but unparseable) on the CLI.

Before:

```
$ python -m ccf.proposal_generator new_user --help
usage: proposal_generator.py new_user [-h] user_cert_path user_data

$ python -m ccf.proposal_generator new_user bob_cert.pem 
usage: proposal_generator.py new_user [-h] user_cert_path user_data
proposal_generator.py new_user: error: the following arguments are required: user_data
```

After
```
$ python -m ccf.proposal_generator new_user --help
usage: proposal_generator.py new_user [-h] user_cert_path [user_data]

# Can call without specifying any user data
$ python -m ccf.proposal_generator new_user bob_cert.pem
[2020-09-10 16:25:08.973] SUCCESS | Writing proposal to ./new_user_proposal.json
[2020-09-10 16:25:08.973] SUCCESS | Wrote vote to ./new_user_vote_for.json

# Or can specify user data as JSON on the CLI
$ python -m ccf.proposal_generator --pretty-print new_user bob_cert.pem '{"friendly_name": "Bob"}'
[2020-09-10 16:25:40.841] SUCCESS | Writing proposal to ./new_user_proposal.json
[2020-09-10 16:25:40.841] SUCCESS | Wrote vote to ./new_user_vote_for.json

$ cat new_user_proposal.json 
{
  "script": {
    "text": "tables, args = ...; return Calls:call(\"new_user\", args)"
  },
  "parameter": {
    "cert": "-----BEGIN CERTIFICATE----- [SNIP] -----END CERTIFICATE-----\n",
    "user_data": {
      "friendly_name": "Bob"
    }
  }
}
```

PR #1603, opened 2020-09-11
Add test of Python CLI

eddyashton:
As noted in #1599, we don't currently have any tests of our Python CLI. We test the proposal generator extensively from code, but we have nothing that tests the CLI.

This PR adds a test of this, run as part of ever `Release` test (so currently in our Daily CI, and for each published version). This is non-exhaustive - we should extend it to cover all the awkward args, in particular for passing files/scripts/modules for JS-related proposals.


tests/test_python_cli.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+set -ex
+
+# This onlt checks that the following commands do not throw errors.
```
achamayou, line 6:
```suggestion
# This only checks that the following commands do not throw errors.
```

achamayou:
Now that we ship a full blown Python package, we could switch to click (https://click.palletsprojects.com/en/7.x/) and use the facility it provides for testing (https://click.palletsprojects.com/en/7.x/testing/).

achamayou (approved):

ashamis (approved):

.azure-pipelines-templates/daily-matrix.yml
```diff
@@ -47,4 +47,6 @@ jobs:
       cmake_args: '${{ parameters.build.SGX.cmake_args }} -DTLS_TEST=ON -DSHUFFLE_SUITE=ON'
       suffix: 'Release'
       artifact_name: 'SGX_Release'
-      ctest_filter: '-LE "benchmark|perf"'
\ No newline at end of file
+      ctest_filter: '-LE "benchmark|perf"'
+  
+# Nop to trigger daily build. TODO: Remove this
```
jumaffre, line 52: Remove this?

tests/test_python_cli.sh
```diff
@@ -0,0 +1,35 @@
...
+# This only checks that the following commands do not throw errors.
+# It is expected that other tests cover correctness of the generated
+# proposals, this just checks the basic usability of the CLI.
+
+keygenerator.sh --help
+keygenerator.sh --name alice
+keygenerator.sh --name bob --gen-enc-key
+
+
+python -m ccf.proposal_generator --help
```
jumaffre, line 15: Perhaps worth annotating some of these and include [from the docs](https://microsoft.github.io/CCF/master/members/proposals.html) directly?

tests/test_python_cli.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
jumaffre, line 1: Can we move this under `python/` as this only tests the python package?

achamayou:
@eddyashton could you address @jumaffre's comments before this is merged?

tests/test_python_cli.sh
eddyashton: I'd rather not - I don't think we gain anything by moving it, and its probably CI-run debugging to check that we copy/reference it correctly from a different path. Its part of CCF testing the package (and in the same style as `test_install.sh`, containing CCF docs annotations), rather than part of the package itself. When/if we move to click, we should probably replace this with a test folder/app/whatever they recommend in the package itself, but I think this is fine for now.

tests/test_python_cli.sh
eddyashton: I was going to do this, but unlike our other annotations we actually want the output of these commands in the docs, not just the input command (or code snippet) itself. If we find a solution for running commands and capturing their output to insert during the sphinx build, then we should definitely replace the proposal_generator docs with this, but I'm not going to tackle that just now.

PR #1604, opened 2020-09-11
js: pass request object as function argument & return response object

letmaik:
...and expose the URL path parameters as well.


letmaik:
I'll extend this PR a bit more to also include returning a response object containing the body, status code, and headers. Then the low-level request/response interface is more or less complete I think.

src/apps/js_generic/js_generic.cpp
```diff
@@ -507,9 +521,10 @@ namespace ccfapp
         }
 
         // Call exported function
-        int argc = 0;
-        JSValueConst* argv = nullptr;
+        int argc = 1;
+        JSValueConst* argv = (JSValueConst *)&request;
```
eddyashton, line 525: Minor: this is supposed to be a (C-style) array, can we make it look more like one? Something like `JSValueConst argv[] = {&request};`?

eddyashton (approved):

src/apps/js_generic/js_generic.cpp
letmaik: I don't think it works like that, compiler is complaining as well: `a value of type "JSValue *" cannot be used to initialize an entity of type "int32_t"`.

achamayou (approved):

achamayou (approved):

PR #1606, opened 2020-09-14
Remove OE 0.10 workarounds

### LINKED ISSUE #1550: Replace oehost with oehostverify in virtual enclave
Left-over from https://github.com/microsoft/CCF/pull/1448.

OE's hostverify library in 0.10 is partly broken (see https://github.com/openenclave/openenclave/issues/3312) and therefore currently we need to link against oehost for virtual enclaves. This is fixed in OE master and we can switch to oehostverify with 0.11.

See CMakeLists.txt and cmake/common.cmake.

letmaik:
Fixes #1550

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,9 +146,7 @@ if("virtual" IN_LIST COMPILE_TARGETS)
            secp256k1.host
            aft.virtual
            sss.host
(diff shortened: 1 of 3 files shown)


eddyashton (approved):

achamayou (approved):
Thank you!

PR #1610, opened 2020-09-15
Proposing more compact logging in tests

achamayou:
While debugging tests, I've found that our logging format is extremely verbose, to the extent that it's difficult to find out what's going on. I think we can make it more compact without losing too much detail, and interpret test failures quicker and more accurately as a result.

This proposes two main things:

1. Colours/depth offset and description shortening
2. Hiding retries

I expect 2. to be controversial, but I think it really helps shorten things to readable traces.

The big drawback is that we can't really do this without modifying the public ccf.clients API. I'm not wildly happy about this, but I think it's just enough of a win for debugging to be worth doing still.


python/ccf/clients.py
```diff
@@ -26,13 +26,28 @@
...
 
-def truncate(string: str, max_len: int = 256):
+def truncate(string: str, max_len: int = 128):
     if len(string) > max_len:
         return f"{string[: max_len]} + {len(string) - max_len} chars"
     else:
         return string
 
 
+def info_or_capture(line, log_capture=None, depth=0):
```
achamayou, line 36: This needs to be split to avoid a deps cycle, but please ignore that for now.

python/ccf/clients.py
```diff
@@ -26,13 +26,28 @@
...
+    if log_capture is None:
+        LOG.opt(colors=True, depth=depth + 1).info(line)
+    else:
+        log_capture.append(line)
+
+
+def flush_info(lines, log_capture=None, depth=0):
```
achamayou, line 43: This (and the call above) needs to be split to avoid a deps cycle, but please ignore that for now.

python/ccf/clients.py
```diff
@@ -575,33 +598,40 @@ def call(
...
-                # call _direct_call
+                # call _call
                 self.is_connected = True
+                flush_info(logs, log_capture, 2)
                 return response
             except (CCFConnectionException, TimeoutError) as e:
                 # If the initial connection fails (e.g. due to node certificate
                 # not yet being endorsed by the network) sleep briefly and try again
                 if time.time() > end_time:
+                    flush_info(logs, log_capture, 2)
```
achamayou, line 631: Here and in other calls, the intention is to hold off logging until the last iteration essentially, so as not to print retries. There's at least two things it could do better:

1. mangling timestamps to keep them correct, right now the request stamp is essentially the same as the response stamp. So if the last request took forever, that's not reported accurately.
2. reporting a retry count "After X retries...". This isn't super straightforward because these things chain up, and then you have to wonder if you report the retries of one step below or all the way down.

python/ccf/commit.py
```diff
@@ -4,11 +4,22 @@
...
 
+from typing import Optional
 
 from ccf.tx_status import TxStatus
 
+from loguru import logger as LOG  # type: ignore
 
-def wait_for_commit(client, seqno: int, view: int, timeout: int = 3) -> None:
+
+def flush_info(lines, log_capture=None, depth=0):
```
achamayou, line 14: See above, ignore the duplicate.

eddyashton:
Can you add a screenshot of this? I assume the depth option is indenting the content of the lines, after the loguru preamble?

tests/infra/node.py
```diff
@@ -266,7 +266,7 @@ def client(self, identity=None, **kwargs):
             if identity
             else None,
             "ca": os.path.join(self.common_dir, "networkcert.pem"),
-            "description": f"node {self.node_id} as {identity or 'unauthenticated'}",
+            "description": f"[{self.node_id}|{identity or '?'}]",
```
jumaffre, line 269: `?` looks a little confusing for unauthenticated connections in the logs `2020-09-15 16:32:35.589 | INFO     | infra.node:wait_for_node_to_join:247 - [2|?] GET /node/commit`. Perhaps stick to "unauthenticated" or "anonymous"?

python/ccf/clients.py
```diff
@@ -537,24 +553,30 @@ def _response(self, response: Response) -> Response:
...
         http_verb: str = "POST",
         headers: Optional[dict] = None,
         signed: bool = False,
         timeout: int = DEFAULT_REQUEST_TIMEOUT_SEC,
+        log_capture: Optional[list] = None,
     ) -> Response:
         description = ""
         if self.description:
-            description = f"({self.description})" + (" [signed]" if signed else "")
+            description = f"{self.description}" + (" (sig)" if signed else "")
```
jumaffre, line 568: I think `signed` is slightly more precise than `sig`?

achamayou: Yes, but it's also longer. It takes precious line space away from display more interesting stuff like the message itself so I'd like to reduce it as much as possible. I was thinking something like `[0|u0]s ...` would be nice (to node 0, as user 0, signed request).

tests/infra/node.py
achamayou: That's super long, and extremely common. I do not want to print lots of "unauthenticated" that force me to truncate interesting payloads.

achamayou:
@eddyashton sure, so this is a good example of not printing 16 seconds worth of retries:

python/ccf/clients.py
eddyashton: "the request stamp is essentially the same as the response stamp. So if the last request took forever, that's not reported accurately"

This seems like a big loss. Can we fix it?

ashamis (approved):

achamayou:
Unfortunately I don't think it's possible to store logs on the client, because several of the upper layers do not have access to a client (some of them are short lived in inner loops), so I'm going to stay with the existing log_capture mechanism.

python/ccf/clients.py
achamayou: I may have found a way to do this.

achamayou: No I haven't. I'm going to give up for now, because I think I've already spent too much time on it.

PR #1612, opened 2020-09-16
Produce OpenAPI document describing CCF's endpoints

eddyashton:
Had this on the side for a while, finally got it into a place where (I think) it mostly works. Opening as a draft PR to see what's still missing, includes a few TODOs.

The core change is that the `/api` endpoint now returns an OpenAPI document, rather than just a list of paths. This involves some changes to our JSON schema-generating macros, to populate an OpenAPI document rather than returning isolated elements. For full customisability (ie - so the apps can, if they wish, store a separate document or elements and merge them in), we manipulate the OpenAPI document as a raw JSON document, with some helper functions to try and put things in the right place. This is more error prone than a strongly typed wrapper, but I think the flexibility is necessary.

There are 2 test for this. There's an `openapi_test` unit test, which really only tests compilation of the helper functions + custom types, and does minimal validation that the result looks like a plausible document. Then the existing `schema.py` e2e test now uses a 3rd party OpenAPI validation library to check that the produced documents are valid OpenAPI. This seems unhappy with some things the online validator was happy with (`$schema` elements in particular), so I've removed a few useful-but-non-essential features to support it. This test also dumps the documents like it does the individual schema responses, and I've been running those documents through the online swagger validator - this is _almost_ happy, but doesn't like nlohmann::json's "maps are lists of tuples" representations, as OpenAPI schema breaks from JSON schema in disallowing tuples.


python/setup.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: Minor: this file should not be checked in as generated by cmake from `setup.py.in`. I think I'll change this as this is confusing and have `setup.py` include `version.py.in` which specifies the version.

eddyashton: I've removed it. Don't know how it got included, since its in `.gitignore`. Dodgy merge at some point.

CMakeLists.txt
```diff
@@ -265,7 +270,7 @@ if(BUILD_TESTS)
     )
     use_client_mbedtls(kv_test)
     target_link_libraries(
-      kv_test PRIVATE ${CMAKE_THREAD_LIBS_INIT} secp256k1.host
+      kv_test PRIVATE ${CMAKE_THREAD_LIBS_INIT} secp256k1.host http_parser.host
```
jumaffre, line 273: Why does `http_parser.host` need to be added here?

achamayou:
@eddyashton I was under the impression that OpenAPI supported a strict superset of recent JSON schema, since https://github.com/OAI/OpenAPI-Specification/pull/1977

tests/schema.py
```diff
@@ -29,18 +31,25 @@ def run(args):
...
         check(
-            list_response, error=lambda status, msg: status == http.HTTPStatus.OK.value
+            api_response, error=lambda status, msg: status == http.HTTPStatus.OK.value
         )
-        methods = list_response.body["endpoints"]
-        all_methods.extend([m["path"] for m in methods])
 
-        for method in [m["path"] for m in methods]:
+        paths = api_response.body["paths"]
+        all_methods.extend([key for key in paths.keys()])
```
achamayou, line 45: Why not just:

```suggestion
        all_methods.extend(paths.keys())
```

achamayou (approved):
Very nice!

eddyashton:
> @eddyashton I was under the impression that OpenAPI supported a strict superset of recent JSON schema, since [OAI/OpenAPI-Specification#1977](https://github.com/OAI/OpenAPI-Specification/pull/1977)

I believe that's the case for v3.1, which is still pre-release/RC. The part we want that isn't supported is described in the current v3.0 spec [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#schema-object): "items - Value MUST be an object and not an array". Our items are sometimes arrays of schemas, to describe a tuple. The [equivalent part of the v3.1 spec](https://github.com/OAI/OpenAPI-Specification/blob/v3.1.0-dev/versions/3.1.0.md#schema-object) is far more concise, claiming to be a clear superset, so hopefully it should be supported once that is standard (and the relevant tools support it).

achamayou:
@eddyashton do we want to target that directly?

scripts/ci-checks.sh
```diff
@@ -16,7 +16,7 @@ echo "Shell scripts"
 find . -type f -regex ".*\.sh$" | grep -E -v "^./3rdparty/" | xargs shellcheck -s bash -e SC2044,SC2002,SC1091,SC2181
 
 echo "TODOs"
-"$SCRIPT_DIR"/check-todo.sh src
+#"$SCRIPT_DIR"/check-todo.sh src
```
ashamis, line 19: i think this should not be commented out

src/ds/json_schema.h
```diff
@@ -83,18 +200,28 @@ namespace ds
...
         nonstd::is_specialization<T, std::unordered_map>::value)
       {
-        // Nlohmann serialises maps to an array of (K, V) pairs
-        schema["type"] = "array";
-        auto items = nlohmann::json::object();
+        // Nlohmann serialises maps to an array of (K, V) pairs...
+        if (std::is_same<typename T::key_type, std::string>::value)
         {
-          items["type"] = "array";
+          // ...unless the keys are strings!
```
achamayou, line 206: That's because json only allows strings as keys in maps:

image

src/ds/openapi.h
```diff
@@ -0,0 +1,393 @@
...
+          auto items = nlohmann::json::array();
+          items.push_back(add_schema_component<typename T::first_type>());
+          items.push_back(add_schema_component<typename T::second_type>());
+          schema["items"] = items;
+          return add_schema_to_components(
+            document, ds::json::schema_name<T>(), schema);
+        }
+        else if constexpr (
+          std::is_same<T, std::string>::value || std::is_same<T, bool>::value ||
+          std::is_same<T, uint8_t>::value || std::is_same<T, uint16_t>::value ||
```
achamayou, line 264: This seems like it could be shorter if it used is_arithmetic (unless I'm missing something we want to exclude explicitly?)

CMakeLists.txt
eddyashton: `openapi.h` calls `http_method_str` to convert `http_method`s to strings. This comes from `http_parser`. So anything that includes `openapi.h` needs to link `http_parser` to get the implementation.

src/ds/openapi.h
eddyashton: Good suggestion! This was a copy-paste reduction from `json_schema.h`, where we want to name each type individually, but `is_arithmetic` is much shorter.

src/ds/json_schema.h
eddyashton: Yes - I understand why nlohmann does this, just trying to make sure the code explains these branches as its not obvious they're describing an nlohmann implementation detail (it _could_ have decided that string-keyed maps were also pair-lists, not objects).

eddyashton:
> @eddyashton do we want to target that directly?

I don't think we should target it while its pre-release, and the validators (even SwaggerHub) don't currently support 3.1. I _assume_ an official release is imminent (its been RC for several months), but can't find a date. Once that's done, we may be waiting for an update to the Python validator package before we can move.

achamayou (approved):

PR #1613, opened 2020-09-16
AFT: verify signature on the backups

ashamis:
When a BFT backup receives a signature append entry it will validate the signature and compare it to its own merkle root (same as with cft).

### DIFF
--- a/src/kv/store.h
+++ b/src/kv/store.h
@@ -699,7 +699,7 @@ namespace kv
 
         auto h = get_history();
 
-        auto search = views.find("ccf.signatures");
+        auto search = views.find(ccf::Tables::SIGNATURES);
         if (search != views.end())
         {
           // Transactions containing a signature must only contain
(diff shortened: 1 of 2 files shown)


src/kv/store.h
```diff
@@ -749,6 +749,14 @@ namespace kv
             return success;
           }
           auto h = get_history();
+          if (!h->verify(term_))
```
jumaffre, line 752: The paths for CFT and BFT seem pretty much identical now. Can we refactor?

ashamis: Unfortunately they are not :( This is the smallest/simplest variant of the code I have come up with.

src/kv/store.h
```diff
@@ -749,6 +749,14 @@ namespace kv
...
+          {
+            LOG_FAIL_FMT("Failed to deserialise");
+            LOG_DEBUG_FMT("Signature in transaction {} failed to verify", v);
+            throw std::logic_error(
+              "Failed to verify signature, view-changes not implemented");
+            return DeserialiseSuccess::FAILED;
```
achamayou, line 758: why is this returning just after throwing?

achamayou: Also what does signature verification have to do with view changes?

achamayou (approved):

src/kv/store.h
ashamis: It is throwing because i want the test to fail when this is happens.
In regards to view-changes, if the primary send a signature that does not verify we want to change the primary i.e. a view change which is not implemented yet

PR #1614, opened 2020-09-16
Add tsoa-based TypeScript app

letmaik:


tests/modules.py
```diff
@@ -205,6 +206,71 @@ def test_npm_app(network, args):
...
+    with primary.client("user0") as c:
+        body = [1, 2, 3, 4]
+        r = c.post("/app/partition", body)
+        assert r.status_code == http.HTTPStatus.OK, r.status_code
+        assert r.body == [[1, 3], [2, 4]], r.body
+
+        r = c.post("/app/proto", body)
+        assert r.status_code == http.HTTPStatus.OK, r.status_code
+        assert r.headers["content-type"] == "application/x-protobuf"
+        # We could now decode the protobuf message but given all the machinery
```
achamayou, line 264: How is it encoded without this machinery?

letmaik: This PR uses the same endpoints as in the basic npm app. Any improvements etc to that should probably be a different issue/PR.

tests/npm-tsoa-app/README.md
```diff
@@ -0,0 +1,19 @@
+# Node.js/npm CCF app using tsoa
```
jumaffre, line 1: I am wary of writing more docs in Markdown files that reference other Markdown files. Could we move all the build-your-JS-app docs to a new dedicated section [here](https://microsoft.github.io/CCF/master/developers/index.html) (not necessarily in this PR!)?

tests/npm-tsoa-app/tsoa-support/routes.ts.tmpl
```diff
@@ -0,0 +1,191 @@
+// A tsoa routes template for CCF. Needs to be post-processed by postprocess.js.
```
jumaffre, line 1: This comment seems to have been added manually but the rest of the file seems to have been generated by `tsoa`?

letmaik: No, this whole file is a template that gets processed by tsoa. It's not generated.

tests/npm-tsoa-app/README.md
achamayou, line 1: I would prefer to have this integrated in the main Sphinx doc, where it's much more easily searchable and cross-referenced.

tests/npm-tsoa-app/package-lock.json
```diff
@@ -0,0 +1,1765 @@
+{
```
achamayou, line 1: Do we really need a lockfile for this?

letmaik: It's common practice. Adding it to .gitignore will give the impression that this is what you're supposed to do for your own projects. I'll leave it to you though, we can also add a note somewhere that the package-lock.json file is only git-ignored because this is a test and not a real app and we don't want to blow up the repo.

tests/npm-tsoa-app/README.md
letmaik: Duplicate of https://github.com/microsoft/CCF/pull/1614#discussion_r490070508.

tests/modules.py
achamayou: I think you can use the reflection API to parser/extract the value the same way the npm app does it (ie. without code gen): https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message

letmaik: The existing npm app doesn't do it either. I don't think it's essential to this test and it's not what people would do in reality.

jumaffre (approved):

tests/npm-tsoa-app/package-lock.json
achamayou: This is a sample though, and we probably won't refresh this often if at all, I would prefer to have it in .gitignore and use recent dependencies in the CI.

Other than the bloat, pinning means we'll stay on some arbitrary set of old stuff in practice (these are not real dependencies for us), and users that picks this up may get breaks against modern versions (or be forced into needlessly old dependencies instead). We make the same tradeoff for the Python tests, and I think it is the right tradeoff.

achamayou (approved):

tests/npm-tsoa-app/README.md
achamayou: Tracked as #1622

PR #1616, opened 2020-09-16
Include view history in snapshot

### LINKED ISSUE #1607: Include view history in snapshot
Once a snapshot has successfully deserialised a snapshot, it should still be able to respond to historical transaction status (`tx` endpoint). To do so, we should include the view history in every snapshot and restore it whenever it is deserialised.

jumaffre:
Resolves #1607 

We now include the full view history in the snapshot so that backups that resumed from a snapshot can also serve transaction status via the `/tx` endpoint.


src/consensus/aft/test/view_history.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    history.initialise({3, 3, 3, 5, 6, 12});
+    CHECK(history.view_at(4) == 3);
+    CHECK(history.view_at(8) == 5);
+    CHECK(history.view_at(19) == 6);
+    CHECK(history.view_at(20) == 6);
+  }
+}
+
+TEST_CASE(
```
jumaffre, line 173: This is the only test case that was added after the file rename.

tests/suite/test_suite.py
```diff
@@ -65,7 +65,9 @@
     e2e_logging.test,
     e2e_logging.test,
     e2e_logging.test,
+    e2e_logging.test,
     reconfiguration.test_add_node_from_snapshot,
+    e2e_logging.test_view_history,
```
jumaffre, line 70: This actually tests that all nodes have the same view history. This test would have failed before this PR.

ashamis (approved):

achamayou (approved):

Issue #1615, opened 2020-09-16
KV opacity can be broken by compaction


eddyashton:
If we have a long-running transaction that gets views at different points, it can break opacity if a compaction happens while it is in-flight:

```
    kv::Tx tx;

    auto view_a = tx.get_view(map_a);
    // This transaction does something slow. Meanwhile...

    // ...another transaction commits...
    increment_vals();
    // ...and is compacted...
    kv_store.compact(kv_store.current_version());

    // ...then the original transaction proceeds, expecting to read a single version
    auto view_b = tx.get_view(map_b);

    auto a_opt = view_a->get(key_a);
    auto b_opt = view_b->get(key_b);
    
    REQUIRE(a_opt == b_opt);
```

The call to `get_view(map_b)` eventually calls `create_view(Version version)` in `untyped_map.h`, using the `read_version` assigned in this `tx` for the `get_view` on `map_a)`. But this version is _older_ than anything in `map_b`'s roll - the requested version was deleted during compact. We currently have an odd case at the end of `create_view` which returns _a_ view, but this is definitely wrong. We're not sure why this is ever valid, so should be removed.

Fixes:
- If we can't find a suitable version in the roll during `create_view`, return nullptr or an error. Somewhere this becomes an error which propagates to the caller (past the transaction logic) to retry the entire transaction.
- Explore extending the range of states which are kept in the roll at compaction. Perhaps we can track exactly which versions are still potentially requestable (a pending transaction has got them as `read_version`), but we still need an upper limit on the length of the roll for memory growth.

achamayou:
Raising in create_view() when a suitable version can't be found seems like the right thing to do.

The second bullet point makes perfect sense, but is an optimisation rather than a fix, I think.

eddyashton:
Closing this issue as resolved in #1817. Once we support multi-threaded execution and have long-running, potentially asynchronous transactions, we may revisit the optimisation mentioned in the second bullet point, but it is not a priority currently.

PR #1617, opened 2020-09-16
Fix opacity issue around KV compaction

eddyashton:
Fix for #1615.

Adds a failing test (`"Mid-tx compaction"`) and a fix (don't return views at a higher version than was requested, throw if we don't have an appropriate view).

Also fixes a related issue revealed by this, where `swap_private_maps` was called with stores at different versions. This was previously hidden as we'd happily start a transaction over a store at v2, then return a view over a map in it at v4.


achamayou (approved):

src/node/rpc/frontend.h
```diff
@@ -436,6 +436,13 @@ namespace ccf
             }
           }
         }
+        catch (const kv::CompactedVersionConflict& e)
+        {
+          // The executing transaction failed because of a conflicting
+          // compaction. Reset and retry
+          tx.reset();
```
ashamis, line 443: I might be wrong but this concerns me for a bit.
If a tx that fails with CompactedVersionConflict takes a minute to run and we compact every 10 seconds then we could end up in a scenario where this tx will realistically never commit but eat system resources (Assuming this is some background load in the system). Should we consider failing this transaction after X retries?

src/node/rpc/frontend.h
```diff
@@ -436,6 +436,13 @@ namespace ccf
             }
           }
         }
+        catch (const kv::CompactedVersionConflict& e)
```
ashamis, line 439: logging e.what() could be useful for debugging

ashamis (approved):

src/kv/store.h
```diff
@@ -997,6 +997,19 @@ namespace kv
      **/
     void swap_private_maps(Store& store)
     {
+      {
+        const auto source_version = store.current_version();
+        const auto target_version = current_version();
+        if (source_version != target_version)
```
jumaffre, line 1003: I feel like this check will break recovery where we have the CCF KV (public-only at this point, i.e. private maps are empty) and the recovery KV (both private and public). I don't think that the two stores are guaranteed to be at the same version when the swap is performed.

src/node/rpc/frontend.h
eddyashton: Wrote a long blurb here but I changed my mind. In short: This is similar to the existing conflict case (`kv::CommitSuccess::CONFLICT`). We only get conflicts in either case if another transaction write transaction is submitted while we're executing (if there are no writes since the last compact, this map is unaffected by the compact, so the requested 'old' version is still available). So any transaction that takes 1 minute _may_ never complete, but only if other transactions are submitted in the interim.

I think a retry limit is a sensible fix for either case. Perhaps this should be caught by a timeout instead - if a transaction takes too long (because it was forwarded, or because it was retried 10 times, or just because its really slow) we may want some global system to kill it and return an error to the user? Let's discuss at standup.

src/kv/test/kv_test.cpp
```diff
@@ -1001,3 +1015,110 @@ TEST_CASE("Conflict resolution")
...
+      // This transaction does something slow. Meanwhile...
+
+      // ...another transaction commits...
+      increment_vals();
+      // ...and is compacted...
+      kv_store.compact(kv_store.current_version());
+
+      // ...then the original transaction proceeds, expecting to read a single
+      // version
+      auto view_b = tx.get_view(map_b);
```
jumaffre, line 1105: Can we write a comment here or only `try catch` this call instead of the whole block?

src/node/rpc/frontend.h
achamayou: +1 for a retry cap.

Also, an explicit mention in the doc to tell users _not_ to catch all/or `CompactedVersionConflict` in code that calls get_view()...

src/kv/store.h
achamayou: Yeah that's definitely not guaranteed, the public store is likely further ahead, which is fine (it's guaranteed not have any private state after the version the private store reached when it finished recovery).

src/node/rpc/frontend.h
eddyashton: I'll add a retry cap.

With regard to avoiding accidental catches, I've made `CompactedVersionConflict` not inherit from any standard exception type, so its slightly harder to catch accidentally (you'd have to do the unusual `catch (...)` rather than `catch (const std::exception& e)`). I wonder if we should just avoid mentioning this particular foot-cannon in the docs at all, rather than saying "please don't load it". To expand on that: I'm not sure where this would live in the doc, and it opens the door to a wider class of hard-to-enumerate-but-incredibly-dangerous things people _could_ do in the C++ API (eg - creating their own `Tx`, calling `commit()`, flipping bits in the kv). In particular, they could already break opacity by talking about mid-execution transaction results in member fields of the app frontend, and I think that catching this (and ignoring it!?) would only allow opacity breaches and nothing else?

achamayou: I briefly tried searching for `catch (...)` but quickly gave up for obvious and yet sad reasons. I'm not sure it's as rare as that (it was certainly plentiful in some code bases I've worked on that used a mix of RogueWave and STL), but I can't prove it. As often/always, it's not so much someone writing this that's likely to happen first, it's someone using a library that does, and passing a callback/functor, which will be unexpectedly wrapped in `catch (...)`. That's why it's often worth having an FAQ keyed by symptoms and offering helpful suggestions. "You see weird thing A:  check you've got the right B".

It may be good to at least spell out positive guidelines/best practices such as keep all state in the KV and nowhere else, I'm not sure that avoiding talking about the limitations of our programming model altogether is such a good thing. It also seems to me that mentioning the existence of the foot gun in an open source project creates more opportunity for a passer-by to propose a way to remove it.

Anyway, that's all fluffy opinion, I've got no facts. I do think making not inherit a standard exception type is an excellent idea in any case.

src/kv/store.h
eddyashton: This check is now `if (source_version > target_version)` - we always expect to copy private maps _from an earlier (or equal) version_, meaning the versions in the rolls we swap in should correctly be lower than the target's current version.

src/kv/test/kv_test.cpp
eddyashton: There's now a comment on this:

```
      // This should throw a CompactedVersionConflict error
```

I prefer try catching the whole block to indicate what the expected behaviour is in _real_ code. We don't want anyone try-catching around individual `get_view`s, we want to catch and retry the entire transaction.

Issue #278, opened 2019-08-05
Better format for JSON byte arrays


eddyashton:
JSON lists of bytes are unreadable. Many of the large arrays we're passing around are really PEMs - these should be passed as strings. Where we have legitimate binary blobs (quotes), they could still be hex-encoded for easier reading/editing.

eddyashton:
#1374 converts our byte-array certificates to JSON strings containing PEM-encodings, but doesn't touch quotes. We should hex- or base64-encode these as well.

achamayou:
`/user_id` needs fixing, it returns a vector of bytes right now, where users expect a PEM string: https://github.com/microsoft/CCF/blob/master/doc/schemas/gov_openapi.json#L1141

achamayou:
List of places in the API that still use vector of bytes that are effectively lists of integers in JSON:

- [ ] receipts
- [ ] ack/update_ack (digest)
- [ ] scripts, when using bytecode

achamayou:
Remaining work captured in follow up tasks listed above.

PR #1621, opened 2020-09-17
Expose response body in Python client similar to request body in JS endpoints

### LINKED ISSUE #1584: Expose response body in Python client similar to request body in JS endpoints
Follow-up from https://github.com/microsoft/CCF/pull/1575.

In JS endpoints, the request body is a `Body` object which is similar to https://developer.mozilla.org/en-US/docs/Web/API/Body, except that it has a synchronous API and only offers `.text()`, `.json()`, and `.arrayBuffer()`.

In the Python client, for HTTP, the response content type determines the type that is returned, see `Response.from_requests_response()`, and for WS, given the lack of headers, an attempt is made to parse JSON with fall-back to return `bytes`.

It probably makes sense to align the handling of response bodies in the Python client to how JS endpoints handle request bodies. This gives the developer more control over how they access the body.

letmaik:
Fixes #1584.

Start reviewing at python/ccf/clients.py.

@achamayou @jumaffre Are there other places except tests that use the python client? Docs?


jumaffre:
LGTM! Thanks for updating this 👍 

It seems that the `/node/tx` endpoint is not too happy:
```
AssertionError: /node/tx should return a JSON object
```

Also, I believe you will also need to update `tutorial.py` in `python/`. To run `tutorial.py` manually (it is only run as part of the daily CI), run `start_test_network.sh ... --network-info-file some_file` and then in a separate terminal `tutorial.py some_file`

jumaffre (approved):

python/ccf/clients.py
```diff
@@ -81,6 +82,60 @@ def makefile(self, *args, **kwargs):
...
+    def json(self) -> Any:
+        pass
+
+    def __len__(self):
+        return len(self.data())
+
+    def __str__(self):
+        try:
+            return self.text()
+        except UnicodeDecodeError:
```
achamayou, line 104: I don't think we should catch that, if a user asks for text, they should get the UnicodeDecodeError and handle it how they want, this is making it very difficult for them to do so.

letmaik: I don't think `str(..)` is ever supposed to throw?

achamayou: No? I can't find a reference either way, this isn't mentioned in https://docs.python.org/3/reference/datamodel.html#object.__str__

At least in 2.x, it definitely used to happen on encoding failures, but that particular behaviour is gone from 3.x.

Could we default to __repr__ on exception perhaps?

achamayou (approved):

achamayou:
@letmaik this is part of the public interface of the ccf python package, so I think we ought to mention the change in release notes.

letmaik:
I'm running the full test suite locally now.
I noticed that /gov/read returns a list of integers for binary data, instead of the raw binary data. Is that some current limitation?

achamayou:
nlohmann::json serialises std::vector<uint8_t> (our internal representation for most raw buffers) as a list of numbers in json: https://github.com/microsoft/CCF/issues/278

python/ccf/clients.py
letmaik: I added the repr fallback.

letmaik:
> 
> 
> nlohmann::json serialises std::vector<uint8_t> (our internal representation for most raw buffers) as a list of numbers in json: #278

Alright, I'll deal with it in the test but I'll open an issue as that's weird.

PR #1626, opened 2020-09-18
Remove uv idle handle

eddyashton:
This swaps `HandleRingbuffer` from being a `uv_idle_t` (which we wrap as `EveryIO`) to a `uv_check_t` (wrapped as the new `AfterIO`).

`uv_idle_t` reduces the UV timeout to 0, so we are constantly executing the (often empty) loop, and hitting 100% useage from the host CPU core. Without any idles, we get a [sensible automatic timeout](http://docs.libuv.org/en/v1.x/design.html#the-i-o-loop) and massively lower CPU%.

The cost of this change is that we increase the delay in processing outbound ringuffer messages, as they only happen after something else has woken up the loop (currently still _fairly_ quick because we have 1ms timer jobs). This may break something unexpected, and ideally we'd find a way to wake the loop every time an outbound message is written.

ADDENDUM: Of course if we want this to happen every ms, we shouldn't rely on something _else_ happening every ms and trust that this is run afterwards - we should just run this itself every ms. That's what we now do: `HandleRingbuffer` is a `Timer` with a 1ms fuse.


achamayou (approved):

src/host/handle_ring_buffer.h
```diff
@@ -65,7 +65,7 @@ namespace asynchost
         });
     }
 
-    void every()
+    void on_timer()
     {
       // On each uv loop iteration...
```
jumaffre, line 70: Comment needs updating.

src/host/after_io.h
```diff
@@ -0,0 +1,46 @@
...
+    {
+      int rc;
+
+      if ((rc = uv_check_init(uv_default_loop(), &uv_handle)) < 0)
+      {
+        LOG_FAIL_FMT("uv_check_init failed: {}", uv_strerror(rc));
+        throw std::logic_error("uv_check_init failed");
+      }
+
+      uv_handle.data = this;
```
ashamis, line 27: Should uv_handle.data be unset if throw on line 32. If it is not unset it is pointing to invalid memory

ashamis (approved):

src/host/after_io.h
eddyashton: This is the same pattern we do for all our libuv handles. I don't think its worth fixing here, and not an issue in practice because any throw from these constructors are fatal. And `uv_handle` should only be accessed if `uv_check_start` succeeded, so I think its fine for it to be invalid in this case. I believe we have an issue in the backlog to do a pass making sure our constructors don't throw/don't leave partially initialised state.

Issue #1559, opened 2020-09-02
Maintain CHANGELOG.md


letmaik:
It would be nice if there was a `CHANGELOG.md` in the repo root in addition to the notes in the GitHub releases. Makes it easier to grep and allows to review changelog changes via PRs.

If the file has a fixed structure (headings = version numbers) then this could be easily integrated with the automated CI release process (https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/github-release?view=azure-devops) to avoid having to copy-paste the notes to the GitHub release each time. The idea would be to run a small script before running the release task that extracts the notes matching the git tag from the Markdown file.

PR #1631, opened 2020-09-21
Exported GitHub release notes to CHANGELOG.md & automate release note extraction

letmaik:
I used https://pypi.org/project/release-exporter/ plus some manual editing.

Fixes https://github.com/microsoft/CCF/issues/1559.


jumaffre:
Awesome! I assume that the Pipelines GitHub release task doesn't automatically pick up new release notes from `Changelog.md`?

letmaik:
Not yet :)

letmaik:
@jumaffre I added the CI automation.

scripts/extract-release-notes.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: Nice! I wonder if the script should also validate that the version number is right. This would prevent us from releasing CCF version `N` with the wrong release notes if CCF version `N-1` is the latest documented release. 

We've already got the `cmake` machinery for this (the CCF version number is derived from the git tag, see `version.cmake`) so we could just add a check for this too?

scripts/extract-release-notes.py
```diff
@@ -0,0 +1,24 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument("out_path")
+args = parser.parse_args()
+
+with open("CHANGELOG.md") as f:
+    changelog = f.read()
+
+marker = "\n## "
+
+idx_unreleased = changelog.find(marker)
```
eddyashton, line 15: This seems error prone. If we rename `Unreleased` to `v1.2.3` in a commit before we make the tag, this will pull the correct section. But if we keep an `Unreleased` section before that, it won't.

scripts/extract-release-notes.py
letmaik: Sounds good, yes.

scripts/extract-release-notes.py
letmaik: My assumption was that the first version is always Unreleased, and then the newest version, which you're about to release. As long as we decide on a convention it should be fine, but I'll also add the suggestion by @jumaffre to validate that the version matches.

letmaik: I've rewritten the script, it now searches for the git version number.

scripts/extract-release-notes.py
```diff
@@ -0,0 +1,40 @@
...
+# Licensed under the Apache 2.0 License.
+
+import argparse
+import re
+import subprocess
+
+parser = argparse.ArgumentParser()
+parser.add_argument("out_path")
+args = parser.parse_args()
+
```
achamayou, line 11: I'd prefer to have this in a main() and to keep import side-effect free. I know it's just a script, but for debugging, some tools etc, it's sometimes nice.

CHANGELOG.md
```diff
@@ -0,0 +1,381 @@
...
+This pre-release implements the genesis model described in the TR, with a distinct service opening phase. See https://microsoft.github.io/CCF/start_network.html for details.
+
+Some discrepancies with the TR remain, and are being tracked under https://github.com/microsoft/CCF/milestone/2
+
+## 0.2 - 2019-06-07
+
+Initial pre-release
+
+
+[Unreleased]: https://github.com/microsoft/CCF/compare/ccf-0.13.3...HEAD
```
achamayou, line 356: What does this do?

achamayou (approved):
Thank you!

CHANGELOG.md
letmaik: Not sure what you mean. It's markdown, and the link shows the diff between the last version and HEAD:

scripts/extract-release-notes.py
letmaik: OK, done.

letmaik:
@achamayou The CI failed for an unrelated issue. This is good to go from my side.

jumaffre:
@letmaik CI is fixed now.

Issue #1627, opened 2020-09-18
AFT: Nodes can collect the required signatures over the merkle root


ashamis:
When running AFT in BFT

- [x] The backups need to produce a signature for every signature send by the primary
- [x] The nodes need to collect the required number of signatures to commit
- [x] The primary needs to create a new transaction that writes the signatures to the ledger

PR #1632, opened 2020-09-21
Aft: In BFT mode backups sign requests

ashamis:
When running in BFT mode every time a backup receives a signature tx from the primary it will verify the signature and then sign the same Merkle root and send it back to the primary.

I am creating a new ProgressTracker class which is responsible to accumulating these signatures and eventually will be responsible for persisting them to the ledger. This persistence will be done in a future commit.

Part of #1627


src/consensus/aft/raft.h
```diff
@@ -825,6 +836,67 @@ namespace aft
...
+        LOG_FAIL_FMT(err.what());
+        return;
+      }
+
+      auto node = nodes.find(r.from_node);
+      if (node == nodes.end())
+      {
+        // Ignore if we don't recognise the node.
+        LOG_FAIL_FMT(
+          "Recv append entries response to {} from {}: unknown node",
```
jumaffre, line 886: Nit: "Recv **signed** append entries response".

src/kv/store.h
```diff
@@ -126,6 +129,17 @@ namespace kv
       history = history_;
     }
 
+    std::shared_ptr<ccf::ProgressTracker> get_progress_tracker()
```
jumaffre, line 132: If the store doesn't use `progress_tracker`, why is it set on the store and not on `AFT` directly?

src/node/progress_tracker.h
```diff
@@ -0,0 +1,179 @@
...
+      {
+        if (seqno == rhs.seqno)
+        {
+          return view < rhs.view;
+        }
+        return seqno < rhs.seqno;
+      }
+    };
+
+    struct Certificate
```
jumaffre, line 148: Minor: Is this the name used in the PBFT terminology? I wonder if this should be evidence as we've already got a few things called certificates in CCF?

src/consensus/aft/raft_types.h
```diff
@@ -141,6 +158,14 @@ namespace aft
     bool success;
   };
 
+  struct SignedAppendEntriesResponse : RaftHeader
+  {
+    Term term;
+    Index last_log_idx;
+    uint32_t signature_size;
```
jumaffre, line 165: Why does the size of the signature need to be passed in if the signature is an `std::array`?

src/node/progress_tracker.h
```diff
@@ -0,0 +1,179 @@
...
+  public:
+    ProgressTracker(kv::NodeId id_, ccf::Nodes& nodes_) : id(id_), nodes(nodes_)
+    {}
+
+    void add_signature(
+      kv::Consensus::View view,
+      kv::Consensus::SeqNo seqno,
+      kv::NodeId node_id,
+      uint32_t signature_size,
+      std::array<uint8_t, MBEDTLS_ECDSA_MAX_LEN>& sig)
```
jumaffre, line 29: Should this be a `ccf::Signature` as defined in `signatures.h`?

src/node/progress_tracker.h
achamayou: Perhaps TxCertificate, TxCert.. ?

ashamis: Sure i can use CommitCert. This is a private struct scoped within ccf::ProgressTracker so i don't think there is much chance of confusion

src/consensus/aft/raft_types.h
ashamis: ECDSA signatures are variable sized. The macro MBEDTLS_ECDSA_MAX_LEN defines the maximum possible length and signature_size is the size of the actual signature. The max size is 141 bytes so its no actually that big

src/kv/store.h
ashamis: Store is where we store history & consensus and this class is meant to have the same scope. Importantly, store is where other classes get access to the history and consensus classes

src/node/progress_tracker.h
ashamis: ccf::Signature is not the best named class because it is more than just a signature it contains other information.

Also this being passed the raw format from the network message. If we just pass this in a different format then we would have to add additional copies.

src/consensus/aft/raft.h
```diff
@@ -779,6 +786,10 @@ namespace aft
               state->view_history.update(state->commit_idx + 1, sig_term);
               commit_if_possible(r.leader_commit_idx);
             }
+            if (consensus_type == ConsensusType::BFT)
+            {
+              send_append_entries_signed_response(r.from_node, sig);
```
achamayou, line 791: This is probably something to discuss later, but we could have a config level where those signatures are produced, sent and recorded just for offline audit but without gating the commit level of the consensus.

src/consensus/aft/raft.h
```diff
@@ -825,6 +836,67 @@ namespace aft
...
+        to,
+        state->last_idx);
+
+      SignedAppendEntriesResponse r = {
+        {raft_append_entries_signed_response, state->my_node_id},
+        state->current_view,
+        state->last_idx,
+        static_cast<uint32_t>(sig.sig.size()),
+        {}};
+      r.sig.fill(0);
```
achamayou, line 853: Why fill with zeros before we copy?

src/consensus/aft/raft.h
```diff
@@ -825,6 +836,67 @@ namespace aft
...
+      SignedAppendEntriesResponse r;
+
+      try
+      {
+        r = channels->template recv_authenticated<SignedAppendEntriesResponse>(
+          data, size);
+      }
+      catch (const std::logic_error& err)
+      {
+        LOG_FAIL_FMT(err.what());
```
achamayou, line 877: Not sure what's in what() but it may be prudent to have a hardcoded FAIL message and have this only in DEBUG, like we've done elsewhere, for confidentiality reasons.

src/consensus/aft/raft_types.h
achamayou: I think Julien's point was that std::array already has a size()?

src/node/progress_tracker.h
```diff
@@ -0,0 +1,179 @@
...
+    };
+    std::map<CertKey, CommitCert> certificates;
+
+    bool verify_signature(
+      kv::NodeId node_id,
+      crypto::Sha256Hash& root,
+      uint32_t sig_size,
+      uint8_t* sig)
+    {
+      kv::Tx tx;
```
achamayou, line 164: Creating a Tx at no particular version doesn't seem right here, I would have thought we'd check all the signatures as of a well defined version instead?

achamayou (approved):

src/node/progress_tracker.h
ashamis: Discussed offline

jumaffre (approved):

PR #1639, opened 2020-09-22
Only send Tick messages to live threads

eddyashton:
Fixes a memory growth issue, where messages would be queued indefinitely for inactive threads.

The real fix is to loop `< thread_count` rather than `: tasks`. The rest of the diff is from pulling the inline lambda into `tick_cb`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- Fixed infinite memory growth issue (#1639)
     
 ## [0.13.3] - 2020-09-18
 ### Added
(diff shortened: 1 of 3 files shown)


jumaffre (approved):

ashamis (approved):

achamayou:
@eddyashton I suggest we inaugurate adding to the changelog on this PR.

eddyashton:
> @eddyashton I suggest we inaugurate adding to the changelog on this PR.

Done. There's a slightly awkward chicken-and-egg issue here - we want to refer to PR numbers in the changelog, but the numbers don't exist until the PR is created. So if we know we're going to put the current change into the Changelog, we need at least one post-PR-creation update to the branch. Minor, but worth noting.

achamayou:
@eddyashton this is less of a problem if we point to issues or documentation, but a good point nonetheless.

PR #1641, opened 2020-09-22
Fix cft election

### LINKED ISSUE #589: Raft persistence bug

achamayou:
Fixes #589

In a nutshell, the change amounts to:

1. Publishing a last committable index on election, and the term of that committable index
2. Resolving the election on that basis, rather than using the last seen commit index and associated term
3. Only rolling back uncommittable transactions, rather than transactions not yet uncommitted


src/consensus/aft/raft.h
```diff
@@ -1253,8 +1264,7 @@ namespace aft
       votes_for_me.clear();
 
       // Rollback unreplicated commits.
```
eddyashton, line 1266: Update this comment

jumaffre:
As `recovery.py` mentions #589, could you remove the check there as well? https://github.com/microsoft/CCF/blob/3ca2946c676ffb3f904a66b95d8ade57b3193922/tests/recovery.py#L59-L63

src/consensus/aft/raft.h
```diff
@@ -1062,7 +1072,8 @@ namespace aft
 
       auto answer = (r.last_commit_term > last_commit_term) ||
```
jumaffre, line 1073: Comment on line `1070` may require updating too.

src/node/rpc/node_frontend.h
```diff
@@ -354,6 +354,30 @@ namespace ccf
...
+          {
+            nlohmann::json n;
+            n["address"] = fmt::format("{}:{}", ninfo.hostname, ninfo.port);
+            c[fmt::format("{}", nid)] = n;
+          }
+          args.rpc_ctx->set_response_body(c.dump());
+        }
+        else
+        {
+          args.rpc_ctx->set_response_status(HTTP_STATUS_NOT_FOUND);
```
jumaffre, line 374: Perhaps add some error message here as well?

src/node/rpc/node_frontend.h
```diff
@@ -354,6 +354,30 @@ namespace ccf
       make_read_only_endpoint("primary", HTTP_HEAD, is_primary)
         .set_forwarding_required(ForwardingRequired::Never)
         .install();
+
+      auto consensus_config = [this](CommandEndpointContext& args) {
+        // Query node for configurations, separate current from pending
+        if (consensus != nullptr)
+        {
+          auto cfg = consensus->get_latest_configuration();
```
jumaffre, line 362: Isn't this awkward during configuration changes when there are more than one configuration active at a time? I think we should print all configurations since we pop unused ones on global commit. The command will still return one configuration in most cases.

src/consensus/aft/test/main.cpp
```diff
@@ -1281,6 +1278,7 @@ DOCTEST_TEST_CASE(
           DOCTEST_REQUIRE(msg.term == 2);
           DOCTEST_REQUIRE(msg.last_commit_idx == 1);
           DOCTEST_REQUIRE(msg.last_commit_term == 1);
+          DOCTEST_REQUIRE(msg.last_committable_idx == 2);
```
jumaffre, line 1281: This is only true because `committable` is set to true in `DOCTEST_REQUIRE(r0.replicate(kv::BatchVector{{2, second_entry, true}}, 1));`. I wonder if we should replicate another non-committable entry after 2?

jumaffre:
The solution in this PR seems simpler than what is described in https://github.com/microsoft/CCF/issues/589#issuecomment-688411856. I think we should record the "why" of the changes in this PR (either in the original ticket or in this PR) as well as the "how" as this is an important change.

src/consensus/aft/raft.h
achamayou: I think it stands, we're just changing the definition of "up to date".

tests/reconfiguration.py
```diff
@@ -9,6 +9,26 @@
 from loguru import logger as LOG
 
 
+def node_configs(network):
+    configs = {}
+    for node in network.nodes:
+        try:
+            with node.client() as nc:
+                configs[node.node_id] = nc.get("/node/config").body.json()
+        except Exception:
```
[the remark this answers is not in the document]
achamayou: Because we don't clean up nodes in the tests, so there's typically dead nodes in there that'll time out.

src/node/rpc/node_frontend.h
achamayou: Yeah right now this is only used after a global commit gate in the tests, but in general we should have the historical version too, I agree.

achamayou:
@eddyashton pointed out that this probably isn't right, and that the election should strictly be based on committable_idx/term(committable_idx). I don't have time just now, but I'll update the PR and describe that case later.

src/node/rpc/node_frontend.h
achamayou: Done, but to be honest this isn't overly likely to happen.

src/consensus/aft/test/main.cpp
achamayou: At the moment the raft tests only deal with committable indices (because the notion of uncommittable index is alien to raft). I don't want to extend them too much because they are quite impractical, and I would like to replace them with a better interface instead.

achamayou:
PR is up to date, please review. There are a few more tests I want to add before it's ready.

src/consensus/aft/raft.h
```diff
@@ -987,9 +993,12 @@ namespace aft
...
       LOG_INFO_FMT("Send request vote from {} to {}", state->my_node_id, to);
 
+      auto last_committable_idx = last_committable_index();
+      CCF_ASSERT(last_committable_idx >= state->commit_idx, "lci < ci");
+
       RequestVote rv = {{raft_request_vote, state->my_node_id},
                         state->current_view,
-                        state->commit_idx,
+                        last_committable_idx,
                         get_term_internal(state->commit_idx)};
```
eddyashton, line 1002:
```suggestion
                        get_term_internal(last_committable_idx)};
```

eddyashton (approved):

tests/rotation.py
```diff
@@ -0,0 +1,56 @@
...
+            test_suspend_primary(network, args)
+
+
+if __name__ == "__main__":
+
+    def add(parser):
+        parser.add_argument(
+            "-p",
+            "--package",
+            help="The enclave package to load (e.g., libsimplebank)",
```
jumaffre, line 50: Super minor: `libsimplebank` isn't a thing anymore.

jumaffre (approved):

tests/rotation.py
achamayou: Haha, yes, will fix.

achamayou:
> As `recovery.py` mentions #589, could you remove the check there as well?
> 
> https://github.com/microsoft/CCF/blob/3ca2946c676ffb3f904a66b95d8ade57b3193922/tests/recovery.py#L59-L63

Yep, testing that now.

achamayou:
Interestingly, the rotation test has failed in CI in an unexpected way: during node addition:

`
45: Traceback (most recent call last):
45:   File "/mnt/build/2/s/tests/rotation.py", line 56, in <module>
45:     run(args)
45:   File "/mnt/build/2/s/tests/rotation.py", line 35, in run
45:     reconfiguration.test_add_node(network, args)
45:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 20, in wrapper
45:     return func(*args, **kwargs)
45:   File "/mnt/build/2/s/tests/reconfiguration.py", line 47, in test_add_node
45:     new_node = network.create_and_trust_node(args.package, "localhost", args)
45:   File "/mnt/build/2/s/tests/infra/network.py", line 506, in create_and_trust_node
45:     self.wait_for_all_nodes_to_catch_up(primary)
45:   File "/mnt/build/2/s/tests/infra/network.py", line 653, in wait_for_all_nodes_to_catch_up
45:     assert len(caught_up_nodes) == len(
45: AssertionError: Only 2 (out of 3) nodes have joined the network
33/49 Test #45: rotation_test ....................***Failed   87.37 sec
`

achamayou:
Looking at the test failure, the election change invalidates our current view history reconstruction logic. In particular, we can no longer assume that commit index is in the term being replayed, and we cannot issue signature transactions until we have "caught up" with previous terms.

I am making necessary changes to address this and will change the PR status again once it is ready.

achamayou:
raft_test is still failing because we replicate transactions marked as committable but that do not deserialise as signatures, so committable indices aren't correct on followers, and now that our term history depends on committable rather than committed indices, it's also inaccurate. Will fix tomorrow.

src/consensus/aft/test/main.cpp
```diff
@@ -1040,366 +1042,4 @@ DOCTEST_TEST_CASE("Exceed append entries limit")
...
-    DOCTEST_CHECK(r1.get_term() == 2);
-    DOCTEST_CHECK(r1.get_commit_idx() == 2);
-    DOCTEST_CHECK(r1.get_last_idx() == 3);
-
-    DOCTEST_CHECK(r2.get_term() == 2);
-    DOCTEST_CHECK(r2.get_commit_idx() == 2);
-    DOCTEST_CHECK(r2.get_last_idx() == 3);
-  }
-}
+}
```
achamayou, line 1045: Axed in favour of the end to end committable.py for now. The existing raft test scaffolding doesn't have a way to accurately replicate signatures at the moment.

src/consensus/aft/raft.h
```diff
@@ -74,6 +74,7 @@ namespace aft
 
     ReplicaState replica_state;
     std::chrono::milliseconds timeout_elapsed;
+    kv::Version election_index = 0;
```
eddyashton, line 77: Can you add a comment explaining what this index is?

src/node/history.h
```diff
@@ -676,67 +676,72 @@ namespace ccf
...
-          auto consensus = store.get_consensus();
-          if (consensus != nullptr && consensus->type() == ConsensusType::BFT)
-          {
-            auto progress_tracker = store.get_progress_tracker();
-            CCF_ASSERT(
-              progress_tracker != nullptr, "progress_tracker is not set");
-            auto r = progress_tracker->record_primary(
-              txid.term, txid.version, id, root, hashed_nonce);
-            if (r != kv::TxHistory::Result::OK)
+      if (signable_txid.has_value())
```
eddyashton, line 681: Minor: This would be match the pattern above (and a smaller diff) with a comment, negative check and early return. ie

```
// Signatures are only emitted when the consensus knows what is safely signable
if (!signable_txid.has_value())
{
  return;
}
```

tests/committable.py
```diff
@@ -0,0 +1,63 @@
...
+    with infra.network.network(
+        hosts, args.binary_dir, args.debug_nodes, args.perf_nodes, pdb=args.pdb
+    ) as network:
+        network.start_and_join(args)
+        primary, backups = network.find_nodes()
+
+        # Suspend three of the backups to prevent commit
+        backups[1].suspend()
+        backups[2].suspend()
+        backups[3].stop()
```
eddyashton, line 23: What's the difference between `suspend()` and `stop()` here? (We suspend 2 and stop 1?)

tests/committable.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # Kill the primary, restore other backups
+        primary.stop()
+        backups[1].resume()
+        backups[2].resume()
+        new_primary, new_term = network.wait_for_new_primary(primary.node_id)
+        LOG.debug(f"New primary is {new_primary.node_id} in term {new_term}")
+        assert new_primary.node_id == backups[0].node_id
+
+        # Check that uncommitted but committsble suffix is preserved
```
eddyashton, line 44:
```suggestion
        # Check that uncommitted but committable suffix is preserved
```

tests/committable.py
```diff
@@ -0,0 +1,63 @@
...
+
+        txs = []
+        # Run some transactions that can't be committed
+        with primary.client("user0") as uc:
+            for i in range(10):
+                txs.append(
+                    uc.post("/app/log/private", {"id": 100 + i, "msg": "Hello world"})
+                )
+
+        # Wait for a signature to ensure those transactions are committable
```
eddyashton, line 33: This kind of sleep is something we do in several places. It now potentially doesn't do what was intended - if we're in the fancy election state and Raft is not returning a `signable` index, we're simply skipping signatures and not producing them at regular intervals.

This is an argument in favour of continuing to produce signatures (making a max-lengthed block of transactions _committable_) without any global commit claim in them, _but_ that's uglier for ledger parsing. I'm happy to keep the current behaviour (signatures are sometimes skipped) for this PR and discuss alternative approaches offline/in future.

eddyashton (approved):

src/consensus/aft/raft.h
achamayou: Sure

src/node/history.h
achamayou: Good point

tests/committable.py
achamayou: Suspend allows resuming, stop is killing the node. We do not need 3 anymore, since we want only three nodes to force a unanimous election, and it's unnecessary to operate after that. We only ever start backup 4 because we want f=2.

tests/committable.py
achamayou: Yes that's true although that will not cause the test to silently pass.

The issue here is that we do not have a way to observe what's committable, the real condition would be, the last of those tx is committable on backup[1]. Even if we shortened the sigx-tx-interval, we wouldn't know that a signature was produced and replicated to backup[1].

tests/committable.py
achamayou: Yes thank you.

achamayou:
Reconfiguration suite failure in nightly

```
45: 20:22:39.960 | INFO     | reconfiguration:node_configs:17 - 200 {"0":{"address":"127.65.38.252:45699"},"1":{"address":"127.114.112.214:42871"},"2":{"address":"127.114.247.179:39983"},"3":{"add + 69 chars
45: 20:22:39.960 | ERROR    | __main__:run:88 - Test reconfiguration.test_retire_primary failed
45: Traceback (most recent call last):
45: 
45:   File "/mnt/build/2/s/tests/e2e_suite.py", line 163, in <module>
45:     run(args)
45:     │   └ Namespace(app_script=None, binary_dir='.', consensus='cft', debug_nodes=[], domain=None, enclave_type='release', enforce_reqs...
45:     └ <function run at 0x7fe830d0d1f0>
45: 
45: > File "/mnt/build/2/s/tests/e2e_suite.py", line 78, in run
45:     new_network = test(network, args)
45:                   │    │        └ Namespace(app_script=None, binary_dir='.', consensus='cft', debug_nodes=[], domain=None, enclave_type='release', enforce_reqs...
45:                   │    └ <infra.network.Network object at 0x7fe8292eb3d0>
45:                   └ <function test_retire_primary at 0x7fe8293598b0>
45: 
45:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 20, in wrapper
45:     return func(*args, **kwargs)
45:            │     │       └ {}
45:            │     └ (<infra.network.Network object at 0x7fe8292eb3d0>, Namespace(app_script=None, binary_dir='.', consensus='cft', debug_nodes=[]...
45:            └ <function test_retire_primary at 0x7fe8293599d0>
45: 
45:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 43, in wrapper
45:     return func(network, args, *nargs, **kwargs)
45:            │    │        │      │        └ {}
45:            │    │        │      └ ()
45:            │    │        └ Namespace(app_script=None, binary_dir='.', consensus='cft', debug_nodes=[], domain=None, enclave_type='release', enforce_reqs...
45:            │    └ <infra.network.Network object at 0x7fe8292eb3d0>
45:            └ <function test_retire_primary at 0x7fe829359940>
45: 
45:   File "/mnt/build/2/s/tests/reconfiguration.py", line 125, in test_retire_primary
45:     pre_count = count_nodes(node_configs(network))
45:                 │           │            └ <infra.network.Network object at 0x7fe8292eb3d0>
45:                 │           └ <function node_configs at 0x7fe8293569d0>
45:                 └ <function count_nodes at 0x7fe829356a60>
45: 
```

PR #1643, opened 2020-09-22
Ensure that we cannot send message to an "unused" threads queue

ashamis:
Always call get_task when getting and task queue and then have an assert in said function to ensure the queue we are returning will eventually be polled

### DIFF
--- a/src/ds/thread_messaging.h
+++ b/src/ds/thread_messaging.h
@@ -267,7 +267,7 @@ namespace threading
 
     void run()
     {
-      Task& task = tasks[get_current_thread_id()];
+      Task& task = get_task(get_current_thread_id());
 
       while (!is_finished())
       {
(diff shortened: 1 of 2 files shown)


src/ds/thread_messaging.h
```diff
@@ -267,16 +267,21 @@ namespace threading
...
       {
         task.run_next_task();
       }
     }
 
-    Task& get_task(uint16_t tid)
+    inline Task& get_task(uint16_t tid)
     {
+      CCF_ASSERT_FMT(
+        tid <= thread_count,
```
eddyashton, line 281: Should this be `<` rather than `<=`? If we have 3 threads, the highest valid (0-indexed) `tid` is `2`?

ashamis, line 281:
```suggestion
        tid < thread_count,
```

ashamis: thank you. Fixed

eddyashton (approved):

PR #1650, opened 2020-09-24
Updates to VSCode debugging

eddyashton:
Fixing 2 things I noticed while trying to use the VSCode debugging:

- Symbols for virtual enclaves weren't loaded so they weren't debuggable. Semi-fixed by adding an `additionalSOLibSearchPath`, but this makes assumptions on where the build directory is. Should be fixable with a smarter `cwd`, but I haven't cracked it yet.
- The `binary_dir` in Python isn't always writable - when using an installed CCF its intended that we only _read_ from the CCF binary directory, which is not the current build/workspace - so we can't write `vscode-gdb.sh` there. What I really want to do here is have the `inputs` in `launch.json` ask for the target node directory. We can reliably write a single node debug command there, and then the debugger configuration can load that and whatever else it needs. But I can't find a suitable `command` which _just_ opens the "Open Folder" dialog (ideally from a sensible root), and doesn't try to then _open_ that folder in a new window. So I gave up, and we use `/tmp` as a predictable, writable home for a temporary helper file.


achamayou (approved):

letmaik (approved):

letmaik:
Thanks for fixing the virtual mode debugging... I was hitting the same issue and thought I'm going mad.

PR #1651, opened 2020-09-24
Add JS app bundle support

### LINKED ISSUE #1642: JS app bundle & deploy_js_app proposal
Deployment of JS apps is currently a bit fiddly, mostly because there are multiple low-level steps involved.

I propose to introduce the concept of a JS app bundle, which is a folder containing two things:
- `endpoints.json`
- `src/` (tree of .js files)

This folder can optionally be zipped up, at which point it becomes a JS app bundle archive. Both the folder and the zip file (useful for a developer/CI to deliver the built app to the operator for deployment) are accepted by CCF's Python package in a new proposal type `deploy_js_app` (this may or may not be a collection of existing low-level proposals).

An initial sketch of `endpoints.json` is here: https://github.com/microsoft/CCF/blob/master/tests/npm-tsoa-app/endpoints.json.

The idea of this bundle format is that it is simple enough for manually written code and also suits itself for more advanced scenarios like npm-based typescript apps with dependencies.

Does this make sense?

letmaik:
Fixes #1642.

I originally planned to only change the Python client but I realized that I might as well do it properly.

~~This also allows deploying multiple apps. It's based on using an *app name* to construct the JS module prefix and the URL prefix. However, there is a special case: the root app. It has the magic name `__root__` and is exposed without URL prefix. By giving the root app a name like all other apps, the same code for adding/removing apps can be used and essentially everything is namespaced and contained. If an endpoint of the root app conflicts with a non-root app prefix then the non-root app wins. An alternative would be to forbid conflicts and let deployment fail. I expect this to happen very rarely though.~~

Follow-ups after this PR:
- Deprecate/remove the `set_js_app` proposal type.
- ~~Add non-npm JS sample which is a hand-written JS bundle, mostly for documentation purposes.~~ done
- Remove requirement for inline endpoint script.
- Store per-endpoint metadata (e.g. readonly), currently ignored.


src/node/rpc/member_frontend.h
```diff
@@ -144,9 +145,60 @@ namespace ccf
...
+    std::string name;
+    JsBundle bundle;
+  };
+  DECLARE_JSON_TYPE(DeployJsApp)
+  DECLARE_JSON_REQUIRED_FIELDS(DeployJsApp, name, bundle)
+
   class MemberEndpoints : public CommonEndpointRegistry
   {
   private:
+    static bool startswith(const std::string& s, const std::string& prefix)
```
letmaik, line 184: If there's a better place for such utility functions, let me know.

eddyashton: `ds/nonstd.h`

src/ds/nonstd.h
```diff
@@ -62,6 +62,14 @@ namespace nonstd
   template <class T>
   using remove_cvref_t = typename remove_cvref<T>::type;
 
+  /** a more generic std::string member function is present in C++20
+   */
+  static inline bool starts_with(
+    const std::string& s, const std::string& prefix)
+  {
+    return s.rfind(prefix, 0) == 0;
```
eddyashton, line 70: Minor: We have a few other places in the code that do this inline. Can you please replace them with `nonstd::starts_with`?

src/node/rpc/member_frontend.h
```diff
@@ -192,6 +227,96 @@ namespace ccf
...
+              info.js_module);
+            return false;
+          }
+
+          // CCF currently requires each endpoint to have an inline JS module.
+          std::string method_uppercase = method;
+          nonstd::to_upper(method_uppercase);
+          std::string key =
+            fmt::format("{} {}{}", method_uppercase, url_prefix, url);
+          std::string script = "import {" + info.js_function + " as f}" +
```
eddyashton, line 273: Can you build this with `fmt::format`?

tests/npm-tsoa-app/src/ccf.ts
```diff
@@ -0,0 +1,27 @@
+export interface Body<T> {
```
eddyashton, line 1: Can we put this somewhere shared, rather than repeating it in each app?

letmaik: That's a tricky one. I thought about it, but I'd like to keep the test apps self-contained so that someone can just copy-paste them. I don't want to go down the path (yet) of creating a ccf npm package which could contain this stuff.

src/node/rpc/member_frontend.h
letmaik: If you tell me how to escape curly braces, then yes :) Keep in mind that this is temporary code that will hopefully be killed soon, and it's not in the critical path.

eddyashton: I believe its `{{` to produce a literal `{`.

tests/npm-tsoa-app/src/ccf.ts
eddyashton: Ok - the copy-paste argument wins me over for now.

eddyashton (approved):

letmaik:
After chatting with @achamayou we decided to not add support for multiple apps. I'll remove that again from this PR.

achamayou:
To clarify, I am not against this, I am just worried that we don't really have a well-defined plan for how to support OpenAPI and in particular shared types across multiple apps. We talked about namespacing, but it's somewhat fuzzy and untested.

In order of priority, I think it's preferable to support single apps with dynamic endpoints and OpenAPI, and to extend that to multiple apps later, than it is to add support for multiple apps if we don't have a good plan for OpenAPI.

Issue #1648, opened 2020-09-24
Support for historical queries after recovery from snapshot


jumaffre:
When resuming from a snapshot or after recovery, a node doesn't have the history of the ledger secrets as ledger secrets are stored _at the same key_ in the `ccf.shares` table. Assuming that the historical ledger is later moved to that node (see https://github.com/microsoft/CCF/issues/1527), if this node has to serve historical queries, it won't be able to decrypt or verify the integrity of these historical entries. 

To be able to support historical queries, the node should process the re-sharing entries in the ledger backwards, from the last available re-sharing entry (available in the snapshot). **Caveat here: on recovery, the re-sharing entry is written at a version later than the one at which the associated ledger secret is applicable from so this mechanism should account for that**. 

Ledger secrets can be decrypted one-by-one using the previously restored ledger secret. There is no need to restore ledger secret prior to the latest re-sharing entry prior to the start index of the historical query and no need to keep in memory ledger secrets more recent than the first one after the end index of the historical query (included).

achamayou:
A slightly tangent point that came up in discussion but may be helpful in implementation is that reading through historical secrets going back can be quite quick (each one contains the version of its predecessor) and does not require ledger integrity validation because ledger secrets are encrypted and integrity protected with AES GCM.

A possible optimisation, should this turn out to be a significant cost, may be to include the last X ledger secrets in snapshots (similarly to what's being done for the term history).

jumaffre:
Removing checkpoint milestone and re-phrasing issue slightly as this is not snapshot-specific but also applies to recoveries.

jumaffre:
FYI, the minimal test suite to reproduce the issue is:
```python
suite_rekey_recovery = [
    e2e_logging.test_historical_query,
    rekey.test,
    rekey.test,
    recovery.test, # Assuming this starts from snapshot
]
```

The recovery test will fail verifying historical queries emitted before the first rekey.

Issue #1608, opened 2020-09-14
End-to-end testing of ledger between nodes


jumaffre:
As part of our end-to-end infrastructure, we should check that the ledger files produced by different nodes are identical at the end of each run. This is important especially that snapshots now trigger the creation of a new ledger chunk.

PR #1656, opened 2020-09-25
Recovery from snapshot

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1303

This PR adds support for resuming from a snapshot on recovery, making the whole recovery procedure a lot faster (from 1 minute to 11 seconds in this example):
```
# Start a liblogging service ...

# Issue 5000 transactions at the same key (i.e. long ledger but tiny state)
$ for ITER in {1..5000}; do curl https://127.212.141.250:38787/app/log/private -H "Content-Type: application/json" --data '{"id":1,"msg":"xyz"}' ...  ; done
...
x-ccf-global-commit: 5304
x-ccf-tx-seqno: 5316
x-ccf-tx-view: 2

# Kill the service. Copy ledger and snapshot directories and network encryption key.

# Recover service only using ledger (i.e. no snapshot): ~1 minute
$ ../start_test_network.sh -p liblogging --recover ...
Python environment successfully setup
[15:02:38.758] Starting 3 CCF nodes...
[15:02:38.758] Recovering network from:
[15:02:38.758]  - Defunct network public encryption key: ./network_enc_pubk.pem
[15:02:38.758]  - Common directory: ./workspace/test_network_common/
[15:02:38.758]  - Ledger: 0.ledger/
[15:02:38.758] No available snapshot to recover from. Entire transaction history will be replayed. <<<<< SLOW!
[15:03:37.505] Started CCF network with the following nodes:
[15:03:37.505]   Node [ 3] = 127.160.143.234:33661
[15:03:37.505]   Node [ 4] = 127.153.80.188:44753
[15:03:37.505]   Node [ 5] = 127.81.23.35:39129

# Same recovery with snapshots: ~11 seconds
$ ../start_test_network.sh -p liblogging --recover ... --snapshot-dir snapshots
[15:07:08.529] Starting 3 CCF nodes...
[15:07:08.529] Recovering network from:
[15:07:08.530]  - Defunct network public encryption key: ./network_enc_pubk.pem
[15:07:08.530]  - Common directory: ./workspace/test_network_common/
[15:07:08.530]  - Ledger: 0.ledger/
[15:07:08.530]  - Snapshots: snapshots/    <<<<< FAST!
[15:07:19.469] Started CCF network with the following nodes:
[15:07:19.469]   Node [ 3] = 127.185.213.5:45741
[15:07:19.469]   Node [ 4] = 127.116.172.203:35685
[15:07:19.469]   Node [ 5] = 127.187.176.189:44623
```

**Implementation details:**
1. The snapshot directory can now be passed to `cchost` on recovery (just like when joining a new network).
2. On recovery, the public part of the snapshot is first applied. Subsequent entries in the ledger are read until the end of the ledger is reached (speed-up part I.). The snapshot is kept in enclave memory for now.
3. Once the recovery shares have been submitted, the ledger secrets from the ones contained in the snapshot till the current ones are decrypted. The private part of the snapshot is applied and subsequent ledger entries are also read (speed-up part II.). The snapshot is then discarded.

**Notes:**
- New `recovery_snapshot_test` e2e test with corresponding changes to the Python infra.
- `start_test_network.sh` can recover from a snapshot.
- Snapshots are not generated when the service is in public mode. This is because such snapshots would have to be generated without direct integrity protection (because the ledger key is not recoverable until the public recovery is finished). We may want to revisit this in the future if this is an issue.
- Getting further snapshots generated by the recovered service at the right intervals requires some juggling of indices, etc. I believe this is working now but will be systematically tested when https://github.com/microsoft/CCF/issues/1608 is worked on.
- If a node has recovered from a snapshot, all other nodes should join from a snapshot that is at least as recent as that recovery snapshot. This is because the entire history of ledger secrets is contained in the ledger. This isn't yet automatically checked on join but will be addressed in #1648.


src/consensus/aft/raft.h
```diff
@@ -205,23 +205,6 @@ namespace aft
       become_leader();
     }
 
-    void force_become_leader(Index index, Term term, Index commit_idx_)
```
jumaffre, line 208: This was unused.

livehtml.sh
```diff
@@ -2,4 +2,4 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache 2.0 License.
 
-sphinx-autobuild -b html doc doc/html -H 0.0.0.0
\ No newline at end of file
+sphinx-autobuild -b html doc doc/html --host 0.0.0.0
```
jumaffre, line 5: Looks like `-H` is deprecated.

src/node/node_state.h
```diff
@@ -314,7 +320,35 @@ namespace ccf
...
+
+          if (from_snapshot)
+          {
+            // Keep a reference to snapshot for private recovery
+            recovery_snapshot =
+              std::make_unique<std::vector<uint8_t>>(args.config.snapshot);
+
+            LOG_DEBUG_FMT(
+              "Deserialising public snapshot ({})", recovery_snapshot->size());
+            auto rc = network.tables->deserialise_snapshot(
```
jumaffre, line 338: I had to change the interface to `deserialise_snapshot()` to return the `view_history` contained in the snapshot rather than set it on the consensus directly. This is because here, to have a consensus, we need a `State` and to have a `State`, we need to our node ID... which we don't have until we have deserialised the snapshot. We could add setters on the state to set the node ID after having created the consensus but I think this solution is neater.

src/node/node_state.h
```diff
@@ -164,12 +164,19 @@ namespace ccf
     std::shared_ptr<kv::Store> recovery_store;
     std::shared_ptr<kv::TxHistory> recovery_history;
     std::shared_ptr<kv::AbstractTxEncryptor> recovery_encryptor;
+
     kv::Version recovery_v;
     crypto::Sha256Hash recovery_root;
-    std::vector<kv::Version> term_history;
-    kv::Version last_recovered_commit_idx = 1;
+    std::vector<kv::Version> view_history;
+    kv::Version last_recovered_signed_idx = 1;
```
jumaffre, line 171: Some harmless renaming here and throughout the file.

src/kv/store.h
```diff
@@ -394,11 +399,10 @@ namespace kv
...
       }
 
-      std::vector<Version> view_history;
-      auto c = get_consensus();
-      if (c)
+      std::vector<Version> view_history_;
+      if (view_history)
       {
-        view_history = d.deserialise_view_history();
+        view_history_ = d.deserialise_view_history();
```
ashamis, line 405: minor: can you simplify the code by passing view_history by reference (into the funtion) as you always want to return the view history and then just deserializing into view_history?

src/enclave/interface.h
```diff
@@ -40,6 +40,9 @@ struct CCFConfig
   std::string domain;
   size_t snapshot_tx_interval;
 
+  // Only if joining or recovering
```
ashamis, line 43: nic: can you please expand on this comment.

src/node/rpc/member_frontend.h
```diff
@@ -1241,11 +1241,13 @@ namespace ccf
           submitted_shares_count = share_manager.submit_recovery_share(
             args.tx, args.caller_id, raw_recovery_share);
         }
-        catch (const std::logic_error& e)
+        catch (const std::exception& e)
         {
+          auto error_msg =
```
ashamis, line 1246: Can you please change this to follow the pattern we are using for other errors we catch
have an error log that prints somethign went wrong, followed by a debug log that prints e.what()
This ensures if there is something confidentual in e.what() we do not leak it

src/node/rpc/member_frontend.h
```diff
@@ -1268,13 +1270,15 @@ namespace ccf
         {
           node.initiate_private_recovery(args.tx);
         }
-        catch (const std::logic_error& e)
+        catch (const std::exception& e)
         {
           // For now, clear the submitted shares if combination fails.
+          auto error_msg =
+            fmt::format("Failed to initiate private recovery: {}", e.what());
+          LOG_FAIL_FMT(error_msg);
```
ashamis, line 1278: Same as above regarding the logs

src/node/rpc/member_frontend.h
```diff
@@ -1268,13 +1270,15 @@ namespace ccf
...
           share_manager.clear_submitted_recovery_shares(args.tx);
           args.rpc_ctx->set_response_status(HTTP_STATUS_INTERNAL_SERVER_ERROR);
-          args.rpc_ctx->set_response_body(
-            fmt::format("Failed to initiate private recovery: {}", e.what()));
+          args.rpc_ctx->set_response_body(std::move(error_msg));
```
ashamis, line 1281: Since we are catching everything should be be sending it back to the user? this might leak confidential info

CMakeLists.txt
```diff
@@ -527,6 +527,22 @@ if(BUILD_TESTS)
...
 
+  add_e2e_test(
+    NAME recovery_snapshot_test
+    PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/recovery.py
+    CONSENSUS cft
+    ADDITIONAL_ARGS
+      --recovery
+      2
+      # Shorten Raft election timeout to speed up test when it kills a node on
+      # purpose to check that a recovery network is robust to a view change.
```
achamayou, line 538: We haven't a lot of timeouts due to quote verification since the dcap client started caching more aggressively, so I wonder if we shouldn't switch to a shorter election timeout across all our tests anyway.

src/enclave/interface.h
achamayou: Unless I'm missing something, this is where we store the snapshot we start/restore from, hence only on joining or recovering. That's hopefully something we can reflect in the name, maybe 'startup_snapshot' or 'snapshot_to_initialise_from'...

src/node/node_state.h
```diff
@@ -166,12 +166,19 @@ namespace ccf
...
     std::list<RecoveredLedgerSecret> recovery_ledger_secrets;
 
+    // This should really be a reference but because there are two ecalls,
+    // this is currently not possible:
+    // https://github.com/microsoft/CCF/issues/857
```
achamayou, line 178: Maybe it's time for 857 then.

achamayou (approved):

src/kv/store.h
jumaffre: I don't think we want to pass a reference here as we don't necessarily want to deserialise the view history. If the source KV store doesn't have a consensus, then the view history won't be included in the snapshot. This is useful to test snapshots on the KV in isolation of other components.

src/enclave/interface.h
jumaffre: Good point. Renamed to `startup_snapshot`.

src/node/rpc/member_frontend.h
jumaffre: Very good point! I was definitely useful for debugging but should not be included in default logging.

CMakeLists.txt
jumaffre: Yes. I'll lower this in a separate PR so we can test the stability of this in isolation.

Issue #1597, opened 2020-09-10
Improve operating member update process


achamayou:
The operating member constitution sample (https://microsoft.github.io/CCF/master/members/constitution.html#constitution) currently makes use of the constitution to decide which members are operating members: https://github.com/microsoft/CCF/blob/master/src/runtime_config/operator_gov.lua#L16

This means that updating the set of operating members requires a constitution replacement, which is a fairly heavy operation.

We want to improve on that process to make it less heavy, and in particular to make it possible for an operator to effectively replace itself, so it can roll its identity key without needing involvement from more senior members.

achamayou:
Summary of team discussion on this matter: this can be implemented with two changes to the current governance flow.

Firstly, adding a set_member_data/member_data analog to set_user_data/user_data, allowing an arbitrary JSON blob to be associated with a member identity. This can contain an "is_operator" property, which the constitution can use to distinguish operating members from higher privilege operators.

Secondly, passing as an extra argument to the constitution script the identity of the member who created the proposal being evaluated.

achamayou:
Roughly speaking, it would allow the following addition to https://github.com/microsoft/CCF/blob/master/src/runtime_config/operator_gov.lua

```
tables, calls, votes, proposer_id = ...

...

proposer_is_operator = tables["ccf.members"].get(proposer_id).user_data.is_operator

...

  -- a vote is an operator vote if it's only making operator calls
  operator_vote = true
  for _, call in pairs(calls) do
    if not operator_calls[call.func] then
      
      # Adding and removing operators is an operator call
      if (call.func == "add_member" or call.func == "remove_member") and call.args[2].is_operator then
        break
      end

      operator_vote = false
      break
    end
  end

...
```

PR #1657, opened 2020-09-25
Add generic member_data to simplify operator governance

eddyashton:
As described in #1597.

This adds a `member_data` field for each member, in the same pattern as `user_data`, storing arbitrary JSON for the governance scripts to interpret how they wish. `operator_gov.lua` is updated to use this, creating an object with `is_operator = true` for operating members, rather than hardcoding them in the constitution. 

#1597 mentioned exposing the proposer's ID to the constitution, but I don't think this is actually necessary. We don't currently block proposals on any grounds, so we'd have to add that if we wanted to gate the _creation_ of specific proposals to members/non-members. Instead we care about the operatoricity of each _voter_. In the case of proposals submitted by operators (which are freely submitted like any other members'), their single yes vote is sufficient to pass an 'operator proposal' (one of the constitution-defined calls), but this works purely by vote-iteration and doesn't need to consider the proposer separately.


tests/memberclient.py
```diff
@@ -38,8 +38,12 @@ def test_set_recovery_threshold(network, args, recovery_threshold=None):
 def test_add_member(network, args):
     primary, _ = network.find_primary()
 
+    member_data = {"name": "test_added_member", "roles": ["OWNER", "CONTRIBUTOR"]}
```
jumaffre, line 41: Super minor: When reading this, I genuinely thought that `OWNER` and `CONTRIBUTOR` were actual roles with specific semantics in CCF. For anyone (e.g. CCF operators) reading this test in isolation (e.g. to find examples of governance operations), could we replace these with dummy strings so that it's clear that these roles don't actually exist?

src/node/rpc/test/member_voting_test.cpp
```diff
@@ -1478,9 +1491,60 @@ DOCTEST_TEST_CASE("Passing operator vote")
     DOCTEST_CHECK(proposer_vote != votes.end());
     DOCTEST_CHECK(proposer_vote->second == vote_for);
   }
+
+  auto new_operator_kp = tls::make_key_pair();
```
jumaffre, line 1495: Should we also add tests for the negative paths, i.e. operator member trying to add or retire a non-operator member?

jumaffre (approved):

achamayou (approved):

src/node/rpc/test/member_voting_test.cpp
eddyashton: Good call - I've added a few negative tests as well.

Issue #1655, opened 2020-09-25
AFT: Collect and commit evidence of signature validation


ashamis:
- [x] Nodes send acknowledgement that they received 2f+1 signatures
- [x] Nodes receives 2f+1 acknowledgements
- [x] nodes send acknowledgements of acknowledgements when they receive 2f+1 acknowledgements
- [x] Primary should write the acknowledgements of acknowledgements to the ledger
- [x] Come up with better names for these messages

PR #1658, opened 2020-09-27
Aft: In BFT mode propagate backup signatures and backups to signatures 

ashamis:
Part of #1655, #1627

In this PR AFT backup nodes will send their signatures of the primary signed merkle root to the other replicas and propagate their their acks of receiving said signatures.
The next PR will involve writing this info to the ledger. This should allow for a node to catchup that is behind the 2f+1 up to date bft nodes.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -871,7 +871,7 @@ namespace aft
       auto h = progress_tracker->get_my_hashed_nonce(
         state->current_view, state->last_idx);
 
-      std::array<uint8_t, 32> hashed_nonce;
+      Nonce hashed_nonce;
       std::copy(h.begin(), h.end(), hashed_nonce.begin());
 
       SignedAppendEntriesResponse r = {
@@ -898,7 +898,8 @@ namespace aft
         auto send_to = it->first;
         if (send_to != state->my_node_id)
         {
-          channels->send_authenticated(ccf::NodeMsgType::consensus_msg, send_to, r);
+          channels->send_authenticated(
+            ccf::NodeMsgType::consensus_msg, send_to, r);
         }
       }
     }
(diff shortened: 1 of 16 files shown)


src/node/history.h
```diff
@@ -576,22 +576,36 @@ namespace ccf
...
+      if (!verify(term, &sig))
       {
-        return false;
+        return kv::TxHistory::Result::FAIL;
+      }
+
+      kv::TxHistory::Result result = kv::TxHistory::Result::OK;
+
+      auto progress_tracker = store.get_progress_tracker();
+      if (progress_tracker)
```
jumaffre, line 590: Should `verify_and_sign()`return an error if `progress_tracker` is not set?

src/consensus/aft/raft.h
```diff
@@ -334,6 +334,11 @@ namespace aft
       return configurations.back().nodes;
     }
 
+    uint32_t node_count() const
+    {
+      return nodes.size();
```
achamayou, line 339: We have to be careful with this, because this is the union of all currently active nodes, it is not necessarily equal to the number of nodes in the current configuration. If for example we're reconfiguring from a 4 node network to another 4 node network, this could be 7 (or any value between 5 and 8).

src/consensus/aft/raft.h
```diff
@@ -845,22 +858,41 @@ namespace aft
         to,
         state->last_idx);
 
+      auto progress_tracker = store->get_progress_tracker();
+      auto h = progress_tracker->get_my_hashed_nonce(
+        state->current_view, state->last_idx);
+
+      Nonce hashed_nonce;
```
achamayou, line 865: no copy ctor on Nonce?

src/consensus/aft/raft.h
```diff
@@ -893,8 +925,184 @@ namespace aft
...
-        progress_tracker->add_signature(
-          r.term, r.last_log_idx, r.from_node, r.signature_size, r.sig);
+        auto result = progress_tracker->add_signature(
+          r.term,
+          r.last_log_idx,
+          r.from_node,
+          r.signature_size,
+          r.sig,
+          r.hashed_nonce,
+          nodes.size());
```
achamayou, line 935: I suspect this value is not what we want during reconfigurations.

src/consensus/aft/raft.h
```diff
@@ -1461,7 +1669,7 @@ namespace aft
 
       for (auto node_id : to_remove)
       {
-        if (replica_state == Leader)
```
achamayou, line 1464: We probably ought to move to a more detailed state, which would have a role (Leader/Follower/...) but also properties (executes, signs, ...) which can be set independently. The protocol variant would define which role has which proprties.

achamayou (approved):

src/node/history.h
ashamis: Yes good catch. This should just not work if it progress_tracker is not set. This is an invalid setup.

src/consensus/aft/raft.h
ashamis: Nonce is currently an alias for std::array. I am trying to resist creating a new type but this might change in the near future

PR #1659, opened 2020-09-27
Implement remove operation for CHAMP map

### LINKED ISSUE #1623: Champ map should support entry deletion
We currently keep tombstone entries for deleted keys in the champ map (key unchanged, negative version, empty value). This will cause unbounded memory growth as more nodes, members, users, etc. are rotated during the lifetime of the service. This is especially bad for reverse lookup tables for members and users certificates.

We should add support for deleting keys deterministically in the champ. The original paper (https://michael.steindorfer.name/publications/oopsla15.pdf, pp. 746-747) suggests that it should be feasible.

Theodus:
This PR adds support for removing entries from the CHAMP hash map. The CHAMP implementation is now tested against the C++ standard `unordered_map` in order to avoid adding similar support to the red-black tree map. This PR does not include support for returning maps to their canonical form after entries have been removed. The canonical form of the CHAMP structure implies that non-root `SubNode`s (internal nodes) contain only one child. Returning a CHAMP to its canonical form is likely a useful feature to reduce memory consumption and cache misses. However, discussion is needed to determine if this should be implemented in every call to the new remove operation, or if it should be a separate operation to avoid hysteresis on repeated insertion and removal of entries with colliding key prefixes.

Closes #1623


achamayou:
Thanks @Theodus for taking the time to do this. It sounds like we'd probably want to canonicalise the CHAMP on compact(), but we may want some benchmarks before we decide.

/azp run

achamayou (approved):

src/ds/champ_map.h
```diff
@@ -165,6 +182,8 @@ namespace champ
   template <class K, class V, class H>
   struct SubNodes
   {
+    // TODO: Reduce node size by replacing vector with `Node<K, V, H>*`, since
+    // the length is equal to `node_map.pop() + data_map.pop()`
```
eddyashton, line 186: We currently have a policy of 0 TODOs in the code. If this is a small addition, please add it to this PR, otherwise please create a new GitHub issue with the `TODO` tag to track this.

eddyashton (approved):
Thanks so much for doing @Theodus.

With regards to canonicalisation, I don't think we actually _need_ canonicalisation at any point. We want the reduced memory footprint of real deletions (rather than tombstones), but we never (currently!) have a situation where we need `put("foo")` and `put("foo"); put("bar"); delete("bar")` to produce identical champs. We can serialise and deserialise a non-canonical champ, as long as its reasonably compact and has a deterministic serialisation.

If we add canonicalisation we should profile it, but can presumably amortise it during compaction or assume hysteresis-prone removals are rare.

src/ds/champ_map.h
Theodus: Issue opened: https://github.com/microsoft/CCF/issues/1661, and TODO comment removed

achamayou:
@Theodus sorry, the formatter is bit unhappy, could you run `./scripts/ci-checks.sh -f` and that ought to fix it (or give an error if it can't).

PR #1665, opened 2020-09-29
Fix initialisation order memory safety issue from OpenAPI info

eddyashton:
The daily SAN build caught an issue introduced by the OpenAPI change. The ownership chain of structs in `RpcFrontend` construction is non-obvious, and sometimes the references we're passed are unconstructed members of something further down the chain. So we should only assign to them in `open`, rather than in the constructor.


ashamis (approved):

achamayou (approved):
@eddyashton can you kick off the nightly on this change?

eddyashton:
@achamayou: I don't think we can easily run the Daily CI on these as they're on another fork (unless there's a way to do this I've missed?) Would you like me to re-create this PR from a `microsoft/` branch, or just run the Daily once this is merged?

jumaffre:
@eddyashton You can manually kick off the Daily manually against PRs on the main repo by using `refs/pull/1656/merge` as the branch name:
image

achamayou:
The other way to do it is to touch one of the trigger files for the daily like .daily.yml.

eddyashton:
Daily CI [here](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=13251&view=results).

PR #1667, opened 2020-09-29
JS app docs

### LINKED ISSUE #1622: Move all the build-your-JS-app docs to a new dedicated section in sphinx docs
https://microsoft.github.io/CCF/master/developers/index.html

Existing text:
- tests/npm-app/README.md
- tests/npm-tsoa-app/README.md

letmaik:
Fixes #1622.

FYI: GitHub has basic ReST support in the "rich diff" view.


doc/developers/js_app.rst
```diff
@@ -0,0 +1,167 @@
...
+
+    $ tree --dirsfirst my-app
+    my-app
+    ├── src
+    │   └── app.js
+    └── app.json
+
+We also call this a `JavaScript app bundle` since it is the format used for deployment.
+It consists of metadata (``app.json``) and one or more JavaScript modules (``src/``)
+-- more details in later sections.
```
jumaffre, line 29: Better to link to the other sections directly? FYI, we use full paths for references, e.g.:
```
:ref:`issue new business requests <users/issue_commands:Issuing Commands>`
```

doc/developers/js_app.rst
```diff
@@ -0,0 +1,167 @@
...
+.. code-block:: bash
+
+    $ python -m ccf.proposal_generator deploy_js_app my-app/
+    SUCCESS | Writing proposal to ./deploy_js_app_proposal.json
+    SUCCESS | Wrote vote to ./deploy_js_app_vote_for.json
+
+Once accepted, a ``deploy_js_app`` proposal atomically (re-)deploys the complete JavaScript application.
+Any existing application endpoints and JavaScript modules are removed.
+
+In the following sections we go into details of the metadata file and
```
jumaffre, line 46: I think we should try to keep one sentence per line wherever possible?

doc/developers/js_app.rst
```diff
@@ -0,0 +1,167 @@
...
+
+We also call this a `JavaScript app bundle` since it is the format used for deployment.
+It consists of metadata (``app.json``) and one or more JavaScript modules (``src/``)
+-- more details in later sections.
+
+You can find an example app bundle in the
+`tests/js-app-bundle <https://github.com/microsoft/CCF/tree/master/tests/js-app-bundle>`_
+folder of the CCF git repository.
+
+An app bundle can be turned into a proposal with the Python client:
```
achamayou, line 35:
```suggestion
An app bundle can be wrapped into a governance proposal with the Python client for deployment:
```

doc/developers/js_app.rst
```diff
@@ -0,0 +1,167 @@
...
+
+CCF includes a native ``js_generic`` application which can execute JavaScript applications proposed through governance.
+
+CCF's JavaScript environment is similar to that in browsers, with the following differences:
+
+- only `JavaScript modules <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules>`_ are supported,
+- no `Web APIs <https://developer.mozilla.org/en-US/docs/Web/API>`_ are available,
+- module imports cannot be URLs and must be relative paths.
+
+.. _js-app-bundle:
```
jumaffre, line 12: Better to refer `App bundle` header directly (see comment below for an example).

doc/developers/js_app.rst
```diff
@@ -0,0 +1,167 @@
...
+
+.. code-block:: bash
+
+    $ tree --dirsfirst my-app
+    my-app
+    ├── src
+    │   └── app.js
+    └── app.json
+
+We also call this a `JavaScript app bundle` since it is the format used for deployment.
```
jumaffre, line 27: I think we should try to avoid "we" and "I" in this doc, but maybe it's just me ;)

doc/developers/js_app.rst
```diff
@@ -0,0 +1,167 @@
...
+              ...
+            }
+          },
+          ...
+        },
+        ...
+      }
+    }
+
+``"endpoints"`` contains endpoint descriptions nested by REST API URL and HTTP method.
```
jumaffre, line 94: I feel like we should introduce the concept of "endpoints" earlier in this page. Perhaps saying something like: "A JavaScript application is a collection of arbitrary JavaScript endpoints that CCF users can call via HTTP requests. These endpoints can include a variety of JavaScript modules and read and write to CCF's replicated key-value store."?

doc/developers/js_app.rst
```diff
@@ -0,0 +1,167 @@
...
+Metadata
+--------
+
+The ``app.json`` file of an app bundle has the following structure:
+
+.. code-block:: json
+
+    {
+      "endpoints": {
+        "/foo": {
```
jumaffre, line 75: Perhaps explain that the endpoint can be reached at this URI? And maybe include a sample `curl` command and its response for a very basic endpoint?

jumaffre:
Looks very good! Thanks for doing this. 

Related to this, now that we ship `libjs_generic` in the CCF install, I wonder if we should also ship a demo JS app and use it as an example in the docs?

doc/developers/js_app.rst
achamayou: While I agree, we also do not have documentation guidelines (yet!) or consensus on what they ought to be.

achamayou:
This looks great, it's definitely a step in the right direction!

doc/developers/js_app.rst
letmaik: Is there a way to reference a page itself?
This looks really ugly:
```
See the :ref:`tsoa-based app example <developers/js_app_tsoa_npm:TypeScript Application using Node.js, npm, and tsoa>` on how to generate OpenAPI definitions with TypeScript.
`` 
And it breaks if I change the page title.

doc/developers/js_app.rst
letmaik: Hmm, even if the line becomes too long? I typically break them so that git diff works alright. Also, the VS Code diff view does not support word wrapping.

doc/developers/js_app.rst
achamayou: You don't have to link to the section if you just want to link to the page, but it is more precise that way. You can create named refs if you prefer too: https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#the-ref-role

letmaik: Doesn't work if I omit the section title. I think it's a sphinx (plugin) issue/limitation when it creates the labels.

achamayou: In autoref you mean? I'll take a look.

achamayou (approved):

doc/developers/js_app.rst
letmaik: I think sphinx.ext.autosectionlabel

doc/developers/js_app.rst
jumaffre: My preference is one sentence per line as it's easier to read in VSCode when editing the file. I agree that it makes the diffs in VSCode a little impractical but GitHub renders the changes alright. No a big deal in any case, so feel free to leave it as it is.

jumaffre (approved):

doc/developers/js_app.rst
letmaik: Let's leave it as it is for now. I'll probably make further changes to these docs anyway in the near future.

Issue #1539, opened 2020-08-27
Snapshot auditability


jumaffre:
Follow up from https://github.com/microsoft/CCF/issues/1302

Snapshots are currently generated at regular intervals for a state that is globally committed. However, the snapshot evidence (hash of snapshot) is only committed _after_ the snapshot has been generated. The snapshot is written to disk as soon as it is generated.

This first implementation means that the evidence of a snapshot that is available for new joiners to resume from (i.e. an operator can copy the snapshot file and start a new joiner from it straight away) can actually be rolled back. In this case, the snapshot would be blameless as there's no evidence for it in the ledger.

What we should do instead is:
- [x] Only write the snapshot to disk _once_ the snapshot evidence is globally committed. This means we would have to keep the serialised snapshot around (+ the `version` of the evidence) until `version` is globally committed. 

However, this may not be enough to guarantee that a joiner that resumed from a snapshot can join the consensus:
- [ ] From its perspective, it should not become part of the network (e.g. process app requests) until it has seen the globally committed evidence of the snapshot it has resumed from.
- From the existing network perspective, this new node shouldn't count as part of the Raft quorum until it has confirmed that it has resumed from a trustworthy snapshot.

jumaffre:
On recovery, we also need to make sure that we do not rollback the snapshot evidence.

jumaffre:
Discussed further with @achamayou today. The simplest solution is probably for new joiners to be passed the snapshot and the ledger suffix until the next signature that confirms that the snapshot evidence has been globally committed. The scheme is then very similar to the existing recovery scheme:

1) The new joiner deserialises the snapshot and the ledger suffix, until its see the signature that confirms that the snapshot evidence has been globally committed. It also verifies that the evidence matches the snapshot (SHA256). 
2) It joins the new network from there and then initiates the deserialisation of the snapshot and the private ledger suffix, but this time in private mode.
3) When the private ledger has been deserialised at the same level than the public store (i.e. signature that validates the snapshot evidence), the Merkle roots of both stores are compared and the private maps are swapped in the public store and the node now becomes part of the consensus.

The main drawback are:
- A node cannot join from just a snapshot but also needs an arbitrarily long ledger suffix: there's no way for operators to tell in which ledger chunk the snapshot evidence is committed although it is likely to be in the chunk that immediately follows the snapshot. 
[...]

PR #1668, opened 2020-09-29
Validate snapshot when snapshot evidence is committed

jumaffre:
Part I of #1539 

Until now, snapshots were written to disk as soon as they were serialised. A new joiner node could then use the snapshot directory and resume from the latest snapshot before joining. However, it was possible for the evidence of the snapshot (committed in a standalone transaction right after the snapshot was generated) to be rolled back (e.g. on election), rendering the snapshot blameless. 

Snapshots stored on disk are now marked as committed (`.committed` extension, similar to ledger chunk files) whenever the evidence for that snapshot has been (globally) committed. New joiners resuming from an existing `snapshots/` directory now skip un-committed snapshot files. 

In other words, if operators and nodes behave non-maliciously, only blame-able snapshots can be used by joiners or for recovery. We still need to add verification of snapshots when new joiners resume from a snapshot: they can only trust the new network once the evidence of the snapshot is validated by a signature transaction (part II of #1539).


src/ds/thread_messaging.h
```diff
@@ -285,8 +285,9 @@ namespace threading
       return tasks[tid];
     }
 
-    bool run_one(Task& task)
+    bool run_one()
```
jumaffre, line 288: I believe it is neater if this function does not take any argument and the task automatically deduced from the current thread ID calling `run_one()`.

tests/infra/node.py
```diff
@@ -255,7 +256,24 @@ def get_ledger(self):
         return self.remote.get_ledger()
 
     def get_snapshots(self):
-        return self.remote.get_snapshots()
+        # Wait for all available snapshot files to be committed before
+        # copying snapshot directory
+        def wait_for_snapshots_to_be_committed(src_dir, list_src_dir_func, timeout=3):
```
jumaffre, line 261: Because snapshot evidences are committed asynchronously, it is better to wait for all available snapshots to be committed before attempting to resume from one. This function is a little awkward as it has to work for both `Local` and `SSH` remotes which need to pass the `list_src_dir_func` lambda in.

tests/infra/node.py
```diff
@@ -255,7 +256,24 @@ def get_ledger(self):
...
+            while time.time() < end_time:
+                committed = True
+                for f in list_src_dir_func(src_dir):
+                    committed &= f.endswith(".committed")
+                if committed:
+                    break
+                time.sleep(0.1)
+            if not committed:
+                raise RuntimeError(
+                    f"All snapshots were not committed after {timeout}s in {src_dir}"
```
achamayou, line 273: Should that read "Not all snapshots..."? Printing the set of uncommitted ones may be quite helpful too.

tests/infra/remote.py
```diff
@@ -213,6 +220,10 @@ def get(self, file_name, dst_path, timeout=FILE_TIMEOUT, target_name=None):
                         if os.path.exists(dst_dir):
                             shutil.rmtree(dst_dir)
                         os.makedirs(dst_dir)
+                        if pre_condition_func is not None:
+                            pre_condition_func(
```
achamayou, line 224: I would expect a pre_condition_func to cause an operation to abort if it returns false, I guess the idea is to raise here instead?

tests/infra/remote.py
```diff
@@ -189,7 +189,14 @@ def _setup_files(self):
...
         session.close()
 
-    def get(self, file_name, dst_path, timeout=FILE_TIMEOUT, target_name=None):
+    def get(
+        self,
+        file_name,
+        dst_path,
+        timeout=FILE_TIMEOUT,
+        target_name=None,
+        pre_condition_func=None,
```
achamayou, line 198:
```suggestion
        pre_condition_func=lambda src_dir, ???: True
```

may be more readable and avoid a branch further down.

achamayou (approved):

src/ds/thread_messaging.h
ashamis: So there is a reason for this, and this change introduces a performance hit on a very hot path

calling get_current_thread_id() is a is a map lookup, this changes running a task from O(1) to O(log(n)).

When this was done we had an issue with thread_local and OE. This should have been resolved after the work Eddy did a couple of months ago. So the correct fix is to create a thread local variable that is this threads task.

ashamis (changes requested):
Incorrect comment

src/ds/thread_messaging.h
eddyashton: Afaict, the only place we were calling this `run_one` was from `enclave.h`, where we were also calling exactly the same pattern of functions so incurring the same lookup cost? So this shouldn't be any slower, and looks like a safer API (you can't pass another thread's `Task`).

ashamis (approved):

src/ds/thread_messaging.h
jumaffre: This is followed-up in #1672

PR #1644, opened 2020-09-23
Added a ledger chunk class to manage the Transactions in a single chunk

msftsettiy:
This PR contains changes to split the Ledger class into two. The Ledger class is responsible for iterating over the ledger chunks and the Ledgerchunk class is responsible for iterating over the Transactions in a chunk.

PR #1670, opened 2020-09-29
Fix tutorial

achamayou:
Fix follow-up to #1644


.daily.yml
```diff
@@ -7,6 +7,7 @@ pr:
       - .daily.yml
       - .azure-pipelines-templates/*
 
+
```
achamayou, line 10: Tutorial is only run by release job in daily.

PR #1673, opened 2020-09-30
Reintroduce 487a098 and fix use-after-free in CHAMP map

Theodus:


jumaffre:
Daily CI which will run the SAN tests: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=13376&view=results

achamayou (approved):

PR #1674, opened 2020-09-30
Document dynamic table creation, and use dynamic-friendly Txs everywhere

eddyashton:
The important bit of this PR is that we no longer have `get_view2`: Now we just `get_view`, which can take a `string` rather than a `Map&`. And (almost) everywhere we use a `Tx` it is created from a `Store` so supports dynamic table creation.

This includes brief docs on dynamic table creation, and the `"public:"` prefixing scheme. These are brief as we're still in an intermediate state - the vast majority of our code talks about statically created tables, so we still support and document that code path. As we move more of our code to talk about dynamically created tables, I'll move the docs too (in particular, whif we use dynamic tables in the logging app, the app docs will talk only about dynamic tables).

2 wrinkles worth calling out:
- We probably want to keep some `Map`-like definition on the C++ API holding the name _and_ the expected type, to avoid the code duplication and error-space of repeating the type every time we access. ie - to avoid issues like:
```
auto v1 = tx.get_view<std::string, std::string>("my_map");
auto v2 = tx.get_view<size_t, size_t>("my_map");
```
we may still want something that looks like:
```
Map my_map<std::string, std::string>("my_map");
auto v1 = tx.get_view(my_map);
```
But this should be a safety convenience rather than a requirement on the C++ side.
- There's still a `TODO` in `progress_tracker.h`, which is the one place in the code we create a `Tx` without easy access to a `Store`. Adding a `Store` is tricky since it introduces a circular dependency. This bit explicitly uses a `get_view_old` code path for now - I'll create a `TODO` issue to replace this with an abstract class or something smarter in future. (EDIT: #1675)


doc/developers/kv/kv_how_to.rst
```diff
@@ -48,6 +48,23 @@ For each ``Map`` that a Transaction wants to write to or read from, a :cpp:class
...
+    
+    // Two Views created at the same time, over different public and private maps
+    auto [view_map2, view_map3] =
+        tx.get_view<kv::Map<string, string>, kv::Map<uint64_t, string>>("public:map2", "map3");
+
+This supports dynamic creation of maps - if the requested map did not exist previously, it will be created during this transaction. Any writes to a newly created ``Map`` will be persisted when the transaction commits, and future transactions will be able to access this ``Map`` by name.
+
+.. note::
+
+    When accessing a ``Map`` by name, the confidentiality is encoded in the map's name with a "public:" prefix. For example "public:foo" refers to a public ``Map`` while "foo" is a private ``Map``. The latter is encrypted before writing to the ledger, whereas the former is written unencrypted so can be read by external tools with access to the ledger. These maps are distinct; writes to "public:foo" have no impact on "foo".
```
achamayou, line 66:
```suggestion
    When accessing a ``Map`` by name, the confidentiality is encoded in the map's name with a "public:" prefix. For example "public:foo" refers to a public ``Map`` while "foo" is a private ``Map``. The latter is encrypted before writing to the ledger and all access to the table must happen through application code. The former is written unencrypted so can be read by external tools with access to the ledger. These maps are distinct; writes to "public:foo" have no impact on "foo".
```

achamayou (approved):

doc/developers/kv/kv_how_to.rst
```diff
@@ -48,6 +48,23 @@ For each ``Map`` that a Transaction wants to write to or read from, a :cpp:class
...
+.. code-block:: cpp
+
+    // View on map1
+    auto view_map1 = tx.get_view<kv::Map<string, string>>("map1");
+    
+    // Two Views created at the same time, over different public and private maps
+    auto [view_map2, view_map3] =
+        tx.get_view<kv::Map<string, string>, kv::Map<uint64_t, string>>("public:map2", "map3");
+
+This supports dynamic creation of maps - if the requested map did not exist previously, it will be created during this transaction. Any writes to a newly created ``Map`` will be persisted when the transaction commits, and future transactions will be able to access this ``Map`` by name.
```
jumaffre, line 62:
```suggestion
This supports dynamic creation of maps - if the requested map did not exist previously, it will be created by this transaction. Any writes to a newly created ``Map`` will be persisted when the transaction commits, and future transactions will be able to access this ``Map`` by name.
```

doc/developers/kv/kv_how_to.rst
```diff
@@ -48,6 +48,23 @@ For each ``Map`` that a Transaction wants to write to or read from, a :cpp:class
...
+
+.. note::
+
+    When accessing a ``Map`` by name, the confidentiality is encoded in the map's name with a "public:" prefix. For example "public:foo" refers to a public ``Map`` while "foo" is a private ``Map``. The latter is encrypted before writing to the ledger and all access to the table must happen through application code. The former is written unencrypted so can be read by external tools with access to the ledger. These maps are distinct; writes to "public:foo" have no impact on "foo".
```
jumaffre, line 66:
```suggestion
    When accessing a ``Map`` by name, the confidentiality is encoded in the map's name with a "public:" prefix. For example "public:foo" refers to a public ``Map`` while "foo" is a private ``Map``. The latter is encrypted before writing to the ledger and all access to the table must happen through application code. The former is written unencrypted so can be read by external tools with access to the ledger. These maps are distinct; writes to "public:foo" have no impact on "foo" and vice versa.
```

jumaffre (approved):

PR #1685, opened 2020-10-01
Tag images as latest after release

achamayou:


jumaffre (approved):

.github/workflows/containers.yml
```diff
@@ -30,3 +30,15 @@ jobs:
 
       - name: Push App Run container
         run: docker push ccfciteam/ccf-app-run:${{steps.tref.outputs.tag}}
+
+      - name: Tag latest CI container
+        run: docker tag ccfciteam/ccf-app-ci:${{steps.tref.outputs.tag}} ccfciteam/ccf-app-ci:latest
```
eddyashton, line 35: We can add this tag when we `docker build`, with multiple `-t` args. That seems less error-prone that repeating the original tag name construction here.

achamayou: Ah great, I didn't know you could -t more than once.

PR #1694, opened 2020-10-02
Complex lua literals in proposal generator

### LINKED ISSUE #1687: Careful vote serialization does not handle special characters
The JS modules e2e test currently uses the simple voting mechanism (`return true`).
The proposal body for deploying JS apps contains JSON object keys starting with `/`, which is not handled properly when using careful votes. The generated votes contain code like `if args.bundle.metadata.endpoints./compute == nil`. Using a different syntax that can handle special characters probably makes more sense.

The reason I'm using careful votes is that I'm extending the `infra/` code to allow deploying a JS app bundle on start up. For that I'm adding a new method to `Consortium` and all existing methods use careful votes, so I figured I have to comply. Does it even make sense to use careful votes for JS app deployments? It would mean that the source code of all modules is replicated in each vote.

eddyashton:
Fixes #1687.

In our generated votes, we try to write a Lua script which compares the arguments to the current proposal with the expected argument, to confirm we're voting on the same proposal (or an identical one). To do this, we take the Python object representing these arguments and try to build a Lua literal with the same encoded value. Previously we only did that for a few simple types, this extends it to support far more types, in particular the nested lists-of-objects that are present in a `deploy_js_app` proposal.


src/lua_interp/tx_script_runner.h
```diff
@@ -174,25 +174,25 @@ namespace ccf
...
       {
-        LOG_DEBUG_FMT(get_var_string_from_args(l));
+        LOG_DEBUG_FMT("{}", get_var_string_from_args(l));
         return 0;
       }
 
       static int lua_log_info(lua_State* l)
       {
-        LOG_INFO_FMT(get_var_string_from_args(l));
+        LOG_INFO_FMT("{}", get_var_string_from_args(l));
```
eddyashton, line 189: If a lua script says `LOG_INFO("I'm hungry :{   }"`, we want to log that string as a value, and not treat the `{}` as a libfmt format specifier.

achamayou (approved):

Issue #1693, opened 2020-10-02
Application Sandbox


achamayou:
This is a follow up to #493 and #1028, which added `start_test_network.sh` in the release package.

After discussion with @letmaik, I think there's a number of improvements we'd like to make to this mechanism designed to allow application developers targeting CCF to quickly spin up their application for testing, be it a C++ or JS app.

- [ ] Change default to be a single local node, single member. That's what most users will want, and it's quickest.
- [ ] Default to loading the virtual build of js_generic on startup.
- [ ] Rename `start_test_network.sh` to `sandbox.sh`, which is shorter and does not imply more than one is started.
- [ ] Add a trivial constitution, where all votes automatically pass.
- [ ] Install the sandbox script and the sample constitution so they are part of the .deb package.
- [ ] Install a VERSION file, which can be used by the script to work both under a build directory and an install directory
- [ ] Adjust defaults such as ledger size as necessary for the sandbox use case.

PR #1695, opened 2020-10-02
Add application sandbox

letmaik:
Implements #1693.


letmaik:
I should note that this invalidates a bunch of docs but I'd prefer if that's done in a follow-up PR.

tests/sandbox/sandbox_gov.lua
```diff
@@ -0,0 +1,60 @@
...
+-- This file defines the default initial contents (ie, Lua scripts) of the governance scripts table.
+return {
+  pass = [[
+  tables, calls, votes = ...
+
+  -- interface definitions
+  PASSED = 1
+  PENDING = 0
+  REJECTED = -1
+
```
achamayou, line 13: A comment similar to:

The constitution in a real CCF application deployment would at least count votes and compare to a threshold of members, but in the sandbox sample, all votes pass automatically.

achamayou (approved):

tests/sandbox/sandbox.sh
```diff
@@ -0,0 +1,41 @@
...
+    START_NETWORK_SCRIPT="${PATH_HERE}"/../start_network.py
+    pip install -q -U -e "${PATH_HERE}"/../../python/
+    pip install -q -U -r "${PATH_HERE}"/../requirements.txt
+fi
+
+echo "Python environment successfully setup"
+
+CURL_CLIENT=ON \
+    python "${START_NETWORK_SCRIPT}" \
+    --enclave-type virtual \
```
achamayou, line 35: It's unfortunately awkward for a user to override this if they want to run with sgx.

tests/sandbox/sandbox.sh
```diff
@@ -0,0 +1,41 @@
...
+    --initial-member-count 1 \
+    --initial-user-count 1 \
+    --gov-script "${GOV_SCRIPT}" \
```
achamayou, line 38: Likewise if they want to test a different governance. I think these two options should allow an override.

tests/sandbox/sandbox.sh
letmaik: why? just pass --enclave-type debug, later argument override earlier ones

achamayou: I did not know you could do that, great!

letmaik:
Running a nightly ci run to be sure I didn't break anything https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=13608&view=results

tests/sandbox/sandbox.sh
```diff
@@ -0,0 +1,41 @@
...
+PATH_HERE=$(dirname "$(realpath -s "$0")")
+VERSION_FILE="${PATH_HERE}"/../share/VERSION
+GOV_SCRIPT="${PATH_HERE}"/sandbox_gov.lua
+
+echo "Setting up Python environment..."
+
+if [ ! -f "sandbox_env/bin/activate" ]; then
+    python3.8 -m venv sandbox_env
+fi
+source sandbox_env/bin/activate
```
eddyashton, line 16: Can we add `ccf_` to this name? Just thinking about the actual user flow of running this from a CCF install, if we're going to add anything to the local directory we should name it clearly. I don't know how common this impromptu venv creation is - should this be under `/tmp`?

eddyashton (approved):
Can you add something to the `CHANGELOG` in this PR, and open a TODO issue to make sure we remember the docs update?

letmaik:
@eddyashton I created an issue regarding docs update. I think the changelog is probably best updated later and can then also link to the docs.

tests/sandbox/sandbox.sh
letmaik: I changed it to `.venv_ccf_sandbox` (the [Python docs](https://docs.python.org/3/tutorial/venv.html#creating-virtual-environments) say that `.venv` is a common name) and allow overriding it by setting the `VENV_DIR` environment variable, if someone really wants to.

tests/sandbox/sandbox.sh
```diff
@@ -15,6 +15,7 @@ echo "Setting up Python environment..."
 if [ ! -f "${VENV_DIR}/bin/activate" ]; then
     python3.8 -m venv "${VENV_DIR}"
 fi
+# shellcheck source=/dev/null
```
achamayou, line 18: Is this the "source only from hardcoded paths" warning? If so we can turn it off globally I think.

letmaik: Yeah, I think turning it off makes sense.

Issue #1678, opened 2020-09-30
Re-think "tables" global in JS apps


letmaik:
The `tables` global exposed to JS apps is a bit awkward, mostly because it has a very "normal" name. Renaming to `ccfTables` will avoid conflicts with application code.

I think keeping it as a global makes sense, since it's equivalent to `localStorage`/`sessionStorage` in browsers.

achamayou:
It might make sense to create a `ccf` global even (`ccf.maps`), since there may be more things we want to have in there.

letmaik:
Possibly. And it would be good to decide on terminology. Store? maps? tables?

achamayou:
We've settled on Maps and done more of the renaming already, tables suggested more potential complexity than our strictly key, value maps will ever support. If we also aim to be close to ES Maps, that's even more reason to go there in the JS app context.

letmaik:
After internal discussion, we settled on `ccf.kv`.

PR #1714, opened 2020-10-06
Modernise KV interface in JS/TS

### LINKED ISSUE #1530: Make JS KV API look like standard JS types
We should reduce friction for new developers by basing our API on existing types. This should remove learning humps and also make it easier to build a stub KV for sandboxed testing.

The key question is how we should expose a `TxView`. It looks like the closest approach is to build something that looks quite like a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), although we cannot offer the same or stable iteration order.

We can at least rename our methods to `set`, `get`, and `delete`. We can also easily add a `has` method.

Beyond this, we should be able to support `forEach`, and `size` (not currently tracked in C++, but could be). We may also be able to provide `@@iterator` support.

letmaik:
Fixes #1530 and #1678.


letmaik:
The reason why `bufToJsonCompatible` fails sometimes is because `JS_ParseJSON` expects a null-terminated string for some reason, even though the string size is passed as argument. See line 20636 of quickjs.c. `JS_NewStringLen` does not have that requirement and I think it's an unfortunate inconsistency. To avoid an extra copy, there are two options: patch quickjs or store null-terminated strings in KV. I have a feeling the latter may break other parts of CCF or tooling. @achamayou @eddyashton Any opinions?

src/apps/js_generic/js_generic.cpp
```diff
@@ -165,7 +165,12 @@ namespace ccfapp
     if (!buf)
       return JS_ThrowTypeError(ctx, "Argument must be an ArrayBuffer");
 
-    JSValue obj = JS_ParseJSON(ctx, (char*)buf, buf_size, "<json>");
+    std::vector<uint8_t> buf_null_terminated;
+    buf_null_terminated.reserve(buf_size + 1);
```
achamayou, line 169:
```suggestion
    std::vector<uint8_t> buf_null_terminated(buf_size + 1);
```

achamayou: Also it's probably good to have a comment here to explain why we do this.

letmaik: I think this is slower as that constructor default-initializes all elements.

src/apps/js_generic/js_generic.cpp
```diff
@@ -80,118 +80,213 @@ namespace ccfapp
...
+    LOG_TRACE_FMT("Looking for kv map '{}'", property_name);
 
-    auto tx_ptr = static_cast<kv::Tx*>(JS_GetOpaque(this_val, tables_class_id));
-    auto view = tx_ptr->get_view<Table>(property_name);
+    auto tx_ptr = static_cast<kv::Tx*>(JS_GetOpaque(this_val, kv_class_id));
+    auto view = tx_ptr->get_view<KVMap>(property_name);
 
-    auto view_val = JS_NewObjectClass(ctx, view_class_id);
+    auto view_val = JS_NewObjectClass(ctx, kv_map_view_class_id);
     JS_SetOpaque(view_val, view);
```
achamayou, line 274: A comment suggesting this follows Map with a link would be good in my opinion.

achamayou (approved):

achamayou (approved):

eddyashton (approved):

PR #1717, opened 2020-10-07
AFT: In BFT mode use Byzantine commit

### LINKED ISSUE #1716: AFT: Use Byzantine commitment to set global commit
- [ ] Work out when we have obtained global commit
- [ ] Use Byzantine commitment for global commit if using AFT in BFT mode
- [ ] Ensure a network can run with less than 3 nodes when the network is opening

ashamis:
Resolves #1716

### DIFF
--- a/src/consensus/aft/impl/state.h
+++ b/src/consensus/aft/impl/state.h
@@ -98,7 +98,8 @@ namespace aft
       my_node_id(my_node_id_),
       current_view(0),
       last_idx(0),
-      commit_idx(0)
+      commit_idx(0),
+      should_start_using_byz_commit(false)
     {}
 
     SpinLock lock;
@@ -110,5 +111,7 @@ namespace aft
     kv::Version commit_idx;
 
     ViewHistory view_history;
+
+    bool should_start_using_byz_commit;
   };
 }
--- a/src/consensus/aft/impl/state.h
+++ b/src/consensus/aft/impl/state.h
@@ -98,8 +98,7 @@ namespace aft
       my_node_id(my_node_id_),
       current_view(0),
       last_idx(0),
-      commit_idx(0),
-      should_start_using_byz_commit(false)
+      commit_idx(0)
     {}
 
     SpinLock lock;
@@ -110,8 +109,9 @@ namespace aft
     kv::Version last_idx;
     kv::Version commit_idx;
 
-    ViewHistory view_history;
+    kv::Version cft_watermark_idx;
+    kv::Version bft_watermark_idx;
 
-    bool should_start_using_byz_commit;
+    ViewHistory view_history;
   };
 }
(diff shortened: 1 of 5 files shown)


src/consensus/aft/raft.h
```diff
@@ -1534,34 +1537,49 @@ namespace aft
...
-        // The majority must be checked separately for each active
-        // configuration.
-        std::vector<Index> match;
-        match.reserve(c.nodes.size() + 1);
-
-        for (auto node : c.nodes)
+        auto progress_tracker = store->get_progress_tracker();
+        CCF_ASSERT(progress_tracker != nullptr, "progress_tracker is not set");
+        new_commit_idx = progress_tracker->get_highest_commit_level();
+        state->should_start_using_byz_commit = true;
```
jumaffre, line 1551: Something I am missing here: at the very beginning, `node_count() < min_bft_node_count` so we only get here if `state->should_start_using_byz_commit == true`. However, it seems that `state->should_start_using_byz_commit` is initially `false` and only becomes `true` if we get here, which isn't possible because `node_count() < min_bft_node_count`?

jumaffre: As discussed, this is actually working the other way around: the commit logic defaults to `CFT` until we have `min_bft_node_count` nodes and then uses `BFT` even if `node_count()` goes under `min_bft_node_count`. We should probably expand the comment to reflect that.

jumaffre (approved):

src/consensus/aft/impl/state.h
```diff
@@ -98,7 +98,8 @@ namespace aft
       my_node_id(my_node_id_),
       current_view(0),
       last_idx(0),
-      commit_idx(0)
+      commit_idx(0),
+      should_start_using_byz_commit(false)
```
achamayou, line 102: Is that really `using_byz_commit`? Presumably we either are or aren't?

src/node/progress_tracker.h
```diff
@@ -505,9 +505,32 @@ namespace ccf
       return hash;
     }
 
+    kv::Consensus::SeqNo get_highest_commit_level()
```
achamayou, line 508: I think a name like `committed_nonce_watermark` or `highest_committed_nonce` would be more descriptive.

achamayou:
@ashamis The logic looks good but I would much prefer if we split `commit_idx` into `cft_commit_idx` and `byz_commit_idx` or `bft_commit_idx` in our state. I think it'd be clearer, easier to test, and we may even want to expose both. `byz_commit_idx` starting out being the same as `cft_commit_idx` while the number of nodes is below 3 seems fine (and/or we could take the view that it's sentinel and that users have to rely on `cft_commit_idx` during that time instead).

In any case, I think keeping track of each separately would be good.

achamayou (approved):

Issue #1564, opened 2020-09-04
Remove dead code


eddyashton:
We should cull unused code. Please add anything else you find to this list.

- [x] Notification system (#1582)
- [x] user_code governance flow (#1611)
- [x] flatbuffers usage in benchmarks and flatbuffers dependency (#1593)
- [x] valijson
- [x] txregulator test
- [x] lua generic
- [x] mpscq commented out code
- [x] mod_mbedtls commented out code
- [x] flatbuffers
- [x] k6

ashamis:
I am removing the following off the unused code list:
* BFT Tx Encryptor (#1538)
* BFT-specific paths in frontend/kv
They are needed for AFT and I am actively working on integrating back into said code

achamayou:
The sig fwd test doesn't seem to be running any faster than the regular client sig test. We don't currently have much of a use case for client signatures, so I think we could just take it out.

achamayou:
The batched test needs to be migrated to JS before we can removed lua_generic.

letmaik:
With https://github.com/microsoft/CCF/pull/1905 it looks like this issue is complete?

eddyashton:
Correct!

PR #1719, opened 2020-10-07
Remove dead mbedtls code

achamayou:
More #1564


src/tls/context.h
```diff
@@ -17,15 +17,8 @@ namespace tls
     EntropyPtr entropy;
 
 #ifndef NO_STRICT_TLS_CIPHERSUITES
-#  ifdef MOD_MBEDTLS
```
jumaffre, line 20: Could you also remove the dead value in the `ParticipantsCurve` enum in `network.py`?

jumaffre (approved):

src/tls/context.h
achamayou: Sure, will do!

achamayou: I'll go around the doc and scrub mentions of 25519 too while I'm at it, because we just don't support it at the moment.

eddyashton (approved):

PR #1722, opened 2020-10-07
Remove api/schema endpoint

eddyashton:
`/api/schema` just returns a subset of the information in `/api`, but in a custom format. We have to jump through some hoops internally to make sure our schema can be returned in either form. Its now unused externally, we only care what `/api` says, so I think we should just remove `/api/schema` completely. If we want to filter to part of the `/api` document we can offer that (maybe there's a standard request format?), but its simpler if it remains OpenAPI.


letmaik (approved):

ashamis (approved):

achamayou:
@eddyashton since this is an RPC removal, I think it's a good thing to mention in the changelog.

Issue #1523, opened 2020-08-21
Enforce role-based/execution context access control to KV tables


eddyashton:
We currently have a whitelisting system for preventing access to sensitive tables during Lua execution. We should extend this to apply to all table accesses elsewhere in CCF. There are several categories of table:

- Public service tables. These are unencrypted in the ledger, and readable by all transactions. They should only be written to in a service execution context. An example is the Signatures table.
- Internal service tables. These are encrypted in the ledger, and only readable by privileged node-created transactions. They should only be written to in a service execution context. An example is the PBFT Requests table.
- Governance tables. These are unencrypted in the ledger, and readable by all transactions. They should only be written to in a member execution context. An example is the CCF Users table.
- Application tables. These may be encrypted or unencrypted. They should only be readable and writable in a user execution context.

We should always know what context a transaction is executing, and pass this when requesting tables. Attempting to access table which is not accessible in the current context will raise an error.

We may need to narrow/alter the execution context within a single transaction. For instance it may begin in a service context, able to record a Request, but be narrowed to a user-context (preventing access to private service tables while granting access to application tables) before being passed to the application endpoint. There may be additional access checks imposed on top - for instance in the member frontend we want to prevent constitution-defined actions from breaking invariants on core governance tables.

eddyashton:
As first part of this we need a way to identify the different types of tables. Following dynamic table creation, we believe it should be possible to encode this information purely in the table's name, in the same way we currently use the `public:` prefix to indicate unencrypted tables. This will let us add quick ACL support to the `js_generic` runtime.

We already have a `ccf.` prefix for all the governance tables, though we also use this for the special 'service' tables. We will need some way to distinguish those, and should try to settle on those names now. We should also spread the `public:` prefixing requirements to _all_ table names, even while some are still statically created.

I propose:
- `public:ccf.members` - a governance table, only written to by auditable member-driven proposals
- `log_records` - a private encrypted app table
- `public:_ccf.signatures` - a publically readable table, but written to by the service/node-started transactions
- `_ccf.bft_requests` - a private, encrypted service-written table

`public:ccf.` is a little redundant - the `ccf.` prefix _always_ indicates a public table, but the `public:` prefix is included for consistency.
`_ccf` is a distinct prefix indicating extra-special service tables, optionally public or private.

eddyashton:
We had a discussion and settled on some different prefices:

- `ccf.gov.` for governance tables
- `ccf.internal.` for internal service tables
[...]

PR #1727, opened 2020-10-08
Encode privacy domain in all table names

eddyashton:
Steps towards enforcing table ACLs in different execution contexts. This PR just covers renaming as described in #1523, encoding the `PrivacyDomain` and (new) `AccessCategory` in all table names.

The interesting changes are under `kv/`, where I've deprecated the variant of `create` that takes a `PrivacyDomain` argument and replaced it with one which parses this from the table name. Although the default argument is removed, this matches the old behaviour (if you called `create("map")` before it was a private map, its still a private map now).


CHANGELOG.md
```diff
@@ -6,14 +6,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Changed
+- Public tables in the KV must now indicate this in their name (with a `public:` prefix), and internal tables have been renamed. Any governance or auditing scripts which operate over internal tables must use the new names (eg - `ccf.members` is now `public:ccf.gov.members`).
```
achamayou, line 10: Can we also indicate that passing the domain explicitly is deprecated (and may go away in 0.16)?

achamayou (approved):

jumaffre (approved):

Issue #1712, opened 2020-10-06
Allow setting member_data on the CLI for initial set of members


achamayou:
Follow up to #1657, requested by @lynshi 

Extend

```
--member-info=member2_cert.pem,member2_enc_pubk.pem
```

To support

```
--member-info=member2_cert.pem,member2_enc_pubk.pem,member2_data.json
```

PR #1728, opened 2020-10-08
Add member data init CLI

achamayou:
Address #1712


tests/governance.py
```diff
@@ -91,6 +91,24 @@ def test_no_quote(network, args, notifications_queue=None, verify=True):
...
+
+        def member_info(mid):
+            return mc.post(
+                "/gov/read", {"table": "ccf.members", "key": mid}
+            ).body.json()
+
+        assert "member_data" not in member_info(0)
+        assert "member_data" not in member_info(1)
+        assert "member_data" not in member_info(2)
+        assert member_info(3)["member_data"] == {"is_operator": True}
```
eddyashton, line 110: This seems fragile, linked at a distance to the `--initial-operator-count` argument defined in CMake and the default `--initial-member-count`. Can we write this as a loop over `args.initial_member_count` and `args.initial_operator_count`?

src/host/main.cpp
```diff
@@ -361,7 +361,7 @@ int main(int argc, char** argv)
     members_info,
     "--member-info",
     "Initial consortium members information (public identity,recovery public "
-    "key)")
+    "ke,member data)")
```
jumaffre, line 364:
```suggestion
    "key,member data)")
```

jumaffre (approved):

tests/governance.py
achamayou: Yeah, idk, I think that's fragile differently because if the initial_operator_count becomes zero for some reason (for example in an accidental CMake change), then the test passes even though it checks nothing.

I've made the check more generic, but added an assert on the initial count.

src/host/main.cpp
achamayou: Thank you!

ashamis (approved):

eddyashton (approved):

Issue #48, opened 2019-05-13
Signature trigger should happen centrally, not per frontend


achamayou:
Node signatures are currently triggered both by count and time mechanism, and occasionally cause signatures that are closer together than they could be. This is somewhat inefficient, and results in slightly lower throughput than expected in a real system.

achamayou:
Discussed with @ashamis today, History seems like a good place to do this now that it receives tick()s. To be discussed further on Monday.

PR #1732, opened 2020-10-09
Issue signatures from history

### LINKED ISSUE #1736: Recovery_history eventually has a deleted reference to recovery_store
In node_state::recover_private_ledger_end_unsafe we delete recovery_store, however recovery_history has a reference to recovery_store.

ashamis:
Resolves #1736, #48
Links: #1737

There is a long held desire to move the instigation of signatures from the frontends. This change address this by moving the issuance of signatures to history. Please review this carefully, not much code changes here but this is a more fundamental change where bugs can be harder to find later.

### DIFF
--- a/src/ds/spin_lock.h
+++ b/src/ds/spin_lock.h
@@ -5,6 +5,9 @@
 #ifdef INSIDE_ENCLAVE
 #  include <pthread.h>
 
+// OpenEnclave, at this time, does not provide pthread_spin_trylock. There is
+// currently a PR that will introduce said function and this should be removed
+// when said function is in OpenEnclave.
 #  ifndef VIRTUAL_ENCLAVE
 static unsigned int _spin_set_locked(pthread_spinlock_t* spinlock)
 {
--- a/src/ds/spin_lock.h
+++ b/src/ds/spin_lock.h
@@ -4,6 +4,18 @@
 
 #ifdef INSIDE_ENCLAVE
 #  include <pthread.h>
+   static unsigned int _spin_set_locked(pthread_spinlock_t* spinlock)
+{
+    unsigned int value = 1;
+
+    asm volatile("lock xchg %0, %1;"
+                 : "=r"(value)     /* %0 */
+                 : "m"(*spinlock), /* %1 */
+                   "0"(value)      /* also %2 */
+                 : "memory");
+
+    return value;
+}
 
 class SpinLock
 {
@@ -24,6 +36,11 @@ public:
   void lock()
   {
     pthread_spin_lock(&sl);
+  }
+
+  bool try_lock()
+  {
+    return (_spin_set_locked(&sl) == 0);
   }
 
   void unlock()
--- a/src/ds/spin_lock.h
+++ b/src/ds/spin_lock.h
@@ -4,18 +4,21 @@
 
 #ifdef INSIDE_ENCLAVE
 #  include <pthread.h>
-   static unsigned int _spin_set_locked(pthread_spinlock_t* spinlock)
+
+#  ifndef VIRTUAL_ENCLAVE
+static unsigned int _spin_set_locked(pthread_spinlock_t* spinlock)
 {
-    unsigned int value = 1;
+  unsigned int value = 1;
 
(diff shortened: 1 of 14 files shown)


achamayou:
Changed to draft to avoid any accidental merging (CI runs on draft too).

src/ds/spin_lock.h
```diff
@@ -5,6 +5,24 @@
 #ifdef INSIDE_ENCLAVE
 #  include <pthread.h>
 
+// OpenEnclave, at this time, does not provide pthread_spin_trylock. There is
+// currently a PR that will introduce said function and this should be removed
+// when said function is in OpenEnclave.
```
eddyashton, line 10: Can we include a link to the PR in this comment?

achamayou, line 10:
```suggestion
// when said function is in OpenEnclave. https://github.com/openenclave/openenclave/pull/3641
```

src/node/history.h
```diff
@@ -490,13 +500,80 @@ namespace ccf
...
+          {
+            auto time = threading::ThreadMessaging::thread_messaging
+                          .get_current_time_offset();
+
+            auto consensus = self->store.get_consensus();
+            if (
+              (consensus != nullptr) && consensus->is_primary() &&
+              self->store.commit_gap() > 0 &&
+              time.count() > self->time_of_last_signature.count() &&
+              (time.count() - self->time_of_last_signature.count()) >
```
achamayou, line 545: So long as the interval is positive (which we hopefully check upfront), :545 always implies :544, which can be removed.

src/node/history.h
```diff
@@ -490,13 +500,80 @@ namespace ccf
...
+                sig_ms_interval)
+            {
+              msg->data.self->emit_signature();
+            }
+
+            time_since_last_sig =
```
achamayou, line 551: This isn't time since last sig? (time - self->time_of_last_signature) is time since last sig or am I missing something?

achamayou: I guess that's time_to_next_sig? I don't understand why it's not sig_msg_interval every time though.

ashamis: This function is called every sig_msg_interval but we could have signed in between callbacks. We will therefore skip this signature (line 545) and so we do not want to schedule the next signature to be sig_msg_interval after now but rather after the last signature. 

Please confirm that my logic is correct because i could be really wrong here.

src/ds/thread_messaging.h
```diff
@@ -95,10 +76,40 @@ namespace threading
       } while (!item_head.compare_exchange_strong(tmp_head, item));
     }
 
+    struct TimerEntry
```
achamayou, line 79: The timer handling logic has expanded somewhat, and isn't completely trivial, it would be good to have some unit tests for it (there's a thread_messaging.cpp but it only has a single test to check no messages are leaked).

src/node/history.h
```diff
@@ -490,13 +500,80 @@ namespace ccf
...
+          {
+            auto time = threading::ThreadMessaging::thread_messaging
+                          .get_current_time_offset()
+                          .count();
+            auto time_of_last_signature = self->time_of_last_signature.count();
+
+            auto consensus = self->store.get_consensus();
+            if (
+              (consensus != nullptr) && consensus->is_primary() &&
+              self->store.commit_gap() > 0 && time > time_of_last_signature &&
```
achamayou, line 545: I still think that `(time - time_of_last_signature) > sig_ms_interval` is stronger than `time > time_of_last_signature` and so we remove the latter for clarity.

src/node/history.h
achamayou: That's fine logic-wise, but the name is confusing is. Time since last sig is `time - self->time_of_last_signature.count()`.

`sig_ms_interval - time_since_last_sig` should be called `time_to_next_sig`.

ashamis: time_to_next_sig

achamayou (approved):

src/ds/thread_messaging.h
ashamis: #1758

PR #1733, opened 2020-10-09
Add read-only restrictions when accessing governance tables from JS

eddyashton:
Following #1727, use the `AccessCategory` and `PrivacyDomain` to decide whether a table should be writeable, readable, or entirely hidden, from JS.

This is roughly the pattern we want to apply for all C++ access as well, so this can be seen as a first pass test. In particular, I don't think we can enforce these restrictions statically since the tables are accessed purely by name, so we'll have to throw an error _when you call `set`_ rather than removing the `set` method entirely.


src/apps/js_generic/js_generic.cpp
```diff
@@ -268,6 +280,42 @@ namespace ccfapp
...
+      case kv::AccessCategory::INTERNAL:
+      {
+        if (security_domain == kv::SecurityDomain::PUBLIC)
+        {
+          read_only = true;
+        }
+        else
+        {
+          throw std::runtime_error(fmt::format(
+            "JS application cannot access private internal CCF table '{}'",
```
achamayou, line 298: I think "Application" may be more accurate than "JS Application", although it is the JS App here, the reason it's denied access if that it's app code, whether it's JS or not isn't relevant.

achamayou (approved):

PR #1735, opened 2020-10-09
Add historical queries capabilities to infra logging app 

jumaffre:
The historical queries test was until now only tested in the `e2e_logging` test and couldn't be integrated in the test suite (spoiler alert: it still isn't). This PR adds support for historical queries in the `LoggingTxs` test class which records all transactions ever sent to the network (useful when checking that all messages are recovered after a recovery for example). 

To do so, we keep track of all log messages ever sent to the network, even the ones that overwrite previous messages, recorded under the same `id`. When verifying messages, we check that the most recent messages are available in the KV (`/log/private` endpoint) and use the `/log/private/historical` endpoint for overwritten entries.

The historical query still can't be introduced in the test suite because of #1648: after recovery/snapshot, the node doesn't have the ledger secrets for the historical entries.


src/apps/logging/logging.cpp
```diff
@@ -200,6 +200,7 @@ namespace loggingapp
       };
       make_endpoint(
         "log/private/prefix_cert", HTTP_POST, log_record_prefix_cert)
+        .set_auto_schema<LoggingRecord::In, bool>()
```
jumaffre, line 203: These endpoints are not listed in `/api` if they don't have a schema.

tests/suite/test_requirements.py
```diff
@@ -98,7 +98,7 @@ def check(network, args, *nargs, **kwargs):
 def can_kill_n_nodes(nodes_to_kill_count):
     def check(network, args, *nargs, **kwargs):
         primary, _ = network.find_primary()
-        with primary.client("member0") as c:
```
jumaffre, line 101: It turns out that `member0` is now retired in one of the tests.

src/apps/logging/logging.cpp
achamayou: That sounds like a bug, we can and should at least have an entry with the path, the verb etc

achamayou (approved):

src/apps/logging/logging.cpp
eddyashton: https://github.com/microsoft/CCF/pull/1738

tests/infra/logging_app.py
```diff
@@ -4,82 +4,163 @@
...
-                check_commit(rep_pub, result=True)
+                self.priv[self.idx].append(
+                    {"msg": priv_msg, "seqno": rep_priv.seqno, "view": rep_priv.view}
+                )
 
-                self.priv[self.next_priv_index] = priv_msg
-                self.pub[self.next_pub_index] = pub_msg
-                self.next_priv_index += 1
-                self.next_pub_index += 1
+                # Public records do not handle historical queries
```
eddyashton, line 55: We don't have an endpoint to query the public table in the logging app, but there's nothing preventing us from adding it. I think this client code should do the same thing for writing to public and private tables (ie - remove the `if not repeat`), and simply skip the `/historical` calls when verifying the public entries. In particular, this level of repeat tests something that we're otherwise _not_ testing, that entries are overwritten when we reuse IDs.

eddyashton (approved):

tests/infra/logging_app.py
jumaffre: Good point. I didn't want to add a historical endpoint for public records as well as this feature is still in progress and the historical endpoint is fairly verbose for now. I've updated the `logging_app.py` though so that public records are repeated as well but only the latest submitted entry is verified. We can revisit this later when the support for historical queries is complete.

PR #1740, opened 2020-10-12
Expose AES key generation to JS

letmaik:
Later on this may be exposed as [SubtleCrypto.generateKey](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey) using the same underlying C++ function.

### DIFF
--- a/src/apps/js_generic/js_generic.cpp
+++ b/src/apps/js_generic/js_generic.cpp
@@ -3,6 +3,7 @@
 #include "enclave/app_interface.h"
 #include "kv/untyped_map.h"
 #include "node/rpc/user_frontend.h"
+#include "tls/entropy.h"
 
 #include <memory>
 #include <quickjs/quickjs-exports.h>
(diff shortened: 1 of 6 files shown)


achamayou (approved):

src/apps/js_generic/js_generic.cpp
```diff
@@ -80,6 +81,31 @@ namespace ccfapp
...
+      return JS_ThrowTypeError(
+        ctx, "Passed %d arguments, but expected 1", argc);
+
+    int32_t key_size;
+    if (JS_ToInt32(ctx, &key_size, argv[0]) < 0)
+    {
+      js_dump_error(ctx);
+      return JS_EXCEPTION;
+    }
+    if (key_size != 128 && key_size != 192 && key_size != 256)
```
achamayou, line 97: Comment explaining that these are the supported key size for AES-128.

letmaik: Done.

achamayou: Thank you!

jumaffre (approved):

Issue #1653, opened 2020-09-25
URL path templates for JS endpoints


letmaik:
Currently path templates are only supported for C++ endpoints.

Support for templates in JS endpoints will implicitly come once https://github.com/microsoft/CCF/issues/1460 is done.

PR #1744, opened 2020-10-12
Initial dynamic endpoints in JS app

eddyashton:
This PR adds support for dynamically defined endpoints (ie - JS scripts added at runtime) using all of the forwarding and dispatch logic of a standard endpoint.

EDIT: To clarify - we already supported JS endpoints which were added dynamically, rather than the statically defined endpoints in C++. But their dispatch logic was custom and limited, as there was 'really' a single, static endpoint that then delegated to dynamically defined scripts. This PR promotes those dynamic endpoints to first-class citizens, with their dispatch related properties also defined dynamically and processed transactionally.

In short, `find_endpoint` returns a smart pointer managed _definition of an endpoint_, rather than a concrete `Endpoint` instance. In the case of all our old statically defined endpoints this is still an `Endpoint` owned by the `EndpointRegistry` (ie - probably an entry in the `fully_qualified_endpoints` map), but for other runtimes like JS we override this to return our _own_ definition, using `EndpointProperties` looked up from the KV, which will later give us everything we need to `execute_endpoint`.

Opening as a `DRAFT` as there's still some `TODO`s to clear up.


src/node/rpc/endpoint.h
```diff
@@ -0,0 +1,83 @@
...
+
+namespace ccf
+{
+  namespace endpoints
+  {
+    using URI = std::string;
+
+    struct EndpointKey
+    {
+      URI method;
```
achamayou, line 19: I think we should avoid the term method here if we can (I know we've used it historically because of JSON RPC), perhaps `uri_suffix` or something?

src/node/rpc/frontend.h
```diff
@@ -551,15 +545,20 @@ namespace ccf
...
+      const bool is_bft =
+        consensus != nullptr && consensus->type() == ConsensusType::BFT;
+      const bool is_local =
+        endpoint != nullptr && endpoint->properties.execute_locally;
+      const bool should_bft_distribute = is_bft && !is_local &&
+        (ctx->execute_on_node || consensus->is_primary());
+
+      // TODO: This decision is based on several things read from the KV
+      // (cert->caller_id, request->is_local) which are true _now_ but may not
+      // be true when this is actually received/executed
```
achamayou, line 559: It's likely that we will need something like an endpoint definition digest which we capture upfront and re-verify at execution time in contexts like forwarding.

eddyashton: Good point, I've added a mention of the plan for digest comparisons.

achamayou (approved):

letmaik:
Does this PR solve https://github.com/microsoft/CCF/issues/1653 as well?

Issue #1527, opened 2020-08-24
Node can read ledger from multiple paths


jumaffre:
Follow up from https://github.com/microsoft/CCF/pull/1518#discussion_r474621559 

We should let operators specify a list of ledger paths for a node. For example, this will allow a node that has joined from a snapshot to read historical entries from a read-only mount point (e.g. rsync'ed by the operator) while writing/reading newer entries from its local ledger directory.

Issue #1705, opened 2020-10-05
Improve historical query support


eddyashton:
We have initial support for historical queries. This issue tracks the remaining tasks:

- [x] Retrieve secrets to decrypt historical entries, after snapshot recovery (#1648)
- [x] Additional validation of historical entries. Currently we check the signature over the target ledger entry, but don't check that the signing node is still an active participant at the point of the signature. We should work out if this is possible.
- [ ] Explore queries from snapshots. Currently we can query the state written _in_ specific target transactions, but not general state. Snapshots make it much cheaper to restore an entire historical store for arbitrary queries.

eddyashton:
To validate signatures over historical queries, we need broader historical state. Specifically, we want to know the active nodes at the point the signature was produced. We discussed alternatives such as having each signature include a service-endorsed copy of the signer's public key, but this would still require additional historical state to know the expected service identity at each signature.

So we need to be able to replay a full KV state (as in every transaction, though we may only _keep_ specific tables), and to make this manageable we need to do it via snapshots. When asked for a historical query at N, we don't just need N and the signature over N, we also need the full range of transactions from the snapshot before N to the signature over the evidence of this snapshot. We need to check both signatures are globally committed (vs the node's knowledge of consensus commit), and check that both signatures are valid (they are produced by a known, active node, as decided in the reconstructed store), to trust the contents of the transaction at N.

eddyashton:
After further discussion, we think that even snapshots don't get us out of this regression loop - we also need historical service identities (and the signatures over those, validated recursively) to fully verify a historical entry internally. Its possible we can store all service identities in the active KV as a workaround, or chain them in the same way we do with the ledger secrets, but this needs further thought.

However, there's a neater solution for all of this - we can now return receipts for historical transactions, and the caller can then verify the signatures in those receipts (rooted in a historical node or service they trust). This may not give them the same validation we could do online, but we believe is sufficient for current requirements,

achamayou:
- [ ] Explore the perf impact of including the (endorsed) node cert in each signature, and (transitively) checking the signature against the current service identity. Once we have support for full-state/snapshot, we can keep that and remove the cert inclusion in the signature, and tighten it further by checking the node state at the time.

eddyashton:
[...]

PR #1748, opened 2020-10-13
Python infra support for read-only ledger

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1527

Now that a node can use the `--read-only-ledger-dir` option to specify a read-only ledger directory (see #1725), this PR adds end-to-end testing for this feature. 

To do so, the `test_add_node_from_snapshot` test in `reconfiguration.py` now takes an extra `copy_ledger_read_only` argument to copy the target node's ledger to the new node's read-only ledger directory. 

The new `@req.add_from_snapshot` decorator also guarantees that there are some historical entries followed by a committed snapshot before the new node joins. The decorator also make sure that the historical entries are verified on the new node after it has joined. Using this decorator will allow us to test this easily in the test suite. Note that the `test_add_node_from_snapshot` test fails if `copy_ledger_read_only` is `False`, which we test for in the `reconfiguration.py` test.

### DIFF
--- a/src/apps/logging/logging.cpp
+++ b/src/apps/logging/logging.cpp
@@ -342,6 +342,7 @@ namespace loggingapp
         ccf::historical::adapter(
           get_historical, context.get_historical_state(), is_tx_committed))
         .set_auto_schema<LoggingGetHistorical>()
+        .set_forwarding_required(ccf::ForwardingRequired::Never)
         .install();
 
       auto record_admin_only =
--- a/tests/infra/logging_app.py
+++ b/tests/infra/logging_app.py
@@ -23,6 +23,7 @@ class LoggingTxs:
         self.priv = defaultdict(list)
         self.idx = 0
         self.user = f"user{user_id}"
+        self.network = None
 
     def get_last_tx(self, priv=True):
         txs = self.priv if priv else self.pub
@@ -36,6 +37,7 @@ class LoggingTxs:
         on_backup=False,
         repeat=False,
     ):
+        self.network = network
         remote_node, _ = network.find_primary()
         if on_backup:
             remote_node = network.find_any_backup()
@@ -81,9 +83,11 @@ class LoggingTxs:
 
         network.wait_for_node_commit_sync()
 
-    def verify(self, network, node=None, timeout=3):
+    def verify(self, network=None, node=None, timeout=3):
         LOG.info("Verifying all logging txs")
-        nodes = network.get_joined_nodes() if node is None else [node]
+        if network is not None:
+            self.network = network
+        nodes = self.network.get_joined_nodes() if node is None else [node]
         for node in nodes:
             for pub_idx, pub_value in self.pub.items():
                 # As public records do not yet handle historical queries,
--- a/tests/infra/logging_app.py
+++ b/tests/infra/logging_app.py
@@ -17,6 +17,11 @@ class LoggingTxs:
         self.idx = 0
         self.user = f"user{user_id}"
 
+    def get_last_tx(self, priv=True):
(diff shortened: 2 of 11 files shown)


src/apps/logging/logging.cpp
```diff
@@ -342,6 +342,7 @@ namespace loggingapp
         ccf::historical::adapter(
           get_historical, context.get_historical_state(), is_tx_committed))
         .set_auto_schema<LoggingGetHistorical>()
+        .set_forwarding_required(ccf::ForwardingRequired::Never)
```
jumaffre, line 345: Even though the historical query endpoint is `HTTP_GET`, it is created with `make_endpoint()` (as opposed to `make_read_only_endpoint()`) so the command was always be forwarded. We may want to revisit this in #1705.

tests/infra/logging_app.py
```diff
@@ -69,37 +83,39 @@ def issue(
...
-            )
-
-        for priv_idx, priv_value in self.priv.items():
-            for v in priv_value:
+        if network is not None:
+            self.network = network
+        nodes = self.network.get_joined_nodes() if node is None else [node]
+        for node in nodes:
+            for pub_idx, pub_value in self.pub.items():
+                # As public records do not yet handle historical queries,
```
achamayou, line 93: Is that right? I thought private ones didn't?

tests/infra/logging_app.py
```diff
@@ -145,7 +161,7 @@ def _verify_tx(
                         LOG.info(
                             f"Sleeping for {retry_after}s waiting for historical query processing..."
                         )
-                        timeout += retry_after
+                        timeout += int(retry_after * 0.8)
```
achamayou, line 164: 0.8?

tests/infra/logging_app.py
```diff
@@ -69,37 +83,39 @@ def issue(
...
-                entry["view"],
-                priv=False,
-                timeout=timeout,
-            )
-
-        for priv_idx, priv_value in self.priv.items():
-            for v in priv_value:
+        if network is not None:
+            self.network = network
+        nodes = self.network.get_joined_nodes() if node is None else [node]
```
eddyashton, line 90: It looks like its possible for `self.network` to still be `None` here, can we add a check and avoid accessing if this is the case?

eddyashton (approved):

tests/infra/logging_app.py
jumaffre: At the moment, only private records support historical queries: https://github.com/microsoft/CCF/blob/a59b09e5d9589ec755ca2935a581c7142802febf/src/apps/logging/logging.cpp#L340

tests/infra/logging_app.py
jumaffre: The logic here is that we want to increase the overall loop timeout (which is 3 seconds by default) if we're verifying historical query. It used to be `timeout += retry_after`, which meant it would take many iterations of 3 seconds (30 iterations as the default timeout is 3 and we sleep 0.1 seconds in the loop anyway) to time out. `0.8` seems a nice compromise to go through this loop enough times without causing a ridiculous timeout if no historical entry can be found. I've added an inline comment to explain this.

tests/infra/logging_app.py
jumaffre: Good point. I've added a check for this.

tests/infra/logging_app.py
eddyashton: To be clear, this is just that we only have an endpoint for querying private historical entries, and none for querying public. There's nothing technical preventing us from querying public state, its just a case of code duplication/writing a more general endpoint, so we haven't done it yet.

Issue #1506, opened 2020-08-18
Node RPC port should default to 443


jumaffre:
As CCF nodes are now full-fledged HTTPS servers, we should default the RPC port to 443 when none is specified. 

This will interfere with the current behaviour where port 0 is selected when none is passed in.

achamayou:
We can manually pass 0 in our test infra though?

jumaffre:
> We can manually pass 0 in our test infra though?

Yep, and we will do that. But it would be nice to default to 443 for the `start_test_network.sh` script.

achamayou:
I'm not sure 443 is the right default for `sandbox.sh` considering that it's more likely to be run on dev environments by users that aren't likely to be privileged. Something unprivileged like 8000 may be preferable there.

As far as `cchost` itself is concerned, I agree that 443 is a good choice.

PR #1749, opened 2020-10-13
Change default sandbox port to 8000

achamayou:
See #1506, with a few minor tweaks.


eddyashton (approved):

jumaffre (approved):

tests/sandbox/sandbox.sh
```diff
@@ -41,4 +41,5 @@ CURL_CLIENT=ON \
     --gov-script "${GOV_SCRIPT}" \
     --ledger-chunk-bytes 5MB \
     --label sandbox \
+    -n 127.0.0.1:8000 \
```
letmaik, line 44: Note that this prevents overriding the default since `-n` is additive.

achamayou: That's true, but unfortunately it's difficult to avoid if we want to have the default be 443 the rest of the time, which we do.

A possible solution would be to only add this line if $@ does not contain -n, and to omit it otherwise.

letmaik: If we change `-n` to be of format `-n foo bar` then it's easy to override.

achamayou: True, but it's a big change of style, we use this for all our options here and elsewhere (member info etc), I doubt the benefit offsets the user update costs.

eddyashton: I think preventing overriding is a major issue. I have an alternative approach, will open a PR shortly.

eddyashton: #1756

Issue #1701, opened 2020-10-02
LeakSanitiser break on some shutdowns in Daily build


achamayou:
eg. https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=13608&view=results

```
Tracer caught signal 11: addr=0x0 pc=0x4f4620 sp=0x7f4a60171d30
==1605==LeakSanitizer has encountered a fatal error.
==1605==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==1605==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
```

achamayou:
Turning on the suggested options does not reveal much more:

```
2020-10-09T11:54:12.8473188Z 48: ==2952==AddressSanitizer: libc interceptors initialized
2020-10-09T11:54:12.8473533Z 48: || `[0x10007fff8000, 0x7fffffffffff]` || HighMem    ||
2020-10-09T11:54:12.8473863Z 48: || `[0x02008fff7000, 0x10007fff7fff]` || HighShadow ||
2020-10-09T11:54:12.8474199Z 48: || `[0x00008fff7000, 0x02008fff6fff]` || ShadowGap  ||
2020-10-09T11:54:12.8474523Z 48: || `[0x00007fff8000, 0x00008fff6fff]` || LowShadow  ||
2020-10-09T11:54:12.8474861Z 48: || `[0x000000000000, 0x00007fff7fff]` || LowMem     ||
2020-10-09T11:54:12.8475251Z 48: MemToShadow(shadow): 0x00008fff7000 0x000091ff6dff 0x004091ff6e00 0x02008fff6fff
2020-10-09T11:54:12.8475763Z 48: redzone=16
2020-10-09T11:54:12.8475954Z 48: max_redzone=2048
2020-10-09T11:54:12.8476175Z 48: quarantine_size_mb=256M
2020-10-09T11:54:12.8476418Z 48: thread_local_quarantine_size_kb=1024K
2020-10-09T11:54:12.8476660Z 48: malloc_context_size=30
2020-10-09T11:54:12.8476887Z 48: SHADOW_SCALE: 3
2020-10-09T11:54:12.8477185Z 48: SHADOW_GRANULARITY: 8
2020-10-09T11:54:12.8477559Z 48: SHADOW_OFFSET: 0x7fff8000
2020-10-09T11:54:12.8477816Z 48: ==2952==Installed the sigaction for signal 11
[...]

PR #1757, opened 2020-10-14
Enable SAN on libraries built in separate cmake files

achamayou:
While looking at #1701, I noticed that we do not use the sanitiser when building some external libraries, which means that we potentially ignore some issues.

It's unlikely that we can fix all of them, but we ought to be aware of them at least, and chase the ones that have a serious impact.


cmake/common.cmake
```diff
@@ -73,6 +73,7 @@ find_package(MbedTLS REQUIRED)
 set(CLIENT_MBEDTLS_INCLUDE_DIR "${MBEDTLS_INCLUDE_DIRS}")
 set(CLIENT_MBEDTLS_LIBRARIES "${MBEDTLS_LIBRARIES}")
 
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/tools.cmake)
```
eddyashton, line 76: Looks like this file isn't included in the PR.

achamayou:
I think that's all of them, the only one that looks suspicious is the zero size memcpy happening in the MerkleTree in the snapshot recovery test, which @jumaffre is taking a look at.

This is otherwise locking in the current situation to avoid further regressions.

Separately, I will try further steps to fix the LeakSanitiser break (#1701).

jumaffre:
The MerkleTree issue has been identified and is now tracked here: https://github.com/project-everest/hacl-star/issues/376

eddyashton:
Ignoring unrelated failure in the Daily CI, merging anyway.

PR #1763, opened 2020-10-15
Add Vegeta stress test to CI

eddyashton:
This adds 2 new CTests:

- `vegeta_stress_test`
- `vegeta_long_stress_test`

These use [Vegeta](https://github.com/tsenart/vegeta) to pump a regular rate of transactions into the logging app. The workload it provides isn't very interesting, but hopefully we can generalise this to build more powerful/interesting/custom request streams. 

The `long_stress` test runs for 45m (!), so is put into a custom CTest 'configuration' such that it won't be run by accident/in normal CI. I've set up a separate CI pipeline which runs just this test, once a day. I think its useful to have it separate from the Daily so it can be triggered in isolation, but that's up for discussion.


tests/generate_vegeta_targets.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
eddyashton, line 1: This script is unused, but its an example of what I was using initially to test by hand. I'll remove it now rather than in-a-month-when-we-remember, but if anyone else wants to play with these tests its good to know you _can_ build request streams with jq manipulation.

ashamis: This seems like somethign that is really useful. Would we want to keep something like this in the DevEnv branch?

ashamis (approved):

ashamis (approved):

achamayou (approved):

PR #1766, opened 2020-10-15
Fix sandbox

letmaik:
Backwards compatible changes to the infra code to fix an issue with launching sandbox.sh from arbitrary folders.
This also uses `exec` to launch python so that signals are forwarded when using the script programmatically.

### DIFF
--- a/tests/sandbox/sandbox.sh
+++ b/tests/sandbox/sandbox.sh
@@ -21,15 +21,13 @@ source "${VENV_DIR}"/bin/activate
 if [ -f "${VERSION_FILE}" ]; then
     # install tree
     BINARY_DIR=${PATH_HERE}
-    LIBRARY_DIR="${PATH_HERE}"/../lib
     START_NETWORK_SCRIPT="${PATH_HERE}"/start_network.py
     VERSION=$(<"${VERSION_FILE}")
     pip install --disable-pip-version-check -q -U ccf=="$VERSION"
     pip install --disable-pip-version-check -q -U -r "${PATH_HERE}"/requirements.txt
 else
     # source tree
-    BINARY_DIR="${PATH_HERE}"/../../build
-    LIBRARY_DIR=${BINARY_DIR}
+    BINARY_DIR=.
     START_NETWORK_SCRIPT="${PATH_HERE}"/../start_network.py
     pip install --disable-pip-version-check -q -U -e "${PATH_HERE}"/../../python/
     pip install --disable-pip-version-check -q -U -r "${PATH_HERE}"/../requirements.txt
@@ -40,7 +38,6 @@ echo "Python environment successfully setup"
 export CURL_CLIENT=ON
 exec python "${START_NETWORK_SCRIPT}" \
     --binary-dir "${BINARY_DIR}" \
-    --library-dir "${LIBRARY_DIR}" \
     --enclave-type virtual \
     --initial-member-count 1 \
     --initial-user-count 1 \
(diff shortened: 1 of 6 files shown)


tests/sandbox/sandbox.sh
```diff
@@ -35,6 +39,8 @@ echo "Python environment successfully setup"
 
 CURL_CLIENT=ON \
     python "${START_NETWORK_SCRIPT}" \
+    --binary-dir "${BINARY_DIR}" \
+    --library-dir "${LIBRARY_DIR}" \
```
eddyashton, line 43: I'm not sure that the `--library-dir` is the right way to fix this. We already have a lot of logic to detect/build the library name, and this is adding yet another layer. I think we get the same result by passing the full path to `js_generic` as the `--package` arg here?

tests/sandbox/sandbox.sh
```diff
@@ -20,12 +20,16 @@ source "${VENV_DIR}"/bin/activate
...
     # install tree
+    BINARY_DIR=${PATH_HERE}
+    LIBRARY_DIR="${PATH_HERE}"/../lib
     START_NETWORK_SCRIPT="${PATH_HERE}"/start_network.py
     VERSION=$(<"${VERSION_FILE}")
     pip install --disable-pip-version-check -q -U ccf=="$VERSION"
     pip install --disable-pip-version-check -q -U -r "${PATH_HERE}"/requirements.txt
 else
     # source tree
+    BINARY_DIR="${PATH_HERE}"/../../build
```
eddyashton, line 31:
```suggestion
    BINARY_DIR="."
```

Minor, but if I'm in a folder called `build.debug` or `build.san`, I'd like to be able to use `sandbox.sh` with my local binaries. If I'm outside of any `build` directory, I _can_ still use `sandbox.sh`, but I need to specify `--binary-dir` explicitly to say which `cchost` I want to use.

tests/sandbox/sandbox.sh
letmaik: I think you would specialize it too much in that case, it wouldn't be easy to switch between different enclave types then for example.

letmaik:
If anyone has an idea why library-dir is turned into an empty list in tests, please shout.

tests/sandbox/sandbox.sh
eddyashton: Sorry, the full _prefix_ to `js_generic`, but not the full path. So it would be either `-p /opt/ccf/lib/libjs_generic` or `-p ../../build/libjs_generic`, and this will still get the appropriate suffix for different enclave types.

ie
```
>>> build_lib_path("/opt/ccf/lib/libjs_generic")
'/opt/ccf/lib/libjs_generic.enclave.so.signed'
>>> build_lib_path("../../libjs_generic", enclave_type="virtual")
'./../../libjs_generic.virtual.so'
```

letmaik: My goal is to have a half-way sane dev/user experience. In the long run I think binary/library dir should be removed and replaced by some ccf_dir or similar, which then works on an install prefix and a build folder, by making the build folder layout the same as the install folder. Lots of projects actually do this as it simplifies these scenarios.
In the short run I'd like the following:
- Don't make the existing dev experience worse. As you said, running from a custom build folder should still work.
- Keep it easy to choose a different base path, ideally just changing one folder.
- Keep the semantics of package as they are, don't pretend it can be a full path prefix, it's hard to explain to people.

I made some changes in the PR which I think accomplish all of that.
- Library dir is still a thing but an advanced option that doesn't have to be touched, it will be derived from binary dir in both cases (install/dev tree). Having library dir in the code base makes sense as it's orthogonal to binary dir. 
- In sandbox.sh, binary dir defaults to "." when in a dev tree. Changing it to a custom build folder works fine and will locate the libraries as well.
- Specifying a different package still works as before, just using the simple name.

jumaffre (approved):

eddyashton (approved):

PR #1769, opened 2020-10-15
Remove old Store-less Tx constructor

eddyashton:
Thanks to the recent changes to `ProgressTracker`, we no longer need the old (unsafe!) `Tx` constructor any more, so this removes it.


achamayou (approved):
Fantastic!

ashamis (approved):

Issue #1709, opened 2020-10-05
AFT: View changes


ashamis:
Get AFT view-changes to work

Trigger view-change
- [x] Detect if the primary is not issuing transactions in a timely manner (#1770)
- [x] Detect if the primary is not creating signatures or adding evidence to the ledger in a timely manner (#1813)

Execute view-change
- [x] Every replica needs to create and send a "view-change" message (#1835)
- [x] When the new proposed primary collects enough evidence it needs to send a "new-view" message (#1861)
- [x] Before applying a view-change the replica needs to validate the view-change message on the backups (#1861, #1864)
- [ ] When a replica receives an append entry it needs to validate that is comes from a valid primary  (#1864, #1881)

ashamis:
For the bullet point "Before applying a view-change the replica needs to validate the view-change message on the backups". The replicas needs to make sure the new-view tx has:
- The that the replica received the correct number of signatures.
- Ensure that the progress tracker agrees for which seqno it is applying the new-view tx for.

ashamis:
for the bullet point " When a replica receives an append entry it needs to validate that is comes from a valid primary". There are some checks we need to do.
- Check if the replica that is sending said data can be the primary for that view i.e. we determine the primary deterministically based on the view
- Then check if we have the required evidence for said view

Issue #1690, opened 2020-10-01
Create ability to change default hashing algorithm


ashamis:
We need a way to change the default hashing algorithm in CCF. This can probably be done via a compile time macro.

ashamis:
Closing this work item, I believe everything was move to crypto::Sha256Hash class which should let us switch to a different hashing algorithm if we want (regardless of the class name :))

PR #1770, opened 2020-10-15
AFT: Send request hash when forwarding request

ashamis:
Part of #1709

When a replica forwards a request it will send a hash of the forwarded request to all the other backup nodes. If said request is not executed within a specific period if time we will issue a view-change. This PR introduces the RequestTracker class which is responsible for matching up the hashes which are sent to the requests that are seen by the backups.

### DIFF
--- a/src/consensus/aft/impl/execution.cpp
+++ b/src/consensus/aft/impl/execution.cpp
@@ -50,7 +50,9 @@ namespace aft
   }
 
   kv::Version ExecutorImpl::execute_request(
-    std::unique_ptr<RequestMessage> request, bool is_create_request)
+    std::unique_ptr<RequestMessage> request,
+    bool is_create_request,
+    std::shared_ptr<aft::RequestTracker> request_tracker)
   {
     std::shared_ptr<enclave::RpcContext>& ctx = request->get_request_ctx().ctx;
     std::shared_ptr<enclave::RpcHandler>& frontend =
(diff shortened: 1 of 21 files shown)


src/consensus/aft/impl/execution.cpp
```diff
@@ -60,6 +62,24 @@ namespace aft
...
       NoNode, ctx->pbft_raw.data(), ctx->pbft_raw.size());
 
+    if (request_tracker != nullptr)
+    {
+      const auto& raw_request = ctx->get_serialised_request();
+      auto data_ = raw_request.data();
+      auto size_ = raw_request.size();
+
+      std::array<uint8_t, 32> hash;
+      tls::do_hash(data_, size_, hash, MBEDTLS_MD_SHA256);
```
achamayou, line 72: We have a bunch of `crypto::Sha256Hash` in other places, we ought to tidy up our hashing and have it all go through a single interface, which isn't too difficult to switch to a different digest algorithm if/when we need to.

src/consensus/aft/raft.h
```diff
@@ -176,6 +184,22 @@ namespace aft
       return leader_id;
     }
 
+    std::set<NodeId> backups()
```
achamayou, line 187: I'm a little worried about this, because this is really the latest backups, but not necessarily all the backups a given Tx ought to go to, for example if a node has been removed but is behind. Am I missing something?

src/node/request_tracker.h
```diff
@@ -0,0 +1,168 @@
...
+    void insert(
+      const std::array<uint8_t, 32>& hash,
+      std::chrono::milliseconds time,
+      std::multiset<Request*, RequestComp>& requests_,
+      snmalloc::DLList<Request, std::nullptr_t, true>& requests_list_)
+    {
+      CCF_ASSERT_FMT(
+        requests_list_.get_tail() == nullptr ||
+          requests_list_.get_tail()->time <= time,
+        "items not entred in the correct order. last:{}, time:{}",
```
achamayou, line 133:
```suggestion
        "items not entered in the correct order. last:{}, time:{}",
```

src/node/request_tracker.h
```diff
@@ -0,0 +1,168 @@
...
+    {
+      return requests.empty() && requests_list.is_empty() &&
+        deleted_requests.empty() && deleted_requests_list.is_empty();
+    }
+
+  private:
+    std::multiset<Request*, RequestComp> requests;
+    snmalloc::DLList<Request, std::nullptr_t, true> requests_list;
+
+    std::multiset<Request*, RequestComp> deleted_requests;
```
achamayou, line 121: I don't think I understand correctly what a deleted request is, @ashamis if you could spare a few minutes to walk me though this, I'd appreciate it :)

src/consensus/aft/raft.h
achamayou: I think this should all_active_backups() and be the union of all backups in active configurations.

achamayou (approved):

src/consensus/aft/impl/execution.cpp
ashamis: https://github.com/microsoft/CCF/issues/1690

Issue #1742, opened 2020-10-12
Pass proposer id to the constitution and remove implicit voting


achamayou:
If the proposer id is passed to the constitution, it can implement implicit voting if/when desired, and there is no need for an explicit governance-level mechanism for it.

- [x] Pass proposer id
- [x] Convert operator sample governance to use that rather than implicit voting to decide when to pass
- [x] Remove implicit voting

PR #1773, opened 2020-10-16
Wait for commit on open_network proposal

jumaffre:
@letmaik noticed that with the new `sandbox.sh` script, the network was reported as open before it actually was: because the sandbox only creates one member, the Python infra logic skipped the wait for commit logic because no vote was issued. We now explicitly wait for the proposal to be committed. 

Note that this might need to be revisited once #1742 is implemented.


eddyashton (changes requested):
Rather than adding a `wait_for_global_commit` parameter to another function, I think we should change `vote_using_majority(wait_for_global_commit=True)` so that it always waits, even if it hasn't voted? I think we could replace the wait inside `member.vote()` with a wait at the end of `vote_using_majority`, waiting on either the last response to a vote _or_ just waiting for the current `/commit` to be global, and get the behaviour we want everywhere?

jumaffre:
@eddyashton Good point! Let me fix this now.

tests/infra/member.py
```diff
@@ -81,7 +79,12 @@ def set_active(self):
         # Use this with caution (i.e. only when the network is opening)
         self.status_code = MemberStatus.ACTIVE
 
-    def propose(self, remote_node, proposal, has_proposer_voted_for=True):
+    def propose(
+        self,
+        remote_node,
+        proposal,
+        has_proposer_voted_for=True,
```
eddyashton, line 86: Unrelated nitpick: Its a little unfortunate that _Black_ does this. If it ever inserts a trailing comma, it won't remove it. There's several issues around this on their repo. If I was writing an opinionated, unconfigurable formatter with an aim to "produce the smallest diffs possible", I'd enforce that this trailing-comma was _always_ present, and lists of all kinds were always split over multiple lines. Remove the spill/unspill hysteresis entirely. Rant over.

eddyashton (approved):

PR #1774, opened 2020-10-16
Implement has(k) on TxView

eddyashton:
Adds a native `.has()` method. This is more efficient than `.get()` as it doesn't deserialise/copy the value.

### DIFF
--- a/src/kv/untyped_tx_view.h
+++ b/src/kv/untyped_tx_view.h
@@ -23,15 +23,61 @@ namespace kv::untyped
 
   class TxView
   {
-  protected:
-    ChangeSet& tx_changes;
-
   public:
     // Expose these types so that other code can use them as MyTx::KeyType or
     // MyMap::TxView::KeyType, templated on the TxView or Map type
     using KeyType = SerialisedEntry;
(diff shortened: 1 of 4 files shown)


src/kv/untyped_tx_view.h
```diff
@@ -112,6 +112,52 @@ namespace kv::untyped
...
+     * key has been updated (written to or deleted) in the current transaction,
+     * that update will be reflected in the return of this call.
+     *
+     * @param key Key
+     *
+     * @return bool true iff key exists
+     */
+    bool has(const KeyType& key)
+    {
+      // NB: In terms of dependencies, this is a read equivalent to a get
```
jumaffre, line 127: Why not calling `get(key).has_value()` here?

ashamis (approved):

src/kv/untyped_tx_view.h
eddyashton: Because `get(key)` constructs a copy of the value, which we'd rather not do if we don't have to. I'll try to rewrite this so they both call the same access non-copying function to avoid this duplication.

PR #1778, opened 2020-10-19
Fix Daily CI - pass absolute package path

eddyashton:
The recent `--library-dir` means that relative package paths must be relative to the binary dir, not the working directory. For simplicity, the path in `test_install.sh` is now an absolute path.

Unrelated tweak in the same script - wait for `network_info.txt` to appear as a sign the network is open.

### DIFF
--- a/tests/test_install.sh
+++ b/tests/test_install.sh
@@ -28,13 +28,13 @@ python -m pip install ../../../python
 # Start ephemeral network in the background
 network_info_file="network_info.txt"
 
-package="$(pwd)"/../../../build/liblogging
 
 network_live_time=30
 timeout --signal=SIGINT --kill-after=${network_live_time}s --preserve-status ${network_live_time}s \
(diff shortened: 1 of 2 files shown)


achamayou (approved):

tests/test_install.sh
```diff
@@ -26,26 +28,38 @@ python -m pip install ../../../python
...
+sleep 5
 
 # Recover network
 cp -r ./workspace/start_network_0/0.ledger .
 cp ./workspace/start_network_0/network_enc_pubk.pem .
 
-timeout --signal=SIGINT --kill-after=30s --preserve-status 30s \
+timeout --signal=SIGINT --kill-after=${network_live_time}s --preserve-status ${network_live_time}s \
 python "$INSTALL_PREFIX"/bin/start_network.py \
-    -p ../../../build/liblogging \
```
letmaik, line 48: Why don't you set library-dir instead? That's exactly the use case here

PR #1781, opened 2020-10-20
Truly delete things in the KV, and add tests of heap use

eddyashton:
The core of this is to use the new `remove` in the champ, to really delete things rather than maintaining negative-version tombstones.

To validate this, there's a new test in `kv_serialisation.cpp` which confirms that snapshots get bigger/smaller/minimal as we add/remove/clear keys.

There's also an e2e `memory_test`. Currently this just confirms that the contents do what we expect (max is constant, peak never decreases). There's a few commented-out relationships we'd like to be true, but currently aren't - we see slight growth even when we should be working with a smaller working set, I don't know if this is just to a known limit or a sign of something broken.


eddyashton:
Converting this to a draft while I explore a potential issue (do we no longer have a conflict with a written-then-deleted key we read, while we're executing, which conflicted previously?)

eddyashton:
Parking this until we have some free time (and stable tests) to confirm the precise conflict behaviour. In particular, there's 2 cases we want to consider:

- If transaction A reads that K doesn't exist then tries to write to it, then other transactions create K then delete K (both steps being committed), should it conflict? Put another way: A sees that K doesn't exist at Version N. At version N+1 K is created by another transaction, at version N+2 it is deleted. Then A attempts to commit at version N+3. Logically, it seems that its dependence on K _not existing_ (with "never existed" and "was deleted" being equivalent, as far as the transaction logic is concerned) is still true, so it can safely commit?
- Do we correctly do read-your-own-write dependence on keys we deleted ourselves, whether or not they previously existed?

eddyashton:
Closing as I don't have time to complete this currently - tracking the task in #1992.

PR #1783, opened 2020-10-20
Support templated paths for JS endpoints

eddyashton:
Follow up to #1744, adding support for OpenAPI-style templated paths (ie - `/app/my_endpoint/{id}/{foo}`) in JS endpoints.

In the `js_generic` implementation of `find_endpoint` we do similar work to the base implementation, iterating through a set of templated paths and seeing if we can build a regex from them that matches the current URI path. The static case is more efficient since we know these templated paths in advance, so can pre-build the regexes and iterate through only those endpoints. In the dynamic case, where we're always reading the current paths from the KV, we to iterate through all defined endpoints and construct the regex on-demand.

While doing this I've added error handling whenever there are ambiguous templated paths (the current path _could_ match one of multiple endpoints). Ideally this is something we'd catch at installation time, or when the dynamic app is deployed, and report an error then, but for now we'll make do with reporting when its seen in dispatch. In other words if you have `/app/AAA/{foo}` and `/app/AAA/{bar}`, you'll only see an error if and when you call `/app/AAA/...`.

To test this, there's an additional templated endpoint defined in the `js-app-bundle` test. Perhaps we should build a second `app2.json` to hold all our edge case test cases, so we can keep a minimal `app.json` as a Hello World example.

### DIFF
--- a/.azure-pipelines-templates/test.yml
+++ b/.azure-pipelines-templates/test.yml
@@ -5,7 +5,7 @@ parameters:
 
 steps:
 - script: |
-    ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
+    ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }} -R modules_test --repeat-until-fail 20
   env:
     METRICS_MONGO_CONNECTION: $(METRICS_MONGO_CONNECTION)
   displayName: CTest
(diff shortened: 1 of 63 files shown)


.azure-pipelines-templates/test.yml
```diff
@@ -5,7 +5,7 @@ parameters:
 
 steps:
 - script: |
-    ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
+    ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }} -R modules_test --repeat-until-fail 20
```
achamayou, line 8: Debugging the modules test? Please don't forget to remove it before merging :)

achamayou (approved):

PR #1787, opened 2020-10-21
Forum sample

letmaik:


samples/apps/forum/app.tmpl.json
```diff
@@ -0,0 +1,106 @@
+{
+  "endpoints": {
+    "/polls": {
+      "post": {
```
achamayou, line 4: This is a fairly random, unrelated thought, but I wonder if we couldn't:

1. default all the flags the same way CCF does, so they only need specifiying if they are different
2. Allow "js_module" to be inherited from a parent scope, because it'll often be the same for all endpoints.

@letmaik @eddyashton what do you think?

samples/apps/forum/src/controllers/poll.ts
```diff
@@ -0,0 +1,305 @@
...
+        } catch (e) {
+            throw new Error(`Poll with topic '${topic}' does not exist`)
+        }
+
+        const opinionCountAboveThreshold = Object.keys(poll.opinions).length >= MINIMUM_OPINION_THRESHOLD
+
+        let response: GetPollResponse
+        // TODO can repetition be avoided while maintaining type checking?
+        if (poll.type == "string") {
+            response = {
```
achamayou, line 278: Seems like at least response = { type: poll.type } should hoist out without trouble?

samples/apps/forum/test/demo/generate-opinions.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
achamayou, line 1: Closing #1784 then.

tests/npm-tsoa-app/src/types/ccf.ts
```diff
@@ -70,7 +70,7 @@ export interface CCF {
 export const ccf = globalThis.ccf as CCF
 
 // Additional functionality on top of C++:
-export const kv = ccf.kv
+export const kv = ccf ? ccf.kv : undefined
```
achamayou, line 73: For testing I guess?

achamayou (approved):
LGTM, thank you @letmaik! I'll add a separate demo_user_opinions.csv and expand on the doc/move it to Sphinx.

samples/apps/forum/app.tmpl.json
letmaik: You could introduce a new top-level `"endpoint_defaults": {..}` object which has the defaults. I wouldn't inherit from CCF directly, explicit is better here I think.

tests/npm-tsoa-app/src/types/ccf.ts
letmaik: Added a comment.

samples/apps/forum/app.tmpl.json
letmaik: These are not ccf defaults, rather common options set by the app developer.

achamayou: They are common options, but right now they are specified every time even when they use the default CCF value, which makes the configuration quite verbose, at least in my opinion :)

In C++, that's not typically the case, the path and the verb are mandatory when creating the endpoint, everything else is optional and can be `.set...()`, I think that's reasonable here too. Happy to discuss thought.

letmaik: I think you misunderstand. I'm happy with making the per-endpoint options optional, but only in exchange for a separate top-level definition with those common options. Like:
```json
{
  "endpoint_common": {
        "forwarding_required": "always",
        "execute_locally": false,
        "require_client_identity": false,
        "require_client_signature": false,
        "readonly": false
  },
  "endpoints": {
    "/polls": {
      "post": {
        "js_module": "build/PollControllerProxy.js",
        "js_function": "createPoll",
      },
...
```

eddyashton: FWIW, I think its more common (and reasonable) for config files like this to be unnecessarily verbose and explicit, since it makes modifications obvious and in-place (what options are available? What's the current value?). Whereas in the code definition, its more reasonable to say "I'd like to save space and don't care what my options are, or what their defaults are". 

That said, I'm fine with eliding the full config here so long as the available options are visible somewhere (ie - we have a test/sample that's verbose, that we know we can go to for "what exactly is this property called?" without looking it up directly in the equivalent C++ struct). The `"endpoint_defaults": {...}` isn't a pattern we've used elsewhere, but we did discuss doing something similar to group endpoint properties - I think its a nice balance of maintaining local definitions without obscuring the default values.

achamayou: @letmaik no, that's exactly what I'd understood, and what I'd prefer not to have, but I can see that I am in the minority here :)

Apart from the verbosity, which is in my opinion excessive (particularly since it's repeated for every endpoint) but I accept may be a matter of taste, I am worried about changing the set of available options in CCF and the impact on configs. I don't expect we will modify/remove options all that much, but I fully expect we will add to them.

Having users specify their delta to defaults is trivially forward compatible, forcing them to specify the default (once per endpoint!) for a new key sounds quite unpleasant.
I'm trying to picture what our Azure Pipeline configs would look like if we had to specify every last value, and I shudder.

letmaik: "forcing them to specify the default (once per endpoint!)" -> It's not per endpoint, its for all endpoints, see my sample above.

PR #1789, opened 2020-10-21
Split stub to avoid importing unnecessary symbols in evercrypt.a

achamayou:
As pointed out by @soren-da, the fact that stub.c contains stubs meant for quickjs and evercrypt leads to `evercrypt.a` containing symbols such as `gettimeofday()` that collide with symbols from Open Enclave when enabling the corresponding syscalls in the EDL. CCF itself does not make use of those syscalls, but application code may want to.

This needs to be revisited once quickjs moves from `js_generic` to become a core dependency (although it will remain a separate .`a`), but for now this resolves the issue.


cmake/crypto.cmake
```diff
@@ -18,7 +18,7 @@ file(GLOB_RECURSE EVERCRYPT_SRC "${EVERCRYPT_PREFIX}/*.[cS]")
 
 if("sgx" IN_LIST COMPILE_TARGETS)
   add_library(
-    evercrypt.enclave STATIC ${EVERCRYPT_SRC} ${CCF_DIR}/3rdparty/stub/stub.c
+    evercrypt.enclave STATIC ${EVERCRYPT_SRC} ${CCF_DIR}/3rdparty/stub/mem.c
```
eddyashton, line 21: Looks like this file isn't included in the PR? `stub.c` removed, `time.c` added, but no `mem.c`

achamayou: Weird, I could have sworn I'd git mv'ed stub.c mem.c, and that's why I added only the other one. Oh well. Thanks!

eddyashton (approved):
I've tested this with an external repo (which actually called `gettimeofday` to trigger the error), and can confirm that this fixes it.

PR #1793, opened 2020-10-22
Add Microsoft login to Forum sample

letmaik:
Note that this is still without token validation.

### DIFF
--- a/samples/apps/forum/src/controllers/site.ts
+++ b/samples/apps/forum/src/controllers/site.ts
@@ -34,6 +34,10 @@ const HEADER_HTML = `
         padding: 3rem 1.5rem;
         text-align: center;
       }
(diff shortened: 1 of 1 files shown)


achamayou (approved):

samples/apps/forum/src/controllers/site.ts
```diff
@@ -34,6 +34,10 @@ const HEADER_HTML = `
         padding: 3rem 1.5rem;
         text-align: center;
       }
+      #login-btn {
+          display: none;
+          margin-left: 15px;
```
achamayou, line 39: Since we're using rem/em, I suggest we do here too rather than px :)

letmaik: Done.

PR #1794, opened 2020-10-22
Forum sample: switch to URL templates & add more tests

letmaik:


achamayou:
@letmaik is it still happening now that #1795 is merged?

letmaik:
All working.

samples/apps/forum/src/controllers/poll.ts
```diff
@@ -71,7 +70,8 @@ interface GetPollsResponse {
 
 // Export REST API request/response types for unit tests
 export {
```
achamayou, line 72: Do we use an autoformatter for TS? If not, should we pick one and hook it up to `ci-checks.sh`?

achamayou (approved):

PR #1803, opened 2020-10-22
Explicit node flags

achamayou:
The current logic used to decide whether to start locally or to use the SSHRemote is based on the ability to bind to the address we target. Since we've switched to a harcoded default for sandbox, this raises the chances that we will silently use the SSH implementation where we don't intend.

This changes forces the caller to explicitly spell out what implementation it intends to use, rather than extend the logic further with code that discovers which interfaces are available locally (it's not just localhost). There's enough complexity in the test infra as it is.

Before the change:

`./tests.sh -n 127.0.0.1 -n remote.domain.com -n thisbox.domain.com`, where `thisbox.domain.com` resolves to a local interface.

After the change:

`./tests.sh -n local://127.0.0.1 -n ssh://remote.domain.com -n local://thisbox.domain.com`


letmaik:
`samples/apps/forum/test/util.ts` should be adapted as well.

achamayou:
@letmaik done.

eddyashton (approved):

tests/ws_scaffold.py
```diff
@@ -47,18 +47,15 @@ def test(network, args):
...
         network.start_and_join(args)
         test(network, args)
 
 
 if __name__ == "__main__":
-
     args = infra.e2e_args.cli_args()
     args.package = args.app_script or "liblogging"
-
+    args.nodes = infra.e2e_args.min_nodes(args, f=1)
```
achamayou, line 60: Not equivalent to what we had before, where we rather bizarrely had f=1 in bft and f=0 in cft. I think this is what we want in this case though, because we're checking transactions are committed.

achamayou:
This is also being a little more explicit about the network layout we want, there are essentially three configs we tend to use:

1. A single node, when testing the schema for example, or aspects of a node that do not involve replication.
2. `max_nodes(f=0)` tests where we want to exercise replication, and we want commit to get stuck if replication fails. This should be our default in most cases, it's also the cheapest in terms of setup with replication.
3. `min_nodes(f=1)` or higher, when the test is deliberately testing node faults.

There is a single exception for now, which is the committable test, hardcoded to 5 nodes. That should be `min_nodes(f=2)` and the logic should be parametrised on f (#1806).

Issue #1780, opened 2020-10-20
JS: Return undefined for non-existing keys in <kvmap>.get


letmaik:
Currently `<kvmap>.get(key)` throws an exception if the key doesn't exist. This is different to how all storage-related JS APIs behave, e.g. `Map` and `Storage`. It also makes things like `<kvmap>.get(key) || default_val` impossible. We should consider returning `undefined` instead.

eddyashton:
I definitely think `undefined` is preferable here. As well as consistency with similar JS APIs, its equivalent to the C++ API (which returns a `std::optional`). Does anyone have an argument in favour of the current approach?

achamayou:
+1 for compliance with ES standards, although as of ES2020 you probably want `<kvmap>.get(key) ?? default_val` rather than `||`.

letmaik:
There is a similar inconsistency with `delete()` which always returns `undefined` and throws an exception if the key doesn't exist, whereas [`Map.delete()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete) returns "true if an element in the Map object existed and has been removed, or false if the element does not exist."
I suggest to change this method as well.

PR #1807, opened 2020-10-22
Align js kvmap API to standard Map

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1780.

Changes:
- `get(key)` returns `undefined` instead of throwing an exception if `key` does not exist
- `delete(key)` returns a `false` instead of throwing an exception if `key` does not exist, and `true` instead of `undefined` otherwise 
- `set(key, val)` returns the map object instead of `undefined`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - `node/quote` endpoint now returns a single JSON object containing the node's quote (#1761).
 - Calling `foreach` on a `TxView` now iterates over the entries which previously existed, ignoring any modifications made by the functor while iterating.
+- JS: `ccf.kv.<map>.get(key)` returns `undefined` instead of throwing an exception if `key` does not exist
+- JS: `ccf.kv.<map>.delete(key)` returns `false` instead of throwing an exception if `key` does not exist, and `true` instead of `undefined` otherwise 
+- JS: `ccf.kv.<map>.set(key, val)` returns the map object instead of `undefined`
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - `node/quote` endpoint now returns a single JSON object containing the node's quote (#1761).
 - Calling `foreach` on a `TxView` now iterates over the entries which previously existed, ignoring any modifications made by the functor while iterating.
+- JS: `ccf.kv.<map>.get(key)` returns `undefined` instead of throwing an exception if `key` does not exist
+- JS: `ccf.kv.<map>.delete(key)` returns `false` instead of throwing an exception if `key` does not exist, and `true` instead of `undefined` otherwise 
+- JS: `ccf.kv.<map>.set(key, val)` returns the map object instead of `undefined`
```
eddyashton, line 15:
```suggestion
- JS: `ccf.kv.<map>.get(key)` returns `undefined` instead of throwing an exception if `key` does not exist.
- JS: `ccf.kv.<map>.delete(key)` returns `false` instead of throwing an exception if `key` does not exist, and `true` instead of `undefined` otherwise.
- JS: `ccf.kv.<map>.set(key, val)` returns the map object instead of `undefined`.
```

Terminating periods for consistency with the rest of the changelog.

I also wonder if we should be recommending `ccf.kv["map"]` over `ccf.kv.map`, since we have some map names with awkward characters? But it doesn't really matter here.

achamayou (approved):

PR #1808, opened 2020-10-22
Spell checking pass over the RST docs

eddyashton:
Ran [this spellchecker extension](https://sphinxcontrib-spelling.readthedocs.io/en/latest/index.html#) over our docs, and fixed all of the real typos I spotted.

I don't think this is clean enough to run in our CI yet - we'd need to do a pass standardi~~s~~**z**ing on American Englizh, add even more technical terms to the exceptions lists, decide how many semi-official portmanteaus we allow, and still get some false positives from an oddly-useless tokenizer. I've included the wordlist I was using in case anyone else wants to go further with this. For an estimate of the effort-benefit ratio, note that the (incomplete) manual exception list is longer than the list of corrections - our spelling was msolty pterty godo.


doc/conf.py
```diff
@@ -215,6 +215,12 @@
     "member-order": "bysource",
 }
 
+# sphinxcontrib.spelling options
```
achamayou, line 218: That presumably needs to be added to doc/requirements.txt.

achamayou: And to extensions? How does this even work?

eddyashton: Its not added to our extensions or requirements, its not run as part of our docs build. I've run it by hand, here's the options if anyone else wants to run it by hand, and we may decide to enforce it in the near future.

achamayou: Oh ok, is there a reason we shouldn't add it as part of the doc build? It seems useful and you've built the extra dictionary?

eddyashton: So I didn't want to add it as it still reports a bunch of warnings (we have a mix of American and British English, and still a lot of words that aren't in the extra dictionary), and I don't think its worth including it in the CI build until its Green. But it turns out that adding it to the `extensions` list doesn't actually run it as part of the normal build - it always need to be run manually/explicitly with `-b spelling`. So I'll add it to the requirements and extensions, but not as a separate CI step yet.

doc/requirements.txt
```diff
@@ -8,4 +8,5 @@ sphinxcontrib-mermaid
 sphinx-multiversion
 pydata-sphinx-theme
 sphinx-copybutton
-sphinxcontrib.openapi
\ No newline at end of file
+sphinxcontrib.openapi
+sphinxcontrib.spelling
```
achamayou, line 12:
```suggestion
sphinxcontrib.spelling
enchant
```

doc/conf.py
eddyashton: So this actually [relies on](https://pyenchant.github.io/pyenchant/install.html#installing-the-enchant-c-library) `libenchant` which is not installed in the CI docker image. I don't think its worth rebuilding these until this is an actual CI requirement, so I've removed it from the extensions again - lets revisit if/when we want to enforce, and update the Docker images then.

achamayou (approved):

Issue #1737, opened 2020-10-10
Remove temp implementation of pthread_spin_trylock


ashamis:
OpenEnclave, at this time, does not provide pthread_spin_trylock. https://github.com/openenclave/openenclave/pull/3641 will bring this into OpenEnclave but for now spin_lock.h implementes it. When the PR is committed we should remove the temp version in spin_lock.h

PR #1814, opened 2020-10-22
Upgrade Open Enclave to latest (0.12rc2)

achamayou:
Unfortunately, I got confused and thought this would bring us the fix for #1737, but I was wrong and we'll have to wait 0.13.


ashamis (approved):
Should the playbooks be also updated?

achamayou:
@ashamis they have, it's the `getting_started/setup_vm/roles/openenclave/vars/common.yml` file in this PR.

If you mean applied to the perf kit too, then yes, but I'm waiting for the CI to be green (and ideally for the PR to merge) before doing it.

Issue #1755, opened 2020-10-13
Fix js_generic leaks in error conditions


letmaik:
There are a few places in the handler function of js_generic that can leak memory if an error condition occurs. In https://github.com/microsoft/CCF/pull/1745 I added a wrapper for quickjs values that frees them when they go out of scope. I applied it only for the new code in that PR but it may be useful in general to tackle this issue. Obviously, for trivial cases a direct call makes more sense.

PR #1816, opened 2020-10-23
js_generic cleanup

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1755.

Tip: Use the "split" view to review this PR.

- Moved JS object creation (`globalThis.ccf`, `globalThis.console`, `request`) into separate functions.
- Separated handler code into runtime/context parts.
- Use `JSAutoFree` to avoid memory leaks in error conditions with tricky control flow.
- Return error if `statusCode` is not an integer (previously silently ignored).


eddyashton (approved):

src/apps/js_generic/js_generic.cpp
```diff
@@ -693,99 +697,9 @@ namespace ccfapp
...
-      auto console = JS_NewObject(ctx);
-      JS_SetPropertyStr(ctx, global_obj, "console", console);
-
-      JS_SetPropertyStr(
-        ctx, console, "log", JS_NewCFunction(ctx, ccfapp::js_print, "log", 1));
-
       auto ccf = JS_NewObject(ctx);
-      JS_SetPropertyStr(ctx, global_obj, "ccf", ccf);
 
       JS_SetPropertyStr(
```
achamayou, line 704: Since we're exposing a growing amount of custom APIs, we ought to document them. I don't think sphinx-js can help us because it wants to parse javascript, but we can use the Javascript domain to expose them manually: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-javascript-domain

achamayou (approved):

src/apps/js_generic/js_generic.cpp
letmaik: Can you create an issue for this?

PR #1819, opened 2020-10-23
Demo tweaks

achamayou:
A few tweaks to the demo, some to get it to work on my machine, others to make a bit smoother in my opinion.


samples/apps/forum/src/controllers/site.ts
```diff
@@ -480,6 +556,7 @@ ${HEADER_HTML}
 
 <script>
 async function main() {
+    await initUser()
```
achamayou, line 559: There seems to be a race here with the call that's made on DOMContentLoaded, and I lose it about half the time. At which point the page sends a request with an empty jwt and gets a rather unfriendly error from the server.

This isn't an issue on the form pages because of the human-induced delay.

awaiting initUser at the beginning of each main() is tempting, but I suspect this is not what a sophisticated web developer would do, so I defer to @letmaik :)

samples/apps/forum/test/demo/submit-opinions.ts
```diff
@@ -44,7 +44,7 @@ async function main() {
         }
         console.log('Submitting opinions for user ' + user)
         try {
-            await bent('PUT', 204)(`${ENDPOINT_URL}/all`, req, getAuth(jwt))
+            await bent('PUT', 204)(`${ENDPOINT_URL}`, req, getAuth(jwt))
```
achamayou, line 47: This caused the submission to yield a slightly funky validation error about unexpected opinions.

samples/apps/forum/src/controllers/site.ts
letmaik: You are referring to the statistics page I assume. Yeah, it's all a bit hacky... definitely a race there. Your suggestion is fine. In a proper project you'd use some framework that deals with all of this.

letmaik (approved):

eddyashton (approved):

PR #1826, opened 2020-10-26
Run forum tests in CI

achamayou:
Virtual mode and CFT only right now, adding more flag passing to enable the rest.


CMakeLists.txt
```diff
@@ -865,4 +865,18 @@ if(BUILD_TESTS)
       10000
       --use-websockets
   )
+
+  if(BUILD_END_TO_END_TESTS)
+    if("virtual" IN_LIST COMPILE_TARGETS)
+      add_test(NAME forum_app
+               COMMAND ${CMAKE_CURRENT_LIST_DIR}/samples/apps/forum/ci.sh
```
eddyashton, line 872: Missing file from this PR.

achamayou: Yep, sorry. It's 2020, and not only does my car not drive itself, but my RCS can't even tell what I mean. I want my AI money back.

CMakeLists.txt
```diff
@@ -822,6 +822,17 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS -p liblogging --duration 45m
   )
 
+  add_e2e_sandbox_test(
```
eddyashton, line 825: If you add an entry for this to `.cmake-format.py` it will be forwarded like our other CMake functions.

eddyashton (approved):

CMakeLists.txt
achamayou: Got it, it's forwarded now :)

eddyashton (approved):

PR #1813, opened 2020-10-22
Aft: Measure interval between global commits

ashamis:
Part of #1709 

Add a check to makes sure too much time or too many transactions do not occur between global commits

PR #1835, opened 2020-10-29
AFT: Send view change messages

ashamis:
Part of #1709 

When running in BFT mode we sometimes decide to send view change messages (#1770, #1813). This PR picks up at the point when we are ready to send a view change message. It includes:
- Create a view-change message will all the relevant evidence and send it to all the other replicas
- Receiving replicas validates the incoming message
- Receiving replicas update their progress tracker with the new evidence
- Rate limit how often a replica will send a view-change message

### DIFF
--- a/src/consensus/aft/impl/view_change_tracker.h
+++ b/src/consensus/aft/impl/view_change_tracker.h
@@ -34,12 +34,17 @@ namespace aft
         ViewChange vc;
         vc.received_view_changes.emplace(my_node_id);
         ++last_view_change_sent;
-        view_changes.insert(std::pair<kv::Consensus::View, ViewChange>(
-          last_view_change_sent, std::move(vc)));
+        view_changes.emplace(last_view_change_sent, std::move(vc));
         time_previous_view_change_increment = time;
         return true;
       }
       return false;
+    }
+
+    bool is_view_change_in_progress(std::chrono::milliseconds time)
+    {
+      return time <=
+        (time_between_attempts + time_previous_view_change_increment);
     }
 
     kv::Consensus::View get_target_view() const
--- a/src/consensus/aft/impl/view_change_tracker.h
+++ b/src/consensus/aft/impl/view_change_tracker.h
@@ -27,12 +27,9 @@ namespace aft
       time_between_attempts(time_between_attempts_)
     {}
 
-    bool should_send_view_change(
-      std::chrono::milliseconds time)
+    bool should_send_view_change(std::chrono::milliseconds time)
     {
-      if (
-        time > time_between_attempts +
-          time_previous_view_change_increment)
+      if (time > time_between_attempts + time_previous_view_change_increment)
       {
         ViewChange vc;
         vc.received_view_changes.emplace(my_node_id);
--- a/src/consensus/aft/impl/view_change_tracker.h
+++ b/src/consensus/aft/impl/view_change_tracker.h
@@ -1,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "consensus/aft/raft_types.h"
+
+#include <chrono>
+#include <map>
+#include <set>
+
+namespace aft
+{
+  class ViewChangeTracker
+  {
+    struct ViewChange
+    {
+      std::set<kv::NodeId> received_view_changes;
+    };
+
+  public:
+    ViewChangeTracker(
+      kv::NodeId my_node_id_,
+      kv::Consensus::View current_view,
+      std::chrono::milliseconds time_between_attempts_) :
+      my_node_id(my_node_id_),
+      last_view_change_sent(current_view),
+      time_between_attempts(time_between_attempts_)
+    {}
+
+    bool should_send_view_change(
+      std::chrono::milliseconds time)
+    {
+      if (
+        time > time_between_attempts +
+          time_previous_view_change_increment)
+      {
+        ViewChange vc;
+        vc.received_view_changes.emplace(my_node_id);
+        ++last_view_change_sent;
+        view_changes.insert(std::pair<kv::Consensus::View, ViewChange>(
+          last_view_change_sent, std::move(vc)));
+        time_previous_view_change_increment = time;
+        return true;
+      }
+      return false;
+    }
+
+    kv::Consensus::View get_target_view() const
+    {
+      return last_view_change_sent;
+    }
+
+    void set_current_view_change(kv::Consensus::View view)
+    {
+      view_changes.clear();
+      last_view_change_sent = view;
+    }
+
+  private:
+    kv::NodeId my_node_id;
+    std::map<kv::Consensus::View, ViewChange> view_changes;
+    std::chrono::milliseconds time_previous_view_change_increment =
+      std::chrono::milliseconds(0);
+    kv::Consensus::View last_view_change_sent = 0;
+    const std::chrono::milliseconds time_between_attempts;
+  };
+}
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -107,6 +107,8 @@ namespace aft
     static constexpr int batch_window_size = 100;
     int batch_window_sum = 0;
 
+    static constexpr size_t starting_view_change = 2;
+
     // Indices that are eligible for global commit, from a Node's perspective
     std::deque<Index> committable_indices;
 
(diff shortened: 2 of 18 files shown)


src/consensus/aft/raft.h
```diff
@@ -157,12 +157,12 @@ namespace aft
       state(state_),
       executor(executor_),
       request_tracker(request_tracker_),
+      view_change_tracker(state_->my_node_id, 2, view_change_timeout_),
```
jumaffre, line 160: We should probably define the initial view as 2 somewhere in this file as it is now used in a quite a few places.

src/crypto/hash.cpp
```diff
@@ -12,45 +12,108 @@ extern "C"
...
-  // update_multi takes complete blocks
-  EverCrypt_Hash_update_multi(state, data_begin, full_blocks_size);
+    // update_last takes start of last chunk (NOT a full block!), and _total
+    // size_
+    EverCrypt_Hash_update_last(state, full_blocks_end, size);
+  }
 
-  // update_last takes start of last chunk (NOT a full block!), and _total size_
-  EverCrypt_Hash_update_last(state, full_blocks_end, size);
+  void Sha256Hash::evercrypt_sha256(const CBuffer& data, uint8_t* h)
```
jumaffre, line 48: Could this function use the new `CSha256Impl` class to avoid calling the Evercrypt code from two places?

src/crypto/hash.cpp
```diff
@@ -12,45 +12,108 @@ extern "C"
...
+  {
+    if (p == nullptr)
+    {
+      throw std::logic_error("Attempting to use hash after it was finalized");
+    }
+
+    Sha256Hash h;
+    p->finalize(h.h);
+    p = nullptr;
+    return std::move(h);
```
jumaffre, line 117: No need to `std::move` local variable thanks to copy elision?

src/node/progress_tracker.h
```diff
@@ -215,13 +216,48 @@ namespace ccf
...
+      {
+        if (!cert.second.is_primary)
+        {
+          continue;
+        }
+
+        if (!cert.second.sig.empty())
+        {
+          LOG_FAIL_FMT(
+            "primary is not empty view:{}, seqno:{}",
```
jumaffre, line 249: Primary's certificate is not empty?

src/consensus/aft/impl/view_change_tracker.h
```diff
@@ -0,0 +1,64 @@
...
+
+    bool should_send_view_change(std::chrono::milliseconds time)
+    {
+      if (time > time_between_attempts + time_previous_view_change_increment)
+      {
+        ViewChange vc;
+        vc.received_view_changes.emplace(my_node_id);
+        ++last_view_change_sent;
+        view_changes.insert(std::pair<kv::Consensus::View, ViewChange>(
+          last_view_change_sent, std::move(vc)));
```
achamayou, line 38:
```suggestion
        view_changes.emplace(last_view_change_sent, std::move(vc));
```

src/consensus/aft/impl/view_change_tracker.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <chrono>
+#include <map>
+#include <set>
+
+namespace aft
+{
+  class ViewChangeTracker
+  {
+    struct ViewChange
```
achamayou, line 15: I'm not entirely clear on what this is. Is that keeping track of acks for a view change proposal effectively?

src/consensus/aft/raft.h
```diff
@@ -533,14 +537,45 @@ namespace aft
                       .get_current_time_offset();
         request_tracker->tick(time);
 
-        if (is_follower() && has_bft_timeout_occurred(time))
+        if (
+          is_follower() &&
+          (state->view_change_in_progress || has_bft_timeout_occurred(time)) &&
+          view_change_tracker.should_send_view_change(time))
         {
+          state->view_change_in_progress = true;
```
achamayou, line 545: Seems odd that this bit of state isn't delegated to the tracker, which seems to also know that there is a pending view change?

src/consensus/aft/raft.h
```diff
@@ -533,14 +537,45 @@ namespace aft
...
+          std::vector<uint8_t> m;
+          m.resize(sizeof(ViewChangeMsg) + vc_size);
+
+          uint8_t* data = m.data();
+          size_t size = m.size();
+
+          serialized::write(
+            data, size, reinterpret_cast<uint8_t*>(&vcm), sizeof(vcm));
+          vc->serialize(data, size);
+          CCF_ASSERT_FMT(size == 0, "Did not write to everything");
```
achamayou, line 567:
```suggestion
          CCF_ASSERT_FMT(size == 0, "Did not write everything");
```

src/consensus/aft/raft_types.h
```diff
@@ -197,6 +198,11 @@ namespace aft
     Nonce nonce;
   };
 
+  struct ViewChangeMsg : RaftHeader
```
achamayou, line 201: Why is a view change message a size?

src/node/node_to_node.h
```diff
@@ -64,6 +64,28 @@ namespace ccf
       return t;
     }
 
+    template <class T>
+    const T& recv_authenticated_with_load(const uint8_t*& data, size_t& size)
```
achamayou, line 68: I don't understand why this is necessary, can't this just be another message type?

tests/election.py
```diff
@@ -79,7 +79,11 @@ def run(args):
         # Number of nodes F to stop until network cannot make progress
         nodes_to_stop = math.ceil(len(args.nodes) / 2)
         if args.consensus == "bft":
-            nodes_to_stop = math.ceil(len(args.nodes) / 3)
+            # nodes_to_stop = math.ceil(len(args.nodes) / 3)
```
achamayou, line 82: f=1 (line 132), so this should be 1 currently anyway?

src/node/progress_tracker.h
```diff
@@ -525,12 +563,102 @@ namespace ccf
...
 
+    std::unique_ptr<ViewChange> get_view_change_message(
+      kv::Consensus::View view)
+    {
+      auto it = certificates.find(highest_prepared_level.version);
+      if (it == certificates.end())
+      {
+        throw ccf::ccf_logic_error(fmt::format(
+          "Invalid prepared level, view:{}, seqno:{}",
+          highest_prepared_level.term,
```
achamayou, line 574: term?

tests/election.py
ashamis: This is actually 2. When view-changes are fully implemented we start the test with 4 nodes, the first view-change will success and the second will fail.

src/node/progress_tracker.h
ashamis: thanks!

src/consensus/aft/impl/view_change_tracker.h
ashamis: It is keeping track of the actual messages (evidence). The way the new primary proves that it can be the new primary is that it sends the evidence to the other replicas and then puts said evidence onto the ledger when it is the primary. This is the final part of #1709

src/consensus/aft/raft.h
ashamis: thank you, I missed deleting this :) the tracker does all this as it should

src/node/progress_tracker.h
ashamis: We are standardizing on size right?

src/consensus/aft/impl/view_change_tracker.h
achamayou: Got it, as discussed I think it would be good to have those be called ViewChangeProposal, ViewChangeRequest, as opposed to ViewChangeComplete,  ViewChangeEvidence or whatever we end up using to refer to the transaction executed by the primary to record the completion of the view change to the ledger once enough valid ViewChangeProposal/Request/Votes have been gathered.

src/node/node_to_node.h
achamayou: As discussed, let's have the seqno and view in the header, and the rest in the payload? That'll faciliate testing.

src/node/progress_tracker.h
achamayou: Yes, let's do that. As discussed, get_view() in the KV is going away, so the opportunity for confusion should be temporary at most.

tests/election.py
achamayou: Ok, we parametrise on f already, so perhaps we just pass that in to run() and use it directly here rather than back it out of the number of nodes.

src/node/node_to_node.h
ashamis: discussed on a call with the group

src/consensus/aft/impl/view_change_tracker.h
ashamis: I ended up going with RequestViewChange but same thing

achamayou (approved):

PR #1840, opened 2020-10-29
Returning the previous signed request when updating the digest is neither necessary nor useful

achamayou:


jumaffre (approved):

src/node/rpc/member_frontend.h
```diff
@@ -1349,7 +1349,7 @@ namespace ccf
         };
```
eddyashton, line 1349: I can't make a suggestion on the previous line, but I believe it should be:
```
return make_success(ma->state_digest);
```

This has changed the reported schema but not the value we actually return, which is still a `MemberAck` rather than a `StateDigest`.

achamayou: Yes, thank you, for some reason I thought the json wrapper would error out, but that doesn't make sense.

PR #1841, opened 2020-10-29
Member recovery shares encryption with RSA-OAEP-256

### LINKED ISSUE #1720: Member recovery share encryption should support RSA encryption
Recovery shares are currently encrypted in the `ccf.shares` table using members' encryption keys. These keys are asymmetric keys generated on the `25519` ECDH curve and the encryption algorithm is https://nacl.cr.yp.to/box.html, provided by Evercrypt.

If member keys (both for signature/identity and encryption) are stored in HSMs (e.g. [Azure Managed HSM](https://docs.microsoft.com/en-gb/azure/key-vault/managed-hsm/)), the keys and algorithms we use should be compatible with the capabilities of HSMs. 

In particular, managed HSMs only seem to support [asymmetric encryption using RSA](https://docs.microsoft.com/en-gb/azure/key-vault/keys/about-keys#wrapkeyunwrapkey-encryptdecrypt) so we should aim to support these too.

jumaffre:
Resolves #1720 

Member recovery shares are now encrypted in the `public:ccf.shares` table with RSA, using  (as opposed to NaCl's crypto box). 

This has the following implications:
- The network's encryption key is no longer required (as we only encrypt with the target's - i.e. member's - public key).
- The `submit_recovery_share.sh` script has been updated and `step-cli` is no longer required (`openssl` does RSA-OAEP-256 key unwrapping fine).
- The 25519 and crypto box specific code has been removed.
- New `RSAKeyPair` class that can be used for unwrapping and corresponding `RSAPublicKey` for unwrapping (see `tls/rsa_key_pair.h`).

Also:
- Renamed `keyshare` in the `public:ccf.members` table to `encryption_pub_key`, as the previous name was confusing. 

Next:
- Update `js_generic.cpp` to use new `tls::RSAPublicKey` to wrap keys.


jumaffre:
This has been tested manually with AKV and the steps to wrap/unwrap a key are available here: https://github.com/microsoft/CCF/issues/1720#issuecomment-718878854. The unwrapping operation is performed by the `submit_recovery_share.sh` script using `openssl`.

tests/start_network.py
```diff
@@ -159,24 +156,14 @@ def add(parser):
...
     args = infra.e2e_args.cli_args(add)
-    if args.recover and (
-        args.ledger_dir is None
-        or args.common_dir is None
-        or args.network_enc_pubk is None
-    ):
-        print(
-            "Error: --recover requires --ledger-dir, --network-enc-pubk and --common-dir arguments."
-        )
+    if args.recover and (args.ledger_dir is None or args.common_dir is None):
```
achamayou, line 165:
```suggestion
    if args.recover and not all(args.ledger_dir, args.common_dir):
```

Not strictly equivalent, but empty paths aren't valid, and arguably more readable.

achamayou (approved):
API schema needs to be updated

PR #1847, opened 2020-11-02
Always create maps dynamically

eddyashton:
To reduce our API surface and complexity, we don't want to support multiple approaches to map creation. This removes the old 'static' map creation, and all maps are created dynamically (ie - implicitly in the first transaction which references them).

This means you don't need to do `kv::Map<K, V>& my_map = kv_store.create<K, V>("my_map_name");`, and don't need to store or pass references or pointers to these maps. Instead you can do:

```
auto view = tx.get_view<K, V>("my _map_name"); // Retrieve purely by name, never create an explicit Map
```

or

```
kv::Map<K, V> my_map("my_map_name"); // Constructed anywhere, without reference to a Store
auto view = tx.get_view(my_map);
```

The latter case lets us keep a pairing of map-type and map-name, so is recommended, but has some semantic differences from the `create` flow, namely:

```
kv::Map<K1, V1> my_map1("my_map_name");
kv::Map<K2, V2> my_map2("my_map_name");
auto view1 = tx.get_view(my_map1);
auto view2 = tx.get_view(my_map2);
// view1 and view2 are over the _same_ contents because they use the same name!
// Even though the types are different (BAD!)

auto tx1 = kv_store1.create_tx();
auto tx2 = kv_store1.create_tx();
auto view1 = tx1.get_view(my_map);
auto view2 = tx2.get_view(my_map);
// Can use the same `kv::Map` over _different stores_, because under the hood they're
// just using the name. Whereas previously we needed to do a `kv::Store::get()` thing
```

There's a few nice knock-ons effects for the implementation:
- We can store `kv::untyped::Map`s rather than `AbstractMap`s in a `Store`, because everything is now a dynamically created `AbstractMap`
- We can store just (untyped) `ChangeSet`s in a `Tx`, and construct whatever kind of `TxView` the caller wants over that `ChangeSet` on-demand
- As a result of the previous point, the `TxView`s that handler code gets from `get_view` contain only what is necessary, and the apps can't call nonsense like `commit` or `prepare` on them

Big PR because it touches a lot of tests.


achamayou:
@eddyashton this change seems to have a material (if moderate) effect on KV serialisation/deserialisation numbers? Is that expected?

doc/developers/kv/kv_how_to.rst
```diff
@@ -10,45 +10,25 @@ The Key-Value :cpp:class:`kv::Store` is a collection of :cpp:class:`kv::Maps` th
 Creating a Map
 --------------
 
-A :cpp:type:`kv::Map` (often referred to as a ``Table``) is created in the constructor of an application. It maps a unique ``key`` to a ``value``.
+A :cpp:type:`kv::Map` (often referred to as a ``Table``) is a collection key-value pairs of a given type. The :cpp:type:`kv::Map` itself is identified by its name, which is used to lookup the map :cpp:type:`kv::Map` in a `cpp:class:`kv::Store` during a transaction.
```
achamayou, line 13:
```suggestion
A :cpp:type:`kv::Map` (often referred to as a ``Table``) is a collection of key-value pairs of a given type. The :cpp:type:`kv::Map` itself is identified by its name, which is used to lookup the map :cpp:type:`kv::Map` in a `cpp:class:`kv::Store` during a transaction.
```

achamayou (approved):

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Added `tools.cmake` to the install , which `ccf_app.cmake` depends on and was missing from the previous release.
 
+### Deprecated
+- `kv::Store::create` is deprecated, and will be removed in a future release. It is no longer necessary to create a `kv::Map` from a `Store`, it can be constructed locally (`kv::Map<K, V> my_map("my_map_name");`) or accessed purely by name (`auto view = tx.get_view<K, V>("my_map_name");`).
```
jumaffre, line 24:
```suggestion
- `kv::Store::create` is deprecated, and will be removed in a future release. It is no longer necessary to create a `kv::Map` from a `Store`, it can be constructed locally (`kv::Map<K, V> my_map("my_map_name");`) or accessed purely by name (`auto view = tx.get_view<K, V>("my_map_name");`) (#1847).
```

doc/developers/kv/kv_how_to.rst
```diff
@@ -10,45 +10,25 @@ The Key-Value :cpp:class:`kv::Store` is a collection of :cpp:class:`kv::Maps` th
...
 
 
 Accessing the ``Transaction``
 -----------------------------
 
 A :cpp:class:`kv::Tx` corresponds to the atomic operations that can be executed on the Key-Value ``Store``. A transaction can affect one or multiple ``Map`` and are automatically committed by CCF once the endpoint's handler returns successfully.
 
-A single ``Transaction`` (``tx``) is passed to all the end-points of an application and should be used to interact with the Key-Value ``Store``.
+A single ``Transaction`` (``tx``) is passed to each end-point of an application and should be used to interact with the Key-Value ``Store``.
```
jumaffre, line 25:
```suggestion
A single ``Transaction`` (``tx``) is passed to each endpoint of an application and should be used to interact with the Key-Value ``Store``.
```

jumaffre (approved):

Issue #683, opened 2020-01-08
Utility to verify node's quote


jumaffre:
It should be possible for members to verify the quote of a given node via a host-only utility (i.e. without requiring to spin up an enclave). This is especially useful before opening the network since the quote of the first node is never verified.

The utility should: 
1. Run `getQuotes` against a given node to retrieve its quote.
2. Verify the collateral associated with the quote (once the call is supported by OE). 
3. Verify the identity of the node quoted (hash the node's cert and verify the quoted data).
4. Verify the quote's `mrenclave` against a trusted enclave measurement.

achamayou:
Open Enclave already has oecertdump: https://github.com/openenclave/openenclave/tree/master/tests/tools/oecertdump

We should:

1. Ask/help with making it part of the release (it's an internal test tool right now)
2. Write a small wrapper script that runs getQuotes, getCode, executes verification (3/4 in @jumaffre's comment) and displays a nice summary

achamayou:
A more minimal and perhaps better starting point: https://github.com/openenclave/openenclave/pull/2865

achamayou:
Link https://github.com/openenclave/openenclave/issues/2724

jumaffre:
Waiting for the `host_verify` utility to be release as part of Open Enclave to be able to test this.

achamayou:
What's still missing is that `host_verify` does not display mrenclave, that needs to be done.

achamayou:
@jumaffre could you open an issue on the openenclave side and link it here?

jumaffre:
https://github.com/openenclave/openenclave/issues/3455

achamayou:
Change complete upstream (thank you @jumaffre!), pending release now.

PR #1850, opened 2020-11-02
Quote verification CLI

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/683

Now that we've got OE 0.12 and #1842, we can wrap the entire quote verification process in one new unique script `verify_quote.sh`. The script takes as argument the node's RPC address, an optional `--mrenclave` and extra curl options. The script then retrieves the quote associated with the node (`/node/quote`), verifies it with `oeverify` (which returns the SGX mrenclave and report data), and verifies the SGX mrenclave against a trusted list of mrenclave (either using the specified `--mrenclave` or calling `/node/code` directly) as well as the report data against the hash of the public key presented by TLS in the certificate of the target node. 

Documentation has been updated accordingly.


livehtml.sh
```diff
@@ -2,4 +2,4 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache 2.0 License.
 
-sphinx-autobuild -b html doc doc/html --host 0.0.0.0
\ No newline at end of file
+sphinx-autobuild -b html doc doc/html --host 0.0.0.0 --port 8080
```
eddyashton, line 5: Does VSCode Remote automatically forward this connection for us? If not, should we change the host or echo it so that it will be forwarded?

jumaffre: It does: 
image

I've updated this port as it clashed with the sandbox.

CMakeLists.txt
```diff
@@ -692,7 +692,7 @@ if(BUILD_TESTS)
       NAME code_update_test
       PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/code_update.py
       CONSENSUS cft
-      ADDITIONAL_ARGS --oe-binary ${OE_BINDIR} --raft-election-timeout 20000
+      ADDITIONAL_ARGS --oe-binary ${OE_BINDIR} --raft-election-timeout 4000
```
jumaffre, line 695: This seemed very high. Let's see if this still works and if not, let's find out why not.

python/utils/verify_quote.sh
```diff
@@ -0,0 +1,112 @@
...
+    echo "Reported quote data does not match node certificate public key:"
+    echo "${extracted_report_data} != ${node_pubk_hash}"
+    exit 1
+elif [ ${is_mrenclave_valid} != true ]; then
+    echo "Error: quote verification failed."
+    echo "Reported mrenclave ${extracted_mrenclave} is not trusted. List of trusted mrenclave:"
+    echo "[${trusted_mrenclaves}]"
+    exit 1
+else
+    echo "Quote verification successful."
```
eddyashton, line 110: I think we should always print the `mrenclave` extracted from the node quote, even when it has successfully matched.

python/requirements.txt
```diff
@@ -1,6 +1,6 @@
 msgpack==1.0.0
 loguru==0.5.3
 requests==2.24.0
-requests-http-signature==0.1.0
+requests-http-signature==0.2.0
```
jumaffre, line 4: A new version of this was released a while back: https://github.com/pyauth/requests-http-signature/releases (at least on PyPi).

python/utils/verify_quote.sh
```diff
@@ -0,0 +1,112 @@
...
+# uses old attestation API to generate quotes
+quote_format="LEGACY_REPORT_REMOTE"
+
+function usage()
+{
+    echo "Usage:"""
+    echo "  $0 https://<node-address> [--mrenclave <mrenclave_hex>] [CURL_OPTIONS]"
+    echo "Verify target node's remote attestation quote."
+    echo "For the verification to be successful, the public key of the node certificate should also match the SGX report data and the corresponding mrenclave should be trusted."
+    echo "If no trusted mrenclave is specified (--mrenclave), the quote measurement should match one of the service's currently accepted code versions."
```
eddyashton, line 20: I think the `should` here is confusing - it sounds like its describing a user error, but actually describing what the script expects _from the service_.

Perhaps something like:

```suggestion
    echo "Verification involves confirming that the node's certificate matches the SGX report data, and that the mrenclave included in the quote is trusted."
    echo "A specific trusted mrenclave may be provided with --mrenclave. If specified, the quoted mrenclave must match this exactly. If unspecified the service's currently accepted code versions will be retrieved from the target node, and verification will succeed only if the quoted mrenclave is present in this list."
```

livehtml.sh
achamayou: If it does, then let’s change it to localhost rather than 0.0.0.0

achamayou (approved):

CMakeLists.txt
achamayou: This is deliberately high because we've historically suffered from unwanted elections caused by very slow quote validation. With recent DCAP cache changes, it's been less of a problem.

doc/users/verify_quote.rst
```diff
@@ -1,33 +1,34 @@
...
+    Quote verification successful.
+
+The script verifies the target CCF node's SGX quote by:
+
+1. Retrieving the node's SGX quote via the ``/node/quote`` endpoint.
+2. Verifying that the cryptographic hash of the node's identity public key (as presented in the node's certificate in the TLS session) matches the quote's report data.
+3. Verifying that the quote's measurement (``MRENCLAVE``) matches one of the trusted enclave measurements.
 
-The raw quote should be decoded and output to file for verification via the Open Enclave ``oeverify`` command-line utility:
+A specific trusted SGX enclave measurement can be specified ``--mrenclave <mrenclave_hex>``.
```
achamayou, line 19: Maybe add a reference to how to run `oesign -dump` to extract mrenclave for an enclave file?

doc/users/verify_quote.rst
```diff
@@ -1,33 +1,34 @@
...
 
-    $ /opt/openenclave/bin/oeverify -r ccf_node_quote.bin -f LEGACY_REPORT_REMOTE
-    Verifying evidence ccf_node_quote.bin...
-    Claims:
-    Enclave unique_id: <ccf_node_mrenclave>
-    Enclave signer_id: <ccf_node_mrsigner>
-    Enclave product_id: <ccf_node_product_id>
-    Enclave sgx_report_data: <ccf_node_report_data>
-    Evidence verification succeeded (0)
+If no enclave measurement is specified, the ``verify_quote.sh`` script automatically verifies it against the code versions currently trusted by the CCF service, as returned by the ``/node/code`` endpoint.
```
achamayou, line 27:
```suggestion
If no code measurement is specified, the ``verify_quote.sh`` script automatically verifies it against the code versions currently trusted by the CCF service, as returned by the ``/node/code`` endpoint.
```

livehtml.sh
achamayou, line 5:
```suggestion
sphinx-autobuild -b html doc doc/html --port 8080
```

achamayou (approved):

doc/users/verify_quote.rst
jumaffre: I believe this `mrenclave` is usually referred to as "enclave measurement" rather than "code measurement" (e.g. [here](https://01.org/sites/default/files/documentation/intel_sgx_developer_guide_pdf.pdf))?

python/utils/verify_quote.sh
jumaffre: Good point. I've re-phrased this accordingly (only slight difference from what you suggested).

doc/users/verify_quote.rst
achamayou: Fair enough, it's just we talk a lot about the code and not much about initial state here, but ok.

Issue #1834, opened 2020-10-28
New JWT authentication proposal types and tables


letmaik:
Step 1 of https://github.com/microsoft/CCF/issues/1833.

The existing `update_ca_cert` proposal type is too flexible (arbitrary key naming scheme, scope too wide) and misnamed (not meant for CA certs but rather token signing keys/certs). This issue proposes a replacement which eventually allows CCF itself to do token signature validation. Token body claims would be passed to the app for further validation/processing.

(Some inspiration on concepts came from https://www.vaultproject.io/docs/auth/jwt)

New tables:
- `ccf.gov.jwt_issuers`
  - key: `<issuer>` (string)  -- URL if OIDC-compliant, otherwise arbitrary name, e.g. `https://login.microsoftonline.com/common/v2.0`
  - value: 
    ```js
    {
      "key_filter": ..., // optional, see below
      "key_policy": ... // optional, see below
    }
    ```
- `ccf.gov.jwt_public_signing_keys`
  - key: `<key_id>` (string) -- `81dsh2B8UUKHZ2JXSOQn-VwfQMU`
  - value: DER-encoded X.509 cert (binary)
- `ccf.gov.jwt_public_signing_key_issuer`
  - key: `<key_id>` (string) -- `81dsh2B8UUKHZ2JXSOQn-VwfQMU`
  - value: `<issuer>` (string)

Having a table keyed by key id (`jwt_public_signing_keys`) allows to efficiently query a key during token signature validation. This then requires another table `jwt_public_signing_key_issuer` to keep the association to issuers, which is needed when updating keys and removing old ones, but also for letting the app know which issuer has signed the token. Note that the table `jwt_issuers`  is only touched during execution of the proposals described below.

New proposal types:
- `set_jwt_issuer` with arguments
  - `issuer` (string) -- URL if OIDC-compliant, otherwise arbitrary name
  - `key_filter` (optional, string, enum), either `all` (default) or `sgx`
    - For `set_jwt_public_signing_keys` proposals, only those keys are stored which satisfy the filter. `all` stores all, `sgx` stores only those that have an embedded SGX/OE quote. A use case is [MAA](https://shareduks.uks.attest.azure.net/certs) which advertises keys for multiple use cases, one of them being SGX, other keys do not have embedded SGX quotes.
  - `key_policy` (optional, object)
    ```js
    {
      "sgx_claims": {
        "signer_id": "12345",
        "attributes": "0300000000000000" 
       }
    }
    ```
[...]

Issue #1031, opened 2020-04-03
Proposal completion is not atomic


eddyashton:
The introduction of a `FAILED` state for member proposals has opened up a possible buggy state, where a proposal has been partially applied. Member proposals produce a list of `ProposedCalls`, which call the hardcoded funcs in `member_frontend`. These functions are called, making changes to the KV, iff the proposal gets enough votes to pass the governance script. If one of these proposed calls _fails_ by returning `false`, then we change the proposal's state to `FAILED`, halt further execution, and send a response to the user (a `HTTP 200 OK` containing the proposal state). Any writes made in this proposed call (or prior proposed calls!) will still be applied to the KV, alongside the recorded vote and state change. This is a bad state - proposals should be applied atomically.

A few early thoughts on solutions:
1) `ProposedCall`s don't get to return false, they can only fail by throwing exceptions. This is close to the old behaviour. The vote returns an error response and makes no writes. Only the 'final' voter gets the exception message, their vote is not recorded.
2) `ProposedCall`s are reduced to a form where they cannot fail. Perhaps they are split into a precondition check and some writes. More simply, I think they all become writes, which will be applied blindly on the state at which their proposal is completed. Any checking of preconditions or error reporting must be done by the proposal script and/or votes.
3) We build a richer KV interface that allows a subset of writes to be dropped, so that we can change the proposal state to `FAILED` while not applying any writes from the `ProposedCall`s. This also fixes some potential issues with our conflict-retry logic, where we currently clear the view but this may lose writes from the framework level (ie - client signatures).

achamayou:
1. sounds attractive because it's consistent with what would happen if the vote evaluation itself did fail: an attempt at voting was made, but could not successfully complete. In some cases that means recording the vote, in others that's recording and proposal execution because the vote would pass, but the behaviour is essential the same: the vote and all its side effects are rolled back.

achamayou:
As discussed, let's go with 1.

achamayou:
Addressing this bug can be done without API impact, so it's moved to post 1.0.

achamayou:
So although this would be easier with sub-transaction support, I think we could implement the following flow specifically for `apply()`:

1. try apply()
2. on catch exception
  a. tx.reset() (yes, we'd have to add it back)
  b. record_voting_history() once more
  c. write the ballot again to proposals_info.ballots
  d. write the votes again to proposals_info.final_votes and vote failures to proposals_info.vote_failures
[...]

PR #1851, opened 2020-11-02
JWT auth proposal types and kv maps

letmaik:
Prototype implementation of https://github.com/microsoft/CCF/issues/1834.


src/node/rpc/member_frontend.h
```diff
@@ -330,6 +382,162 @@ namespace ccf
...
+    bool set_jwt_public_signing_keys(
+      kv::Tx& tx,
+      ObjectId proposal_id,
+      std::string issuer,
+      const JsonWebKeySet& jwks)
+    {
+      auto issuers = tx.get_view(this->network.jwt_issuers);
+      auto keys = tx.get_view(this->network.jwt_public_signing_keys);
+      auto key_issuer =
+        tx.get_view(this->network.jwt_public_signing_key_issuer);
```
jumaffre, line 411: FYI, you can `auto [view1, view2, view3] = tx.get_view(map1, map2, map3);` in one line.

src/node/rpc/member_frontend.h
```diff
@@ -330,6 +382,162 @@ namespace ccf
...
+        bool has_key_policy_sgx_claims =
+          issuer_metadata.key_policy.has_value() &&
+          issuer_metadata.key_policy.value().sgx_claims.has_value() &&
+          !issuer_metadata.key_policy.value().sgx_claims.value().empty();
+        if (
+          issuer_metadata.key_filter == JwtIssuerKeyFilter::SGX ||
+          has_key_policy_sgx_claims)
+        {
+          oe_verifier_initialize();
+          oe_verify_attestation_certificate_with_evidence(
```
jumaffre, line 460: Check return code?

src/node/rpc/member_frontend.h
```diff
@@ -330,6 +382,162 @@ namespace ccf
...
+          {
+            LOG_FAIL_FMT(
+              "Proposal {}: JWKS kid {} has an invalid X.509 certificate: "
+              "{}",
+              proposal_id,
+              jwk.kid,
+              tls::error_string(rc));
+            return false;
+          }
+          mbedtls_x509_crt_free(&cert);
```
jumaffre, line 522: Minor: Could the `mbedtls`-specific code be moved to `tls/key_pair.h`. We also already parse certificates (see `public_key_pem_from_cert()`) so we can probably refactor this to only call into `mbedtls` once.

src/node/rpc/member_frontend.h
```diff
@@ -330,6 +382,162 @@ namespace ccf
...
+          if (v == issuer)
+          {
+            keys->remove(k);
+            key_issuer->remove(k);
+          }
+          return true;
+        });
+    }
+
+    bool set_jwt_public_signing_keys(
```
jumaffre, line 402: Proposals so far were fairly simple and only had a few reasons to fail. However, this one is probably the most complex one we've had so far. Proposal execution currently suffers from https://github.com/microsoft/CCF/issues/1031, which means for example that even if the proposal fails, what has been written to a view will be written to the KV anyway. 

This is bad and should be fixed. I think this PR makes me think we should fix this sooner than later.

tests/npm-app/src/endpoints/jwt.ts
```diff
@@ -0,0 +1,88 @@
...
+        return unauthorized('kid missing in header claims')
+    }
+
+    // Get the stored signing key to validate the token.
+    const keysMap = new ccf.TypedKVMap(ccf.kv['public:ccf.gov.jwt_public_signing_keys'], ccf.string, ccf.typedArray(Uint8Array))
+    const publicKeyDer = keysMap.get(signingKeyId)
+    if (publicKeyDer === undefined) {
+        return unauthorized(`token signing key not found: ${signingKeyId}`)
+    }
+    // jsrsasign can only load X.509 certs from PEM strings
```
jumaffre, line 53: Curious: Why not store the certificate directly in PEM then?

letmaik: The JS code will go away and the C++ code will load DER certs directly which has lower parse overhead.

tests/modules.py
```diff
@@ -314,6 +315,62 @@ def test_npm_app(network, args):
...
+        assert body["msg"] == "jwt validation failed", r.body
+
+        jwt = infra.crypto.create_jwt({}, jwt_key_priv_pem, jwt_kid)
+        r = c.get("/app/jwt", headers={"authorization": "Bearer " + jwt})
+        assert r.status_code == http.HTTPStatus.UNAUTHORIZED, r.status_code
+        body = r.body.json()
+        assert body["msg"] == "jwt invalid, sub claim missing", r.body
+
+        user_id = "user0"
+        jwt = infra.crypto.create_jwt({"sub": user_id}, jwt_key_priv_pem, jwt_kid)
```
jumaffre, line 368: Curious: why are sub-claims necessary?

jumaffre (approved):
LGTM!

tests/modules.py
letmaik: They aren't in general, this is just testing that the app-specific claim validation works as well. This would be the only piece left in JS once the rest moves to C++.

src/node/rpc/member_frontend.h
```diff
@@ -330,6 +382,162 @@ namespace ccf
...
+          int rc = mbedtls_x509_crt_parse(&cert, der.data(), der.size());
+          if (rc != 0)
+          {
+            LOG_FAIL_FMT(
+              "Proposal {}: JWKS kid {} has an invalid X.509 certificate: "
+              "{}",
+              proposal_id,
+              jwk.kid,
+              tls::error_string(rc));
+            return false;
```
achamayou, line 520: This leaks in case of error.

achamayou (changes requested):
Looks good except for the mbedtls leak.

src/node/rpc/member_frontend.h
letmaik: If the cert isn't an OE attested cert or fails to parse for some other reason then the `claims` map will stay empty, which I use for all further checks. Should be enough?

src/node/rpc/member_frontend.h
letmaik: Fixed.

src/node/rpc/member_frontend.h
letmaik: I worked around this for now by doing all kv writes at the end.

achamayou (approved):

src/node/rpc/member_frontend.h
jumaffre: I believe it is safer to exit here/early if the attestation certificate verification fails?

src/node/rpc/member_frontend.h
```diff
@@ -523,6 +735,45 @@ namespace ccf
            users_view->put(parsed.user_id, user_info.value());
            return true;
          }},
+        {"set_jwt_issuer",
+         [this](ObjectId proposal_id, kv::Tx& tx, const nlohmann::json& args) {
+           const auto parsed = args.get<SetJwtIssuer>();
+           auto issuers = tx.get_view(this->network.jwt_issuers);
+
+           issuers->put(parsed.issuer, parsed);
```
jumaffre, line 743: Following the other comment on proposal failure, should this be moved after the call to `set_jwt_public_signing_keys()`?

jumaffre (approved):

src/node/rpc/member_frontend.h
letmaik: No, this works together with the key_filter. For MAA, the JWKS contains a mix of OE and non-OE certs and this code tries to extract the OE claims. If there are none, it's not an OE cert, which is fine.

src/node/rpc/member_frontend.h
letmaik: I missed that... though it's a bit more tricky here to refactor the code. Not impossible though.

letmaik: Done.

src/node/rpc/member_frontend.h
letmaik: To be honest, I think that's quite weird and should be removed :/

achamayou: Is it? Seems like a nice way to declare all your views in one place. Anyway, either is fine.

PR #1853, opened 2020-11-03
Unbundle ballots

achamayou:
Next part of #1742: remove implicit voting as an explicit mechanism (hah). It's up to the constitution to decide whether that ought to be the case for a proposal or not, and some proposals such as operator changes may not require any explicit votes.

### DIFF
--- a/doc/members/proposals.rst
+++ b/doc/members/proposals.rst
@@ -3,7 +3,7 @@ Proposing and Voting for a Proposal
 
 This page explains how members can submit and vote for proposals.
 
-Proposals and vote ballots are submitted as Lua scripts. These scripts are executed transactionally, able to read from the current KV state but not write directly to it. Proposals return a list of proposed actions which can make writes, but are only applied when the proposal is accepted. Each vote script is given this list of proposed actions, and also able to read from the KV, and returns a  indicating whether it supports or rejects the proposed actions.
+Proposals and vote ballots are submitted as Lua scripts. These scripts are executed transactionally, able to read from the current KV state but not write directly to it. Proposals return a list of proposed actions which can make writes, but are only applied when the proposal is accepted. Each vote script is given this list of proposed actions, and also able to read from the KV, and returns a boolean indicating whether it supports or rejects the proposed actions.
 
 Any member can submit a new proposal. All members can then vote on this proposal using its unique proposal id.
 Each member may alter their vote (by submitting a new vote) any number of times while the proposal is open.
--- a/doc/members/proposals.rst
+++ b/doc/members/proposals.rst
@@ -111,7 +111,7 @@ For example, ``member1`` may submit a proposal to add a new member (``member4``)
       }
     }
 
-    $ ./scurl.sh https://<ccf-node-address>/gov/proposals --cacert network_cert --key member1_privk --cert member1_cert --data-binary @add_member.json -H "content-type: application/json"
+    $ scurl.sh https://<ccf-node-address>/gov/proposals --cacert network_cert --key member1_privk --cert member1_cert --data-binary @add_member.json -H "content-type: application/json"
     {
       "proposal_id": 4,
       "proposer_id": 1,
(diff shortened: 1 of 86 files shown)


tests/infra/consortium.py
```diff
@@ -116,15 +116,14 @@ def generate_and_propose_new_member(self, remote_node, curve, member_data=None):
             new_member_id, curve, self.common_dir, self.share_script, self.key_generator
         )
 
-        proposal_body, careful_vote = self.make_proposal(
+        proposal_body, _ = self.make_proposal(
```
jumaffre, line 119: Do we still use the second value returned by `make_proposal()` at all? If no, can we remove it?

jumaffre:
https://microsoft.github.io/CCF/master/members/proposals.html should be updated slightly as well, e.g.:

>  The member who originally submitted the proposal (the proposer) votes for the proposal by default, but has the option to include a negative or conditional vote like any other member.

tests/infra/consortium.py
achamayou: In infra no, but a client may well decide to use it.

doc/members/proposals.rst
```diff
@@ -3,9 +3,11 @@ Proposing and Voting for a Proposal
 
 This page explains how members can submit and vote for proposals.
 
-Proposals and vote ballots are submitted as Lua scripts. These scripts are executed transactionally, able to read from the current KV state but not write directly to it. Proposals return a list of proposed actions which can make writes, but are only applied when the proposal is accepted. Each vote script is given this list of proposed actions, and also able to read from the KV, and returns a boolean indicating whether it supports or rejects the proposed actions.
+Proposals and vote ballots are submitted as Lua scripts. These scripts are executed transactionally, able to read from the current KV state but not write directly to it. Proposals return a list of proposed actions which can make writes, but are only applied when the proposal is accepted. Each vote script is given this list of proposed actions, and also able to read from the KV, and returns a  indicating whether it supports or rejects the proposed actions.
```
ashamis, line 6: sorry don't know how to do the code snip it thing but 

returns a  indicating ==> returns an indication

tests/governance_history.py
```diff
@@ -79,8 +79,8 @@ def run(args):
         proposals_issued += 1
 
         LOG.info("2/3 members accept the proposal")
-        network.consortium.vote_using_majority(primary, new_member_proposal)
-        votes_issued += 1
+        p = network.consortium.vote_using_majority(primary, new_member_proposal)
+        votes_issued += p.votes_for
```
ashamis, line 83: I could be wrong here but should this be:

votes_issued = p.votes_for

ashamis (approved):

doc/members/proposals.rst
achamayou: Thank you!

tests/governance_history.py
achamayou: We are counting the total number of votes we issue, and then what we find in the ledger by comparison, so I believe this is right. In practice we only issue a single set of votes, so it will come out to the same thing, but it's still an accumulation.

doc/members/proposals.rst
```diff
@@ -23,7 +25,6 @@ Assuming the CCF Python package has been installed in the current Python environ
     $ python -m ccf.proposal_generator
     usage: proposal_generator.py [-h] [-po PROPOSAL_OUTPUT_FILE]
                              [-vo VOTE_OUTPUT_FILE] [-pp] [-i]
-                             [--vote-against] [-v]
```
eddyashton, line 26: `[-v]` (verbose) argument hasn't been removed.

eddyashton (commented):
Why are we no longer using the `careful_vote`s? Whether the ballot is submitted with the original proposal or by later members, I think we still want to test the proposal generator's generated votes, and not always vote with `return true`?

achamayou:
@eddyashton as far as I can tell it was never used by other members?

doc/members/proposals.rst
achamayou: Indeed, sorry.

eddyashton:
@achamayou `consortium.py` `vote_using_majority`, is calling `member.py` `vote`, which is submitting `proposal.vote_for`. So we're using it in every infra governance vote.

achamayou:
@eddyashton ah, that's bad, I missed it. I'll make it an explicit input to vote_using_majority instead.

PR #1855, opened 2020-11-04
Use raw serialiser instead of msgpack when accessing tables from js

letmaik:
Currently, in JS, all data is queried/stored as msgpack-encoded `std::vector<uint8_t>`. This is clearly a mistake and should query/store data as-is.


eddyashton (approved):
To test this, we should have a test which parses the keys and values from a public table in the ledger, and confirm that they're raw/JSON, readable from Python.

achamayou (approved):

PR #1856, opened 2020-11-04
Remove unnecessary waits that are not tied to specific transactions

achamayou:


ashamis (approved):

achamayou:
This causes too many races in tests where we shut down nodes, so I will set it aside for now.

PR #1857, opened 2020-11-04
Expose KV foreach to JS

eddyashton:


src/apps/js_generic/js_generic.cpp
```diff
@@ -436,6 +436,54 @@ namespace ccfapp
...
+      JS_FreeValue(ctx, val);
+
+      return true;
+    });
+
+    JS_FreeValue(ctx, func);
+
+    if (failed)
+    {
+      return JS_ThrowInternalError(ctx, "foreach loop failed");
```
letmaik, line 481: I don't think you need to throw another error since the one from the loop is still current. Probably enough to `return JS_Exception` here.

src/apps/js_generic/js_generic.cpp
```diff
@@ -517,6 +565,12 @@ namespace ccfapp
     JS_SetPropertyStr(
       ctx, view_val, "delete", JS_NewCFunction(ctx, deleter, "delete", 1));
 
+    JS_SetPropertyStr(
+      ctx,
+      view_val,
+      "foreach",
```
letmaik, line 571: Should be `forEach`

src/apps/js_generic/js_generic.cpp
```diff
@@ -436,6 +436,54 @@ namespace ccfapp
...
+
+    JSValue func = argv[0];
+
+    if (!JS_IsFunction(ctx, func))
+      return JS_ThrowTypeError(ctx, "Argument must be a function");
+
+    bool failed = false;
+    map_view->foreach([ctx, func, &failed](const auto& k, const auto& v) {
+      JSValue args[2];
+      args[0] = JS_NewArrayBufferCopy(ctx, k.data(), k.size());
```
letmaik, line 457: Let's keep the order as in JS's Map: value, key.

tests/modules.py
```diff
@@ -314,6 +314,19 @@ def test_npm_app(network, args):
         body = r.body.json()
         assert body["msg"] == "Hello!", r.body
 
+        r = c.post("/app/log?id=42", {"msg": "Saluton!"})
```
letmaik, line 317: Probably should change this to `/logs` and `/logs/{id}` now that JS supports templates (maybe in another PR, just mentioning it since you added `/log/all`).

src/apps/js_generic/js_generic.cpp
eddyashton: Sure, I've actually made it `(value, key, map)` to match [`Map.prototype.forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach), though I haven't included `thisArg`.

src/apps/js_generic/js_generic.cpp
eddyashton: So I thought it should be possible to `return JS_GetException(ctx);` here, but it's not. First we're only allowed to call `GetException` once, and we already do so in `dump_error`. Even if we fiddle to call this only once, it looks like returning this `JSValue` from this function doesn't work, causing errors when QuickJS later tries to free things. It's happy with `JS_ThrowXXX` values, or `JS_EXCEPTION`, but not `JS_GetException`, so sticking with the current result.

letmaik: What's the issue with `JS_EXCEPTION`? This is what `JS_ThrowInternalError` (or any other exception result) returns anyway.

eddyashton: You're right, this `JS_EXCEPTION` const works fine.

tests/npm-app/src/types/ccf.ts
```diff
@@ -263,4 +264,14 @@ export class TypedKVMap<K, V> {
     delete(key: K): boolean {
         return this.kv.delete(this.kt.encode(key));
     }
+    forEach(callback: (value: V, key: K, table: TypedKVMap<K, V>) => void) : void {
+        let kt = this.kt;
+        let vt = this.vt;
+        let typed_table = this;
```
letmaik, line 270: Camel case please :)

letmaik, line 270: You re-introduce the term "table", probably not a good idea.

tests/npm-app/src/types/ccf.ts
```diff
@@ -37,6 +37,7 @@ export interface KVMap {
     get: (key: ArrayBuffer) => ArrayBuffer | undefined
     set: (key: ArrayBuffer, value: ArrayBuffer) => KVMap
     delete: (key: ArrayBuffer) => boolean
+    forEach: (callback: (value: ArrayBuffer, key: ArrayBuffer, table: KVMap) => void) => void
```
letmaik, line 40: `ccf.ts` exists in two more places currently. Can you update it there as well? After the forum sample fully replaces the tsoa test app, there will only be two in total.

tests/npm-app/app.json
```diff
@@ -1,203 +1,272 @@
...
+        "require_client_signature": false,
+        "require_client_identity": true,
+        "readonly": false,
+        "openapi": {
+          "responses": {
+            "200": {
+              "description": "Ok",
+              "content": {
+                "application/json": {
+                  "schema": {
```
letmaik, line 267: Does seem to match the code?

src/apps/js_generic/js_generic.cpp
```diff
@@ -436,6 +436,57 @@ namespace ccfapp
...
+    if (!JS_IsFunction(ctx, func))
+      return JS_ThrowTypeError(ctx, "Argument must be a function");
+
+    bool failed = false;
+    map_view->foreach(
+      [ctx, this_val, func, &failed](const auto& k, const auto& v) {
+        JSValue args[3];
+
+        // JS forEach expects (v, k, map) rather than (k, v)
+        args[0] = JS_NewArrayBufferCopy(ctx, v.data(), v.size());
```
letmaik, line 460: I don't think we need copies here, shouldn't `JS_NewArrayBuffer` work as well?

eddyashton: The problem is that the JS code might make copies of the ArrayBuffer (used entirely outside of the `forEach`), so to support that we need to extend the lifetime of the data and provide a function to eventually free it. The data in `k` and `v` is only safe for the duration of this C++ callback.

letmaik: Got it, ok, I don't see a way around a copy then.

tests/npm-app/app.json
eddyashton: Good spot - intended to make this an object but used a tuple in the code. Now an object, so schema and code should match.

tests/npm-app/src/types/ccf.ts
achamayou: @letmaik @eddyashton unless there's a strong objection, I will set up https://prettier.io/, because it covers everything that's currently not (JS/TS, YAML, Markdown), it requires little config, and it will solve this kind of problem forever.

achamayou (approved):

PR #1860, opened 2020-11-05
DRAFT: KV maps are JSON-encoded by default

eddyashton:
For readability of the ledger, and interop with JS, we'd prefer JSON'd tables to MsgPack'd tables. This changes the default (`kv::Map`) to use JSON for both keys and values. A few governance tables (notably `CertDERs`) get custom encoders, blitting the DER certs directly rather than re-encoding them, for faster lookup.

This gives roughly comparable performance, though a bunch of governance operations are now slower since they do slower JSON encodes/decodes. Separately, I've changed `smallbank` to store raw copies values, rather than to re-serialise. They're only accessed by the app, in a performance test, so should be as fast as possible.

A knock-on effect of this is that the Python `Ledger` class doesn't try to decode specific tables any more, it just leaves all decoding to the caller. For the `governance_history` table this is already a nuisance - the information we want is a bunch of byte-vectors, which were Python `bytes` for free with msgpack, but via JSON they become int-lists. It might be worth storing more of these as base64 strings in JSON, both for performance and useability.

Draft for now to see the perf impact.


src/kv/map.h
```diff
@@ -79,9 +79,10 @@ namespace kv
...
 
-  /** Short name for default-serialised maps, using msgpack serialisers. Support
-   * for custom types can be added through the MSGPACK_DEFINE macro
+  /** Short name for default-serialised maps, using JSON serialisers. Support
+   * for custom types can be added through the DEFINE_JSON_TYPE macros, or by
+   * manually implementing nlohmann's to_json and from_json for the custom type.
    */
   template <typename K, typename V>
-  using Map = MsgPackSerialisedMap<K, V>;
+  using Map = JsonSerialisedMap<K, V>;
```
jumaffre, line 87: Maybe it's just me but it feels like this should be defined in a separate file, that is not kv-specific but CCF-wide. Also, do we really want the KV to default to JSON? I imagine that this will affect the performance of our pico-benchmarks?

src/node/certs.h
```diff
@@ -5,6 +5,6 @@
 
 namespace ccf
 {
-  using CertDERs = kv::Map<Cert, ObjectId>;
-  using CACertDERs = kv::Map<std::string, Cert>;
+  using CertDERs = kv::RawCopySerialisedMap<Cert, ObjectId>;
```
jumaffre, line 8: Minor: Perhaps a quick comment to explain why?

src/kv/map.h
achamayou: We do want as much of the governance as possible to be JSON, to facilitate audit. Where performance is a concern, we can make other choices.

eddyashton:
Closing for now. Tracking in issue #1993.

Issue #1837, opened 2020-10-29
AFT: Track data not currently being cleared


ashamis:
While AFT is being built out there are some data structures that are currently only being appended to but nothing is being removed from them. These data structures need to also have data removed from them by the time AFT is complete.

- [ ] ccf::ProgressTracker::certificates (#1981)
- [x] aft::ViewChangeTracker::view_changes (#1861)

PR #1861, opened 2020-11-05
AFT: receive view-change message & send new-view tx

ashamis:
Part of #1709, #1837
This change picks up from #1835. The replicas have received view change messages from one another and the prospective primary makes itself the primary and issues a new-view transaction that contains the required evidence to show that the view-change is needed and has occurred.

In this change we do the following:
- Do the actual view change
- verify this by getting the election test to work correctly
 
Next:
- There is still some work remaining in verifying that the new-view tx is correct (#1709)

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -407,6 +407,7 @@ namespace aft
       {
         for (auto& [index, data, globally_committable] : entries)
         {
+          LOG_INFO_FMT("TTTTTTTT index:{}, globally_committable:{}", index, globally_committable);
           state->last_idx = index;
           ledger->put_entry(*data, globally_committable, false);
         }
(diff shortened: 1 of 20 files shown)


src/consensus/aft/raft.h
```diff
@@ -224,6 +228,13 @@ namespace aft
       return replica_state == Follower;
     }
 
+    NodeId get_leader(kv::Consensus::View view)
```
achamayou, line 231: get_primary?

src/node/node_state.h
```diff
@@ -1631,8 +1632,12 @@ namespace ccf
     {
       setup_n2n_channels();
       setup_cmd_forwarder();
+      try_setup_tracker_store();
```
achamayou, line 1635: Why is this try?

src/node/view_change.h
```diff
@@ -60,5 +60,30 @@ namespace ccf
 
       return v;
     }
+    MSGPACK_DEFINE(signatures, signature);
   };
+  DECLARE_JSON_TYPE(ViewChange);
+  DECLARE_JSON_REQUIRED_FIELDS(ViewChange, signatures, signature);
+
+  struct NewView
```
achamayou, line 68: imo this would be better named ViewChanges or ViewHistory, and ViewChange is really ViewChangeRequest or ViewChangeVote.

I think having new_view later is particularly confusing.

achamayou (approved):

src/node/view_change.h
ashamis: thanks! I made it ViewChangeRequest and ViewChangeConfirmation. I think that does a good job of describing what is going on.

PR #1862, opened 2020-11-06
Make what we can prettier

achamayou:
We have formatters for C/C++, Python, CMake and bash, but we had nothing for (Type|Java)Script, Markdown, YAML and JSON. So here it is.

It also seems to be the case that the copyright notice header enforcement ignores JavaScript and TypeScript files. We probably ought to fix that.


SECURITY.md
```diff
@@ -2,27 +2,27 @@
 
 Microsoft takes the security of our software products and services seriously, which includes all source code repositories managed through our GitHub organizations, which include [Microsoft](https://github.com/Microsoft), [Azure](https://github.com/Azure), [DotNet](https://github.com/dotnet), [AspNet](https://github.com/aspnet), [Xamarin](https://github.com/xamarin), and [our GitHub organizations](https://opensource.microsoft.com/).
 
-If you believe you have found a security vulnerability in any Microsoft-owned repository that meets Microsoft's [Microsoft's definition of a security vulnerability](https://docs.microsoft.com/en-us/previous-versions/tn-archive/cc751383(v=technet.10)), please report it to us as described below.
+If you believe you have found a security vulnerability in any Microsoft-owned repository that meets Microsoft's [Microsoft's definition of a security vulnerability](<https://docs.microsoft.com/en-us/previous-versions/tn-archive/cc751383(v=technet.10)>), please report it to us as described below.
```
letmaik, line 5: Eh, weird? Is there an advantage to add `<>`?

letmaik: Nevermind, got it.

.codecov.yml
```diff
@@ -17,4 +17,3 @@ comment:
   require_changes: false
   require_base: no
   require_head: yes
-
```
eddyashton, line 20: Minor: We often use newlines in `daily-matrix.yml` to trigger Daily CI builds on a PR. An aggressive formatter means these runs are always going to fail from the formatter check, so this makes it slightly harder to get a green Daily run on a PR. I guess we just need to get in the habit of triggering these builds manually?

.prettierrc.json
```diff
@@ -0,0 +1 @@
+{}
```
eddyashton, line 1: Do we need this empty config file?

SECURITY.md
```diff
@@ -2,27 +2,27 @@
...
-  * Step-by-step instructions to reproduce the issue
-  * Proof-of-concept or exploit code (if possible)
-  * Impact of the issue, including how an attacker might exploit the issue
+- Type of issue (e.g. buffer overflow, SQL injection, cross-site scripting, etc.)
+- Full paths of source file(s) related to the manifestation of the issue
+- The location of the affected source code (tag/branch/commit or direct URL)
+- Any special configuration required to reproduce the issue
+- Step-by-step instructions to reproduce the issue
+- Proof-of-concept or exploit code (if possible)
+- Impact of the issue, including how an attacker might exploit the issue
```
eddyashton, line 25: Minor: Since this is a standard file copied from elsewhere, it could perhaps be ignored and pasted verbatim? When we next take an updated version, we'll have to reformat it again. I have no strong opinions either way.

doc/schemas/app_openapi.json
```diff
@@ -7,16 +7,11 @@
             "$ref": "#/components/schemas/uint64"
           }
         },
-        "required": [
-          "caller_id"
-        ],
+        "required": ["caller_id"],
```
eddyashton, line 10: These openapi files are generated so I think they should be ignored.

.codecov.yml
achamayou: Or we add an empty daily_canary file which we tell the daily to trigger on but prettier to ignore?

.prettierrc.json
achamayou: If the documentation is to be believed, yes.

SECURITY.md
achamayou: Sure, it’s easy to exclude it.

doc/schemas/app_openapi.json
achamayou: Yes, the reformatting actually breaks the test.

scripts/ci-checks.sh
```diff
@@ -25,6 +25,14 @@ else
   "$SCRIPT_DIR"/check-format.sh src samples
 fi
 
+npm --loglevel=error install --save-dev --save-exact prettier 1>/dev/null
```
letmaik, line 28: Why `--save-exact`?

letmaik: In fact, why save at all? Will this create a package.json and/or a node_modules in the repo root?

SECURITY.md
achamayou: Not sure there is, perhaps some parsers are unhappy because of the () in the URL path?

.prettierrc.json
achamayou: Nope, removing.

scripts/ci-checks.sh
achamayou: Just following doc recommendations, and it does create those. I've removed those options.

letmaik (approved):

achamayou:
Daily failure unrelated, filed as: https://github.com/microsoft/CCF/issues/1865

Will merge after the release.

jumaffre (approved):
Related VSCode plugin: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

PR #1864, opened 2020-11-09
AFT: Verify append entry validity

ashamis:
Part of #1709

This is still a work in progress and there will be follow-up PRs. In this PR we do the following:
- A backup checks the validity of the NewViewConfirmation append entry before it applies it
- Every time an append entry arrives a backup will check that it comes from a "valid" primary
- If the validity of a primary cannot be established from the information that the backup already has it will look at the append entry to see if it is a NewViewConfirmation entry that contains said proof.

### DIFF
--- a/src/consensus/aft/impl/state.h
+++ b/src/consensus/aft/impl/state.h
@@ -98,7 +98,8 @@ namespace aft
       my_node_id(my_node_id_),
       current_view(0),
       last_idx(0),
-      commit_idx(0)
+      commit_idx(0),
+      new_view_idx(0)
     {}
 
     SpinLock lock;
@@ -113,5 +114,6 @@ namespace aft
     kv::Version bft_watermark_idx;
 
     ViewHistory view_history;
+    kv::Version new_view_idx;
   };
 }
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -701,7 +701,6 @@ namespace aft
       entries_batch_size = std::max((batch_window_sum / batch_window_size), 1);
     }
 
-    // TODO: we need to find if we committed a new view here
     void append_new_view(kv::Consensus::View view)
     {
       state->current_view = view;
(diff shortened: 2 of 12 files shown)


src/consensus/aft/impl/state.h
```diff
@@ -98,7 +98,8 @@ namespace aft
       my_node_id(my_node_id_),
       current_view(0),
       last_idx(0),
-      commit_idx(0)
+      commit_idx(0),
+      new_view_idx(0)
```
achamayou, line 102: What's this value? Is it what we think the next view should be?

src/consensus/aft/raft.h
```diff
@@ -856,6 +858,53 @@ namespace aft
       // passes the integrity check. This way, entries containing dynamic
       // topology changes that include adding this new leader can be accepted.
 
+      // When we are running with in a Byzantine model we cannot trust that the
+      // replica is sending up this data is correct so we need to validate
+      // additional properties that go above an beyond the non-byzantine
```
achamayou, line 863:
```suggestion
      // additional properties that go above and beyond the non-byzantine
```

src/consensus/aft/raft.h
```diff
@@ -856,6 +858,53 @@ namespace aft
...
+      // scenario.
+      bool confirm_evidence = false;
+      if (consensus_type == ConsensusType::BFT)
+      {
+        if (active_nodes().size() == 0)
+        {
+          // The replica is just starting up, we want to check that this replica
+          // is part of the network we joined but that is dependant on Byzantine
```
achamayou, line 871:
```suggestion
          // is part of the network we joined but that is dependent on Byzantine
```

achamayou (approved):

src/consensus/aft/impl/state.h
ashamis: This is used by the primary to track at which index the last NewViewConfirmation is stored so that it can send a hint when replicating append entries

PR #1866, opened 2020-11-09
Support for shareless members

### LINKED ISSUE #1021: Support for share-less members

jumaffre:
Resolves #1021 

~(Opening as draft are there are still a couple of TODOs)~

This PR adds support for share-less members, i.e. members without a recovery share. Other members (i.e. those with recovery shares) are named "recovery members" in the documentation.

If a member is registered in CCF without a public encryption key, such member is not given a recovery share and thus cannot request/decrypt a recovery share. 

Changes:
- `cchost`'s `--member-info` now doesn't require the path of the member's public encryption key. In other words, `--member-info member_cert.pem`, `--member-info member_cert.pem,,member_data.json` (not the double `,,`) are valid.
- Recovery threshold, adding/removing members logic updated accordingly, so that the CCF service never ends up in a non-recoverable state (i.e. number of active recovery members < recovery threshold).
- Python infra and documentation

I have moved the recovery share end-to-end test code from `memberclient.py` to a new `membership.py` file.


doc/quickstart/test_network.rst
```diff
@@ -1,21 +1,20 @@
...
-The script creates a new one node test CCF network running locally. All the governance requests required to open the network to users are automatically issued.
+The script creates a new one node CCF test network running locally. All the governance requests required to open the network to users are automatically issued.
 
 For example, deploying the ``liblogging`` example application:
 
 .. code-block:: bash
 
-    $ cd CCF/build
-    $ ../sandbox.sh --package ./liblogging.virtual.so
+    $ /opt/ccf/bin/sandbox.sh --package ./liblogging.virtual.so
```
jumaffre, line 17: The paths seemed wrong when using the sandbox from an install of CCF.

python/ccf/proposal_generator.py
```diff
@@ -231,11 +237,6 @@ def new_member(
     return false
     end
 
-    expected_enc_pub_key = [====[{encryption_pub_key}]====]
```
jumaffre, line 234: This is now optional so the sample vote should only check for the member's identity public key.

src/host/main.cpp
```diff
@@ -453,20 +453,36 @@ int main(int argc, char** argv)
 
     if (*start)
     {
+      // Count members with public encryption key as only these members will be
+      // handed a recovery share.
+      // Note that it is acceptable to start a network without any member having
+      // a recovery share. The service will check that at least one recovery
+      // member is added before the service can be opened.
```
jumaffre, line 460: Tested in `membership.py`, starting a CCF service with different combinations of members.

doc/operators/start_network.rst
```diff
@@ -35,7 +35,7 @@ When starting up, the node generates its own key pair and outputs the certificat
...
 
-The certificates and recovery public keys of initial members of the consortium are specified via ``--member-info``. For example, if 3 members should be added to CCF, operators should specify ``--member-info member1_cert.pem,member1_enc_pubk.pem``, ``--member-info member2_cert.pem,member2_enc_pubk.pem``, ``--member-info member3_cert.pem,member3_enc_pubk.pem``.
+The certificates, encryption public keys and member data of initial members of the consortium are specified via ``--member-info``. For example, if 3 members (one recovery member with member data, one recovery member without member data and one non-recovery member with member data) should initially be added to CCF, operators should specify ``--member-info member1_cert.pem,member1_enc_pubk.pem,member1_data.json``, ``--member-info member2_cert.pem,member2_enc_pubk.pem``, ``--member-info member3_cert.pem,,member2_data.json`` (note the empty public encryption key path "``,,``" for the third member).
```
eddyashton, line 38: I think this example is now covering too many things, and they should be separately listed. Something like:

> For example:
> - A recovery member with member data
> `--member-info member1_cert.pem,member1_enc_pubk.pem,member1_data.json`
> - A recovery member without member data
> `--member-info member2_cert.pem,member2_enc_pubk.pem`
> - A non-recovery member with member data (note the empty public encryption key path, indicating a non-recovery member)
> `--member-info member3_cert.pem,,member2_data.json`

python/ccf/proposal_generator.py
eddyashton: I think this should remain, wrapped in `if encryption_pub_key is not None:`. If the proposal _does_ specify an encryption public key, then the careful vote for it should check that encryption public key is present.

eddyashton: Sorry this is inline Lua. So either formatted or hand-appended or something, if we have an encryption key in the Python we should check it in the Lua.

src/node/genesis_gen.h
```diff
@@ -297,12 +315,12 @@ namespace ccf
     {
       auto service_view = tx.get_view(tables.service);
 
-      if (get_active_members_count() < get_recovery_threshold())
+      if (get_active_recovery_members().size() < get_recovery_threshold())
```
eddyashton, line 318: Nit: `get_active_recovery_members()` is expensive, so would prefer:
```
const auto recovery_members_count = get_active_recovery_members().size();
if (recovery_members_count...)
{
  ..., recovery_members_count, ...
}
```

python/ccf/proposal_generator.py
jumaffre: You're absolutely right. I'd assumed that checking for the member's identity certificate was sufficient for some reason anyway. Fixed now.

doc/operators/start_network.rst
jumaffre: Good point. I've also added a bullet point for non-recovery, no member data member, i.e. `--member-info member_cert.pem`.

src/node/genesis_gen.h
jumaffre: Good spot.

src/node/rpc/member_frontend.h
```diff
@@ -1733,7 +1747,16 @@ namespace ccf
           g.add_member(info);
         }
 
-        g.set_recovery_threshold(in.recovery_threshold);
+        // Note that it is acceptable to start a network without any member
+        // having a recovery share. The service will check that at least one
+        // member with a recovery is added before the service is opened.
```
achamayou, line 1752:
```suggestion
        // member with a recovery share is added before the service is opened.
```

achamayou:
I think it would be good to add a check in the operator governance, validating that operators (ie. members with member_info.is_operator) do no have a recovery share, since it is the primary use case for that feature.

achamayou (approved):

jumaffre:
> I think it would be good to add a check in the operator governance, validating that operators (ie. members with member_info.is_operator) do no have a recovery share, since it is the primary use case for that feature.

@achamayou: Done now 👍

Issue #1844, opened 2020-10-30
Proposal signature with key held in HSM


jumaffre:
As members' identity keys may be held in a HSM, proposals should be able to be signed with such keys. CCF currently only supports `ecdsa-sha256` for the HTTP signature scheme. However, Azure HSMs don't seem to support this signature algorithm (see https://docs.microsoft.com/en-us/rest/api/keyvault/sign/sign#jsonwebkeysignaturealgorithm) but only SHA-384 on the 384r1 EC curve.

With regards to the type of signature algorithm we should support, https://tools.ietf.org/html/draft-cavage-http-signatures-12#section-2.1.3 states:

> Implementers SHOULD derive the digital signature algorithm used by an implementation from the key metadata identified by the `keyId` rather than from this field.  If `algorithm` is provided and differs from the key metadata identified by the `keyId`, for example `rsa-sha256` but an EdDSA key is identified via `keyId`, then an implementation MUST produce an error.

> Algorithm Name: ecdsa-sha256
   Status: deprecated, specifying signature algorithm enables attack vector.

> Algorithm Name: hs2019
   Status: active

A solution would then be to sign the http request using P-384 and SHA-384, use the `hs2019` algorithm in the HTTP header and derive the hashing algorithm from the member's identity key retrieved from `keyId`.

Note that the Python package used to sign HTTP requests ([`requests-http-signature`](https://github.com/pyauth/requests-http-signature)) doesn't support `hs2019` and only supports `sha256` as the signing algorithm so we would have to contribute these changes upstream or remove the dependency on this package.

jumaffre:
Steps to generate a key and sign from the command line (to be moved to documentation once this work is finished). Note that this is a little trickier as the `az` CLI doesn't support signing, so it's done with `curl` here:

```bash
# Assumes VAULT_NAME and APP_NAME are set
$ az keyvault key create --vault-name $VAULT_NAME --name "sign-key-1" --curve P-384 --kty EC --ops sign

# Create App
$ az ad sp create-for-rbac --name $APP_NAME
# Returns clientid (appId), client secret (password)

# Get Access Token
$ export AZ_TOKEN=$(curl -X POST -d "grant_type=client_credentials&client_id=$CLIENT_ID&client_secret=$PASSWORD$CLIENT_SECRET&resource=https://vault.azure.net" https://login.microsoftonline.com/<replace_with_tenant_id>/oauth2/token | jq -r .access_token)

[...]

PR #1872, opened 2020-11-10
Support for hs2019 in HTTP signature scheme

jumaffre:
Related to #1844 

If members' keys are held in a HSM, the signing algorithm used to sign HTTP requests should be compatible with the signing algorithm supported by CCF. This PR adds support for the [`hs2019` signature scheme 
](https://tools.ietf.org/html/draft-cavage-http-signatures-12#appendix-E.2), which gives us compatibility with HSM signatures. 

Edit: As pointed by @eddyashton, the signature hashing algorithm to use is now derived from the member's public key used for verification (it used to always be sha-256), using: https://github.com/microsoft/CCF/blob/82e4d007fd54080d7fed9429c2bd48fbca10d7c0/src/tls/curve.h#L62

For now, only `scurl.sh` supports this new scheme. The old `ecdsa-sha256` ([which is actually deprecated](https://tools.ietf.org/html/draft-cavage-http-signatures-12#appendix-E.2)) is still supported as the Python client hasn't yet been updated.

Next: Upgrade Python client and drop `ecdsa-sha256` scheme entirely.


src/http/http_sig.h
```diff
@@ -110,13 +110,13 @@ namespace http
...
 
-    const auto signature = kp->sign(to_sign.value(), MBEDTLS_MD_SHA256);
+    const auto signature = kp->sign(to_sign.value());
 
     auto auth_value = fmt::format(
       "Signature "
       "keyId=\"ignored\",algorithm=\"{}\",headers=\"{}\",signature="
       "\"{}\"",
-      auth::SIGN_ALGORITHM_SHA256,
+      auth::SIGN_ALGORITHM_HS_2019,
```
eddyashton, line 119: I'm surprised that this is the only change around here - where has the actual implementation of the signing changed? I guess this relies on the passed `kp` using HS2019 rather than SHA256 when we call `kp->sign`? Is there a way to request the signing algorithm from the `kp`, to confirm its one of the algorithms we expect?

eddyashton: After discussing, the actual change I was looking for here is the removal of `, MBEDTLS_MD_SHA256` from `kp->sign`. This doesn't mean _no_ hashing, it means the hashing algorithm is _determined from the key_, which is the HS2019 behaviour we're reporting here. I'm happy!

achamayou (approved):

eddyashton (approved):

PR #1873, opened 2020-11-10
Fix #1871

### LINKED ISSUE #1871: Update state digest is incorrectly documented
**Describe the bug**
[`/ack/update_state_digest`](https://github.com/microsoft/CCF/blob/master/doc/schemas/gov_openapi.json#L509) is documented to return a JSON response with a `state_digest` field; however, having called it the response seems to only contain a list that would be the contents of `state_digest`.

**Environment information**
CCF 0.14.3

achamayou:
Fix #1871 

```
41: 16:54:14.612 | INFO     | infra.member:update_ack_state_digest:142 - [0|member0] POST /gov/ack/update_state_digest
41: 16:54:14.612 | INFO     | infra.member:update_ack_state_digest:142 - 200 @2.4 {"state_digest":"2fb5520d4d50d599f056378925de1208e2a88b81666a5cd75e1ca1ff978b8ed6"}
```


src/node/rpc/member_frontend.h
```diff
@@ -1589,7 +1589,7 @@ namespace ccf
         };
       make_endpoint(
         "ack/update_state_digest", HTTP_POST, json_adapter(update_state_digest))
-        .set_auto_schema<void, StateDigest>()
+        .set_auto_schema<void, std::vector<uint8_t>>()
```
jumaffre, line 1592: I think we should make this a base64 string while we're updating it.

achamayou: Should we even make it a string all the way down, and have that _be_ the digest?

jumaffre: That will certainly help for auditing too so I would say yes.

achamayou: We seem to use hex for digests pretty much everywhere else, so I'm going with that for now.

achamayou:
After discussion with @jumaffre, I'm wrapping this again in a top-level JSON object, under a single key. I think we want to apply the following principles across all node and governance RPCs:

- JSON in/out, by default. We may accept, and return (given the right `Accept:`) other content types for convenience, but JSON should always be available, for:
  - Consistency when scripting/interacting from code
  - Extensibility, adding new fields to an object is straightforward
- Digests are always encoded as hex-strings in JSON
- Other contents such as embedded requests are represented as base64

src/node/rpc/member_frontend.h
```diff
@@ -1594,13 +1594,13 @@ namespace ccf
...
-              std::vector<uint8_t>(s->root.h.begin(), s->root.h.end());
-
+            ma->state_digest = s->root.hex_str();
             ma_view->put(caller_id, ma.value());
           }
+          nlohmann::json j;
+          j["state_digest"] = ma->state_digest;
 
-          return make_success(ma->state_digest);
+          return make_success(j);
```
jumaffre, line 1603: Is it possible to do `return make_success(ma->state_digest);` instead of constructing a JSON object?

jumaffre (approved):

src/node/rpc/member_frontend.h
achamayou: No, because the serialisation auto-unwraps state_digest, so that ends up being just a string.

PR #1874, opened 2020-11-11
Include JS OpenAPI from app.json in /api responses

eddyashton:
The endpoints table now stores the `openapi` elements defined for each endpoint in `app.json`, and merges these into the resulting document during `build_api`.

The `modules.py` test retrieves the OpenAPI docs for the tested JS apps, and checks that they're valid OpenAPI. This required a few minor changes to the existing `app.json` files.

To support this I've updated the Lua JSON converters, so they can correctly roundtrip an empty object _or_ empty array. This is done by adding a magic field to the metatable on the pushed table. Its an ugly solution, but does what we need for now, while we're still reliant on Lua.


achamayou (approved):

tests/npm-app/app.json
```diff
@@ -178,7 +178,6 @@
                 "application/json": {
                   "schema": {
                     "items": {
-                      "items": {},
```
letmaik, line 181: Why is this necessary?

eddyashton: The validator we're using doesn't like this empty object - it doesn't think this matches the schema for a JSON schema object. So we say the same thing ("the items have no further type restriction") by omitting the field entirely.

PR #1881, opened 2020-11-13
AFT: New primary send evidence extremely delayed backup

### LINKED ISSUE #1709: AFT: View changes

ashamis:
resolves #1709

This is closing a hole that exists in view-changes where: if there is a network partition a replica that is considerably behind the primary and did receive any of the view-change messages could receive append entries from the new primary and no evidence that supports this new primary is the actual primary.

### DIFF
--- a/src/consensus/aft/impl/view_change_tracker.h
+++ b/src/consensus/aft/impl/view_change_tracker.h
@@ -100,25 +100,59 @@ namespace aft
       return ResultAddView::OK;
     }
 
+    // TODO we generate the view here
     kv::Consensus::SeqNo write_view_change_confirmation_append_entry(
       kv::Consensus::View view)
     {
-      auto it = view_changes.find(view);
-      if (it == view_changes.end())
-      {
-        throw std::logic_error(fmt::format(
-          "Cannot write unknown view-change to ledger, view:{}", view));
-      }
-
-      auto& vc = it->second;
-      ccf::ViewChangeConfirmation nv(vc.view, vc.seqno);
-
-      for (auto it : vc.received_view_changes)
-      {
-        nv.view_change_messages.emplace(it.first, it.second);
-      }
-
+      ccf::ViewChangeConfirmation nv =
+        create_view_change_confirmation_msg(view);
       return store->write_view_change_confirmation(nv);
+    }
+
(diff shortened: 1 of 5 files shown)


src/consensus/aft/impl/view_change_tracker.h
```diff
@@ -103,22 +103,55 @@ namespace aft
...
+      kv::Consensus::View view)
+    {
+      ccf::ViewChangeConfirmation nv =
+        create_view_change_confirmation_msg(view);
+      nlohmann::json j;
+      to_json(j, nv);
+      std::string s = j.dump();
+
+      std::vector<uint8_t> ret;
+      ret.resize(s.size() + 1);
```
achamayou, line 121: Why +1?

src/consensus/aft/raft.h
```diff
@@ -673,6 +677,48 @@ namespace aft
...
+          state->my_node_id,
+          r.from_node);
+        return;
+      }
+
+      if (r.from_node != state->requested_evidence_from)
+      {
+        // Ignore if we didn't request this evidence.
+        LOG_FAIL_FMT(
+          "we received evidence which we did not request, ignoring it {}",
```
achamayou, line 712: There are no formal guidelines (yet?) but we generally capitalize error messages, and tend to omit the pronoun.

achamayou (approved):

src/consensus/aft/impl/view_change_tracker.h
ashamis: I cleaned this up to use a vector ctor but the + 1 is needed so that we copy the string terminator.

PR #1884, opened 2020-11-13
Document HSM usage for members identity and encryption keys

### LINKED ISSUE #1844: Proposal signature with key held in HSM

### LINKED ISSUE #1845: Document use of Azure HSM to store member identity and encryption keys
Follow-up from https://github.com/microsoft/CCF/issues/1720#issuecomment-718878854, we should test and document how members keys can be held in AKV, registered in CCF and used for signing proposals and decrypting recovery shares.

jumaffre:
Resolves #1844
Resolves #1845 

This PR adds documentation for how members can store their identity keys/certificates and encryption keys inside Azure Key Vault and how to use them with CCF. 

Most changes are documentation, except a minor change to `scurl.sh` which now prints the digest to be signed as well as the request headers if `--print-digest-to-sign. This is useful when constructing signed requests with a HSM (see new documentation page for an example)..


achamayou (approved):
Looks good!

achamayou:
One of the json files needs reformatting.

doc/members/adding_member.rst
```diff
@@ -3,11 +3,13 @@ Adding New Members
 
 It is possible for existing members to add new members to the consortium after a CCF network has been started.
 
-.. note:: The maximum number of active consortium members at any given time is 255.
+.. note:: The maximum number of active recovery members (i.e. those with a recovery share) at any given time is 255.
```
ashamis, line 6: you probably want to say something like "The maximum number of allowed active ...."

doc/members/hsm_keys.rst
```diff
@@ -0,0 +1,126 @@
+Using Member Keys Stored in HSM
+===============================
+
+This page explains how members' identity certificates and encryption keys stored in an `HSM <https://en.wikipedia.org/wiki/Hardware_security_module>`_ can be used with CCF. The following guide describes the usage of `Azure Key Vault <https://azure.microsoft.com/en-gb/services/key-vault>`_
+
+.. note::
+
+    It is assumed that the CCF member has already access to an existing Azure Key Vault. See `here <https://docs.microsoft.com/en-us/azure/key-vault/secrets/quick-create-portal#create-a-vault>`_ for more details on how to create one. Using the `Azure CLI <https://docs.microsoft.com/en-us/cli/azure/install-azure-cli>`_, it is possible to check the list of available Key Vault instances:
```
ashamis, line 8: "that CCF members already have"

ashamis (approved):

PR #1885, opened 2020-11-13
Correctly report GetQuote schema

### LINKED ISSUE #1876: /node/quote no longer returns a list but is still documented as such
**Describe the bug**
#1761 changed `/node/quote` to return the quote info directly instead of in an array, but the API schema still lists the response as [`GetQuotes__Out`](https://github.com/microsoft/CCF/blob/master/doc/schemas/node_openapi.json#L673) which contains an array of quotes.

achamayou:
Fix #1876

### DIFF
--- a/doc/schemas/node_openapi.json
+++ b/doc/schemas/node_openapi.json
@@ -670,7 +670,7 @@
             "content": {
               "application/json": {
                 "schema": {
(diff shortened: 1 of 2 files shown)


jumaffre (approved):
My bad! Thanks for fixing this.

Issue #1818, opened 2020-10-23
"Starting a Test Network" doc page without parent


letmaik:
https://microsoft.github.io/CCF/master/quickstart/test_network.html

There is no navigation on the left.

There are links to it from https://microsoft.github.io/CCF/master/developers/run_app.html and https://microsoft.github.io/CCF/master/members/accept_recovery.html.

I think the instructions should also be changed to not refer to a build tree but rather an installation tree.

achamayou:
Agreed, I intend to replace this page with one called "Sandbox" which will go under "Build Apps".

achamayou:
So that's now under Build Apps, as https://microsoft.github.io/CCF/master/developers/run_app.html, and refers to an installation tree. There's more to be done, as tracked under #1700, but this particular issue is complete as far as I can tell.

PR #1892, opened 2020-11-17
Remove references to test network script in documentation

achamayou:
Addresses #1818 

I want to make further changes, in particular the fact that we have two menu structures, one in the toctree/top bar and one on the home page is very confusing to me. I would like to establish consensus on those though, whereas this feels like an obvious good change.


doc/developers/run_app.rst
```diff
@@ -2,7 +2,73 @@
 Running CCF Applications
 ========================
```
achamayou, line 3: That should be renamed to Sandbox in my opinion, and expanded to cover the various usage scenarios it addresses.

eddyashton (approved):

PR #1895, opened 2020-11-18
Replace logging_js.lua with app.json

eddyashton:
Historically we had to wrap JS apps in Lua, and install them with a `set_js_app` proposal. Now we have a `deploy_js_app`, which is preferable as it lets developers work directly with `.js` sources.

This removes `logging_js.lua` with the modern approach, as a standalone example of how this transition should be done.

In a separate PR I'll remove the remaining uses of `set_js_app` in the infra and see if we can mark the old approach as deprecated.


src/apps/logging/js/src/logging.js
```diff
@@ -0,0 +1,58 @@
+function get_id_from_request_query(request) {
+  const elements = request.query.split("&");
+  for (const kv of elements) {
+    const [k, v] = kv.split("=");
+    if (k == "id") {
+      return ccf.strToBuf(v);
+    }
+  }
+  throw new Error("Could not find 'id' in query");
```
jumaffre, line 9: Could we also add an example of logging to the node's log, just to showcase that feature?

CMakeLists.txt
```diff
@@ -741,8 +741,8 @@ if(BUILD_TESTS)
     NAME js_e2e_logging_raft
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/e2e_logging.py
     CONSENSUS cft
-    ADDITIONAL_ARGS --js-app-script
```
jumaffre, line 744: It seems that we're not using `--js-app-script` for anything anymore and that the app bundle should be used instead as it exports the ccf module for access to KV and typing?

achamayou (approved):

src/apps/logging/js/src/logging.js
eddyashton: I don't think this JS app is (yet) a showcase for the full CCF-JS API. I think we could make it that (approach full parity with the C++ implementation, templated endpoints, different endpoint properties, etc), but for now I'd prefer to keep it as a very simple demo. The sample JS/TS apps do include `console.log` logging, so we show its possible and test it runs elsewhere.

CMakeLists.txt
eddyashton: Agreed - planning to do this in a future PR (deprecate rather than remove if possible, for `--js-app-script`, `set_js_app` proposals, &c).

achamayou:
@eddyashton the formatter needs to be run.

jumaffre (approved):

PR #1897, opened 2020-11-18
Wait for app frontend to be opened after recovery

jumaffre:
On recovery, operators should wait for the `/app` frontend to be open. This is different from the classic "open network" scenario where members can guarantee that the `/app` frontend is open by checking that the `open_network` proposal execution is globally committed. On recovery, the `/app` frontend is opened much later that when the last recovery share is submitted as the private ledger needs to be recovered first. 

This PR adds a quick check in the Python infra to make sure that the `/app` frontend is open to users before letting the service be used by users. Note that there's no easy way for operators to check that `/app` frontend is open apart from issuing a request on that frontend themselves.

This edge-case could occur when recovering large ledgers.


src/consensus/aft/raft_consensus.h
```diff
@@ -20,14 +20,12 @@ namespace aft
   private:
     std::unique_ptr<Aft<T...>> aft;
     ConsensusType consensus_type;
-    bool is_open;
```
jumaffre, line 23: This was unused.

achamayou (approved):

ashamis (approved):

PR #1875, opened 2020-11-11
JWT docs

letmaik:
@achamayou Feel free to fill in the "Certificate Authentication" page, either here or in a subsequent PR.

Note: Some URLs depend on https://github.com/microsoft/CCF/pull/1867 being merged.

PR #1898, opened 2020-11-18
Update certificate auth page

achamayou:
Follow up to #1875


doc/developers/auth/cert.rst
```diff
@@ -1,8 +1,12 @@
...
+User identities in CCF are X.509 certificates. They can be added or removed via governance proposals, which are subject to the consortium constitution rules (see :ref:`members/open_network:Adding Users`).
 
-Users should have previously been trusted by the consortium of members for them to use this authentication mechanism (see :ref:`members/open_network.html:Adding Users`). 
+Requests emitted by users can be authenticated one of two ways:
 
-This can be disabled for a specific application endpoint, via the ``set_require_client_identity`` property. If certificate authentication is disabled, the application should authenticate and authorise users based on other mechanisms (e.g. :doc:`/developers/auth/jwt`).
+- via the TLS handshake, in which a client uses the user private key to prove its identity (e.g. using the ``--key`` and ``--cert`` argument to ``curl``)
+- by signing the request contents with the user private key, following the scheme described `here <https://tools.ietf.org/html/draft-cavage-http-signatures-12>`_
```
jumaffre, line 9: Add reference to https://microsoft.github.io/CCF/master/users/issue_commands.html#signing perhaps?

jumaffre (approved):

jumaffre (approved):

doc/developers/auth/cert.rst
```diff
@@ -1,8 +1,12 @@
...
 
-Users should have previously been trusted by the consortium of members for them to use this authentication mechanism (see :ref:`members/open_network.html:Adding Users`). 
+Requests emitted by users can be authenticated one of two ways:
 
-This can be disabled for a specific application endpoint, via the ``set_require_client_identity`` property. If certificate authentication is disabled, the application should authenticate and authorise users based on other mechanisms (e.g. :doc:`/developers/auth/jwt`).
+- via the TLS handshake, in which a client uses the user private key to prove its identity (e.g. using the ``--key`` and ``--cert`` argument to ``curl``)
+- by :ref:`users/issue_commands:Signing` the request contents with the user private key, following the scheme described `here <https://tools.ietf.org/html/draft-cavage-http-signatures-12>`_.
+
+This certificate-based authentication can be disabled for a specific application endpoint, via the :cpp:func:`ccf::EndpointRegistry::Endpoint::set_require_client_identity()`.
```
eddyashton, line 11: This comment is misleading. `require_client_identity` only interacts with the first point above (while `require_client_signature` affects the latter). If `require_client_identity = true`, an unknown TLS handshake will not currently be acceptable auth (even if this key is known).

doc/developers/auth/cert.rst
```diff
@@ -1,8 +1,12 @@
 Certificate Authentication
 ==========================
 
-CCF's primary authentication mechanism relies on users specifying their certificate and private key in the TLS handshake (e.g. using the ``--key`` and ``--cert`` argument to ``curl``).
+User identities in CCF are X.509 certificates. They can be added or removed via governance proposals, which are subject to the consortium constitution rules (see :ref:`members/open_network:Adding Users`).
 
-Users should have previously been trusted by the consortium of members for them to use this authentication mechanism (see :ref:`members/open_network.html:Adding Users`). 
+Requests emitted by users can be authenticated one of two ways:
```
ashamis, line 6: minor: consider changing emitted to sent, i think this is more precise

ashamis (approved):

doc/developers/auth/cert.rst
achamayou: Yes, agreed, I'd just updated the link in that sentence but I will refresh it. `require_client_identity` is probably not a great name given the current behaviour, since a signed request is effectively tied to a client identity, more strongly than one over a authenticated connection if anything.

As discussed with @letmaik, we probably want to revisit this, and default to a policy that amounts to "valid user identity, via signature or session, or valid jwt", at least for app endpoints.

doc/developers/auth/cert.rst
achamayou: Done

PR #1900, opened 2020-11-18
Include evidence seqno in committed snapshot filename

### LINKED ISSUE #1724: Snapshot file name should include evidence seqno
To serve historical queries, it is much quicker to read the latest snapshot before the historical target `target_seqno`. However, it is not guaranteed that the evidence of this snapshot is present at a seqno included between `snapshot_seqno` and `target_seqno`. For example, if `snapshot_seqno < target_seqno < next_signature_seqno < evidence_seqno`, we wouldn't be able to verify that the snapshot we deserialised is valid.

By including the snapshot evidence `evidence_seqno` in the snapshot name, we can make sure that we only deserialise a snapshot whose evidence is verifiable. In other words, we can guarantee that `snapshot_seqno < evidence_seqno < target_seqno < next_signature_seqno`.

jumaffre:
Resolves #1724 

We now include the seqno at which the snapshot evidence was committed to the corresponding snapshot file name. This will be helpful for #1539 to select the correct snapshot file to recovery/join from and make sure we fail early/fall back to a previous snapshot if the evidence is not included in a known ledger file.

Before:

```bash
$ ls snapshots/
snapshot_10.committed
snapshot_20.committed
```

After:

```bash
$ ls snapshots/
snapshot_10.committed_12
snapshot_20.committed_21
```


ashamis (approved):

src/host/snapshot.h
```diff
@@ -57,7 +57,7 @@ namespace asynchost
       if (fs::exists(full_snapshot_path))
       {
         throw std::logic_error(fmt::format(
-          "Cannot write snapshot at {} since file already exists: {}",
+          "Error: Cannot write snapshot at {} since file already exists: {}",
```
achamayou, line 60: We don't consistently prefix our logic_errors with "Error:", I'm not sure we should, but I think we shouldn't do it just for some of them.

achamayou (approved):

PR #1903, opened 2020-11-19
Add benchmark covering digest algorithms

achamayou:


src/crypto/test/digest_bench.cpp
```diff
@@ -0,0 +1,50 @@
...
+  }
+  s.stop_timer();
+}
+
+const std::vector<int> hash_sizes = {2 << 6, 2 << 8, 2 << 12, 2 << 16, 2 << 18};
+
+PICOBENCH_SUITE("SHA-256");
+
+PICOBENCH(evercrypt_sha256).iterations(hash_sizes).baseline();
+PICOBENCH(mbedtls_sha256).iterations(hash_sizes).baseline();
```
eddyashton, line 50:
```suggestion
PICOBENCH(mbedtls_sha256).iterations(hash_sizes);
```

src/crypto/test/digest_bench.cpp
```diff
@@ -0,0 +1,50 @@
...
+  std::vector<uint8_t> v(s.iterations());
+  for (size_t i = 0; i < v.size(); ++i)
+  {
+    v.data()[i] = rand();
+  }
+
+  crypto::Sha256Hash h;
+
+  s.start_timer();
+  for (size_t i = 0; i < 1000; ++i)
```
eddyashton, line 20: Is the performance difference as stark if we don't repeat this 1000 times? I think we care more about the single-iteration speed than hot-cache looping.

src/crypto/test/digest_bench.cpp
```diff
@@ -0,0 +1,50 @@
...
+  {
+    crypto::Sha256Hash::mbedtls_sha256(v, h.h.data());
```
eddyashton, line 40: Nit: We could combine these so we have less duplicated code to keep in sync, but its a little awkward. Something like:

```
template<HashImpl IMPL>
static void digest_bench(...)
{
...
if constexpr (IMPL == HashImpl::mbedtls)
{
  crypto::Sha256Hash::mbedtls_sha256(v, h.h.data());
}
else if constexpr (IMPL == HashImpl::evercrypt)
{
  crypto::Sha256Hash::evercrypt_sha256(v, h.h.data());
}
...
}

auto mbedtls_sha256 = digest_bench<HashImpl::mbedtls>;
auto evercrypt_sha256 = digest_bench<HashImpl::evercrypt>;
```

eddyashton (commented):
I'm surprised this isn't already compared by `history_bench`. It looks like we're separately benchmarking the evercrypt and mbedtls implementations in that, but not directly comparing them in the same benchmark.

src/crypto/test/digest_bench.cpp
achamayou: I don't know, but it's guaranteed to be much less stable and therefore much harder to compare. We do hash a fair bit in the merkle tree, and this is only going to increase with multi-label receipts, so while we won't do 1000 in a row, we'll do more than the odd one shot digest.

src/crypto/test/digest_bench.cpp
achamayou: Yeah, I almost did it, and then I realised it was actually just as much code as duplicating. I don't mind.

achamayou:
@eddyashton I'll have a look at the history bench to see if we can easily mod it to test the tree with either hash variant there too.

src/crypto/test/digest_bench.cpp
achamayou: The answer seems to be yes, by the way:

```
amchamay@amchamay2:/data/amchamay/CCF/build$ ./digest_bench 
SHA-256:
evercrypt_digest_sha256 * |     128 |     0.001 |       7 |      - |142222222.2
    mbedtls_digest_sha256 |     128 |     0.001 |       5 |  0.778 |182857142.9
evercrypt_digest_sha256 * |     512 |     0.003 |       5 |      - |196923076.9
    mbedtls_digest_sha256 |     512 |     0.002 |       3 |  0.769 |256000000.0
evercrypt_digest_sha256 * |    8192 |     0.034 |       4 |      - |243086053.4
    mbedtls_digest_sha256 |    8192 |     0.026 |       3 |  0.760 |320000000.0
evercrypt_digest_sha256 * |  131072 |     0.534 |       4 |      - |245497317.9
    mbedtls_digest_sha256 |  131072 |     0.402 |       3 |  0.754 |325804182.4
evercrypt_digest_sha256 * |  524288 |     2.133 |       4 |      - |245808086.2
    mbedtls_digest_sha256 |  524288 |     1.614 |       3 |  0.757 |324794807.1
```

ashamis (approved):

Issue #1893, opened 2020-11-17
JWT public signing key background refresh


letmaik:
This is step 3 of https://github.com/microsoft/CCF/issues/1833.

## Summary

Instead of manually refreshing the JWT signing keys, the idea is to instead only manually refresh the (Root) CA certs of the JWT signing authority and fetch the keys from there periodically in the background, while verifying the TLS connection is genuine by checking against the stored CA cert (chain). The advantage is that CA certs are updated much more rarely (years) than signing keys (days), so fewer proposals have to be issued.

### CA certs

CCF has the `update_ca_cert` proposal type which stores a DER-encoded public key / cert under a given name in the `ccf.gov.ca_cert_ders` table. This was originally meant for storing JWT signing keys but has since been re-designed and uses different tables. Given that the name matches nicely, we can re-purpose it for storing general TLS CA certs.

### JWT issuer metadata

The existing JWT issuer metadata stored in `ccf.gov.jwt_issuers` is extended with:
- `ca_cert_name` -- name of cert in `ccf.gov.ca_cert_ders` table; used for auto-refreshes only
- `auto_refresh` -- true/false

Note: There is no option for configuring the key refresh interval, it'll be 30min and apply for all issuers.

## Implementation design

CCF already has timer functionality which is used for the join protocol, see `node/node_state.h::join`. This is equivalent to JavaScript's `setTimeout`, meaning to achieve periodic refreshes, the timer task callback schedules a new task to keep going. The timer runs on all nodes, but should only refresh keys on the primary, so it should skip any processing if the node is not the primary. If there is a network/HTTP error during key fetching, the failure should be recorded in some way and the key fetching is retried at the next scheduled event.

For now, we ignore the case of triggering a background key refresh based on seeing a JWT with an unknown key id. This is generally done to handle the case where a token issuer updates its signing keys more frequent than expected (in case of emergency or similar). The current workaround is to refresh keys more often than needed, e.g. every 1-2h.

Once the callback of the timer has fetched the new keys it sends them to a new internal member RPC endpoint (see `node/node_state.h::send_create_request/serialize_create_request` for an example) which in turn stores them in the kv in a transaction. The endpoint is unauthenticated but checks that the request originated from the current node by checking the caller cert. Using the endpoint abstraction for this provides a convenient existing mechanism to implement this.

[...]

Issue #1486, opened 2020-08-06
cchost does not print an error when trying to access an invalid target RPC address


eddyashton:
When launching a new node to join an existing network, we run `cchost ... join --target-rpc-address <address>`. If the address is invalid or unreachable, the joining node appears to hang and, with the default logging configuration, does not indicate that anything is wrong:

```
2020-08-06T12:29:39.506015Z        100 [info ] ../src/host/node_connections.h:161   | Listening for node-to-node on 127.98.160.202:37457
2020-08-06T12:29:39.506406Z        100 [info ] ../src/host/rpc_connections.h:75     | Listening for RPCs on 127.98.160.202:33653
2020-08-06T12:29:39.506562Z        100 [info ] ../src/host/main.cpp:608             | Creating new node - joining existing network at 127.1.2.3:1234
Azure Quote Provider: libdcap_quoteprov.so [ERROR]: Could not retreive environment variable for 'AZDCAP_DEBUG_LOG_LEVEL'
2020-08-06T12:29:39.553100Z        100 [info ] ../src/host/main.cpp:638             | Created new node
2020-08-06T12:29:39.553507Z -0.047 0   [info ] ../src/node/node_state.h:1071        | Node TLS connections now accepted
2020-08-06T12:29:39.563806Z -0.057 0   [info ] ../src/enclave/main.cpp:119          | Starting thread: 0
2020-08-06T12:29:39.563917Z -0.058 0   [info ] ../src/enclave/main.cpp:126          | All threads are ready!
```

Under the hood, the enclave is repeatedly sending join requests every time its timer expires, which attempts to create a connection via `tls::tls_connect`, which fails on the host side with all failure diagnostics at `DEBUG` and below.

We should make this failure more visible. At a minimum we should increase the verbosity of the RPC connection failure logging, and make it clear when this address appears invalid. We may also be able to specify an explicit behaviour for this connection which informs the _enclave_ of total connection failures.

achamayou:
Slightly different but related: #733

putyoutoosleep00:
Fuc you

achamayou:
Offtopic comment containing inappropriate language deleted.

letmaik:
I'm running into the same problem now when trying to figure out if JWT key refreshes were successful or not, which in the case of connection issues is impossible. I think the http client in the enclave should either have a second error callback, or have special values passed in the current callback, e.g. status=0.

rengolin:
[...]

Issue #1904, opened 2020-11-19
Support declaration of multiple auth policies for each endpoint


eddyashton:
Following on from https://github.com/microsoft/CCF/pull/1898#discussion_r526373762 and related discussion.

### Problem outline

We will soon support at least 3 broad auth policies: TLS cert, HTTP signature, JWT bearer token. We'd like a single endpoint to accept several of these, and the default is likely to be supporting _any_ of these. We need a way to declare what auth policies an endpoint requires, and to retrieve the authenticated identity information afterwards, inside the endpoint.

As a related point, we currently duplicate authentication policies for members and users, and choose which applies at a frontend-wide level - this seems like a good opportunity to make this an explicit and per-endpoint property, as a step towards eventually abolishing the frontend.

We want these policies to be extensible in C++. App writers should be able to write their own custom AuthN policy, though they may not be able to cache the parsed identity info they produced on the way.

When multiple policies are acceptable for an endpoint, we should try them in order and halt after the first successful auth. We need a way for the endpoint to learn which policy was applied (and thus which identity is being used, and which user metadata should be used for additional authorization decisions).

A complication with the current code is that for cert-based auth we try to do the check early on backups, before forwarding. As an optimisation, we don't forward the caller's cert in this case. We either need to do this consistently, forwarding all of the constructed metadata, or re-parse it on the forwardee, or abandon backup-auth.

### Code sketches

This proposes that each policy is just a functor, setting all return state on some `AuthCtx` and maybe HTTP response errors on the original request context. I think this is simpler than making it a richer self-contained object and re-extracting those fields. I think this could be a `ReadOnlyTx` - separately we want a `Endpoint::set_record_all_requests(true)` decorator that would be used for signed governance transactions, but could be used for others regardless of the auth scheme to dump the raw request to the ledger.

These include some open questions. In particular I've not decided the best way to set error messages. Maybe the auth functor sets response code + www-auth header directly, potentially overwritten by later functors, and on success overwritten by the 'real' result?

```c++
// Populates auth if request is acceptable, based on current state read from tx
// Should perhaps be ReadOnlyTx?
using AuthPolicy = std::function<void(
  Tx& tx, const std::shared_ptr<enclave::RpcContext>& request, AuthCtx& auth)>;
```

Then here's what an auth policy looks like, associated with some string identifier
```c++
static constexpr auto user_cert_policy_name = "user-cert";
void require_user_cert(
[...]

PR #1908, opened 2020-11-20
Add JWT public signing key auto-refresh

letmaik:
Implements https://github.com/microsoft/CCF/issues/1893.

Follow-up PRs (not necessarily all by myself):
- [Expose connection failures](https://github.com/microsoft/CCF/issues/1486#issuecomment-731081498) to enclave to be able to record key refresh errors in those situations
- In case of any error during key refresh, schedule next task with lower time interval, maybe gradually backing off
- Do an initial immediate key refresh after KV is synced (not sure where the best place is to trigger this)
- Move `update_ca_cert` proposal to C++, currently in Lua. Also, rename to `set_ca_cert` and add `remove_ca_cert` proposal.


src/node/node_state.h
```diff
@@ -596,6 +599,282 @@ namespace ccf
         std::chrono::milliseconds(config.joining.join_timer));
     }
 
+    template <typename T>
```
jumaffre, line 602: Can we move this new code to a different file? We should aim to have `node_state.h` only deal with the internal state of the node.

src/node/node_state.h
```diff
@@ -596,6 +599,282 @@ namespace ccf
...
+      http::Request request(fmt::format(
+        "/{}/{}",
+        ccf::get_actor_prefix(ccf::ActorsType::members),
+        "jwt_keys/refresh"));
+      request.set_header(
+        http::headers::CONTENT_TYPE, http::headervalues::contenttype::JSON);
+      request.set_body(&body);
+      // Need a custom authentication policy that accepts only node certs.
+      // See https://github.com/microsoft/CCF/issues/1904
+      // http::sign_request(request, node_sign_kp);
```
jumaffre, line 616: I might be missing something but if the signed request doesn't need to be recorded, presenting the `node_cert` in the `enclave::SessionContext` is probably sufficient to authenticate the node? A client wouldn't be able to present the node certificate in the TLS session as it would also require presenting the node private key which never leaves the node's enclave.

src/node/node_state.h
```diff
@@ -596,6 +599,282 @@ namespace ccf
...
+
+      const auto actor = rpc_map->resolve(actor_opt.value());
+      auto frontend_opt = this->rpc_map->find(actor);
+      if (!frontend_opt.has_value())
+      {
+        throw std::logic_error(
+          "RpcMap::find returned invalid (empty) frontend");
+      }
+      auto frontend = frontend_opt.value();
+      frontend->process(ctx);
```
jumaffre, line 637: Check that this has a return value?

src/node/rpc/endpoint_registry.h
```diff
@@ -626,6 +626,9 @@ namespace ccf
 
       for (const auto& [path, verb_endpoints] : fully_qualified_endpoints)
       {
+        // Special endpoint, can only be called from the node.
+        if (path == "jwt_keys/refresh")
```
jumaffre, line 630: Perhaps we should add an endpoint property to disable automatic OpenAPI generation?

src/node/rpc/member_frontend.h
```diff
@@ -1804,6 +1866,56 @@ namespace ccf
       make_endpoint("create", HTTP_POST, json_adapter(create))
         .set_require_client_identity(false)
         .install();
+
+      // Only called from node. See node_state.h.
+      auto refresh_jwt_keys = [this](
+                                EndpointContext& args, nlohmann::json&& body) {
+        // All errors are server errors since the client is the server.
```
jumaffre, line 1873: Here, we should check that `args.rpc_ctx.session.caller_cert` is a `TRUSTED` node in the nodes table.

src/runtime_config/gov.lua
```diff
@@ -98,4 +98,11 @@ return {
   end
   return true]],
 
+  update_ca_cert = [[
```
jumaffre, line 101: Should this be added to all governance sample files?

src/node/node_state.h
letmaik: The endpoint doesn't return anything, why should I check?

src/node/rpc/endpoint_registry.h
letmaik: @eddyashton said it's fine to hard-code for now. Should probably be discussed in a separate issue as it's related to hiding OpenAPI info when unauthenticated.

src/node/node_state.h
letmaik: Yeah you may be right. If I keep auth disabled but then check the cert in the endpoint manually then that's probably good enough.

src/runtime_config/gov.lua
letmaik: Theoretically yes, but I will rip this out in the next PR.

src/node/rpc/member_frontend.h
letmaik: Can I also check whether the request is coming from the same node?

src/node/node_state.h
letmaik: Sure, I identified the following things that I use from the node state instance:
- `std::lock_guard<SpinLock> guard(lock);`
- `rpcsessions->create_client`
- `network.tables->create_read_only_tx();`

I would leave the `auto_refresh_jwt_keys` method in node_state and move the rest out, passing rpcsessions and tx around. The only thing I'm not sure about is the lock. When do I need that?

src/host/main.cpp
```diff
@@ -318,6 +318,14 @@ int main(int argc, char** argv)
     "Subject Alternative Name in node certificate. Can be either "
     "iPAddress:xxx.xxx.xxx.xxx, or dNSName:sub.domain.tld");
 
+  size_t jwt_key_refresh_interval_s = 1800;
```
jumaffre, line 321: What motivated the default for this value? I can't seem to find how often JWKS document are typically updated and how long the overlap between the old and new keys is.

letmaik: It's generally one day or more, so this is very conservative. 1800 comes from https://github.com/AzureAD/passport-azure-ad/blob/5ce0f199af305ab3568facdfa4a4d31156d5bc2b/lib/bearerstrategy.js#L42

src/node/node_state.h
jumaffre: The `lock` is to protect the node's state so as long as you don't touch those (which I believe you don't), you should be fine to move this to a new class/file.

src/node/node_state.h
jumaffre: You should check that the endpoint returned `HTTP_STATUS_OK` at least?

src/node/node_state.h
letmaik: Some more thing:
- `node_cert`
- `rpc_map`

src/node/node_state.h
```diff
@@ -596,6 +599,282 @@ namespace ccf
...
+
+      LOG_DEBUG_FMT(
+        "JWT key auto-refresh: Received JWKS for issuer '{}'", issuer);
+
+      JsonWebKeySet jwks;
+      try
+      {
+        jwks = nlohmann::json::parse(data).get<JsonWebKeySet>();
+      }
+      catch (std::exception& e)
```
achamayou, line 676:
```suggestion
      catch (const std::exception& e)
```

src/node/node_state.h
```diff
@@ -596,6 +599,282 @@ namespace ccf
...
+        "JWT key auto-refresh: Received OpenID metadata for issuer '{}'",
+        issuer);
+
+      std::string jwks_url_str;
+      try
+      {
+        auto metadata = nlohmann::json::parse(data);
+        jwks_url_str = metadata.at("jwks_uri").get<std::string>();
+      }
+      catch (std::exception& e)
```
achamayou, line 723:
```suggestion
      catch (const std::exception& e)
```

src/node/node_state.h
letmaik: The endpoint either returns 200 or 500, and both are expected results.

src/node/node_state.h
letmaik: Let's defer this until #1904 is done.

src/node/rpc/member_frontend.h
letmaik: Let's defer this until #1904 is done. It's more tricky than it seems.

src/node/node_state.h
```diff
@@ -596,6 +599,282 @@ namespace ccf
...
+          e.what());
+        send_refresh_jwt_keys_error();
+        return;
+      }
+      http::URL jwks_url;
+      try
+      {
+        jwks_url = http::parse_url_full(jwks_url_str);
+      }
+      catch (std::invalid_argument& e)
```
achamayou, line 738:
```suggestion
      catch (const std::invalid_argument& e)
```

src/node/rpc/member_frontend.h
```diff
@@ -499,27 +516,23 @@ namespace ccf
           {
             tls::check_is_cert(der);
           }
-          catch (std::exception& exc)
+          catch (std::invalid_argument& exc)
```
achamayou, line 519:
```suggestion
          catch (const std::invalid_argument& exc)
```

tests/jwt_test.py
```diff
@@ -252,6 +260,156 @@ def test_jwt_with_sgx_key_filter(network, args):
...
+        assert m["failures"] > 0, m["failures"]
+
+    LOG.info("Restart OpenID endpoint server with new keys")
+    kid2 = "my_kid_2"
+    key2_priv_pem, _ = infra.crypto.generate_rsa_keypair(2048)
+    cert2_pem = infra.crypto.generate_cert(key2_priv_pem, cn=issuer_host)
+    jwks = create_jwks(kid2, cert2_pem)
+    with OpenIDProviderServer(issuer_port, key_priv_pem, cert_pem, jwks):
+        LOG.info("Wait for key refresh to happen")
+        time.sleep(2)
```
achamayou, line 406: We do not want one-shot waits for asynchronous operations they're brittle. This should instead be implemented as a loop on a concrete condition with an eventual timeout that's long enough to work even when the CI is a bit slow.

achamayou (changes requested):

tests/jwt_test.py
letmaik: Done

jumaffre (approved):

achamayou (approved):

Issue #1833, opened 2020-10-28
JWT authentication


letmaik:
This issue is a meta-issue to document overall ideas/design. Concrete tasks will have separate issues.

CCF has the `update_ca_cert` proposal type which stores a DER-encoded public key / cert under a given name in the `ccf.gov.ca_cert_ders` table. The purpose of this is for apps to be able to validate the signature of JWTs (=tokens), which are part of the `authorization` HTTP header as part of JWT authentication. Naming of keys is left to the app developer. Note that the proposal and table name is suboptimal as the keys are not of certificate authorities (meaning the key is used to sign a token, not to sign another cert that signs a token). Implementing JWT auth in apps is already possible but cumbersome.

Some notes on terminology:
- "JWT auth" is not something that is directly defined as part of an RFC. The closest is [RFC6750](https://tools.ietf.org/html/rfc6750) which defines how to call APIs with a "bearer" access token, where the bearer token can be anything, and in our case we require JWTs as bearer tokens. See also the [Web APIs](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-app-types#web-apis) section in the auth protocols docs of Microsoft. Note that this is distinct from supporting [OpenID Connect](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-app-types#web-apps) authentication, which also deals with JWT tokens but in this case the app/server is directly involved in the login flow of the user (traditional web apps). A modern web app type is the [single-page app](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-app-types#single-page-apps-javascript) where the auth flow is driven completely by the client using JavaScript (our Forum sample), and the app is called as in the Web API case above using bearer tokens, without knowing about the login flow of the client.
- Validating JWT bearer tokens is similarly not standardized, however, at a minimum, the JWT/JWS validation rules from [RFC7519](https://tools.ietf.org/html/rfc7519#section-7.2) and [RFC7515](https://tools.ietf.org/html/rfc7515#section-5.2) should be followed. Depending on the issuer of the tokens, additional non-app-specific validation of some body claims may be done. If the token issuer is an OpenID Connect compatible provider (which is typically the case), then this typically means checking that the `iss` claim matches the issuer URL. An issuer URL is a concept from OpenID Connect and is also part of the OpenID provider metadata.

Given that JWT bearer token authentication is by now an industry standard it should be supported more directly in CCF. This means:
- Apps should not do token signature validation themselves.
[...]

PR #1912, opened 2020-11-23
Native JWT validation

letmaik:
Step 2 of https://github.com/microsoft/CCF/issues/1833.

This PR adds JWT auth in the style of the other auth methods. It will be refactored once https://github.com/microsoft/CCF/issues/1904 is in but provides an early testbed.

I created a new struct `Jwt` to hold post-validation data of the token, but haven't actually extended `EndpointContext` with it yet as it may be a bit premature, depending on how #1904 unfolds. Opinions welcome.


src/node/rpc/frontend.h
```diff
@@ -352,6 +354,48 @@ namespace ccf
...
+            error_reason = "JWT signature is invalid";
+          }
+          else
+          {
+            key_issuer = key_issuer_view->get(key_id).value();
+          }
+        }
+        if (!error_reason.empty())
+        {
+          set_response_unauthorized(
```
eddyashton, line 388: This sets a header telling the client to use `Signature` auth. In this case we should instead indicate that this requires `Bearer` auth.

achamayou (approved):

src/node/rpc/endpoint.h
```diff
@@ -70,7 +72,8 @@ namespace ccf
       execute_locally,
       require_client_signature,
       require_client_identity);
-    DECLARE_JSON_OPTIONAL_FIELDS(EndpointProperties, openapi);
+    DECLARE_JSON_OPTIONAL_FIELDS(
+      EndpointProperties, require_jwt_authentication, openapi);
```
jumaffre, line 76: Why is this optional?

tests/js-authentication/authentication.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: We've already got a `jwt_test.py` test. Perhaps add it there instead?

src/tls/base64.h
```diff
@@ -39,6 +39,32 @@ namespace tls
...
+          break;
+      }
+    }
+    auto padding =
+      b64_string.size() % 4 == 2 ? 2 : b64_string.size() % 4 == 3 ? 1 : 0;
+    while (padding > 0)
+    {
+      b64_string.push_back('=');
+      padding--;
+    }
```
eddyashton, line 64: I think this padding calculation is missing the `size % 4 == 1` case? And there must be a neater way to express this, I think:

```suggestion
    // Add =s chars to pad string to a multiple of 4
    const auto padding = 3 - ((b64_string.size() - 1) % 4);
    b64_string += std::string(padding, '=');
```

We have some unit tests for b64 encode/decode in `tls/test/main.cpp` - can you add one for `raw_from_b64url` testing different padding sizes?

eddyashton: Alternatively, perhaps the most readable:

```suggestion
    while (b64_string.size() % 4 != 0)
    {
      b64_string.push_back('=');
    }
```

src/http/http_jwt.h
```diff
@@ -0,0 +1,162 @@
...
+  DECLARE_JSON_TYPE(JwtHeader)
+  DECLARE_JSON_REQUIRED_FIELDS(JwtHeader, alg, kid)
+
+  class JwtVerifier
+  {
+  public:
+    struct Token
+    {
+      nlohmann::json header;
+      JwtHeader header_typed;
```
eddyashton, line 38: Do we need to keep both of these around? If we can get away with it, I think we should store just the JSON `header` in here - the verifier can extract whatever fields it needs when verifying, but these don't need to be retained from parsing?

letmaik: It roughly goes: parse JWT, get keyid from token, lookup cert in kv matching keyid, validate token using cert. So, you need a typed variant to get the key id out. Of course I could move the following into `frontend.h` but that's suboptimal since I wanted all the parsing logic in http_jwt.h:
```
      JwtHeader header_typed;
      try
      {
        header_typed = header.get<JwtHeader>();
      }
      catch (const nlohmann::json::exception& e)
      {
        error_reason =
          fmt::format("JWT header does not follow schema: {}", e.what());
        return std::nullopt;
      }
```

src/node/rpc/endpoint.h
letmaik: So that existing app.json's don't break.

tests/js-authentication/authentication.py
letmaik: The new test is jwt only, yes, but the intended purpose is to cover other auth schemes as well, looking at #1904. Since we have folders now for each JS test app it makes sense to have the test module in that folder as well, like a self-contained package. Another argument is that jwt_test.py is for governance things, where the new one is for app/runtime logic.

src/tls/base64.h
letmaik: https://en.wikipedia.org/wiki/Base64#Output_padding. You pad at most two characters.

letmaik: I added a test case for base64url for different sizes, like for base64.

src/node/rpc/endpoint.h
jumaffre: OK, that's what I suspected. Is the plan to move this to "required" once the JS apps are updated?

tests/js-authentication/authentication.py
jumaffre: OK, makes sense. Does the `js-authentication` JS app currently double verify the token (once natively, once in JS)?

src/node/rpc/endpoint.h
letmaik: With #1904 this will be replaced by something else.

tests/js-authentication/authentication.py
letmaik: No, just natively.

jumaffre (approved):

src/node/rpc/endpoint.h
achamayou: This is a good illustration of why expanding all those fields in the app.json was not a good idea...

PR #1913, opened 2020-11-23
Replace http-parser with llhttp

### LINKED ISSUE #1910: Explore replacing http-parser
The [http-parser](https://github.com/nodejs/http-parser) lib we currently use is [no longer maintained](https://github.com/nodejs/http-parser/issues/522). This isn't an urgent issue, but we should look at moving to a supported replacement, perhaps the recommended [llhttp](https://github.com/nodejs/llhttp).

eddyashton:
Closes #1910.

Replaces [http-parser](https://github.com/nodejs/http-parser) with [llhttp](https://github.com/nodejs/llhttp). Specifically v2.2.0, from the release package which contains the extracted C sources.

Most of the changes are renames (replacing `http_` with `llhttp_` for various types and functions). `http_status.h` defines `HTTP_STATUS_XXX` constants that we use widely, but aren't included in `llhttp`. The bit of URL parsing we need (splitting query from path) is implemented inline by us rather than the lib. `execute` no longer returns the count of bytes parsed (it always parses everything given), so we remove a bunch of error checking around calls.


src/http/http_endpoint.h
```diff
@@ -99,34 +99,10 @@ namespace http
...
-            else
-            {
-              // Used some bytes - pass over these and retry with remainder
-              data += used;
-              n_read -= used;
-            }
+            p.execute(data, n_read);
+
+            // Used all provided bytes - check if more are available
+            n_read = read(buf.data(), buf.size(), false);
```
achamayou, line 105: Nice!

achamayou (approved):

jumaffre (approved):

ashamis (approved):

PR #1923, opened 2020-11-25
Print CLI options on every node launch

eddyashton:
I find the connection between the Python output and the node output is sometimes hard to track, especially when comparing across multiple `out` files. Printing the CLI options and node ID makes the node output self-describing, so they disambiguated by searching for "self node ID" (and IP before that) rather than filename.


jumaffre (approved):
In the Python infra, should we also prefix the `out` and `err` files with the node id?

achamayou:
@jumaffre they should already be in separate files.

@eddyashton I looked at this, but relented because it doesn't work well if a config is passed. It's probably still a good idea for our tests though.

achamayou (approved):

eddyashton:
@achamayou AFAICT, config file parsing is slightly busted regardless of this change? If I try to use this dumped config, it includes a `version=false` line which is apparently sufficient to trigger the `print_version` flag, so that line needs to be deleted. And then it gets confused about the presence of multiple `network-cert-file` args (`start.network-cert-file` has a default value, so we parse it but then insist it shouldn't exist, even though the command was `join`). But if I trim enough options out of the config, I can get it to launch from this config (and print the full set of options) just the same as launching with all args on the CLI.

jumaffre:
> @jumaffre they should already be in separate files.

@achamayou I'm not sure what you mean. Prefixing the `out` and `err` with the node id would help when downloading files from a failed CI run. At the moment for example, all node output files for different nodes are named `out`, which makes it hard to debug efficiently when jumping between different output files.

achamayou:
@eddyashton if that's the case, then we need to fix it. I think you mean the config generated that way is busted, I would hope that the parsing of well-formed config is ok.

@jumaffre we're going to have sha256 as node ids in the not-too-distant future, so while I agree that the current situation is imperfect, I am not sure this will be a significant improvement.

PR #1925, opened 2020-11-25
Verify snapshot evidence on join/recover

### LINKED ISSUE #1539: Snapshot auditability

jumaffre:
Resolves #1539 

When joining/recovering from a snapshot, we used to automatically accept the snapshot, even if the snapshot evidence wasn't present in the ledger. We now verify that the snapshot evidence is present in the ledger and that there is proof that this evidence was committed, i.e. that there's a signature that proves that the commit seqno was greater than the evidence seqno.

On recovery, this is pretty much the same scheme as before but we make sure the snapshot evidence is in the ledger when recovering the public part. On join, we now apply the snapshot in public mode when on start-up and deserialise the ledger suffix (in public mode) until we've found the proof of the snapshot evidence.

Operation-wise, this now means that node joining from a snapshot should have access to the ledger suffix containing the snapshot evidence (+ signature proving that the evidence was committed). I'll update the documentation accordingly in a follow-up PR. 

The Python infra requires some changes when joining/recovering from a snapshot. In particular, we now wait for the latest available snapshot to have ledger proof of the commit of its evidence (see inline comment).


tests/infra/network.py
```diff
@@ -154,6 +154,34 @@ def _adjust_local_node_ids(self, primary):
             for n in self.nodes:
                 n.node_id = n.node_id + first_node_id
 
+    def wait_for_snapshot_evidence_commit_proof(self, node, snapshot, timeout=3):
```
jumaffre, line 157: This is a little awkward but I believe the easiest way to programmatically test that the commit of the evidence of given snapshot is covered by a signature in the ledger.

src/host/snapshot.h
```diff
@@ -13,13 +13,42 @@ namespace fs = std::filesystem;
...
+    auto pos = file_name.find(
+      fmt::format("{}{}", snapshot_committed_suffix, snapshot_idx_delimiter));
+    if (pos == std::string::npos)
+    {
+      // Snapshot is not yet committed
+      return std::nullopt;
+    }
+
+    auto committed_suffix = file_name.substr(pos);
+    return std::stol(committed_suffix.substr(
```
achamayou, line 44: This is returning a long as a size_t, so this would technically accept a negative value which would come out as a large unsigned. `std::from_chars()` is more precise and would validate this away.

This is untrusted code anyway, but since it's not much more verbose, I think it's still reasonable to make the change.

tests/infra/network.py
```diff
@@ -169,16 +200,30 @@ def create_node(self, host):
...
 
+    def get_committed_snapshots(self, node):
+        """
+        Returns committed snapshots, waiting for the latest available
+        snapshot to have evidence proof in ledger
+        """
+        snapshot_dir = node.get_committed_snapshots()
+        assert (
+            len(os.listdir(snapshot_dir)) > 0
+        ), f"There are no snapshots to resume from in directory {snapshot_dir}"
```
achamayou, line 211:
```suggestion
        assert (
            os.listdir(snapshot_dir)
        ), f"There are no snapshots to resume from in directory {snapshot_dir}"
```

achamayou (approved):

PR #1928, opened 2020-11-25
Do not use global hook to open service

achamayou:
This removes our sole use of the global commit hook, although not the mechanism itself in this particular PR. If we're happy with it, I want to take it out.


src/node/rpc/frontend.h
```diff
@@ -591,19 +593,44 @@ namespace ccf
...
     }
 
-    bool is_open() override
+    bool is_open(kv::Tx& tx) override
     {
       std::lock_guard<SpinLock> mguard(open_lock);
+      if (!is_open_)
+      {
+        auto sv = tx.get_view<Service>(Tables::SERVICE);
+        auto s = sv->get_globally_committed(0);
```
eddyashton, line 623: Why does this need to be globally committed? If it's readable, then the current transaction is ordered after that identity was written, and everything works? We're hoping to drop `get_globally_comitted` (and make the entire _transaction_ read-only for committed state), so hopefully this isn't necessary.

Separately, why do we only do this for the user frontend? I see that node and members are not currently opened transactionally, but I think there's no harm in making them consistent, and all read an explicit `frontend_state` table? (Where members and nodes would be written as open during the `create` transaction, and users during `open_network`).

achamayou: Because if it's not globally committed, that transaction could conceivably be rolled back, but the user frontend would still be open, which is not what we want. I know we're hoping to drop it from the user facing API, and I still think we should, I'm less convinced it's necessary to drop it as an internal call, since it's useful here.

There's no reason to ever have the node frontend be closed, or even to wait for a KV to be there. It's sufficiently low level that it should really be on pretty much all the time (as soon as it's initialised enough no to crash anyway). Similarly, the member frontend is necessary to vote on open.
Why is the user frontend open delayed is the part that's less clear to me. It's explicitly a design decision (see TR), and it's perhaps to protect users from using an unapproved network?

src/node/node_state.h
```diff
@@ -532,6 +532,10 @@ namespace ccf
               "Node has now joined the network as node {}: {}",
               self,
               (resp.network_info.public_only ? "public only" : "all domains"));
+
+            // The network identity is now know, the user frontend can be opened
```
jumaffre, line 536:
```suggestion
            // The network identity is now known, the user frontend can be opened
```

src/node/rpc/frontend.h
achamayou: As discussed with @eddyashton, there are three alternatives here:

1. Read this every time, which I contend will be noticeably expensive
2. Do a local read, at the expense of building a rollback mechanism. It's feasible, but more complex with virtually no user benefit.
3. Do it this way, either through get_globally_committed(), or by comparing the versions manually the establish the same thing (which I think amounts to a slightly watered down 2).

It's arguable whether the notion of service state is even useful, because members explicitly enable users or token CAs. It's used incidentally for C++ endpoint management, and that would need to change if it was removed. The cost/benefit ratio does not justify its removal.

So this is 3. for now. It's functionally identical to the status quo, does not incur a performance penalty, and does remove an execution path we are wary of for thread safety reasons (hooks) in favour of plain tx execution (albeit with a local cache, but there is well-understood precedent for that such as the verifiers).

src/node/rpc/frontend.h
```diff
@@ -591,19 +593,44 @@ namespace ccf
       cmd_forwarder = cmd_forwarder_;
     }
 
-    void open() override
+    void open(std::optional<tls::Pem*> identity = std::nullopt) override
     {
       std::lock_guard<SpinLock> mguard(open_lock);
-      if (!is_open_)
+      // open() without an identity unconditionally opens the frontend
+      // if an identity is passed, the frontend must instead wait for
```
jumaffre, line 600:
```suggestion
      // open() without an identity unconditionally opens the frontend.
      // If an identity is passed, the frontend must instead wait for
```

jumaffre (approved):

src/node/rpc/frontend.h
```diff
@@ -688,6 +721,10 @@ namespace ccf
       std::shared_ptr<enclave::RpcContext> ctx) override
     {
       auto tx = tables.create_tx();
+      if (!is_open(tx))
+      {
+        throw std::logic_error("Transaction failed");
```
ashamis, line 726: Please add a comment here that this state can only be reached with a malicious primary so we should trigger a view change when catching this

ashamis (approved):

src/node/rpc/frontend.h
achamayou: Done.

PR #1932, opened 2020-11-25
Install logging JS app

jumaffre:
The primary goal of this PR is to install the logging JS application so that a test CCF service running a simple application can be run with the sandbox. In other words, it is now possible to install the CCF debian package and run the `sandbox.sh` script (with no arguments) to get a virtual CCF test service running the logging JS app.

Other changes:
- Moved `logging` app from `src/apps/` to `samples/apps/`
- Moved `simplebank` app from `samples/apps` to `src/apps`
- `test_install.sh` script now makes use of sandbox
- Fixed issue with JS logging app where the `/log/public` endpoints weren't reachable


tests/test_install.sh
```diff
@@ -17,50 +17,32 @@ rm -rf "$working_dir"
 mkdir -p "$working_dir"
 cd "$working_dir"
 
+# Start ephemeral network in the background
+network_live_time=60
+timeout --signal=SIGINT --kill-after=${network_live_time}s --preserve-status ${network_live_time}s \
+"$INSTALL_PREFIX"/bin/sandbox.sh --verbose &
+
+# Wait for service to be open
+sleep 45
```
achamayou, line 26: That seems like a long time, but I guess we don't have a convenient file to wait on anymore.

tests/test_install.sh
```diff
@@ -17,50 +17,32 @@ rm -rf "$working_dir"
...
+
+# # Issue tutorial transactions to ephemeral network
 python3.8 -m venv env
 source env/bin/activate
-python -m pip install -U -r "$INSTALL_PREFIX"/bin/requirements.txt
 python -m pip install ../../../python
+python ../../../python/tutorial.py ./workspace/sandbox_0/0.ledger/ ./workspace/sandbox_common/
```
achamayou, line 32: Maybe the waiting ought to happen at the top of tutorial, using the client timeout logic we have in Python anyway?

tests/test_install.sh
```diff
@@ -17,50 +17,32 @@ rm -rf "$working_dir"
...
-# Issue tutorial transactions to ephemeral network
-python ../../../python/tutorial.py "$network_info_file"
+../../../tests//test_python_cli.sh > test_python_cli.out
 
 # Wait until original network has died
-sleep ${network_live_time}
-
-# ...and a tad longer to be sure
-sleep 5
+sleep 20
```
achamayou, line 38: Likewise, that seems like something the Python code in tutorial could more easily and precisely wait for.

achamayou (approved):

tests/test_install.sh
eddyashton: Yes I think we should find a better way to wait. We could grep the output of `sandbox.sh`, or wait for some file to be written by the infra, or have a separate util (in bash or Python) which polls the node until it seems active?

jumaffre: Good point. I've added a quick check for the service being open with `curl`.

tests/test_install.sh
jumaffre: As discussed earlier, we wait for service opening directly inside `test_install.sh` (where the `sleep ...` was). I believe this is neater and keeps the `tutorial.py` untouched.

jumaffre:
Looks like the daily build failed when testing the install: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=15807&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=fe98d859-2537-5570-75af-de67d4290f3a

Investigating this now to find out what the issue is.

tests/sandbox/sandbox.sh
```diff
@@ -10,6 +10,21 @@ PATH_HERE=$(dirname "$(realpath -s "$0")")
...
 
+is_package_specified=false
+for item in "$@" ; do
+    if [ "$item" == "-p" ] || [ "$item" == "--package" ]; then
+        is_package_specified=true
+    fi
+done
+
+extra_args=("$@")
+if [ ${is_package_specified} == false ] && [ -f "${VERSION_FILE}" ]; then
```
letmaik, line 21: This change makes behaviour different for build vs install tree when only `--js-app-bundle` is given because the extra args here are added *after* the user args, overriding them and starting the wrong app.

PR #1945, opened 2020-11-27
Build both virtual and sgx on daily build and test SGX with test_install.sh

jumaffre:
Follow up from #1932

The daily build was failing as we only used to build the SGX target (unlike the "real" Release job in the GitHub CI pipeline).


achamayou (approved):

jumaffre:
Daily build is failing because of other issues (see `Perf Instrumented` job) so merging this PR regardless.

PR #1946, opened 2020-11-27
Move perf client out of samples directory

jumaffre:
The `plot_tx_times.py` script stays in the `samples/` directory.


achamayou:
Why do we keep that script there as opposed to tests or scripts?

achamayou (approved):

jumaffre:
@achamayou Using the script is documented on https://microsoft.github.io/CCF/master/design/performance.html?highlight=perf#end-to-end-performance-tests so I thought having it under `samples/` would make it easier for people to find. Happy to move it elsewhere.

achamayou:
@jumaffre ah good point, samples is good place then.

ashamis (approved):

PR #1953, opened 2020-11-27
Recover ledger last idx from read-only ledger directory on recovery

jumaffre:
Fix for the CI issues experienced on #1925 

When recovering a ledger, we assumed that there was at least one ledger entry in the main ledger directory that could inform the ledger of the idx of the last entry (`last_idx`). However, it is possible that only a read-only ledger directory (`--read-only-ledger-dir` option to `cchost`) was passed in. 

This PR fixes this issue by looking up all the ledger files in read-only ledger directories on start-up as well.


ashamis (approved):

src/host/ledger.h
```diff
@@ -625,6 +625,35 @@ namespace asynchost
...
+            fmt::format("\"{}\" is not a directory", read_dir));
+        }
+
+        for (auto const& f : fs::directory_iterator(read_dir))
+        {
+          auto last_idx_ = get_last_idx_from_file_name(f.path().filename());
+          if (!last_idx_.has_value())
+          {
+            LOG_DEBUG_FMT(
+              "Read only ledger file \"{}\" is ignored as not committed",
```
achamayou, line 644:
```suggestion
              "Read-only ledger file \"{}\" is ignored as not committed",
```

I don't feel strongly either way, but we hyphenate everywhere I can see, and I think consistency would be good.

achamayou (approved):

PR #1959, opened 2020-11-30
Free JS_Runtime leaked on exception paths

eddyashton:
We recently added new return paths in `execute_request` which weren't calling `JS_FreeRuntime`, causing memory leaks (triggered by the `js_batched_stress_test` and failing the Daily CI).

Rather than adding the appropriate `FreeXXX` calls on every path, we just wrap the `JSRuntime` and `JSContext` in auto-free structs.

I recommend viewing the diff with ignored whitespace, eg via [this link](https://github.com/microsoft/CCF/pull/1959/files?diff=split&w=1).


achamayou (approved):

src/apps/js_generic/js_generic.cpp
```diff
@@ -82,9 +82,26 @@ namespace ccfapp
     JS_FreeValue(ctx, exception_val);
   }
 
-  struct JSAutoFree
+  struct JSAutoFreeRuntime
   {
-    JSAutoFree(JSContext* ctx) : ctx(ctx) {}
+    JSRuntime* rt;
+
+    JSAutoFreeRuntime(JSRuntime* rt) : rt(rt) {}
```
ashamis, line 89: can you do the same things with a std::unique_ptr that specifics a custom deleter?

ashamis (approved):

src/apps/js_generic/js_generic.cpp
achamayou: You can, but it's not super satisfying because you pay for an extra pointer on every object, and you can't use make_unique, so it's both a little awkward to use and a somewhat expensive.

PR #1960, opened 2020-11-30
Ledger and snapshot management doc

jumaffre:
We used to have some documentation for operators on how to manage the ledger but it was 1) very succinct and 2) in the wrong place. 

This PR hopefully fixes that, adding more detail on the different ledger directories and how snapshots work with all of that. I've also added a quick "Best Practices" section at the bottom of the new documentation page to give some guidelines to operators when setting certain values when starting a CCF node.


doc/operators/ledger_snapshot.rst
```diff
@@ -0,0 +1,90 @@
...
+
+To validate the snapshot a node is added from, the node first replays the transactions in the ledger following the snapshot until the proof that the snapshot was committed by the service to join is found. This process requires operators to copy the ledger suffix to the node's ledger directory. The validation procedure is generally quick and the node will automatically join the service one the snapshot has been validated. On recovery, the snapshot is automatically verified as part of the usual ledger recovery procedure.
+
+For example, if a node is added using the ``snapshot_1000.committed_1250`` snapshot file, operators should copy the ledger files containing the sequence numbers ``1000`` to ``1250`` to the directories specified by ``--ledger-dir`` (or ``--read-only ledger-dir``). This would involve copying the ledger files following the snapshot sequence number ``1000`` until the evidence sequence number ``1250``, e.g. ``ledger_1001-1200.committed`` and ``ledger_1201-1500.committed``, to the joining node's ledger directory.
```
jumaffre, line 72: This is wrong as of #1925. We currently mark a snapshot as "committed" once its evidence has been (globally) committed but a recovery/join node only accepts a snapshot once there's proof in the ledger that its evidence has been (globally) committed. I'm in the process of fixing this but apart from the snapshot file name (which will probably include another seqno), the working of these two paragraphs will remain the same.

doc/audit/ledger.rst
```diff
@@ -1,50 +1,19 @@
...
-    -rw-rw-r-- 1 user user 1.1M Jun 16 14:08 ledger_92502-97501.committed
-    -rw-rw-r-- 1 user user 553K Jun 16 14:08 ledger_97502
-
 Ledger Encryption
 -----------------
 
 Each entry in the ledger corresponds to a transaction committed by the primary node.
 
-When a transaction is committed, each affected ``Store::Map`` is serialised in different security domains (i.e. public or private), based on the policy set when the ``Store::Map`` was created (default is private). A public ``Store::Map`` is serialised and stored in the ledger as plaintext while a private ``Store::Map`` is serialised and encrypted before being stored.
+When a transaction is committed, each affected ``Store::Map`` is serialised in different security domains (i.e. public or private), based on the name of the Map when it was created (default is private). A public ``Store::Map`` is serialised and stored in the ledger as plaintext while a private ``Store::Map`` is serialised and encrypted before being stored.
```
achamayou, line 15:
```suggestion
When a transaction is committed, each affected ``Store::Map`` is serialised in different security domains (i.e. public or private), based on the name of the Map when it was created (default is private). A public ``Store::Map`` (ie. one whose name starts with "public:") is serialised and stored in the ledger as plaintext while a private ``Store::Map`` is serialised and encrypted before being stored.
```

doc/operators/ledger_snapshot.rst
```diff
@@ -0,0 +1,90 @@
...
+===============================
+
+The ledger and snapshot files written by CCF nodes to disk should be managed by operators to allow for safe backup of the service and application state as well as fast join and recovery procedures. This section describes how these files are generated and how operators should manage them effectively.
+
+.. note:: See the :doc:`/audit/index` section to read about offline ledger auditability.
+
+Ledger
+------
+
+The ledger is the persistent replicated append-only record of the transactions that have been executed by the CCF service. It is written by the primary node when a transaction is executed and replicated to all backups which maintain their own duplicated copy. Each node in a network creates and maintains its own local copy of the ledger. Committed entries are always identical between :ref:`a majority <design/consensus:CFT Consensus Protocol>` of nodes, but a node may be more or less up to date, and uncommitted entries may differ.
```
achamayou, line 11: We're a little vague about transactions here, and whether that's reads/writes. We could/should perhaps elaborate for the BFT and CFT cases.

achamayou (approved):

PR #1964, opened 2020-11-30
Cleanup tls client

### LINKED ISSUE #738: Memory leaks in crypto code in error paths
(Hello CCF, sorry to cause trouble.)

**Description**
In CCF-level crypto code, there are multiple examples of non-trivial ctors that throw exceptions without freeing resources thus far created. (When c++ objects fail during construction, the object's dtor will *not* be called.) For example, the class TlsClient's ctor calls member function connect(), which does quite a bit of work and throws exceptions on error. Any resources created in the ctor prior to the exception will not be freed.

ccf\src\clients\tls_client.h

There are other examples of this pattern in CCF's higher-level crypto code. (By "higher-level" I mean in CCF proper, that is, not in 3rd-party code.)

**Expected behavior**
Memory leaks will not occur, including during failure cases.

achamayou:
fix #738


achamayou:
Although this doesn’t solve #1701, I think this is still a worthwhile change.

src/clients/tls_client.h
```diff
@@ -21,19 +21,85 @@
 #include <string>
 #include <vector>
 
+struct auto_free_mbedtls_net_context
```
eddyashton, line 24: These structs should be `CamelCase` for consistency with the rest of our code, ie `AutoFreeMbedtlsNetContext`.

eddyashton (changes requested):
I don't think this fixes the original linked issue. If we throw in `connect()`, we throw in the constructor of `TlsClient`, meaning we don't call the default destructor `~TlsClient()`, and in turn we don't call the destructors of the _members_ or the relevant `_free` functions. We either need to pepper `connect()` with `_free`s, or maybe:

```
TlsClient()
{
  try
  {
    connect();
  }
  catch (const std::exception&)
  {
    disconnect();
    throw;
  }
}
```

achamayou:
@eddyashton yes, you're right, I completely missed that.

src/clients/tls_client.h
```diff
@@ -109,18 +84,66 @@ class TlsClient
...
-  void disconnect()
+  void teardown()
   {
-    // Signal the end of the connection
-    mbedtls_ssl_close_notify(&ssl);
+    mbedtls_net_init(&server_fd);
+    mbedtls_ssl_init(&ssl);
+    mbedtls_ssl_config_init(&conf);
+    mbedtls_entropy_init(&entropy);
+    mbedtls_ctr_drbg_init(&ctr_drbg);
```
eddyashton, line 96:
```suggestion
    mbedtls_net_free(&server_fd);
    mbedtls_ssl_free(&ssl);
    mbedtls_ssl_config_free(&conf);
    mbedtls_ctr_drbg_free(&ctr_drbg);
    mbedtls_entropy_free(&entropy);
```

eddyashton (approved):

PR #1975, opened 2020-12-02
Only retire backup when test suite can progress

eddyashton:
I was trying to work out what failed in this test (a specific seed for `full_test_suite`):
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=16059&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=63d660c3-6f76-573a-0154-fd0bcb096ded&l=11078

I thought maybe we had a bug where we're not able to retire the final backup, as the special case progress for a single-node network wasn't triggered? But that's not the case - I tried retiring all backups and the final lone primary is happy to proceed.

The actual fix is much simpler. This seed had previously run the test which _kills_ the primary, without retiring it. So the configuration currently contains 3 nodes, 1 of whom is not participating, and sure enough the primary can't reach consensus if the sole live backup is retired. So we should only retire this backup if `can_kill_n_nodes` (which counts the service's `TRUSTED` nodes directly, not just the Python clients running nodes) says we can.

### DIFF
--- a/tests/reconfiguration.py
+++ b/tests/reconfiguration.py
@@ -43,24 +43,14 @@ def check_can_progress(node, timeout=3):
         assert False, f"Stuck at {r}"
 
 
-def add_node(network, args):
+@reqs.description("Adding a valid node from primary")
+def test_add_node(network, args):
     new_node = network.create_and_trust_node(args.package, "local://localhost", args)
-    assert new_node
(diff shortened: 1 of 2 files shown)


achamayou (approved):
Good catch!

Issue #1919, opened 2020-11-24
Use OData-style error responses


letmaik:
Part of https://github.com/microsoft/CCF/issues/1918.

Following https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#7102-error-condition-responses, error responses should follow [OData](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091) and look like that at a minimum:

```json
{
  "error": {
    "code": "BadArgument",
    "message": "Previous passwords may not be reused"
  }
}
```

Tasks:
- Establish a centralised list of error codes
  - The standard also allows grouping errors into base errors where the detailed error is in `"innererror"`, mostly for clients to handle categories of errors more easily
- Change all endpoints to return errors like above
- Adapt clients if needed

NOTE: Error codes are not defined as enumeration but rather just string and hence do not have to be [lowerCamelCase](https://github.com/microsoft/CCF/issues/1940). See https://github.com/microsoft/api-guidelines/issues/100.

PR #1980, opened 2020-12-03
Unified error responses

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1919.

Any and all feedback welcome, nothing is set in stone.

- Introduces `ctx->set_error()` to set an OData-style error response.
- Introduces error codes in `src/node/rpc/error.h`.
- Changed HTTP response status for some endpoints:
  - All requests: if tx failed to replicate or the commit attempt threshold was reached, then a `503 Service Unavailable` is now returned instead of `500 Internal Server Error` or `409 Conflict` (conflicts are always user errors that require intervention, but here it's a server issue), since it roughly means the server is busy and the situation can be fixed by trying again later. The "failed to replicate" case is a bit trickier (since you might have been talking to a primary that is not primary anymore, in which case retrying won't help, but it will help if you talked to a backup) but it definitely isn't a 500.
  - ../complete now returns 404 instead of 400 if the proposal is not found. Not that important since this endpoint will go away soon.

TODO / questions:
- Is it worth creating error codes for endpoints that are only called by nodes like /join and /create? E.g. `ConsensusTypeMismatch`.
  - Yes, for consistency.
- Should applications be able to influence the error response format for built-in endpoints?
  - Not now, but maybe later.
- Should `ctx->set_error()` be renamed or made into an internal utility function like `set_error(ctx, ...)`?
  - Fine as-is.
- Expose OData error in OpenAPI as schema and add as default response to all built-in endpoints.
  - Yes: https://github.com/microsoft/CCF/issues/2004


src/enclave/rpc_context.h
```diff
@@ -214,12 +215,31 @@ namespace enclave
       set_response_header(name, fmt::format("{}", n));
     }
 
+    virtual void set_error(http_status status, std::string&& msg)
```
ashamis, line 218: should this not be called set_response as 200 is a valid error code also

tests/js-modules/modules.py
```diff
@@ -370,11 +372,11 @@ def run(args):
...
         network.start_and_join(args)
-        network = test_module_set_and_remove(network, args)
-        network = test_module_import(network, args)
-        network = test_app_bundle(network, args)
+        #network = test_module_set_and_remove(network, args)
+        #network = test_module_import(network, args)
+        #network = test_app_bundle(network, args)
         network = test_dynamic_endpoints(network, args)
-        network = test_npm_app(network, args)
+        #network = test_npm_app(network, args)
```
ashamis, line 379: typo?

achamayou:
> The "failed to replicate" case is a bit trickier (since you might have been talking to a primary that is not primary anymore, in which case retrying won't help, but it will help if you talked to a backup) but it definitely isn't a 500.

Re-trying, even on a former primary turned backup could help because we do forward writes. It depends if the issue that caused the election was transient or permanent. I don't see why it shouldn't be a 503 too, it's an issue on the server side, which is likely temporary in nature and can be resolved by re-trying.

letmaik:
> 
> 
> > The "failed to replicate" case is a bit trickier (since you might have been talking to a primary that is not primary anymore, in which case retrying won't help, but it will help if you talked to a backup) but it definitely isn't a 500.
> 
> Re-trying, even on a former primary turned backup could help because we do forward writes. It depends if the issue that caused the election was transient or permanent. I don't see why it shouldn't be a 503 too, it's an issue on the server side, which is likely temporary in nature and can be resolved by re-trying.

Alright, 503 it is! (it's already like that in this PR)

src/enclave/rpc_context.h
```diff
@@ -214,12 +215,26 @@ namespace enclave
...
+      http_status status, const std::string& code, std::string&& msg)
+    {
+      set_error({status, code, std::move(msg)});
+    }
+
+    virtual void set_error(ccf::ErrorDetails&& error)
+    {
+      nlohmann::json body = ccf::ODataErrorResponse{
+        ccf::ODataError{std::move(error.code), std::move(error.msg)}};
+      const auto s = fmt::format("{}\n", body.dump());
```
achamayou, line 228: Not sure we need the new line here.

src/http/http_rpc_context.h
```diff
@@ -35,18 +36,27 @@ namespace http
...
 
-  static std::vector<uint8_t> error(size_t code, const std::string& msg)
+  inline std::vector<uint8_t> error(ccf::ErrorDetails&& error)
   {
-    http_status status = (http_status)code;
-    std::vector<uint8_t> data(msg.begin(), msg.end());
-    auto response = http::Response(status);
+    nlohmann::json body = ccf::ODataErrorResponse{
+      ccf::ODataError{std::move(error.code), std::move(error.msg)}};
+    const auto s = fmt::format("{}\n", body.dump());
```
achamayou, line 43: Another new line we maybe don't need.

src/http/ws_rpc_context.h
```diff
@@ -20,11 +21,21 @@ namespace ws
...
 
-  static std::vector<uint8_t> error(size_t code, const std::string& msg)
+  inline std::vector<uint8_t> error(ccf::ErrorDetails&& error)
   {
-    std::vector<uint8_t> ev(msg.begin(), msg.end());
-    return serialise(code, ev);
-  };
+    nlohmann::json body = ccf::ODataErrorResponse{
+      ccf::ODataError{std::move(error.code), std::move(error.msg)}};
+    const auto s = fmt::format("{}\n", body.dump());
```
achamayou, line 28: Ditto

src/node/rpc/member_frontend.h
```diff
@@ -1556,14 +1601,17 @@ namespace ccf
...
         }
 
         const auto digest = params.get<StateDigest>();
         if (ma->state_digest != digest.state_digest)
         {
           return make_error(
-            HTTP_STATUS_BAD_REQUEST, "Submitted state digest is not valid");
+            HTTP_STATUS_BAD_REQUEST,
+            ccf::errors::StateDigestMismatch,
+            "Submitted state digest is not valid");
```
achamayou, line 1614: Period/no period?

src/node/rpc/member_frontend.h
```diff
@@ -1855,7 +1914,9 @@ namespace ccf
         {
           LOG_FAIL_FMT("JWT key auto-refresh: no consensus available");
           return make_error(
-            HTTP_STATUS_INTERNAL_SERVER_ERROR, "no consensus available");
+            HTTP_STATUS_INTERNAL_SERVER_ERROR,
+            ccf::errors::InternalError,
+            "no consensus available");
```
achamayou, line 1919: Capitalisation?

achamayou (approved):

src/enclave/rpc_context.h
```diff
@@ -214,12 +215,26 @@ namespace enclave
...
+      http_status status, const std::string& code, std::string&& msg)
+    {
+      set_error({status, code, std::move(msg)});
+    }
+
+    virtual void set_error(ccf::ErrorDetails&& error)
+    {
+      nlohmann::json body = ccf::ODataErrorResponse{
+        ccf::ODataError{std::move(error.code), std::move(error.msg)}};
+      const auto s = fmt::format("{}", body.dump());
```
achamayou, line 228: Do we need the format then?

src/http/http_rpc_context.h
```diff
@@ -35,18 +36,27 @@ namespace http
...
 
-  static std::vector<uint8_t> error(size_t code, const std::string& msg)
+  inline std::vector<uint8_t> error(ccf::ErrorDetails&& error)
   {
-    http_status status = (http_status)code;
-    std::vector<uint8_t> data(msg.begin(), msg.end());
-    auto response = http::Response(status);
+    nlohmann::json body = ccf::ODataErrorResponse{
+      ccf::ODataError{std::move(error.code), std::move(error.msg)}};
+    const auto s = fmt::format("{}", body.dump());
```
achamayou, line 43: Ditto

src/http/ws_rpc_context.h
```diff
@@ -20,11 +21,21 @@ namespace ws
...
 
-  static std::vector<uint8_t> error(size_t code, const std::string& msg)
+  inline std::vector<uint8_t> error(ccf::ErrorDetails&& error)
   {
-    std::vector<uint8_t> ev(msg.begin(), msg.end());
-    return serialise(code, ev);
-  };
+    nlohmann::json body = ccf::ODataErrorResponse{
+      ccf::ODataError{std::move(error.code), std::move(error.msg)}};
+    const auto s = fmt::format("{}", body.dump());
```
achamayou, line 28: Ditto

src/enclave/rpc_context.h
letmaik: Absolutely not.

PR #1982, opened 2020-12-04
Fix SAN-caught errors

eddyashton:
- substr() second arg is length, not end
- Don't assign temp string to string_view


.daily_canary
```diff
@@ -1 +1,2 @@
 ##
+#
```
achamayou, line 2: I like to think that this is the beginning of a beautiful ascii-art canary.

achamayou (approved):

jumaffre (approved):
Thanks!

achamayou:
Full test suite seems to have run out of memory near the end (26/28), seed was 1607078119

PR #1984, opened 2020-12-04
Add memory stats in suite

achamayou:
Print memory usage on each live node at the end of each test in the suite.

image

Not useful to debug a bad_alloc this time, but could be in the future.

### DIFF
--- a/tests/e2e_suite.py
+++ b/tests/e2e_suite.py
@@ -26,8 +26,7 @@ def mem_stats(network):
     mem = {}
     for node in network.get_joined_nodes():
         try:
-            primary, _ = network.find_primary()
-            with primary.client() as c:
+            with node.client() as c:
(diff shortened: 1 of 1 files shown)


tests/e2e_suite.py
```diff
@@ -22,6 +22,19 @@ class TestStatus(Enum):
     skipped = 3
 
 
+def mem_stats(network):
+    mem = {}
+    for node in network.get_joined_nodes():
+        try:
+            primary, _ = network.find_primary()
+            with primary.client() as c:
```
eddyashton, line 30: Don't think we should always be contacting the primary?

```suggestion
            with node.client() as c:
```

achamayou: Indeed no, that got leftover from before.

eddyashton (approved):

PR #1985, opened 2020-12-04
Remove pbft notice

achamayou:
All the contents of this file except for a few squiggly brackets has been replaced (*), none of the code to which the original notice applied still exists.

(*) See file history in git, but mostly #1261, #1309 and #1633

This change simplifies the notice check, and removes the sole MIT-licensed file from the otherwise all-Apache 2.0 `src/`.

### DIFF
--- a/src/ds/stacktrace_utils.h
+++ b/src/ds/stacktrace_utils.h
@@ -1,7 +1,5 @@
-// Copyright (c) Microsoft Corporation.
-// Copyright (c) 1999 Miguel Castro, Barbara Liskov.
-// Copyright (c) 2000, 2001 Miguel Castro, Rodrigo Rodrigues, Barbara Liskov.
-// Licensed under the MIT license.
+// Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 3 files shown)


eddyashton (approved):

src/ds/stacktrace_utils.h
```diff
@@ -1,7 +1,5 @@
-// Copyright (c) Microsoft Corporation.
-// Copyright (c) 1999 Miguel Castro, Barbara Liskov.
-// Copyright (c) 2000, 2001 Miguel Castro, Rodrigo Rodrigues, Barbara Liskov.
-// Licensed under the MIT license.
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: Looks like `ccf_assert.h` needs updating too.

jumaffre (approved):

src/ds/stacktrace_utils.h
achamayou: Indeed, thank you.

PR #1988, opened 2020-12-07
Compile-time check of logger messages format

jumaffre:
We used to not verify the format strings passed to our `LOG_..._FMT()` logging macros at compile-time. For example, `LOG_FAIL_FMT("Error: CCF failure on node {}")` would compile OK but fail with an argument error exception at run-time (in this case the, the node ID is missing). I think this is bad as these error lines may only be written late in the life-cycle of a node, possibly just before dumping many more useful logs when crashing unexpectedly. 

We now use the [`FMT_STRING`](https://fmt.dev/latest/index.html#safety) macro to verify the format string at compile-time. Because this relies on `constexpr`, this means that the first argument to the logging macro must be known at compile-time (hence the few changes around our code base).

I think this is worth adding but happy to discuss this further (draft PR for now).


src/ds/logger.h
```diff
@@ -508,28 +508,38 @@ namespace logger
...
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
+
 #define LOG_TRACE \
   logger::config::ok(logger::TRACE) && \
     logger::Out() == logger::LogLine(logger::TRACE, __FILE__, __LINE__)
-#define LOG_TRACE_FMT(...) LOG_TRACE << fmt::format(__VA_ARGS__) << std::endl
+#define LOG_TRACE_FMT(s, ...) \
+  LOG_TRACE << fmt::format(FMT_STRING(s), ##__VA_ARGS__) << std::endl
```
jumaffre, line 518: We'll have to wait until C++20 to use `__VA_OPT__`.

src/consensus/aft/raft.h
```diff
@@ -1626,7 +1627,7 @@ namespace aft
       }
       catch (const std::logic_error& err)
       {
-        LOG_FAIL_FMT(err.what());
+        LOG_FAIL_FMT("Error recvo authenticated request vote: {}", err.what());
```
eddyashton, line 1630:
```suggestion
        LOG_FAIL_FMT("Error recv authenticated request vote: {}", err.what());
```

src/node/node_state.h
```diff
@@ -119,7 +119,8 @@ namespace ccf
     template <typename T>
     static Result<T> Fail(const char* s)
     {
-      LOG_DEBUG_FMT(s);
+      (void)s;
+      // LOG_DEBUG_FMT(s);
```
eddyashton, line 123:
```suggestion
      LOG_DEBUG_FMT("{}", s);
```

eddyashton (approved):
Nice! Just to note another benefit of this change - `LOG_INFO_FMT(s)` with a non-constexpr `s` is often unsafe, because we want to show the string verbatim but it'll throw an error when it contains format markers (eg `s = "I (ran [out <of {brackets}>])"`). So the correct way to do that is `LOG("{}", s)`, at which point you should probably add a helpful prefix, and this PR now fixes all of those and adds any new ones at compile time.

src/consensus/aft/raft.h
```diff
@@ -889,7 +889,7 @@ namespace aft
       }
       catch (const std::logic_error& err)
       {
-        LOG_FAIL_FMT(err.what());
+        LOG_FAIL_FMT("Error recv authenticated append entries: {}", err.what());
```
achamayou, line 892: These often end up saying "Error", which is redundant with "fail", and something that's essentially the name of the call. So I wonder if we shouldn't just use __PRETTY_FUNCTION__ instead and have a shortcut for that like LOG_FAIL_EXC(err).

achamayou:
@jumaffre that's an excellent change, thank you!

achamayou (approved):

src/consensus/aft/raft.h
jumaffre: Good point. I'll add this!

achamayou: Maybe “Exception in {}: {}”, passing the function name and what().

jumaffre: Done. I've also subsumed the existing `LOG_FAIL_FMT("Exception!"); LOG_DEBUG_FMT("Exception: {}", err.what());` pattern that we had into one macro.

PR #1989, opened 2020-12-08
New Merkle trees

wintersteiger:
This adds a new Merkle tree implementation, wired into History and Receipts.


wintersteiger:
All checks passed; I'm un-drafting this PR as it's technically ok to merge. More features will follow, but they can go into new PRs.

src/libmerklecpp/merklecpp.h
```diff
@@ -0,0 +1,1294 @@
...
+#  endif
+#  undef max
+#endif
+
+#ifdef WITH_TRACE
+#  include <iostream>
+#  define TRACE_HASH_SIZE 3
+#  define TRACE(X) \
+    { \
+      X; \
```
achamayou, line 36: It'd be better if this way defined to use the logging macros (LOG_TRACE_FMT here), since they know to use the ringbuffer when inside the enclave and there is no std::out. That perhaps belong in a separate file that's effectively CCF integration if we want to keep this independent/spin it off at some point, but I think it would still be a good thing.

src/libmerklecpp/merklecpp.h
```diff
@@ -0,0 +1,1294 @@
...
+      std::cout.flush(); \
+    };
+#else
+#  define TRACE(X)
+#endif
+namespace Merkle
```
achamayou, line 42: Our namespaces are all lower-case, I think we would prefer for this one to be too. This isn't enforced by the formatters, but it is something we stick to consistently.

wintersteiger: Sure, no problem.

src/libmerklecpp/test/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+if(EVERCRYPT)
+  add_subdirectory(evercrypt)
+endif()
+
+function(add_merkle_test NAME SRC)
+  add_executable(${NAME} ${SRC})
+  target_include_directories(${NAME} PRIVATE ${MERKLE++_DIR})
```
achamayou, line 9: MERKLECPP?

src/libmerklecpp/test/demo_tree.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <merklecpp.h>
+
+#define PRINT_HASH_SIZE 3
+
+int main()
+{
+  try
+  {
+    const size_t num_leaves = 6;
+    // for (size_t num_leaves = 1; num_leaves < 121; num_leaves++)
```
achamayou, line 23: TDB?

wintersteiger: To select how much and which ones you want :)

achamayou (approved):

src/libmerklecpp/merklecpp.h
```diff
@@ -0,0 +1,1294 @@
...
+
+    size_t size() const
+    {
+      return SIZE;
+    }
+
+    size_t byte_size() const
+    {
+      return SIZE;
+    }
```
jumaffre, line 116: Two functions returning the same value?

src/libmerklecpp/merklecpp.h
```diff
@@ -0,0 +1,1294 @@
...
+
+    std::string to_string(size_t num_bytes = SIZE) const
```
jumaffre, line 118: Should this be using `fmt::format` for consistency with other hash code, e.g. https://github.com/microsoft/CCF/blob/master/src/crypto/hash.h#L26-L40?

src/libmerklecpp/merklecpp.h
```diff
@@ -0,0 +1,1294 @@
...
+    }
+
+    bool operator!=(const HashT<SIZE>& other) const
+    {
+      return memcmp(bytes, other.bytes, SIZE) != 0;
+    }
+
+    void serialise(std::vector<uint8_t>& buffer) const
+    {
+      for (auto& b : bytes)
```
jumaffre, line 145: What if `buffer` is not already empty? Should this use `buffer.assign()`?

src/libmerklecpp/merklecpp.h
achamayou: That should be added in a separate file with the LOG_ macros that are specific to CCF. That way if/when this gets spun off into its own header-only library in a separate repo, it won't have a dependency on fmtlib just for this one thing.

src/libmerklecpp/merklecpp.h
```diff
@@ -0,0 +1,1294 @@
...
+      elements(elements)
+    {}
+    PathT(const PathT& other)
+    {
+      _leaf = other._leaf;
+      elements = other.elements;
+    }
+    PathT(PathT&& other)
+    {
+      _leaf = std::move(other._leaf);
```
jumaffre, line 209: Does `HashT` need a move constructor?

src/libmerklecpp/test/hashes.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+  merkle::Hash h4b54;
+  H(h4b,
+    h54,
+    h4b54); // 3D26DF03794B5E2781B5ACE81CFDC7A12F0C73C65F5AF44684FE81D6EFB3FAE9
+}
+
+int main()
+{
+  // check1();
```
jumaffre, line 67: Should this not be commented out?

jumaffre (approved):

src/libmerklecpp/merklecpp.h
wintersteiger: No, it appends to an existing buffer, e.g. see use in the other serialisation functions further down.

src/libmerklecpp/merklecpp.h
wintersteiger: I don't think so. HashT has no members that need to be moved, so it would be the same as the copy constructor. But I can add one if you think it's useful.

src/libmerklecpp/test/hashes.cpp
wintersteiger: Actually, this whole file can go. I just used that to compute some temporary hashes during debugging.

src/libmerklecpp/merklecpp.h
wintersteiger: Yeah, I usually use `byte_size` for the size in bytes of the serialised form, but in the case of hashes they are the same. Would it be more consistent with the rest of the codebase to make one return the other?

achamayou: Perhaps serialised_size() would be more explicit?

wintersteiger: Sure, sounds good!

PR #1994, opened 2020-12-09
Code update test rewrite

achamayou:
The code update test is one of the older end to end tests we have. Since it predates the test_suites, it doesn't compose nicely with other tests, and the pattern of code update it demonstrates is not how we expect a network to proceed.

This is closer to what we want, with some caveats:

1. Split out the bad code test
2. At f=1, swap acceptable code versions on join (*)
3. Spin up the same amount of nodes, running new code
~~4. Wait for catchup (**)~~
5. Retire old nodes one by one (***)
6. Confirm the network can still execute application transactions

Follow ups:

- Make the bad code test and the code update executable in any order. That requires a change to the code_id table, which instead of being ever growing, should only have valid entries. The replay mitigation afforded by the current scheme is no longer necessary with new proposal ids. #1996
- (*) Make the code swap a single proposal, which is how we expect it to be done, rather than two consecutive ones. #1997
~~- (**) The current wait is too strong, I think waiting for the last join to be globally committed is the right criterion instead.~~
- (***) We should do them all in one go. This is a bit awkward with the current proposal/test infra.

(**) create_and_trust_node does contain that wait anyway, so it's unnecessary to repeat it. It is still too strong for this particular use case.


tests/code_update.py
```diff
@@ -176,8 +149,8 @@ def run(args):
         network.start_and_join(args)
 
         test_verify_quotes(network, args)
+        test_add_node_with_bad_code(network, args)
         test_update_all_nodes(network, args)
-        test_verify_quotes(network, args)
```
jumaffre, line 180: I thought having this test at the end was useful, to make sure we can verify quotes after all nodes have been updated.

jumaffre (approved):

tests/code_update.py
achamayou: It's not clear to me what this is testing that the first one isn't: nodes are already added and we check their quotes. This is doing it on a different set of nodes, but the procedure to add them is identical and checks for quote exactly the same way.

I think this was meaningful in the fat genesis day, when we did not check the quotes of the original set of nodes at all. Now, not so much. Anyway, I've put it back, in case I've missed a reason why it's useful :)

eddyashton (approved):

PR #2005, opened 2020-12-10
Fix test_install.sh when testing a new CCF release

jumaffre:
When testing a new release on the CI, the CCF Python package won't be published until `test_install.sh` has passed. However, `test_install.sh`, which now uses `sandbox.sh`, assumed that `pip install ccf=<version>` works. To fix this, we can now pass a `PYTHON_PACKAGE_PATH` env var to both `test_install.sh` and `sandbox.sh` so that the Python package is installed from there instead.

### DIFF
--- a/src/host/main.cpp
+++ b/src/host/main.cpp
@@ -743,7 +743,7 @@ int main(int argc, char** argv)
       else
       {
         LOG_FAIL_FMT(
-          "No snapshot found. Node will request all historical transactions.");
+          "No snapshot found: Node will request all historical transactions");
       }
     }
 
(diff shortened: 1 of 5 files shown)


src/host/main.cpp
```diff
@@ -743,8 +743,7 @@ int main(int argc, char** argv)
       else
       {
         LOG_FAIL_FMT(
-          "No snapshot found. Node will request transactions all historical "
-          "transactions");
+          "No snapshot found. Node will request all historical transactions.");
```
eddyashton, line 746:
```suggestion
          "No snapshot found. Node will request all historical transactions");
```

As discussed this morning, I think we decided that our log lines don't end in periods? We could follow libfmt's lead and also try to enforce this at compile-time...

achamayou (approved):

Issue #1979, opened 2020-12-03
Don't return 500 on input validation errors for built-in endpoints


letmaik:
Currently built-in endpoints return 500 if the input (body/query parameters/..) doesn't match the schema.

Instead a 4xx response (likely 400) should be returned.

The ZAP fuzzer also complains about it. After fixing this, remove 100000 from `tests/zap.config`.

Notes:
- There's `JsonParseError` which is handled and treated as 400 in `frontend.h`, but this is only thrown by CCF for some specific custom conversions / checks. It's not what's thrown by nlohmann::json::parse generally and hence those other exception fall-back to the generic 500 handler.

PR #2006, opened 2020-12-10
Return 400 instead of 500 for user errors

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1979.

### DIFF
--- a/src/node/rpc/frontend.h
+++ b/src/node/rpc/frontend.h
@@ -543,6 +543,20 @@ namespace ccf
           update_metrics(ctx, metrics);
           return ctx->serialise_response();
         }
+        catch (const nlohmann::json::exception& e)
+        {
+          ctx->set_response_status(HTTP_STATUS_BAD_REQUEST);
+          ctx->set_response_body(e.what());
+          update_metrics(ctx, metrics);
+          return ctx->serialise_response();
+        }
+        catch (const UrlQueryParseError& e)
+        {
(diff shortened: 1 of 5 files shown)


achamayou (approved):

src/node/rpc/frontend.h
```diff
@@ -543,6 +543,20 @@ namespace ccf
           update_metrics(ctx, metrics);
           return ctx->serialise_response();
         }
+        catch (const nlohmann::json::exception& e)
+        {
+          ctx->set_response_status(HTTP_STATUS_BAD_REQUEST);
+          ctx->set_response_body(e.what());
```
eddyashton, line 549: I was going to suggest these bodies should all be prefixed so we can more easily tell what 'kind' of error happened. I think this is better achieved by setting the OData error code once this is integrated with #1980? So (assuming we merge that first), can we distinguish JSON parse errors from URL query errors from (...) with separate error codes?

letmaik: Yes, #1980 should be merged first, and then I'll adapt this PR. The problem is that it's not easy to figure out the different types of parse errors in CCF because it's all a bit scattered around and some errors are indistinguishable in terms of url param vs body parse errors.

achamayou: #1980 is in now.

letmaik: Done. I used `InvalidQueryParameterValue` for the `UrlQueryParseError` exception and `InvalidInput` for the rest.

Issue #1933, opened 2020-11-25
Don't expose internal /join endpoint in OpenAPI definition


letmaik:
Part of https://github.com/microsoft/CCF/issues/1918.

~~Should be a simple matter of adding `.set_auto_schema(JoinNetworkNodeToNode)`.~~ see below, issue title changed.

The OpenAPI definition should describe the public REST API, which implies that special endpoints which cannot be called directly (/join) should not be included.

It probably makes sense to remove the existing hack used for the JWT endpoint and replace it with something like `set_openapi_visible(false)` as new endpoint property.

achamayou:
/join is an internal RPC, one that’s called by a new node joining the network. There is no expectation that anything other than our code would ever make that call.

letmaik:
Then it shouldn't be published in the OpenAPI schema, like the new internal RPC for JWT key refresh.

achamayou:
Agreed.

PR #2008, opened 2020-12-10
Allow endpoints to be omitted from OpenAPI

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1933.

For C++, there's a new `set_openapi_hidden()` method.

For JS, there's a new `"openapi_hidden"` field in `app.json`.

The three internal endpoints /create, /join/ and /jwt_keys/refresh are now hidden.


achamayou:
Thank you! Could you perhaps add an entry in the CHANGELOG?

achamayou (approved):

letmaik:
> 
> 
> Thank you! Could you perhaps add an entry in the CHANGELOG?

Done

achamayou:
@letmaik 

The js batched stress test is failing:

```
51: Traceback (most recent call last):
51:   File "/mnt/build/2/s/tests/e2e_batched.py", line 138, in <module>
51:     run(args)
51:   File "/mnt/build/2/s/tests/e2e_batched.py", line 67, in run
51:     network = test(network, args, batch_size=1)
51:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 20, in wrapper
51:     return func(*args, **kwargs)
51:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 35, in wrapper
51:     check_reqs(network, args, *nargs, **kwargs)
51:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 67, in check
51:     raise TestRequirementsNotMet(f"Missing required methods: {concat}")
51: suite.test_requirements.TestRequirementsNotMet: Missing required methods: batch/submit, batch/fetch
```

letmaik:
@achamayou
Right, so the default in C++ is to include endpoints in OpenAPI except if `set_openapi_hidden(true)` is set, which is backwards-compatible. The default in JS now is to include OpenAPI only if `"openapi"` is set, which is different then the default before (always include an incomplete variant). We could retain the old behaviour for JS endpoints by leaving `"openapi"` alone and adding a new field `"openapi_hidden"` which would default to false.

achamayou:
@letmaik it seems to me that:

1. C++/JS endpoint behaviour should be consistent as much as possible, unless there's a good reason for it not be. I don't think I understand what the reason is here.
2. It is preferable to have insufficiently precise OpenAPI documentation than none at all.

I don't understand why we need to add a new field for this. It seems to me that we can just change the default value for that field to be "true" rather than "false"?

letmaik:
The reason for not introducing a new field was just to keep it simple, but if we want the incomplete definition by default then we can't do that, but instead we introduce, like c++, a new optional `openapi_hidden` field defaulting to false.

PR #2009, opened 2020-12-10
Clear member submitted shares on error

jumaffre:
Spotted by @eddyashton 

On recovery, when a threshold of member shares have been submitted and one (or more) of the recovery shares was bogus, the share combination fails and the endpoint returns an error. In this scenario, we want to clear all the submitted recovery shares from the `ccf.gov.submitted_shares` table to let members try again (and because it is hard to find out which share was bogus). 

Because the endpoint returned an error, the write set wasn't actually applied, which meant the submitted recovery shares weren't cleared. This is now the case, via `ctx->set_apply_writes(true)`.

### DIFF
--- a/src/node/rpc/test/member_voting_test.cpp
+++ b/src/node/rpc/test/member_voting_test.cpp
@@ -2092,12 +2092,27 @@ DOCTEST_TEST_CASE("Submit recovery shares")
 
       submitted_shares_count++;
 
+      auto tx = network.tables->create_tx();
+      auto submitted_shares = tx.get_view(network.submitted_shares);
       // Share submission should only complete when the recovery threshold
       // has been reached
       if (submitted_shares_count >= recovery_threshold)
       {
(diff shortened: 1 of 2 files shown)


achamayou (approved):

src/node/rpc/test/member_voting_test.cpp
```diff
@@ -2092,13 +2092,28 @@ DOCTEST_TEST_CASE("Submit recovery shares")
...
 
+      auto tx = network.tables->create_tx();
+      auto submitted_shares = tx.get_view(network.submitted_shares);
       // Share submission should only complete when the recovery threshold
       // has been reached
       if (submitted_shares_count >= recovery_threshold)
       {
         check_error(rep, HTTP_STATUS_INTERNAL_SERVER_ERROR);
+
+        // On error, all submitted shares should have been clearer
```
eddyashton, line 2103:
```suggestion
        // On error, all submitted shares should have been cleared
```

eddyashton (approved):

PR #2010, opened 2020-12-10
Support multiple auth policies, decouple auth from frontends

eddyashton:
Implementation of #1904.

The core goal is that instead of an ever-growing list of `set_FOO_required` properties on each endpoint, with a) a poorly understood matrix of interactions between them and b) a chain of checks+supporting code in `frontend.h`, we abstract the auth handling as a list of policies. If any policy passes, the endpoint executes, and can retrieve the identity established by that auth check. _The frontend does not know or use any of this identity information._ This means caller IDs and sigs are _not_ used by the frontend, instead it passes the caller cert and request around, and the processor makes authentication decisions consistently, after endpoint dispatch, using the current KV state. This has some knock-on effects that should be called out:

- The `member_client_signatures` and `user_client_signatures` are deleted, along with the code in `frontend.h` which populated them and associated unit tests. The framework doesn't prefer signed requests or privilege them. Instead, the endpoints which care about storing these for audit purposes (namely, the governance endpoints like `/propose` and `/vote` which require member-signed requests) store this themselves (in the previously-duplicated `governance_history` table)
- We always forward the original caller's cert when forwarding, and don't do auth/sig checks on the backups. I couldn't find a way to maintain this code and convince myself it was doing the right thing, and AFAICT we're no longer even running the perf test which showed the benefit of this. If we want to restore this, I think it makes sense to implement as a more generic execution sharding system (backups can execute some prefix of a transaction, and forward the results of this _along with a summary of the KV version it was executed on_ to the primary).
- Signatures or TLS handshake or JWT validity are only checked _if that's the policy we use_. The complications in `frontend.h` were largely along the separation of "if we have a sig, we should validate it", from "if we have a sig, we should store it", from "this endpoint is using the caller ID from here". For instance if you have a junk JWT, and an anonymous TLS connection, but a _good HTTP signature_, and policies saying "any of these is fine", then we will ignore the junk and accept the signed request, but we ensure that if the endpoint wants to use any identity it must use the one derived from the signature.

Remaining tasks for this PR:
- [x] Remove TODOs
- [x] Document these changes in `CHANGELOG.md`
- [x] Document in Sphinx
- [x] Drop existing flags completely, rather than deprecating. Currently the deprecation is misleading, because calling `set_client_signature_required` does _nothing_. I had a deprecation shim here ("if there are no policies, look at the flags and add the appropriate (per-frontend!) policies"), but it failed through a lot of our test cases and I opted for deleting it to make progress. I think we want to switch to "no auth is the default", and don't see a backwards-friendly way to do this. Thoughts welcome.
- [x] Investigate performance test differences
  - [x] logging WS does not auth at all
  - [x] SB does not auth at all
  - [x] SB sig should auth via signatures
- [x] Always set an explicit policy, even when it's empty_auth_policy, and log a warning when none was set. This will become an error in the future if we find a good way to communicate it (#2032).

Follow up items, out of scope for this PR: #2032


samples/apps/logging/logging.cpp
```diff
@@ -218,7 +225,120 @@ namespace loggingapp
...
+        }
+      };
+      make_endpoint("multi_auth", HTTP_GET, multi_auth)
+        .set_auto_schema<void, std::string>()
+        .add_authentication_policy(user_cert_auth_policy)
+        .add_authentication_policy(user_signature_auth_policy)
+        .add_authentication_policy(member_cert_auth_policy)
+        .add_authentication_policy(member_signature_auth_policy)
+        .add_authentication_policy(jwt_auth_policy)
+        .add_authentication_policy(empty_auth_plicy)
```
eddyashton, line 341: This (and the definition of the endpoint lambda above) are a good starting point - this is what we're trying to support. I think this is a sensible pattern for when there are multiple possible types of identity - try to get each in sequence, process the identity contained with it, potentially execute some common prefix/suffix.

Note that this uses `try_get_caller` (a friendly name for a `dynamic_cast`), since it expects to fail when another auth type was used. Most endpoints will have a single auth type and can safely call `get_caller` (or no auth, in which case they shouldn't call `get_caller` at all).

src/consensus/aft/request.h
```diff
@@ -13,19 +13,18 @@ namespace aft
...
   struct Request
   {
-    uint64_t caller_id;
     kv::TxHistory::RequestID rid;
     std::vector<uint8_t> caller_cert;
     std::vector<uint8_t> raw;
     uint8_t frame_format = enclave::FrameFormat::http;
 
-    MSGPACK_DEFINE(caller_id, rid, caller_cert, raw, frame_format);
+    MSGPACK_DEFINE(rid, caller_cert, raw, frame_format);
```
eddyashton, line 21: Knock-on effect - `caller_id` isn't guaranteed to exist, so isn't passed through `aft::Request` (and since nothing broke, nothing was actually using it). This ID (or another kind of identity) can (and should?) be derived indepdently, from the original request, on each executor.

src/enclave/rpc_context.h
```diff
@@ -121,35 +121,13 @@ namespace enclave
...
-
-      Forwarded(size_t client_session_id_, ccf::CallerId caller_id_) :
-        client_session_id(client_session_id_),
-        caller_id(caller_id_)
-      {}
-    };
-    std::optional<Forwarded> original_caller = std::nullopt;
-
-    // Constructor used for non-forwarded RPC
+    bool is_forwarded = false;
```
eddyashton, line 124: So I removed `caller_id` as mentioned in the description, and instead the cert is always present. And then `original_caller` was an optional struct containing only a `client_session_id`, but when this is filled the 'local' `client_session_id` above was invalid? So instead, we just reuse this same property for local or forwarded transactions. On the forwardee we do nothing with it other than returning it to the caller.

src/enclave/rpc_context.h
```diff
@@ -193,7 +172,6 @@ namespace enclave
       const std::string_view& name) = 0;
 
     virtual const std::vector<uint8_t>& get_serialised_request() = 0;
-    virtual std::optional<ccf::SignedReq> get_signed_request() = 0;
```
eddyashton, line 196: `SignedReq` structs can only be retrieved if we used that signature for auth, and have a signed identity. They're not stored on the request context, so this getter is goner.

src/http/authentication/authentication_types.h
```diff
@@ -0,0 +1,68 @@
...
+#pragma once
+
+#include "enclave/rpc_context.h"
+#include "kv/tx.h"
+
+#include <memory>
+
+namespace ccf
+{
+  struct AuthnIdentity
```
eddyashton, line 12: I didn't feel like wearing out my keyboard writing `Authentication` everywhere, but wanted to distinguish the ancient enemies AuthN and AuthZ. So this and similar types get a little `n`. But some variable names use the full `authentication`, because I hate consistency. Leaning towards replacing `authn` everywhere with `authentication`, unless anyone objects to the verbosity?

src/http/authentication/cert_auth.h
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    const OpenAPISecuritySchema& get_openapi_security_schema() const override
+    {
+      // TODO: There's no OpenAPI-compliant way to describe this cert auth?
+      return unauthenticated_schema;
+    }
+  };
+
+  struct MemberCertAuthnIdentity : public AuthnIdentity
```
eddyashton, line 74: Yes this is near identical to `UserCertAuthnIdentity`, and fully identical to `MemberSignatureAuthnIdentity`. I think we want to distinguish exactly where the identity came from, but maybe it turns out this doesn't matter if the results are identical? But I think the error-checking of distinguishing `user_` from `member_` (these are fundamentally different objects in different domains!) is useful.

src/http/http_sig.h
```diff
@@ -112,6 +112,7 @@ namespace http
 
     const auto signature = kp->sign(to_sign.value());
 
+    // TODO: Should insert correct keyId, but we don't have a cert here!
```
eddyashton, line 115: Our C++ signer produces signatures we probably shouldn't be accepting. But because they _also_ come with a TLS handshake identity, everything works. This is real bad for audit - the endpoint thinks it has (and stores) a signed request that doesn't let anyone verify the request. As this TODO notes, this is non-trivial to fix because we only have a key-pair at this point, not the specific _cert_ that we treat as the caller's identity. Perhaps an argument for using public keys as identities rather than certs? But we can plumb through the needed value for this instance.

src/perf_client/timing.h
```diff
@@ -163,25 +163,26 @@ namespace timing
     vector<PerRound> per_round;
   };
 
-  static CommitIDs parse_commit_ids(const RpcTlsClient::Response& response)
+  static std::optional<CommitIDs> parse_commit_ids(
```
eddyashton, line 166: This one is subtle. Because we previously always did some kind of auth check, even if we didn't use/require it, every request resulted in KV reads so got a TxID it could report in the headers (including the read-only stuff). But now our un-authed command endpoints (and all the common ones are now un-authed) will legitimately do _no interaction with the the KV at all_, even in the framework code. So they get no version headers, and some client code that expected these were always present needs to be updated.

src/http/http_sig.h
eddyashton: Tracking this in #2018, so I'll remove the TODO here.

src/http/authentication/authentication_types.h
achamayou: Fine with me, I think AuthN (as opposed to Authn) would have been quite acceptable too.

src/http/authentication/cert_auth.h
achamayou: Agreed.

src/node/rpc/common_endpoint_registry.h
```diff
@@ -96,46 +96,63 @@ namespace ccf
...
-      }
+        return make_success(out);
+      };
+      make_read_only_endpoint(
+        "user_id", HTTP_GET, json_read_only_adapter(user_id))
+        .set_auto_schema<GetUserId::In, GetUserId::Out>()
+        .add_authentication_policy(user_cert_auth_policy)
+        .add_authentication_policy(user_signature_auth_policy)
+        .add_authentication_policy(member_cert_auth_policy)
+        .add_authentication_policy(member_signature_auth_policy)
```
achamayou, line 154:
```suggestion
        .add_authn(member_signature_policy)
```

Feels a little less redundant. I think `member_sig_policy` could even be justified, seeing how the alternative is `cert`.

tests/infra/checker.py
```diff
@@ -30,7 +30,8 @@ def __call__(self, rpc_result, result=None, error=None, timeout=2):
                     result, rpc_result.body
                 )
 
-            assert rpc_result.seqno >= 0 and rpc_result.view >= 0, rpc_result
+            if rpc_result.seqno is not None or rpc_result.view is not None:
```
achamayou, line 33: Why?

eddyashton: Some requests now don't interact with the `Tx` at all, so don't get TxID headers:
https://github.com/microsoft/CCF/pull/2010#discussion_r540426983

letmaik:
> Listing these auth policies in OpenAPI. I've sketched the correct type of JSON objects, but not added them yet. A slight annoyance here is it looks like there's not a way to describe our 'cert auth' policy in OpenAPI? If anyone can see a better solution than "don't mention it in the OpenAPI", please let me know.

OpenAPI 3.1 will support `mutualTLS` as security scheme: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md#securitySchemeObject. For now the right thing to do is probably to leave it out completely.

src/http/authentication/authentication_types.h
```diff
@@ -0,0 +1,74 @@
...
+      kv::ReadOnlyTx& tx,
+      const std::shared_ptr<enclave::RpcContext>& ctx,
+      std::string& error_reason) = 0;
+
+    virtual void set_unauthenticated_error(
+      std::shared_ptr<enclave::RpcContext>& ctx, std::string&& error_reason)
+    {
+      ctx->set_error(
+        HTTP_STATUS_UNAUTHORIZED,
+        ccf::errors::AuthorizationFailed,
```
letmaik, line 35: Should probably be `InvalidAuthenticationInfo`.

src/http/authentication/jwt_auth.h
```diff
@@ -0,0 +1,89 @@
...
+        else if (!http::JwtVerifier::validate_token_signature(
+                   token.value(), token_key.value()))
+        {
+          error_reason = "JWT signature is invalid";
+        }
+        else
+        {
+          auto identity = std::make_unique<JwtAuthnIdentity>();
+          identity->key_issuer = key_issuer_view->get(key_id).value();
+          identity->header = token->header;
```
letmaik, line 55: `std::move` for header and body.

src/http/authentication/jwt_auth.h
```diff
@@ -0,0 +1,89 @@
...
+      return nullptr;
+    }
+
+    void set_unauthenticated_error(
+      std::shared_ptr<enclave::RpcContext>& ctx,
+      std::string&& error_reason) override
+    {
+      ctx->set_error(
+        HTTP_STATUS_UNAUTHORIZED,
+        ccf::errors::AuthorizationFailed,
```
letmaik, line 70: Should be `InvalidAuthenticationInfo`.

tests/e2e_logging.py
```diff
@@ -550,24 +625,25 @@ def run(args):
...
-        network = test_view_history(network, args)
-        network = test_primary(network, args)
-        network = test_metrics(network, args)
-        network = test_memory(network, args)
+        # network = test(
+        #     network,
+        #     args,
+        #     verify=args.package != "libjs_generic",
+        # )
+        # network = test_illegal(network, args, verify=args.package != "libjs_generic")
```
letmaik, line 633: Enable those again

src/node/rpc/common_endpoint_registry.h
letmaik: I like the long one a bit more.

src/http/authentication/sig_auth.h
```diff
@@ -0,0 +1,210 @@
...
+      return nullptr;
+    }
+
+    void set_unauthenticated_error(
+      std::shared_ptr<enclave::RpcContext>& ctx,
+      std::string&& error_reason) override
+    {
+      ctx->set_error(
+        HTTP_STATUS_UNAUTHORIZED,
+        ccf::errors::AuthorizationFailed,
```
letmaik, line 95: Should be `InvalidAuthenticationInfo`.

src/http/authentication/sig_auth.h
```diff
@@ -0,0 +1,210 @@
...
```
letmaik, line 189: Should be `InvalidAuthenticationInfo`.

src/node/rpc/frontend.h
```diff
@@ -169,54 +144,6 @@ namespace ccf
...
-            std::make_pair(caller_id, tls::make_verifier(caller)));
-        }
-        verifier = verifiers[caller_id];
-      }
-
-      return verifier->verify(
-        signed_request.req, signed_request.sig, signed_request.md);
-    }
-
     void set_response_unauthorized(
```
letmaik, line 147: Dead code?

src/node/rpc/frontend.h
```diff
@@ -292,136 +218,40 @@ namespace ccf
...
         }
-        if (!error_reason.empty())
+
+        if (identity == nullptr)
         {
-          set_response_unauthorized_jwt(
-            ctx, fmt::format("'{}' {}", ctx->get_method(), error_reason));
+          // If none were accepted, let the first set an error
+          endpoint->authn_policies[0]->set_unauthenticated_error(
+            ctx, std::move(auth_error_reason));
```
letmaik, line 240: This will use the error message of the last policy and pass it to the first policy. Needs extra logic to prevent filling the error message after the first policy.

src/node/rpc/common_endpoint_registry.h
achamayou: Can we at least not repeat policy?

tests/infra/checker.py
achamayou: Sure, but then they should break here, checking they have committed is meaningless.

src/node/rpc/frontend.h
achamayou: Yes

src/node/rpc/frontend.h
achamayou: Or just use the last policy :)

src/http/authentication/authentication_types.h
achamayou: Done

src/http/authentication/jwt_auth.h
achamayou: Done

src/http/authentication/jwt_auth.h
achamayou: Done

src/http/authentication/sig_auth.h
achamayou: Done

tests/e2e_logging.py
achamayou: Done

src/http/authentication/sig_auth.h
achamayou: Done

src/node/rpc/frontend.h
achamayou: Yep

src/node/rpc/frontend.h
achamayou: Made it the last.

src/node/rpc/endpoint_registry.h
```diff
@@ -351,28 +392,43 @@ namespace ccf
...
       }
 
+      /** Add an authentication policy which will be checked before executing
+       * this endpoint.
+       *
+       * When multiple policies are specified, any single successful check is
+       * sufficient to grant access, even if others fail. If all policies fail,
+       * the first can set an error status on the response, and the endpoint
+       * will not be invoked. If no policies are specified, then by default the
+       * endpoint accepts any request without an authentication check.
```
jumaffre, line 416: I guess this makes sense for new apps to default to no authentication but for existing apps that rely on the current default user cert check, if they don't read the release notes carefully, upgrading CCF will silently open all their app endpoints to unauthenticated clients. 

Should we default to user cert check for the next release at least?

achamayou: That's a good question, I'd be tempted to say yes, but on the other hand it looks like the popular default is really going to be jwt, not user certs...

tests/e2e_logging.py
```diff
@@ -197,6 +200,80 @@ def test_anonymous_caller(network, args):
     return network
 
 
+@reqs.description("Use multiple auth types on the same endpoint")
+@reqs.supports_methods("multi_auth")
+def test_multi_auth(network, args):
+    if args.package == "liblogging":
```
jumaffre, line 206: Minor: perhaps log a warning if it's not the case.

achamayou: I don't know why we have this, it seems to me that the req at 204 should be enough?

tests/infra/rates.py
```diff
@@ -62,7 +62,7 @@ def process_next(self):
 
     def get_metrics(self):
         with self.primary.client() as client:
-            rv = client.get("/node/metrics")
+            rv = client.get("/app/metrics")
```
jumaffre, line 65: Why this change?

src/node/rpc/endpoint_registry.h
achamayou: After discussion with @jumaffre, I am going to make the following changes:

1. Add validation in install() to check at least one authentication policy is present, throw otherwise. We have not found a good way to do this at compile time, but if there is one it can always be done later.
2. Add an explicit empty_auth policy to all anonymous endpoints.

Apart from this being what seems right now, it's also a good starting point: if we relax it, user code won't break.

tests/infra/rates.py
achamayou: This was just wrong before, we want metrics for application calls!

tests/e2e_logging.py
achamayou: As discussed, we probably should enforce reqs in tests (as opposed to suites), instead of duplicating this logic.

src/node/rpc/endpoint_registry.h
achamayou: Throwing in install() is very awkward: it cause the enclave_create_node ECALL to fail, and there's essentially no way to tell the user what went wrong. I am going to make this a log warning for now, and defer to #2032 finding a better enforcement mechanism.

achamayou (approved):

jumaffre (approved):

PR #2011, opened 2020-12-10
Only store valid Code IDs

### LINKED ISSUE #1996: Code IDs should be reusable, and should not be kept forever
We currently store code_ids forever, and their status is either RETIRED or ACTIVE.

1. This prevents adding code_ids more than once, which may be necessary during operations
2. This causes the code_ids table to grow indefinitely

Instead of marking code_ids as RETIRED, they should be removed. Instead of being marked ACTIVE, we should label them "ALLOW_TO_JOIN". We may decided to extend this if we check code_ids more frequently than at join time in the future.

achamayou:
Fix #1996

### DIFF
--- a/doc/audit/builtin_maps.rst
+++ b/doc/audit/builtin_maps.rst
@@ -1,0 +1,29 @@
+Built-in Maps
+=============
+
+`public:ccf.gov.nodes.code_ids`
+-------------------------------
+
+This table contains all the versions of the code allowed to join the current network.
+
+Key
+~~~
+
+base64 string representation of MRENCLAVE
+
+Value
+~~~~~
+
+.. doxygenenum:: ccf::CodeStatus
+   :project: CCF
+
+Example
+~~~~~~~
+
+.. list-table::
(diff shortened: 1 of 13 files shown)


doc/audit/builtin_maps.rst
```diff
@@ -0,0 +1,29 @@
...
+Example
+~~~~~~~
+
+.. list-table::
+   :header-rows: 1
+
+   * - Code ID
+     - Status
+   * - `cae46d1...bb908b64e`
+     - `ALLOWED_TO_JOIN`
```
jumaffre, line 29: What does this look like?

jumaffre (approved):

doc/audit/builtin_maps.rst
achamayou: image

achamayou: It's not necessarily the final word on how we want to document the KV, but it's a start.

PR #2015, opened 2020-12-11
Sandbox improvements

jumaffre:
As pointed by @letmaik, the recent changes to `sandbox.sh` meant that if `--js-app-bundle` was set but `--package` wasn't, the logging JS app would be run anyway. This PR fixes it, and adds some basic support for equal-separated arguments.


tests/sandbox/sandbox.sh
```diff
@@ -11,19 +11,30 @@ VERSION_FILE="${PATH_HERE}"/../share/VERSION
...
+            ;;
+        --js-app-bundle)
+            is_js_bundle_specified=true
+            shift
+            ;;
+        --js-app-bundle=*)
+            is_js_bundle_specified=true
+            ;;
+        *)
+            break
```
letmaik, line 34: This will break the loop which is probably not what you want.

tests/sandbox/sandbox.sh
```diff
@@ -11,19 +11,30 @@ VERSION_FILE="${PATH_HERE}"/../share/VERSION
...
 
 is_package_specified=false
-for item in "$@" ; do
-    if [ "$item" == "-p" ] || [ "$item" == "--package" ]; then
-        is_package_specified=true
-    fi
-done
+is_js_bundle_specified=false
 
 extra_args=("$@")
```
eddyashton, line 16: So we don't hit this issue again every time we try to modify these args, I think we want all the args produced by this script to be overridable (ie - to happen _before_ user-supplied args). So the final command should be
```
exec python
  --binary-dir ...
  ...
  "${extra_args[@]}"
  "$@"
```
rather than combining the user-supplied args within `extra_args`.

tests/sandbox/sandbox.sh
jumaffre: Thanks!

achamayou (approved):

Issue #1942, opened 2020-11-26
Respect Accept: header for content negotiation in built-in endpoints


letmaik:
Part of https://github.com/microsoft/CCF/issues/1918.

https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#7101-clients-specified-response-format

> In HTTP, response format SHOULD be requested by the client using the Accept header. This is a hint, and the server MAY ignore it if it chooses to, even if this isn't typical of well-behaved servers. Clients MAY send multiple Accept headers and the service MAY choose one of them.
The default response format (no Accept header provided) SHOULD be application/json, and all services MUST support application/json.

Currently, CCF matches the response content type based on what the request content type is. For `GET` requests without a body, this means that including `content-type` is currently required to switch between JSON and msgpack which is suboptimal.

The default should always be JSON, except when a different preference is given through `Accept`.

PR #2016, opened 2020-12-11
Allow content negotiation via Accept header

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1942.

After reading the HTTP spec again it's clear that the existing behaviour (using the request payload format and defaulting to json if no body and `content-type` is sent) is fine as-is:
> Selection is based on the available representations of the response (the dimensions over which it can vary; e.g. language, content-coding, etc.) and the contents of particular header fields in the request message or on other information pertaining to the request (such as the network address of the client). 

It still makes sense to support `Accept`-driven content negotiation for two reasons:
- A `GET` endpoint wouldn't have a request payload, so specifying `content-type` in the request is odd (and likely wrong) and instead `accept` should be used for content negotiation.
- A client may actually want a different request vs response format.

Changes:
- If the `accept` header is given, then it gets priority in determining the response format, otherwise the behaviour is as before.
- For unsupported request `content-type` values, `415 Unsupported Media Type` instead of `400` is returned.
- For unsupported request `accept` values: `406 Not Acceptable` is returned.
- `RpcException` is now a (useful) thing. It contains an `ErrorDetails` object and can be used instead of having to create more exception classes (and catch clauses) in cases where no further handling is required.
- Removed newline after JSON responses and hence use of `fmt::format`.

Final note: The Microsoft REST API guidelines say that JSON should be the default format if no `accept` header is provided. This PR doesn't fix that, msgpack will be returned if msgpack was sent in the request body. I think this is ok though since one has to opt-in to msgpack explicitly.

### DIFF
--- a/src/node/rpc/json_handler.h
+++ b/src/node/rpc/json_handler.h
@@ -64,6 +64,12 @@ namespace ccf
    * });
    */
 
+  class UrlQueryParseError : public std::invalid_argument
+  {
+  public:
+    using std::invalid_argument::invalid_argument;
+  };
+
   namespace jsonhandler
   {
     using JsonAdapterResponse = std::variant<ErrorDetails, nlohmann::json>;
@@ -185,7 +191,7 @@ namespace ccf
         const auto field_split = this_entry.find('=');
         if (field_split == std::string::npos)
         {
-          throw std::runtime_error(
+          throw UrlQueryParseError(
             fmt::format("No k=v in URL query fragment: {}", query));
         }
 
@@ -197,7 +203,7 @@ namespace ccf
         }
         catch (const std::exception& e)
         {
-          throw std::runtime_error(fmt::format(
+          throw UrlQueryParseError(fmt::format(
             "Unable to parse URL query value: {} ({})", query, e.what()));
         }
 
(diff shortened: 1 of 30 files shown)


src/node/rpc/json_handler.h
```diff
@@ -128,6 +132,46 @@ namespace ccf
...
+          packing = request_pack;
+        }
+        else
+        {
+          throw RpcException(
+            HTTP_STATUS_NOT_ACCEPTABLE,
+            ccf::errors::UnsupportedContentType,
+            fmt::format(
+              "Unsupported content type {} in accept header. Only {} and {} "
+              "are currently "
```
achamayou, line 164: That's a weird line break, but fine.

achamayou (approved):

achamayou:
Thank you for looking into this and fixing it.

src/node/rpc/json_handler.h
letmaik: Yeah, clang-format did that. Looks like it can introduce line breaks to satisfy the rules, but it doesn't re-format the whole string. I fixed it.

PR #2023, opened 2020-12-14
Doc versions nice-ness

jumaffre:
Some documentation nice-ness to conclude 2020:

**Before**

**After** (the selected item always matches the one of the current branch)


doc/_templates/sidebar-nav-bs.html
```diff
@@ -30,19 +55,14 @@
         </ul>
 
         {% if versions %}
-        <h5> Versions </h5>
-        <ul>
+        <h6> Versions </h6>
+        <select name="versions" id="versions" style="background:#F8F9FA;width:50%; border-radius:6px;border:1px solid #e5e5e5;font-family: inherit;font-size: inherit;padding: 0.4em 0.4em 0.4em 0.4em" onchange="onVersionChange()">
```
achamayou, line 59: That's... quite specific :) Are some generic bootstrap classes we can use instead?

Just asking, if the answer is no, that's fine too of course.

achamayou:
That's very nice! I did try to do that with HTML/CSS only once, but utterly failed.

Is it "sticky" when you navigate? Does it initialise correctly if you set/send the URL?

achamayou (approved):

doc/_templates/sidebar-nav-bs.html
jumaffre: Very specific! I'll have a look but I'm sure I'll find something more generic ;)

jumaffre: Done.

jumaffre:
> That's very nice! I did try to do that with HTML/CSS only once, but utterly failed.
> 
> Is it "sticky" when you navigate? Does it initialise correctly if you set/send the URL?

Yes, it it sticky on navigate and updates correctly from a URL as the selected option depends on the URL itself (defaults to `master` only if `ccf-X` is not found in the URL).

Issue #2013, opened 2020-12-11
ccf-dev.yml playbook .deb install link is broken


achamayou:
https://github.com/microsoft/CCF/releases/latest/download/ccf_latest_amd64.deb is redirecting to: https://github.com/microsoft/CCF/releases/download/ccf-0.16.0/ccf_latest_amd64.deb and not https://github.com/microsoft/CCF/releases/download/ccf-0.16.0/ccf_0.16.0_amd64.deb

The 'latest' link works so long as the asset names are fixed/don't contain a version. That was true for 'ccf.tar.gz' but not for .deb.

The playbook needs to query github to work out what the latest version is instead, for example with:
```
$ curl -s https://api.github.com/repos/microsoft/ccf/releases/latest | egrep 'https://.*\.deb' | cut -d\" -f4
https://github.com/microsoft/CCF/releases/download/ccf-0.16.0/ccf_0.16.0_amd64.deb
```
or
```
$ curl -s https://api.github.com/repos/microsoft/ccf/releases/latest | jq -r '.assets | .[]  | select(.browser_download_url | endswith(".deb")) | .browser_download_url'
https://github.com/microsoft/CCF/releases/download/ccf-0.16.0/ccf_0.16.0_amd64.deb
```

PR #2028, opened 2020-12-15
Query latest deb url from github

letmaik:
Fixes https://github.com/microsoft/CCF/issues/2013


getting_started/setup_vm/roles/ccf_install/tasks/deb_install.yml
```diff
@@ -1,9 +1,19 @@
 - name: Include vars
   include_vars: common.yml
 
+- name: Get package url
```
achamayou, line 4: I think you could have left the var for versioned ccf_ver case, and just added the curl gated with a `when: "{{ ccf_ver }} = "latest"`, trading off an extra line of YAML in return for keeping the bash a one-liner.

The end result is exactly the same of course.

achamayou (approved):

getting_started/setup_vm/roles/ccf_install/tasks/deb_install.yml
letmaik: Yeah I thought about that but I wanted the URL construction for both cases in a single place. Also you have the issue that in one case you need `.stdout`.

achamayou: Ah true, you'd have needed a separate `set_fact`, with the same `when` to assign the `.stdout` to the same name.

Issue #1986, opened 2020-12-04
Getting compacted keys from a later primary does not work


eddyashton:
When a node joins, it is given the list of compacted ledger secrets from the current primary. This lets it replay the ledger/incoming append entries, and while doing so it calls `encryptor->compact` at each signature transaction, and tries to compact its own list of ledger secrets. This call to `record_compacted_keys` simply appends the given secret to the same list we received from the primary, so we end up with duplicates. If the primary told us secrets at versions `[1, 32, 46, 85]`, we'll end up with `[1, 32, 46, 85, 32, 46, 85]`. If we then become primary, this is the list we'll send to new joiners, and the duplicates will cause them to loop forever.

This is the cause of [this test failure](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=16293&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=63d660c3-6f76-573a-0154-fd0bcb096ded&l=13261) - node 8 gets stuck in a permanent loop adding to a container, eventually timing out or hitting a `bad_alloc`. This has a high repro rate on this seed, but is not 100%. We should add a reconfiguration suite which exhibits this crash consistently.

There's several things we should or could fix here:
- When we get a list of secrets from the primary, we should make sure that the versions they apply from are unique and ordered.
- When we `record_compacted_keys`, we should check if this is a secret we already know (if so, it should match!), rather than apppending out-of-order.
- `TxEncryptor::compact` should be rewritten to be clearer, and to remove the possibility of looping forever.
- We should add some logging on rekeying/key compaction, so that verbose logs show us the lifetime of a secret.
- The 'active' `encryption_keys` and compacted `ledger_secrets` could be combined into a single, ordered container. We want to keep the lookup on `get_encryption_key` cheap, but we can do this by looking at the final item first (which is always the secret we want on the hot path), and only running the `std::upper_bound` search if that fails.
- We should investigate what we're doing with very old ledger secrets. With snapshotting, we don't want to keep them around forever, but this might mean declaring that only snapshots created in the last N secrets can be used for joining (for anything older, the primary doesn't remember how you should decrypt it, and finding out by trawling backwards through the ledger is too expensive).

eddyashton:
For a simple repro, add a rekey to the `reconfiguration` test_suite:
```
--- a/tests/suite/test_suite.py
+++ b/tests/suite/test_suite.py
[...]

PR #2029, opened 2020-12-15
Generate snapshots by default

### LINKED ISSUE #1956: Enable snapshotting by default
Snapshots are currently an opt-in feature (`cchost`'s `--snapshot-tx-interval` is set to `std::numeric_limits<std::size_t>::max()`). Now that all snapshotting features are complete, we should enable them by default and change the Python infra so that joining from snapshots are default when adding a node to the service.

Tasks:
- [ ] Change `cchost` default value for `--snapshot-tx-interval`.
- [ ] Python infra changes so that all nodes join from snapshot (unless specifically disabled).

jumaffre:
Resolves #1956 

Snapshots are now enabled by default, and generated by the current primary node every 10,000 transactions (why 10,000? It provides a good compromise between not-too-frequent snapshot generation but frequent enough for fast recovery/join procedures). 

The Python infra has been updated accordingly:
- Snapshots are generated every 10 transaction for end-to-end tests, and 10,000 transactions for performance tests
- Snapshot-specific end-to-end tests/suite have been removed
- Adding a new node to the service now defaults to using a snapshot, unless explicitly disabled


CHANGELOG.md
```diff
@@ -17,10 +17,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Error responses of built-in endpoints are now JSON and follow the OData schema (#1919).
 - Code ids are now deleted rather than marked as `RETIRED`. `ACTIVE` is replaced with the more precise `ALLOWED_TO_JOIN`. (#1996)
 - CCF now depends on [Open Enclave 0.13](https://github.com/openenclave/openenclave/releases/tag/v0.13.0-rc2).
+- Snapshots are generated by default on the current primary node, every `10,000` committed transaction (#2029).
 
 ## Fixed
 
-- Scenario perf client `--msg-ser-fmt text` option works with C++ apps.
+- Scenario perf client `--msg-ser-fmt text` option works with C++ apps (#2026).
```
achamayou, line 24: Please remove this from the changelog. The perfclient is strictly an internal tool, it is not released, and there is no plan to change that or to support it.

cmake/common.cmake
```diff
@@ -555,7 +555,8 @@ function(add_perf_test)
       ${PYTHON} ${PARSED_ARGS_PYTHON_SCRIPT} -b . -c ${PARSED_ARGS_CLIENT_BIN}
       ${CCF_NETWORK_TEST_ARGS} --consensus ${PARSED_ARGS_CONSENSUS} -g
       ${PARSED_ARGS_GOV_SCRIPT} --write-tx-times ${VERIFICATION_ARG} --label
-      ${LABEL_ARG} ${PARSED_ARGS_ADDITIONAL_ARGS} ${NODES}
+      ${LABEL_ARG} --snapshot-tx-interval 10000 ${PARSED_ARGS_ADDITIONAL_ARGS}
```
achamayou, line 558: If this is the default, why pass it explictly?

tests/infra/e2e_args.py
```diff
@@ -246,7 +246,8 @@ def cli_args(add=lambda x: None, parser=None, accept_unknown=False):
     parser.add_argument(
         "--snapshot-tx-interval",
         help="Number of transactions between two snapshots",
-        default=None,
+        type=int,
+        default=10,
```
achamayou, line 250: Why 10?

tests/reconfiguration.py
```diff
@@ -56,14 +60,25 @@ def test_add_node(network, args):
...
+        args,
+        target_node=backup,
+        snapshot_dir=snapshot_dir,
     )
     assert new_node
     return network
 
 
+# Note: this test cannot be included in the test suite yet as
+# add_from_snapshot() decorator makes use of historical queries
```
achamayou, line 81: Link to #1986?

achamayou (approved):

cmake/common.cmake
jumaffre: Because `e2e_args` defaults to `10` so that all e2e tests generate snapshots very often, so that new nodes join from snapshots wherever possible.

tests/infra/e2e_args.py
jumaffre: As explained above, we want snapshots to be generated frequently enough that nodes automatically join from snapshots whenever possible.

tests/reconfiguration.py
jumaffre: This is actually #1648 but yes!

tests/infra/e2e_args.py
achamayou: Got it.

tests/reconfiguration.py
achamayou: Indeed, definitely a good thing to link then :)

achamayou:
At least the log_scenario_SGX_CFT throughput seems to be rather substantially affected by this change. Are we snapshotting too often?

jumaffre:
> At least the log_scenario_SGX_CFT throughput seems to be rather substantially affected by this change. Are we snapshotting too often?

Well spotted. Investigating this now.

CMakeLists.txt
```diff
@@ -881,7 +843,7 @@ if(BUILD_TESTS)
       --max-writes-ahead
       1000
       --repetitions
-      800
+      5000
```
achamayou, line 846: This seems to time out, you will need to bump up the test timeout from the current 90s for this to pass.

tests/reconfiguration.py
```diff
@@ -56,14 +61,25 @@ def test_add_node(network, args):
...
+        args,
+        target_node=backup,
+        snapshot_dir=snapshot_dir,
     )
     assert new_node
     return network
 
 
+# Note: this test cannot be included in the full test suite yet as
+# add_from_snapshot() decorator makes use of historical queries
```
achamayou, line 82:
```suggestion
# add_from_snapshot() decorator makes use of historical queries (#1648)
```

CMakeLists.txt
jumaffre: Weird, I'd pushed a fix for this yesterday. This should run fine now.

PR #2031, opened 2020-12-16
Merkle trees: add OpenSSL and mbedTLS hash functions. Switch to full SHA256.

wintersteiger:
This adds the OpenSSL and mbedTLS full SHA256 and SHA256-compression functions to the Merkle trees and switches the history to full SHA256.


achamayou (approved):
This is great, it will make the receipts and the ledger much easier to audit for users!

jumaffre (approved):

PR #2038, opened 2020-12-18
Delete quickstart section and move sub-pages to other sections

jumaffre:
~The quickstart documentation had become orphan so re-add it and link it from documentation homepage.~

Quickstart pages have now been moved to other sections.


achamayou:
As discussed, I would prefer to shuffle this into the existing sections (Build Apps, Use Apps, Operations) rather than have it as a separate section.

.gitignore
```diff
@@ -1,5 +1,6 @@
 build*/
 _build/
+!build_apps/
```
eddyashton, line 3: I don't think this is necessary - it wasn't previously ignored?

doc/quickstart/build.rst
```diff
@@ -69,7 +69,7 @@ Although CCF's unit tests can be run through ``ctest`` directly, the end-to-end
 
     .. code-block:: bash
 
-        $ TEST_ENCLAVE=virtual ./tests.sh -VV
+        $ TEST_ENCLAVE=virtual ./tests.sh [-VV]
```
eddyashton, line 72: I think the bash code blocks are commands which could be pasted and run verbatim, so this shouldn't have the documenting-an-optional-arg brackets. If we think users don't want `-VV` all the time, we should just remove it (the comment above makes it clear that this is just a wrapper for `ctest`, so hopefully `--help` and `-VV` etc follow from there).

eddyashton (approved):

.gitignore
jumaffre: It is necessary as any file created under `build_apps/` will not be visible to git, because of the `build*/` exclusion rule two lines above (note that it doesn't affect existing files). I'm modifying this to `doc/build_apps/` to be more specific though.

jumaffre:
> As discussed, I would prefer to shuffle this into the existing sections (Build Apps, Use Apps, Operations) rather than have it as a separate section.

I was writing a blurb about how we could move/split the existing pages under `quickstart/` to the other sections but I've managed to convince myself that I prefer having a distinct quickstart section (maybe to be renamed to "Start Here"). It's easier for users of CCF, on first use, to install CCF first and then figure out the technical features than the other way around. There is also an overlap between the different roles and what the quickstart section currently contains (e.g. the Python package is applicable to all roles). Also, some pages in the quickstart section don't belong to any role (e.g. contributing to CCF) so we may as well have a new section for the quickstart than a specific one for contributors.

I also suggest we move the overview page under the new quickstart and move the other overview sub-pages under the relevant sections (e.g. "Constitution" page moves to "Governance" section).

achamayou (approved):

PR #2039, opened 2020-12-18
Faster end-to-end tests in the CI

jumaffre:
- Rotation test only runs 3 times in the CI (10 times on daily, with the new `-DLONG_TESTS=ON` cmake flag)
- Python logging app only checks for commit after submitting a batch of entries
- Only register 1 user in most end-to-end tests (now the default value)
- Simplify reconfiguration test suite


tests/rotation.py
```diff
@@ -28,19 +28,23 @@ def run(args):
     ) as network:
         network.start_and_join(args)
 
+        iterations = 10 if args.long_tests else 3
```
achamayou, line 31: If this is the only thing long_tests does, I think we may as well specify ROTATION_ITERATIONS directly? Just a thought.

achamayou (approved):

tests/rotation.py
jumaffre: I feel like we may want to keep this option generic enough to be used by other tests too? I'm thinking if we aim to keep the CI as quick as possible and only run longer tests to the Daily, this may be a pattern that we use more.

jumaffre:
Down from 
image
to 
image


The daily is failing with the well-known node shutdown issues so I will merge regardless.

PR #2044, opened 2021-01-04
Fix signature auth on read-only member endpoints

achamayou:
~~The change itself should be fine, but the hack applied to the test infra clearly isn't. Trying to work out how to do it nicely, suggestions welcome and appreciated.~~

Fix + tests.

1. Introduce a --disable-member-session-auth in infra, which turns off TLS session authentication for all consortium/member RPCs. governance_test is now run:
  a. The same as before, with session auth and signatures on writes
  b. As in the TPAL deployment scenario, where keys are kept in HSM, with no session auth and signatures only
2. CCFClient has changed to support this, and now takes two identities, a session identity and a signing identity, which can be distinct. The presence of the signing identity controls signing, there is no use case now where we decide to sign per-RPC, and I'm suspicious of the default to False, so I've removed that.
3. scurl.sh has also changed in the same way, this was a little fiddly, reviews especially welcome.

Reported by @lynshi


lynshi:
As we've seen so far, the tests have missed one or the other kind of authentication. I think it would be best to do separate tests for the different authentication mechanisms (i.e. all signed requests vs all TLS auth + signing when necessary). This would also help with testing JWT authentication, when it is available, completely on all the endpoints. Maybe the tests can take a callable that can wrap the client calls to inject the authentication mechanism?

achamayou:
@lynshi the problem is that there is a substantial amount of infra code for governance orchestration that creates its own clients and executed sequences of requests on its own, and is called by the tests as high level operations.

Code that tests application logic is much shallower and easier to parametrise, and JWT auth can be tacked on to messages without changing much to the client.

src/node/rpc/endpoint_registry.h
```diff
@@ -604,23 +604,6 @@ namespace ccf
      */
     void install(Endpoint& endpoint)
     {
-      if (endpoint.authn_policies.empty())
```
achamayou, line 607: The rest of the fix was mostly duplicated in #2048, but this was forgotten there, so scrubbing it here.

eddyashton: Thanks, was just about to open another PR for this one.

python/ccf/clients.py
```diff
@@ -605,13 +592,12 @@ def _call(
...
         http_verb: str = "POST",
         headers: Optional[dict] = None,
-        signed: bool = False,
         timeout: int = DEFAULT_REQUEST_TIMEOUT_SEC,
         log_capture: Optional[list] = None,
     ) -> Response:
         description = ""
         if self.description:
-            description = f"{self.description}{signed * 's'}"
+            description = f"{self.description}{bool(self.sign) * 's'}"
```
eddyashton, line 600:
```suggestion
            description = f"{self.description}{int(self.sign) * 's'}"
```

`self.sign` is already boolified on assignment, so we don't need to recast it here (`self.sign * 's'`) should work too. But I think casting to int is more readable.

src/node/rpc/endpoint_registry.h
```diff
@@ -604,23 +604,6 @@ namespace ccf
...
-      }
-
-      // A single empty auth policy is semantically equivalent to no policy, but
-      // no policy is faster
-      if (
-        endpoint.authn_policies.size() == 1 &&
-        endpoint.authn_policies.back() == empty_auth_policy)
-      {
-        endpoint.authn_policies.pop_back();
-      }
```
eddyashton, line 622: I think this section should be kept?

python/ccf/clients.py
```diff
@@ -572,28 +563,24 @@ def __init__(
...
         connection_timeout: int = DEFAULT_CONNECTION_TIMEOUT_SEC,
         description: Optional[str] = None,
         ws: bool = False,
-        disable_client_auth: bool = False,
     ):
         self.connection_timeout = connection_timeout
         self.description = description
         self.name = f"[{host}:{port}]"
         self.is_connected = False
+        self.sign = bool(signing_auth)
```
eddyashton, line 576: For my own understanding - this is equivalent to `signing_auth is not None`?

python/ccf/clients.py
achamayou: Yes, good point, this is a leftover from before sign was there. I think this can be simplified further actually.

src/node/rpc/endpoint_registry.h
achamayou: I thought empty_auth was just an empty vector now?

python/ccf/clients.py
achamayou: No, that's sign is falsy, so False, "", [], {} etc would also yield False. Anyway, this is removed now.

src/node/rpc/endpoint_registry.h
achamayou: Ah, no!

eddyashton: `no_auth_required` _is_ an empty vector, but if we end up with the vector `{empty_auth_policy}` through some other way, we can optimise it here?

python/ccf/clients.py
```diff
@@ -572,28 +563,32 @@ def __init__(
...
-        self.description = description
         self.name = f"[{host}:{port}]"
+        self.description = description
+        if self.description is not None:
+            if session_auth is not None:
+                self.description += "a"
+            if signing_auth is not None:
+                self.description += "s"
+        else:
+            self.description = self.name
```
eddyashton, line 581: This produces prefixes like:

```
[0|member0]as
```

So the `a` is just repeating the fact that `member0` is there. I think this should be rolled into the construction of the initial description in `node.py`. If we have different auth and signing identities, they should both be present. Something like:

```
[0|member1] # I'm talking to node 0 as member1
[0|member1s] # I'm talking to node 0 as member1, and signing everything
[0|member1|member2s] # I'm talking to node 0 as member1, and signing everything as member2
```

eddyashton:
This change adds functionality to the Python CCFClient to use different identities for TLS and signatures, but afaict doesn't add tests for this. I think we should add a case to the `multi_auth` test in `e2e_logging.py` which checks the request is submitted successfully when these identities are both present but distinct.

python/tutorial.py
```diff
@@ -99,11 +99,16 @@
 # >>> proposal
 # {'script': {'text': 'return Calls:call("open_network")'}}
 
-member_client = ccf.clients.CCFClient(host, port, ca, member_cert, member_key)
+member_client = ccf.clients.CCFClient(
```
jumaffre, line 102: Does this mean that once a client has been created, all requests sent will be signed or not, depending on the value passed to the `signing_auth` argument here? This seems a little restricting, especially that these are now set at the granularity of a handler. I prefer the current way where `signed` is set per request. Of course, we can still set a `signing_auth` in the constructor of `CCFClient` so that `signing_auth` can be different from `session_auth`.

python/ccf/clients.py
achamayou: Yeah good idea.

achamayou:
@eddyashton this is tested by governance_no_session_auth_test

tests/infra/e2e_args.py
```diff
@@ -255,7 +255,12 @@ def cli_args(add=lambda x: None, parser=None, accept_unknown=False):
         default=None,
     )
     parser.add_argument(
-        "--long-tests", help="Enable extended tests", action="store_true", default=False
+        "--long-tests", help="Enable extended tests", action="store_true"
```
jumaffre, line 258: Does this default to `False`? It might be worth leaving it the default value to be more explicit here?

achamayou: That's always the case for store_true

eddyashton:
@achamayou If I'm reading correctly, `--disable-member-session-auth` means we test `(None, member0)` requests. And through unsigned and anonymous requests we tests every other combination of `None` and `member0` for TLS and sig. But we don't test `(member0, member1)` auth, so we won't catch anything that explodes when they're both present but distinct.

python/tutorial.py
achamayou: Yes, I mentioned this in the description. I do not believe this is restricting, it's never used at the moment.

achamayou:
@eddyashton True, but that's a use case we have now?

eddyashton:
@achamayou I agree its not a targeted use case or something we need to document, but it is something we now support (and this API makes it _look_ like a deliberately supported option), and its very likely we or a user of the library are going to do it sooner rather than later.

python/ccf/clients.py
achamayou: I'm going with positions instead, the s suffix inside the bracket could be part of the name, I don't think it's a very happy choice.

eddyashton (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - To avoid accidentally unauthenticated endpoints, a vector of authentication policies must now be specified at construction (as a new argument to `make_endpoint`) rather than by calling `add_authentication`. The value `ccf::no_auth_required` must be used to explicitly indicate an unauthenticated endpoint.
+- All `/gov` endpoints accept signature authentication alone correctly, regardless of session authentication.
+- `ccf.CCFClient` now allows separate `session_auth` and `signing_auth` to be passed as construction time. `ccf.CCFClient.call()` now longer takes a `signed` argument, clients with a `signing_auth` always sign. Similarly, the `disable_session_auth` constructor argument is removed, the same effect can be achieved by setting `session_auth` to `None`.
```
eddyashton, line 14:
```suggestion
- `ccf.CCFClient` now allows separate `session_auth` and `signing_auth` to be passed as construction time. `ccf.CCFClient.call()` no longer takes a `signed` argument, clients with a `signing_auth` always sign. Similarly, the `disable_session_auth` constructor argument is removed, the same effect can be achieved by setting `session_auth` to `None`.
```

s/now/no

achamayou: Argh, thank you.

Issue #2042, opened 2021-01-04
forum_app test failure


achamayou:
In https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=17040&view=logs&j=ee9484ec-b476-577b-aa0e-d5ddfcbcd149&t=e2c1a4cf-16d0-5038-f9f6-343f9806d38b

```
66: TypeError [ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING]: A dynamic import callback was not specified.
66:     at exports.importModuleDynamicallyCallback (internal/process/esm_loader.js:34:9)
66:     at formattedImport (/mnt/build/1/s/samples/apps/forum/node_modules/mocha/lib/esm-utils.js:6:23)
66:     at Object.exports.requireOrImport (/mnt/build/1/s/samples/apps/forum/node_modules/mocha/lib/esm-utils.js:23:14)
66:     at Object.exports.loadFilesAsync (/mnt/build/1/s/samples/apps/forum/node_modules/mocha/lib/esm-utils.js:33:34)
66:     at Mocha.loadFilesAsync (/mnt/build/1/s/samples/apps/forum/node_modules/mocha/lib/mocha.js:431:19)
66:     at singleRun (/mnt/build/1/s/samples/apps/forum/node_modules/mocha/lib/cli/run-helpers.js:125:15)
66:     at exports.runMocha (/mnt/build/1/s/samples/apps/forum/node_modules/mocha/lib/cli/run-helpers.js:190:10)
66:     at Object.exports.handler (/mnt/build/1/s/samples/apps/forum/node_modules/mocha/lib/cli/run.js:362:11)
66:     at /mnt/build/1/s/samples/apps/forum/node_modules/mocha/node_modules/yargs/lib/command.js:241:49
66:     at process.runNextTicks [as _tickCallback] (internal/process/task_queues.js:58:5)
66:     at /mnt/build/1/s/samples/apps/forum/node_modules/esm/esm.js:1:34535
66:     at /mnt/build/1/s/samples/apps/forum/node_modules/esm/esm.js:1:34176
66:     at process.<anonymous> (/mnt/build/1/s/samples/apps/forum/node_modules/esm/esm.js:1:34506)
66:     at Function.<anonymous> (/mnt/build/1/s/samples/apps/forum/node_modules/esm/esm.js:1:296856)
[...]

PR #2045, opened 2021-01-05
Fix forum app test

letmaik:
Fixes https://github.com/microsoft/CCF/issues/2042.

This works around an issue related to the experimental ESM support used with mocha. By not importing the (runtime) constant from the ccf app code it doesn't try to import the module and its dependencies which caused the issue (NOTE: the error message in the issue is not the real error because that's swallowed by mocha, see also mocha/lib/esm-utils.js::requireOrImport).

### DIFF
--- a/samples/apps/forum/test/poll.test.ts
+++ b/samples/apps/forum/test/poll.test.ts
@@ -18,9 +18,10 @@ import {
   SubmitOpinionsRequest,
   NumericPollResponse,
   StringPollResponse,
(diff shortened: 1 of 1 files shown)


achamayou:
"If a dependency fails to import in the forest, but nobody uses it, does it produce an error?" said Master @letmaik.

And upon these words, the novice was enlightened.

achamayou (approved):

eddyashton (approved):

PR #2046, opened 2021-01-05
Pin Python package dependencies more weakly

letmaik:
Currently, all `ccf` Python package dependencies are pinned to exact versions. This is fine if ccf is used via CLI only in a fresh venv. However, `ccf` is also a library that can end up in a bigger dependency tree and would cause version conflict errors if another sibling package requires a different version, e.g. an application that uses ccf may define a `requirements.txt` file like:

```
ccf==0.16.2
requests==2.25.1
```
which would fail to install as ccf has `requests` currently pinned to 2.24.0.

This PR weakens the version constraints and takes some good guesses to prevent accidental breaks while still allowing flexibility.


python/requirements.txt
```diff
@@ -1,6 +1,6 @@
-msgpack==1.0.0
-loguru==0.5.3
-requests==2.24.0
-requests-http-signature==0.2.0
-websocket-client==0.57.0
-cryptography==3.2
+msgpack==1.*
+loguru
+requests==2.*
```
achamayou, line 3: I think this is a little too relaxed, we haven't tested with old requests versions, and it's not impossible that they wouldn't work, particularly with signatures.

python/requirements.txt
```diff
@@ -1,6 +1,6 @@
...
+requests-http-signature==0.2.*
+websocket-client==0.*
```
achamayou, line 5: Likewise, this is too loose.

python/requirements.txt
```diff
@@ -1,6 +1,6 @@
...
+cryptography==3.*
```
achamayou, line 6: Same here.

letmaik:
> 
> 
> @letmaik I wholeheartedly agree with the spirit of this change, I think it's just a little too lax in some cases.

Do you have better version constraints in mind? I would leave the granularity as-is but additionally specify lower bounds.

letmaik:
> 
> 
> @letmaik exactly what you say, switch to >=

I had `requests >= 2.24, == 2.*` in mind, as a new major version may break things, but I'm ok with just `>=` as well if we think the likelihood of breaking changes for the APIs we use is too low to bother.

achamayou:
@letmaik that's a good point, I hadn't thought of that. I think that's a good constraint.

letmaik:
@achamayou Could you have another look at the new constraints?

achamayou (approved):

Issue #2032, opened 2020-12-17
Pluggable Auth policy follow up items


achamayou:
- [X] Listing these auth policies in OpenAPI. I've sketched the correct type of JSON objects, but not added them yet. A slight annoyance here is it looks like there's not a way to describe our 'cert auth' policy in OpenAPI? If anyone can see a better solution than "don't mention it in the OpenAPI", please let me know. (#2049)
- [x] Exposing these auth policies through js_generic by name (in app.json and the user object). On the C++ side I've opted entirely for type-based comparisons, but since we need unique names for JS we should see if it makes sense to use these in C++ as well.
- [X] A test demonstrating a custom policy. (#2050)
- [X] installing endpoints without any policy should give a clear error, as early as possible, to avoid accidentally unauthenticated enpoints. Endpoints that deliberately want no authentication should set empty_auth_policy. (#2048)

lynshi:
Regarding your first point, we have come across the same thing - OpenAPI doesn't support cert-based auth. However, the feature request for it has been open for a while, and it looks like a `mutualTLS` option has made it into 3.1: https://github.com/OAI/OpenAPI-Specification/pull/1764.

achamayou:
@lynshi what’s your view on CCF using OpenAPI 3.1 as soon as it is released?

lynshi:
I think that's fine. Using the latest specification can't hurt for readability/maintenance going forward.

We have to write our own Swagger anyway to submit to the [Azure REST API specs repo](https://github.com/Azure/azure-rest-api-specs), which basically decouples us from CCF for anything documentation related.

PR #2048, opened 2021-01-05
Set auth policies at construction

eddyashton:
While we want to support unauthenticated endpoints, we want to make sure this is explicitly chosen and not the default. So rather than setting auth policies with calls to `add_authentication`, they are now declared at construction in `make_endpoint`. To keep unauthenticated endpoints easily greppable, there's a named `no_auth_required` rather than `{}`.

This is part of #2032.


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- To avoid accidentally unauthenticated endpoints, a vector authentication policies must now be specified at construction (as a new argument to `make_endpoint`) rather than by calling `add_authentication`. The value `ccf::no_auth_required` may be used to explicitly indicate an unauthenticated endpoint.
```
achamayou, line 12:
```suggestion
- To avoid accidentally unauthenticated endpoints, a vector of authentication policies must now be specified at construction (as a new argument to `make_endpoint`) rather than by calling `add_authentication`. The value `ccf::no_auth_required` must be used to explicitly indicate an unauthenticated endpoint.
```

achamayou (approved):

PR #2049, opened 2021-01-06
Document auth policies in OpenAPI

eddyashton:
Part of #2032.

Adds `components/securitySchemes` and per-path-operation `security` fields to generated OpenAPI documents.

### DIFF
--- a/src/ds/openapi.h
+++ b/src/ds/openapi.h
@@ -9,6 +9,7 @@
 #include <llhttp/llhttp.h>
 #include <nlohmann/json.hpp>
 #include <string>
+#include <regex>
 
 #pragma clang diagnostic push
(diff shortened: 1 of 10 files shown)


src/ds/openapi.h
```diff
@@ -49,16 +49,24 @@ namespace ds
...
     {
+      // From the OpenAPI spec:
+      // All the fixed fields declared above are objects that MUST use keys that
+      // match the regular expression: ^[a-zA-Z0-9\.\-_]+$."
       std::string s(s_);
 
       for (auto& c : s)
       {
-        if (c == ':')
+        if (
```
achamayou, line 62: I didn't think I'd ever say that, but I wonder if a std::regexp would be more readable here.

achamayou (approved):

Issue #789, opened 2020-02-04
Frontend should store a fixed number of verifiers


achamayou:
Verifiers are somewhat expensive to construct, but also to store, so we should cap the amount we create, and preferably use an LRU map or something similar.

PR #2051, opened 2021-01-07
Fix signature verification

eddyashton:
In the auth refactor, we lost the step of actually verifying HTTP signatures. There is validation of the structure of the headers and the digest to construct a `SignedReq` object, but we weren't verifying that signed req against the expected caller cert.

### DIFF
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -376,6 +376,8 @@ class RequestClient:
     CCF default client and wrapper around Python Requests, handling HTTP signatures.
     """
 
+    _auth_provider = HTTPSignatureAuth_AlwaysDigest
+
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


achamayou (approved):

jumaffre (approved):

eddyashton:
This now includes a test - `test_corrupted_signature` in `memberclient.py` - which fiddles with `RequestClient` so that it can modify the signed request in various ways. Specifically, it makes the `signature="..."` entry in the `Authorization` header invalid in various ways, and checks they all produce auth errors.

The infinitely growing verifier collection of #789 is now in `VerifierCache` in `sig_auth.h`, and per-policy rather than per-frontend.

achamayou (approved):

PR #2052, opened 2021-01-07
Update tutorial for authentication policies

jumaffre:
Follow-up to the authentication changes to the Python `CCFClient` class:

- Tutorial update to cover authentication by HTTP signature (+ doc changes)
- The tutorial is now run as part of the CI, as it was previously run only by the Release job, daily.
- Fix an issue when issuing `GET` signed requests with the `Requests` Python client, where a "Content-Length" of `0` wasn't supported (basically the same hack than in `scurl.sh`).


doc/use_apps/python_tutorial.rst
```diff
@@ -6,7 +6,7 @@ Install
 
 The CCF Python tools package can be used to interact with an existing running service and provides utilities to:
 
-- Issue HTTP requests over TLS to CCF applications
+- Issue HTTP requests over TLS to CCF
```
achamayou, line 9: Well it is to CCF application or CCF services, technically. CCF is just the framework, it is not a running instance of a service.

eddyashton:
```suggestion
- Issue HTTP requests over TLS to a CCF application
```

python/ccf/clients.py
```diff
@@ -409,6 +413,16 @@ def request(
 
         auth_value = None
         if self.signing_auth is not None:
+            # Add content length of 0 when signing a GET request
+            if request.http_verb == "GET":
+                if "Content-Length" in extra_headers and extra_headers[
```
achamayou, line 418:
```suggestion
                if extra_headers.get("Content-Length") != "0"
```

python/ccf/clients.py
```diff
@@ -409,6 +413,16 @@ def request(
...
+                    "Content-Length"
+                ] != str(0):
+                    raise ValueError(
+                        "Content-Length should be set to 0 for GET requests"
+                    )
+                else:
+                    extra_headers["Content-Length"] = str(0)
```
achamayou, line 425:
```suggestion
                    extra_headers["Content-Length"] = "0"
```

Always use literals where possible, they're not only faster to execute (no lookup), but also can't be shadowed (no lookup!).

achamayou (approved):

doc/use_apps/python_tutorial.rst
jumaffre: CCF application is too specific IMO - governance and operator requests can also be issued with the Python client. I'd rather go with "a CCF service".

python/ccf/clients.py
```diff
@@ -417,13 +417,17 @@ def request(
         if self.signing_auth is not None:
             # Add content length of 0 when signing a GET request
             if request.http_verb == "GET":
-                if extra_headers.get("Content-Length") != "0":
+                LOG.error(extra_headers.get("Content-Length"))
```
eddyashton, line 420: Don't think we want to be logging this everytime?

jumaffre: Good spot! Thanks 👍

PR #2053, opened 2021-01-07
Merkle trees: partial serialisation

wintersteiger:
This adds partial serialisation to the Merkle trees. Since only the relevant bits of an existing tree are extracted, this should give us sufficiently efficient receipt extraction. Verification is a little more expensive because the intermediate hashes have to be reconstructed, but this is usually done on small trees.


achamayou (approved):
Thank you! Could you update the code in emit_signature() to serialise just the right sub-tree (from the previous signature to the current signature)?

wintersteiger:
Sure, with pleasure!

src/node/history.h
```diff
@@ -745,7 +757,8 @@ namespace ccf
             root,
             hashed_nonce,
             primary_sig,
-            replicated_state_tree.serialise());
+            replicated_state_tree.serialise(
+              commit_txid.second, txid.version - 1));
```
achamayou, line 761: This isn't quite right, it suffers from the same problem as now: the commit index may be quite a bit behind the last signature.

wintersteiger: I was wondering about that too. How do I find the index of the previous signature?

wintersteiger: See below; is https://github.com/microsoft/CCF/pull/2053/commits/c53485564b1f67c08ebabe7efdf4cfbb7fabd19c the right way to get that index?

wintersteiger: All done! I think I have the right indices now and the CI passed all tests.

src/consensus/aft/raft.h
```diff
@@ -336,13 +336,17 @@ namespace aft
...
-    std::optional<std::pair<Term, Index>> get_signable_commit_term_and_idx()
+    std::optional<kv::Consensus::SignableTXIndices>
+    get_signable_commit_term_and_idx()
     {
       std::lock_guard<SpinLock> guard(state->lock);
       if (state->commit_idx >= election_index)
       {
-        return std::pair<Term, Index>{get_term_internal(state->commit_idx),
-                                      state->commit_idx};
+        kv::Consensus::SignableTXIndices r;
```
achamayou, line 345: We use Tx everywhere I believe, so this should probably be a SignableTxIndices, technically.

src/node/history.h
achamayou: Thank you, looks good!

Issue #2055, opened 2021-01-07
AFT: Support parallel execution on the backup


ashamis:
The primary is able to execute transactions in parallel and we want backups to be able to do the same

[ ] - primary needs to send the execution dependencies to backups
[ ] - backups can execute transaction on the non-primary thread

...
more to come

PR #2056, opened 2021-01-08
Track the max version a transaction would conflict with

ashamis:
part of #2055

To allow parallelism on the backup we need to know where conflict could occur. Step 1 is tracking the max version that a transaction reads from.

### DIFF
--- a/src/kv/generic_serialise_wrapper.h
+++ b/src/kv/generic_serialise_wrapper.h
@@ -312,7 +312,8 @@ namespace kv
       domain_restriction(domain_restriction)
     {}
 
-    std::optional<Version> init(const uint8_t* data, size_t size)
+    std::optional<std::tuple<Version, Version>> init(
+      const uint8_t* data, size_t size)
     {
       current_reader = &public_reader;
       auto data_ = data;
@@ -324,7 +325,7 @@ namespace kv
       {
         public_reader.init(data, size);
         read_public_header();
-        return version;
+        return std::make_tuple(version, max_conflict_version);
       }
 
       // Skip gcm hdr and read length of public domain
@@ -343,7 +344,7 @@ namespace kv
         domain_restriction.has_value() &&
         domain_restriction.value() == SecurityDomain::PUBLIC)
       {
-        return version;
+        return std::make_tuple(version, max_conflict_version);
       }
 
       // Go to start of private domain
@@ -362,7 +363,7 @@ namespace kv
 
       // Set private reader
       private_reader.init(decrypted_buffer.data(), decrypted_buffer.size());
-      return version;
+      return std::make_tuple(version, max_conflict_version);
     }
 
     std::optional<std::string> start_map()
--- a/src/kv/generic_serialise_wrapper.h
+++ b/src/kv/generic_serialise_wrapper.h
@@ -50,6 +50,7 @@ namespace kv
     W private_writer;
     W* current_writer;
     Version version;
+    Version max_conflict_version;
     bool is_snapshot;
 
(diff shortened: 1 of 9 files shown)


src/kv/generic_serialise_wrapper.h
```diff
@@ -50,6 +50,7 @@ namespace kv
     W private_writer;
     W* current_writer;
     Version version;
+    Version max_conflict_version;
```
jumaffre, line 53: As it is not updated automatically yet, could you update the table in https://microsoft.github.io/CCF/master/build_apps/kv/kv_serialisation.html#serialised-transaction-format as well?

src/kv/untyped_map.h
```diff
@@ -189,6 +189,11 @@ namespace kv::untyped
               LOG_DEBUG_FMT("Read depends on invalid version of entry");
               return false;
             }
+
+            if (max_conflict_version < search.value().version)
+            {
+              max_conflict_version = search.value().version;
```
jumaffre, line 195:
```suggestion
            if (max_conflict_version < search->version)
            {
              max_conflict_version = search->version;
```

src/kv/snapshot.h
```diff
@@ -41,7 +41,7 @@ namespace kv
     std::vector<uint8_t> serialise(
       std::shared_ptr<AbstractTxEncryptor> encryptor)
     {
-      KvStoreSerialiser serialiser(encryptor, version, true);
+      KvStoreSerialiser serialiser(encryptor, version, version - 1, true);
```
jumaffre, line 44: Perhaps add a short comment here to explain that the value doesn't really matter for snapshots?

src/kv/untyped_map.h
```diff
@@ -141,7 +141,7 @@ namespace kv::untyped
         return committed_writes || change_set.has_writes();
       }
 
-      bool prepare() override
+      bool prepare(kv::Version& max_conflict_version) override
```
jumaffre, line 144: Nit: AFAIA, the KV code so far doesn't use out parameters. Could `prepare()` return an `std::optional<kv::Version>` instead?

ashamis: Its not done often but in some places like commit_deserialised.

The version of this code where we return an optional is a fair bit messier which is why i went with the out param

src/kv/snapshot.h
```diff
@@ -41,7 +41,10 @@ namespace kv
     std::vector<uint8_t> serialise(
       std::shared_ptr<AbstractTxEncryptor> encryptor)
     {
-      KvStoreSerialiser serialiser(encryptor, version, true);
+      // Set the execution dependency for the snapshot to be the version
+      // previous to said snaphot to ensure that the correct snaphot is
```
achamayou, line 45:
```suggestion
      // previous to said snapshot to ensure that the correct snapshot is
```

src/kv/tx.h
```diff
@@ -388,7 +389,13 @@ namespace kv
       auto map = all_changes.begin()->second.map;
       auto e = map->get_store()->get_encryptor();
 
-      KvStoreSerialiser replicated_serialiser(e, version);
+      kv::Version mcv = max_conflict_version;
```
achamayou, line 392: Why create a distinct mcv? I thought it might be to avoid writing to an argument, but we do that at 298, so might as well set max_conflict_version again here?

src/kv/untyped_map.h
achamayou: Same thing for hooks fwiw.

achamayou (approved):

Issue #1954, opened 2020-11-27
Re-organize /network, /network_info, /node/ids, /primary, /primary_info


letmaik:
The network-related endpoints seem a bit random in their structure.

As a user I would have expected a hierarchical structure like:
- `/network`
  - general network info without node details, maybe just node count
  - ```json
    { "service_status": "OPENING|OPEN|WAITING_FOR_RECOVERY_SHARES",
      "current_view": 123,
      "primary_id": 456
    }
    ```
- `/network/nodes`
  - collection of nodes
  - the primary node has an extra field `"primary": true`
  - optional filtering by public host/port: `?host=abc&port=444`
  - ```js
    { "nodes": [ .. ] } // list of /network/nodes/{id}
    ```
- `/network/nodes/{id}`
  - ```js
    { "node_id": "...",
      "status": "PENDING|TRUSTED|RETIRED",
      "host": "...", // public host
      "port": "...", // public port
      "rpc_host": "...", // internal host
      "rpc_port": "...", // internal port
      "primary": true // optional
     }
    ```
- `/network/nodes/self`
  - `307 Temporary Redirect`
  - `Location: /network/nodes/123`
- `/network/nodes/primary`
  - `307 Temporary Redirect`
  - `Location: /network/nodes/456`
  - Note: This is not a redirect to the remote node's endpoint.

While doing this, I would unify the response data models for nodes and not have different ones for non-primary and primary.

NOTE: is_primary (status code based) is still needed!

letmaik:
@eddyashton Would the URL scheme above work? There is an overload of `/network/nodes/{id}` and `/network/nodes/self` / `/network/nodes/primary` which in practice would be fine but I'm not sure if CCF has extra restrictions on that.

eddyashton:
@letmaik These overloads should work correctly - in `EndpointRegistry::find_endpoint()`, we first look for an exact match if it exists (which will find `/nodes/self` and `/nodes/primary`), and only look for a templated match (`/nodes/{id}`) if we don't have an exact match. I don't think we have anything testing this currently, but it should work.

PR #2059, opened 2021-01-08
Re-organize network endpoints

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1954.


src/node/rpc/call_types.h
```diff
@@ -65,38 +55,42 @@ namespace ccf
...
   };
 
-  struct GetNodesByRPCAddress
+  struct GetNode
   {
     struct NodeInfo
     {
       NodeId node_id;
       NodeStatus status;
+      std::string host;
```
jumaffre, line 72: Why not use `NodeInfoNetwork` here directly? Some fields (e.g. public interface) are missing for example.

jumaffre: Ah, checked the implementation of the endpoints that populate this struct and we do return the public port. In fact, we'd report the public and private RPC interfaces. Is this expected? If my understand is correct, one seems interesting for operators (currently `rpc_host` and `rpc_port`) while the public interface is mostly for users/members?

src/node/rpc/node_frontend.h
```diff
@@ -332,22 +332,184 @@ namespace ccf
...
         json_read_only_adapter(network_status),
         no_auth_required)
+        .set_auto_schema<void, GetNetworkInfo::Out>()
+        .install();
+
+      auto get_nodes = [this](auto& args, nlohmann::json&& params) {
+        const auto in = params.get<GetNodes::In>();
+        GetNodes::Out out;
+
+        auto nodes_view =
```
jumaffre, line 372: You can probably remove the awkward template here by calling `get_read_only_view` directly on the map instance (i.e. args.tx.get_read_only_view(this->network.nodes)`).

src/node/rpc/call_types.h
letmaik: I used a new type to provide a boundary between internal and public interface, since we don't want to break the latter even if the internal one changes. 
And yes, returning the public and private addresses is expected, this was an ask from @lynshi and I don't see any harm in it either way.
We should discuss naming, rpc_* doesn't seem like a great choice.

src/node/rpc/node_frontend.h
letmaik: Done.

src/node/rpc/call_types.h
jumaffre: These could be exposed as `private_host/port` perhaps?

achamayou: `local` would preferable to `private`, these are the local interfaces the nodes were told to bind to on startup. Whether they differ from `public` depends on whether there is some routing/NATing going on, but they aren't really `private`, at least not in way that compares to other things being private in CCF.

letmaik: What about `nodehost`/`nodeport`? Does it make sense to expose that as well?

jumaffre: @letmaik These are only used for node-to-node communications and aren't exposed to clients directly so my hunch is that they shouldn't be exposed directly here.

letmaik: OK, I renamed it to `local_*`

letmaik:
github pages build fails with:
```
2021-01-12T14:38:38.6194169Z reading sources... [ 71%] operations/operator_rpc_api
2021-01-12T14:38:38.7029753Z /tmp/tmpql0jqllv/7524c1c8e1d80930a16c1dbc0719b6c0d196ab04/doc/governance/hsm_keys.rst:2: WARNING: Duplicate explicit target name: "here".
2021-01-12T14:38:38.7030487Z /tmp/tmpql0jqllv/7524c1c8e1d80930a16c1dbc0719b6c0d196ab04/doc/governance/hsm_keys.rst:2: WARNING: Duplicate explicit target name: "here".
2021-01-12T14:38:38.9414601Z 
2021-01-12T14:38:38.9415027Z Exception occurred:
2021-01-12T14:38:38.9416063Z   File "/__w/1/s/env/lib/python3.8/site-packages/sphinxcontrib/openapi/openapi30.py", line 281, in _httpresource
2021-01-12T14:38:38.9416591Z     type=param['schema']['type'],
2021-01-12T14:38:38.9416959Z KeyError: 'type'
```
@eddyashton Any idea? The schema seems valid according to https://apitools.dev/swagger-parser/online/

eddyashton:
@letmaik I agree this looks like valid OpenAPI. We run the generated API documents through a validator in the e2e tests, and that seems happy, and SwaggerHub is also happy with `node_openapi.json`, so I'm not sure why this sphinx plugin isn't... Do you know which element it breaks on? I'm guessing its the `status` parameter in `GET /network/nodes`, which now has just an `enum` field but no `type` field?

letmaik:
Yes, you're right. Looks like a bug in the library unfortunately. It's the only enum query parameter we have so far. And `type` is definitely not needed there.

letmaik:
Although it would be fine to add `"type": "string"` for enums, since all our enums are string-based. Should we just do that?

https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values

eddyashton:
> Yes, you're right. Looks like a bug in the library unfortunately. It's the only enum query parameter we have so far. And `type` is definitely not needed there.

Can you raise an issue on the library? As a workaround for now, I think we can just add a `"type": "string"` field to all our enum schemas (this isn't enforced to be true but is true in practice, I believe). At the bottom of `json.h` in `fill_enum_schema` is where we construct these objects, so it should be as simple as adding `j["type"] = "string";` after `j["enum"] = ...`. If you want to be more precise, it could guarded by a check that every element in `enums` _is_ a string (these are nlohmann::json objects, so `.is_string()`), but this is probably unnecessary. If we can get a fix in the sphinx lib that would be much better - this is a less precise constraint than provided by the `enum` field.

letmaik:
https://github.com/sphinx-contrib/openapi/issues/106

I'll add your proposed fix to json.h.

achamayou (approved):

PR #2060, opened 2021-01-08
Implement ConsensusHook

### LINKED ISSUE #1368: On primary, KV local hooks are not triggered in total transaction order
On the primary node, KV hooks are not currently triggered in the same order as the total order of the transactions they correspond to. For example, for 2 concurrent RPCs executed by the primary and which trigger a local hook, it is possible that the hook of the first transaction (executed a version `V`) is triggered _after_ the hook of the second transaction (executed at version `V+1`). 

Note that this issue does not happen on the backups as all write sets are applied sequentially and by the same thread, thanks to the Raft lock in `recv_append_entries()`. This is not an issue for global hooks either which are directly triggered by the consensus thread (on consensus commit). 

In particular, this is an issue for Raft reconfiguration in which new nodes are added to a new configuration whenever (local hook) a new node is `TRUSTED` by the KV. The `add_configuration` function below, which specifies that it should be called when the Raft lock is taken, is actually called _without_ acquiring lock on the primary. Concurrent rollback/compact on the consensus thread would cause concurrent access to the `configurations` and `nodes` members.
https://github.com/microsoft/CCF/blob/ffdb7d56df8425c861d5ad278e59969983ba3917/src/consensus/raft/raft.h#L287-L292

achamayou:
This replaces `local_hooks` with `map_hooks`, which are very much like `local_hooks` (*) except they're allowed to return a `ConsensusHook`. A `ConsensusHook` is a functor that is executed:

- in total order
- under consensus lock

and is a passed a `ConfigurableConsensus`. That's voluntarily scoped, but could in principle be extended to other aspects of the consensus we think we want to manipulate in consensus hooks.

(*) ie. they execute at Tx::commit() time.

This fixes #1368.

I have not removed global commit hooks (yet), but I would like to discuss implementing them by adding an attribute to `ConsensusHook`. This is to implement post-reconfiguration forwarding.


src/kv/kv_types.h
```diff
@@ -50,8 +50,58 @@ namespace kv
...
+  // ordering
+  using SeqNo = int64_t;
+
+  struct Configuration
+  {
+    struct NodeInfo
+    {
+      std::string hostname;
+      std::string port;
+      tls::Pem cert = {};
```
jumaffre, line 63: Depending on a type from `tls` isn't ideal here. I'm aware this was already the case. Could the configuration be defined in a different file, outside of the `kv` namespace?

src/node/hooks.h
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ds/logger.h"
+#include "kv/change_set.h"
```
jumaffre, line 6: Why this include in particular? Should `kv_types.h` be all that's necessary here?

src/kv/kv_types.h
```diff
@@ -50,8 +50,58 @@ namespace kv
...
+    SeqNo idx;
+    Nodes nodes;
+  };
+
+  class ConfigurableConsensus
+  {
+  public:
+    virtual void add_configuration(
+      SeqNo seqno, const Configuration::Nodes& conf) = 0;
+    virtual Configuration::Nodes get_latest_configuration() const = 0;
```
jumaffre, line 88: Looking at `add_configuration()` in `raft.h`, should `configurations` be protected by its own lock? I can see a few places where it is used, and could potentially be used by different threads simultaneously?

src/kv/kv_types.h
achamayou: Agreed, it's not clear to me why this needs to be a Pem. Then I had a look and as best I can tell it's not used at all, so I will remove it.

src/kv/kv_types.h
achamayou: That's currently assumed to be called under the Consensus lock (I'll add a comment in case it was not obvious). There's no real concurrency in the consensus itself right now, and I don't think making the config separately lockable is going to help that. Am I missing something obvious?

src/node/hooks.h
achamayou: That's only a path dependency, I'll remove it, thank you.

src/consensus/aft/raft.h
```diff
@@ -187,6 +187,8 @@ namespace aft
       }
     }
 
+    virtual ~Aft() {}
```
ashamis, line 190: nit: "=default"

src/consensus/aft/raft.h
```diff
@@ -395,12 +397,20 @@ namespace aft
       return get_latest_configuration().size();
     }
 
-    template <typename T>
+    template <typename T, typename H>
```
ashamis, line 400: Is it possible to not use a template type here? I would like to remote T in the future

src/kv/kv_types.h
```diff
@@ -50,8 +50,53 @@ namespace kv
...
+  // SeqNo indexes transactions processed by the consensus protocol providing
+  // ordering
+  using SeqNo = int64_t;
+
+  struct Configuration
+  {
+    struct NodeInfo
+    {
+      std::string hostname;
+      std::string port;
```
ashamis, line 62: should this not be a uint16_t?

src/kv/kv_types.h
ashamis: I am going to change many things with locking with multi-threading on the backup so this seems fine for now?

ashamis (approved):

src/kv/kv_types.h
achamayou: Removed

src/node/hooks.h
achamayou: Removed

src/kv/kv_types.h
achamayou: Yes, let's revisit locking if we think there's something to be gained from changing granularity. As I understand the changes you're making though, only execution will become concurrent, actual consensus logic remains sequential?

src/consensus/aft/raft.h
achamayou: That's what I'd gone for at first, but it created a dependency cycle. A bunch of stuff has changed, let me try again.

src/kv/kv_types.h
achamayou: It's resolved further down, this is still raw from the KV at this point. Technically, Unix allows named ports (/etc/services), and I expect libuv probably handles them (or rather, does nothing to prevent them from working). They're rather out of fashion these days.

src/consensus/aft/raft.h
achamayou: Yes

src/consensus/aft/raft.h
achamayou: We can because of ConfigurableConsensus, done.

jumaffre (approved):

eddyashton (approved):

PR #2062, opened 2021-01-11
AFT: Commands should not run within the consensus

ashamis:
Part of #2055, #1601

Commands are not passed a transaction and run outside the consensus. This is the case for CFT but was not the case for BFT. In this change I change the bool which says if a request runs locally locally, locally on the primary, or always through the consensus protocol.


src/node/rpc/endpoint.h
```diff
@@ -47,12 +55,18 @@ namespace ccf
        {ForwardingRequired::Always, "always"},
        {ForwardingRequired::Never, "never"}});
 
+    DECLARE_JSON_ENUM(
+      ExecuteOutsideConsensus,
+      {{ExecuteOutsideConsensus::Never, "never"},
+       {ExecuteOutsideConsensus::Locally, "locally"},
+       {ExecuteOutsideConsensus::Primary, "primary"}});
```
eddyashton, line 62: I'm trying to understand why there's 3 cases here - I would have expected just `Always` and `Never` - and how they interact with the forwarding options. What does a backup do with a `Primary` transaction - does this override the `ForwardingRequired` property and always get forwarded?

It feels like the default for a "command" should be "always execute locally". Forwarding to the primary, then executing outside of consensus, feels like an attempt to get a service-level answer for something which is really node-local (if there was a service-safe answer, it would be in the KV).

ashamis: Discussed offline - I renamed things as discussed. There will be followup following this PR (as discussed)

doc/build_apps/js_app_bundle.rst
```diff
@@ -55,7 +55,7 @@ The ``app.json`` file of an app bundle has the following structure:
             "js_module": "app.js",
             "js_function": "foo_post",
             "forwarding_required": "never",    
-            "execute_locally": true,           
+            "execute_outside_consensus": true,
```
eddyashton, line 58:
```suggestion
            "execute_outside_consensus": "never",           
```

eddyashton (approved):

Issue #1468, opened 2020-07-31
Update attestation API and cache quote endorsements


jumaffre:
Open Enclave exposes a new richer attestation API (`oe_get_evidence`, etc.) that also returns the endorsements for a quote. We should use this API and cache the quote endorsements in the `ccf.nodes` table (along with the quote) and return it to clients who want to verify a CCF node's quote (via `/node/quote(s)`). 

This is particularly important for clients/auditors who want to verify quotes from "older" nodes in the networks whose quote might no longer verify in the present (e.g. they ran on a CPU whose SVN was bumped up since). With the endorsements attached, they could verify these quotes without retrieving the endorsements from Azure's caching server.

I believe this is currently blocked by https://github.com/openenclave/openenclave/issues/3333

jumaffre:
As of OE 0.11, `host_verify` uses the new attestation API so we are now able to update to the new attestation API.

PR #2063, opened 2021-01-12
Expose versioned C++ API for re-exposing endpoints

eddyashton:
This is a draft as it needs discussion and improvement.

The goal is to build a scheme for exposing C++ implementations of some of our endpoints, which we will commit to support long term. App writers can choose not to expose the standard/common endpoints, but if they want similar functionality under a different endpoint they can call these functions.

These functions rely on some state that is currently part of an `EndpointRegistry` (`consensus`, `history`, and - with substantial plumbing to make this available everywhere - `node`), so they are implemented as methods. An advantage of this is it minimises the _types_ that are part of the supported interface: if `kv::Consensus` goes away, we can change the implementation of `get_status_for_txid_v1(view, seqno)` while maintaining the same function signature, whereas a free-standing `get_status_for_txid_v1(consensus, view, seqno)` would tie us to keeping `kv::Consensus` for much longer. However these methods bloat the already busy `common_endpoint_registry.h` - ideally they'd be free-standing functions for maintainability and to make their selection+use explicit. So we need to decide how these API implementations access standard services like `Consensus` and `History`, while hiding those services where possible from the API signature:
1) Implemented as methods, so `kv::Consensus` or its replacement is a member of `EndpointRegistry`. Means we either need to pass this at construction of the app, or set it via the frontend in `node_state.h`.
2) Implemented as free functions, taking an opaque context which retrieves those services. `EndpointRegistry`s, and in turn app code, has access to some opaque `api_context` object. All the API functions take `ApiContext*` as their first arg, and know they can cast it/call `->get_consensus()` on it. Its either a property bag or a service locator, depending how much complexity we need. If we do this _right_, it gives us a lot of future proofing - new services can be plumbed in, and if apps can define multiple services or in some other way need to see dynamic state, they can get a tweaked `ApiContext`.
3) Implemented as free functions, and these services become singletons or otherwise statically-lookupable. In practice all of the calls we currently need to make ("what is my node ID?", "what consensus algorithm is this node running?") are fairly static, and could sensibly come from `Consensus::get()->type()`. This is a slightly hairy rewrite now, and "singletons-bad", but may be simpler to reason about/extend/maintain going forward.

_EDIT - After some discussion we've opted for keeping these as methods, as a clean way of hiding this state dependency. I've moved them into a `BaseEndpointRegistry` in the hierarchy, so they're easily reused but maintainably distinct from `CommonEndpointRegistry`._

The style of these APIs is currently inconsistent - some return values, some `std::optional`s, some throw, some have `error_reason` strings. This is just a paste of their existing implementation, but we should consolidate these into some standard style. In particular we shouldn't be returning the custom `Quote` struct, just the raw quote itself.

_EDIT - I've opted for "everything returns an error string", with the requested data returned in `out&` parameters. This leads to a simple and consistent calling pattern: `const auto error_reason = foo_v1([tx,] input..., output...); if (error_reason.empty()) { return success; } else { return error; }`. A caveat here is that I thought they should all take a `tx`, as this feels like sensible future proofing for ways their implementation may have to change, but some of these are `tx`-less commands so this isn't possible._

_EDIT2 - Switched these return values to an `ApiResult` enum. Its switchable, avoids us tweaking the returned strings and breaking code that was trying to grep them. The error strings are now trace-logged, but very little of this contains anything specific to these APIs (ie - it is just surfacing an error from further down the stack, and even then those errors should be rare)_

TODO:
- [X] Doxygen describing this API, each of the supported methods and any supporting types. Probably a blurb on the general style too.
- [X] Test and document that we can build an app that doesn't expose the common endpoints, and rewraps some of these APIs in its own endpoints.
- [X] Mention this in the `CHANGELOG`


src/node/rpc/common_endpoint_registry.h
```diff
@@ -7,31 +7,171 @@
...
+        Quote q;
+        q.node_id = node_id;
+
+        if (node_info->status == ccf::NodeStatus::TRUSTED)
+        {
+          q.raw = fmt::format("{:02x}", fmt::join(node_info->quote, ""));
+
+#ifdef GET_QUOTE
+          // TODO: Why don't we include this in BFT?
+          if (consensus != nullptr && consensus->type() != ConsensusType::BFT)
```
eddyashton, line 119: @jumaffre @ashamis This was originally in `node_state.h` but moved here. Can you explain why the BFT path doesn't extract the code ID from the quote? We should add a comment if this is something we want to maintain long term.

ashamis: This was performance related. The BFT replays all the tx so every replica would attest every other replica. At one point attestation was very slow at times (not sure if this was the case) so the BFT tests were taking way too long

achamayou: That needs re-enabling now, good catch: #2064

src/node/rpc/common_endpoint_registry.h
```diff
@@ -7,31 +7,174 @@
...
 
 namespace ccf
 {
+  struct Quote
+  {
+    NodeId node_id = {};
+    std::string raw = {}; // < Hex-encoded
+
+    std::string error = {};
+    std::string mrenclave = {}; // < Hex-encoded
```
achamayou, line 21: Can we make those doxygen comments? That way they should show up when we expose the type in the doc.

src/node/rpc/common_endpoint_registry.h
```diff
@@ -7,31 +7,174 @@
 #include "http/ws_consts.h"
 #include "json_handler.h"
 #include "node/code_id.h"
+#include "node/quote.h"
+#include "node/rpc/node_interface.h"
 
 namespace ccf
 {
+  struct Quote
+  {
```
achamayou, line 16: Let's wrap this in a top-level + a version field, set to "oe_sgx_v1" or something along those lines, for future proofing.

src/node/rpc/common_endpoint_registry.h
eddyashton: I've changed the API here. `get_quote_for_this_node_v1` just returns the raw quote as `std::vector<uint8_t>` (with a `QuoteFormat` enum of `oe_sgx_v1`). Then this `Quote` struct only lives in the `node_frontend`, as part of our HTTP API, but not as part of the supported C++ API. Separately I've also remove the `error` field from this struct in its new home, I think it was an odd way of indicating quote errors.

eddyashton: But I will add doxygen comments on these methods, since we want them to be well documented. I'll add a TODO list...

src/node/rpc/base_endpoint_registry.h
```diff
@@ -0,0 +1,148 @@
...
+#pragma once
+
+#include "ds/json.h"
+#include "endpoint_registry.h"
+#include "node/quote.h"
+#include "node/rpc/node_interface.h"
+
+namespace ccf
+{
+  enum class QuoteFormat
```
jumaffre, line 12: FYI, until https://github.com/microsoft/CCF/issues/1468 is complete, we're still using (what OE refers to) as the legacy quote format (see https://github.com/openenclave/openenclave/blob/master/include/openenclave/attestation/sgx/evidence.h#L83-L123). I think we should upgrade sooner than later (I've got a half-finished branch somewhere to use the new API).

I don't think this impacts this enum though, as CCF v1 should just choose a quote format and stick with it.

eddyashton: So this enum is just trying to capture what CCF did to generate the quote. I assume OE's format tag is somewhere inside the quote blob? Or should we also be extracting that and passing it around to support later verification?

achamayou: Not all OE quote formats are tagged (the newer ones are), but CCF must have its own separate tag because the quote may or may not come from OE.

src/node/rpc/base_endpoint_registry.h
```diff
@@ -66,6 +78,8 @@ namespace ccf
       }
     }
 
+    /** Get the ID of latest transaction known to be committed.
+     */
     std::string get_last_committed_txid_v1(
```
achamayou, line 82: Why go with returning an error string over returning the data type and throwing on error?

achamayou (approved):

PR #2065, opened 2021-01-12
AFT: Resolve recursive lock issues with BFT

### LINKED ISSUE #1601: Fix re-entrant lock issues BFT has with the AFT class

ashamis:
Resolves #1601
Part of #2055

This resolves 2 issues:
- The "primary_info" endpoint should run outside of BFT consensus on the node that receives the request
- After executing a transaction the frontend returns the last global commit. This change adds a check to see if this value has been set and only sets it in the frontend if it has not been set. This allows for BFT to set this value at an appropriate time.

Minor: Change log level for a info log in ledger.h

### DIFF
--- a/tests/election.py
+++ b/tests/election.py
@@ -42,7 +42,7 @@ def test_kill_primary(network, args):
                 )
         except CCFConnectionException:
             LOG.warning(
-                f"Could not successfully connect to node {backup.node_id}. Retrying..."
+                f"Could not successfully connect to node {backup.node_id}."
             )
 
     new_primary, new_term = network.wait_for_new_primary(primary.node_id)
--- a/tests/election.py
+++ b/tests/election.py
@@ -41,9 +41,7 @@ def test_kill_primary(network, args):
                     },
                 )
         except CCFConnectionException:
-            LOG.warning(
-                f"Could not successfully connect to node {backup.node_id}."
-            )
+            LOG.warning(f"Could not successfully connect to node {backup.node_id}.")
 
     new_primary, new_term = network.wait_for_new_primary(primary.node_id)
     LOG.debug(f"New primary is {new_primary.node_id} in term {new_term}")
--- a/tests/election.py
+++ b/tests/election.py
@@ -131,13 +131,8 @@ def run(args):
 
             try:
                 test_kill_primary(network, args)
-            except (PrimaryNotFound, TimeoutError) as e:
+            except PrimaryNotFound:
                 if node_to_stop < nodes_to_stop - 1:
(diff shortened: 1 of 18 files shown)


achamayou:
> After executing a transaction the frontend returns the last global commit.

This is something we intend to remove, it's only there as an artifact of the current performance test implementation.

tests/election.py
```diff
@@ -23,8 +24,27 @@
 @reqs.description("Stopping current primary and waiting for a new one to be elected")
 @reqs.can_kill_n_nodes(1)
 def test_kill_primary(network, args):
-    primary, _ = network.find_primary()
+    primary, backup = network.find_primary_and_any_backup()
     primary.stop()
+
+    # When the consensus is BFT there is no status message timer that triggers a new election.
```
achamayou, line 30: I am confused by this:

1. Didn't this test pass before? Which part of this PR changed that?
2. Shouldn't "X seconds without a signature from the primary" also trigger a view change? Why is a dummy request necessary?

tests/election.py
```diff
@@ -111,9 +131,14 @@ def run(args):
 
             try:
                 test_kill_primary(network, args)
-            except PrimaryNotFound:
+            except (PrimaryNotFound, TimeoutError) as e:
                 if node_to_stop < nodes_to_stop - 1:
                     raise
+                elif not (
+                    (type(e) == PrimaryNotFound and args.consensus == "cft")
+                    or (type(e) == TimeoutError and args.consensus == "bft")
```
achamayou, line 139: I don't understand why this distinction is necessary or useful here. test_kill_primary should not exhibit consensus-specific behaviour here, only nodes_to_stop (calculated accordingly) should be allowed to differ?

tests/election.py
ashamis: Previously primary info triggered a view change because it was going through the consensus now it is not so after killing the primary we need some request to go through the consensus, timeout, triggering a view change.

achamayou: Oh, I see.

achamayou (approved):

PR #2070, opened 2021-01-13
Use thread messaging round robin for snapshot generation

jumaffre:
Fixes issues related to multithreading (see #2069)

Asynchronous snapshot generation used its own round robin function which wasn't correct in a multi-threading setting (it would dispatch snapshots to the last thread + 1). Instead, use the one from `thread_messaging.h` directly.

### DIFF
--- a/src/node/snapshotter.h
+++ b/src/node/snapshotter.h
@@ -53,22 +53,6 @@ namespace ccf
     // Indices at which a snapshot will be next generated
     std::deque<consensus::Index> next_snapshot_indices;
 
-    size_t get_execution_thread()
-    {
-      // Generate on main thread if there are no worker threads. Otherwise,
-      // round robin on worker threads.
-      if (threading::ThreadMessaging::thread_count > 1)
-      {
-        static size_t generation_count = 0;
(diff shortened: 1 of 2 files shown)


src/node/snapshotter.h
```diff
@@ -194,8 +178,11 @@ namespace ccf
         msg->data.snapshot = network.tables->snapshot(idx);
 
         last_snapshot_idx = idx;
+
+        static uint32_t generation_count = 0;
         threading::ThreadMessaging::thread_messaging.add_task(
-          get_execution_thread(), std::move(msg));
+          threading::ThreadMessaging::get_execution_thread(generation_count++),
+          std::move(msg));
```
ashamis, line 185: you probably do not want to run on thread 0 if there is more than 1 thread

ashamis (approved):

src/node/snapshotter.h
jumaffre: Doesn't `++tid;` in `thread_messaging::get_execution_thread()` prevent that?

ashamis: yes you are right

achamayou (approved):

PR #2071, opened 2021-01-13
Add timeouts to Merkle tree tests

wintersteiger:
This adjusts the numbers on Merkle tree tests so that they don't take too much time during CI runs and it adds timeout checking so no spurious failures are reported.


src/libmerklecpp/test/flush.cpp
```diff
@@ -12,24 +12,27 @@
 
 int main()
 {
+  auto test_start_time = std::chrono::high_resolution_clock::now();
+  double timeout = get_timeout();
```
achamayou, line 16: That's oddly specific but ok :)

achamayou (approved):
Thank you, that will prevent the tests from timing out in Daily while running completely in CI context.

src/libmerklecpp/test/flush.cpp
wintersteiger: How so? It gets the timeout from the environment variable.

achamayou: I meant making it explicitly a double when everything around is auto :)

achamayou:
This probably needs a force merge too since it fixes an issue in the Daily.

wintersteiger:
Good point, yes. It threw a different error the last time, I'll check whether it still does so.

src/libmerklecpp/test/flush.cpp
wintersteiger: Oh yes, that's unnecessary :)

Issue #2020, opened 2020-12-11
Ledger Secrets Management


jumaffre:
This issue should summarise some of the recent discussions on ledger secret management (e.g. #1986). I've added tasks inline to cover the work that needs to be completed.

TL;DR: We need to fix #1986, pass the locally committed ledger secrets on join, and add support for historical queries after rekeys/snapshots.

## Ledger Rekey

Rekeying the ledger rotates the key used to encrypt the ledger entries. Rekey at `seqno` `N` writes new secret for all nodes in KV (in the `ccf.gov.secrets` table), including the primary node (encrypted with each node's encryption key) so that ledger secret are used from `N+1` onwards.

Because txs don't take a lock between commit (`apply_changes()`) and serialise, two parallel rekey txs may interleave commit and serialise, e.g. commit(tx0 at `N`) -> commit(tx1 at `N+1`) -> serialise(tx1 at `N+1`) -> serialise(tx0 at `N`). Because of this, the encryptor needs to keep track of the ledger secrets, along with the `seqno` at which they're applicable from.

When a node no longer needs to use a secret (i.e. global commit past the `seqno` of the _next_ ledger secret), it can safely discard it from the encryptor. More precisely, the node will no longer need to use the key for ledger encryption/decryption (\*) so the key can be removed from the main store's encryptor so that key lookup on encryption/decryption is as cheap as possible (it is based on binary search). However, the ledger secret needs to be kept saved to be passed on to new nodes on join (see `Sharing Ledger Secrets` section).

(\*): It may still be used for historical queries, but the historical query mechanism uses a different store/encryptor. See `Snapshots and Historical Queries` section below.

## Recovery

When recovering a service, the first recovery node creates a ledger secret that is used to integrity protect the entries in the public-only service. When the recovery is complete, the same secret is used to encrypt/decrypt private entries.

The history of ledger secrets is stored in the `ccf.gov.shares` table, at a unique key (`0`), where each ledger secret is recorded, encrypted with the next ledger secret. The history of ledger secrets is recovered when the public ledger is read. On private recovery, the latest ledger secret is decrypted from the recovery shares, which can decrypt the previous ledger secret, and so on.

In short, the historical ledger secrets can be recovered from the ledger by looking up entries backwards.

## Sharing Ledger Secrets

As mentioned above, on rekey, the primary writes the new ledger secret for each trusted node in the KV at the time in the `ccf.secrets` table, encrypted with each node's encryption public key.

[...]

PR #2073, opened 2021-01-13
Re-hauling encryptor for multi-threading

### LINKED ISSUE #1567: Combine the CFT and BFT encryptors
We currently have a BFT and CFT encryptors. We need to look into combining them so that we only have 1 encryptor, if this is at all possible.

### LINKED ISSUE #1986: Getting compacted keys from a later primary does not work

jumaffre:
Fixes #1986
Fixes #1567 
And item 1. from #2020 

As per #1986 and #2020, there were some issues with the way CCF was managing ledger secrets. This PR fixes those as well as combine the CFT and BFT encryptors which used a different IV. In particular:

- CFT and BFT encryptors are unified. The IV used now depends on `{KV seqno, term, is_snapshot}` so that the IV isn't re-used but yet produces the same encrypted write sets from the same input in BFT mode (see `encryptor.h`) 
- The `ccf.secrets` table is now keyed by node id so that it easy for a node to take a read dependency on that table when accessing the ledger secrets (see following point)
- Accessing the ledger secrets now require a `kv::Tx` which takes a read dependency on the `ccf.secrets` table so that concurrent rekey (which write the new encrypted secret to that table for all nodes) operations will result in a KV conflict.
- Because the join procedure is a three-step procedure (added as pending -> node is trusted by consortium -> node requests network secrets) and the new joiner only needs the ledger secrets up to the seqno it is trusted, CCF records the version of the latest ledger secret when a node is trusted and only release ledger secrets until then. Future ledger secrets are passed through the `ccf.secrets` table.


src/crypto/symmetric_key.h
```diff
@@ -19,7 +19,7 @@ namespace crypto
     uint8_t iv[SIZE_IV] = {};
 
     // 12 bytes IV with 8 LSB are unique sequence number
-    // and 4 MSB are 4 LSB of unique id (Node id or View)
+    // and 4 MSB are 4 LSB of term (with last bit indicating a snapshot)
```
achamayou, line 22: Can we use "view" rather than "term" here, since this is a cross-consensus concern.

src/kv/encryptor.h
```diff
@@ -2,144 +2,82 @@
...
-
-      return search->key;
+      // IV is function of seqno, term and snapshot so that:
+      // - Same seqno across rollbacks does not reuse IV
+      // - Snapshots do not reuse IV
+      // - If all nodes execute the _same_ tx (or generate the same snapshot),
+      // the same IV will be used
+
+      hdr.set_iv_seq(version);
+      hdr.set_iv_term(term);
```
achamayou, line 28: "view" here and elsewhere please.

src/kv/generic_serialise_wrapper.h
```diff
@@ -50,6 +50,7 @@ namespace kv
     W private_writer;
     W* current_writer;
     Version version;
+    Term term; // Used for encryption IV only
```
achamayou, line 53: Taking a TxID down that path may end up being a little cleaner.

src/node/ledger_secrets.h
```diff
@@ -3,118 +3,282 @@
...
+    {
+      // Ledger secrets are not stored in the KV. Instead, they are
+      // cached in a unique LedgerSecrets instance that can be accessed without
+      // reading the KV. However, it is possible that the ledger secrets are
+      // updated (e.g. rekey tx) concurrently to their access by another tx. To
+      // prevent conflicts, accessing the ledger secrets require access to a tx
+      // object, which must take a dependency on the secrets table.
+      auto v = tx.get_view(secrets_table);
+
+      // Taking a depency on the key at self, which would get updated on rekey
```
achamayou, line 109:
```suggestion
      // Taking a read dependency on the key at self, which would get updated on rekey
```

src/node/node_state.h
```diff
@@ -1597,85 +1552,65 @@ namespace ccf
...
+                   ledger_secret_set.encrypted_secrets)
               {
-                if (encrypted_secret_for_node.node_id == self)
+                auto plain_ledger_secret = LedgerSecretsBroadcast::decrypt(
+                  node_encrypt_kp,
+                  tls::make_public_key(
+                    ledger_secret_set.primary_public_encryption_key),
+                  encrypted_ledger_secret.encrypted_secret);
+
+                // On rekey, the version is infered from the version at which
```
achamayou, line 1585:
```suggestion
                // On rekey, the version is inferred from the version at which
```

src/crypto/symmetric_key.h
jumaffre: As discussed, let's stick to "term" as this comes from the KV directly, which refers to consensus view as "term" so that it's not confused with a `kv::Tx`'s view.

src/kv/encryptor.h
jumaffre: Same as above.

src/kv/generic_serialise_wrapper.h
jumaffre: Done.

src/crypto/symmetric_key.h
```diff
@@ -40,16 +40,16 @@ namespace crypto
       *reinterpret_cast<uint64_t*>(iv) = seq;
     }
 
-    void set_iv_id(uint64_t id)
+    void set_iv_term(uint64_t term)
```
ashamis, line 43: nit: I think this should be view

jumaffre: As discussed in another comment, the KV uses `term` rather than view so that it's not confused with a `kv::Tx`'s view. Because the KV drives the encryptor, I think it is preferable to use `term` here.

achamayou: If/when we kill off usage of the "view" in "get_view" in the KV and the associated view types, we can do this rename. But until then, we're better off not having two different concepts called "views" in KV.

Another possibility, is to use consensus_view/ConsensusView for this in the KV.

achamayou (approved):

src/kv/encryptor.h
```diff
@@ -2,144 +2,85 @@
...
-     *
-     * @param[in]   plain             Plaintext to encrypt
-     * @param[in]   additional_data   Additional data to tag
-     * @param[out]  serialised_header Serialised header (iv + tag)
-     * @param[out]  cipher            Encrypted ciphertext
-     * @param[in]   version           Version used to retrieve the corresponding
-     * encryption key
-     * @param[in]   is_snapshot       Indicates that the entry is a snapshot (to
-     * avoid IV re-use)
-     */
```
eddyashton, line 97: Why have we lost these docs?

src/node/ledger_secrets.h
```diff
@@ -3,118 +3,283 @@
...
+
+    LedgerSecret(const LedgerSecret& other) :
+      raw_key(other.raw_key),
+      key(std::make_shared<crypto::KeyAesGcm>(other.raw_key))
+    {}
+
+    LedgerSecret(std::vector<uint8_t>&& raw_key_) :
+      raw_key(raw_key_),
+      key(std::make_shared<crypto::KeyAesGcm>(std::move(raw_key_)))
+    {}
```
eddyashton, line 37: Why do we call `make_shared` here - can't we copy-construct/move the original `key` directly?

```suggestion
    LedgerSecret(const LedgerSecret& other) :
      raw_key(other.raw_key),
      key(other.key)
    {}

    LedgerSecret(std::vector<uint8_t>&& raw_key_) :
      raw_key(raw_key_),
      key(std::move(other.key))
    {}
```

src/node/ledger_secrets.h
```diff
@@ -3,118 +3,283 @@
...
+
+  inline LedgerSecret make_ledger_secret()
+  {
+    return LedgerSecret(tls::create_entropy()->random(crypto::GCM_SIZE_KEY));
+  }
+
+  using VersionedLedgerSecret = std::pair<kv::Version, LedgerSecret>;
+  using LedgerSecretsMap = std::map<kv::Version, LedgerSecret>;
+
+  struct VersionedLedgerSecrets
```
eddyashton, line 48: These names are a little confusing. I'd expect `VersionedLedgerSecrets` to be a collection of `VersionedLedgerSecret`s, but its actually a thin wrapper around `LedgerSecretsMap`. Why do we need this wrapper rather than `LedgerSecretsMap` directly? And if we only use `VersionedLedgerSecret` (singular) rarely/internally, `LedgerSecretsMap::value_type` may be clearer?

src/node/ledger_secrets.h
```diff
@@ -3,118 +3,283 @@
...
+      if (search == search_begin)
+      {
+        throw std::logic_error(fmt::format(
+          "kv::TxEncryptor: could not find ledger encryption key for seqno {}",
+          version));
+      }
+      return --search;
+    }
+
+    void take_dependency_on_secrets(kv::Tx& tx)
```
eddyashton, line 99: I think this should be `ReadOnlyTx` and `get_read_only_view`. We don't use this in many places but if we did, we could get `const`-like type-safety by using `ReadOnlyTx` throughout some execution stacks.

src/node/ledger_secrets.h
```diff
@@ -3,118 +3,283 @@
...
       }
-      return secrets_list.back().secret;
+
+      auto latest_ledger_secret = ledger_secrets.secrets.rbegin();
+      return std::make_pair(
+        latest_ledger_secret->first, latest_ledger_secret->second);
     }
 
-    std::optional<LedgerSecret> get_penultimate()
+    std::optional<LedgerSecret> get_penultimate(kv::Tx& tx)
```
eddyashton, line 170: I think we should change this function to fetch both latest and penultimate under the same lock. If a secret is added after fetching one and before fetching the other, we encrypt junk.

src/node/ledger_secrets.h
```diff
@@ -3,118 +3,283 @@
...
+
+    void take_dependency_on_secrets(kv::Tx& tx)
+    {
+      // Ledger secrets are not stored in the KV. Instead, they are
+      // cached in a unique LedgerSecrets instance that can be accessed without
+      // reading the KV. However, it is possible that the ledger secrets are
+      // updated (e.g. rekey tx) concurrently to their access by another tx. To
+      // prevent conflicts, accessing the ledger secrets require access to a tx
+      // object, which must take a dependency on the secrets table.
+      auto v = tx.get_view(secrets_table);
```
eddyashton, line 107:
```suggestion
      auto v = tx.get_view<Secrets>(Tables::SECRETS);
```

If we lookup this table by name, we can remove the ref member, and remove the constructor argument, and remove some of the changes to test cases where this no longer needs to be passed.

src/kv/encryptor.h
jumaffre: Because I re-wrote this class from scratch and forgot to add this back - well spotted! I've added it back.

src/node/ledger_secrets.h
jumaffre: Good point. Changed now.

src/node/ledger_secrets.h
jumaffre: Good point. Done now.

src/node/ledger_secrets.h
jumaffre: Ah! True. I forgot we could do that 👍

src/node/ledger_secrets.h
jumaffre: Done.

src/node/ledger_secrets.h
jumaffre: For the copy constructor, yes. The second one isn't the move constructor so we still need the `make_shared` to create the key.

PR #2076, opened 2021-01-14
Merkle tree fixes

wintersteiger:
Some minor fixes for the Merkle trees and removal of parent pointers for performance.


achamayou:
If possible/not too complicated, it would be nice to have unit tests to describe/lock-in the fixes.

achamayou (approved):

wintersteiger:
That's a good idea; these particular fixes are just about empty trees, but I'll add some doctests anyways.

PR #2077, opened 2021-01-14
Merkle tree fuzzer

wintersteiger:
Here's an example of a fuzzer for the Merkle trees based on LLVM's [libFuzzer](https://www.llvm.org/docs/LibFuzzer.html).

This helped me to find some corner case issues in the Merkle tree API, but I don't think it will be useful much beyond that. So, just an example for now, I don't intend to actually merge this.

### DIFF
--- a/src/libmerklecpp/test/fuzzer/Makefile
+++ b/src/libmerklecpp/test/fuzzer/Makefile
@@ -1,0 +1,11 @@
+all: fuzzer
+
+fuzzer: fuzzer.cpp
+	clang++-8 -g -O1 -fsanitize=fuzzer,address -fno-omit-frame-pointer -fprofile-instr-generate -fcoverage-mapping  $< -o $@
+
+clean:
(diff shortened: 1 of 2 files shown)


achamayou:
Nice! For what it's worth, https://github.com/google/AFL is very good and not too difficult to set up.

wintersteiger:
Yeah, libFuzzer is compatible with AFL. You can run both on the same binary and test input corpus [[1](https://www.llvm.org/docs/LibFuzzer.html#id35)].

achamayou:
@wintersteiger ah that's interesting, I did not know that!

achamayou:
@wintersteiger is it ok to close this, or did you intend to make changes to merge the change into one of the CI jobs?

wintersteiger:
Sure, that's perfectly fine. I did this just to see how and whether it works and it helped me to find some simple/shallow bugs, but I don't think it will be of continued value to run it in the CI.

PR #2083, opened 2021-01-15
Allow redirects by default, but make this overridable

eddyashton:
Hopefully fixing `Test installed CCF` in the Daily CI.

### DIFF
--- a/python/ccf/clients.py
+++ b/python/ccf/clients.py
@@ -638,7 +638,7 @@ class CCFClient:
         headers: Optional[dict] = None,
         timeout: int = DEFAULT_REQUEST_TIMEOUT_SEC,
         log_capture: Optional[list] = None,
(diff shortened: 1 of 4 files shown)


python/ccf/clients.py
```diff
@@ -632,6 +638,7 @@ def call(
         headers: Optional[dict] = None,
         timeout: int = DEFAULT_REQUEST_TIMEOUT_SEC,
         log_capture: Optional[list] = None,
+        allow_redirects=True,
```
jumaffre, line 641: To be added to the docstring below.

jumaffre (approved):

achamayou (approved):

python/ccf/clients.py
eddyashton: Done.

PR #2085, opened 2021-01-15
Add on: pull_request trigger to CodeQL workflow

rneatherway:
From February 2021, in order to provide feedback on pull requests, Code Scanning workflows must be configured with both `push` and `pull_request` triggers. This is because Code Scanning compares the results from a pull request against the results for the base branch to tell you only what has changed between the two.

Early in the beta period we supported displaying results on pull requests for workflows with only `push` triggers, but have discontinued support as this proved to be less robust.

See https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#scanning-pull-requests for more information on how best to configure your Code Scanning workflows.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -8,6 +8,8 @@ name: "CodeQL"
 on:
   push:
     branches: [master]
+  pull_request:
+    branches: [master]
 
 jobs:
   analyze:


achamayou:
Hi @rneatherway, thank you for letting us know. We deliberately do not have CodeQL enabled on PRs because it takes too long to run at the moment.

The main reason it is so slow for us is because we have a fair few dependencies, and the script that installs them takes quite a long time to run. If we could use docker images like we do for CI elsewhere, our CodeQL job would be a lot faster and we would certainly consider running it on every PR!

rneatherway:
Ah, no problem! Apologies for the noise. I tried to filter out those that had deliberately turned off PR analysis but missed your repository.

Issue #1649, opened 2020-09-24
Remove TLS authentication requirement from member frontend


achamayou:
The current member frontend in CCF (ie. RPCs prefixed under '/gov') requires client authentication, and that the client identity be that of one of the members. This requirement makes it difficult to keep member identity keys in HSMs at all times, which is likely to be a requirement for high-privilege member identities.

Governance RPCs fall under two categories: writes, such as proposals, votes, acks, all of which must be signed for offline audit reasons and reads that allow convenient access to public ledger tables. The latter do not require authentication, since the information they return is public, and the former are already authenticated by virtue of being signed.

Removing authentication opens governance to the possibility of replay attacks however, which is particularly dangerous for votes and proposals. To prevent this, CCF will keep an auto-incrementing per-member write count, which the member must include in their (signed) write request.

achamayou:
Summary of discussion with @jumaffre and @sylvanc: a write count isn't sufficient to cover all replay attacks. If a rollback of a proposal and the write counter occurs, a signed vote can be reused for a proposal with different contents.

Two possibilities were discussed:

1. Adding the view to proposal ids. This isn't sufficient either in fact, because the objects that the proposal talks about such as node ids are subject to the same problem, so we would need to introduce the view in every id. It also has very unpleasant consequences on governance auditing, which would now be tied to consensus.
2. Binding the proposal id to the current state (proposal id as the hash of the proposal contents and merkle root so far). This is attractive from an audit perspective, but it's a bit messy to implement (it needs a special transaction like signatures), and id contiguity is lost. I can't remember why contiguity mattered, but I seem to recall from #1378 that there was some interest in it. Unlike the first proposal, this removes the need to modify the ids of the objects described by the proposals.

This requires further discussion before implementation.

achamayou:
Further discussion happened, and consensus is currently that #2 with some nuances is the way to go.

1. Proposal id is a digest of:
  a. The digest of the canonicalised proposal request (as covered by the user signature)
  b. The root of the Merkle Tree as of the read version of the proposal creation transaction
2. The proposal transaction must read from the proposals table at that ID. This is to ensure that two otherwise identical proposals submitted in very close succession do not end up with the same ID, and the KV conflict resolution ends up assigning a different read_version to the second one instead.

This binds proposals to current state completely, so they do not require a per-member counter, since they are inherently non-replayable on anything but the same observed prefix.

[...]

PR #2087, opened 2021-01-18
Replay-immune votes

achamayou:
Draft open for discussion. Final part of #1649.

This is using only a digest of the request, not bound to the root of the tree. There is a difficulty in obtaining past_root(read_version) in a thread-safe way: compaction may have occurred between taking read_version and calling past_root() on the history.

I am not clear on the best way to solve this, my intuition is that it is necessary to create a variant of the initial get_view() call that atomically marks the version in the history as being a compaction watermark, and to release that watermark once past_root() has been called. That's fairly awkward, because the initial get_view() happens deep in our code, before the user transaction. In the case of BFT, it's even worse, the read version is set before we can possibly know this is a "propose", on the initial read from the request table.

Ideas would be most welcome.


eddyashton:
We should add a note in the `CHANGELOG`.

For root retrieval:
- Since we already store a range of recent nodes (for receipt retrieval), maybe we can rely on this rather than additional locking? ie - assume that by the time a proposal gets to this point, the flushed index won't be `MAX_HISTORY_LEN` ahead of this transaction's read index, so the past root will still be retrievable. I don't know what the safety net is when this heuristic fails - I guess if `past_root` says "that's too early", this transaction needs to throw an "I feel like I've conflicted please retry me" exception? But this feels like opening a dangerous situation where the service is so busy (and making constant progress on business transactions!) that you can't open any proposals...
- If we're trying to lock versions in the history so they won't be compacted, my gut says we'd need to do it on every transaction, rather than trying to work out which are proposals before doing any reads. Maybe we can identify proposals early in the execution (ie - just after reading from the requests table in the BFT case), and try to lock a version then (shrinking the gap before we call `past_root` so it hopefully isn't compacted, and only doing the lock for the few transactions that need it), but this would still be best-effort and need error handling like above for when the tree had been flushed in the interim.

I'll keep thinking about this.

achamayou:
- [ ] Add note in `CHANGELOG.md`

Relaxing/changing the history interval is good, but only helps probabilistically as you rightly point out. Conflicting on failure in this case is awkward because there is no guarantee the retry will succeed. In pure CFT, it might be ok to do that as a first pass and then do something more expensive the second time, but that will need to happen every time on backups in BFT.

I will set up a meeting to discuss this further, but please feel free to mention any ideas here too.

eddyashton:
I think there's a simpler answer here. Rather than trying to intercept the first call to `get_view` (where the read version is currently set), we can just add a new call on the `Tx`, immediately after its constructed but before we've done anything else to it, which both _gets the root and sets the read version_. The `Tx`s are now always constructed inside `frontend.h`, when we have an `RpcContext` to query to distinguish proposals. Something like:

```
// In frontend.h, everywhere we call `create_tx()`
auto tx = tables.create_tx();
if (request_needs_iniital_merkle_root(ctx))
{
  tx.fetch_merkle_root(); // Maybe needs to take history as an arg? 
}

...

bool request_needs_initial_merkle_root(...)
{
  return ctx->get_request_path().ends_with("proposal");
}

...

const auto tx_read_merkle_root = tx.get_read_merkle_root();
if (!tx_read_merkle_root.has_value())
{
 // Uh oh
}
proposal_id = fmt::format(...);
```

achamayou:
@eddyashton that sounds good, there are a few implications:

1. We need a lock on history
2. History needs to keep track of the current version, because we don't want a lock that also covers the store to be able to grab this in sync
3. BFT backups need to reset the read version after they've deserialised the request (but they probably do that already?). I don't know how we can make sure they haven't truncated their tree though.

achamayou:
After discussion, let's go with this with the twist that you proposed: add a read dependency on signatures. That way we're guaranteed never to have a signature (and therefore a possible compaction) between our read_version and our commit version. This should hold even on BFT backups, without having to bundle any additional meta data.

achamayou:
I am going to split this change in at least the following items:

- [ ] A bugfix in the existing implementation of the store rollback, which I came across while making the changes
- [ ] Turning the Proposal Ids from integers to digests of the requests alone
- [ ] Using the root of the Merkle Tree at the read version when running in CFT
- [ ] A ticket describing what must be done for BFT, as discussed with @ashamis, specifically
  - [ ] Adding a "flush" flag to the read version for backup re-execution, which will be useful for both signatures and proposals, since they both need to express not just KV state dependencies but also History state dependencies
  - [ ] Adding a mechanism to capture non-deterministic values during primary execution, likely in a KV table, and to send them to secondaries for re-execution. That's one way to capture the merkle tree read version, and the mechanism would be generally useful once we add support for REST requests from CCF etc.

Issue #1944, opened 2020-11-26
Use JSON objects as top-level request/response types in REST API


letmaik:
Related to https://github.com/microsoft/CCF/issues/1918 but not directly mentioned by the guides. Still a good thing to do.

Some built-in endpoints currently return a primitive JSON type as value:
- `POST /ack`: `true`
- `GET|POST /recovery_share`: `"base64share"`

This is uncommon and prevents adding more properties in the future. If there's no good reason against it, the top level should always be an object.

PR #2088, opened 2021-01-18
Return 204 on /ack success instead of JSON true

letmaik:
Addresses first point of https://github.com/microsoft/CCF/issues/1944.

### DIFF
--- a/tests/infra/member.py
+++ b/tests/infra/member.py
@@ -148,7 +148,7 @@ class Member:
         state_digest = self.update_ack_state_digest(remote_node)
         with remote_node.client(*self.auth(write=True)) as mc:
             r = mc.post("/gov/ack", body=state_digest)
-            assert r.status_code == 204, f"Error ACK: {r}"
+            assert r.status_code == http.HTTPStatus.NO_CONTENT, f"Error ACK: {r}"
             self.status_code = MemberStatus.ACTIVE
             return r
 
--- a/tests/infra/member.py
+++ b/tests/infra/member.py
@@ -148,7 +148,7 @@ class Member:
         state_digest = self.update_ack_state_digest(remote_node)
         with remote_node.client(*self.auth(write=True)) as mc:
             r = mc.post("/gov/ack", body=state_digest)
-            assert r.status_code == 200, f"Error ACK: {r}"
+            assert r.status_code == 204, f"Error ACK: {r}"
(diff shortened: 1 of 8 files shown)


achamayou:
@letmaik thank you, that looks good. Since this is a REST API change, would you mind adding an entry to CHANGELOG.md?

We could use https://github.com/openapi-tools/open-api-diff, the markdown diff looks quite reasonable, and may even be generated automatically on release.

letmaik:
> 
> 
> @letmaik thank you, that looks good. Since this is a REST API change, would you mind adding an entry to CHANGELOG.md?

Will do, I always forget.

I just realized that CCF only supports `200` when using `set_auto_schema`. Any suggestions on how to extend that?

achamayou:
Pinging @eddyashton, I think it's just a matter of adding a default argument to set_auto_schema() and using it line 337?

eddyashton:
Yes - what @achamayou should be sufficient.

tests/infra/member.py
```diff
@@ -148,7 +148,7 @@ def ack(self, remote_node):
         state_digest = self.update_ack_state_digest(remote_node)
         with remote_node.client(*self.auth(write=True)) as mc:
             r = mc.post("/gov/ack", body=state_digest)
-            assert r.status_code == 200, f"Error ACK: {r}"
+            assert r.status_code == 204, f"Error ACK: {r}"
```
eddyashton, line 151: We have a mix of int literals and `http.HTTPStatus` values in our Python, but I think we should prefer the latter?

```suggestion
            assert r.status_code == http.HTTPStatus.NO_CONTENT, f"Error ACK: {r}"
```

letmaik:
@achamayou @eddyashton I had to do a bit more plumbing but I think I ended up at something good. If `set_auto_schema` is called with a `void` response schema then it defaults to `204` now.

achamayou:
@letmaik that does seem like a sensible default, but we may still want to support user-defined 2xx, like Created etc in the future. Agree it's out of scope for this change though.

achamayou (approved):

letmaik:
> 
> 
> @letmaik that does seem like a sensible default, but we may still want to support user-defined 2xx, like Created etc in the future. Agree it's out of scope for this change though.

This PR allows specifiying a custom status code. Isn't that what you mean?

PR #2089, opened 2021-01-18
Change schemas of /recovery_share GET/POST endpoints

letmaik:
Addresses the second point of https://github.com/microsoft/CCF/issues/1944.

GET /recovery_share now returns `{"encrypted_share": "..base64"}` while POST /recovery_share has as input `{"share": "..base64"}` and returns `{"message": "status message..."}`.

### DIFF
--- a/src/node/rpc/call_types.h
+++ b/src/node/rpc/call_types.h
@@ -158,19 +158,20 @@ namespace ccf
 
   struct GetRecoveryShare
   {
-    struct RecoveryShare
-    {
-      std::string share = {};
-    };
-
     using In = void;
 
-    using Out = RecoveryShare;
+    struct Out
+    {
+      std::string encrypted_share;
+    };
   };
 
   struct SubmitRecoveryShare
   {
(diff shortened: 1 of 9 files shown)


src/node/rpc/call_types.h
```diff
@@ -155,4 +155,26 @@ namespace ccf
       bool valid = false;
     };
   };
+
+  struct GetRecoveryShare
```
jumaffre, line 159: I believe the `submit_recovery_share.sh` script will need updating.

letmaik: Ah, that means semantically what you get and what you send is not the same actually. One is encrypted, one is plaintext. I'll make the fields `"encrypted_share"` and `"share"`, maybe?

jumaffre: Yes, that seems even better!

letmaik: Done.

achamayou (approved):

PR #2090, opened 2021-01-19
Fix SAN error on rollback during recovery

jumaffre:
At the end of the public recovery, the store is rolled back to discard any unsigned entries. The rollback propagates to the ledger secrets which haven't yet been set, which caused a SAN error as we were accessing the first ledger secret.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Time is a flat circle
+Time is a flat circle.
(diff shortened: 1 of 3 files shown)


eddyashton (approved):

achamayou (approved):

jumaffre:
Force merging this as this is a fix for the failed daily build.

PR #2091, opened 2021-01-19
Do not start the signature timer on the recovery store

achamayou:
That store is a pure replay construct to recover private state, emitting signatures over it neither makes sense nor works.

### DIFF
--- a/src/node/history.h
+++ b/src/node/history.h
@@ -513,14 +513,18 @@ namespace ccf
       NodeId id_,
       tls::KeyPair& kp_,
       size_t sig_tx_interval_ = 0,
-      size_t sig_ms_interval_ = 0) :
+      size_t sig_ms_interval_ = 0,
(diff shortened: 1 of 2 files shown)


src/node/history.h
```diff
@@ -513,14 +513,18 @@ namespace ccf
       NodeId id_,
       tls::KeyPair& kp_,
       size_t sig_tx_interval_ = 0,
```
eddyashton, line 515: Will we still use this interval to try to produce signatures on this `History`? I think if we can get away with it, we should avoid passing any of these timers to this store if it doesn't use them.

eddyashton (approved):

jumaffre (approved):

src/node/history.h
achamayou: No we won't, but I'm not sure what you have in mind?

PR #2094, opened 2021-01-19
List authn policies in JS

### LINKED ISSUE #1775: JS: Expose user details in Request object
Currently JS apps have no access to any user details like the client certificate.

I suggest to extend the `Request` object with a new optional `user` field that is of `any` type. If CCF authentication is enabled (`require_client_identity`/`require_client_signature`) then it would be an object like `{certificate: ...}`.

The reason I don't want to strongly type the `user` field is that this can come in handy with JS-side authentication "middleware", when CCF-side authentication is disabled (I don't see a case when both would be enabled). For example, in tsoa [authentication middleware](https://tsoa-community.github.io/docs/authentication.html) can be provided which returns an arbitrary value. This value is typically set as `user` property in the [request object](https://github.com/lukeautry/tsoa/blob/2e1859f011ddd4e7c2290ea0b84a16f22a79ab44/packages/cli/src/routeGeneration/templates/koa.hbs#L106). In the case of JWT, it could be the decoded token.

eddyashton:
Final part of #2032. Closes #1775.

~~Draft initially for further discussion.~~

Replaces JS endpoints' multiple boolean properties (`require_client_identity`, `require_client_signature`) with a list of policy names (strings). Adds the extracted `Identity` as `request.caller`*.

`named_auth_policies.h` has function to do name lookups both ways - from "`app.json` listed these values in `authn_policies`" to "a `ccf::AuthnPolicy` we can execute", and from "this kind of identity was found by the policy checks" to "a string to tell JS which policy was accepted".

Adds a `multi_auth` endpoint in `js-authentication/src/endpoints.js`, roughly equivalent to the C++ one in `logging.cpp`, to check that we can extract all the fields we think we can.

This doesn't include support for defining custom auth policies in JS. I have an idea how to do that, but I'll hold it until I have more time/a concrete use case.

\* I went with `caller` rather than the previous `user`, because `user` is a slightly more specific term in CCF, and I think we want to re-use as much of this as possible for writing member endpoints in governance. Could be convinced this is unhelpful.


src/node/rpc/endpoint.h
```diff
@@ -68,32 +68,23 @@ namespace ccf
...
     DECLARE_JSON_REQUIRED_FIELDS(
-      EndpointProperties,
-      forwarding_required,
-      execute_outside_consensus,
-      require_client_signature,
-      require_client_identity);
-    DECLARE_JSON_OPTIONAL_FIELDS(
-      EndpointProperties, require_jwt_authentication, openapi, openapi_hidden);
+      EndpointProperties, forwarding_required, authn_policies);
+    DECLARE_JSON_OPTIONAL_FIELDS(EndpointProperties, openapi, openapi_hidden);
```
jumaffre, line 87: Why are these optional? I assume this was because we didn't want to break existing applications when these fields were added but we may want to set them as required now as the changes in this PR are breaking anyway?

src/apps/js_generic/js_generic.cpp
```diff
@@ -781,6 +782,114 @@ namespace ccfapp
...
+          JS_NewString(ctx, get_policy_name_from_ident(empty_ident)));
+        return caller;
+      }
+
+      char const* policy_name = nullptr;
+      size_t id;
+      nlohmann::json data;
+      std::string cert_s;
+
+      if (
```
achamayou, line 839: It would have seemed more natural to follow the order given by the auth list here, in case some of the values end up overwriting each other. Either or separate the output of each policy so they can be fetched separately.

If I specify [sig, cert], and they yield different identities, I will get the cert identity, not because I specified it last but because of this hidden hardcoded order. That doesn't seem right? Am I missing something?

src/apps/js_generic/named_auth_policies.h
```diff
@@ -0,0 +1,72 @@
...
+#pragma once
+
+#include "node/rpc/endpoint_registry.h"
+
+namespace ccfapp
+{
+  static std::shared_ptr<ccf::AuthnPolicy> get_policy_by_name(
+    const std::string& name)
+  {
+    if (name == ccf::UserCertAuthnPolicy::SECURITY_SCHEME_NAME)
```
achamayou, line 13: Can we not have a map/something O(1) here?

src/apps/js_generic/js_generic.cpp
eddyashton: Once the first policy passes, we won't execute any others, so (at-most) one identity is set. Only one of these branches can be executed. The priority is determined by the order the policies are listed, either in `make_endpoint()` or `app.json`.

If you specify `[sig, cert]`, and the sig is valid, you'll get the sig identity. If there's no sig but a valid cert, you'll get the cert identity.

src/apps/js_generic/named_auth_policies.h
eddyashton: Yes, I'll convert this to a map.

achamayou (approved):

PR #2095, opened 2021-01-20
Remove dead code

achamayou:
Remove left-over code from the PBFT days, noticed while making changes to the History. And a few other bits of bobs too, while I'm at it.

### DIFF
--- a/src/node/history.h
+++ b/src/node/history.h
@@ -174,11 +174,7 @@ namespace ccf
       kv::TxHistory::RequestID, kv::Version, const std::vector<uint8_t>&)
     {}
 
-    void add_pending(
-      kv::TxHistory::RequestID,
-      kv::Version,
-      std::shared_ptr<std::vector<uint8_t>>) override
-    {}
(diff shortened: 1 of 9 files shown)


src/node/history.h
```diff
@@ -858,105 +784,9 @@ namespace ccf
...
-          results[id] = {version, root};
-          on_result.value()({id, version, root});
-        }
-      }
-    }
-
-    void add_response(
-      kv::TxHistory::RequestID id,
-      const std::vector<uint8_t>& response) override
+    void append(std::shared_ptr<std::vector<uint8_t>> replicated) override
```
eddyashton, line 787: Nit: No reason to copy this shared_ptr and deal with ref-count changes. Could be `const std::vector<uint8_t>&` for current needs, or at least `const std::shared_ptr...&` if we may want to extend its lifetime in future.

achamayou: Good point!

ashamis (approved):

eddyashton (approved):

PR #2096, opened 2021-01-20
Add JWT support to Perf Client

sidmore:
rpc_tls_client.h changes:
```
std::vector<uint8_t> gen_http_request_internal(
    const std::string& method,
    const CBuffer params,
    const std::string& content_type,
    llhttp_method verb,
    const char* jwt = nullptr)
```

perf_client.h changes:
``` 
app.add_option("--jwt", jwt)
        ->required(false)
```


eddyashton:
I don't see where the `std::string` `jwt` in `perf_client.h` is converted to a `char const*` (`.c_str()`) and passed to any calls to `rpc_tls_client.h` - what am I missing?

sidmore:
> I don't see where the `std::string` `jwt` in `perf_client.h` is converted to a `char const*` (`.c_str()`) and passed to any calls to `rpc_tls_client.h` - what am I missing?

actually there are couple of methods....with gen_request being called...was not sure which all need to be updated. I will update the code accordingly.

sidmore:
> > I don't see where the `std::string` `jwt` in `perf_client.h` is converted to a `char const*` (`.c_str()`) and passed to any calls to `rpc_tls_client.h` - what am I missing?
> 
> actually there are couple of methods....with gen_request being called...was not sure which all need to be updated. I will update the code accordingly.

I have implemented the change to Pass JWT from argument to the Perfclient.

sidmore:
> > > I don't see where the `std::string` `jwt` in `perf_client.h` is converted to a `char const*` (`.c_str()`) and passed to any calls to `rpc_tls_client.h` - what am I missing?
> > 
> > 
> > actually there are couple of methods....with gen_request being called...was not sure which all need to be updated. I will update the code accordingly.
> 
> I have implemented the change to Pass JWT from argument to the Perfclient.

src/clients/rpc_tls_client.h
```diff
@@ -175,15 +183,16 @@ class HttpRpcTlsClient : public TlsClient, public http::ResponseProcessor
       method,
       {body.data(), body.size()},
       http::headervalues::contenttype::MSGPACK,
-      verb);
+      verb,
+      jwt);
```
syhamza, line 187: Seems like the id_token (jwt) is not being set with Bearer scheme? Also, where do you specify that this belongs to Authorization header?

src/clients/rpc_tls_client.h
```diff
@@ -60,7 +60,8 @@ class HttpRpcTlsClient : public TlsClient, public http::ResponseProcessor
     const std::string& method,
     const CBuffer params,
     const std::string& content_type,
-    llhttp_method verb)
+    llhttp_method verb,
+    const char* jwt = nullptr)
```
syhamza, line 64: >jwt [](start = 16, length = 3)

I wonder if you should name this as auth_token or id_token?

syhamza: jwt is an implementation detail. 

---
In reply to: [561288919](https://github.com/microsoft/CCF/pull/2096#discussion_r561288919) [](ancestors = 561288919)

syhamza (approved):
:shipit:

src/perf_client/perf_client.h
```diff
@@ -115,6 +115,8 @@ namespace client
         ->required(true)
         ->check(CLI::ExistingFile);
       app.add_option("--ca", ca_file)->required(true)->check(CLI::ExistingFile);
+      app.add_option("--jwt", jwt)
```
achamayou, line 118: Since this is setting the Bearer header, and isn't jwt-specific, can we make this --bearer or --bearer-token?

achamayou:
@sidmore thank you for making this change, it's an excellent idea to add a test for this. Do you think we could add an instance of the perf_test running this in the CI, so it could be tracked along with other configs by cimetrics? What authority/token combination do you use for testing at the moment?

achamayou:
Could you also run `scripts/ci-checks.sh -f`? That should automatically reformat your changes so they comply with the repo guidelines (you will need to commit the result). To make sure it's worked, you can run `scripts/ci-checks.sh`, there may be some rare cases where the auto formatting fails, but it's generally quite good.

sidmore:
> Could you also run `scripts/ci-checks.sh -f`? That should automatically reformat your changes so they comply with the repo guidelines (you will need to commit the result). To make sure it's worked, you can run `scripts/ci-checks.sh`, there may be some rare cases where the auto formatting fails, but it's generally quite good.

yup @achamayou will do. Thanks for the pointers.

src/clients/rpc_tls_client.h
sidmore: changed to auth_token

achamayou (approved):

PR #2097, opened 2021-01-21
AFT: Async execution on backup

ashamis:
Part of #2055

This is the first set step of parallel execution on the backup. This PR does a bit of ground work in separating append entry deserialization & execution. There are also moved the deserialization logic out of store.h in deserialise.h this makes things a bit cleaner :)

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1073,7 +1073,6 @@ namespace aft
         r.prev_idx);
 
       std::vector<std::tuple<std::unique_ptr<kv::IExecutionWrapper>, kv::Version>> foobar;
-      // TODO: we now know this is safe so start looking from here
       // Finally, deserialise each entry in the batch
       for (Index i = r.prev_idx + 1; i <= r.idx; i++)
       {
@@ -1109,15 +1108,14 @@ namespace aft
           store->deserialise_views_async(entry, consensus_type, public_only);
         if (ds == nullptr)
         {
-          //throw std::logic_error("foobar - we failed to deserialise");
-          // TODO: this is not awesome
-          LOG_INFO_FMT("BBBBBBBB skipping because we failed to apply");
-          break;
+          LOG_DEBUG_FMT("failed to deserialize we failed to apply");
+          send_append_entries_response(
+            r.from_node, AppendEntriesResponseType::FAIL);
+          return;
         }
         foobar.push_back(std::make_tuple(std::move(ds), i));
       }
 
-      LOG_INFO_FMT("starting new loop");
       for (auto& d : foobar)
       {
         auto& [ds, i] = d;
@@ -1130,7 +1128,6 @@ namespace aft
           deserialise_success = ds->Execute();
           if (deserialise_success == kv::DeserialiseSuccess::FAILED)
           {
-            LOG_INFO_FMT("XXXXXXX Execute failed - i:{}", i);
             state->last_idx--;
             ledger->truncate(state->last_idx);
 
@@ -1148,16 +1145,7 @@ namespace aft
               r.from_node, AppendEntriesResponseType::FAIL);
             return;
           }
-          else
-          {
-            LOG_INFO_FMT("YYYYYYY Execute result - {}, i:{}", deserialise_success, i);
-          }
         }
-        else
-        {
-          LOG_INFO_FMT("ds is null at i:{}", i);
-        }
-
 
         for (auto& hook : ds->get_hooks())
         {
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1114,8 +1114,19 @@ namespace aft
         kv::ConsensusHookPtrs hooks;
         if (consensus_type == ConsensusType::BFT)
         {
+            /*
           deserialise_success = store->deserialise_views(
             entry, hooks, public_only, &sig_term, &sig_index, &tx, &sig);
+          */
+          auto r = store->deserialise_views_async(
+            entry, hooks, public_only, &sig_term, &sig_index, &tx, &sig);
+
+          deserialise_success = kv::DeserialiseSuccess::FAILED;
+          if (r != nullptr)
+          {
+            deserialise_success = r->Execute();
+          }
+
         }
         else
         {
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1072,6 +1072,7 @@ namespace aft
         r.idx,
         r.prev_idx);
 
+      // TODO: we now know this is safe so start looking from here
       // Finally, deserialise each entry in the batch
       for (Index i = r.prev_idx + 1; i <= r.idx; i++)
       {
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1119,7 +1119,6 @@ namespace aft
       for (auto& d : foobar)
       {
         auto& [ds, i] = d;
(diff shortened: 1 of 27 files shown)


src/kv/kv_types.h
```diff
@@ -528,6 +528,30 @@ namespace kv
     virtual void swap(AbstractMap* map) = 0;
   };
 
+  class Tx;
+
+  class AbstractExecutionWrapper
+  {
+  public:
+    virtual ~AbstractExecutionWrapper() = default;
+    virtual kv::DeserialiseSuccess Execute() = 0;
```
jumaffre, line 537: `Execute()` -> `execute()`?

src/consensus/aft/test/logging_stub.h
```diff
@@ -222,6 +222,58 @@ namespace aft
       return kv::DeserialiseSuccess::PASS;
     }
 
+    class ExecutionWrapper : public kv::AbstractExecutionWrapper
+    {
+    public:
+      ExecutionWrapper(const std::vector<uint8_t>& data_) : data(data_) {}
+
+      kv::DeserialiseSuccess Execute() override
```
achamayou, line 230: Our naming convention is lower-case, underscore-separated for functions/methods.

src/consensus/aft/raft.h
```diff
@@ -1103,26 +1106,49 @@ namespace aft
...
         {
-          deserialise_success =
-            store->deserialise(entry, hooks, public_only, &sig_term);
+          deserialise_success = ds->Execute();
+          if (deserialise_success == kv::DeserialiseSuccess::FAILED)
+          {
+            state->last_idx--;
+            ledger->truncate(state->last_idx);
+
+            commit_if_possible(state->last_idx);
```
achamayou, line 1135: Why are we calling commit_if_possible here? It seems like we've failed to deserialise, so I'm not sure I understand why we think the commit would have advanced.

src/consensus/aft/raft.h
```diff
@@ -1103,26 +1106,49 @@ namespace aft
...
+
+            // The term may have changed, and we have not have seen a signature
+            // yet.
+            auto lci = last_committable_index();
+            if (r.term_of_idx == aft::ViewHistory::InvalidView)
```
achamayou, line 1140: Is this now duplicated? Also, we don't have a warning for this until clang 10, but I think we're trying to avoid meaningful indentation.

src/consensus/aft/raft.h
```diff
@@ -1103,26 +1106,49 @@ namespace aft
...
+        auto& [ds, i] = ae;
+        state->last_idx = i;
+
+        kv::DeserialiseSuccess deserialise_success =
+          kv::DeserialiseSuccess::FAILED;
+        if (ds != nullptr)
         {
-          deserialise_success =
-            store->deserialise(entry, hooks, public_only, &sig_term);
+          deserialise_success = ds->Execute();
```
achamayou, line 1129: It's probably time to rename deserialise() to apply(), because we're clearly doing much more than deserialisation when the replicas re-execute.

src/consensus/aft/raft_types.h
```diff
@@ -129,20 +105,17 @@ namespace aft
...
 
-    S deserialise_views(
-      const std::vector<uint8_t>& data,
-      kv::ConsensusHookPtrs& hooks,
-      bool public_only = false,
-      kv::Term* term = nullptr,
-      kv::Version* index = nullptr,
-      kv::Tx* tx = nullptr,
-      ccf::PrimarySignature* sig = nullptr) override
+    std::unique_ptr<kv::AbstractExecutionWrapper> deserialise(
```
achamayou, line 108: That does look a lot nicer :)

src/kv/kv_types.h
```diff
@@ -528,6 +528,30 @@ namespace kv
...
+
+  class AbstractExecutionWrapper
+  {
+  public:
+    virtual ~AbstractExecutionWrapper() = default;
+    virtual kv::ApplySuccess execute() = 0;
+    virtual kv::ConsensusHookPtrs& get_hooks() = 0;
+    virtual const std::vector<uint8_t>& get_entry() = 0;
+    virtual kv::Term get_term() = 0;
+    virtual kv::Version get_index()
```
achamayou, line 541: Why do we have stubs in an Abstract class? The test stubs seem to reimplement those as well, so it seems like they could be pure virtual, unless I've missed something?

src/consensus/aft/test/logging_stub.h
```diff
@@ -219,7 +219,59 @@ namespace aft
...
+        return 0;
+      }
+
+      kv::Version get_index() override
+      {
+        return 0;
+      }
+      ccf::PrimarySignature& get_signature() override
+      {
+        throw std::logic_error("3. Not Implemented");
```
achamayou, line 256: What do the numbers mean?

ashamis: Thanks - removing that it was my debugging :)

src/kv/kv_types.h
ashamis: yes good point, I am making it pure virtual

src/kv/store.h
```diff
@@ -759,123 +727,73 @@ namespace kv
...
+            get_history(),
+            get_progress_tracker(),
+            get_consensus(),
+            std::move(data),
+            public_only,
+            v,
+            std::move(changes),
+            std::move(new_maps));
         }
         else if (changes.find(ccf::Tables::NONCES) != changes.end())
```
achamayou, line 759: Do we ever check in these branches that there's nothing in the AE but the tables we're looking for? That seems to be done for signatures, and not elsewhere.

achamayou (approved):

src/kv/store.h
ashamis: In store.h:723 we make sure there is only 1 table in changes

src/consensus/aft/raft.h
```diff
@@ -1128,7 +1128,7 @@ namespace aft
           apply_success = ds->execute();
           if (apply_success == kv::ApplySuccess::FAILED)
           {
-            rollback(last_committable_index());
+            state->last_idx = i - 1;
```
achamayou, line 1131: Why not rollback here, what's the distinction?

src/consensus/aft/raft.h
```diff
@@ -1103,43 +1106,55 @@ namespace aft
...
+            ledger->truncate(state->last_idx);
+            send_append_entries_response(
+              r.from_node, AppendEntriesResponseType::FAIL);
+            return;
+          }
         }
 
-        for (auto& hook : hooks)
+        for (auto& hook : ds->get_hooks())
```
jumaffre, line 1139: Will this segfault if `ds` is `nullptr`?

src/consensus/aft/raft.h
jumaffre: I believe there's no need for this path anymore: simply move `state->last_idx = i;` after this check and only return a failed append entries response (no need for `ledger->truncate(...)` either).

ashamis: As discussed I added a comment what is going here

src/consensus/aft/raft.h
ashamis: ds cannot be nullptr because we check this on line 1110. I am removing the check on 1126 at it is not needed

jumaffre:
It seems that the JS performance test has suffered a hit: 
image
Maybe just a coincidence as the other perf runs look alright but it may be worth running this a few more times to make cimetrics happy?

PR #2098, opened 2021-01-21
Historical queries after rekey

jumaffre:
Part of #1648 

~Because the ledger secrets that have been compacted are not considered for KV encryption/decryption, historical queries prior to a rekey would fail. This is now fixed by passing a special flag down to the decryption path on the historical store (which was already special cased).~

We now consider all ledger secrets on encryption/decryption (even compacted ones). On the hot path however (the version to encrypt/decrypt is later than the version of the latest ledger secret), we use the latest ledger secret directly. The consequence of this is that historical queries now work after rekey(s).

Next: historical queries should work after recoveries.


src/kv/encryptor.h
```diff
@@ -91,6 +91,8 @@ namespace kv
      * @param[out]  plain             Decrypted plaintext
      * @param[in]   version           Version used to retrieve the corresponding
      * encryption key
+     * @param[in]   is_historical     If set, only consider encryption keys
+     * that have not been compacted away (used to decrypt historical queries)
```
achamayou, line 95: I don't understand why we need this flag. Shouldn't we just treat the secrets as a cache? Why do we need to communicate this to decrypt?

src/node/ledger_secrets.h
```diff
@@ -54,28 +54,38 @@ namespace ccf
 
     SpinLock lock;
     LedgerSecretsMap ledger_secrets;
-    std::optional<LedgerSecretsMap::iterator> commit_key_it = std::nullopt;
 
-    LedgerSecretsMap::iterator get_encryption_key_it(kv::Version version)
+    LedgerSecret get_secret_for_version(kv::Version version)
```
achamayou, line 58: Can we have some unit tests for this please?

achamayou (commented):
Looks good in principle, thank you for removing the flag. The lookup would benefit from a small unit test in my opinion.

src/node/ledger_secrets.h
jumaffre: This is already tested via the encryptor's unit test which makes use of this class (see `encryptor_test.cpp`).

src/node/ledger_secrets.h
```diff
@@ -54,28 +54,38 @@ namespace ccf
...
+      if (search == ledger_secrets.begin())
       {
-        throw std::logic_error(fmt::format(
-          "kv::TxEncryptor: could not find ledger encryption key for seqno {}",
-          version));
+        throw std::logic_error(
+          fmt::format("Could not find ledger secret for seqno {}", version));
       }
-      return --search;
+      return (--search)->second;
```
eddyashton, line 88:
```suggestion
      return std::prev(search)->second;
```

I normally prefer `++` and `--` for iterator manipulation, but for some reason the additional lookup here makes it look like dangerous pointer chasing, so `std::prev` would make my eyes happier. Very minor!

eddyashton (approved):

achamayou (approved):

PR #2104, opened 2021-01-22
Use digests for proposal ids, rather than incrementing integers

achamayou:
This is part of #1649, and a chunk of #2087.

Proposal ids are updated to become hex-encoded digest strings, rather than integers. The digest only covers the proposal request itself at the moment, and votes are therefore no more replay-proof than under the previous integer scheme unless the user includes a nonce in their proposal.

Once necessary changes have been made to the history/KV/frontend, this digest will be combined with the root of the transactions' Merkle Tree as of the transaction read_version. Users will no longer need to include a nonce in their proposal to ensure votes cannot be reused.

- [x] Update proposals included in documentation to user new ids

### DIFF
--- a/doc/governance/open_network.rst
+++ b/doc/governance/open_network.rst
@@ -22,7 +22,7 @@ Then, the certificates of trusted users should be registered in CCF via the memb
 
     $ scurl.sh https://<ccf-node-address>/gov/proposals --cacert network_cert --key member0_privk --cert member0_cert --data-binary @add_user.json -H "content-type: application/json"
     {
-        "proposal_id": 5,
+        "proposal_id": "f665047e3d1eb184a7b7921944a8ab543cfff117aab5b6358dc87f9e70278253",
         "proposer_id": 0,
         "state": "OPEN"
     }
@@ -38,9 +38,9 @@ Other members are then allowed to vote for the proposal, using the proposal id r
         }
     }
 
-    $ scurl.sh https://<ccf-node-address>/gov/proposals/5/votes --cacert network_cert --key member1_privk --cert member1_cert --data-binary @vote_accept.json -H "content-type: application/json"
+    $ scurl.sh https://<ccf-node-address>/gov/proposals/f665047e3d1eb184a7b7921944a8ab543cfff117aab5b6358dc87f9e70278253/votes --cacert network_cert --key member1_privk --cert member1_cert --data-binary @vote_accept.json -H "content-type: application/json"
     {
-        "proposal_id": 5,
+        "proposal_id": "f665047e3d1eb184a7b7921944a8ab543cfff117aab5b6358dc87f9e70278253",
         "proposer_id": 0,
         "state": "OPEN"
     }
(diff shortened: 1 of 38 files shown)


src/node/rpc/member_frontend.h
```diff
@@ -554,41 +554,41 @@ namespace ccf
     //! Table of functions that proposal scripts can propose to invoke
     const std::unordered_map<
       std::string,
-      std::function<bool(ObjectId, kv::Tx&, const nlohmann::json&)>>
+      std::function<bool(ProposalId, kv::Tx&, const nlohmann::json&)>>
```
eddyashton, line 557: Minor: Now that these are `std::string` rather than `size_t`, we should (where possible) pass them by `const&`.

doc/governance/open_network.rst
```diff
@@ -38,9 +38,9 @@ Other members are then allowed to vote for the proposal, using the proposal id r
         }
     }
 
-    $ scurl.sh https://<ccf-node-address>/gov/proposals/5/votes --cacert network_cert --key member1_privk --cert member1_cert --data-binary @vote_accept.json -H "content-type: application/json"
+    $ scurl.sh https://<ccf-node-address>/gov/proposals/f665047e3d1eb184a7b7921944a8ab543cfff117aab5b6358dc87f9e70278253/votes --cacert network_cert --key member1_privk --cert member1_cert --data-binary @vote_accept.json -H "content-type: application/json"
```
eddyashton, line 41: How many bytes of this digest do we need to keep? Could we truncate these IDs to ~12 chars so they're more readable in these URLs?
(If we wanted to get fancy I think we could actually do a git-style thing where the public ID is "as few chars as possible while still unique", down to a minimum of N chars, by checking against each existing proposal ID when we generate a new one...)

src/node/rpc/member_frontend.h
achamayou: Good point, will fix.

doc/governance/open_network.rst
achamayou: These IDs need to be properly unique, collisions are a security issue, so it's not obvious to me that there is any scope to truncate them. Unlike git, we very much want not to keep (or access) the whole history of proposals ids and have to scan it to find collisions.

achamayou: Git seems to scan for collisions when picking the short sha: https://lkml.org/lkml/2013/9/30/365, which isn't something we want to have to do. If we don't do that, we increase the chance of collision.

src/node/rpc/member_frontend.h
```diff
@@ -1967,7 +2008,7 @@ namespace ccf
         }
 
         if (!set_jwt_public_signing_keys(
-              ctx.tx, INVALID_ID, parsed.issuer, issuer_metadata, parsed.jwks))
+              ctx.tx, "", parsed.issuer, issuer_metadata, parsed.jwks))
```
eddyashton, line 2011: Nit: Would prefer if `""` was `INVALID_PROPOSAL_ID`, set to something like `INVALID_PROPOSAL_ID = "INVALID"` in `proposals.h`, so we can catch if this ever leaks somewhere unexpected.

eddyashton (approved):

jumaffre (approved):

src/node/rpc/member_frontend.h
achamayou: Ok

letmaik:
> Once necessary changes have been made to the history/KV/frontend, this digest will be combined with the root of the transactions' Merkle Tree as of the transaction read_version. Users will no longer need to include a nonce in their proposal to ensure votes cannot be reused.

@achamayou Is this tracked somewhere?

Issue #274, opened 2019-07-26
CCF's kv should expose Values as well as Maps


eddyashton:
We keep finding situations where we the kv to store a single value of a certain type, rather than a map of several values. The current solution is to create a `Map` from `size_t` to the value type, and only ever access the value at key `0`. This is an ugly paradigm to expose to app writers.

Instead, the kv could expose a `Value` type, parameterised only by a `TValue` and no `TKey`. Under the hood, we can implement this as a wrapper around a `Map`, either keyed by a `size_t` or a single-value struct.

eddyashton:
I also found a case where I really wanted a _set_, where I only cared about the existence of keys and not their corresponding values. I think we can support this similarly, with a `kv::Set<T>` type wrapping a `kv::Map<T, Mono>` (cf `kv::Value<T>` as `kv::Map<Mono, T>`), where `Mono` is probably a handmade single-valued type or perhaps `nullptr_t`.

achamayou:
@eddyashton looking at this in the post-flat-KV and JS apps world, it seems to me that:

1. This can be implemented completely in user space as either a C++ or JS wrapper (or indeed, both).
2. There is not much we can do at the framework level to support this, if we really want to keep serialisation user-defined and not tag values.

So I'm tempted to conclude that resolving this issue is a matter of proving (and using) a sample wrapper? Am I missing something?

PR #2109, opened 2021-01-25
Replace TxView with MapHandle

eddyashton:
The use of the typename `TxView` in the KV conflicts with `View` in our consensus terminology, and makes code which has to deal with both (ie - endpoints which return consensus state, or trying to write consensus state to the KV) difficult to read. This renames `TxView` to `MapHandle`, to remove this conflict.

This touches a lot of files, and I've attempted to keep variable names and comments up-to-date where I've made this rename. Most of these changes are the function `Tx::get_view` to `Tx::get_handle`. I'm not set on `get_handle` - I'd prefer `get_map` (with half an eye on #274 then offering `get_value`, `get_set`) but we already use `get_map` for a few other things, notably "ask the Store if this Map already exists".

~~There was also a suggestion of using `operator[]`, which is tempting but has a few issues to resolve:~~

~~1) Is it too hard to grep? I think `tx[` would find all the instances in real code, but we'd need to dig to find the few remaining cases in unit tests. And going back to a named method from an operator, if we change our mind, is hard.
2) How do we retrieve read-only handles? Currently we have named methods `ReadOnlyTx::get_read_only_handle` and `Tx::get_handle`. Perhaps `ReadOnlyTx::operator[]` returns read-only handles, and `Tx` gets `ReadOnlyTx& read_only() { return this; }`, so some calls look like `auto read_only_handle = tx.read_only[foo]`.
3) You need a `Map` instance to retrieve handles. There's no equivalent of `tx.get_handle<TMap>("map_name")`, since you can't (afaict?) specify a template-argument to `operator[]`. So instead you need a concrete `TMap map_def("map_name")` somewhere (which is what we were trying to get away from), or at least the odd-looking inline construction `auto handle = tx[TMap("map_name")]`.~~

I've gone with `tx.rw` as a concise-but-greppable replacement for `tx.get_view`, with `.ro` and `.wo` to get read-only and write-only handles. This PR also adds write-only handles*, and updates a bunch of docs to makes the user-visible KV doxygen useful.

\* But _not_ write-only `Tx`s, because; (a) I think I've established that we can't actually construct a `ReadOnlyTx` for the lifetime of a real user request, so don't expect we would use a `WriteOnlyTx`, and (b) the inheritance pattern makes this non-trivial. Maybe later, whif we need it.


achamayou:
Though the issues with using `operator[]` are good points, I'm of the opinion that we should still try to go short, because users neither expect nor care about the intermediary type. I cannot think of a KV store interface that does expose one. So I would suggest `.rw()` and `.ro()` instead of `.get_handle()` and `.get_read_only_handle()`. Digression: is there a use-case for a `.wo()`, when you want to write but be explicit/accident-proof about not having a read dependency?

As far as #273 is concerned, it seems to me that both extension types can be done purely as wrappers and I don't find it obvious that making them a first class part of the API would be beneficial.

eddyashton:
> So I would suggest `.rw()` and `.ro()` instead of `.get_handle()` and `.get_read_only_handle()`.

These work for me.

> Digression: is there a use-case for a `.wo()`, when you want to write but be explicit/accident-proof about not having a read dependency?

Yeah, I'll add this for symmetry with read-only, getting type-safety that keeps you within expected/required dependencies.

doc/build_apps/kv/kv_how_to.rst
```diff
@@ -17,88 +17,109 @@ If a ``Map`` with the given name did not previously exist, it will be created in
...
 
 A single ``Transaction`` (``tx``) is passed to each endpoint of an application and should be used to interact with the Key-Value ``Store``.
 
 When the end-point successfully completes, the node on which the end-point was triggered attempts to commit the transaction to apply the changes to the Store. Once the transaction is committed successfully, it is automatically replicated by CCF and should globally commit.
 
-For each ``Map`` that a Transaction wants to write to or read from, a :cpp:class:`kv::Map::TxView` should first be acquired. This may be retrieved either purely by name (in which case the desired type must be explicitly specified as template parameters), or by using a ``Map`` which contains both the name and the desired types.
+For each :cpp:class:`kv::Map` that a transaction wants to write to or read from, a :cpp:class:`kv::MapHandle` must first be acquired. These are acquired from the :cpp:func:`kv::Tx::rw` (`read-write`) method. These may be axquired either by name (in which case the desired type must be explicitly specified as a template parameter), or by using a :cpp:class:`kv::Map` instance which defines both the map's name and key-value types.
```
jumaffre, line 29:
```suggestion
For each :cpp:class:`kv::Map` that a transaction wants to write to or read from, a :cpp:class:`kv::MapHandle` must first be acquired. These are acquired from the :cpp:func:`kv::Tx::rw` (`read-write`) method. These may be acquired either by name (in which case the desired type must be explicitly specified as a template parameter), or by using a :cpp:class:`kv::Map` instance which defines both the map's name and key-value types.
```

src/node/genesis_gen.h
```diff
@@ -64,36 +64,35 @@ namespace ccf
...
     {
-      auto cv = tx.get_view(tables.consensus);
+      auto cv = tx.rw(tables.consensus);
       cv->put(0, consensus_type);
     }
 
     auto get_active_recovery_members()
     {
-      auto members_view = tx.get_view(tables.members);
+      auto members = tx.rw(tables.members);
```
jumaffre, line 92:
```suggestion
      auto members = tx.ro(tables.members);
```

src/node/genesis_gen.h
```diff
@@ -220,7 +218,7 @@ namespace ccf
 
     std::optional<MemberInfo> get_member_info(MemberId member_id)
     {
-      auto m = tx.get_view(tables.members);
+      auto m = tx.rw(tables.members);
```
jumaffre, line 221:
```suggestion
      auto m = tx.ro(tables.members);
```

src/node/genesis_gen.h
```diff
@@ -290,11 +290,10 @@ namespace ccf
       // self_to_exclude is not included in the list of returned nodes.
       std::map<NodeId, NodeInfo> active_nodes;
 
-      auto [nodes_view, secrets_view] =
-        tx.get_view(tables.nodes, tables.secrets);
+      auto nodes = tx.rw(tables.nodes);
```
jumaffre, line 293:
```suggestion
      auto nodes = tx.ro(tables.nodes);
```

src/node/genesis_gen.h
```diff
@@ -310,19 +309,19 @@ namespace ccf
...
-      service_view->put(0, {network_cert, ServiceStatus::OPENING});
+      auto service = tx.rw(tables.service);
+      service->put(0, {network_cert, ServiceStatus::OPENING});
     }
 
     bool is_service_created()
     {
-      auto service_view = tx.get_view(tables.service);
-      return service_view->get(0).has_value();
+      auto service = tx.rw(tables.service);
```
jumaffre, line 318:
```suggestion
      auto service = tx.ro(tables.service);
```

src/node/genesis_gen.h
```diff
@@ -351,15 +350,15 @@ namespace ccf
...
+      service->put(0, active_service.value());
 
       return true;
     }
 
     std::optional<ServiceStatus> get_service_status()
     {
-      auto service_view = tx.get_view(tables.service);
-      auto active_service = service_view->get(0);
+      auto service = tx.rw(tables.service);
```
jumaffre, line 360:
```suggestion
      auto service = tx.ro(tables.service);
```

src/node/genesis_gen.h
```diff
@@ -410,20 +409,20 @@ namespace ccf
...
+      nodes->put(node_id, node_info.value());
 
       LOG_INFO_FMT("Node {} is now {}", node_id, node_info->status);
     }
 
     auto get_last_signature()
     {
-      auto sig_view = tx.get_view(tables.signatures);
-      return sig_view->get(0);
+      auto sig = tx.rw(tables.signatures);
```
jumaffre, line 419:
```suggestion
      auto sig = tx.ro(tables.signatures);
```

src/node/genesis_gen.h
```diff
@@ -492,20 +491,20 @@ namespace ccf
...
+      config->put(0, {threshold});
       return true;
     }
 
     size_t get_recovery_threshold()
     {
-      auto config_view = tx.get_view(tables.config);
-      auto config = config_view->get(0);
-      if (!config.has_value())
+      auto config = tx.rw(tables.config);
```
jumaffre, line 500:
```suggestion
      auto config = tx.ro(tables.config);
```

src/node/historical_queries.h
```diff
@@ -95,8 +95,8 @@ namespace ccf::historical
       const StorePtr& sig_store)
     {
       auto tx = sig_store->create_tx();
-      auto sig_view = tx.get_view<ccf::Signatures>(ccf::Tables::SIGNATURES);
-      return sig_view->get(0);
+      auto signatures = tx.rw<ccf::Signatures>(ccf::Tables::SIGNATURES);
```
jumaffre, line 98:
```suggestion
      auto signatures = tx.ro<ccf::Signatures>(ccf::Tables::SIGNATURES);
```

src/node/historical_queries.h
```diff
@@ -106,8 +106,8 @@ namespace ccf::historical
       // makes no check that the signing node was active at the point it
       // produced this signature
       auto tx = source_store.create_tx();
-      auto nodes_view = tx.get_view<ccf::Nodes>(ccf::Tables::NODES);
-      return nodes_view->get(node_id);
+      auto nodes = tx.rw<ccf::Nodes>(ccf::Tables::NODES);
```
jumaffre, line 109:
```suggestion
      auto nodes = tx.ro<ccf::Nodes>(ccf::Tables::NODES);
```

src/node/history.h
```diff
@@ -619,9 +619,10 @@ namespace ccf
       kv::Term* term = nullptr, PrimarySignature* signature = nullptr) override
     {
       auto tx = store.create_tx();
-      auto [sig_tv, ni_tv] = tx.template get_view<ccf::Signatures, ccf::Nodes>(
-        ccf::Tables::SIGNATURES, ccf::Tables::NODES);
-      auto sig = sig_tv->get(0);
+      auto signatures =
+        tx.template rw<ccf::Signatures>(ccf::Tables::SIGNATURES);
```
jumaffre, line 623:
```suggestion
        tx.template ro<ccf::Signatures>(ccf::Tables::SIGNATURES);
```

src/node/history.h
```diff
@@ -619,9 +619,10 @@ namespace ccf
...
+      auto nodes = tx.template rw<ccf::Nodes>(ccf::Tables::NODES);
```
jumaffre, line 624:
```suggestion
      auto nodes = tx.template ro<ccf::Nodes>(ccf::Tables::NODES);
```

src/node/quote.h
```diff
@@ -109,8 +109,8 @@ namespace ccf
     {
       auto code_digest = get_digest_from_parsed_quote(parsed_quote);
 
-      auto codeid_view = tx.get_view(code_ids_table);
-      auto code_id_status = codeid_view->get(code_digest);
+      auto code_ids = tx.rw(code_ids_table);
```
jumaffre, line 112:
```suggestion
      auto code_ids = tx.ro(code_ids_table);
```

src/node/rpc/frontend.h
```diff
@@ -122,8 +122,8 @@ namespace ccf
         {
           NodeId primary_id = consensus->primary();
           auto tx = tables.create_tx();
-          auto nodes_view = tx.get_view<Nodes>(Tables::NODES);
-          auto info = nodes_view->get(primary_id);
+          auto nodes = tx.rw<Nodes>(Tables::NODES);
```
jumaffre, line 125:
```suggestion
          auto nodes = tx.ro<Nodes>(Tables::NODES);
```

src/node/rpc/frontend.h
```diff
@@ -444,8 +444,8 @@ namespace ccf
       std::lock_guard<SpinLock> mguard(open_lock);
       if (!is_open_)
       {
-        auto sv = tx.get_view<Service>(Tables::SERVICE);
-        auto s = sv->get_globally_committed(0);
+        auto service = tx.rw<Service>(Tables::SERVICE);
```
jumaffre, line 447:
```suggestion
        auto service = tx.ro<Service>(Tables::SERVICE);
```

jumaffre (approved):

achamayou (approved):

achamayou:
@eddyashton @letmaik what do we think of adjusting the JS API to match?

doc/build_apps/kv/kv_how_to.rst
```diff
@@ -17,88 +17,109 @@ If a ``Map`` with the given name did not previously exist, it will be created in
...
+    auto map1_handle_wo = tx.wo(map_priv);
+
+    // Write to that handle
+    map1_handle_wo->put("key1", "value2");
+
+    // Reads are blocked at compile time
+    map1_handle_wo->has("key1"); // Does not compile
+    map1_handle_wo->get("key1"); // Does not compile
+
+Note that, as in the sample above, it is possible to acquire different kinds of handle at different points within your transaction's execution. So if you need to read in one location and write in another you can retrieve multiple distinct handles and get local type-safety, while the resulting transaction correctly handles all reads and writes made.
```
letmaik, line 122:
```suggestion
Note that, as in the sample above, it is possible to acquire different kinds of handles at different points within your transaction's execution. So if you need to read in one location and write in another you can retrieve multiple distinct handles and get local type-safety, while the resulting transaction correctly handles all reads and writes made.
```

src/apps/js_generic/js_generic.cpp
```diff
@@ -675,7 +674,7 @@ namespace ccfapp
 
     auto arg = (JSModuleLoaderArg*)opaque;
 
-    const auto modules = arg->tx->get_view(arg->network->modules);
+    const auto modules = arg->tx->rw(arg->network->modules);
```
letmaik, line 677:
```suggestion
    const auto modules = arg->tx->ro(arg->network->modules);
```

src/apps/js_generic/js_generic.cpp
```diff
@@ -940,7 +939,7 @@ namespace ccfapp
     {
       const auto local_method = method.substr(method.find_first_not_of('/'));
 
-      const auto scripts = args.tx.get_view(this->network.app_scripts);
+      const auto scripts = args.tx.rw(this->network.app_scripts);
```
letmaik, line 942:
```suggestion
      const auto scripts = args.tx.ro(this->network.app_scripts);
```

src/apps/js_generic/js_generic.cpp
```diff
@@ -1292,13 +1291,13 @@ namespace ccfapp
       const auto method = fmt::format("/{}", rpc_ctx.get_method());
       const auto verb = rpc_ctx.get_request_verb();
 
-      auto endpoints_view =
-        tx.get_view<ccf::endpoints::EndpointsMap>(ccf::Tables::ENDPOINTS);
+      auto endpoints =
+        tx.rw<ccf::endpoints::EndpointsMap>(ccf::Tables::ENDPOINTS);
```
letmaik, line 1295:
```suggestion
        tx.ro<ccf::endpoints::EndpointsMap>(ccf::Tables::ENDPOINTS);
```

src/node/rpc/member_frontend.h
```diff
@@ -184,7 +184,7 @@ namespace ccf
   private:
     Script get_script(kv::Tx& tx, std::string name)
     {
-      const auto s = tx.get_view(network.gov_scripts)->get(name);
+      const auto s = tx.rw(network.gov_scripts)->get(name);
```
letmaik, line 187:
```suggestion
      const auto s = tx.ro(network.gov_scripts)->get(name);
```

letmaik:
> @eddyashton @letmaik what do we think of adjusting the JS API to match?

I would leave it as it is for simplicity, mostly because it's convenient to initialize all your handles once in a class constructor for example, and I have a feeling most people would just use `rw` for convenience. There's already the `"readonly"` option in `app.json` which at least exposes all-or-nothing read-only vs read/write for specific endpoints.

PR #2110, opened 2021-01-25
Fix ledger chunks discrepancy between primary and backups

### LINKED ISSUE #1608: End-to-end testing of ledger between nodes

### LINKED ISSUE #2107: Secondaries chunk on every signature after the initial chunk range has elapsed
Observed by @syhamza and @lynshi. This seems to be caused by deserialise() calling requires_snapshot https://github.com/microsoft/CCF/blob/2ad841ac616c5f9359b39ed9b51370f483ed1544/src/node/snapshotter.h#L230 and next_snapshot_indices not being reset correctly.

jumaffre:
Fix #2107
Fix #1608 

This PR fixes the discrepancy of ledger chunk sizes between the primary and backup nodes observed in #2107, and adds support for ledger consistency in our end-to-end test infra by implementing #1608 (i.e. checking that all committed ledger chunks on a stopped node are the same that on the primary).

The main issues were:
- On a non-recovery network, only the primary node was made aware of when to snapshot (and thus chunk the ledger). 
- In recovery, nodes were not keeping track on when the previous network had generated snapshots, which would cause different ledger frequencies on different nodes.

### DIFF
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -540,6 +540,7 @@ namespace ccf
               last_recovered_signed_idx =
                 resp.network_info.last_recovered_signed_idx;
               setup_recovery_hook(startup_snapshot_info != nullptr);
+              snapshotter->set_snapshot_generation(false);
 
               sm.advance(State::partOfPublicNetwork);
             }
@@ -874,6 +875,7 @@ namespace ccf
 
       // Set snapshot interval but do not actually create snapshots!
       snapshotter->init_after_public_recovery();
+      snapshotter->set_snapshot_generation(false);
 
       kv::Version index = 0;
       kv::Term view = 0;
@@ -1001,7 +1003,7 @@ namespace ccf
       encryptor->disable_recovery();
 
       // Snapshots are only generated after recovery is complete
-      // TODO: Enable snapshot generation
+      snapshotter->set_snapshot_generation(true);
 
       // Open the service
       if (consensus->is_primary())
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -163,7 +163,6 @@ namespace ccf
     consensus::Index last_recovered_signed_idx = 1;
     std::list<RecoveredLedgerSecret> recovery_ledger_secrets;
     consensus::Index ledger_idx = 0;
-    // size_t recovery_snapshot_tx_interval = Snapshotter::max_tx_interval;
 
     struct StartupSnapshotInfo
     {
@@ -243,8 +242,7 @@ namespace ccf
       to_host(writer_factory.create_writer_to_outside()),
       network(network),
       rpcsessions(rpcsessions),
-      share_manager(share_manager),
-      snapshotter(std::make_shared<Snapshotter>(writer_factory, network))
+      share_manager(share_manager)
     {}
 
     //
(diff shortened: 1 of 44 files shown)


src/node/node_state.h
```diff
@@ -769,6 +762,12 @@ namespace ccf
         {
           startup_snapshot_info->is_evidence_committed = true;
         }
+
+        // Inform snapshotter of all signature entries so that this node can
+        // continue generating snapshots at the correct interval once the
+        // recovery is complete
+        snapshotter->requires_snapshot(ledger_idx);
```
achamayou, line 769: requires_snapshot() doesn't seem like the right method name here...

src/node/snapshotter.h
```diff
@@ -158,8 +173,11 @@ namespace ccf
       next_snapshot_indices.push_back(last_snapshot_idx);
     }
 
-    void snapshot(consensus::Index idx)
+    void snapshot(consensus::Index idx, bool generate_snapshot)
```
achamayou, line 176: `snapshot()` doesn't seem to cover what this does, which isn't to create a snapshot most of the time. `update()` is maybe more accurate.

src/node/snapshotter.h
```diff
@@ -173,32 +191,47 @@ namespace ccf
...
-        msg->data.self = shared_from_this();
-        msg->data.snapshot = network.tables->snapshot(idx);
-
         last_snapshot_idx = idx;
 
-        static uint32_t generation_count = 0;
-        threading::ThreadMessaging::thread_messaging.add_task(
-          threading::ThreadMessaging::get_execution_thread(generation_count++),
-          std::move(msg));
+        // Snapshots are only generated on primary node
```
achamayou, line 196: Either we pass in is_primary, and that combined with the flag decides what happens and the comment is fair. Or we decide that we abstract away the consensus and we pass `should_emit`, and this comment should not mention anything to do with the primary.

tests/infra/network.py
```diff
@@ -483,14 +483,62 @@ def recover(self, args):
     def ignore_errors_on_shutdown(self):
         self.ignoring_shutdown_errors = True
 
+    def stop_node(self, node, primary=None):
+        if primary is None:
+            primary, _ = self.find_primary()
```
achamayou, line 488: This doesn't seem ideal, because we may not always know who's primary on shutdown, and the error cases where we don't may be interesting anyway. This is also done while other nodes are still running, and is potentially incomplete as a result.

I think we'd be better off doing this once all the nodes are stopped, by taking the longest ledger and checking everything else is a subset. Errors will only show up if there's a problem in the logic, or a fork.

tests/infra/path.py
```diff
@@ -86,6 +87,14 @@ def copy_dir(src_path, dst_path):
     copy2(src_path, dst_path)
 
 
+def compute_file_checksum(file_name):
```
achamayou, line 90: fwiw, https://pypi.org/project/filehash/ can do a directory at once with a pattern filter.

tests/infra/network.py
jumaffre: As discussed, this is now addressed by selecting the node with the longest ledger and comparing other nodes against that instead.

src/node/node_state.h
jumaffre: Updated to `record_committable`.

src/node/snapshotter.h
jumaffre: Good point. Updated to `update()`.

tests/infra/path.py
jumaffre: I haven't used this just yet but I'm happy to looking into this further and improve the verification logic in a separate issue!

achamayou (approved):

PR #2111, opened 2021-01-25
Keep accurate term in history on the primary

achamayou:
Next part of #2087

This is necessary so the frontend can atomically capture a (term, version, root) and initialise the Tx with an up to date (term, read_version) and the corresponding merkle tree root.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1858,7 +1858,7 @@ namespace aft
     {
       election_index = last_committable_index();
       LOG_DEBUG_FMT(
-        "Election index is {} in term", election_index, state->current_view);
+        "Election index is {} in term {}", election_index, state->current_view);
       // Discard any un-committable updates we may hold,
       // since we have no signature for them. Except at startup,
       // where we do not want to roll back the genesis transaction.
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1857,7 +1857,8 @@ namespace aft
     void become_leader()
     {
       election_index = last_committable_index();
-      LOG_DEBUG_FMT("Election index is {}", election_index);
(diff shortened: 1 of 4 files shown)


src/consensus/aft/raft.h
```diff
@@ -1857,7 +1857,8 @@ namespace aft
     void become_leader()
     {
       election_index = last_committable_index();
-      LOG_DEBUG_FMT("Election index is {}", election_index);
+      LOG_DEBUG_FMT(
+        "Election index is {} in term", election_index, state->current_view);
```
ashamis, line 1861: I think you want a second {} somewhere

src/kv/store.h
```diff
@@ -549,14 +542,24 @@ namespace kv
             compacted));
         }
 
+        // The term should always be updated on rollback() when passed
+        // regardless of whether version needs to be updated or not
+        if (t.has_value())
```
ashamis, line 547: nit: i think our coding style states braces with if statements

ashamis (approved):

src/consensus/aft/raft.h
achamayou: Yes, thank you.

src/kv/store.h
achamayou: Yes, absolutely.

src/kv/store.h
```diff
@@ -599,6 +610,9 @@ namespace kv
     {
       std::lock_guard<SpinLock> vguard(version_lock);
       term = t;
+      auto h = get_history();
+      if (h)
+        h->set_term(term);
```
eddyashton, line 615:
```suggestion
      if (h)
      {
        h->set_term(term);
      }
```

eddyashton (approved):

PR #2123, opened 2021-01-27
Hash harmonisation and signature exploration. Fixes #2106.

### LINKED ISSUE #2106: Unify hashing
Since Merkle Trees now use OpenSSL's SHA-256 by default, we should also be able to use this hash in other parts of CCF. This will require some refactoring so that there's only one hashing module in CCF and whose underlying implementation can be changed at build time.

Note: we should have a look at signature verification performance of OpenSSL vs. mbedTLS when looking into this

wintersteiger:
Refactored hashes. This implementation lets us use both mbedTLS and OpenSSL at runtime, which may be unnecessary. Also, I think I should kill all the `#ifdef HAVE_OPENSSL` since it's always available in the enclave and it's de-facto a requirement for the host too.

Added signature benchmarks; OpenSSL's secp384r1 signatures appear to be ~4x faster than mbedTLS. The benchmark runs only on the host though; I'd really like to run these benchmarks in an enclave to see what kind of difference it makes.

Does not contain a refactoring of `KeyPair` yet, but that should be the next step.


achamayou:
@wintersteiger that's awesome, the best comparison would be the SB_sig_SGX_CFT test which uses client signatures, that's our most signature-heavy end to end benchmark. The jwt benchmarks are also interesting for similar reasons.

Micro-benchmarks are good too, but positive impact on the end to end numbers is what matters most ultimately.

src/crypto/hash.cpp
```diff
@@ -23,57 +26,93 @@ namespace crypto
     mbedtls_sha256_free(&ctx);
   }
 
-  class MBSha256HashImpl
+#ifdef HAVE_OPENSSL
```
achamayou, line 29: As you pointed out, we can get rid of those, we depend on a version of Open Enclave (>= 0.13) that guarantees this.

wintersteiger:
I'm opening this one for review, although it's probably easier if I walk you through on Monday. The default `KeyPair` is still mbedTLS's because I don't have an OpenSSL version of `KeyExchangeContext` yet, but I'm not sure whether we want that at this point anyways.

wintersteiger:
Basically done, ready for review. The one test failure is a performance problem. The results are correct, that test just take a lot longer now. I think it picks up the mbedTLS secp256r1 instead of the one from OpenSSL, but not sure yet.

jumaffre:
The `sb_sig_sgx_cft` performance test seems to be timing out, even though it is using the fast curve (`secp256r1`). Is this because it is running with the mbedtls implementation?

python/ccf/clients.py
```diff
@@ -355,7 +349,7 @@ def request(self, request, timeout=DEFAULT_REQUEST_TIMEOUT_SEC):
 
 class TlsAdapter(HTTPAdapter):
```
jumaffre, line 350: I believe this class was only used to tweak the Python requests client to use the `secp256k1` curve. Is it still necessary to have when connection to a `secp256r1` node?

src/crypto/hash.h
```diff
@@ -4,23 +4,137 @@
...
   class Sha256Hash
   {
   public:
     static constexpr size_t SIZE = 256 / 8;
-    Sha256Hash();
-    Sha256Hash(const CBuffer& data);
+    Sha256Hash() : h{0} {}
+    Sha256Hash(const CBuffer& data) : h{0}
+    {
+      // const auto info = mbedtls_md_info_from_type(MBEDTLS_MD_SHA256);
```
jumaffre, line 122: Should go?

src/host/main.cpp
```diff
@@ -422,6 +422,14 @@ int main(int argc, char** argv)
     ->capture_default_str()
     ->check(CLI::NonexistentPath);
 
+  CurveID curve_id = CurveID::SECP384R1;
+  std::vector<std::pair<std::string, CurveID>> curve_id_map = {
+    {"secp384r1", CurveID::SECP384R1}, {"secp256r1", CurveID::SECP256R1}};
+  app
+    .add_option("--curve-id", curve_id, "Elliptic curve to use for signatures")
```
jumaffre, line 429:
```suggestion
    .add_option("--curve-id", curve_id, "Elliptic curve to use for node's certificate (used for TLS and ledger signatures")
```

src/tls/hash.h
```diff
@@ -2,91 +2,14 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
```
jumaffre, line 3: Should this file go?

src/node/node_state.h
```diff
@@ -119,8 +119,9 @@ namespace ccf
     static constexpr NodeId invalid_node_id = -1;
     NodeId self = invalid_node_id;
     tls::KeyPairPtr node_sign_kp;
-    tls::KeyPairPtr node_encrypt_kp;
+    std::shared_ptr<tls::KeyPair_mbedTLS> node_encrypt_kp;
```
jumaffre, line 122: I've probably missed it but should we define the following and have it configurable from `cmake` so that we can easily switch between provider?

```C++
#ifdef CRYPTO_PROVIDER_IS_MBEDTLS
    using KeyPair = KeyPair_mbedTLS;
#else
    using KeyPair = KeyPair_openSSL;
#endif
```

wintersteiger: Sure, makes sense!

python/ccf/clients.py
wintersteiger: Good point. I'm not completely sure, but I'll see if all the tests still pass after I remove it.

wintersteiger:
Got it! The verifier was using mbedTLS. Fixing that also reminded me that I should add an RSAKeyPair for OpenSSL, but we don't have to have that immediately.

wintersteiger:
All checks passed, but there seem to be a couple more performance issues that I don't think are noise.

src/node/node_state.h
```diff
@@ -236,11 +237,12 @@ namespace ccf
...
       std::shared_ptr<enclave::RPCSessions> rpcsessions,
-      ShareManager& share_manager) :
+      ShareManager& share_manager,
+      const CurveID& curve_id) :
       sm(State::uninitialized),
       self(INVALID_ID),
-      node_sign_kp(tls::make_key_pair()),
-      node_encrypt_kp(tls::make_key_pair()),
+      node_sign_kp(tls::make_key_pair(curve_id)),
+      node_encrypt_kp(std::make_shared<tls::KeyPair_mbedTLS>(curve_id)),
```
jumaffre, line 245: This still seems hardcoded to mbedTLS, as opposed to a generic compile-time configurable type?

wintersteiger: Yes, that's on purpose. This particular key pair is used in key_exchange, which we decided not to port to OpenSSL yet. It's pretty much the only thing left though - maybe I should take the time and port that too.

src/crypto/hash.h
```diff
@@ -4,23 +4,128 @@
...
+      return nullptr;
+    }
+
+    virtual HashBytes Hash(const uint8_t* data, size_t size, MDType type) const
+    {
+      HashBytes r;
+      unsigned int len = 0;
+      auto o_md_type = get_md_type(type);
+
+      r.resize(EVP_MD_size(o_md_type));
```
achamayou, line 104: I couldn't work out from a quick search if this was guaranteed to make a single heap allocation of the right size, or if line 100 would allocate a default, and then attempt to realloc on 104. I'd be tempted to create r only on line 104 with the right size straight away, just in case it's the latter, but I'm hopefully needlessly paranoid.

src/enclave/enclave.h
```diff
@@ -86,10 +90,22 @@ namespace enclave
       logger::config::msg() = AdminMessage::log_msg;
       logger::config::writer() = writer_factory.create_writer_to_outside();
 
+      // From
+      // https://software.intel.com/content/www/us/en/develop/articles/how-to-use-the-rdrand-engine-in-openssl-for-random-number-generation.html
+      if (
+        ENGINE_load_rdrand() != 1 ||
+        (rdrand_engine = ENGINE_by_id("rdrand")) == nullptr ||
+        ENGINE_init(rdrand_engine) != 1 ||
+        ENGINE_set_default(rdrand_engine, ENGINE_METHOD_RAND) != 1)
```
achamayou, line 99: If ENGINE_set_default() fails, are we leaking rdrand_engine?

src/host/main.cpp
```diff
@@ -422,6 +422,18 @@ int main(int argc, char** argv)
...
     ->check(CLI::NonexistentPath);
 
+  CurveID curve_id = CurveID::SECP384R1;
+  std::vector<std::pair<std::string, CurveID>> curve_id_map = {
+    {"secp384r1", CurveID::SECP384R1}, {"secp256r1", CurveID::SECP256R1}};
+  app
+    .add_option(
+      "--curve-id",
+      curve_id,
+      "Elliptic curve to use for node's certificate (used for TLS and ledger "
```
achamayou, line 432: Node and network identity in the case of a node starting in new or recovery mode.

src/tls/curve.h
```diff
@@ -2,76 +2,69 @@
...
+      case CurveID::SECP384R1:
+        return MDType::SHA384;
+      case CurveID::SECP256R1:
+        return MDType::SHA256;
       default:
       {
-        throw std::logic_error(
-          "Unhandled curve type: " +
-          std::to_string(static_cast<size_t>(curve)));
+        if (allow_none)
```
achamayou, line 47: When do we use allow_none?

src/crypto/hash.h
wintersteiger: Good observation - I'll move the allocation up so it's guaranteed only once.

achamayou (approved):

src/enclave/enclave.h
wintersteiger: Yes, but everything will be aborted anyways and there wasn't an existing error-path. I can add that though.

src/tls/curve.h
wintersteiger: In `verify_hash`, but I don't like it either, I'll see if I can get rid of it.

PR #2124, opened 2021-01-27
On backup run work async

ashamis:
part of #2055

Run execution/applying table data on backups on the non main thread (if it exists).

### DIFF
--- a/src/consensus/aft/async_execution.h
+++ b/src/consensus/aft/async_execution.h
@@ -6,7 +6,7 @@
 
 namespace aft
 {
-  class AbstractExecEntryStore
+  class AbstractExecMsgStore
   {
   public:
     virtual void recv_append_entries(
@@ -18,22 +18,24 @@ namespace aft
     virtual void recv_request_vote_response(RequestVoteResponse r) = 0;
     virtual void recv_signature_received_ack(SignaturesReceivedAck r) = 0;
     virtual void recv_nonce_reveal(NonceRevealMsg r) = 0;
-    virtual void recv_view_change(RequestViewChangeMsg r, const uint8_t* data, size_t size) = 0;
-    virtual void recv_view_change_evidence(ViewChangeEvidenceMsg r, const uint8_t* data, size_t size) = 0;
+    virtual void recv_view_change(
+      RequestViewChangeMsg r, const uint8_t* data, size_t size) = 0;
+    virtual void recv_view_change_evidence(
+      ViewChangeEvidenceMsg r, const uint8_t* data, size_t size) = 0;
   };
 
-  class AbstractExecEntry
+  class AbstractExecMsg
   {
   public:
-    virtual ~AbstractExecEntry() = default;
+    virtual ~AbstractExecMsg() = default;
     virtual void execute() = 0;
   };
 
-  class AppendEntryExecEntry : public AbstractExecEntry
+  class AppendEntryExecEntry : public AbstractExecMsg
   {
   public:
     AppendEntryExecEntry(
-      AbstractExecEntryStore* store_,
+      AbstractExecMsgStore* store_,
       AppendEntries&& hdr_,
       const uint8_t* data_,
       size_t size_,
@@ -51,19 +53,18 @@ namespace aft
     }
 
   private:
-    AbstractExecEntryStore* store;
+    AbstractExecMsgStore* store;
     AppendEntries hdr;
     const uint8_t* data;
     size_t size;
     OArray oarray;
   };
 
-  class AppendEntryResponseExecEntry : public AbstractExecEntry
+  class AppendEntryResponseExecEntry : public AbstractExecMsg
   {
   public:
     AppendEntryResponseExecEntry(
-      AbstractExecEntryStore* store_,
-      AppendEntriesResponse&& hdr_) :
+      AbstractExecMsgStore* store_, AppendEntriesResponse&& hdr_) :
       store(store_),
       hdr(std::move(hdr_))
     {}
@@ -74,16 +75,15 @@ namespace aft
     }
 
   private:
-    AbstractExecEntryStore* store;
+    AbstractExecMsgStore* store;
     AppendEntriesResponse hdr;
   };
 
-  class SignedAppendEntryResponseExecEntry : public AbstractExecEntry
+  class SignedAppendEntryResponseExecEntry : public AbstractExecMsg
   {
   public:
     SignedAppendEntryResponseExecEntry(
-      AbstractExecEntryStore* store_,
-      SignedAppendEntriesResponse&& hdr_) :
+      AbstractExecMsgStore* store_, SignedAppendEntriesResponse&& hdr_) :
       store(store_),
       hdr(std::move(hdr_))
     {}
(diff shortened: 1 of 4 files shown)


src/consensus/aft/raft.h
```diff
@@ -505,59 +508,132 @@ namespace aft
...
+          aee->execute();
+        }
+        else
+        {
+          pending_execution.push_back(std::move(aee));
         }
       }
+      catch (const std::logic_error& err)
+      {
+        LOG_FAIL_FMT("error type:{}, err.what:{}", type, err.what());
```
jumaffre, line 612: Please use `LOG_FAIL_EXC(err.what())` here

src/consensus/aft/raft.h
```diff
@@ -1117,6 +1178,61 @@ namespace aft
...
+        threading::ThreadMessaging::thread_messaging.add_task(
+          threading::ThreadMessaging::get_execution_thread(
+            threading::MAIN_THREAD_ID),
+          std::move(msg));
+      }
+      else
+      {
+        guard.unlock();
+        msg->cb(std::move(msg));
+      }
```
jumaffre, line 1199: As this pattern is repeated below, could we create a new wrapper in `thread_messaging.h` that adds an async task if `thread_count > 1` and otherwise executes the callback synchronously?

src/consensus/aft/raft.h
```diff
@@ -1117,6 +1178,61 @@ namespace aft
...
+        execute_append_entries_cb,
+        this,
+        std::move(append_entries),
+        std::move(r),
+        confirm_evidence);
+      if (threading::ThreadMessaging::thread_count > 1)
+      {
+        threading::ThreadMessaging::thread_messaging.add_task(
+          threading::ThreadMessaging::get_execution_thread(
+            threading::MAIN_THREAD_ID),
```
jumaffre, line 1192:
```suggestion
            threading::MAIN_THREAD_ID,
```

src/consensus/aft/raft.h
```diff
@@ -1117,6 +1178,61 @@ namespace aft
...
+      else
+      {
+        msg->cb(std::move(msg));
+      }
+    }
+
+    static void continue_execution(
+      std::unique_ptr<threading::Tmsg<AsyncExecution>> msg)
+    {
+      msg->data.self->is_execution_pending = false;
```
jumaffre, line 1224: Please correct me if I'm wrong but this doesn't seem to be protected by any lock.

src/consensus/aft/raft.h
```diff
@@ -1117,6 +1178,61 @@ namespace aft
         append_entries.push_back(std::make_tuple(std::move(ds), i));
       }
 
+      is_execution_pending = true;
+      auto msg = std::make_unique<threading::Tmsg<AsyncExecution>>(
+        execute_append_entries_cb,
+        this,
```
jumaffre, line 1184: Minor: Why not capturing a reference here instead, i.e. `*this`?

src/consensus/aft/raft.h
```diff
@@ -83,6 +84,8 @@ namespace aft
     // or even previous terms, and can therefore not meaningfully sign
     // over the commit level.
     kv::Version election_index = 0;
+    bool is_execution_pending = false;
+    std::list<std::unique_ptr<AbstractExecMsg>> pending_execution;
```
jumaffre, line 88:
```suggestion
    std::list<std::unique_ptr<AbstractExecMsg>> pending_executions;
```

src/consensus/aft/raft.h
```diff
@@ -505,59 +508,132 @@ namespace aft
...
-        case raft_append_entries_signed_response:
-          recv_append_entries_signed_response(data, size);
-          break;
+      try
+      {
+        // The host does a CALLIN to this when a Aft message
+        // is received. Invalid or malformed messages are ignored
+        // without informing the host. Messages are idempotent,
+        // so it is not necessary to defend against replay attacks.
+        if (type == raft_append_entries)
```
jumaffre, line 522: Why has the switch statement been replaced by `if {} else if {} ...` here?

src/consensus/aft/raft.h
```diff
@@ -505,59 +508,132 @@ namespace aft
...
 
-        case bft_view_change_evidence:
-          recv_view_change_evidence(data, size);
-          break;
+          aee = std::make_unique<ViewChangeEvidenceExecEntry>(
+            this, std::move(r), data, size, std::move(d));
+        }
+        else
+        {
+          throw std::logic_error("Unknown message type");
```
jumaffre, line 598: Also include `type` in error message here.

jumaffre: In fact, this shouldn't be a `throw` but simply a `LOG_FAIL` as we do not want to crash a node if a message is corrupted (integrity verification only happens if the message type is valid).

src/consensus/aft/async_execution.h
```diff
@@ -0,0 +1,232 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "raft_types.h"
+
+namespace aft
+{
+  class AbstractExecMsgStore
```
achamayou, line 9: I think I'm confused by the name of this class, this is presumably not a store?

achamayou:
I'm nervous about this change because:

1. It adds a lot of complexity to the CFT case for no expected performance benefit
2. There's no evidence that it works, all the tests are run without threads and no specific test was added

src/consensus/aft/async_execution.h
ashamis: AbstractConsensusCb

src/consensus/aft/async_execution.h
```diff
@@ -0,0 +1,232 @@
...
+  };
+
+  class AbstractExecMsg
+  {
+  public:
+    virtual ~AbstractExecMsg() = default;
+    virtual void execute() = 0;
+  };
+
+  class AppendEntryExecEntry : public AbstractExecMsg
```
ashamis, line 34: call callbacks....

src/consensus/aft/raft.h
```diff
@@ -505,59 +508,132 @@ namespace aft
...
 
-        case raft_append_entries_signed_response:
-          recv_append_entries_signed_response(data, size);
-          break;
+      try
+      {
+        // The host does a CALLIN to this when a Aft message
+        // is received. Invalid or malformed messages are ignored
+        // without informing the host. Messages are idempotent,
+        // so it is not necessary to defend against replay attacks.
```
ashamis, line 521: delete comment

src/consensus/aft/raft.h
```diff
@@ -505,59 +508,132 @@ namespace aft
...
+      {
+        LOG_FAIL_FMT("error type:{}, err.what:{}", type, err.what());
+      }
+
+      try_execute_pending();
+    }
+
+    void try_execute_pending()
+    {
+      while (!is_execution_pending && !pending_execution.empty())
```
ashamis, line 620: call periodic

src/consensus/aft/raft.h
```diff
@@ -1117,6 +1178,61 @@ namespace aft
...
       }
 
+      is_execution_pending = true;
+      auto msg = std::make_unique<threading::Tmsg<AsyncExecution>>(
+        execute_append_entries_cb,
+        this,
+        std::move(append_entries),
+        std::move(r),
+        confirm_evidence);
+      if (threading::ThreadMessaging::thread_count > 1)
```
ashamis, line 1188: Add comment here about control synch

src/consensus/aft/raft.h
ashamis: discussed offline

achamayou (approved):

PR #2126, opened 2021-01-28
Remove drift in enclave logging time

eddyashton:
I just noticed there was a drift in the time we're tracking within the enclave, meaning the time offset was constantly increasing. For instance, on [a healthy node which had been running for 45 minutes](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=18082&view=logs&j=b69c18ba-eebc-5249-7cb0-46eb355ef417&t=42332f68-aaa8-559e-eed6-ff9102874193&l=5507):

```
79: 17:53:57.014 | INFO     | infra.remote:log_errors:77 - 2021-01-27T17:08:45.806587Z -0.068 0   [info ] ../src/consensus/aft/raft.h:2237     | Removed raft self 2
79: 17:53:57.014 | INFO     | infra.remote:log_errors:77 - 2021-01-27T17:53:56.898778Z        100 [info ] ../src/host/sig_term.h:24            | SIGTERM: Shutting down enclave gracefully...
79: 17:53:57.014 | INFO     | infra.remote:log_errors:77 - 2021-01-27T17:53:56.924282Z -149.550 0   [info ] ../src/enclave/enclave.h:385         | Enclave stopped successfully. Stopping host...
```

This `-149.550` is wrong - the enclave asked for this to be logged a few microseconds ago, not 150s. (You can see the same thing by adding a `LOG_INFO` line in the handler for the `tick` message, and starting an idle sandbox network - even with no load the enclave is reported as falling constantly behind the host).

The root cause is the rounding of elapsed time (down to ms precision) when passed to `tick`. The host/real time advances 1.1ms, and tells the enclave to advance its count by 1ms. These add up and we drift. My fix is to remove the elapsed time argument from `tick` entirely - we have a better in-enclave time source now, in a memory location written to by a host thread. So each tick we pull a microsecond-accurate current time, and use that. The other systems are still happy with elapsed_ms approximations, but the logging gets a straight-up "time now" to be later compared against the host time. I have a feeling this could be simplified further, but don't want to get too deep into these weeds.


achamayou (approved):
That's a really good catch!

PR #2130, opened 2021-01-28
Fix Signatures Table Name in Python Ledger module

syhamza:
Fixing Signatures table name as defined in CCF here: https://github.com/microsoft/CCF/blob/master/src/node/entities.h#L96


achamayou (approved):
Thank you! We unfortunately do not have an end to end test that validates signatures from the ledger using the Python API, which would have caught this.

jumaffre (approved):

eddyashton (approved):

jumaffre:
Merging this as the failure on the Daily CI is unrelated to this change.

PR #2131, opened 2021-01-29
Tweak metrics layout

achamayou:
Experiment changing the layout slightly because recent additions have resulted in a squished-downed vertical scale, and so it's harder to see changes now.


achamayou:
Ok, I think the plots themselves are more readable, but the titles clearly aren't.

PR #2132, opened 2021-01-29
Shallow repo clone in CI

jumaffre:
Each CI job spends a non-negligible (~1 minute) amount of time cloning the entire CCF repo. Instead, we should only clone the latest commit, bringing the clone time down to ~10 seconds.

### DIFF
--- a/.azure-pipelines-gh-pages.yml
+++ b/.azure-pipelines-gh-pages.yml
@@ -14,6 +14,7 @@ jobs:
     steps:
       - checkout: self
         clean: true
+        fetchDepth: 1
 
       - script: |
           env
--- a/.azure-pipelines-gh-pages.yml
+++ b/.azure-pipelines-gh-pages.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - checkout: self
         clean: true
-        fetchDepth: 1000
+        fetchDepth: 10000
 
       - script: |
           env
--- a/.azure-pipelines-gh-pages.yml
+++ b/.azure-pipelines-gh-pages.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - checkout: self
         clean: true
-        fetchDepth: 1
+        fetchDepth: 1000
 
       - script: |
           env
--- a/.azure-pipelines-templates/checks.yml
+++ b/.azure-pipelines-templates/checks.yml
@@ -7,7 +7,7 @@ jobs:
     steps:
       - checkout: self
         clean: true
-        submodules: true
+        fetchDepth: 1
 
       - script: ./scripts/ci-checks.sh
         displayName: "Checks"
--- a/.azure-pipelines-templates/common.yml
+++ b/.azure-pipelines-templates/common.yml
@@ -10,6 +10,7 @@ jobs:
     steps:
       - checkout: self
         clean: true
+        fetchDepth: 1
 
       - template: cmake.yml
         parameters:
--- a/.azure-pipelines-templates/common.yml
+++ b/.azure-pipelines-templates/common.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - checkout: self
         clean: true
-        fetchDepth: 1
+        fetchDepth: 1000
 
       - template: cmake.yml
         parameters:
--- a/.azure-pipelines-templates/common.yml
+++ b/.azure-pipelines-templates/common.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - checkout: self
         clean: true
-        fetchDepth: 1000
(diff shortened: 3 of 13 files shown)


jumaffre:
Argh, this fails because of the magic version detection from the latest git tag we have in cmake...

achamayou:
I was going to say... I think we could get much the same effect by compacting the cimetrics branch again.

jumaffre:
> Argh, this fails because of the magic version detection from the latest git tag we have in cmake...

Setting the `depth` to `1000` still yields big improvements as only the branch of interest gets cloned. 

> I was going to say... I think we could get much the same effect by compacting the cimetrics branch again.

Yes, it would speed up manual clones I guess. I think we still want to set the clone `depth` so we know only the target branch is cloned.

achamayou:
@jumaffre ok, and the failure mode is cmake breaking if there's been more than 1000 commits since the last tag?

achamayou:
It sounds like --single-branch may be what we want instead?

jumaffre:
> It sounds like --single-branch may be what we want instead?

Unfortunately, it isn't a valid option for [the Azure Pipelines YML](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#checkout). We could convert all "checkout" steps to a manual git clone but it seems far from ideal...

1,000 seems deep enough that we wouldn't hit this issue with the current rate of commits on this repo (`master` currently has 1,406 commits, for 46 releases) and the checkout step is now down to 2s.

achamayou:
Ah, I hadn't noticed that, that's a shame because it seems like a useful option. It may be worth opening a ticket? Setting the depth to 1000 sounds like an acceptable workaround for now. How much we do save that way?

eddyashton:
image
becomes
image
That's a good win! Is it worth making the error case in CMake a little more obvious? What happens if we try our git detection on a repo with no tags?

jumaffre:
> Ah, I hadn't noticed that, that's a shame because it seems like a useful option. It may be worth opening a ticket? Setting the depth to 1000 sounds like an acceptable workaround for now. How much we do save that way?

Tested locally:

```bash
$ time git clone git@github.com:microsoft/CCF.git
Cloning into 'CCF'...
remote: Enumerating objects: 43, done.
remote: Counting objects: 100% (43/43), done.
remote: Compressing objects: 100% (34/34), done.
remote: Total 38402 (delta 8), reused 1 (delta 0), pack-reused 38359
Receiving objects: 100% (38402/38402), 1.48 GiB | 15.47 MiB/s, done.
Resolving deltas: 100% (22851/22851), done.

real    1m51.878s
user    0m24.551s
sys     0m9.170s
$ rm -rf CCF
$ time git clone git@github.com:microsoft/CCF.git --depth 1000
Cloning into 'CCF'...
remote: Enumerating objects: 18076, done.
remote: Counting objects: 100% (18076/18076), done.
remote: Compressing objects: 100% (5734/5734), done.
remote: Total 18076 (delta 13244), reused 16512 (delta 12015), pack-reused 0
Receiving objects: 100% (18076/18076), 12.81 MiB | 14.83 MiB/s, done.
Resolving deltas: 100% (13244/13244), done.

real    0m3.457s
user    0m0.781s
sys     0m0.208s
```

I've also tested this with no depth limitations and only the master branch and the speedup is similar so I'm assuming the long clone comes from other branches (and most certainly the `cimetrics/` one).

jumaffre:
> That's a good win! Is it worth making the error case in CMake a little more obvious? What happens if we try our git detection on a repo with no tags?

Good point! I'll add too. 

I was looking at the speed up on the CI until I saw [this run](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=18142&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e) and I'm now assuming that the repo gets cached on the CI agent somehow?

eddyashton:
> I was looking at the speed up on the CI until I saw [this run](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=18142&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e) and I'm now assuming that the repo gets cached on the CI agent somehow?

Looks like its cached per-run, but there's always one slow checkout each run. Usually the slow job is Format Checks since it runs first, but in that Daily its SGX Release.

As another mitigation to delay this biting us with awkward CI errors, shall we do 10,000 rather than 1,000? We're barely over 1,000 now, so it won't get slower for a while yet.

achamayou (approved):

jumaffre:
Merging this regardless of the daily CI status which is now green.

PR #2135, opened 2021-01-29
Include merkle root at read version in proposal ids

achamayou:
Final part of #1649.

- [x] More specific path comparison
- [x] Add comment on conflict behaviour
- [x] Add a reliable conflict test


src/node/rpc/member_frontend.h
```diff
@@ -1332,16 +1332,56 @@ namespace ccf
...
+        else
+        {
+          proposal_id = fmt::format(
+            "{:02x}", fmt::join(caller_identity.request_digest, ""));
+        }
+
+        Proposal proposal(in.script, in.parameter, caller_identity.member_id);
         auto proposals = ctx.tx.rw(this->network.proposals);
-        proposals->put(proposal_id, proposal);
+        if (!proposals->has(proposal_id))
```
jumaffre, line 1378: Is the `else` path here covered by a conflict when the tx is committed? It might be worth adding a short comment about this?

achamayou: No, it's not. What this does is if the proposal already exists, skip trying to create it, but still evaluate it in order to return an accurate status.

jumaffre: AIUI, the only way the proposal ID could already exist (apart from hash collisions) is that two identical proposals were created at the same read version... which isn't possible because of this very line which takes a read dependency at this key on the proposal table (i.e. the second proposal creation would conflict).

achamayou: Yes, you're absolutely right, this can't even happen! I'll add a comment.

src/node/rpc/test/proposal_id_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+      std::ref(frontend),
+      std::ref(propose),
+      std::ref(proposer_cert),
+      std::ref(out2));
+    t1.join();
+    t2.join();
+
+    DOCTEST_CHECK(out1.state == ProposalState::OPEN);
+    DOCTEST_CHECK(out2.state == ProposalState::OPEN);
+    DOCTEST_CHECK(out1.proposal_id != out2.proposal_id);
```
achamayou, line 321: This needs an extra check to confirm we have triggered a conflict, perhaps looking at a (yet to be added) conflict counter in the endpoint metrics.

src/node/rpc/test/proposal_id_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+string get_script_path(string name)
+{
+  auto default_dir = "../src/runtime_config";
+  auto dir = getenv("RUNTIME_CONFIG_DIR");
+  stringstream ss;
+  ss << (dir ? dir : default_dir) << "/" << name;
+  return ss.str();
+}
+const auto gov_script_file = files::slurp_string(get_script_path("gov.lua"));
```
achamayou, line 53: There's a fair bit of stuff that's not needed here, I'll trim it off.

src/node/rpc/member_frontend.h
```diff
@@ -1332,16 +1332,61 @@ namespace ccf
...
+            HTTP_STATUS_INTERNAL_SERVER_ERROR,
+            ccf::errors::InternalError,
+            "No consensus available.");
+        }
+
+        std::string proposal_id;
 
+        if (consensus->type() == ConsensusType::CFT)
+        {
+          auto root_at_read = ctx.tx.get_root_at_read_version();
```
jumaffre, line 1348: I thought this was supposed to take a read dependency on the signatures table to avoid the compaction issue discussed a couple of weeks ago?

achamayou: That may be an implementation detail in #2136 for BFT, but unnecessary for CFT, since it's not a historical root.

CMakeLists.txt
```diff
@@ -399,6 +399,15 @@ if(BUILD_TESTS)
                                  secp256k1.host http_parser.host sss.host
     )
 
+    # add_unit_test(
```
achamayou, line 402: Turned off for now because of transient instability, which I suspect is caused by the lack of safety in metrics and endpoint lookups, to be fixed separately.

src/node/rpc/test/member_voting_test.cpp
```diff
@@ -1053,6 +1073,11 @@ DOCTEST_TEST_CASE("Propose raw writes")
...
       NetworkState network;
       network.tables->set_encryptor(encryptor);
+      auto history =
+        std::make_shared<ccf::NullTxHistory>(*network.tables, 0, *kp);
+      network.tables->set_history(history);
+      auto consensus = std::make_shared<aft::StubCFTConsensus>();
+      network.tables->set_consensus(consensus);
       auto gen_tx = network.tables->create_tx();
       GenesisGenerator gen(network, gen_tx);
       gen.init_values();
```
eddyashton, line 1083: Can we bury this (and `gen.create_service()`, and maybe some other lines that are _often_ called + harmless where they're currently not called) in a `init_network` or similar?

src/node/rpc/test/stub_consensus.h
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
eddyashton, line 1: Can this file be replaced by `kv/test/stub_consensus.h`? Or is it doing something different I'm missing?

src/node/rpc/member_frontend.h
```diff
@@ -1332,16 +1332,61 @@ namespace ccf
...
+
+        Proposal proposal(in.script, in.parameter, caller_identity.member_id);
         auto proposals = ctx.tx.rw(this->network.proposals);
-        proposals->put(proposal_id, proposal);
+        // The existence of this check guarantees that we always enter this
+        // block because it creates a read-dependency on that proposal_id, and
+        // so any other request with an identical digest and the same
+        // read_version will conflict, be re-based, acquire a new read_version
+        // and therefore a new proposal_id.
+        if (!proposals->has(proposal_id))
```
eddyashton, line 1383: This behaviour is surprisingly subtle. Even reading the comment and dicsussion, I spent a while down the "what if this check fails" path (and so we never write and so this whole proposal becomes potentially read-only and so we get a duplicate proposal ID?) before convincing myself that can never happen.

What we're really saying is "the other case is a hash collision", so I think we should invert this test and document that:
```
// Introduce a read dependency, so that if identical proposal creations are in-flight
// and reading at the same version, the later conflicts and is re-executed. If we ever
// produce a proposal ID which already exists, we have a hash collision
if (proposals->hash(proposal_id))
{
 return make_error("HASH COLLISION: The sky is falling");
}

proposals->put(...)
record...
```
(proposed comment wording is inspirational and unmaintained)

src/node/rpc/test/member_voting_test.cpp
achamayou: Sure

src/node/rpc/member_frontend.h
achamayou: What I had in mind was that dependency handling could be broken in the KV. The defensive and correct thing to do is to avoid overwriting the key, since the proposal is structurally identical. Adding an assert() so this gets caught in debug would be better of course.

You're right that a hash collision could happen here (and in many other places). I do not know how to rate the relative probabilities of that, versus dependency handling breaking, but I'm happy to make the change you suggest if it improves readability.

Technically, we could read and compare the actual proposal contents, and work out which it is. I am not convinced it is useful.

src/node/rpc/test/stub_consensus.h
achamayou: There's at least two things, one is the node_id/node_set is consistent in the new file, but that can probably be fixed. The other is that the new stub has signatures turned off. I'll try to merge them still because they are quite similar, I did not know we had this one.

eddyashton (approved):

PR #2137, opened 2021-01-29
Add method to get the Version when a key was last written to

eddyashton:
See #2134 for discussion.

This implements option 2, with some unit tests for basic correctness and contention (checking we produce a monotonic chain after concurrent execution).

The name of this method is still up for grabs.


src/kv/test/kv_test.cpp
```diff
@@ -149,6 +161,126 @@ TEST_CASE("Reads/writes and deletions")
   }
 }
 
+TEST_CASE("get_previous_version")
```
jumaffre, line 164: Can we also check that this is true across a primary and backup store and after deserialising snapshots?

src/kv/map_handle.h
```diff
@@ -88,6 +88,26 @@ namespace kv
       return read_handle.has(KSerialiser::to_serialised(key));
     }
 
+    /** Get version when this key was last written to, by a previous
```
jumaffre, line 91: I think `kv/api.rst` will need to include `get_previous_version` as a member of `ReadableMapHandle` to show up in the docs.

src/kv/test/kv_test.cpp
eddyashton: Modified the basic tests in `kv_serialisation.cpp` and `kv_snapshot.cpp` to test this - let me know if you think of any other awkward cases we should be testing.

src/kv/test/kv_snapshot.cpp
```diff
@@ -79,21 +95,34 @@ TEST_CASE("Simple snapshot" * doctest::test_suite("snapshot"))
...
-    REQUIRE_EQ(new_store.current_version(), 2);
+    REQUIRE_EQ(new_store.current_version(), second_snapshot_version);
 
     auto tx1 = new_store.create_tx();
-    auto handle = tx1.rw<MapTypes::StringString>("public:string_map");
 
-    auto v = handle->get("foo");
-    REQUIRE(v.has_value());
-    REQUIRE_EQ(v.value(), "bar");
+    {
```
jumaffre, line 102: Should we also test for key deletion in the second snapshot?

jumaffre (approved):

achamayou:
@eddyashton can we have a changelog entry for this?

doc/build_apps/kv/api.rst
```diff
@@ -43,7 +43,7 @@ Handle
 
 .. doxygenclass:: kv::ReadableMapHandle
    :project: CCF
-   :members: get, has, foreach
+   :members: get, has, foreach, get_version_of_previous_write
```
achamayou, line 46: fwiw I do like that name.

src/kv/map_handle.h
```diff
@@ -88,6 +88,26 @@ namespace kv
...
+     * Returns nullopt when there is no value, because the key has no value
+     * (never existed or has been removed). Note that this is always talking
+     * about the version of previously committed state and not the same values
+     * as @c get or @c has - this transaction's pending writes have no version
+     * yet, and this method does not talk about them.
+     *
+     * @param key Key to read
+     *
+     * @return Optional containing version of applied transaction which last
+     * wrote at this key, or nullopt of this key has no associated value
```
achamayou, line 103:
```suggestion
     * wrote at this key, or nullopt if this key has no associated value
```

achamayou (approved):

Issue #230, opened 2019-07-05
Documentation for internal tables


achamayou:
As well as being prefixed (#140), framework tables (ie. starting with `ccf.`) ought to be documented:

1. Schema
2. Access category
3. Description
4. Serialisation format

achamayou:
To start with, governance tables that are expected to be audited by users should be documented.

CH-ND-N:
r I am interested please assign this to me. I am new to this but have knowledge of C, C++, and proper markdown knowledge.I will need guidance.

achamayou:
@CH-ND-N thank you for helping, may I suggest you start by reading the “Governance” section of the online documentation, before having a look at member_frontend.h? That’s where most of the logic interacting with internal tables lives. If you have any questions, please do not hesitate to ask!

jumaffre:
Assigning this to myself as I've been looking into (in the background for now). Ideally, we would automatically generate the documentation from the existing definition of CCF's built-in tables. Doxygen annotations is probably the best option I can think of.

lynshi:
Bumping this since it's a little more useful now, as we're querying the CCF tables in our endpoints, and I ended up here while searching for schemas (incidentally, tables are defined [here](https://github.com/microsoft/CCF/blob/main/src/node/network_tables.h), and you can follow the types to infer the schemas). @eddyashton has also mentioned having C++ APIs wrapping table querying to decouple application code from schema changes, which seems even more useful than a published schema.

achamayou:
All governance tables are documented now, but still TODO:

- [ ] document internal tables

PR #2140, opened 2021-02-01
Split shares KV table

jumaffre:
Part of #1648 

To be able to retrieve historical ledger secrets, CCF must know the version at which a ledger secret was recorded in the KV. Once #2137 is merged, we will be able to record this in the store and recover historical ledger secrets. 

The issue was that the `ccf.shares` table was also updated on pure re-share (i.e. reshuffling the recovery shares without refreshing the ledger secret), which meant that we would have to do extra round trips to recover the ledger secrets.

The `ccf.shares` now is split into the `ccf.encrypted_ledger_secrets` to record the previous encrypted ledger secret + version + version at which it was recorded in the store (unused for now, waiting for #2137) + version of the next ledger secret (useful for recovery). The `ccf.shares` still contains the latest ledger secret encrypted with the ledger secret wrapping key and the corresponding recovery shares, encrypted with each member's public encryption key. 

Also tidied up and simplified of bunch of things so that the recovery flow is hopefully a little cleaner.


src/node/entities.h
```diff
@@ -81,6 +81,8 @@ namespace ccf
     static constexpr auto GOV_HISTORY = "public:ccf.gov.governance.history";
     static constexpr auto SERVICE = "public:ccf.gov.service";
     static constexpr auto SHARES = "public:ccf.gov.shares";
+    static constexpr auto ENCRYPTED_LEDGER_SECRETS =
```
achamayou, line 84: This is a new map, can we have it be documented? Also, this feels much more like a internal rather than a gov table.

src/node/node_state.h
```diff
@@ -1595,68 +1596,50 @@ namespace ccf
...
-            for (const auto& [k, opt_v] : w)
+            if (w.size() > 1)
             {
-              if (!opt_v.has_value())
-              {
-                throw std::logic_error(
-                  fmt::format("Unexpected: removal from shares table ({})", k));
-              }
+              throw std::logic_error(fmt::format(
+                "Unexpected: multiple writes to {} table",
```
achamayou, line 1610: I think I would prefer "Transaction contains {} writes to map {}, expected one".

src/node/node_state.h
```diff
@@ -1595,68 +1596,50 @@ namespace ccf
...
+                network.encrypted_ledger_secrets.get_name()));
+            }
 
-              const auto& v = opt_v.value();
+            auto encrypted_ledger_secret_info = w.at(0);
+            if (!encrypted_ledger_secret_info.has_value())
+            {
+              throw std::logic_error(fmt::format(
+                "Unexpected: removal from {} table",
```
achamayou, line 1618: "Unexpected" is implied by logic_error in my opinion.

src/node/share_manager.h
```diff
@@ -153,44 +146,65 @@ namespace ccf
...
-      // wrapping key, allocating a new share for each active member. Finally,
-      // encrypt each share with the public key of each member and record it in
-      // the shares table.
+      // wrapping key, allocating a new share for each active recovery member.
+      // Finally, encrypt each share with the public key of each member and
+      // record it in the shares table.
 
-      auto ls_wrapping_key = LedgerSecretWrappingKey();
-      auto wrapped_latest_ls = ls_wrapping_key.wrap(latest_ledger_secret);
+      _shuffle_recovery_shares(tx, latest_ledger_secret);
```
achamayou, line 173: Technically this is reserved, and we don't use leading underscore anywhere else in the code base that I know of, I would prefer not to introduce it unless it's a deliberate and well-defined convention.

src/node/share_manager.h
```diff
@@ -266,61 +280,72 @@ namespace ccf
...
 
-      set_recovery_shares_info(tx, latest, penultimate, latest_ls_version);
+      set_recovery_shares_info(tx, latest.second, penultimate, latest.first);
     }
 
-    void issue_shares_on_rekey(
+    void issue_recovery_shares(
       kv::Tx& tx, const LedgerSecret& new_ledger_secret)
     {
+      // Issue new recovery shares of the new ledger secret, recording the
```
achamayou, line 297: Here and below, this would be better as a doxygen block for the function. We don't built an internal API doc (yet?), but if we ever do, we'll be happier that way and it's not much more verbose. It's great to have descriptive comments though, thank you for taking the time to add them!

src/node/share_manager.h
```diff
@@ -266,61 +280,72 @@ namespace ccf
...
       // First, re-assemble the ledger secret wrapping key from the submitted
       // encrypted shares. Then, unwrap the latest ledger secret and use it to
-      // decrypt the previous ledger secret and so on.
+      // decrypt the sequence of recovered ledger secrets, from the last one.
 
-      auto ls_wrapping_key = combine_from_submitted_shares(tx);
+      if (recovery_ledger_secrets.empty())
+      {
+        throw std::logic_error(
+          "There should be at least one encrypted ledger secret to recover");
```
achamayou, line 345: "Could not find  any ledger secrets in the ledger"? It seems odd to have that check here, it feel like this is a transform over a sequence and it's more appropriate for the caller to have an opinion about whether empty is a problem.

src/node/share_manager.h
```diff
@@ -329,25 +354,36 @@ namespace ccf
...
+        recovery_shares_info->wrapped_latest_ledger_secret);
       auto decryption_key = restored_ls.raw_key;
 
+      LOG_DEBUG_FMT(
+        "Recovering {} encrypted ledger secrets",
+        recovery_ledger_secrets.size());
+
+      auto& current_ledger_secret_version =
+        recovery_ledger_secrets.back().next_version;
+      if (!current_ledger_secret_version.has_value())
```
achamayou, line 367: If that should always be set, why is that optional? When is it ever unset?

achamayou (approved):

src/node/entities.h
jumaffre: If this map is internal (as your other comment suggests, which I also believe is right), do we still need to document it?

src/node/share_manager.h
jumaffre: Good point. I've updated the error message.

An empty `recovery_ledger_secrets` list here is an issue as the version of the current ledger secret (in the `shares` table) is indicated by the last entry in `recovery_ledger_secrets`. If no ledger secret were recovered from the ledger, the recovery of the ledger secrets cannot be done. I believe it is the responsibility of this class, rather than the caller.

src/node/share_manager.h
jumaffre: It is optional because the elements in `recovery_ledger_secrets` are of the same type as the values in the `encrypted_ledger_secrets` table (i.e. `PreviousLedgerSecretInfo`), for convenience mostly. The recovery hook on that table make sure that this value is set though. It's not set in the case of a rekey, when CCF doesn't know the value at which the new ledger secret is applicable when it writes the transaction, and the value is then set to the recovery hook version (see comment in the recovery hook in `node_state.h`)

achamayou: That seems unfortunate, perhaps we ought to consider splitting the type between the tables so we can be more precise?

jumaffre: As discussed, I've added a comment to explain why this optional here is always set by the recovery hook. It is optional in the first place because the value for that table contains an optional.

src/node/entities.h
jumaffre: As discussed, this needs documenting as part of #230

eddyashton (approved):

PR #2141, opened 2021-02-01
Infra: Raft election timeout defaults to realistic value

jumaffre:
It looks like only some tests were run with a realistic raft election timeout value while most were set to such a high value that no election would ever occur in case of the primary failing.


CMakeLists.txt
```diff
@@ -577,35 +565,41 @@ if(BUILD_TESTS)
...
   add_e2e_test(
     NAME js_batched_stress_test
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/e2e_batched.py
     CONSENSUS cft
-    ADDITIONAL_ARGS --js-app-bundle ${CMAKE_SOURCE_DIR}/src/apps/batched
+    ADDITIONAL_ARGS
+      --js-app-bundle
+      ${CMAKE_SOURCE_DIR}/src/apps/batched
+      --raft-election-timeout-ms
+      10000 # Larger election timeout as recording large JS applications may
```
jumaffre, line 591: As discussed with @letmaik, both the JS batched and module tests submit large payloads to CCF which, in a single enclave thread scenario, may cause leadership changes. I've observed these changes on a virtual _debug_ build only (see previous CI runs on this PR).

achamayou (approved):

Issue #2139, opened 2021-02-01
Test out-of-process attestation


achamayou:
With Intel upstreaming the driver, this is going to become default in OpenEnclave. Images need the following additional packages:

  - "libsgx-aesm-ecdsa-plugin"
  - "libsgx-aesm-pce-plugin"
  - "libsgx-aesm-quote-ex-plugin"
  - "libsgx-quote-ex"
  - "sgx-aesm-service"

And SGX_AESM_ADDR must be set. The next release will make this default, although opting out with SGX_USE_IN_PROCESS_QUOTING will be possible.

After the kernel upgrade, staying with in process quoting will require permission changes (https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/driver/linux/README.md) which are very unlikely to be desirable for typical CCF usage.

achamayou:
Also see: https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-nodes-out-of-proc-attestation

This seems to work locally, we'll update the playbooks and pick up the new default at the next OpenEnclave release.

PR #2146, opened 2021-02-01
Update dependencies for quote-ex

achamayou:
Preparation for the switch to quote-ex in the next OpenEnclave release (#2139).


getting_started/setup_vm/roles/intel/vars/common.yml
```diff
@@ -20,6 +20,8 @@ intel_sgx_dev_packages:
 intel_dcap_packages:
   - "libsgx-dcap-ql"
   - "libsgx-urts"
+  - "libsgx-quote-ex"
 
 intel_dcap_dev_packages:
   - "libsgx-dcap-ql-dev"
```
achamayou, line 26: Deliberately only on dev images, the expectation for the app-run image is that it's run under Kubernetes, where a single externally deployed instance of AESM is expected, eg. https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-nodes-out-of-proc-attestation#how-does-the-out-of-proc-attestation-mode-work-for-confidential-workloads-scenario

eddyashton (approved):
Because this triggers all of the CIs, including Daily and Stress, it is very expensive to rerun. The branch is currently out-of-date, but once we get a green run we should probably update, squash and merge, overriding the CI checks. But we should preferably do that at the end of the day so the master CI runs don't block other work.

PR #2161, opened 2021-02-03
Update quote format and record endorsements in KV

### LINKED ISSUE #1468: Update attestation API and cache quote endorsements

jumaffre:
Fixes #1468 

This PR updates the API used by CCF nodes to generate and verify quotes, and now uses the latest/generic `oe_get_evidence`/`oe_verify_evidence` API. This means that the node's quote format is now different, but still verifiable with the  `oeverify` CLI (*). 

We also now record the quote endorsements in the nodes table, along with the quote. Clients can retrieve these endorsements using the existing `/node/quotes/self` endpoint and verify those directly. 

(*) Because we now serialise the report data (i.e. hash of node's public key) as a "custom claim", it is not printed by the current version of `oeverify` (see fix https://github.com/openenclave/openenclave/pull/3817). I have disabled this check in `verify_quote.sh` (via the `CCF_VERIFY_QUOTED_CERT` env var) and we will change the default behaviour once the update is included in a release of Open Enclave.


src/node/rpc/node_frontend.h
```diff
@@ -187,7 +185,8 @@ namespace ccf
 
           if (
             !this->node.is_part_of_network() &&
-            !this->node.is_part_of_public_network())
+            !this->node.is_part_of_public_network() &&
+            !this->node.is_reading_private_ledger())
```
jumaffre, line 189: Nodes should be able to join even when a node is recovering the private ledger!

src/node/rpc/node_frontend.h
```diff
@@ -326,23 +325,22 @@ namespace ccf
 
       auto get_quote = [this](auto& args, nlohmann::json&&) {
         QuoteFormat format;
-        std::vector<uint8_t> raw_quote;
+        QuoteInfo node_quote_info;
         const auto result =
-          get_quote_for_this_node_v1(args.tx, format, raw_quote);
+          get_quote_for_this_node_v1(args.tx, format, node_quote_info);
```
jumaffre, line 330: Breaking change here. I believe this is fine as we are not v1.0 just yet? I've noted the change in the changelog.

tests/infra/network.py
```diff
@@ -588,10 +584,8 @@ def create_and_add_pending_node(
             if errors:
                 # Throw accurate exceptions if known errors found in
                 for error in errors:
-                    if "CODE_ID_NOT_FOUND" in error:
```
jumaffre, line 591: Simplifying the logic here as code ids are now either in the KV or not.

.azure-pipelines-gh-pages.yml
```diff
@@ -7,7 +7,7 @@ trigger:
 
 jobs:
   - job: build_and_publish_docs
-    container: ccfciteam/ccf-ci:oe0.13.0-quote-ex
+    container: ccfciteam/ccf-ci:oe0.13.0-xxd
```
jumaffre, line 10: Required to decode the hex-encoded quote and endorsements and replace the call to `perl` in `verify_quote.sh`.

achamayou (approved):

PR #2162, opened 2021-02-03
persistent js runtime

letmaik:
very experimental, work-in-progress

TODO:
- re-create runtime whenever js app is redeployed (to avoid using old cached modules)
- quickjs [does not support](https://www.freelists.org/post/quickjs-devel/stackoverflow,2) being used from different threads -> evaluate against ccf's upcoming multi-threading support
  - maintain separate runtimes per thread
- evaluate perf improvements


src/apps/js_generic/js_generic.cpp
```diff
@@ -1004,6 +967,47 @@ namespace ccfapp
...
+            fmt::format(
+              "No handler script found for method '{}'.", verb_prefixed));
+          return;
+        }
+      }
+
+      // Ideally this should happen in the constructor but that leads
+      // to issues with quick's stack overflow checks.
+      // This is only safe as long as this instance of JSHandlers
+      // is only called from a single persistent thread.
```
achamayou, line 1004: That's fine, the runtime can be created once per thread and owned by a unique pointer in thread-local storage for example.

letmaik:
It looks like quickjs' stack overflow test is not compatible with storing a runtime instance and then later entering the same path again if the stack size is not constant. Since CCF doesn't guarantee a constant stack size when reaching an endpoint handler the only solution would be to disable the stack overflow check, although this would mean infinite loops in JS would crash the node instead of being detected. Since C++ apps also do not have this protection it may make sense to disable the check.

src/apps/js_generic/js_generic.cpp
letmaik: I made it thread-local, could you have a quick look if this is fine?  I also found a way to deal with the stack overflow check, see comments in the code.

letmaik:
I think I'm going to abandon this. I did some benchmarks locally with the forum app and could not see improvements.

Issue #1378, opened 2020-07-03
Re-evaluate the use of KV-driven incrementing IDs for users, members, nodes and proposals


eddyashton:
Our incrementing IDs expose CCF users to dangerous mistakes in the presence of rollbacks.

We could remove these IDs entirely. In most places we can instead use the digest of the caller's cert, and the only case this doesn't work is Proposals for which we should be able to generate a UID (on the node, while processing `/gov/propose`).

eddyashton:
There is a major caveat to replacing the incrementing IDs, which is that they make it harder to detect and blame replay attacks. We should look at these closely, and work out which are safe or can be made safe (the IDs are only accepted/used once they're globally committed), and which should be replaced by `View.IncrementingIndex` or by generated UIDs.

achamayou:
@eddyashton that's only true for UUIDs (ie. proposals), where we can use identity digests there is no problem and it's a clear improvement in rollback situations.

achamayou:
To clarify, there are a few different items here:

1. Proposals
2. Nodes
3. Users and Members

In the case of proposals, the main risk is confusion. If proposal X comes up, but gets rolled back, it's possible to lodge a vote for/against X that really applies to what's now a different proposal, if another one gets submitted in the interval. In theory, the vote script ought to be precise enough for that not to matter, but it's good defense in depth to make this difficult.

To make this difficult, it's ideal for proposal ids never to be re-used in time (even if they aren't in the committed order of course), and probably good for them not to be too predictable either.

The same applies to node ids, for less important reasons: to avoid operator confusion (not a security problem, but one that affects availability nonetheless), and to make the reconfiguration code more obviously correct. The reconfiguration code involves non-trivial interactions between KV and consensus (see #1368), making it deal with unique ids (again, in time, not in committed order) is a clear simplicity win.

Users are also at risk of confusion to some extent in case of rollback, assuming they are removed by id, for the same reasons as proposals. There is no member removal mechanism at the moment (they are simply flagged), but the same reason would apply if/once there is.

Two more factors to consider are:

1. The benefit of having a consistent id mechanism across identities, be they members, nodes etc.
2. The information leaked in non-random ids, which allows guessing the rate at which identities are added, and potentially the total amount.

letmaik:
Proposal IDs are digests as of #2104. 

For nodes, users, and members, using digests of certs/public keys makes sense.

[...]

PR #2166, opened 2021-02-04
Rename internal tables

jumaffre:
Part of #230 

This is very much a discussion PR - I thought it would be more concrete to discuss this on a PR directly, rather than an issue/discussion.

Most CCF's internal tables were prefixed with `public:ccf.gov.` but not all of them are actually useful for audit or written/read by the governance directly. I've made an attempt at renaming them consistently, following these principles:

1. Tables that are not written by governance operations are `ccf.internal` (e.g. signatures, snapshot evidence). 
2. Tables that are written by governance operations but are implementation details are also `ccf.internal` (e.g. submitted shares, secrets). 
3. Tables that reflect other governance tables are also `ccf.internal` (e.g. reverse lookup for member/user certs).
4. All the other tables are prefixed with `ccf.gov` as they can be read by the constitution/proposals/votes (*) and are useful for audit. 

I've also tried to namespace tables by groups but this is also open for discussions. 

(*) Do we actually want to let members write to `ccf.gov` tables from JS proposals directly, without going through the well defined C++ proposal API?


src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
     static constexpr auto NODES = "public:ccf.gov.nodes";
-    static constexpr auto VALUES = "public:ccf.gov.values";
-    static constexpr auto CONSENSUS = "public:ccf.gov.consensus";
-    static constexpr auto WHITELISTS = "public:ccf.gov.whitelists";
+    static constexpr auto NODE_CODE_IDS = "public:ccf.gov.nodes.code_ids";
+
+    // Service information
+    static constexpr auto SERVICE = "public:ccf.gov.service";
+    static constexpr auto CONFIGURATION = "public:ccf.gov.service.config";
+    static constexpr auto VALUES = "public:ccf.internal.values";
```
jumaffre, line 88: This will go as part of https://github.com/microsoft/CCF/issues/1378

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    static constexpr auto SHARES = "public:ccf.internal.shares";
+    static constexpr auto SUBMITTED_SHARES =
+      "public:ccf.intenal.encrypted_submitted_shares";
+    static constexpr auto SNAPSHOT_EVIDENCE =
+      "public:ccf.internal.snapshot_evidence";
     static constexpr auto SIGNATURES = "public:ccf.internal.signatures";
 
+    // Consensus
+    static constexpr auto CONSENSUS = "public:ccf.internal.consensus";
+    static constexpr auto AFT_REQUESTS = "ccf.internal.consensus.requests";
```
jumaffre, line 120: This was public, which didn't seem right. @ashamis?

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    static constexpr auto NODE_CODE_IDS = "public:ccf.gov.nodes.code_ids";
+
+    // Service information
+    static constexpr auto SERVICE = "public:ccf.gov.service";
+    static constexpr auto CONFIGURATION = "public:ccf.gov.service.config";
+    static constexpr auto VALUES = "public:ccf.internal.values";
+
+    // Governance
     static constexpr auto PROPOSALS = "public:ccf.gov.proposals";
     static constexpr auto GOV_SCRIPTS = "public:ccf.gov.governance.scripts";
```
jumaffre, line 92: Should this just be `public:ccf.gov.scripts`?

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    static constexpr auto MEMBER_DIGESTS =
+      "public:ccf.internal.members.digests";
+
+    // User identities
     static constexpr auto USERS = "public:ccf.gov.users";
-    static constexpr auto USER_CERT_DERS = "public:ccf.gov.user_cert_ders";
-    static constexpr auto USER_DIGESTS = "public:ccf.gov.user_digests";
+    static constexpr auto USER_CERT_DERS =
+      "public:ccf.internal.users.certs_der";
+    static constexpr auto USER_DIGESTS = "public:internal.gov.users.digests";
```
eddyashton, line 77: This doesn't match the naming conventions of other tables.

```suggestion
    static constexpr auto USER_DIGESTS = "public:ccf.internal.users.digests";
```

src/node/entities.h
ashamis: I did not realize this was public, this should not be. 

I created #2167 to resolve this. There maybe some carry on affects there so I will do this in a separate PR

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
     static constexpr auto SERVICE_PRINCIPALS =
```
eddyashton, line 78: Quite a general thought, but raising it here:

We want to allow apps to define their own proposal actions, on their own tables. This is actually an instance of a table which should really be in the app, not the framework. I think the JS application tables are the same - they shouldn't be in every app. And if they're not in every app, I don't think they should be `ccf.` prefixed - the point is to reserve that for names which the _service_ needs, and apps should _not_ be allowed to modify any `ccf.` tables. So if we're trying to get to final naming, maybe we should remove `ccf.` from these? And then I'm not sure whether it makes sense to keep `gov.` - its probably still a helpful distinction, but is it a guideline or a requirement for app proposed tables?

achamayou:
> Do we actually want to let members write to ccf.gov tables from JS proposals directly, without going through the well defined C++ proposal API

For user-defined tables, such as service_principals, yes.

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    // JWT issuers
+    static constexpr auto CA_CERT_DERS = "public:ccf.gov.jwt.ca_certs_der";
+    static constexpr auto JWT_ISSUERS = "public:ccf.gov.jwt.issuers";
     static constexpr auto JWT_PUBLIC_SIGNING_KEYS =
-      "public:ccf.gov.jwt_public_signing_keys";
+      "public:ccf.gov.jwt.public_signing_keys";
     static constexpr auto JWT_PUBLIC_SIGNING_KEY_ISSUER =
-      "public:ccf.gov.jwt_public_signing_key_issuer";
-    static constexpr auto ENDPOINTS = "public:ccf.gov.endpoints";
+      "public:ccf.gov.jwt.public_signing_key_issuer";
```
eddyashton, line 106: Another general discussion point, without an actionable recommendation yet:

The `JWT` and `users` tables don't need to be in every service. There's a combination of proposals and tables and auth policies which all apps _may_ use, and since they're framework-standard I think its right that they get `ccf.gov.` prefixes. But if an app purely uses its own auth policies, ideally it should be able to opt out of these entirely. The maps are fine - they're simply never written to so never appear in the ledger. But we should think about hiding the proposals too, and encapsulating all of these definitions in a more obviously optional package.

achamayou: Unless the governance decides to use them, the overhead is nil. If it was just tables + proposals, we could make them pure JS after the conversion, but there's the polling job which needs to be done natively, so it's likely to stay that way in the foreseeable future.

achamayou:
This needs a changelog entry, because it breaks ledger compatibility.

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    static constexpr auto USER_DIGESTS = "public:ccf.internal.users.digests";
     static constexpr auto SERVICE_PRINCIPALS =
       "public:ccf.gov.service_principals";
+
+    // Nodes identities and allowed code ids
     static constexpr auto NODES = "public:ccf.gov.nodes";
-    static constexpr auto VALUES = "public:ccf.gov.values";
-    static constexpr auto CONSENSUS = "public:ccf.gov.consensus";
-    static constexpr auto WHITELISTS = "public:ccf.gov.whitelists";
+    static constexpr auto NODE_CODE_IDS = "public:ccf.gov.nodes.code_ids";
```
achamayou, line 83: Seems really confusing to have `nodes` be a and a namespace.

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+
+    // Service information
+    static constexpr auto SERVICE = "public:ccf.gov.service";
+    static constexpr auto CONFIGURATION = "public:ccf.gov.service.config";
```
achamayou, line 87: Likewise for service.

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
     static constexpr auto GOV_HISTORY = "public:ccf.gov.governance.history";
-    static constexpr auto SERVICE = "public:ccf.gov.service";
-    static constexpr auto SHARES = "public:ccf.gov.shares";
-    static constexpr auto CONFIGURATION = "public:ccf.gov.config";
-    static constexpr auto SUBMITTED_SHARES = "public:ccf.gov.submitted_shares";
-    static constexpr auto SNAPSHOT_EVIDENCE =
-      "public:ccf.gov.snapshot_evidence";
-    static constexpr auto CA_CERT_DERS = "public:ccf.gov.ca_cert_ders";
-    static constexpr auto JWT_ISSUERS = "public:ccf.gov.jwt_issuers";
+    static constexpr auto WHITELISTS = "public:ccf.gov.whitelists";
```
achamayou, line 93: Is this member writable? It also needs renaming.

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
     static constexpr auto JWT_PUBLIC_SIGNING_KEYS =
-      "public:ccf.gov.jwt_public_signing_keys";
+      "public:ccf.gov.jwt.public_signing_keys";
     static constexpr auto JWT_PUBLIC_SIGNING_KEY_ISSUER =
-      "public:ccf.gov.jwt_public_signing_key_issuer";
-    static constexpr auto ENDPOINTS = "public:ccf.gov.endpoints";
+      "public:ccf.gov.jwt.public_signing_key_issuer";
 
+    // Internal only
+    static constexpr auto SECRETS = "public:ccf.internal.secrets";
```
achamayou, line 109: ledger.secrets, ledger_secrets?

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    static constexpr auto SHARES = "public:ccf.internal.shares";
```
achamayou, line 110: recovery_shares?

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    static constexpr auto SUBMITTED_SHARES =
+      "public:ccf.intenal.encrypted_submitted_shares";
```
achamayou, line 112:
```suggestion
      "public:ccf.internal.encrypted_submitted_shares";
```

src/node/entities.h
```diff
@@ -60,50 +60,68 @@ namespace ccf
...
+    static constexpr auto SNAPSHOT_EVIDENCE =
+      "public:ccf.internal.snapshot_evidence";
     static constexpr auto SIGNATURES = "public:ccf.internal.signatures";
+    static constexpr auto VALUES = "public:ccf.internal.values";
 
+    // Consensus
+    static constexpr auto CONSENSUS = "public:ccf.internal.consensus";
+    static constexpr auto AFT_REQUESTS = "ccf.internal.consensus.requests";
+    static constexpr auto NEW_VIEWS = "public:ccf.internal.consensus.new_views";
```
achamayou, line 121: is this new_views or really just views?

achamayou: Or even view_evidence, like snapshot_evidence?

src/node/entities.h
achamayou: encrypted_ledger_secrets

src/node/entities.h
jumaffre: Thanks for the clarification. I've removed the `public` prefix so the table should be private now.

src/node/entities.h
jumaffre: I believe it is. It is set to a default value (see `default_whitelists.h`) but I'm not sure we have any end-to-end test that modify this table. I'm assuming that this will change/disappear once we have JS governance?

src/node/entities.h
jumaffre: I went for it.

src/node/entities.h
```diff
@@ -60,52 +60,70 @@ namespace ccf
...
+    // User identities
+    static constexpr auto USERS = "public:ccf.gov.users.info";
+    static constexpr auto USER_CERT_DERS =
+      "public:ccf.internal.users.certs_der";
+    static constexpr auto USER_DIGESTS = "public:ccf.internal.users.digests";
     static constexpr auto SERVICE_PRINCIPALS =
       "public:ccf.gov.service_principals";
-    static constexpr auto NODES = "public:ccf.gov.nodes";
-    static constexpr auto VALUES = "public:ccf.gov.values";
-    static constexpr auto CONSENSUS = "public:ccf.gov.consensus";
```
jumaffre, line 75: I've merged this table with `CONFIGURATION` and deleted it.

src/node/entities.h
jumaffre: For this specific table (`SERVICE_PRINCIPALS`), my understanding is that it will be removed once we move to JS governance and dynamic tables are supported (what you described above). Renaming it to `public:ccf.app.service_principals` makes sense though.

Are you also referring to the 3 JS tables (scripts, modules and endpoints)? If yes, it seems that they are very much hardcoded in the member frontend to be able to bootstrap a JS app. I've also prefixed them with `public:ccf.app.`.

src/node/entities.h
jumaffre: @ashamis: any preference?

src/node/entities.h
ashamis: @jumaffre - can you please leave it as public for now. This will conflict with the current work I am doing.

src/node/entities.h
eddyashton: They're in `member_frontend` currently, but those proposals don't need to be there anymore than `SERVICE_PRINCIPALS` (well maybe `modules` is reused for governance modules, but the gist remains). The proposals could be defined in the governance script, or even in `js_generic.cpp`. Eventually, hopefully, ideally.

jumaffre: Good points. I wonder whether they should still be prefixed (either `public:ccf.app` or `public:app.meta`) so that they don't clash with app-defined tables though? I believe we don't want app endpoints to read/write in these tables?

eddyashton: I'd lean towards giving them a `gov.` prefix. So proposals can write to (and are the _only_ thing that can write to) `gov.` and `ccf.gov.`. Then apps can't write to these tables because they're "governance", rather than because they're CCF.

jumaffre: Using `gov.` prefix now.

src/node/entities.h
achamayou: Yes

achamayou (approved):

PR #2173, opened 2021-02-05
Only run fuzzer on daily build

achamayou:


.azure-pipelines-templates/matrix.yml
```diff
@@ -61,17 +61,6 @@ jobs:
       artifact_name: "SGX_Perf"
       ctest_filter: "${{ parameters.test.perf.ctest_args }}"
 
-  # Fuzzing
```
eddyashton, line 64: This looks like its just removing the Fuzzing test entirely - where is it added to the Daily?

achamayou: Sorry, git mistake, thanks for catching, will fix.

PR #2174, opened 2021-02-05
Record version at which the previous ledger secret was written to store

jumaffre:
Follow up from #2140 

Now that we have #2137, actually record the version at which the previous ledger secret was written to the store in the encrypted past ledger secret table. This will be useful to query the ledger for historical secrets after recovery from a snapshot.


achamayou (approved):

eddyashton (approved):

jumaffre:
Merging this as the daily failure (rotation test) is unrelated.

PR #2175, opened 2021-02-05
Reorder and cull perf metrics

eddyashton:
Since we're renaming our CIMetrics labels and losing all history, take this opportunity to also:

- Remove `secp256k1`, since we don't care about its signing rate enough any more (and have the full picobench when we do)
- Move the `mem` decorator to suffix rather than prefix, so we get the memory use of each test adjacent to the throughput, rather than all mem followed by all throughput

I'm also wondering if `ls` is too brief and opaque a name for what was `logging_scenario`, is `log` better? Scream if you want to go longer.

### DIFF
--- a/tests/infra/runner.py
+++ b/tests/infra/runner.py
@@ -199,9 +199,10 @@ def run(get_command, args):
                         tx_rates.insert_metrics(**results)
 
                         # Construct name for heap metric, removing ^ suffix if present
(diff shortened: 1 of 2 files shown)


achamayou:
FWIW, I think `ls` and `sb` are fine, they're never going to be descriptive enough anyway to carry the full meaning of the benchmark, so they might as well be short.

achamayou (approved):

Issue #1459, opened 2020-07-29
JS Runtime execution efficiency and sandboxing improvements


achamayou:
Performance items:

- [x] ~~Share runtime across execution contexts? Per-session?~~ (experiment: https://github.com/microsoft/CCF/pull/2162)
- [x] Investigate performance benefit from caching pre-compiled bytecode? Implemented in #2643.

Sandboxing/Fairness:
- [x] Expose context memory limit to either governance or endpoint schema (JS_SetMemoryLimit) #4344
- [x] Add execution cap (JS_SetInterruptHandler) #4396

letmaik:
For the limits, I would suggest to store global (not per-endpoint) settings. I don't think there's a good use case for per-endpoint settings and keeping it separate from the app deployment proposal (which would include per-endpoint metadata) makes it more visible. I suggest a new proposal `set_js_engine_options` with fields `max_stack_size`, `max_heap_size`, and  `max_execution_time`.

achamayou:
+1 on the names.

I think there may eventually be a case for split values because I can imagine a scenario where public/quasi public-endpoints have lower limits than endpoints for which access is restricted to a small number of admin-like identities who occasionally need to run more expensive tasks (GC-like, analytics...). Right now though, a global limit is fine.

letmaik:
The interrupt handler set by `JS_SetInterruptHandler` is called (via `js_poll_interrupts()`) at the following points, mostly to avoid going into infinite loops/recursion without allowing "ctrl-c":
- function calls
- `instanceof`
- some opcodes: goto (for exceptions etc.), if_true, if_false

Given that, we could enforce a limit based roughly on number of function calls, but this would be a very weak guarantee.
The alternative is to use untrusted time to measure execution time, which likely makes more sense.
Any opinions?

eddyashton:
As additional motivation for sharing runtime/context, we've recently seen a measurable drop in the `ls_js_sgx_cft` benchmark from a seemingly innocuous change - #2569:

![image](https://user-images.githubusercontent.com/6000239/117957737-44ccfc80-b312-11eb-9076-4036f73866d4.png)

After further investigation, it turns out this is wholly from the increase in the size of the JS script, where this PR adds a few functions to the end. We can see a similar drop from anything that increases the size of the JS script, for instance adding a large string constant to the start:

```
let s = `Lorem ipsum <SNIP>`; // A 12K string
```
Local test throughput dropped from ~1.8kTPS to ~1.4kTPS.

[...]

PR #2177, opened 2021-02-05
Add js heap size limit & tests

letmaik:
Part of https://github.com/microsoft/CCF/issues/1459. Next step will be to make the hard-coded limits configurable.


achamayou (approved):

eddyashton (approved):

letmaik:
The SGX stack overflow test still crashes, likely because the enclave runs out of stack space. It has 4 MiB in total, and quickjs's limit is set to 1 MiB. That means for it to crash the rest of CCF needs to use more than 3 MiB? I wonder if there is a way to check stack usage at the point when the endpoint handler is called, simply to fail hard if the requested quickjs stack limit + existing usage would exceed the enclave limit.

letmaik:
Turns out the stack check wasn't enabled for the enclave build. @achamayou I had to patch quickjs.c to enable this (see changes). As part of https://github.com/microsoft/CCF/pull/2162 I moved the patches to a separate folder, and although this patch doesn't allow this easily I think the next-best thing would be to let the build fail if `CONFIG_STACK_CHECK` is not defined (which would indicated a faulty merge of newer quickjs versions).

cmake/quickjs.cmake
```diff
@@ -29,7 +29,7 @@ if("sgx" IN_LIST COMPILE_TARGETS)
   )
   target_compile_options(
     quickjs.enclave PUBLIC -nostdinc -DCONFIG_VERSION="${QUICKJS_VERSION}"
-                           -DEMSCRIPTEN
+                           -DEMSCRIPTEN -DENABLE_STACK_CHECK
```
achamayou, line 32: define CONFIG_STACK_CHECK and avoid having to patch quickjs?

letmaik: Genius.

src/apps/js_generic/js_generic.cpp
```diff
@@ -969,6 +969,7 @@ namespace ccfapp
       JSAutoFreeRuntime auto_free_rt(rt);
 
       JS_SetMaxStackSize(rt, 1024 * 1024);
+      JS_SetMemoryLimit(rt, 20 * 1024 * 1024);
```
achamayou, line 972: This seems low, we typically run with 500Mb of heap, can we make it 100Mb at least?

achamayou (approved):

PR #2179, opened 2021-02-07
Consolidate some end to end testing envs

achamayou:
Checked on rekey CFT, out of a roughly 10 seconds end to end test, there's about 4 or 5 seconds of setup. Another step I suspect we could take is to merge all the JS tests into a single environment, and switch app rather than create an fresh environment.

We have about 30 end to end envs in total, so if we consolidate down to say about 10, that's two minutes saved every CI.

This also enables the websockets and receipts tests for BFT.


tests/e2e_logging.py
```diff
@@ -799,6 +875,9 @@ def run(args):
         network = test_network_node_info(network, args)
         network = test_metrics(network, args)
         network = test_memory(network, args)
+        if args.package == "liblogging":
```
achamayou, line 878: What we save in setup time, we spend again testing in bft, which we didn't before. Still a win.

tests/e2e_logging.py
```diff
@@ -799,6 +895,13 @@ def run(args):
         network = test_network_node_info(network, args)
         network = test_metrics(network, args)
         network = test_memory(network, args)
+        if args.consensus == "cft":
```
achamayou, line 898: Re-key causes BFT to hang right now.

jumaffre (approved):

PR #2180, opened 2021-02-08
[multi-thread] Reset history after verifying snapshot

### LINKED ISSUE #2186: Do not use main store when verifying snapshot evidence
Follow-up from https://github.com/microsoft/CCF/pull/2180.

CCF uses the main store to verify the snapshot evidence (so that is uses the same code path as the public recovery). This adds some complexity on the verification logic where the history has to be hidden from other modules (e.g. frontend) until the snapshot is verified. 

We should simplify this and use a different store to check the validity of the snapshot, with its own history.

jumaffre:
Resolves #2186 

When joining from a snapshot in recovery, the node ends up with two histories (one used to verify the snapshot, and another one initialised once the join request has been successful). ~We now reset the history once the snapshot has been verified.~

Edit: We now use a separate store to verify the snapshot when joining from one, and discard it once the snapshot is verified.


cmake/common.cmake
```diff
@@ -339,7 +339,7 @@ install(
 
 # Common test args for Python scripts starting up CCF networks
 set(WORKER_THREADS
-    0
+    2
```
jumaffre, line 342:
```suggestion
    0
```

src/node/node_state.h
```diff
@@ -1770,29 +1808,23 @@ namespace ccf
 
     void setup_history()
     {
-      // This function can be called once the node has started up and before
-      // it has joined the service.
+      // Signature generation is disabled on temporary history
```
jumaffre, line 1811:
```suggestion
```

src/node/snapshotter.h
```diff
@@ -27,7 +28,8 @@ namespace ccf
     ringbuffer::WriterPtr to_host;
     SpinLock lock;
 
-    NetworkState& network;
+    // NetworkState& network;
```
jumaffre, line 31:
```suggestion
```

achamayou (approved):

Issue #2153, opened 2021-02-02
Accept plain strings for string query parameters of C++ endpoints


letmaik:
Currently, it is `/network/nodes?host="123.456..."` because everything is JSON, but this is clearly undesirable and also conflicts with the generated OpenAPI schema.

achamayou:
Yes, this aspect of the json wrapper is a (the last?) leftover from the jsonrpc days, it clearly needs removing now.

eddyashton:
I took another look at this, convinced myself it was a trivial fix, and instantly hit the "oh no this doesn't work" wall.

My attempted fix was just "if parsing as JSON fails, keep the string". So in `get_params_from_query`,
```
catch (const std::exception& e)
{
  throw UrlQueryParseError(fmt::format(
    "Unable to parse URL query value: {} ({})", query, e.what()));
}
```
became
```
catch (const nlohmann::json::parse_error& e)
{
  params[std::string(key)] = value;
}
```
Easy peasy. But!

If you remove the quotes from `?host=1.2.3.4`, then our attempt to parse it succeeds in the wrong way - it reads a JSON number, not a string!

So this 2-step "try and find a JSON type" now butts against the "I expected you to find the right JSON type for my conversion" later. I don't see an escape hatch here, and I think the answer is to drop any JSON magic conversion for query parameters. We can build a string->string map for you, but from there the parsing is up to the app.

achamayou:
> I think the answer is to drop any JSON magic conversion for query parameters. We can build a string->string map for you, but from there the parsing is up to the app.

I completely agree.

PR #2188, opened 2021-02-09
Fix '/node/network/nodes' endpoint to return all nodes when no filter is used

### LINKED ISSUE #2185: /network/nodes API returning error when filter isn't passed as query parameter
**Describe the bug**
/network/nodes API is returning error when filter isn't passed in the query parameter. However, when the call is made with host filter, for e.g.` curl 'https://127.0.0.1:8000/node/network/nodes?host="127.0.0.1"' -X GET`, expected response is returned.

**To Reproduce**
Call /network/nodes API on deployed CCF network.
`curl 'https://127.0.0.1:8000/node/network/nodes' -X GET`
The following error is returned:
`
{
  "error": {
    "code": "InvalidInput",
    "message": "At #/: Expected object, found: null"
  }
}
`

**Expected behavior**
The expected response should contain a list of all nodes in the network if query parameter doesn't contain host or port filter.

**Environment information**
Same issue whether CCF network is running in virtual environment or on enclaves

**Additional context**
Changes to the network API was made as part of the following release: https://github.com/microsoft/CCF/releases/tag/ccf-0.17.0

achamayou:
fix #2185 

Longer term, we need to get to #2153

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -17,7 +17,6 @@ import tempfile
 import base64
 import json
 import ccf.clients
-import pprint
 
 from loguru import logger as LOG
 
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -17,6 +17,7 @@ import tempfile
 import base64
 import json
 import ccf.clients
+import pprint
 
 from loguru import logger as LOG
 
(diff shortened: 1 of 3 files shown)


letmaik (approved):

tests/e2e_logging.py
```diff
@@ -17,6 +17,7 @@
 import base64
 import json
 import ccf.clients
+import pprint
```
jumaffre, line 20: To remove?

jumaffre (approved):

tests/e2e_logging.py
achamayou: Uh yeah, thanks!

Issue #2189, opened 2021-02-10
rotation_test failure


eddyashton:
`rotation_test` has been failing regularly in the CI, example runs [here](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=18162&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=bbcdfa66-260b-5619-96df-9537b476c35e) and [here](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=18599&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=bbcdfa66-260b-5619-96df-9537b476c35e). Primarily, but not exclusively, in the `NoSGX_Instrumented` build, suggesting it is a timing issue. The failure appears to be an election deadlock; the surviving nodes get into a cycle of elections, and although they pass we continue to start subsequent elections.

We disabled it [here](https://github.com/microsoft/CCF/pull/2178) to avoid blocking PRs with unrelated CI failures.

I've finally obtained a repro on my local machine, so here's a representative log walkthrough. Just looking at the logs and pointing out anything odd for now, will clarify this with the code later.

#### Infra
```
47: 09:56:20.958 | INFO     | infra.node:resume:355 - Node 11 has resumed from suspension.
47: 09:56:20.976 | INFO     | reconfiguration:check_can_progress:35 - [12||] GET /node/commit
47: 09:56:20.976 | INFO     | reconfiguration:check_can_progress:35 - 200 {"seqno":137,"view":17}
47: 09:56:20.999 | INFO     | reconfiguration:check_can_progress:37 - [12|user0|] POST /app/log/private {'id': 42, 'msg': 'Hello world'}
47: 09:56:20.999 | INFO     | reconfiguration:check_can_progress:37 - 200 @17.138 true
47: 09:56:21.001 | INFO     | reconfiguration:check_can_progress:40 - [12||] GET /node/commit
47: 09:56:21.001 | INFO     | reconfiguration:check_can_progress:40 - 200 {"seqno":137,"view":17}
<-SNIP->
47: 09:56:22.020 | INFO     | reconfiguration:check_can_progress:40 - [12||] GET /node/commit
47: 09:56:22.020 | INFO     | reconfiguration:check_can_progress:40 - 200 {"seqno":139,"view":17}
47: 09:56:22.020 | SUCCESS  | suite.test_requirements:wrapper:19 - Suspend and resume primary
[...]

PR #2192, opened 2021-02-12
Raft election bugfix: Accept post-election rollbacks from new primary

eddyashton:
A potential fix for the issue identified and discussed in #2189.

The real change here is the modification to `raft.h`, adding an additional rollback on a single path when our uncommitted suffix doesn't match the primary's. This seems to fix the stalling behaviour that was previously seen, but needs some more thought to confirm its safe in general/doesn't break some other property.

I've also removed `--long-tests` as an `e2e_arg`. It wasn't used anywhere other than `rotation.py`, so now the `LONG_TESTS` CMake arg is used to set per-test options, rather than hardcoding the iterations in the test itself. This makes it easier to run `rotation_test` with 100 suspensions but no retirements (a fast way to trigger the crash) by hand.

This includes a temporary new job which should just run `rotation_test` in a loop. The bare minimum for this PR is that it convincingly makes `rotation_test` stable, but we'll also aim to write some reasoning about the change before merging it.

TODO:
- [ ] See how the CI feels about this change
- [x] Remove new rotation_test-in-a-loop job
- [x] Add a safety check in `rollback` that forbids rolling back committed state
- [x] Convince ourselves that the new rollback doesn't drop any entries it should have retained
- [ ] ~~Rename rather than delete truncated ledger files on the host (this is generally a good idea for operational safety, though actually reconciling the forked ledger suffices is left as an exercise for the reader)~~ Discussion to be resolved in future PR: #2197


src/consensus/aft/raft.h
```diff
@@ -1008,12 +1014,13 @@ namespace aft
     {
       std::unique_lock<SpinLock> guard(state->lock);
       LOG_DEBUG_FMT(
-        "Received pt: {} pi: {} t: {} i: {} toi: {}",
+        "Received append entries: {}.{} to {}.{} (primary is {} in term {})",
```
eddyashton, line 1017: I think this rewording makes this log line much more readable. I'm tempted to do a general rewrite of all the logging in `raft.h` to make it more concise and consistent, but I'm wary of losing any information/patterns that other people rely on. In any case it likely remains incredibly spammy and verbose, so probably not worth doing.

src/consensus/aft/raft.h
```diff
@@ -1167,6 +1174,24 @@ namespace aft
...
+      if (is_new_follower)
+      {
+        if (state->last_idx != r.prev_idx)
+        {
+          LOG_DEBUG_FMT(
+            "New follower received first append entries with mismatch - "
+            "rolling back from {} to {}",
+            state->last_idx,
+            r.prev_idx);
+          rollback(r.prev_idx);
```
jumaffre, line 1186: This shouldn't ever happen but could we `throw` in `rollback()` if we rollback before `commit_idx`?

eddyashton: Yeah, I've added this as `TODO` on this PR.

src/consensus/aft/raft.h
```diff
@@ -1167,6 +1174,24 @@ namespace aft
         r.idx,
         r.prev_idx);
 
+      if (is_new_follower)
+      {
+        if (state->last_idx != r.prev_idx)
```
jumaffre, line 1179: If this node was a lagging follower in the previous term and is told by the new primary about `r.prev_idx`, which is in the future (`r.prev_idx > state->last_idx`), the we would rollback in the future? I'm assuming this is fine as all actions performed by `rollback()` have no effect in this case, but I wonder if the check here should be `state->last_idx > r.prev_idx` instead?

src/consensus/aft/raft.h
achamayou: All the shorthand stuff is old and from the driver test, I like the view.seqno notation better.

src/consensus/aft/raft.h
```diff
@@ -102,6 +102,12 @@ namespace aft
     bool is_execution_pending = false;
     std::list<std::unique_ptr<AbstractMsgCallback>> pending_executions;
 
+    // When we receive append entries from a new primary, we should rollback
```
achamayou, line 105: Well, we should not roll back, to avoid a potential loss of persistence.

src/consensus/aft/raft.h
eddyashton: I've added this in `rollback()`, though as a `LOG_FAIL` rather than a throw. I think we want to avoid crashing in this case if we 
can so opted for logging, but in practice this likely means something later breaks because we're in an unexpected state. Does anyone have a strong argument for throwing here instead?

src/consensus/aft/raft.h
eddyashton: Changed this to `state->last_idx > r.prev_idx`.

tests/infra/network.py
```diff
@@ -802,12 +802,14 @@ def wait_for_node_commit_sync(self, timeout=3):
             for node in self.get_joined_nodes():
                 with node.client() as c:
                     r = c.get("/node/commit")
-                    commits.append(f"{r.view}.{r.seqno}")
+                    assert r.status_code == http.HTTPStatus.OK.value
```
eddyashton, line 805: This function `wait_for_node_commit_sync` has been broken for a while, because `/node/commit` no longer returns view+seqno in headers, so we were constructing a list of `None.None` strings which is identical enough to pass the check. This change extracts view and seqno from the JSON body.

tests/rotation.py
```diff
@@ -22,28 +26,112 @@ def test_suspend_primary(network, args):
     return network
 
 
+@reqs.description("Suspend and resume all nodes while submitting constant client load")
+@reqs.at_least_n_nodes(3)
+def test_rolling_disconnections(network, args):
```
eddyashton, line 31: This was an attempt at writing a test which repros the issue more consistently than repeated calls to `test_suspend_primary`. In practice its still a probabilistic repro, in an idiosyncratic and expensive test. I'm going to remove it since we now have a better test in `committable.py`. Long-term, we should probably run some suite of tests while separately driving load on the service via `vegeta`, but we shouldn't apply that load per-test like this.

src/consensus/aft/raft.h
```diff
@@ -102,10 +102,14 @@ namespace aft
     bool is_execution_pending = false;
     std::list<std::unique_ptr<AbstractMsgCallback>> pending_executions;
 
-    // When we receive append entries from a new primary, we should rollback
-    // to the index they're telling us about, and replay from here. However
-    // for complex reasons we want to do this once only, and expect to be
-    // correct without doing this thereafter.
+    // When this node receives append entries from a new primary, it may need to
+    // roll back an uncommitted suffix it holds. It must trust the new primary
```
achamayou, line 106: committable but uncommitted, especially

src/consensus/aft/raft.h
```diff
@@ -102,10 +102,14 @@ namespace aft
...
+    // to dictate this index, and trust it to be honest about its approximation
```
achamayou, line 107: I don't think approximation of committed state is the right description here, the new primary sends its committable suffix, and does not make a statement about whether that is committed or not.

src/consensus/aft/raft.h
```diff
@@ -102,10 +102,14 @@ namespace aft
...
+    // of committed state. To minimise the window where a malicious node could
+    // manipulate this, and to avoid retransmission of a suffix already hold,
+    // this node only executes this rollback rollback instruction on the first
```
achamayou, line 110: I'm not sure if the double rollback is here for comical effect, but I like it. I'm wary of talking about malicious nodes here and elsewhere in CFT, where we do not make guarantees about them. We rightly take this step because it's the conservative thing to do, and anything else is a bug.

tests/infra/network.py
achamayou: Thank you for catching that!

tests/committable.py
```diff
@@ -50,33 +50,72 @@ def run(args):
...
+        last_tx = committable_txs[-1]
+        sig_view, sig_seqno = last_tx.view, last_tx.seqno + 1
         with backups[0].client() as bc:
             wait_for_pending(bc, sig_view, sig_seqno)
 
-        # Kill the primary, restore other backups
-        primary.stop()
+        # Suspend the final backup and run some transactions which only the primary hears, which should be discarded by the new primary
+        backups[0].suspend()
+        uncommittable_txs = []
```
achamayou, line 68: I suggest a different name, because we call a committable transaction anything that has a valid signature after it. Maybe post_partition_txs or something like that.

src/consensus/aft/raft.h
jumaffre: I'd say `throw` as we would be able to detect the error from a unit test rather than logging an error without failing the test. Also, I think most of the Raft logic (`recv_message()`) is under a big `try {} catch {}`.

tests/committable.py
```diff
@@ -50,33 +50,72 @@ def run(args):
...
 
-        sig_view, sig_seqno = txs[-1].view, txs[-1].seqno + 1
+        last_tx = committable_txs[-1]
+        sig_view, sig_seqno = last_tx.view, last_tx.seqno + 1
         with backups[0].client() as bc:
             wait_for_pending(bc, sig_view, sig_seqno)
 
-        # Kill the primary, restore other backups
-        primary.stop()
+        # Suspend the final backup and run some transactions which only the partitioned primary hears, which should be discarded by the new primary
```
jumaffre, line 66: Minor: very long line?

src/consensus/aft/raft.h
```diff
@@ -102,6 +102,17 @@ namespace aft
...
+    // When this node receives append entries from a new primary, it may need to
+    // roll back a committable but uncommitted suffix it holds. The
+    // new primary dictates the index where this suffix begins, which
+    // following the Raft election rules must be at least as high as the highest
+    // commit index reported by the previous primary. The window in which this
+    // rollback could be accepted is minimised to avoid unnecessary
+    // retransmissions - this node only executes this rollback instruction on
+    // the first append entries after it became a follower. As with any append
+    // entries, the initial index will not advance until this node acks.
+    bool is_new_follower = false;
```
achamayou, line 114: This is an accurate state flag name, but I wonder if a condition-based name like `waiting_for_initial_view_seqno` or `pending_end_of_view_rollback` would be more descriptive. Given the extensive comment, this is nitpicking at best.

src/consensus/aft/raft.h
achamayou: I don't think we should throw here unless we're sure this is not recoverable, which it may be. It's good to fail early in tests, but in production we want to be defensive.

tests/infra/node.py
```diff
@@ -254,6 +255,8 @@ def _read_ports(self):
 
     def stop(self):
         if self.remote and self.network_state is not NodeNetworkState.stopped:
+            if self.suspended:
```
achamayou, line 258: Why?

tests/infra/node.py
```diff
@@ -347,12 +350,16 @@ def client(self, identity=None, signing_identity=None, **kwargs):
         return ccf.clients.client(self.pubhost, self.pubport, **akwargs)
 
     def suspend(self):
-        self.remote.suspend()
-        LOG.info(f"Node {self.node_id} suspended...")
+        if not self.suspended:
```
achamayou, line 353: I think we want to raise here, this is test infra, we don't want it silently not suspending or resuming.

achamayou (approved):

tests/infra/node.py
eddyashton: We terminate with `SIGTERM`, which is ignored if the process is currently suspended, so if the infra tries to teardown the network (because something failed) while a node is suspended, it will hang waiting for a node termination that never happens. There may be a watchdog timeout we can set elsewhere, but this seems sufficient.

tests/infra/node.py
eddyashton: Added asserts.

tests/committable.py
eddyashton: Manually reflowed this over multiple lines. Wary of doing this too often, and surprised that black has no option to do it automatically.

src/node/channels.h
```diff
@@ -476,18 +476,25 @@ namespace ccf
...
+        channels.emplace_hint(search, peer_id, std::move(channel));
+      }
+      else if (!search->second->is_outgoing())
       {
         // Channel with peer already exists but is incoming. Create host
         // outgoing connection.
+        LOG_DEBUG_FMT("Setting existing channel to {} as outgoing", peer_id);
         search->second->set_outgoing(hostname, service);
         return;
       }
```
eddyashton, line 497: @achamayou + @jumaffre: Highlighting this related bugfix that this PR exposed. The `try_emplace` was unafe; if the channel already exist then we briefly construct a temporary `Channel`, and then immediately call its destructor which sends a message to the host closing the outgoing connection.

I don't fully grok why this is happening so often with these PR changes - something something candidate-to-candidate transitions are more common something something? But hopefully this clears up the CI instability on this PR, so I'm rolling the fix in here.

Issue #1993, opened 2020-12-08
KV should serialise governance tables as JSON


eddyashton:
Currently, the default is to serialise all tables as msgpack. For easier auditing and interoperability with JS, we'd like to default to JSON - at least for our governance tables and perhaps for (almost) everything. Some tables may need faster serialisers for rapid lookup, such as from caller's DER to caller ID.

There's an incomplete PR attempting this, with several unexpected but seemingly required additions, here: #1860.

achamayou:
After discussion with @syhamza, there's interest in getting at least the signature and nodes tables to a stable, auditable state as soon as possible.

Some thoughts on Signatures:

1. Most fields are well-suited to the conversion with little performance hit
2. The root of the tree needs conversion to hex, to be consistent with our representation of digests throughout the API.
3. Storage of the tree snapshot is a very poor fit, since it's a large binary blob. I think we can split it to a separate table (public:ccf.internal.tree), and write it raw. This isn't totally trivial, some checks we have on signatures will need adjusting, but I'm exploring the impact.

@jumaffre @eddyashton, feedback welcome. I'll look at the nodes once I'm done with signatures.

jumaffre:
For the serialised Merkle tree in the signature tx, two of the main impacts I can think of are:
1. When applying a snapshot, we assume that the snapshot was taken at the `seqno` of a signature, which means it is safe to read from the signatures table after applying the snapshot, and seed the new history with the serialised Merkle tree contained in the signature. 
2. For historical queries, we deserialise the Merkle tree contained in the signature to find out if the target tx is indeed validated by that signature.

I think both would be fine with a separate table, as long as the serialised Merkle tree is written to the store at the `seqno` as the signature.

achamayou:
Yes, writes would occur in the same transaction of course.

achamayou:
Dupe #2237

PR #2193, opened 2021-02-12
Add ability for implicit ledger verification in ccf/ledger.py

syhamza:
This PR adds the ability for ledger verification in ccf/ledger.py module. It also adds helper MerkleTree module that LedgerValidator uses to validate the ledger.


sidmore:
should there be tests for this?

syhamza:
I didn't find tests currently existing for this module in CCF. Ideally, I would have piggy backed on those existing tests. I don't think writing UTs for ledger.py would make sense because it'll be complicated to mock ledger chunks. However, we can add UTs for MerkleTree implementation.
Waiting to see what CCF folks suggest.

---
In reply to: [778509161](https://github.com/microsoft/CCF/pull/2193#issuecomment-778509161) [](ancestors = 778509161)

python/ledger_tutorial.py
```diff
@@ -41,4 +41,15 @@
         if target_table in public_tables:
             for key, value in public_tables[target_table].items():
                 target_table_changes += 1  # A key was changed
+
+        # extracting transactions on a sample table
+        # Ledger verification is happening implicitly in ccf.ledger.Ledger()
+        if "public:sample.logs" in public_tables:
+            public_tpal_table = public_tables["public:sample.logs"]
```
jumaffre, line 48: This references a table that doesn't exist in the framework. In this case, using the `target_table` (defined on line `30`) would be simpler/more generic.

python/ccf/merkletree.py
```diff
@@ -0,0 +1,75 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: Any reason a custom implementation of a Merkle tree vs. e.g. https://github.com/Tierion/pymerkletools was preferred?

python/ccf/ledger.py
```diff
@@ -19,6 +29,12 @@
 
 UNPACK_ARGS = {"raw": True, "strict_map_key": False}
 
+# Public table names as defined in CCF
+# https://github.com/microsoft/CCF/blob/main/src/node/entities.h#L96
```
jumaffre, line 33: Minor: better not to reference an exact line in the C++ source code as it is likely to change in the future.

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
...
+    # Signature table contains PrimarySignature which extends NodeSignature. NodeId should be at index 1 in the serialized Node
+    # https://github.com/microsoft/CCF/blob/main/src/node/signatures.h#L23
+    EXPECTED_NODE_SIGNATURE_INDEX = 0
+    EXPECTED_NODE_VIEW_INDEX = 2
+    EXPECTED_ROOT_INDEX = 5
+    # https://github.com/microsoft/CCF/blob/main/src/node/node_signature.h#L76
+    EXPECTED_SIGNING_NODE_ID_INDEX = 1
+    EXPECTED_SIGNATURE_INDEX = 0
+    # Constant for the size of a hashed transaction
+    SHA_256_HASH_SIZE = 32
```
jumaffre, line 214: Unused?

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
...
+        1) The merkle proof is signed by a TRUSTED node in the given network
+        2) The merkle root and signature are verified with the node cert
+        3) The merkle proof is correct for each set of transactions
+        4) The last transaction in the LedgerChunk should be that of merkle tree signing
+    """
+
+    # The node that is expected to sign the signature transaction
+    # The certificate used to sign the signature transaction
+    # https://github.com/microsoft/CCF/blob/main/src/node/nodes.h#L42
+    EXPECTED_NODE_CERT_INDEX = 1
```
jumaffre, line 201: FYI, the deserialisation logic for the built-in tables will be simplified greatly once https://github.com/microsoft/CCF/issues/1993 is implemented.

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
...
+
+    def add_transaction(self, transaction):
+        """
+        To validate the ledger, ledger transactions need to be added via this method.
+        Depending on the tables that were part of the transaction, it does different things.
+        When transaction contains signature table, it starts the verification process and verifies that the root of merkle tree was signed by a node which was part of the network.
+        It also matches the root of the merkle tree that this class maintains with the one extracted from the ledger.
+        If any of the above checks fail, this method throws.
+        """
+        # Start with empty bytes array. CCF MerkleTree uses an empty array as the first leaf of it's merkle tree.
```
jumaffre, line 236: Minor: this comment should be move to the constructor.

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
...
+
+                # validations for 1, 2 and 3
+                # throws if ledger validation failed.
+                if self._verify_tx_set(tx_info):
+                    continue
+                else:
+                    raise ValueError("Ledger validation failed")
+
+        # Checks complete, add this transaction to tree
+        self.merkle.add_leaf(transaction.get_public_tx())
```
jumaffre, line 289: Could you rename `get_public_tx()` as it returns the entire serialised transaction as opposed to just the public doamin?

python/ccf/ledger.py
```diff
@@ -297,19 +520,81 @@ def __init__(self, directory: str):
             ),
         )
 
+        # Verifies there are no missing ledger chunks
+        self._verify_commited_files(sorted_ledgers)
```
jumaffre, line 524: The existing `governance_history_test` is failing on the CI (see https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=18863&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=efdbc34d-c6b6-5dc6-e05c-1751d4c47c20&l=1166) and this call seems to be causing the issue. 

The `governance_history_test` end-to-end test counts the number of governance transactions in the ledger (see https://github.com/microsoft/CCF/blob/main/tests/governance_history.py#L22). At this point, the ledger is likely to be short, with at least one non-committed ledger suffix chunk. Is this supported by the current verification code?

python/ccf/ledger.py
```diff
@@ -241,9 +452,15 @@ def __next__(self):
         try:
             self._complete_read()
             self._read_header()
+
+            # Adds every transaction to the ledger validator
+            # LedgerValidator does verification for every added transaction and throws when it finds any anamoly.
```
jumaffre, line 457:
```suggestion
            # LedgerValidator does verification for every added transaction and throws when it finds any anomaly.
```

jumaffre:
> I didn't find tests currently existing for this module in CCF. Ideally, I would have piggy backed on those existing tests. I don't think writing UTs for ledger.py would make sense because it'll be complicated to mock ledger chunks. However, we can add UTs for MerkleTree implementation.
> Waiting to see what CCF folks suggest.
> 
> In reply to: [778509161](https://github.com/microsoft/CCF/pull/2193#issuecomment-778509161) [](ancestors = 778509161)

We don't currently have any unit test for the ledger Python parser but we use it in a couple of places to parse the ledger created by a CCF node (the Python ledger tutorial and `governance_history_test`). If we want to check the integrity of a larger ledger, I suggest that verifying the CCF ledger at the end of a performance test (in `runner.py`), once the network has been shutdown, would be a great choice.

jumaffre:
The other test failure (https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=18863&view=logs&j=ee9484ec-b476-577b-aa0e-d5ddfcbcd149&t=e2c1a4cf-16d0-5038-f9f6-343f9806d38b&l=9207) is unrelated to your changes.

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
     return ret
 
 
+class NodeStatus(IntEnum):
+    """These are the corresponding status meanings from the ccf.nodes table"""
+
```
achamayou, line 169: Although not covered by pep8, since they came later, Enums tend to be overwhelmingly all caps in Python (https://docs.python.org/3/library/enum.html#creating-an-enum, https://mail.python.org/pipermail/python-ideas/2016-September/042340.html), and they are consistently all caps in CCF in both Python and C++, so I would prefer if we could make this one the same.

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
...
+class TxBundleInfo(NamedTuple):
+    """Bundle for transaction information required for validation """
+
+    merkle_tree: MerkleTree
+    existing_root: bytes
+    node_cert: bytes
+    signature: bytes
+    node_activity: dict
+    signing_node: int
```
achamayou, line 183:
```suggestion
    signing_node: bytes
```

Node ids are going to be hex digests of their public key shortly, the fact that they are integers at the moment isn't going to last, they're best treated as opaque strings.

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
...
+    def __init__(self):
+        self.node_certificates = {}
+        self.node_activity_status = {}
+        self.signature_count = 0
+        self.signature_ending = False
+        self.chosen_hash = ec.ECDSA(utils.Prehashed(hashes.SHA256()))
+
+        # Initialize the Merkle tree with an empty transaction.
+        self.merkle = MerkleTree()
+        empty_bytes_array = bytearray(32)
```
achamayou, line 225:
```suggestion
        empty_bytes_array = bytearray(SHA_256_HASH_SIZE )
```

python/ccf/ledger.py
```diff
@@ -149,6 +164,200 @@ def _byte_read_safe(file, num_of_bytes):
...
+        if not self.signature_ending:
+            LOG.error(
+                "The last transaction in the ledger chunk wasn't a signature transaction which is unexpected."
+            )
+            raise SignatureEndingException
+
+    def _verify_tx_set(self, tx_info: TxBundleInfo) -> bool:
+        """Verify items 1, 2, and 3 for a the transactions up until a signature"""
+        # 1) The merkle root is signed by a TRUSTED node in the given network
+        if self._verify_node_status(tx_info):
```
achamayou, line 311: Rather than returning a boolean, throwing a specific exception or at least logging which condition failed would probably be a good thing.

achamayou: Commented too soon, but this still looks odd, because it suggests the calls can return False, which they happen never to do. If all errors are represented by exceptions (which I think is right here), then it seems to me the return status should be None.

python/ccf/ledger.py
achamayou, line 524:
```suggestion
        self._verify_committed_files(sorted_ledgers)
```

python/ccf/ledger.py
```diff
@@ -297,19 +520,81 @@ def __init__(self, directory: str):
...
+                self._filenames[self._fileindex], self._ledger_validator
+            )
             return self._current_chunk
         else:
             raise StopIteration
 
     def __iter__(self):
         return self
+
+    def _verify_commited_files(self, sorted_ledgers: list):
```
achamayou, line 548:
```suggestion
    def _verify_committed_files(self, sorted_ledgers: list):
```

python/ccf/ledger.py
```diff
@@ -297,19 +520,81 @@ def __init__(self, directory: str):
...
+        expected_commit = 1
+
+        # initialize end commit id to 1
+        end_commit_id = -1
+
+        for filename in sorted_ledgers:
+            match = re.search(self.LEDGER_COMMITTED_FILE_NAME_REGEX, filename)
+            if match:
+                start_commit_id = match.group(1)
+                if (len(match.group(2))) > 0:
```
achamayou, line 564: Python supports named groups (https://docs.python.org/3/library/re.html#regular-expression-syntax, with `(?P<name>...)` and `m.group('name')`. It's mildly more readable, and less prone to breaks when the regexp changes.

python/ccf/merkletree.py
syhamza: This implementation was inspired by the above and few other merkle tree implementations. The way CCF does hashing was a bit different from the hashing techniques employed by the other solutions. Also, the added MerkleTree solution has more customized logging and has been simplified given our use-case.

python/ccf/ledger.py
syhamza: makes sense

python/ccf/ledger.py
syhamza: Sounds good. Changed the signature of this and all the methods being called to return void and throw respective exceptions when verification failed.

python/ccf/ledger.py
syhamza: Oh, ok. Will take a look

achamayou (approved):

python/ccf/ledger.py
syhamza: So Julien, what we can do is to only consider files with .committed suffix in the ledger directory. As of now, we are reading all the files and report failure if ledger chunk doesn't have .committed extension. What are your thoughts on it. I am assuming if we don't consider the non-committed chunk, the verification should still work fine and this test should succeed as well.

python/ccf/ledger.py
```diff
@@ -303,13 +503,50 @@ def __init__(self, directory: str):
...
+class SignatureCountException(Exception):
+    """When a ledger has no signature transactions"""
+
+
+class SignatureEndingException(Exception):
+    """When a signature transaction is not the last transaction of a ledger"""
+
+
+class InvalidRootSignatureException(Exception):
+    """When a signature transactions certificate cannot be validated against the node signature"""
```
eddyashton, line 544: Can we change the tense of these to match the `Exception`s below (tell the user what went wrong, rather than describing when they would be thrown?

```suggestion
class InvalidRootException(Exception):
    """Merkle root constructed from transaction digests doesn't match the root in signature transaction"""


class SignatureCountException(Exception):
    """The ledger contains no signature transactions"""


class SignatureEndingException(Exception):
    """The last transaction in the ledger is not a signature transaction"""


class InvalidRootSignatureException(Exception):
    """Signature transaction contains invalid signature"""
```

eddyashton (approved):

python/ccf/ledger.py
```diff
@@ -303,13 +503,50 @@ def __init__(self, directory: str):
                     LOG.warning(f"The file {chunk} has not been committed")
                 self._filenames.append(os.path.join(directory, chunk))
 
+        # Initialize LedgerValidator isntance which will be passed to LedgerChunks.
```
jumaffre, line 506:
```suggestion
        # Initialize LedgerValidator instance which will be passed to LedgerChunks.
```

jumaffre (approved):

tests/governance_history.py
```diff
@@ -25,17 +25,23 @@ def count_governance_operations(ledger):
...
+                    members[member_id_unpacked] = cert_unpacked
 
             if "public:ccf.gov.history" in tables:
                 governance_history_table = tables["public:ccf.gov.history"]
                 for member_id, signed_request in governance_history_table.items():
+                    member_id_unpacked = ccf.ledger.extract_msgpacked_data(member_id)
+                    signed_request_unpacked = ccf.ledger.extract_msgpacked_data(
+                        signed_request
+                    )
                     assert member_id in members
```
jumaffre, line 39:
```suggestion
                    assert member_id_unpacked in members
```

jumaffre:
@syhamza The `governance_history_test` should now be fixed (see https://github.com/microsoft/CCF/pull/2193/commits/df34a86bea89f4f5f5bc8e8d32056bc2edfd587a). I've also updated the log messages and level slightly so that it is not too verbose by default.

PR #2194, opened 2021-02-13
fix race in NodeToNode

ashamis:
There is a race that exists on a backup where the primary can create a connection from 2 different source, a connection from a client and a connection to another node (if CFT then only the primary). If these 2 things happened at the same time then an unordered_map in the ChannelManager is concurrently modified.

### DIFF
--- a/src/node/node_to_node.h
+++ b/src/node/node_to_node.h
@@ -130,16 +130,20 @@ namespace ccf
     NodeId self;
     std::unique_ptr<ChannelManager> channels;
     ringbuffer::AbstractWriterFactory& writer_factory;
-    SpinLock lock;
 
   public:
     NodeToNodeImpl(ringbuffer::AbstractWriterFactory& writer_factory_) :
+      self(INVALID_ID),
       writer_factory(writer_factory_)
     {}
 
     void initialize(NodeId self_id, const tls::Pem& network_pkey) override
     {
-      std::unique_lock<SpinLock> guard(lock);
+      CCF_ASSERT_FMT(
+        self_id == INVALID_ID,
+        "Calling initialize more than once, previous id:{}, new id:{}",
+        self,
+        self_id);
       self = self_id;
       channels =
         std::make_unique<ChannelManager>(writer_factory, network_pkey, self);
@@ -155,7 +159,6 @@ namespace ccf
         return;
       }
 
-      std::unique_lock<SpinLock> guard(lock);
       channels->create_channel(peer_id, hostname, service);
     }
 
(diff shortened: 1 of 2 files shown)


src/node/node_to_node.h
```diff
@@ -138,6 +139,7 @@ namespace ccf
 
     void initialize(NodeId self_id, const tls::Pem& network_pkey) override
     {
+      std::unique_lock<SpinLock> guard(lock);
```
achamayou, line 142: Why not `std::lock_guard`? It doesn't look like we need to unlock it manually?

achamayou (approved):
Thanks for fixing this!

jumaffre:
Good spot and thanks for fixing this! After a brief look at the existing code, it seems to be that the lock should be in the `ChannelManager` class directly (to protect all accesses to the `channels` member). 

If a channel can be destroyed concurrently to its access by another thread, then I also suggest that `ChannelManager::get()` returns a `std::shared_ptr<Channel>` instead of a `Channel&`. If the message ends up being sent to the other node on a closed channel, the host will discard it anyway as the node-to-node connection on the host will have been closed.

src/node/node_to_node.h
```diff
@@ -182,15 +188,23 @@ namespace ccf
...
       NodeId from_node, CBuffer cb, const uint8_t*& data, size_t& size) override
     {
-      auto& n2n_channel = channels->get(from_node);
-      return n2n_channel.recv_authenticated(cb, data, size);
+      ccf::Channel* n2n_channel;
+      {
+        std::unique_lock<SpinLock> guard(lock);
+        n2n_channel = &channels->get(from_node);
+      }
+      return n2n_channel->recv_authenticated(cb, data, size);
```
eddyashton, line 207: This pattern is unsafe. If we call `destroy_channel` or `destroy_all_channels` from another thread, after `get` but before `n2n_channel->`, then we are dereferencing a deleted object here.

src/node/node_to_node.h
eddyashton, line 142: This guards assignment of `self`, but we don't have the same guard on later reads of it. So if it was possible to call `initialize` multiple times, then the value could change during execution of (eg) `create_channel`. I think we should convince ourselves that `initialize` is only called once in a node's lifetime, while the enclave is single-threaded, before any other calls to this object, and then we can remove this lock.

eddyashton (changes requested):
Seconding @jumaffre's comment - good catch, but I think this locking should be done within `ChannelManager`, and `ChannelManager::get` should return `shared_ptr`s.

jumaffre (approved):

eddyashton (approved):

PR #2198, opened 2021-02-16
Parallel execution on the backup

ashamis:
Part of #2055

This change allows execution of transactions in parallel on the back-ups when running with BFT.
The CFT case does not change as it does not make sense to apply append entries across multiple threads.

This change does the following:
- We start tracking the last max read version for every key-value pair so that we can correctly build a dependency order. We use this watermark as the barrier which must be passed for a tx to be executed on the backup
- In raft we chunk up the append entries and if possible execute them in parallel
- We also are verifying signatures sent by backups in parallel

### DIFF
--- a/src/consensus/aft/async_execution.h
+++ b/src/consensus/aft/async_execution.h
@@ -13,8 +13,7 @@ namespace aft
       AppendEntries r, const uint8_t* data, size_t size) = 0;
     virtual void recv_append_entries_response(AppendEntriesResponse r) = 0;
     virtual bool recv_append_entries_signed_response(
-      SignedAppendEntriesResponse r,
-      bool is_pre_exec) = 0;
+      SignedAppendEntriesResponse r, bool is_pre_exec) = 0;
     virtual void recv_request_vote(RequestVote r) = 0;
     virtual void recv_request_vote_response(RequestVoteResponse r) = 0;
     virtual void recv_signature_received_ack(SignaturesReceivedAck r) = 0;
--- a/src/consensus/aft/async_execution.h
+++ b/src/consensus/aft/async_execution.h
@@ -12,8 +12,9 @@ namespace aft
     virtual void recv_append_entries(
       AppendEntries r, const uint8_t* data, size_t size) = 0;
     virtual void recv_append_entries_response(AppendEntriesResponse r) = 0;
-    virtual void recv_append_entries_signed_response(
-      SignedAppendEntriesResponse r) = 0;
+    virtual bool recv_append_entries_signed_response(
+      SignedAppendEntriesResponse r,
+      bool is_pre_exec) = 0;
     virtual void recv_request_vote(RequestVote r) = 0;
     virtual void recv_request_vote_response(RequestVoteResponse r) = 0;
     virtual void recv_signature_received_ack(SignaturesReceivedAck r) = 0;
@@ -29,6 +30,7 @@ namespace aft
   public:
     virtual ~AbstractMsgCallback() = default;
     virtual void execute() = 0;
+    virtual void async_execute() {}
   };
 
   class AppendEntryCallback : public AbstractMsgCallback
@@ -88,14 +90,23 @@ namespace aft
       hdr(std::move(hdr_))
     {}
 
+    void async_execute() override
+    {
+      async_exec_result = store.recv_append_entries_signed_response(hdr, true);
+    }
+
     void execute() override
     {
-      store.recv_append_entries_signed_response(hdr);
+      if (async_exec_result)
+      {
+        store.recv_append_entries_signed_response(hdr, false);
+      }
     }
 
   private:
     AbstractConsensusCallback& store;
     SignedAppendEntriesResponse hdr;
+    bool async_exec_result = false;
   };
 
   class RequestVoteCallback : public AbstractMsgCallback
--- a/src/consensus/aft/async_execution_coordinator.h
+++ b/src/consensus/aft/async_execution_coordinator.h
@@ -1,0 +1,88 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ds/thread_messaging.h"
+
+namespace aft
+{
+  enum class AsyncExecutionResult
+  {
+    PENDING,
+    COMPLETE
+  };
+
+  class AsyncExecutionCoordinator
+  {
+  public:
+    AsyncExecutionCoordinator(uint16_t thread_count) :
+      pending_cbs(0), run_sync(thread_count == 1)
+    {}
+
+    void increment_pending()
+    {
+      ++pending_cbs;
+    }
+
+    AsyncExecutionResult decrement_pending()
+    {
+      --pending_cbs;
+      return execution_status();
+    }
+
+    AsyncExecutionResult execution_status()
+    {
+      if (pending_cbs == 0) 
+      {
+        return AsyncExecutionResult::COMPLETE;
+      }
+      return AsyncExecutionResult::PENDING;
+    }
+
+    void start_next_execution_round(kv::Version last_idx)
+    {
+      is_first = true;
+      must_break = false;
+      execution_start_idx = last_idx;
+    }
+    
+    bool should_exec_next_append_entry(bool support_async_execution, uint64_t max_conflict_version)
+    {
+      if (!run_sync)
+      {
+        if (must_break)
+        {
+          return false;
+        }
+
+        if (execution_status() == AsyncExecutionResult::PENDING)
+        {
+          if (!support_async_execution && !is_first)
+          {
+            return false;
+          }
+
+          if (
+            support_async_execution &&
+            max_conflict_version >= execution_start_idx)
+          {
+            return false;
+          }
+        }
+      }
+
+      is_first = false;
+      must_break = !support_async_execution &&
+        (execution_status() == AsyncExecutionResult::PENDING);
+
+      return true;
+    }
+
+  private:
+    uint32_t pending_cbs = 0;
+    bool run_sync = threading::ThreadMessaging::thread_count == 1;
+    bool is_first;
+    bool must_break;
+    uint64_t execution_start_idx;
+  };
+}
--- a/src/consensus/aft/async_execution_coordinator.h
+++ b/src/consensus/aft/async_execution_coordinator.h
@@ -12,11 +12,19 @@ namespace aft
     COMPLETE
   };
 
-  class AsyncExecutionCoordinator
+  class AsyncExecutor
   {
+  private:
+    uint32_t pending_cbs = 0;
+    bool run_sync = threading::ThreadMessaging::thread_count == 1;
+    bool is_first;
+    bool must_break;
+    uint64_t execution_start_idx;
+
   public:
-    AsyncExecutionCoordinator(uint16_t thread_count) :
-      pending_cbs(0), run_sync(thread_count == 1)
+    AsyncExecutor(uint16_t thread_count) :
+      pending_cbs(0),
+      run_sync(thread_count == 1)
     {}
 
     void increment_pending()
@@ -32,21 +40,22 @@ namespace aft
 
     AsyncExecutionResult execution_status()
     {
-      if (pending_cbs == 0) 
+      if (pending_cbs == 0)
       {
         return AsyncExecutionResult::COMPLETE;
       }
       return AsyncExecutionResult::PENDING;
     }
 
-    void start_next_execution_round(kv::Version last_idx)
+    void execute_as_far_as_possible(kv::Version start_idx)
     {
       is_first = true;
       must_break = false;
-      execution_start_idx = last_idx;
+      execution_start_idx = start_idx;
     }
-    
-    bool should_exec_next_append_entry(bool support_async_execution, uint64_t max_conflict_version)
+
+    bool should_exec_next_append_entry(
+      bool support_async_execution, uint64_t max_conflict_version)
     {
       if (!run_sync)
       {
@@ -77,12 +86,5 @@ namespace aft
 
       return true;
     }
-
-  private:
-    uint32_t pending_cbs = 0;
-    bool run_sync = threading::ThreadMessaging::thread_count == 1;
-    bool is_first;
-    bool must_break;
-    uint64_t execution_start_idx;
   };
 }
--- a/src/consensus/aft/impl/execution.h
+++ b/src/consensus/aft/impl/execution.h
@@ -37,16 +37,20 @@ namespace aft
     virtual kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      std::shared_ptr<aft::RequestTracker> request_tracker = nullptr) = 0;
+      kv::Consensus::SeqNo last_idx,
+      std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
+      kv::Consensus::SeqNo max_conflict_version = -1) = 0;
 
     virtual std::unique_ptr<aft::RequestMessage> create_request_message(
       const kv::TxHistory::RequestCallbackArgs& args,
       kv::Consensus::SeqNo committed_seqno) = 0;
 
     virtual kv::Version commit_replayed_request(
-      kv::Tx& tx,
+      aft::Request& request,
       std::shared_ptr<aft::RequestTracker> request_tracker,
-      kv::Consensus::SeqNo committed_seqno) = 0;
+      kv::Consensus::SeqNo last_idx,
+      kv::Consensus::SeqNo committed_seqno,
+      kv::Consensus::SeqNo max_conflict_version) = 0;
   };
 
   class ExecutorImpl : public Executor
@@ -69,16 +73,20 @@ namespace aft
     kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      std::shared_ptr<aft::RequestTracker> request_tracker = nullptr) override;
+      kv::Consensus::SeqNo last_idx,
+      std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
+      kv::Consensus::SeqNo max_conflict_version = -1) override;
 
     std::unique_ptr<aft::RequestMessage> create_request_message(
       const kv::TxHistory::RequestCallbackArgs& args,
       kv::Consensus::SeqNo committed_seqno) override;
 
     kv::Version commit_replayed_request(
-      kv::Tx& tx,
+      aft::Request& request,
       std::shared_ptr<aft::RequestTracker> request_tracker,
-      kv::Consensus::SeqNo committed_seqno) override;
+      kv::Consensus::SeqNo last_idx,
+      kv::Consensus::SeqNo committed_seqno,
+      kv::Consensus::SeqNo max_conflict_version) override;
 
   private:
     std::shared_ptr<State> state;
--- a/src/consensus/aft/impl/execution.h
+++ b/src/consensus/aft/impl/execution.h
@@ -37,9 +37,9 @@ namespace aft
     virtual kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      kv::Consensus::SeqNo commit_version = -1,
+      kv::Consensus::SeqNo commit_version = kv::NoVersion,
       std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
-      kv::Consensus::SeqNo max_conflict_version = -1) = 0;
+      kv::Consensus::SeqNo max_conflict_version = kv::NoVersion) = 0;
 
     virtual std::unique_ptr<aft::RequestMessage> create_request_message(
       const kv::TxHistory::RequestCallbackArgs& args,
@@ -73,9 +73,9 @@ namespace aft
     kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      kv::Consensus::SeqNo last_idx = -1,
+      kv::Consensus::SeqNo last_idx = kv::NoVersion,
       std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
-      kv::Consensus::SeqNo max_conflict_version = -1) override;
+      kv::Consensus::SeqNo max_conflict_version = kv::NoVersion) override;
 
     std::unique_ptr<aft::RequestMessage> create_request_message(
       const kv::TxHistory::RequestCallbackArgs& args,
--- a/src/consensus/aft/impl/execution.h
+++ b/src/consensus/aft/impl/execution.h
@@ -37,7 +37,7 @@ namespace aft
     virtual kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      kv::Consensus::SeqNo last_idx,
+      kv::Consensus::SeqNo last_idx = -1,
       std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
       kv::Consensus::SeqNo max_conflict_version = -1) = 0;
 
@@ -73,7 +73,7 @@ namespace aft
     kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      kv::Consensus::SeqNo last_idx,
+      kv::Consensus::SeqNo last_idx = -1,
       std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
       kv::Consensus::SeqNo max_conflict_version = -1) override;
 
--- a/src/consensus/aft/impl/execution.h
+++ b/src/consensus/aft/impl/execution.h
@@ -37,7 +37,7 @@ namespace aft
     virtual kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      kv::Consensus::SeqNo last_idx = -1,
+      kv::Consensus::SeqNo commit_version = -1,
       std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
       kv::Consensus::SeqNo max_conflict_version = -1) = 0;
 
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -680,41 +680,41 @@ namespace aft
         return;
       }
 
-        if (execute_async)
-        {
-          auto async_pending_msg =
-            std::make_unique<threading::Tmsg<AsyncPendingExec>>(
-              [](std::unique_ptr<threading::Tmsg<AsyncPendingExec>> msg) {
-                msg->data.pending_execution->async_execute();
+      if (execute_async)
+      {
+        auto async_pending_msg =
+          std::make_unique<threading::Tmsg<AsyncPendingExec>>(
+            [](std::unique_ptr<threading::Tmsg<AsyncPendingExec>> msg) {
+              msg->data.pending_execution->async_execute();
 
-                msg->reset_cb(add_to_pending_execution_cb);
-                threading::ThreadMessaging::thread_messaging.add_task(
-                  threading::MAIN_THREAD_ID, std::move(msg));
-              },
-              this,
-              std::move(aee));
+              msg->reset_cb(add_to_pending_execution_cb);
+              threading::ThreadMessaging::thread_messaging.add_task(
+                threading::MAIN_THREAD_ID, std::move(msg));
+            },
+            this,
+            std::move(aee));
 
-          if (threading::ThreadMessaging::thread_count > 1)
-          {
-            threading::ThreadMessaging::thread_messaging.add_task(
-              threading::ThreadMessaging::get_execution_thread(
-                ++next_exec_thread),
-              std::move(async_pending_msg));
-          }
-          else
-          {
-            async_pending_msg->data.pending_execution->async_execute();
-            add_to_pending_execution_cb(std::move(async_pending_msg));
-          }
-        }
-        else if (!is_execution_pending)
+        if (threading::ThreadMessaging::thread_count > 1)
         {
-          aee->execute();
+          threading::ThreadMessaging::thread_messaging.add_task(
+            threading::ThreadMessaging::get_execution_thread(
+              ++next_exec_thread),
+            std::move(async_pending_msg));
         }
         else
         {
-          pending_executions.push_back(std::move(aee));
+          async_pending_msg->data.pending_execution->async_execute();
+          add_to_pending_execution_cb(std::move(async_pending_msg));
         }
+      }
+      else if (!is_execution_pending)
+      {
+        aee->execute();
+      }
+      else
+      {
+        pending_executions.push_back(std::move(aee));
+      }
 
       try_execute_pending();
     }
@@ -1403,9 +1403,7 @@ namespace aft
 
     struct AsyncExecutionCtx
     {
-      AsyncExecutionCtx() :
-        pending_cbs(0)
-      {}
+      AsyncExecutionCtx() : pending_cbs(0) {}
 
       uint32_t pending_cbs;
     };
@@ -1447,7 +1445,7 @@ namespace aft
       AppendEntries& r = msg->data.r;
       bool confirm_evidence = msg->data.confirm_evidence;
 
-      for(auto& ae : append_entries)
+      for (auto& ae : append_entries)
       {
         auto& [ds, i] = ae;
         state->last_idx = i;
@@ -1587,7 +1585,7 @@ namespace aft
         must_break =
           !ds->support_async_execution() && (async_exec.pending_cbs > 0);
 
-       ++async_exec_msg->data.index;
+        ++async_exec_msg->data.index;
         state->last_idx = i;
 
         kv::ApplyResult apply_success = ds->apply();
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1332,8 +1332,7 @@ namespace aft
       }
       else
       {
-        guard.unlock();
-        msg->cb(std::move(msg));
+        execute_append_entries(std::move(msg));
       }
     }
 
@@ -1351,10 +1350,11 @@ namespace aft
       std::unique_ptr<threading::Tmsg<AsyncExecution>> msg)
     {
       auto self = msg->data.self;
-      self->execute_append_entries_cb_aaa(std::move(msg));
+      std::unique_lock<SpinLock> guard(self->state->lock);
+      self->execute_append_entries(std::move(msg));
     }
 
-    void execute_append_entries_cb_aaa(
+    void execute_append_entries(
       std::unique_ptr<threading::Tmsg<AsyncExecution>> msg)
     {
       auto self = msg->data.self;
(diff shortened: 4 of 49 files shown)


src/consensus/aft/raft.h
```diff
@@ -633,7 +668,34 @@ namespace aft
           }
         }
 
-        if (!is_execution_pending)
+        if (always_execute_async)
```
ashamis, line 671: execute_async

src/consensus/aft/raft.h
```diff
@@ -1259,47 +1334,284 @@ namespace aft
...
+      self->execute_append_entries(std::move(msg));
+    }
+
+    void execute_append_entries(
+      std::unique_ptr<threading::Tmsg<AsyncExecution>> msg)
+    {
+      auto self = msg->data.self;
+      if (self->consensus_type == ConsensusType::CFT)
+      {
+        self->execute_append_entries_cft(msg);
```
ashamis, line 1365: execute_append_entries_sync vs async

src/consensus/aft/raft.h
```diff
@@ -1259,47 +1334,284 @@ namespace aft
...
+            {
+              run_async_execution(pending_requests);
+              return false;
+            }
+          }
+        }
+
+        is_first = false;
+        must_break =
+          !wrapper->support_asyc_execution() && (async_exec->pending_cbs > 0);
```
ashamis, line 1611: async

src/consensus/aft/raft.h
```diff
@@ -1259,47 +1334,284 @@ namespace aft
...
+
+        auto ae = std::move(append_entries.front());
+        append_entries.pop_front();
         auto& [ds, i] = ae;
         state->last_idx = i;
```
ashamis, line 1617: not execute ~> resolve

src/kv/tx.h
```diff
@@ -251,7 +268,9 @@ namespace kv
      *
      * @return transaction outcome
      */
-    CommitResult commit()
+    CommitResult commit(
+      std::function<Version()> f = nullptr,
```
ashamis, line 272: f ~~> version_resolver

src/consensus/aft/async_execution.h
```diff
@@ -12,8 +12,8 @@ namespace aft
     virtual void recv_append_entries(
       AppendEntries r, const uint8_t* data, size_t size) = 0;
     virtual void recv_append_entries_response(AppendEntriesResponse r) = 0;
-    virtual void recv_append_entries_signed_response(
-      SignedAppendEntriesResponse r) = 0;
+    virtual bool recv_append_entries_signed_response(
+      SignedAppendEntriesResponse r, bool is_pre_exec) = 0;
```
achamayou, line 16: I don't think I understand what the flag does, recv_append_entries_signed_response() is the consensus behaviour on receiving a signed response (ack/nack) to an append entries, but what does pre-execution mean in this context? Hasn't the request always been executed on the primary by that point?

src/consensus/aft/impl/execution.h
```diff
@@ -37,16 +37,20 @@ namespace aft
     virtual kv::Version execute_request(
       std::unique_ptr<RequestMessage> request,
       bool is_create_request,
-      std::shared_ptr<aft::RequestTracker> request_tracker = nullptr) = 0;
+      kv::Consensus::SeqNo last_idx = -1,
```
achamayou, line 40: `max_conflict_version` is the version the store must be at least at, for the request to execute correctly, as I understand it. It's not clear to me what `last_idx` is at this point?

src/consensus/aft/raft.h
```diff
@@ -408,6 +419,13 @@ namespace aft
 
     void add_configuration(Index idx, const Configuration::Nodes& conf)
     {
+      std::unique_lock<SpinLock> guard(state->lock, std::defer_lock);
+      if (
+        consensus_type == ConsensusType::BFT && is_follower() &&
```
achamayou, line 424: How safe is it to call is_follower() without the state lock at this point?

src/consensus/aft/raft.h
```diff
@@ -1013,6 +1095,7 @@ namespace aft
         append_entries;
       AppendEntries r;
       bool confirm_evidence;
+      uint32_t index = 0;
```
achamayou, line 1098: What is this index? why is it 32 bits?

src/consensus/aft/raft.h
```diff
@@ -1259,51 +1342,253 @@ namespace aft
...
-        msg->data.append_entries, msg->data.r, msg->data.confirm_evidence);
-      msg->cb =
-        reinterpret_cast<void (*)(std::unique_ptr<threading::ThreadMsg>)>(
-          continue_execution);
+      auto self = msg->data.self;
+      std::unique_lock<SpinLock> guard(self->state->lock);
+      self->execute_append_entries(std::move(msg));
+    }
+
+    void execute_append_entries(
```
achamayou, line 1367: Can we go with the more general apply()? There's no execution in CFT.

src/consensus/aft/raft.h
```diff
@@ -1259,51 +1342,253 @@ namespace aft
...
+        commit_idx(commit_idx_),
+        home_thread(home_thread_)
+      {}
+
+      Aft<LedgerProxy, ChannelProxy, SnapshotterProxy>* self;
+      std::unique_ptr<kv::AbstractExecutionWrapper> ds;
+      std::unique_ptr<threading::Tmsg<AsyncExecution>> msg;
+      kv::Version last_idx;
+      kv::Version commit_idx;
+      uint16_t home_thread;
```
achamayou, line 1433: dispatch_thread? scheduler_thread? If we're going to use home_thread, it needs to be defined somewhere, it's not self-descriptive?

src/consensus/aft/raft.h
```diff
@@ -1259,51 +1342,253 @@ namespace aft
...
+      std::shared_ptr<AsyncExecutionCtx> ctx;
+    };
+
+    uint64_t next_exec_thread = 0;
+
+    bool execute_append_entries_sync(
```
achamayou, line 1439: That's a 200+ lines function, it's quite unfortunate, is there a way we can break this down?

src/consensus/aft/raft.h
```diff
@@ -1259,51 +1342,253 @@ namespace aft
...
+          {
+            throw std::logic_error("Unknown ApplyResult value");
+          }
+        }
+      }
+
+      return execute_append_entries_finish(confirm_evidence, r);
+    }
+
+    bool execute_append_entries_async(
```
achamayou, line 1540: That's a 200+ lines function, can we break it down?

src/consensus/aft/raft.h
```diff
@@ -1259,51 +1342,253 @@ namespace aft
...
+      std::unique_ptr<threading::Tmsg<AsyncExecution>>& msg)
+    {
+      // This function is responsible for selection the next batch of
+      // transactions we can execute concurrently and then starting said
+      // execution. If the are more pending entries after we select the batch we
```
achamayou, line 1545:
```suggestion
      // execution. If there are more pending entries after we select the batch we
```

src/consensus/aft/raft.h
```diff
@@ -1259,51 +1342,253 @@ namespace aft
...
+      bool must_break = false;
+      uint64_t before_state_idx = state->last_idx;
+      bool run_sync = threading::ThreadMessaging::thread_count == 1;
+
+      while (async_exec_msg->data.index != append_entries.size())
+      {
+        auto& [ds, i] = append_entries[async_exec_msg->data.index];
+        if (!run_sync)
+        {
+          if (must_break)
```
achamayou, line 1563: Can we document the return type of the call (it looks like false means we need to call again, which isn't necessarily what one might expect), and what must_break means (we've hit a serialisation point because of transaction dependencies, as far as I can tell?).

src/consensus/aft/raft.h
```diff
@@ -1532,30 +1877,50 @@ namespace aft
       try_send_sig_ack({r.term, r.last_log_idx}, result);
     }
 
-    void recv_append_entries_signed_response(SignedAppendEntriesResponse r)
+    bool recv_append_entries_signed_response(
```
achamayou, line 1880: What does the new boolean return value mean? Is this re-execution status or error?

src/consensus/aft/test/logging_stub.h
```diff
@@ -261,6 +261,22 @@ namespace aft
         throw std::logic_error("Not Implemented");
       }
 
+      bool support_async_execution() override
+      {
+        return true;
+        throw std::logic_error("Not Implemented");
```
achamayou, line 267: Surely this should return false?

src/kv/change_set.h
```diff
@@ -18,7 +18,7 @@ namespace kv
   using Snapshot = champ::Snapshot<K, VersionV<V>, H>;
 
   template <typename K>
-  using Read = std::map<K, Version>;
+  using Read = std::map<K, std::tuple<Version, Version>>;
```
achamayou, line 21: This really needs documenting. With one version, it's reasonably obvious, but not with two.

src/kv/deserialise.h
```diff
@@ -78,6 +80,11 @@ namespace kv
       throw std::logic_error("get_index not implemented");
     }
 
+    aft::Request& get_request() override
+    {
+      throw std::logic_error("get_index not implemented");
```
achamayou, line 85:
```suggestion
      throw std::logic_error("get_request not implemented");
```

Could async execution be implemented as a Mix-in? It might be easier to test too.

src/kv/deserialise.h
```diff
@@ -583,22 +626,50 @@ namespace kv
...
+             Term term,
+             OrderedChanges& changes,
+             aft::Request& req_) -> ApplyResult {
       tx->set_change_list(std::move(changes), term);
+
+      auto aft_requests = tx->rw<aft::RequestsMap>(ccf::Tables::AFT_REQUESTS);
+      auto req_v = aft_requests->get(0);
+      CCF_ASSERT(
+        req_v.has_value(),
+        "Deserialised request but it was not found in the requests map");
```
achamayou, line 656: Deserialised append entry, but requests map is empty?

src/kv/untyped_map.h
```diff
@@ -152,36 +156,99 @@ namespace kv::untyped
...
+      }
+
+      void commit(Version v, kv::Version& max_conflict_version) override
+      {
+        auto& roll = map.get_roll();
+        auto current = roll.commits->get_tail();
+        auto state = current->state;
+        bool is_writing_to_new_key = false;
+        auto consensus = map.store->get_consensus();
+        bool track_commit = consensus != nullptr &&
```
achamayou, line 180: I am very uncomfortable with making commit() consensus-aware and dependent. Can we return the necessary version information to the caller instead? If not, can this at least be an optional callback provided by the caller?

src/node/progress_tracker.h
```diff
@@ -710,13 +635,146 @@ namespace ccf
       return true;
     }
 
+    Nonce get_my_nonce(kv::TxID tx_id)
```
achamayou, line 638: get_node_nonce, get_local_node_nonce? I don't believe we use "my" anywhere else, and we certainly don't in the external APIs.

src/consensus/aft/raft.h
```diff
@@ -1431,6 +1764,15 @@ namespace aft
         }
       }
 
+      if (async_exec.pending_cbs == 0)
+      {
+        return execute_append_entries_finish(confirm_evidence, r);
```
achamayou, line 1769: I think it would be really good if we could separate the sync/async execution out of the consensus, where the implementation of the async flow does not belong, and have unit tests on the async flow to make sure it produces a valid order, which seem to be missing at the moment.

src/consensus/aft/raft.h
ashamis: Yes you can change the value of is_follower during a view change when there are no concurrent tx

src/consensus/aft/raft.h
```diff
@@ -632,24 +673,65 @@ namespace aft
...
+          threading::ThreadMessaging::thread_messaging.add_task(
+            threading::ThreadMessaging::get_execution_thread(
+              ++next_exec_thread),
+            std::move(async_pending_msg));
         }
         else
         {
-          pending_executions.push_back(std::move(aee));
+          async_pending_msg->data.pending_execution->async_execute();
+          add_to_pending_execution_cb(std::move(async_pending_msg));
```
eddyashton, line 707: I really can't parse this. Can you please add an explanatory comment?

We create a task which calls `async_execute`, and if we have multiple threads we send that task round-robin to another execution thread - I think I get that bit. But the task also calls `reset_cb`? And then sends itself to the main thread? And if we're single-threaded we execute inline and then add a message which is going to call `async_execute` again? If `execute_async` is true, I'd expect this to go through the messaging system (ie - be called later, asynchronously), even when single-threaded.

src/consensus/aft/test/main.cpp
```diff
@@ -21,7 +21,7 @@ using StoreSig = aft::LoggingStubStoreSig;
 using Adaptor = aft::Adaptor<Store>;
 
 threading::ThreadMessaging threading::ThreadMessaging::thread_messaging;
-std::atomic<uint16_t> threading::ThreadMessaging::thread_count = 0;
+std::atomic<uint16_t> threading::ThreadMessaging::thread_count = 1;
```
eddyashton, line 24: Tangential to this PR: Its confusing to have a possible value here (`0`) that most of the systems won't handle, and then have a load of `thread_count > 1` checks. I wonder if we can instead phrase this as `worker_thread_count`, and implicitly there's _always_ a main thread with 0-or-more worker threads.

eddyashton (commented):
Initial comments, still need to look at the KV changes and batch detection logic.

src/kv/apply_changes.h
```diff
@@ -113,6 +119,11 @@ namespace kv
       // Get the version number to be used for this commit.
       version = f();
 
+      if (!new_maps.empty() && version > 0)
```
eddyashton, line 122: Please add a comment explaining this. Is it "If this creates any maps, it must be fully serialised"?

src/kv/apply_changes.h
```diff
@@ -43,7 +44,8 @@ namespace kv
     std::function<Version()> f,
     kv::ConsensusHookPtrs& hooks,
     const MapCollection& new_maps = {},
-    const std::optional<Version>& new_maps_conflict_version = std::nullopt)
+    const std::optional<Version>& new_maps_conflict_version = std::nullopt,
+    bool track_conflicts = false)
```
eddyashton, line 48: Can you add a comment explaining the behaviour of `track_conflicts`?

src/kv/change_set.h
eddyashton: Seconded, ideally as a named `struct`.

src/kv/deserialise.h
```diff
@@ -583,22 +626,50 @@ namespace kv
...
 
     std::function<ApplyResult(
-      std::unique_ptr<Tx>& tx, Term term, OrderedChanges& changes)>
-      fn = [](std::unique_ptr<Tx>& tx, Term term, OrderedChanges& changes)
-      -> ApplyResult {
+      std::unique_ptr<Tx>& tx,
+      Term term,
+      OrderedChanges& changes,
+      aft::Request& req_)>
+      fn = [](
```
eddyashton, line 645: Is this a public `std::function` member called `fn`? Why is it implemented like this, can't this simply be inlined into `apply`? Or at least be a method rather than a member?

eddyashton: Reading again, I really don't follow this - it operates entirely on its own members. and the real result is that it populates `tx` and `req`. What am I missing?

src/kv/test/kv_test.cpp
```diff
@@ -1138,7 +1137,7 @@ TEST_CASE("Deserialise return status")
     REQUIRE(success == kv::CommitResult::SUCCESS);
 
     REQUIRE(
-      store.apply(data, ConsensusType::CFT)->execute() ==
+      store.apply(data, ConsensusType::CFT)->apply() ==
```
eddyashton, line 1140: On the one hand, I'm in favour of the function that returns an `ApplyResult` being called `apply()`. But `apply()->apply()` is very odd. Should the first be `parse()` or `deserialise()` or something?

src/kv/tx.h
```diff
@@ -2,9 +2,26 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+// The transaction engine supports producing a transaction dependencies. This is
+// materialized by providing a sequence number after which the current
+// transaction must be run. The current use case for dependency tracking is to
+// enable parallel execution of transactions on the backup, and as such we only
+// track dependencies when running BFT consensus protocol.
+//
+// Depency tracking follows the following pseudocode
```
eddyashton, line 11:
```suggestion
// Dependency tracking follows the following pseudocode
```

src/kv/tx.h
```diff
@@ -2,9 +2,26 @@
...
+//   foreach Accessed Key-Value pair:
+//     MaxSeenReadVersion = max(MaxSeenReadVersion, pair.last_read_version)
+//     pair.last_read_version = pair.seqno
+//
+//   TxSerialize(pairs, MaxSeenReadVersion)
+
 #include "apply_changes.h"
 #include "ds/ccf_assert.h"
 #include "ds/ccf_deprecated.h"
+#include "ds/thread_messaging.h"
```
eddyashton, line 24: I think this is unused and can be removed?

src/node/history.h
```diff
@@ -260,18 +261,21 @@ namespace ccf
       kv::Consensus::SignableTxIndices commit_txid_,
       kv::Store& store_,
       T& replicated_state_tree_,
+      SpinLock& state_lock_,
```
eddyashton, line 264: Whose lock is this?

src/node/progress_tracker.h
```diff
@@ -710,13 +635,146 @@ namespace ccf
...
   private:
     kv::NodeId id;
     std::shared_ptr<tls::Entropy> entropy;
     kv::Consensus::SeqNo highest_commit_level = 0;
     kv::TxID highest_prepared_level = {0, 0};
 
     std::map<kv::Consensus::SeqNo, CommitCert> certificates;
+    mutable SpinLock lock;
+
+    kv::TxHistory::Result add_signature_(
```
eddyashton, line 653: Please call this `add_signature_internal` or `add_signature_impl` or something.

src/node/request_tracker.h
```diff
@@ -152,19 +162,22 @@ namespace aft
...
-        requests_list_.get_tail() == nullptr ||
-          requests_list_.get_tail()->time <= time,
-        "items not entered in the correct order. last:{}, time:{}",
-        requests_list_.get_tail()->time,
-        time);
+      if (
+        requests_list_.get_tail() != nullptr &&
+        requests_list_.get_tail()->time > time)
+      {
+        // Time is an not a precise measurement and can be different be
```
eddyashton, line 177:
```suggestion
        // Time is not a precise measurement and can be different
```

I don't follow what this time is for. If we're sometimes overriding it, why doesn't it always come from `requests_list_.get_tail()->time`? Separately, if I'm reading correctly this could be a `static` function since it doesn't interact with any member state?

src/node/rpc/frontend.h
```diff
@@ -284,7 +286,18 @@ namespace ccf
             return ctx->serialise_response();
           }
 
-          switch (tx.commit())
+          std::function<kv::Version()> version_resolver = nullptr;
+          if (reserved != -1)
+          {
+            CCF_ASSERT(
+              consensus->type() == ConsensusType::BFT, "Wrong consensus type");
+            version_resolver = [&]() {
```
eddyashton, line 294: I'm a little worried about what this means for anything else trying to interact with the Store. If I'm reading correctly, the idea is that we call `next_version()` n times, but the `reserved` Version we _claim_ may happen out of order? But this seems like a complication to the `Store` API - if any other system is trying to read "where are you", it gets junk and needs to know that these are applied out-of-order. It would be simpler if we could execute our batch on the side (not touching the KV so other systems can safely call its old values), and then update it with the final state all at once? Obviously a large architectural change, but we need to think about what the semantics of any of these calls actually are.

src/node/rpc/frontend.h
```diff
@@ -592,7 +610,7 @@ namespace ccf
            ctx.frame_format()});
       };
 
-      auto rep = process_command(ctx, tx, fn);
+      auto rep = process_command(ctx, tx, fn, last_idx, max_conflict_version);
```
eddyashton, line 613: This is sometimes `last_idx` and sometimes `reserved` - these don't sound like the same property.

src/node/rpc/frontend.h
ashamis: I do not agree with this statement. The only guarantee that we provide our users is around globally committed state in regards to stability and serializability. In regards to correctness in a BFT system everything that reads from the KV needs to go through the consensus otherwise the result you have is junk anyway.

src/node/request_tracker.h
ashamis: Added more info to the comment and made the function static

src/consensus/aft/raft.h
```diff
@@ -632,24 +673,65 @@ namespace aft
...
       }
-      catch (const std::logic_error& err)
+      else if (!is_execution_pending)
       {
-        LOG_FAIL_EXC(err.what());
+        aee->execute();
+      }
+      else
+      {
+        pending_executions.push_back(std::move(aee));
```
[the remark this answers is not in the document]
ashamis: pending_executions is the backlog of pending execution but its possible to not have a backlog and have something currently executing, the later is tracked by is_execution_pending. I renamed pending_executions to execution_backlog to make this distinction clearer

src/node/history.h
ashamis: This is the state lock which protected replicated_state_tree_ this is a race condition that exists in multi-threading that exists before this PR.

src/consensus/aft/impl/execution.h
```diff
@@ -37,16 +37,20 @@ namespace aft
...
+      std::shared_ptr<aft::RequestTracker> request_tracker = nullptr,
+      kv::Consensus::SeqNo max_conflict_version = -1) = 0;
```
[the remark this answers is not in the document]
ashamis: nothing fundamental here, changing to kv::NoVersion

src/node/rpc/frontend.h
ashamis: fix the naming

src/kv/deserialise.h
ashamis: ok so i have been working with this code for too long :(, the artifact is fixed, fn no longer exists :)

So what you are doing in this function is separating 2 steps, step 1 is creating the tx object (from the append entry) and in an asynchronous step running it.

src/consensus/aft/raft.h
ashamis: Added a comment and broke this apart to make it easier to undestand

src/consensus/aft/raft.h
ashamis: i tried to doing this please have a look and tell me what you think

src/consensus/aft/async_execution.h
achamayou: Sorry, I still don't understand what pre_exec means here, and there's no comment either, could it be clarified for my benefit please?

src/consensus/aft/async_execution_coordinator.h
```diff
@@ -0,0 +1,88 @@
...
+
+namespace aft
+{
+  enum class AsyncExecutionResult
+  {
+    PENDING,
+    COMPLETE
+  };
+
+  class AsyncExecutionCoordinator
```
achamayou, line 15: Could this be AsyncExecutor? This is just conducting async execution, there is no coordination with any other component is there?

src/consensus/aft/async_execution_coordinator.h
```diff
@@ -0,0 +1,88 @@
...
+    AsyncExecutionResult execution_status()
+    {
+      if (pending_cbs == 0) 
+      {
+        return AsyncExecutionResult::COMPLETE;
+      }
+      return AsyncExecutionResult::PENDING;
+    }
+
+    void start_next_execution_round(kv::Version last_idx)
```
achamayou, line 42: What's a round? Is this just "execute as far as possible"?

src/consensus/aft/async_execution_coordinator.h
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      is_first = false;
+      must_break = !support_async_execution &&
+        (execution_status() == AsyncExecutionResult::PENDING);
+
+      return true;
+    }
+
+  private:
```
achamayou, line 81: We don't enforce this, but I think having the data members at the beginning of the class is better, and we do it in most places. Coming across them in the member functions and having to scroll down for them is not very intuitive.

src/consensus/aft/async_execution_coordinator.h
achamayou: last_idx -> start_idx?

src/kv/tx.h
```diff
@@ -2,6 +2,22 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+// The transaction engine supports producing a transaction dependencies. This is
+// materialized by providing a sequence number after which the current
+// transaction must be run. The current use case for dependency tracking is to
+// enable parallel execution of transactions on the backup, and as such we only
+// track dependencies when running BFT consensus protocol.
```
eddyashton, line 9: I must be missing something, but this sounds incomplete. This gives a lower bound ("this transaction must read at at-least this version, which is the version _I_ read it at"), but to ensure all executions read the same value we also need an upper bound - this transaction must not execute too late or your value may have changed by the time you execute this. How is that dependency tracked and enforced?

src/node/rpc/frontend.h
eddyashton: Sorry I don't mean to users, in the endpoint interface. I mean other bits of code within CCF. A lot of the instance I was thinking of are only in unit tests, so don't really matter.

But for instance, how does a snapshot work when `current_version()` doesn't actually report what's in the KV? We also ask for `current_version()` several times during recovery - how do we ensure that's not affected by this change? (There may be answers already for these! Just want to make sure they're considered and documented)

src/node/rpc/frontend.h
```diff
@@ -284,7 +286,18 @@ namespace ccf
...
+          {
+            CCF_ASSERT(
+              consensus->type() == ConsensusType::BFT, "Wrong consensus type");
+            version_resolver = [&]() {
+              tables.next_version();
+              return prescribed_commit_version;
+            };
+          }
+
+          switch (tx.commit(version_resolver, max_conflict_version))
```
eddyashton, line 300: Can you rewrite this as:

```
kv::CommitResult result;
if (prescribed_commit_version != kv::NoVersion)
{
  version_resolver = ...;
  result = tx.commit(version_resolver, max_conflict_version);
}
else
{
  result = tx.commit();
}

switch (result)
```

I think it helps readability to make it clear there's really a simpler `tx.commit()` path here, without working backwards to compare the default values and confirm that's what's being called.

src/node/rpc/test/frontend_test.cpp
```diff
@@ -522,7 +522,7 @@ TEST_CASE("process_bft")
     enclave::InvalidSessionId, user_caller_der);
   auto ctx = enclave::make_rpc_context(session, request.raw);
   ctx->execute_on_node = true;
-  frontend.process_bft(ctx);
+  frontend.process_bft(ctx, 1, 1);
```
eddyashton, line 525:
```suggestion
  constexpr auto prescribed_commit_version = 1;
  constexpr auto max_conflict_version = 1;
  frontend.process_bft(ctx, prescribed_commit_version, max_conflict_version);
```

And just to confirm - these are unrelated to the `{1, 1}` used to construct `rid` above?

src/enclave/rpc_handler.h
```diff
@@ -42,8 +42,13 @@ namespace enclave
...
     virtual ProcessBftResp process_bft(
-      std::shared_ptr<enclave::RpcContext> ctx) = 0;
+      std::shared_ptr<enclave::RpcContext> ctx,
+      kv::Consensus::SeqNo last_idx,
+      kv::Consensus::SeqNo max_conflict_version) = 0;
     virtual ProcessBftResp process_bft(
-      std::shared_ptr<enclave::RpcContext> ctx, kv::Tx& tx) = 0;
+      std::shared_ptr<enclave::RpcContext> ctx,
+      kv::Tx& tx,
+      kv::Consensus::SeqNo last_idx = -1,
```
eddyashton, line 51: Please use the same names we use in the implementation (`prescribed_commit_version`), and `kv::NoVersion` as default value.

src/kv/tx.h
ashamis: Ill add it to the comment but  transaction execution is started in the total order

src/node/rpc/test/frontend_test.cpp
ashamis: yes the two are not related

src/node/rpc/frontend.h
ashamis: This is done the same way that we deal with view-changes and signature transactions on the backup in this PR. We ensure that there in no pending execution when we trigger these cases.

src/consensus/aft/test/main.cpp
```diff
@@ -1096,4 +1096,47 @@ DOCTEST_TEST_CASE("Exceed append entries limit")
...
+    }
+  }
+
+  DOCTEST_INFO("multithreaded run upto sync point");
+  {
+    aft::AsyncExecutor aec(2);
+    aec.execute_as_far_as_possible(5);
+    for (uint32_t i = 0; i < 4; ++i)
+    {
+      DOCTEST_REQUIRE(aec.should_exec_next_append_entry(true, i));
```
ashamis, line 1121: when we have the first entry and we have false

ashamis: move bool into ctor

src/consensus/aft/raft.h
```diff
@@ -1380,57 +1519,254 @@ namespace aft
...
+          throw std::logic_error("Unknown ApplyResult value");
+        }
+      }
+      return AsyncExecutionResult::COMPLETE;
+    }
+
+    AsyncExecutionResult execute_append_entries_async(
+      std::unique_ptr<threading::Tmsg<AsyncExecution>>& msg)
+    {
+      // This function is responsible for selection the next batch of
```
achamayou, line 1726:
```suggestion
      // This function is responsible for selecting the next batch of
```

src/consensus/aft/raft.h
```diff
@@ -1380,57 +1519,254 @@ namespace aft
...
 
+      if (async_executor.execution_status() == AsyncExecutionResult::COMPLETE)
+      {
+        execute_append_entries_finish(confirm_evidence, r);
+        return AsyncExecutionResult::COMPLETE;
+      }
+      return AsyncExecutionResult::PENDING;
+    }
+
+    bool execute_append_entries_finish(bool confirm_evidence, AppendEntries& r)
```
achamayou, line 1768: void

src/consensus/aft/async_execution.h
achamayou: To be split as two void calls as discussed.

src/consensus/aft/raft.h
```diff
@@ -632,41 +677,90 @@ namespace aft
...
-        }
       }
       catch (const std::logic_error& err)
       {
         LOG_FAIL_EXC(err.what());
+        return;
+      }
+
+      // There are two different choices on how to deal with the request.
+      // Either have to execute the request in guarantteed serialized order
```
achamayou, line 688:
```suggestion
      // Either have to execute the request in guaranteed serialized order
```

src/consensus/aft/raft.h
eddyashton, line 1768: Not clear here where `confirm_evidence` comes from, and reading the code it suggests we will `commit_if_possible` _without_ confirming, if this bool can be set to false. I think it could be removed, and we can always call `view_change_tracker->check_evidence(r.term)`? That would be a nop (I've already checked and I'm happy) for subsequent calls in the same term, but the first call (per-term) it does the real evidence check.

src/consensus/aft/raft.h
```diff
@@ -1380,57 +1519,254 @@ namespace aft
...
+
+    AsyncExecutionResult process_async_execution(
+      kv::ApplyResult apply_result,
+      std::unique_ptr<kv::AbstractExecutionWrapper>& ds,
+      kv::Version i,
+      AppendEntries& r)
+    {
+      if (apply_result == kv::ApplyResult::FAIL)
+      {
+        // Setting last_idx to i-1 is a work around that should be fixed
```
achamayou, line 1550: ref to ticket, s/we/backup

src/consensus/aft/test/logging_stub.h
```diff
@@ -261,6 +261,21 @@ namespace aft
...
+      }
+
+      uint64_t get_max_conflict_version() override
+      {
+        throw std::logic_error("get_index not implemented");
+      }
+
+      aft::Request& get_request() override
+      {
+        throw std::logic_error("get_index not implemented");
```
achamayou, line 276:
```suggestion
        throw std::logic_error("get_request not implemented");
```

src/consensus/aft/test/main.cpp
```diff
@@ -1096,4 +1096,47 @@ DOCTEST_TEST_CASE("Exceed append entries limit")
     (sent_entries > num_small_entries_sent &&
      sent_entries <= num_small_entries_sent + num_big_entries));
   DOCTEST_REQUIRE(r2.ledger->ledger.size() == individual_entries);
+}
+
+DOCTEST_TEST_CASE("Test Asynchronous Execution Coordinator")
```
ashamis, line 1101: - first tx with false as its sync
- use primary to create sequence
  - no shuffle
  - shuffle but happy exec 
- on primary to create order -  shuffle - see conflict occurred

src/kv/test/kv_test.cpp
```diff
@@ -1456,13 +1454,44 @@ TEST_CASE("Conflict resolution")
...
   REQUIRE(tx1.commit_version() > tx2.commit_version());
-  REQUIRE(tx2.get_read_version() >= tx2.get_read_version());
+  REQUIRE(tx1.get_read_version() >= tx2.get_read_version());
 
   // Re-running a _committed_ transaction is exceptionally bad
   REQUIRE_THROWS(tx1.commit());
   REQUIRE_THROWS(tx2.commit());
 }
 
+TEST_CASE("Backup conflict detection")
```
ashamis, line 1464: - first tx that must not be parallel with other (next tx must wait)
- primary runs a bunch of tx with no dependency - apply order to backup have no dependencies
- - primary runs a bunch of tx with dependencies - apply in correct & incorrect order

src/kv/test/kv_test.cpp
```diff
@@ -1456,13 +1454,44 @@ TEST_CASE("Conflict resolution")
   REQUIRE(res3 == kv::CommitResult::SUCCESS);
 
   REQUIRE(tx1.commit_version() > tx2.commit_version());
-  REQUIRE(tx2.get_read_version() >= tx2.get_read_version());
+  REQUIRE(tx1.get_read_version() >= tx2.get_read_version());
```
ashamis, line 1457: tx2 > tx3

ashamis:
Closing as this was committed as part of another PR

Issue #2199, opened 2021-02-16
Not all ledger secrets should be kept in enclave memory


jumaffre:
Follow up from #1648. Now that we have a mechanism to recover historical ledger secrets from the ledger, we shouldn't keep in enclave memory all the historical ledger secrets that we've ever known about as we can conveniently retrieve them from the ledger when necessary.

achamayou:
@jumaffre that seems like something that could make use of the LFS mechanism @eddyashton built for indexing?

achamayou:
Accounting for all the various overhead in `LedgerSecret`, each entry comes out at about 320 bytes, and so 3000+ will fit in ~1Mb of memory. For a ten-year ledger, that would be a rekey every day.

There is no need to look at paging to disk for this.

PR #2200, opened 2021-02-16
Recover historical ledger secrets from the ledger

### LINKED ISSUE #1648: Support for historical queries after recovery from snapshot

jumaffre:
Resolves #1648 

After recovering from a snapshot, a node doesn't have the history of ledger secrets required to deserialise historical ledger entries. This PR adds support for historical ledger secrets recovery from the ledger.

This is done by hoping through the ledger backwards, at the exact `seqno` at which the encrypted previous ledger secret was written to the `public:ccf.internal.historical_encrypted_ledger_secret` table, decrypting the previous ledger with the last recovered one, etc. The historical state cache keeps track of all recovered ledger secrets in `network.historical_ledger_secrets` so that further requests won't need to recover those ledger secrets (*). 

To do so, each `LedgerSecret` now keeps track of the `seqno` at which the previous ledger secret was written to the store. This is fairly trivial for new ledger secrets on rekey but isn't for the first ledger secret after recovery. To solve this, I've had to introduce a new hook on the `public:ccf.internal.historical_encrypted_ledger_secret` table which only triggers once, on all nodes, after the private recovery is complete and the new ledger secret has been written to the ledger. 

Edit: Also updated the schema for the `public:ccf.internal.encrypted_ledger_secrets` table which is now keyed at `0`. This prevents infinite growth as more nodes are added to the service and simplifies the `LedgerSecrets` class which no longer needs to know about the node id. 

(*) For now, the main and historical ledger secrets aren't compacted/garbage collected (follow-up work: https://github.com/microsoft/CCF/issues/2199)


src/kv/deserialise.h
```diff
@@ -139,6 +139,12 @@ namespace kv
         success = ApplyResult::PASS_SNAPSHOT_EVIDENCE;
       }
 
+      search = changes.find(ccf::Tables::ENCRYPTED_PAST_LEDGER_SECRET);
+      if (search != changes.end())
+      {
+        success = ApplyResult::PASS_ENCRYPTED_PAST_LEDGER_SECRET;
```
eddyashton, line 145: Do we write to this table in transactions that touch multiple tables? This return scheme originally worked for `PASS_SIGNATURES`, which worked because signatures is a special table with a bunch of other requirements on it. I'm wary of repeating this pattern for too many tables, especially if we're not checking that there's only one table/value written.

Probably out of scope for this PR but: since we now split `deserialise` from application, and its returned value is really the raw `changes`, we could remove all of these enum values and just return `PASS`/`FAIL`, and leave the "did this produce some tables I care about" decision up to the caller.

src/node/historical_queries.h
```diff
@@ -18,36 +22,74 @@ namespace ccf::historical
...
     std::list<consensus::Index> recent_requests;
 
     // To trust an index, we currently need to fetch a sequence of entries
     // around it - these aren't user requests, so we don't store them, but we do
     // need to distinguish things-we-asked-for from junk-from-the-host
     std::set<consensus::Index> pending_fetches;
 
+    ccf::VersionedLedgerSecret get_first_known_ledger_secret()
+    {
+      auto tx = network.tables->create_read_only_tx();
```
eddyashton, line 79: I think we should avoid creating this tx here, and instead have `get_latest`/`get_first` variants that don't take a `Tx` and don't have a transactional dependency, iff doing this non-transactionally is really safe.

At the very least, we should create 2 different transactions for passing to the different stores: The assert means this `Tx` holds views over 2 different tables of the same name on different stores (in Debug only!). This is completely unexpected behaviour, and its probably doing something wrong under-the-hood.

src/node/historical_queries.h
```diff
@@ -64,12 +106,44 @@ namespace ccf::historical
...
-      // Try to insert new request
-      const auto ib = requests.insert(std::make_pair(idx, Request{}));
+      Request request;
+      request.current_stage = RequestStage::Fetching;
+      consensus::Index first_idx_to_fetch = idx;
+
+      // If the target historical entry cannot be deserialised with the first
+      // known ledger secret, record the target idx and fetch the previous
+      // historical ledger secret.
+      auto first_known_ledger_secret = get_first_known_ledger_secret();
```
eddyashton, line 116: `first_known_ledger_secret` is already a long variable name, and then we actually use `.first` and `.second` from it - I have no idea what those are. Can we destructure this here, something like:
```
auto [first_secret_idx, first_secret] = get_first_known_ledger_secret();
```

src/node/secret_broadcast.h
```diff
@@ -60,40 +62,50 @@ namespace ccf
...
-        secrets->put(
-          nid,
-          {encryption_key->public_key_pem().raw(),
-           std::move(ledger_secrets_for_node)});
+        secrets_for_nodes.emplace(nid, std::move(ledger_secrets_for_node));
+
+        // secrets->put(
+        //   nid,
+        //   {encryption_key->public_key_pem().raw(),
+        //    std::move(ledger_secrets_for_node)});
```
eddyashton, line 74: Should this be deleted?

tests/infra/network.py
```diff
@@ -480,6 +480,10 @@ def ignore_errors_on_shutdown(self):
         self.ignoring_shutdown_errors = True
 
     def stop_all_nodes(self):
+        # Verify that all apps that all app txs committed on the service can be read
```
eddyashton, line 483:
```suggestion
        # Verify that all txs committed on the service can be read
```

src/node/ledger_secrets.h
```diff
@@ -124,22 +91,13 @@ namespace ccf
...
-
-      // Taking a read dependency on the key at self, which would get updated
-      // on rekey
-      if (!self.has_value())
-      {
-        throw std::logic_error(
-          "Node id should be set before taking dependency on secrets table");
-      }
-      secrets->get(self.value());
+      secrets->get(0);
```
eddyashton, line 94: I'm unable to grok how this relates to the rest of the PR - why do we no longer store secrets per-node?

jumaffre: We still do. The previous schema of the `secrets` table was `NodeId` to `EncryptedLedgerSecrets`. This meant two things: `LedgerSecrets` had to know their node ID and the table was forever growing, even when will we delete retired nodes from the store. Passing the node ID to the historical ledger secrets was awkward so I've changed the schema for the table so that we only store secrets at key `0`.

src/node/secret_broadcast.h
jumaffre: Good spot!

src/kv/deserialise.h
jumaffre: Agreed on the broader store's `deserialise()` refactor (I've raised https://github.com/microsoft/CCF/issues/2208).

Currently, the encrypted past ledger secret map isn't the only map touched by this transaction but the other maps touched aren't considered in the deserialise return code, so I think this change is safe for now. If in the future another map of interest is touched by the rekey transaction, the unit test for historical queries will fail.

src/node/historical_queries.h
jumaffre: The transaction created here is read-only and never commits so it is only created to match the current interface. Unless I'm missing something, there's only one store involved here though (the main one, i.e. `network.tables`) so I think it is preferable to keep the current interface as it is?

eddyashton: Yeah you're right, sorry I was way off here. I thought `LedgerSecrets` had a reference to the `Store` and was getting secrets from that store's tables, but of course this doesn't happen - the tables are only referred to by name, they're not per-Store.

I'm tempted to say that `get_first` doesn't need a `Tx` though, and then this `tx` is still unnecessary? `get_latest` is a cache/interpretation of the current value, so needs a transactional dependency to make sure its still 'current'. But `first` is an independent cache unrelated to the transaction - if `first` changes, it doesn't need to invalidate any in-flight transactions?

src/kv/deserialise.h
achamayou: Why can't this be done with a hook?

src/node/historical_queries.h
jumaffre: Yep. I've now removed the `tx` argument to `get_first()`.

eddyashton (approved):

src/kv/deserialise.h
jumaffre: With the new hook interface, it seems that we can use them but I'd rather do that in a separate PR.

achamayou: Separate PR is fine of course, it's the trend I'm worried about.

achamayou (approved):

PR #2202, opened 2021-02-17
History lock

ashamis:
We have had a couple of places in the code where we access the history without a lock, version or term lock. This causes issues when there is multi-threading. Discussed this with @achamayou  and we decided to convert the term_lock in history to protect more than just the term.


achamayou (approved):

ashamis:
The daily build is failing for unrelated reasons. I am pushing this so not to slow down development

PR #2209, opened 2021-02-18
Do not crash node if ledger file cannot be committed

jumaffre:
When a ledger file is committed, it is renamed to include the end index and the `.committed` suffix. If the file cannot be renamed (e.g. if the file was deleted), the node used to crash. Instead, we now log an error and continue.


achamayou (approved):

src/host/ledger.h
```diff
@@ -147,11 +148,11 @@ namespace asynchost
...
     // Used when recovering an existing ledger file
-    LedgerFile(const std::string& dir, const std::string& file_name) :
+    LedgerFile(const std::string& dir, const std::string& file_name_) :
       dir(dir),
-      file(nullptr)
+      file_name(file_name_)
     {
-      auto full_path = (fs::path(dir) / fs::path(file_name));
+      auto file_path = (fs::path(dir) / fs::path(file_name));
       file = fopen(full_path.c_str(), "r+b");
```
achamayou, line 156:
```suggestion
      file = fopen(file_path.c_str(), "r+b");
```

src/host/ledger.h
```diff
@@ -147,12 +148,12 @@ namespace asynchost
...
-    LedgerFile(const std::string& dir, const std::string& file_name) :
+    LedgerFile(const std::string& dir, const std::string& file_name_) :
       dir(dir),
-      file(nullptr)
+      file_name(file_name_)
     {
-      auto full_path = (fs::path(dir) / fs::path(file_name));
-      file = fopen(full_path.c_str(), "r+b");
+      auto file_path = (fs::path(dir) / fs::path(file_name));
+      file = fopen(file_path.c_str(), "r+b");
```
achamayou, line 156: There's a few more s/full_path/file_path/ but the github editor won't let me change them from here.

jumaffre: Thanks! It should be fixed now 👍

eddyashton (approved):

PR #2210, opened 2021-02-18
Fix memory leaks in RSAKeyPair

wintersteiger:


jumaffre (approved):

src/tls/rsa_key_pair.h
```diff
@@ -213,6 +213,7 @@ namespace tls
         EVP_PKEY_encrypt(ctx, output.data(), &olen, input, input_size));
 
       output.resize(olen);
+      EVP_PKEY_CTX_free(ctx);
```
achamayou, line 216: Unless I'm missing something, OPENSSL_CHECK1() can throw, so we're still leaking in all error cases. Can we wrap the ctx into a unique_ptr with a custom deleter instead?

src/tls/rsa_key_pair.h
```diff
@@ -429,6 +432,7 @@ namespace tls
         ctx, output.data(), &olen, input.data(), input.size()));
 
       output.resize(olen);
+      EVP_PKEY_CTX_free(ctx);
```
achamayou, line 435: Same as above.

src/tls/rsa_key_pair.h
wintersteiger: Sure, can do. Failures of `OPENSSL_CHECK1` are pretty much fatal, so I didn't spend too much effort on cleanup, but I'll have another look around.

achamayou: We have a big catch around transaction logic though, in case anything gets thrown by user code. So these will get caught there, they won’t cause an abort. I’m happy for that to go in a follow up PR, but I would prefer if it was fixed.

achamayou (approved):

achamayou:
There is still a leak in the ca_cert test it seems:

57: =================================================================
57: ==5024==ERROR: LeakSanitizer: detected memory leaks
57: 
57: Direct leak of 120 byte(s) in 1 object(s) allocated from:
57:     #0 0x4ced43 in __interceptor_malloc (/mnt/build/1/s/build/cchost.virtual+0x4ced43)
57:     #1 0x7f3831660e58 in CRYPTO_zalloc (/usr/lib/x86_64-linux-gnu/libcrypto.so.1.1+0x17ae58)
57: 
57: Indirect leak of 136 byte(s) in 4 object(s) allocated from:
57:     #0 0x4ced43 in __interceptor_malloc (/mnt/build/1/s/build/cchost.virtual+0x4ced43)
57:     #1 0x7f3831660e58 in CRYPTO_zalloc (/usr/lib/x86_64-linux-gnu/libcrypto.so.1.1+0x17ae58)
57: 
57: SUMMARY: AddressSanitizer: 256 byte(s) leaked in 5 allocation(s).
57:

achamayou (approved):
There's still a leak in the ca_certs_test it seems.

wintersteiger:
Yeah, that one's giving me headaches. I fixed a number of them yesterday, but I can't reproduce the ca_certs_test failure on my machine and the stack traces are useless. I'll dig deeper shortly.

PR #2211, opened 2021-02-19
Resolve snmalloc zero memory issue

ashamis:
Snmalloc which we use in virtual mode added a test configuration in which it zeros out memory in the POSIX PAL. It is turned off in Release mode and was meant to be used in the snmalloc CI. What happened with CCF is that we primarily run in RelWithDebInfo and setting was not turned off in this configuration.

Matt is going to look into fixing this in the snmalloc repo but for us this is a quick fix by adding a setting to the cmake.


cmake/common.cmake
```diff
@@ -256,6 +256,8 @@ if("virtual" IN_LIST COMPILE_TARGETS)
     set(SNMALLOC_CPP)
   else()
     set(SNMALLOC_ONLY_HEADER_LIBRARY ON)
```
achamayou, line 258:
```suggestion

    # Remove once we upgrade to snmalloc 0.5.4
    set(SNMALLOC_ONLY_HEADER_LIBRARY ON)
```

achamayou (approved):

ashamis:
Ignore daily build failure, this will be fixed in #2210

Issue #2191, opened 2021-02-10
Cannot deploy JS application


msftsettiy:
After upgrading to v0.18.1, a JS application could not be deployed by submitting a "deploy_js_app" proposal. The first time the proposal was submitted, CCF responded with a 500 Internal server error. Following is the request details. Some data is masked with 'x'. 

_call_potentially_signed:84 - [xxxxxx:443] POST /gov/proposals {'content-type': 'text/plain', 'Digest': 'xxxxxxx', 'Authorization': 'Signature keyId=\"xxxxxx\",algorithm=\"hs2019\",headers=\"(request-target) digest content-length\",signature=\"xxxxxxx\"', 'content-length': '353231'} {\"script\": {\"text\": \"tables, args = ...;\\nreturn Calls:call(\\\"deploy_js_app\\\"xxxxxxx", args)\"}, \"parameter\": {\"bundle\": {\"metadata\": {\"e + 353103 chars\n"

Response from CCF:

_call_potentially_signed:84 - 500 {\"error\":{\"code\":\"InternalError\",\"message\":\"Proposal failed to bind to state.\"}}\n"

The proposal was retried.  Following is the request.

_call_potentially_signed:84 - [xxxxxx:443] POST /gov/proposals {'content-type': 'text/plain', 'Digest': xxxxxx', 'Authorization': 'Signature keyId=\"xxxxxx\",algorithm=\"hs2019\",headers=\"(request-target) digest content-length\",signature=\"xxxxx\"', 'content-length': '353231'} {\"script\": {\"text\": \"tables, args = ...;\\nreturn Calls:call(\\\"deploy_js_app\\\", args)\"}, \"parameter\": {\"bundle\": {\"metadata\": {\"e + 353103 chars\n"

Response from CCF:

_call_potentially_signed:84 - 415 {\"code\":\"UnsupportedContentType\",\"message\":\"Unsupported content type text/plain. Only application/json and application/msgpack are currently supported\"}}

In v0.16.3, the "UnsupportedContentType" error was observed the first time a proposal to deploy a JS application was sent. Upon retry, the proposal succeeded and the JS application was deployed. But in v0.18.1, the error is consistent and the JS application could not be deployed after multiple retries. 

CCF is deployed on a VM running Ubuntu Linux 18.04.

letmaik:
Regarding the second error, it seems like you're sending the wrong content type, which should be application/json.

msftsettiy:
[...]

PR #2212, opened 2021-02-19
Do not set content type on empty requests

achamayou:
And add test for empty POSTs (described in #2191), which failed to spot any problems:

```
./tests.sh -VV -R governance_no_session_auth_test --repeat-until-fail 100
...
1/1 Test #56: governance_no_session_auth_test ...   Passed   26.78 sec

The following tests passed:
        governance_no_session_auth_test

100% tests passed, 0 tests failed out of 1

Label Time Summary:
cft    =  26.78 sec*proc (1 test)
e2e    =  26.78 sec*proc (1 test)

Total Test time (real) = 2830.34 sec
```


tests/governance.py
```diff
@@ -229,6 +229,13 @@ def read_service_principal():
     return network
 
 
+@reqs.description("Test ack state digest updates")
+def test_ack_state_digest_update(network, args):
+    for _ in range(50):
```
wintersteiger, line 234: Seems like an arbitrary number - is it necessary to try this that often?

wintersteiger (approved):

tests/governance.py
achamayou: It is arbitrary, but users have reported in #2191 that they sometimes see validation failures even when they have an empty body/content-length zero. Since there is as far as I can tell no way for that to happen, this is all I can think to do to increase our chances of catching this issue if it exists. Ideas welcome.

wintersteiger: I see, thanks for the explanation! The only thing I can think of is to add random (small) delays on various operations so we get unusual execution sequences. Not sure that's really a factor here though.

achamayou: Come to think of it what we should test is mismatched content length and sizes probably.

PR #2213, opened 2021-02-19
Experiment with moving to JSON serialisation for public built-in tables

achamayou:
Draft for #1993


achamayou:
So it's not just the tree, the root and the signature, there's also a hashed_nonce every time, which serialised to a large list of zeroes every time.

I think we can split out the tree, and either hex or base64 encode the rest of the values.

PR #2216, opened 2021-02-19
Fix JWT auto-refresh on backup nodes

jumaffre:
I noticed that `auto_refresh_jwt_keys()` was called too early on join and recover, before the consensus was set, which meant that all nodes except the initial primary were refreshing the JWT keys automatically. This PR should fix it + some minor refactor to have the `CCFConfig` held by `NodeState` instead of `enclave::Enclave` to provide more convenient access to it.


src/node/node_state.h
```diff
@@ -292,15 +292,15 @@ namespace ccf
...
     //
-    NodeCreateInfo create(StartType start_type, CCFConfig& config)
+    NodeCreateInfo create(StartType start_type, CCFConfig& config_)
     {
       std::lock_guard<SpinLock> guard(lock);
       sm.expect(State::initialized);
 
-      create_node_cert(config);
-      open_frontend(ActorsType::nodes);
+      config = std::move(config_);
```
letmaik, line 300: Is this ok if the arg is not `&&`?

eddyashton: This is well-defined, but idiosyncratic. By passing as `&`, the caller lets us modify the value, but probably doesn't expect it to be moved from (potentially leaving their instance as complete junk). We should pass this as `&&` if we want to move from it.

jumaffre: Good spot! It would still use the move constructor but it is confusing for the caller. Fixed to now use `&&`.

letmaik (approved):

achamayou (approved):

PR #2217, opened 2021-02-19
Add temporary multi-threading CI

jumaffre:
I will create the pipeline once this is merged on `main`. The pipeline has run against my fork (see https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=19284&view=logs&j=a45bc589-b130-5dc2-b16b-03ef91a336e9), with two test failures (but at least, the YAML is sound!).


.multi-thread.yml
```diff
@@ -0,0 +1,33 @@
...
+  - template: .azure-pipelines-templates/common.yml
+    parameters:
+      target: SGX
+      env:
+        container: sgx
+        pool: Ubuntu-1804-DC8_v2
+      cmake_args: "-DCOMPILE_TARGETS=sgx -DWORKER_THREADS=2"
+      suffix: "MultiThread"
+      artifact_name: "MultiThread"
+      ctest_filter: '-LE "perf"'
```
jumaffre, line 31: Excluding perf for now since the perf infra doesn't run in docker and the perf numbers are not correct with multi-threading (only one client is spawned).

achamayou (approved):

.multi-thread.yml
```diff
@@ -0,0 +1,33 @@
+trigger:
+  batch: true
+  branches:
+    include:
+      - main
+
+pr:
```
achamayou, line 7: Let's not have this for PRs for now please, there isn't enough CI capacity.

jumaffre: Good point!

PR #2218, opened 2021-02-19
Add LRU for signature Verifiers

### LINKED ISSUE #789: Frontend should store a fixed number of verifiers

eddyashton:
Resolves #789.

- Adds a basic LRU implementation under `ds/`
- Adds some unit tests for this, checking it drops the correct old entries at the correct time
- Use this LRU to store contents of each `VerifierCache` from `sig_auth.h`

I hacked up an e2e test to try and confirm that signatures from users in the cache were faster than those not in the cache, but it got ugly quickly so I dropped it (need to correct for the slow first transaction on each socket establishment, need a contextlib stack of clients, need to ignore slow outliers/do average comparisons, and the resulting difference we're trying to measure ends up very small).


jumaffre (approved):
As CCF is moving towards multi-threading, should we mention somewhere that this LRU is not thread safe?

eddyashton:
> As CCF is moving towards multi-threading, should we mention somewhere that this LRU is not thread safe?

I don't think so, for a few reasons:
- The default assumption should be that containers aren't internally threadsafe (cf the standard library), and instead the few which _are_ threadsafe need to call this out
- There's already a disclaimer saying "if you're using the LRU, check it does what you want", and its small enough to read and see there's no synchronisation logic
- Anybody who wanted to use the LRU would presumably search to see how its currently used, and see that its explicitly guarded by an external `SpinLock`

achamayou (approved):

PR #2222, opened 2021-02-22
Support CA cert bundles

### LINKED ISSUE #2221: Allow storing CA cert bundles
Currently, `set_ca_cert` allows to store single named certs, which are used for the JWT auto-refresh feature.
It makes sense to generalize this to allow storing named lists of CA certs. This supports scenarios where a trusted/curated Root CA list is imported wholesale, e.g. from the Microsoft Trust Root Program or the Mozilla CA Cert Program. Note that the existing use case of a single cert will be supported as before (as that is just a special case).

The following changes are necessary:
- `set_ca_cert` proposal: Store PEM as-is (multiple certs are a concatenation, no change to payload) instead of converting to DER (there is no such thing as concatenated DER certs, see also docs on `mbedtls_x509_crt_parse`)
- Rename proposal `{set|remove}_ca_cert` to `{set|remove}_ca_cert_bundle`
- Rename kv table from `public:ccf.gov.jwt.ca_certs_der` to `public:ccf.gov.tls.ca_cert_bundles`
- Rename `ca_cert_name` in JWT Issuer metadata to `ca_cert_bundle_name`

Note that there won't be a performance impact from this change since JWT key auto-refresh is not on the critical path.

letmaik:
Closes #2221.


eddyashton:
Since a bundle is "just" a concatenation, I wonder if this should be just the plural `set_ca_certs` rather than `set_ca_certs_bundle`? If I have a single cert I want to add, it may not be clear to me how I combine/wrap those into a bundle? But if I saw `set_ca_certs` I'd probably blindly submit my single entry there? And in the reverse case where I already have a bundle (and know its a mere concatenation), I'd still be happy with `set_ca_certs`? This bike shed has a very nice felt roof.

letmaik:
I think semantically there's a big difference and I would rather be more explicit at the possible cost of confusing people so that they have to check the docs.

tests/infra/ca_certs.py
```diff
@@ -0,0 +1,109 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
achamayou, line 1: infra/ is for infrastructure used by tests, not tests themselves, why is this moved here?

letmaik: It's a duplicate and I've no idea how it ended up there. Gone now.

achamayou (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed format of `notBefore` and `notAfter` in node and network certificates (#2243).
 - CCF now depends on [Open Enclave 0.14](https://github.com/openenclave/openenclave/releases/tag/v0.14.0).
 
+### Changed
```
achamayou, line 15: @letmaik can you create a 0.18.4 section and move this up?

letmaik: done

PR #2225, opened 2021-02-22
Refactor crypto library

wintersteiger:
Refactoring of the crypto code into a proper library with implementations hidden behind abstract interfaces.


wintersteiger:
(Build failures from main)

CMakeLists.txt
```diff
@@ -479,24 +469,23 @@ if(BUILD_TESTS)
   add_picobench(
     merkle_bench
     SRCS src/node/test/merkle_bench.cpp
-    LINK_LIBS ccfcrypto.host crypto
+    LINK_LIBS
```
achamayou, line 472:
```suggestion
```

CMakeLists.txt
```diff
@@ -479,24 +469,23 @@ if(BUILD_TESTS)
...
   )
   add_picobench(
     history_bench
     SRCS src/node/test/history_bench.cpp
-    LINK_LIBS ccfcrypto.host crypto
+    LINK_LIBS
```
achamayou, line 477:
```suggestion
```

CMakeLists.txt
```diff
@@ -479,24 +469,23 @@ if(BUILD_TESTS)
...
   )
   add_picobench(
     kv_bench
-    SRCS src/kv/test/kv_bench.cpp src/crypto/symmetric_key.cpp
-         src/enclave/thread_local.cpp
-    LINK_LIBS ccfcrypto.host
+    SRCS src/kv/test/kv_bench.cpp src/enclave/thread_local.cpp
+    LINK_LIBS
```
achamayou, line 482:
```suggestion
```

CMakeLists.txt
```diff
@@ -479,24 +469,23 @@ if(BUILD_TESTS)
...
   )
   add_picobench(hash_bench SRCS src/ds/test/hash_bench.cpp)
   add_picobench(
     digest_bench
     SRCS src/crypto/test/digest_bench.cpp
-    LINK_LIBS ccfcrypto.host
+    LINK_LIBS
```
achamayou, line 486:
```suggestion
```

cmake/crypto.cmake
```diff
@@ -28,8 +47,9 @@ endif()
 
 add_library(ccfcrypto.host STATIC ${CCFCRYPTO_SRC})
 add_san(ccfcrypto.host)
-target_compile_options(ccfcrypto.host PRIVATE -stdlib=libc++)
-target_link_libraries(ccfcrypto.host PRIVATE crypto)
+target_compile_options(ccfcrypto.host PUBLIC -stdlib=libc++)
+target_link_options(ccfcrypto.host PUBLIC -stdlib=libc++)
```
achamayou, line 51: Maybe we could define a use_libcxx(), like we've done for mbedtls etc.

src/crypto/test/crypto.cpp
```diff
@@ -375,15 +384,120 @@ static const vector<uint8_t>& getRawKey()
...
-  REQUIRE(k2.decrypt(h.get_iv(), h.tag, p, nullb, p.p));
+  auto k2 = crypto::make_key_aes_gcm(getRawKey());
+  REQUIRE(k2->decrypt(h.get_iv(), h.tag, p, nullb, p.p));
+}
+
+TEST_CASE("AES mbedTLS vs OpenSSL")
+{
+  auto key = getRawKey();
+
+  // std::cout << "Contents: ";
```
achamayou, line 404: Commented out code.

src/crypto/test/crypto.cpp
```diff
@@ -375,15 +384,120 @@ static const vector<uint8_t>& getRawKey()
...
+    REQUIRE(decrypted.size() == contents_.size());
+    REQUIRE(memcmp(decrypted.data(), contents_.data(), sizeof(contents_)) == 0);
+  }
+}
+
+TEST_CASE("AES mbedTLS vs OpenSSL + AAD")
+{
+  auto key = getRawKey();
+
+  // std::cout << "Contents: ";
```
achamayou, line 456: Commented out code

cmake/crypto.cmake
wintersteiger: There is `LINK_LIBCXX` in `common.cmake`, should I use that?

src/crypto/test/crypto.cpp
```diff
@@ -375,15 +384,120 @@ static const vector<uint8_t>& getRawKey()
...
+      h.tag,
+      {encrypted.data(), encrypted.size()},
+      aad,
+      decrypted.data()));
+
+    REQUIRE(decrypted.size() == contents_.size());
+    REQUIRE(memcmp(decrypted.data(), contents_.data(), sizeof(contents_)) == 0);
+  }
+
+  { // OpenSSL -> mbedTLS
```
achamayou, line 486: Doctest has a nice INFO() macro we often use, which if you run with -s will get logged to stdout. Here it'd be nice to have INFO("OpenSSL vs mbedTLS") or something like that rather than a comment.

src/crypto/rsa_key_pair.cpp
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  RSAPublicKeyPtr make_rsa_public_key(const std::vector<uint8_t>& der)
+  {
+    return std::make_shared<RSAPublicKeyImpl>(der);
+  }
+
+  RSAPublicKeyPtr make_rsa_public_key(const uint8_t* data, size_t size)
+  {
+    if (size < 10 || strncmp("-----BEGIN", (char*)data, 10) != 0)
```
achamayou, line 31: It'd be good to have the string a constexpr, the len as a name derived constexpr rather than literals inline.

cmake/crypto.cmake
achamayou: Yes!

achamayou (approved):

src/crypto/rsa_key_pair.cpp
wintersteiger: Sure, np. What do mean by "name derived"?

achamayou: BEGIN and BEGIN_LEN for example? Looking into it, it seems that strlen(literal) is optimized out anyway.

achamayou: image

achamayou: image

wintersteiger: Oh sure, the names are fine. In the meanwhile I found out that there is `std::char_traits<char>::length(...);` for this and similar purposes now and it seems to work.

PR #2232, opened 2021-02-24
Driver upgrade

achamayou:


getting_started/setup_vm/roles/intel/vars/common.yml
```diff
@@ -3,8 +3,8 @@
 
 ---
 flc_enabled: true
-intel_sgx_w_flc_driver_url: "https://download.01.org/intel-sgx/sgx-dcap/1.9/linux/distro/ubuntu18.04-server/sgx_linux_x64_driver_1.36.2.bin"
-intel_sgx1_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.12/distro/ubuntu18.04-server/sgx_linux_x64_driver_1.36.2.bin"
+intel_sgx_w_flc_driver_url: "https://download.01.org/intel-sgx/sgx-dcap/1.7/linux/distro/ubuntu18.04-server/sgx_linux_x64_driver_1.35.bin"
+intel_sgx1_driver_url: "https://download.01.org/intel-sgx/sgx-linux/2.13/distro/ubuntu18.04-server/sgx_linux_x64_driver_2.11.0_0373e2e.bin"
```
eddyashton, line 7: What are all of these versions? `1.9` to `1.7` and `1.36.2` to `1.35` look like downgrades rather than upgrades? `2.12` to `2.13` looks like an upgrade, but then we're getting `2.11` from it?

(I can believe the answer here is just "Intel has named everything daftly", just looking for some context/explanation)

achamayou: We pick this up from OE, there is:

1. the non FLC stuff (Skylake), which we do not care about at all. We could remove it altogether.
2. the FLC stuff. That's got a separate version for the driver and for the sgx-linux project. Why Intel went from 1.x to 2.x between 2.11 and 2.13 I don't know. This is supposed to go away quite soon because of the upstreaming, but until 20.04, OE continues to have a specific version they install and test, and I think it's right to stay aligned with that (or at least to make it easy for users to do so).

eddyashton (approved):

PR #2233, opened 2021-02-24
Historic queries: Range queries and app-dictated handles

eddyashton:
This adds 2 main features to make the historical query system far more useful:

- Adds the ability to ask for a range of entries rather than a single point
- Replaces the hard-coded LRU-keyed-by-requested-index cache with a fully app-controlled system

It also adds a big lock around all calls into `StateCache` so that it can be used from multiple threads, and automatic garbage collection of historical queries after a (configurable) expiry time. There are new tests for range queries and multi-threaded queries, and a refactoring of the unit tests for readability.

#### TODOs, some for this PR and some for future
- [x] Add a sample endpoint to the logging app using range queries
- [x] Describe this in `CHANGELOG`
- [ ] Do some performance testing to get an idea how fast we can service historical queries
- [x] Add some `.rst` docs, or at least make sure the Doxygen is exposed
- [ ] Explore asking for/receiving ranges over the ringbuffer, rather than individual entries (I'm wary of this leading to arbitrarily large ringbuffer messages, left to the app to limit, or a duplication of fragmenting/batching logic)
- [ ] Explore making the app-defined handle a string or byte-array, so they have more range and flexibility than `size_t`? But perhaps we just recommend they hash any complex keys

There are a few followup tasks that I'll document elsewhere


src/node/test/historical_queries.cpp
```diff
@@ -48,50 +126,76 @@ void initialise_store(kv::Store& store, bool initialise_ledger_rekey = false)
...
 
-  LOG_DEBUG_FMT("Recording entry at {}", idx);
+kv::Version rekey(
+  kv::Store& kv_store,
+  const std::shared_ptr<ccf::LedgerSecrets>& ledger_secrets)
+{
+  // This isn't really used, but is needed for ShareManager, so can be recreated each time here
+  ccf::NetworkState network;
+  network.ledger_secrets = ledger_secrets;
+  ccf::ShareManager share_manager(network);
```
eddyashton: (Totally out-of-scope for this PR) This pattern is a little odd. I think `ShareManager` should really be constructed with `ledger_secrets` directly, but it (like many other systems) relies on `network` to give it table definitions. We can replace every `tx.rw(network.foo)` with `tx.rw<Foo>(Tables::FOO)`, but its a bit riskier (needing to keep typename and table-name in sync, rather than using a pre-constructed pairing). I think we can get around this with statically named tables (so you could write `tx.rw<Nodes>()` for instance, since we know the name the `Nodes` table should _always_ have), but its not a priority fix right now.

src/node/test/historical_queries.cpp
```diff
@@ -462,21 +881,19 @@ TEST_CASE("Recover historical ledger secrets")
     // Provide target and subsequent entries until next signature
     for (size_t i = second_index; i <= signature_index; ++i)
     {
-      REQUIRE(provide_ledger_entry(i));
+      provide_ledger_entry(i);
```
eddyashton, line 884: Several of these `provide_ledger_entries` can no longer be `REQUIRE`d because the `StateCache` may already have this entry from a previous request, and not have asked for it again. We could be precise about which were asked for (the return value of this should always match "did the Cache ask for it"), but for simplicity for now I've just dropped checking some of these return values.

src/node/historical_queries.h
```diff
@@ -50,128 +58,296 @@ namespace ccf::historical
...
+      std::vector<StoreDetailsPtr> requested_stores;
+      std::chrono::milliseconds time_to_expiry;
 
-      // Add request to front of list, most recently requested
-      recent_requests.emplace_front(idx);
+      // An entry from outside the requested range containing the next signature
+      // may be needed to trust this range. It is stored here, distinct from
+      // user-requested stores.
+      std::optional<std::pair<consensus::Index, StoreDetailsPtr>>
+        supporting_signature;
```
achamayou, line 97: `evidence` would be preferable, for CFT this is a single signature, but this is really "consensus evidence" in general.

src/node/historical_queries.h
```diff
@@ -50,128 +58,296 @@ namespace ccf::historical
...
+      StorePtr store = nullptr;
+      bool is_signature = false;
+    };
+    using StoreDetailsPtr = std::shared_ptr<StoreDetails>;
+
+    struct Request
     {
-      // To avoid duplicates, remove index if it was already requested
-      recent_requests.remove(idx);
+      consensus::Index first_requested_idx = 0;
```
achamayou, line 88: Index is the Raft term, we want to use seqno in consensus contexts, and version in the KV and possibly the ledger, but I think we would like to avoid introducing new references to index as much as possible.

src/node/historical_queries_interface.h
```diff
@@ -6,144 +6,100 @@
 #include "kv/store.h"
 #include "node/rpc/endpoint_registry.h"
 
+#include <chrono>
```
achamayou, line 9: Can this move to a newly created include/, which contains our public includes please?

src/node/historical_queries_interface.h
```diff
@@ -6,144 +6,100 @@
...
   public:
     virtual ~AbstractStateCache() = default;
 
-    virtual StorePtr get_store_at(consensus::Index idx) = 0;
+    /** Set the default time after which a request's state will be deleted, and
+     * will not be accessible without retrieving it again from the ledger. Any
+     * call to get_store_XXX which doesn't pass an explicit expire_after will
+     * reset the timer to this default duration.
+     */
+    virtual void set_default_expiry_duration(ExpiryDuration duration) = 0;
```
achamayou, line 47:
```suggestion
    virtual void set_default_expiry_duration(ExpiryDuration seconds) = 0;
```

achamayou:
This needs the first of the TODO items, but I'd really like to have to some end to end tests for this.

src/node/historical_queries.h
eddyashton: I agree - I'll go through and change these in historical queries from `consensus::Index` to `kv::Consensus::SeqNo`, and rename all the `_idx` to `_seqno`. Thinking more broadly, `consensus::Index` should really be renamed to `consensus::SeqNo`, and `kv::Consensus::SeqNo` shouldn't exist (it should just be `consensus::SeqNo`. But that's a lot of renames, so I'll punt for this PR.

achamayou: Yeah, not everything needs cleaning up in the same PR, but we can at least avoid introducing more things that'll need cleaning up.

src/node/ledger_secrets.h
```diff
@@ -111,6 +111,16 @@ namespace ccf
       ledger_secrets.emplace(initial_version, make_ledger_secret());
     }
 
+    void init_from_map(LedgerSecretsMap&& ledger_secrets_)
```
jumaffre, line 114: Could you remove the `LedgerSecrets` constructor that takes a `LedgerSecretsMap&&`?

src/node/historical_queries.h
```diff
@@ -19,18 +19,26 @@
 
 namespace ccf::historical
 {
+  static std::optional<ccf::PrimarySignature> get_signature(
+    const StorePtr& sig_store)
+  {
+    auto tx = sig_store->create_tx();
```
jumaffre, line 25:
```suggestion
    auto tx = sig_store->create_read_only_tx();
```

src/node/historical_queries.h
eddyashton: I've now done this. This involved a lot more casting than I expected, because `kv::SeqNo` is currently signed. I think we want `SeqNo` to be unsigned and only `Version` needs to (maybe) be signed, to handle deletions/other special values? Its possible both can be unsigned with `0` or `MAX` as `NoVersion`.

src/node/historical_queries_interface.h
eddyashton: Done. This is now `include/ccf/historical_queries_interface.h`, and we have a global `include_directories(${CCF_DIR}/include)` in `common.cmake`.

eddyashton:
Going round in circles trying to fix the new test failure. I'm fighting the automated verification done by `LoggingTxs`. It wants to submit _every_ transaction to the service - if somebody else logs something different, it gets very confused when the "current" value doesn't match what it expects. And it can't just fetch everything as a historical query, because `js` and `bft` tests don't support that (sidebar: I think we have some fragility where so far we're lucky that we've not logged multiple things to the same ID in either of these tests, and nothing explicit is preventing them from calling a non-existent `/log/private/historical`?). I've tried adding flags to make my e2e range test go through `LoggingTxs`, but this adds several dodgy modifications to stop it waiting for commit of every transaction. I think I'm near a solution now, but this may need an axe/revisiting in future.

samples/apps/logging/logging.cpp
```diff
@@ -581,6 +582,140 @@ namespace loggingapp
...
+      auto get_historical_range = [&, this](ccf::EndpointContext& args) {
+        // Parse request body
+        const auto query_j =
+          ccf::jsonhandler::get_params_from_query(args.rpc_ctx);
+        const auto in = query_j.get<LoggingGetHistoricalRange::In>();
+
+        // Range must be in order
+        if (in.to_seqno < in.from_seqno)
+        {
+          args.rpc_ctx->set_response_status(HTTP_STATUS_INTERNAL_SERVER_ERROR);
```
letmaik, line 594: All of those should use `args.rpc_ctx->set_error()`.

letmaik, line 594: Not a server error, but a client error.

samples/apps/logging/logging.cpp
```diff
@@ -581,6 +582,140 @@ namespace loggingapp
...
+
+          if (v.has_value())
+          {
+            LoggingGetHistoricalRange::Entry e;
+            e.seqno = store_seqno;
+            e.id = in.id;
+            e.msg = v.value();
+            response.entries.push_back(e);
+          }
+          // We do not include an entry when the given key wasn't modified at
```
letmaik, line 694: Does that mean I cannot access other store data at that seqno if it hasn't been modified? Will this be possible in the future?

samples/apps/logging/logging.cpp
```diff
@@ -581,6 +582,140 @@ namespace loggingapp
...
+        nlohmann::json j_response = response;
+        args.rpc_ctx->set_response_status(HTTP_STATUS_OK);
+        args.rpc_ctx->set_response_header(
+          http::headers::CONTENT_TYPE, http::headervalues::contenttype::TEXT);
+        args.rpc_ctx->set_response_body(j_response.dump());
+
+        // ALSO: Assume this response makes it all the way to the client, and
+        // they're finished with it, so we can drop the retrieved state. In a
+        // real app this should be driven by a separate client request!
+        historical_cache.drop_request(handle);
```
achamayou, line 708: I think it's more likely that a real app would want a cap of some sort on the historical cache size, and let stuff drop when either that or the expiry time hits.

achamayou (approved):

samples/apps/logging/logging.cpp
eddyashton: Thanks - I've replaced all of the `set_response_status(ERROR)` blocks in `logging.cpp` with `set_error(ERROR)`

samples/apps/logging/logging.cpp
eddyashton: Yes - this is how the historical query system currently operates. The retrieved store is constructed from the deserialised write set at that entry, so it only contains the state _written by_ that seqno. It does not contain the complete state that was visible at that seqno. The latter is possible to build with snapshots, but isn't needed so we haven't built it yet.

samples/apps/logging/logging.cpp
eddyashton: I agree. I've extended the comment to suggest this. I was aiming to show here "you should think about and call this somewhere", but also "this approach is a bodge for this demo and not something you should copy verbatim".

Issue #1852, opened 2020-11-02
AFT: Support reconfiguration


ashamis:
This is the top level ticket for reconfiguration. This will be broken down into multiple tickets and PRs.

- [ ] Change the BFT node_id to be the same as in CFT. To do this we will need to keep a node_id -> ordered_id mapping and have a data structure that can allow us to deterministically decide which node is the primary in any given view within a configuration.
- [ ] Change BFT to use replica ids which have the same NodeIds that are derived from the public cert like CFT

jumaffre:
As discussed with @ashamis and @achamayou and follow-up from #2241: we still use monotonic node IDs in BFT so that the primary node ID in a given view can be computed deterministically by all nodes in the network. Once reconfiguration works in the BFT variant of the consensus (and we can skip nodes during a view change, so that a new added node doesn't try to become primary when it is just added), we should change the node IDs in BFT to also be the hash of the node's public key.

PR #2241, opened 2021-02-25
Use hash of node's public key as unique node identifier

jumaffre:
Part of #1378 

~Very much draft for now as there are still some issues with BFT and two of the unit tests that will need further refactoring, but I want to measure the performance impact of this change first.~ **Now read for review**

Node IDs used to be a monotonic counter, which could lead to replay attacks, and weren't convenient for operators who wouldn't know the ID of their node until this one had actually joined the network. This PR aims to solve these issues by using a fixed identifier (hex-encoded string of hash of node's identity public key) as the node's ID.

Notable changes:
- We now use a `std::optional<NodeId>` (instead of `NoNode`) in the consensus to differentiate whether we know the current leader node. This has some minor effects on the endpoints that return the current primary ID.
- We only display the first `10` characters of the node ID in the node's log, e.g.:
```
2021-03-03T14:07:34.809510Z -0.001 0   [debug] ../src/consensus/aft/raft.h:994      | Send append entries from dc84b80910 to faeac68820: 11 to 10 (2)
```
- Node-to-node channel refactoring so that the `NodeId` is no longer included in the message headers (e.g. Raft's append entries). This means we don't pay the cost of integrity-protect it, which is fine because the `NodeId` identifies the channel to use for decryption.
- The Python infra is mostly unchanged: nodes are still labelled incrementally (e.g. the node folder is still `/workspace/test_0` as we have to create those in advance). The infra keeps track of the "real" new node ID too though.
- BFT currently relies on a stable order of node IDs so that the primary at a given view can be computed deterministically by all nodes. So in BFT, we still use monotonic node IDs for now (see https://github.com/microsoft/CCF/issues/1852 for more detail).


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- Node unique identifier is hex-encoded string of node's identity public key (#2241).
```
jumaffre, line 12:
```suggestion
- Node unique identifier is the hex-encoded string of the SHA-256 digest of the node's identity public key (#2241).
```

doc/build_apps/demo.rst
```diff
@@ -31,7 +31,7 @@ The following command will run a simple one node test network on a single machin
     [16:14:05.294] Starting 1 CCF node...
     [16:14:05.295] Virtual mode enabled
     [16:14:10.010] Started CCF network with the following nodes:
-    [16:14:10.011]   Node [0] = https://127.0.0.1:8000
+    [16:14:10.011]   Node [50211327a77fc16dd2fba8fae5fffac3df909fceeb307cf804a4125ae2679007] = https://127.0.0.1:8000
```
eddyashton, line 34: Should this also be a limited prefix rather than the full digest? The interesting information on this line is the URL, and the prefix should be sufficient to go find this node's real ID if needed.

CMakeLists.txt
```diff
@@ -443,6 +443,7 @@ if(BUILD_TESTS)
       raft_driver ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/driver.cpp
     )
     target_link_libraries(raft_driver PRIVATE ccfcrypto.host)
+    use_client_mbedtls(raft_driver)
```
achamayou, line 446: ?

src/consensus/aft/async_execution.h
```diff
@@ -64,112 +79,137 @@ namespace aft
...
     AppendEntriesResponse hdr;
   };
 
   class SignedAppendEntryResponseCallback : public AbstractMsgCallback
   {
   public:
     SignedAppendEntryResponseCallback(
-      AbstractConsensusCallback& store_, SignedAppendEntriesResponse&& hdr_) :
+      AbstractConsensusCallback& store_,
+      NodeId from_,
```
eddyashton, line 106: Several `NodeId from_`s that could be `const NodeId& from_` in this file.

src/consensus/aft/raft_types.h
```diff
@@ -138,11 +137,9 @@ namespace aft
   struct RaftHeader
   {
     RaftMsgType msg;
```
eddyashton, line 139: Why is the `from_node` removed from here, and passed separately as an argument in so many places (this is why we're now doing `recv_append_entries(node_id, append_entries, ...)` rather than just `recv_append_entries(append_entries, ...)`)?

src/consensus/aft/raft_types.h
```diff
@@ -20,7 +21,7 @@ namespace aft
 {
   using Index = int64_t;
   using Term = int64_t;
-  using NodeId = uint64_t;
+  using NodeId = ccf::NodeId;
```
eddyashton, line 24: Do we still need this alias, or should everything be using `ccf::NodeId` rather than `aft::NodeId`?

src/crypto/public_key.h
```diff
@@ -14,6 +14,12 @@
 
 namespace crypto
 {
+  static inline std::string get_public_key_pem_hash_hex(
```
letmaik, line 17: Can you have a look at https://github.com/microsoft/CCF/issues/1378#issuecomment-771581216? It seems like this is taking the hash over the PEM and not DER.

src/ds/json.h
```diff
@@ -418,8 +418,8 @@ namespace std
  * transformation. Missing elements will cause errors to be raised. This assumes
  * that from_json, to_json, are implemented for each member
  * field type, either manually or through these macros. Additionally, you will
- * schema_name, fill_json_schema, and add_schema_components to be defined for
- * OpenAPI schema generation.
+ * need schema_name, fill_json_schema, and add_schema_components to be defined
```
eddyashton, line 421: While you're here, could you please remove the duplicate `fill_json_schema` from the first line of this comment?

src/host/node_connections.h
```diff
@@ -17,20 +17,28 @@ namespace asynchost
...
         node(node)
       {}
 
       void on_read(size_t len, uint8_t*& incoming)
       {
-        LOG_DEBUG_FMT("from node {} received {} bytes", node, len);
+        // LOG_DEBUG_FMT(
+        //   "from node {} received {} bytes",
+        //   node.value_or(UnassociatedNode),
+        //   len);
```
eddyashton, line 41: Should this log (and the one on line 64) be uncommented?

src/kv/test/stub_consensus.h
```diff
@@ -11,17 +11,22 @@
 
 namespace kv
 {
+  static NodeId PrimaryNodeId = std::string("PrimaryNodeId");
+  static NodeId FirstBackupNodeId = std::string("FirstBackupNodeId");
+  static NodeId SecondBackupNodeId = std::string("SecondBackupNodeId");
+  static NodeId ThirdBackupNodeId = std::string("ThirdBackupNodeId");
```
eddyashton, line 17: If these are just used for testing, they shouldn't be in the `kv` namespace. That's probably true of the existing things in this namespace too, but `kv::PrimaryNodeId` is more obviously something which shouldn't exist.

src/host/node_connections.h
```diff
@@ -94,56 +111,58 @@ namespace asynchost
...
       {
-        LOG_DEBUG_FMT("node connect failed {}", node);
+        LOG_DEBUG_FMT("node connect failed {}", node.value());
         reconnect();
       }
 
       void on_disconnect()
       {
-        LOG_DEBUG_FMT("node disconnect failed {}", node);
+        LOG_DEBUG_FMT("node disconnect failed {}", node.value());
```
eddyashton, line 159: I think this should be `node` rather than `node.value()`, to use the custom formatter which prints only the first 10 characters?

src/node/node_id.h
```diff
@@ -0,0 +1,153 @@
...
+  inline void to_json(nlohmann::json& j, const NodeId& node_id)
+  {
+    j = node_id.id;
+  }
+
+  inline void from_json(const nlohmann::json& j, NodeId& node_id)
+  {
+    if (j.is_string())
+    {
+      node_id = j.get<std::string>();
```
eddyashton, line 76: Should we check that the JSON string looks like a digest here - that its the correct length at least, and contains only expected characters?

src/node/node_id.h
```diff
@@ -0,0 +1,153 @@
...
+    }
+    else
+    {
+      throw std::runtime_error(
```
eddyashton, line 80: Minor: If you throw a `JsonParseError` here, then the frontend might wrap it in a more helpful message (with a JSON pointer to where in the nested structure was wrong). This is undocumented and not used elsewhere though, so perhaps not worth relying on.

CMakeLists.txt
wintersteiger: `use_*_mbedtls` should not be necessary when ccfcrypto.host/.enclave is linked in. I think we could remove those functions, but haven't actually tried to do so yet.

doc/build_apps/demo.rst
jumaffre: Good point. I wrote this as the short hash at first but I believe that the short hash is for internal use only (e.g. it is the full hash that has to be included in a proposal to retire a node). But then the sandbox isn't used for node operation I don't think so maybe that's OK? Alternatively, I'd be happy to leave this as the `node.local_node_id` (i.e. `0`) too to avoid confusion. What do you think?

CMakeLists.txt
jumaffre: Merge conflict gone wrong.

src/consensus/aft/raft_types.h
jumaffre: The `NodeId` used to be included in all consensus message headers, which was integrity protected by each channel separately. With the previous scheme, this data was fairly small but the new `NodeId` scheme means that we would be integrity-protecting a much bigger object, for each target node. Since the peer's `NodeId` is already included in each channel and is already an input to the integrity verification (i.e. `from_node` already routes to the correct channel), I've removed it from all the consensus headers. I don't have benchmarks to back this up but at the price of a "small" refactoring (which ended up being bigger than I thought), we don't do this anymore.

src/node/node_to_node.h
```diff
@@ -184,89 +182,92 @@ namespace ccf
...
-    void complete_key_exchange(const uint8_t* data, size_t size)
+    void complete_key_exchange(
+      const NodeId& from, const uint8_t* data, size_t size)
     {
       // Called on channel initiator when a key exchange response message is
       // received from the target
-      const auto& ke = serialized::overlay<ChannelHeader>(data, size);
-
-      auto n2n_channel = channels->get(ke.from_node);
+      // const auto& ke = serialized::overlay<ChannelHeader>(data, size);
```
eddyashton, line 251: Commented out, can be removed?

src/crypto/key_pair.h
```diff
@@ -3,10 +3,12 @@
 #pragma once
 
 #include "curve.h"
+#include "ds/logger.h" // TODO: Remove!
```
achamayou, line 6: =_=

src/consensus/aft/async_execution.h
jumaffre: Thanks! I thought I'd caught them all but clearly didn't. This should be fixed now.

src/crypto/public_key.h
jumaffre: This is now done: we use the SHA-256 hash of the DER-encoded public key (hex-encoded) as the node ID (this seems to be standard, e.g. https://stackoverflow.com/questions/40404963/how-do-i-get-public-key-hash-for-ssl-pinning). We also use the same value in the SGX quote's report data.

src/host/node_connections.h
jumaffre: `node` here is actually a `std::optional` (since incoming channels are unassociated when created), so this will print the short hash.

src/node/node_id.h
jumaffre: We now check for both length and hex encoding.

src/kv/test/stub_consensus.h
jumaffre: Renamed this namespace from `kv` to `kv::test`/

doc/build_apps/demo.rst
jumaffre: As discussed, let's go for the local Python infra monotonic node ID as it is easier to find out in which directory the node is started.

python/utils/verify_quote.sh
```diff
@@ -8,20 +8,12 @@ quote_file_name="quote.bin"
...
-    verify_quoted_cert=true
-fi
-
 function usage()
 {
     echo "Usage:"
     echo "  $0 https://<node-address> [--mrenclave <mrenclave_hex>] [CURL_OPTIONS]"
     echo "Verify target node's remote attestation quote."
-    echo "Verification involves confirming that the public key of the node certificate matches the SGX report data and that the mrenclave included in the quote is trusted."
+    echo "Verification involves confirming that the public key (DER encoded) of the node certificate matches the SGX report data and that the mrenclave included in the quote is trusted."
```
achamayou, line 16:
```suggestion
    echo "Verification involves confirming that the public key (DER encoded) of the node certificate matches the SGX report data and that the MRENCLAVE included in the quote is trusted."
```

achamayou (approved):

PR #2246, opened 2021-02-26
More key wrapping

### LINKED ISSUE #2230: Add AES-KW key wrap algorithm to wrapKey JS API
Currently, only RSA-OEAP-256 is supported in the JS API `ccf.wrapKey`. Adding `AES-KW` would be a good addition, as it allows to implement CKM_RSA_AES_KEY_WRAP (see also https://github.com/microsoft/CCF/issues/1204). 

mbedtls has support for AES-KW: https://github.com/ARMmbed/mbedtls/blob/a0fd0f85314de324cda024422e20e2d80294b5e6/library/cipher_wrap.c#L2179-L2210

OpenSSL supports it as well: https://www.openssl.org/docs/man1.1.1/man3/EVP_aes_256_ccm.html

wintersteiger:
This adds more key wrapping methods to the crypto library and the JS app. Fixes #2230.


src/crypto/key_wrap.cpp
```diff
@@ -0,0 +1,159 @@
...
+{
+  // With inspiration from
+  // http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/cos01/pkcs11-curr-v2.40-cos01.html
+
+  std::vector<uint8_t> ckm_rsa_pkcs_oaep_wrap(
+    RSAPublicKeyPtr wrapping_key,
+    const std::vector<uint8_t>& unwrapped,
+    const std::vector<uint8_t>& label)
+  {
+    return wrapping_key->wrap(unwrapped, label);
```
letmaik, line 25: Probably `wrap` should be renamed now to make it more clear which alg it is.

src/crypto/key_wrap.cpp
```diff
@@ -0,0 +1,159 @@
...
+  {
+    auto pk = make_rsa_key_pair(wrapping_key);
+    return ckm_rsa_pkcs_oaep_unwrap(pk, wrapped, label);
+  }
+
+  std::vector<uint8_t> ckm_aes_key_wrap_pad(
+    const std::vector<uint8_t>& wrapping_key,
+    const std::vector<uint8_t>& unwrapped)
+  {
+    auto ossl = std::make_unique<KeyAesGcm_OpenSSL>(wrapping_key);
```
letmaik, line 58: Can we avoid hard-coding to OpenSSL somehow?

samples/apps/forum/src/types/ccf.ts
```diff
@@ -68,7 +70,7 @@ export interface CCF {
   wrapKey(
     key: ArrayBuffer,
     wrappingKey: ArrayBuffer,
-    wrapAlgo: WrapAlgo
```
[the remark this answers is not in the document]
wintersteiger: That's as close as possible. There are additional parameters that need to be passed somewhere.

src/apps/js_generic/js_generic.cpp
```diff
@@ -224,26 +226,81 @@ namespace ccfapp
...
+
+        auto wrapped_key = crypto::ckm_rsa_pkcs_oaep_wrap(
+          Pem(wrapping_key, wrapping_key_size),
+          {key, key + key_size},
+          {label_buf, label_buf + label_buf_size});
+
+        return JS_NewArrayBufferCopy(
+          ctx, wrapped_key.data(), wrapped_key.size());
+      }
+      else if (algo_name == "AES-KWP")
```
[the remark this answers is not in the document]
wintersteiger: Correct, and we don't need/want `AES-KW` at this time.

src/crypto/key_wrap.cpp
wintersteiger: No, we can't. The algorithm comes with mbedTLS too, but the Open Enclave config disables it. That's not a problem though, we have OpenSSL available now and forever and there are no knock-on effects like having to export/import PEMs to switch between the two.

tests/npm-app/src/endpoints/crypto.ts
```diff
@@ -27,24 +27,47 @@ export function generateAesKey(
...
+    const l = p.label ? b64ToBuf(p.label) : undefined;
+    const new_p = { name: p.name, label: l };
+    const wrappedKey = ccf.ccf.wrapKey(key, wrappingKey, new_p);
+    return { body: wrappedKey };
+  } else if (r.parameters.name == "RSA-OAEP-AES-KWP") {
+    const p = r.parameters as RsaOaepAesParams;
+    const l = p.label ? b64ToBuf(p.label) : undefined;
+    const new_p = { name: p.name, aes_key_size: p.aes_key_size, label: l };
+    const wrappedKey = ccf.ccf.wrapKey(key, wrappingKey, new_p);
+    return { body: wrappedKey };
```
wintersteiger, line 66: @letmaik I hate the base64 decoding going on here, especially for the labels. Is there a way to avoid that or make the infrastructure do it without having to drill into the `WrapAlgoParams`? Maybe just move the base64 decoding into C++ so we don't have to replicate the logic here?

samples/apps/forum/src/types/ccf.ts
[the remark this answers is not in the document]
letmaik: Sorry, I should have been more clear. Can we keep the old name `wrapAlgo` like in the Web Crypto API?

src/apps/js_generic/js_generic.cpp
```diff
@@ -224,26 +226,81 @@ namespace ccfapp
...
 
-    // key can in principle be arbitrary data (see note on maximum size
-    // in rsa_key_pair.h). wrapping_key is a public RSA key.
+        return JS_NewArrayBufferCopy(
+          ctx, wrapped_key.data(), wrapped_key.size());
+      }
+      else if (algo_name == "RSA-OAEP-AES-KWP")
+      {
+        auto aes_key_size_value =
+          auto_free(JS_GetPropertyStr(ctx, parameters, "aes_key_size"));
```
letmaik, line 261: For JS APIs we use camelCase, so this should be `aesKeySize`.

tests/npm-app/src/endpoints/crypto.ts
letmaik: Whether JS apps use Base64 for the label is up to them, we shouldn't prescribe that. I think it's ok as you have it.

samples/apps/forum/src/types/ccf.ts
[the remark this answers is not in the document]
wintersteiger: Oh sure, we can keep the name!

tests/js-modules/modules.py
```diff
@@ -268,11 +268,14 @@ def test_npm_app(network, args):
         )
         label = "label42"
         r = c.post(
-            "/app/wrapKeyRsaOaep",
+            "/app/wrapKey",
             {
                 "key": b64encode(aes_key_to_wrap).decode(),
-                "wrappingKey": wrapping_key_pub_pem,
-                "label": label,
+                "wrappingKey": b64encode(bytes(wrapping_key_pub_pem, "ascii")).decode(),
```
letmaik, line 274: Why do you b64 encode the wrapping key? PEM should be fine as a string.

letmaik: Nevermind, makes sense, to keep it uniform with AES.

wintersteiger: It would be very unusual to put a symmetric (AES) key into a PEM and I suspect most libraries would simply reject it because they don't handle `----- BEGIN AES `...

wintersteiger: Clarification: that's for the AES wrap and I just wanted to have the same format for all three wrapping functions, so I re-encode the PEM too. I guess we could try and look into the wrapAlgo and make it dependent on that.

letmaik: Makes sense, let's leave it as it is. It's just a test anyway.

tests/npm-app/src/types/ccf.ts
```diff
@@ -68,7 +75,7 @@ export interface CCF {
   wrapKey(
     key: ArrayBuffer,
     wrappingKey: ArrayBuffer,
-    wrapAlgo: WrapAlgo
+    parameters: WrapAlgoParams
```
letmaik, line 78: Missed this one.

letmaik (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Governance proposals can be submitted successfully against secondaries (#2247)
 - `set_ca_cert`/`remove_ca_cert` proposals have been renamed `set_ca_cert_bundle`/`remove_ca_cert_bundle` and now also accept a bundle of certificates encoded as concatenated PEM string (#2221).
 
+### Added
+
+- Support for multiple key wrapping algorithms for C++ and JavaScript applications (#2246)
```
achamayou, line 23: Sphinx has a js domain (https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#the-javascript-domain) as we expand our JS API it would be good to make use of it to document what we provide.

achamayou (approved):

eddyashton (approved):

PR #2249, opened 2021-02-28
Track kv dependencies 

ashamis:
Part of #2055

This is part of parallelizing backup execution.

In the PR the KVs dependency tracking extends to track the last read of every transaction and use that information when building the dependency order (i.e. provide boundaries for possible valid serializations of a subsection of the transaction total order).

Note: in the case of a malicious primary the backups will re-create this tracking information to detect linearizability violations.

### DIFF
--- a/src/kv/apply_changes.h
+++ b/src/kv/apply_changes.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include "kv_types.h"
+#include "node/entities.h"
 #include "untyped_map_handle.h"
 
 #include <functional>
@@ -37,13 +38,18 @@ namespace kv
   // sets to their underlying Maps. Calls f() at most once, iff the writes are
   // applied, to retrieve a unique Version for the write set and return the max
   // version which can have a conflict with the transaction.
+  //
+  // The track_conflicts parameter tells the store if it needs to track the last
+  // read version for every key. This is required for backup execution described
+  // at the top of tx.h
 
   static inline std::optional<std::tuple<Version, Version>> apply_changes(
     OrderedChanges& changes,
-    std::function<Version()> f,
+    std::function<std::tuple<Version, Version>(bool)> f,
     kv::ConsensusHookPtrs& hooks,
     const MapCollection& new_maps = {},
-    const std::optional<Version>& new_maps_conflict_version = std::nullopt)
+    const std::optional<Version>& new_maps_conflict_version = std::nullopt,
+    bool track_conflicts = false)
   {
     // All maps with pending writes are locked, transactions are prepared
     // and possibly committed, and then all maps with pending writes are
@@ -51,6 +57,8 @@ namespace kv
     // interleaved fashion.
     Version version = 0;
     bool has_writes = false;
+
+    kv::Version max_conflict_version = kv::NoVersion;
 
     std::map<std::string, std::unique_ptr<AbstractCommitter>> views;
     for (const auto& [map_name, mc] : changes)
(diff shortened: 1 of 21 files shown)


src/kv/apply_changes.h
```diff
@@ -111,38 +124,65 @@ namespace kv
...
+      if (version > max_conflict_version || !track_conflicts)
       {
-        const auto it = views.find(map_name);
-        if (it != views.end() && it->second->has_writes())
+        // Since the tracking of a read version is done in the key-value pair it
+        // is not possible to track the dependencies of two transactions that
+        // depend on a key-value pair on a map that does not exist yet.
+        // Thus, execution is gated on map creation.
+        //
+        // Additionally, if the prepare set max_conflict_version to NoVersion
```
achamayou, line 139:
```suggestion
        // Additionally, if the prepare sets max_conflict_version to NoVersion
```

src/kv/apply_changes.h
```diff
@@ -37,20 +37,26 @@ namespace kv
   // sets to their underlying Maps. Calls f() at most once, iff the writes are
   // applied, to retrieve a unique Version for the write set and return the max
   // version which can have a conflict with the transaction.
+  //
+  // The track_conflicts parameter tells the store if it needs to track the last
```
achamayou, line 41: This description makes me think we should rename `track_conflicts` to `track_read_versions`, because it's more precise, and it helps clarify that this is separate from the flow that sets max_conflict_version on the primary.

I also think the comment here or in tx.h should mention that the reason for this tracking is to protect against a malicious primary setting conflict information incorrectly (ie. why this is required).

src/kv/apply_changes.h
```diff
@@ -37,20 +37,26 @@ namespace kv
...
+  // read version for every key. This is required for backup execution as
+  // described at the top of tx.h
 
   static inline std::optional<std::tuple<Version, Version>> apply_changes(
     OrderedChanges& changes,
-    std::function<Version()> f,
+    std::function<std::tuple<Version, Version>(bool)> f,
```
achamayou, line 47: It's quite hard to tell what f is here, I think a better name and a named type alias would be nice.

src/kv/change_set.h
```diff
@@ -17,8 +17,10 @@ namespace kv
   template <typename K, typename V, typename H>
   using Snapshot = champ::Snapshot<K, VersionV<V>, H>;
 
+  // This is a map of keys and with a tuple of the key's write version and the
+  // version of last transaction which read the key and committed successfully
   template <typename K>
-  using Read = std::map<K, Version>;
+  using Read = std::map<K, std::tuple<Version, Version>>;
```
achamayou, line 23: using LastReadVersion = Version;

.... std::pair<Version, LastReadVersion> ...

would be even better imho, although the comment does describe the two fields.

achamayou:
Unless I am missing something, the SB benchmark in BFT, but also in CFT to some extent, seem to take a noticeable hit?

src/kv/test/kv_test.cpp
```diff
@@ -1463,6 +1463,156 @@ TEST_CASE("Conflict resolution")
...
+      handle->put(std::to_string(info.id), std::to_string(info.id));
+      auto version_resolver = [&](bool) {
+        return std::make_tuple(info.id, kv::NoVersion);
+      };
+      REQUIRE(
+        tx.commit(
+          true, version_resolver, info.replicated_max_conflict_version) ==
+        kv::CommitResult::SUCCESS);
+    }
+  }
```
achamayou, line 1531: Check end state?

src/kv/apply_changes.h
ashamis: prepare()

src/kv/store.h
```diff
@@ -118,6 +122,17 @@ namespace kv
       return false;
     }
 
+    Version next_version_internal()
+    {
+      // Get the next global version. If we would go negative, wrap to 0.
```
eddyashton, line 127:
```suggestion
      // Get the next global version. If version would become negative, wrap to 0.
```

src/kv/store.h
```diff
@@ -31,6 +31,7 @@ namespace kv
 
     SpinLock version_lock;
     Version version = 0;
+    Version last_new_map = -1;
```
eddyashton, line 34:
```suggestion
    Version last_new_map = kv::NoVersion;
```

achamayou (approved):

Issue #2247, opened 2021-02-26
Error: Proposal failed to bind to state


lynshi:
**Describe the bug**
In a 3 node CCF network, when submitting a proposal, the following error is returned.
```
{'error': {'code': 'InternalError', 'message': 'Proposal failed to bind to state.'}}
```

To debug, we sent proposals to each CCF node individually instead of going through the load balancer. The above error is observed on all the secondary nodes. Meanwhile, the primary returns a response like the following:
```
{'proposal_id': '56c1acb4247462a9201b254bdcd84958accf1bb22ac942e26d8a545fa9dffa20', 'proposer_id': 0, 'state': 'OPEN'}
```

The commit was observed to increment on all nodes from 9.22 to 9.24, so it appears that replication from the primary is functional.

Some time later, we observed a change in the primary. Sending the proposal to each node individually again uncovered the same issue; only the (new) primary could "bind state".

**To Reproduce**
Sadly, this appears to be a transient error and we're not able to reproduce consistently; it just happens occasionally after we start up the network. I'm not able to point to a possible cause yet.

For this case in particular, we were actually able to submit proposals before this, because checking `/node/network` shows the service status as `OPEN`. We may have gotten lucky with the load balancing and hit the primary though.

We are submitting the proposals using signature authentication, with the signing occurring in Azure Key Vault.

**Expected behavior**
The proposal can be submitted on the secondary nodes.

**Environment information**
CCF version: 0.18.2
Start node config:
```
consensus = cft
enclave-file = 
enclave-type = release
ledger-chunk-bytes = 104857600
ledger-dir = 
log-format-json = true
node-address = 10.240.0.112:16384
public-rpc-address = 10.240.0.112:16385
read-only-ledger-dir = 
rpc-address = 10.240.0.112:16385
snapshot-dir = 
snapshot-tx-interval = 10000

[start]
gov-script = 
network-cert-file =
member-info = 
```

Joining node config:
```
consensus = cft
enclave-file =
enclave-type = release
ledger-chunk-bytes = 104857600
ledger-dir = 
log-format-json = true
node-address = 10.240.0.9:16384
public-rpc-address = 10.240.0.9:16385
read-only-ledger-dir = 
rpc-address = 10.240.0.9:16385
snapshot-dir =
snapshot-tx-interval = 10000

[join]
network-cert-file = 
target-rpc-address = 10.240.0.112:16385
```

oe_sign.conf:
```
# Enclave settings:
Debug=0
[...]

PR #2252, opened 2021-02-28
Set root on forwarded proposals

achamayou:
See #2247 

- [x] Expand test coverage so that governance operations aren't exclusively executed against primaries


tests/governance.py
```diff
@@ -174,13 +175,13 @@ def test_node_ids(network, args):
 
 @reqs.description("Checking service principals proposals")
 def test_service_principals(network, args):
-    primary, _ = network.find_nodes()
+    node = args.choose_node(network)
```
eddyashton, line 178: Why is this `args.choose_node` rather than `network.find_any_node()`?

achamayou: This is 50/50 primary/backup, and will break if the network does not in fact have backups, so it's a bit better in making sure both cases are equally tested.

achamayou: Moved as discussed.

eddyashton (approved):

PR #2257, opened 2021-03-01
Combine view and seqno into single HTTP header

### LINKED ISSUE #2256: Transaction ID should be returned in a single header
Currently we return 2 headers (view and seqno) - these should be combined into a single header. Other APIs which take these fields should be updated to take a single string.

eddyashton:
Closes #2256.

The format of this header is `{view}.{seqno}`, with no other decoration. Any decoration we do add to aid parsing should be safe to insert as a URL query parameter.

Also creates public header file for this `ccf::TxID`. This should be used more broadly within CCF, replacing a few similar duplicated structs, but I don't want that to pollute this PR.

There are other places in the API (notably `/tx`) where we still talk about view and seqno separately. These could be consolidated into a single field, but they're not directly tied to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- `x-ccf-tx-view` and `x-ccf-tx-seqno` response headers have been removed, and replaced with `x-ms-ccf-transaction-id`. This includes both original fields, separated by a single `.`. Historical queries using `ccf::historical::adapter` should also pass a single combined `x-ms-ccf-transaction-id` header (#2257).
+
 ## [0.18.5]
 
 ### Changed
 
 - Historical query system now supports range queries.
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Historical query system now supports range queries.
+- `x-ccf-tx-view` and `x-ccf-tx-seqno` response headers have been removed, and replaced with `x-ms-ccf-transaction-id`. This includes both original fields, separated by a single `.`. Historical queries using `ccf::historical::adapter` should also pass a single combined `x-ms-ccf-transaction-id` header.
```
achamayou, line 13: That needs to be moved out of 0.18.5

python/ccf/clients.py
```diff
@@ -91,8 +90,14 @@ class Identity:
     description: str
 
 
-def int_or_none(v):
-    return int(v) if v is not None else None
+def parse_tx_id(s: Optional[str]):
+    try:
+        if s is not None:
```
achamayou, line 95: Since AttributeError is caught, this check isn't necessary.
```suggestion
```

achamayou (approved):

python/ccf/clients.py
eddyashton: That's what I thought, but `mypy` can't tell that:

```
python/ccf/clients.py:96: error: Item "None" of "Optional[str]" has no attribute "split"
Found 1 error in 1 file (checked 1 source file)
```

achamayou: Type checking, type schmecking...

Issue #2248, opened 2021-02-27
Receipts change over time


lynshi:
**Describe the bug**
The receipt for a transaction changes when the sequence number increases. Based on the [documentation here](https://microsoft.github.io/CCF/main/use_apps/verify_tx.html#transaction-receipts) and my previous understanding, receipts should be constant over the lifetime of the ledger.

**To Reproduce**
Given a 1-node CCF network,
1. Get the receipt for an existing transaction (`GET /app/receipt?commit=42`).
2. Verify the receipt (`POST /app/receipt/verify`). This passes.
3. Create a transaction that increments the sequence number (i.e. write to a KV store). 
4. Verify the previous receipt. Now, the response says `valid: False`.
5. Get a receipt for the original transaction again. The receipt is now different.

**Expected behavior**
The original receipt should always be valid, and subsequent retrievals should return the same receipt.

**Environment information**
CCF 0.18.1

achamayou:
These problems are known issues of the current implementation, which only issues receipts based on the current tree. Instead, a historical tree snapshot must be used. I will address this as soon as #2233 is merged.

Separately from that, it is worth noting that there may be as many valid receipts for a transaction as there are signatures in the ledger that follow the transaction in question. The one we aim to consistently produce is the shortest one, ie. the one rooted in the very next signature, but the others are equally valid, if longer.

In some situations, for example after a catastrophic recovery (and therefore a network identity change), a user may explicitly want a longer receipt, rooted in the latest identity.

lynshi:
Got it, so returned receipts can be expected to change over time as the Merkle tree is modified, since they're issued off the latest state. However, the known issue is that previously issued receipts should still be valid because they can be verified over the historical tree.

Given a receipt, is it possible to tell from which historical tree it was issued? Or will we also need to provide other metadata so CCF can figure out which historical tree to load?

achamayou:
There is only one tree, it just gets compacted gradually to avoid taking all the memory. It's possible to load older views using historical queries however. We will default to returning receipts rooted in the next signature, which are stable in time.

Note that verifying receipts against the service, as opposed to independently offline, is not a particularly meaningful operation. If you trust the service, then asking for a tx status provides the same information. If you do not, then the value you can attribute to its answer is very unclear.

wintersteiger:
[...]

PR #2259, opened 2021-03-01
Enable support for historical receipt verification

wintersteiger:


src/node/history.h
```diff
@@ -231,8 +231,15 @@ namespace ccf
 
     bool verify(HistoryTree* tree) const
     {
-      return tree->max_index() == path->max_index() && tree->root() == root &&
-        path->verify(root);
+      if (path->max_index() > tree->max_index())
```
achamayou, line 234: We try to use {} everywhere, although we haven't found a good way to enforce it so far.

achamayou (approved):

src/node/history.h
wintersteiger: Not sure what you're referring to. Do you mean you'd like it to work for empty or all-zero `this->root`?

src/node/history.h
```diff
@@ -231,8 +231,15 @@ namespace ccf
...
+        return false;
+      else if (tree->max_index() == path->max_index())
+        return tree->root() == root && path->verify(root);
```
eddyashton, line 237:
```suggestion
      if (path->max_index() > tree->max_index())
      {
        return false;
      }
      else if (tree->max_index() == path->max_index())
      {
        return tree->root() == root && path->verify(root);
      }
```
This is what @achamayou was referring to - we try to avoid unscoped single-line blocks.

eddyashton (approved):
Looks great, thanks @wintersteiger.

This is part of the issue noted #2248 (namely point 4, with the previously valid receipt now being marked as invalid).

Not critical but would be nice: we have an endpoint `/receipt/verify` for online receipt verification, which is tested in `test_receipts` is `e2e_logging.py`. I'm surprised this wasn't already testing this, but we actually never call the endpoint again after a write. Would be good to cover this behaviour there, similarly to what we do in the unit test (save `r.body.json()["receipt"]`, submit more `/app/log/private` commands, call `/receipt/verify` again).

src/node/history.h
achamayou: Er no, sorry, what Eddy said about if/else blocks :)

wintersteiger:
Oh the branch braces! Yeah, the auto-formatter doesn't do that and my brain likes it better without, so some of them slip through :-)

Sure, I'll take a look at the JS API and the additional tests.

PR #2260, opened 2021-03-01
Add convenience functions and docstrings to crypto API.

wintersteiger:
Also contains some general clean-up around related code.


cmake/sss.cmake
```diff
@@ -7,8 +7,9 @@ set(SSS_PREFIX
 )
 message(STATUS "Using sss at ${SSS_PREFIX}")
 
-set(SSS_SRC ${SSS_PREFIX}/sss.c ${SSS_PREFIX}/hazmat.c
-            ${SSS_PREFIX}/tweetnacl.c
+set(SSS_SRC
+    ${SSS_PREFIX}/sss.c ${SSS_PREFIX}/hazmat.c ${SSS_PREFIX}/tweetnacl.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/node/test/sss_random.cpp
```
achamayou, line 12: That does not seem right, we do not want test code in the library.

src/crypto/symmetric_key.cpp
```diff
@@ -19,4 +19,44 @@ namespace crypto
...
 #endif
   }
+
+  std::vector<uint8_t> aes_gcm_encrypt(
+    const std::vector<uint8_t>& key,
+    std::vector<uint8_t>& plaintext,
+    const std::vector<uint8_t>& iv,
+    const std::vector<uint8_t>& aad)
+  {
+    if (key.size() != 16 && key.size() != 24 && key.size() != 32)
```
achamayou, line 29: Can we move this out to a single checking call? I'm uncomfortable with having literals like that in multiple places.

src/node/test/sss_random.cpp
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
achamayou, line 1: Is this only used for testing purposes?

cmake/sss.cmake
wintersteiger: I just moved it there because it shouldn't be part of ccfcrypto and I didn't know where else to put it. libsss expects a function with the prototype in this file (which used to be in a header). @jumaffre I think you last touched these bits - do you know where they should rather go?

src/crypto/symmetric_key.cpp
wintersteiger: Sure, np.

src/node/test/sss_random.cpp
wintersteiger: libsss is linked into the enclave image, but I don't know whether it's actually used.

jumaffre, line 1: Is this still required?

wintersteiger: No, gone now. Just wasn't done with my changes yet (more docs coming).

achamayou:
@wintersteiger this looks good to me, is it ready for review?

wintersteiger:
I think what's there is ready for review, but I'm planning to add more to it.

achamayou:
@wintersteiger in that case it's best to update the branch and change the status from Draft to PR.

achamayou (approved):

eddyashton (approved):

PR #2262, opened 2021-03-01
Install public headers

eddyashton:
Missing CMake install that broke the attempted 0.18.5 release.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,6 +183,14 @@ install(
   PATTERN "*/test*" EXCLUDE
 )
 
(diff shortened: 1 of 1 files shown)


achamayou (approved):

achamayou:
fwiw, tested locally and does seem to work:

```
$ ls /opt/ccf-0.18.4-4-g45ee19982/include/ccf/
apps     consensus  ds       historical_queries_interface.h  http  libmerklecpp  node         runtime_config
clients  crypto     enclave  host                            kv    lua_interp    perf_client  tls
```

PR #2265, opened 2021-03-02
Add regexp filter for E2E test suite

wintersteiger:
A simple filter to select which tests of the suite to run.

### DIFF
--- a/tests/e2e_suite.py
+++ b/tests/e2e_suite.py
@@ -12,6 +12,7 @@ import sys
 from enum import Enum
 import random
 import os
+import re
 
 from loguru import logger as LOG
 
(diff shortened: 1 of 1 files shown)


tests/e2e_suite.py
```diff
@@ -72,7 +73,17 @@ def run(args):
...
     elapsed = args.test_duration
 
-    for i, test in enumerate(chosen_suite):
+    if args.filter is not None:
+        filter_re = re.compile(args.filter)
+
+        def filter_fun(x):
+            return filter_re is None or filter_re.match(x[1].__name__)
+
+        tests_to_run = filter(filter_fun, enumerate(chosen_suite))
```
achamayou, line 82:
```
tests_to_run = filter(lambda x: re.compile(args.filter or '.*').match(x[1].__name__), enumerate(chosen_suite))
```
was quite tempting, but this is more readable :)

wintersteiger: Guess what I had before making it readable :)

achamayou (approved):

PR #2242, opened 2021-02-26
JSON receipts

letmaik:
Partially implements https://github.com/microsoft/CCF/issues/1891.

What's missing is the signature over the root, however that wasn't part of the old receipts either, so it can be a follow-up PR.

The new receipt format is as below:

```json
{
  "seqno": 14,
  "root": "FCE9A4CBA103EB9996C5B6D20D9B9647471F87F99F8B3F7B0C1975C6EBA68012",
  "path": [
    {
      "right": "E56341B3BB1E6106856389D8DA661289C2620477DFEBFF184B0624FCA83B730E"
    },
    {
      "left": "230D59672A8C64F171FACFA0B7EDF355A2AB311FFD9F64CA59CB16153351C5E1"
    },
    {
      "left": "3911417095DACAEC9FBBBEEFC210EEE30753D1600FBD5F73CC30ADDC5DB0167D"
    },
    {
      "left": "E3589D648A6E0DF70BDA3CBD25430E1D19BF0690097ADE339A8F5E8E4340DEB7"
    },
    {
      "right": "5C0643BD736C70982F659B69ECB46F227B8B0D913F4AD01BF834E03339E3B0DB"
    }
  ],
  "leaf": "44EA24B4BA7D61CFFFD31A996902DA24F50A92D4B80BC4D09F789B518B2B0B7E"
}
```

PR #2270, opened 2021-03-02
Historical receipts

achamayou:
Still very much a draft, but enables historical queries to fetch receipts associated with the indices they cover. This is useful on its own for historical receipts, but also for custom claims. Next items:

- [x] Break down path to match #2242
- [x] Verify merkle proof in test
- [x] Add node id to receipt to allow cert lookup
- [x] Verify signature in test
- [x] Replace `get_receipt`, remove `verify_receipt`
- [x] Clean up and document offline receipt verification API/steps
- [x] Clean up and document `get_receipt` public API
- [x] Changelog entry

To be done in later PRs:

- Make receipt generation optional
- Find a sensible way to support ranges

```
75: 18:09:50.012 | INFO     | __main__:test_receipts:993 - [0|user0|] GET /app/receipt?transaction_id=2.365
75: 18:09:50.012 | INFO     | __main__:test_receipts:993 - 200 {"leaf":"fdc977c49d3a8bdf986176984e9432a09b5f6fe0c04e0b1c2dd177c03fdca9ec","node_id":"682c161e1bc0aec694cac58a6ea456e1caa6c9c56d + 715 chars
75: {'leaf': 'fdc977c49d3a8bdf986176984e9432a09b5f6fe0c04e0b1c2dd177c03fdca9ec',
75:  'node_id': '682c161e1bc0aec694cac58a6ea456e1caa6c9c56d8dd873da9455c341947065',
75:  'proof': [{'left': 'f847e5efe3965b0dacb5c15c666602807a11fdecd465d0976779eed27121ffa3'},
75:            {'left': 'a56ce9efb73957f561f12d60513281fd2aaf16440234e2fd56e7d3d2ff4be8b0'},
75:            {'left': 'd91c982f525302244b13b6add92cd0925e1e0fb621ff2a7bb408ecc51be8528e'},
75:            {'left': '6d87faceda763ce65914f95dfcc04b37ea3f26bc552764752a0f2720039f76be'},
75:            {'left': 'e0cc83ea2fae6c535fc44605fb25ba9fdfb319e0e577b3541760f9a3565c549b'},
75:            {'left': 'f0e95ed85f5f6c0197aed4f6685b93dc56edd823a2532bd717558a5ab77267cb'}],
75:  'root': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
75:  'signature': 'MGQCMACklXqd0ge+gBS8WzewrwtwzRzSKy+bfrLZVx0YHmQvtsqs7dExYESsqrUrB8ZcKwIwS3NPKaGq0w2QlPlCqUC3vQoQvhcZgPHPu2GkFYa7JEOdSKLknNPHaCRv80zx2RGF'}
```


tests/e2e_logging.py
```diff
@@ -434,6 +434,28 @@ def test_historical_query(network, args):
...
+def test_historical_receipts(network, args):
+    if args.consensus == "bft":
+        LOG.warning("Skipping historical queries in BFT")
+        return network
+
+    if args.package == "liblogging":
+        node = network.find_node_by_role()
+        network.txs.issue(network, number_txs=1)
+        first_msg = network.txs.priv[1][0]
+        first_receipt = network.txs.get_receipt(
```
eddyashton, line 448: Can we add a call to `/receipt/verify` to confirm this is a valid receipt?

achamayou: Won't work all that well until that supports historical queries as well, unless it's still in cache. I'd rather do the offline validation in the test code first, which I think is much more useful and is a better test as well.

achamayou: Done

tests/e2e_logging.py
```diff
@@ -434,6 +439,41 @@ def test_historical_query(network, args):
     return network
 
 
+@reqs.description("Read historical receipts")
+@reqs.supports_methods("log/private", "log/private/historical_receipt")
+def test_historical_receipts(network, args):
```
jumaffre, line 444: Could you add this to the test suite as well?

python/ccf/receipt.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def root(leaf: str, proof: List[dict]):
+    current = bytes.fromhex(leaf)
+    for n in proof:
+        if "left" in n:
+            current = sha256(bytes.fromhex(n["left"]) + current).digest()
+        else:
+            current = sha256(current + bytes.fromhex(n["right"])).digest()
+    return current.hex().upper()
```
jumaffre, line 15: Why `.upper()`? This seems inconsistent with the way we already report hex-encoded strings in CCF.

src/node/historical_queries_adapter.h
```diff
@@ -17,9 +17,131 @@ namespace ccf::historical
...
+    AbstractStateCache& state_cache,
+    const CheckAvailability& available)
+  {
+    return [f, &state_cache, available](EndpointContext& args) {
+      // Extract the requested transaction ID
+      kv::Consensus::View target_view;
+      kv::SeqNo target_seqno;
+
+      {
+        const auto target_view_opt =
```
jumaffre, line 41: Most of these checks are repeated in the other historical adapter. Could we unify those?

python/ccf/receipt.py
achamayou: Because the merklecpp built-in does upper, but I'll make that configurable to match the rest of CCF, or even switch the default there.
@wintersteiger do you feel strongly about hex string casing? Do you mind if we change the merklecpp default to match ccf?

tests/e2e_logging.py
achamayou: Yes!

src/node/historical_queries_adapter.h
achamayou: Hopefully, I started out by templatizing it, but I got into a bit an error rabbit hole, so I just forked it. I'll have another look now that everything is working and I'm cleaning up.

python/ccf/receipt.py
wintersteiger: No, I don't care about the casing at all! Happy to change it to lower-case if you prefer that. The Hash constructor from hex string should accept both as it is. I think I'll add a parameter with default lower-case; might be helpful for other applications.

src/node/historical_queries_adapter.h
achamayou: Discussed with @eddyashton, merged with a single StatePtr interface.

tests/e2e_logging.py
```diff
@@ -919,58 +968,42 @@ def test_ws(network, args):
 
 
 @reqs.description("Running transactions against logging app")
-@reqs.supports_methods("receipt", "receipt/verify", "log/private")
+@reqs.supports_methods("receipt", "log/private")
 @reqs.at_least_n_nodes(2)
 def test_receipts(network, args):
     primary, _ = network.find_primary_and_any_backup()
+    cert = os.path.join(primary.common_dir, f"{primary.local_node_id}.pem")
+    with open(cert) as c:
```
letmaik, line 976:
```suggestion
    cert_path = os.path.join(primary.common_dir, f"{primary.local_node_id}.pem")
    with open(cert_path) as c:
```

letmaik (approved):

src/node/rpc/common_endpoint_registry.h
```diff
@@ -209,63 +248,61 @@ namespace ccf
...
+
+      auto get_receipt = [](
+                           ccf::EndpointContext& args,
+                           ccf::historical::StatePtr historical_state,
+                           kv::Consensus::View,
+                           kv::Consensus::SeqNo) {
+        const auto [pack, params] =
+          ccf::jsonhandler::get_json_params(args.rpc_ctx);
+
+        // const auto in = params.get<GetReceipt::In>();
```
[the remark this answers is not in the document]
achamayou: It's effectively unused, the adapter has to parse the query string instead. I don't know that there is a good way to turn that into a type that's auto-documented, the current query parsing scheme to JSON isn't all that good to begin with. I would like to have that in OpenAPI though, I hope I can just add it manually.

samples/apps/logging/logging_schema.h
```diff
@@ -37,13 +37,37 @@ namespace loggingapp
...
+    {
+      size_t id;
+    };
+
+    struct Out
+    {
+      std::string msg;
+      std::string signature;
+      std::string root;
+      std::vector<nlohmann::json> proof;
```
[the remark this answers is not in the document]
achamayou: Hadn't thought of that, it's a good point. Perhaps there's even a way to get OpenAPI to say "oneOf".

src/node/rpc/common_endpoint_registry.h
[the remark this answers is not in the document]
letmaik: Sorry, I commented on the wrong line, I meant the following lines. So, why use untyped nlohmann::json here for the generic receipt endpoint?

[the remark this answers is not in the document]
achamayou: Because I didn't think there was a good way to do it with the macros, but as you pointed out, there is, so I've done it now.

include/ccf/historical_queries_interface.h
```diff
@@ -4,15 +4,77 @@
...
+      std::shared_ptr<ccf::HistoryTree::Path> p_,
+      const NodeId& n_) :
+      signature(s_),
+      root(r_),
+      path(p_),
+      node_id(n_)
+    {}
+  };
+
+  inline void to_json(nlohmann::json& j, const TxReceipt& r)
```
letmaik, line 34: Do we still need this?

achamayou: Nope

src/node/rpc/call_types.h
```diff
@@ -139,12 +139,43 @@ namespace ccf
...
+      std::optional<std::string> left = std::nullopt;
+      std::optional<std::string> right = std::nullopt;
     };
 
     struct Out
     {
-      std::vector<std::uint8_t> receipt = {};
+      std::string signature;
+      std::string root;
+      std::vector<Node> proof = {};
```
eddyashton, line 155: Think this type should be something other than `Node`, since its nothing to do with the `ccf::NodeId node_id` just below. Maybe `MerkleStep` or `ProofElement` or `Link` or similar?

src/node/rpc/common_endpoint_registry.h
```diff
@@ -3,13 +3,52 @@
...
+  {
+    const std::string prefix("transaction_id=");
+    const auto& path_params = args.rpc_ctx->get_request_query();
+
+    if (!nonstd::starts_with(path_params, prefix))
+    {
+      args.rpc_ctx->set_error(
+        HTTP_STATUS_BAD_REQUEST,
+        ccf::errors::InvalidInput,
+        "Query string must contain a single \'transaction_id\' parameter");
```
eddyashton, line 27: Probably a stupid question: Why do we need the escapes on these single quotes?

src/node/history.h
```diff
@@ -7,6 +7,7 @@
 #include "ds/dl_list.h"
 #include "ds/logger.h"
 #include "ds/thread_messaging.h"
+#include "endian.h"
```
eddyashton, line 10: Unused?

samples/apps/logging/logging.cpp
```diff
@@ -7,6 +7,7 @@
 #include "node/quote.h"
 #include "node/rpc/metrics_tracker.h"
 #include "node/rpc/user_frontend.h"
+#include "tls/base64.h"
```
eddyashton, line 10: Unused, as far as I can tell?

include/ccf/historical_queries_interface.h
```diff
@@ -4,15 +4,54 @@
 
 #include "consensus/ledger_enclave_types.h"
 #include "kv/store.h"
-#include "node/rpc/endpoint_registry.h"
+#include "node/history.h"
+#include "tls/base64.h"
```
eddyashton, line 8: Also unused, as far as I can tell?

tests/e2e_logging.py
```diff
@@ -434,6 +439,50 @@ def test_historical_query(network, args):
...
 
+@reqs.description("Read historical receipts")
+@reqs.supports_methods("log/private", "log/private/historical_receipt")
+def test_historical_receipts(network, args):
+    if args.consensus == "bft":
+        LOG.warning("Skipping historical queries in BFT")
+        return network
+
+    if args.package == "liblogging":
+        node, _ = network.find_primary()
```
jumaffre, line 450: Could we test this on all nodes in the network?

src/node/rpc/serialization.h
```diff
@@ -111,9 +111,17 @@ namespace ccf
...
   DECLARE_JSON_TYPE(GetReceipt::In)
-  DECLARE_JSON_REQUIRED_FIELDS(GetReceipt::In, commit)
+  DECLARE_JSON_REQUIRED_FIELDS(GetReceipt::In, transaction_id)
+  DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(GetReceipt::Node)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+#pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
+  DECLARE_JSON_REQUIRED_FIELDS(GetReceipt::Node);
+#pragma clang diagnostic pop
+  DECLARE_JSON_OPTIONAL_FIELDS(GetReceipt::Node, left, right)
```
eddyashton, line 121: More of a comment: I agree this should end up as a `oneOf` in OpenAPI. I'm not sure how we'd represent this in the macro, and adding support is not worth the effort at the moment. I think this is a great example of a type that's complicated enough it should have custom `to_json`+schema builder functions, but this is close enough for now.

tests/e2e_logging.py
```diff
@@ -434,6 +439,50 @@ def test_historical_query(network, args):
...
+        with open(cert_path) as c:
+            node_cert = load_pem_x509_certificate(
+                c.read().encode("ascii"), default_backend()
+            )
+
+        TXS_COUNT = 5
+        network.txs.issue(network, number_txs=5)
+        for idx in range(1, TXS_COUNT + 1):
+            first_msg = network.txs.priv[idx][0]
+            first_receipt = network.txs.get_receipt(
```
jumaffre, line 461: As the receipt verification has been added to the `LoggingTxs` object, could we also verify the receipts for all the transactions when calling `network.txs.verify()` (which is called at the end of each end-to-end test run)? If it takes too long, perhaps just on the full test suite?

tests/e2e_logging.py
```diff
@@ -919,58 +968,42 @@ def test_ws(network, args):
...
+                            receipt["leaf"], receipt["proof"]
+                        )
+                        ccf.receipt.verify(
+                            receipt["root"], receipt["signature"], node_cert
+                        )
+                        break
+                    elif rc.status_code == http.HTTPStatus.ACCEPTED:
+                        time.sleep(0.5)
+                    else:
+                        assert False, rc
```
eddyashton, line 1006: Can you please split this over 2 loops - one which submits the transactions (bonus points if it waits for commit half-way), and then another which retrieves and verifies receipts? With this single loop, we're always fetching the receipt for the 'most recent' tx at that point, and we may sleep long enough at it gets its own signature. To make sure we're not benefitting from any special cases, we should test receipts for 'old' transactions, ideally from multiple signatures.

eddyashton (approved):
Minor comments on unused includes and naming, but I'm happy with the changes.

tests/e2e_logging.py
eddyashton: Of course this is `test_receipts` and not `test_historical_receipts` so my points are basically moot - this is already tested elsewhere. Ignore this.

tests/e2e_logging.py
eddyashton: RE: "If it takes too long" - I recently added sampling logging to `LoggingTxs.verify`, so we only verify a small sample of historical transactions rather than every one individually. So this shouldn't take too long, though we may want to extend that sampling to "only some of the keys" as well as the current "only some of the historical entries (in each key)".

src/node/rpc/call_types.h
```diff
@@ -139,12 +139,43 @@ namespace ccf
...
+          Node n;
+          if (node.direction == ccf::HistoryTree::Path::Direction::PATH_LEFT)
+          {
+            n.left = node.hash.to_string();
+          }
+          else
+          {
+            n.right = node.hash.to_string();
+          }
+          proof.push_back(std::move(n));
```
jumaffre, line 174: emplace_back?

jumaffre (approved):

src/node/rpc/call_types.h
achamayou: It's a node in the MerkleTree though and a GetReceipt::Node, not a ccf::Node. If we had an accurate alternative I would agree, but I don't feel like the other terms proposed are more precise.

eddyashton: `MerkleNode`?

wintersteiger: Path step, proof step, path element, tree node. Between two nodes there's also a "branch"?

achamayou: I was quite tempted by `NeighbourNode` but there's the whole British/American thing, so maybe `Element`. It still feels like a failure to namespace more than a legitimate change.

tests/e2e_logging.py
achamayou: Yes

tests/e2e_logging.py
achamayou: Will have a look, but I do not like the magic verification thing that's called outside testcases. I think only testcases should have assertions and infrastructure should remain neutral.

src/node/history.h
achamayou: Surprisingly used by the base64 code, will check if that's still the case.

samples/apps/logging/logging.cpp
achamayou: Probably now yes, will check.

include/ccf/historical_queries_interface.h
achamayou: Maybe, see above.

src/node/rpc/serialization.h
achamayou: Agreed.

src/node/rpc/common_endpoint_registry.h
achamayou: We don't.

src/node/history.h
achamayou: Nope, but used by merklecpp:

```
In file included from ../src/node/history.h:25:
../src/libmerklecpp/./merklecpp.h:64:22: error: use of undeclared identifier 'htobe64'
    size_t size_be = htobe64(size);
                     ^
../src/libmerklecpp/./merklecpp.h:77:12: error: use of undeclared identifier 'be64toh'
    return be64toh(result);
           ^
../src/libmerklecpp/./merklecpp.h:1529:16: error: use of undeclared identifier 'be32toh'
      cws[i] = be32toh(((int32_t *)block)[i]);
               ^
../src/libmerklecpp/./merklecpp.h:1561:35: error: use of undeclared identifier 'htobe32'
      ((uint32_t*)out.bytes)[i] = htobe32(s[i] + h[i]);
                                  ^
../src/libmerklecpp/./merklecpp.h:1580:35: error: use of undeclared identifier 'htobe32'
      ((uint32_t*)out.bytes)[i] = htobe32(((uint32_t*)ctx.h)[i]);
                                  ^
../src/libmerklecpp/./merklecpp.h:1611:35: error: use of undeclared identifier 'htobe32'
      ((uint32_t*)out.bytes)[i] = htobe32(ctx.state[i]);
                                  ^
6 errors generated.
```

wintersteiger: Interesting! merklecpp uses those legitimately, but I didn't need that #include so far. I'll look for a more portable solution tomorrow.

CHANGELOG.md
```diff
@@ -5,12 +5,13 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.19.0]
 
 ### Changed
 
 - `x-ccf-tx-view` and `x-ccf-tx-seqno` response headers have been removed, and replaced with `x-ms-ccf-transaction-id`. This includes both original fields, separated by a single `.`. Historical queries using `ccf::historical::adapter` should also pass a single combined `x-ms-ccf-transaction-id` header (#2257).
 - Node unique identifier is the hex-encoded string of the SHA-256 digest of the node's identity public key, which is also used as the node's quote report data (#2241).
+- The receipt interface has changed, `/app/receipt?commit=23` is replaced by `/app/receipt?transaction_id=2.23`. Receipt fetching is now implemented as a historical query, which means that the first reponse(s) may be a 202 with a Retry-After header. Receipts are now structured JSON, and `/app/receipt/verify` has been removed in favour of an [offline verification sample]().
```
eddyashton, line 14: There's still a TODO at the end of this line, and we should also say:
- `get_receipt_for_seqno_v1` has been removed and should now be implemented through the historical query system
- Constructor for `UserFrontend`/`XXXEndpointRegistry` etc have changed, and should now be passed `AbstractNodeContext`

achamayou: Fixed now

tests/e2e_logging.py
achamayou: Not going to do this now, this requires keeping track of elections, in another PR maybe.

PR #2277, opened 2021-03-08
Fix version of Python in ci-checks.sh

wintersteiger:
The Python linter throws errors if run with, e.g., Python 3.6 instead of 3.8. This change makes sure we always use 3.8.

### DIFF
--- a/scripts/ci-checks.sh
+++ b/scripts/ci-checks.sh
@@ -12,6 +12,8 @@ fi
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
+PYTHON=python3.8
+
 echo "Shell scripts"
 git ls-files | grep -e '\.sh$' | grep -E -v "^3rdparty" | xargs shellcheck -s bash -e SC2044,SC2002,SC1091,SC2181
 
@@ -38,7 +40,7 @@ echo "OpenAPI"
 find doc/schemas/*.json -exec npx swagger-cli validate {} \;
 
 echo "Copyright notice headers"
-python3.8 "$SCRIPT_DIR"/notice-check.py
+$PYTHON "$SCRIPT_DIR"/notice-check.py
 
 echo "CMake format"
 if [ $FIX -ne 0 ]; then
@@ -50,7 +52,7 @@ fi
 # Virtual Environment w/ dependencies for Python steps
 if [ ! -f "scripts/env/bin/activate" ]
     then
-        python3.8 -m venv scripts/env
+        $PYTHON -m venv scripts/env
 fi
 
 source scripts/env/bin/activate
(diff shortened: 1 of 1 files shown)


scripts/ci-checks.sh
```diff
@@ -50,7 +52,7 @@ fi
 # Virtual Environment w/ dependencies for Python steps
 if [ ! -f "scripts/env/bin/activate" ]
     then
-        python3.8 -m venv scripts/env
+        $PYTHON -m venv scripts/env
 fi
 
 source scripts/env/bin/activate
```
letmaik, line 58: After this line, `python` will refer to the one from the venv, so I don't really understand why the fix below would make a difference.

wintersteiger: That's a good point. It does so anyways; mine used Python 3.6 which doesn't have the `dirs_exist_ok` argument on `shutil.copytree`.

wintersteiger: And it still does so if I change line 73 back to using `python`: `191,30,error,unexpected-keyword-arg:Unexpected keyword argument 'dirs_exist_ok' in function call`

letmaik: When you run `source scripts/env/bin/activate` manually, what do the following print?
```
python -c "import sys;print(sys.executable)"
python3.8 -c "import sys;print(sys.executable)"
```

wintersteiger:
```
...$ . scripts/env/bin/activate
(env) ...$ python -c "import sys;print(sys.executable)"
/usr/bin/python
(env) ...$ python3.8 -c "import sys;print(sys.executable)"
/usr/bin/python3.8
(env) ...$ ls -la /usr/bin/python
lrwxrwxrwx 1 root root 24 Mar  8 12:36 /usr/bin/python -> /etc/alternatives/python
(env) ...$ ls -la /etc/alternatives/python
lrwxrwxrwx 1 root root 18 Mar  8 12:36 /etc/alternatives/python -> /usr/bin/python3.6
```

letmaik: Weird, something is very broken on your system, let's figure this out "offline".

wintersteiger: It's the venv; for some reason I had a root-owned `scripts/env` and never got any error messages relating to that. Now that I tried to manually recreate it, I got the permission error and I was able to create a new venv that works as expected. 

But yes, my system is broken, because I tried to get rid of Python 2.x while too much stuff still depends on it.

Still a good idea to have a `$PYTHON` I think.

letmaik: Glad you figured it out. `$PYTHON` makes only sense before venv activation, otherwise you would also have to patch up related calls to `pip` or Python entrypoints like `mypy`. I think whenever we update to a newer Python version we search-replace all instances of `pythonx.y`, so it wouldn't really be less maintenance I think.

wintersteiger: Yeah, that's right, all the `pip` references would have to change too. Is `python` inside of a venv guaranteed (or at least intended) to be the same version of python that created the venv?

letmaik: Yes, that's guaranteed.

PR #2278, opened 2021-03-08
Fix SAN error from recent node ID changes and BFT

jumaffre:
The recent node ID changes (#2241) introduced a memory misalignment with BFT. This PR fixes the issue by padding the monotic node ID in this case. Note that this is a quick fix and that the BFT node ID scheme will eventually be unified with CFT (see #1852).

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-A daily, just for good measure.
+Just want to be making daily records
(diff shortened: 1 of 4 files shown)


ashamis (approved):

jumaffre:
Merging this as it contains the fix for the failed Daily build.

PR #2279, opened 2021-03-08
Use hash of certificate for user and member IDs

### LINKED ISSUE #1378: Re-evaluate the use of KV-driven incrementing IDs for users, members, nodes and proposals

jumaffre:
Fixes #1378 (part II)

We now use a hex-encoded string of the SHA-256 hash of the DER-encoded certificate (a.k.a [fingerprint](https://ldapwiki.com/wiki/Certificate%20Fingerprint#:~:text=The%20Certificate%20Fingerprint%20is%20the,entire%20certificate%20in%20DER%20form.)) for the member and user IDs. 

This has the following implication: the user and member certs and digests KV tables are gone and the `keyId` in signed HTTP requests is set to the Member/User ID. 

The Python infra still tracks members/users as `0`, `1`, etc. (`local_id`) but also knows their CCF IDs (`service_id`).


src/node/client_signatures.h
```diff
@@ -54,6 +54,4 @@ namespace ccf
   };
   DECLARE_JSON_TYPE(SignedReq)
   DECLARE_JSON_REQUIRED_FIELDS(SignedReq, sig, req, request_body, md, key_id)
-  // this maps client-id to latest SignedReq
-  using ClientSignatures = kv::Map<CallerId, SignedReq>;
```
jumaffre, line 58: This was unused.

doc/schemas/gov_openapi.json
```diff
@@ -66,6 +66,28 @@
...
+                "$ref": "#/components/schemas/Script"
+              }
+            ]
+          },
+          "maxItems": 2,
+          "minItems": 2,
+          "type": "array"
+        },
+        "type": "array"
+      },
```
eddyashton, line 90: I think it should be possible to convert this `std::map` to an object (since the keys are strings), but since we wrap them in another type `nlohmann::json` will produce an array of pairs, which at least we correctly describe here.

eddyashton: This is another "more of a comment than a suggestion", don't think there's any way around this with our current JSON abstractions.

src/node/certs.h
```diff
@@ -1,12 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the Apache 2.0 License.
 #pragma once
+
 #include "entities.h"
 
 namespace ccf
 {
-  using CertDERs = kv::Map<Cert, ObjectId>;
+  using CertDERs = kv::Map<Cert, CallerId>;
```
eddyashton, line 9: I think this table can also be removed, since we can also recompute this `Cert -> CallerId` when needed?

jumaffre: Ah! Very good point.

tests/e2e_logging.py
```diff
@@ -241,15 +244,15 @@ def require_new_response(r):
                 require_new_response(r)
 
             LOG.info("Authenticate as a member, via TLS cert")
-            with primary.client("member0") as c:
+            with primary.client(f"member{member.local_id}") as c:
```
eddyashton, line 247: Since this construction is so common, should we instead just store this full string as `local_id`? ie - if we never use the `local_id` of `0` in isolation, its only ever to construct `"member0"` or `"user0"`, then we may as well store that string directly (with the bonus that when we've generated a key for `"alice"`, this works, and we don't need to copy her keys to `useralice.pem` and `memberalice.pem`).

tests/start_network.py
```diff
@@ -55,6 +55,7 @@ def run(args):
             network.start_and_join(args)
 
         primary, backups = network.find_nodes()
+        max_len = len(str(len(backups)))
```
eddyashton, line 58:
```suggestion
        max_len = max(len(str(backup.local_node_id)) for backup in backups)
```

Else this breaks when we have 3 nodes left, but they're 9, 10, and 11 (or 1, 100, and 100000000).

src/node/certs.h
jumaffre: Done now. Let's see if hashing the DER certificate on every request has any impact on performance.

src/node/node_id.h
```diff
@@ -105,32 +58,25 @@ namespace ccf
...
     }
   };
-
 }
 
-// Only display the first node_id_truncation_max_char_count character when
-// printing the node id to the node's log
+// Node ids are printed in many places (e.g. consensus) so only display the
+// first node_id_truncation_max_char_count characters when printing it to the
+// node's log (e.g. using the LOG_..._FMT() macros)
```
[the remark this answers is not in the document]
jumaffre: Good point. My first though was that we don't print the member/user IDs very often so the full hash would be OK then. But I'd rather go with a unified way of dealing with/printing entities in the logs, so I'll merge the two.

doc/schemas/gov_openapi.json
jumaffre: Good spot. I didn't pay attention to this as we already had something similar when `MemberId`s were `uint64_t` (see `uint64_to_Script` on left-hand side of diff). Do you think there's a quick fix for this?

eddyashton: I don't think there's a quick fix, it should be possible with some explicit template instantiations but not worth investigating this now. Can you just check that this array-of-pairs is actually what's produced when we convert one of these maps to JSON? I notice that the nlohmann docs say any type "whose keys can construct an std::string" will convert to a JSON object, so its possible that your `operator std::string` on `EntityID` actually produces an object in the serialisation (in which case this auto-generated schema is wrong...).

src/node/entity_id.h
```diff
@@ -0,0 +1,167 @@
...
+
+    auto& value() const
+    {
+      return id;
+    }
+
+    auto data() const
+    {
+      return id.data();
+    }
```
eddyashton, line 69:
```suggestion
    Value& value()
    {
      return id;
    }

    const Value& value() const
    {
      return id;
    }

    char const* data() const
    {
      return id.data();
    }
```

It wasn't obvious to me that `auto& foo() const` was returning a `const Foo&` (rather than the dangerously broken `Foo&`-from-a-const-method), so I think we should spell this ourselves in the return types. And if we intend to use these for all accesses, `id` should be `protected`.

CHANGELOG.md
```diff
@@ -10,7 +10,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - `x-ccf-tx-view` and `x-ccf-tx-seqno` response headers have been removed, and replaced with `x-ms-ccf-transaction-id`. This includes both original fields, separated by a single `.`. Historical queries using `ccf::historical::adapter` should also pass a single combined `x-ms-ccf-transaction-id` header (#2257).
-- Node unique identifier is the hex-encoded string of the SHA-256 digest of the node's identity public key, which is also used as the node's quote report data (#2241).
+- Node unique identifier is the hex-encoded string of the SHA-256 digest of the node's DER-encoded identity public key, which is also used as the node's quote report data (#2241).
+- Members and users unique identifier is the hex-encoded string of the SHA-256 digest of their DER-encoded identity certificate (i.e. fingerprint), which has to be specified as the `keyId` field for signed HTTP requests (#2279).
+
+### Removed
+
+- `/caller_id` endpoint. Members and users can compute their unique identifier without interacting with CCF (#2279).
```
```suggestion
- `/caller_id` endpoint. Members and users can now compute their unique identifier without interacting with CCF (#2279).
```

doc/governance/adding_member.rst
```diff
@@ -30,6 +30,13 @@ Members that are registered in CCF `with` a public encryption key are recovery m
 
 The member’s identity and encryption private keys (e.g. ``member_name_privk.pem`` and ``member_name_enc_privk.pem``) should be stored on a trusted device (e.g. HSM) while the certificate (e.g. ``member_name_cert.pem``) and public encryption key (e.g. ``member_name_enc_pubk.pem``) should be registered in CCF by members.
 
+The CCF unique member identity is the hex-encoded string of the SHA-256 hash of the DER-encoded certificate, and can be computed without interacting with CCF:
```
achamayou, line 33:
```suggestion
The CCF unique member identity is the hex-encoded string of the SHA-256 hash of the DER-encoded certificate, and can be computed from the certificate alone, without interacting with CCF:
```

src/node/certs.h
achamayou: In the unlikely event that it does, we can always memoise it like we do for verifiers.

src/node/entity_id.h
```diff
@@ -0,0 +1,167 @@
...
+  template <typename ParseContext>
+  auto parse(ParseContext& ctx)
+  {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext>
+  auto format(const ccf::EntityId& entity_id, FormatContext& ctx)
+    -> decltype(ctx.out())
+  {
```
achamayou, line 159: Why are we systematically truncating those in format?

src/node/certs.h
jumaffre: As per https://github.com/microsoft/CCF/pull/2279#issuecomment-792863287, these changes don't impact performance 👍

doc/schemas/gov_openapi.json
jumaffre: As discussed, because of the `operator std::string()`, this serialised the map as an object. I've removed the operator and used `.value()` instead so that the schema is in line with what the serialisation code actually does.

src/node/entity_id.h
jumaffre: I've removed the custom formatting and added an explicit `trim()` member which is called from `raft.h` when the node ID needs to be printed (only on "sunny-day" messages).

eddyashton (approved):

PR #2283, opened 2021-03-09
Add API for getting View from SeqNo

eddyashton:
Rather than accessing `consensus` directly, we want to provide a stable API for apps to access this common functionality. This PR adds `get_view_for_seqno_v1`, and an endpoint in `nobuiltins` which uses it.


samples/apps/nobuiltins/nobuiltins.cpp
```diff
@@ -131,6 +133,67 @@ namespace nobuiltins
...
+            }
+
+            kv::Consensus::View view;
+            const auto result = get_view_for_seqno_v1(seqno, view);
+            if (result == ccf::ApiResult::OK)
+            {
+              ccf::TxID tx_id;
+              tx_id.view = view;
+              tx_id.seqno = seqno;
+              nlohmann::json out = nlohmann::json::object();
```
jumaffre, line 167: Why not using an out schema for this endpoint too?

jumaffre (approved):
Could you add a quick end-to-test for this new endpoint?

samples/apps/nobuiltins/nobuiltins.cpp
eddyashton: No real reason, I've added this now.

eddyashton:
> Could you add a quick end-to-test for this new endpoint?

We actually didn't have an e2e test for this nobuiltins app at all, but I've added it now.

tests/nobuiltins.py
```diff
@@ -0,0 +1,54 @@
...
+    with infra.network.network(
+        args.nodes, args.binary_dir, args.debug_nodes, args.perf_nodes, pdb=args.pdb
+    ) as network:
+        network.start_and_join(args)
+
+        test_nobuiltins_endpoints(network, args)
+
+
+if __name__ == "__main__":
+    LOG.warning("Hello")
```
achamayou, line 48: This line and line 51 are presumably temporary?

achamayou (approved):

tests/nobuiltins.py
eddyashton: Haha, yes! Removed.

PR #2284, opened 2021-03-09
Fix find primary timeout

jumaffre:
In #2241, I incorrectly changed the logic of `network.find_primary()` so that we don't exit early if the primary is found. This meant that some end-to-end tests that rely heavily on this function took longer, and started to fail in the Daily pipeline. This PR should fix this.

### DIFF
--- a/tests/infra/network.py
+++ b/tests/infra/network.py
@@ -695,14 +695,17 @@ class Network:
                     try:
                         logs = []
                         res = c.get("/node/network", log_capture=logs)
-                        if res.status_code != 200:
-                            continue
+                        assert res.status_code == http.HTTPStatus.OK.value, res
+
                         body = res.body.json()
                         view = body["current_view"]
                         if "primary_id" not in body:
                             continue
+
                         view_change_in_progress = body["view_change_in_progress"]
                         primary_id = body["primary_id"]
+                        if primary_id is not None:
+                            break
 
                     except CCFConnectionException:
                         LOG.warning(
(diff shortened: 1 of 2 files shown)


tests/infra/network.py
```diff
@@ -695,14 +695,17 @@ def find_primary(self, timeout=3, log_capture=None):
                     try:
                         logs = []
                         res = c.get("/node/network", log_capture=logs)
-                        if res.status_code != 200:
-                            continue
+                        assert res.status_code == http.HTTPStatus.OK.value, res
```
jumaffre, line 698: The `/node/network` endpoint isn't guaranteed to return `200` but at this stage, we only loop through nodes that have joined the service so this assert is actually safe.

tests/infra/network.py
```diff
@@ -695,14 +695,17 @@ def find_primary(self, timeout=3, log_capture=None):
...
+
                         body = res.body.json()
                         view = body["current_view"]
                         if "primary_id" not in body:
                             continue
+
                         view_change_in_progress = body["view_change_in_progress"]
                         primary_id = body["primary_id"]
+                        if primary_id is not None:
```
eddyashton, line 707: Is this `if` necessary? It looks like line 702 has already checked that `"primary_id"` is present, and then we've assigned it here. Or is it possible to have a `"primary_id": null` in the response object?

eddyashton (approved):

tests/infra/network.py
achamayou: It's not now, but when @jumaffre and I talked about it, I suggested we should always have a primary_id and it should be set to null when it's unknown, rather than not have the key at all when we don't know which node is primary.

There's a more general point here about what do for optional fields in our JSON responses, we can talk about it.

PR #2285, opened 2021-03-09
Historical JS endpoints

letmaik:


src/apps/js_generic/js_generic.cpp
```diff
@@ -813,9 +814,22 @@ namespace ccfapp
...
       auto kv = JS_NewObjectClass(ctx, kv_class_id);
-      JS_SetOpaque(kv, &args.tx);
+      JS_SetOpaque(kv, &tx);
       JS_SetPropertyStr(ctx, ccf, "kv", kv);
 
+      auto state = JS_NewObject(ctx);
+      auto kv_tx_id = tx.get_read_tx_id();
+      ccf::TxID tx_id;
+      tx_id.view = static_cast<ccf::View>(kv_tx_id.term);
+      tx_id.seqno = static_cast<ccf::SeqNo>(kv_tx_id.version);
```
letmaik, line 824: This is very annoying. I'm doing this so that I can `to_str()`.  The casts are needed as `kv::TxId` uses signed, while `ccf::TxID` uses unsigned ints.

samples/apps/forum/src/types/ccf.ts
```diff
@@ -74,6 +99,7 @@ export interface CCF {
   ): ArrayBuffer;
 
   kv: KVMaps;
+  state: State;
```
letmaik, line 102: Does this make sense?

samples/apps/forum/src/types/ccf.ts
```diff
@@ -48,6 +48,26 @@ export interface KVMap {
...
+export interface Receipt {
+  signature: string;
+  root: string;
+  proof: Proof;
+  leaf: string;
+  nodeId: string;
+}
+
+export interface State {
+  transactionId: string;
```
letmaik, line 67: Always defined, it's the *read* transaction ID.

samples/apps/forum/src/types/ccf.ts
```diff
@@ -48,6 +48,26 @@ export interface KVMap {
...
 
+export interface ProofElement {
+  left?: string
+  right?: string
+}
+
+export type Proof = ProofElement[];
+
+export interface Receipt {
+  signature: string;
```
letmaik, line 59: I haven't implemented this yet, but, there's the choice of providing `ArrayBuffer`s, or Base64/hex-encoded strings. I'm leaning slightly towards `ArrayBuffer` to avoid having an opinion on final encoding. It would require a bit more work from the app developer, but nothing too dramatic.

samples/apps/forum/src/types/ccf.ts
```diff
@@ -74,6 +99,7 @@ export interface CCF {
   ): ArrayBuffer;
 
   kv: KVMaps;
+  state?: State;
```
letmaik, line 102: Only for historical queries. Bad name?

achamayou: `historical.state` or `historicalState` would be preferable.

samples/apps/logging/js/app.json
```diff
@@ -29,6 +29,30 @@
...
+    },
+    "/log/private/historical_receipt": {
+      "get": {
+        "js_module": "logging.js",
+        "js_function": "get_historical_with_receipt",
+        "forwarding_required": "never",
+        "execute_outside_consensus": "never",
+        "authn_policies": ["jwt", "user_cert"],
+        "historical": true,
+        "readonly": true,
```
achamayou, line 52: Historical implies readonly, it would be elegant if the config reflected that.

src/apps/js_generic/js_generic.cpp
```diff
@@ -568,7 +568,9 @@ namespace ccfapp
     JSValueConst this_val,
     JSAtom property)
   {
-    const auto property_name = JS_AtomToCString(ctx, property);
+    const auto property_name_c = JS_AtomToCString(ctx, property);
+    const auto property_name = std::string(property_name_c);
```
achamayou, line 572:
```suggestion
    const std::string property_name(property_name_c);
```

src/apps/js_generic/js_generic.cpp
```diff
@@ -994,6 +1062,70 @@ namespace ccfapp
...
       EndpointContext& args)
+    {
+      // Is this a historical endpoint?
+      auto endpoints =
+        args.tx.ro<ccf::endpoints::EndpointsMap>(ccf::Tables::ENDPOINTS);
+      auto info = endpoints->get(ccf::endpoints::EndpointKey{method, verb});
+
+      if (info.has_value() && info.value().historical)
+      {
+        // (copied from samples/apps/logging/logging.cpp)
```
achamayou, line 1073: Let's factor that out.

achamayou:
Looks good! Can the logging JS section in the doc and the changelog be expanded accordingly please?

samples/apps/logging/js/app.json
letmaik: `"mode": "readonly|writable|historical"` ? Or were you thinking of something different?

letmaik:
> 
> 
> Looks good! Can the logging JS section in the doc and the changelog be expanded accordingly please?

There's only a C++ section for logging currently. Let's do docs in another PR.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- Single-transaction historical query support has been added to JavaScript endpoints (#2285).
+
```
achamayou, line 13: Does single-transaction mean point queries as opposed to range queries?

letmaik: Yes. I'll change it.

cmake/quickjs.cmake
```diff
@@ -28,8 +28,10 @@ if("sgx" IN_LIST COMPILE_TARGETS)
     quickjs.enclave STATIC ${QUICKJS_SRC} ${CCF_DIR}/3rdparty/stub/time.c
   )
   target_compile_options(
-    quickjs.enclave PUBLIC -nostdinc -DCONFIG_VERSION="${QUICKJS_VERSION}"
-                           -DEMSCRIPTEN -DCONFIG_STACK_CHECK
+    quickjs.enclave
+    PUBLIC -nostdinc -DCONFIG_VERSION="${QUICKJS_VERSION}" -DEMSCRIPTEN
+           -DCONFIG_STACK_CHECK
+    PRIVATE $<$<CONFIG:Debug>:-DDUMP_LEAKS>
```
achamayou, line 34: Just to check, this is saying "when building this target in Debug, set DUMP_LEAKS", right?

samples/apps/forum/src/types/ccf.ts
```diff
@@ -48,6 +48,26 @@ export interface KVMap {
...
 
+export interface ProofElement {
+  left?: string;
+  right?: string;
+}
+
+export type Proof = ProofElement[];
+
+export interface Receipt {
+  signature: string;
```
achamayou, line 59: Could we add a jsdoc comments like /** base64-encoded signature of the root by the node identified by nodeId */, and if we did, would it get picked up by sphinx-js? Just wondering, not requesting :)

samples/apps/forum/src/types/ccf.ts
```diff
@@ -61,6 +81,11 @@ export interface AESKWPParams extends WrapAlgoParams {
   name: "AES-KWP";
 }
 
+export interface RsaOaepAESKWPParams extends WrapAlgoParams {
```
achamayou, line 84: I'd say either RSAOAEPAESKWParams or RsaOaepAesKwpParams, but this is a ConfusingWAYToDOIT.

wintersteiger: Yep, that part slipped through; entirely my fault :-)

samples/apps/logging/js/app.json
achamayou: Yeah I think that'd be good. Paging @eddyashton for API input, but I think this would have the merit of ruling out historical: true, readonly: false and the validation and error that'd need to go with it.

cmake/quickjs.cmake
letmaik: Yeah

samples/apps/forum/src/types/ccf.ts
letmaik: Yes and yes, see the CryptoKeyPair interface in https://github.com/microsoft/CCF/pull/2313

samples/apps/forum/src/types/ccf.ts
letmaik: Not related to this PR, this was just an inconsistency between the two ccf.ts copies.

samples/apps/logging/js/app.json
eddyashton: Yes, I think this is a tri-state enum. I like `readonly` and `historical` but not convinced by `writable`. I think `readwrite` is what we've used elsewhere (eg `tx.rw`)? I could also go for `canwrite` or `maywrite` or just `write`, but I think `readwrite` is best.

letmaik: `readwrite` is fine. I'll do the change.

achamayou (approved):

samples/apps/logging/js/app.json
letmaik: @achamayou @eddyashton Done, please have a final look if the change looks good.

PR #2287, opened 2021-03-09
Use pushd to always run ci-checks.sh at the project root

eddyashton:


scripts/ci-checks.sh
```diff
@@ -12,6 +12,9 @@ fi
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
+ROOT_DIR=$( dirname "$SCRIPT_DIR" )
+pushd "$ROOT_DIR"
```
wintersteiger, line 16: Without a `popd`, would this leave the caller in `$ROOT_DIR` too?

eddyashton: No - this dir stack is local to the script, so when we exit (exceptionally or otherwise) the stack is unchanged and the caller remains where they were.

achamayou (approved):

Issue #1765, opened 2020-10-15
Update JS API docs


letmaik:
https://microsoft.github.io/CCF/master/developers/js_app_bundle.html#globals

- renamed: `tables` -> `ccf.kv`
- added: `<map>.has()`
- added: `ccf.strToBuf`, `ccf.bufToStr`, `ccf.jsonCompatibleToBuf`, `ccf.bufToJsonCompatible`, `ccf.generateAesKey`, `ccf.wrapKey`

PR #2293, opened 2021-03-10
Add JS endpoint for RSA key generation

wintersteiger:


src/apps/js_generic/js_generic.cpp
```diff
@@ -185,6 +186,51 @@ namespace ccfapp
...
+
+    std::shared_ptr<RSAKeyPair> k;
+    if (argc == 1)
+      k = crypto::make_rsa_key_pair(key_size);
+    else
+      k = crypto::make_rsa_key_pair(key_size, key_exponent);
+
+    Pem p = k->private_key_pem();
+
+    return JS_NewArrayBufferCopy(ctx, p.data(), p.size());
```
letmaik, line 231: Can we return the public one as well? Maybe using the same field names as in https://developer.mozilla.org/en-US/docs/Web/API/CryptoKeyPair.

src/apps/js_generic/js_generic.cpp
```diff
@@ -806,6 +852,11 @@ namespace ccfapp
         ccf,
         "generateAesKey",
         JS_NewCFunction(ctx, ccfapp::js_generate_aes_key, "generateAesKey", 1));
+      JS_SetPropertyStr(
+        ctx,
+        ccf,
+        "generateRsaKey",
```
letmaik, line 858: `generateRsaKeyPair`?

samples/apps/forum/src/types/ccf.ts
```diff
@@ -67,6 +67,7 @@ export interface CCF {
   jsonCompatibleToBuf<T extends JsonCompatible<T>>(v: T): ArrayBuffer;
   bufToJsonCompatible<T extends JsonCompatible<T>>(v: ArrayBuffer): T;
   generateAesKey(size: number): ArrayBuffer;
+  generateRsaKey(size: number, exponent?: number): ArrayBuffer;
```
letmaik, line 70: Let's return strings here since it is a PEM.

tests/js-modules/modules.py
```diff
@@ -262,6 +262,10 @@ def test_npm_app(network, args):
         assert len(r.body.data()) == key_size // 8
         assert r.body.data() != b"\x00" * (key_size // 8)
 
+        r = c.post("/app/generateRsaKey", {"size": 2048})
+        assert r.status_code == http.HTTPStatus.OK, r.status_code
+        assert not all(v == b"\x00" for v in r.body.data())
```
letmaik, line 267: This test should probably try to actually parse the returned key, likely by adding some new functions to infra/crypto.py.

src/apps/js_generic/js_generic.cpp
```diff
@@ -185,6 +186,51 @@ namespace ccfapp
...
+    }
+
+    if (argc == 2 && JS_ToInt32(ctx, &key_exponent, argv[1]) < 0)
+    {
+      js_dump_error(ctx);
+      return JS_EXCEPTION;
+    }
+
+    // Supported key sizes for RSA.
+    if (key_size < 1024 || key_size >= MBEDTLS_MPI_MAX_BITS)
```
letmaik, line 210: This file shouldn't know about mbedtls. Better to move this check into a function `is_supported_rsa_key_size(n)` in the crypto namespace.

src/apps/js_generic/js_generic.cpp
```diff
@@ -185,6 +186,51 @@ namespace ccfapp
...
   }
 
+  static JSValue js_generate_rsa_key(
+    JSContext* ctx, JSValueConst, int argc, JSValueConst* argv)
+  {
+    if (argc != 1 && argc != 2)
+      return JS_ThrowTypeError(
+        ctx, "Passed %d arguments, but expected 1 or 2", argc);
+
+    int32_t key_size, key_exponent;
```
letmaik, line 196: Why not `uint32_t` together with `JS_ToUint32()`?

src/apps/js_generic/js_generic.cpp
```diff
@@ -185,6 +186,51 @@ namespace ccfapp
...
+    }
+
+    // Supported key sizes for RSA.
+    if (key_size < 1024 || key_size >= MBEDTLS_MPI_MAX_BITS)
+    {
+      JS_ThrowRangeError(ctx, "invalid key size");
+      js_dump_error(ctx);
+      return JS_EXCEPTION;
+    }
+    else if (argc == 2 && (key_exponent <= 1 || key_exponent % 2 == 0))
```
letmaik, line 216: Similarly, this logic should be in a helper function in the crypto namespace.

src/apps/js_generic/js_generic.cpp
wintersteiger: Yes, that's exactly what I wanted anyways, thanks for the link!

src/apps/js_generic/js_generic.cpp
wintersteiger: mbedTLS has tighter limits, but all of these checks are unnecessary anyways. I just added a few of them to catch common problems, but I'll remove them and just return whatever `make_rsa_key_pair` throws.

src/apps/js_generic/js_generic.cpp
wintersteiger: Because I copied that part from `js_generate_aes_key` :-)

samples/apps/forum/src/types/ccf.ts
```diff
@@ -67,6 +67,7 @@ export interface CCF {
   jsonCompatibleToBuf<T extends JsonCompatible<T>>(v: T): ArrayBuffer;
   bufToJsonCompatible<T extends JsonCompatible<T>>(v: ArrayBuffer): T;
   generateAesKey(size: number): ArrayBuffer;
```
achamayou, line 69: I think now is a good time to start documenting the js API.

src/apps/js_generic/js_generic.cpp
```diff
@@ -185,6 +186,51 @@ namespace ccfapp
...
+      JS_ThrowRangeError(ctx, "invalid key exponent");
+      js_dump_error(ctx);
+      return JS_EXCEPTION;
+    }
+
+    std::shared_ptr<RSAKeyPair> k;
+    if (argc == 1)
+      k = crypto::make_rsa_key_pair(key_size);
+    else
+      k = crypto::make_rsa_key_pair(key_size, key_exponent);
```
achamayou, line 227:
```suggestion
    if (argc == 1)
    {
      k = crypto::make_rsa_key_pair(key_size);
    }
    else
    {
      k = crypto::make_rsa_key_pair(key_size, key_exponent);
     }
```

achamayou (approved):
LGTM but please consider documenting the growing JS API in Sphinx.

wintersteiger:
Yes, I agree, we should start documenting this properly. I have trouble seeing where exactly "the JS API" begins and ends and what shape or form the documentation should take. Should there be a general `CCF.ts` file with docstrings on the functions which we distribute instead of copying it into samples? @letmaik can you take the lead on this, if you're not too busy? I'm happy to write the text for the crypto bits or to set up the infrastructure if you tell me how you'd like to do this.

wintersteiger:
See also #1765

tests/infra/crypto.py
```diff
@@ -173,3 +173,13 @@ def compute_public_key_der_hash_hex_from_pem(pem: str):
...
+
+
+def check_key_pair_pem(private: str, public: str, password=None) -> bool:
+    prv = load_pem_private_key(private.encode(), password=password)
+    pub = load_pem_public_key(public.encode())
+    prv_pub_der = prv.public_key().public_bytes(
+        Encoding.DER, PublicFormat.SubjectPublicKeyInfo
+    )
+    pub_der = pub.public_bytes(Encoding.DER, PublicFormat.SubjectPublicKeyInfo)
+    return prv_pub_der == pub_der
```
letmaik, line 185:
```suggestion
    return prv.public_key() == pub
```

tests/npm-app/app.json
```diff
@@ -118,6 +118,45 @@
...
+                  "type": "object"
+                }
+              }
+            }
+          },
+          "responses": {
+            "200": {
+              "description": "Generated key pair",
+              "content": {
+                "application/octet-stream": {}
```
letmaik, line 153:
```suggestion
                "application/json": {
                  "schema": {
                    "properties": {
                       "privateKey": {
                         "type": "string"
                       },
                       "publicKey": {
                         "type": "string"
                       }
                    },
                    "required": ["privateKey", "publicKey"],
                    "type": "object"
                  }
                }
```

letmaik (approved):

letmaik:
Looks good to me, let's tackle the docs in a separate PR.

tests/infra/crypto.py
wintersteiger: That's not right; that comparison is always false because it compares two different key objects (which hold the same key). I initially tried the same and then had to add the complicated conversion to make it work.

wintersteiger:
Yeah, I agree, the docs can go into a separate PR.

tests/infra/crypto.py
letmaik: Really? I could have sworn I used that in the past, hmm..

PR #2294, opened 2021-03-10
Enable component governance

wintersteiger:


.cg.yml
```diff
@@ -0,0 +1,40 @@
...
+
+pr:
+  autoCancel: true
+  branches:
+    include:
+      - main
+  paths:
+    include:
+      - .daily_canary
+      - .cg.yml
```
achamayou, line 16:
```suggestion
      - .cg.yml
      - cgmanifest.json
```

achamayou: If we update the manifest, we probably want to re-run as well.

.cg.yml
```diff
@@ -0,0 +1,40 @@
...
+    displayName: "Component Governance"
+
+    steps:
+      - checkout: self
+        clean: true
+        fetchDepth: 1
+
+      - task: ComponentGovernanceComponentDetection@0
+        inputs:
+          ignoreDirectories: 'tests'
```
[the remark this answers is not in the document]
achamayou: Only one of them is, the logging sample, and it has no dependencies. We do not ship the dependencies themselves in any case (except for C++), not even for the python/ module.

Issue #2237, opened 2021-02-24
Service KV tables schema evolution


jumaffre:
See https://github.com/microsoft/CCF/discussions/2181#discussioncomment-350194 for interesting problems this issue is aiming to prevent.

Whenever possible, CCF service tables should follow the following pattern:
- [ ] The key should be a unique field which is blit-serialised.
- [ ] The value should be serialised as JSON
- [ ] Only new _optional_ fields can be added to the value of an existing table (no deletion), so that newer nodes don't throw an error when deserialising an old entry that doesn't contain newer fields (*)

(*) We may be able to do something smarter here with a better `from_json` deserialiser, so that fields can be deleted.

Note that it is the responsibility of application writers to use serialisation schemes that allow for versioning and backwards compatibility.

PR #2301, opened 2021-03-11
Change serialisation format for service maps

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/2237 
Largely inspired by https://github.com/microsoft/CCF/pull/1860

In this PR, I've converted most service tables from [msgpack -> msgpack] to [raw -> JSON], with some tables changed to [raw -> raw] for performance reasons.

Notable changes:
- All binary data is serialised to base64, except for SHA-256 hashes that are hex-encoded
- The serialised Merkle tree which used to be included in each signature transaction is now written to a separate table `public:ccf.internal.tree` (raw -> raw) as encoding in the JSON signature was way too costly.
- The members and users table have been split so that the certificates (raw -> raw) can be accessed quickly in the TLS certificate authn schemes.
- The user and member data is no longer passed as part of the `AuthnIdentity` but should instead be retrieved via two new API calls: `get_user_data_v1()` and `get_member_data_v1()`. 
- Parsing the ledger from the Python `ledger` module is much more convenient.
- New `ds/` functions to encode/decode hex strings.


src/ds/test/hex.cpp
```diff
@@ -0,0 +1,46 @@
...
+  {
+    INFO("Uppercase hex string gets lowercased");
+
+    std::string uppercase_hex_str("0123456789ABCDEF");
+    std::string lowercase_hex_str = uppercase_hex_str;
+    std::transform(
+      lowercase_hex_str.begin(),
+      lowercase_hex_str.end(),
+      lowercase_hex_str.begin(),
+      [](unsigned char c) { return std::tolower(c); });
```
eddyashton, line 40:
```suggestion
    nonstd::to_lower(lowercase_hex_str);
```

python/ccf/clients.py
```diff
@@ -35,7 +35,7 @@ def escape_loguru_tags(s):
     return loguru_tag_regex.sub(lambda match: f"\\{match[0]}", s)
 
 
-def truncate(string: str, max_len: int = 128):
+def truncate(string: str, max_len: int = 256):
```
jumaffre, line 38: Since the member/user/node ids change, many responses were truncated too aggressively to display any useful information.

samples/apps/nobuiltins/nobuiltins.cpp
```diff
@@ -56,10 +56,8 @@ namespace nobuiltins
           }
 
           summary.quote_format = quote_info.format;
-          summary.quote =
-            fmt::format("{:02x}", fmt::join(quote_info.quote, ""));
-          summary.endorsements =
-            fmt::format("{:02x}", fmt::join(quote_info.endorsements, ""));
+          summary.quote = ds::to_hex(quote_info.quote);
```
jumaffre, line 59: I can't seem to recall the reason why we encode these as hex and not base64, but it feels like the latter would be more consistent with the rest of binary data in CCF. @achamayou?

src/http/authentication/cert_auth.h
```diff
@@ -124,31 +116,23 @@ namespace ccf
       const std::shared_ptr<enclave::RpcContext>& ctx,
       std::string& error_reason) override
     {
-      const auto caller_cert_pem =
-        crypto::cert_der_to_pem(ctx->session->caller_cert);
-
-      std::unique_ptr<NodeCertAuthnIdentity> identity = nullptr;
+      auto caller_public_key_der =
```
jumaffre, line 119: Unrelated to this PR but fixed it anyway: since the node ID changes, we no longer need to `foreach` the nodes table to get the information for that node. We simply extract the public key from the cert, hash it, and use the hash as the key in the nodes table.

src/kv/deserialise.h
```diff
@@ -81,9 +81,11 @@ namespace kv
       auto search = changes.find(ccf::Tables::SIGNATURES);
       if (search != changes.end())
       {
-        // Transactions containing a signature must only contain
-        // a signature and must be verified
-        if (changes.size() > 1)
+        // Transactions containing a signature must only contain the signature
+        // and the serialised Merkle tree and must be verified
+        if (
```
jumaffre, line 86: The serialised tree is not directly signed by the signature transaction but the historical queries module verifies that the root of the serialised tree matches the root that's signed in the signature transaction.

src/node/certs.h
```diff
@@ -2,10 +2,9 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
-#include "entities.h"
+#include "service_map.h"
 
 namespace ccf
 {
-  using CertDERs = kv::Map<Cert, CallerId>;
```
jumaffre, line 9: Unused since the member/user id changes.

src/node/channels.h
```diff
@@ -417,6 +417,15 @@ namespace ccf
       // the non-authenticated plaintex payload. data contains payload first,
       // then GCM header
 
+      if (status != ESTABLISHED)
```
jumaffre, line 420: Unrelated but discussed with @wintersteiger: we shouldn't accept messages if the channel isn't ready.

src/node/modules.h
```diff
@@ -10,18 +12,6 @@
...
-    Module(const std::string& js_) : js(js_) {}
-
-    MSGPACK_DEFINE(js);
-  };
-  DECLARE_JSON_TYPE(Module)
-  DECLARE_JSON_REQUIRED_FIELDS(Module, js)
-
-  using Modules = kv::Map<std::string, Module>;
+  using Module = std::string;
+  using Modules = kv::RawCopySerialisedMap<std::string, Module>;
```
jumaffre, line 16: FYI, using `ServiceMap` has some non-negligible impact (~-15%) on the `ls_js_sgx_cft` perf test.

src/node/rpc/member_frontend.h
```diff
@@ -2117,16 +2104,14 @@ namespace ccf
   {
   protected:
     MemberEndpoints member_endpoints;
-    Members* members;
```
jumaffre, line 2120: This was unused.

src/node/service_map.h
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "crypto/pem.h"
+#include "kv/map.h"
+
+namespace ccf
+{
+  template <typename K, typename V>
+  using ServiceMap = kv::MapSerialisedWith<
```
jumaffre, line 11: Should this just become the new `kv::Map`?

src/apps/smallbank/app/smallbank.cpp
```diff
@@ -15,6 +15,8 @@ namespace ccfapp
 {
   struct SmallBankTables
   {
+    // TODO: Update to RawCopySerialisedMap last, once all service maps have
```
jumaffre, line 18: I'll update this last, once we have a few stable perf runs with the service maps changes only.

python/ccf/ledger.py
```diff
@@ -162,7 +150,7 @@ def _byte_read_safe(file, num_of_bytes):
     return ret
 
 
-class NodeStatus(IntEnum):
+class NodeStatus(Enum):
     """These are the corresponding status meanings from the ccf.nodes table"""
 
     PENDING = 0
```
eddyashton, line 156: Shouldn't these now be strings? I assume the integer values were the raw message pack encoding, but these will now be JSON so should be converted from strings?

python/ledger_tutorial.py
```diff
@@ -42,9 +42,5 @@
             # Ledger verification is happening implicitly in ccf.ledger.Ledger()
             for key, value in public_tables[target_table].items():
                 target_table_changes += 1  # A key was changed
-                # Log the key and value for the transaction on the target table
-                # The target_table: 'public:ccf.gov.nodes.info' has already been decoded in ledger.py
-                # For other tables knowledge of serialization scheme used is important.
-                # If the table was using msgpack, use ccf.ledger.extract_msgpacked_data(data)
```
eddyashton, line 48: Can we restore some of this comment? Is the logging line below showing something interesting because the target table is JSON-to-JSON, or just dumping some binary? We should still say that serialization schemes are per-table and must be known to interpret anything useful here.

samples/apps/logging/logging.cpp
```diff
@@ -878,7 +904,26 @@ namespace loggingapp
...
-            const nlohmann::json user_data = caller_ident.user_data;
+            nlohmann::json user_data = nullptr;
+            auto result =
+              get_user_data_v1(ctx.tx, caller_ident.user_id, user_data);
+            if (result == ccf::ApiResult::NotFound)
+            {
+              return ccf::make_error(
+                HTTP_STATUS_NOT_FOUND,
+                ccf::errors::ResourceNotFound,
+                fmt::format("User {} does not exist", caller_ident.user_id));
```
eddyashton, line 915: I don't think this message is correct? Its possible to be a real, known user, but have no user data, in which case we'd get a `NotFound`?

src/consensus/aft/request.h
```diff
@@ -20,32 +20,44 @@ namespace aft
 
     MSGPACK_DEFINE(rid, caller_cert, raw, frame_format);
 
-    std::vector<uint8_t> serialise()
+    size_t serialised_size() const
     {
-      bool include_caller = false;
-      size_t size = sizeof(rid) + sizeof(include_caller) + sizeof(size_t) +
-        raw.size() + sizeof(enclave::FrameFormat);
+      size_t size = sizeof(rid) + sizeof(bool) + sizeof(size_t) + raw.size() +
```
eddyashton, line 25: Not for this PR but its been bothering me for a while and I don't think we wrote it anywhere: We should really add a `serialized::size` which does this calculation, since we have a similar pattern in so many places. What we're really saying is "I know I'm going to call `serialized::write(data, size, VECTOR)` later, and I know _that_ is going to write a `size_t` prefix then VECTOR.size bytes", and if we decide to try to pack that a whole bunch of hard-to-grep code is going to break. If we instead used `serialized::size(VECTOR)` as a precursor to `serialized::write`, the prefixing logic could be contained.

src/ds/json.h
```diff
@@ -61,19 +62,49 @@ namespace std
   template <typename T>
   inline void to_json(nlohmann::json& j, const std::vector<T>& t)
   {
-    j = nlohmann::json::array();
-    for (const auto& e : t)
+    if constexpr (std::is_same_v<T, uint8_t>)
```
eddyashton, line 65: It looks like we don't have an equivalent path in the schema generation, so where we're doing this automatic conversion to a base64 string the schema will still claim a list of uint8s?

src/http/authentication/cert_auth.h
```diff
@@ -34,21 +32,17 @@ namespace ccf
       const auto& caller_cert = ctx->session->caller_cert;
       auto caller_id = crypto::Sha256Hash(caller_cert).hex_str();
 
-      auto users = tx.ro<Users>(Tables::USERS);
-      const auto user = users->get(caller_id);
-      if (user.has_value())
+      auto user_certs = tx.ro<UserCerts>(Tables::USER_CERTS);
+      const auto user_cert = user_certs->get(caller_id);
```
eddyashton, line 36: If we're retrieving (and deserialising) the user data lazily, it seems like we could also get a slight perf win (/consistency gain) from making the user cert lazy as well? Or in fact, we could use the `caller_cert` here, and not deserialise from the `UserCerts` table, since they "must" match!

src/kv/deserialise.h
```diff
@@ -81,9 +81,11 @@ namespace kv
...
+          changes.size() > 2 ||
+          changes.find(ccf::Tables::SERIALISED_MERKLE_TREE) == changes.end())
```
eddyashton, line 88: Should we be finding both expected table names here?

src/node/modules.h
eddyashton: This table is a prime candidate to benefit from #2220. Large strings to deserialise, changed rarely, used by every endpoint invocation.

src/node/network_tables.h
```diff
@@ -100,9 +103,12 @@ namespace ccf
     Service service;
     Values values;
     Secrets secrets;
-    Signatures signatures;
     SnapshotEvidence snapshot_evidence;
 
+    // The signatures and serialised_tree tables should always be in sync
```
eddyashton, line 108: What does in-sync mean here? They both contain a single entry at key 0, and `signatures` is a signature over the root of the entry written to `serialised_tree` at the same version?

src/node/service_map.h
eddyashton: No - we shouldn't change the schema for app tables like this (though maybe we deprecate `kv::Map` at some point if we have a few sensibly named alternatives we want them to pick from, to make the serialization choice more explicit). And keeping a separate `ServiceMap` name is very useful for distinguishing 'our' tables.

src/node/service_map.h
```diff
@@ -0,0 +1,34 @@
...
+      const auto& data = pem.raw();
+      return SerialisedEntry(data.begin(), data.end());
+    }
+
+    static crypto::Pem from_serialised(const SerialisedEntry& data)
+    {
+      return crypto::Pem(data.data(), data.size());
+    }
+  };
+}
```
eddyashton, line 34: Odd place for this - don't we have a file defining the cert-keyed tables where this could go?

samples/apps/nobuiltins/nobuiltins.cpp
achamayou: I don't think there's a good reason other than this was done before we decided on the "hex for digests, base64 for everything else"-policy, it's definitely a good idea to change it.

src/node/modules.h
achamayou: There's also a considerable amount of unnecessary duplication because of the export hack at the moment.

src/http/authentication/cert_auth.h
jumaffre: I was wondering about that. My original hunch was it's nicer to always present the PEM cert to users, rather than the DER-encoded `caller_cert`. The `UserCerts` lookup is basically a cheap (*) way to encode the caller certificate as PEM. I'm not sure there are any advantages to have the caller cert as PEM rather than DER anymore though, so I'm keen to simplify this too.

(*) I've checked and a pure DER to PEM conversion (crypto::cert_der_to_pem(der)`) costs ~2/3 of performance on the smallbank benchmark.

src/kv/deserialise.h
jumaffre: We already have found the signatures table in `changes` on line `81`.

python/ledger_tutorial.py
jumaffre: Good point. I've added a comment to explain this. Both `key` and `value` are raw bytes, and it's up to this code to know how to deserialise them.

src/node/network_tables.h
jumaffre: Good point. I've clarified the comment.

src/node/service_map.h
jumaffre: I swear there was a `TODO` around this at some point! I'll move this and all other `BlitSerialiser` definitions to a new file.

src/http/authentication/cert_auth.h
eddyashton: So I think this is an extra argument for making this lazy. We're not sure if we should be presenting the cert as DER or PEM, because we don't know what an app might want to use, and in fact we have no apps which use either. I think we should stop trying to present one 'for free' (because we've seen that even this deserialisation from the `UserCerts` lookup is not free), and instead have `get_cert_for_user_v1` (or perhaps more precisely `get_pem_for_user_v1` and/or `get_der_for_user_v1`, but I think `get_cert` is fine and `crypto::` is part of our supported API).

src/ds/json.h
jumaffre: Done now. The associated strings are marked as "base64" in the OpenAPI spec too.

src/http/authentication/cert_auth.h
jumaffre: Done. Added two new `get_user_cert_v1()` and `get_member_cert_v1()` API calls.

achamayou (approved):

achamayou:
@jumaffre can you kick off a few CI runs just so we can get a few more data points on performance with the latest changes?

PR #2303, opened 2021-03-11
Add component governance CI job

wintersteiger:
This adds a separate pipeline for component governance checks, so that we're not dependant on auto-injection anymore. The way it is set up in this PR, it produces 5 correct alerts and I'm working with CELA to get those resolved (problems on their end).

### DIFF
--- a/.cg.yml
+++ b/.cg.yml
@@ -4,7 +4,6 @@ trigger:
     include:
       - main
       - "refs/tags/ccf-*"
-      - cwinter_com_gov
 
 pr:
   autoCancel: true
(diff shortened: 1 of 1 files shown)


wintersteiger:
(Regarding performance: the new job (see Checks) runs in parallel and currently takes 27 seconds, although that will likely go up when new components that aren't cached are added.)

.cg.yml
```diff
@@ -0,0 +1,37 @@
+trigger:
+  batch: true
+  branches:
+    include:
+      - main
+      - "refs/tags/ccf-*"
+      - cwinter_com_gov
```
achamayou, line 7: Presumably to be removed before merging?

achamayou (approved):

.cg.yml
wintersteiger: Yep, thanks!

PR #2305, opened 2021-03-11
Extend certificate validity range

### LINKED ISSUE #2190: Certificate validity period is fixed
As discussed with @wintersteiger, the validity period of the node's certificates is currently hardcoded to a set date, which isn't great:
https://github.com/microsoft/CCF/blob/acc4ad227777431e111d164d305b4f60bb35efad/src/tls/key_pair.h#L581-L586

Instead, this should be set by the operator when at node startup, with perhaps a default sets by the host to the current date minus a few days.

achamayou:
Fix #2190 for now.

Longer term, we probably want to add the following features:

1. Make node and network cert validity ranges user-configurable
2. Add a mechanism allowing a periodic refresh of the self-signed network certificate without creating a new identity

### DIFF
--- a/src/crypto/mbedtls/key_pair.cpp
+++ b/src/crypto/mbedtls/key_pair.cpp
@@ -460,7 +460,7 @@ namespace crypto
     // Note: 825-day validity range
     // https://support.apple.com/en-us/HT210176
     MCHK(mbedtls_x509write_crt_set_validity(
-      crt.get(), "20191101000000", "20211231235959"));
+      crt.get(), "20210311000000", "20230611235959"));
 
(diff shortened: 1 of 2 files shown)


src/crypto/mbedtls/key_pair.cpp
```diff
@@ -460,7 +460,7 @@ namespace crypto
     // Note: 825-day validity range
     // https://support.apple.com/en-us/HT210176
     MCHK(mbedtls_x509write_crt_set_validity(
```
jumaffre, line 462: Can you update the `openssl` implementation as well?

wintersteiger (approved):

eddyashton (approved):

jumaffre (approved):

PR #2306, opened 2021-03-11
Split JS Wrapper

achamayou:
Minimal changes necessary to make the JS runtime available in both application code and built-in endpoints.


letmaik (approved):

src/js/wrap.h
```diff
@@ -0,0 +1,172 @@
...
+
+  extern JSClassDef kv_class_def;
+  extern JSClassExoticMethods kv_exotic_methods;
+  extern JSClassDef kv_map_handle_class_def;
+  extern JSClassDef body_class_def;
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wc99-extensions"
+
+  extern void register_class_ids();
```
letmaik, line 28: I haven't seen `extern` used like that for function declarations before. Probably can be omitted?

letmaik (approved):

src/js/wrap.h
achamayou: Yes, extern is the default, bad search and replace, fixed.

eddyashton (approved):

jumaffre (approved):

PR #2309, opened 2021-03-11
Remove auto-JSONification of query params, allowing them to be unquoted

### LINKED ISSUE #2153: Accept plain strings for string query parameters of C++ endpoints

eddyashton:
Resolves #2153.

The `json_handler` adapter no longer parses the query string. Any request parameters which are included in the query string must be extracted manually by the handler, and we will not try to read/parse them beforehand. We do add some helper functions in `http_query.h` to remove some of the boiler plate in query parsing.

The main benefit of this is that when those query parameters are strings they no longer need to be quoted. Additionally, this gives the app more precise control of the OpenAPI description (in a form we want to extend in future). It also makes parsing of `GET` and `POST` requests more consistent and removes the requirement that we are opinionated on `GET`s having no body (the pre-parsed JSON will now contain the body from either method, and query strings must be manually parsed in either method).


doc/schemas/app_openapi.json
```diff
@@ -622,11 +618,9 @@
           {
             "in": "query",
             "name": "id",
-            "required": false,
+            "required": true,
```
eddyashton, line 621: It was a bug in the previous API generation that marked all of these query params as `"required": false` - most would actually throw a JSON parsing error if the field wasn't present. Now they're correctly marked as required, and the error if they're missing more clearly refers to the query string.

jumaffre (approved):

src/node/rpc/endpoint_registry.h
```diff
@@ -375,6 +354,40 @@ namespace ccf
         return set_auto_schema<typename T::In, typename T::Out>(status);
```
letmaik, line 354: Docstring is outdated a bit: "Sets the schema that the request parameters and response must comply with"

src/node/rpc/endpoint_registry.h
```diff
@@ -375,6 +354,40 @@ namespace ccf
...
       }
 
+      template <typename T>
```
letmaik, line 357: Docstring?

letmaik (approved):

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.19.1]
```
achamayou, line 8:
```suggestion
## [0.19.2]
```

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - `"readonly"` has been replaced by `"mode"` in `app.json` in JavaScript apps (#2285).
+- String values in query parameters no longer need to be quoted. For instance, you should now call `/network/nodes?host=127.0.0.1` rather than `/network/nodes?host="127.0.0.1"`.
```
achamayou, line 17: Needs shifting up.

CHANGELOG.md
```diff
@@ -738,6 +741,7 @@ Some discrepancies with the TR remain, and are being tracked under https://githu
 
 Initial pre-release
 
+[0.19.1]: https://github.com/microsoft/CCF/releases/tag/ccf-0.19.1
```
achamayou, line 744:
```suggestion
[0.19.2]: https://github.com/microsoft/CCF/releases/tag/ccf-0.19.2
[0.19.1]: https://github.com/microsoft/CCF/releases/tag/ccf-0.19.1
```

achamayou (approved):
I think it would be good to write some unit tests for the query parsing, to cover corner cases on separators etc.

eddyashton:
> I think it would be good to write some unit tests for the query parsing, to cover corner cases on separators etc.

I've added a unit test in `http_test.cpp`. This confirms the behaviour of a few corner cases, let me know if you'd like anything else tested.

PR #2311, opened 2021-03-12
Only build recent versions of docs

eddyashton:
~~The doc-building CI pipeline is broken this morning. I haven't found an immediate fix - in fact trying to repro locally I got a completely different error than the one seen on the CI. As an optimistic temporary fix, try pinning to an older version of Sphinx while I investigate further, maybe this unblocks other PRs.~~

We tracked the failure down to [this new release](https://github.com/pydata/pydata-sphinx-theme/releases/tag/v0.5.0), which isn't happy building our docs between versions `ccf-0.14.1` and `ccf-0.16.2`. We have a lot of very old versions that we don't want to support for ever, so taking this opportunity to cull some of the oldest and only build the docs since `ccf-0.16.3`. Its unfortunate that this needs to be expressed as a regex, but that's what we have. Please have a look and see if I've missed any tags (including expected/possible future tags!) that we want to be documented: regexr.com/5oe2d

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -194,7 +194,7 @@ breathe_default_project = "CCF"
 # Set up multiversion extension
 
 # Build tags from ccf-0.16.3 onwards
-smv_tag_whitelist = r"^ccf-(0\.(1([6-9].[3-9]|[7-9].*)|[2-9].*)|[1-9].*)$"
+smv_tag_whitelist = r"^ccf-(0\.(1([6-9]\.[3-9]|[7-9].*)|[2-9].*)|[1-9].*)$"
 smv_branch_whitelist = r"^main$"
 smv_remote_whitelist = None
 smv_outputdir_format = "{ref.name}"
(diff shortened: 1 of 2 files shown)


wintersteiger:
The failing build says

    WARNING: You are using pip version 20.2.3; however, version 21.0.1 is available.
    ...
    Running Sphinx v3.5.1
    ...

I just upgraded to 3.5.2 on my VM and it builds everything, but throws 6 warnings.

wintersteiger:
@achamayou did you forget to check in `doc/developers`?

eddyashton:
We've tracked the issue down to the new version of `pydata-sphinx-theme` (v0.5.0, released yesterday), which isn't happy building our `ccf-0.14.1` docs. Trying pinning that to an old version, but we may instead drop versions from our docs older than `ccf-0.15.0`

jumaffre (approved):

eddyashton:
Bonus: Build time has dropped from >19mins to 8mins.

PR #2312, opened 2021-03-12
Move some more headers into public include folder

eddyashton:
Opening as a draft for now, while I see if there's more changes worth putting in this PR. On the one hand I don't want too many PRs that rename headers, but I also want to make sure this is a manageable amount of work - twice this has approached the can-o-worms of "maybe we don't need multiple frontends anymore", but unpicking that is really thorny.

This PR just moves a few more headers that are definitely public to `include/ccf`. Additionally, it includes these as `ccf/` (to distinguish them from all the other private headers, which are not prefixed). The `ccf` CMake targets add both `/opt/ccf-vvv/include` and `/opt/ccf-vvv/include/ccf` as include directories so that these can be found when building against the install.

Additional refactorings, probably for a future PR:
- Distinguish the code for core node functionality from the frontend-endpoint dispatch machinery from the "utils for writing endpoints". Only the latter needs to be public.
- Remove `commit` and similar from the `Tx` type that's exposed to the app - it should only have `rw`, `ro`.
- Move the implementation of the definitely-public things (`EndpointRegistry` in particular) into `.cpp`, hopefully removing the need for most of the transitive private includes. This `.cpp` can be built into `ccf.enclave`/`ccf.virtual`.


CMakeLists.txt
```diff
@@ -77,9 +77,13 @@ if("sgx" IN_LIST COMPILE_TARGETS)
...
-    PUBLIC $<BUILD_INTERFACE:${CCF_GENERATED_DIR}>
-           $<INSTALL_INTERFACE:include/ccf/>
-           $<INSTALL_INTERFACE:include/3rdparty/>
+    PUBLIC
+      $<BUILD_INTERFACE:${CCF_GENERATED_DIR}>
+      $<INSTALL_INTERFACE:include/ccf/> #< This contains the private headers
+                                        #< which are currently under src, and
+                                        #< should be removed or renamed
+      $<INSTALL_INTERFACE:include/>
+      $<INSTALL_INTERFACE:include/3rdparty/>
```
achamayou, line 86: I think we would ideally split out the test (doctest, trompeloeil, picobench) and even the host-only 3party headers (CLI11) so they aren't installed.

include/ccf/historical_queries_interface.h
```diff
@@ -28,6 +30,27 @@ namespace ccf::historical
       path(p_),
       node_id(n_)
     {}
+
+    void describe_receipt(ccf::Receipt& r)
```
achamayou, line 34: Arguably describe(), since this is already a TxReceipt?

src/apps/smallbank/app/smallbank.cpp
```diff
@@ -452,29 +452,27 @@ namespace ccfapp
...
   public:
     SmallBank(kv::Store& store, AbstractNodeContext& context) :
-      UserRpcFrontend(store, sb_handlers),
+      RpcFrontend(store, sb_handlers),
       sb_handlers(context)
     {}
   };
 
-  std::shared_ptr<ccf::UserRpcFrontend> get_rpc_handler(
+  std::shared_ptr<ccf::RpcFrontend> get_rpc_handler(
```
achamayou, line 475: This needs a changelog entry

samples/apps/logging/logging_schema.h
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 #include "ds/json.h"
+#include "node/rpc/call_types.h" // For GetReceipt
```
achamayou, line 5: That seems less narrowly scoped than it could, why not ccf/receipt.h?

eddyashton:
Closing this PR as its superceded by a more thorough (though still incomplete) attempt in #2328.

samples/apps/logging/logging_schema.h
eddyashton: Absolutely - that's now done [here](https://github.com/microsoft/CCF/pull/2328/files#diff-35569e31aa8a27272198e8492dc20ef17a2731da8ff3bb337c44e500ca28471eR4).

include/ccf/historical_queries_interface.h
eddyashton: Sure, this is now called [describe()](https://github.com/microsoft/CCF/pull/2328/files#diff-ddb414d7c3e5af83a2abf8d566044b3f34577fd3428ed6229f2c29d13f0cfbdfR34). I wonder if we should rename one of these types, but haven't found better names. One is the public, serialisable representation of a `Receipt` you can use in your HTTP API, the other is the in-memory representation you can ask for validation. Another approach may be to flatten these, and just write a custom serialiser/API-descriptors for the latter, rather than producing a separate type. Not critical, but I may do this at some point.

CMakeLists.txt
eddyashton: Yes. I've added this as a TODO on #2328. It looks like CMake doesn't make this easy - the `REGEX`/`PATTERN` always installs the directories, even if they're empty, so this becomes quite a verbose description opting in every library we want to keep. There's also a few libraries that we _could_ hide as an implementation detail (`small_vector`, `sss`, `snmalloc`?) once our public includes are clean, but I won't try to unpick that right now.

achamayou: I would have thought we just create sub-directories like '3rdparty/export' and '3rdparty/internal'? I agree this isn't urgent by any means.

include/ccf/historical_queries_interface.h
achamayou: Yeah I wondered about that too, but equally did not find any good names. I think this is fine for now.

PR #2313, opened 2021-03-12
Auto-generated JS API docs & TS cleanup

letmaik:
Fixes https://github.com/microsoft/CCF/issues/1765.

This uses sphinx-js and typedoc to generate API docs for the built-in parts of CCF. It does not (yet?) document the utility TypeScript module that can optionally be used for dealing with type conversions of KV map keys/values.

sphinx-js' TypeScript support is basic and the project acknowledges that and doesn't intend to work on it (since Mozilla doesn't use TypeScript themselves). Any improvements typically come from community PRs. I tried to work around various issues but it's not as pretty as it could be.

I also took the opportunity to split the built-in from the utility parts and move them into separate modules. There are no "automatic" breaking changes for existing apps since the original `ccf.ts` module has to be vendored.

screenshot


.azure-pipelines-gh-pages.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           pip install wheel
           pip install -U -r doc/requirements.txt
           pip install -U -e ./python
+          npm install typescript typedoc@0.19.2
```
letmaik, line 37: sphinx-js doesn't support newer versions yet.

doc/conf.py
```diff
@@ -227,6 +228,10 @@
 tokenizer_lang = "en_UK"
 spelling_word_list_filename = ["spelling_wordlist.txt"]
 
+# sphinx_js options
+js_language = 'typescript'
+js_source_path = '../tests/npm-app/src/ccf'
```
letmaik, line 233: Not great, but also not *very bad*. Should be done differently in the future.

samples/apps/forum/src/ccf/util.ts
```diff
@@ -1,94 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
letmaik, line 1: Probably `util` is not the best name. Since these are all KV helpers, maybe something with kv in the name?

tests/npm-app/src/ccf/builtin.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  export type KVMaps = { [key: string]: KVMap };
+
+  export interface WrapAlgoParams {
+    name: string;
+  }
+
+  /**
+   * TODO
```
letmaik, line 120: @wintersteiger This is for you :)

doc/conf.py
wintersteiger: Can we move the sources out of the tests directory? Maybe simply `src/js`?

letmaik: Once this is distributed as a library on npm it can move to a separate folder, but for now let's keep it there, otherwise the JS sample apps are not copy-pasteable.

achamayou: We could move it to src/js and have a symlink in the samples.

samples/apps/forum/src/ccf/util.ts
achamayou: app.ts to match the C++ ccfapp namespace?

achamayou:
Two things that are unrelated to documentation but which this PR made obvious to me:

1. We seem to have taken a rather random view to capitalising acronyms, so we have RsaOaep-like things and AESKW-like things. I don't particularly care which one we go with (although the latter makes more sense to me), but it's important to be consistent.
2. All the crypto calls like generate key, wrap etc should be under `ccf.crypto`, and we should perhaps have `ccf.conv` for the conversion calls.

achamayou (approved):
Thanks you for setting this up, I expect it will make the JS/TS interface considerably easier to use in practice.

doc/conf.py
wintersteiger: Yeah, I like that too.

doc/build_apps/js_app_bundle.rst
```diff
@@ -12,7 +12,7 @@ This page documents the components of a bundle and the JavaScript API available
     `npm <https://www.npmjs.com/>`_, and
     `TypeScript <https://www.typescriptlang.org/>`_.
     CCF provides multiple example apps built with these tools.
-    They involve a `build` step that generates an app bundle suitable for CCF.
+    They involve a `build` step that generates an app bundle suitable for
```
jumaffre, line 15: Unfinished sentence?

doc/conf.py
letmaik: OK, let's do that.. if you then  `cp -rf` a sample you will still have a symlink to src/js which then takes effort to un-symlink.

samples/apps/forum/src/ccf/util.ts
letmaik: I don't think that's a good name. We may have more modules in the future and collectively they may be in the "ccfapp" namespace but modules != namespace.

doc/build_apps/js_app_bundle.rst
letmaik: Eh, yeah, I... fixed.

tests/npm-app/src/ccf/builtin.ts
letmaik: @wintersteiger I moved the sources to src/js. Do you want to fill in the jsdoc in this PR or in a separate one? I don't mind either way.

letmaik:
> 
> 
> Two things that are unrelated to documentation but which this PR made obvious to me:
> 
>     1. We seem to have taken a rather random view to capitalising acronyms, so we have RsaOaep-like things and AESKW-like things. I don't particularly care which one we go with (although the latter makes more sense to me), but it's important to be consistent.
> 
>     2. All the crypto calls like generate key, wrap etc should be under `ccf.crypto`, and we should perhaps have `ccf.conv` for the conversion calls.

Yep, I opened https://github.com/microsoft/CCF/issues/2315 for that.

tests/npm-app/src/ccf/builtin.ts
wintersteiger: Let's do it in a separate one, easier to review.

letmaik: https://github.com/microsoft/CCF/issues/2317

samples/apps/forum/src/ccf/util.ts
letmaik: Let's ignore this for now and think about it again when we make it into a library.

PR #2316, opened 2021-03-15
Fix component governance for QuickJS

wintersteiger:
This fixes the component governance scans, which report QuickJS as an unsupported component.


wintersteiger:
(Looks like something's broken on the other end, will retry later.)

achamayou:
What's the problem exactly? There's now a github repo, so we could switch to that if it helps.

wintersteiger:
That's what this PR does. The CG scanner throws an uninterpretable error, I suspect something's wrong with the upload of the scanner results.

wintersteiger:
Also, the removal of `#include "quickjs-exports.c"` upstream seems to have broken stuff.

achamayou:
@wintersteiger that's ours, it wasn't removed from upstream so much as it never was there :)

wintersteiger:
Oooh, I see, makes sense now. Are we tracking modifications to 3rdparty  somewhere?

achamayou:
@wintersteiger not really, there's very few. The pattern so far on update PRs has been to make the 3rdparty pull a separate commit from our changes, so the next upgrade can be done by cherry picking the second commit (sort of, it never works that well, but it's a start).

I'll do a review tomorrow, but I believe only quickjs has meaningful patches at this point, which are in those two separate files and just included across.

I've had a look at the rest of 3rdparty and we're up to date everywhere except for llhttp, which needs upgrading.

wintersteiger:
Argh, CG doesn't work on PRs from a fork.

PR #2318, opened 2021-03-15
Node to node channel improvements. Fixes #2092.

### LINKED ISSUE #2092: Node-to-node channels improvement
Further improvements to the node-to-node channel protocol:

- [x] 1. To prevent replay attacks of the authenticated key exchange protocol, each node should not only sign its own public ECDH key share but also the one of the other node. This implies adding an extra message as the initiator node doesn't have the other node's public key share at the start of the protocol.
- [x]  2. The shared secret (TLS's pre master key) should not directly be used as the AES GCM key but instead be used as the input of a KDF. The KDF should also take a label identifying the pair of nodes in that channel, and the encryption key used in one direction should be different from the one used for encryption in the opposite direction. 
- [x] 3. Encryption keys should be refreshed periodically (as per https://tools.ietf.org/html/rfc8446#section-5.5). The key refresh can be initiated by any node reaching that limit, and the other node should update its encryption key accordingly. 
- [x] 4. Nodes should sign the ECDH key shares with their own signing keys (as opposed to the service's) and attach their certificate, endorsed by the service, to the key establishment messages (relates to #408).

wintersteiger:
Draft for items 1 and 4 of #2092.

### DIFF
--- a/src/crypto/hash.h
+++ b/src/crypto/hash.h
@@ -2,9 +2,7 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
-#include "mbedtls/hash.h"
-
-#include "openssl/hash.h"
+#include "hash_provider.h"
 
 #define FMT_HEADER_ONLY
 #include <fmt/format.h>
@@ -23,19 +21,13 @@ namespace crypto
   /** Create a default incremental SHA256 hash provider */
   std::shared_ptr<ISha256Hash> make_incremental_sha256();
 
-  inline std::vector<uint8_t> hkdf(
+  /** Perform HKDF key derivation */
+  std::vector<uint8_t> hkdf(
     MDType md_type,
     size_t length,
     const std::vector<uint8_t>& ikm,
     const std::vector<uint8_t>& salt = {},
-    const std::vector<uint8_t>& info = {})
-  {
-#if defined(CRYPTO_PROVIDER_IS_MBEDTLS) && defined(INSIDE_ENCLAVE)
-    return mbedtls::hkdf(md_type, length, ikm, salt, info);
-#else
-    return OpenSSL::hkdf(md_type, length, ikm, salt, info);
-#endif
-  }
+    const std::vector<uint8_t>& info = {});
 }
 
 namespace fmt
--- a/src/crypto/hash.h
+++ b/src/crypto/hash.h
@@ -22,6 +22,20 @@ namespace crypto
 
   /** Create a default incremental SHA256 hash provider */
   std::shared_ptr<ISha256Hash> make_incremental_sha256();
+
+  inline std::vector<uint8_t> hkdf(
+    MDType md_type,
+    size_t length,
+    const std::vector<uint8_t>& ikm,
+    const std::vector<uint8_t>& salt = {},
+    const std::vector<uint8_t>& info = {})
+  {
+#if defined(CRYPTO_PROVIDER_IS_MBEDTLS) && defined(INSIDE_ENCLAVE)
+    return mbedtls::hkdf(md_type, length, ikm, salt, info);
+#else
+    return OpenSSL::hkdf(md_type, length, ikm, salt, info);
+#endif
+  }
 }
 
 namespace fmt
--- a/src/crypto/mbedtls/hash.cpp
+++ b/src/crypto/mbedtls/hash.cpp
@@ -9,8 +9,40 @@
 #include <mbedtls/sha256.h>
 #include <stdexcept>
 
+#ifdef INSIDE_ENCLAVE
+// Note: Older system packages of mbedTLS may not support HKDF.
+#  include <mbedtls/hkdf.h>
+#endif
+
 namespace crypto
 {
+  namespace mbedtls
+  {
+#ifdef INSIDE_ENCLAVE
+    std::vector<uint8_t> hkdf(
+      MDType md_type,
+      size_t length,
+      const std::vector<uint8_t>& ikm,
+      const std::vector<uint8_t>& salt,
+      const std::vector<uint8_t>& info)
+    {
+      auto md = mbedtls_md_info_from_type(get_md_type(md_type));
+      std::vector<uint8_t> okm(length);
+      int rc = mbedtls_hkdf(
+        md,
+        salt.data(),
+        salt.size(),
+        ikm.data(),
+        ikm.size(),
+        info.data(),
+        info.size(),
+        okm.data(),
+        okm.size());
+      return okm;
+    }
+#endif
+  }
+
   using namespace mbedtls;
 
   void mbedtls_sha256(const CBuffer& data, uint8_t* h)
--- a/src/crypto/mbedtls/hash.cpp
+++ b/src/crypto/mbedtls/hash.cpp
@@ -9,16 +9,14 @@
 #include <mbedtls/sha256.h>
 #include <stdexcept>
 
-#ifdef INSIDE_ENCLAVE
-// Note: Older system packages of mbedTLS may not support HKDF.
-#  include <mbedtls/hkdf.h>
-#endif
+// Note: Older system packages of mbedTLS may not support HKDF; if this include
+// file is not found, consider upgrading your mbedTLS.
+#include <mbedtls/hkdf.h>
 
 namespace crypto
 {
   namespace mbedtls
   {
-#ifdef INSIDE_ENCLAVE
     std::vector<uint8_t> hkdf(
       MDType md_type,
       size_t length,
@@ -40,7 +38,6 @@ namespace crypto
         okm.size());
       return okm;
     }
-#endif
   }
 
   using namespace mbedtls;
--- a/src/crypto/mbedtls/verifier.cpp
+++ b/src/crypto/mbedtls/verifier.cpp
@@ -105,4 +105,9 @@ namespace crypto
 
     return Pem(buf, len);
   }
+
+  bool Verifier_mbedTLS::validate_certificate(const Pem& ca_pem)
+  {
+    return true;
+  }
 }
--- a/src/crypto/mbedtls/verifier.cpp
+++ b/src/crypto/mbedtls/verifier.cpp
@@ -106,8 +106,22 @@ namespace crypto
     return Pem(buf, len);
   }
 
-  bool Verifier_mbedTLS::validate_certificate(const Pem& ca_pem)
+  bool Verifier_mbedTLS::verify_certificate(
+    const std::vector<const Pem*>& trusted_certs)
   {
+    // TODO
     return true;
+  }
+
+  bool Verifier_mbedTLS::is_self_signed() const
+  {
+    // TODO
+    return true;
+  }
+
+  std::string Verifier_mbedTLS::serial_number() const
+  {
+    // TODO
+    return "";
   }
 }
--- a/src/crypto/mbedtls/verifier.cpp
+++ b/src/crypto/mbedtls/verifier.cpp
@@ -7,6 +7,7 @@
 #include "rsa_key_pair.h"
 
 #include <mbedtls/pem.h>
+#include <mbedtls/x509_crt.h>
 
 namespace crypto
 {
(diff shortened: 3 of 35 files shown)


src/crypto/openssl/key_pair.cpp
```diff
@@ -44,10 +45,17 @@ namespace crypto
     std::vector<std::pair<std::string, std::string>> r;
```
achamayou, line 45: That looks almost entirely like http::parse_query() except for the separator and the raw pointers. Could we maybe parametrise the implementation of http::parse_query() on the separator and use it in both places?

src/node/channels.h
```diff
@@ -246,89 +243,293 @@ namespace ccf
...
-      auto data_ = serialised_signed_public.data();
-      serialized::write(data_, space, own_public.size());
-      serialized::write(data_, space, own_public.data(), own_public.size());
+      // Serialise channel key share, signature, and certificate and
+      // length-prefix them
+      auto space =
+        ks.size() + signature.size() + node_cert.size() + 3 * sizeof(size_t);
+      std::vector<uint8_t> serialised_signed_key_share(space);
+      auto data_ = serialised_signed_key_share.data();
+      serialized::write(data_, space, ks.size());
```
achamayou, line 265: For stuff whose length fits in 16bits, we have write_lps.

src/crypto/openssl/key_pair.cpp
wintersteiger: Can't find `parse_query`, do you mean `parse_delimited_string`?

src/node/channels.h
wintersteiger: There are no `uint16_t`s here and `write_lps` just assumes that the value fits in 16 bits without checking it. In practice those values will be small, but I'd have to add checks on all the vector sizes myself then.

src/crypto/openssl/key_pair.cpp
achamayou: Yes, I think it was called that in an earlier PR, but I'm obviously behind, sorry.

src/node/node_to_node.h
```diff
@@ -171,7 +171,8 @@ namespace ccf
 
     virtual void restart() override
     {
-      channels->restart();
+      if (channels)
+        channels->restart();
```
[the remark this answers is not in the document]
wintersteiger: Yeah, I'll take a pass over everything before merging it.

src/enclave/rpc_sessions.h
```diff
@@ -97,7 +97,7 @@ namespace enclave
       auto search = sessions.find(id);
       if (search == sessions.end())
       {
-        LOG_FAIL_FMT("Replying to unknown session {}", id);
+        LOG_FAIL_FMT("Refusing to reply to unknown session {}", id);
```
jumaffre, line 100: Unrelated to this PR: this should probably be a `LOG_DEBUG_FMT()` as this can happen often in a normal scenario (e.g. if a client disconnects before it has received the response to its forwarded request).

wintersteiger: Yeah, you're right, I'll make it a `_DEBUG_`.

src/node/node_state.h
```diff
@@ -1441,13 +1454,26 @@ namespace ccf
       // Accept TLS connections, presenting node certificate signed by network
       // certificate
 
-      auto nw = crypto::make_key_pair(network.identity->priv_key);
-      auto csr = node_sign_kp->create_csr(config.subject_name);
-      auto sans = get_subject_alternative_names();
-      auto endorsed_node_cert = nw->sign_csr(network.identity->cert, csr, sans);
+      if (node_cert.empty())
```
jumaffre, line 1457: I'm surprised that this is here. I thought this function was always called at the right times, i.e. always after the node certificate had been endorsed.

src/node/test/channels.cpp
```diff
@@ -112,49 +136,139 @@ auto read_node_msgs(ringbuffer::Circuit& circuit)
...
+  auto outbound_msgs = read_outbound_msgs<MsgType>(circuit);
+  REQUIRE(outbound_msgs.size() == 1);
+  auto msg = outbound_msgs[0];
+  const auto* data_ = msg.payload.data();
+  auto size_ = msg.payload.size();
+  REQUIRE(msg.type == msg_type);
+  return msg;
+}
+
+void hexdump(const char* hdr, const std::vector<uint8_t>& data)
```
jumaffre, line 151: FYI, there's a new `ds/hex.h` that should handle all conversions from/to hex strings.

src/node/test/channels.cpp
```diff
@@ -254,24 +368,115 @@ TEST_CASE("Client/Server key exchange")
   }
 }
 
+TEST_CASE("Switch to endorsed certificates")
```
jumaffre, line 371: I don't understand the use case for switching to endorsed certificate. The channel shouldn't be established until the node certificates of both peers have been endorsed. What am I missing?

src/tls/key_exchange.h
```diff
@@ -3,37 +3,30 @@
...
 
 namespace tls
 {
   class KeyExchangeContext
   {
   private:
     crypto::mbedtls::ECDHContext ctx = nullptr;
-    std::vector<uint8_t> own_public;
-    crypto::EntropyPtr entropy;
+    std::vector<uint8_t> key_share, peer_key_share;
```
jumaffre, line 23: Minor: I'd prefer the definition on two separate lines as it's more consistent with the existing code base, and less prone to initialisation errors.

src/tls/key_exchange.h
```diff
@@ -42,11 +35,15 @@ namespace tls
         throw std::logic_error(error_string(rc));
       }
 
+      crypto::EntropyPtr entropy = crypto::create_entropy();
```
jumaffre, line 38: Curious: why is the entropy created in two places now?

src/node/node_state.h
wintersteiger: This is all in flux, I haven't decided where/how to move the calls to `accept_network_tls_connections`. E.g. the version that's currently checked in erroneously re-generates already endorsed certificates.

src/node/test/channels.cpp
wintersteiger: Thanks, found that and started using it in other places.

src/node/test/channels.cpp
wintersteiger: Possible. I don't understand the logic for when it's okay to accept self-signed certificates. During the joining process, nodes seem to want to communicate with self-signed certificates until they are properly joined. Are you saying that they shouldn't do that?

src/tls/key_exchange.h
wintersteiger: Because it's an mbedTLS implementation detail. Once I add the OpenSSL version, there will be no entropy objects in that respective implementation and/or the base class.

wintersteiger:
1/4 of #2092 are working. Currently adding 2. We should have all the bits for key rotation (4), but it isn't currently triggered yet.

src/crypto/hash.h
```diff
@@ -22,6 +22,20 @@ namespace crypto
...
   std::shared_ptr<ISha256Hash> make_incremental_sha256();
+
+  inline std::vector<uint8_t> hkdf(
+    MDType md_type,
+    size_t length,
+    const std::vector<uint8_t>& ikm,
+    const std::vector<uint8_t>& salt = {},
+    const std::vector<uint8_t>& info = {})
+  {
+#if defined(CRYPTO_PROVIDER_IS_MBEDTLS) && defined(INSIDE_ENCLAVE)
```
jumaffre, line 33: As discussed, I believe `&& defined(INSIDE_ENCLAVE)` should go before merging this PR?

src/crypto/mbedtls/hash.cpp
```diff
@@ -9,8 +9,40 @@
 #include <mbedtls/sha256.h>
 #include <stdexcept>
 
+#ifdef INSIDE_ENCLAVE
+// Note: Older system packages of mbedTLS may not support HKDF.
+#  include <mbedtls/hkdf.h>
+#endif
```
jumaffre, line 15: Ditto, I think this should go.

src/crypto/mbedtls/verifier.cpp
```diff
@@ -105,4 +106,64 @@ namespace crypto
...
+        mbedtls_x509_crt_init(&chain[i]);
+      }
+
+      for (size_t i = 0; i < trusted_certs.size(); i++)
+      {
+        auto& tc = trusted_certs[i];
+        if (
+          (rc = mbedtls_x509_crt_parse(&chain[i], tc->data(), tc->size())) != 0)
+        {
+          goto exit;
```
jumaffre, line 132: Could we avoid the C-style `goto` and `exit` label and use the mbedtls wrappers if possible?

src/crypto/hash.h
wintersteiger: It would work on the CI machines, but necessarily on all dev VMs. I'll remove it and install a new one on my VM and if we get complaints that it doesn't build just send them my way :)

src/node/channels.h
```diff
@@ -186,19 +189,28 @@ namespace ccf
     {
       RINGBUFFER_WRITE_MESSAGE(
         ccf::add_node, to_host, peer_id.value(), peer_hostname, peer_service);
+      auto e = crypto::create_entropy();
+      hkdf_salt = e->random(32);
```
jumaffre, line 193: Can `32` be a named `static constexpr` in this file?

src/node/channels.h
```diff
@@ -246,93 +258,346 @@ namespace ccf
...
-      // Serialise channel public and network signature and length-prefix both
-      auto space = own_public.size() + signature.size() + 2 * sizeof(size_t);
-      std::vector<uint8_t> serialised_signed_public(space);
-      auto data_ = serialised_signed_public.data();
-      serialized::write(data_, space, own_public.size());
-      serialized::write(data_, space, own_public.data(), own_public.size());
+      // Serialise channel key share, signature, and certificate and
+      // length-prefix them
+      auto space =
+        ks.size() + signature.size() + node_cert.size() + +4 * sizeof(size_t);
```
jumaffre, line 278:
```suggestion
        ks.size() + signature.size() + node_cert.size() + 4 * sizeof(size_t);
```

src/crypto/mbedtls/verifier.cpp
wintersteiger: Not easy; mbedTLS expects a buffer full of it's own data structures, it wouldn't work if we need more space per object in the buffer. In theory we could add a new wrapper for this type of set, but this is the only place we need it.

src/node/channels.h
```diff
@@ -246,93 +258,346 @@ namespace ccf
...
+    bool consume_responder_key_share(const std::vector<uint8_t>& data)
+    {
+      return consume_responder_key_share(data.data(), data.size());
+    }
 
-      auto signature_size = serialized::read<size_t>(data, size);
-      auto signature_start = data;
+    bool consume_responder_key_share(const uint8_t* data, size_t size)
+    {
+      LOG_DEBUG_FMT("status == {}", status);
```
jumaffre, line 391: Can we make this, and all other log messages in this file (except the channel initiation and establishment) `LOG_TRACE_FMT`?

src/node/channels.h
```diff
@@ -246,93 +258,346 @@ namespace ccf
...
+        LOG_FAIL_FMT("{} exccess bytes remaining", size);
         return false;
       }
 
-      if (!network_pubk->verify(
-            peer_public_start,
-            peer_public_size,
-            signature_start,
-            signature_size))
+      if (peer_version != protocol_version || ks.n == 0 || sig.n == 0)
```
jumaffre, line 417: It might be worth having a specific `LOG_FAIL` message if the protocols mismatch?

src/node/channels.h
```diff
@@ -514,6 +837,14 @@ namespace ccf
       }
     }
 
+    void restart()
```
jumaffre, line 840: I cannot find where this is called?

src/node/test/channels.cpp
```diff
@@ -40,6 +47,21 @@ struct NodeOutboundMsg
...
+    std::vector<uint8_t> r;
+    r.insert(r.end(), authenticated_hdr.begin(), authenticated_hdr.end());
+    r.insert(r.end(), payload.begin(), payload.end());
+    return r;
+  }
+
+  std::vector<uint8_t> unauthenticated_data() const
+  {
+    std::vector<uint8_t> r = data();
+    r.erase(r.begin(), r.begin() + 8); // Skip 8 bytes of message type
```
jumaffre, line 62: As discussed, this is because of the `ChannelMsg` enum included in each node-to-node channel. Can we read it more cleanly?

jumaffre (approved):

src/crypto/hash.h
jumaffre: We tell contributors to use the playbooks in this repo to setup their development environment (see https://microsoft.github.io/CCF/main/build_apps/build_setup.html), which installs a recent-enough version of mbedtls so I think we're good :)

src/node/channels.h
wintersteiger: Only the channels test so far, I'm adding the logic for triggering it on the message counter now.

src/node/channels.h
```diff
@@ -246,93 +258,346 @@ namespace ccf
...
+        label = self.value() + peer_id.value();
+      }
+      else
+      {
+        label = peer_id.value() + self.value();
+      }
+      std::vector<uint8_t> info = {label.data(), label.data() + label.size()};
+      LOG_DEBUG_FMT("salt={}", ds::to_hex(hkdf_salt));
+      auto key_bytes = crypto::hkdf(
+        crypto::MDType::SHA256, 32, shared_secret, hkdf_salt, info);
```
jumaffre, line 597: Could `32` be defined somewhere?

PR #2319, opened 2021-03-15
TPCC benchmark

ashamis:
Adding the TPCC benchmark ==> http://www.tpc.org/tpcc/
This is a port of https://github.com/evanj/tpccbench.

There are still some missing bits where we need to find the last row in a couple of tables. To do this correctly we need to implement a table index which I will do in a future PR.


wintersteiger:
Should this really go into the main CCF repo? AFAICT this is an application that may use CCF, but not something that CCF depends on, right?

ashamis:
> Should this really go into the main CCF repo? AFAICT this is an application that may use CCF, but not something that CCF depends on, right?

I am happy to put this elsewhere but this seems to be a better perf benchmark then smallbank. I would argue that we keep this in the main repo and move smallbank out if we want to do that.

Regardless, of this PR we should form a policy on how we add such benchmarks and future samples.

wintersteiger:
Yeah, that's a good point, we should think about that. I remember that, at some point, there was a second repo, not sure what became of that.

CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ option(ZAP_TEST
        "ZAP fuzz test using https://www.zaproxy.org/docs/docker/api-scan/" OFF
 )
 option(BUILD_SMALLBANK "Build SmallBank sample app and clients" ON)
+option(BUILD_TPCC "Build tppc sample app and clients" ON)
 
 # Build common library for CCF enclaves
```
achamayou, line 61: So we are building smallbank by default, and we have to have a -DBUILD_SMALLBANK=OFF in our CI Debug config (matrix.yml) as a result. We can either do the same for TPCC, or default both to OFF and add a flag to enable them in the Perf job. I think I mildly prefer the latter, but so long as we do one of those I think I'm happy.

achamayou (approved):

PR #2322, opened 2021-03-15
Missing locks

achamayou:
Spotted by @jumaffre earlier, the first instance seems to have happened accidentally in the move to async execution, where we forgot to lower the lock for recv_request_vote_response specifically (it was done correctly on other message types).


ashamis:
you should probably run the threading tests as part of this PR

ashamis (approved):

jumaffre (approved):

PR #2323, opened 2021-03-15
PascalCase enum members in REST API responses

### LINKED ISSUE #2152: PascalCase enum members for endpoint responses
The `State` enum members are lowerCamelCase. Other enums are PascalCase (e.g. `TxStatus`). The JSON serialization converts some of the enums to UPPERCASE via custom functions.

The Microsoft REST API guidelines say that enum members should be lowerCamelCase. I suggest either standardizing on lowerCameCase in the code (even though it looks a bit funny) and removing any custom conversion functions, or using PascalCase in the code and adding/changing conversion functions everywhere to make them lowerCamelCase in endpoint responses.

This also has an effect on enum members used as part of a request, e.g. `status` query param in /network/nodes, so it has to work both directions.

letmaik:
Fixes #2152.


src/node/rpc/serialization.h
```diff
@@ -11,14 +11,14 @@ namespace ccf
...
     ccf::State,
-    {{ccf::State::uninitialized, "uninitialized"},
-     {ccf::State::initialized, "initialized"},
-     {ccf::State::pending, "pending"},
-     {ccf::State::partOfPublicNetwork, "partOfPublicNetwork"},
-     {ccf::State::partOfNetwork, "partOfNetwork"},
-     {ccf::State::readingPublicLedger, "readingPublicLedger"},
-     {ccf::State::readingPrivateLedger, "readingPrivateLedger"},
-     {ccf::State::verifyingSnapshot, "verifyingSnapshot"}})
+    {{ccf::State::uninitialized, "Uninitialized"},
```
ashamis, line 14: should this be ccf::State::UNINITIALIZED, "Uninitialized"?

src/node/rpc/tx_status.h
```diff
@@ -36,19 +36,19 @@ namespace ccf
     {
       case TxStatus::Unknown:
```
ashamis, line 37: should Unknown be uppercase?

src/node/rpc/serialization.h
letmaik: I'm not changing any C++ enum names in this PR, just serialization. But yes, this should eventually be changed.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- Enum values returned by built-in REST API endpoints are now PascalCase (#2152).
```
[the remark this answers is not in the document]
letmaik: It's actually the same, only the values are updated, but the enum names are the same.

src/runtime_config/gov.lua
```diff
@@ -10,7 +10,7 @@ return {
   PASSED = 1
   PENDING = 0
   REJECTED = -1
-  STATE_ACTIVE = "ACTIVE"
```
letmaik, line 13: Reviewers, take note!

tests/infra/network.py
```diff
@@ -286,7 +286,7 @@ def _start_all_nodes(
                         )
                         self.wait_for_state(
                             node,
-                            "partOfPublicNetwork",
+                            "PartOfPublicNetwork",
```
eddyashton, line 289: While we're here: we should make this a Python enum defining all of the possible values, rather than a string comparison.

eddyashton (approved):

tests/infra/proposal.py
```diff
@@ -18,11 +18,11 @@ def __init__(self, proposal):
...
-    Open = "OPEN"
-    Accepted = "ACCEPTED"
-    Withdrawn = "WITHDRAWN"
-    Rejected = "REJECTED"
-    Failed = "FAILED"
+    OPEN = "Open"
+    ACCEPTED = "Accepted"
+    WITHDRAWN = "Withdrawn"
+    REJECTED = "Rejected"
+    FAILED = "Failed"
```
eddyashton, line 25: None of the other Python enum names have changed, only this one. If we're not touching the code names, only the values, the left hands here should be reverted.

letmaik: Hm, I didn't want to change C++ enums but Python test code seems ok? Is there anything that could break?

eddyashton: Its harmless, just seems odd to do a partial consistency fix so I assume it was a mistake.

letmaik: I fixed the ones that are in sync with C++. There are some local ones which I didn't touch.

PR #2327, opened 2021-03-17
Python infra: fix MemberStatus enum parsing

letmaik:
Fixes daily build which [is broken](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=20791&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=fe98d859-2537-5570-75af-de67d4290f3a)

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Happy daily trigger :)
+Happy daily trigger :))
--- a/tests/infra/consortium.py
+++ b/tests/infra/consortium.py
@@ -99,7 +99,7 @@ class Consortium:
                     member = self.get_member_by_service_id(member_service_id)
                     if member:
                         if (
-                            infra.member.MemberStatus[status]
+                            infra.member.MemberStatus(status)
                             == infra.member.MemberStatus.ACTIVE
                         ):
                             member.set_active()


jumaffre (approved):

jumaffre:
You may want to touch the https://github.com/microsoft/CCF/blob/main/.daily_canary file to make sure the Daily is indeed fixed before merging this.

achamayou (approved):

achamayou:
Modifying the Daily Build Status job to grep the PR title/branch name for "[Fix daily]"/fix_daily and pass automatically when that's the case to avoid having to use the red button to fix the daily build: Crazy or not?

eddyashton:
> Modifying the Daily Build Status job to grep the PR title/branch name for "[Fix daily]"/fix_daily and pass automatically when that's the case to avoid having to use the red button to fix the daily build: Crazy or not?

Could we at least combine this with a check that the daily canary is modified? So Daily Status passes if (Daily CI is green) OR ((this PR says its fixing the Daily) AND (this PR will be gated by its own Daily run))?

eddyashton (approved):

achamayou:
@eddyashton I like that!

Issue #697, opened 2020-01-16
Headers should be split into include and private folders


eddyashton:
As a first step towards minimising our public headers, and to make an installable package, we should start declaring some of our headers as private.

achamayou:
Related to #66

achamayou:
- [ ] 3rdparty/(internal|external|test)

PR #2328, opened 2021-03-17
Define more headers as public, and split endpoints into separate library

eddyashton:
This is a more-complete replacement for #2312.

This PR makes it so that public headers are included as `ccf/` both locally and against the install, to distinguish them from the private headers (which are still installed, for now).

Additionally, it creates a `ccf_endpoints` library to break some of the include chains from app code into our implementation details, with the nice benefit of massively reducing build times for some kinds of change. I think its worth repeating this pattern for other parts of the code but I'm not sure what the next target is - I'd love to do it for the kv but not sure it'll work due to the heavy templating.

TODO:
- [x] Write a CHANGELOG
- [x] Trim the 3rdparty dependencies that don't need to be public ([here](https://github.com/microsoft/CCF/pull/2328/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR175-R194))


samples/apps/logging/logging.cpp
```diff
@@ -1,11 +1,15 @@
...
-#include "node/historical_queries_adapter.h"
-#include "node/quote.h"
-#include "node/rpc/metrics_tracker.h"
+
+// CCF
+#include "apps/utils/metrics_tracker.h"
+#include "ccf/app_interface.h"
+#include "ccf/historical_queries_adapter.h"
+#include "http/http_query.h"
 #include "node/rpc/user_frontend.h"
```
eddyashton, line 13: I think the core of "what are we trying to do" here, is to reach a point where all of the CCF includes here (and in other apps) are of public `ccf/` files. We can make an exception for some shared utils that several of our apps use, like `metrics_tracker.h`. The next step after that, to stop installing our private headers, is to add enough forward declaration/hidden implementation that the public headers also only include public headers, though that's not so urgent.

CMakeLists.txt
```diff
@@ -117,7 +117,8 @@ endif()
 
 if("virtual" IN_LIST COMPILE_TARGETS)
   # virtual version
-  add_library(ccf.virtual STATIC ${CCF_IMPL_SOURCE})
+  add_library(ccf.virtual STATIC ${CCF_IMPL_SOURCE}
```
achamayou, line 120: is that really cmake_format?

eddyashton:
This is part of #697. This doesn't yet include the split of `3rdparty` into multiple folders - I have that on a separate branch but would prefer not to pollute this PR (it is almost purely a file rename, so easier to read in isolation).

cmake/common.cmake
```diff
@@ -584,8 +608,6 @@ function(add_picobench name)
 
   add_executable(${name} ${PARSED_ARGS_SRCS})
 
-  add_lvi_mitigations(${name})
```
jumaffre, line 587: I'm assuming that removing this causes the performance improvement for all our micro benchmarks?

samples/apps/logging/logging.cpp
```diff
@@ -164,7 +168,7 @@ namespace loggingapp
         .install();
 
       // SNIPPET_START: get
-      auto get = [this](ccf::ReadOnlyEndpointContext& args, nlohmann::json&&) {
+      auto get = [this](auto& args, nlohmann::json&&) {
```
jumaffre, line 171: Nitpick: Some of these are called `args`, others `ctx`. Should we unify?

achamayou: That's a great post-1.0 task #2345

achamayou (approved):

PR #2329, opened 2021-03-17
Upgrade (indirect dependency on) pygments

wintersteiger:


achamayou:
If we could exclude this from the scope of the CG, that'd be better. This isn't part of our released artifacts, it's only used for analysis of perf test results.

achamayou (approved):

wintersteiger:
Yes, I agree! What's the best way to determine whether something's included in the released artifacts? CMake install targets maybe?

Issue #1401, opened 2020-07-13
Some KV tables are forever growing


jumaffre:
Some KV tables, e.g. `ccf.nodes`, `ccf.members`, grow as more actors are added to the service (nodes, members, etc.) but are not freed when these actors are retired (actor is only marked as `RETIRED`). This gives a live CCF service auditing properties (e.g. "which nodes have already been retired?") at the cost of memory growth inside the enclave.

achamayou:
Addressing this will require converting some queries, such as the list of proposal, to historical queries.

jumaffre:
Once this is done, `POST /gov/proposals/{proposal_id}/withdraw` can be turned into `DELETE /gov/proposals/{proposal_id}.

jumaffre:
Closing this issue as remaining work is covered by other issues or WIP PRs.

- Users and members are now removed from the store when they are retired (in fact, the concept of retiring a member/user no longer exists) - #2320.
- For retired proposals, the work will be done as part of the new JS governance work (w/ the two proposals and proposals info tables) - #2286 
- For nodes, the work will be covered in #2320, post 1.0.

PR #2330, opened 2021-03-19
Remove retired members from the store

jumaffre:
Part of #1401 

Retired members are now removed the store instead of being marked as `Retired`. We also remove all their associated data in the store, including their governance history.

The renamed `remove_member` proposal is also idempotent.


doc/governance/accept_recovery.rst
```diff
@@ -77,7 +77,7 @@ When the recovery threshold is reached, the ``POST recovery_share`` RPC returns
 
 Once the recovery of the private ledger is complete on a quorum of nodes that have joined the new network, the ledger is fully recovered and users are able to continue issuing business transactions.
 
-.. note:: Recovery shares are updated every time a new recovery member is added or retired and when the ledger is rekeyed. It also possible for members to update the recovery shares via the ``update_recovery_shares`` proposal.
+.. note:: Recovery shares are updated every time a new recovery member is added or remove and when the ledger is rekeyed. It also possible for members to update the recovery shares via the ``update_recovery_shares`` proposal.
```
achamayou, line 80:
```suggestion
.. note:: Recovery shares are updated every time a new recovery member is added or removed and when the ledger is rekeyed. It also possible for members to update the recovery shares via the ``update_recovery_shares`` proposal.
```

python/ccf/clients.py
```diff
@@ -181,7 +181,7 @@ class Response:
 
     def __str__(self):
         versioned = (self.view, self.seqno) != (None, None)
-        status_color = "red" if self.status_code / 100 in (4, 5) else "green"
+        status_color = "red" if int(self.status_code / 100) in (4, 5) else "green"
```
achamayou, line 184: Is status code now a string?

achamayou (approved):

python/ccf/clients.py
eddyashton: I think this is just
```suggestion
        status_color = "red" if self.status_code // 100 in (4, 5) else "green"
```

jumaffre: This used to be a float division, which only worked for `400` and `500`. I didn't know about `//` which is neater!

PR #2331, opened 2021-03-19
npm package & updated js docs

letmaik:
This PR moves the old TypeScript helpers into an actual npm package called `ccf-app` that can be consumed when writing CCF JS apps (works with and without TypeScript). It also contains a `polyfill` module that provides Node.js implementations for most of the global `ccf` object functionality. This allows to write unit tests for an app which are run in Node.js instead of CCF. The package itself has unit tests for the polyfill and the higher-level helpers. The latter are tested by relying on the polyfill.

API docs for the package are done with typedoc instead of sphinx-js due to limitations of its TypeScript support. New roles for sphinx have been added to cross-link to the typedoc docs. The default typedoc theme has been extended slightly to add a version selector and a link back to the main CCF docs (the mechanism to inject those two things is quite hacky but self-contained without duplicating existing templates -- typedoc is in the process of improving the theme system to allow for easier modifications, at which point we should make use of it). 
PREVIEW: https://letmaik.github.io/ccf-docs-staging/main/js/ccf-app/

The package is published as tarball in GitHub Releases in addition to npmjs.org. 

The forum sample has been cleaned up and in particular the JWT auth now uses the native "jwt" policy from CCF.


doc/conf.py
```diff
@@ -249,6 +249,11 @@ def setup(self):
...
     jsdoc_config_path = str(srcdir / jsdoc_config_path)
 
-    # disable sphinx-js for old ccf versions
     if not os.path.exists(jsdoc_config_path):
-        jsdoc_config_path = None
+        # try old location
+        js_source_path = str(srcdir / "../src/js")
+        jsdoc_config_path = str(srcdir / "../src/js/tsconfig.json")
+        
+        # disable sphinx-js for old ccf versions
```
achamayou, line 257:
```suggestion
        # disable sphinx-js for old ccf versions (prior to 0.19.4 ???)
```

Will be useful to know when we can remove this.

achamayou (approved):

doc/build_apps/js_app_bundle.rst
```diff
@@ -133,9 +133,6 @@ It is an object implementing the :js:class:`CCF <CCF>` interface documented belo
 .. js:autoclass:: ProofElement
    :members:
 
-.. js:autoclass:: WrapAlgoParams
```
letmaik, line 136: This is now a union `type` instead of `interface`, and sphinx-js doesn't support those.

.azure-pipelines-templates/release.yml
```diff
@@ -40,3 +40,8 @@ jobs:
           twine upload -u __token__ -p $(pypi_token) $(Build.ArtifactStagingDirectory)/**/*.whl
         workingDirectory: python
         displayName: PyPi https://pypi.org/project/ccf/
+
+      - script: |
+          npm publish $(Build.ArtifactStagingDirectory)/*.tgz
```
letmaik, line 45: TBD

doc/build_apps/js_app_bundle.rst
achamayou: So that can revert now presumably?

letmaik: Hm? Union is the correct type for this. I think it's fine as it is.

achamayou: Sorry, mis-read that first comment, ignore me.

tests/npm-app/package.json
```diff
@@ -1,11 +1,11 @@
...
   "scripts": {
-    "build": "del-cli -f dist/ && rollup --config && cp app.json dist/",
-    "demo": "npm run build && http-server -o demo/index.html"
+    "build": "del-cli -f dist/ && rollup --config && cp app.json dist/"
   },
   "type": "module",
   "dependencies": {
+    "@microsoft/ccf-app": "file:../../js/ccf-app",
     "js-base64": "^3.5.2",
     "jsrsasign": "^10.0.4",
```
wintersteiger, line 10: Do we really need jsrsasign? Feels like it should be easy to re-route into C++ and we would get rid of a dependency with recent CVEs.

wintersteiger (approved):
I haven't checked all the details, but I totally approve of this!

doc/build_apps/js_app_bundle.rst
```diff
@@ -113,51 +108,15 @@ JavaScript provides a set of built-in
 `global functions, objects, and values <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects>`_.
 
 CCF provides the additional global variable ``ccf`` to access native CCF functionality.
-It is an object implementing the :js:class:`CCF <CCF>` interface documented below.
+It is an object implementing the :typedoc:interface:`CCF <ccf-app/global/CCF>` interface.
```
achamayou, line 111: This makes me inexplicably happy

js/ccf-app/README.md
```diff
@@ -0,0 +1,9 @@
+# ccf-app
+
+ccf-app is a support package for developing CCF JavaScript apps. It is published as [@microsoft/ccf-app](https://www.npmjs.com/package/@microsoft/ccf-app) on npm.
```
achamayou, line 3:
```suggestion
ccf-app is a support package for developing [CCF]( https://github.com/microsoft/CCF) JavaScript or TypeScript apps. It is published as [@microsoft/ccf-app](https://www.npmjs.com/package/@microsoft/ccf-app) on npm.
```

achamayou (approved):

tests/npm-app/package.json
letmaik: The dependencies of the npm-app test are really just a test to see whether they work in CCF. I don't think CCF will ever expose all of the functionality of jsrsasign, so including it in the test is still useful as it's a popular library that people may use, and we don't want to accidently do something that breaks it. I think in the long run there should be a more dedicated test suite to test a large number of popular npm packages.

Issue #1791, opened 2020-10-21
Idempotent proposal functions


lynshi:
**Is your feature request related to a problem? Please describe.**
Proposal functions are not idempotent, and this can cause errors if a proposal is submitted to create a state that already exists. For example, [`open_network`](https://github.com/microsoft/CCF/blob/master/src/node/rpc/member_frontend.h#L636) calls [`open_service`](https://github.com/microsoft/CCF/blob/master/src/node/genesis_gen.h#L296), which returns `false` if the network is already open, which causes `open_network` to indicate the proposal failed. Similarly, [`new_user`](https://github.com/microsoft/CCF/blob/master/src/node/rpc/member_frontend.h#L490) calls [`add_user`](https://github.com/microsoft/CCF/blob/master/src/node/genesis_gen.h#L209), which fails if the user certificate already exists.

This is problematic because errors for a resubmitted proposal will cause the client's backoff code to retry proposals unnecessarily and eventually fail despite the fact that the desired state is already achieved. In a multi-member consortium, it's also possible that proposals with the same effect are accepted and executed concurrently, in which case the later-to-execute proposal may fail though its intended effect has been applied.

**Describe the solution you'd like**
Proposal functions hardcoded in C++ should be idempotent so that they can run successfully if accepted by members despite having no effect. This avoids requiring additional logic in the constitution, proposal, or vote to protect against misleading proposal failures.

**Additional context**
As discussed with @achamayou and @eddyashton, this issue can be mitigated initially with a side-by-side introduction of idempotent functions equivalent to existing proposal functions, such as `user_present` for adding a user.

achamayou:
We will introduce idempotent variants of the calls that exist now, and phase out the old ones over time.

achamayou:
Link #2169

jumaffre:
Write-up on how idempotence should be implemented: https://github.com/microsoft/CCF/pull/2332#discussion_r598889381

achamayou:
Done in JS governance.

PR #2332, opened 2021-03-19
Unify open network and accept recovery proposals

jumaffre:
Part of https://github.com/microsoft/CCF/issues/1791

The `open_network` and `accept_recovery` proposals, which were historically separate, are now unified into a single `transition_service_to_open` idempotent proposal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-- `accept_recovery` and `open_network` proposals have been merged into a single `transition_service_to_open`.
+- `accept_recovery` and `open_network` proposals have been merged into a single idempotent `transition_service_to_open` proposal.
 
 ## [0.19.2]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- `accept_recovery` and `open_network` proposals have been merged into a single `transition_service_to_open`.
+
 ## [0.19.2]
 
 ### Added
--- a/src/node/genesis_gen.h
+++ b/src/node/genesis_gen.h
@@ -351,6 +351,12 @@ namespace ccf
         return false;
       }
 
+      if (active_service->status == ServiceStatus::OPEN)
+      {
+        // If the service is already open, return with no effect
+        return true;
+      }
+
       if (
         active_service->status != ServiceStatus::OPENING &&
         active_service->status != ServiceStatus::WAITING_FOR_RECOVERY_SHARES)
--- a/src/node/rpc/member_frontend.h
+++ b/src/node/rpc/member_frontend.h
@@ -880,7 +880,7 @@ namespace ccf
              this->network.node_code_ids,
              proposal_id);
          }},
-        {"transition_network_to_open",
+        {"transition_service_to_open",
          [this](
            const ProposalId& proposal_id, kv::Tx& tx, const nlohmann::json&) {
            auto service = tx.ro<Service>(Tables::SERVICE)->get(0);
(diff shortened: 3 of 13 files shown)


src/node/rpc/member_frontend.h
```diff
@@ -1110,8 +1131,21 @@ namespace ccf
...
-      // if the vote was successful, update the proposal's state
-      proposal.state = ProposalState::ACCEPTED;
+      if (!unknown_call.has_value())
+      {
+        // if the vote was successful, update the proposal's state
+        proposal.state = ProposalState::ACCEPTED;
+      }
+      else
+      {
+        // If any function in the proposal is unknown, mark the proposal as
```
jumaffre, line 1141: I noticed that proposals which contained unknown function calls were still marked as `Accepted` on completion. I'm assuming that this will be fixed by the JS governance changes (proposals verified on creation?) but I've fixed it here in the meantime.

achamayou (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
```
eddyashton, line 8: Can you please call this `0.19.4` (and add the link to its not-yet-existing release notes at the bottom)? Then we can cut a release whenever we like, without an additional push to `main` to give this the correct name. If we're wrong about the name of the next release (and we decide its actually `0.20.0`) then we'll still need a renaming commit, but its no worse than if we'd had `Unreleased`.

src/node/genesis_gen.h
```diff
@@ -351,6 +351,12 @@ namespace ccf
         return false;
       }
 
+      if (active_service->status == ServiceStatus::OPEN)
+      {
+        // If the service is already open, return with no effect
+        return true;
+      }
```
eddyashton, line 358: As a general principle for making these proposals idempotent I think we should prefer "do it again", rather than "early out" (ie - adding `ServiceState::OPEN` as one of the acceptable states in the `if` below). I see 3 main arguments for this:
- Simplifying audit. If a proposal says `do_foo` and I expect that to write `did: FOO` to the ledger, then that write should always be present in a transaction which passes a `do_foo` transaction. I shouldn't need to track the invariants against existing state to reproduce the service's write logic, I should wherever possible be able to reason about these writes as atoms.
- Supporting re-application across code versions. If we change this proposal to _also_ write `magic_flag: 42` to the KV, on line 370, then it looks like that magic flag is set in every service which has opened. But services which migrate from the old code version to the new won't have this, and may have no way to set it. Rather than handling this manually, or with a separate proposal, the neatest solution seems to be to allow the members to "open the service again", and trust that always applies the latest invariants for an open service. This is harder where it might involve merging with current state, but that can be handled case-by-case.
- Consistent tracking of write conflicts. Early outs remove some of our interaction with the KV, so that the retry semantics of the original and re-applied proposal are different. I don't have a full story for the transaction-logic bug this introduces, but there's enough of a smell that I expect its there (or at least - I know there's something unknown there).

eddyashton: We discussed this offline, and there's some major arguments in favour of the "idempotence means some proposals are successful nops" approach:
- Smaller ledger. If you're deploying through some abstract infrastructure that ends up spamming your proposals, its safe because they're idempotent, and better if they don't produce spam in the ledger (maybe they can even become complete nops, not writing to the proposals table if they instantly pass and have no effect, we shouldn't rule this possibility out).
- Moving the complexity elsewhere. If we re-apply changes, then every hook that expects to be called once needs to be robust to this.
- Precondition checks need to be precise in either case. Most of the second case above boils down to being able to distinguish "have I done this before", and what that means across multiple code versions with different effects. Avoiding writes may be safer than always writing, since the bad outcome is "can't do the thing I wanted" rather than "produced inconsistent state", until the constitution is updated/fixed.
- The example as written _would_ produce a conflict, because A read `service_status` to determine whether it had already been applied. A similar hole could occur if the reads and writes were not aligned - by reading `X` I decide not to write `Y`, and thus miss a conflict with `Y`. But this devolves to the "need to check your preconditions precisely" noted above.

In summary we'd prefer the early-out approach for current use cases, and if we find a service that would prefer re-application we'll need to provide it in open form, without ruling out the optimisations possible from the nop approach.

jumaffre: @eddyashton Thanks for the write-up! I've linked it in the parent issue: https://github.com/microsoft/CCF/issues/1791#issuecomment-804820088

PR #2336, opened 2021-03-22
Fix and test node status filter

achamayou:
@eddyashton probably has a better way to fix the string -> enum conversion.


eddyashton:
I have an alternative fix in #2338 (reusing the existing JSON parsing rather than requiring these hardcoded strings to stay in-sync). I've included your modifications to the e2e test in that PR, so I'm closing this as redundant.

PR #2343, opened 2021-03-23
Remove msgpack from framework - Part I

jumaffre:
Edit: Draft for now as I'm investigating the performance regression for the KV benchmarks.

This PR contains the first part of the removal of `msgpack-c` from the framework code.

The main change is the serialisation of the ledger frames which now uses `ds::serialized` functions instead of `msgpack-c` (with related changes in `ledger.py`). The node configuration passed in the initial ECALL is now also serialised to JSON. I have also removed most of the `MSGPACK_DEFINE` macros from the service table definitions.

Next:
- Actually remove `msgpack-c` from the framework code (still used for the serialisation of the default `kv::Map`).
- Have one of our sample apps (TPPC or SmallBank) use `msgpack` (both for performance but also as a sample app).


src/kv/raw_serialise.h
```diff
@@ -0,0 +1,195 @@
...
+
+namespace kv
+{
+  class RawWriter
+  {
+  private:
+    using WriterData = std::vector<uint8_t>;
+
+    // Number of bytes to reserve initially for storage of transaction frame
+    static constexpr size_t initial_reserve_size = 72;
```
achamayou, line 21: How did we get to this value?

achamayou (approved):

achamayou:
@jumaffre there seems to be substantial hit to the KV serialisation performance, which we don't expect, could you investigate please?

jumaffre:
> @jumaffre there seems to be substantial hit to the KV serialisation performance, which we don't expect, could you investigate please?

Yep, I'm aware and that's why I've changed the PR to draft for now.

src/kv/msgpack_serialise.h
```diff
@@ -124,4 +124,4 @@ namespace kv
```
jumaffre, line 124: To delete.

doc/build_apps/kv/kv_serialisation.rst
```diff
@@ -30,8 +30,7 @@ The following table describes the structure of a serialised KV Store transaction
 +          +------------------------------------------+-------------------------------------------------------------------------+
 |          | **Repeating [0..n]**                     | With ``n`` the number of maps in the transaction                        |
 +          +-----+------------------------------------+-------------------------------------------------------------------------+
-|          |     | | ``KOT_MAP_START_INDICATOR``      | | Indicates the start of a new serialised :cpp:type:`kv::Map`           |
```
jumaffre, line 33: Note that the `KOT_MAP_START_INDICATOR` is now gone as no longer necessary.

jumaffre:
With some tweaks, the KV benchmarks are now much closer to what we used to have with `msgpack-c` so this is now ready for review.

src/kv/raw_serialise.h
```diff
@@ -0,0 +1,200 @@
...
+      data_offset = 0;
+      data_ptr = data_in_ptr;
+      data_size = data_in_size;
+    }
+
+    template <typename T>
+    T read_next()
+    {
+      if constexpr (
+        nonstd::is_std_vector<T>::value ||
```
eddyashton, line 161: Do we ever actually use this for `std::vector<T>`, rather than just `std::vector<uint8_t>`? If we don't need arbitrary types, I think this and `std::string` can share more code and this can be simpler (no `sizeof` corrections, which I think is wrong for any struct with padding in it).

jumaffre: We actually do for the serialisation of the view history in snapshots. 

Good point for the `sizeof` and padding! We could check that `T` is integral, or maybe use https://en.cppreference.com/w/cpp/types/has_unique_object_representations?

eddyashton: After staring at this some more its actually safe - although we're writing directly to a `std::vector`'s `.data()`, we're only doing it with the contents we previously dumped _from_ a `std::vector`'s `.data()`, so it should be safe.

PR #2347, opened 2021-03-23
/tx and /commit use a single transaction_id rather than separate view and seqno

eddyashton:
We introduced a unified `TxID` recently but we were not consistently using it in the HTTP API. This removes what I believe are the last instances of separate `view` and `seqno` fields, in the query parameter for `/tx` and the response body for `/commit`.


src/ds/openapi.h
```diff
@@ -240,6 +240,19 @@ namespace ds
...
+    // to that). This default implementation simply falls back to
+    // fill_json_schema, which already exists to describe leaf types. A
+    // recursive implementation for struct-to-object types is created by the
+    // json.h macros, and this could be implemented manually for other types.
+    template <typename Doc, typename T>
+    void add_schema_components(Doc&, nlohmann::json& j, const T& t)
+    {
+      fill_json_schema(j, t);
+    }
+
```
eddyashton, line 255: This lets us remove the `add_schema_components` for `EntityId`, and removes one class of horrid template error when trying to auto-document C++ types - you only need to implement `fill_json_schema`, not this very-similar-but-subtley-different function too.

eddyashton: Oh and the side-benefit that I hadn't quite twigged is that it results in more `#ref`s and fewer duplicated schema in the resulting OpenAPI doc.

tests/reconfiguration.py
```diff
@@ -33,11 +34,15 @@ def count_nodes(configs, network):
 def check_can_progress(node, timeout=3):
     with node.client() as c:
         r = c.get("/node/commit")
+        original_seqno = ccf.clients.parse_tx_id(r.body.json()["transaction_id"])[1]
```
achamayou, line 37: I think that if we're going to have a `parse_tx_id` that's longer than `split('.')`, we should have it return a TxID with named fields

achamayou (approved):

PR #2348, opened 2021-03-23
Extend testing of custom auth policies

eddyashton:
Adds a test that if a custom authn policy throws, whether forwarded or 'normal', we catch this and convert it to a HTTP error response.


tests/e2e_logging.py
```diff
@@ -324,38 +324,70 @@ def require_new_response(r):
...
-            assert r.status_code == http.HTTPStatus.UNAUTHORIZED.value, r.status_code
-
-            LOG.info("Requests with unacceptable header contents are refused")
-            r = c.get("/app/custom_auth", headers={name_header: "", age_header: "42"})
-            assert r.status_code == http.HTTPStatus.UNAUTHORIZED.value, r.status_code
-            r = c.get(
-                "/app/custom_auth", headers={name_header: "Bob", age_header: "12"}
-            )
-            assert r.status_code == http.HTTPStatus.UNAUTHORIZED.value, r.status_code
+        for node in (primary, other):
```
eddyashton, line 329: The below is just whitespace+formatting adjustment, because of the additional loop here.

achamayou (approved):
Thank you, this is an excellent test to add.

jumaffre (approved):

PR #2350, opened 2021-03-23
No-op JS gov

achamayou:
This PR aims to cover the endpoint surface for the new governance runtime, and to implement the proposal/voting flow up to the final status resolution but without executing the actions (yet).

- Proposals
  - [x] Validation
  - [x] Storage
- Ballots
  - [x] Validation
  - [x]  Storage
  - [x] Evaluation

(Previously) missing tests:

- [x] Voter-identity-dependent evaluation
- [x] Multiple votes
- [x] KV-dependent resolution

Out of scope for this PR:

- Action execution
- Error reporting during evaluation

TODO after review:

- [x] Return ProposalInfoSummary on submit/vote
- [x] Don't assume proposal is a string (generic buffer)


CMakeLists.txt
```diff
@@ -56,6 +56,10 @@ option(ZAP_TEST
        "ZAP fuzz test using https://www.zaproxy.org/docs/docker/api-scan/" OFF
 )
 option(BUILD_SMALLBANK "Build SmallBank sample app and clients" ON)
+option(ENABLE_JS_GOV "Disable JS Governance" ON)
```
letmaik, line 59:
```suggestion
option(ENABLE_JS_GOV "Enable JS Governance" ON)
```

Should this be off by default?

tests/governance_js.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def action(name, **args):
+    return {"name": name, "args": args}
+
+
+@reqs.description("Test proposal validation")
+def test_proposal_validation(network, args):
+    primary, _ = network.find_nodes()
+    add_member = [action("add_member", cert="", enc_pubk="", member_data={})]
```
letmaik, line 18: Should be wrapped in `"actions": ...` but probably best to get the PR in and do any further work iteratively.

letmaik (approved):

CMakeLists.txt
achamayou: Started out that way, but now it's on by default and only disabled in release mode.

tests/governance_js.py
achamayou: So let's talk about this, but wrapping is awkward as discussed this morning if we think that being precise about the proposal is important (and that's still my opinion) because it implies partial parsing/serialisation.

achamayou: Ok, so here's a suggestion:

Proposals _are_ {"actions": []}, and are stored as such in `public:ccf.gov.proposals.js` (all `.js` suffixes are temporary for disambiguation where necessary). That adds a bit to the validation in the constitution, but it's not too bad.

Then there are:

`GET /gov/proposals.js/{id}/proposal`, which returns the proposal verbatim, `GET /gov/proposals.js/{id}/state`, which returns the contents of `public:ccf.gov.proposals_info`, ie. proposer id, state and votes.

This eliminates the need for partial serialisation/deserialisation.

letmaik: I would probably make it `GET /gov/proposals.js/{id}` which has the proposer id and state (and maybe vote count), and then `GET /gov/proposals.js/{id}/actions` which returns what was submitted, and `GET /gov/proposals.js/{id}/votes` for the votes.

achamayou: Ok, I did nearly that, except I kept the ballots in the first endpoint for now. I think we can default to returning a count and having a url param to ask for expansion.

letmaik: Works for me.

src/node/gov.h
```diff
@@ -5,26 +5,58 @@
...
       ccf::MemberId proposer_id;
+      ccf::ProposalState state;
       std::unordered_map<ccf::MemberId, std::string> ballots = {};
     };
-    DECLARE_JSON_TYPE(ProposalInfo)
-    DECLARE_JSON_REQUIRED_FIELDS(ProposalInfo, proposer_id, ballots);
+    DECLARE_JSON_TYPE(ProposalInfo);
+    DECLARE_JSON_REQUIRED_FIELDS(ProposalInfo, proposer_id, state, ballots);
+
+    struct ProposalSubmitted
```
achamayou, line 25: This is the type we return on submission of a proposal or a ballot. It's really a proposal state, but of course `ccf::ProposalState` is already the enum. So maybe `ProposalInstanceState`? Or `jsgov::ProposalState`, not to be confused with `ccf::ProposalState`? Input welcome.

src/node/gov.h
```diff
@@ -5,26 +5,58 @@
...
+    DECLARE_JSON_REQUIRED_FIELDS(Action, name, args);
+
+    struct Proposal
+    {
+      std::vector<Action> actions;
+    };
+    DECLARE_JSON_TYPE(Proposal);
+    DECLARE_JSON_REQUIRED_FIELDS(Proposal, actions);
+
+    struct Ballot
```
achamayou, line 55: I've tried to create, and would like to enforce a distinction between `ballot`, the script submitted by a member, and `vote`, the outcome of that ballot applied to a `(proposal, kv)`. Thoughts?

src/node/gov.h
```diff
@@ -5,26 +5,58 @@
...
+    {
+      ProposalId proposal_id;
+      ccf::ProposalState state;
+    };
+    DECLARE_JSON_TYPE(ProposalSubmitted);
+    DECLARE_JSON_REQUIRED_FIELDS(ProposalSubmitted, proposal_id, state);
 
-    using ProposalMap = kv::RawCopySerialisedMap<ProposalId, Proposal>;
+    using ProposalMap = kv::RawCopySerialisedMap<ProposalId, std::string>;
     using ProposalInfoMap = kv::MapSerialisedWith<
```
jumaffre, line 34: Use `ccf::ServiceMap`?

achamayou: How do you mean? The types are structurally identical, but that's a coincidence, not a deliberate choice.

src/node/gov.h
letmaik: Why not return `ProposalInfo`?

src/node/rpc/member_frontend.h
```diff
@@ -2110,6 +2241,533 @@ namespace ccf
...
+            ctx.get_caller<ccf::MemberSignatureAuthnIdentity>();
+          if (!check_member_active(ctx.tx, caller_identity.member_id))
+          {
+            return make_error(
+              HTTP_STATUS_FORBIDDEN,
+              ccf::errors::AuthorizationFailed,
+              "Member is not active.");
+          }
+
+          // Take expand=ballots, return eg. "ballots": 3 if not set
```
letmaik, line 2420: Never use the same field for different things. Better to name it `"ballot_count"` or similar.

src/node/gov.h
achamayou: Sticking to that for now, we agree about the distinction, although `vote` may get renamed to something better.

src/node/rpc/member_frontend.h
```diff
@@ -1152,6 +1152,137 @@ namespace ccf
...
+        argv[1] = prop_id;
+
+        auto vs = JS_NewArray(context);
+        size_t index = 0;
+        for (auto& [mid, vote] : votes)
+        {
+          auto v = JS_NewObject(context);
+          auto member_id = JS_NewStringLen(context, mid.data(), mid.size());
+          JS_DefinePropertyValueStr(
+            context, v, "member_id", member_id, JS_PROP_C_W_E);
```
letmaik, line 1231:
```suggestion
            context, v, "memberId", member_id, JS_PROP_C_W_E);
```

achamayou: I'm hesitant about this, because it's member_data, and is_operator and absolutely everything is_still_like_this. I think we do it all or we don't do it, but I don't want half-half.

src/node/rpc/member_frontend.h
achamayou: Ok

src/node/gov.h
achamayou: Gotcha

src/node/gov.h
achamayou: Going with:

```
    struct ProposalInfoSummary
    {
      ProposalId proposal_id;
      ccf::MemberId proposer_id;
      ccf::ProposalState state;
      size_t ballot_count;
    };
```

src/node/rpc/member_frontend.h
letmaik: OK, not a big deal.

eddyashton (approved):

PR #2352, opened 2021-03-24
Exception resilience when processing ring buffer messages

jumaffre:
It was possible that exceptions thrown during the processing of ring buffer messages in the enclave bubbled up to the main processing loop (`enclave.h`) which caused the node to log the exception and shutdown. This shouldn't be the case if the node can actually recover from the exception. 

This PR aims to address this by catching as many exceptions as possible in each sub-module that processes ring buffer messages (e.g. forwarder, node-to-node channels) or replacing exceptions with `LOG_FAIL` logs. We also log the type and size of the message that causes the exception (`messaging.h`).


src/node/node_state.h
```diff
@@ -878,13 +873,27 @@ namespace ccf
     void verify_snapshot_end()
     {
       std::lock_guard<SpinLock> guard(lock);
-      sm.expect(State::verifyingSnapshot);
+      if (!sm.check(State::verifyingSnapshot))
+      {
+        LOG_FAIL_FMT(
+          "Node is state {} cannot finalise snapshot verification", s.value());
```
eddyashton, line 879:
```suggestion
          "Node in state {} cannot finalise snapshot verification", s.value());
```

eddyashton (approved):

PR #2354, opened 2021-03-25
Add named TxID type to Python

eddyashton:
Moving `ccf.clients.parse_tx_id` to its own file since its now used in several places, and returning a named type. I vacillated between returning `None` or `TxID(None, None)` when trying to parse an invalid string - I've opted for the latter for now, because I think it will print cleaner errors with our existing code.


python/ccf/tx_id.py
```diff
@@ -0,0 +1,30 @@
...
+        return self.view is not None
+
+    @staticmethod
+    def from_str(s: str):
+        return TxID(*TxID.parse(s))
+
+    @staticmethod
+    def parse(s: Optional[str]):
+        try:
+            if s is not None:
```
achamayou, line 25: That if isn't necessary since you catch AttributeError anyway.

achamayou (approved):

PR #2355, opened 2021-03-25
Run network init sequence across a mix of primary/followers

achamayou:
We've seen errors happening specifically with forwarded requests before, this is making the init sequence use a mix of nodes rather than always the primary.


achamayou:
50/50 did not work because we have a few single-node environments.

eddyashton (approved):

tests/infra/network.py
```diff
@@ -403,21 +403,23 @@ def start_and_join(self, args):
                 "cp", args.js_app_script, args.binary_dir
             ).check_returncode()
             self.consortium.set_js_app(
```
jumaffre, line 405: Could we do the same thing for the recovery sequence, in the `recover()` function below?

jumaffre (approved):

tests/infra/network.py
achamayou: Good idea, let's find out :)

PR #2357, opened 2021-03-25
Add N2N channel key rotation

wintersteiger:
This adds the facilities to trigger key exchanges for node-to-node channels when the message limits have been reached. This is item 3 of #2092.

### DIFF
--- a/src/node/channels.h
+++ b/src/node/channels.h
@@ -23,6 +23,13 @@ namespace ccf
   using SeqNo = uint64_t;
   using GcmHdr = crypto::GcmHeader<sizeof(SeqNo)>;
 
+#ifndef NDEBUG
+  static constexpr size_t message_limit = 25;
+#else
+  // 2**24.5 as per RFC8446 Section 5.5
+  static constexpr size_t message_limit = 23726566;
+#endif
+
   struct RecvNonce
   {
     uint8_t tid;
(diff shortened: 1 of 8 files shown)


wintersteiger:
In debug mode, the first CI run used a very small message limit, i.e. a new key exchange is performed every 25 messages. This was just for the first test to see whether I can trigger any errors or assertions, but I couldn't. I'll raise that limit now so that there's no impact on performance.

src/node/channels.h
```diff
@@ -631,24 +681,29 @@ namespace ccf
...
+      if (status == WAITING_FOR_FINAL)
         return;
 
       LOG_INFO_FMT("Initiating node channel with {}.", peer_id);
 
-      status = INITIATED;
+      key_exchange_in_progress = true;
+
+      if (status != ESTABLISHED)
+        status = INITIATED;
```
achamayou, line 696:
```suggestion
      }
        status = INITIATED;
      }
```

achamayou (approved):

src/node/channels.h
wintersteiger: Awww, I thought I caught all of them this time!

src/node/test/channels.cpp
```diff
@@ -501,20 +514,20 @@ TEST_CASE("Replay and out-of-order")
 
   INFO("Restart, message, resume");
```
jumaffre, line 515: Please correct me if I'm wrong but I don't think we test for the auto-refreshing of the channel when the message limit has been reached. If the message limit is passed as an argument to the `Channel` constructor, this should be easy to set it to a low value in this unit test and see if the `key_exchange_init` message is produced.

src/node/channels.h
```diff
@@ -23,6 +23,13 @@ namespace ccf
   using SeqNo = uint64_t;
   using GcmHdr = crypto::GcmHeader<sizeof(SeqNo)>;
 
+#ifndef NDEBUG
+  static constexpr size_t message_limit = 2048;
```
jumaffre, line 27: Perhaps add a comment here to explain why we do this (i.e. see a few refresh during our perf test)?

jumaffre (approved):

src/node/test/channels.cpp
wintersteiger: Good point, I'll add a new test for that!

PR #2359, opened 2021-03-25
Disable BFT in CCF releases

jumaffre:
As BFT support is still experimental, this PR disables it for CCF releases via a new CMake `ENABLE_BFT` flag (defaults to `ON`), which is set to `OFF` in our release CI job (tested in `test_install.sh`). 

The fact that BFT is enabled or not is captured at compile time so that it is part of the enclave measurement, so that operators cannot enable BFT for an already-trusted measurement.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,11 @@ configure_file(
 )
 configure_file(${CCF_DIR}/python/setup.py.in ${CCF_DIR}/python/setup.py @ONLY)
 
-set(CONSENSUSES cft bft)
+if(${ENABLE_BFT})
+  set(CONSENSUSES cft bft)
+else()
+  set(CONSENSUSES cft)
+endif()
 
 option(BUILD_TESTS "Build tests" ON)
 option(BUILD_UNIT_TESTS "Build unit tests" ON)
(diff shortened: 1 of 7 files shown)


wintersteiger (approved):
LGTM!

ashamis (approved):

cmake/common.cmake
```diff
@@ -50,6 +50,11 @@ option(COVERAGE "Enable coverage mapping" OFF)
 option(SHUFFLE_SUITE "Shuffle end to end test suite" OFF)
 option(LONG_TESTS "Enable long end-to-end tests" OFF)
 
+option(ENABLE_BFT "Disable experimental BFT consensus at compile time" ON)
```
achamayou, line 53:
```suggestion
option(ENABLE_BFT "Enable experimental BFT consensus at compile time" ON)
```

achamayou (approved):

CMakeLists.txt
```diff
@@ -47,7 +47,11 @@ configure_file(
 )
 configure_file(${CCF_DIR}/python/setup.py.in ${CCF_DIR}/python/setup.py @ONLY)
 
-set(CONSENSUSES cft bft)
+if(${ENABLE_BFT})
```
jumaffre, line 50:
```suggestion
if(ENABLE_BFT)
```

PR #2360, opened 2021-03-25
Separate 3rdparty includes into exported, internal, test

eddyashton:
More of these are exported than are really necessary - I think `merklecpp` and `snmalloc` _should_ be `internal`, but they need to be `exported` for now because their headers are visible.


wintersteiger:
Many thanks for taking this on! Do we really need to export the merklecpp header? It's easy enough for others to copy it from the GitHub page if someone wants to validate our trees.

achamayou:
@wintersteiger I think we expose some types through our own headers in the receipts (PathT iirc) so that’s why they need to be exported right now.

That’s something we should try to change as @eddyashton said.

wintersteiger:
Right, good point! Let me know when/if/how you want that changed, I'm happy to help out, especially for the Merkle trees, which shouldn't take too much effort.

jumaffre (approved):

achamayou (approved):

PR #2363, opened 2021-03-26
Non-deterministic/randomized ECDSA

wintersteiger:
ECDSA signatures are non-deterministic/randomized, i.e. signing the same message twice produces different signatures (and both verify correctly). Our implementations of signing with mbedTLS and OpenSSL don't agree on that; mbedTLS is deterministic and OpenSSL is not. mbedTLS makes it hard to sign non-deterministically (see code changes here), while OpenSSL doesn't support deterministic signatures (work in progress: https://github.com/openssl/openssl/pull/9223). 

I believe we should be using non-deterministic signatures because we have a good enough source of randomness, at least in enclaves, via the Intel DRNG (see also https://tools.ietf.org/html/rfc6979), so that's what I opted to implement. Our tests tolerate either, so I can be convinced otherwise (and perhaps help out with the OpenSSL implementation at their end to speed up integration).

### DIFF
--- a/src/crypto/mbedtls/key_pair.cpp
+++ b/src/crypto/mbedtls/key_pair.cpp
@@ -15,6 +15,7 @@
 #include <mbedtls/asn1write.h>
 #include <mbedtls/bignum.h>
 #include <mbedtls/error.h>
+#include <mbedtls/md.h>
 #include <mbedtls/oid.h>
 #include <mbedtls/pem.h>
 #include <mbedtls/pk.h>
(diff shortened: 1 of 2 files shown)


eddyashton:
Any non-determinism is a problem for BFT, where we expect to be able to reproduce identical results on every node. Where an app needs non-determinism, the vague plan is that we'll be able to capture a non-deterministic result on the primary and accept this on the secondaries; to use that here we'd need an additional API to say "sign with _this_ RNG sequence". Alternatively we could capture the signature itself as the input for the re-executing nodes.

I think in general we should prefer the deterministic implementation, unless there's some benefit to the non-determinism that I'm missing?

wintersteiger:
Thanks @eddyashton for pointing us toward BFT. You're right that this could become important at some point for some applications. I think for now, it's better for us to go with non-deterministic signatures because they provide slightly better security, e.g. for node-to-node channels. We still have both implementations, so for BFT, we can use the deterministic mbedTLS version instead of the default `KeyPair`. In the not too distant future, OpenSSL will also support this out of the box, we can make that optional for both implementations. 

If we really need deterministic OpenSSL signatures before they become officially supported, we could indeed try to determinize the current algorithms by fixing the RNG, e.g. to a hash of the input, which some others seem to do.

achamayou (approved):

PR #2364, opened 2021-03-26
New Python read ledger tool

jumaffre:
As suggested by @achamayou, it would be convenient to have a lightweight command line to display the contents of a ledger. I created this quickly while checking that all tables were serialised correctly in the ledger, so contributing this to the main repo. All feedback welcome.

Sample output:
image


eddyashton (approved):
This is great. Is there any summary we can print of the private tables?

jumaffre:
> This is great. Is there any summary we can print of the private tables?

From the ledger files, we only know the size of private domain in the transaction, which may be useful for debugging purposes?

achamayou:
This is good, but I thought that emitting a single large JSON document worked well with VSCode, because you can so easily collapse/expand sections, and with jq. I think emitting a document per chunk may be a nice balance to avoid documents getting too large.

eddyashton:
> This is good, but I thought that emitting a single large JSON document worked well with VSCode, because you can so easily collapse/expand sections, and with jq. I think emitting a document per chunk may be a nice balance to avoid documents getting too large.

FWIW, adding identation or braces (per-entry and per-table) will make this collapsible in VSCode just as nicely as JSON, though not jq-parseable.

wintersteiger:
This is awsome, thanks @jumaffre!

ashamis:
This is great! thank you for doing this. If we are putting this in the main repo i think there needs to be a test that verifies that we do not accidentally break this as this is now "officially" supported? Alternatively if we add a "samples" repo we can put this there.

achamayou:
@ashamis `ccf.ledger.Ledger` is exercised by the governance_history_test where it's used to audit governance steps offline.

jumaffre:
Merging this now as it is useful to implement JS governance. I'm happy to revisit and improve this tool at a later stage.

PR #2366, opened 2021-03-26
Decorate vectors and arrays through libfmt

eddyashton:
I noticed these custom formatters while investigating something else. This PR removes the unnecessary copy when formatting `std::array`s, generalises to any length of array, and decorates the types so its easier to grep where the hexification has come from.

```
  std::vector<uint8_t> v{0xde, 0xad, 0xbe, 0xef};
  LOG_INFO_FMT("{}", v);

  std::array<uint8_t, 7> a{1, 2, 3, 4, 5, 6, 7};
  LOG_INFO_FMT("What about my keys? {}!", a);
```
```
[info ] ... | <vec[4]: de ad be ef>
[info ] ... | What about my keys? <arr[7]: 01 02 03 04 05 06 07>!
```

### DIFF
--- a/src/ds/logger_formatters.h
+++ b/src/ds/logger_formatters.h
@@ -21,7 +21,8 @@ namespace fmt
     template <typename FormatContext>
     auto format(const std::vector<uint8_t>& v, FormatContext& ctx)
     {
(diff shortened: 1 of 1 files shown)


achamayou (approved):

wintersteiger (approved):
Nice!

ashamis (approved):

Issue #1992, opened 2020-12-08
Permanently delete items in the KV champ maps


eddyashton:
Currently we keep deleted items around forever with a tombstone entry. The value is minimal, but the entire key must be kept. We now have support for removing things from the underlying champ, so should use this. This will require some extra testing to make sure the semantics are correct and well understood.

There was an initial attempt in this PR, #1781. The comments there summarise the open questions.

eddyashton:
I've been thinking about this further, and convinced myself again that its actually safe to permanently delete things, as any other 'real' conflicts would be caught by other dependencies, or else there's no conflict. Also experimenting with diagrams to explain these tx conflicts.

#### Normal conflicts
So a normal conflict looks like this:
```
        1              2               3
A [ K=1 ]              |               |
B       | [ a=K? K=a+1 ]               |
C       |  [ a=K? K=a+1  X[ a=K? K=a+1 ]
```
Time goes left-to-right, versions are listed at the top. Transaction A writes a value to key K, and applies this at version 1. B and C try to do the same workload which is "increment K by 1". They both start after 1 is applied, so read the state written by version 1. C is shown as running slightly later, but really they're concurrent and their only ordering is who tries to commit first. B is applied first, creating version 2. Then when C tries to apply, it conflicts (marked by `X`), because its read of K (`K?`) was at version 1, but K has now been written at version 2. So C is retried, now reading the state at version 2 (written by B), and eventually applies at version 3.

The goal here is to produce the same results as if the transactions executed in sequence. So if 3 hadn't read K, but had read some other keys, we can execute it with a read version of 1 but we get the same result as if we'd had a read version of 2.

#### Deletions
Now we deal with _deleting_ keys, written as `!K`, and here's the case that I was concerned about:
```
       1         2      3 
A [ !K ]         |      | 
B      |   [ K=1 ]      | 
C      |         | [ !K ] 
D      | [ a=K? K'=a?     ]
```
D is executed reading at version 1, checks if K exists, and writes this somewhere else in the KV. While its executing, other transactions create/modify K, maybe multiple times, and eventually delete it before D completes.

When D completes, does it need to conflict and retry, because K was modified while it was executing? If it does, then we need to store a tombstone for K in the KV, for at least as long as D is executing, so that when D comes to conflict we can compare "you depended on deleted-at-1" against "K is now deleted-at-3".

[...]

PR #2367, opened 2021-03-26
Consolidate View and SeqNo typedefs

### LINKED ISSUE #2282: Consolidate View and SeqNo typedefs
We currently have a lot of different typedefs for `View` and `SeqNo` (under `kv::`, `kv::Consensus`, etc), as well as some older terminology referring to `Version`, `Index`, and `Term`. As part of #697, the former should be consolidated into a single alias (`ccf::View`, `ccf::SeqNo`, from `ccf/tx_id.h`), and the latter should only be used inside the KV/Raft and not exposed beyond them.

eddyashton:
Resolves #2282.

Opening as a draft so I remember to take another look at this on Monday, and remember exactly what I did.

This removes our duplicate definitions of core type aliases, preferring `ccf::SeqNo` and `ccf::View` from `tx_id.h` wherever possible. The exception is the KV and things which touch it closely, which get to keep `Term`, `Version`, and `kv::TxID`, though these also converted to unsigned in most places.

Big effects:
- `VIEW_UNKNOWN` and `NoVersion` are now `0`, rather than large negatives. I think this is sound - these values were never valid (the first real `TxID` must be at least `1.1`), but this may have unintended consequences on some default-initialisations/comparisons.


src/kv/test/kv_test.cpp
```diff
@@ -1471,169 +1471,202 @@ TEST_CASE("Conflict resolution")
   REQUIRE_THROWS(tx2.commit());
 }
 
-TEST_CASE("Primary can create correct execution order")
+std::string rand_string(size_t i)
+{
+  return fmt::format("{}: {}", i, rand());
+}
+
+TEST_CASE("Max conflict version tracks execution order")
```
eddyashton, line 1479: Took me a while to debug why this test was broken, and it turned out to be because it was using `id` as a `kv::Version` (return value from `version_resolver`). While fixing that I combined the good and bad tests into 1 to reduce the duplicated code, removed a type confusion between primary and backup, and added 2 entries to the test matrix testing in-primary-order execution (for both independent and strictly-ordered transactions).

include/ccf/tx_id.h
```diff
@@ -14,9 +14,26 @@
 
 namespace ccf
 {
+  // Transactions occur within a fixed View. Each View generally spans a range
```
wintersteiger, line 17: The consolidation and these comments and are great, I was always a bit confused by the respective numbers/terms. Does this mean that the documentation needs updating too?

src/kv/kv_types.h
```diff
@@ -30,12 +31,15 @@ namespace aft
...
-  using Version = int64_t;
-  static const Version NoVersion = std::numeric_limits<Version>::min();
+  // Version indexes modifications to the local kv store.
+  using Version = uint64_t;
+  static constexpr Version NoVersion = 0u;
 
-  static bool is_deleted(Version version)
+  // DeletableVersion describes the version of an individual key within each
+  // table, which may be negative to indicate a deletion
+  using DeletableVersion = int64_t;
```
wintersteiger, line 40: Possible overflow issues. Does this mean SeqNo is effectively limited to 2^31 and are there checks to ensure this? E.g. `static_cast<DeletableVersion>(v_)` in `src/kv/untyped_map.h` smells fishy to me.

src/node/channels.h
```diff
@@ -20,13 +20,13 @@
 
 namespace ccf
 {
-  using SeqNo = uint64_t;
-  using GcmHdr = crypto::GcmHeader<sizeof(SeqNo)>;
+  using SendNonce = uint64_t;
```
wintersteiger, line 23: This is actually used like a sequence number (number of messages since last key rotation, which is limited to 2^24.5 now). Not sure we need a type for this at all.

eddyashton: I'm just renaming the existing type so it doesn't squat on `ccf::SeqNo` for now. I agree it may be simpler to remove this if we don't have a clean name for it, though in general I think its a good idea to add name for these simple types whenever its used for a distinct purpose (to reduce some errors where they're used in the wrong place, or at least make it possible to unpick the mistake after the fact).

include/ccf/tx_id.h
eddyashton: This is mostly a rewording of the existing comments we had in the kv, but that's a good point that these should go in the docs - I'll doxygenate and add them now.

src/kv/kv_types.h
eddyashton: Good spot, I'll amend the wrapping check we already had. In general these casts are probably going to remain looking unsafe/unguarded, but they should be temporary (once #1992 is done, `DeletableVersion` can be removed and we no longer need _any_ signed versions).

include/ccf/tx_id.h
wintersteiger: Awesome, many thanks!

include/ccf/tx_id.h
```diff
@@ -14,9 +14,26 @@
...
+  // View transitions. This matches the order in which transactions are
+  // applied, where a higher SeqNo means that a transaction executed later.
+  // SeqNos are unique during normal operation, but around elections it is
+  // possible for distinct transactions in separate Views to have the same
+  // SeqNo. Only one of these transactions will ever commit, and the others are
+  // ephemeral.
   using SeqNo = uint64_t;
 
+  // No transactions occur in View 0.
+  constexpr View VIEW_UNKNOWN = 0;
```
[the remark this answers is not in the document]
achamayou: The change does make a lot of sense though.

doc/build_apps/api.rst
```diff
@@ -65,6 +65,12 @@ Supporting Types
 .. doxygenenum:: ccf::TxStatus
```
achamayou, line 65: TxID should be there too

include/ccf/base_endpoint_registry.h
```diff
@@ -88,14 +88,11 @@ namespace ccf
      * @see ccf::TxStatus
      */
     ApiResult get_status_for_txid_v1(
-      kv::Consensus::View view,
-      kv::Consensus::SeqNo seqno,
-      ccf::TxStatus& tx_status);
+      ccf::View view, ccf::SeqNo seqno, ccf::TxStatus& tx_status);
```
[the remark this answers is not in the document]
achamayou: +1 for keeping the components here

src/kv/tx.h
```diff
@@ -66,7 +66,8 @@ namespace kv
 
     bool committed = false;
     bool success = false;
-    Version read_version = NoVersion;
+
+    std::optional<Version> read_version = std::nullopt;
```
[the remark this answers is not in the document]
achamayou: I would add that as a comment in the code, it's a very useful explanation.

src/kv/kv_types.h
achamayou: It might be worth bumping up the importance of #1992 post-1.0 then because a 2^31 transaction limit is clearly going to be a problem sooner rather than later.

achamayou (approved):

src/kv/kv_types.h
wintersteiger: +1. I had only one half of my brain engaged it seems - it's of course 2^63 :-)

doc/build_apps/api.rst
eddyashton: Too many branches, I thought it was - have added it now.

src/kv/kv_types.h
eddyashton: I've updated the check in `next_version_internal()` to track this limit correctly - and yes its 2^63 rather than 2^31. We will wrap to 0 and log - this will break any service trying where that actually happens, but retains the semantics we've always had. In a few hundred years we can replace `uint64_t` with a bignum type.

PR #2368, opened 2021-03-29
JS governance: resolve functions for existing constitutions

letmaik:


src/runtime_config/resolve_default.js
```diff
@@ -0,0 +1,23 @@
+export function resolve(proposal, proposerId, votes) {
+    let memberVoteCount = 0;
```
achamayou, line 2: const memberVoteCount = votes.filter(v => v.vote).length;

src/runtime_config/resolve_veto.js
```diff
@@ -0,0 +1,27 @@
+export function resolve(proposal, proposerId, votes) {
+    let memberVoteCount = 0;
+    for (const vote of votes) {
+        if (vote.vote) {
```
eddyashton, line 4: Looking at this example, I think this field should be `vote.in_favour`, or `vote.for`, or `vote.supports`, or something. To avoid the duplicated `vote.vote`, and so we can more clearly distinguish "did you submit ballot executed to completion" from "do you support this proposal".

achamayou: Agreed, although `for` is probably not going to work.

achamayou (approved):

PR #2370, opened 2021-03-29
Add JS proposals and votes to proposal_generator

eddyashton:
The generation is much simpler than for Lua since we can trivially inline values with `json.dumps`.

This is enabled with an env var read into a Python module variable, so we can test it from the CLI and turn it on for specific tests.

Minimal testing so far, just generating an equivalent of one of the existing proposals with a more detailed vote. The real testing for this will come when we use it for submitting existing proposals in JS form.


python/ccf/proposal_generator.py
```diff
@@ -18,6 +18,9 @@
 from loguru import logger as LOG  # type: ignore
 
 
+GENERATE_JS_PROPOSALS=os.getenv("JS_GOVERNANCE")
```
wintersteiger, line 21: Maybe add a CLI flag instead of the env var?

eddyashton: I prefer the env var since its simpler to turn on globally for an entire test (ie - at some point soon we can switch export this env var for our e2e suites, and they should all work with the new proposal style). The CLI invocations of this are pretty rare, so I don't think its worse to have the env var than a flag during the transition period.

python/ccf/proposal_generator.py
```diff
@@ -144,35 +147,70 @@ def build_proposal(
...
+    vote: Dict[str, Any] = {}
+
+    if GENERATE_JS_PROPOSALS:
+        action = {"name": proposed_call, "args": args}
+        actions = [action]
+        proposal = {"actions": actions}
+
+        vote_lines = []
+        vote_lines.append("function vote (raw_proposal, proposer_id) {")
+        vote_lines.append("  let proposal = JSON.parse(input)")
```
letmaik, line 160:
```suggestion
        vote_lines.append("  let proposal = JSON.parse(raw_proposal)")
```

python/ccf/proposal_generator.py
```diff
@@ -144,35 +147,70 @@ def build_proposal(
...
+        vote_lines.append("  if (!'actions' in proposal) { return false }")
```
letmaik, line 161:
```suggestion
        vote_lines.append("  if (!('actions' in proposal)) { return false }")
```

python/ccf/proposal_generator.py
```diff
@@ -144,35 +147,70 @@ def build_proposal(
...
+        vote_lines.append("  let actions = proposal['actions']")
+        vote_lines.append("  if (actions.length !== 1) { return false }")
+        vote_lines.append("  let action = actions[0]")
+        vote_lines.append("  if (!'name' in action) { return false }")
```
letmaik, line 165:
```suggestion
        vote_lines.append("  if (!('name' in action)) { return false }")
```

python/ccf/proposal_generator.py
```diff
@@ -144,35 +147,70 @@ def build_proposal(
...
+        vote_lines.append(
+            f"  if (action.name !== '{proposed_call}') {{ return false }}"
+        )
+
+        if args is not None:
+            vote_lines.append("  if (!'args' in action) { return false }")
```
letmaik, line 171:
```suggestion
            vote_lines.append("  if (!('args' in action)) { return false }")
```

python/ccf/proposal_generator.py
```diff
@@ -144,35 +147,70 @@ def build_proposal(
...
+            vote_lines.append("  let args = action.args")
+
+            for name, body in args.items():
+                vote_lines.append(f"  if (!'{name}' in args) {{ return false }}")
+                vote_lines.append(f"  let expected = {json.dumps(body)}")
+                vote_lines.append(f"  if (args.{name} !== expected) {{ return false }}")
```
letmaik, line 177:
```suggestion
                vote_lines.append(f"  if (args['{name}'] !== expected) {{ return false }}")
```

python/ccf/proposal_generator.py
```diff
@@ -144,35 +147,70 @@ def build_proposal(
...
+        vote_lines.append(
+            f"  if (action.name !== '{proposed_call}') {{ return false }}"
+        )
+
+        if args is not None:
+            vote_lines.append("  if (!'args' in action) { return false }")
+            vote_lines.append("  let args = action.args")
+
+            for name, body in args.items():
+                vote_lines.append(f"  if (!'{name}' in args) {{ return false }}")
```
letmaik, line 175:
```suggestion
                vote_lines.append(f"  if (!('{name}' in args)) {{ return false }}")
```

achamayou (approved):

PR #2371, opened 2021-03-29
Apply JS gov

achamayou:
Apply actions from JS governance.


src/js/wrap.h
```diff
@@ -22,14 +22,27 @@ namespace js
   extern JSClassDef kv_map_handle_class_def;
   extern JSClassDef body_class_def;
 
+  enum TxAccess
```
letmaik, line 25: enum class?

src/runtime_config/constitution.js
```diff
@@ -104,8 +88,18 @@ const actions = new Map([
...
+    ),
+  ],
+  [
+    "remove_user",
+    new Action(
+      function (args) {
+        return typeof args.user_id === "string";
+      },
+      function (args) {
+        ccf.kv["public:ccf.gov.users.certs"].delete(args.user_id);
```
letmaik, line 101: Needs strToBuf

achamayou: Yes, thank you.

src/js/wrap.h
achamayou: Sorry?

letmaik (approved):

src/runtime_config/constitution.js
```diff
@@ -104,8 +88,19 @@ const actions = new Map([
...
-      function (args, tx) {
-        return true;
+      function (args) {}
+    ),
+  ],
+  [
+    "remove_user",
+    new Action(
+      function (args) {
+        return typeof args.user_id === "string";
```
jumaffre, line 98: We should probably add a function to check that a giver id is valid (i.e. hex-encoded string of SHA-256).

jumaffre (approved):

src/runtime_config/constitution.js
achamayou: Yes, let's refine the actions as we implement them, this is just a testcase for the apply path, not necessarily the final remove_user action.

Issue #669, opened 2020-01-05
Support JS constitution scripts


achamayou:
We should support JS scripts for the constitution, for proposals, and for vote ballots. This will involve replacing the lua interpreter in the member frontend with a JS interpreter, and replacing our sample scripts with JS alternatives.

achamayou:
After discussion with @eddyashton, @letmaik and @jumaffre, these are the main differences with the current Lua governance we have now that I think we want to implement in the Javascript replacement.

First, proposals become a static list of actions rather than a script returning a list of actions. We've not seen any use of the dynamic nature of proposals (ie. their use of KV state as input), on the contrary users so far prefer idempotent actions (#1791, #2169). But we know that it makes error reporting and handling difficult, since actions are not materialised and validation can only happen at voting time.

So this:

```
$ cat new_member.json
{
  "parameter": {
    "cert": "-----BEGIN CERTIFICATE-----\nMIIBdzCCARygAwIBAgIURwD6S1/rcb2TbHhQLnTNh/7WyYYwCgYIKoZIzj0EAwIw\nEjEQMA4GA1UEAwwHbWVtYmVyNDAeFw0yMDEwMjkxNjI2NTNaFw0yMTEwMjkxNjI2\nNTNaMBIxEDAOBgNVBAMMB21lbWJlcjQwVjAQBgcqhkjOPQIBBgUrgQQACgNCAARG\nwqj2ZD7vA+h4KoTdh3if3tVO/yks+xtLU1tXAFsbeWSQfDxK3nnA65uX6n/25A20\nJcAQMDHYH2NdLOLra9lxo1MwUTAdBgNVHQ4EFgQUQQDC71N60r/a9c+EGXrzr5l6\nIDQwHwYDVR0jBBgwFoAUQQDC71N60r/a9c+EGXrzr5l6IDQwDwYDVR0TAQH/BAUw\nAwEB/zAKBggqhkjOPQQDAgNJADBGAiEAkvP0AuAU7y0b3z4rhvoOkCBKoH4G3vh/\nPJpLFdWcEu4CIQCSnEYpDaDTP2zoWTheqchZ+/BdTzM2j2s9ILpvSVYMxg==\n-----END CERTIFICATE-----\n",
[...]

PR #2372, opened 2021-03-29
JS proposal actions

jumaffre:
Part of #669 

This PR defines two JS actions (`set_member_data` and `rekey_ledger`). 

Until https://github.com/microsoft/CCF/pull/2371 is merged, the two new JS actions are executed on each proposal resolve, i.e. way before the proposal is voted for.


src/js/wrap.cpp
```diff
@@ -201,41 +203,41 @@ namespace js
...
-      {
-        break;
-      }
-      default:
-      {
-        throw std::logic_error(fmt::format(
-          "Unhandled AccessCategory for table '{}'", property_name));
-      }
-    }
+    // switch (access_category)
```
jumaffre, line 210: As discussed, commented out for now so that we can mutate the KV from JS actions.

src/runtime_config/constitution.js
```diff
@@ -186,6 +225,22 @@ function resolve(proposal, proposer_id, votes) {
     ) {
       return "Rejected";
     }
+    if (actions_[0].name == "set_member_data") {
```
jumaffre, line 228: The following will need to move to `apply()`.

src/js/wrap.cpp
```diff
@@ -328,6 +330,41 @@ namespace js
     return body_;
   }
 
+  JSValue js_rekey_ledger(
```
letmaik, line 333:
```suggestion
  JSValue js_node_rekey_ledger(
```

src/runtime_config/constitution.js
```diff
@@ -6,6 +6,58 @@ class Action {
...
+          typeof args.member_data == "object"
+        );
+      },
+
+      function (args) {
+        let member_id = ccf.strToBuf(args.member_id);
+        let members_info = ccf.kv["public:ccf.gov.members.info"];
+        let member_info = members_info.get(member_id);
+        if (member_info === undefined) {
+          console.log("Member " + args.member_id + " does not exist");
```
achamayou, line 25:
```suggestion
          console.log(`Member ${args.member_id} does not exist");
```

achamayou (approved):

src/runtime_config/constitution.js
```diff
@@ -5,7 +5,101 @@ class Action {
   }
 }
 
+function toHexString(byteArray) {
+  console.log("To hex: " + byteArray.byteLength);
+  return Array.from(byteArray, function (byte) {
+    console.log("Byte: " + byte);
+    // return ("0" + (byte & 0xff).toString(16)).slice(-2);
```
achamayou, line 12: Needs fixing?

achamayou (approved):

src/runtime_config/constitution.js
jumaffre: No longer needed as we do the conversion from PEM to user ID in a wrapped C++ call.

src/runtime_config/constitution.js
```diff
@@ -6,6 +6,92 @@ class Action {
...
+    new Action(
+      function (args) {
+        return true; // Check that args is null?
+      },
+
+      function (args) {
+        let user_id = ccf.pemToId(args.cert);
+        let raw_user_id = ccf.strToBuf(user_id);
+
+        if (ccf.kv["ccf.gov.users.certs"].get(raw_user_id) != undefined) {
```
eddyashton, line 72: A general JS question: I'd write `!== undefined` here, because the gremlins haunt my nightmares and I like to pretend that JS's `!=` doesn't exist. Is this case safe? Perhaps in the same way that we can use `is None` in Python, where the righthand-side is known to be a single instance?

letmaik (approved):

src/runtime_config/constitution.js
letmaik: Why not `has()`?

src/runtime_config/constitution.js
```diff
@@ -6,6 +6,92 @@ class Action {
...
+            return false; // Internal error
+          }
+
+          ccf.kv["ccf.gov.users.info"].set(
+            raw_user_id,
+            ccf.jsonCompatibleToBuf(args.user_data)
+          );
+        }
+
+        return true;
```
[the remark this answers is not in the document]
jumaffre: My understanding is that if two proposals to add the same user but with different user data are emitted, they are in fact different proposals. So if a member exists already (based on its unique ID, i.e. from its identity cert), the second proposal should fail. As per https://github.com/microsoft/CCF/discussions/2169#discussioncomment-373490, we will have a separate `PATCH` call to change the user data on a member. 

I wonder if this has any implication for the proposal creation conflict logic... Will think about this more tomorrow.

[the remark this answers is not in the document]
eddyashton: I fully agree that they're different proposals, but I think they should both pass.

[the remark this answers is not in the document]
achamayou: Let's put a pin in that to get it the PR merged, I share Eddy's concern fwiw, but we need to make progress.

PR #2374, opened 2021-03-29
JS Digest

achamayou:
Close to https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest, but not quite exactly since the call is synchronous and doesn't return a Promise.


letmaik (approved):

src/crypto/hash.cpp
```diff
@@ -22,6 +22,15 @@ namespace crypto
     return r;
   }
 
+  std::vector<uint8_t> SHA256(const uint8_t* data, size_t len)
```
jumaffre, line 25: Should the `SHA256(const std::vector<uint8_t>&)` function above call into this new one to avoid code duplication?

jumaffre (approved):

src/crypto/hash.cpp
achamayou: I nearly did that, but it involves creating an additional CBuffer that may or may not be optimised out, and it's very little code, so I decided against it. I could be convinced otherwise.

letmaik: No?
```
  std::vector<uint8_t> SHA256(const std::vector<uint8_t>& data)
  {
    return SHA256(data.data(), data.size());
  }

PR #2379, opened 2021-03-30
PEM validation

achamayou:


src/node/rpc/member_frontend.h
```diff
@@ -2354,6 +2352,11 @@ namespace ccf
           "{}\n export default (input) => validate(input);",
           constitution.value());
 
+        js::Runtime rt;
+        js::Context context(rt);
+        rt.add_ccf_classdefs();
+        js::populate_global_ccf(nullptr, std::nullopt, nullptr, context);
```
eddyashton, line 2358: More of a comment:

I thought this line meant we were now exposing a KV access to `validate`, where we hadn't previously. After reading more closely this isn't the case - the first `nullptr` arg results in us _not_ creating a `kv` field on the `ccf` object. I think these relationships aren't very clear, and we should more verbose initialisers so its obvious what's accessible in each JS context (something like `js::add_ccf_node()`, `js::add_ccf_kv()`).

letmaik (approved):

eddyashton (approved):

src/runtime_config/constitution.js
```diff
@@ -104,6 +104,15 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "valid_pem",
+    new Action(
+      function (args) {
+        return ccf.isValidPem(args.pem);
```
letmaik, line 111: What is this checking exactly? Any PEM-style thing? Or just X509? Public key/private key?

achamayou: Anything that makes mbedtls_x509_crt_parse happy, so any chain of x509 certs. `isValidX509Chain`?

jumaffre (approved):

PR #2380, opened 2021-03-30
Support concatenation of constitution shards

achamayou:
Instead of requiring a single file passed to `--constitution`, `cchost` can now take several files (eg. `--constitution c1 --constitution c2 ...`), which it will concatenate into a single constitution to be stored in the KV.

This enables mixing sets of actions with different resolve calls for example and minimises code duplication without introducing the complexity of the full JS module system, which I believe is overkill for governance purposes.


cmake/common.cmake
```diff
@@ -446,7 +447,7 @@ function(add_e2e_test)
       NAME ${PARSED_ARGS_NAME}
       COMMAND
         ${PYTHON} ${PARSED_ARGS_PYTHON_SCRIPT} -b . --label ${PARSED_ARGS_NAME}
-        ${CCF_NETWORK_TEST_ARGS} -g ${PARSED_ARGS_GOV_SCRIPT} --constitution
+        ${CCF_NETWORK_TEST_ARGS} -g ${PARSED_ARGS_GOV_SCRIPT}
         ${PARSED_ARGS_CONSTITUTION} --consensus ${PARSED_ARGS_CONSENSUS}
```
eddyashton, line 451: Should this also remove `${PARSED_ARG_CONSTITUTION}`? It appears to be left as a floating arg, no idea how it now gets processed but looks unintentional?

achamayou: No, this is on purpose, the idea being that it can now be used to pass in stuff like "--constitution fragment1.js --constitution fragment2.js"?

eddyashton: Ah I see: like `CCF_NETWORK_TEST_DEFAULT_CONSTITUTION` does - `PARSED_ARG_CONSTITUTION` now contains multiple elements, including the `--constitution` bit. Makes sense!

eddyashton (approved):

jumaffre (approved):

PR #2381, opened 2021-03-30
Hide details of Tx implementation and commit logic

eddyashton:
The vast majority of the code interacting with a `kv::Tx` only needs to call `.ro`, `.rw` etc on it to access maps. A rarefied smattering of systems (primarily `Frontend` and unit tests) need to call `.commit()`, and then maybe related functions like `.commit_version()` to find out what happened. This splits those types, moving the commit logic out of `kv::Tx` so it is not exposed to the apps.

This public type to be used in endpoints remains named `kv::Tx`, and is moved to `ccf/tx.h` as a public header. The commit logic is in the derived `kv::CommittableTx` - this is what gets created under the hood, but most things see it as a `kv::Tx&`.

Because of the templating, the code which remains in `ccf/tx.h` (`get_handle_by_name<THandle>`) still needs some of the type definitions from `kv_types.h` - ideally these would be private but I don't see a way to break that chain yet.


include/ccf/tx.h
```diff
@@ -0,0 +1,305 @@
...
+
+    // In most places we use NoVersion to indicate an invalid version. In this
+    // case, NoVersion is a valid value - it is the version that the first
+    // transaction in the service will read from, before anything has been
+    // applied to the KV. So we need an additional special value to distinguish
+    // "haven't yet fetched a read_version" from "have fetched a read_version,
+    // and it is NoVersion", and we get that by wrapping this in a
+    // std::optional with nullopt representing "not yet fetched".
+    std::optional<Version> read_version = std::nullopt;
+    ccf::View view = ccf::VIEW_UNKNOWN;
```
eddyashton, line 71: I renamed `Tx.term` to `Tx.view` here, because it made sense at the time (I thought this wouldn't see `Term` any more), but that's an incomplete migration. I'm planning to do a broader rename of these (the variables, types, and getters on `Tx`) to `View` and `SeqNo`, but this PR shows its not part of the public API so its not urgent and can be done later.

src/kv/tx.h
```diff
@@ -1,679 +0,0 @@
...
-    {
-      return ro<M>(m);
-    }
-
-    template <class M>
-    CCF_DEPRECATED("Replace with ro")
-    typename M::ReadOnlyHandle* get_read_only_view(const std::string& map_name)
-    {
-      return ro<M>(map_name);
-    }
```
eddyashton, line 559: These have been present but deprecated for a while - I've removed them now.

src/node/genesis_gen.h
```diff
@@ -59,11 +59,6 @@ namespace ccf
       }
     }
 
-    auto finalize()
-    {
-      return tx.commit();
-    }
-
```
eddyashton, line 66: This meant `GenesisGen` (and everyone creating one) needed a `CommittableTx`, just so we could have hide this final call. Instead I've kept it so `GenesisGen` sees the (now-reduced) `kv::Tx&`, and the caller is responsible for calling `tx.commit()`. In practice this is fine; all of the callers have their `gen_tx` directly available _and_ already know it as a `kv::CommittableTx`.

src/kv/test/kv_dynamic_tables.cpp
```diff
@@ -331,6 +329,8 @@ TEST_CASE("Mixed map dependencies" * doctest::test_suite("dynamic"))
     REQUIRE(tx1.commit() == kv::CommitResult::SUCCESS);
     REQUIRE(tx2.commit() == kv::CommitResult::FAIL_CONFLICT);
 
+    tx2 = kv_store.create_tx();
```
jumaffre, line 332: It seems odd to re-use a transaction here?

src/node/rpc/frontend.h
```diff
@@ -361,7 +363,8 @@ namespace ccf
           // compaction. Reset and retry
           LOG_DEBUG_FMT(
             "Transaction execution conflicted with compaction: {}", e.what());
-          tx.reset();
```
eddyashton, line 364: If I remember correctly, the only reason we had `.reset()` was because `Tx` had no copy-constructor. But the desired behaviour was "set this back to a freshly constructed state", and we had to that by keeping the implementation of `.reset()` in-sync has new members were added? Instead, I've added a move-assign operator to the `Tx`, so now we can do this reset more consistently as a "build me a new one" (but we still have no copy-constructor, so the compiler encourages us to pass-by-ref and not accidentally copy the `Tx`).

jumaffre (approved):

src/kv/kv_types.h
```diff
@@ -372,17 +372,14 @@ namespace kv
   struct PendingTxInfo
   {
     CommitResult success;
-    TxHistory::RequestID reqid;
```
eddyashton, line 375: This was unused, so I've removed it. There's still a `reqid` on each `Tx`, which we put into the `AFT_REQUESTS` table, but we didn't use it when we got it out of this `PendingTxInfo`.

src/kv/test/kv_dynamic_tables.cpp
eddyashton: This is deliberate to mimic the flow in `frontend.h` where the same name is re-used. Previously we'd call `tx2.reset()` inside the conflicting `commit()` on line 330, now we have to "manually reset" by assigning a freshly-constructed transaction. This is equivalent to saying `auto tx4 = ...` here, but by re-using `tx2` we also test the move-assignment that will be used in `frontend.h`.

src/kv/committable_tx.h
```diff
@@ -0,0 +1,363 @@
...
+    void set_root_at_read_version(const crypto::Sha256Hash& r)
+    {
+      root_at_read_version = r;
+    }
+  };
+
+  // Used by frontend for reserved transactions. These are constructed with a
+  // pre-reserved Version, and _must succeed_ to fulfil this version. Otherwise
+  // they create a hole in the transaction order, and no future transactions can
+  // complete.
```
achamayou, line 328:
```suggestion
  // complete. These transactions are used internally by CCF for the sole
  // purpose of recording node signatures and are safe in this particular
  // situation because they never perform any reads and therefore can
  // never conflict.
```

achamayou (approved):
Hiding these details is great, and the inline comments are excellent, thank you!

PR #2382, opened 2021-03-30
js gov: exceptions as errors

letmaik:


src/runtime_config/constitution.js
```diff
@@ -5,64 +5,85 @@ class Action {
...
+  }
+}
+
+function checkEnum(value, members, field) {
+  if (!members.contains(value)) {
+    throw new Error(`${field} must be one of ${members}`);
+  }
+}
+
+function checkBounds(value, low, high, field) {
```
achamayou, line 31: That always puts NaN in bounds, which isn't wrong as such, but seems counter intuitive to me.

achamayou (approved):

src/runtime_config/constitution.js
achamayou: Ah, true!

PR #2386, opened 2021-03-31
Enable code update and reconfiguration test with virtual enclaves

jumaffre:
These two end-to-end tests were disabled for virtual enclaves. For the code update test specifically, we skip the addition/removal of the new code IDs but add the new nodes anyway, and expected the test to succeed.


wintersteiger (approved):
Should `test_verify_quotes` also be guarded?

PR #2387, opened 2021-03-31
Use different channel keys in each direction

wintersteiger:
During the recent changes for #2092 we accidentally skipped the change to different channel keys for each direction of the channel. Also improves error reporting a little.


src/node/channels.h
```diff
@@ -917,14 +928,32 @@ namespace ccf
...
         return;
       }
+      else if (!search->second)
+      {
+        LOG_DEBUG_FMT(
+          "Re-creating new outbound channel to {} ({}:{})",
+          peer_id,
+          hostname,
+          service);
+        search->second = std::make_shared<Channel>(
```
wintersteiger, line 946: @jumaffre I'm not 100% sure we need this branch, but if channels can be closed and re-opened later, then we do.

achamayou (approved):

PR #2388, opened 2021-03-31
js gov: ca cert and jwt proposals

letmaik:
Note that I didn't move over the code of `set_jwt_public_signing_keys` / `remove_jwt_public_signing_keys` since those are still used in C++ for JWT auto-refresh and I didn't want to duplicate the code. Apart from that, `set_jwt_public_signing_keys` is quite complex, involving OE and cert parsing etc.

### DIFF
--- a/src/js/wrap.cpp
+++ b/src/js/wrap.cpp
@@ -7,6 +7,7 @@
 #include "js/conv.cpp"
 #include "js/crypto.cpp"
 #include "kv/untyped_map.h"
+#include "node/jwt.h"
 #include "node/rpc/call_types.h"
 #include "node/rpc/node_interface.h"
 #include "tls/base64.h"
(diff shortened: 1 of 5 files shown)


src/js/wrap.cpp
```diff
@@ -410,6 +411,131 @@ namespace js
...
+        ccf::set_jwt_public_signing_keys(tx, "<js>", issuer, metadata, jwks);
+      if (!success)
+      {
+        return JS_ThrowInternalError(
+          ctx, "set_jwt_public_signing_keys() failed");
+      }
+    }
+    catch (std::exception& exc)
+    {
+      return JS_ThrowInternalError(ctx, "Unexpected error: %s", exc.what());
```
achamayou, line 486: I think "Error" is fine, unexpected seems like a reasonable default assumption in an exception context.

achamayou (approved):

PR #2393, opened 2021-03-31
In JS proposal generator, use JSON.stringify for object equality

eddyashton:
JS has no built-in object equality. This should work, because the fields in the object are defined in the same order (coming from the same Python `args` object).


achamayou (approved):

achamayou:
A deep equality check is a surprising amount of code in JS: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L3267

PR #2394, opened 2021-03-31
Disable perf CI in PRs for now (main only)

achamayou:


wintersteiger (approved):
I think there's a `condition: ...` property that allows you to do that with less clutter, but I might be wrong.

PR #2395, opened 2021-03-31
Idempotent user and member actions

achamayou:


src/runtime_config/actions.js
```diff
@@ -144,25 +150,16 @@ const actions = new Map([
...
-
         ccf.kv["ccf.gov.users.certs"].set(raw_user_id, ccf.strToBuf(args.cert));
 
-        if (args.user_data != null) {
-          if (ccf.kv["ccf.gov.users.info"].has(raw_user_id)) {
-            throw new Error(`User info for ${user_id} already exists`);
-            // Internal error
-          }
-
+        if (args.user_data !== null && args.user_data !== undefined) {
```
eddyashton, line 162: I think this should have an else clause:

```js
if (args.user_data !== null && args.user_data !== undefined) {
  ...
} else {
  ccf.kv["ccf.gov.users.info"].delete(raw_user_id);
}
```

achamayou:
@jumaffre will merge these changes into #2393

PR #2397, opened 2021-04-01
Split test and default actions

achamayou:


tests/governance_js.py
```diff
@@ -345,7 +345,7 @@ def test_actions(network, args):
         r = c.post("/gov/proposals.js", valid_set_member_data)
         assert r.status_code == 200, r.body.text()
 
-        valid_rekey_ledger = proposal(action("rekey_ledger"))
+        valid_rekey_ledger = proposal(action("trigger_ledger_rekey"))
```
achamayou, line 348: Minor drive-by rename to match what we specc'ed in https://github.com/microsoft/CCF/discussions/2169#discussioncomment-373490

src/runtime_config/default_actions.js
```diff
@@ -250,32 +168,11 @@ const actions = new Map([
...
-      function (args) {
-        return true;
-      },
-      function (args) {
-        throw new Error("Error message");
-      }
-    ),
-  ],
   [
     "set_ca_cert_bundle",
```
achamayou, line 172: Note, this doesn't quite match the spec, where we didn't have _bundle: https://github.com/microsoft/CCF/discussions/2169#discussioncomment-373490

I've left it as is because it seems reasonable anyway, @letmaik feel free to correct the other way if you prefer.

letmaik: I've changed it in the spec.

eddyashton (approved):

PR #2399, opened 2021-04-01
Verbose kv logging

letmaik:


src/kv/untyped_map_handle.h
```diff
@@ -82,16 +83,23 @@ namespace kv::untyped
...
+      tx_changes(cs),
+      map_name(map_name)
+    {}
 
     std::optional<ValueType> get(const KeyType& key)
     {
       auto value_p = read_key(key);
-      if (value_p == nullptr)
+      auto found = value_p != nullptr;
+#ifdef VERBOSE_LOGGING
```
eddyashton, line 95: These don't need to be guarded - instead the internals of `LOG_TRACE_FMT` will be enabled when `VERBOSE_LOGGING` is on (but potentially in future also from other things).

eddyashton (approved):

src/kv/untyped_map_handle.h
letmaik: I want to avoid constructing the string etc., even if it is not logged. `VERBOSE_LOGGING` is only used to determine the logging level currently, it should probably also make `LOG_TRACE` a no-op in that case.

letmaik: I removed the guards and we'll do this in the logging macro directly (separate PR).

achamayou (approved):

PR #2400, opened 2021-04-01
js gov: set_js_app / remove_js_app

letmaik:
- set_js_app / remove_js_app as js gov actions
- Build quickjs with bigint support
- Key of endpoints table changed to serialize as string `"<METHOD> <PATH>"`, e.g. "POST /foo"
- removed app_scripts table and instead store js_module & js_function in endpoints table as extra properties
- rename `public:gov.modules` to `public:ccf.gov.modules`
- rename `public:gov.endpoints` to `public:ccf.gov.endpoints`
- rename `public:gov.service_principals` to `public:ccf.gov.service_principals`


achamayou (approved):

cmake/quickjs.cmake
```diff
@@ -30,7 +30,7 @@ if("sgx" IN_LIST COMPILE_TARGETS)
   target_compile_options(
     quickjs.enclave
     PUBLIC -nostdinc -DCONFIG_VERSION="${QUICKJS_VERSION}" -DEMSCRIPTEN
-           -DCONFIG_STACK_CHECK
+           -DCONFIG_STACK_CHECK -DCONFIG_BIGNUM
```
achamayou, line 33: This used to bring in libgmp iirc, which I why I'd turned it off. If it works happily now, great!

Issue #2378, opened 2021-03-30
code_update_test fails on CI


eddyashton:
We have occasional failures in the `code_update_test` during CI runs, which I believe has started recently:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=22126&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=efdbc34d-c6b6-5dc6-e05c-1751d4c47c20&l=16704
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=22138&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=efdbc34d-c6b6-5dc6-e05c-1751d4c47c20&l=16789

Both of these are in SGX Debug. The latter has a segfault node crash, while the former doesn't. Both appear to have a failure to establish a node-to-node channel:

```
62: 10:15:33.702 | ERROR    | infra.remote:log_errors:70 - /mnt/build/1/s/build/workspace/code_update_test_2/out: 2021-03-30T10:15:26.927312Z -0.001 0   [fail ] ../src/node/channels.h:781           | Node channel with 22176ee69ffcc7b157e9ef4b154b0235e04f47f61205b5cb6471692489e0b61f cannot receive authenticated message: not yet established, status=0
62: 10:15:33.702 | ERROR    | infra.remote:log_errors:70 - /mnt/build/1/s/build/workspace/code_update_test_2/out: 2021-03-30T10:15:26.927339Z -0.001 0   [fail ] ../src/consensus/aft/raft.h:679      | Exception in void aft::Aft<consensus::LedgerEnclave, ccf::NodeToNode, ccf::Snapshotter>::recv_message(const ccf::NodeId &, OArray &&) [LedgerProxy = consensus::LedgerEnclave, ChannelProxy = ccf::NodeToNode, SnapshotterProxy = ccf::Snapshotter]
62: 10:15:33.702 | ERROR    | infra.remote:log_errors:70 - /mnt/build/1/s/build/workspace/code_update_test_2/out: 2021-03-30T10:15:27.729962Z -0.001 0   [fail ] ../src/node/channels.h:781           | Node channel with 8c52ba6383d40b50af43c38c87f85d79a754a5af694da0a056bf53a15acfd115 cannot receive authenticated message: not yet established, status=0
[...]

PR #2403, opened 2021-04-01
Fix N2N channel destruction/recreation and logging

wintersteiger:
Fixes at least one more part of #2378. Attempting to send on a channel that has previously been closed is now allowed and triggers a new key exchange.

Also, less `[fail]` logging.

### DIFF
--- a/src/host/node_connections.h
+++ b/src/host/node_connections.h
@@ -318,7 +318,7 @@ namespace asynchost
       decltype(reconnect_queue) local_queue;
       std::swap(reconnect_queue, local_queue);
 
-      for (const auto node : local_queue)
+      for (const auto& node : local_queue)
       {
         LOG_DEBUG_FMT("reconnecting node {}", node);
         auto s = outgoing.find(node);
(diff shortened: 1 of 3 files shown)


achamayou (approved):

jumaffre (approved):

jumaffre:
@wintersteiger Did you manage to identify precisely when this error condition occurred?

wintersteiger:
Yes, our assumption about sending after channel closure being a bug was just wrong. When a node switches roles (e.g. becoming leader) it destroys all channels and later it may create new ones, e.g. when it changes it's role again. So our fix to not `.erase()` the channel meant that it couldn't re-create new ones at that later time, meaning that all messages were dropped and tests failed because nodes didn't respond anymore. With this fix, the `Channel` object are kept alive, but they are `.reset()` (i.e. they forget the keys, etc) and they will correctly resume communication by initiating a new key exchange, if necessary.

eddyashton (approved):

PR #2404, opened 2021-04-01
Make TRACE and DEBUG options only compile-time enabled

eddyashton:
We currently enable `TRACE` and `DEBUG` logging purely at compile time, with `VERBOSE_LOGGING`. However this was only by a run-time short-circuit - the string construction code was still compiled, and could be easily within an enclave at runtime. We'd prefer if it was not possible to do this by accident, so this fully disables those levels at compile-time (their definitions do not exist, their macros are empty nops).

The main knock-on effect of this is a whole bunch of unused variables, parameters, and functions, which were only used by these macros. We can't easily `(void)foo` these, and even if we did it loses the intended benefit of removing their code entirely at compilation. Rather than peppering `[[maybe_unused]]` throughout the code I think we should just disable those unused warnings. We should discuss this, its a big reduction in the lint-style sanity checks the compiler runs - does anyone prefer that we keep them enabled?


cmake/preproject.cmake
```diff
@@ -63,7 +63,14 @@ endif()
...
 function(add_warning_checks name)
   target_compile_options(
-    ${name} PRIVATE -Wall -Wextra -Werror -Wundef -Wpedantic
+    ${name}
+    PRIVATE -Wall
+            -Wextra
+            -Werror
+            -Wundef
+            -Wpedantic
+            -Wno-unused
```
wintersteiger, line 72: Instead, we could disable those warnings just for the logging code via `#pragma`s (like @jumaffre just did in another PR for a different purpose).

wintersteiger (approved):
LGTM. I can live without automatic unused variable/parameter errors.

cmake/preproject.cmake
eddyashton: I don't think we can do this. The pragma disables warnings in a specific scope, but the unused variable exists at an outer scope from the macro

ie
```cpp
void foo(Widget w)
{
  LOG_DEBUG_FMT("My widget is {}", w);
}
```
becomes
```cpp
void foo(Widget w)
{
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wunused-parameter" ...
// No body nop
#pragma clang diagnostic pop
}
```
and the warning remains on the first line. The `Wgnu-zero-variadic-macro-arguments` is a special case since the warning arises on the macro implementation line, so it can be disabled there rather than at the call-site.

achamayou (approved):

jumaffre (approved):

src/ds/logger.h
```diff
@@ -511,17 +504,27 @@ namespace logger
...
+
+#  define LOG_DEBUG \
+    logger::config::ok(logger::DEBUG) && \
+      logger::Out() == logger::LogLine(logger::DEBUG, __FILE__, __LINE__)
+#  define LOG_DEBUG_FMT(s, ...) \
+    LOG_DEBUG << fmt::format(FMT_STRING(s), ##__VA_ARGS__) << std::endl
+#else
+// Without compile-time VERBOSE_LOGGING option, these logging macros are
+// compile-time nops (and cannot be enabled by accident or malice)
+#  define LOG_TRACE
```
letmaik, line 522: This will break if someone actually uses `LOG_TRACE` directly. The only place I found was in merklecpp, and that is currently disabled (see `src/node/history.h`: `// #include "merklecpp_trace.h"`). Given that you would only enable this if you also have `VERBOSE_LOGGING` enabled it should be fine.

cmake/preproject.cmake
wintersteiger: You're right, the macro doesn't include that scope, this wouldn't work.

src/ds/logger.h
wintersteiger: I think we can just drop that file. If we ever need tree-traces it's pretty easy to figure out how to enable them...

PR #2396, opened 2021-04-01
More js actions

jumaffre:
Adds JS actions for node, members and users operations.

PR #2405, opened 2021-04-01
Node gov actions

achamayou:
- transition_node_to_trusted
- remove_node


src/js/wrap.cpp
```diff
@@ -954,6 +1028,16 @@ namespace js
           "Failed to register JS class definition for node");
       }
     }
+
+    // Register class for node
+    {
+      auto ret = JS_NewClass(rt, network_class_id, &network_class_def);
```
jumaffre, line 1034: Should we just go through `NodeState` to access the ledger secrets, instead of registering another class?

src/runtime_config/default_actions.js
```diff
@@ -125,19 +125,22 @@ const actions = new Map([
         let user_id = ccf.pemToId(args.cert);
         let raw_user_id = ccf.strToBuf(user_id);
 
-        if (ccf.kv["ccf.gov.users.certs"].has(raw_user_id)) {
+        if (ccf.kv["public:ccf.gov.users.certs"].has(raw_user_id)) {
```
jumaffre, line 128: As discussed, should also be fixed in #2396

src/runtime_config/default_actions.js
```diff
@@ -418,4 +421,49 @@ const actions = new Map([
...
+    new Action(
+      function (args) {
+        checkType(args.node_id, "string", "node_id");
+      },
+      function (args) {
+        const node = ccf.kv["public:ccf.gov.nodes.info"].get(
+          ccf.strToBuf(args.node_id)
+        );
+        if (node !== undefined) {
+          let node_obj = ccf.bufToJsonCompatible(node);
```
jumaffre, line 435:
```suggestion
          let nodeObj = ccf.bufToJsonCompatible(node);
```

src/js/wrap.cpp
```diff
@@ -892,6 +943,27 @@ namespace js
...
+        throw std::logic_error("Tx should be set to set network context");
+      }
+
+      auto network = JS_NewObjectClass(ctx, network_class_id);
+      JS_SetOpaque(network, network_state);
+      JS_SetPropertyStr(ctx, ccf, "network", network);
+      JS_SetPropertyStr(
+        ctx,
+        network,
+        "latestLedgerSecretSeqno",
```
letmaik, line 959: `getLatestLedgerSecretSeqno` ?

src/js/wrap.cpp
```diff
@@ -954,6 +1028,16 @@ namespace js
           "Failed to register JS class definition for node");
       }
     }
+
+    // Register class for node
```
letmaik, line 1032:
```suggestion
    // Register class for network
```

tests/infra/consortium.py
```diff
@@ -302,9 +302,14 @@ def get_proposals(self, remote_node):
 
     def retire_node(self, remote_node, node_to_retire):
         LOG.info(f"Retiring node {node_to_retire.local_node_id}")
-        proposal_body, careful_vote = self.make_proposal(
-            "retire_node", node_to_retire.node_id
-        )
+        if os.getenv("JS_GOVERNANCE"):
```
jumaffre, line 305: We might as well rename the hard-coded C++ proposal too, so that the user experiencer is somehow smoother if we keep both Lua and JS governance together in the next release?

jumaffre (approved):

src/js/wrap.cpp
achamayou: That seemed awkward because:

1. the abstract node state doesn't expose the network state, which seems deliberate
2. they're both scoped at the same level under Enclave

But I also don't have strong feelings about it, I could be convinced otherwise.

src/runtime_config/default_actions.js
```diff
@@ -418,4 +421,49 @@ const actions = new Map([
...
+    "transition_node_to_trusted",
+    new Action(
+      function (args) {
+        checkType(args.node_id, "string", "node_id");
+      },
+      function (args) {
+        const node = ccf.kv["public:ccf.gov.nodes.info"].get(
+          ccf.strToBuf(args.node_id)
+        );
+        if (node !== undefined) {
```
letmaik, line 434: I would remove this. It should be an error if the node doesn't exist.

src/runtime_config/default_actions.js
```diff
@@ -418,4 +421,49 @@ const actions = new Map([
...
+          let node_obj = ccf.bufToJsonCompatible(node);
+          if (node_obj.status === "Pending") {
+            node_obj.status = "Trusted";
+            node_obj.ledger_secret_seqno = ccf.network.latestLedgerSecretSeqno();
+            ccf.kv["public:ccf.gov.nodes.info"].set(
+              ccf.strToBuf(args.node_id),
+              ccf.jsonCompatibleToBuf(node_obj)
+            );
```
letmaik, line 442:
```suggestion
          const nodeInfo = ccf.bufToJsonCompatible(node);
          if (nodeInfo.status === "Pending") {
            nodeInfo.status = "Trusted";
            nodeInfo.ledger_secret_seqno = ccf.network.latestLedgerSecretSeqno();
            ccf.kv["public:ccf.gov.nodes.info"].set(
              ccf.strToBuf(args.node_id),
              ccf.jsonCompatibleToBuf(nodeInfo)
            );
```

src/runtime_config/default_actions.js
```diff
@@ -418,4 +421,49 @@ const actions = new Map([
...
+    "remove_node",
+    new Action(
+      function (args) {
+        checkType(args.node_id, "string", "node_id");
+      },
+      function (args) {
+        const node = ccf.kv["public:ccf.gov.nodes.info"].get(
+          ccf.strToBuf(args.node_id)
+        );
+        if (node !== undefined) {
```
letmaik, line 458: When is this the case? Are retired nodes deleted at some point? If not, then this check should go.

src/js/wrap.cpp
achamayou: I find it needlessly verbose, change my mind :)

tests/infra/consortium.py
achamayou: I disagree, the point of keeping the old governance around is that it hasn't changed and still works. Changes are only necessary when changing to JS.

src/runtime_config/default_actions.js
achamayou: Ok

src/runtime_config/default_actions.js
achamayou: Nodes aren't deleted now, but they will be. Regardless, a user may pass a node id for which there is no node?

eddyashton: I don't think node deletion matters: The check prevents us writing something if we have a junk `node_id`, for a typo or a node which never existed. `remove_node(Bob)` should not write `Bob: status = Retired` to the KV.

src/js/wrap.cpp
letmaik: I think it would be more consistent with the rest of the exposed functions. With the name `latestLedgerSecretSeqno` I would expect a value property and not a function.

achamayou: Ok, fine.

PR #2406, opened 2021-04-01
Code id gov + misc fixes

achamayou:
All the remaining bits of JS Governance as far as I can tell, including enabling it by default for all the tests, the sandbox etc.

- add_code_id
- remove_code_id
- set_service_principal
- remove_service_principal
- sandbox gov


src/runtime_config/default_actions.js
```diff
@@ -418,4 +418,29 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "add_node_code",
```
jumaffre, line 422: Ditto, should we rename the existing hardcoded C++ proposals to match these too?

achamayou: Let's keep the Lua gov as compatible as possible, as discussed.

CMakeLists.txt
```diff
@@ -500,9 +503,16 @@ if(BUILD_TESTS)
 
   if(ENABLE_JS_GOV)
     set(CONSTITUTION_ARGS
-        --constitution ${CCF_DIR}/src/runtime_config/default_actions.js
-        --constitution ${CCF_DIR}/src/runtime_config/test_actions.js
-        --constitution ${CCF_DIR}/src/runtime_config/constitution.js
+        --constitution
+        ${CCF_DIR}/src/runtime_config/default/actions.js
+        --constitution
+        ${CCF_DIR}/src/runtime_config/test/test_actions.js
```
achamayou, line 509: I had this as "actions.js" first, but for some bizarre reason, CLI11 seems to deduplicate it with `default/actions`. I will look at this later.

metrics.yml
```diff
@@ -2,7 +2,7 @@ db: "ccf"
 collection: "perf"
 view: "trend"
 columns: 4
-span: 20
+span: 150
```
achamayou, line 5: Temporary, to display a longer history, to be removed before merging.

CMakeLists.txt
achamayou: It’s the copy in remote.py.

CMakeLists.txt
```diff
@@ -220,7 +220,10 @@ install(
 
 install(PROGRAMS tests/sandbox/sandbox.sh DESTINATION bin)
 install(FILES tests/sandbox/sandbox_gov.lua DESTINATION bin)
-install(FILES tests/sandbox/sandbox_constitution.js DESTINATION bin)
+install(FILES src/runtime_config/default/actions.js DESTINATION bin)
+install(FILES src/runtime_config/default/validate.js DESTINATION bin)
+install(FILES tests/sandbox/resolve.js DESTINATION bin)
```
achamayou, line 225:
```suggestion
install(FILES src/runtime_config/sandbox/resolve.js DESTINATION bin)
```

eddyashton (approved):

PR #2407, opened 2021-04-01
Remove need for pragma around empty DECLARE_JSON_REQUIRED_FIELDS

eddyashton:
I thought this wasn't possible, because `#define`s can't include `#pragma`s. Today I discovered `_Pragma` and TADA: we can suppress this warning _inside_ the macro that triggers it, rather than wrapping every call-site.


wintersteiger:
Nice! This is new to me too, even though it's apparently in C99.

wintersteiger (approved):

PR #2408, opened 2021-04-02
Add in-enclave crypto benchmark

wintersteiger:
I'm not intending to merge this, this is just an example of how to run picobench benchmarks in an enclave, in case we ever need to run something similar.


achamayou:
I think we may want to merge this after all, the current situation suggests it may be quite useful!

PR #2422, opened 2021-04-07
Join shouldn't be allowed from stale snapshots

jumaffre:
Resolves item 2. in https://github.com/microsoft/CCF/issues/2020 

The snapshot used by new nodes to join the network shouldn't be older than the snapshot that was used by the current primary to join (either on recovery, or a previous join). Otherwise, if the snapshot used to join is older, the new joiner may not be given ledger secrets recent enough to decrypt ledger entries. 

Note that this may be too strict (ledger rekeys may not have happened between an old and a new snapshot), so this PR gives a clear and early error message for operators starting new nodes from snapshots.

### DIFF
--- a/tests/reconfiguration.py
+++ b/tests/reconfiguration.py
@@ -7,6 +7,9 @@ import infra.logging_app as app
 from ccf.tx_id import TxID
 import suite.test_requirements as reqs
 import time
+import tempfile
+from shutil import copy
+import os
 
 from loguru import logger as LOG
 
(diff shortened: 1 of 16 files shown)


tests/reconfiguration.py
```diff
@@ -215,10 +241,70 @@ def run(args):
...
+
 if __name__ == "__main__":
 
     args = infra.e2e_args.cli_args()
     args.package = "liblogging"
     args.nodes = infra.e2e_args.max_nodes(args, f=0)
     args.initial_user_count = 1
-    run(args)
+
+    # run(args)
```
jumaffre, line 309: To re-enable once I'm finished investigating one intermittent issue with the new `run_join_old_snapshot` test function.

tests/reconfiguration.py
```diff
@@ -78,10 +84,19 @@ def test_add_node_from_backup(network, args):
 
 @reqs.description("Adding a valid node from snapshot")
 @reqs.at_least_n_nodes(2)
-@reqs.add_from_snapshot()
 def test_add_node_from_snapshot(
     network, args, copy_ledger_read_only=True, from_backup=False
 ):
+    # Before adding the node from a snapshot, override at least one app entry
```
jumaffre, line 90: Moving this from `test_requirements.py` which was only used historically to avoid code duplication.

achamayou (approved):

PR #2431, opened 2021-04-08
Add indentation to make read_ledger.py collapsible

eddyashton:
As suggested in a comment on original PR, I think this aids navigation of large ledgers. I've also changed the formatting of bytes (to avoid using non-printable characters, make enum values more readable) and removed all the info logging from `ledger.py`. Current output looks like this:

image


jumaffre (approved):
Nice!

wintersteiger (approved):
Awesome, very nice!

PR #2433, opened 2021-04-08
Move service_principals proposals to standalone constitution shard

eddyashton:
The proposals that affect the `service_principal` tables are not default proposals which we expect on every service, instead they're samples of how apps may want to extend the constitution to talk about their own custom tables. This wasn't possible in lua, but is now. This PR moves those proposals to their own constitution shard, and removes any reference to the tables from the core service.


eddyashton:
Parking this as a draft for now - tests for this rely on Lua access for `/gov/read` which is being rewritten in a separate PR. I'll revisit once that's in (and ideally once Lua governance is removed completely).

achamayou (approved):
LGTM but let's mention this in CHANGELOG because it will otherwise almost certainly cause disruption for @lynshi

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.0.0]
```
achamayou, line 8:
```suggestion
## [0.99.2]
```

PR #2434, opened 2021-04-08
Remove oesign shenanigans from CMake

eddyashton:
I added these CMake bodges a while ago to work around `oesign`. `oesign` insisted on a fixed output path. We want debuggable and non-debuggable variants of our enclaves, so I needed to create a temp directory to sign both, otherwise every build would overwrite its own files and cause unnecessary rebuilds. Additionally, I added a `grep|sed|echo` pipe to add a `Debug=0` or `Debug=1` to the `.conf` files used to sign, regardless of what the original `.conf` said.

`oesign` now takes a custom output path as an argument, so we can create the `.debuggable` enclave without overwriting any other files. Additionally, I realised my pipe was unnecessarily complicated - rather than trying to overwrite an existing line, we can just delete any existing `Debug=` lines and add our own at the end. This has a rather hilarious improvement to `cmake-format` time, which really _hated_ this piped `COMMAND` (even when it was correctly formatted):

**Before**
```
$ time cmake-format -i ../cmake/ccf_app.cmake 

real    0m26.391s
user    0m26.362s
sys     0m0.028s
```

**After**
```
$ time cmake-format -i ../cmake/ccf_app.cmake 

real    0m0.380s
user    0m0.372s
sys     0m0.008s
```

### DIFF
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -93,8 +93,10 @@ function(sign_app_library name app_oe_conf_path enclave_sign_key_path)
       OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/lib${name}.so.debuggable
       # Copy conf file locally
       COMMAND cp ${app_oe_conf_path} ${DEBUG_CONF_NAME}
-      # Replace any existing Debug= lines with Debug=1
-      COMMAND sed -i "s/^Debug=\.*/Debug=1/" ${DEBUG_CONF_NAME}
+      # Remove any existing Debug= lines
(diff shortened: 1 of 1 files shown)


wintersteiger (approved):
Hah, nice work! 

Instead of asking sed to just remove the `Debug=` lines, you might as well make it do the substitution?

eddyashton:
> Hah, nice work!
> 
> Instead of asking sed to just remove the `Debug=` lines, you might as well make it do the substitution?

I was going to do this, but have opted against it because the semantics are slightly different (if there's no `Debug=` in the original `.conf`, which some existing CCF apps might have since we're doing this magic for them).

PR #2440, opened 2021-04-09
[Don't Merge] Emulate bad network in CI

eddyashton:
Something we don't currently have great tests for is how CCF performs in adverse network conditions. This is a brute force attempt at a smoke test - use `tc` to emulate a poor connection for everything on localhost (within this Docker image), and then run all of our tests.

This will not be merged, and is simply to see what our CI picks up.


eddyashton:
Doesn't work, `tc` is unavailable in our CI images.

PR #2441, opened 2021-04-09
Add check for double uv_close

wintersteiger:
This adds another (necessary) check on uv_handles to avoid double-`uv_close`s. Pertains to #2378.

### DIFF
--- a/src/host/proxy.h
+++ b/src/host/proxy.h
@@ -79,7 +79,10 @@ namespace asynchost
 
     void close()
     {
-      uv_close((uv_handle_t*)&uv_handle, on_close);
+      if (!uv_is_closing((uv_handle_t*)&uv_handle))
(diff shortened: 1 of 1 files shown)


eddyashton:
Just putting this in writing so we have it for posterity:

Our `uv_handle`s are managed in `close_ptr`s, which call `uv_close` in their destructor. This is safe for most of our handles, which are inited in the constructor then closed in the destructor. The only exception is `TcpImpl`, which is one of these handles that also closes/re-inits the `uv_handle` during normal operation (during reconnection attempts). If we destroy one of these after it has closed, and before it has been inited for a new connection attempt, then we were previously calling `uv_close` in an unsafe case and crashing. This check avoids that.

eddyashton (approved):

Issue #2411, opened 2021-04-06
Remove Lua


eddyashton:
Now that we have JS governance, we can remove Lua governance and the Lua interpreter from CCF.

achamayou:
#2465

PR #2442, opened 2021-04-09
Remove `/gov/read` and `/gov/query` endpoints

jumaffre:
Part of #2411 

This PR removes the existing `/gov/read` and `/gov/query` endpoints which used the Lua interpreter and were mostly used for testing. 

Instead, `ccf.ledger.Ledger` now has two new members `get_transaction(seqno)` to return the `ccf.ledger.Transaction` recorded in the ledger at a specific `seqno` and `get_latest_public_state()` to return the latest public state. The former is used to check that a specific table was updated by a proposal (hence tracking the proposal completion `seqno`) while the latter is used to check the final public state (e.g. to check that JWT refresh works).


src/node/rpc/test/member_voting_test.cpp
```diff
@@ -312,238 +192,6 @@ struct TestNewMember
   crypto::Pem cert;
 };
 
-DOCTEST_TEST_CASE("Add new members until there are 7 then reject")
```
jumaffre, line 315: This made use of `query` and `read` but I decided to scrap it as it is already covered by end-to-end tests.

python/ccf/ledger.py
```diff
@@ -513,6 +518,68 @@ def last_verified_txid(self):
...
+
+        public_tables: Dict[str, Dict] = {}
+        latest_seqno = 0
+        for chunk in self:
+            for tx in chunk:
+                latest_seqno = tx.get_public_domain().get_seqno()
+                for table_name, records in tx.get_public_domain().get_tables().items():
+                    if table_name in public_tables:
+                        public_tables[table_name].update(records)
+                        # Remove deleted keys
```
jumaffre, line 572: Note that the latest state does not include keys deleted keys (value of `None` in delta), although we do include empty tables.

letmaik (approved):

python/ccf/ledger.py
```diff
@@ -513,6 +518,68 @@ def last_verified_txid(self):
...
+        return transaction
+
+    def get_latest_public_state(self) -> Tuple[dict, int]:
+        """
+        Returns the current public state of the service.
+
+        Note that the public state returned may not yet be verified by a
+        signature transaction nor committed by the service.
+
+        :return: Tuple[Dict, int]: Tuple containing a dictionnary of public tables and their values and the seqno of the state read from the ledger.
```
eddyashton, line 558:
```suggestion
        :return: Tuple[Dict, int]: Tuple containing a dictionary of public tables and their values and the seqno of the state read from the ledger.
```

python/ccf/ledger.py
```diff
@@ -513,6 +518,68 @@ def last_verified_txid(self):
...
+        """
+        if seqno < 1:
+            raise ValueError("Ledger first seqno is 1")
+
+        self._reset_iterators()
+
+        transaction = None
+        try:
+            for chunk in self:
+                for tx in chunk:
```
eddyashton, line 538: Non-critical: This iteration is going to be very slow. Could we rely on the filenames here to scan to the relevant chunk, and parse it in isolation? And even within that, could we skip over other transactions rather than parsing them? Not needed for this PR or current tests, but will likely bite us if we start using this `ledger` util in more places/on real service ledgers.

python/ledger_tutorial.py
```diff
@@ -44,3 +45,13 @@
...
+
+# Read state of ledger
+_, latest_seqno = ledger.get_latest_public_state()
+
+seqnos = [1, 2, 3, latest_seqno // 2, latest_seqno]
+random.shuffle(seqnos)
+for seqno in seqnos:
+    transaction = ledger.get_transaction(seqno)
+
+ledger.get_latest_public_state()
```
eddyashton, line 57:
```suggestion
latest_state, latest_seqno = ledger.get_latest_public_state()

seqnos = [1, 2, 3, latest_seqno // 2, latest_seqno]
random.shuffle(seqnos)
for seqno in seqnos:
    transaction = ledger.get_transaction(seqno)

# Confirm latest state can still be accessed, and is unchanged
latest_state1, latest_seqno1 = ledger.get_latest_public_state()
assert(latest_seqno == latest_seqno1)
assert(latest_state == latest_state1)
```

eddyashton (approved):

python/ccf/ledger.py
jumaffre: Yep, I should have mentioned it in the comment. I wanted the fetching to still verify the integrity of the ledger and we currently have no way to do that without starting from the very beginning of the ledger. I'll create a separate ticket and add a comment in the docstring.

jumaffre: @eddyashton See first item in https://github.com/microsoft/CCF/issues/2453 for follow-up improvements.

PR #2443, opened 2021-04-09
Add set_constitution proposal

eddyashton:
This is a proposal which I thought we already had, but seem to never have implemented. It allows the governance scripts to be updated while the service is live, by a proposal. This is only implemented for JS governance. This is one of the special proposals which invalidates all other proposals since it may change their semantics. I have added this `Invalidated` as a new state because I think its worth distinguishing from other types of closed proposal.

### DIFF
--- a/src/node/constitution.h
+++ b/src/node/constitution.h
@@ -4,5 +4,5 @@
 
 namespace ccf
 {
-  using Constitution = kv::JsonSerialisedMap<size_t, std::string>;
+  using Constitution = ServiceMap<size_t, std::string>;
 }
--- a/src/node/constitution.h
+++ b/src/node/constitution.h
@@ -2,6 +2,8 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+#include "service_map.h"
+
 namespace ccf
 {
   using Constitution = ServiceMap<size_t, std::string>;
--- a/src/node/proposals.h
+++ b/src/node/proposals.h
@@ -54,6 +54,9 @@ namespace ccf
     REJECTED, //< Proposal was rejected by vote, will never be enacted
     FAILED, //< Proposal passed a successful vote, but its proposed actions
             // failed, will never be enacted
+    INVALIDATED, //< Proposal was open when its semantics were potentially
+                 // changed (code or constitution were modified), so it was
+                 // automatically invalidated
   };
   DECLARE_JSON_ENUM(
     ProposalState,
@@ -61,7 +64,8 @@ namespace ccf
      {ProposalState::ACCEPTED, "Accepted"},
      {ProposalState::WITHDRAWN, "Withdrawn"},
      {ProposalState::REJECTED, "Rejected"},
-     {ProposalState::FAILED, "Failed"}});
+     {ProposalState::FAILED, "Failed"},
+     {ProposalState::INVALIDATED, "Invalidated"}});
 
   struct Proposal
   {
--- a/src/node/proposals.h
+++ b/src/node/proposals.h
@@ -54,9 +54,9 @@ namespace ccf
     REJECTED, //< Proposal was rejected by vote, will never be enacted
     FAILED, //< Proposal passed a successful vote, but its proposed actions
             // failed, will never be enacted
-    INVALIDATED, //< Proposal was open when its semantics were potentially
-                 // changed (code or constitution were modified), so it was
-                 // automatically invalidated
+    DROPPED, //< Proposal was open when its semantics were potentially changed
+             // (code or constitution were modified), so it was automatically
+             // invalidated and dropped
   };
   DECLARE_JSON_ENUM(
     ProposalState,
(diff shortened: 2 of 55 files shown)


src/node/rpc/member_frontend.h
```diff
@@ -973,16 +975,29 @@ namespace ccf
...
+            argv[1] = prop_id;
+
             auto val = js_context(
-              JS_Call(js_context, apply_func, JS_UNDEFINED, 1, &prop));
+              JS_Call(js_context, apply_func, JS_UNDEFINED, 2, argv));
+
+            // Execution of apply may make arbitrary changes to KV. Refetch
+            // current proposal info in case it has been modified
+            // NB: Can't do this, because we've only modified the .state on our
+            // local copy! pi_ = pi->get(proposal_id);
```
eddyashton, line 995: Reminder to self: We should do this. Maybe not in this PR.

The use case we currently have is "`apply` may modify _every other_ proposal", but this is only a hair's breadth from "`apply` may modify _this_ proposal", and I think we should respect that modification here.

I'd like to audit anywhere else we may have this pattern of a long-held local copy of the KV value - we may need to `put` before `JS_Call`, and then `get` again afterwards.

src/runtime_config/default/actions.js
```diff
@@ -672,6 +711,9 @@ const actions = new Map([
         const codeId = ccf.strToBuf(args.code_id);
         const ALLOWED = ccf.jsonCompatibleToBuf("AllowedToJoin");
         ccf.kv["public:ccf.gov.nodes.code_ids"].set(codeId, ALLOWED);
+
+        // Adding a new allowed code ID changes the semantics of any other open proposals, so invalidate them to avoid confusion or malicious vote modification
+        invalidateOtherOpenProposals(proposalId);
```
jumaffre, line 716: Can we also test for this in `code_update.py`?

src/node/constitution.h
```diff
@@ -4,5 +4,5 @@
 
 namespace ccf
 {
-  using Constitution = kv::JsonSerialisedMap<size_t, std::string>;
+  using Constitution = ServiceMap<size_t, std::string>;
```
eddyashton, line 7: `read_ledger.py` made it trivial to test this, without a bunch of trial-and-error. I think this was `JsonSerialisedMap` rather than `ServiceMap` purely by accident? And now that its accessed from JS, its much easier if it is `ServiceMap` like everything else.

The KV writes to `public:ccf.gov.constitution` from the genesis transaction (with a string constructed in `cchost`) and from this new proposal (with a string constructed in `proposal_generator.py`) are now byte-identical.

jumaffre (approved):

src/node/constitution.h
jumaffre: Spotted by @letmaik, I've included the same change in https://github.com/microsoft/CCF/pull/2442

src/runtime_config/default/actions.js
eddyashton: Yes, I'll add this.

src/node/proposals.h
```diff
@@ -54,14 +54,18 @@ namespace ccf
     REJECTED, //< Proposal was rejected by vote, will never be enacted
     FAILED, //< Proposal passed a successful vote, but its proposed actions
             // failed, will never be enacted
+    INVALIDATED, //< Proposal was open when its semantics were potentially
+                 // changed (code or constitution were modified), so it was
+                 // automatically invalidated
```
eddyashton, line 59: I've just realised this is a potentially confusing name, since one of the constitution functions is `validate` and this is totally unrelated to that concept of validity. Anyone have a better, but still concise, name for "dead because semantics may have changed".

wintersteiger: Obsolete, out-dated, superannuated, antediluvian?

src/runtime_config/default/actions.js
```diff
@@ -123,7 +123,46 @@ function checkX509CertChain(value, field) {
...
+  [
+    "set_constitution",
+    new Action(
+      function (args) {
+        checkType(args.constitution, "string");
+      },
+      function (args, proposalId) {
+        ccf.kv["public:ccf.gov.constitution"].set(
+          getSingletonKvKey(),
+          ccf.strToBuf(JSON.stringify(args.constitution))
```
letmaik, line 158:
```suggestion
          ccf.jsonCompatibleToBuf(args.constitution)
```

src/node/proposals.h
letmaik: stale?

achamayou: Dropped?

src/runtime_config/default/actions.js
achamayou: Why? constitution is a string.

achamayou (approved):

src/node/rpc/member_frontend.h
eddyashton: Tracked in #2454.

src/runtime_config/default/actions.js
letmaik: Yeah, but all ServiceMap's are JSON values.

PR #2449, opened 2021-04-12
Remove msgpack from framework - Part II

jumaffre:
Follow-up from #2343 

`kv::Map` is no longer serialised with msgpack by default. 

TODO: 
- [x] Update docs.
- [x] Move `msgpack` dependency folder to `test/` as we still use `msgpack` for TPCC.


achamayou (approved):

CHANGELOG.md
```diff
@@ -9,11 +9,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-- The service certificate is now returned as part of the `/node/network/` endpoint response.
+- The service certificate is now returned as part of the `/node/network/` endpoint response (#2442).
+
+### Changed
+
+- `kv::Map` is now an alias to `kv::JsonSerialisedMap`, which means all existing applications using `kv::Map` will now require `DECLARE_JSON...` macros (instead of `MSGPACK_DEFINE`) for custom key and value types. Note that this change may affect throughput of existing applications (#2449).
```
eddyashton, line 16: I think we should also mention that msgpack is no longer installed or visible in their include paths. So any `MSGPACK_DEFINE`s in app code, that were previously necessary, actually need to be _removed_ (and replaced with `DECLARE_JSON_...` if they didn't exist.

Also perhaps a suggestion to restore the lost throughput - "use `Blit` if you have simple types, or a custom serialiser"?

src/apps/tpcc/app/tpcc_setup.h
```diff
@@ -131,7 +131,7 @@ namespace tpcc
       {
         bool is_original = selected_rows.find(i) != selected_rows.end();
         Stock s = generate_stock(i, wh_id, is_original);
-        auto stocks = args.tx.rw(tpcc::TpccTables::stocks);
+        auto stocks = args.tx.rw(TpccTables::stocks);
```
eddyashton, line 134: Nit: I think all of the changes in this file are unnecessary? Or am I missing something?

eddyashton (approved):
Code looks good, just think this needs a slightly more explicit warning in the CHANGELOG.

PR #2460, opened 2021-04-13
Document tables, part I

achamayou:
Aiming for complete coverage of `public:ccf.gov.*`

#230


wintersteiger (approved):
LGTM. I guess there isn't an easy way to connect the docs to the code so that we get notified if a table is renamed/removed/added/etc?

achamayou:
@wintersteiger not very easily, but a sanity check may be to scrape the public tables from a ledger post-test case and check they’re at least all mentioned on this page, if not the other way around.

doc/audit/builtin_maps.rst
```diff
@@ -1,29 +1,264 @@
 Built-in Maps
 =============
 
-`public:ccf.gov.nodes.code_ids`
--------------------------------
+`public:ccf.gov.`
```
jumaffre, line 4: Nitpick^2: Why italic?

doc/audit/builtin_maps.rst
```diff
@@ -1,29 +1,264 @@
...
+**Key** Member ID: SHA-256 digest of the member certificate, represented as a hex-encoded string.
+
+**Value** Member certificate, represented as a PEM-encoded string.
+
+`encryption_public_keys`
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Public encryption keys submitted by members to the network. The recovery share for each member is encrypted by the key they have submitted.
+
+**Key** Member ID: SHA-256 digest of the member certificate, represented as a hex-encoded string.
```
jumaffre, line 21: Here and elsewhere:
```suggestion
**Key** Member ID: SHA-256 digest of the DER-encoded member certificate, represented as a hex-encoded string.
```

jumaffre (approved):

doc/audit/builtin_maps.rst
achamayou: I was actually going for monospace, but I markdowned. Monospace seems good for table names? What do we think?

doc/audit/builtin_maps.rst
achamayou: Yes, absolutely, very well spotted, I will fix this everywhere.

doc/audit/builtin_maps.rst
jumaffre: Monospace is great!

doc/audit/builtin_maps.rst
jumaffre: It's also often referred to as the SHA-256 fingerprint of the certificate, which is more concise.

Issue #2448, opened 2021-04-12
Make `ccf` global injection in JS Context modular


achamayou:
Split populate_global_ccf() per namespace, and enable the relevant namespaces only for different governance contexts

letmaik:
I'm bumping against this in #2461 where the new `ccf.node.triggerHostProcessLaunch()` should be available in apps but not all the other `ccf.node.*` functions.

achamayou:
That's not really what I had in mind when I wrote this, I expected namespace granularity would be the correct level of modularity. But your usecase shows that it isn't sufficient, and something more similar to KV access is needed instead (calling context x access level -> capability).

achamayou:
As discussed, let's create `ccf.host.triggerSubprocess()` for this, separately from `ccf.node`.

achamayou:
(very) related to #2961.

achamayou:
Time is better spent working on the containerized API now.

PR #2461, opened 2021-04-13
Allow to launch host processes

letmaik:
This PR allows to trigger host processes from an app endpoint. The app endpoint does not block on it.

- What if the host process sends a response back to CCF (via HTTP) and the relevant app endpoint depends on the transaction that triggered the host process being finished/committed? Does it make sense to investigate an after-commit behaviour?
  - This is ok, assumption is that things are idempotent, so retrying a failed request in this situation would be fine.

- To expose it to JS I temporarily expose the node state as a second argument during `ccf` population. That's a work-around. Related to https://github.com/microsoft/CCF/issues/2448


achamayou:
This looks good to me, as discussed it would be easier and preferable to use libuv to manage spawning and reaping processes.

The execution is best triggered in the transaction itself rather than hooks, so long as it is idempotent there will be no functional ill-effects.

src/host/process_launcher.h
```diff
@@ -0,0 +1,159 @@
...
+#include <chrono>
+#include <queue>
+#include <unordered_map>
+#include <uv.h>
+
+namespace asynchost
+{
+  class ProcessLauncher
+  {
+    static constexpr size_t max_processes = 8;
```
achamayou, line 17: That could really benefit from being a launch-time option.

letmaik: Sure, but since it's an experimental feature let's wait for feedback first.

achamayou: I think the best way to know that it's an experimental feature would be for it to be spelled out as such in the changelog ;)

achamayou (approved):

Issue #2456, opened 2021-04-12
Retired nodes becoming followers


wintersteiger:
The rotation_test (and likely others) frequently results in nodes changing from the retired state to the follower state. In the log files I see things like these:

```
... $ for i in workspace/rotation_test_*/out; do echo $i; grep Becoming $i; done
workspace/rotation_test_0/out
2021-04-12T14:18:34.197290Z -0.091 0   [info ] ../src/consensus/aft/raft.h:2347     | Becoming leader 8ea41e38682cfd5c87528bef3337c7f08464f93e68ffce423e8ba79edb3bfc33: 2
2021-04-12T14:18:45.575498Z -0.003 0   [info ] ../src/consensus/aft/raft.h:2398     | Becoming retired 8ea41e38682cfd5c87528bef3337c7f08464f93e68ffce423e8ba79edb3bfc33: 2
2021-04-12T14:18:47.943018Z -0.002 0   [info ] ../src/consensus/aft/raft.h:2384     | Becoming follower 8ea41e38682cfd5c87528bef3337c7f08464f93e68ffce423e8ba79edb3bfc33: 3
workspace/rotation_test_1/out
2021-04-12T14:18:36.278080Z -0.002 0   [info ] ../src/consensus/aft/raft.h:2384     | Becoming follower d8c06b7581f77496f9c46e1b3fc76c0d4fb9e819e27beb12a4d70585fd5ab6ff: 2
2021-04-12T14:18:47.915849Z -0.002 0   [info ] ../src/consensus/aft/raft.h:2307     | Becoming candidate d8c06b7581f77496f9c46e1b3fc76c0d4fb9e819e27beb12a4d70585fd5ab6ff: 3
2021-04-12T14:18:48.076984Z -0.002 0   [info ] ../src/consensus/aft/raft.h:2347     | Becoming leader d8c06b7581f77496f9c46e1b3fc76c0d4fb9e819e27beb12a4d70585fd5ab6ff: 3
2021-04-12T14:18:55.822770Z -0.058 0   [info ] ../src/consensus/aft/raft.h:2398     | Becoming retired d8c06b7581f77496f9c46e1b3fc76c0d4fb9e819e27beb12a4d70585fd5ab6ff: 3
[...]

PR #2464, opened 2021-04-13
Fix N2N channel exception behaviour

wintersteiger:
This reverts part of a previous change to the N2N channel exception behaviour. There is a bug in `Node2Node::recv_authenticated`; it should not return `t` when a message is dropped. This caused outdated (and potentially unauthenticated) messages to be delivered, which is likely the only reason for #2456. 

The ugly part is that this brings back failure log messages that aren't really failures; for instance:

```
43: 17:39:49.084 | ERROR    | infra.remote:log_errors:70 - /data/cwinter/CCF/build/workspace/rotation_test_3/out: 2021-04-13T17:39:48.114058Z -0.005 0   [fail ] ../src/consensus/aft/raft.h:678      | Exception in void aft::Aft<consensus::LedgerEnclave, ccf::NodeToNode, ccf::Snapshotter>::recv_message(const ccf::NodeId &, OArray &&) [LedgerProxy = consensus::LedgerEnclave, ChannelProxy = ccf::NodeToNode, SnapshotterProxy = ccf::Snapshotter]
```

This is hard to interpret for users and I would much prefer these to be info log messages (about dropping the message) or to be completely silent. So, perhaps 

```
if (!recv_authenticated(from, asCb(t), data, size))
{
  LOG_INFO_FMT("Dropping invalid authenticated message from {}", from);
  t = T();
}
```

is preferable?

### DIFF
--- a/src/node/node_to_node.h
+++ b/src/node/node_to_node.h
@@ -19,6 +19,13 @@ namespace ccf
   public:
     virtual ~NodeToNode() = default;
 
+    class DroppedMessageException : public std::exception
+    {
+    public:
+      NodeId from;
+      DroppedMessageException(const NodeId& from) : from(from) {}
+    };
+
     virtual void create_channel(
       const NodeId& peer_id,
       const std::string& peer_hostname,
@@ -56,9 +63,7 @@ namespace ccf
 
       if (!recv_authenticated(from, asCb(t), data, size))
       {
-        throw std::logic_error(fmt::format(
-          "Invalid authenticated node2node message with load from node {}",
-          from));
+        throw DroppedMessageException(from);
       }
 
       return t;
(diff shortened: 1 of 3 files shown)


achamayou (approved):

src/node/node_to_node.h
```diff
@@ -56,7 +56,9 @@ namespace ccf
 
       if (!recv_authenticated(from, asCb(t), data, size))
       {
-        LOG_INFO_FMT("Dropping invalid authenticated message from {}", from);
+        throw std::logic_error(fmt::format(
```
jumaffre, line 59: We could always throw a new node-to-node-specific exception here and catch it separately in `raft.h`'s `recv_message()`, and display the "Dropping invalid message from..." message there?

jumaffre (approved):

src/node/node_to_node.h
wintersteiger: Sure, can do. We don't catch any other custom exceptions in `recv_message()` - would it be acceptable to do that there?

Issue #2437, opened 2021-04-08
CCF Crashing under load


rajdhandus:
**Describe the bug**
CCF Nodes crashing under load

**To Reproduce**
Run 100 TPS transactions through a CCF Network installed in a Kubernetes environment

**Environment information**
Version of the code being used, versions of dependencies, relevant environment info where applicable (OS, SGX driver...).

Ubuntu 18.04
CCF 0.19.3

**Additional context**

```
[...]

PR #2467, opened 2021-04-14
Expand connections test to better understand fd-exhaustion behaviour

eddyashton:
Potential fix for #2437, although I've not seen exactly the same error case.

Our connection test set a low file descriptor (fd) limit the primary, and then created enough connections to this node that we actually used all available fds, and checked that the client saw this as a timeout error but the node survived. However the node only survived because we also disabled other fd-acquisition, by increasing the ledger chunking and snapshot creation thresholds so they were never used. In practice, if we have so many sessions that we hit our fd-cap, the node may legitimately be stuck when it tries to do file operations, and we should test that behaviour. The operational mitigation for this is to limit the number of sessions to be lower than the fd limit, so that sessions are refused _before_ they get in the way of the node's file operations.

This PR adds a `--max-open-sessions` param so that we can adjust the session limit at service launch time, rather than only at compile-time. It also modifies `connections_test` to test this behaviour more precisely. As long as `max_open_sessions` is << `max_fds`, the nodes will survive even after they're refusing client connections. If `max_open_sessions` > `max_fds`, then client connections can use up the entire fd-quota, and the node will likely crash the next time it tries to create a ledger chunk. Avoiding this crash would non-trivial, since it involves changing the chunking logic to store a single uncommitted chunk and potentially having different chunks on different nodes.

I've also reduced the severity of one place where we might notice we have no fds. If we throw in `commit_snapshot` (or any snapshot operation) we can simply report a `LOG_FAIL` and continue - snapshots are best-effort and non-critical.


jumaffre (approved):

src/enclave/rpc_sessions.h
```diff
@@ -48,6 +48,13 @@ namespace enclave
       to_host = writer_factory.create_writer_to_outside();
     }
 
+    void set_max_open_sessions(size_t n)
+    {
+      max_open_sessions = n;
+
+      LOG_INFO_FMT("Setting max open sessions to {}", n);
```
ashamis, line 55: should this be debug level?

src/host/snapshot.h
```diff
@@ -165,38 +165,46 @@ namespace asynchost
...
+          }
         }
-      }
 
-      LOG_FAIL_FMT("Could not find snapshot to commit at {}", snapshot_idx);
+        LOG_FAIL_FMT("Could not find snapshot to commit at {}", snapshot_idx);
+      }
+      catch (std::exception& e)
+      {
+        LOG_FAIL_FMT(
```
ashamis, line 205: this message should probably include the snapshot_idx

ashamis (approved):

src/enclave/rpc_sessions.h
eddyashton: No - this is set by the host (so non-confidential), but happens after construction (so useful to log when it _has_ happened).

PR #2476, opened 2021-04-15
Record final votes and collect complete proposals

achamayou:
Final bits of #669, still missing:

- [x] record vote failures in final tally
- [x] tests, details below

In `member_voting_test.cpp`
- [x] "Proposer ballot": Create proposal, vote, check storage and status. Vote again, check applied. Covered by `test_proposals_with_votes`, `test_proposal_storage` and `test_ballot_storage`
- [x] "Reject duplicate vote": Create proposal, vote, try to vote again.
- [x] "Accept node": Propose trusting node, vote, propose removal, vote. Observe state, check that node cannot be trusted or retired again. N/A once nodes are completely deleted, tests a specific action, not governance.
- [x] "Propose raw writes": Test "raw_puts" on sensitive tables. N/A in JS governance.
- [x] "Remove proposal": Attempt to remove non-existent or disallowed proposals. Legitimate withdrawal. Covered by `test_proposal_withdrawal`.
- [x] "Vetoed proposal gets rejected". Tests a specific single-vote reject, but we test rejects in several places such as `test_proposals_with_votes`.
- [x] "Add and remove user via proposed calls". Tests a specific action, not governance, covered by `test_actions`.
- [x] "Passing members ballot with operator". Tests operator proposal flow. Covered by `test_operator_proposals_and_votes`.
- [x] "Passing operator change". Tests operator proposal flow. Covered by `test_operator_proposals_and_votes`.
- [x] "Members passing an operator change". Tests details of the operator constitution, not governance.
- [x] "User data". Tests specific proposal, not governance, covered by `test_actions`.
- [x] "Submit recovery shares". Covered by recovery test and `test_actions`.
- [x] "Number of active members with recovery shares limits", covered by `test_actions`.


src/node/gov.h
```diff
@@ -12,33 +12,53 @@ namespace ccf
...
+    {
+      std::string reason;
+      std::string trace;
+      bool operator==(const Failure& rhs) const
+      {
+        return reason == rhs.reason && trace == rhs.trace;
+      }
+    };
+    DECLARE_JSON_TYPE(Failure);
+    DECLARE_JSON_REQUIRED_FIELDS(Failure, reason, trace);
```
eddyashton, line 28: Minor: Since `trace` is sometimes/often the empty string (if `js_error_message` doesn't return one), I think it should be an optional field here so we don't insert unnecessary empty string fields in the JSON serialisation.

eddyashton (approved):

src/node/gov.h
achamayou: Yes, good point.

PR #2478, opened 2021-04-16
Remove OE logging

eddyashton:
Swallow the OE logging so it doesn't log "logging is not enabled" in the wrong format.

Goodbye forever to this message:

```
2021-04-16T16:39:20+0100.498499Z [(H)WARN] tid(0x7ff6502ee600) | In-enclave logging is not supported. To enable, please add 

from "openenclave/edl/logging.edl" import *;

in the edl file.
 [/source/openenclave/host/sgx/create.c:oe_create_enclave:1178]
```

### DIFF
--- a/src/host/enclave.h
+++ b/src/host/enclave.h
@@ -14,7 +14,21 @@
 #  include <ccf_u.h>
 #  include <openenclave/bits/result.h>
 #  include <openenclave/host.h>
+#  include <openenclave/trace.h>
 #endif
+
+extern "C"
+{
+  void nop_oe_logger(
+    void* context,
+    bool is_enclave,
+    const struct tm* t,
+    long int usecs,
+    oe_log_level_t level,
+    uint64_t host_thread_id,
+    const char* message)
+  {}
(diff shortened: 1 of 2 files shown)


wintersteiger (approved):
❤️

src/host/enclave.h
```diff
@@ -56,6 +70,8 @@ namespace host
       }
       else
       {
+        oe_log_set_callback(nullptr, nop_oe_logger);
```
achamayou, line 73: Can we keep oe logging when VERBOSE_LOGGING is on?

achamayou (approved):

src/host/enclave.h
letmaik: What's the resolution here?

eddyashton: I missed this - will disable this nop logger when `VERBOSE_LOGGING` is on so we restore the old behaviour. I believe that's still a single line though, unless apps can turn on more OE logging themselves?

PR #2480, opened 2021-04-19
Do not stop enclave when execution of thread message throws

jumaffre:


wintersteiger (approved):

wintersteiger:
I don't think it's the case here, but it could conceivably enter an infinite loop, constantly re-throwing the same exception?

achamayou (changes requested):
What's the context for this? Are we sure it's safe?

jumaffre:
As discussed with @achamayou, abandoning this as we should still shut down the enclave if an exception bubbles up to here, as per the previous changes in https://github.com/microsoft/CCF/pull/2352.

PR #2482, opened 2021-04-19
Read multiple directories to construct Python ledger

eddyashton:
Fix for Daily CI. When a node's ledger is split across multiple directories, we should read from all of them to reconstruct the full ledger.

### DIFF
--- a/python/ccf/ledger.py
+++ b/python/ccf/ledger.py
@@ -500,22 +500,21 @@ class Ledger:
 
         self._filenames = []
 
-        ledgers = []
+        ledger_files = []
         for directory in directories:
-            ledgers += os.listdir(directory)
+            for path in os.listdir(directory):
+                chunk = os.path.join(directory, path)
+                if os.path.isfile(chunk):
+                    ledger_files.append(chunk)
 
         # Sorts the list based off the first number after ledger_ so that
         # the ledger is verified in sequence
-        sorted_ledgers = sorted(
-            ledgers,
+        self._filenames = sorted(
+            ledger_files,
             key=lambda x: int(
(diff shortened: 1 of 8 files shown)


python/ccf/ledger.py
```diff
@@ -496,24 +496,29 @@ def _reset_iterators(self):
...
-    def __init__(self, directory: str):
+    def __init__(self, directories: List[str]):
 
         self._filenames = []
 
-        ledgers = os.listdir(directory)
+        ledger_files = []
+        for directory in directories:
+            for path in os.listdir(directory):
+                chunk = os.path.join(directory, path)
```
jumaffre, line 506: Should we be defensive here and check that `path` hasn't already been read, e.g. if the same directory is passed twice, or if the same ledger chunk (based on its name only) exists in two directories?

jumaffre (approved):

python/ccf/ledger.py
achamayou: If the same chunk occurs twice, there will be a validation error in the signatures. Properly making sure there are no duplicates (ie. resolving symlinks etc) may be more trouble than it's worth in practice.

achamayou (approved):

python/ccf/ledger.py
jumaffre: Fair point. I thought we could easily prevent suspicious signature verification errors here, but resolving symlinks, etc. seems too convoluted indeed.

achamayou: I suspect we can use the parsing of the sequence numbers to implement an easy but still helpful check.

PR #2483, opened 2021-04-19
Add release policy detail

### LINKED ISSUE #2234: Document release schedule and compatibility
We should document how often CCF versions are released and which level of compatibility users should expect between these versions. 

In particular, with `N` being a major release, and with some degree of overlap:

0. Minor releases `N.x` should be compatible: nodes built from the same major release but different minor releases can run within the same service (backward and forward compatibility between these).
1. Users should expect backward compatibility of the ledger forever: a ledger produced by version `N.x` can be read by all versions > `N.x`.
2. Users should expect forward compatibility of the ledger across minor releases: a ledger produced by versions `N.x` can be read by versions < `N.x` and >= `N.0`.
3. Snapshots can only be compatible across incremental versions (going from `N.x` to `N+1.x`), but as they use the same serialisation format as the ledger entries, this should be true forever too (although not used much in practice).
4. Receipts include a version format (e.g. to indicate which fields are signed by the service, etc.).

achamayou:
Fix #2234.


doc/contribute/index.rst
```diff
@@ -28,10 +28,18 @@ The CCF team welcome contributions to any part of the framework, including this
 
     Build CCF from source.
 
+    ---
+
+    :fa:`rocket` :doc:`release_policy`
```
jumaffre, line 33: This seems like an odd page for this tile as release support is also shared across members, users and operators. Should this go on the main index, or under "overview"?

doc/contribute/release_policy.rst
```diff
@@ -0,0 +1,59 @@
...
+-----------------------------
+
+REST endpoints exposed by CCF itself fall into three categories:
+
+User-facing API
+~~~~~~~~~~~~~~~
+
+As defined under :ref:`use_apps/rpc_api:Built-ins`.
+
+We expect these endpoints to be quite stable, but it is difficult for CCF itself to version them because its own versioning scheme may not be in-sync with a given application: release dates and numbers will differ.
```
jumaffre, line 14: Avoid first-person pronouns in docs?

doc/contribute/release_policy.rst
```diff
@@ -0,0 +1,59 @@
...
+Fields/input arguments are never be modified/deleted unless exceptionally and explicitly notified in advance to users.
+
+Member-facing API
+~~~~~~~~~~~~~~~~~
+
+As defined under :ref:`governance/member_rpc_api`.
+
+Same as operator-facing API.
+
+Support policy
```
jumaffre, line 48: Can we be a little more specific here, e.g. mentioning some of the items on https://github.com/microsoft/CCF/issues/2234?

doc/contribute/index.rst
achamayou: I hesitated between here and overview, but overview seemed crowded, and if you squint a bit it's a project matter. I'm happy to switch it though, I don't feel strongly about it.

doc/contribute/release_policy.rst
achamayou: Yes, thank you!

doc/contribute/release_policy.rst
achamayou: Ok, will expand

doc/contribute/release_policy.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Same as operator-facing API.
+
+Operations compatibility
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Patches are compatible: nodes built from the same major release but different patches releases can run within the same service (backward and forward compatibility between these).
+- The last patch in a major release (`N.0.x`) is compatible with the first patch of the next major release (`N+1.0.x`): nodes built against both versions can be run within the same service.
+- Users can expect open-ended ledger backward compatibility: a ledger produced by version `N.0.x` can be read by all versions `> N.0.x`.
+- Users can expect forward compatibility of the ledger across patches: a ledger produced by version `N.0.x` can be read by versions `>= N.0.*`.
```
jumaffre, line 55: Isn't this a subset of the previous point?

doc/contribute/release_policy.rst
```diff
@@ -0,0 +1,69 @@
...
+As defined under :ref:`governance/member_rpc_api:Member RPC API`.
+
+Same as operator-facing API.
+
+Operations compatibility
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Patches are compatible: nodes built from the same major release but different patches releases can run within the same service (backward and forward compatibility between these).
+- The last patch in a major release (`N.0.x`) is compatible with the first patch of the next major release (`N+1.0.x`): nodes built against both versions can be run within the same service.
+- Users can expect open-ended ledger backward compatibility: a ledger produced by version `N.0.x` can be read by all versions `> N.0.x`.
```
jumaffre, line 54: Nitpick: could we simplify the sentences here, e.g. "- Open-ended ledger backward compatibility: ..."

doc/contribute/release_policy.rst
achamayou: The previous point is strict backwards compatibility, this point tries to be forward compatibility within a major release, ie. a ledger produced by N.0.5 can be read by N.0.2. I will try to re-phrase it.

doc/contribute/release_policy.rst
achamayou: Yes.

Point of order: I propose we never write the word "Nitpick". Instead reviewer either asks for a change (small changes are fine!), or if they feel that it is may be perceived as a nitpick they can use "Insert a suggestion" to make it a zero-cost change for the reviewee.

doc/contribute/release_policy.rst
```diff
@@ -0,0 +1,69 @@
...
+
+APIs use ``_v${INCREMENTING_INTEGER}`` as a symbol suffix, starting at 1. Symbol versions are connected with CCF releases by release notes “eg. CCF release X.Y.Z introduces call_v2”.
+
+Operator-facing API
+~~~~~~~~~~~~~~~~~~~
+
+As defined under :ref:`operations/operator_rpc_api:Operator RPC API`.
+
+This is the API used to monitor the network topology, memory usage, endpoint metrics etc. The intention is to keep this API compatible without explicit versioning, by making sure that all changes are strict additions (ie. new fields, new arguments with default values that behave identically to the old call).
+Fields/input arguments are never be modified/deleted unless exceptionally and explicitly notified in advance to users.
```
eddyashton, line 40:
```suggestion
Fields/input arguments will never be modified/deleted unless exceptionally and explicitly notified in advance to users.
```

doc/contribute/release_policy.rst
```diff
@@ -0,0 +1,69 @@
...
+
+As defined under :ref:`governance/member_rpc_api:Member RPC API`.
+
+Same as operator-facing API.
+
+Operations compatibility
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Patches are compatible: nodes built from the same major release but different patches releases can run within the same service (backward and forward compatibility between these).
+- The last patch in a major release (`N.0.x`) is compatible with the first patch of the next major release (`N+1.0.x`): nodes built against both versions can be run within the same service.
```
eddyashton, line 53:
```suggestion
- The last patch in a major release (`N.0.x`) is compatible with the first patch of the next major release (`N+1.0.0`): nodes built against both versions can be run within the same service.
```

Shouldn't re-use `x` for different values, I think the 'first patch of the next major release' will always be `.0`?

doc/contribute/index.rst
achamayou: Moved now

doc/contribute/release_policy.rst
achamayou: Done

doc/contribute/release_policy.rst
achamayou: Re-phrased, hopefully more clearly.

jumaffre (approved):

PR #2486, opened 2021-04-20
Validate entity IDs in default proposal actions

### LINKED ISSUE #2444: Add more validation to default governance Actions
The `validate` checks for current proposals are fairly loose, and allow many invalid proposals. We should tighten these up, for instance by checking that all `EntityID`s contain only hex chars and are of the correct length.

eddyashton:
Resolves #2444.


src/runtime_config/default/actions.js
```diff
@@ -74,6 +74,17 @@ function checkNone(args) {
   }
 }
 
+function checkEntityId(value, field) {
+  checkType(value, "string", field);
+  if (value.length !== 64) {
```
jumaffre, line 79: Perhaps define `64` as a `const` around this line, along with a comment to explain this is the hex-encoded string of a SHA-256 hash?

jumaffre (approved):

achamayou (approved):

PR #2487, opened 2021-04-20
Update docs

eddyashton:
Primarily removing references to `/votes` endpoint that doesn't exist, also updating a few samples to use/talk about JS gov rather than Lua.


achamayou (approved):

src/node/rpc/test/frontend_test_infra.h
```diff
@@ -139,7 +139,7 @@ auto get_vote(
 {
   const auto getter = create_request(
     nullptr,
-    fmt::format("proposals/{}/votes/{}", proposal_id, voter),
```
achamayou, line 142: That sounds unused.

eddyashton: Yes - there's several functions in `frontend_test_infra.h` that are only used by `member_voting_test`, so can be removed with it.

achamayou: Ok, I'll have a look for them.

Issue #2235, opened 2021-02-24
Ledger entry compatibility


jumaffre:
Our ledger frames (currently documented [here](https://microsoft.github.io/CCF/main/build_apps/kv/kv_serialisation.html#serialised-transaction-format)) should:
- [ ] Be versioned so that an old ledger can be deserialised by a newer node (point 1. in #2234)
- [ ] Include some headers so that a single ledger entry can include multiple elements (e.g. hash of the request and response), as opposed to just the serialised transaction write set.

PR #2488, opened 2021-04-20
Ledger versioning

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/2235

This PR adds support for ledger versioning, with a new 8-byte header (constructed by the KV, i.e. in enclave) which contains:
- The version of the serialisation format (1 byte)
- A set of flags, unused for now (1 byte)
- The size of the serialised entry (6 bytes)

Most of the changes are in the `ledger.h` which no longer automatically prefixes entries with their size, and in the KV serialisation code, plus the mandatory changes in `ledger.py`.


src/kv/deserialise.h
```diff
@@ -118,7 +118,9 @@ namespace kv
 
       if (history)
       {
-        history->append(data);
+        history->append(
```
jumaffre, line 121: These are the kind of changes required because we decided not to include this new header in the Merkle tree (see also `historical_queries.h`, `store.h`). In hindsight, while it wouldn't provide much security guarantee, I think it's pretty inelegant and I'm now keen to include it in the hash of the Merkle Tree. @achamayou, @eddyashton thoughts?

wintersteiger: +1, I also think the header should be included in the data. Makes everything much easier to read and interpret. I suppose the ledger would grow a little because of the copies of identical headers, but that's acceptable I think. I have no idea what I'm talking about, but it seems to me that it does at least provide a record of ill behaviour on the ledger if someone tries to play games with old version numbers (but that's probably not really necessary). I think it also means that ledger verification is less stateful because we don't need to remember headers/versions/flags of previous entries.

eddyashton: I agree this is inelegant. The problem here is that we have some internal structure that we're accessing in lots of places. For argument's sake, I see 2 other approaches other than adding the header to the Merkle tree:
- Hide this indirection in a new type. We could return this type from `serialised` (rather than the raw `std::vector`), or it could just have a bunch of static functions that operate over a `std::vector`. So this would be something like `history->append(LedgerEntry::hashed_body(data))`, but with a better name for "the bit that goes in the Merkle tree" than `hashed_body`.
- Extract these internal bits _inside_ `history->append`. So `append` takes a `SerialisedLedgerEntry` (aka a `std::vector`, but to be clear this is an entire `Entry`), and skips the prefix. If we have raw `append`s elsewhere, we could do this with 2 functions (`append_raw` vs `append_from_entry`? Naming is hard).

doc/audit/ledger.rst
```diff
@@ -5,31 +5,33 @@ Ledger
...
+The ledger is split into multiple files (or chunks). Each ledger file starts with an 8-byte offset pointing to the end of file where a vector of positions is stored for fast fetching of transactions (only set when a file is complete, internal use only).
 
-Ledger entries are integrity-protected and encrypted using a symmetric key shared by all trusted nodes (see :doc:`/overview/cryptography`). This key is kept secure inside each enclave. See :ref:`governance/common_member_operations:Rekeying Ledger` for details on how members can rotate the ledger encryption key.
+The serialised transactions follow the 8-byte offset, as described in :ref:`audit/ledger:Transaction Format`.
 
-Note that even if a transaction only affects a private ``Store::Map``, unencrypted information such as the sequence number is always present in the serialised entry.
+.. note:: A completed ledger file always ends on a signature transaction.
```
achamayou, line 18:
```suggestion
.. note:: A complete ledger file always ends on a signature transaction.
```

doc/audit/ledger.rst
```diff
@@ -71,4 +73,15 @@ The following table describes the structure of a serialised KV Store transaction
...
-+----------+--------------------------------------------------------------------------------------------------------------------+
\ No newline at end of file
++----------+--------------------------------------------------------------------------------------------------------------------+
+
+Transaction Encryption
+----------------------
+
+Each entry in the ledger corresponds to a transaction committed by the primary node.
+
+When a transaction is committed, each affected ``Store::Map`` is serialised in different security domains (i.e. public or private), based on the name of the Map when it was created (default is private). A public ``Store::Map`` (i.e. one whose name starts with "public:") is serialised and stored in the ledger as plaintext while a private ``Store::Map`` is serialised and encrypted before being stored.
```
achamayou, line 83:
```suggestion
When a transaction is committed, each ``Store::Map`` containing writes is serialised in different security domains (i.e. public or private), based on the name of the Map when it was created (default is private). A public ``Store::Map`` (i.e. one whose name starts with "public:") is serialised and stored in the ledger as plaintext while a private ``Store::Map`` is serialised and encrypted before being stored.
```

doc/audit/ledger.rst
```diff
@@ -71,4 +73,15 @@ The following table describes the structure of a serialised KV Store transaction
...
+
+Ledger entries are integrity-protected and encrypted using a symmetric key shared by all trusted nodes (see :doc:`/overview/cryptography`). This key is kept secure inside each enclave. See :ref:`governance/common_member_operations:Rekeying Ledger` for details on how members can rotate the ledger encryption key.
+
+Note that even if a transaction only affects a private ``Store::Map``, unencrypted information such as the sequence number is always present in the serialised entry.
```
achamayou, line 87:
```suggestion
Note that even if a transaction only writes to a private ``Store::Map``, unencrypted information such as the sequence number is always present in the serialised entry.
```

src/host/ledger.h
```diff
@@ -205,29 +205,36 @@ namespace asynchost
...
-          len -= frame_header_size;
+          len -= kv::serialised_entry_header_size;
 
+          const auto& entry_size = entry_header.size;
           if (len < entry_size)
           {
-            throw std::logic_error(
-              fmt::format("Malformed ledger file {}", file_path));
+            throw std::logic_error(fmt::format(
+              "Malformed uncompleted ledger file {} (expecting entry of size "
```
achamayou, line 226:
```suggestion
              "Malformed incomplete ledger file {} (expecting entry of size "
```

src/host/test/ledger.cpp
```diff
@@ -111,15 +105,16 @@ void verify_framed_entries_range(
   const std::vector<uint8_t>& framed_entries, size_t from, size_t to)
 {
   size_t idx = from;
-  for (int i = 0; i < framed_entries.size();)
+  for (int pos = 0; pos < framed_entries.size();)
```
achamayou, line 108:
```suggestion
  for (size_t pos = 0; pos < framed_entries.size();)
```

src/kv/deserialise.h
achamayou: > I think it also means that ledger verification is less stateful because we don't need to remember headers/versions/flags of previous entries.

The primary motivation for the introduction of flags is to enable custom claims, in which case the ledger verification will necessarily require more interpretation (likely a small mini-tree for each entry), so this isn't going to stay simple for very long.

Since these flags will (among other things) communicate to an auditor how to build the tree, putting them inside is quite circular and does not obviously prevent bad behaviour.

achamayou (approved):

src/kv/deserialise.h
```diff
@@ -118,7 +118,9 @@ namespace kv
...
+          data.data() + kv::serialised_entry_header_size,
+          data.size() - kv::serialised_entry_header_size);
```
achamayou, line 123: @jumaffre This is causing the history_bench to crash because it's inserting entries that are smaller than serialised_entry_header_size. The easiest fix is probably just to make them larger.

src/kv/deserialise.h
jumaffre: @eddyashton: Agreed that we could do this in a cleaner way if we decide to not include the headers in the Merkle tree. My concern is that you still have to remember to do this every time you receive an entry from the KV/ledger and want to append it to the tree (e.g. for historical queries), as opposed to having everything that comes out of the KV or read from the ledger as the input to the Merkle tree. Because of this, I'm still leaning towards including the header in the Merkle tree.

src/kv/deserialise.h
jumaffre: Good spot! I didn't realise that these benches are no longer built/run by default (guarded by our cmake's `LONG_TESTS` flag).

src/kv/deserialise.h
achamayou: As discussed with the broader team, let's include the prefix in what we hash. There isn't a material upside at the moment other than the code is a little cleaner, but we don't think there is a downside either.

jumaffre: Done.

PR #2489, opened 2021-04-20
Add regex table filter to read_ledger.py

wintersteiger:
Also added an argparse parser as I expect more to be added soon!

### DIFF
--- a/python/ccf/read_ledger.py
+++ b/python/ccf/read_ledger.py
@@ -52,7 +52,7 @@ if __name__ == "__main__":
     )
     args = parser.parse_args()
 
-    ledger_dirs = args.directories
+    ledger_dirs = args.directory
     ledger = ccf.ledger.Ledger(ledger_dirs)
     table_filter = re.compile(args.tables)
 
--- a/python/ccf/read_ledger.py
+++ b/python/ccf/read_ledger.py
@@ -45,7 +45,7 @@ if __name__ == "__main__":
         format="<level>{message}</level>",
     )
 
-    parser = argparse.ArgumentParser(description="Read a CCF ledger directory")
+    parser = argparse.ArgumentParser(description="Read CCF ledger directories")
     parser.add_argument("directory", help="a ledger directory", nargs="+")
     parser.add_argument(
         "-t", "--tables", help="regex filter for tables", type=str, default=".*"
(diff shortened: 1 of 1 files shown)


python/ccf/read_ledger.py
```diff
@@ -43,12 +45,16 @@ def counted_string(l, name):
         format="<level>{message}</level>",
     )
 
-    if len(sys.argv) < 2:
-        LOG.error("At least one CCF ledger directory must be passed as an argument")
-        sys.exit(1)
+    parser = argparse.ArgumentParser(description="Read a CCF ledger directory")
+    parser.add_argument("directory", help="a ledger directory", nargs='+')
+    parser.add_argument(
```
achamayou, line 50: I promise this the last time I bring it up, but I think this is exactly why outputting raw JSON would be preferable: we can pipe into jq (https://stedolan.github.io/jq/) which is extremely powerful and quite easy to use rather than rebuild a very small portion of the functionality bit by bit.

It's a tiny bit more pain at first, but I suspect it will soon pay off!

And now I'll shut up about this forever :)

wintersteiger: Sure, that sounds like a good idea. Are there any known problems/inconveniences with that?

python/ccf/read_ledger.py
```diff
@@ -43,12 +45,16 @@ def counted_string(l, name):
         format="<level>{message}</level>",
     )
 
-    if len(sys.argv) < 2:
-        LOG.error("At least one CCF ledger directory must be passed as an argument")
-        sys.exit(1)
+    parser = argparse.ArgumentParser(description="Read a CCF ledger directory")
```
jumaffre, line 48: This should say "directories".

python/ccf/read_ledger.py
eddyashton: I think there's some minor inconveniences with JSON stringification*, but nothing major. I think we should add JSON as an option, but I'd like to keep the text output as default. The colour coding and summaries (`N writes` etc) make it far more human skimmable.

\* We don't do this currently but we _could_ decorate all of the text representations to make it clear what type they were originally. ie - if we added `"` around strings we managed to decode, we could distinguish `<u64: 0>` (8 zero bytes) from `"<u64: 0>"` (8 ASCII bytes that happen to look confusingly similar). If we're dumping to JSON we'd probably _not_ be re-encoding the things that are already raw JSONable, so I think we'd lose this distinction.

eddyashton (approved):

PR #2493, opened 2021-04-21
Add `MapHandle::size()`

eddyashton:
The implementation of this is currently expensive, calling `foreach` (~~and thus potentially introducing a read dependency!~~*), but will be improved in future. This is just to make the API available.

\* After staring at this a bit I realised this _needs_ a read dependency, for the same reason as `clear()`. If all we do is return a size, we need to return the version of the size we read. And if TxA writes `foo.size()` to table `bar`, and `TxB` writes to table `foo`, then TxA must happen _entirely before or after_ B; if it reads before but writes after it is wrong.


src/kv/map_handle.h
```diff
@@ -147,6 +147,19 @@ namespace kv
       };
       read_handle.foreach(g);
     }
+
+    /** Returns number of entries in this map.
+     *
+     * This is the count of all currently present keys, including both those
+     * which were already committed and any modifications (new additions or
```
achamayou, line 154:
```suggestion
     * which were already committed and any modifications (taking into account new additions or
```

achamayou (approved):

jumaffre (approved):

PR #2494, opened 2021-04-21
Add `MapHandle::clear()`

eddyashton:
Same motivation as #2493 - create this API so it can be used now, and maybe we'll improve the implementation in future.

Note the second test `SUBCASE` - `clear()` actually needs to introduce a read dependency in the current implementation, since if its ordered after another operation it may miss new keys. If we find a different way to express the clear in the underlying implementation then we may not need to introduce this read dependency (ie - an app endpoint that just says `clear()` and does no reads _can_ be ordered before or after any other), but we'd need to ensure that we remove all keys which are present at _commit_ time.


achamayou:
So long as the write sets are {key, value} maps themselves, I don't think it's possible to get rid of the read dependency.

achamayou (approved):

PR #2496, opened 2021-04-22
Turn off SAN tracer error hiding

achamayou:
Another look at #1701, after various upgrades but before clang 10.


achamayou:
20210422.4 failure caused by the partial ledger reading issue, not LeakSan

achamayou:
In the vegeta stress test:

```
Tracer caught signal 11: addr=0x0 pc=0x4f5070 sp=0x7efd9a5dad30
==8160==LeakSanitizer has encountered a fatal error.
==8160==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==8160==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
```

PR #2497, opened 2021-04-22
Improved documentation for join protocol

jumaffre:
Our documentation for the join protocol was not comprehensive enough for operators to know when a node has actually joined the service. This PR resolves this. 

image


achamayou (approved):

doc/operations/start_network.rst
```diff
@@ -138,54 +160,28 @@ Using a Configuration File
...
+Two options are provided to that end:
 
+- ``--sig-tx-interval``: number of transactions between two signatures
+- ``--sig-ms-interval``: time in milliseconds between two signatures
 
-        loop Business transactions
-            Users->>+Node 0: HTTP Request (any node)
-            Node 0-->>Users: HTTP Response (any node)
-        end
+.. note:: These options specify the intervals at which the generation of signature transactions is `triggered`. However, because of the parallel execution of transactions, it is possible that signature transactions are recorded in the ledger at a slightly higher interval than the specified values.
```
eddyashton, line 184:
```suggestion
.. note:: These options specify the intervals at which the generation of signature transactions is `triggered`. However, because of the parallel execution of transactions, the actual intervals between signature transactions may be slightly larger.
```

I think "at a slightly higher interval" is odd wording, prefer _larger_ interval. Reworded to clarify this.

doc/operations/start_network.rst
```diff
@@ -90,17 +65,64 @@ To add a new node to an existing opening network, other nodes should be started
...
+Otherwise, if the network has already been opened to users, members need to trust the joining node before it can become part of the network and participate in the consensus (see :ref:`governance/common_member_operations:Trusting a New Node`).
+
+The ``Pending`` joining node will automatically poll the service (interval configurable via ``--join-timer`` option) until the members have successfully transitioned the node to the ``Trusted`` state. It is only then that the joining node will transition to the ``PartOfNetwork`` state and start updating its ledger.
 
-If the network has already been opened to users, members need to trust the joining node before it can become part of the network (see :ref:`governance/common_member_operations:Trusting a New Node`).
+.. tip:: After the node has been trusted by members, operators should poll the ``/node/state`` endpoint on the newly added node until the ``{"state": "PartOfNetwork"}`` is reported. This status confirms that the replication of the ledger has started on this node.
```
eddyashton, line 80: Should we talk about the difference between `PartOfNetwork` and "fully caught up" - ie, you may want to poll the node's commit index and only direct users to it once it has _finished_ replication?

achamayou: What does "finished" mean?

jumaffre: Yeah, it's awkward to define "finished" here. If we decide to go the "[learner](https://etcd.io/docs/v3.3/learning/learner/#features-in-v35)" way at some point, we could have some thresholds to define when a node actually becomes part of the service (e.g. `N` transactions after it's seen its own addition to the store).

achamayou: I would prefer it not be mentioned until we have a proper definition for it.

eddyashton: I think finished is up to the operator, but if we're clarifying the meaning of these states we should also clarify that a `PartOfNetwork` node may still be serving reads at `1.10`, despite your service being at `10.999999`?

achamayou: Sure, but that's always true, nodes can lag behind etc.

eddyashton: Sure, so if you were monitoring all of your nodes and noticed it was lagging, you may want to remove it from your load balancer while you investigated. This lag is going to be most pronounced, and always present even in a well-behaved network, on a new joiner.

achamayou: If they join from a recent snapshot, it's going to be quite slight and rapidly plugged. Adding a reminder is fine, and the operator may have a QoS policy, but clients still need to be robust to followers lagging.

PR #2498, opened 2021-04-22
Add `get_id_for_this_node_v1`

eddyashton:
Adds a versioned API to fetch the executing node's ID.

Question: Should we insert the node ID into a response header for the default endpoints?


achamayou (approved):

achamayou:
I don't think we should include this header as default, applications that wish to do so can easily. This sounds like unnecessary bloat in most cases, a normal HTTPS server would not typically include its identity fingerprint in response headers.

Issue #1276, opened 2020-06-11
Commit a ledger chunk early


jumaffre:
Follow-up from #1135 

Now that the ledger is split into different chunks that get committed (i.e. we are sure that they will never be written to/truncated) when 1) they have grown enough that they contain a signature that has passed the chunking threshold and 2) the last index they contain has been globally committed by the underlying consensus. 

However, if a chunk has not yet passed the threshold, it may be nice to let operators force a new chunk to be created so that on the next global commit, that chunk is committed. For example, this would be useful if the system is not under heavy load for some time and operators want to backup all existing ledger chunks so far.

achamayou:
@jumaffre this seems like a subset of #3155?

jumaffre:
@achamayou I hadn't see your comment until now. Yes, I'll close this issue as it's already tracked by #3155.

PR #2499, opened 2021-04-22
Do not try to parse uncommitted chunks by default

achamayou:
Sparked by https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=23939&view=logs&j=ee9484ec-b476-577b-aa0e-d5ddfcbcd149&t=e2c1a4cf-16d0-5038-f9f6-343f9806d38b

```
36: Traceback (most recent call last):
36:   File "/mnt/build/2/s/tests/governance_history.py", line 182, in <module>
36:     run(args)
36:   File "/mnt/build/2/s/tests/governance_history.py", line 155, in run
36:     test_ledger_is_readable(network, args)
36:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 21, in wrapper
36:     return func(*args, **kwargs)
36:   File "/mnt/build/2/s/tests/governance_history.py", line 96, in test_ledger_is_readable
36:     for chunk in ledger:
36:   File "/mnt/build/2/s/python/ccf/ledger.py", line 570, in __next__
36:     self._current_chunk = LedgerChunk(
36:   File "/mnt/build/2/s/python/ccf/ledger.py", line 509, in __init__
36:     self._current_tx = Transaction(name, ledger_validator)
36:   File "/mnt/build/2/s/python/ccf/ledger.py", line 395, in __init__
36:     self._file = open(filename, mode="rb")
36: FileNotFoundError: [Errno 2] No such file or directory: '/mnt/build/2/s/build/workspace/governance_history_test_1/1.ledger/ledger_19'
36: Exception ignored in: <function Transaction.__del__ at 0x7f355f84bf70>
36: Traceback (most recent call last):
36:   File "/mnt/build/2/s/python/ccf/ledger.py", line 416, in __del__
36:     self._file.close()
36: AttributeError: 'NoneType' object has no attribute 'close'
30/62 Test #36: governance_history_test ...........***Failed    8.32 sec
```

If we attempt to read an uncommitted chunk where the ledger is read, it might disappear between listing and reading. By restricting the read to committed chunks only (which are immutable), we can avoid this issue.


achamayou:
The change in its current form breaks a number of tests because it restricts our ability to read fresh, but not yet immutable state.

After discussion with @eddyashton, we've come to the conclusion that:

1. trying to parse not yet immutable state reliable is very difficult, copying the files aside doesn't really help when there are several uncommitted chunks for example.
2. the easiest way to make sure the state we want to see in those tests becomes available in ".committed" chunks is to append more than a chunk's worth of data to the ledger.

This solution is less elegant and general than #1276, but considering the scope of the requirement (our tests only), it's adequate.

tests/infra/consortium.py
```diff
@@ -347,6 +347,14 @@ def add_user(self, remote_node, user_id, user_data=None):
         proposal = self.get_any_active_member().propose(remote_node, proposal)
         return self.vote_using_majority(remote_node, proposal, careful_vote)
 
+    def create_and_withdraw_large_proposal(self, remote_node):
```
jumaffre, line 350: I'd suggest adding a short comment here to explain why someone would ever want to use this.

jumaffre (approved):

PR #2500, opened 2021-04-22
Two-step reconfiguration

wintersteiger:
This adds two-step reconfigurations, which will allow us to easily plug in Byzantine signing.

I expect there are some things that should be moved between files/folders, just let me know if something looks like it's in the wrong place.

The whole process is triggered by `nodeInfo.status = "CatchingUp";` in `actions.js`, which I will revert back to `Trusted` once all the tests pass, so that the current behaviour doesn't change too much. Is there an easy way to make this conditional on the fact that we're running with BFT consensus?


src/consensus/aft/raft.h
```diff
@@ -292,7 +316,17 @@ namespace aft
...
+      if (!catching_up && !retiring)
+      {
+        active_nodes_.insert(id());
+      }
+#ifndef NDEBUG
+      std::stringstream ss;
+      for (auto& m : active_nodes_)
+        ss << " " << m;
+      LOG_DEBUG_FMT("Active nodes:{}", ss.str());
+#endif
```
eddyashton, line 329: You can build this concatenation with `fmt::join`. Something like

```suggestion
      LOG_DEBUG_FMT("Active nodes: {}", fmt::join(active_nodes_, ", "));
```

src/consensus/aft/raft.h
```diff
@@ -284,6 +282,32 @@ namespace aft
...
+      size_t r = (catching_up || retiring) ? 0 : 1;
+      for (const auto& [_, state] : nodes)
+      {
+        if (!state.catching_up)
+        {
+          r++;
+        }
+      }
+      return r;
+    }
```
eddyashton, line 309: Is there a reason this duplicates some of `is_eligible_voter`? While less efficient, I think this would be much more readable as

```suggestion
    size_t num_eligigble_voters() const
    {
      size_t r = 0;
      for (const auto& id, _] : nodes)
      {
        if (is_eligible_voter(id))
        {
          r++;
        }
      }
      return r;
    }
```

This could also be replaced by `return std::count_if(...)`.

src/consensus/aft/raft.h
```diff
@@ -284,6 +282,32 @@ namespace aft
...
+    bool is_eligible_voter(const ccf::NodeId& from) const
+    {
+      if (from == state->my_node_id)
+      {
+        return !catching_up && !retiring;
+      }
+      else
+      {
+        auto it = nodes.find(from);
+        return it != nodes.end() && !it->second.catching_up;
```
eddyashton, line 294: A few questions, to help me understand the implementation:

Does the local state need to be different than the state for all the other nodes?
Do we not track whether other nodes are retiring?
Do we have an entry for `state->my_node_id` in `nodes`, and if so does `nodes[me].catching_up` always match `catching_up`?

src/consensus/aft/raft.h
```diff
@@ -430,6 +469,11 @@ namespace aft
 
     void add_configuration(Index idx, const Configuration::Nodes& conf)
     {
+      for (auto& [id, info] : conf)
+      {
+        LOG_TRACE_FMT(
+          "CADD {} {}:{} {}", id, info.hostname, info.port, info.catching_up);
+      }
```
eddyashton, line 476: I think the logs are more readable if we're more verbose:

```suggestion
      LOG_TRACE_FMT("Adding a configuration at {} of {} nodes", idx,  conf.size());
      for (auto& [id, info] : conf)
      {
        LOG_TRACE_FMT(
          "  {} = {}:{} ({})", id, info.hostname, info.port, info.catching_up ? "passive" : "active");
      }
```

src/consensus/aft/raft.h
```diff
@@ -2393,11 +2529,12 @@ namespace aft
 
     void become_candidate()
     {
-      if (replica_state == kv::ReplicaState::Retired)
+      if (replica_state == kv::ReplicaState::Retired || retiring)
```
eddyashton, line 2532: I'm wary of having a few bools in addition to the replica state machine. It might be the right approach, but we should confirm that.

src/consensus/aft/raft.h
```diff
@@ -2515,19 +2676,61 @@ namespace aft
...
+      catching_up = false;
 
-      LOG_INFO_FMT(
-        "Becoming retired {}: {}", state->my_node_id, state->current_view);
+#if 0
+      // To guarantee liveness during reconfiguration, we need something of this
+      // sort (e.g. when replacing the whole network in one reconfiguration).
+      // This is not enough though; we would need to check that n/2+1 of the new
+      // nodes have seen enough of the new configuration to make sure we don't
+      // lose any transactions.
```
eddyashton, line 2686: I agree with the comment, but I don't think a local decision to delay retirement will work. My understanding is; when a node knows locally of its own removal, it should stop producing writes but otherwise continues as a normal follower/primary (it is valid and correct for a node to become leader after its removal, and share this suffix describing its own removal with the other nodes). Once this is globally committed, the node can transition to `Retired` and cease to participate (with a slight complication that if it is primary, it _should_ stick a round a little longer to try to catch up _all_ nodes, and do what it can to ensure the next primary is live in the latest configuration).

I'm not sure how this changes with catch-up. I think if a node hears global commit of its removal, it can still retire instantly - this means there are a quorum of nodes in the new configuration who can take over, so they don't need this specific node to stay active any longer.

src/ds/thread_messaging.h
```diff
@@ -67,7 +69,14 @@ namespace threading
       ThreadMsg* current = local_msg;
       local_msg = local_msg->next;
 
-      current->cb(std::unique_ptr<ThreadMsg>(current));
+      if (current->cb)
+        current->cb(std::unique_ptr<ThreadMsg>(current));
+      else
+      {
+        current->f();
```
eddyashton, line 76: Can we give `f` a better name? It looks like its a fire-once-then-kill-me callback or something?

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+  public:
+    const NodeId& node_id;
+    ConsensusType consensus_type;
+    std::shared_ptr<kv::AbstractStore> store;
+    std::shared_ptr<enclave::RPCSessions> rpcsessions;
+    std::shared_ptr<enclave::RPCMap> rpc_map;
+    crypto::KeyPairPtr node_sign_kp;
+    const crypto::Pem& node_cert;
+    std::unordered_map<ccf::NodeId, aft::NodeState>& nodes;
+    std::list<kv::Configuration> configurations;
```
eddyashton, line 65: I don't think `node_id` and `node_cert` should be references, they should just be copied locally (they may be referring to a heap-allocated object's members right now, but they could easily come from the stack and this would break). `nodes` I don't know, that may really be referring to an map owned by someone else? But that should then be some kind of smart pointer, not a reference.

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+      rpcsessions(rpcsessions_),
+      rpc_map(rpc_map_),
+      node_sign_kp(node_sign_kp_),
+      node_cert(node_cert_),
+      nodes(nodes_)
+    {
+      if (!node_sign_kp_)
+        LOG_INFO_FMT("NO SIGNING KEY!");
+      if (!rpc_map)
+        LOG_INFO_FMT("NO RPC MAP!");
```
eddyashton, line 89: Should these be errors?

CMakeLists.txt
```diff
@@ -679,6 +682,13 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS --ccf-version ${CCF_VERSION}
   )
 
+  add_e2e_test(
```
achamayou, line 685: Better to move that to ~803 in the loop.

python/ccf/tx_id.py
```diff
@@ -29,3 +29,8 @@ def parse(s: Optional[str]):
         except (AttributeError, ValueError):
             pass
         return None, None
+
+    def __lt__(self, other):
```
achamayou, line 33: This isn't obviously correct?

tests/rotation.py
```diff
@@ -16,6 +16,7 @@
 @reqs.can_kill_n_nodes(1)
 def test_suspend_primary(network, args):
     primary, backup = network.find_primary_and_any_backup()
+    network.wait_for_trusted_and_committed(primary, primary)
```
achamayou, line 19: Can a node be primary without being trusted and committed?

src/ds/thread_messaging.h
```diff
@@ -14,10 +14,12 @@ namespace threading
 {
   struct ThreadMsg
   {
-    void (*cb)(std::unique_ptr<ThreadMsg>);
+    void (*cb)(std::unique_ptr<ThreadMsg>) = nullptr;
+    std::function<void()> f = nullptr;
```
ashamis, line 18: why is this specialization needed? we should try to minimize the number of ways an async function should be called

src/ds/thread_messaging.h
```diff
@@ -276,6 +293,20 @@ namespace threading
...
+      task.add_task(new ThreadMsg(std::move(f)));
+    }
+
+    void add_task(std::function<void()>&& f)
+    {
+      add_task(
+        threading::ThreadMessaging::get_execution_thread(
+          threading::MAIN_THREAD_ID),
+        std::move(f));
+    }
```
ashamis, line 308: consider something like add_task_on_main as to me an add task without a tid implies you call it on the current thread

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+#include "node/nodes.h"
+#include "node/rpc/node_call_types.h"
+#include "node/rpc/serdes.h"
+#include "rpc/serialization.h"
+
+using namespace ccf;
+
+namespace aft
+{
+  using Index = uint64_t;
```
ashamis, line 24: This should be called seqno and can't we reuse SeqNo?

src/node/nodes.h
```diff
@@ -17,12 +17,14 @@ namespace ccf
   enum class NodeStatus
   {
     PENDING = 0,
-    TRUSTED = 1,
-    RETIRED = 2
+    CATCHING_UP = 1,
+    TRUSTED = 2,
+    RETIRED = 3
```
ashamis, line 22: If NodeStatus is serialized anywhere and sent as a network msg or put in the ledger reordering the numbers like this will break back compat

src/consensus/aft/raft.h
```diff
@@ -1891,6 +1978,19 @@ namespace aft
...
+        response.success == AppendEntriesResponseType::OK)
+      {
+        NodeCaughtUpMsg ncamsg = {
+          {raft_node_caught_up}, state->current_view, state->last_idx};
+        for (const auto& [node_id, _] : nodes)
+        {
+          channels->send_authenticated(
+            node_id, ccf::NodeMsgType::consensus_msg, ncamsg);
+        }
+      }
```
ashamis, line 1993: should we be only sending this if the node is catching up?

src/consensus/aft/raft.h
```diff
@@ -2017,6 +2120,13 @@ namespace aft
         return;
       }
 
+      if (!is_eligible_voter(from))
+      {
+        LOG_INFO_FMT(
```
ashamis, line 2125: FAIL message type?

src/consensus/aft/raft.h
```diff
@@ -2198,6 +2307,19 @@ namespace aft
         state->my_node_id.trim(),
         from.trim(),
         r.last_log_idx);
+
+      if (is_primary() && node->second.catching_up)
```
ashamis, line 2311: line 2213 already checks if the node is the leader

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+        }
+      }
+    }
+
+    static void promote_cb(
+      ConfigurationTracker& configuration_tracker,
+      NodeId from,
+      TxID txid,
+      TxID primary_txid)
+    {
```
ashamis, line 402: this is already occurring in an async callback why is another needed?

src/consensus/aft/raft.h
```diff
@@ -2198,6 +2307,19 @@ namespace aft
...
+      {
+        threading::ThreadMessaging::thread_messaging.add_task(
+          [this, from, r]() {
```
ashamis, line 2314: why do you need to capture this? also this seems unsafe

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+          else if (nit != nodes.end() && nit->second.catching_up)
+            ss << "*";
+        }
+        ss << "]@" << c.idx << " ";
+      }
+      return ss.str();
+    }
+
+    bool record_promotion(const NodeId& node_id)
+    {
```
ashamis, line 319: Sorry i might be misunderstanding what is going on here, why does this need to go through the frontend?

src/node/genesis_gen.h
```diff
@@ -269,8 +269,13 @@ namespace ccf
 
     void add_node(const NodeId& id, const NodeInfo& node_info)
     {
-      // Increment the node id (only used in BFT)
-      get_next_id(tx.rw(tables.values), ValueIds::NEXT_NODE_ID);
+      auto config = tx.ro(tables.config);
+      auto c0 = config->get(0);
+      if (c0 && c0->consensus == BFT)
```
ashamis, line 274: Should we not throw if there is no config?

src/consensus/aft/test/driver.cpp
```diff
@@ -98,5 +98,6 @@ int main(int argc, char** argv)
     ++lineno;
   }
 
+  threading::ThreadMessaging::thread_messaging.drop_tasks();
```
ashamis, line 101: Why is this required?

src/consensus/aft/async_execution.h
```diff
@@ -301,4 +303,30 @@ namespace aft
     size_t size;
     OArray oarray;
   };
+
+  class NodeCatchUpCallback : public AbstractMsgCallback
+  {
+  public:
+    NodeCatchUpCallback(
```
ashamis, line 310: This is not following the pattern of the rest of the code. You can add a virtual method to store update_node_progress

src/consensus/aft/raft.h
```diff
@@ -141,8 +131,10 @@ namespace aft
     size_t sig_tx_interval;
 
     // Configurations
-    std::list<Configuration> configurations;
-    std::unordered_map<ccf::NodeId, NodeState> nodes;
+    aft::ConfigurationTracker configuration_tracker;
+    std::unordered_map<ccf::NodeId, aft::NodeState> nodes;
+    bool catching_up = false;
```
ashamis, line 136: Should this be an enum? I assume you cannot be both catching up and retiring?

ashamis (approved):

include/ccf/json_handler.h
```diff
@@ -261,6 +261,9 @@ namespace ccf
   inline jsonhandler::JsonAdapterResponse make_error(
     http_status status, const std::string& code, const std::string& msg)
   {
+    // Debug-only log for BFT, which may swallow frontend error results.
```
ashamis, line 264: This comment does not match the implementation, maybe make keep this both CFT and BFT but i am not sure everyone would agree on this

src/consensus/aft/raft.h
```diff
@@ -380,6 +414,11 @@ namespace aft
       return state->commit_idx;
     }
 
+    Index get_commit_idx_unsafe()
```
ashamis, line 417: Can you please change the unsafe version to just aquire a lock and then call the unsafe function

src/consensus/aft/raft.h
```diff
@@ -1853,6 +1939,7 @@ namespace aft
 
       // After entries have been deserialised, try to commit the leader's
       // commit index and update our term history accordingly
+      LOG_DEBUG_FMT("Leader's commit index: {}", r.leader_commit_idx);
```
ashamis, line 1942: Do we need to keep this log line?

src/consensus/aft/raft.h
```diff
@@ -2409,23 +2546,36 @@ namespace aft
 
       LOG_INFO_FMT(
         "Becoming candidate {}: {}", state->my_node_id, state->current_view);
+      LOG_DEBUG_FMT("Configurations: {}", configuration_tracker.to_string());
```
ashamis, line 2549: This might be a bit too spammy, if there is not configuration changes in progress do we really need to print this log line?

ashamis (commented):
Something odd is going on with the github vscode pluggin - testing to see if this fixes it

ashamis (changes requested):
sorry i did not mean to signoff - still going through the review

src/ds/nonstd.h
```diff
@@ -7,6 +7,7 @@
 #include <cctype>
 #include <string>
 #include <string_view>
+#include <tuple>
```
ashamis, line 10: I might be missing something but is this required?

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+      }
+
+      LOG_TRACE_FMT("Configurations: {}", to_string());
+
+      return r;
+    }
+
+    bool rollback(Index idx)
+    {
+      LOG_TRACE_FMT("Configurations: rollback to {}", idx);
```
ashamis, line 125: Should there not be at least 1 configuration left after the rollback? (The globally committed configuration)

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+        configurations.pop_front();
+        r = true;
+      }
+
+      LOG_TRACE_FMT("Configurations: {}", to_string());
+
+      return r;
+    }
+
+    bool rollback(Index idx)
```
ashamis, line 123: This is nit: but i find this API confusing, returning false to me implies that the rollback failed not that there was nothing to rollback

doc/schemas/node_openapi.json
```diff
@@ -277,6 +277,7 @@
       "NodeStatus": {
         "enum": [
           "Pending",
+          "CatchingUp",
```
achamayou, line 280: I'm a bit worried about this naming, Trusted formerly implied two things:
  1. that the node was participating in replication and election quorums
  2. that it had received a copy of the ledger secrets (ie. trusted from a confidentiality perspective).

`CatchingUp` effectively corresponds to the latter but not the former, but the name doesn't convey it strongly, and the existence of a separate `Trusted` stage even suggests otherwise.

I propose:

`CatchingUp` -> `Replicating`
`Trusted` -> `Participating`

or:

`CatchingUp` -> `Learner`
`Trusted` -> `Replica`

include/ccf/json_handler.h
achamayou: In CFT, this results in an error returned to the client, which is why we don't typically log it. In BFT of course, most execution doesn't yield anything to the client, and so this become useful for debugging.

What happens at least once per request would typically be TRACE rather than DEBUG, but I have no objection to adding this otherwise in all cases. We know we have too much logging and we need to look into named loggers.

include/ccf/tx_id.h
```diff
@@ -82,6 +82,11 @@ namespace ccf
 
       return tx_id;
     }
+
+    bool operator<(const TxID& other) const
+    {
+      return view < other.view || (view == other.view && seqno < other.seqno);
```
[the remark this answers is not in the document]
achamayou: I am also worried about this, I don't think there is a natural order for transaction ids. The answer is probably "it depends exactly what you do with it", in which case operator overloading doesn't seem like a natural choice.

src/consensus/aft/raft.h
achamayou: +1 for rolling this into the state enum unless there's an explicit reason why not.

src/consensus/aft/raft.h
```diff
@@ -141,8 +131,10 @@ namespace aft
     size_t sig_tx_interval;
 
     // Configurations
-    std::list<Configuration> configurations;
-    std::unordered_map<ccf::NodeId, NodeState> nodes;
+    aft::ConfigurationTracker configuration_tracker;
+    std::unordered_map<ccf::NodeId, aft::NodeState> nodes;
+    bool catching_up = false;
+    bool retiring = false;
```
achamayou, line 137: I'll put up a separate PR for this, but I think a flag isn't enough and we need to keep track of more than that to ensure a single signature retirement.

src/consensus/aft/raft.h
```diff
@@ -483,6 +525,12 @@ namespace aft
       return get_latest_configuration_unsafe().size();
     }
 
+    ccf::SeqNo get_confirmed_matching_index(const NodeId& id) const
```
achamayou, line 528: We've almost entirely moved away from magic values and towards optionals, I think we should try to stick to that.

include/ccf/tx_id.h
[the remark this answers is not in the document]
wintersteiger: Yeah, this is incorrect. I think it works only on the primary, but an overloaded operator is a bad choice regardless.

src/ds/thread_messaging.h
wintersteiger: It isn't. I just wanted to use lambdas for the functions, but I think I ended up not using this at all.

src/consensus/aft/raft.h
wintersteiger: Good point, I think you're right, we only need this during catchup.

src/node/configuration_tracker.h
wintersteiger: Another what? Am I checking some condition twice or something like that?

src/node/configuration_tracker.h
wintersteiger: Recommendation by @achamayou. This makes sure the transaction is properly recorded on the ledger.

include/ccf/json_handler.h
wintersteiger: Yeah "for BFT" means "I'm adding this because of BFT" :-)

src/consensus/aft/raft.h
wintersteiger: Not sure what you mean. The safe version gets the lock, the unsafe version is not supposed to get the lock?

src/consensus/aft/raft.h
wintersteiger: No, and there are lots of other logs that should be removed too.

src/consensus/aft/raft.h
wintersteiger: I need it, but nobody else does :-)

src/node/configuration_tracker.h
wintersteiger: And there isn't? I don't think the rollback logic changed.

src/consensus/aft/raft.h
wintersteiger: Yes, retirement changes aren't implemented here, because I thought we wouldn't need it. We'll need a lot more to do that properly.

doc/schemas/node_openapi.json
wintersteiger: I agree, we need to change that. To me, "Participating" sounds a bit too weak, but maybe it would be a good thing to remove the connotation with confidentiality. Maybe Replicating/Replica? Or Cloning/Replica?

src/node/node_to_node.h
```diff
@@ -209,9 +209,16 @@ namespace ccf
       size_t size) override
     {
       auto n2n_channel = channels->get(to);
-      // Sending after a channel has been destroyed is a bug.
-      assert(n2n_channel);
-      return n2n_channel->send(type, {data, size});
+      if (!n2n_channel)
+      {
+        LOG_DEBUG_FMT(
```
jumaffre, line 214: Curious: why has this changed in this PR?

wintersteiger: That's because BFT assumes that it can reach all the nodes at all times and this was the quickest way to make it work. I'll take another look at this after the ongoing changes and if possible I'd like to make it a failure again.

src/node/rpc/node_frontend.h
```diff
@@ -409,6 +416,93 @@ namespace ccf
...
+
+        if (!consensus)
+        {
+          return make_error(
+            HTTP_STATUS_INTERNAL_SERVER_ERROR,
+            ccf::errors::InternalError,
+            "No configured consensus");
+        }
+
+        if (
```
jumaffre, line 441: This should already be handled by the HTTP endpoint framework as the endpoint is a write endpoint.

src/consensus/aft/raft.h
wintersteiger: Not sure, yes, and no. I had tried not to disrupt the existing code too much (e.g. `me` not in `nodes`), but after the latest discussions I'm now changing this.

src/ds/thread_messaging.h
wintersteiger: `f` is the `f`unction. I just wanted to use lambdas for thread functions and add a way to make them automatically retry upon failure. I might remove all of that though.

src/node/rpc/node_frontend.h
```diff
@@ -409,6 +416,93 @@ namespace ccf
...
+
+        auto nodes = args.tx.rw(this->network.nodes);
+
+        auto node_info = nodes->get(in.node_id);
+        if (!node_info.has_value())
+        {
+          return make_error(
+            HTTP_STATUS_INTERNAL_SERVER_ERROR,
+            ccf::errors::InternalError,
+            "Missing node state");
```
jumaffre, line 493:
```suggestion
            fmt::format("Unknown node {}", in.node_id));
```

src/node/configuration_tracker.h
wintersteiger: The whole class is just a mental separation of the configuration-relevant parts of `Aft`. It's not helpful though, so I'll move all of this back into `Aft`.

src/node/configuration_tracker.h
wintersteiger: Yes, but I'll remove this anyway.

tests/infra/consortium.py
```diff
@@ -516,12 +518,33 @@ def check_for_service(self, remote_node, status):
                 current_status == status.value
             ), f"Service status {current_status} (expected {status.value})"
 
-    def _check_node_exists(self, remote_node, node_id, node_status=None):
-        with remote_node.client() as c:
-            r = c.get(f"/node/network/nodes/{node_id}")
+    def _check_node_exists(self, remote_node, node_id, node_status=None, retries=5):
```
jumaffre, line 521: The Python infra usually deals with `timeout` (in secs) rather than `retries`.

src/consensus/aft/raft.h
```diff
@@ -2409,23 +2542,36 @@ namespace aft
...
+      LOG_DEBUG_FMT("Configurations: {}", configuration_tracker.to_string());
 
       if (consensus_type != ConsensusType::BFT)
       {
         for (auto it = nodes.begin(); it != nodes.end(); ++it)
         {
+          if (it->first == state->my_node_id)
+            continue;
+
           channels->create_channel(
```
jumaffre, line 2554: Do we need to create channels to nodes that aren't eligible?

wintersteiger: Yes we do, during catchup we need to send them append entries and in BFT they also gossip their state to the other nodes.

src/node/rpc/node_frontend.h
```diff
@@ -409,6 +416,93 @@ namespace ccf
...
+            fmt::format("Unknown node {}", in.node_id));
+        }
+
+        node_info->status = NodeStatus::TRUSTED;
+        nodes->put(in.node_id, *node_info);
+        return make_success(true);
+      };
+
+      make_endpoint(
+        "promote", HTTP_POST, json_adapter(promote), no_auth_required)
```
jumaffre, line 502:
```suggestion
        "promote", HTTP_POST, json_adapter(promote), {std::make_shared<NodeCertAuthnPolicy>()})
```

Otherwise anyone could promote a node.

wintersteiger: Great many thanks, I just didn't know what to use here!

src/consensus/aft/raft.h
```diff
@@ -646,6 +697,20 @@ namespace aft
             break;
           }
 
+          case raft_node_caught_up:
+          {
+            NodeCaughtUpMsg r =
+              channels->template recv_authenticated_with_load<NodeCaughtUpMsg>(
```
jumaffre, line 703:
```suggestion
              channels->template recv_authenticated<NodeCaughtUpMsg>(
```
I think?

src/consensus/aft/raft.h
```diff
@@ -1891,6 +1979,19 @@ namespace aft
 
       channels->send_authenticated(
         to, ccf::NodeMsgType::consensus_msg, response);
+
+      if (
+        consensus_type == ConsensusType::BFT &&
+        response.success == AppendEntriesResponseType::OK)
+      {
+        NodeCaughtUpMsg ncamsg = {
+          {raft_node_caught_up}, state->current_view, state->last_idx};
```
jumaffre, line 1988: AIUI, this new message only contains the current view and last idx, which are already sent as part of the `AppendEntriesResponse`. Couldn't we use the `AppendEntriesResponse` only for this purpose instead of creating a new message type?

src/consensus/aft/raft.h
```diff
@@ -646,6 +697,20 @@ namespace aft
...
+          {
+            NodeCaughtUpMsg r =
+              channels->template recv_authenticated_with_load<NodeCaughtUpMsg>(
+                from, data, size);
+
+            aee = std::make_unique<NodeCatchUpCallback>(
+              from,
+              configuration_tracker,
+              TxID({r.view, r.seqno}),
+              TxID({state->current_view, state->last_idx}));
```
jumaffre, line 710: This is a pair of seqno/view but I don't think this is a valid `TxID`: `last_idx` could have been committed in the previous view but we're not at `current_view`.

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+        !(txid < primary_txid) && store != nullptr)
+      {
+        return record_promotion(id);
+      }
+
+      return true;
+    }
+
+    void update_node_progress(
+      const NodeId& from, const TxID& txid, const TxID& node_txid)
```
jumaffre, line 175: `node_txid` is unused?

src/consensus/aft/raft.h
```diff
@@ -2198,6 +2308,19 @@ namespace aft
...
+
+      if (is_primary() && node->second.catching_up)
+      {
+        threading::ThreadMessaging::thread_messaging.add_task(
+          [this, from, r]() {
+            ConfigurationTracker::promote_cb(
+              configuration_tracker,
+              from,
+              {r.term, r.last_log_idx},
+              {state->current_view, state->commit_idx});
```
jumaffre, line 2320: Same comment here: `current_view` and `commit_idx` are not a valid `TxID`.

src/node/configuration_tracker.h
```diff
@@ -0,0 +1,413 @@
...
+      TxID txid,
+      TxID primary_txid)
+    {
+      if (!configuration_tracker.promote_if_possible(from, txid, primary_txid))
+      {
+        threading::ThreadMessaging::thread_messaging.add_task(
+          [&configuration_tracker, from, txid, primary_txid]() {
+            ConfigurationTracker::promote_cb(
+              configuration_tracker, from, txid, primary_txid);
+          });
```
jumaffre, line 409: I'm not sure whether we want to retry forever in the unlikely event that the promotion execution fails. This will fire the next time an Append Entries Response message will be received from the peer node so maybe that's sufficient?

src/node/node_to_node.h
ashamis: In what circumstances are we attempting to send to an unknown node?

wintersteiger:
Replaced by a series of smaller PRs.

PR #2501, opened 2021-04-22
Use `CKM_RSA_PKCS_OAEP` to encrypt ledger secrets during ledger rekey

jumaffre:
We used to encrypt ledger secrets in the `public:ccf.internal.encrypted_ledger_secrets` table with a custom scheme based on the key exchange context. Instead, we now use the standard `CKM_RSA_PKCS_OAEP` wrapping algorithm for it, which is already used to wrap member recovery shares.


achamayou:
It may be good to mention this briefly in https://microsoft.github.io/CCF/main/overview/cryptography.html as an exception to our usage of EC for all node-to-node communication.

src/node/node_state.h
```diff
@@ -1724,18 +1724,15 @@ namespace ccf
                 network.secrets.get_name()));
             }
 
-            auto encrypted_ledger_secrets = w.at(0);
-            if (!encrypted_ledger_secrets.has_value())
+            auto secrets_for_nodes = w.at(0);
```
achamayou, line 1727: Keeping "ledger" in the name would be good, imo, we have several different kind of secrets.

src/node/secret_broadcast.h
```diff
@@ -91,43 +60,26 @@ namespace ccf
...
-            gcmcipher.hdr.tag,
-            gcmcipher.cipher,
-            nullb,
-            plain.data()))
-      {
-        throw std::logic_error("Decryption of ledger secret failed");
-      }
+      auto unwrapped = crypto::ckm_rsa_pkcs_oaep_unwrap(encryption_key, cipher);
+
+      LOG_FAIL_FMT("Unwrapped: {}", tls::b64_from_raw(unwrapped));
```
achamayou, line 80: To be removed, presumably?

jumaffre: Yes! Thanks.

achamayou (approved):

PR #2505, opened 2021-04-23
Reset RPC response on conflict

### LINKED ISSUE #2490: Reset ` HttpRpcContext` response when re-executing endpoints
When an endpoint needs be re-executed (e.g. KV conflict), we should reset the ` HttpRpcContext`'s HTTP response as well as resetting the `kv::Tx` object as the failed endpoint execution may have modified the HTTP response, and the re-execution may go through a different code path.

jumaffre:
Resolves #2490 

We now reset the RPC response on conflict. For example, if the endpoint set a HTTP response header in its first execution with resulted in a KV conflict, the next (conflict-free) execution may go through a different code path and not set the same HTTP response header and the headers shouldn't be returned in the response.


src/node/rpc/frontend.h
```diff
@@ -273,6 +273,10 @@ namespace ccf
       {
         if (attempts > 0)
```
jumaffre, line 274: The retry loop should be moved higher up in this function, so that we lookup the endpoint (useful for JS apps) and authenticate the caller based on a new KV `read_version`. I believe this is safe but haven't included the change in this PR. In particular:
- If the request was forwarded to the primary, the transaction will not commit so no retry there.
- If the request was forwarded to us (primary), the transaction may conflict but that's fine too.

Am I missing anything?

eddyashton (approved):

src/http/http_rpc_context.h
```diff
@@ -303,6 +303,14 @@ namespace http
       return status_success(response_status);
     }
 
+    virtual void reset_response() override
```
achamayou, line 306: It would be nice to have unit tests that construct two contexts, run a request + reset_response on one, and perform a structural equality check to make sure we're not missing anything here if&when we add state in the future.

src/node/rpc/frontend.h
achamayou: That sounds right, but I seem to remember there were some problems with that. One thing that comes to mind was discovering from re-reading the endpoint that it shouldn't forward (anymore), while executing it in a forwarded context. Let's discuss this with @eddyashton, I agree it's best done separately.

achamayou (approved):

PR #2508, opened 2021-04-26
Fix python format

eddyashton:
Format check is failing on `main` because somehow this was checked in unformatted.


jumaffre (approved):

jumaffre:
This seems to be related to a new release of `black`: https://github.com/psf/black/releases/tag/21.4b0

PR #2509, opened 2021-04-26
Pass system clock for enclave time

eddyashton:
This should have basically no affect. It changes the value we pass as untrusted host time from "microseconds since the node started" to "microseconds since Unix epoch", so it can be converted back more easily on the host side. I also changed the offset calculation code in `logger.h` - I believe its now simpler but functionally equivalent.


src/ds/logger.h
```diff
@@ -94,17 +94,26 @@ namespace logger
...
+        if (enc_ts.tv_nsec > ns_per_s)
+        {
+          enc_ts.tv_sec += 1;
+          enc_ts.tv_nsec -= ns_per_s;
+        }
+
         std::tm enclave_tm;
-        auto enc_ts = enclave_ts.value();
-        ::timespec_get(&enc_ts, TIME_UTC);
-        ::gmtime_r(&enc_ts.tv_sec, &enclave_tm);
```
eddyashton, line 107: This was just wrong. In the JSON logging path, we ignored the passed enclave/offset time, and just re-fetched the host time again!

src/ds/logger.h
```diff
@@ -438,30 +427,24 @@ namespace logger
...
 
-      // Enclave time is recomputed every time. If multiple threads
-      // log inside the enclave, offsets may not always increase
-      ::timespec enclave_ts{logger::config::start.tv_sec + elapsed_s,
-                            logger::config::start.tv_nsec + elapsed_ns};
+      // Represent offset as a real (counting seconds) to handle both small
+      // negative _and_ positive numbers. Since the system clock used is not
+      // monotonic, the offset we calculate could go in either direction, and tm
+      // can't represent small negative values.
+      std::optional<double> offset_time = std::nullopt;
```
eddyashton, line 435: To expand on this:

We previously represented this offset with a `timespec`, which has an `int` seconds and a `size_t` nanoseconds. So `+0.2us` is representable (`{0, 2'000}`), but `-0.2us` is tricky. There's no `-0`, so we'd need to encode this as something like "nanoseconds always count up, ignoring sign of seconds" (ie `-0.2us` is `{-1, 998'000}`). This is awkward to work with/correct for, if our non-monotonic clock ever actually says the enclave is further ahead than the host! So instead I now represent this offset in a `double`.

src/ds/logger.h
eddyashton: Previously:
```
{"h_ts":"2021-04-26T10:26:02.030865Z","e_ts":"2021-04-26T10:26:02.030865Z",
{"h_ts":"2021-04-26T10:26:03.402864Z","e_ts":"2021-04-26T10:26:03.402865Z",
```
Identical times!

Now:
```
{"h_ts":"2021-04-26T10:27:12.872847Z","e_ts":"2021-04-26T10:27:12.871473Z",
{"h_ts":"2021-04-26T10:27:13.094136Z","e_ts":"2021-04-26T10:27:13.092686Z",
```
Enclave timestamp is slightly behind host timestamp, as in non-JSON logging.

src/ds/logger.h
achamayou: Alternatively we could extend timespec, but given the level of precision we expect here, using a double should be fine.

src/host/time_updater.h
```diff
@@ -11,24 +11,25 @@ namespace asynchost
 {
   class TimeUpdaterImpl
   {
-    using TClock = std::chrono::steady_clock;
-    TClock::time_point creation_time;
+    using TClock = std::chrono::system_clock;
```
achamayou, line 14: This is changing from a monotonic clock to one that is not. We need to carefully check that all logic that derives elapsed deltas (eg. consensus timeouts) is correctly clamping negative values to zero, or expose this in addition to the monotonic clock used for timeout purposes.

eddyashton: The only accessor to this within the enclave, `enclave::get_enclave_time()`, already does this clamping to ensure the observed time never moves backwards.

achamayou (approved):

Issue #2206, opened 2021-02-18
Add get_quotes_for_all_nodes C++ API


eddyashton:
We currently have `get_quote_for_this_node_v1`, but we also need to support a `get_quotes_for_all_nodes`, to retrieve quotes across all nodes. This API will expose node IDs, which we had previously tried to avoid making part of the C++ API.

PR #2511, opened 2021-04-26
Add `get_quotes_for_all_trusted_nodes_v1` API

eddyashton:
Completes #2206.

Opening as a draft for now because I think this API needs a little more thought. This initial version includes _all known_ node quotes directly from the KV, with no indication as to the node's current status. If the aim is to allow apps to build their own status endpoints, they will need to see node status (and perhaps other node data, like public RPC address?). I'm tempted to just make a `get_status_for_node_v1`, to keep these queries cleanly separated, and perhaps even make this `get_quote_for_node_v1` with a separate `get_ids_for_all_nodes_v1`, but the core question remains - what do we mean by "all nodes"? The current behaviour, where all historic nodes happen to remain in the KV, is supposed to be temporary. So maybe this API should ignore `Retired` nodes already? Do we need to provide an API to retrieve the quote for `Retired` nodes, which will eventually/in general require a historic query?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+
+- Added `get_untrusted_host_time_v1` API. This can be used to retrieve a timestamp during endpoint execution, accurate to within a few milliseconds. Note that this timestamp comes directly from the host so is not trusted, and should not be used to make sensitive decisions within a transaction (#2550).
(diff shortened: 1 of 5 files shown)


achamayou:
Paging in @lynshi who requested this initially.

I don't think we should include Retired nodes, that will eventually need to be done via a historical query and a different API. I think the idea is to get a view of the network from a user perspective at a given point, and validate participating nodes, so restricting this to trusted nodes is the right thing to do.

Members already have separate and more detailed access to this, they aren't the target audience afaict.

lynshi:
That's right, it's sufficient to retrieve only trusted nodes at that point in time.

achamayou (approved):

PR #2516, opened 2021-04-27
Add test for `secp256r1` as service and node identities curve

eddyashton:
NB: This still doesn't support _mixed curves_ within a service. If a node tries to join with an identity generated on a different curve, it will currently be unable to form channels with other nodes.

### DIFF
--- a/src/host/main.cpp
+++ b/src/host/main.cpp
@@ -347,6 +347,19 @@ int main(int argc, char** argv)
       )
     ->capture_default_str();
 
+  crypto::CurveID curve_id = crypto::CurveID::SECP384R1;
+  std::vector<std::pair<std::string, crypto::CurveID>> curve_id_map = {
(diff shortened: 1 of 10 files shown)


src/host/main.cpp
```diff
@@ -347,6 +347,19 @@ int main(int argc, char** argv)
...
+    {"secp256r1", crypto::CurveID::SECP256R1}};
+  app
+    .add_option(
+      "--curve-id",
+      curve_id,
+      "Elliptic curve to use as for node and network identities (used for TLS "
+      "and ledger signatures)")
+    ->transform(CLI::CheckedTransformer(curve_id_map, CLI::ignore_case))
+    ->capture_default_str();
+
```
eddyashton, line 362: Just moving this up, since its common to all subcommands (was previously amongst the `start` specific args).

achamayou (approved):

PR #2525, opened 2021-04-28
Add tests for mixed curves for node and service identities

eddyashton:
Followup to #2516.

This fixes a bug when nodes are on different curves (unable to create channels because `Verifier` got the wrong `md_type`, so signature verification failed), and adds tests for this in both the `channels_test` unit test and the end-to-end suites.

### DIFF
--- a/src/crypto/verifier.h
+++ b/src/crypto/verifier.h
@@ -13,7 +13,6 @@ namespace crypto
   {
   protected:
     std::shared_ptr<PublicKey> public_key;
-    MDType md_type = MDType::NONE;
 
   public:
     Verifier() : public_key(nullptr) {}
@@ -37,11 +36,6 @@ namespace crypto
       size_t sig_size,
       MDType md_type = MDType::NONE) const
     {
-      if (md_type == MDType::NONE)
-      {
-        md_type = this->md_type;
-      }
-
       return public_key->verify(
         contents, contents_size, sig, sig_size, md_type);
     }
@@ -75,9 +69,6 @@ namespace crypto
       MDType md_type,
       HashBytes& hash_bytes) const
     {
-      if (md_type == MDType::NONE)
-        md_type = this->md_type;
-
       return public_key->verify(
         contents, contents_size, sig, sig_size, md_type, hash_bytes);
     }
@@ -138,9 +129,6 @@ namespace crypto
       size_t sig_size,
       MDType md_type = MDType::NONE)
     {
-      if (md_type == MDType::NONE)
-        md_type = this->md_type;
-
       return public_key->verify_hash(hash, hash_size, sig, sig_size, md_type);
     }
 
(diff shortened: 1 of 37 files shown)


achamayou (approved):

src/node/test/channels.cpp
```diff
@@ -41,6 +41,8 @@ using MsgType = std::array<uint8_t, msg_size>;
 static NodeId self = std::string("self");
 static NodeId peer = std::string("peer");
 
+static constexpr auto default_curve = crypto::CurveID::SECP384R1;
```
wintersteiger, line 44: There's also `crypto::service_identity_curve_choice` as the default choice, maybe we should renamed that though.

src/crypto/verifier.h
```diff
@@ -138,9 +129,6 @@ namespace crypto
       size_t sig_size,
       MDType md_type = MDType::NONE)
     {
-      if (md_type == MDType::NONE)
-        md_type = this->md_type;
-
       return public_key->verify_hash(hash, hash_size, sig, sig_size, md_type);
```
wintersteiger, line 132: There may be a problem with `md_type == None` here. I think it works fine with OpenSSL, but the corresponding mbedTLS version won't even attempt to verify anything and return an error. Their docs say `md_alg may be MBEDTLS_MD_NONE, only if hash_len != 0`, but that's incorrect, it will throw an error even when `hash_len != 0`. Could you try running the tests with `-DCRYPTO_PROVIDER_IS_MBEDTLS`?

wintersteiger (approved):

src/node/test/channels.cpp
eddyashton: Yes - I'd like to remove that value entirely, since its now badly named. But doing so involves touching a bunch of unit tests with this kind of change, so I'll save it for a separate PR.

src/crypto/verifier.h
eddyashton: I believe we always grab a valid `md_type` later, in both the OpenSSL and mbedTLS implementations

```
    if (md_type == MDType::NONE)
    {
      md_type = get_md_for_ec(get_curve_id());
    }
```

I've tried several unit tests (`channels_test`, `crypto_test`, `tls_test`, `secret_sharing_test`) with `CRYPTO_PROVIDER_IS_MBEDTLS` and they're all happy.

src/node/test/channels.cpp
wintersteiger: Sure, sounds good!

tests/suite/test_suite.py
```diff
@@ -58,6 +58,21 @@
 ]
 suites["reconfiguration"] = suite_reconfiguration
 
+# This suite tests that the service supports nodes on different curves, and
```
jumaffre, line 61: This seems like quite an expensive way to test this. Could we add this to one of the existing test suite for example, e.g. `rekey_recovery`?

achamayou:
@eddyashton could you add an entry in the changelog, perhaps "The curve-id selected for the identity of joining nodes no longer needs to match that of the network"

Issue #2491, opened 2021-04-21
Reconfiguration test: ledger files don't match up between nodes


jumaffre:
At the end of the `reconfiguration_test_suite`, the ledger files between the primary don't seem to match up https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=23901&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=bbcdfa66-260b-5619-96df-9537b476c35e&l=8328:

```bash
43: Traceback (most recent call last):
43:   File "/mnt/build/2/s/tests/e2e_suite.py", line 190, in <module>
43:     run(args)
43:   File "/mnt/build/2/s/tests/e2e_suite.py", line 143, in run
43:     network.stop_all_nodes()
43:   File "/mnt/build/2/s/tests/infra/network.py", line 534, in stop_all_nodes
43:     raise Exception(
43: Exception: Ledger file on node 5326a975bc39a7ce108a29e4e58eb3135604a4cdbef4a78cfbdcd674051bbc0d does not exist on most up-to-date node 3d6694b7225c584786308d2c1295a499d057e9a15a8bd3db84f51ca1f980d26e: ledger_83-95.committed
```

We should also print clearer information about the ledger files present on each node at the end of the test run in case of error.

eddyashton:
Another repro in the Daily:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=25156&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=bbcdfa66-260b-5619-96df-9537b476c35e&l=11603

jumaffre:
I've managed to get a local repro for this on the `reconfiguration_test_suite` with JWT auto-refresh set to 1s (see #2600). Investigating this now.

PR #2526, opened 2021-04-28
Investigate ledger files not matching up between nodes

jumaffre:
See #2491


achamayou:
@jumaffre did we figure out what was causing this?

jumaffre:
Fixed in #2607

PR #2530, opened 2021-04-28
Change install prefix and cmake version parsing

eddyashton:
This does 2 main things:
- Always installs under `/opt/ccf` rather than `/opt/ccf-x.y.z`. This seems more common, can be overridden when necessary, and removes confusion around release candidates*.
- Changes how we parse the `git describe --tags` produced version name, and retains any suffix for later use. The real use case for this is producing a `ccf-2.3.4~rc1.deb`, but anything else that can usefully make a distinction based on this suffix could also do so.

Some expected parses for posterity:

```
On git tag ccf-3.4.5
-- CCF version = ccf-3.4.5
-- CCF release version = 3.4.5
-- CCF version suffix =

On git tag ccf-3.4.5-dev6
-- CCF version = ccf-3.4.5-dev6
-- CCF release version = 3.4.5
-- CCF version suffix = dev6

3 commits later, on a non-tag
-- CCF version = ccf-3.4.5-dev6-3-g7afa8f5f3
-- CCF release version = 3.4.5
-- CCF version suffix = dev6-3-g7afa8f5f3
```

Opening as a draft while I see if I can expand on this:
- [X] Make sure this full version is visible in the install directory. I'd like to just dump this to the `VERSION` file we already write, but this seems to be used to version our `npm` and `pip` packages as well, and I'm not sure about their naming conventions? _This is now done, with `VERSION` containing `3.4.5` and `VERSION_LONG` containing `ccf-3.4.5-dev2`_
- [x] Use the "version suffix" (eg - `rc1`) in more places, not just the Debian package name. _Done. See new `CCF_PYTHON_VERSION` variable which tries to construct a name that pip will be happy with (eg - `1.0.0`, `1.0.0.rc1`, `1.0.0.rc1+rc1-5-gfa7a7`), and npm using `VERSION_LONG` (without the "ccf-" prefix). These may break in annoying ways if we ever have adventurous tag names, but should work correctly for `-rc1`, `-dev2` etc that we're currently considering._
- [x] Replicate this logic in the source-directory-parsing path. If we can keep the full name in the tarball, it should also be parseable here? _Done. Just refactored this to use most of the existing logic. I'd thought this was to support finding the version from the install directory, but its not. The tarball is named from the tag, and the tag has the precise data we're using in the in-repo case._

\* If we have `2.0.0` and `2.0.0-rc1`, its hard to tell cmake about this distinction. It will look under `/opt/ccf` and `/opt/ccf-2.0.0`, but not `/opt/ccf-2.0.0-rc1`. Apps can't easily require an `rc` version, their dependency will just look for version `2.0.0` and find whichever it's given. So we try to upfront this and say you _always_ have a single CCF version; the last one you installed, under `/opt/ccf`.


achamayou (approved):

cmake/version.cmake
```diff
@@ -76,6 +76,31 @@ execute_process(
   RESULT_VARIABLE "VERSION_IS_SEMVER"
 )
 
+# Produce a valid version for the Python package
+set(CCF_PYTHON_VERSION "${CCF_RELEASE_VERSION}")
```
achamayou, line 79: I think I'd have been tempted to do this in Python to be easily able to add a few unit tests.

eddyashton: That's a really good point. I had convinced myself that CMake was a hammer, when it is clearly made from silly string.

python/versionifier.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+
+from setuptools.extern.packaging.version import Version, InvalidVersion
+
+def remove_prefix(s, prefix):
```
achamayou, line 6: A reason to upgrade to 3.9 if there ever was one.

cmake/version.cmake
eddyashton: This is now done by a Python script, so handles even more cases with helpful auto-canonicalisation by the Python `packaging.version`. In particular, it helpfully turns the git describe `1.2.3-42-deadbeef` (where 42 is the number of commits since `1.2.3`) into `<Version('1.2.3.post42+deadbeef')>`.

python/versionifier.py
eddyashton: We only have this in 2 places (3 if you include the C++ `nonstd` version...). Once its redefined in >5% of our files, then we upgrade/burn it all down.

achamayou: Joking aside, we have the suffix variant in quite a few more places. Still not a reason :)

Issue #1307, opened 2020-06-17
Enclave-host compatibility check


jumaffre:
We currently do not check whether a given enclave and host are compatible. For example, a recent enclave could write a new message on the ring buffer that is not supported by the host, which would cause the CCF node to crash, possibly long after the node has been started.

Instead, we should enforce that such situations can never occur. To do so, we should make sure that the version of `cchost` is recorded in the binary (so that `cchost --version` returns the release tag + commit). 

We should also record the version of CCF that an enclave was built against. It is not clear whether i) this should be recorded as a metadata to the enclave or ii) included in, e.g., `libccf.enclave.a`. If i), `cchost` could enforce this check without spinning up the enclave. If ii), the enclave would have to be spinned up to retrieve this and compare it to the version of `cchost`.

jumaffre:
As discussed with @achamayou, there are two approaches to this:

- The first approach is a `==` comparison between the `cchost` version and the enclave version. This guarantees that the same version of Open Enclave is used for both (so the `oe_create_xxx_enclave()` API is in sync), that the ecalls used are the same and that all ring buffer messages are the same. This should be achieved ahead of the enclave creation so we would either a) try to embed the CCF version in the enclave metadata (not clear how to do this) or b) use `ccf_version` in the enclave as well and read the symbols from the enclave before loading the enclave (preferable). Option b) should be preferable for simplicity. This method is too strict (perhaps two enclave/host pairs across two different CCF versions are compatible but we would prevent such combinations) but easy to implement. 

- The second is to have a more complex scheme where we (CCF devs) maintain an internal version number on the host and the enclave and update it whenever a breaking change is introduced. This would allow us to provide compatibility across different hosts and enclaves at the cost of higher complexity.

The first approach (strict equality) seems preferable for now as it gives us what we want (prevent breakages between enclave and host at runtime) at minimum cost. The cost of upgrading the enclave or the host is also cheap enough.

achamayou:
Let's go with the strict approach for now and for the foreseeable future. We can revisit this if:

1. There is a clear backwards compatibility story for Open Enclave OCalls
2. There is demand for difference cchost and enclave release cycles

PR #2532, opened 2021-04-28
Enclave host compatibility check

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/1307

We now check that the enclave and the host have been built from the exact same commit.

The check is run by the enclave so that we don't have to deal with the awkward ECALL handling of `out` strings (see https://github.com/openenclave/openenclave/blob/cd72fd7069488ba6f453c8f5f47bd9fd9a6e6c0d/docs/GettingStartedDocs/Edger8rGettingStarted.md#you-are-on-your-own), ~but at the cost of not being able to report a useful error message when the version check fails (enclave log message is not output by the host as too early in the enclave creation process). I also tried to delay the point at which the version check is done, so that we can output a useful error message, but it seems too late and the node has already done a fair bit of work by then.~ Edit: the enclave now returns a specific error code on the first ECALL, that the host can print.

As always, feedback welcome!


wintersteiger (approved):
I think the only way to get nice/meaningful error messages for early failures like this, is to return an error code (or string?) from `enclave_create_node` instead of just a bool.

src/enclave/main.cpp
```diff
@@ -59,6 +61,13 @@ extern "C"
     }
 #endif
 
+    // Host and enclave versions must match. Otherwise the node may crash much
+    // later (e.g. unhandled ring buffer message on either end)
+    if (std::string(host_version) != ccf::ccf_version)
+    {
+      return false;
```
achamayou, line 68: +1 for a custom return code, this is too difficult to debug otherwise.

src/enclave/main.cpp
```diff
@@ -59,6 +61,13 @@ extern "C"
     }
 #endif
 
+    // Host and enclave versions must match. Otherwise the node may crash much
+    // later (e.g. unhandled ring buffer message on either end)
+    if (std::string(host_version) != ccf::ccf_version)
```
ashamis, line 66: I might be missing something here but the comment in the PR says you compare commits but this is comparing versions. Not that I think there is any issue with comparing versions.

achamayou (approved):

src/enclave/main.cpp
jumaffre: Good point. The `ccf::ccf_version` is retrieved from `common/version.h` which is generated by `cmake` based on the long version string which is derived by `git describe --tags`, and changes for every commit (`e.g. "ccf-1.0.0-rc1-40-ge1d549c04"` for my current local checkout). So yeah, it's referred to as version because it will be a nice and neat version for major releases (e.g. `ccf-1.0.0`) but is more specific for everything in between.

src/enclave/main.cpp
jumaffre: Done

src/common/enclave_interface_types.h
```diff
@@ -0,0 +1,61 @@
...
+#pragma once
+
+enum CreateNodeStatus
+{
+  OK = 0,
+  InternalError = 1,
+  NodeAlreadyCreated = 2,
+  VersionMismatch = 3,
+  ConsensusNotAllowed = 4,
+  TooManyThreads = 5,
```
achamayou, line 12: MoreThreadsThanTCSCount would be a little more precise here

src/common/enclave_interface_types.h
```diff
@@ -0,0 +1,61 @@
...
+  MemoryNotInEnclave = 6,
```
achamayou, line 13: I think I can guess what this is, but I suspect it and the other values could do with comments.

src/common/enclave_interface_types.h
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+}
+
+enum StartType
+{
+  New = 1,
+  Join = 2,
+  Recover = 3,
+  Unknown = 100
```
achamayou, line 60: Why do we have this value?

achamayou (approved):

src/enclave/main.cpp
```diff
@@ -1,12 +1,14 @@
...
 // Licensed under the Apache 2.0 License.
+#include "common/enclave_interface_types.h"
 #include "ds/json.h"
 #include "ds/logger.h"
 #include "ds/spin_lock.h"
 #include "ds/stacktrace_utils.h"
 #include "enclave.h"
 #include "enclave_time.h"
 #include "oe_shim.h"
+#include "version.h"
```
eddyashton, line 11: Just a note, but don't think its actually a problem at the moment.

Most of our directory-less includes are in the same directory as the current file. This is an exception, since its in `build/generated`. This is not installed, so is not visible to an app. Its unclear that this is different and may confuse us when it breaks in future - perhaps we should put these all under `build/_gen/generated`, so we can call them `generated/version.h` here (but keeping an extra layer so we're adding `build/_gen` to the include dirs, not `build/`).

This won't break at the moment because apps don't actually build this file (its built into `libccf`, and they link to that). But now that it _looks_ like we have a `ccf_version` that can be used within the enclave, its probably only a matter of time before we try to use that in an app, and find that the include is surprisingly unavailable.

I'm not sure what we'd do to fix this. Maybe I'm saying `version.h` should be put under `include/ccf` as a public header, even though its generated?

eddyashton:
I think a check on the host would be more natural than a check in the enclave. If we ever want to use mixed versions, I think this will always be a newer host trying to talk to an older enclave (ie - the host is cheaper to rebuild and redeploy than the enclave, so if we can patch just the host that's preferable). If the enclave is checking versions, this new host needs to impersonate an _old_ host, and potentially different old host-versions depending on which enclave its talking to. If the host did the check, this could be more naturally expressed in the host code ("I'm 2.4.5, but I know I can still talk to an enclave on 2.4.3 or 2.4.4").

RE: Dealing with out-strings - I think we just create a big-enough buffer (say 128 chars), and give the enclave an out-ptr and size as we do for certs.

src/enclave/main.cpp
achamayou: This a very good point, I think distributing `version.h` as a public include is an excellent idea.

jumaffre:
> I think a check on the host would be more natural than a check in the enclave. If we ever want to use mixed versions, I think this will always be a newer host trying to talk to an older enclave (ie - the host is cheaper to rebuild and redeploy than the enclave, so if we can patch just the host that's preferable). If the enclave is checking versions, this new host needs to impersonate an _old_ host, and potentially different old host-versions depending on which enclave its talking to. If the host did the check, this could be more naturally expressed in the host code ("I'm 2.4.5, but I know I can still talk to an enclave on 2.4.3 or 2.4.4").
> 
> RE: Dealing with out-strings - I think we just create a big-enough buffer (say 128 chars), and give the enclave an out-ptr and size as we do for certs.

Good point. This is what I initially intended but the idea somehow got lost along the way. This is now done.

src/common/enclave_interface_types.h
jumaffre: I couldn't find out why so I've removed it.

src/common/enclave_interface_types.h
jumaffre: This seems to be hardcoded to `24` in our code base, independently of the `oe_sign.conf` `NumTCS`. Am I missing something?

eddyashton (approved):

src/enclave/main.cpp
jumaffre: See #2562

PR #2547, opened 2021-04-30
Continue to send append entries for current term when retiring

achamayou:
Related to #2456, rather than aggressively shut down all channels upon noticing that a node's retirement is committed, instead:

1. Continue to serve append entries up to this point
2. Stop sending heartbeats, so that nodes that have caught up will time out and stage an election

This diminishes the amount of catch-up that the new primary needs to perform.


src/consensus/aft/raft.h
```diff
@@ -95,6 +95,16 @@ namespace aft
...
     std::unordered_set<ccf::NodeId> votes_for_me;
 
+    // Replicas start in state Follower. Apart from a single forced
+    // transition from Follower to Leader on the initial node at startup,
+    // the state machine is made up of the following transitions:
+    //
+    // Follower -> Candidate, when election timeout expires
+    // Follower -> Retired, when commit advances past the last config containing
+    // the node Candidate -> Leader, upon collecting enough votes Leader ->
+    // Retired, when commit advances past the last config containing the node
```
jumaffre, line 105:
```suggestion
    // Follower -> Retired, when commit advances past the last config containing
    // the node 
    // Candidate -> Leader, upon collecting enough votes 
    // Leader -> Retired, when commit advances past the last config containing the node
```

src/consensus/aft/raft.h
```diff
@@ -1022,6 +1032,13 @@ namespace aft
       const ccf::NodeId& to, Index start_idx, Index end_idx)
     {
       const auto prev_idx = start_idx - 1;
+
+      if (replica_state == Retired && start_idx >= end_idx)
+      {
+        channels->destroy_channel(to);
```
jumaffre, line 1038: Could we add a log line here to say that node `to` has been caught up after "our" retirement?

src/consensus/aft/raft.h
```diff
@@ -2406,7 +2438,6 @@ namespace aft
 
       LOG_INFO_FMT(
```
jumaffre, line 2439: Can we add a test for this in `reconfiguration.py:test_retire_primary`? We should be able to observe that the commit level on majority of backups has gone past the `seqno` at which the retire node proposal was accepted before the election is started.

src/consensus/aft/raft.h
achamayou: Sure

src/consensus/aft/raft.h
achamayou: I have no idea how to do this without a serious race condition I'm afraid, do you have something specific in mind?

jumaffre: Fair enough. The election may have already happened by the time we check for commit of the primary's retirement.

jumaffre (approved):

PR #2550, opened 2021-04-30
Add C++ API to get untrusted host time

### LINKED ISSUE #2523: Expose `get_untrusted_time_v1()`
Enabled by #2509, may be used to implement #2468.

eddyashton:
Resolves #2523.

Adds the API `get_untrusted_host_time_v1`, which returns the host-written (UTC) time as a `timespec`.

Also adds an example endpoint using this in the `nobuiltins` app, and a few test calls to confirm the reported time is close enough to the true time.

### DIFF
--- a/samples/apps/nobuiltins/nobuiltins.cpp
+++ b/samples/apps/nobuiltins/nobuiltins.cpp
@@ -37,6 +37,14 @@ namespace nobuiltins
   DECLARE_JSON_TYPE(TransactionIDResponse)
   DECLARE_JSON_REQUIRED_FIELDS(TransactionIDResponse, transaction_id)
 
+  struct TimeResponse
+  {
+    std::string timestamp;
+  };
+
+  DECLARE_JSON_TYPE(TimeResponse)
+  DECLARE_JSON_REQUIRED_FIELDS(TimeResponse, timestamp)
+
   // SNIPPET: registry_inheritance
   class NoBuiltinsRegistry : public ccf::BaseEndpointRegistry
   {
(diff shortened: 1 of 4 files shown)


src/endpoints/base_endpoint_registry.cpp
```diff
@@ -243,4 +244,18 @@ namespace ccf
...
+  ApiResult BaseEndpointRegistry::get_untrusted_host_time_v1(::timespec& time)
+  {
+    const std::chrono::microseconds now_us = enclave::get_enclave_time();
+
+    constexpr auto us_per_s = 1'000'000;
+    time.tv_sec = now_us.count() / us_per_s;
+    time.tv_nsec = (now_us.count() % us_per_s) * 1'000;
+
+    LOG_INFO_FMT(
+      "{} -> {{ {}, {} }}", now_us.count(), time.tv_sec, time.tv_nsec);
```
ashamis, line 257: this should probably be at a lower log level

samples/apps/nobuiltins/nobuiltins.cpp
```diff
@@ -227,6 +235,50 @@ namespace nobuiltins
...
+        }
+
+        std::tm calendar_time;
+        gmtime_r(&time.tv_sec, &calendar_time);
+
+        // 20 characters for a (timezoneless) ISO 8601 datetime, plus a
+        // terminating null
+        constexpr size_t buf_size = 21;
+        char buf[buf_size];
+        if (strftime(buf, buf_size, "%Y-%m-%dT%H:%M:%S", &calendar_time) == 0)
```
ashamis, line 258: consider adding milliseconds? We may not need it now but adding it later would be a breaking change

ashamis (approved):

src/endpoints/base_endpoint_registry.cpp
eddyashton: This was a debugging line, has now been removed.

samples/apps/nobuiltins/nobuiltins.cpp
eddyashton: I add microseconds below (there's no specifier for milliseconds with strftime, and `calendar_time` only goes down to second precision). Also we don't care about breaking changes to this endpoint - its just a sample endpoint so we can tweak the sample formatting here if we want to recommend something else. The API guarantee is on the `get_untrusted_host_time_v1` API, which returns a `timespec`.

achamayou (approved):

jumaffre (approved):

PR #2553, opened 2021-05-05
Python infra for testing node partitions

jumaffre:
This PR introduces a new Python module to test for network partitions as part of our existing end-to-end test infra.

The partitioning relies on the privileged [`iptables`](https://en.wikipedia.org/wiki/Iptables) to isolate nodes from the network (hence the new CI docker image and docker CI container new runtime capabilities). 

The only notable change to `cchost` is the new `--node-client-host` CLI option to specify the hostname/IP node-to-node client connections should bind to. This allows the infra to specify a known IP for each node so that it is possible to isolate a node from another one (as opposed to a node from _all_ the other ones). Without this, all nodes would bind to `127.0.0.1:<random port>` for each node outgoing connection, which is awkward to discover.

Examples on how to use the new `infra.partitions` module is in `partitions_test.py`, but in short:
- You can now isolate a single node from another one: `isolate_node(node, other)` or from the rest of the network `isolate_node(node`). 
- You can create arbitrary partitions of nodes: `partition([node0, node 1], [node 2])`.

These methods return a `Rules` object that can be manually `drop()`ed or used with a context manager.

I've still got a few pending questions that I'll post inline in this PR.


src/host/node_connections.h
```diff
@@ -344,6 +348,15 @@ namespace asynchost
       TCP s;
       s->set_behaviour(std::make_unique<OutgoingBehaviour>(*this, node));
 
+      if (client_host.has_value())
```
jumaffre, line 351: Should/could this binding to the client host be more integrated in the existing TCP flow? If the binding fail, should we retry? 

In fact, just writing this comment makes me realise that this connection would proceed even if `uv_tcp_bind` fails, so I should integrate this in the existing flow better.

tests/infra/partitions.py
```diff
@@ -0,0 +1,196 @@
...
+    All member functions return a :py:class:`infra.partitions.Rules` object that can be used
+    in a context manager and which will be automatically dropped when the object goes out of
+    scope.
+
+    Note: It should be managed by a :py:class:`infra.network.Network` instance so that rules
+    outlive nodes to avoid spurious log messages when the network is shutdown.
+    """
+
+    @staticmethod
+    def cleanup():
```
jumaffre, line 69: The iptables rules are cleaned up by the `infra.network.Network` on shutdown. However, is the test is killed halfway through, the rules won't currently be cleaned up. This is kind of OK if we keep re-running partitions test as the rules are destroyed on creation (see constructor a few lines down) but may affect non-partition tests (in the unlikely but still possible event of port collisions). 

I'd started with adding signal handlers but then remembered that signals are not forwarded by `ctest` to the Python interpreter. A few things we could do:
- Adding a wrapper in `cmake`'s `add_e2e_test` function to save and restore the current iptables rules before and after a partition test. 
- Still add signal handlers so that the rules are probably cleaned up when running without `ctest`. 

Thoughts?

tests/infra/network.py
```diff
@@ -147,6 +148,11 @@ def __init__(
         self.dbg_nodes = dbg_nodes
         self.perf_nodes = perf_nodes
 
+        # Requires admin privileges
+        self.partitioner = (
```
jumaffre, line 152: Because this required admin privileges, should we move all partition tests under a different `ctest` label so that it is easy to exclude them when running all tests from the command line? In other words, `ctest -L e2e` should always work, but `ctest -L partitions` would require `sudo`?

src/host/main.cpp
```diff
@@ -107,6 +107,13 @@ int main(int argc, char** argv)
     "Path to which the node's node-to-node address (including potentially "
     "auto-assigned port) will be written. If empty (default), write nothing");
 
+  std::string node_client_host = {};
+  app.add_option(
+    "--node-client-host",
+    node_client_host,
+    "Host on which to connect to for commands sent to other nodes. If "
```
achamayou, line 114:
```suggestion
    "Host on which to bind to for commands sent to other nodes. If "
```

I think this would be better as "address", it would be more consistent with our other flags, "host" is a bit overloaded.

src/host/node_connections.h
achamayou: Yes! :)

.azure-pipelines-templates/matrix.yml
```diff
@@ -50,15 +50,15 @@ jobs:
...
-          ctest_filter: "${{ parameters.test.perf.ctest_args }}"
+  # - ${{ if eq(parameters.perf_tests, 'run') }}:
+  #     - template: common.yml
+  #       parameters:
+  #         target: SGX
+  #         env: ${{ parameters.env.Perf }}
+  #         cmake_args: "${{ parameters.build.common.cmake_args }} ${{ parameters.build.perf.cmake_args }} ${{ parameters.build.SGX.cmake_args }}"
+  #         suffix: "Perf"
+  #         artifact_name: "SGX_Perf"
+  #         ctest_filter: "${{ parameters.test.perf.ctest_args }}"
```
eddyashton, line 61: TODO: Re-enable these before committing.

achamayou (approved):
This is great!

I think we'd like to add a variant of the rotation test that isolates nodes in turn using this.

src/host/main.cpp
jumaffre: It seems that we already define an address as a `host:port` pair. In this specific case, we don't want operators to specify a port as _many_ outgoing n2n connections are created on the node. 

(Note: I actually tried to use a hardcoded port here to see how it would fail but to my surprise, it didn't. I believe this is because libuv uses [`SO_REUSEADDR`](https://github.com/libuv/libuv/blob/v1.x/src/unix/tcp.c#L164)).

achamayou: Good point, how about "interface" then?

tests/infra/partitions.py
achamayou: Is there any way to tag the rules we set up so we can recognise them somehow?

I would otherwise be in favour of cleaning up any pre-existing rules at test startup, at least for the hosts we know we want to use. Trying to ensure cleanup happens on shutdown every time is difficult to guarantee with absolute certainty.

tests/infra/network.py
achamayou: Yes, this is an excellent idea.

tests/infra/partitions.py
jumaffre: The rules are all under the `CCF-TEST` iptables chain which is now cleaned up on startup. The only other issue is when the privileged `partitions_test` is interrupted and doesn't cleanup the chain, and a _non-privileged_ test runs after it but can't cleanup the left-over chain. 

If this happens on the CI, this would actually be OK as the `partitions_test` runs last. If this happens locally, we may have failing tests but maybe that's not such a big deal as re-running the `partitions_test` would hopefully solve this?

tests/infra/network.py
jumaffre: Done. As per `test.yml`, we use this to run the partitions test with `sudo` in a separate step than the other non-privileged end-to-end tests.

tests/infra/partitions.py
```diff
@@ -0,0 +1,196 @@
...
+        self,
+        *args: List[infra.node.Node],
+        **kwargs,
+    ):
+        """
+        Creates an arbitrary number of partitions of :py:class:`infra.node.Node`. All other joined nodes in the
+        :py:class:`infra.network.Network` are also isolated in their own partition.
+
+        :param List[infra.node.Node] *args: A variable length argument list of :py:class:`infra.node.Node` (i.e. partitions) to isolate.
+        :param str **name: Name of the partition (optional).
```
eddyashton, line 151: What's the use case for setting a name here? It seems like we want the names to be canonical, so we can reliably detect when we already have a collision, so its risky for the caller to override this? Perhaps this name is not used for clearing things later.

If it needs to be canonical, I think the partitions should be specified as `set`s rather than `list`s, since the partitions `[n1, n2]` and `[n2, n1]` are logically equivalent and we want to give them the same name.

tests/partitions_test.py
```diff
@@ -0,0 +1,104 @@
...
+    return network
+
+
+@reqs.description("Partition primary + f nodes")
+def test_partition_majority(network, args):
+    primary, backups = network.find_nodes()
+
+    # Create a partition with primary + half remaining nodes (i.e. majority)
+    partition = [backups[0]]
+    partition.extend(backups[len(backups) // 2 :])
```
eddyashton, line 49: What am I missing here?

It looks like this creates a majority partition that doesn't include the primary (it includes the first backup, and the last half-rounded-up backups), which should cause an election?

tests/partitions_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Note: Context manager
+    with network.partitioner.partition(partition):
+        try:
+            network.wait_for_new_primary(primary.node_id)
+            assert False, "No new primary should be elected when partitioning majority"
+        except TimeoutError:
+            pass
+
+    return network
```
eddyashton, line 59: Minor: Although there's no election during this partition, there will be one when the network anneals. Any node that has been partitioned from the primary will start calling elections, and when it rejoins the network it will continue to call them. It can't win if more transactions have come in (someone in the majority partition that has the up-to-date state will win*), but it may still cause a leader change. This might be a problem for whichever test starts next, so I think we should wait for all nodes to settle down after removing partitions.

\* This is probably worth testing in this test. While the primary is in a majority partition, we can actually submit transaction to the primary or any of the in-partition backups, and these will commit and persist.

tests/partitions_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Isolate first backup from primary so that first backup becomes candidate
+    # in a new term and wins the election
+    # Note: Managed manually
+    rules = network.partitioner.isolate_node(primary, backups[0])
+
+    network.wait_for_new_primary(backups[0].node_id)
+
+    # Explicitly drop rules before continuing
+    rules.drop()
```
eddyashton, line 74: Again I may be missing something about the partitioning logic here, but I think the comment's inaccurate. If we partition the primary and the first backup, then if there's a majority of _remaining_ backups, one of them will win an election. That must be what's happening - `wait_for_new_primary(backups[0].node_id)` means wait for anyone _except `backups[0]`_ to become primary. I'm also not sure why this is managed manually. But we don't gain anything by partitioning `backups[0]` as well, that just increases the required number of nodes (equivalent to crashing 2 nodes rather only 1 primary)?

```suggestion
    # Isolate first backup from primary so that first backup becomes candidate
    # in a new term and wins the election
    # Note: Managed manually
    rules = network.partitioner.isolate_node(primary)

    network.wait_for_new_primary(primary.node_id)

    # Explicitly drop rules before continuing
    rules.drop()
```

eddyashton: (Ignore this, I'd completely missed that this call was `isolate_node` rather than `partition`)

tests/infra/partitions.py
achamayou: Can we add a non-privileged check at the beginning of all tests, that tells us if anything is in that chain and suggests a clean-up command?

src/host/node_connections.h
jumaffre: Done.

tests/infra/partitions.py
jumaffre: Unfortunately, even listing the iptables rules requires privileges (unless we want to fiddle with capabilities. which I don't think we want to). See https://unix.stackexchange.com/questions/385109/can-you-list-iptables-as-a-non-root-user-and-why. 

I will list a `iptables` CLI command to cleanup the chain manually in `partition.py` as an alternative, so that it is easy to recover from if this ever happens.

tests/infra/partitions.py
jumaffre: As discussed, the name doesn't have to be canonical as the rules are already identified by their value (iptables rule dict). It is only optionally used to easily name/identify partition rules from the caller. I'm leaning towards leaving it in as our use case for testing network partitions isn't fully defined yet.

tests/partitions_test.py
jumaffre: Good spot. Changed to `partiton = [primary]` to match the comment and intent of the test.

tests/infra/partitions.py
jumaffre: I've added a new `cleanup_iptables` Python CLI to cleanup the iptables manually, in case something goes wrong on a local run.

jumaffre:
I'm investigating some of the flakiness associated with the new partitions end-to-end tests, which I guess are expected since this is what we're building this test infra for! We may have to comment some of those out before merging this in.

jumaffre:
> I'm investigating some of the flakiness associated with the new partitions end-to-end tests, which I guess are expected since this is what we're building this test infra for! We may have to comment some of those out before merging this in. I'll create follow-up tickets if that's the case, also including the rotation test that @achamayou mentioned.

The tests should now be stable. The issue was that a signature is emitted by a freshly-elected primary, even if no write transaction were processed since it was elected. This is because the time-based signature emission logic doesn't know about the time the previous transaction was emitted (in the previous term, by a different node). This is actually OK (this gives us a quick ledger evidence that a new node was elected) but wasn't something I'd consider before.

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "prettier": "^2.3.0"
+  }
+}
```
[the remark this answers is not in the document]
jumaffre: Good spot!

jumaffre:
The CI failures on the multi-threading job aren't related to these changes so merging this PR now as we've had four successful CI runs in a row. Follow-up work will start in #2577.

PR #2556, opened 2021-05-06
Modify sample historical range query endpoint to better handle subranges

eddyashton:
Adds a table storing the version where each key (private record) was first written to, so that we can start a historical query from there. As part of this, makes `from_seqno` and `to_seqno` optional parameters, so they can be defaulted to tight values where possible.

So previously if you wanted the full range you had to say `GET /historical/range?id=42&from_seqno=1&to_seqno=10000`. You'd need to know the first ever seqno for `from_seqno`, and set the current/high value as `to_seqno`. We can easily make these optional and default those values, so you'd submit `GET /historical/range?id=42` and internally do `from=1, to=<committed>`. But if your ID `42` was actually only written to 5 times, between seqno '1050' and seqno `1070', this query is very inefficient - you needlessly retrieve the entire historical range. Using this table that tracks the first write, paired with `version_of_previous_write`, we can take `GET /historical/range?id=42` and internally compute the default values `from=1050, to=1070`, to only fetch the necessary range.

Caveats:
- If the key is deleted we don't know the end of the range, because we don't want to track when every key was deleted, so we need to resort to the current commit point again
- This optimisation is much faster if your ID is only used in a small range of seqnos. It doesn't help for arbitrary sparseness, for instance if you only have 5 writes but the first is at `12` and the last is at `99984`, we need to fetch the entire seqno range. We could handle that with a more detailed index, extending the first-write to "written in this epoch of seqnos", either stored in the KV or node-locally. That's out-of-scope for this PR.


achamayou (approved):

samples/apps/logging/logging.cpp
```diff
@@ -157,8 +169,10 @@ namespace loggingapp
             "Cannot record an empty log message.");
         }
 
-        auto records_handle = ctx.tx.rw(records);
+        // SNIPPET: private_table_access
+        auto records_handle = ctx.tx.template rw<RecordsTable>("records");
```
jumaffre, line 173: Why changing those everywhere in this app? Having a unique reference to the map object seems like good practice for apps (avoid typo in table name, etc.).

jumaffre (approved):

samples/apps/logging/logging.cpp
eddyashton: So this pattern of having a single instance of a `kv::Map` isn't something we're now doing in most places, so I don't think we should recommend it in the sample app. In particular, it was used in the docs to suggest this creates a `Map`, which is no longer true under the hood (does the distinction matter? Maybe not).

On the one hand, giving this repeated string a name to avoid typos seems like a good idea. On the other, I'm keen to reinforce to anyone reading this that a `Map` can be accessed from anywhere, purely by name, and hopefully with the same type. You don't need a single value (that could have some lookup magic under the hood), you can construct your string names on-the-fly. So I think there's value in having these redundant, and perhaps even not-best-practice, for clarity when reading a single endpoint? Arguable.

jumaffre: Perhaps just defining the name of the table once instead of repeating the name of the table in multiple places is a good compromise?

PR #2566, opened 2021-05-07
Sample consistency pass

### LINKED ISSUE #2345: Sample consistency pass
App samples have been mutated over time, and aren't necessarily a coherent whole at this point, particularly in terms of naming conventions (eg. `args` vs `ctx`).

We ought to review them and make sure they're consistent and only expose best practices.

eddyashton:
As outlined in #2345, our samples have got a mix of old and new styles in many places. In particular still using the old generic `args` variable name for the `EndpointContext` object. This replaces those with `ctx`, and removes some string duplication.

EDIT: Resolves #2345.

### DIFF
--- a/samples/apps/logging/logging.cpp
+++ b/samples/apps/logging/logging.cpp
@@ -23,13 +23,13 @@ using namespace nlohmann;
 namespace loggingapp
 {
   // SNIPPET: table_definition
-  using RecordsTable = kv::Map<size_t, string>;
+  using RecordsMap = kv::Map<size_t, string>;
   static constexpr auto PUBLIC_RECORDS = "public:records";
   static constexpr auto PRIVATE_RECORDS = "records";
 
   // Stores the index at which each key was first written to. Must be written by
   // the _next_ write transaction to that key.
-  using FirstWritesTable = kv::Map<size_t, ccf::SeqNo>;
+  using FirstWritesMap = kv::Map<size_t, ccf::SeqNo>;
   static constexpr auto FIRST_WRITES = "first_write_version";
 
   // SNIPPET_START: custom_identity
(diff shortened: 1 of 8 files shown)


samples/apps/logging/logging.cpp
```diff
@@ -250,7 +253,7 @@ namespace loggingapp
             std::move(error_reason));
         }
 
-        auto records_handle = ctx.tx.template rw<RecordsTable>("records");
+        auto records_handle = ctx.tx.template rw<RecordsTable>(PRIVATE_RECORDS);
```
achamayou, line 256: I feel like we used to have 80/20 `tables`/`maps`, and we're about 40/60 now? I have a mild inclination to go all `maps`, but what does everyone else think?

src/node/rpc/test/frontend_test.cpp
```diff
@@ -208,15 +208,15 @@ class TestAlternativeHandlerTypes : public BaseTestFrontend
   {
     open();
 
-    auto command = [this](auto& args) {
-      args.rpc_ctx->set_response_status(HTTP_STATUS_OK);
+    auto command = [this](auto& ctx) {
+      ctx.rpc_ctx->set_response_status(HTTP_STATUS_OK);
```
achamayou, line 212: I wonder to what extent splitting `ctx.rpc_ctx` into `ctx.request` and `ctx.response` would be justifiable.

achamayou (approved):

samples/apps/logging/logging.cpp
eddyashton: I'm wary of `map` because its overloaded (verb vs noun, type vs operation, existing std type), but I think since we've settled on `Map` for the core type you're right we should probably consolidate on `map` for the specialized types and variables.

src/node/rpc/test/frontend_test.cpp
eddyashton: Yeah, we should do this. It was slightly awkward in the original HTTP migration where there was some shared state (eg - the packing, whether that was JSON or msgpack, being the same for both request and response). I believe these are now cleanly separated - one may inform the other but they're tracked in distinct fields. This should also make `reset_response` simpler + safer.

eddyashton: https://github.com/microsoft/CCF/issues/2568

samples/apps/logging/logging.cpp
eddyashton: So I've renamed `Table` to `Map` for these types in `logging.cpp`, but going to punt on a wider rename. Renaming `NetworkTables` and `ccf::Tables` is a breaking change, so deserves a bit more thought (maybe I roll it into the statically-named-maps PR I was working on, which touches most of these references anyway...)

achamayou: I think that all other things being equal I'd rather be consistent with what we've decided too, but I do agree that we have to consider if the break is worth the improvement.

jumaffre (approved):

Issue #2358, opened 2021-03-25
Write a test that exercises retiring a node and starting another node on the same ip:port shortly after


achamayou:
Reported by @lynshi, not something we test at the moment because of our deliberate host:port randomisation to avoid test collisions.

achamayou:
We're also missing a test for retiring a pending, but not trusted node.

PR #2567, opened 2021-05-07
Test nodes can be replaced successfully with same listening addresses

achamayou:
One of two items under #2358.

There is no suggestion that this is a problem since we've switched to using digests of public keys for node ids, but it should be tested nonetheless.


wintersteiger (approved):
Not sure whether replay attacks are a problem here, but if so, should we add a negative test to check what happens if the node id is "accidentally" the same as the old one?

eddyashton (approved):

wintersteiger:
My question was about a public key being re-used, e.g. in a replay attack where someone tries to impersonate an old or existing node. It's probably fine, I just thought we could add a test to make sure that at least one of the safeguards triggers in these circumstances.

Conceivably, there could also be applications that overwrite the keys via custom ecalls (smart operator injecting the same key into all their enclaves, or attempts to get keys via secure key release, ...). They keys in the node tables wouldn't match up though, so I guess this would be caught very early.

achamayou:
@wintersteiger the replay attack here would involve replaying the TLS handshake for the joining RPC successfully, unless I am missing something?

I don't believe we claim any properties in the face of attackers able to read and mutate arbitrary enclave memory, it's essentially game over for all transactions from that point onward.

tests/replacement.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: Perhaps just include this in the existing `reconfiguration.py` test?

jumaffre (approved):

tests/replacement.py
achamayou: Yes, you're right.

achamayou: So that turns out to be awkward, because the test leaves the network with f dead nodes. But then that's probably a bad thing in the suite sooner or later too, so I think it's best to address that.

PR #2569, opened 2021-05-07
Add `.size` and `.clear()` to JS maps

### LINKED ISSUE #2512: Expose Map.size|clear to JS
https://github.com/microsoft/CCF/pull/2493/files
https://github.com/microsoft/CCF/pull/2494/files

`.size` should be a property, `.clear()` a method, in line with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

eddyashton:
Resolves #2512.

I also added endpoints to the logging app, in both C++ and JS, to test this. We test the KV function calls in unit tests but don't have any unit tests for the JS wrappers, so these are tested by a full e2e flow.


js/ccf-app/src/polyfill.ts
```diff
@@ -52,13 +52,20 @@ class KvMapPolyfill implements KvMap {
...
   forEach(
     callback: (value: ArrayBuffer, key: ArrayBuffer, kvmap: KvMap) => void
   ): void {
     this.map.forEach((value, key, _) => {
       callback(value, unbase64(key), this);
     });
   }
+  get size(): number {
+    return this.map.size;
+  }
```
eddyashton, line 68: Questions for @letmaik:

- JS says that `Map.clear` [returns undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/clear#return_value), but the TS `Map` returns void. I think we want to match the documented JS API, but maybe I'm missing something.
- Is this a sane way to implement this property getter? I was hoping I could just write an alias like `readonly size = this.map.size;`, but it seems the `readonly` that's fine in the `interface` means something very different (const, assigned-on-construction) in the implementations.

samples/apps/logging/js/app.json
```diff
@@ -26,6 +26,26 @@
...
+    "/log/private/count": {
+      "get": {
+        "js_module": "logging.js",
+        "js_function": "count_private",
+        "forwarding_required": "always",
+        "authn_policies": ["jwt", "user_cert"],
+        "mode": "readonly",
+        "openapi": {}
+      }
+    },
```
eddyashton, line 48: Not particularly RESTful, but a) its 'just' a sample, b) a reasonable extension to what's already there?

js/ccf-app/src/global.ts
```diff
@@ -41,9 +41,11 @@ export interface KvMap {
   get(key: ArrayBuffer): ArrayBuffer | undefined;
   set(key: ArrayBuffer, value: ArrayBuffer): KvMap;
   delete(key: ArrayBuffer): boolean;
+  clear(): undefined;
```
letmaik, line 44: This should be `void` since the result should never be assigned to a variable. Note that this is a TS thing, JS does not have void and the real thing that's returned is `undefined`, but the correct TS type here is `void`.

js/ccf-app/src/polyfill.ts
letmaik: Regarding the second point, the getters look fine. I don't think there's a substantially nicer way to do this.

samples/apps/logging/logging.cpp
```diff
@@ -262,6 +262,29 @@ namespace loggingapp
...
         .install();
 
+      auto clear = [this](auto& ctx, nlohmann::json&&) {
+        auto records_handle = ctx.tx.template rw<RecordsTable>("records");
+        records_handle->foreach([&ctx](const auto& id, const auto&) {
+          update_first_write(ctx.tx, id);
+          return true;
+        });
+        records_handle->clear();
+        return ccf::make_success(true);
```
eddyashton, line 272: This is rubbish, looping twice. But necessary to keep the optimisation from #2556 correct - whenever we write to a private ID (even if we don't know it uniquely), we _may_ need to update the first write table. Still calling `clear()` afterwards since we want to test its implementation e2e. Expensive but correct, which is what we want for these samples.

js/ccf-app/src/global.ts
eddyashton: Thanks, that makes sense. I'd started from the raw-JS implementation, don't know why I was trying to extend its types into our TS API!

achamayou (approved):

PR #2579, opened 2021-05-10
js: add function to validate a cert chain

### LINKED ISSUE #2503: js: add ccf.isValidX509CertChain
`ccf.isValidX509CertChain(chain, caBundle)` where both `chain` and `caBundle` are strings of concatenated PEMs.

I tried to use existing npm libraries but the best candidate (node-forge) unfortunately doesn't support EC keys (there's an open issue from 2014). jsrsasign doesn't have chain validation at all.

Prior art:
- [Forge](https://github.com/digitalbazaar/forge/blob/c666282c812d6dc18e97b419b152dd6ad98c802c/lib/x509.js#L2944):
   ```js
   const caCerts = [...]; // array of PEM-encoded certs
   const chain = [...]; // array of PEM-encoded certs, order is significant
   const caStore = pki.createCaStore(caCerts);
   const valid = pki.verifyCertificateChain(caStore, chain, {
     validityCheckDate: null // null = no check; default = current date; Date object = specified date
   });
   ```
- OpenSSL:
  - main API: https://www.openssl.org/docs/manmaster/man3/X509_verify_cert.html
  - validation options: https://www.openssl.org/docs/manmaster/man3/X509_VERIFY_PARAM_set_flags.html

letmaik:
Closes #2503.

This also adds a function to validate that a string is exactly a single PEM cert.

Note: The existing mbedtls `CertificateChain` was wrongly implemented and effectively only supported chains of size one, which happened to be what CCF uses.


src/crypto/mbedtls/verifier.cpp
```diff
@@ -108,52 +108,68 @@ namespace crypto
...
-        {
-          mbedtls_x509_crt_init(&raw[i]);
-        }
+      mbedtls_x509_crt_init(&raw);
+    }
 
-        for (size_t i = 0; i < certs.size(); i++)
+    void add(const std::vector<const Pem*>& certs)
+    {
+      for (size_t i = 0; i < certs.size(); i++)
```
achamayou, line 120: for(auto & cert: certs) perhaps, since you don't seem to use the index? Alternatively, include the index in the exception message (could not parse PEM in chain at position {i}...)?

eddyashton (approved):

achamayou (approved):

PR #2575, opened 2021-05-10
Install doc improvements

fritzalder:
This PR fixes two minor issues in the install docs: An old version number (now a dynamic pull of the latest release) and a wrong ansible script (app-run instead of app-dev).

PR #2581, opened 2021-05-11
Fix broken code block in install doc

fritzalder:
This fixes the code block that broke with the earlier #2575 Thanks @jumaffre for noticing


jumaffre (approved):
/azp run

PR #2582, opened 2021-05-11
Add new `/node/version` endpoint

jumaffre:
Follow-up from #2562 

This PR adds a new `/node/version` endpoint to return the CCF version a node was built from. This is useful for operators during a code update to check which nodes run the new version of CCF, without having to deal with the opaque `mrenclave`.

### DIFF
--- a/tests/reconfiguration.py
+++ b/tests/reconfiguration.py
@@ -242,6 +242,16 @@ def test_node_filter(network, args):
     return network
 
 
+@reqs.description("Get node CCF version")
+def test_version(network, args):
+    nodes = network.get_joined_nodes()
+
+    for node in nodes:
+        with node.client() as c:
+            r = c.get("/node/version")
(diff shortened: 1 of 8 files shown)


tests/reconfiguration.py
```diff
@@ -242,6 +242,16 @@ def get_nodes(status):
...
 
 
+@reqs.description("Get node CCF version")
+def test_version(network, args):
+    nodes = network.get_joined_nodes()
+
+    for node in nodes:
+        with node.client() as c:
+            r = c.get("/node/version")
+            assert r.body.json()["version"] == args.ccf_version
```
achamayou, line 252: The naming feels a little vague here, the "node" is effectively CCF+app, and this only talks about the CCF version, so I think either the path or the name should mention "ccf" or "framework" somewhere.

Issue #2521, opened 2021-04-28
Split interfaces with separate hard and soft connection caps


achamayou:
The following changes would help handle the situation where user sessions are saturating the RPC interface:

1. Allow cchost to listen to multiple interfaces with individual fd/session quotas, so that the network operator can reserve one for their usage even when a user-facing interface is at capacity.
2. Add the notion of a soft cap, which if reached causes new sessions to return a 429 and close. The existing hard cap will continue to exist and will still cause a connection closure at the TCP level.

The basic requirement is for two interfaces, one for operators, the other for users. While not necessary, it may be nice to introduce some additional routing restriction (ie. the interface exposed to users can only access the User frontend).

Since adding support for any number of interfaces is probably similar in complexity to adding two, we may as well do that.

PR #2583, opened 2021-05-12
Add a soft session cap, returning HTTP errors rather than closed connections

eddyashton:
Part of #2521.

Note that this returns errors for _new sessions_, but we don't do anything to discourage/disconnect existing connections. We could develop some scheme to return errors on existing sessions for some kind of fairness, but I don't know if its worth the complexity?

This makes the session exhaustion behaviour far friendlier to clients, and changes what we're able to easily test in `connections_test_cft`.

TODO:
- [x] Add to CHANGELOG
- [X] Check error code. _The issue says 429, but I think 503 (returned in this PR) is more accurate. This error is "the server is overloaded" rather than "you have sent too many requests"._


tests/connections.py
```diff
@@ -127,6 +147,13 @@ def create_connections_until_exhaustion(target):
         to_create = max_fds - num_fds + 1
         num_fds = create_connections_until_exhaustion(to_create)
 
+        try:
+            create_connections_until_exhaustion(to_create, True)
+        except AllConnectionsCreatedException as e:
+            # This is fine! The soft cap means this test no longer causes reaches the hard cap.
```
jumaffre, line 153:
```suggestion
            # This is fine! The soft cap means this test no longer reaches the hard cap.
```

src/enclave/rpc_sessions.h
```diff
@@ -75,26 +114,41 @@ namespace enclave
...
+          max_open_sessions_hard,
           id);
 
         RINGBUFFER_WRITE_MESSAGE(
           tls::tls_stop, to_host, id, std::string("Session refused"));
-        return;
       }
+      else if (sessions.size() >= max_open_sessions_soft)
+      {
+        LOG_INFO_FMT(
```
jumaffre, line 131: This seems rather verbose?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - The curve-id selected for the identity of joining nodes no longer needs to match that of the network (#2525).
+- The per-node session cap behaviour has changed. The `--max-open-sessions` is now a soft cap on the number of sessions. Beyond this, new sessions will receive a HTTP 503 error immediately after completing the TLS handshake. The existing hard cap (where sessions are closed immediately, before the TLS handshake) is still available, under the argument `--max-open-sessions-hard` (#2583).
```
jumaffre, line 20:
```suggestion
- The per-node session cap behaviour has changed. The `--max-open-sessions` is now a soft cap on the number of sessions. Beyond this, new sessions will receive a HTTP 503 error immediately after completing the TLS handshake. The existing hard cap (where sessions are closed immediately, before the TLS handshake) is still available, under the new argument `--max-open-sessions-hard` (#2583).
```

jumaffre (approved):

achamayou:
@eddyashton doing this only for new sessions seems fine. Shutting down existing sessions to give priority to fresh ones doesn't sound great, a single malicious actor could easily DoS the service, whereas right now at least existing connections would continue.

src/enclave/rpc_sessions.h
```diff
@@ -38,6 +39,43 @@ namespace enclave
...
+
+        if (get_status() == Status::ready)
+        {
+          // Send HTTP response describing soft session limit
+          auto http_response = http::Response(HTTP_STATUS_SERVICE_UNAVAILABLE);
+          http_response.set_header(
+            http::headers::CONTENT_TYPE, http::headervalues::contenttype::TEXT);
+          const auto response_body = fmt::format(
+            "Service is currently busy and unable to serve new connections");
+          http_response.set_body(
```
achamayou, line 64: Seems like we could and perhaps even should have a set_body(const std::string&)?

src/enclave/rpc_sessions.h
eddyashton: I have this at `INFO` to match the existing logging when the cap is reached. I think it shouldn't happen under normal operation, and whenever it _does_ happen the operator should be informed. Especially with the other change from #2521, where the operator is likely to be using a separate interface and not seeing this cap, its important that they have some signal that the service is overloaded. Maybe with #2522 they have another way to find this, but I think the log is still valuable.

achamayou (approved):

src/enclave/rpc_sessions.h
eddyashton: I don't think this is worth adding. Accessing the raw `http::Response` like this is pretty rare, and its interface surface is currently pretty small. I don't think this conversion is too onerous here, or repeated elsewhere.

PR #2584, opened 2021-05-12
Rename value returned by `/node/version` endpoint

jumaffre:
Leftover from https://github.com/microsoft/CCF/pull/2582#discussion_r630312522 which wasn't included in that PR.

### DIFF
--- a/doc/schemas/node_openapi.json
+++ b/doc/schemas/node_openapi.json
@@ -235,12 +235,12 @@
       },
       "GetVersion__Out": {
         "properties": {
-          "version": {
+          "ccf_version": {
             "$ref": "#/components/schemas/string"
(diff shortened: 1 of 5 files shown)


eddyashton:
I was thinking this endpoint should simply be called `/node/ccf_version`. Are we ever going to add other fields to it?

jumaffre:
> I was thinking this endpoint should simply be called `/node/ccf_version`. Are we ever going to add other fields to it?

I believe it's likely, e.g. including the underlying enclave SDK version (I couldn't find any exposed Open Enclave version in their headers). Leaving as is a generic `version` gives us that flexibility in the future?

eddyashton (approved):

PR #2586, opened 2021-05-12
Add isolation/reconnection rotation test

achamayou:
Follow up to #2553

### DIFF
--- a/tests/partitions_test.py
+++ b/tests/partitions_test.py
@@ -6,9 +6,9 @@ import infra.e2e_args
 import infra.partitions
 import infra.logging_app as app
 import suite.test_requirements as reqs
-import reconfiguration
 from ccf.tx_id import TxID
 import time
+from infra.checker import check_can_progress
 
 
 @reqs.description("Invalid partitions are not allowed")
@@ -113,7 +113,7 @@ def test_isolate_and_reconnect_primary(network, args):
         new_primary, _ = network.wait_for_new_primary(
             primary, nodes=backups, timeout_multiplier=6
         )
-        new_tx = reconfiguration.check_can_progress(new_primary)
+        new_tx = check_can_progress(new_primary)
 
     # Check reconnected former primary has caught up
     with primary.client() as c:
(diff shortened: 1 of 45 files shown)


tests/partitions_test.py
```diff
@@ -103,6 +106,33 @@ def test_isolate_primary(network, args):
...
+def test_isolate_and_reconnect_primary(network, args):
+    primary, backups = network.find_nodes()
+    rules = network.partitioner.isolate_node(primary)
+
+    new_primary, new_view = network.wait_for_new_primary(
+        primary, nodes=backups, timeout_multiplier=6
+    )
+    new_tx = reconfiguration.check_can_progress(new_primary)
+
+    # Drop rules to allow primary to reconnect
```
jumaffre, line 119: It might be better to use the context manager for managing rules. The explicit `drop()` in the previous test was more of an example for the partioning API.

tests/partitions_test.py
```diff
@@ -103,6 +106,30 @@ def test_isolate_primary(network, args):
...
 
 
+@reqs.description("Isolate and reconnect primary")
+def test_isolate_and_reconnect_primary(network, args):
+    primary, backups = network.find_nodes()
+    with network.partitioner.partition(backups):
+        new_primary, _ = network.wait_for_new_primary(
+            primary, nodes=backups, timeout_multiplier=6
+        )
+        new_tx = reconfiguration.check_can_progress(new_primary)
```
jumaffre, line 116: I'm not a big fan of test including other tests. If the functionality is generic enough, it should probably be moved to `infra/`? Here, I suggest that we use the `wait_for_commit()` method in `CCFClient` to achieve the same result.

jumaffre (approved):

tests/partitions_test.py
achamayou: `check_can_progress` isn't a test, it's just a utility method, I'll move it to infra (it's also used in rotation and other places). `wait_for_commit()` doesn't submit transactions,

eddyashton (approved):

Issue #2592, opened 2021-05-13
Signed request verification uses unescaped URLs


achamayou:
`construct_raw_signed_string()` incorrectly takes and uses an escaped query, where it should be using the raw url instead.

achamayou:
After discussion, we have decided that the signature verification should not make an attempt to strip a fragment if it's present, and to use the URL as sent. Sending fragments to the server is unexpected and a breach of RFCs, it is not possible to effectively second guess how such a client expects signatures to work.

PR #2594, opened 2021-05-14
Fix signed requests with escaped urls

achamayou:
Fix for #2592. The signature verification code previously re-assembled decoded path and query string to construct the canonical payload. This change makes it use the original URL instead.

- [x] Confirm we need to drop fragment and drop it: we don't and we won't, details on ticket.
- [x] CHANGELOG

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -765,15 +765,38 @@ def test_forwarding_frontends(network, args):
         check(c.get(f"/app/log/private?id={log_id}"), result={"msg": msg})
 
         if args.package == "liblogging":
-            for query_string in samples:
-                r = c.get(f"/app/log/request_query?{query_string}")
-                assert r.body.text() == query_string, r.body.text()
+            samples = [
+                {"this": "that"},
+                {"this": "that", "other": "with spaces"},
+                {"this with spaces": "with spaces"},
+                {
+                    "arg": 'This has many@many many \\% " AWKWARD :;-=?!& characters %20%20'
+                },
+            ]
+            for query in samples:
+                unescaped_query = "&".join([f"{k}={v}" for k, v in query.items()])
+                query_to_send = unescaped_query
+                if os.getenv("CURL_CLIENT"):
+                    query_to_send = urllib.parse.urlencode(query)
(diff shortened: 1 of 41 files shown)


tests/e2e_logging.py
```diff
@@ -746,6 +746,13 @@ def test_forwarding_frontends(network, args):
         ack = network.consortium.get_any_active_member().ack(backup)
         check_commit(ack)
 
+    samples = [
+        "this=that",
+        "this=that&other=with spaces",
+        "this with spaces=with spaces",
+        'arg=This has many@many many \\% " AWKWARD :;-=?!& characters %20%20',
+    ]
+
```
eddyashton, line 755:
```suggestion
```
Believe this is now defined on 768.

tests/e2e_logging.py
```diff
@@ -793,6 +799,53 @@ def test_forwarding_frontends(network, args):
...
+@reqs.description("Testing signed queries with escaped queries")
+@reqs.at_least_n_nodes(2)
+def test_signed_escapes(network, args):
+    samples = [
+        "this=that",
+        "this=that&other=with spaces",
+        "this with spaces=with spaces",
+        'arg=This has many@many many \\% " AWKWARD :;-=?!& characters %20%20',
+    ]
+
```
eddyashton, line 811:
```suggestion
```
Now defined on 815.

tests/e2e_logging.py
```diff
@@ -793,6 +799,53 @@ def test_forwarding_frontends(network, args):
     return network
 
 
+@reqs.description("Testing signed queries with escaped queries")
+@reqs.at_least_n_nodes(2)
+def test_signed_escapes(network, args):
```
eddyashton, line 804: This duplicates a lot of `test_forwarding_frontends`, which could be factored out into something like `def escaped_query_tests(client, url_base)`.

jumaffre (approved):

eddyashton (approved):

tests/e2e_logging.py
achamayou: Yes, good point.

PR #2595, opened 2021-05-14
Compatibility end-to-end tests

### LINKED ISSUE #2238: Continuous testing of compatibility in the CI
We should test for compatibility of the latest commit on the `main` branch against the latest major version as part of our CI pipeline. To do so, a new end-to-test test should at least:

1. Pull and install the latest major release from GitHub. 
2. Run a short-lived service with that install (we'd have to rebuild the current version of our sample app against that)
3. Perform a code upgrade with nodes built from the latest commit on `main`, possibly killing the current (old) primary node until a new node is elected as primary.
4. Run the full suite of end-to-end tests on this hybrid network.
5. Retire all old nodes.

jumaffre:
Resolves #2238 

~**Still WIP but feedback very welcome at this stage!**~ Now ready for review!

This PR adds a new end-to-end test to check that we don't break the compatibility between CCF versions that we've outlined [here](https://microsoft.github.io/CCF/main/overview/release_policy.html#operations-compatibility) and that are (ticked boxes are currently covered by the new tests):

1. [x] Patches are compatible: nodes built from the same major release but different patches releases can run within the same service (backward and forward compatibility between these): **code upgrade from the latest available LTS to the version of the local checkout, including primary rotation & historical queries**.
2. [x] The last patch in a major release (N.0.final) is compatible with the first patch of the next major release (N+1.0.0): nodes built against both versions can be run within the same service: **when on a release branch, code upgrade from the latest LTS to the first release of the next LTS**
3. [x] Open-ended ledger backward compatibility: a ledger produced by version N.0.x can be read by all versions > N.0.x: **successive recoveries from the very first LTS to the next LTS, etc., till the local checkout and (daily build only)**.
4. [x] Forward compatibility of the ledger across patches: a ledger produced by version N.0.x can be read by all N.0.* patches in the same major release: **same as 2., covered by historical queries**.
5. [x] Snapshots are compatible across incremental major releases (going from N.0.x to N+1.0.x): **same as 3., but recovering from snapshots this time**.

Interesting things to note:
- Had to refactor how the Python infra start new nodes (i.e. I broke down the existing `create_and..._new_node()` methods so that we can pass specific args to the node). I think this is an improvement as the orchestration API now hopefully reflects more what operators/members should do.
- The `infra.network.Network` and `infra.node.Node` need to know which major version they are as new CLI arguments will be rejected by older nodes.
- We rely on `PyGithub` to interface with our GitHub repo. Note that the GitHub API has some rate limit, which may cause the test to fail if run repeatedly on the same VM. The limit is especially low for unauthenticated clients but this shouldn't cause any issues when running in the CI? (see https://docs.github.com/en/rest/reference/rate-limit). The `GITHUB_COMPATIBILITY_TOKEN` envvar can be used to pass a valid token there.


tests/lts_compatibility.py
```diff
@@ -0,0 +1,253 @@
...
+
+def issue_activity_on_live_service(network, args):
+
+    log_capture = []
+    network.txs.issue(
+        network, number_txs=args.snapshot_tx_interval * 2, log_capture=log_capture
+    )
+
+
+def run_live_compatibility_since_last(args):
```
jumaffre, line 30: Example of how to break this test: add a new field to Raft's `AppendEntries`.

tests/lts_compatibility.py
```diff
@@ -0,0 +1,253 @@
...
+        )
+        network.consortium.retire_code(primary, old_code_id)
+        for node in lts_nodes:
+            network.retire_node(new_primary, node)
+            node.stop()
+
+        issue_activity_on_live_service(network, args)
+
+
+def run_ledger_compatibility_since_first(args, use_snapshot):
```
jumaffre, line 143: TODO: 
- [ ] Test JWTs.
- [ ] Since KV table schemas are only used when a table is read, a successful recovery doesn't mean that the new version can handle the old schema. So after each recovery, we should issue requests that read those tables (which may be tricky for some).

tests/lts_compatibility.py
```diff
@@ -0,0 +1,253 @@
...
+    The recovery process uses snapshot is `use_snapshot` is True. Otherwise, the
+    entire historical ledger is used.
+    """
+    repo = infra.gh_helper.Repository()
+    lts_releases = repo.get_lts_releases()
+
+    # TODO: Remove fakeness
+    lts_releases_fake = {}
+    lts_releases_fake["1.0"] = lts_releases["release/1.x"]
+    lts_releases_fake["2.0"] = lts_releases["release/1.x"]
```
jumaffre, line 157: To be removed. This pretends that we have more than one LTS branch so far.

tests/lts_compatibility.py
```diff
@@ -0,0 +1,253 @@
...
+            # version, lts_install_path = repo.install_release(lts_release)
+            # TODO: Fix
+            version = "1.0.0"
+            lts_install_path = "/data/git/CCF/build/ccf_install_1.0.0/opt/ccf"
+            binary_dir = os.path.join(lts_install_path, "bin")
+            library_dir = os.path.join(lts_install_path, "lib")
+            major_version = Version(version).release[0]
+            # Explictly test the logging app as it was packaged in the very first LTS
+            # TODO: Is this the right approach?
+            args.js_app_bundle = os.path.join(lts_install_path, "samples/logging/js")
```
jumaffre, line 176: Do we want to test the latest version of the JS logging app here, or the one that was shipped with the very first LTS? I don't think we make claims about compatibility for JS apps yet, but a JS app built for `1.x` should just work (tm) on `N.x`?

tests/lts_compatibility.py
```diff
@@ -0,0 +1,253 @@
...
+    # JS generic is the only app included in CCF install
+    args.package = "libjs_generic"
+    args.nodes = infra.e2e_args.max_nodes(args, f=0)
+
+    # Hardcoded because host only accepts from info on release builds
+    args.host_log_level = "info"
+
+    run_live_compatibility_since_last(args)
+    run_ledger_compatibility_since_first(args, use_snapshot=False)
+    run_ledger_compatibility_since_first(args, use_snapshot=True)
```
jumaffre, line 253: Running this test twice with and without snapshots seems wasteful. We could randomly alternate between recovery from snapshot/without snapshot in the test instead?

tests/lts_compatibility.py
achamayou: I agree that testing the one that shipped with the first LTS is the way to go, it's more likely to be run against a recent release than the latest sample against an LTS. The latest sample may legitimately make use of new features that aren't in the LTS.

tests/lts_compatibility.py
achamayou: It depends. In the main Daily, yes. On a release branch when we run once before release, we probably want to be exhaustive if we can.

jumaffre: Agreed. I'll use our cmake's `LONG_TESTS` flag to indicate this.

.gitignore
```diff
@@ -29,4 +29,5 @@ node_modules
 package-lock.json
 .cache
 include/ccf/version.h
-python/version.py
\ No newline at end of file
+python/version.py
+tests/compatibility_report.json
```
achamayou, line 33: Could that be made to go under build/workspace, like other test output?

tests/lts_compatibility.py
achamayou: Only for built-in functionality though, not for application tables (eg. logging sample).

.gitignore
jumaffre: Yep, good idea!

tests/lts_compatibility.py
```diff
@@ -0,0 +1,368 @@
...
+            for _ in range(0, len(network.get_joined_nodes()) + 1):
+                new_node = network.create_node(
+                    "local://localhost",
+                    binary_dir=to_binary_dir,
+                    library_dir=to_library_dir,
+                    version=to_major_version,
+                )
+                network.join_node(new_node, args.package, args)
+                network.trust_node(new_node, args)
+                from_snapshot = not from_snapshot
```
eddyashton, line 105: This is unused.

jumaffre: Good spot!

achamayou (approved):

PR #2596, opened 2021-05-14
Add session metrics endpoint

### LINKED ISSUE #2522: Add active and high watermark session counts to metrics endpoint

eddyashton:
Resolves #2522.

I opted to put this under a new `/node/metrics` rather than `/*/api/metrics`, because it is exclusively node-local. `/api/metrics` is as well, but due to an implementation quirk needs to exposed separately on each frontend (each counts only its own endpoints). Its an API minor version bump in any case (our first since #2589), and this way it only affects the `/node` API.

### DIFF
--- a/src/enclave/rpc_sessions.h
+++ b/src/enclave/rpc_sessions.h
@@ -33,6 +33,7 @@ namespace enclave
 
     SpinLock lock;
     std::unordered_map<size_t, std::shared_ptr<Endpoint>> sessions;
+    size_t sessions_peak;
 
     // Upper half of sessions range is reserved for those originating from
     // the enclave via create_client().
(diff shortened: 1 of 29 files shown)


src/enclave/rpc_sessions.h
```diff
@@ -191,6 +205,9 @@ namespace enclave
       // this type of session to be rare and want it to succeed even when we are
       // busy.
       sessions.insert(std::make_pair(id, session));
+
+      sessions_peak = std::max(sessions_peak, sessions.size());
```
jumaffre, line 209: Should we separate client from server sessions in the stats?

jumaffre (approved):
It might also be interesting to give stats on the number of requests that have been forwarded by this node?

src/enclave/rpc_sessions.h
achamayou: Unless we give them different quotas, it won't make a much difference to an operator, particularly since they have little ability to influence client sessions.

achamayou (approved):
This definitely deserves a CHANGELOG entry in my opinion.

eddyashton:
> It might also be interesting to give stats on the number of requests that have been forwarded by this node?

That's a good idea, but could/should probably be done in the existing `/api/metrics`. We count failed and succeeded requests, and in the same code path we sometimes decide to forward so could easily count those as well.

src/enclave/rpc_sessions.h
jumaffre: Since client session are not affected by the session limits, I think we shouldn't count them here? Otherwise, we could technically report a number of open session that's actually over the limit.

eddyashton: So our caps count both kinds of session because both affect our fds, and could eventually cause the node to crash. If we ever create several hundred client sessions at the same time. Although the operator doesn't currently have any dials to turn to _stop_ the node creating client sessions, I think they want to see the same warning signal in this case ("your active sessions are at or near the cap, so users are going to get errors"). Separating these counts is an extra complication, and (I think) hides the information the operator cares about in the common case.

In practice, there's only ever 1 client session at a time currently (when and if we add more, we can work out how _they're_ capped and maybe how they're reported here), and I think its fine to (correctly) report that the open sessions are over the given limit if this happens. Its an artifact of the fact we don't apply the limit to client sessions, and there's no motivation to revisit that here.

jumaffre: @eddyashton That makes sense to me, and as you said, even with the JWT refresh job, the number of clients is only ever going to be one, at least as of 1.x.

PR #2598, opened 2021-05-17
AFT: add skip view

### LINKED ISSUE #2590: AFT: add skip-view to bft view-changes
When a view-change occurs the upon receiving a new-view message the new primary will know if it has the most up to date state or not. If it does not we can bypass needing to performing catch-up by letting that new primary surrender its primariness to the next replica.

- [ ] the proposed primary needs to work out if it wants to skip being the primary for a view-change message that it was sent
- [ ] the proposed primary sends a skip-view message to any replica with sends it a view-change message which has a prepared seqno higher than its own
- [ ] upon receiving a valid skip-view message in response to a view-change message the replica will attempt to view-change to the next view

ashamis:
resolves #2590

Look at the issue for details

### DIFF
--- a/src/consensus/aft/impl/view_change_tracker.h
+++ b/src/consensus/aft/impl/view_change_tracker.h
@@ -66,6 +66,23 @@ namespace aft
       last_view_change_sent = view;
     }
 
+    void received_skip_view(const SkipViewMsg& r)
+    {
+      if (last_view_change_sent != r.view)
+      {
+        LOG_FAIL_FMT(
+          "Received skip view message for not the latest view, "
+          "last_view_change_sent:{}, r.view:{}",
+          last_view_change_sent,
+          r.view);
+        return;
+      }
+
+      time_previous_view_change_increment = std::min(
+        time_previous_view_change_increment - time_between_attempts,
+        std::chrono::milliseconds(0));
+    }
+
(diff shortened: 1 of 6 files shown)


src/node/progress_tracker.h
```diff
@@ -494,17 +494,34 @@ namespace ccf
       return std::make_tuple(std::move(m), highest_prepared_level.seqno);
     }
 
-    bool apply_view_change_message(
+    enum class ApplyViewChangeMessageResult
+    {
+      OK,
```
ashamis, line 499: start at 1

src/consensus/aft/impl/view_change_tracker.h
```diff
@@ -66,6 +66,23 @@ namespace aft
...
+      {
+        LOG_FAIL_FMT(
+          "Received skip view message for not the latest view, "
+          "last_view_change_sent:{}, r.view:{}",
+          last_view_change_sent,
+          r.view);
+        return;
+      }
+
+      time_previous_view_change_increment = std::min(
```
ashamis, line 81: just set this to 0

ashamis: add comment

achamayou (approved):

PR #2599, opened 2021-05-17
Add `kv::Value` and `kv::Set`

### LINKED ISSUE #274: CCF's kv should expose Values as well as Maps

eddyashton:
Resolves #274.

Here's an implementation of a feature that's been on my mind for a while, adding `kv::Value` and `kv::Set` alongside `kv::Map`. 

There's a comment on #274 suggesting this could be done in user-space. This version doesn't actually need to be in `kv` - nothing in `kv` depends on them. But it does use internal `kv` details, and I think we should provide a standard implementation. This makes them look like first-class citizens at the app level (`tx.rw(my_set)` just like `tx.rw(my_map)`, whereas I believe anything that doesn't use internal `kv` details would need to look like `MySet(tx, "my_set_name")`). This doesn't actually require any changes to the existing `kv` code - the only changes in existing `kv` headers are renames, to make it clear they're actually abstract `Handle`s and don't need to represent `Map`s at all.

Opening as a draft for early input. There's still several things to do before this is ready for merging:
- [ ] Explore possible alternative APIs
- - [x] At the very least, find sensible names for the method calls that aren't just direct copies of the methods on `kv::Map`, and evaluate the names `Unit` and `Value`
- - [ ] At the extreme other end, see if we can make `kv::Value` act more like `std::optional` (in particular adding some assignment operators - almost certainly a dangerous idea)
- - [x] Also want to consider making the unit-type serialisation customisable, since all other serialisation choices are. This may make it easier to upgrade from a `Value` to a `Map` (if it was at 0 rather than null, its easier to decide you actually have entries at 1 and 2... but is that worthwhile, or are you just as likely to want an entry at "old" and "new"?)
- [x] Use these new types for our current value tables (`Signatures`, ...)
- [x] Add doxygen for the new types
- [ ] Describe new types in rst docs
- [x] CHANGELOG

### DIFF
--- a/src/kv/unit.h
+++ b/src/kv/unit.h
@@ -8,7 +8,7 @@ namespace kv
 {
   // A single-valued type, used as a utility type to convert kv::Maps to
   // kv::Values and kv::Sets. Specifically, these are implemented as wrappers so
-  // that kv::Value<T> is implemented as kv::Map<Unit, T>, and kv::Set<T> is
+  // that kv::Value<T> is essentially kv::Map<Unit, T>, and kv::Set<T> is
   // kv::Map<T, Unit>.
   struct Unit
   {
--- a/src/kv/unit.h
+++ b/src/kv/unit.h
@@ -12,14 +12,15 @@ namespace kv
   // kv::Map<T, Unit>.
   struct Unit
   {
-    static kv::serialisers::SerialisedEntry get()
+    static const kv::serialisers::SerialisedEntry& get()
     {
       // TODO: Should this be a 0? Our existing mono-valued tables had a single
       // value at key 0, we could remain ledger-compatible if we produce the
       // same value here. But that only works where we can guess the
       // serialisation format of that 0, and we're stuck with that inefficiency
       // forever.
-      return {};
+      static kv::serialisers::SerialisedEntry the_value = {};
+      return the_value;
     }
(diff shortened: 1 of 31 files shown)


src/kv/unit.h
```diff
@@ -6,21 +6,31 @@
...
+
+  // This is the default UnitCreator, returning 8 null bytes for compatibility
+  // with old ledgers (where Values were previously Maps with a single entry
+  // at key 0, serialised as a uint64_t)
+  struct UnitCreator
+  {
+    static kv::serialisers::SerialisedEntry get()
+    {
+      kv::serialisers::SerialisedEntry e;
+      e.assign(8, 0u);
```
achamayou, line 24:
```suggestion
      e.assign(sizeof(uint64_t), 0u);
```

Possibly?

achamayou (approved):

PR #2600, opened 2021-05-18
Jwt infra refactor

jumaffre:
Part of #2595 

JWTs make use of core service tables that should be covered by #2595. Our Python test infra only tested JWT features in limited scenarios. This PR refactors the JWT testing features in a new `infra.jwt_issuer` module that can be used to create a JWT issuer, start the corresponding OpenID server for JWKS auto-refresh, etc. 

JWTs are now the user authentication mechanism for the test suite, and the new `test_refresh_jwt_issuer` tests that auto-refresh works in the test suite.

### DIFF
--- a/tests/infra/jwt.py
+++ b/tests/infra/jwt.py
@@ -3,7 +3,6 @@
 import infra.crypto
 import base64
 from http.server import HTTPServer, BaseHTTPRequestHandler
-from socketserver import ThreadingMixIn
 from http import HTTPStatus
 import ssl
 import threading
@@ -185,7 +184,7 @@ class JwtIssuer:
         server.set_jwks(self.create_jwks(kid))
         server.restart()
 
-    def issue_jwt(self, kid=TEST_JWT_KID, claims={}):
+    def issue_jwt(self, kid=TEST_JWT_KID, claims=None):
         return infra.crypto.create_jwt(claims, self.key_priv_pem, kid)
 
     def wait_for_refresh(self, network, kid=TEST_JWT_KID, timeout=3):
--- a/tests/infra/jwt.py
+++ b/tests/infra/jwt.py
@@ -143,9 +143,7 @@ class JwtIssuer:
         ).decode("ascii")
         return {"keys": [{"kty": "RSA", "kid": kid, "x5c": [der_b64]}]}
 
-    def register_test_kid(
-        self, network, kid=TEST_JWT_KID, ca_bundle_name=TEST_CA_BUNDLE_NAME
-    ):
+    def register(self, network, kid=TEST_JWT_KID, ca_bundle_name=TEST_CA_BUNDLE_NAME):
         primary, _ = network.find_primary()
 
         if self.auto_refresh:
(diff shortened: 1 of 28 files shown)


tests/infra/jwt.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def issue_jwt(self, kid=TEST_JWT_KID, claims={}):
+        return infra.crypto.create_jwt(claims, self.key_priv_pem, kid)
+
+    def wait_for_refresh(self, network, kid=TEST_JWT_KID, timeout=3):
+        time.sleep(timeout)
+        return
+
+        # TODO: Fix!
+        LOG.info(f"Waiting {timeout}s for JWT key refresh")
```
jumaffre, line 196: This currently fails with a ledger root mismatch error. Investigating now, but the rest of the PR is ready for review.

tests/infra/jwt.py
```diff
@@ -0,0 +1,208 @@
...
+        self.port = port
+        self.jwks = jwks
+        self.tls_key_pem = tls_key_pem
+        self.tls_cert_pem = tls_cert_pem
+        self.self_ = self
+        self.bind_port = None
+        self.started = False
+        self.start(self.port)
+
+    def start(self, port):
```
letmaik, line 56: What's the difference between port and bind_port?

jumaffre: To avoid port collisions, we now ask to bind to `0` by default. Once this is done, we get the port the server actually bound to (`bind_port`) and use that to register the issuer URL to CCF.

tests/jwt_test.py
```diff
@@ -4,50 +4,48 @@
...
-    der_b64 = base64.b64encode(
-        infra.crypto.cert_pem_to_der(cert_pem)
-        if not test_invalid_is_key
-        else infra.crypto.pub_key_pem_to_der(cert_pem)
-    ).decode("ascii")
-    return {"keys": [{"kty": "RSA", "kid": kid, "x5c": [der_b64]}]}
+@reqs.description("Refresh JWT issuer")
+def test_refresh_jwt_issuer(network, args):
+    assert network.jwt_issuer.server, "JWT server is not started"
+    jwt_server = network.jwt_issuer.server
```
letmaik, line 23: I'm sure there's a reason, but why is restarting the server necessary? Shouldn't the job of `refresh_keys` also be to refresh the jwks?

jumaffre: I only did this this to avoid race conditions as the OpenID server listens on a separate thread, and refreshing the jwks while the server is serving those may (very unlikely but still) cause CI instabilities.

letmaik: Global Interpreter Lock :)

achamayou (approved):

tests/infra/jwt_issuer.py
```diff
@@ -0,0 +1,190 @@
...
+import ssl
+import threading
+from contextlib import AbstractContextManager
+import tempfile
+import json
+import time
+from loguru import logger as LOG
+
+
+def make_authn_bearer_header(jwt):
```
eddyashton, line 16: This should be `authz`, not `authn`, since its setting the `authorization` header? Or perhaps just `make_bearer_header`.

tests/infra/network.py
```diff
@@ -990,7 +996,7 @@ def _get_ledger_public_view_at(self, node, call, seqno, timeout):
                 return call(seqno)
             except Exception:
                 self.consortium.create_and_withdraw_large_proposal(node)
-
+                time.sleep(0.1)
```
eddyashton, line 999: What's this sleep doing?

eddyashton (approved):

jumaffre:
To be merged after https://github.com/microsoft/CCF/pull/2607

tests/infra/jwt_issuer.py
jumaffre: Good point! Changed it `make_bearer_header`.

tests/infra/network.py
jumaffre: Not much. I was just worried of doing an action in a loop without waiting at least a little bit between each iteration.

PR #2607, opened 2021-05-19
Fix for ledger chunking and snapshot out-of-sync

### LINKED ISSUE #2491: Reconfiguration test: ledger files don't match up between nodes

jumaffre:
Resolves #2491 

I'd been tracking down this bug for a while, [without success](https://github.com/microsoft/CCF/pull/2526), but activating JWT auto-refresh for the test suites (see https://github.com/microsoft/CCF/pull/2600) triggers the bug much more consistently.

This is the proposed fix for it. This issue was that, while we record "snapshottable" `seqno`s as we replicate entries (on both primary and backups) based on the value of `snapshot_tx_interval`, nodes would generate snapshots on (global) commit at the `seqno` of that commit. However, commits can happen at any signature `seqno`, and don't necessarily match the "snapshottable" `seqno` which was record on the replication code path. In other words, for a `snapshot_tx_interval` of `10`, we may have recorded `seqnos` 10, 20 and 30 as "snapshottable" `seqno`s (and thus forced new ledger chunks at those) but the global commit would come at 15 and we would create a snapshot at 15, instead of 10. Then, when a new node joins from the snapshot at 15, the ledger chunks it creates would be "chunked" at different intervals than those of the primary node. 

This is very obvious in hindsight, and the snapshotter unit test now covers my original misunderstanding. I'm also very surprised that this hasn't showed up in other tests, but I'm assuming our end-to-end tests are simple enough that a service commits all committable `seqno`s?

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -2567,12 +2567,10 @@ namespace aft
       state->commit_idx = idx;
 
       LOG_DEBUG_FMT("Compacting...");
-      snapshotter->commit(idx);
-      if (consensus_type == ConsensusType::CFT)
-      {
-        // Snapshots are not yet supported with BFT
-        snapshotter->update(idx, replica_state == Leader);
-      }
+      // Snapshots are not yet supported with BFT
+      snapshotter->commit(
+        idx, replica_state == Leader && consensus_type == ConsensusType::CFT);
+
       store->compact(idx);
       ledger->commit(idx);
 
--- a/tests/infra/logging_app.py
+++ b/tests/infra/logging_app.py
@@ -203,11 +203,6 @@ class LoggingTxs:
                             raise ValueError(
                                 f"Response with status {rep.status_code} is missing 'retry-after' header"
                             )
-                        sleep_time = 0.5
-                        LOG.info(
-                            f"Sleeping for {sleep_time}s waiting for historical query processing..."
-                        )
-                        time.sleep(sleep_time)
                     elif rep.status_code == http.HTTPStatus.NO_CONTENT:
                         raise ValueError(
                             f"Historical query response claims there was no write to {idx} at {view}.{seqno}"
--- a/tests/infra/logging_app.py
+++ b/tests/infra/logging_app.py
@@ -244,11 +244,6 @@ class LoggingTxs:
                         raise ValueError(
                             f"Response with status {rep.status_code} is missing 'retry-after' header"
                         )
-                    sleep_time = 0.5
-                    LOG.info(
-                        f"Sleeping for {sleep_time}s waiting for historical query processing..."
-                    )
-                    time.sleep(sleep_time)
                 elif rep.status_code == http.HTTPStatus.NO_CONTENT:
                     raise ValueError(
                         f"Historical query response claims there was no write to {idx} at {view}.{seqno}"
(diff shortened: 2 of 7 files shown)


tests/infra/logging_app.py
```diff
@@ -203,11 +203,6 @@ def _verify_tx(
                             raise ValueError(
                                 f"Response with status {rep.status_code} is missing 'retry-after' header"
                             )
-                        sleep_time = 0.5
```
achamayou, line 206: Why is this removed?

src/consensus/aft/raft.h
```diff
@@ -2567,12 +2567,10 @@ namespace aft
       state->commit_idx = idx;
 
       LOG_DEBUG_FMT("Compacting...");
-      snapshotter->commit(idx);
-      if (consensus_type == ConsensusType::CFT)
-      {
-        // Snapshots are not yet supported with BFT
-        snapshotter->update(idx, replica_state == Leader);
-      }
+      // Snapshots are not yet supported with BFT
```
achamayou, line 2570: While we don't support resuming from them in BFT, it seems to me that there should be no harm in emitting them?

tests/infra/logging_app.py
jumaffre: There's already another sleep (line `219`) in this loop. While clients should pause for `retry-after` value returned by the historical query endpoint, in the interest of time, I suggest our e2e tests shouldn't actually wait for that long as it's likely the historical queries will be processed quickly enough?

src/consensus/aft/raft.h
jumaffre: Yes, I was wondering that too and gave it a try but this currently doesn't work as the snapshot evidence generation logic doesn't go through the frontend like BFT would expect (instead, it creates a local `kv::Tx`, records the evidence and commits it "manually"). Is creating a new endpoint for this what we want? My assumption was that all nodes should generate the same snapshot in BFT, and record identical evidences in the store when they do so.

achamayou: If this involved creating an endpoint, let's not do it for now.

achamayou (approved):

achamayou (approved):

PR #2613, opened 2021-05-21
Make join robust to primary retirement

### LINKED ISSUE #2610: Join protocol is not robust to primary retiring after Pending but before Trusted
This one of the problems highlighted by #2606, the following sequence fails:

1. Joinee sends initial join RPC to Primary
2. Joinee receives Pending status
3. Joinee polls Primary
4. Transaction causes Joinee's state to advance to Trusted in KV and Primary to become Retired
5. Joinee continues to poll ex-Primary
6. ex-Primary fails to forward join RPC because node-to-node channel to new primary may not exist

The ex-Primary does not in fact need to forward the RPC in this case: it is a pure read. The secrets sequence it will return may be somewhat outdated at this point, but the new primary will have use the on-ledger distribution mechanism if a rekey has taken place.

As proposed by @eddyashton, the join handler can be modified to be NeverForward to solve this issue: when the node is primary it can successfully execute both write and read variants. When it isn't, it can execute the read variant and send a HTTP redirect if possible in the write case.

This should be testable by passing a combined Primary Retire + Trust Node proposal.

achamayou:
Fix #2610 

This change modifies the join endpoint to:

1. Never forward, but...
2. Execute locally when possible (`add_node` when primary, return Trusted status and secrets when applicable regardless of consensus role)
3. Redirect to the new primary otherwise

This typically wouldn't have been a problem under the old governance, which did not allow multiple actions in the same proposal. It is one of (several) problems spotted by #2606.

### DIFF
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -500,7 +500,7 @@ namespace ccf
 
           if (status != HTTP_STATUS_OK)
           {
-            const auto& location = headers.find("location");
+            const auto& location = headers.find(http::headers::LOCATION);
             if (
               status == HTTP_STATUS_PERMANENT_REDIRECT &&
               location != headers.end())
@@ -508,7 +508,7 @@ namespace ccf
               const auto& url = http::parse_url_full(location->second);
               config.joining.target_host = url.host;
               config.joining.target_port = url.port;
-              LOG_FAIL_FMT("Target node redirected to {}", location->second);
+              LOG_INFO_FMT("Target node redirected to {}", location->second);
             }
             else
             {
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -489,7 +489,9 @@ namespace ccf
         config.joining.target_host,
         config.joining.target_port,
         [this](
-          http_status status, http::HeaderMap&&, std::vector<uint8_t>&& data) {
+          http_status status,
+          http::HeaderMap&& headers,
+          std::vector<uint8_t>&& data) {
           std::lock_guard<SpinLock> guard(lock);
           if (!sm.check(State::pending))
           {
@@ -498,13 +500,26 @@ namespace ccf
 
           if (status != HTTP_STATUS_OK)
           {
-            LOG_FAIL_FMT(
-              "An error occurred while joining the network: {} {}{}",
-              status,
-              http_status_str(status),
-              data.empty() ?
-                "" :
-                fmt::format("  '{}'", std::string(data.begin(), data.end())));
+            const auto& location = headers.find("location");
+            if (
+              status == HTTP_STATUS_PERMANENT_REDIRECT &&
+              location != headers.end())
+            {
+              const auto& url = http::parse_url_full(location->second);
+              config.joining.target_host = url.host;
+              config.joining.target_port = url.port;
+              LOG_FAIL_FMT("Target node redirected to {}", location->second);
+            }
+            else
+            {
+              LOG_FAIL_FMT(
+                "An error occurred while joining the network: {} {}{}",
+                status,
+                http_status_str(status),
+                data.empty() ?
+                  "" :
+                  fmt::format("  '{}'", std::string(data.begin(), data.end())));
+            }
             return false;
           }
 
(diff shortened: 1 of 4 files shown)


src/node/rpc/node_frontend.h
```diff
@@ -289,6 +289,35 @@ namespace ccf
             return make_success(rep);
           }
 
+          if (!this->context.get_node_state().is_primary())
```
jumaffre, line 292: It seems to me that we could avoid all this code here if we created a new `endpoints::ForwardingRequired::RedirectOnly` enum that specified that the endpoint would never forward, but would redirect if the primary isn't us. The generic frontend could then handle this logic. Thoughts?

achamayou: Unless I'm missing something, it wouldn't help much because we still want to run much of the logic even if we're not primary, and in cases where we know the node is trusted, respond accordingly. The forwarding policy is accurately "NeverForward", if we were going to make this generic it might take the form of a wrapper, catching a RedirectException and returning the appropriate response.

I'm skeptical about going generic here though, this is a special case and we don't seem to have any other situation where this applicable?

src/node/rpc/node_frontend.h
```diff
@@ -289,6 +289,35 @@ namespace ccf
...
+              if (info)
+              {
+                args.rpc_ctx->set_response_header(
+                  "Location",
+                  fmt::format(
+                    "https://{}:{}/node/join", info->pubhost, info->pubport));
+
+                return make_error(
+                  HTTP_STATUS_PERMANENT_REDIRECT,
+                  ccf::errors::NodeRetired,
```
jumaffre, line 310: This isn't always true, is it? We could get here if the joiner's target node is not/no longer primary, but isn't necessarily retired.

src/node/rpc/node_frontend.h
jumaffre: Fair enough. It's true that the read-only join can be served by any node, even if that node is retired. The code here looked at lot like the generic forwarding/redirect logic in `frontend.h`, hence my original comment.

src/node/rpc/node_frontend.h
achamayou: Yes!

src/node/node_state.h
```diff
@@ -498,13 +500,26 @@ namespace ccf
...
           if (status != HTTP_STATUS_OK)
           {
-            LOG_FAIL_FMT(
-              "An error occurred while joining the network: {} {}{}",
-              status,
-              http_status_str(status),
-              data.empty() ?
-                "" :
-                fmt::format("  '{}'", std::string(data.begin(), data.end())));
+            const auto& location = headers.find("location");
```
jumaffre, line 503:
```suggestion
            const auto& location = headers.find(http::headers::LOCATION);
```

src/node/node_state.h
```diff
@@ -498,13 +500,26 @@ namespace ccf
...
+            if (
+              status == HTTP_STATUS_PERMANENT_REDIRECT &&
+              location != headers.end())
+            {
+              const auto& url = http::parse_url_full(location->second);
+              config.joining.target_host = url.host;
+              config.joining.target_port = url.port;
+              LOG_FAIL_FMT("Target node redirected to {}", location->second);
```
jumaffre, line 511:
```suggestion
              LOG_INFO_FMT("Target node redirected to {}", location->second);
```

tests/reconfiguration.py
```diff
@@ -275,6 +275,46 @@ def test_node_replacement(network, args):
...
+    )
+    network.consortium.vote_using_majority(
+        primary,
+        proposal,
+        {"ballot": "export function vote (proposal, proposer_id) { return true }"},
+        timeout=10,
+    )
+
+    network.wait_for_new_primary(primary)
+    new_node.wait_for_node_to_join(timeout=10)
```
jumaffre, line 311: Can we assert that the node was redirected to the new primary here, or is it not guaranteed if the join happens quickly?

jumaffre (approved):

tests/reconfiguration.py
achamayou: Here we don't think the node should be redirected, on the contrary, we expect the old primary to be able to serve the read-only query and the join to happen that way.

It's the join from follower test that exercises the redirect, now that the RPC isn't forwarded.

Issue #2616, opened 2021-05-24
Partitioning end-to-end has become unstable


jumaffre:
The partitioning end-to-end test has become unstable, with two different errors:
- The old primary is sometimes not able to make progress after its partition is lifted (e.g. https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=25837&view=logs&j=ee9484ec-b476-577b-aa0e-d5ddfcbcd149&t=8e7d12cf-69d2-5704-84b7-b6fd04033d7b&l=841). I managed to get a local repro for this but the logs are unclear: it seems that the old primary doesn't receive append entries from the new primary and triggers a new election, which eventually is won by the third node.
- A libuv error at shutdown (https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=25834&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=cf4e34b3-24ee-5d97-0df9-ba9302d9fe90&l=855), already tracked by #2571. 

I am deactivating this test for now, until these issues are fixed.

jumaffre:
https://github.com/microsoft/CCF/pull/2618 fixes 3 issues that made the partitioning test unstable. 

There still is one outstanding issue with the previous primary not being able to catch up with the new primary. The reason for this is that the outgoing connection cannot be established between the new primary and the old one. This is infrequent yet can be reproduced by running the test about 20 times in a loop. More specifically, the `on_connect` callback isn't fired when the partition is lifted, which means the new primary cannot send AEs to the old one, which then becomes candidate and doesn't catch up. 

I've upgraded libuv to the latest release (1.41.0) but the issue still persists. The `iptables` rules have also been cleaned up at this point and the `TCP` connection on the host is still alive so I'm not sure why the `on_connect` callback isn't triggered. Ideas welcome!

jumaffre:
The issue described above was due to the exponential connection retry for outgoing TCP connections. In some cases, the partition was up for slightly too long and the partition was lifted in between two long connection retries, which exceeded the CFT election timeout (more details in the SYN-SENT section: https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/). The fix for this is in #2618: the timeout is now configurable by the operator (defaults to 1s) so that we reconnect much more aggressively if the connection cannot be established.

jumaffre:
Fixed by #2618

Issue #2571, opened 2021-05-07
Investigate node shutdown exception during network partitioning 


jumaffre:
Testing the partitioning of nodes in #2553 has revealed another libuv-related exception on shutdown (see [CI failure](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=25011&view=logs&j=ee9484ec-b476-577b-aa0e-d5ddfcbcd149&t=8e7d12cf-69d2-5704-84b7-b6fd04033d7b&l=207)):
```
63: 11:05:28.864 | ERROR    | infra.remote:log_errors:86 - Contents of /mnt/build/2/s/build/workspace/partitions_cft_2/err:
63: Stack trace (most recent call last):
63: #23   Object "[0xffffffffffffffff]", at 0xffffffffffffffff, in 
63: #22   Object "./cchost.virtual", at 0x408719, in 
63: #21   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fb9d6a66bf6, in __libc_start_main
63: #20   Object "./cchost.virtual", at 0x423189, in 
63: #19   Object "/usr/lib/x86_64-linux-gnu/libuv.so.1", at 0x7fb9d7ce5ddd, in uv_run
63: #18   Object "/usr/lib/x86_64-linux-gnu/libuv.so.1", at 0x7fb9d7cf05db, in uv__stream_destroy
63: #17   Object "./cchost.virtual", at 0x52cdf7, in 
63: #16   Object "./cchost.virtual", at 0x52ce51, in 
63: #15   Object "./cchost.virtual", at 0x52c29a, in 
63: #14   Object "./cchost.virtual", at 0x5c1974, in 
63: #13   Object "./cchost.virtual", at 0x5c1b0c, in 
63: #12   Object "./cchost.virtual", at 0x5c1b4e, in 
63: #11   Object "./cchost.virtual", at 0x5c1b7c, in 
63: #10   Object "./cchost.virtual", at 0x5c1bd3, in 
63: #9    Object "./cchost.virtual", at 0x5c1c8c, in 
63: #8    Object "./cchost.virtual", at 0x5c1dc9, in 
63: #7    Object "./cchost.virtual", at 0x5c2250, in 
63: #6    Object "./cchost.virtual", at 0x5c230c, in 
63: #5    Object "./cchost.virtual", at 0x4b3e4c, in 
63: #4    Object "./cchost.virtual", at 0x4b3e83, in 
[...]

PR #2618, opened 2021-05-24
Fix for host connection bugs triggered by partitioning test

jumaffre:
Fixes 4 issues triggered by the partition test:
- #2571. More details and related libuv docs (http://docs.libuv.org/en/v1.x/handle.html#c.uv_close): 
> In-progress requests, like uv_connect_t or uv_write_t, are cancelled and have their callbacks called asynchronously with status=UV_ECANCELED.
- Snapshots were enabled, which meant the assumption that a partitioned node would become primary when the partition is lifted wasn't true if a snapshot was generated while the partition was up. 
- Double node connection issue on the host when a follower node would be unable to send responses back to the primary because no node connection was still associated.
- Added a new `--client-connection-timeout-ms` CLI option to `cchost` to specify the TCP timeout for all outgoing TCP connections. This fixes the issue described https://github.com/microsoft/CCF/issues/2616#issuecomment-852253179.

### DIFF
--- a/.azure-pipelines-templates/common.yml
+++ b/.azure-pipelines-templates/common.yml
@@ -6,6 +6,10 @@ jobs:
     ${{ insert }}: ${{ parameters.env }}
 
     steps:
+      - script: |
+          sudo rm -rf build
+        displayName: "Cleanup"
+
       - checkout: self
         clean: true
         fetchDepth: 10000
--- a/.azure-pipelines-templates/test.yml
+++ b/.azure-pipelines-templates/test.yml
@@ -5,10 +5,9 @@ parameters:
   suffix: ""
 
 steps:
-  # TODO: Re-enable before merging. Only commented out to get more timeout on partition tests
   - script: |
       set -ex
-      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test -R tutorial
+      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
     env:
       METRICS_MONGO_CONNECTION: $(METRICS_MONGO_CONNECTION)
       CTEST_TIMEOUT: "${{ parameters.ctest_timeout }}"
@@ -19,7 +18,7 @@ steps:
   - ${{ if ne(parameters.suffix, 'Perf') }}:
       - script: |
           set -ex
-          sudo bash -c "source env/bin/activate && ctest -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test -L partitions --repeat-until-fail 100"
+          sudo bash -c "source env/bin/activate && ctest -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test -L partitions"
         env:
           CTEST_TIMEOUT: "${{ parameters.ctest_timeout }}"
         condition: succeededOrFailed()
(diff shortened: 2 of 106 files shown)


src/host/tcp.h
```diff
@@ -161,6 +175,29 @@ namespace asynchost
       return service;
     }
 
+    void client_bind()
```
jumaffre, line 178: Some minor refactor of the client bind as we didn't re-bind to the client interface in the disconnection case.

.azure-pipelines-templates/test.yml
```diff
@@ -5,37 +5,36 @@ parameters:
   suffix: ""
 
 steps:
+  # TODO: Re-enable before merging. Only commented out to get more timeout on partition tests
```
jumaffre, line 8: Revert this before merging this PR.

src/host/main.cpp
```diff
@@ -298,6 +295,17 @@ int main(int argc, char** argv)
       "defined by this timer interval.")
     ->capture_default_str();
 
+  size_t client_connection_timeout = 1000;
```
achamayou, line 298: 1 second seems very aggressive compared to the default (exponential up to 64s)?

src/host/node_connections.h
```diff
@@ -129,9 +126,21 @@ namespace asynchost
...
       {
+        // It is possible that a peer terminates a connection and opens a new
+        // one but the termination is seen by us _after_ messages on the new
+        // connection are received. We re-associate the connection on the latest
+        // received message, so that oubound messages are routed to the correct,
+        // most up-to-date, incoming connection.
+        auto search = parent.associated.find(n);
+        if (search != parent.associated.end() && search->second.first == id)
+        {
+          // Incoming connection is already associated n
```
achamayou, line 137:
```suggestion
          // Incoming connection is already associated with n
```

src/host/tcp.h
```diff
@@ -327,14 +360,37 @@ namespace asynchost
...
         LOG_FAIL_FMT("uv_tcp_init failed: {}", uv_strerror(rc));
         return false;
       }
 
+      if (is_client)
+      {
+        uv_os_sock_t sock;
+        if ((sock = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP)) == -1)
+        {
+          LOG_FAIL_FMT("socket creation failed: {}", strerror(errno));
```
achamayou, line 376: If we have the address easily available, it'd be really good to add it to this log line.

achamayou (approved):

src/host/main.cpp
jumaffre: As discussed, this is now 2s so we get at least one retry from the OS, while keeping this value below the default raft election timeout.

src/host/tcp.h
jumaffre: It's too early at this point as (the address is specified later, on connect.

.azure-pipelines-templates/common.yml
```diff
@@ -6,6 +6,10 @@ jobs:
     ${{ insert }}: ${{ parameters.env }}
 
     steps:
+      - script: |
+          sudo rm -rf build
```
ashamis, line 10: This comment could be wrong but should the delete be for . ?

ashamis (approved):

.azure-pipelines-templates/common.yml
achamayou: No, that's the git repo and will be 1. untouched by the tests and 2. reset correctly by DevOps, selectively nuking build/ is the right thing to do.

jumaffre: This cleanup step happens before the checkout so deleting `.` would work too but as @achamayou said, the test doesn't touch anything apart from `build/` (`build/workspace/` in fact) so this should be sufficient.

PR #2620, opened 2021-05-24
AFT: Get BFT variant of the rotation tests to work

ashamis:
This change gets the rotation test to work with BFT.

- Need to rollback the current state of the progress tracker to ensure that after the view-change we do not maintain any unnecessary state from the previous view.
- Track view-change messages by seqno and view to ensure that when we check the signatures we use the correct merkle root.
- We were not correctly tracking if we needed evidence from a new primary to saying that it has 2f+1 replicas agreeing it is the primary.
- some minor fixes around checking invariants.

### DIFF
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -30,7 +30,7 @@ if((NOT ${IS_VALID_TARGET}))
 endif()
 
 # Find OpenEnclave package
-find_package(OpenEnclave 0.15 CONFIG REQUIRED)
+find_package(OpenEnclave 0.13 CONFIG REQUIRED)
 # As well as pulling in openenclave:: targets, this sets variables which can be
 # used for our edge cases (eg - for virtual libraries). These do not follow the
 # standard naming patterns, for example use OE_INCLUDEDIR rather than
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -30,7 +30,7 @@ if((NOT ${IS_VALID_TARGET}))
 endif()
 
 # Find OpenEnclave package
-find_package(OpenEnclave 0.16.1 CONFIG REQUIRED)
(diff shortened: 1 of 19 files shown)


achamayou (approved):

cmake/ccf_app.cmake
```diff
@@ -30,7 +30,7 @@ if((NOT ${IS_VALID_TARGET}))
 endif()
 
 # Find OpenEnclave package
-find_package(OpenEnclave 0.16.1 CONFIG REQUIRED)
+find_package(OpenEnclave 0.13 CONFIG REQUIRED)
```
achamayou, line 33: I assume this is in error? We are not rolling back to OE 0.13.

achamayou (changes requested):

src/node/progress_tracker.h
```diff
@@ -635,18 +635,6 @@ namespace ccf
     void rollback(ccf::SeqNo rollback_seqno, ccf::View view)
     {
       std::unique_lock<SpinLock> guard(lock);
-      if (rollback_seqno < highest_commit_level)
```
achamayou: That seemed like a good sanity check, is it gone because it's not in fact correct for a reason that's not obvious to me now, or just because it's redundant with a check somewhere else?

cmake/ccf_app.cmake
ashamis: yes definitely in error. Will rollback.

src/node/progress_tracker.h
ashamis: its wrong here, commit its a local property so only less than f+1 machines may have reached commit. The correct check is in raft.h (same as for CFT) where we do not rollback anything that is globally committed.

cmake/ccf_app.cmake
ashamis: I need to rebuild my machine :(

PR #2621, opened 2021-05-25
Add a '/node/consensus' RPC to facilitate operations/debug

achamayou:
Sample output:

```
43: {'details': {'acks': {'416cb1527799a424b1d88f7608995ff86c7140a7524d47681aae65bc83ff925c': 29,
43:                       '41cedc7034123dd155c1e9dd2f891fd20b440f2dd4d9dda879285cf86334fbc0': 29},
43:              'configs': [{'idx': 28,
43:                           'nodes': {'416cb1527799a424b1d88f7608995ff86c7140a7524d47681aae65bc83ff925c': {'address': '127.99.126.246:37135'},
43:                                     '41cedc7034123dd155c1e9dd2f891fd20b440f2dd4d9dda879285cf86334fbc0': {'address': '127.154.179.176:34905'},
43:                                     'dac56ab51c8242c4609f34575f175ef83e8fd929b876d8307822aeb574b3303f': {'address': '127.227.163.145:39541'}}}],
43:              'state': 'Leader'}}
```

### DIFF
--- a/src/kv/kv_types.h
+++ b/src/kv/kv_types.h
@@ -93,6 +93,11 @@ namespace kv
     Nodes nodes;
   };
 
+  DECLARE_JSON_TYPE(Configuration::NodeInfo);
+  DECLARE_JSON_REQUIRED_FIELDS(Configuration::NodeInfo, hostname, port);
+  DECLARE_JSON_TYPE(Configuration);
+  DECLARE_JSON_REQUIRED_FIELDS(Configuration, idx, nodes);
+
   class ConfigurableConsensus
   {
   public:
@@ -100,6 +105,7 @@ namespace kv
       ccf::SeqNo seqno, const Configuration::Nodes& conf) = 0;
     virtual Configuration::Nodes get_latest_configuration() = 0;
     virtual Configuration::Nodes get_latest_configuration_unsafe() const = 0;
+    virtual std::vector<Configuration> get_active_configurations() = 0;
   };
 
   class ConsensusHook
--- a/src/kv/kv_types.h
+++ b/src/kv/kv_types.h
@@ -93,10 +93,46 @@ namespace kv
     Nodes nodes;
   };
 
-  DECLARE_JSON_TYPE(Configuration::NodeInfo);
-  DECLARE_JSON_REQUIRED_FIELDS(Configuration::NodeInfo, hostname, port);
+  inline void to_json(nlohmann::json& j, const Configuration::NodeInfo& ni)
+  {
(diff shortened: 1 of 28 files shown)


jumaffre (approved):

eddyashton (approved):
This new endpoint is never called by the e2e tests, so will rot quickly. Can we add a few calls before and after elections, to confirm it returns a 200 all the time and has the expected shape?

src/kv/kv_types.h
```diff
@@ -93,13 +93,55 @@ namespace kv
...
+
+  DECLARE_JSON_ENUM(
+    ReplicaState,
+    {{ReplicaState::Leader, "Leader"},
+     {ReplicaState::Follower, "Follower"},
+     {ReplicaState::Candidate, "Candidate"},
+     {ReplicaState::Retired, "Retired"}});
+
+  DECLARE_JSON_TYPE(Configuration);
+  DECLARE_JSON_REQUIRED_FIELDS(Configuration, idx, nodes);
```
eddyashton, line 125: Its not obvious in the HTTP response what this `idx` is - is that the seqno where the configuration was created? Or the highest one it has acked?

achamayou: It's the creation/applicability seqno for the config.

src/kv/kv_types.h
```diff
@@ -93,13 +93,55 @@ namespace kv
...
+
+  struct ConsensusDetails
+  {
+    std::vector<Configuration> configs = {};
+    std::unordered_map<ccf::NodeId, ccf::SeqNo> acks = {};
```
wintersteiger, line 130: Is the idea that only trusted nodes are listed in the configs? For me it would be convenient to add a node state field to indicate whether they are pending/joining/exiting/etc.

wintersteiger (approved):

src/kv/kv_types.h
achamayou: Yes, only trusted nodes in a given config are listed here, because that's all consensus ever deals with. Nodes that aren't trusted yet definitely do not have the ability to verify or decrypt append entries, so talking to them isn't very useful.

Nodes that are retiring are in a config, but not in later configs, so they show up here up until the point where their removal is committed and the config dropped.

src/kv/kv_types.h
achamayou: @eddyashton do you have a suggestion for a better name? I'd be happy to take it.

wintersteiger: `start_idx`, `valid_from`, `as_of`?

Issue #2609, opened 2021-05-21
Upgrade to OpenEnclave 0.16.1


achamayou:
https://github.com/openenclave/openenclave/releases/tag/v0.16.0
https://github.com/openenclave/openenclave/releases/tag/v0.16.1

Note: oe_verify_attestation_certificate_with_evidence() has been deprecated because it has been deemed insufficient for security. Use the new, experimental oe_verify_attestation_certificate_with_evidence_v2() instead to generate a self-signed certificate for use in the TLS handshaking process.

achamayou:
Going to hold off on the oe_verify_attestation_certificate_with_evidence_v2() upgrade and stick with oe_verify_attestation_certificate_with_evidence() until https://github.com/openenclave/openenclave/issues/4048 is clarified.

PR #2622, opened 2021-05-25
Upgrade to OE 0.16.1

achamayou:
#2609 

Looks good on container envs, updated perf env too.


achamayou:
The Threading CI failures are not specific to this change and unfortunately typical on the main branch right now.

eddyashton (approved):

Issue #2612, opened 2021-05-21
Move cchost configuration from CLI/TOML to JSON configuration


achamayou:
Implementing #2521, it has become increasingly apparent that we are stretching CLI11's abilities. For each interface, that we think would ideally be named, we need to specify a local address, a public address, a soft quota and a hard quota.

While this is all possible with comma separated values, like member info, it's not particularly readable. We suspect there may soon be a requirement for policy assigning certain frontends to certain interfaces (eg. /members is only accessible through the admin interface), which will add to the situation.

CLI11 doesn't have great options for nesting things further, and after discussion we've narrowed things to a couple of possibilities:

1. go to a JSON config file. This entails a one-time conversion from existing CLI or TOML configs, and would preclude the use of inline comments. On the plus side, this is a format we already handle, so we would be net negative in terms of code/dependencies, and passing on that config to the KV as a "node_cfg" akin to "user_data" would be trivial. This may become subject to a flexible policy validation step during join in the future (eg. a `validate_node_config()` defined in constitution).

2. stick with TOML for config, but lose the limited CLI, toml++ (https://marzer.github.io/tomlplusplus/) is a plausible option. This could be made completely compatible with the current config format, and while nesting isn't quite nice as in JSON it still works well enough for our purposes. There's no TCB impact, this remains host-side only, but it's another dependency. Because the config can be serialised to JSON, storage in the KV and policy still can be done, if somewhat more indirectly.

After discussion with @eddyashton, @jumaffre and @lynshi, we have decided to go for option 1.

jumaffre:
List of remaining tasks once #3209 is merged (the PR is already too large):
- [ ] Merge `{"hostname": "...", "port": "..."}` to `host:port` (requires substantial refactoring)
- [ ] Size and time strings (+ remove unit from entry name)
- [ ] JSON schema

jumaffre:
Closing this issue as most of the work is now done. Follow-up work is now in #3354.

PR #2628, opened 2021-05-28
Listen for client RPCs on multiple interfaces on each node

### LINKED ISSUE #2521: Split interfaces with separate hard and soft connection caps

eddyashton:
Resolves #2521.

Currently expressed as a compound comma-separated CLI argument - this will be neater when #2612 is completed. In the Python infra, the main argument for starting each node is a `host` argument. This is now a `HostSpec` object rather than a string, but still represents a single desired host. The only test which starts multiple interfaces, `connections.py`, does so by passing `additional_raw_node_args` to append to the `cchost` CLI. I think this is a temporary solution, and once we have JSON node configuration it will be much simpler for the Python infra to read/modify it, without maintaining multiple intermediate formats.

Draft for discussion/TODOs:
- [X] Mention this in the `CHANGELOG`
- [X] Decide if we're keeping the old arguments for a single interface, or if these should be removed _Keeping these for now. Will revisit as part of porting them to JSON document_
- [x] Decide if `public_rpc_address` should be part of each interface, or a separate concern _Keeping these per-interface. This is likely what we will want in future if interfaces become per-frontend_


doc/operations/recovery.rst
```diff
@@ -25,7 +25,6 @@ To initiate the first phase of the recovery procedure, one or several nodes shou
     --node-address node_ip:node_port
     --rpc-address <ccf-node-address>
     --public-rpc-address <ccf-node-public-address>
-    [--domain domain]
```
eddyashton, line 28: We deprecated `--domain` a long time ago (0.13.0), but it still had semantic impact on `cchost`. This removes it entirely.

src/enclave/client_endpoint.h
```diff
@@ -17,12 +17,12 @@ namespace enclave
 
     HandleDataCallback handle_data_cb;
 
-    size_t session_id;
+    int64_t session_id;
```
eddyashton, line 20: Our `session_id`s were a mix of `size_t` and `int64_t` - mostly signed on the host-side and unsigned in the enclave. This was an unnecessary complication. In fact we have 2 classes of sessions - originating inside and outside - and these were already distinguished by the high-bit. But in-enclave we treated these as "very large unsigneds", when they're more simply "negative numbers".

src/enclave/rpc_sessions.h
```diff
@@ -23,28 +25,38 @@ namespace enclave
...
+      std::pair<ListenInterfaceID, std::shared_ptr<Endpoint>>>
+      sessions;
     size_t sessions_peak;
 
-    // Upper half of sessions range is reserved for those originating from
+    // Negative sessions are reserved for those originating from
     // the enclave via create_client().
-    std::atomic<size_t> next_client_session_id =
-      std::numeric_limits<size_t>::max() / 2;
+    std::atomic<tls::ConnID> next_client_session_id = -1;
```
eddyashton, line 53: Client sessions are now small negative numbers (counting down from `-1`), rather than large unsigneds (counting up from `MAX / 2`), making them more readable in the logs.

src/host/main.cpp
```diff
@@ -475,9 +500,28 @@ int main(int argc, char** argv)
...
+  if (!(*mosh_option))
   {
-    max_open_sessions_hard = max_open_sessions + hard_session_cap_diff;
+    max_open_sessions_hard =
+      max_open_sessions + cli::ParsedRpcInterface::default_mosh_diff;
+  }
+
+  // If --rpc-address etc were specified, they populate a single object at the
+  // start of the rpc_interfaces list
+  if (*rpc_address_option)
```
eddyashton, line 517: We still have the original CLI args `--rpc-address`, `--max-open-sessions` etc. They allow you to specify a single interface, that could have been a `--rpc-interface` option. I think its worth keeping this complication for now, assuming that when we do #2612 and have a tool to convert CLI to JSON, we will flatten into a single form there.

src/host/main.cpp
```diff
@@ -661,16 +697,22 @@ int main(int argc, char** argv)
...
-    if (!rpc_address_file.empty())
+
+    std::string rpc_addresses;
+    for (auto& interface : rpc_interfaces)
     {
-      files::dump(
-        fmt::format("{}\n{}", rpc_address.hostname, rpc_address.port),
-        rpc_address_file);
+      rpc.listen(0, interface.rpc_address.hostname, interface.rpc_address.port);
+      rpc_addresses += fmt::format(
```
eddyashton, line 705: When we write the (possibly resolved) RPC addresses to file, we write all of them.

src/node/node_info_network.h
```diff
@@ -11,14 +11,38 @@ namespace ccf
...
+      bool operator==(const NetAddress& other) const
+      {
+        return hostname == other.hostname && port == other.port;
+      }
+    };
+
+    struct RpcAddresses
+    {
+      NetAddress rpc_address;
+      NetAddress public_rpc_address;
```
eddyashton, line 28: I keep going back-and-forth on whether this should be here. We never use the public address of anything but the first interface, so maybe we just need 1 public address. Its up to the operator how the locally-bound interfaces are exposed on the public internet, and then they need to give the node a single name for this for redirection/node info purposes? But if we end up segregating specific frontends or endpoints on each interface, we want finer control of public interfaces ("if you're a user looking for node 0, the public address is FOO, if you're a member/looking for operator functionality, its BAR").

src/node/node_state.h
```diff
@@ -1513,27 +1513,55 @@ namespace ccf
...
+        if (c < '0' || c > '9')
+        {
+          return false;
+        }
+      }
+
+      return true;
     }
 
     std::vector<crypto::SubjectAltName> get_subject_alternative_names()
```
eddyashton, line 1547: This used to be "everything passed to `--san`, plus one extra element (either the IP address on `--rpc-address`, or the DNS name on `--domain`". It is now "everything passed to `--san`, unless `--san` is empty in which case the default is each RPC address from `--rpc-interface` (auto-detected as IP address or DNS name)". I think that gives operators everything they need (the default should work for local testing, then you override it with `--san` for deployment, but needs documenting.

src/ds/cli_helper.h
```diff
@@ -19,11 +19,45 @@ namespace cli
...
+  {
+    auto found = addr.find_last_of(":");
+    auto hostname = addr.substr(0, found);
+
+    const auto port =
+      found == std::string::npos ? default_port : addr.substr(found + 1);
+
+    // Check if port is in valid range
+    int port_int;
+    try
```
achamayou, line 36: uint16_t port_int = 0;
std::from_chars(port_str.data(), port_str.data() + port_str.size(), port_int);
...

to save the manual validation perhaps?

src/enclave/rpc_sessions.h
```diff
@@ -118,54 +167,87 @@ namespace enclave
...
+        per_listen_interface.open_sessions >=
+        per_listen_interface.max_open_sessions_soft)
       {
         LOG_INFO_FMT(
-          "Soft refusing a session inside the enclave - already have {} "
-          "sessions and limit is {}: {}",
-          sessions.size(),
-          max_open_sessions_soft,
-          id);
+          "Soft refusing session {} inside the enclave - already have {} "
```
achamayou, line 212:
```suggestion
          "Soft refusing session {} (returning 503) inside the enclave - already have {} "
```

src/node/node_info_network.h
achamayou: Different interfaces may be routed differently, in particular if we ever get to multiple apps and more complex policy rules. I agree for now that it's largely just one interface that will have a public variant, but it seems reasonable to keep it specific just in case?

src/enclave/rpc_sessions.h
```diff
@@ -77,6 +89,24 @@ namespace enclave
...
     };
 
+    tls::ConnID get_next_id()
+    {
+      auto id = next_client_session_id--;
+
+      if (next_client_session_id > 0)
+        next_client_session_id = -1;
+
+      while (sessions.find(id) != sessions.end())
```
jumaffre, line 99: Could this loop forever in the (very unlikely) case that all outgoing ids are already taken?

src/enclave/rpc_sessions.h
```diff
@@ -77,6 +89,24 @@ namespace enclave
       }
     };
 
+    tls::ConnID get_next_id()
```
jumaffre, line 92:
```suggestion
    tls::ConnID get_next_client_id()
```

jumaffre (approved):

src/enclave/rpc_sessions.h
eddyashton: Yes, it could. It is based on the similar loop in the host-side `RPCConnections::get_next_id()`. This can only happen if we have 2^63 sessions in one of these maps, at which point we have bigger problems. But I've added a throw in both these functions, just for clarity.

achamayou (approved):

eddyashton:
There are 2 issues where this PR currently breaks the `lts_compatibility_test`:

- Firstly it tries to change the CLI arguments that are used to launch all nodes, from `--rpc-address` etc to `--rpc-interface`. That's fine, that was a step towards clean deprecation but is unnecessary - we can still use the old arg names in the infra.
- Secondly it tries to change the format of `NodeInfoNetwork`, a type that's stored in the ledger, considerably. The transformation in the current PR isn't the usual pattern - adding some new fields that need to be optional because old nodes might not have them - it is entirely renaming/restructuring existing fields. I need to think more about how to do this in a backwards-compatible way - either we need a hybrid structure that keeps the first value in the old field names, or to make all the new values optional and write duplicate entries to the old field names during serialisation?

CHANGELOG.md
```diff
@@ -39,6 +43,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Nodes code digests are now extracted and cached at network join time in `public:ccf.gov.nodes.info`, and the `/node/quotes` and `/node/quotes/self` endpoints will use this cached value whenever possible (#2651).
 
+### Added
```
jumaffre, line 46: Duplicate?

src/ds/cli_helper.h
```diff
@@ -19,11 +20,50 @@ namespace cli
...
+    const std::string& option_name,
+    const std::string& default_port = "0")
+  {
+    auto found = addr.find_last_of(":");
+    auto hostname = addr.substr(0, found);
+
+    const auto port =
+      found == std::string::npos ? default_port : addr.substr(found + 1);
+
+    // Check if port is in valid range
```
achamayou, line 35: I don't think we should effectively forbid the use of service names, they're still very widely supported and sometimes used. If we want to have an integer type internally, we can resolve early for all cases except 0 (https://linux.die.net/man/3/getservbyname).

src/ds/cli_helper.h
```diff
@@ -19,11 +20,50 @@ namespace cli
...
+
+    return true;
+  }
+
   CLI::Option* add_address_option(
     CLI::App& app,
     ParsedAddress& parsed,
     const std::string& option_name,
-    const std::string& option_desc,
+    const std::string& option_desc = "",
```
achamayou, line 66: Is this right? I would have thought that we always want a non-empty description.

src/node/node_info_network.h
```diff
@@ -18,7 +18,115 @@ namespace ccf
...
+    NodeInfoNetwork_v1, rpchost, pubhost, nodehost, nodeport, rpcport, pubport);
+
+  struct NodeInfoNetwork_v2
+  {
+    struct NetAddress
+    {
+      std::string hostname;
+      std::string port;
+
+      bool operator==(const NetAddress& other) const
```
achamayou, line 32: Can be default now.

src/node/node_info_network.h
```diff
@@ -18,7 +18,115 @@ namespace ccf
...
+    {
+      NetAddress rpc_address;
+      NetAddress public_rpc_address;
+
+      size_t max_open_sessions_soft;
+      size_t max_open_sessions_hard;
+
+      bool operator==(const RpcAddresses& other) const
+      {
+        return rpc_address == other.rpc_address &&
```
achamayou, line 48: Ditto

tests/connections.py
```diff
@@ -174,6 +202,15 @@ def create_connections_until_exhaustion(target, continue_to_hard_cap=False):
         to_create = max_fds - num_fds + 1
         num_fds = create_connections_until_exhaustion(to_create)
 
+        # Check that lower caps are enforced on each interface
+        for i, (address, cap) in enumerate(caps.items()):
+            create_connections_until_exhaustion(
+                cap + 1,
+                client_fn=lambda **kwargs: primary.client(
```
achamayou, line 209: client_fn=functools.partial(primary.client, interface_idx=i + 1)

tests/infra/node.py
```diff
@@ -88,27 +88,34 @@ def __init__(
...
         self.node_client_host = None
+        self.interfaces = []
         self.version = version
         self.consensus = None
 
-        if host.startswith("local://"):
+        if isinstance(host, str):
+            host = infra.e2e_args.HostSpec.from_str(host)
+
+        if type(host) == str:
```
achamayou, line 98: lines 95 will already do this for all subclasses of str, this is presumably intended to be removed?

tests/infra/node.py
```diff
@@ -397,7 +417,9 @@ def signing_auth(self, name=None):
             "signing_auth": self.identity(name),
         }
 
-    def client(self, identity=None, signing_identity=None, **kwargs):
+    def client(
+        self, identity=None, signing_identity=None, interface_idx=None, **kwargs
```
achamayou, line 421: I think we almost certainly want to name interface.

achamayou (approved):

src/node/node_info_network.h
eddyashton: I think if we start using `= default` we're enforcing C++20 (rather than supporting both C++20 and C++17), which we don't want to do yet?

src/ds/cli_helper.h
eddyashton: Yeah good catch, I think this must have been laziness while adding these options?

tests/infra/node.py
eddyashton: I'm not sure what you mean here, you want the arg to be `interface` (and eventually an identifiable string) rather than `interface_idx`?

tests/infra/node.py
eddyashton: Good spot, don't know I ended up with both. I'll blame a bad merge.

src/ds/cli_helper.h
eddyashton: This is just a move/rewrite of the existing logic, that was previously lines 43-52. We can add service name support if we want, but we didn't have it previously so I think we can make that a separate PR.

tests/infra/node.py
achamayou: Yes, I think this will be useful for definition, monitoring etc. But we can re-visit that once we change the config format.

src/node/node_info_network.h
achamayou: True, although we could also ifdef and it'd make the old stuff easier to drop later.

PR #2633, opened 2021-06-01
Highlight entity IDs in Python output

eddyashton:
From an idea discussed briefly with @achamayou. This finds all entity IDs (or any long-enough hex-string) in the Python logging, and gives it a colour based on the string so it can be easily spotted/distinguished in the output:

image

This is a broad approach, giving consistency (every instance of a proposal ID is highlighted) at perhaps the cost of readability (its not so easy to identify a URL when part of it is coloured). 

This also produces some hard-to-read colours - I think we could mitigate this by scaling to avoid too-bright/too-dark colours, or just choosing from a dict.

I've also just noticed that it appears to have lost all _other_ colour coding in our messages, so I'll fix that.

### DIFF
--- a/tests/infra/e2e_args.py
+++ b/tests/infra/e2e_args.py
@@ -5,6 +5,7 @@ import os
 import infra.path
 import infra.network
 import sys
+import re
 
 from loguru import logger as LOG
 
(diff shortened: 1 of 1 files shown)


eddyashton:
This approach doesn't work. Loguru does the colour tag substitution in a slightly odd order, so if we try to modify the 'message' like this we get the stripped version, losing the colour tags that were already _in_ 'message'. We can still add new tags (ie - highlight entity IDs), but not in parallel with other in-message colouring.

I'll see if I can fix this by moving the highlighting earlier in the logging evaluation, otherwise we may have to settle for enabling this with an ENV switch.

eddyashton:
I don't see a way to get our existing colours _and_ these entity ID highlights. Have raised an issue on loguru in case I'm missing something:

https://github.com/Delgan/loguru/issues/457

eddyashton:
Closing this. We'll instead write a post-processing tool that does this highlighting on arbitrary text, so we can use it for both the Python output and the node output files.

PR #2634, opened 2021-06-01
Add decoration around EntityID in node output

eddyashton:
`EntityID`s are currently printed as long hex strings, and it can be hard to identify them in the output and hard to distinguish what domain they're in. This adds decoration to make grepping and domain-distinction easier.

So rather than a line in the node output saying:

```
012345 has voted to add 123456 on 234567
```

we'd see

```
=[012345]= has voted to add -[123456]- on <[234567]>
```

and (hopefully) learn to quickly recognise `=[members]=`, `-[users]-` and (most commonly) `<[nodes]>`.

This distinction relies on the strong typing of (now-distinct) `EntityID`s, so it won't work in places that see these IDs as raw strings (ie - in JS, or for ProposalIDs, or in places where we deliberately support any kind of `EntityID`).

### DIFF
--- a/include/ccf/entity_id.h
+++ b/include/ccf/entity_id.h
@@ -136,7 +136,7 @@ namespace ccf
   {
     static std::string format(const std::string& core)
     {
-      return fmt::format("=[{}]=", core);
+      return fmt::format("m[{}]", core);
     }
   };
   using MemberId = EntityId<MemberIdFormatter>;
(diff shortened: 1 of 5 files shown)


wintersteiger (approved):

ashamis (approved):

achamayou (approved):

achamayou:
Issue in code_update_test is unrelated, and this is a change to help debug exactly that kind of problem.

PR #2635, opened 2021-06-02
AFT: fix race conditions

ashamis:
This change resolves 2 timing based race conditions.
- For a ViewChangeConfirmation message we were storing 1 seqno this was done to reduce the number of times we store the seqno. Unfortunately, this is not correct - f replicas could be behind the other 2f+1 replicas and thus would have produced their ViewChangeRequest message with a different seqno and merkle root, in this case their ViewChangeRequest message is correct but we would have lost the ability to verify the ViewChangeRequest message.

- If a primary is behind and becomes a backup (without becoming aware of this) the old primary could attempt to sign a seqno that the new primary signed. If the old primary received prepare messages from other replicas based on the new primary's signature we would get a situation where the same thread obtains the store lock twice. Changing this so that if a primary signs a seqno any other prepare messages received for said seqno are ignored, this is the correct behavior and removes the scenario in which the deadlock occurs.

- When running in BFT, on the backup, we execute and then add the hash of the encrypted append-entry to the merkle tree. This is a problem for a node that is catching up append entries written 1 or more views ago because we use the node's current view in the iv when encrypting the append-entry. In this PR we obtain the view from the iv of the primary created append entry and use this view when encrypting on the backup.

Lots of view-change logging added to making finding view-change bugs easier.

### DIFF
--- a/include/ccf/tx.h
+++ b/include/ccf/tx.h
@@ -69,7 +69,6 @@ namespace kv
     // std::optional with nullopt representing "not yet fetched".
     std::optional<Version> read_version = std::nullopt;
     ccf::View view = ccf::VIEW_UNKNOWN;
-    // std::optional<ccf::View> replicated_view;
 
     std::map<std::string, std::shared_ptr<AbstractMap>> created_maps;
 
--- a/include/ccf/tx.h
+++ b/include/ccf/tx.h
@@ -69,6 +69,7 @@ namespace kv
     // std::optional with nullopt representing "not yet fetched".
     std::optional<Version> read_version = std::nullopt;
     ccf::View view = ccf::VIEW_UNKNOWN;
+    std::optional<ccf::View> replicated_view;
 
     std::map<std::string, std::shared_ptr<AbstractMap>> created_maps;
 
--- a/include/ccf/tx.h
+++ b/include/ccf/tx.h
@@ -69,7 +69,7 @@ namespace kv
     // std::optional with nullopt representing "not yet fetched".
     std::optional<Version> read_version = std::nullopt;
     ccf::View view = ccf::VIEW_UNKNOWN;
-    std::optional<ccf::View> replicated_view;
+    // std::optional<ccf::View> replicated_view;
 
     std::map<std::string, std::shared_ptr<AbstractMap>> created_maps;
 
--- a/include/ccf/tx.h
+++ b/include/ccf/tx.h
@@ -69,7 +69,6 @@ namespace kv
     // std::optional with nullopt representing "not yet fetched".
     std::optional<Version> read_version = std::nullopt;
     ccf::View view = ccf::VIEW_UNKNOWN;
-    std::optional<ccf::View> replicated_view;
 
     std::map<std::string, std::shared_ptr<AbstractMap>> created_maps;
 
--- a/src/crypto/symmetric_key.h
+++ b/src/crypto/symmetric_key.h
@@ -40,7 +40,6 @@ namespace crypto
       *reinterpret_cast<uint64_t*>(iv) = seq;
     }
 
-    // TODO: inverse of this
     void set_iv_term(uint64_t term)
     {
       if (term > 0x7FFFFFFF)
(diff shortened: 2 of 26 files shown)


include/ccf/tx.h
```diff
@@ -69,6 +69,7 @@ namespace kv
     // std::optional with nullopt representing "not yet fetched".
     std::optional<Version> read_version = std::nullopt;
     ccf::View view = ccf::VIEW_UNKNOWN;
+    std::optional<ccf::View> replicated_view;
```
achamayou, line 72: It feels like maybe those could be combined into a single view/replicated_view?

src/node/progress_tracker.h
```diff
@@ -74,8 +75,13 @@ namespace ccf
         sig);
 
       auto it = certificates.find(tx_id.seqno);
-      if (it == certificates.end())
+      if (it == certificates.end() || am_i_primary)
```
achamayou, line 78: Comment

src/node/progress_tracker.h
```diff
@@ -558,57 +557,35 @@ namespace ccf
...
             seqno,
             sig.node,
-            it->second.root,
+            view_change.root,
             sig.sig,
             sig.sig.size());
           verified_signatures = false;
           continue;
         }
-
```
achamayou, line 567: ?

src/node/test/progress_tracker.cpp
```diff
@@ -325,24 +326,71 @@ TEST_CASE("Request tracker")
 
 TEST_CASE("Record primary signature")
 {
+  using trompeloeil::_;
+
   kv::NodeId my_node_id = kv::test::PrimaryNodeId;
-  ccf::View view = 0;
+  ccf::View view = 1;
```
achamayou, line 332:
```suggestion
  ccf::View view = 0;
```

achamayou (approved):

src/node/progress_tracker.h
```diff
@@ -74,8 +75,17 @@ namespace ccf
         sig);
 
       auto it = certificates.find(tx_id.seqno);
-      if (it == certificates.end())
+      if (it == certificates.end() || am_i_primary)
       {
+        // If a primary is behind and becomes a backup (without becoming aware
+        // of this) the old primary could attempt to sign a seqno that the new
+        // primary signed. In this case clear an prepares that we could have
```
wintersteiger, line 82:
```suggestion
        // primary signed. In this case clear any prepares that we could have
```

src/node/progress_tracker.h
```diff
@@ -104,30 +114,27 @@ namespace ccf
...
+              sig->second.sig.size(),
+              sig->second.sig.data()))
           {
-            // NOTE: We need to handle this case but for now having this make a
-            // test fail will be very handy
-            throw ccf::ccf_logic_error(fmt::format(
-              "record_primary: Signature verification from {} FAILED, view:{}, "
-              "seqno:{}",
-              sig.first,
+            sig = cert.sigs.erase(sig);
```
wintersteiger, line 127: Can't see the type of `cert.sigs` here, but it looks like the iterator may be invalidate here.

src/node/progress_tracker_types.h
```diff
@@ -251,14 +250,9 @@ namespace ccf
...
-          "Failed to write new_view, view:{}, seqno:{}",
-          new_view.view,
-          new_view.seqno);
-        throw ccf_logic_error(fmt::format(
-          "Failed to write new_view, view:{}, seqno:{}",
-          new_view.view,
-          new_view.seqno));
+        LOG_FAIL_FMT("Failed to write new_view, view:{}", new_view.view);
+        throw ccf_logic_error(
+          fmt::format("Failed to write new_view, view:{}", new_view.view));
```
wintersteiger, line 255: Don't duplicate error string?

wintersteiger (approved):

src/crypto/symmetric_key.h
```diff
@@ -52,6 +52,11 @@ namespace crypto
         static_cast<uint32_t>(term);
     }
 
+    uint64_t get_term() const
+    {
+      return *reinterpret_cast<const uint32_t*>(iv + IV_DELIMITER);
```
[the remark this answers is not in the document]
ashamis: I am going to leave this as is to make sure it is symmetrical with how we put in the term

PR #2644, opened 2021-06-04
Add ability to parse snapshots with Python `ccf.ledger` module

### LINKED ISSUE #2484: Python `ledger` module should also parse snapshots
Follow-up from https://github.com/microsoft/CCF/pull/2482 

It should also be possible for the `ledger` module to read snapshots so that partial ledgers (e.g. when a node joined or recovered from a snapshot) can be verified. Note that this should be straightforward enough as the format of snapshots is similar to serialised transactions.

jumaffre:
Fixes #2484 

We can now also parse snapshots with the Python `ccf.ledger` module as follows:

```python
import ccf.ledger

with ccf.ledger.Snapshot("/path/to/snapshot/file") as snapshot:
    tables = snapshot.get_public_domain().get_tables()  # Just like for transactions in the ledger
    # Loop through tables, etc.
```

Also updated the ledger docs that had rotten and updated the `read_ledger.py` utility to also display snapshots (via `--snapshot` flag).


doc/audit/python_library.rst
```diff
@@ -9,37 +9,49 @@ This page describes the Python API of the :py:class:`ccf.ledger` module which ca
...
         $ read_ledger.py /path/to/ledger/dir
-        Initialised CCF ledger from directory '/path/to/ledger/dir' (found 6 files)
-        seqno 1 (11 public tables)
-        table "public:ccf.gov.constitution":
+        Reading ledger from ['/path/to/ledger/dir']
+        Contains 9 chunks
+        chunk /path/to/first/ledger/chunk (committed)
+          seqno 1 (9 public tables)
+        ...
+        Ledger verification complete. Found 15 signatures, and verified till 2.52
```
achamayou, line 22:
```suggestion
        Ledger verification complete. Found 15 signatures, and verified until 2.52
```

python/ccf/ledger.py
```diff
@@ -375,56 +439,25 @@ def get_size():
...
-    _next_offset: int = LEDGER_HEADER_SIZE
     _public_domain: Optional[PublicDomain] = None
     _file_size: int = 0
     gcm_header: Optional[GcmHeader] = None
-    _tx_offset: int = 0
-    _ledger_validator: LedgerValidator
 
-    def __init__(self, filename: str, ledger_validator: LedgerValidator):
-        self._ledger_validator = ledger_validator
+    @abc.abstractmethod
```
achamayou, line 450: An implemented abstractmethod is... weird?

python/ccf/ledger.py
```diff
@@ -493,6 +567,20 @@ def __next__(self):
...
+    """
+    Utility used to parse the content of a snapshot file.
+    """
+
+    def __init__(self, filename: str):
+        super().__init__(filename)
+        self._file_size = os.path.getsize(filename)
+        super()._read_header()
+
+    def __del__(self):
```
achamayou, line 580: Not guaranteed to execute, not a good way to do cleanup (or anything in fact).

achamayou (approved):

python/ccf/ledger.py
jumaffre: Good point. That had been there forever but is now fixed.

python/ccf/ledger.py
jumaffre: Yeah, I've remove the use of the `abc` module.

PR #2650, opened 2021-06-10
Remove redundant suspension-based rotation test

achamayou:
Effectively replaced by the more accurate #2586.


eddyashton (approved):
We should take a look at other uses of `.suspend()` within the e2e tests, and work out if they can/should be replaced by partitions.

PR #2656, opened 2021-06-11
Return details of HTTP parse errors in HTTP response

### LINKED ISSUE #2652: Provide details on HTTP parse failures rather than just close the connection
Upon failing to parse an HTTP request, we currently close the connection. We do not want to log details, to avoid impacting confidentiality, but we could assemble an HTTP response with details and send it back. Users will hopefully be able to either interpret or report this response, suitably sanitised if necessary.

Additionally, we want to log failures to parse forwarded requests in a distinct way, so we can detect if our re-serialisation is causing problems in some contexts.

eddyashton:
Resolves #2652.

Previously we closed this connection without responding to the user, meaning there was no way to debug _what_ was unparseable. Now we return a raw exception message to the caller in a textual HTTP response, which hopefully makes it clear what was wrong with their request.

I investigated doing a similar thing for forwarded requests, but have decided not to. These requests need to make it through this stage of parsing on the first node before they are forwarded, so any failures that only arise on the forwardee would have to come from either a) a mismatch in parsing behaviour between nodes, or b) an error in our re-serialisation of the request. Neither of these is likely to be fixable by the caller, so returning them a "this isn't valid HTTP" response (with HTTP they maybe didn't send!) seems rude. Further, this is only part of a more general inconsistency; there are a great many failure states that produce a nice error on the first node, but a generic forwarding failed-with-timeout error if they occur on the forwardee. We should ideally convert all of those to respond all the way to the caller, ideally with minimal distinction indicating they were forwarded, but that's a far more involved and risky change than is warranted here. In particular we may want to cherry pick _this_ but not _that_ change to the 1.0.x release.


achamayou (approved):

achamayou:
@eddyashton can we make parsing errors on requests that were forwarded distinct in the log though? That would help us figure out if they happen at all.

If they do, there's very likely a problem in the way we do our re-serialisation.

eddyashton:
> @eddyashton can we make parsing errors on requests that were forwarded distinct in the log though?

@achamayou Yes, we already do. Forwarded requests go through a very different path, and parse errors thrown from `HttpParser::execute()` (which would have logged `Error parsing HTTP request` when not forwarded) will log `Invalid forwarded command` (and perhaps other forwarding-specific messages) if they only fail when forwarded.

achamayou:
@eddyashton let's highlight this particular error message to Greg along with the release containing this patch, so we can completely rule out problems with forwarding.

achamayou:
@eddyashton can we have a changelog entry for this please (we'll need one for the LTS as well)?

Issue #2315, opened 2021-03-15
JS API cleanup


letmaik:
From https://github.com/microsoft/CCF/pull/2313#issuecomment-799237637 by @achamayou :

1. We seem to have taken a rather random view to capitalising acronyms, so we have RsaOaep-like things and AESKW-like things. I don't particularly care which one we go with (although the latter makes more sense to me), but it's important to be consistent. DONE: https://github.com/microsoft/CCF/pull/2325
2. All the crypto calls like generate key, wrap etc should be under `ccf.crypto`, and we should perhaps have `ccf.conv` for the conversion calls.

letmaik:
Regarding naming, I think PascalCase for acronyms is better, since it provides separation between two subsequent acronyms, like AES and KW -> AesKw. This is in line with the web crypto api: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/wrapKey. Currently we have the following outliers: `CCF`, `KVMap`, and `AESKWPParams`. These are TypeScript interfaces. There are no inconsistencies in the actual JS API.

achamayou:
Ok let’s align on the web crypto API naming then.

achamayou:
Only `CCF` remains, which I think is acceptable since it's a single, short acronym.

PR #2661, opened 2021-06-14
js: Add `ccf.crypto.verifySignature`

letmaik:
Loosely modeled after https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/verify.


js/ccf-app/src/crypto.ts
```diff
@@ -32,6 +32,11 @@ export const generateRsaKeyPair = ccf.generateRsaKeyPair;
  */
 export const wrapKey = ccf.wrapKey;
 
+/**
+ * @inheritDoc CCF.verifySignature
+ */
+export const verifySignature = ccf.verifySignature;
```
achamayou, line 38: #2315, can this go under crypto now please, the ccf namespace is crowded enough.

tests/infra/crypto.py
```diff
@@ -79,6 +79,25 @@ def generate_rsa_keypair(key_size: int) -> Tuple[str, str]:
     return priv_pem, pub_pem
 
 
+def generate_ec_keypair(curve_name: str) -> Tuple[str, str]:
+    if curve_name == "secp256k1":
```
achamayou, line 83: I believe we've dropped support for secp256k1?

wintersteiger: That's right, we removed that. Currently we support only secp384r1 and secp256r1. In general JS applications might want to use it for other purposes in their application, so it would still make a little bit of sense to keep in the API, but it's not necessary.

js/ccf-app/src/crypto.ts
letmaik: Done.

tests/infra/crypto.py
letmaik: I switched to `secp256r1`.

js/ccf-app/src/global.ts
```diff
@@ -251,6 +263,25 @@ export interface CCF {
    */
   isValidX509CertChain(chain: string, trusted: string): boolean;
 
+  crypto: {
+    /**
+     * Returns whether digital signature is valid.
```
achamayou, line 268: It would be good to list the RSA key sizes and EC curve names this supports for clarity, either here or in the definitions of the SigningAlgorithm interfaces.

achamayou (approved):
@letmaik thank you for this useful API addition. Since this is making new functionality available to users, could you please document it in the Changelog?

js/ccf-app/src/global.ts
letmaik: There is no restriction on the RSA key sizes or EC curves. Anything that OE-OpenSSL supports will work. The only limitation is in the supported digest algorithms which is already a TypeScript type and self-documented. I'd like to extend that from SHA-256 eventually to SHA-384 but want to do it across the board, so including ccf.digest(), and currently CCF doesn't have native SHA-384 digest support.
I found https://github.com/openenclave/openenclave/blob/master/docs/OpenSSLSupport.md but it's not that useful.

achamayou: Ok, thank you.

letmaik:
If this PR is ok I'd like to get it merged and if possible kick off a dev release to consume it together with the new bytecode cache feature.

Issue #2136, opened 2021-01-29
Bind Proposal Ids to root of Merkle Tree when running BFT


achamayou:
Follow-up to #1649 and #2087. Proposal Ids are bound to the state at the read version, and read version selection isn't deterministic in the presence of concurrent transactions. So a possible way to resolve that for BFT is to:

- [ ] Add a "flush" flag to the read version for backup re-execution, which will be useful for both signatures and proposals, since they both need to express not just KV state dependencies but also History state dependencies

~~Add a mechanism to capture non-deterministic values during primary execution, likely in a KV table, and to send them to secondaries for re-execution. That's one way to capture the merkle tree read version, and the mechanism would be generally useful once we add support for REST requests from CCF etc~~

ashamis:
Discussing this with @achamayou  the current proposal is to have a pipeline execution stall for transactions that write to the proposals table on backups. This will ensure that the transaction will see the correct read version.

PR #2663, opened 2021-06-15
BFT recovery tests

ashamis:
This PR is the beginning of recovery for a BFT CCF network. BFT will replicate all tables so that we can align with how CFT recovery works, but uses BFT verification when verifying the public ledger. The small bank tests do not show a performance degradation.

### DIFF
--- a/samples/apps/logging/logging.cpp
+++ b/samples/apps/logging/logging.cpp
@@ -756,6 +756,8 @@ namespace loggingapp
         .set_auto_schema<void, LoggingGetHistorical::Out>()
         .add_query_parameter<size_t>("id")
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
+        .set_execute_outside_consensus(
+          ccf::endpoints::ExecuteOutsideConsensus::Locally)
         .install();
       // SNIPPET_END: get_historical
 
@@ -809,6 +811,8 @@ namespace loggingapp
         .set_auto_schema<void, LoggingGetReceipt::Out>()
         .add_query_parameter<size_t>("id")
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
+        .set_execute_outside_consensus(
+          ccf::endpoints::ExecuteOutsideConsensus::Locally)
         .install();
       // SNIPPET_END: get_historical_with_receipt
 
@@ -1068,6 +1072,8 @@ namespace loggingapp
           "to_seqno", ccf::endpoints::QueryParamPresence::OptionalParameter)
         .add_query_parameter<size_t>("id")
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
+        .set_execute_outside_consensus(
+          ccf::endpoints::ExecuteOutsideConsensus::Locally)
         .install();
 
       auto record_admin_only = [this](
--- a/src/consensus/aft/impl/state.h
+++ b/src/consensus/aft/impl/state.h
@@ -99,6 +99,8 @@ namespace aft
       current_view(0),
       last_idx(0),
       commit_idx(0),
+      cft_watermark_idx(0),
+      bft_watermark_idx(0),
       new_view_idx(0)
     {}
 
--- a/src/consensus/aft/impl/state.h
+++ b/src/consensus/aft/impl/state.h
@@ -118,6 +118,7 @@ namespace aft
     ViewHistory view_history;
     kv::Version new_view_idx;
     std::optional<ccf::NodeId> requested_evidence_from = std::nullopt;
-    std::optional<std::tuple<ccf::NodeId, ccf::View>> public_only_from = std::nullopt;
+    std::optional<std::tuple<ccf::NodeId, ccf::View>> public_only_from =
+      std::nullopt;
   };
 }
--- a/src/consensus/aft/impl/state.h
+++ b/src/consensus/aft/impl/state.h
@@ -116,5 +116,6 @@ namespace aft
     ViewHistory view_history;
     kv::Version new_view_idx;
     std::optional<ccf::NodeId> requested_evidence_from = std::nullopt;
+    std::optional<std::tuple<ccf::NodeId, ccf::View>> public_only_from = std::nullopt;
   };
 }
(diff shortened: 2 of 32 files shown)


samples/apps/logging/logging.cpp
```diff
@@ -756,6 +756,8 @@ namespace loggingapp
         .set_auto_schema<void, LoggingGetHistorical::Out>()
         .add_query_parameter<size_t>("id")
         .set_forwarding_required(ccf::endpoints::ForwardingRequired::Never)
+        .set_execute_outside_consensus(
```
achamayou, line 759: I am worried by this. We've historically only used this for node-specific data, which was clearly out of consensus. For business/application endpoints though, the expectation/reality so far is that consensus guarantees apply evenly.

We are lacking clear guidance of when to use this flag and what the implications are, I think this is something we need to address.

src/consensus/aft/impl/state.h
```diff
@@ -116,5 +118,7 @@ namespace aft
     ViewHistory view_history;
     kv::Version new_view_idx;
     std::optional<ccf::NodeId> requested_evidence_from = std::nullopt;
+    std::optional<std::tuple<ccf::NodeId, ccf::View>> initial_primary =
```
achamayou, line 121: What does the view represent here? How does it differ from current_view?

src/consensus/aft/raft.h
```diff
@@ -2095,7 +2124,10 @@ namespace aft
       progress_tracker->add_nonce_reveal(
         {r.term, r.idx}, r.nonce, from, node_count(), is_primary());
 
-      update_commit();
+      if (!public_only)
```
achamayou, line 2127: Why don't we update commit in public only mode? How can we commit governance changes then?

src/enclave/interface.h
```diff
@@ -84,6 +84,7 @@ struct CCFConfig
   size_t jwt_key_refresh_interval_s;
 
   crypto::CurveID curve_id;
+  uint32_t node_id;
```
achamayou, line 87: Why is this being duplicated out of node state?

src/host/main.cpp
```diff
@@ -388,6 +388,14 @@ int main(int argc, char** argv)
     ->transform(CLI::CheckedTransformer(curve_id_map, CLI::ignore_case))
     ->capture_default_str();
 
+  uint32_t node_id = -1;
+  app
+    .add_option(
+      "--bft-node-id",
```
achamayou, line 394: I don't think this should be assigned manually, there's a serious re-use risk. We should work this out from the public ledger recovery instead.

src/node/request_tracker.h
```diff
@@ -149,6 +162,11 @@ namespace aft
       hashes_without_requests_list.clear();
     }
 
+    void start_tracking_requests()
```
achamayou, line 165: No locking here?

src/node/rpc/member_frontend.h
```diff
@@ -1086,7 +1086,10 @@ namespace ccf
         // creations are in-flight and reading at the same version, all except
         // the first conflict and are re-executed. If we ever produce a
         // proposal ID which already exists, we must have a hash collision.
-        if (pm->has(proposal_id))
+        //
+        // This is current broken broken for BFT.
+        // https://github.com/microsoft/CCF/issues/2668
```
achamayou, line 1091: What's the break?

src/consensus/aft/raft.h
ashamis: I will add a comment to the code - What is going on here is during public only mode we do not want to move commitment forward because of nonces via the message channels only from evidence that is written on the ledger.

src/consensus/aft/impl/state.h
achamayou: initial_recovery_primary

achamayou: Please add a comment and a link to the relevant issue.

src/consensus/aft/raft.h
achamayou: As discussed: update_commit should be safe for a sparse history, and we don't need to/want to validate the pre-catastrophe ledger.

src/consensus/aft/raft_tables.h
```diff
@@ -9,14 +9,6 @@
...
-
-  static constexpr auto replicate_type_bft = kv::ReplicateType::SOME;
-  static const std::unordered_set<std::string> replicated_tables_bft = {
-    ccf::Tables::AFT_REQUESTS,
-    ccf::Tables::SIGNATURES,
-    ccf::Tables::BACKUP_SIGNATURES,
-    ccf::Tables::NONCES,
-    ccf::Tables::NEW_VIEWS};
+  static constexpr auto replicate_type = kv::ReplicateType::ALL;
+  static const std::unordered_set<std::string> replicated_tables = {};
```
achamayou, line 13: Please split.

src/host/main.cpp
achamayou:
```
      auto tx = network.tables->create_tx();
      GenesisGenerator g(network, tx);
      g.create_service(network.identity->cert);
      g.retire_active_nodes();

      g.add_node(
```

I think we can set self correctly in the gap here, and have a temp value before.

src/node/rpc/member_frontend.h
achamayou: This is really #2136, and apart from the solutions proposed in the ticket, we could add a nonce to duplicate (or all proposals) in BFT as a short term hack.

achamayou (approved):

src/node/node_state.h
```diff
@@ -972,6 +972,20 @@ namespace ccf
       g.create_service(network.identity->cert);
       g.retire_active_nodes();
 
+      if (network.consensus_type == ConsensusType::BFT)
+      {
+        // BFT consensus requires a stable order of node IDs so that the
+        // primary node in a given view can be computed deterministically by
+        // all nodes in the network
+        // See https://github.com/microsoft/CCF/issues/1852
```
achamayou, line 980: @ashamis so as discussed, we will unify ids with CFT, but keep a mapping of node_id -> election_order_id and use that to decide election order, right? If that's right, can it be recorded in #1852.

PR #2659, opened 2021-06-14
[Do not merge] Code update repro

eddyashton:
Unable to get local repro of this: https://github.com/microsoft/CCF/issues/2625

This may be the same issue, though the diagostic has changed slightly: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=26955&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=d3511c0a-aced-5990-4c7d-8e47aebb8200&l=17561

Re-running in CI with, hopefully, a `/node/consensus` + view history dump on failure.

PR #2665, opened 2021-06-15
Turn off code_update_test in instrumented build

achamayou:
This is breaking in the Daily and under investigation in #2659.


ashamis (approved):

eddyashton:
I believe this PR is now unnecessary - the last problematic channel closure was removed in #2654, and there's a fix for (an) underlying issue in #2688. If that's correct, then `code_update_test` should now be stable, and this is a useful test to keep (in particular it seems to be our main canary for channels issues, which will be critical for validating #2672).

achamayou:
@eddyashton excellent, will have a look at #2688.

PR #2666, opened 2021-06-15
Backport LTS compatibility test to `release/1.x` branch

jumaffre:
This PR backports the Python-only changes from #2600 and #2595 to the `release/1.x` branch so that we test for compatibility when releasing patches on this first release branch too.


src/host/main.cpp
```diff
@@ -457,6 +457,7 @@ int main(int argc, char** argv)
   }
 
   const auto cli_config = app.config_to_str(true, false);
+  LOG_INFO_FMT("Version: {}", ccf::ccf_version);
```
jumaffre, line 460: This is the only change to the shipped CCF binaries, which I believe is harmless but useful for operators when performing code upgrades.

achamayou (approved):

PR #2669, opened 2021-06-15
Add documentation for reconfiguration

achamayou:
image


achamayou:
Notes from discussion with @ashamis wrt to the 2 transaction scheme, for discussion with @wintersteiger tomorrow:

1. On voting (3.42), only Cfg 0 is active
2. Promote is triggered once per node, upon that node becoming aware of 3.42 being committed in Cfg 0.
3. Promote transitions a node from Learner to UpToDateLearner in KV
4. The transaction that promotes the 2f+1st Learner to UpToDateLearner is the new *reconfiguration transaction*, and Active Configs becomes [Cfg 0, Cfg 1] at that transaction.
5. Only [Cfg 1] is active for later transactions.

The re-sharing would be kicked off once 3.42 is committed in Cfg 0, and nodes would only trigger a Promote once they meet condition 2. AND they observe through the ledger that the re-sharing is complete.

ashamis (approved):

fritzalder:
I made some rewordings and added some text on leader retirement. Please check for correctness and wording/consistency with the rest of CCF's language.

Issue #2655, opened 2021-06-10
Failures in full_test_suite in jwt_test


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=26871&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=d37f6062-24ee-5eff-15b2-0eb74651c675
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=26906&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=d3511c0a-aced-5990-4c7d-8e47aebb8200

```
42: 16:10:38.268 | ERROR    | __main__:run:117 - Test jwt_test.test_refresh_jwt_issuer failed
42: Traceback (most recent call last):
42: 
42:   File "/mnt/build/2/s/tests/e2e_suite.py", line 200, in <module>
42:     run(args)
42:     │   └ Namespace(bft_view_change_timeout_ms=5000, binary_dir='/mnt/build/2/s/build', ccf_version=None, client_connection_timeout_ms=...
42:     └ <function run at 0x7ff4a5b160d0>
42: 
42: > File "/mnt/build/2/s/tests/e2e_suite.py", line 107, in run
42:     new_network = test(network, args)
42:                   │    │        └ Namespace(bft_view_change_timeout_ms=5000, binary_dir='/mnt/build/2/s/build', ccf_version=None, client_connection_timeout_ms=...
42:                   │    └ <infra.network.Network object at 0x7ff4a5b04b80>
42:                   └ <function test_refresh_jwt_issuer at 0x7ff4a5b12790>
42: 
42:   File "/mnt/build/2/s/tests/suite/test_requirements.py", line 20, in wrapper
42:     return func(*args, **kwargs)
42:            │     │       └ {}
42:            │     └ (<infra.network.Network object at 0x7ff4a5b04b80>, Namespace(bft_view_change_timeout_ms=5000, binary_dir='/mnt/build/2/s/buil...
42:            └ <function test_refresh_jwt_issuer at 0x7ff4a5b12700>
42: 
42:   File "/mnt/build/2/s/tests/jwt_test.py", line 24, in test_refresh_jwt_issuer
42:     network.jwt_issuer.wait_for_refresh(network)
42:     │       │          │                └ <infra.network.Network object at 0x7ff4a5b04b80>
42:     │       │          └ <function JwtIssuer.wait_for_refresh at 0x7ff4a5c94670>
[...]

PR #2674, opened 2021-06-15
Increase JWT refresh timeout in JWT refresh test

jumaffre:
Potential fix for #2655


jumaffre:
Alternative and better fix in #2694

PR #2678, opened 2021-06-16
Use LTS version of JS constitution

jumaffre:
As pointed in https://github.com/microsoft/CCF/pull/2643#issuecomment-859715259, we now use the constitution JS scripts from the LTS rather than from our local checkout, so that we don't break the test by adding new features to the constitution (e.g. addition of JS KV map's `clear()` function).

### DIFF
--- a/tests/lts_compatibility.py
+++ b/tests/lts_compatibility.py
@@ -48,6 +48,38 @@ def get_bin_and_lib_dirs_for_install_path(install_path):
     )
 
 
+def set_js_args(args, from_install_path):
+    # Use from_version's app and constitution as new JS features may not be available
+    # on older versions
+    js_app_directory = (
+        "../samples/apps/logging/js"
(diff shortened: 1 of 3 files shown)


tests/lts_compatibility.py
```diff
@@ -48,6 +48,38 @@ def get_bin_and_lib_dirs_for_install_path(install_path):
...
+    def replace_constitution_fragment(args, fragment_name):
+        args.constitution[:] = [
+            os.path.join(constitution_directory, fragment_name)
+            if fragment_name in f
+            else f
+            for f in args.constitution
+        ]
+
+    # Note: Use resolve.js script from local checkout as only the trivial sandbox
+    # version is included in installation
```
jumaffre, line 77: I'm in favour of shipping the default `resolve.js` script as well as the pass-through sandbox one. Any objections?

achamayou (approved):

Issue #2679, opened 2021-06-16
BFT recovery re-share


achamayou:
Re-sharing with sss currently requires entropy, which leads to re-execution failure in a BFT setting.

We could either fix the seed (digest of the ledger secrets?), or perform execution only on the primary and make the transaction a special case on the followers.

PR #2682, opened 2021-06-17
Change BFT to record all kv entries to the ledger

ashamis:
- There is an issue of with TPCC so it is being removed https://github.com/microsoft/CCF/issues/2662
- Do no record the shares table due to non-determinism https://github.com/microsoft/CCF/issues/2679

### DIFF
--- a/src/consensus/aft/raft_tables.h
+++ b/src/consensus/aft/raft_tables.h
@@ -10,8 +10,8 @@
 namespace aft
 {
   // The KV provides an option to select which tables are replicated and which
-  // are not. This feature is not used but should be required for Byzantine
-  // Identity. https://github.com/microsoft/CCF/issues/2683
+  // are not. This feature is currently not used but should be required for
+  // Byzantine Identity. https://github.com/microsoft/CCF/issues/2683
   static constexpr auto replicate_type = kv::ReplicateType::ALL;
   static const std::unordered_set<std::string> replicated_tables = {};
 }
--- a/src/consensus/aft/raft_tables.h
+++ b/src/consensus/aft/raft_tables.h
@@ -9,14 +9,6 @@
 
 namespace aft
 {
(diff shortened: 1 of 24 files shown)


wintersteiger (approved):
Do I understand correctly that the main change here is that failed requests are now recorded where they weren't before?

src/consensus/aft/raft_tables.h
```diff
@@ -9,14 +9,6 @@
...
-  static const std::unordered_set<std::string> replicated_tables_raft = {};
-
-  static constexpr auto replicate_type_bft = kv::ReplicateType::SOME;
-  static const std::unordered_set<std::string> replicated_tables_bft = {
-    ccf::Tables::AFT_REQUESTS,
-    ccf::Tables::SIGNATURES,
-    ccf::Tables::BACKUP_SIGNATURES,
-    ccf::Tables::NONCES,
-    ccf::Tables::NEW_VIEWS};
+  static constexpr auto replicate_type = kv::ReplicateType::ALL;
```
jumaffre, line 12: Can we remove all the table replication filtering logic in the KV as well, and remove these types altogether? If we plan to use them again in the future, we should add a comment about that here.

src/kv/store.h
```diff
@@ -842,6 +831,21 @@ namespace kv
             std::move(changes),
             std::move(new_maps));
         }
+        else if (
```
jumaffre, line 834: Can we add a comment/link to https://github.com/microsoft/CCF/issues/2679 here?

achamayou (approved):

ashamis:
> Do I understand correctly that the main change here is that failed requests are now recorded where they weren't before?

In this change a BFT network will record all writes to the key-value store. So in this change not committed requests will still not be recorded.

jumaffre (approved):

PR #2688, opened 2021-06-17
Avoid deadlocked channel state after interrupted key exchange

### LINKED ISSUE #2864: raft_test undefined behaviour
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=30111&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=d3511c0a-aced-5990-4c7d-8e47aebb8200

```
8: 2021-08-03T21:54:55.964617Z        100 [info ] ../src/consensus/aft/raft.h:2552     | Send request vote from n[A] to n[B] at 0
8: ../src/consensus/aft/test/test_common.h:77:26: runtime error: load of value 3, which is not a valid value for type 'bool'
8: SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/consensus/aft/test/test_common.h:77:26 in 
```

eddyashton:
This resolves #2864.

The meaningful changes here are 2 removals in `channels.h`: handle channel initiation (on either side) as a restart when we were in state `WAITING_FOR_FINAL`.

I've added a unit test for this which tries dropping each of the key exchange messages, and tries to initiate a fresh key exchange and transmit messages after that. It fails without the `channels.h` changes, and passes with them.

I have an inkling that there's something else a little funky here. When we're re-establishing a new key exchange on a previously-successful channel, we remain in state `ESTABLISHED` and take some other early-out paths. I think this means we can get into a similar position if later messages are dropped (ie, if we `send` during a key re-establishment, what key is used on each end?). I'd like to add another unit test which exercises this, but don't think it needs to be bundled with this PR.

### DIFF
--- a/src/node/test/channels.cpp
+++ b/src/node/test/channels.cpp
@@ -18,7 +18,7 @@
 threading::ThreadMessaging threading::ThreadMessaging::thread_messaging;
 std::atomic<uint16_t> threading::ThreadMessaging::thread_count = 0;
 
-constexpr auto buffer_size = 1024 * 16;
+constexpr auto buffer_size = 1024 * 8;
 
 auto in_buffer_1 = std::make_unique<ringbuffer::TestBuffer>(buffer_size);
 auto out_buffer_1 = std::make_unique<ringbuffer::TestBuffer>(buffer_size);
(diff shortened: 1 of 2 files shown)


src/node/test/channels.cpp
```diff
@@ -18,7 +18,7 @@
 threading::ThreadMessaging threading::ThreadMessaging::thread_messaging;
 std::atomic<uint16_t> threading::ThreadMessaging::thread_count = 0;
 
-constexpr auto buffer_size = 1024 * 16;
+constexpr auto buffer_size = 1024 * 8;
```
eddyashton, line 21: See the comment below about `ringbuffer::read()`. I was getting some very strange failures in this unit test, until I worked out it was just missing a message because it had reached the end of the buffer, and needed `read()` to be called again. So we now call it twice, and prove that's sane by reducing `buffer_size` (and seeing that other tests now fail without the double-read).

eddyashton:
So it seems this fix is too aggressive, and introduces a new failure case: if 2 nodes try to communicate at around the same time, one is likely to initiate (losing state) just after the other believes the channel is established, when it would previously have waited in state `WAITING_FOR_PRIMARY`. So we again end up with a one-way dead channel. I'll see if there's a small fix for this, but its possible we need timeouts + error responses on the node channel to be more robust here.

achamayou (approved):

eddyashton:
Closing - there's a larger fix on its way which supercedes this. The tests from this PR are included in that branch.

Issue #2519, opened 2021-04-27
Expose JWT keys and service configuration via RPCs


achamayou:
Although there is no user demand right now, it is probably useful in general, and it certainly is for tests in particular to have:

1. a way to list active JWT keys
2. a way to extract current service configuration

We rely on reading the ledger offline in our tests in both cases, which is quite inefficient, and there is a very good chance an operator would want these RPCs.

achamayou:
2. is addressed by `/node/consensus`, but 1. remains an issue (#2655, #2674).

Issue #2671, opened 2021-06-15
Support bare keys in JWKS for JWT auth


letmaik:
Currently, CCF only accepts JWKS documents where each key has an `x5c` field. While this is the case for Microsoft, it is not required by the [spec](https://openid.net/specs/openid-connect-discovery-1_0.html):
> The JWK x5c parameter MAY be used to provide X.509 representations of keys provided. When used, the bare key values MUST still be present and MUST match those in the certificate.

An example where only bare keys are used: https://accounts.google.com/.well-known/openid-configuration

At the moment, CCF converts `x5c` into DER encoding and then stores that in the kv. At validation time, it assumes a DER-encoded X.509 cert was stored.

An alternative may be to store a DER-encoded public key instead. This key would have to be created from the individual fields in the JWK, like `kty`, `n`, and `e`. To remain backwards compatible, CCF would have to support both variants for a while. Unfortunately, DER is not self-describing, so it's not possible to easily figure out if something is a public key or cert. An option may be to try parsing as public key first, and fall-back to cert parsing. The fall-back would only be used until the next (auto-)refresh of the keys.

achamayou:
@letmaik is this a blocker for a particular use case?

letmaik:
Google uses bare keys for example, but I don’t have that use case myself.
https://www.googleapis.com/oauth2/v3/certs

achamayou:
Closing for now, since there is no immediate demand for this.

PR #2694, opened 2021-06-18
Add JWT keys endpoint and use it in tests

achamayou:
Part of #2519, preferable to #2674 in my opinion.

### DIFF
--- a/src/node/rpc/member_frontend.h
+++ b/src/node/rpc/member_frontend.h
@@ -488,7 +488,7 @@ namespace ccf
       openapi_info.description =
         "This API is used to submit and query proposals which affect CCF's "
         "public governance tables.";
-      openapi_info.document_version = "1.0.0";
+      openapi_info.document_version = "1.1.0";
     }
 
     static std::optional<MemberId> get_caller_member_id(
@@ -944,6 +944,24 @@ namespace ccf
         json_adapter(refresh_jwt_keys),
         {std::make_shared<NodeCertAuthnPolicy>()})
         .set_openapi_hidden(true)
+        .install();
+
+      using JWTKeyMap = std::map<JwtKeyId, Cert>;
+
+      auto get_jwt_keys = [this](auto& ctx, nlohmann::json&& body) {
+        auto keys = ctx.tx.template ro<JwtPublicSigningKeys>(
+          ccf::Tables::JWT_PUBLIC_SIGNING_KEYS);
+        JWTKeyMap kmap;
+        keys->foreach([&kmap](const auto& kid, const auto& kpem) {
+          kmap[kid] = kpem;
+          return false;
+        });
+
+        return make_success(kmap);
+      };
+      make_endpoint(
+        "jwt_keys/all", HTTP_GET, json_adapter(get_jwt_keys), no_auth_required)
+        .set_auto_schema<void, JWTKeyMap>()
         .install();
 
 #pragma clang diagnostic push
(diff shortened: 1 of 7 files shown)


src/node/rpc/member_frontend.h
```diff
@@ -946,6 +946,24 @@ namespace ccf
         .set_openapi_hidden(true)
         .install();
 
+      using JWTKeyMap = std::map<JwtKeyId, Cert>;
+
+      auto get_jwt_keys = [this](auto& ctx, nlohmann::json&& body) {
```
jumaffre, line 951: We should also use this new endpoint in `jwt_issuer.py:wait_for_refresh`

src/node/rpc/member_frontend.h
```diff
@@ -946,6 +946,24 @@ namespace ccf
...
+        JWTKeyMap kmap;
+        keys->foreach([&kmap](const auto& kid, const auto& kpem) {
+          kmap[kid] = kpem;
+          return false;
+        });
+
+        return make_success(kmap);
+      };
+      make_endpoint(
+        "jwt_keys/all", HTTP_GET, json_adapter(get_jwt_keys), no_auth_required)
```
jumaffre, line 963: Why not just `jwt_keys`?

src/node/rpc/member_frontend.h
achamayou: I'd missed, but yes, absolutely.

src/node/rpc/member_frontend.h
achamayou: No strong reason, other than there's precedent for /all elsewhere it seemed like it left more room for other GET endpoints at the same level (/issuers etc).

Happy to defer to a REST-qualified expert :)

src/node/rpc/member_frontend.h
```diff
@@ -946,6 +946,24 @@ namespace ccf
...
         .install();
 
+      using JWTKeyMap = std::map<JwtKeyId, Cert>;
+
+      auto get_jwt_keys = [this](auto& ctx, nlohmann::json&& body) {
+        auto keys = ctx.tx.template ro<JwtPublicSigningKeys>(
+          ccf::Tables::JWT_PUBLIC_SIGNING_KEYS);
+        JWTKeyMap kmap;
+        keys->foreach([&kmap](const auto& kid, const auto& kpem) {
+          kmap[kid] = kpem;
```
jumaffre, line 956: Could we bundle the issuer here as well, by looking up the `public:ccf.gov.jwt.public_signing_key_issuer` table?

tests/jwt_test.py
```diff
@@ -28,13 +28,20 @@ def test_refresh_jwt_issuer(network, args):
     return network
 
 
+def make_pem_cert(pem_encoded):
```
jumaffre, line 31: Can't we just re-use the existing `infra.crypto.cert_der_to_pem` function?

achamayou: It's not particularly convenient, we have to decode the string first. If anything I was more leaning towards making the endpoint do the wrapping so we return proper PEM certs.

src/node/rpc/member_frontend.h
achamayou: We could, but we won't until there's a use case for it.

tests/jwt_test.py
jumaffre: Returning PEM directly sounds good!

letmaik: +1 on PEM, as that allows to serve public key PEMs as well (if they should come in the future, see #2671) and keep a single response schema,

letmaik (approved):

tests/jwt_test.py
achamayou: Yes I've done that now.

jumaffre (approved):

PR #2695, opened 2021-06-18
Release 1.0.4, take 2

achamayou:
Includes #2693

### DIFF
--- a/tests/lts_compatibility.py
+++ b/tests/lts_compatibility.py
@@ -186,14 +186,18 @@ def run_live_compatibility_with_latest(args, repo, local_branch):
     lts_version, lts_install_path = repo.install_latest_lts_for_branch(
         os.getenv(ENV_VAR_LATEST_LTS_BRANCH_NAME, local_branch)
     )
-    LOG.info(f'From LTS {lts_version} to local "{local_branch}" branch')
-    run_code_upgrade_from(
-        args,
-        from_install_path=lts_install_path,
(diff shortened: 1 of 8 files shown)


tests/lts_compatibility.py
```diff
@@ -259,65 +267,74 @@ def run_ledger_compatibility_since_first(args, use_snapshot):
...
-            # Check that the ledger can be parsed
-            ccf.ledger.Ledger([committed_ledger_dir]).get_latest_public_state()
+                # Check that ledger and snapshots can be parsed
+                ccf.ledger.Ledger([committed_ledger_dir]).get_latest_public_state()
+                if snapshot_dir:
+                    for s in os.listdir(snapshot_dir):
+                        with ccf.ledger.Snapshot(
+                            os.path.join(snapshot_dir, s)
+                        ) as snapshot:
+                            snapshot.get_public_domain()
```
jumaffre, line 337:
```suggestion
```
This will fail because `1.x`'s Python doesn't know about `Snapshot`.

eddyashton (approved):

tests/lts_compatibility.py
achamayou: Ok, so we need to cherry-pick #2644?

jumaffre: Either that or simply skip this check here as this is test-only?

jumaffre (approved):

PR #2707, opened 2021-06-22
Add more design notes on node-to-node channels

eddyashton:
Adds some diagrams of the existing flow, and brief discussion/flowchart for proposed new flow.


doc/contribute/design/index.rst
```diff
@@ -56,14 +58,181 @@ Encryption keys are automatically refreshed periodically (in line with https://d
...
         Backup->>+Primary: Consensus headers response
 
         Backup->>+Primary: Forwarded client HTTP request <br> (encrypted with channel key)
         Primary->>+Backup: Forwarded client HTTP response
 
+This diagram shows the state machine for a channel on each node, with the messages that trigger each transition.
+
+.. mermaid::
+
+    stateDiagram-v2
```
eddyashton, line 71: image

doc/contribute/design/index.rst
```diff
@@ -56,14 +58,181 @@ Encryption keys are automatically refreshed periodically (in line with https://d
...
-        
+        %%INITIATED --> INACTIVE: close()
+        %%WAITING_FOR_PRIMARY --> INACTIVE: close()
+        %%ESTABLISHED --> INACTIVE: close()
+
+The brief sequence diagram of a successful key exchange is as follows.
+
+.. mermaid::
+
+    sequenceDiagram
```
eddyashton, line 88: image

doc/contribute/design/index.rst
```diff
@@ -56,14 +58,181 @@ Encryption keys are automatically refreshed periodically (in line with https://d
...
+        Note over Node1: state := Established
+        Node1->>+Node2: key_exchange_final
+
+        Note over Node2: state := Established
+
+However, if messages are dropped this protocol can reach various deadlocked states. Once the channel is established we are able to process dropped and reordered messages on the receiving end, so we should also be robust to these during channel establishment. Even with perfect network conditions and honest hosts, if we want to support legitimate closure of channels then we should be robust to this happening during key establishment. The following diagram shows some of the issues when key exchange messages are dropped.
+
+.. mermaid::
+
+    sequenceDiagram
```
eddyashton, line 107: image

doc/contribute/design/index.rst
```diff
@@ -56,14 +58,181 @@ Encryption keys are automatically refreshed periodically (in line with https://d
...
+                    end
+                end
+            end
+        end
+
+To be robust to this, I think we need to reason about what a node should do in response to each type of message, in each state. An initial approach to this is summarised by the following proposed flowchart.
+
+.. mermaid::
 
+    graph TD
```
eddyashton, line 179: image

doc/contribute/design/index.rst
achamayou: s/WAITING_FOR_PRIMARY/WAITING_FOR_FINAL/ presumably?

achamayou (approved):

doc/contribute/design/index.rst
```diff
@@ -56,14 +58,181 @@ Encryption keys are automatically refreshed periodically (in line with https://d
...
+
+This doesn't yet deal with key rotation, and I believe we need an establishment-attempt nonce to prevent replay attacks that could re-establish an old, overused key. While the summary above says we start a fresh key-exchange protocol, the actual implementation attempts to do this in parallel with encrypted messages over an existing channel. If we build a protocol which can reliably deal with legitimate reconnection attempts, it would be preferable to re-use that for key rotation - deliberately close an existing connection and start fresh with a new channel - rather than building an additional rotation protocol.
+
+An open question here is whether we need to handle channel closures and re-opening. If we don't, then we can consider all dropped messages here as malicious DoS, but perhaps simplify the protections. One option is to avoid ever closing a channel - if a node has opened a channel to another node, it remains communicating with them forever. Another is to determine a point after which a channel can be safely closed - if the channel is used only for consensus and not forwarding, it may be safe to close after a node's retirement.
```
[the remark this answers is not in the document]
eddyashton: So I'm not 100% on this, but I see 2 reasons why making these channels consensus-only doesn't give us a win:

- The point where we can close a channel (we're done talking to them, and they should never try to talk to us), is later than anything that we currently track. It is when they have ACK'd global commit of their own retirement. At any time before they know their retirement _is globally committed_, they may become candidate and still ask us for votes + require us (and other nodes) to help them rediscover the global commit index.
- I think we should implement key rotation as fresh key exchange, which means we need to handle multiple key exchanges  from the same node, which means we have at least once case of legitimate channel closure and re-opening ("they want to use a new key"). It's possible we can somehow design key rotation as a separate protocol over an existing channel, but I don't see an easy way to do so and it would mean designing+verifying+maintaining 2 protocols.

PR #2711, opened 2021-06-22
Reduce the number of view-changes that occur when a new config is added

ashamis:
Part of #1852

BFT maintains a list of ordered nodes to ensure view-changes are deterministic. This logic needs to also be maintained during reconfiguration. This change adds an offset into every configuration so that if possible the primary will not change between configurations. A complication arises in determining at what view a replica sees a reconfiguration because if a replica is catching up the reconfiguration could have occurred in a different view to the one that the current primary is sending append entries from. To solve this we look at the last committed view-change, this view-change also serves as a deterministic marker on how the offset is created (for auditing purposes)


src/consensus/aft/impl/view_change_tracker.h
```diff
@@ -239,7 +240,7 @@ namespace aft
       }
 
       auto& vc = it->second;
-      ccf::ViewChangeConfirmation nv(vc.view);
+      ccf::ViewChangeConfirmation nv(vc.view, from);
```
achamayou, line 243: Is `from` the node id that issued the view change confirmation? Is that effectively the primary at the time?

achamayou:
I'm having a hard time working out what the test for this should look like, but I think we ought to have one. Perhaps get the primary from a config history, and obtain the same result from an arbitrarily truncated sequence with the correct initial view primary id?

achamayou (approved):

ashamis:
> I'm having a hard time working out what the test for this should look like, but I think we ought to have one. Perhaps get the primary from a config history, and obtain the same result from an arbitrarily truncated sequence with the correct initial view primary id?

Yes I agree with you on this - this is close to what we get with the recovery tests & and in part the rotation/partition tests

src/consensus/aft/impl/view_change_tracker.h
achamayou: can we rename that to primary?

Issue #2719, opened 2021-06-24
Upgrade to Clang 10


achamayou:
We are now using Open Enclave 0.17.0-rc1 (#2701), which is built with and enables clang-10. We want to:

- [x] Update code and cmake to be clang-10 compatible, but keep clang-8 default (#2736)
- [x] Update toolchain in playbooks and clang-format config for 10 (#2739)
- [x] Switch default to clang 10

The intention is to transition both main _and_ LTS to clang-10 after a successful release of the LTS against Open Enclave 0.17.1, to ease future patches and make the LTS benefit from CFG.

achamayou:
Enabling CFG (#819) requires turning on LTO, and will impact virtual mode which makes use of DSOs. We will look at it separately. We will also want to enable C++20 support.

PR #2720, opened 2021-06-24
Clang 10

achamayou:
Look into #2719, as discussed with @eddyashton, going to break this up into:

- A set of changes that allow CCF to build with 10, but still default to 10 (#2736)
- Playbook changes necessary to upgrade the toolchain and clang format config, to be parked for now

The first one can go in now, the second will wait until after we've gone through mandatory patches coming from Open Enclave.


eddyashton (approved):
Does `clang-format-10` cause any formatting changes to our code? If it does, we should choose a point to do a single formatting pass and then start using the new toolchain everywhere.

src/node/code_id.h
```diff
@@ -15,7 +15,7 @@ namespace ccf
     std::array<uint8_t, crypto::Sha256Hash::SIZE> data;
 
     CodeDigest() = default;
-    CodeDigest(const CodeDigest& other) : data(other.data) {}
+    CodeDigest(const CodeDigest&) = default;
```
ashamis, line 18: Doesn't this change the logic here?

src/node/node_signature.h
```diff
@@ -20,11 +20,7 @@ namespace ccf
     /// Hashed nonce created by the node, only used for BFT
     Nonce hashed_nonce;
 
-    NodeSignature(const NodeSignature& ns) :
-      sig(ns.sig),
-      node(ns.node),
-      hashed_nonce(ns.hashed_nonce)
-    {}
+    NodeSignature(const NodeSignature& ns) = default;
```
ashamis, line 23: Doesn't this change the logic here?

achamayou: How so?

src/node/code_id.h
achamayou: No?

src/node/node_signature.h
ashamis: sorry my mistake

ashamis (approved):

achamayou:
@eddyashton I totally missed that, I'll find out now.

Issue #2721, opened 2021-06-24
BFT reconfiguration: 2 node primary retirement sequence diagram


ashamis:
Add a sequence diagram to the consensus protocol docs that illustrates the behaviour of the 2-tx protocol when a primary is retired.

wintersteiger:
Fixed by #2731

PR #2731, opened 2021-06-24
2-tx reconfiguration protocol updates

wintersteiger:
This adds the latest changes to the 2-tx reconfiguration protocol to the documentation. Also includes a diagram for the 2-tx leader retirement. Addresses #2721. 

image
image
image

### DIFF
--- a/doc/overview/consensus.rst
+++ b/doc/overview/consensus.rst
@@ -160,13 +160,15 @@ This sample illustrates the addition of a single node to a one-node network with
         Note right of Node 0: Active configs := [Cfg 0]
         Node 0-->>-Members: Success
 
+        Node 1->>+Node 0: Poll join
+        Node 0-->>-Node 1: Learner
 
         Node 0->>Node 1: Replicate Tx ID 3.42
         Note over Node 1: State in KV := LEARNER
         Node 1->>Node 0: Acknowledge Tx ID 3.42
 
         Node 0->>Node 1: Notify commit 3.42
-        
+
         Node 1->>+Node 0: Up-to-date RPC for Node 1
         Note over Node 0: Node 1 in KV := UP_TO_DATE_LEARNER
         Note over Node 0: All nodes in Cfg 1 in KV := TRUSTED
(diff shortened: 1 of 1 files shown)


wintersteiger:
This is currently missing the second transaction that finalizes the reconfiguration. For that, I think we'll need either a RETIRING state or we could remove the node from the nodes table after it retired.

ashamis:
> This is currently missing the second transaction that finalizes the reconfiguration. For that, I think we'll need either a RETIRING state or we could remove the node from the nodes table after it retired.

Do we need to do that? Can we not just remove the node?

wintersteiger:
We don't have to, but technically it's not a two-transaction reconfiguration if the second one is missing. It may complicate the logic a little as we'd have to check whether another transaction is expected, before switching to a new configuration. We could also issue an empty transaction if that's possible.

wintersteiger:
@fritzalder we've changed the 2-tx reconfiguration protocol today to add the `RETIRING` state and we changed to observation RPCs (ORC) from all nodes in the new configuration instead of promotion RPCs from the learners. Could you take a quick look at these diagrams and see whether this makes sense?

fritzalder (approved):
New diagrams look good!

fritzalder:
I like the new diagrams but these changes also invalidate the next section: "Replica State Machine". It says "Main states and transitions in CCF consensus. Note that while the implementation of the transitions differs between CFT and BFT, the states themselves do not."

I suggest:
1. Reorganize replica state machine section and split up CFT/BFT clearly and add the new state transitions
2. Add a node state overview at the top of the page with a short explanation what each state does and what it is for. For example, right now it is fuzzy what retiring entails. We could make this more clear. 

Not sure whether you want these doc changes in the same PR or in a new PR (since they all still relate to this change).

achamayou:
@fritzalder I think the claim about same states sort of holds, to the extent that some of states (Learner and Retiring) simply aren't used (or are skipped through if you prefer) in CFT. Perhaps it is better to rephrase this as the BFT states are a super-set of the CFT states?

I would like to merge this and some of the implementation steps, but I do agree that a doc reorg is necessary, since this is clearly far too detailed for the Overview to begin with.

achamayou (approved):

fritzalder:
I see your point. I can make a separate PR at the end of this week that splits up the consensus page into three pages (one overview/commonalities and one for each protocol). That would also make it easier to get the concepts of CFT/BFT without needing to read the papers or be an expert, and it would prevent potential confusion between the two protocols. Would that be an idea?

wintersteiger:
That's great, thanks for reviewing this @fritzalder! I agree that this is getting a bit too detailed for the overview and I like your suggestion of splitting this up. I'll start work on the implementation now, but I can also help with the docs later in the week.

PR #2740, opened 2021-06-25
Add learner state, logic, and test for two-tx reconfig

wintersteiger:
This adds the new learner state and logic for the two transaction reconfiguration. Also enables the reconfiguration tests for BFT, but currently runs only one test, which checks that a learner correctly catches up with the network. (But no promotion to trusted yet.)


doc/schemas/node_openapi.json
```diff
@@ -465,7 +466,7 @@
   "info": {
     "description": "This API provides public, uncredentialed access to service and node state.",
     "title": "CCF Public Node API",
-    "version": "1.3.0"
+    "version": "1.3.1"
```
achamayou, line 469: This must be 1.4.0, not 1.3.1 according to SemVer.

src/consensus/aft/raft.h
```diff
@@ -243,6 +245,12 @@ namespace aft
         // Initialize view history for bft. We start on view 2 and the first
         // commit is always 1.
         state->view_history.update(1, starting_view_change);
+        use_two_tx_reconfig = true;
+      }
+
+      if (use_two_tx_reconfig && !public_only)
```
achamayou, line 251: That doesn't seem right to me, whether the node thinks it is in public_only mode or not should have no bearing on whether it becomes learner. It's to do with whether the service is open or not already, regardless of whether it is in recovery or startup mode.

wintersteiger: That's what I thought too. It seems the `public_only` flag is only used for recovery, but I may be wrong about that. Are there other flags that let us determine whether we should be starting as a learner or follower?

src/consensus/aft/raft.h
```diff
@@ -2627,7 +2679,9 @@ namespace aft
 
       is_new_follower = true;
 
-      if (replica_state != kv::ReplicaState::Retired)
+      if (
+        replica_state != kv::ReplicaState::Retired &&
```
achamayou, line 2683: I think we should have calls like takes_part_in_replication() and takes_part_in_election(), and the state conditions inside, for better readability.

tests/reconfiguration.py
```diff
@@ -353,29 +379,36 @@ def run(args):
...
-        with open(latest_node_log, "r+") as log:
-            assert any(
-                "No snapshot found: Node will replay all historical transactions" in l
-                for l in log.readlines()
-            ), "New nodes shouldn't join from snapshot if snapshot evidence cannot be verified"
-
-        test_node_filter(network, args)
-        test_retiring_nodes_emit_at_most_one_signature(network, args)
+
+        if args.consensus == "cft":
```
achamayou, line 384: Historically we've made an exception for "!bft" rather than "cft", and it's helpful to stay consistent because we think of grepping for "bft" to find exceptions, but not for "cft".

src/consensus/aft/raft.h
achamayou: It is only used for recovery, but recovery is not what decides if the node can skip learner or not. It's whether the network is pre-OPEN, which can happen both at startup and in recovery. The return of the join RPC from the network is what tells the node what state it can enter (currently either PENDING or TRUSTED).

tests/reconfiguration.py
```diff
@@ -341,6 +344,29 @@ def test_retiring_nodes_emit_at_most_one_signature(network, args):
     return network
 
 
+@reqs.description("Adding a learner without snapshot")
+def test_learner_catches_up(network, args):
```
jumaffre, line 349: We should check that the new node doesn't count in the election quorum or for commit. I think the best way for this would be to suspend a majority of followers (expect the new node) and make sure the commit is stuck, and that no new node can be elected as primary when the primary is suspended/killed.

src/node/node_state.h
```diff
@@ -541,7 +542,9 @@ namespace ccf
           }
 
           // Set network secrets, node id and become part of network.
-          if (resp.node_status == NodeStatus::TRUSTED)
+          if (
```
jumaffre, line 545: As discussed, we need to make sure that we cover all places where we currently nodes as `TRUSTED`, e.g. in `secret_broadcast.h`.

src/consensus/aft/raft.h
wintersteiger: -> join RPC will tell us

src/consensus/aft/raft.h
```diff
@@ -298,7 +306,8 @@ namespace aft
     bool can_replicate()
     {
       std::unique_lock<std::mutex> guard(state->lock, std::defer_lock);
-      if (!(consensus_type == ConsensusType::BFT && is_follower()))
+      if (!(consensus_type == ConsensusType::BFT &&
```
wintersteiger, line 309: !is_leader()

wintersteiger: lock_unless_bft_rexec

src/consensus/aft/raft.h
```diff
@@ -2601,6 +2650,17 @@ namespace aft
       }
     }
 
+    bool takes_part_in_replication()
+    {
+      return replica_state != kv::ReplicaState::Retired;
+    }
+
+    bool takes_part_in_election()
```
wintersteiger, line 2658: can_endorse_primary

src/consensus/aft/raft.h
```diff
@@ -2601,6 +2650,17 @@ namespace aft
       }
     }
 
+    bool takes_part_in_replication()
```
wintersteiger, line 2653: can_advance_watermark

wintersteiger: + != learner

src/consensus/aft/raft.h
```diff
@@ -507,8 +532,33 @@ namespace aft
...
       configurations.push_back({idx, std::move(conf), offset});
+      if (use_two_tx_reconfig)
+      {
+        if (learners.size() > 0)
+        {
+          LOG_DEBUG_FMT(
+            "Configurations: learners {{{}}}", fmt::join(learners_, ", "));
+          learners.insert(learners_.begin(), learners_.end());
+        }
+        else if (!learners.empty())
```
wintersteiger, line 543: move branch dowwn to !use_two...

src/consensus/aft/raft.h
```diff
@@ -507,8 +532,33 @@ namespace aft
...
+          // At the very start of a fresh network, there is no leader to commit
+          // promotions. There's also nothing to catch up to, so we become a
+          // follower immediately. We could also pass a flag to Aft<>::Aft() to
+          // indicate that we're still starting up.
+          replica_state = kv::ReplicaState::Follower;
+        }
+      }
       backup_nodes.clear();
-      create_and_remove_node_state();
+      create_and_remove_node_state(idx);
```
wintersteiger, line 561: -idx

src/consensus/aft/raft.h
```diff
@@ -2646,11 +2706,33 @@ namespace aft
...
+
+        if (cfg.find(from) == cfg.end())
+        {
+          LOG_INFO_FMT("Ignoring vote from ineligible voter {}", from);
+          return;
+        }
+
+        // Need 50% + 1 of the total nodes in the current config (including us).
+        votes_for_me.insert(from);
+        quorum = (cfg.size() / 2) + 1;
```
wintersteiger, line 2723: enough_trusted

src/consensus/aft/raft.h
```diff
@@ -2800,14 +2911,46 @@ namespace aft
...
+            if (!enough_trusted(*next))
+            {
+              LOG_TRACE_FMT(
+                "Configurations: not enough trusted nodes for next "
+                "configuration");
+              break;
+            }
+            else
+            {
+              // trigger promotion of whole configuration here?
```
wintersteiger, line 2933: Not here!

src/node/network_tables.h
```diff
@@ -79,6 +80,7 @@ namespace ccf
     // Node table
     //
     Nodes nodes;
+    NetworkConfigurations network_configurations;
```
wintersteiger, line 83: remove table, add ID to nodes table

src/node/network_configurations.h
```diff
@@ -0,0 +1,16 @@
...
+#include "entities.h"
+#include "kv/map.h"
+#include "node_info_network.h"
+#include "service_map.h"
+
+#include <set>
+
+namespace ccf
+{
+  using NetworkConfigurations = ServiceMap<size_t, std::unordered_set<NodeId>>;
```
wintersteiger, line 15: seq_no instead of ID

tests/reconfiguration.py
```diff
@@ -446,4 +480,6 @@ def run_join_old_snapshot(args):
     args.initial_user_count = 1
 
     run(args)
-    run_join_old_snapshot(args)
+
+    if args.consensus == "cft":
```
achamayou, line 484:
```suggestion
    if args.consensus != "bft":
```

wintersteiger: Thanks, missed that one!

tests/reconfiguration.py
wintersteiger: Both added!

src/node/node_state.h
wintersteiger: I took another look and added learners in `secret_broadcast.h` and in another place in `node_frontend.h`. I think this covers all the cases we need, but I'd be grateful if you could take another look to make sure I'm not giving secrets to nodes who aren't supposed to get them.

src/consensus/aft/raft.h
```diff
@@ -193,7 +196,8 @@ namespace aft
       std::chrono::milliseconds election_timeout_,
       std::chrono::milliseconds view_change_timeout_,
       size_t sig_tx_interval_ = 0,
-      bool public_only_ = false) :
+      bool public_only_ = false,
+      bool join_as_learner_ = false) :
```
achamayou, line 200: I would mildly prefer if we passed initial_state here, because we could perhaps kill off the force_become_leader() that way, but I'm happy for that to be a separate PR, since this is purely cosmetic.

achamayou (approved):

src/consensus/aft/raft.h
wintersteiger: No problem, that's an easy change and it makes sense!

Issue #2651, opened 2021-06-10
/node/quotes/self fails when some of the collateral becomes expired


achamayou:
`/node/quotes/self` validates the quote to extract claims such as mrenclave. When some of the collateral expires (eg. the CRL), even though it had been valid at join time, users are unable to extract claims and the endpoint returns an error.

The right fix would be to validate the collateral against the timestamps of the quote in this context, rather than the current time, but oe_verify_evidence() doesn't provide that functionality. For now we can parse those claims at join time, store them in the ledger and retrieve them from there.

PR #2742, opened 2021-06-28
Cache code digest

achamayou:
Workaround for #2651, while waiting for https://github.com/openenclave/openenclave/issues/4086 to be fixed.


src/node/rpc/node_frontend.h
```diff
@@ -461,9 +465,24 @@ namespace ccf
...
+          auto nodes = args.tx.ro(network.nodes);
+          auto node_info = nodes->get(context.get_node_state().get_node_id());
+          if (node_info.has_value() && node_info->code_digest.has_value())
+          {
+            q.mrenclave = node_info->code_digest.value();
+          }
+          else
+          {
+            auto code_id =
+              EnclaveAttestationProvider::get_code_id(node_quote_info);
```
eddyashton, line 483: In the current case where this fails, it throws. Should we catch that and turn it into a nicer error?

achamayou: We could do, but this isn't the exact endpoint logic ACL will use anyway, so I think we should discuss it with them and find out what they prefer.

src/node/quote.h
```diff
@@ -212,6 +212,7 @@ namespace ccf
   public:
     static CodeDigest get_code_id(const QuoteInfo& quote_info)
     {
+      LOG_INFO_FMT("GETTING CODE ID");
```
eddyashton, line 215: If we move this log line lower, just before the return, we could log the code ID we retrieved and keep it as a useful logging line, not just for the temporary test?

eddyashton (approved):

src/node/quote.h
achamayou: Agreed but logging that as info seems quite verbose?

achamayou: This would be once per join, once per jwt refresh when using sgx filter, and every time the JS runtime calls verify as well, so as discussed, it's too verbose.

eddyashton (approved):

PR #2746, opened 2021-06-29
Remove websockets support

achamayou:
As discussed inside the team, there appears to be no real user-demand for WebSockets, we think our efforts are better spent elsewhere. This will reduce our TCB, and our CI/maintenance costs slightly to make room for other work.

I have left the RESTVerb abstraction, so we can (relatively) easily plug-in endpoints into alternative protocols in the future.


eddyashton (approved):
This needs a `CHANGELOG` entry

achamayou:
@eddyashton it does indeed.

eddyashton (approved):

ashamis (approved):

PR #2755, opened 2021-07-02
Fix reported view for read-only transaction after election

### LINKED ISSUE #2224: Reported term of read-only tx just after election is incorrect
As discussed with @eddyashton today, for read-only requests (typically `GET`), we currently report the `Tx`'s `term` (retrieved via `tx.commit_term()`)  in the HTTP response header. The `Tx`'s term is set on the `store` on rollback (when becoming primary or backup), to the consensus term to which we rollback to. 

However, the very first read-only `Tx`(s) issued by clients right after the election will report the wrong `TxID` as it will report the `seqno` of the last entry in the now-finished term, but with the _current_ term at which the transaction was executed.

This was identified when looking at (almost) unrelated [issue in the CI](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=19316&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=63d660c3-6f76-573a-0154-fd0bcb096ded&l=9803) and summarised here:

```
// seqno 129 is the last entry committed in term 4

// Killing primary
46: 13:20:59.428 | SUCCESS  | suite.test_requirements:wrapper:19 - Stopping current primary and waiting for a new one to be elected

// New primary 6 is elected in term 6...
46: 13:21:04.502 | INFO     | infra.network:find_primary:730 - [3||] GET /node/network
46: 13:21:04.502 | INFO     | infra.network:find_primary:730 - 200 @6.129 {"current_view":6,"primary_id":6,"service_status":"OPEN","view_change_in_progress":false}

// But @6.129 isn't a valid TxID!
```

In other words, if the user were to pass `6.129` to the `tx/status` endpoint, the result would come back as `Invalid`. 

Instead, the consensus should inform the store of the `term` of the `idx` at which it rolls back. The store should keep track of this "read term" and use it when a `Tx` grabs its `read_version` and `term` (implementation details deliberately vague for now, but it seems feasible in practice). 

Bundled with this change should be a slight refactoring of the `frontend.h` logic (see below) which has been prone to confusion and a source of multiple bugs. Instead, encapsulating most of the commit version/term logic in the `Tx` would be preferable (easier to write unit tests for it, etc.): 
https://github.com/microsoft/CCF/blob/7b6b60780dd55d53546e490596a4834b124d149e/src/node/rpc/frontend.h#L291-L300

jumaffre:
Fix #2224 

The transaction view reported for read-only transactions right after an election used to be incorrect (set to the current view, rather than the view at which the last transaction was committed in). This PR aims to fix this by separating the `term_of_next_version` (view at which future transactions should be committed in) from the `TxID` of valid transactions. On `rollback` the consensus, which keeps track of the view history, informs the store of the `TxID` at which to rollback to, as well as the `term_of_next_version` to use for future commits.

This is still very much WIP as the following items still need to be addressed: 
- [x] Changelog
- [x] Fix view of `0` reported on backup nodes on recovery
- [x] Fix signatures path
- [x] Discuss and address TODOs
- [x] Write end-to-end test 
- [x] How do backup nodes find out about the view of the entry they've just deserialised?

Follow-up items (separate PRs):
- [x] ~Acquire opacity `TxID` in transaction constructor rather than when the first handle is acquired so that all RPCs return a `TxID` for consistency, which also simplifies reasoning about the KV, code and testing.~ 
**We will not do this as this is a feature of the framework to reduce the number of conflicts when an expensive transaction that doesn't need to touch the KV is running**
- [x] ~Rename `term` to `view` in the store for consistency.~
**Let's skip this for now to avoid tricky merges when backporting these changes to the LTS branch**


src/consensus/aft/raft.h
```diff
@@ -3007,7 +3007,7 @@ namespace aft
       }
 
       snapshotter->rollback(idx);
-      store->rollback(idx, state->current_view);
+      store->rollback({get_term_internal(idx), idx}, state->current_view);
```
jumaffre, line 3010: The consensus, which knows about the view history, always tells the store which `TxID` to rollback to, so that further read-only transactions report the correct `TxID`, rather than the "{state->current_view}.{idx}", which is wrong.

src/kv/committable_tx.h
```diff
@@ -61,17 +61,14 @@ namespace kv
         return {};
       }
 
-      // Retrieve encryptor.
```
jumaffre, line 64: No need for this as every `Tx` now knows about its parent store already.

src/kv/committable_tx.h
```diff
@@ -193,6 +190,12 @@ namespace kv
           }
         }
 
+        if (version == NoVersion)
```
jumaffre, line 193: This is to return even earlier in case there's nothing to serialise, i.e. read-only Tx.

python/ccf/clients.py
```diff
@@ -152,6 +152,7 @@ def json(self):
         return json.loads(self._data)
 
 
+# TODO: Merge seqno/view into unique tx_id
```
achamayou, line 155: Not sure this is necessary here, since we would like to remove this wrapper altogether.

src/node/rpc/frontend.h
```diff
@@ -488,14 +484,20 @@ namespace ccf
     void set_root_on_proposals(
       const enclave::RpcContext& ctx, kv::CommittableTx& tx)
     {
-      if (ctx.get_request_path() == "/gov/proposals")
+      if (
```
jumaffre, line 487: As per the comment below, we should only ever use this for one specific endpoint, so let's make sure we proof this for later in case we add a new `GET /gov/proposals`

src/node/node_state.h
```diff
@@ -822,7 +822,6 @@ namespace ccf
       if (result == kv::ApplyResult::FAIL)
       {
         LOG_FAIL_FMT("Failed to deserialise entry in public ledger");
-        store->rollback(ledger_idx - 1);
```
jumaffre, line 825: No need for this as the store is about to be rolled back even further, to the previous signature.

src/kv/deserialise.h
```diff
@@ -75,7 +76,7 @@ namespace kv
         return ApplyResult::FAIL;
       }
 
-      if (!store->commit_deserialised(changes, v, new_maps, hooks))
+      if (!store->commit_deserialised(changes, v, view, new_maps, hooks))
```
achamayou, line 79: All the applys use term, so I would rename that one for local consistency.

src/kv/kv_types.h
```diff
@@ -68,6 +68,11 @@ namespace kv
     {
       return {term, version};
     }
+
+    bool operator==(const TxID& other)
```
achamayou, line 72: default would work

src/kv/kv_types.h
```diff
@@ -315,7 +320,7 @@ namespace kv
     virtual void try_emit_signature() = 0;
     virtual void emit_signature() = 0;
     virtual crypto::Sha256Hash get_replicated_state_root() = 0;
-    virtual std::pair<kv::TxID, crypto::Sha256Hash>
+    virtual std::tuple<kv::TxID, crypto::Sha256Hash, kv::Term>
```
achamayou, line 323:
```suggestion
    virtual std::tuple<kv::TxID, crypto::Sha256Hash, kv::Term /* read_term */>
```

src/kv/store.h
```diff
@@ -135,9 +147,18 @@ namespace kv
         version = 0;
       }
 
+      // Further transactions should read in the commit term
+      read_term = commit_term;
+
       return version;
     }
 
+    TxID current_txid_internal()
```
achamayou, line 156: _unsafe

src/kv/store.h
```diff
@@ -641,14 +661,16 @@ namespace kv
       }
     }
 
-    void set_term(Term t) override
+    void set_commit_term(Term t) override
```
achamayou, line 664: initialise_term() and throw if term is already initialised?

tests/recovery.py
```diff
@@ -113,25 +113,25 @@ def run(args):
...
-                else:
-                    recovered_network = network
-            else:
-                recovered_network = test(network, args, from_snapshot=False)
-            network = recovered_network
-            LOG.success("Recovery complete on all nodes")
+        network.txs.issue(network, number_txs=1)
+        network.txs.issue(network, number_txs=1, repeat=True)
+
+        # Alternate between recovery with primary change and stable primary-ship,
```
achamayou, line 122: TBD?

achamayou (approved):

src/kv/kv_types.h
jumaffre: As discussed, this is only available from C++20, so waiting on #2720.

tests/recovery.py
jumaffre: Unwanted change. Should be reverted now.

eddyashton (approved):

PR #2760, opened 2021-07-05
Fix BFT issues

ashamis:
This is work towards getting the partition tests to work correctly with BFT. We solve several issues here but not all of them so this PR is a checkpoint and more fixes will follow.

### DIFF
--- a/src/consensus/aft/impl/view_change_tracker.h
+++ b/src/consensus/aft/impl/view_change_tracker.h
@@ -159,7 +159,7 @@ namespace aft
       if (!store->verify_view_change_request_confirmation(vc, from))
       {
         LOG_INFO_FMT(
-          "Add unknown evidence - bad view change confirmation, from:{}", from);
+          "Add unknown evidence - bad view change confirmation, from:{}, sig.size:{}, sig:{}", from, vc.signature.size(), vc.signature);
         return false;
       }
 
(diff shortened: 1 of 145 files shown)


src/consensus/aft/impl/view_change_tracker.h
```diff
@@ -258,6 +258,7 @@ namespace aft
         nv.view_change_messages.emplace(it.first, it.second);
       }
 
+      store->sign_view_change_confirmation(nv);
```
achamayou, line 261: Was this not signed at all before, or is this code being moved from somewhere else?

src/consensus/aft/raft.h
```diff
@@ -436,14 +436,25 @@ namespace aft
...
       std::lock_guard<std::mutex> guard(state->lock);
-      return {get_term_internal(state->commit_idx), state->commit_idx};
+      if (consensus_type == ConsensusType::CFT)
+      {
+        return {get_term_internal(state->commit_idx), state->commit_idx};
+      }
+      else
+      {
+        auto progress_tracker = store->get_progress_tracker();
+        auto idx = progress_tracker->get_rollback_seqno();
```
achamayou, line 446: What's the rollback seqno, is the index of the last committed transaction?

ashamis: This is the latest prepared which is >= committed

src/consensus/aft/impl/view_change_tracker.h
ashamis: yeah not signed before, so fun bug here :S

ashamis:
Closing this PR as it has been broken down smaller PRs

Issue #2724, opened 2021-06-24
BFT Reconfiguration: Learners consensus commit hook on reconfig tx and reshare tx + on commit watermark trigger


ashamis:
Add a consensus hook to recognize the completion of byzantine identity resharing and make the commit watermark dependent on that.

achamayou:
Done in #2899

PR #2761, opened 2021-07-05
Resharing hooks and RPC

wintersteiger:
Adds identity hooks and makes reconfiguration dependent on having an identity (if enabled). Addresses  #2724.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -432,9 +432,7 @@ if(BUILD_TESTS)
     add_executable(
       raft_driver ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/driver.cpp
     )
-    target_link_libraries(
-      raft_driver PRIVATE ccfcrypto.host openenclave::oehost
-    )
+    target_link_libraries(raft_driver PRIVATE ccfcrypto.host)
     target_include_directories(raft_driver PRIVATE src/aft)
 
     add_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,7 +301,7 @@ if(BUILD_TESTS)
       ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/view_history.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/committable_suffix.cpp
     )
-    target_link_libraries(raft_test PRIVATE ccfcrypto.host http_parser.host)
+    target_link_libraries(raft_test PRIVATE ccfcrypto.host)
 
     add_unit_test(
       raft_enclave_test
@@ -433,7 +433,7 @@ if(BUILD_TESTS)
       raft_driver ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/driver.cpp
     )
     target_link_libraries(
-      raft_driver PRIVATE ccfcrypto.host openenclave::oehost http_parser.host
+      raft_driver PRIVATE ccfcrypto.host openenclave::oehost
     )
     target_include_directories(raft_driver PRIVATE src/aft)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,7 +301,7 @@ if(BUILD_TESTS)
       ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/view_history.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/committable_suffix.cpp
     )
-    target_link_libraries(raft_test PRIVATE ccfcrypto.host)
+    target_link_libraries(raft_test PRIVATE ccfcrypto.host http_parser.host)
 
     add_unit_test(
       raft_enclave_test
@@ -432,7 +432,9 @@ if(BUILD_TESTS)
     add_executable(
       raft_driver ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/driver.cpp
     )
-    target_link_libraries(raft_driver PRIVATE ccfcrypto.host)
+    target_link_libraries(
+      raft_driver PRIVATE ccfcrypto.host openenclave::oehost http_parser.host
+    )
     target_include_directories(raft_driver PRIVATE src/aft)
 
     add_test(
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -32,11 +32,11 @@
 #include "impl/state.h"
 #include "impl/view_change_tracker.h"
 #include "kv/kv_types.h"
-#include "node/byzantine_identity.h"
 #include "node/node_to_node.h"
 #include "node/node_types.h"
 #include "node/progress_tracker.h"
 #include "node/request_tracker.h"
+#include "node/resharing_tracker.h"
 #include "node/rpc/tx_status.h"
 #include "node/signatures.h"
 #include "raft_types.h"
@@ -148,7 +148,7 @@ namespace aft
     std::unordered_map<ccf::NodeId, ccf::SeqNo> learners;
     bool use_two_tx_reconfig = false;
     bool require_identity_for_reconfig = false;
-    std::shared_ptr<ccf::IdentityTracker> identity_tracker;
+    std::shared_ptr<ccf::ResharingTracker> resharing_tracker;
 
     // Index at which this node observes its retirement
     std::optional<ccf::SeqNo> retirement_idx = std::nullopt;
@@ -195,7 +195,7 @@ namespace aft
       std::shared_ptr<Executor> executor_,
       std::shared_ptr<aft::RequestTracker> request_tracker_,
       std::unique_ptr<aft::ViewChangeTracker> view_change_tracker_,
-      std::shared_ptr<ccf::IdentityTracker> identity_tracker_,
+      std::shared_ptr<ccf::ResharingTracker> resharing_tracker_,
       std::chrono::milliseconds request_timeout_,
       std::chrono::milliseconds election_timeout_,
       std::chrono::milliseconds view_change_timeout_,
@@ -219,7 +219,7 @@ namespace aft
       view_change_timeout(view_change_timeout_),
       sig_tx_interval(sig_tx_interval_),
 
-      identity_tracker(std::move(identity_tracker_)),
+      resharing_tracker(std::move(resharing_tracker_)),
 
       public_only(public_only_),
 
@@ -260,7 +260,7 @@ namespace aft
 
       if (require_identity_for_reconfig)
       {
-        if (!identity_tracker)
+        if (!resharing_tracker)
         {
           throw std::logic_error("missing identity tracker");
         }
@@ -590,15 +590,15 @@ namespace aft
       create_and_remove_node_state();
     }
 
-    void add_identity(
+    void add_resharing_result(
       ccf::SeqNo seqno,
       kv::ReconfigurationId rid,
-      const ccf::Identity& identity)
+      const ccf::ResharingResult& result)
     {
       if (use_two_tx_reconfig && require_identity_for_reconfig)
       {
-        assert(identity_tracker);
-        identity_tracker->add_identity(seqno, rid, identity);
+        assert(resharing_tracker);
+        resharing_tracker->add_resharing_result(seqno, rid, result);
       }
     }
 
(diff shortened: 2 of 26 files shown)


wintersteiger:
Two discussion points for this one:

  - Searching for the reconfiguration id isn't great
  - What is the authentication requirement for nodes that aren't part of the configuration yet, but which need to call RPCs for identity negotiation?

CMakeLists.txt
```diff
@@ -303,7 +303,7 @@ if(BUILD_TESTS)
       raft_test ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/main.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/view_history.cpp
     )
-    target_link_libraries(raft_test PRIVATE ccfcrypto.host)
+    target_link_libraries(raft_test PRIVATE ccfcrypto.host http_parser.host)
```
achamayou, line 306: How does the raft_test now depend on the http parser?

src/consensus/aft/raft.h
```diff
@@ -564,10 +582,23 @@ namespace aft
...
     }
 
+    void add_identity(
+      ccf::SeqNo seqno,
+      kv::ReconfigurationId rid,
+      const ccf::Identity& identity)
+    {
+      if (use_two_tx_reconfig && require_identity_for_reconfig)
+      {
+        assert(byzantine_identity_tracker);
```
achamayou, line 597: On :264, we check systematically, but this is just an assert here. Is this something that always needs checking at runtime or a semi-static guarantee?

src/consensus/aft/raft.h
```diff
@@ -2898,9 +2940,18 @@ namespace aft
       {
         if (
           nodes.find(id) != nodes.end() && learners.find(id) == learners.end())
+        {
+          LOG_TRACE_FMT("trusted: {}", id);
```
achamayou, line 2944:
```suggestion
          LOG_TRACE_FMT("Trusted: {}", id);
```

src/consensus/aft/raft.h
```diff
@@ -2898,9 +2940,18 @@ namespace aft
...
+          r++;
+        }
+        else if (id == state->my_node_id && !is_learner())
         {
           r++;
         }
+        else
+        {
+          LOG_TRACE_FMT("untrusted: {}", id);
```
achamayou, line 2953:
```suggestion
          LOG_TRACE_FMT("Untrusted: {}", id);
```

CMakeLists.txt
wintersteiger: Via `make_request` in `ByzantineIdentityTracker`; the `Aft` class is now able to make HTTP calls. If there's a better way to do this, I'm happy to change it.

src/consensus/aft/raft.h
wintersteiger: Just a sanity check - since we check and throw an exception in line 264, this assertion should always hold.

src/consensus/aft/raft.h
wintersteiger: Unnecessary debug output, I'll remove that.

wintersteiger:
@achamayou is it ok to merge this one or do you want to take another look?

src/consensus/aft/raft.h
```diff
@@ -248,6 +256,15 @@ namespace aft
         // commit is always 1.
         state->view_history.update(1, starting_view_change);
         use_two_tx_reconfig = true;
+        require_identity_for_reconfig = true;
```
ashamis, line 259: Is this variable needed? it just the check if the node is using BFT

src/node/hooks.h
```diff
@@ -103,7 +103,9 @@ namespace ccf
 
     void call(kv::ConfigurableConsensus* consensus) override
     {
-      for (auto nc : configs)
+      // This hook is always executed after the hook for the nodes table above,
```
ashamis, line 106: Is it possible to create an assert/throw if this assumption is broken

src/node/identity.h
```diff
@@ -10,9 +10,20 @@
 
 namespace ccf
 {
-  struct NetworkIdentity
+  struct Identity
   {
+    enum class IdentityType
+    {
+      NORMAL,
```
ashamis, line 17: I don't think NORMAL is very descriptive any chance this can be renamed?

src/consensus/aft/raft.h
wintersteiger: Today it isn't necessary, but if we ever want to use identities for reconfiguration in CFT then it would be necessary.

src/node/hooks.h
wintersteiger: Sure, that's a good idea.

src/node/identity.h
wintersteiger: I agree, but I can't find a good name. What would you call the normal/current concept of identity?

src/node/rpc/node_frontend.h
```diff
@@ -1001,6 +1001,48 @@ namespace ccf
...
+        // For now, just pretend that we're done and that we have a shared
+        // identity.
+        ByzantineIdentity byid;
+        byid.cert = network.identity->cert;
+        ids->put(in.rid, byid);
+        return make_success(true);
+      };
+
+      make_endpoint(
+        "update-identity",
```
ashamis, line 1037: should this be called create-identity? it seems like this will fail if the identity exists

src/node/byzantine_identity.h
```diff
@@ -0,0 +1,331 @@
...
+          return rid;
+        }
+      }
+      return -1;
+    }
+
+    bool have_identity_for(
+      kv::ReconfigurationId rid, ccf::SeqNo commit_idx) const
+    {
+      if (rid == (kv::ReconfigurationId)-1)
```
ashamis, line 145: can (kv::ReconfigurationId)-1 be made into a const or something similar?

src/node/byzantine_identity.h
```diff
@@ -0,0 +1,331 @@
...
+    void call(kv::ConfigurableConsensus* consensus) override
+    {
+      for (auto& [rid, id] : byids)
+      {
+        consensus->add_identity(version, rid, id);
+      }
+    }
+  };
+
+  class ByzantineIdentityTracker
```
ashamis, line 66: Is is possible to unit test this class?

src/node/hooks.h
wintersteiger: Good call, the name I had chosen actually broke that. Fixed now.

src/node/rpc/node_frontend.h
wintersteiger: This is where the Byzantine identity code will go; it will make many calls to this endpoint before the identity is actually established, but today we don't have that yet, so this RPC will immediately create a copy of the existing identity.

src/node/byzantine_identity.h
wintersteiger: It is. That was changed in another PR that's been merged into master, you may be reviewing an old un-rebased version of this PR?

src/node/byzantine_identity.h
wintersteiger: It really doesn't do anything testable yet, it just keeps a list of configuration IDs. This is where the actual Byzantine identity crypto will go, which will of course come with tests.

src/node/byzantine_identity.h
wintersteiger: Ah, no, I was wrong, I missed on occurrence. Thanks!

src/node/rpc/node_frontend.h
eddyashton:
```suggestion
        "update_identity",
```

For consistency with existing endpoint naming style.

src/node/byzantine_identity.h
```diff
@@ -0,0 +1,326 @@
...
+      crypto::KeyPairPtr node_sign_kp,
+      const crypto::Pem& node_cert)
+    {
+      LOG_DEBUG_FMT("Submitting RPC call to update identity");
+      ccf::UpdateIdentity::In ps = {rid};
+
+      http::Request request(fmt::format(
+        "/{}/{}",
+        ccf::get_actor_prefix(ccf::ActorsType::nodes),
+        "update-identity"));
```
eddyashton, line 293:
```suggestion
        "update_identity"));
```

src/node/entities.h
```diff
@@ -66,8 +66,8 @@ namespace ccf
     // Nodes identities and allowed code ids
     static constexpr auto NODES = "public:ccf.gov.nodes.info";
     static constexpr auto NODE_CODE_IDS = "public:ccf.gov.nodes.code_ids";
-    static constexpr auto NETWORK_CONFIGURATIONS =
-      "public:ccf.gov.network.configurations";
+    static constexpr auto NODES_CONFIGURATIONS =
+      "public:ccf.gov.nodes.network.configurations";
```
eddyashton, line 70: What does renaming this table mean for backwards compatibility? If there's not a strong reason to rename, I think we should avoid doing so.

CMakeLists.txt
eddyashton: I think we should avoid this dependency with another layer of abstraction. Either by templating (the tests take a `StubByzantineIdentityTracker` which doesn't do HTTP calls, which we already do for the ledger type for instance), or inheritance (`Aft` talks to a `ByzantineIdentityTracker` that has some stub virtual calls where it would send a HTTP request, and the actual instance is a `ByzantineIdentityTrackerWithDescriptiveName` that overrides those and needs HTTP-parser linkage). Either approach allows stub implementations that would make this unit testable - ie rather than creating HTTP requests, we just create an in-memory stack of requests, and the test can assert that the required calls were made.

src/node/entities.h
wintersteiger: This is a new table that was added very recently, so no backwards compatibility problems yet. The motivation for renaming is that it now appears after the `nodes` table in lexicographic ordering and we rely on that during the execution of table hooks.

CMakeLists.txt
wintersteiger: That's a good suggestion, I'll add a stub tracker!

src/node/identity.h
achamayou: Replicated/Sharded|Distributed?

CMakeLists.txt
```diff
@@ -432,7 +432,9 @@ if(BUILD_TESTS)
     add_executable(
       raft_driver ${CMAKE_CURRENT_SOURCE_DIR}/src/consensus/aft/test/driver.cpp
     )
-    target_link_libraries(raft_driver PRIVATE ccfcrypto.host)
+    target_link_libraries(
+      raft_driver PRIVATE ccfcrypto.host openenclave::oehost
```
achamayou, line 436: Why does the raft_drive have a dependency on OE/oehost?

src/node/byzantine_identity.h
```diff
@@ -0,0 +1,342 @@
...
+      auto msg = std::make_unique<threading::Tmsg<UpdateIdentityTaskMsg>>(
+        update_identity_cb, rid, rpc_map, node_sign_kp, node_cert);
+
+      threading::ThreadMessaging::thread_messaging.add_task(
+        threading::ThreadMessaging::get_execution_thread(
+          threading::MAIN_THREAD_ID),
+        std::move(msg));
+    }
+
+    virtual kv::ReconfigurationId find_reconfiguration(
```
achamayou, line 133: I would prefer an std::optional, that's what we do nearly everywhere in preference to sentinel values.

src/node/byzantine_identity.h
```diff
@@ -0,0 +1,342 @@
...
+    {
+      LOG_DEBUG_FMT("Identities: adding identity for configuration #{}", rid);
+
+      assert(id.type == Identity::IdentityType::BYZANTINE);
+      SeqNoById entry = {seqno, *reinterpret_cast<ByzantineIdentity*>(&id)};
+      identities.emplace(rid, entry);
+
+      sessions.erase(rid);
+
+      // Delete old identities?
```
achamayou, line 176: Yes?

src/node/identity.h
achamayou: FWIW, I think "Byzantine" is also very confusing, because it implies malicious rather than "Byzantine Fault Tolerant". Describing whether the identity is copied/replicated in each node, or sharded/split seems more accurate.

achamayou (approved):

CMakeLists.txt
achamayou: We want to be able to build and run this test even in virtual mode.

PR #2767, opened 2021-07-06
Add table of network configurations

wintersteiger:
This adds a table of network configurations that assigns unique reconfiguration IDs and a hook on that table that will trigger ORC RPCs in the next reconfiguration PR.

### DIFF
--- a/doc/audit/builtin_maps.rst
+++ b/doc/audit/builtin_maps.rst
@@ -111,7 +111,7 @@ Identity, status and attestations (endorsed quotes) of the nodes hosting the net
 ``network_configurations``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The currently valid and in-flight network configurations of the network.
+The currently valid and in-flight network configurations of the network. The entry at -1 contains a dummy configuration that holds the largest ID used so far.
 
 **Key** Reconfiguration ID: a unique identifier of a configuration
 
--- a/doc/audit/builtin_maps.rst
+++ b/doc/audit/builtin_maps.rst
@@ -115,7 +115,7 @@ The currently valid and in-flight network configurations of the network.
 
 **Key** Reconfiguration ID: a unique identifier of a configuration
 
-**Value** A set of node IDs of the nodes in the respective configuration. Represented as JSON.
+**Value** A set of node IDs of the nodes in the respective configuration, represented as a JSON array.
 
 .. doxygenstruct:: ccf::NetworkConfiguration
    :project: CCF
--- a/doc/audit/builtin_maps.rst
+++ b/doc/audit/builtin_maps.rst
@@ -108,7 +108,7 @@ Identity, status and attestations (endorsed quotes) of the nodes hosting the net
 .. doxygenenum:: ccf::QuoteFormat
    :project: CCF
 
-``network_configurations``
+``network.configurations``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The currently valid and in-flight network configurations of the network. The entry at -1 contains a dummy configuration that holds the largest ID used so far.
--- a/doc/audit/builtin_maps.rst
+++ b/doc/audit/builtin_maps.rst
@@ -108,6 +108,19 @@ Identity, status and attestations (endorsed quotes) of the nodes hosting the net
 .. doxygenenum:: ccf::QuoteFormat
    :project: CCF
 
+``network_configurations``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The currently valid and in-flight network configurations of the network.
+
+**Key** Reconfiguration ID: a unique identifier of a configuration
+
+**Value** A set of node IDs of the nodes in the respective configuration. Represented as JSON.
+
+.. doxygenstruct:: ccf::NetworkConfiguration
+   :project: CCF
+   :members:
+
 ``nodes.code_ids``
 ~~~~~~~~~~~~~~~~~~
 
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -326,6 +326,11 @@ namespace aft
       return replica_state == kv::ReplicaState::Learner;
     }
 
+    bool is_retired()
+    {
+      return replica_state == kv::ReplicaState::Retired;
+    }
+
     ccf::NodeId get_primary(ccf::View view)
     {
       CCF_ASSERT_FMT(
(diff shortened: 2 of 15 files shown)


doc/audit/builtin_maps.rst
```diff
@@ -108,6 +108,19 @@ Identity, status and attestations (endorsed quotes) of the nodes hosting the net
...
    :project: CCF
 
+``network_configurations``
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The currently valid and in-flight network configurations of the network.
+
+**Key** Reconfiguration ID: a unique identifier of a configuration
+
+**Value** A set of node IDs of the nodes in the respective configuration. Represented as JSON.
```
achamayou, line 118: JSON doesn't have sets, so it would be good to mention whether that's represented as an array or as a object with special values.

src/node/entities.h
```diff
@@ -66,6 +66,8 @@ namespace ccf
     // Nodes identities and allowed code ids
     static constexpr auto NODES = "public:ccf.gov.nodes.info";
     static constexpr auto NODE_CODE_IDS = "public:ccf.gov.nodes.code_ids";
+    static constexpr auto NETWORK_CONFIGURATIONS =
+      "public:ccf.gov.network_configurations";
```
jumaffre, line 70: `public:ccf.gov.network.configurations` or `public:ccf.gov.nodes.configurations` maybe?

wintersteiger: Both good suggestions. Are there any conventions for table names? I'm sort of inclined to pick the shortest one :)

src/node/reconfig_id.h
```diff
@@ -8,20 +8,42 @@
...
+  inline void add_new_network_reconfiguration(
+    ccf::NetworkTables& tables, kv::Tx& tx, kv::NetworkConfiguration& config)
+  {
+    config.rid = get_next_reconfiguration_id(tables, tx);
+    LOG_DEBUG_FMT(
+      "Configurations: adding new entry to network_configurations table: {}",
+      config);
+    auto nconfigs = tx.rw(tables.network_configurations);
+    nconfigs->put(config.rid, config);
+    nconfigs->put((kv::ReconfigurationId)-1, {config.rid, {}});
```
jumaffre, line 47: I'm a little confused about this. AIUI, we keep track of the latest config id at a well-known key (`-1`) and then we store the config at its id. If that's what we want to do, we should at least add a comment next to the table definition to explain this. This will also cause memory growth in the KV store as more nodes are added to the service. Is this intended?

src/node/rpc/member_frontend.h
```diff
@@ -831,10 +831,14 @@ namespace ccf
            {in.quote_info},
            in.public_encryption_key,
            NodeStatus::TRUSTED,
-           get_next_reconfiguration_id(network, ctx.tx),
            std::nullopt,
            ds::to_hex(in.code_digest.data)});
 
+        kv::NetworkConfiguration nc =
```
jumaffre, line 837: We should move this to `add_node()`, only if the status is `NodeStatus::TRUSTED` so that we don't forget some cases where nodes are added.

src/node/reconfig_id.h
wintersteiger: Sure, will add a comment for `-1`.  And yes, outdated configurations will be removed when we switch to them, so that only the current one and future entries are retained. This isn't in this PR yet, because I couldn't find a way to correlate between old config (nodes table, no ID) and new config (with ID in network_configs) yet, so at the point where we switch to the next config, I don't know the right ID.

src/node/hooks.h
```diff
@@ -76,4 +76,33 @@ namespace ccf
       }
     }
   };
+
+  class NetworkConfigurationsHook : public kv::ConsensusHook
```
jumaffre, line 80: I'm not sure I understand why we have this parallel hook to update the consensus configuration. Could we remove the existing hook on the `nodes.info` table altogether if we're planning to use this new table going forward?

wintersteiger: In theory, yes. For now my plan was to keep both of them to avoid LTS issues. Keeping just one table would definitely make all of this easier.

src/consensus/aft/raft.h
```diff
@@ -555,6 +560,28 @@ namespace aft
...
+    {
+      LOG_DEBUG_FMT("Configurations: new network config: {{{}}}", config);
+      std::unique_lock<std::mutex> guard(state->lock, std::defer_lock);
+
+      if (is_bft_reexecution() && threading::ThreadMessaging::thread_count > 1)
+      {
+        guard.lock();
+      }
+
+      // Hooks may be reordered, so the node info in `configurations` and
```
achamayou, line 574: I don't understand this comment. Consensus hooks are always executed in TxID order. What's not currently predictable is the relative order of hooks on different maps at the same TxID. Is that what you are referring to? If so, this isn't set in stone, and could be changed.

src/node/entities.h
achamayou: The conventions are: lower-case, "ccf.gov." for all tables written to by governance directly or indirectly and deemed important for audit purposes. Nesting is fine, but avoid tables named "foo.bar.baz" if there is a table "foo.bar", ie. no table name should be the prefix of another.

That's about it I think. We try to avoid underscores/multiple words where we can, but it's not always possible, there is precedent for it.

src/node/hooks.h
achamayou: Is that right? I think the learner/trusted and retiring/retired distinction remains important, so unless I'm missing something we'll continue to need both?

src/node/reconfig_id.h
```diff
@@ -8,20 +8,45 @@
 
 namespace ccf
 {
-  inline size_t get_next_reconfiguration_id(
-    ccf::NetworkState& network, kv::ReadOnlyTx& tx)
+  inline kv::ReconfigurationId get_next_reconfiguration_id(
+    ccf::NetworkTables& tables, kv::ReadOnlyTx& tx)
   {
-    auto nodes = tx.ro(network.nodes);
+    // The entry at -1 contains a dummy configuration that holds the largest ID
```
achamayou, line 14: This would be better located in the table documentation.

src/node/reconfig_id.h
```diff
@@ -8,20 +8,45 @@
...
-    size_t max_id = 0;
-    nodes->foreach([&max_id](const auto& node_id, const auto& node_info) {
-      if (node_info.reconfiguration_id.has_value())
-      {
-        max_id = std::max(max_id, node_info.reconfiguration_id.value());
-      }
-      return true;
-    });
+    auto nconfigs = tx.ro(tables.network_configurations);
+    auto e = nconfigs->get((kv::ReconfigurationId)-1);
```
achamayou, line 18: I would suggest naming it, constexpr size_t CONFIG_COUNT_KEY = -1; for clarity.

src/consensus/aft/raft.h
wintersteiger: Yes, that's what I meant to say. I assumed that was set in stone, so I didn't look into it any further. It would be very convenient if we could make sure that the new table hook follows the nodes table hook.

src/node/hooks.h
wintersteiger: Yes, that's right, we need the same amount of information, but we could put all of it into one table instead of two.

src/node/entities.h
wintersteiger: Done, I went with `public:ccf.gov.network.configurations`, but I'm not emotionally attached to it, so feel free to change it if you think of a better name :)

src/consensus/aft/raft.h
achamayou: Ok, we can look at that in a separate PR.

achamayou (approved):

PR #2770, opened 2021-07-07
Fix network.wait_for_new_primary for BFT

ashamis:
BFT view-changes are not timeout for a heartbeat based like CFT but require a timeout of a pending request. In this change the test will wait for a new primary to be elected (when appropriate) and if the consensus protocol is bft it will issue requests.

### DIFF
--- a/tests/infra/network.py
+++ b/tests/infra/network.py
@@ -908,7 +908,7 @@ class Network:
                                     "msg": "This is submitted to force a view change",
                                 },
                             )
-                        time.sleep(5)
+                        time.sleep(1)
                     except CCFConnectionException:
                         LOG.warning(
                             f"Could not successfully connect to node {backup.node_id}."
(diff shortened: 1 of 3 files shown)


jumaffre (approved):

tests/infra/network.py
```diff
@@ -897,6 +897,22 @@ def wait_for_new_primary(self, old_primary, nodes=None, timeout_multiplier=2):
...
+                        with backup.client("user0") as c:
+                            _ = c.post(
+                                "/app/log/private",
+                                {
+                                    "id": -1,
+                                    "msg": "This is submitted to force a view change",
+                                },
+                            )
+                        time.sleep(5)
+                    except CCFConnectionException:
```
achamayou, line 912: I would do another backup = self.find_any_backup() here in the hope to get a happier backup, and perhaps shorten the sleep to 1?

achamayou (approved):

wintersteiger (approved):

PR #2772, opened 2021-07-07
Add retiring state for 2-tx reconfiguration

wintersteiger:
This adds the new `RETIRING` consensus state, the associated logic, and a test.

### DIFF
--- a/src/runtime_config/default/actions.js
+++ b/src/runtime_config/default/actions.js
@@ -829,4 +829,25 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "remove_node_2tx",
+    new Action(
+      function (args) {
+        checkEntityId(args.node_id, "node_id");
+      },
+      function (args) {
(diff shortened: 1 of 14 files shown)


src/runtime_config/default/actions.js
```diff
@@ -829,4 +829,25 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "remove_node_2tx",
```
wintersteiger, line 833: I would like to avoid having a separate gov action here. Is there any way I can find out whether we're in two-tx or BFT mode so I can make "remove_node" dependent on that? Should we perhaps save that information in the service config table?

achamayou: That's an excellent idea, we can do the same thing for adding nodes. The service table is an excellent place for the flag.

wintersteiger: Done!

achamayou (approved):

Issue #2684, opened 2021-06-17
Closing connections during key exchange leads to permanently blocked node-to-node channel


eddyashton:
This was spotted while debugging a recent `code_update_test` instability. This is not present in main right now, because we currently never close connections, but will reappear when we implement #2672.

In summary:
1) Node 3 becomes candidate, and opens channels to every other node to ask for votes.
2) Node 3 sends `key_exchange_init` to nodes it wasn't previously talking to.
3) Node 2 receives this and executes `consume_initiator_key_share()`, transitioning to `WAITING_FOR_FINAL`.
4) Node 2 sends a `key_exchange_response` message to Node 3.
5) Before Node 3 receives this `key_exchange_response` it receives an Append Entries from someone who already won in this term and transitions to follower, _closing its outgoing connections_.
6) Node 3 receives `key_exchange_response` and considers its channel with Node 2 established, but its `key_exchange_final` response is dropped (because the host-side connection has already been closed).
7) Node 2 remains in `WAITING_FOR_FINAL`.
8) Next time Node 2 tries to talk to 3, it fails (state is not `ESTABLISHED`, call `initiate()`, early out because we're in state `WAITING_FOR_FINAL`).

So nodes 2 and 3 are permanently unable to communicate, and progress which relies on one of them voting for the other + sending append entries can never happen.

Transformed logs showing the steps above:
```
1)
45:24.844 |03|       Becoming candidate n[3=3692]: 3 (raft.h:2419)

2)
45:24.857 |03|       Creating new outbound channel to n[2=f53b] (127.216.212.151:42357) (channels.h:940)
45:24.861 |03|       Added node connection with n[2=f53b] (127.216.212.151:42357) (node_connections.h:378)
45:24.861 |03|       Send request vote from n[3=3692] to n[2=f53b] at 78 (raft.h:2219)
45:24.861 |03|       Queuing send of 1 message to n[2=f53b] - status is 0 (channels.h:764)
45:24.861 |03|       Initiating node channel with n[2=f53b]. (channels.h:725)
45:24.862 |03|       node send to n[2=f53b] [1035] (node_connections.h:321)
45:24.862 |03|       key_exchange_init -> n[2=f53b] node serial: 83E423FC96F6DB5787741C2FDD235677 (channels.h:753)

3)
45:24.869 |02|     consume_initiator_key_share from n[3=3692], status == 0 (channels.h:539)

5)
[...]

PR #2775, opened 2021-07-07
[Do not merge] Changes around channel management

eddyashton:
Opening this mega-PR to get CI testing, as the previously problematic tests are now passing locally. This includes many non-critical refactors/reworkings of adjacent bits of code, so I'll factor those into separate PRs.

A brief summary of the changes, as I remember them:
- `NodeToNodeChannelManager` takes some of `node_state.h`, and some of `channels.h`, and is the sole point for external systems to send/receive (they don't open/close/manage individual channels)
- Host-side node-to-node session management isn't at all exposed to the enclave. It doesn't distinguish inbound/outbound sessions, and doesn't try to (asynchronously) open/close sessions. Instead the enclave deals purely with NodeIDs. I believe that either the node  already knows the IP of its recipient (it has been seen in the KV), in which case it has sent an `associate` message to the host and from then on the host can remember the `ID -> IP` mapping, or else it is a response to an inbound message, in which case the host still has a single associate session. Then the host can create TCP sessions on-demand, re-creating them if they were closed by networking issues, etc, and just becomes another step in the best-effort-but-faulty network stack (WIP)
- Consistent, explicit serialisation of `node_outbound` messages over the ringbuffer. We previously benefitted from the fact that writing multiple values to the ringbuffer would blit them in a known format, which happened to match how we later read them with `serialized::read` (partly on the host, partly inside the receiving enclave). We now gives this explicit types like any other message, distinguishing the host-readable bits from the payload. There's a reserialisation penalty here but I think we can minimise it, and there's an associated fix to use `ByteRange`s on these types that may mitigate its effect in practice
- Consistent logging format, with NodeID and current-state prefix, for all channel logging

And the core goal:
- Re-examine the key exchange state machine and parsing logic to avoid deadlocked states


eddyashton:
2 more thoughts:
- This is the proposed fix for #2684, and is a replacement for #2688.
- Splitting this PR into many smaller PRs will make it harder to cherry-pick onto the LTS, and the related fixes may also make unrelated PRs harder to cherry-pick.

eddyashton:
Split up into other PRs: #2777, #2780, #2783, #2801.

PR #2776, opened 2021-07-07
Move `StateMachine` to standalone ds header

eddyashton:
We have a simple `StateMachine` class in `NodeState`. This moves it to `ds/` where it can be reused by other systems (eg - node-to-node channels). Eventually I'd like to add a restricted set of transitions to this implementation, so we can check that we're only transitioning through states in the expected order.

### DIFF
--- a/src/ds/state_machine.h
+++ b/src/ds/state_machine.h
@@ -1,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ds/logger.h"
+
+#include <atomic>
+#include <string>
+
+namespace ds
+{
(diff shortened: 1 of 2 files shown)


src/ds/state_machine.h
```diff
@@ -0,0 +1,48 @@
...
+namespace ds
+{
+  template <typename T>
+  class StateMachine
+  {
+    const std::string label;
+    std::atomic<T> s;
+
+  public:
+    StateMachine(std::string&& l, T s) : label(std::move(l)), s(s) {}
```
achamayou, line 19:
```suggestion
    StateMachine(std::string&& label_, T s) : label(std::move(label_)), s(s) {}
```

achamayou (approved):

PR #2778, opened 2021-07-07
Fix endpoint prefix inconsistency

achamayou:
Follow up to #2771.

This does:

1. Make the internal representation of endpoints consistent across C++ and JS; they are all `/`-prefixed.
2. Preserve `app.json` and `make_.*endpoint` compatibility, and extends it to be consistently permissive.
3. Update samples to always `/`-prefix.

This does _not_:

1. Change metrics, which remain without a `/`-prefix.
2. Change OpenAPI specs, which remain `/`-prefixed.

I believe this is the furthest we can go towards consistency without breaking compatibility.

### DIFF
--- a/src/apps/js_generic/js_generic.cpp
+++ b/src/apps/js_generic/js_generic.cpp
@@ -489,7 +489,7 @@ namespace ccfapp
     ccf::endpoints::EndpointDefinitionPtr find_endpoint(
       kv::Tx& tx, enclave::RpcContext& rpc_ctx) override
     {
-      const auto method = fmt::format("/{}", rpc_ctx.get_method());
+      const auto method = rpc_ctx.get_method();
       const auto verb = rpc_ctx.get_request_verb();
 
       auto endpoints =
--- a/src/endpoints/endpoint_registry.cpp
+++ b/src/endpoints/endpoint_registry.cpp
@@ -82,7 +82,14 @@ namespace ccf::endpoints
     const AuthnPolicies& ap)
   {
     Endpoint endpoint;
-    endpoint.dispatch.uri_path = method;
+    if (nonstd::starts_with(method, "/"))
+    {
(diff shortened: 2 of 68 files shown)


src/apps/js_generic/js_generic.cpp
```diff
@@ -489,7 +489,7 @@ namespace ccfapp
     ccf::endpoints::EndpointDefinitionPtr find_endpoint(
       kv::Tx& tx, enclave::RpcContext& rpc_ctx) override
     {
-      const auto method = fmt::format("/{}", rpc_ctx.get_method());
+      const auto method = rpc_ctx.get_method();
```
achamayou, line 492: Method is already `/`-prefixed.

src/endpoints/endpoint_registry.cpp
```diff
@@ -84,7 +84,14 @@ namespace ccf::endpoints
     const AuthnPolicies& ap)
   {
     Endpoint endpoint;
-    endpoint.dispatch.uri_path = method;
+    if (nonstd::starts_with(method, "/"))
```
achamayou, line 87: `/`-prefix at registration time if necessary.

src/node/rpc/member_frontend.h
```diff
@@ -100,7 +100,11 @@ namespace ccf
           auto info_ = info;
           info_.js_module = module_prefix + info_.js_module;
           auto verb = nlohmann::json(method).get<RESTVerb>();
-          endpoints->put(ccf::endpoints::EndpointKey{url, verb}, info_);
+          endpoints->put(
+            ccf::endpoints::EndpointKey{
+              nonstd::starts_with(url, "/") ? url : fmt::format("/{}", url),
```
achamayou, line 105: Prefix if necessary for JS endpoints.

PR #2780, opened 2021-07-08
Move channel manager into node state

eddyashton:
Another part of #2775. The forwarder and node-to-node channels were owned by the enclave and passed to node state, but the enclave doesn't do anything with them. It makes more sense for them to be owned and constructed within the node state. To make that construction path a little cleaner, the `ConsensusType` in the configuration moves from the top-level to the `consensus::Configuration` struct.

### DIFF
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -315,9 +315,7 @@ namespace ccf
 
       for (auto& [actor, fe] : rpc_map->frontends())
       {
-        fe->set_sig_intervals(
-          sig_tx_interval,
-          sig_ms_interval);
+        fe->set_sig_intervals(sig_tx_interval, sig_ms_interval);
         fe->set_cmd_forwarder(cmd_forwarder);
       }
     }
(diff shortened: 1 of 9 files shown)


src/node/node_state.h
```diff
@@ -1405,41 +1414,45 @@ namespace ccf
...
-
-      const uint8_t* payload_data = data.data();
-      size_t payload_size = data.size();
-
       NodeMsgType msg_type =
         serialized::overlay<NodeMsgType>(payload_data, payload_size);
       NodeId from = serialized::read<NodeId::Value>(payload_data, payload_size);
 
-      switch (msg_type)
+      if (msg_type == ccf::NodeMsgType::forwarded_msg)
```
jumaffre, line 1423: Is it awkward to process forwarded messages but not channel messages? For example, if a forwarded message caused the channel to be re-established on the forwarder node, this node (the forwardee) wouldn't ever receive it?

eddyashton: This is the same behaviour we had previously, just moved from the message handler in `enclave.h` into this function call.

I don't follow the example. A legitimate forwarded message will only ever arrive after channel establishment messages, which will only be accepted once this node is part of network. If we get a forwarded message before then, we will try to process it here but in practice fail to decrypt it later.

achamayou (approved):

src/node/node_state.h
jumaffre: I didn't realise this behaviour already existed - it's now a lot more obvious that forwarded messages are processed even though the node isn't in state `partOfNetwork/partOfPublicNetwork/readingPrivateLedger`. 

I was thinking about a node which transitioned from one of these 3 states to another state e.g. a new `retired` state (but such transitions don't actually exist, and these three states are final or transition to one another).

PR #2781, opened 2021-07-08
Fix some of the BFT rollback related bugs

ashamis:
How far things rollback in CFT and BFT is different. In BFT when rolling back in most cases (the ones in this PR) we should rollback to the last prepared.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -371,7 +371,7 @@ namespace aft
 
       std::lock_guard<std::mutex> guard(state->lock);
       state->current_view += starting_view_change;
-      become_leader();
+      become_leader(true);
     }
 
     void force_become_leader(
@@ -395,7 +395,7 @@ namespace aft
       state->view_history.initialise(terms);
       state->view_history.update(index, term);
       state->current_view += starting_view_change;
-      become_leader();
+      become_leader(true);
     }
 
     void init_as_follower(
@@ -436,14 +436,25 @@ namespace aft
     std::pair<Term, Index> get_commit_term_and_idx()
     {
       std::lock_guard<std::mutex> guard(state->lock);
-      return {get_term_internal(state->commit_idx), state->commit_idx};
+      if (consensus_type == ConsensusType::CFT)
+      {
+        return {get_term_internal(state->commit_idx), state->commit_idx};
+      }
+      else
+      {
+        auto progress_tracker = store->get_progress_tracker();
+        auto idx = progress_tracker->get_highest_committed_level();
+        return {get_term_internal(idx), idx};
+      }
     }
 
     std::optional<kv::Consensus::SignableTxIndices>
     get_signable_commit_term_and_idx()
     {
       std::lock_guard<std::mutex> guard(state->lock);
-      if (state->commit_idx >= election_index)
+      if (
+        consensus_type == ConsensusType::BFT ||
+        state->commit_idx >= election_index)
       {
         kv::Consensus::SignableTxIndices r;
         r.term = get_term_internal(state->commit_idx);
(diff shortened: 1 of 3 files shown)


src/consensus/aft/raft.h
```diff
@@ -436,14 +436,25 @@ namespace aft
...
       std::lock_guard<std::mutex> guard(state->lock);
-      return {get_term_internal(state->commit_idx), state->commit_idx};
+      if (consensus_type == ConsensusType::CFT)
+      {
+        return {get_term_internal(state->commit_idx), state->commit_idx};
+      }
+      else
+      {
+        auto progress_tracker = store->get_progress_tracker();
+        auto idx = progress_tracker->get_highest_committed_level();
```
achamayou, line 446: It seems to me that we ought to have a get_highest_committed_level() at the Consensus level, doing the dispatch to either value, and start using that?

src/consensus/aft/raft.h
```diff
@@ -980,6 +994,7 @@ namespace aft
       const uint8_t* data,
       size_t size)
     {
+      LOG_DEBUG_FMT("Received evidence for view:{}, from {}", r.view, from);
```
achamayou, line 997: Seems weird that we sometimes ":{}" and sometimes " {}" for values. I don't really have a preference, but I would like it if we stuck to one style.

src/consensus/aft/raft.h
```diff
@@ -2640,14 +2671,22 @@ namespace aft
...
-    void become_leader()
+    void become_leader(bool force_become_leader = false)
     {
       if (is_retired())
       {
         return;
       }
 
-      election_index = last_committable_index();
+      if (consensus_type == ConsensusType::BFT && !force_become_leader)
```
achamayou, line 2681: Is the special case for force_become_leader necessary? I would have thought the initial rollback seqno could be returning the right initial value for election index. If that's the case, then last_committable_index() can be doing the switching on BFT/CFT, and this call can stay unchanged and avoid the flag.

src/node/progress_tracker.h
```diff
@@ -352,6 +352,8 @@ namespace ccf
 
       if (can_send_reply_and_nonce(cert, config))
       {
+        LOG_TRACE_FMT(
+          "sending revealed nonce, view:{}, seqno:{}", tx_id.view, tx_id.seqno);
```
achamayou, line 356:
```suggestion
          "Sending revealed nonce, view:{}, seqno:{}", tx_id.view, tx_id.seqno);
```

achamayou (approved):

src/consensus/aft/raft.h
ashamis: I added a comment to the code to explain this

src/consensus/aft/raft.h
```diff
@@ -2640,14 +2671,25 @@ namespace aft
...
-    void become_leader()
+    void become_leader(bool force_become_leader = false)
     {
       if (is_retired())
       {
         return;
       }
 
-      election_index = last_committable_index();
+      // When we force to become the primary we are going are around the
```
achamayou, line 2681:
```suggestion
      // When we force to become the primary we are going around the
```

src/consensus/aft/raft.h
```diff
@@ -2640,14 +2671,25 @@ namespace aft
...
+      // consensus protocol. Therefore in BFT we need to bypass the information in
+      // the progress tracker when working out how far we rollback.
```
achamayou, line 2683: I get that, but I would have thought that we only do this once at startup, when the progress tracker has a well known initial state. Am I missing something?

ashamis: no this is exactly what we are doing

achamayou: Ok, this is so we don't roll back the genesis/recovery tx, I see.

PR #2782, opened 2021-07-09
Fix LTS compatibility issue in service config

wintersteiger:
Fixes the breakage in yesterday's daily build.

### DIFF
--- a/src/node/config.h
+++ b/src/node/config.h
@@ -16,7 +16,7 @@ namespace ccf
 
     ConsensusType consensus = ConsensusType::CFT;
     std::optional<ReconfigurationType> reconfiguration_type =
-      ReconfigurationType::ONE_TRANSACTION;
+      std::nullopt;
   };
   DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(ServiceConfiguration)
   DECLARE_JSON_REQUIRED_FIELDS(
--- a/src/node/config.h
+++ b/src/node/config.h
@@ -15,8 +15,7 @@ namespace ccf
     size_t recovery_threshold = 0;
 
     ConsensusType consensus = ConsensusType::CFT;
-    std::optional<ReconfigurationType> reconfiguration_type =
-      std::nullopt;
+    std::optional<ReconfigurationType> reconfiguration_type = std::nullopt;
   };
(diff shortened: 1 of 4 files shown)


achamayou (approved):

src/node/config.h
```diff
@@ -15,12 +15,13 @@ namespace ccf
     size_t recovery_threshold = 0;
 
     ConsensusType consensus = ConsensusType::CFT;
-    ReconfigurationType reconfiguration_type =
+    std::optional<ReconfigurationType> reconfiguration_type =
```
eddyashton, line 18: Some background here, and why I think this should probably be `= nullopt` with the default-to-`ONE_TRANSACTION` logic handled elsewhere.

The logic is that if a field is declared in `OPTIONAL_FIELDS`, then in deserialisation its acceptable for it not to be present in the JSON object (the C++ field will retain its default value), and in serialisation if it _has_ its default value then it won't be written to the JSON object. Wrapping it in a `std::optional` just give you an additional value (`nullopt`), to distinguish "wasn't in the JSON object" from "was in the JSON object with the default value" (ie - if you have add a `std::optional<size_t> x = std::nullopt`, then you can distinguish `{}` from `{"x" = 0}`, which you can't do if you just had `size_t x`). However, in this case I think we always want to _write_ it, and just be permissive on deserialisation. By setting the default value here, then whenever we have the value `ONE_TRANSACTION`, we won't write that to the JSON object. If instead you write `= std::nullopt` for the default, then whenever it _does_ get a value (by some of the handling code setting this as the default), we'll write it to the JSON object (but by keeping it in `OPTIONAL_FIELDS`, still successfully _parse_ objects without it in).

src/node/config.h
```diff
@@ -15,12 +15,13 @@ namespace ccf
...
       ReconfigurationType::ONE_TRANSACTION;
```
eddyashton, line 19:
```suggestion
    std::optional<ReconfigurationType> reconfiguration_type =
      std::nullopt;
```

src/node/config.h
wintersteiger: Awesome, thanks for the background here, I didn't expect the default value not to be written when I set a default value, we definitely do want that!

eddyashton (approved):

jumaffre (approved):

PR #2783, opened 2021-07-09
Use standard ringbuffer type safety + macros when writing `node_outbound`

eddyashton:
Another cleanup carved off #2775.

`node_outbound` was written to ringbuffers directly, with `to_host->write`, rather than with the `RINGBUFFER_WRITE_MESSAGE` macro that every other instance used. This meant it didn't have the same error message handling and type-checking logic as every other ringbuffer write. This was done deliberately, since the types written were previously more distinct, and to allow appending multiple fields onto the ringbuffer without copying them into a single intermediate buffer. This removes those restrictions, getting the type safety without an additional copy, clarifying the exact types of the fields in this message and what they should be used for.

In particular, this lets us remove the generic 'blit anything' serialisation which occasionally failed when we accidentally passed new types - it is now _only_ used for integral and enum types.

### DIFF
--- a/src/node/channels.h
+++ b/src/node/channels.h
@@ -764,18 +764,20 @@ namespace ccf
       // with the nonce/seqno reset to 1). But, we can immediately start to send
       // messages with the new send_key.
 
+      // TODO: Remove these unnecessary copies
+      std::vector<uint8_t> payload(aad.p, aad.p + aad.n);
+
+      const auto gcm_hdr_serialised = gcm_hdr.serialise();
+      payload.insert(
+        payload.end(), gcm_hdr_serialised.begin(), gcm_hdr_serialised.end());
+
       std::vector<uint8_t> cipher(plain.n);
       send_key->encrypt(
         gcm_hdr.get_iv(), plain, aad, cipher.data(), gcm_hdr.tag);
+      payload.insert(payload.end(), cipher.begin(), cipher.end());
 
-      to_host->write(
-        node_outbound,
-        peer_id.value(),
-        type,
-        self.value(),
-        serializer::ByteRange{aad.p, aad.n},
-        gcm_hdr,
-        cipher);
+      RINGBUFFER_WRITE_MESSAGE(
+        node_outbound, to_host, peer_id.value(), type, self.value(), payload);
 
       LOG_TRACE_FMT(
         "-> {}: node msg with nonce={}", peer_id, (uint64_t)nonce.nonce);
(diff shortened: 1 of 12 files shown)


src/node/channels.h
```diff
@@ -597,31 +589,34 @@ namespace ccf
...
 
-      auto oks = kex_ctx.get_own_key_share();
-      std::vector<uint8_t> pks = {ks.p, ks.p + ks.n};
-      auto serialised_signed_share = sign_key_share(oks, false, &pks);
+      auto space = 1 * sizeof(size_t);
+      std::vector<uint8_t> payload(space);
+      auto data_ = payload.data();
+      serialized::write(data_, space, ChannelMsg::key_exchange_response);
+
+      sign_key_share(payload, kex_ctx.get_own_key_share(), false, peer_ks);
```
eddyashton, line 605: I changed the API of `sign_key_share` to serialise into an out-vector before I added the non-contiguous-ByteArray change for non-key-exchange messages. So this still creates one mega-payload (as it did previously), but could be improved to just use the existing buffers in-place. I'll do that in a future PR.

src/node/channels.h
```diff
@@ -749,31 +751,35 @@ namespace ccf
...
+      // 1) aad
+      // 2) gcm header
+      // 3) ciphertext
+      const serializer::ByteRange payload[] = {
+        {aad.p, aad.n},
+        {gcm_hdr_serialised.data(), gcm_hdr_serialised.size()},
+        {cipher.data(), cipher.size()}};
+
+      RINGBUFFER_WRITE_MESSAGE(
+        node_outbound, to_host, peer_id.value(), type, self.value(), payload);
```
eddyashton, line 782: To make this flow explicit:

A `node_outbound` message is a ringbuffer message that contains 4 fields: `Peer`, `Type`, `Self`, `Payload`.
When the host reads that, it reads the first 3 (because if `Type` has the correct value, it may append an additional payload from the ledger).
Using `Peer` as the key, the host then sends the other 3 to the intended node. This could be serialised or retransmitted in some other carrier format! In practice we reuse the existing serialised form and blit it over TCP directly, but this is a totally distinct transfer.
On the receiving host, we see that incoming message and read the fields in it: `Type`, `From`, `Payload`. Those are written as `node_inbound` over the ringbuffer.

We previously benefitted from several places where luckily the serialisation methods matched up, and I think they're now more explicit.

jumaffre (approved):

PR #2785, opened 2021-07-09
Return receipts for signature transactions

achamayou:
The historical query fetching logic does not populate receipts for signature transactions at the moment, but frontend and user code expects them to be present. This is adding signature receipts of the following shape on signature transactions in historical queries:

For reference, a non-signature receipt looks like:

```
{'leaf': 'fdc977c49d3a8bdf986176984e9432a09b5f6fe0c04e0b1c2dd177c03fdca9ec',
 'node_id': '682c161e1bc0aec694cac58a6ea456e1caa6c9c56d8dd873da9455c341947065',
 'proof': [{'left': 'f847e5efe3965b0dacb5c15c666602807a11fdecd465d0976779eed27121ffa3'},
           {'left': 'a56ce9efb73957f561f12d60513281fd2aaf16440234e2fd56e7d3d2ff4be8b0'},
           {'left': 'd91c982f525302244b13b6add92cd0925e1e0fb621ff2a7bb408ecc51be8528e'},
           {'left': '6d87faceda763ce65914f95dfcc04b37ea3f26bc552764752a0f2720039f76be'},
           {'left': 'e0cc83ea2fae6c535fc44605fb25ba9fdfb319e0e577b3541760f9a3565c549b'},
           {'left': 'f0e95ed85f5f6c0197aed4f6685b93dc56edd823a2532bd717558a5ab77267cb'}],
 'root': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
 'signature': 'MGQCMACklXqd0ge+gBS8WzewrwtwzRzSKy+bfrLZVx0YHmQvtsqs7dExYESsqrUrB8ZcKwIwS3NPKaGq0w2QlPlCqUC3vQoQvhcZgPHPu2GkFYa7JEOdSKLknNPHaCRv80zx2RGF'}
```

Whereas a signature receipt, as introduced in this PR, looks like:

```
{'leaf': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
 'node_id': '682c161e1bc0aec694cac58a6ea456e1caa6c9c56d8dd873da9455c341947065',
 'proof': [],
 'root': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
 'signature': 'MGQCMACklXqd0ge+gBS8WzewrwtwzRzSKy+bfrLZVx0YHmQvtsqs7dExYESsqrUrB8ZcKwIwS3NPKaGq0w2QlPlCqUC3vQoQvhcZgPHPu2GkFYa7JEOdSKLknNPHaCRv80zx2RGF'}
```

Note that root == leaf, and neither quite matches root and leaf concepts in other receipts, instead they are both what is signed directly, ie. root[idx -1]. This allows signature receipts to be verified without special-casing (receipt.py needs no change).

Specifically, this is _not_ a path to the signature transaction write set from the following signature, because:

- That is not meaningfully different in terms of guarantees from the signature itself, they're both an endorsement from the node.
- That would not be usefully verifiable without knowing how to serialise and integrity protect the ledger entry in any case, and leaking schema details.
- That would add latency in most cases, or be downright impossible to produce for the last transaction in an idle ledger.

Open items:

- [x] ~~Do we want to remove `get_store_at()` altogether? No.~~
- [x] CHANGELOG entry, this must be backported to the LTS
- [x] Expand documentation on receipts
- [x] Add test to suite

### DIFF
--- a/doc/use_apps/verify_tx.rst
+++ b/doc/use_apps/verify_tx.rst
@@ -68,6 +68,20 @@ This means that the request may return ``202 Accepted`` at first, with a suggest
      'root': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
      'signature': 'MGQCMACklXqd0ge+gBS8WzewrwtwzRzSKy+bfrLZVx0YHmQvtsqs7dExYESsqrUrB8ZcKwIwS3NPKaGq0w2QlPlCqUC3vQoQvhcZgPHPu2GkFYa7JEOdSKLknNPHaCRv80zx2RGF'}
 
+Note that receipts over signature transactions are a special case, for example:
+
+.. code-block:: bash
+
+    $ curl -X GET "https://<ccf-node-address>/app/receipt?transaction_id=2.35" --cacert networkcert.pem --key user0_privk.pem --cert user0_cert.pem
+    
+    {'leaf': 'fdc977c49d3a8bdf986176984e9432a09b5f6fe0c04e0b1c2dd177c03fdca9ec',
+     'node_id': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
+     'proof': [],
+     'root': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
+     'signature': 'MGQCMACklXqd0ge+gBS8WzewrwtwzRzSKy+bfrLZVx0YHmQvtsqs7dExYESsqrUrB8ZcKwIwS3NPKaGq0w2QlPlCqUC3vQoQvhcZgPHPu2GkFYa7JEOdSKLknNPHaCRv80zx2RGF'}
+
+The proof is empty, and the 'leaf' and 'root' fields are both set to the value being signed, which is the root of the Merkle Tree covering all transactions until the signature.
+
 Verifying a receipt is a two-phase process:
 
   - Combine ``leaf`` with the successive elements in ``proof`` to reproduce the value of ``root``. See :py:func:`ccf.receipt.root` for a reference implementation.
--- a/doc/use_apps/verify_tx.rst
+++ b/doc/use_apps/verify_tx.rst
@@ -81,6 +81,8 @@ Note that receipts over signature transactions are a special case, for example:
      'signature': 'MGQCMACklXqd0ge+gBS8WzewrwtwzRzSKy+bfrLZVx0YHmQvtsqs7dExYESsqrUrB8ZcKwIwS3NPKaGq0w2QlPlCqUC3vQoQvhcZgPHPu2GkFYa7JEOdSKLknNPHaCRv80zx2RGF'}
 
 The proof is empty, and the 'leaf' and 'root' fields are both set to the value being signed, which is the root of the Merkle Tree covering all transactions until the signature.
+This allows writing verification code that handles both regular and signature receipts without special casing, but it is worth noting that the 'leaf' value for signatures is not
+the digest of the signature transaction itself.
 
 Verifying a receipt is a two-phase process:
 
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -1196,6 +1196,43 @@ def test_receipts(network, args):
     return network
 
 
+@reqs.description("Validate all receipts")
+@reqs.supports_methods("receipt", "log/private")
+@reqs.at_least_n_nodes(2)
+def test_all_receipts(network, args):
+    primary, _ = network.find_primary_and_any_backup()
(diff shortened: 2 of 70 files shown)


tests/e2e_logging.py
```diff
@@ -1196,6 +1197,43 @@ def test_receipts(network, args):
...
+    with open(cert_path) as c:
+        node_cert = load_pem_x509_certificate(
+            c.read().encode("ascii"), default_backend()
+        )
+
+    with primary.client("user0") as c:
+        r = c.get("/app/commit")
+        max_view, seqno = [int(e) for e in r.body.json()["transaction_id"].split(".")]
+        view = 2
+        for s in random.sample(range(1, seqno - 1), 50):
```
achamayou, line 1215: Doing them all is fairly expensive it turns out.

lynshi:
Looks good to me; I like that receipt verification of signatures will still work with existing code. The special case will have to be well-documented though as I can imagine users being confused as to why proofs are occasionally missing.

It looks like `get_store_at()` is part of the CCF C++ API (just to confirm this is not an internal method we are using), so would its removal mean using `get_store_range` in all cases?

achamayou:
The proof is empty, since the transaction is a signature, but code that loops over it shouldn't require special-casing, as illustrated by our existing receipt verification code. I will expand the documentation though.

The replacement for `get_store_at()` is `get_state_at()`, which returns a state containing pointers to a store, a receipt and a transaction id. This will happen in a separate PR if at all, and not on the LTS branch in any event, since it is a obviously a breaking change.

tests/e2e_logging.py
eddyashton: A few things here:
- Why is this `seqno - 1`, does that mean we'll never sample the final signature?
- Can we replace `50` with `min(50, seqno)`, to avoid `ValueError: Sample larger than population or is negative`
- `random.sample` doesn't return sorted results, so the view-incrementing logic below is wrong (eg - we're fetching `42`, so count up to `5.42`, but then sample `35`, which is actually `3.35`)

tests/e2e_logging.py
```diff
@@ -1260,6 +1298,7 @@ def run(args):
             network = test_liveness(network, args)
         if args.package == "liblogging":
             network = test_receipts(network, args)
+            network = test_all_receipts(network, args)
```
eddyashton, line 1301: There's a chance that neither of these actually fetches a _signature_ receipt (if the `test_all` runs at the end of a large test where signatures are sparse, and we get unlucky with our sampling). Can we extend one of them to explicitly fetch a signature receipt (for the commit index, whatever it is), and even confirm that it has the structure we expect (empty proof and `leaf == root`)?

achamayou: I had the same thought, but for some reason decided it was hard to deliberately get a receipt for a signature tx, probably because it was late! You're absolutely right, always including the last commit does guarantee it will be a signature.

tests/e2e_logging.py
achamayou: Yes!

eddyashton (approved):

doc/use_apps/verify_tx.rst
```diff
@@ -68,6 +68,20 @@ This means that the request may return ``202 Accepted`` at first, with a suggest
...
+
+    $ curl -X GET "https://<ccf-node-address>/app/receipt?transaction_id=2.35" --cacert networkcert.pem --key user0_privk.pem --cert user0_cert.pem
+    
+    {'leaf': 'fdc977c49d3a8bdf986176984e9432a09b5f6fe0c04e0b1c2dd177c03fdca9ec',
+     'node_id': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
+     'proof': [],
+     'root': '06fef62c80b6471c7005c1b114166fd1b0e077845f5ad544ad4eea4fb1d31f78',
+     'signature': 'MGQCMACklXqd0ge+gBS8WzewrwtwzRzSKy+bfrLZVx0YHmQvtsqs7dExYESsqrUrB8ZcKwIwS3NPKaGq0w2QlPlCqUC3vQoQvhcZgPHPu2GkFYa7JEOdSKLknNPHaCRv80zx2RGF'}
+
+The proof is empty, and the 'leaf' and 'root' fields are both set to the value being signed, which is the root of the Merkle Tree covering all transactions until the signature.
```
eddyashton, line 83: We could be a little more explicit about the implications of this. Namely, you can verify both with the same code, starting from the leaf and iterating through the proof to produce the root, but this isn't really a leaf and you shouldn't use it as such if you're trying to reconstruct a full merkle tree.

achamayou: Ok, I've extended the description accordingly.

eddyashton (approved):

PR #2789, opened 2021-07-12
Use latest version of Doxygen

jumaffre:
Follow-up from #2787 

We now use the latest version of Doxygen, which we have to build from source.

### DIFF
--- a/.azure-pipelines-gh-pages.yml
+++ b/.azure-pipelines-gh-pages.yml
@@ -7,7 +7,7 @@ trigger:
 
 jobs:
   - job: build_and_publish_docs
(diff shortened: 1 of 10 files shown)


achamayou (approved):

achamayou:
@jumaffre we should be able to unpin sphinx in the same release.

eddyashton (approved):
Is this release of Doxygen available in Ubuntu 20.04, or are we going to have to keep building from source for a long time?

achamayou:
@eddyashton sadly no, 20.04 only has 1.8.17, so the answer to your second question is probably yes.

PR #2791, opened 2021-07-13
Table-based formatting of entries in `read ledger.py`

jumaffre:
~`read_ledger.py` didn't cope with `-1` keys (use in the new configuration table). This PR fixes it and also runs the utility in the CI so that we notice if it breaks again.~

We now format the keys and values of each table depending on the name of this one.


python/ccf/read_ledger.py
```diff
@@ -15,13 +15,17 @@ def indent(n):
...
-    s = bs.decode()
-    if s.isprintable():
-        return s
-    if len(bs) > 0 and len(bs) <= 8:
-        n = int.from_bytes(bs, byteorder="little")
-        return f"<u{8 * len(bs)}: {n}>"
-    return bs
+    try:
+        s = bs.decode()
+        if s.isprintable():
```
achamayou, line 20: I'm not convinced this is a good way to pretty print the contents of arbitrary buffers, it will sometimes display short, unprintable sequences as unsigned integers (signed, with a u prefix!?), and printable integers will be shown as strings. I would prefer if we defaulted to something more regular, even if that means having to rely on schema hints to make things nicer for governance.

wintersteiger: Yeah, we would need some sort of type hints to make this nicer. We could also just print it in hex.

achamayou: We could embed hints about governance tables, because we have to keep that schema extremely stable anyway, and allow users to supply their own hints as well. Defaulting to hex is reasonable, I agree.

achamayou:
As discussed, let's go with:

1. Print to keys and values as  hex
2. For public:ccf.gov.*, print keys as strings, values as strings
3. For public:ccf.gov.[tables with unsigned int keys], print keys as unsigned ints

tests/suite/test_suite.py
```diff
@@ -107,13 +107,17 @@
...
-    governance_history.test_tables_doc,
     # curve migration:
     reconfiguration.test_change_curve,
     recovery.test,
     # jwt
     jwt_test.test_refresh_jwt_issuer,
+    #
+    #
+    #
+    # Below tests should always stay last
```
ashamis, line 118: It might be useful to extend the comment explaining why this is the case

achamayou: For test_tables_doc, to goal is to collect all the tables the tests have used, parse the doc, and complain if they aren't mentioned. So naturally, to avoid missing tables, you want it last.
For the ledger is readable, it's the same idea, you want to test as much of it as possible.

Neither test will break if it's not last, but they will be less useful.

jumaffre:
> As discussed, let's go with:
> 
> 1. Print to keys and values as  hex
> 2. For public:ccf.gov.*, print keys as strings, values as strings
> 3. For public:ccf.gov.[tables with unsigned int keys], print keys as unsigned ints
> 
> The top-level call can take a list of (regexp, rendering method) that users can extend/override.

This should now be done and ready for review.

python/ccf/read_ledger.py
```diff
@@ -14,18 +14,49 @@ def indent(n):
...
+        "key": fmt_unsigned_int_little,
+        "value": fmt_json,
+    },
+    "^public:ccf\.gov\.(service|network|constitution).*$": {
+        "key": fmt_unsigned_int_little,
+        "value": fmt_json,
+    },
+    "^public:ccf\.gov\..*$": {"key": fmt_str, "value": fmt_json},
+    ".*": {"key": fmt_hex, "value": fmt_hex},
+}
```
eddyashton, line 48: I think this should be a list of pairs, rather than a dict. We don't need random lookup, we just iterate, and that makes it clearer that the order is intentional and meaningful.

python/ccf/read_ledger.py
```diff
@@ -14,18 +14,51 @@ def indent(n):
...
 
-def print_key(indent_s, k, is_removed=False):
-    k = stringify_bytes(k)
+
+def find_rule(tables_format_rules, target_table_name):
+    for table_name_re, format_rules in tables_format_rules.items():
+        if re.compile(table_name_re).match(target_table_name):
```
eddyashton, line 55: Not critical, since we don't have any performance bounds for this, but recompiling the regex here every time will get very slow for large ledgers/many rules. Should be possible to precompile (convert a `list-of-pairs-of-(string-and-rule)` to a `list-of-pairs-of-(regex-and-rule)`, and pass that here.

python/ccf/read_ledger.py
jumaffre: Yes, you're right, I'll change that (`dict`s are ordered since Python3.7 btw).

tests/suite/test_suite.py
jumaffre: I've added a comment to reflect this.

python/ccf/read_ledger.py
```diff
@@ -14,18 +14,57 @@ def indent(n):
...
 
-def stringify_bytes(bs):
-    s = bs.decode()
-    if s.isprintable():
-        return s
-    if len(bs) > 0 and len(bs) <= 8:
-        n = int.from_bytes(bs, byteorder="little")
-        return f"<u{8 * len(bs)}: {n}>"
-    return bs
+def fmt_unsigned_int_little(data):
```
achamayou, line 17:
```suggestion
def fmt_uint64_le(data):
```

tests/infra/consortium.py
```diff
@@ -364,7 +364,7 @@ def create_and_withdraw_large_proposal(self, remote_node):
         when trying to use ccf.ledger to read ledger entries.
         """
         proposal, _ = self.make_proposal(
-            "set_user", self.user_cert_path("user0"), {"padding": " " * 4096 * 5}
+            "set_user", self.user_cert_path("user0"), {"padding": "x" * 4096 * 5}
```
achamayou, line 367: what's the reason for this change?

achamayou (approved):

achamayou:
@jumaffre this is a user-visible change, can we have a CHANGELOG entry please?

PR #2798, opened 2021-07-14
Remove deprecated `--domain` option

jumaffre:
Taken from https://github.com/microsoft/CCF/pull/2628#discussion_r641426032 

This PR removes the now long-deprecated `--domain` option to `cchost`. This didn't feel particularly urgent but it actually causes a slight cosmetic issue where the same SAN would be included twice in the node certificate when a node was started with `--san iPAddress=<rpc_address>`.

~(To be backported to LTS branch.)~ As @eddyashton rightly pointed out, let's not do that for now as users may be relying on this option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Removed
 
-- Remove long-deprecated `--domain` argument from `cchost`. Node certificate Subject Alternative Names should be passed in via existing `--san` argument.
+- Remove long-deprecated `--domain` argument from `cchost`. Node certificate Subject Alternative Names should be passed in via existing `--san` argument (#2798).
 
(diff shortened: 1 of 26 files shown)


achamayou:
Can we have a changelog entry please?

achamayou (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Upgrade OpenEnclave from 0.17.0 to 0.17.1.
 
+### Removed
+
+- Remove long-deprecated `--domain` argument from `cchost`. Node certificate Subject Alternative Names should be passed in via existing `--san` argument.
```
jumaffre, line 16:
```suggestion
- Remove long-deprecated `--domain` argument from `cchost`. Node certificate Subject Alternative Names should be passed in via existing `--san` argument (#2798).
```

PR #2801, opened 2021-07-15
Decouple channel management from host session management

eddyashton:
This PR includes the final pieces from #2775:

- Node-to-node sessions on the host are not managed by the enclave. Instead they are created on-demand by the host, and should be cullable on idleness independently of the enclave's channel key management
- Key exchange protocol is rewritten to clarify what is serialised in each message, with a few functional tweaks. The core idea is that we more often obey a new `init` message, regardless of what we _were_ doing, and try to proceed with that exchange
- ChannelManager implementation is separated from interface, and only the manager (not the internal `Channel`s) are exposed by the API/used by the tests


achamayou (approved):

achamayou:
I've not managed to reproduce exactly what's happening in CI, but I've hit a run where one of the new nodes essentially fails to catch up, seemingly because it refuses to establish a channel successfully with the current primary (one of the old nodes).

It then stays stuck in WAITING_FOR_FINAL, seemingly forever.

```
2021-08-10T16:58:53.252032Z -0.006 0   [info ] ../src/node/node_state.h:657         | Node n[03b2170c2ad208e6d3a895e37d45c539cd89e0e644549328997e9aaf0aa98187] is waiting for votes of members to be trusted
2021-08-10T16:58:57.301365Z -0.073 0   [info ] ../src/consensus/aft/raft.h:3352     | Added raft node n[7914f7a2941bbf6c0ebcf8c45b165992687c641c5625888d962817e1bd717fc1] (127.168.43.206:40165)
2021-08-10T16:58:57.301419Z -0.073 0   [info ] ../src/consensus/aft/raft.h:3352     | Added raft node n[7a02f4becd3615bf6f92f0c5e0c6a5137cd53002a91210e5301c71c3fc4f5850] (127.109.68.148:37233)
2021-08-10T16:58:57.301442Z -0.073 0   [info ] ../src/consensus/aft/raft.h:3352     | Added raft node n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2] (127.20.34.178:38167)
2021-08-10T16:58:57.301463Z -0.073 0   [info ] ../src/consensus/aft/raft.h:3352     | Added raft node n[d3e5d133b737de6c691c7d79864a5052ab5c483f1335e7ca2f68f4e6e246f85f] (127.114.210.171:42657)
2021-08-10T16:58:57.301515Z        100 [info ] ../src/host/ledger.h:822             | Setting last known/commit index to 131
2021-08-10T16:58:57.301535Z -0.073 0   [info ] ../src/consensus/aft/raft.h:2866     | Becoming follower n[03b2170c2ad208e6d3a895e37d45c539cd89e0e644549328997e9aaf0aa98187]: 2
2021-08-10T16:58:57.307910Z -0.079 0   [info ] ../src/node/node_state.h:626         | Joiner successfully resumed from snapshot at seqno 131 and view 2
2021-08-10T16:58:57.312170Z -0.084 0   [info ] ../src/node/node_state.h:1591        | Network TLS connections now accepted
2021-08-10T16:58:57.313251Z -0.085 0   [info ] ../src/node/node_state.h:647         | Node has now joined the network as node n[03b2170c2ad208e6d3a895e37d45c539cd89e0e644549328997e9aaf0aa98187]: all domains
2021-08-10T16:58:57.331681Z -0.009 0   [info ] ../src/node/channels.h:902           | Resetting channel with n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2]
2021-08-10T16:58:57.402321Z -0.004 0   [info ] ../src/node/channels.h:902           | Resetting channel with n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2]
2021-08-10T16:58:57.408753Z -0.011 0   [fail ] ../src/node/channels.h:531           | <- n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2] (WAITING_FOR_FINAL): Peer certificate verification failed
2021-08-10T16:58:57.408818Z -0.001 0   [info ] ../src/node/channels.h:814           | Node channel with n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2] cannot receive authenticated message: not established, status=WAITING_FOR_FINAL
2021-08-10T16:58:57.409902Z -0.002 0   [info ] ../src/consensus/aft/raft.h:867      | Dropped invalid message from n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2]
2021-08-10T16:58:57.492951Z -0.001 0   [info ] ../src/node/channels.h:814           | Node channel with n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2] cannot receive authenticated message: not established, status=WAITING_FOR_FINAL
2021-08-10T16:58:57.492992Z -0.001 0   [info ] ../src/consensus/aft/raft.h:867      | Dropped invalid message from n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2]
2021-08-10T16:58:57.661553Z -0.001 0   [info ] ../src/node/channels.h:814           | Node channel with n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2] cannot receive authenticated message: not established, status=WAITING_FOR_FINAL
2021-08-10T16:58:57.661602Z -0.001 0   [info ] ../src/consensus/aft/raft.h:867      | Dropped invalid message from n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2]
2021-08-10T16:58:57.801621Z -0.056 0   [info ] ../src/node/channels.h:814           | Node channel with n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2] cannot receive authenticated message: not established, status=WAITING_FOR_FINAL
2021-08-10T16:58:57.801679Z -0.056 0   [info ] ../src/consensus/aft/raft.h:867      | Dropped invalid message from n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2]
2021-08-10T16:58:57.917641Z -0.001 0   [info ] ../src/node/channels.h:814           | Node channel with n[a5085d99316d81c177c5fab55986dbb7dfc5c7113b9e8dafe84bd52f98cd0ea2] cannot receive authenticated message: not established, status=WAITING_FOR_FINAL
```

tests/infra/partitions.py
```diff
@@ -140,17 +161,22 @@ def isolate_node(
...
             name += f" from node {other.local_node_id}"
 
-        if iptc.easy.has_rule("filter", CCF_IPTABLES_CHAIN, server_rule):
-            iptc.easy.delete_rule("filter", CCF_IPTABLES_CHAIN, server_rule)
+        rules = [
+            server_rule,
+            self.reverse_rule(server_rule),
+            client_rule,
+            self.reverse_rule(client_rule),
+        ]
```
eddyashton, line 169: The partitions test was failing, and this is the fix. Adding an explanation here, for posterity (and many thanks to @jumaffre for the help spotting this):

Each of these rules becomes a single `DROP` line in `iptables`, which drops _one direction_ of traffic for _one connection_ (potentially only scoped to specific ports). Generally when you do this, you are either a client _or_ a server, and you create 2 symmetric rules; one blocks 'sends', the other blocks 'responses'*, over the 2 known addresses. In our case, we're running both the client _and_ the server, and the 2 rules we created block the 'send' side _but not the respond_. This worked with the previous channels, because all the interesting messages were sent by a channel initiator as 'send's, and since _they_ were blocked we never produced anything to send over the (still available) 'respond' direction.

With the new channel code, an existing connection will be reused if it still exists. So a channel that was initiated by A to talk to B, when A was primary, may later be used by B to talk to A, when B becomes primary. This uses the 'respond' direction of the original connection, though as far as the code is concerned it's always just a 'send' locally (the same as any Raft response, which isn't distinguished from a 'send' by the time it reaches the host).

So this fix is just saying we always block a connection in 2 directions, both the initiator->recipient direction and recipient->initiator. And since we're doing that with 2 client-server address pairs (since either node may have acted as client at some point to initiate the connection), we end up with 4 total rules.

* A connection always has an initiating source, and a receiving destination. For this discussion, over that connection, anything from source to destination is a send, and anything from destination to source is a respond.

src/host/node_connections.h
```diff
@@ -327,113 +369,57 @@ namespace asynchost
...
+      while (unassociated_incoming.find(id) != unassociated_incoming.end())
       {
         id = next_id++;
       }
 
       return id;
     }
   };
-
-  using NodeConnectionsTickingReconnect = proxy_ptr<Timer<NodeConnections>>;
```
jumaffre, line 438: Why has the timer behaviour been removed from this class again?

PR #2805, opened 2021-07-15
Fix test_random_receipts to actually work across view changes

achamayou:
Fix for https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=29152&view=results

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -1225,9 +1225,15 @@ def test_random_receipts(network, args):
             int(e) for e in r.body.json()["transaction_id"].split(".")
         ]
         view = 2
-        seqnos = range(3, max_seqno)
+        genesis_seqno = 1
+        likely_first_sig_seqno = 2
+        last_sig_seqno = max_seqno
+        interesting_prefix = [genesis_seqno, likely_first_sig_seqno]
+        seqnos = range(len(interesting_prefix) + 1, max_seqno)
         for s in (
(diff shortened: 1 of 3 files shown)


eddyashton (approved):

jumaffre (approved):

tests/e2e_logging.py
```diff
@@ -1202,17 +1204,31 @@ def test_receipts(network, args):
...
         r = c.get("/app/commit")
-        max_view, seqno = [int(e) for e in r.body.json()["transaction_id"].split(".")]
+        max_view, max_seqno = [
+            int(e) for e in r.body.json()["transaction_id"].split(".")
+        ]
         view = 2
-        for s in sorted(random.sample(range(1, seqno), min(50, seqno))) + [seqno]:
+        seqnos = range(3, max_seqno)
+        for s in (
+            [1, 2] + sorted(random.sample(seqnos, min(50, len(seqnos)))) + [max_seqno]
```
ashamis, line 1230: there are some magic numbers here, if at all possible can they become variable/constants

ashamis (approved):

tests/e2e_logging.py
achamayou: Yes, I will do that. Just to explain the rationale here, 1 and 2 are to cover genesis and probably the first signature (it is today, but not guaranteed), because they're always a little special. The last one is to cover at least a signature for sure, since it's a commit point we know it will be.

The rest is just a random sampling, to keep runtime in check.

Issue #2554, opened 2021-05-05
Expose endorsed node certificate


letmaik:
Currently there's no way for apps to access the network-endorsed node certificate. Being able to do that is important for creating self-contained receipts.

achamayou:
The join protocol currently goes through these steps:

1. Joiner connects with self-signed cert, sends request containing quote
2. Joinee verifies the quote, and checks the user-data against the public key of the joiner
3. Joinee writes node_id->node self-signed cert to KV
4. Governance happens, joiner is voted as trusted
5. Joiner polls, eventually receives network secrets and endorses itself, creating endorsed identity

If we changed the end of that sequence to be:

5. Primary joiner in the same transaction that votes it as trusted, writes endorsed identity in KV
6. Joiner polls, eventually receives network secrets
7. On a hook on the endorsed_nodes tables, node picks up its endorsed identity, starts using it for TLS

We would both have a solution to this problem, and a relatively easy way to roll/refresh certificates going forward.

jumaffre:
As discussed with @letmaik and @achamayou, this should be good to implement. Interesting points:
- We should only record the public key (PEM) in the existing nodes table and create a new endorsed_nodes table to record the endorsed certificate
- We will need a new JavaScript binding for governance `transition_node_to_trusted` proposal as it needs to do the endorsement there.
- There will be some compatibility issues around signature verification in a live hybrid network (where is the primary cert I can use to verify their signature?), and endorsement (if a primary runs old code, I will still have to do the endorsement myself). Linked to #2238.

jumaffre:
Some more details after having implemented a first version:
![image](https://user-images.githubusercontent.com/42961061/125604028-bc286af7-bd77-4a1f-ab9a-a16296dbe84b.png)

- The joiner needs to pass its self-signed CSR to the primary on join, and it is this CSR that should be signed by the primary once the members trust the node.
- The CSR should also be recorded in the store in case the primary changes by the time the members trust the new node.
- (*) The SAN should be passed and stored separately (as opposed to being store in the CSR) because of some CSR parsing limitation of the mbedtls crypto backend (see https://github.com/ARMmbed/mbedtls/issues/2912). 
[...]

PR #2818, opened 2021-07-16
Read SubjectAltName field from CSR

jumaffre:
Part of #2554 

Historically, mbedtls was the only crypto backend we used for generating node/service certificates. However, mbedtls [doesn't support parsing x509 v3 extensions from a CSR](https://github.com/ARMmbed/mbedtls/issues/2912) and we circumvented that by specifying the SANs explicitly when the certificate was created from the CSR, rather than reading the SANs from the CSR directly. This was OK while the CSRs and certificates were created on the same nodes, but won't be the case after #2554 (joining node will pass its CSR to the primary which will sign it when the node becomes trusted). 

This PR adds the ability to read the extensions from the CSR and apply them to the certificate directly (**OpenSSL implementation only**). This avoids having to store the SANs for each node in the KV separately (i.e. CSR on its own is sufficient - changes to come in a separate PR). It also disables the creation of certificates with SANs for the mbedtls implementation. This means that the node and service certificates now always use OpenSSL.

### DIFF
--- a/src/crypto/mbedtls/key_pair.cpp
+++ b/src/crypto/mbedtls/key_pair.cpp
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the Apache 2.0 License.
 
+#include "key_pair.h"
+
 #include "curve.h"
 #include "ds/net.h"
 #include "entropy.h"
 #include "hash.h"
-#include "key_pair.h"
 
 #define FMT_HEADER_ONLY
 #include <fmt/format.h>
--- a/src/crypto/mbedtls/key_pair.cpp
+++ b/src/crypto/mbedtls/key_pair.cpp
@@ -1,12 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the Apache 2.0 License.
 
-#include "key_pair.h"
-
 #include "curve.h"
 #include "ds/net.h"
 #include "entropy.h"
 #include "hash.h"
+#include "key_pair.h"
 
 #define FMT_HEADER_ONLY
 #include <fmt/format.h>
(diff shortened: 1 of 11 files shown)


src/crypto/mbedtls/key_pair.cpp
```diff
@@ -286,192 +295,8 @@ namespace crypto
     registeredID = 8
   };
 
-  static inline int x509write_crt_set_subject_alt_name(
```
jumaffre, line 289: This was unused and the sister `x509write_crt_set_subject_alt_names()` (plural) is also deleted.

src/crypto/openssl/key_pair.cpp
```diff
@@ -281,35 +315,21 @@ namespace crypto
...
-      std::string all_alt_names;
-      bool first = true;
-      for (auto san : subject_alt_names)
+      for (size_t i = 0; i < extension_count; i++)
       {
-        if (first)
+        X509_EXTENSION* ext = sk_X509_EXTENSION_value(exts, i);
+        ASN1_OBJECT* obj = X509_EXTENSION_get_object(ext);
+        auto nid = OBJ_obj2nid(obj);
+        if (nid == NID_subject_alt_name)
```
jumaffre, line 328: Here, we only copy the SANs extension from the CSR to the final certificate. I don't think there's a use case for it yet, but should we be more permissive and copy _all_ extensions from the CSR?

src/crypto/openssl/key_pair.cpp
```diff
@@ -187,6 +188,42 @@ namespace crypto
     if (key)
       OpenSSL::CHECK1(X509_REQ_sign(req, key, EVP_sha512()));
 
+    if (!sans.empty())
+    {
+      Unique_STACK_OF_X509_EXTENSIONS exts;
+
+      std::string all_alt_names;
+      bool first = true;
+      for (const auto& san : sans)
```
achamayou, line 197: If the sans had a fmtlib formatter, you could fmt::join(sans, ", ")

src/crypto/openssl/key_pair.cpp
achamayou: Signing things we haven't validated sounds dangerous?

achamayou (approved):

src/crypto/openssl/key_pair.cpp
wintersteiger: +1, I also think we should just keep the SANs.

src/crypto/mbedtls/key_pair.cpp
wintersteiger: We could still set them, even if mbedTLS can't parse them back. Not worth the effort though, I think.

wintersteiger (approved):

PR #2820, opened 2021-07-19
Splitting up consensus protocols in doc

fritzalder:
This PR splits up the consensus document page into two parts:
 - Moves CFT/BFT to individual pages
 - New glossary entries for both
 - Some additional text to describe both on the common parent page

I think I caught all mentions of the old consensus page and its uses in the menu but please double check that there is no stale reference to the old consensus.rst.


doc/overview/consensus/index.rst
```diff
@@ -0,0 +1,86 @@
...
+CFT parameters can be configured when starting up a network (see :doc:`here </operations/start_network>`). The parameters that can be set via the CLI are:
+
+- ``raft-timeout-ms`` is the Raft heartbeat timeout in milliseconds. The Raft leader sends heartbeats to its followers at regular intervals defined by this timeout. This should be set to a significantly lower value than ``--raft-election-timeout-ms``.
+- ``raft-election-timeout-ms`` is the Raft election timeout in milliseconds. If a follower does not receive any heartbeat from the leader after this timeout, the follower triggers a new election.
+
+BFT Consensus Protocol
+----------------------
+.. warning:: CCF with BFT is currently still in development and should not be used in a production environment.
+
+BFT is still under development and should not be enabled in a production environment. More details on this mode is given :doc:`here <bft>`. There is an open research question of `node identity with Byzantine nodes <https://github.com/microsoft/CCF/issues/893>`_.
```
jumaffre, line 24: Repeats the warning above.

doc/overview/consensus/index.rst
```diff
@@ -0,0 +1,86 @@
...
+
+By default CCF runs with CFT. To run CCF with BFT the ``--consensus bft`` CLI argument must be provided when starting up the nodes (see :doc:`/operations/start_network` for starting up a network and nodes).
```
jumaffre, line 26: This is explicitly forbidden/disabled when running a release version of CCF so we may want to mention that it's only when CCF is built from source (link to https://microsoft.github.io/CCF/main/contribute/build_ccf.html).

jumaffre (approved):

doc/glossary.rst
```diff
@@ -9,12 +9,24 @@ Glossary
   Azure DCAP
     Intel SGX Data Centre Attestation Primitives which allows SGX attestation to be used within Microsoft Azure.
 
+  BFT
+    Byzantine Fault Tolerance is a type of fault tolerance that expects some nodes in the network to behave maliciously. This usually requires additional messages and checks on inputs that are received from other servers since these inputs can not be fully trusted (in contrast to :term:`CFT` where the servers are expected to behave honestly but may fail). Read more on BFT :ref:`here <overview/consensus/index:BFT Consensus Protocol>`.
```
achamayou, line 13:
```suggestion
    Byzantine Fault Tolerance is a type of fault tolerance that expects some nodes in the network to behave maliciously. This usually requires additional messages and checks on inputs that are received from other servers since these inputs are not trusted (in contrast to :term:`CFT` where the servers are expected to behave honestly but may fail). Read more on BFT :ref:`here <overview/consensus/index:BFT Consensus Protocol>`.
```

doc/overview/consensus/bft.rst
```diff
@@ -0,0 +1,167 @@
+Byzantine Fault Tolerance
+=========================
+
+Below we discuss the reconfiguration in a BFT setting.
```
achamayou, line 4: Technically the 2-tx scheme is not specific to BFT, it could be used in CFT as well. We want to use it for BFT because:

1. it only starts when the reconfiguration transaction is committed, so a reconfiguration attempt can never roll back
2. because it is not atomic, it has room for additional conditions, such as checking that the byzantine reconfiguration (a multiple transaction protocol) is complete before proceeding to the new config.

fritzalder: I included your comment in the text now 👍

doc/overview/consensus/index.rst
fritzalder: Thanks, it's removed now

doc/overview/consensus/index.rst
fritzalder: Clarified in cc4fff8

achamayou (approved):

PR #2821, opened 2021-07-19
TLA+ model of CCF's Raft

fritzalder:
This PR adds the TLA+ model and its associated docs. This unfortunately makes it a rather large PR but because of several cross-links I did not want to split it up (and have some links go into the nether until both PRs are merged).

Changes are:
- Moved the research page to its subfolder
- Added menu items linking to the papers for the other two research items
- Added TLA+ spec with Readmes on how to run them in the tla subfolder. This includes the full Raft spec and a reduced Raft spec that only does the signatures. Since only modelling signatures is way easier to handle with the runtime (minutes vs days), the reduced model may be helpful in the future.
- Added a TLA+ description and short entry in the research page.


doc/research/index.rst
```diff
@@ -1,11 +1,25 @@
 Research
 ========
 
+:doc:`TLA+ model of CCF's Raft modifications <raft-tla>`
+  CCF implements some modifications of Raft as it was originally proposed by Ongaro and Ousterhou. Specifically, CCF constraints that only appended entries that were signed by the primary can be committed. Any other entry that has not been globally commited is rolled back. Additionally, the CCF implementation introduced a variant of the reconfiguration that is different from the one proposed by the original Raft paper. In CCF CFT, reconfigurations are  done via one transaction (as described :doc:`here </overview/consensus>`).
```
eddyashton, line 5:
```suggestion
  CCF implements some modifications of Raft as it was originally proposed by Ongaro and Ousterhout. Specifically, CCF constrains that only appended entries that were signed by the primary can be committed. Any other entry that has not been globally committed is rolled back. Additionally, the CCF implementation introduced a variant of the reconfiguration that is different from the one proposed by the original Raft paper. In CCF CFT, reconfigurations are  done via one transaction (as described :doc:`here </overview/consensus>`).
```

doc/research/raft-tla.rst
```diff
@@ -0,0 +1,73 @@
...
+This file controls the constants as seen below. In addition to basic settings of how many nodes are to be model checked and their initial configuration, the model allows to place additional limitations on the state space of the program.
+
+.. literalinclude:: ../../tla/raft_spec/MCraft.tla
+    :language: text
+    :start-after: SNIPPET_START: mc_config
+    :end-before: SNIPPET_END: mc_config
+
+Possible state transitions in the model
+---------------------------------------
+During the model check, the model checker will exhaustively search through all possible state transitions. The belo snippet defines the possible state transitions at each step. Each of these transitions has additional constraints that have to be fulfilled for the state to be an allowed step. For example, BecomeLeader is only a possible step if the selected node has enough votes to do so.
```
eddyashton, line 22:
```suggestion
During the model check, the model checker will exhaustively search through all possible state transitions. The below snippet defines the possible state transitions at each step. Each of these transitions has additional constraints that have to be fulfilled for the state to be an allowed step. For example, ``BecomeLeader`` is only a possible step if the selected node has enough votes to do so.
```

doc/research/raft-tla.rst
```diff
@@ -0,0 +1,73 @@
...
+    :language: text
+    :start-after: SNIPPET_START: signing
+    :end-before: SNIPPET_END: signing
+
+Reconfiguration steps
+---------------------
+
+The one transaction reconfiguration is already described :doc:`here </overview/consensus>`. In the TLA model, a reconfiguration is initiated by the Leader which appends an arbitrary new configuration to its own log. This also triggers a change in the ``Configurations`` variable which keeps track of all running configurations.
+
+In the following, this ``Configurations`` variable is then checked to calculate a Quorum and to check which nodes should be contacted or received messages from.
```
eddyashton, line 68:
```suggestion
In the following, this ``Configurations`` variable is then checked to calculate a quorum and to check which nodes should be contacted or received messages from.
```

tla/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+Each model controls its limits in the MCraft.tla file where the constants can be modified to increase or decrease the size of the model checking.
+
+Running TLC on our models can take any time between minutes (for small configurations) and days (especially for the full model with reconfiguration) on a 128 core VM (specifically, we used an [Azure HBv3 instance](https://docs.microsoft.com/en-us/azure/virtual-machines/hbv3-series)).
+
+### A note on TLC performance
+
+TLC works best if it can utilize all system resources. For this, consider using the TLC options `-workers auto` to use all cores and the Java VM options `-Xms435G -Xmx435G` to enforce a specific RAM usage (435GB in this case as provided by the 128 core HBv3 VM).
+
+During development, it helps to use simulation mode which performs a depth-first search of the search tree (instead of the default breadth first that is very slow). Turn on the simulation mode with `-simulate -depth 100000` (using a very large number as a maximum depth). Note that this mode never completes (but will find errors in minutes instead of hours).
```
eddyashton, line 27: Just a comment: These kinds of note are _extremely_ useful for anyone looking to work with the model in future, so thanks for adding them!

tla/raft_spec/MCraft.tla
```diff
@@ -0,0 +1,49 @@
...
+InitialConfig_mc == [tmp EXCEPT ![NodeFour] = Pending,
+                                ![NodeFive] = Pending]
+
+\*  SNIPPET_START: mc_config
+\* Limit the terms that can be reached. Needs to be set to at least 3 to
+\* evaluate all relevant states. If set to only 2, the candidate_quorum
+\* constraint below is too restrictive.
+TermLimit_mc == 4
+
+\* Limit number of requests (new entires) that can be made
```
eddyashton, line 17:
```suggestion
\* Limit number of requests (new entries) that can be made
```

tla/raft_spec/MCraft.tla
```diff
@@ -0,0 +1,49 @@
+
+---------- MODULE MCraft ----------
+EXTENDS ccfraft, TLC
+
+PossibleServer_mc == {NodeOne, NodeTwo, NodeThree, NodeFour, NodeFive}
+InitialServer_mc == {NodeOne, NodeTwo, NodeThree}
+tmp == [s \in PossibleServer_mc |-> Follower]
+InitialConfig_mc == [tmp EXCEPT ![NodeFour] = Pending,
+                                ![NodeFive] = Pending]
```
eddyashton, line 9: The first node launched in a CCF network actually does a `force_become_leader` step, rather than waiting for an initial election. So in practice the initial state for us is `{Primary, Follower, Follower, Pending, Pending}`. Not sure how useful this is, but it could remove some searching of equivalent states early in the execution (ie - once the state is stable after the initial election, we don't care whether the winner was `NodeOne`, `NodeTwo`, or `NodeThree`, so we can avoid searching all of them by just starting _from_ that point with the winner as `NodeOne`)?

tla/raft_spec/MCraft.tla
```diff
@@ -0,0 +1,49 @@
...
+
+\* Limit number of times a RetiredLeader server sends commit notifications
+CommitNotificationLimit_mc == 1
+
+\* We made several restrictions to the state space of Raft. However since we
+\* made these restrictions, Deadlocks can occur at places that Raft would in
+\* real-world deployments handle graciously. 
+\* One example of this is if a Quorum of nodes becomes Candidate but can not
+\* timeout anymore since we constrained the terms. Then, an artificial Deadlock
+\* is reached. We solve this below. If TerLimit is set to any number >2, this is
```
eddyashton, line 37:
```suggestion
\* is reached. We solve this below. If TermLimit is set to any number >2, this is
```

tla/raft_spec/README.md
```diff
@@ -0,0 +1,39 @@
...
+| :------ | :---------- |
+| Signatures | In CCF, only entries signed by the primary can be committed (in contrast to all entries in normal Raft) |
+| Reconfiguration | CCF implements a one-transaction version of reconfiguration for Raft (read the description [here](https://microsoft.github.io/CCF/main/overview/consensus.html)). |
+
+## Checked invariants
+
+| Name | Short description |
+| :--- | :---------------- |
+| LogInv | Committing an entry never changes the previous log indices, neither does it decrease the size of the committed log. |
+| CandidateTermNotInLogInv | Candidates that have a chance to win an election will not be elected to a term that already had a leader that appended entries. |
```
eddyashton, line 26: This sounds a little weaker than it could be. Why do we need the "that appended entries" condition? I believe we should only ever have a single leader in each term - is that not true under our current reconfiguration approach?

tla/raft_spec/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+The given model consists of three files:
+
+- `ccfraft.tla` : The core specification that implements the features as seen below.
+- `MCraft.tla` : The model checking implementation for the spec. Sets the constants and can be modified for each run to
+  fine tune the settings.
+- `MCraft.cfg` : The core configuration which invariants are to be checked etc. Usually stays untouched during normal
+  model checking.
+
+To run this spec, run the `MCraft.tla` with the `-deadlock` parameter to disable liveness checking (see explanation
```
jumaffre, line 11: Would it be worth having a simple `run.sh` script in the repo as well?

tla/raft_spec/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Implemented features
+
+| Feature | Description |
+| :------ | :---------- |
+| Signatures | In CCF, only entries signed by the primary can be committed (in contrast to all entries in normal Raft) |
+| Reconfiguration | CCF implements a one-transaction version of reconfiguration for Raft (read the description [here](https://microsoft.github.io/CCF/main/overview/consensus.html)). |
+
+## Checked invariants
+
+| Name | Short description |
```
jumaffre, line 23: It could be nice to indicate which ones were in the original Raft model and which ones were added by you? Unless they all fall into the latter category, in which case we should probably mention briefly here.

tla/README.md
fritzalder: Thanks! I figured maybe it helps to lessen the pain of the next person :)

doc/research/index.rst
```diff
@@ -1,11 +1,25 @@
 Research
 ========
 
+:doc:`TLA+ model of CCF's Raft modifications <raft-tla>`
+  CCF implements some modifications of Raft as it was originally proposed by Ongaro and Ousterhout. Specifically, CCF constrains that only appended entries that were signed by the primary can be committed. Any other entry that has not been globally committed is rolled back. Additionally, the CCF implementation introduced a variant of the reconfiguration that is different from the one proposed by the original Raft paper. In CCF CFT, reconfigurations are  done via one transaction (as described :doc:`here </overview/consensus>`).
```
achamayou, line 5:
```suggestion
  CCF implements some modifications to Raft as it was originally proposed by Ongaro and Ousterhout. Specifically, CCF constrains that only appended entries that were signed by the primary can be committed. Any other entry that has not been globally committed is rolled back. Additionally, the CCF implementation introduced a variant of the reconfiguration that is different from the one proposed by the original Raft paper. In CCF CFT, reconfigurations are  done via one transaction (as described :doc:`here </overview/consensus>`).
```

doc/research/raft-tla.rst
```diff
@@ -0,0 +1,73 @@
...
+    :end-before: SNIPPET_END: init_values
+
+Basic functions and log changes
+-------------------------------
+Below, we shortly describe some basic functionality of the Raft model. Note that this is only a selection of the model, please refer to the full model for a full picture.
+
+Timeout
+~~~~~~~
+
+Since TLA does not model time, any node can timeout at any moment as a next step. Since this may lead to an infinite state space, we limited the maximum term any node can reach. While this would be overly constraining in any actual program, the model checker will ensure to also explore those states that are feasible within these limits. Since interesting traces can already be generated with one or better two term changes, this approach is feasible to model reconfigurations and check persistence.
```
achamayou, line 46:
```suggestion
Since TLA does not model time, any node can time out at any moment as a next step. Since this may lead to an infinite state space, we limited the maximum term any node can reach. While this would be overly constraining in any actual program, the model checker will ensure to also explore those states that are feasible within these limits. Since interesting traces can already be generated with one or better two term changes, this approach is feasible to model reconfigurations and check persistence.
```

achamayou (approved):

tla/raft_spec/README.md
achamayou: It isn't, because members of an old config that haven't observed commit of the reconfiguration can elect a leader, even after the new (sufficiently non-overlapping) config has elected its own leader.

It is fine for that to happen, because the leader elected in the old config will only be able to "catch it up" to the reconfiguration, at which point its members will retire. So there may be two leaders at the same term, but only one of them would ever eventually be able to perform writes. The other is only re-transmitting in a dead-end config subset.

tla/raft_spec/MCraft.tla
fritzalder: That is a very good point. I'll keep the current model as is but will add a PR later with a slightly modified initialization for node states (we can also already start with a partial log in my opinion to make things interesting). Thanks for the idea!

tla/raft_spec/README.md
fritzalder: Good point, I clarified this

tla/raft_spec/README.md
fritzalder: Good idea, I added scripts in the tla folder. The reduced spec should run in a few seconds, can you try to run it and report back whether it was usable and documented? Maybe both @achamayou and @jumaffre

jumaffre: Tested locally and it worked fine on my local VM. Note that I had to install Java beforehand (`$ sudo apt install default-jre`).

Issue #2803, opened 2021-07-15
Incorrect suffix retained after election


eddyashton:
The changes in #2801 have hit a Raft bug very similar to the one described in #2189 (and believed fixed in #2192). Specifically, that fix to "accept a rollback instruction from the new primary" is not reached if the node rejects the AppendEntries early in processing, for instance because it is not contiguous with its current ledger or it disagrees on the previous term. The former is legitimate, but the latter is not and can lead to permanent commit lock.

From spelunking in some logs, one incorrect flow was observed:
- Node 0 is briefly primary in term 8
- Node 0 produces a signature at 8.25, but doesn't manage to tell anyone about it
- Node 1 becomes primary in term 9
- Node 1 produces a transaction at 9.25, and proceeds to 9.26
- Node 1 sends AppendEntries to 0 containing 9.25 -> 9.26
- Node 0 sees an AppendEntries in term 9, and dutifully rolls back to its `last_committable` (setting `is_new_follower`, but hoping that its 25 is still useful)
- Node 0 processes the AppendEntries and says "no, that looks like junk to me"

0's logs show it rejecting an AppendEntries it should have accepted:
```
24:49.243 |00| Received append entries: 9.25 to 9.26 (from 1=bf9b in term 9) (raft.h:1350)
24:49.243 |00| Previous term for 25 should be 8 (raft.h:1439)
24:49.243 |00| Recv append entries to n[0=92bf] from n[1=bf9b] but our log at 25 has the wrong previous term (ours: 8, theirs: 9) (raft.h:1461)
24:49.243 |00| Send append entries response from 0=92bf to 1=bf9b for index 25: 1 (raft.h:2084)
```

This is a more general problem than that precise flow. For instance the locked node could have been any follower of 0 in term 8. Its not critical that they produced the signature, just that it is the final entry in their log and disagrees with an AppendEntries they should be obeying.

I'm currently considering 2 possible fixes:
- Obey an AppendEntries rollback in more circumstances. This should just involve moving a block of code earlier in the `recv_append_entries` function. However reasoning about exactly where it should go is tricky, as we want to (if possible) keep the current behaviour that it is robust to replay (we only do this immediately after becoming follower, if we did it blindly then a malicious host could trivially cause us to repeatedly roll back to this point and lose state).
[...]

PR #2823, opened 2021-07-20
Refresh Raft driver and unit test

eddyashton:
To repro + test the fix for #2803, I'd like to use some of our existing tests. I'd hoped that it would be as simple as introducing a shuffle operation in some of these scenarios, but that hasn't produced a repro. However to get there I've had to refresh the driver (which had rotted over consensus changes, and was no longer testing anything useful), and believe its now a useful baseline for the Raft behaviour that we can extend with some repros + fixes in future.


.gitignore
```diff
@@ -9,7 +9,7 @@ CMakeFiles/
 .vs/
 libuv/
 raft_tests.md
-tests/raft_scenarios
+tests/raft_scenarios/scenario-*
```
eddyashton, line 12: Making this more precise to allow additional scenarios to be added. These `scenario-X` files are now actually generated in the build directory (or local directory, wherever `raft_scenarios_runner` is executed), but I don't want to add the scenarios you all have sitting around

src/consensus/aft/test/driver.cpp
```diff
@@ -24,12 +24,28 @@ constexpr auto shash = ds::fnv_1a<size_t>;
...
+  {
+    throw std::runtime_error(
+      "Too few arguments - first must be path to scenario");
+  }
+
+  std::ifstream fstream;
+  fstream.open(argv[1]);
+
+  string line;
+  while (getline(fstream, line))
```
eddyashton, line 40: Making this take a filename rather than reading from stdin, so its easier to run through a debugger.

src/consensus/aft/test/driver.cpp
```diff
@@ -65,15 +93,11 @@ int main(int argc, char** argv)
...
-        driver->state_all();
-        break;
       case shash("replicate"):
-        assert(items.size() == 4);
+        assert(items.size() == 3);
         data = std::make_shared<std::vector<uint8_t>>(
-          items[3].begin(), items[3].end());
-        driver->replicate(items[1], stoi(items[2]), data);
+          items[2].begin(), items[2].end());
+        driver->replicate(items[1], data);
```
eddyashton, line 100: This `replicate` step previously tried to dictate the node and index to replicate on and at. I don't think this is a helpful abstraction - it is only valid to `replicate` on a node if they currently believe they are a primary, and the index should always be monotonic. I've changed this to take the `term` to replicate in (its valid and interesting for a partitioned primary, who is still in term X, to replicate _in_ term X, while the other nodes have advance to term Y). Additionally this term can be the special value `latest` to replicate something in the last term-with-a-primary, which is what the scenario generator wants to do.

src/consensus/aft/test/driver.h
```diff
@@ -4,25 +4,60 @@
...
+    RAFT_DRIVER_OUT << "  Node" << _id << "->>Node" << _id
+                    << ": ledger put s: " << stringify(data) << std::endl;
+    aft::LedgerStubProxy::put_entry(data, globally_committable, force_chunk);
+  }
+
+  void truncate(aft::Index idx) override
+  {
+    RAFT_DRIVER_OUT << "  Node" << _id << "->>Node" << _id
+                    << ": truncate i: " << idx << std::endl;
+    aft::LedgerStubProxy::truncate(idx);
```
eddyashton, line 44: Minor: Moving these behaviour changes (additional logging) to "override the virtual method" rather than "define the `STUB_LOG` macro", so that they're localised to these driver files that use them.

src/consensus/aft/test/driver.h
```diff
@@ -74,59 +109,130 @@ class RaftDriver
...
-    log(node_id, tgt_node_id, s.str());
+    const auto s = fmt::format(
+      "append_entries ({}.{}, {}.{}] (term {}, commit {})",
+      ae.prev_term,
+      ae.prev_idx,
+      ae.term_of_idx,
+      ae.idx,
+      ae.term,
+      ae.leader_commit_idx);
+    log(node_id, tgt_node_id, s);
```
eddyashton, line 154: Done a bunch of reformatting of these log lines, to be slightly more concise and match our (or at least my) current understanding of the messages. In particular, AppendEntries contains a non-inclusive _previous_ term + index (TxID), and an inclusive _final_ term + index (TxID). I've started writing this as a range `(2.3, 5.8]`.

src/consensus/aft/test/driver.h
```diff
@@ -302,4 +486,88 @@ class RaftDriver
...
+        RAFT_DRIVER_OUT
+          << fmt::format(
+               "  Note over Node{}: Term {} doesn't match term {} on Node{}",
+               node_id,
+               raft->get_term(),
+               target_term,
+               target_id)
+          << std::endl;
+        all_match = false;
+      }
```
eddyashton, line 518: These were asserts (hence the name `assert_state_sync` for this step), but I realised we really want them to trigger a failure in non-Debug builds, and to produce a verbose description of the failure. So now if any of these conditions aren't met, we output a note in the Mermaid diagram describing the mismatch, which is really helpful in analysing bad traces.

src/consensus/aft/test/logging_stub.h
```diff
@@ -72,15 +69,52 @@ namespace aft
...
-    std::list<std::pair<ccf::NodeId, RequestVote>> sent_request_vote;
-    std::list<std::pair<ccf::NodeId, AppendEntries>> sent_append_entries;
-    std::list<std::pair<ccf::NodeId, RequestVoteResponse>>
-      sent_request_vote_response;
-    std::list<std::pair<ccf::NodeId, AppendEntriesResponse>>
-      sent_append_entries_response;
+    // Using a deque so we can both pop from the front and shuffle
+    using MessageList =
+      std::deque<std::pair<ccf::NodeId, std::vector<uint8_t>>>;
+    MessageList messages;
```
eddyashton, line 75: Collapsing these separate queues into a single combined queue, so its easier to shuffle/drop arbitrary messages. Means we need to peek at the serialised form to rediscover the message type in more places if we're looking for something specific, but I think this is worth it.

src/consensus/aft/test/main.cpp
```diff
@@ -18,14 +18,18 @@ using ms = std::chrono::milliseconds;
...
 
 threading::ThreadMessaging threading::ThreadMessaging::thread_messaging;
 std::atomic<uint16_t> threading::ThreadMessaging::thread_count = 1;
 
 std::vector<uint8_t> cert;
 
+aft::ChannelStubProxy* channel_stub_proxy(const TRaft& r)
+{
+  return (aft::ChannelStubProxy*)r.channels.get();
+}
```
eddyashton, line 31: Replacing the `((aft::ChannelStubProxy*)r0.channels.get())` pattern with `channel_stub_proxy(r0)`, and in most places giving that a name like `r0c` for readability.

tests/raft_scenarios_runner.py
```diff
@@ -1,91 +1,100 @@
...
 
-    generate_scenarios(path)
+    files = expand_files(args.files)
+
+    if args.gen_scenarios:
+        files += generate_scenarios()
+
+    ostream = sys.stdout
 
-    with open(os.path.join(doc, RAFT_TEST_FILE_NAME), "w") as raft:
```
eddyashton, line 69: Making this test more like all our other test apps. Rather than writing to a file, it writes to stdout. It still strips out the mermaid-specific bits, and surrounds them with some Markdown decoration, but it dumps any errors at the end rather than the start (for access from CTest). Can then be piped to a `.md` file to preview the resulting trace diagrams.

src/consensus/aft/test/driver.cpp
achamayou: My long term view is we would discard this frontend in favour of a bit of pybind11, because that's ultimately where we want to drive things from. Never did find the time for it though.

achamayou (approved):

PR #2825, opened 2021-07-21
Raft fix: Don't infer future terms when AppendEntries ends on signature

eddyashton:
This is a fix for another Raft bug discovered as part of #2823.

When we get an AppendEntries that ends between signatures, we use the `term_of_idx` to tell us if the unsigned suffix is in a _new_ term (starting after the last signature we saw). However, we also took this path when the AppendEntries _did_ end in a signature, resulting in us incorrectly assuming the transaction _after_ the signature was in the same term, though it may not even exist yet.


achamayou (approved):

src/consensus/aft/raft.h
```diff
@@ -2099,11 +2099,22 @@ namespace aft
...
+        // the current term. This can only happen before _any_ transactions have
+        // occurred, when processing a heartbeat at index 0, which does not
+        // happen in a real node (due to the genesis transaction executing
+        // before ticks start), but may happen in tests.
         state->view_history.update(1, r.term);
       }
       else
       {
-        state->view_history.update(lci + 1, r.term_of_idx);
+        // The end of this append entries was not a signature, but may be in a
```
achamayou, line 2111:
```suggestion
        // The end of this append entries (r.idx) was not a signature, but may be in a
```

src/consensus/aft/raft.h
```diff
@@ -2099,11 +2099,22 @@ namespace aft
...
+        // new term. If it's a new term, this term started immediately after the
+        // previous signature we saw
```
achamayou, line 2113:
```suggestion
        // previous signature we saw (lci, last committable index)
```

jumaffre (approved):

achamayou:
@eddyashton could you backport this to LTS please?

PR #2828, opened 2021-07-21
Add ability to parse public key from CSR

jumaffre:
Part of #2554

Since joining nodes will pass their CSR to the primary on join, the primary should verify that the public key in the CSR matches the public key of the TLS client. This PR doesn't actually perform the check, but only add the ability to parse the CSR (OpenSSL implementation only).


achamayou:
@jumaffre since the CSR is signed by the joiner, do we want to verify that signature too?

achamayou (approved):

jumaffre:
> @jumaffre since the CSR is signed by the joiner, do we want to verify that signature too?

Very good point. I had assumed that CSR verification was already done by the parsing functions for both OpenSSL and mbedtls but this wasn't the case. I've added the checks to this PR.

PR #2834, opened 2021-07-22
Fix `check_can_progress` in Python infra

jumaffre:
Fix a minor issue in the Python infra when checking for progress on a service. We now wait for a specific write transaction to become committed rather than a previous commit level to advance. Otherwise, as some transactions are automatically emitted by the service (e.g. signatures), the test might move on too quickly and eventually lead to an uncommitted suffix before an election, i.e. #2803.

This resolves issues like, where we a signature transaction gets emitted at seqno `25` but we should wait for commit over `26`. 
image


tests/infra/checker.py
```diff
@@ -40,35 +40,41 @@ def __call__(self, rpc_result, result=None, error=None):
 
 
 def check_can_progress(node, timeout=3):
+    # Check that a write transaction issued on one node is eventually
+    # committed by the service
+    with node.client("user0") as uc:
+        submitted_seqno = uc.post(
+            "/app/log/private", {"id": 42, "msg": "Hello world"}
+        ).seqno
```
eddyashton, line 48: I don't think we should be checking only `seqno` here. While its correct as long as nothing else is talking to the service, it feels like we're leaving a bug to bite us in future (when _this_ thing never happened but something _else_ grabbed the same seqno in a new term, did we really make progress or just get a signature from a new primary? I guess if its committed that's _some_ progress, but still misleading). If we want to match what the comment above says, I think this should be using `/node/tx` rather than `/node/commit`.

eddyashton:
We have the same pattern in `partitions_test.py` in `test_isolate_and_reconnect_primary`:

```python
    # Check reconnected former primary has caught up
    with primary.client() as c:
        timeout = 5
        r = c.get("/node/commit")
        while time.time() < end_time:
            current_tx = TxID.from_str(
                c.get("/node/commit").body.json()["transaction_id"]
            )
            if current_tx.seqno >= new_tx.seqno:
                return network
            time.sleep(0.1)
        details = c.get("/node/consensus").body.json()
        assert False, f"Stuck at {r}: {pprint.pformat(details)}"
```

I believe it should be fixed there as well.

tests/infra/checker.py
achamayou: I agree, and the more I think about it, the more I think we should think of our usage of `/commit` critically. It's easy to misuse.

jumaffre: Good point. We already have the machinery to wait for a commit for a specific `TxID` so I'll update the code to use this instead.

tests/infra/checker.py
```diff
@@ -40,35 +38,26 @@ def __call__(self, rpc_result, result=None, error=None):
...
-        while time.time() < end_time:
-            current_tx = TxID.from_str(
-                c.get("/node/commit").body.json()["transaction_id"]
-            )
-            if current_tx.seqno > original_tx.seqno:
-                assert False, "Commit advanced when it shouldn't have"
-            time.sleep(0.1)
-        return True
+    # Check that a write transaction issued on one node is _not_
+    # committed by the service
```
achamayou, line 55:
```suggestion
    # committed by the service by a timeout
```

tests/infra/checker.py
```diff
@@ -40,35 +38,26 @@ def __call__(self, rpc_result, result=None, error=None):
...
+    with node.client("user0") as c:
+        r = c.post("/app/log/private", {"id": 42, "msg": "Hello world"})
+        try:
+            c.wait_for_commit(r, timeout=timeout)
+        except TimeoutError:
+            pass
+        else:
+            assert False, "Commit advanced when it shouldn't have"
```
achamayou, line 63:
```suggestion
            assert False, "Commit unexpectedly advanced past {r.view}.{r.seqno}"
```

tests/infra/checker.py
```diff
@@ -40,35 +38,26 @@ def __call__(self, rpc_result, result=None, error=None):
...
+    # Check that a write transaction issued on one node is eventually
+    # committed by the service
+    with node.client("user0") as c:
+        r = c.post("/app/log/private", {"id": 42, "msg": "Hello world"})
+        try:
+            c.wait_for_commit(r, timeout=timeout)
+            return r
+        except TimeoutError:
+            details = c.get("/node/consensus").body.json()
+            assert False, f"Stuck at {r.seqno}: {pprint.pformat(details)}"
```
achamayou, line 50:
```suggestion
            assert False, f"Stuck before {r.view}.{r.seqno}: {pprint.pformat(details)}"
```

tests/partitions_test.py
```diff
@@ -111,22 +109,15 @@ def test_isolate_and_reconnect_primary(network, args):
...
-            if current_tx.seqno >= new_tx.seqno:
-                return network
-            time.sleep(0.1)
-        details = c.get("/node/consensus").body.json()
-        assert False, f"Stuck at {r}: {pprint.pformat(details)}"
+        try:
+            c.wait_for_commit(new_tx_resp, timeout=5)
+        except TimeoutError:
+            details = c.get("/node/consensus").body.json()
+            assert False, f"Stuck at {new_tx_resp.seqno}: {pprint.pformat(details)}"
```
achamayou, line 120:
```suggestion
            assert False, f"Stuck before {new_tx_resp.view}.{new_tx_resp.seqno}: {pprint.pformat(details)}"
```

achamayou (approved):

achamayou:
Not a bugfix, but will improve test (and therefore release pipeline) stability, so let's bring it to the LTS at well.

eddyashton (approved):

PR #2835, opened 2021-07-22
Raft fix: Respond with correct index when NACKing AppendEntries

eddyashton:
Fix for #2803.

This fix (pending a full CI run) is very small and self-contained so I'm opening this in a standalone PR so it could be easily cherry-picked.

I have a consistent repro for the original issue, as evidence this fixes it, but it is written as a Raft scenario and relies on the changes in #2823, so I'll add it later once that is merged. If you're interested, I've included it below.

<details>
  <summary>Repro</summary>
  
  ```
nodes,0,1,2
connect,0,1
connect,1,2
connect,0,2

# Node 0 starts first, and begins sending messages first
periodic_one,0,110
dispatch_all

periodic_all,10
dispatch_all

state_all

replicate,1,hello world
periodic_all,10
dispatch_all

replicate,1,saluton mondo
periodic_all,10
dispatch_all

periodic_all,10
dispatch_all

state_all

# Node 0 is partitioned, and replicates another item
disconnect,0,1
disconnect,0,2

replicate,1,drop me 1
periodic_all,10
dispatch_all

state_all

# Node 1 calls an election, and wins
periodic_one,1,110
dispatch_all

periodic_all,10
dispatch_all  #< This AppendEntries starts at a point that Node 1 would accept, but is dropped

state_all

# Node 1 replicates a new entry, resulting in the same seqno as Node 0
replicate,2,keep me 1
periodic_all,10
dispatch_all

replicate,2,keep me 2
periodic_all,10
dispatch_all

state_all

# Network heals
connect,0,1
connect,0,2

# Node 0 is brought up-to-date?
periodic_all,10
dispatch_all   #< This AppendEntries starts from a later point in Node 1's term, which Node 0 refuses!

periodic_all,10
dispatch_all

state_all
assert_state_sync
```
</details>

PR #2840, opened 2021-07-23
Add additional Raft scenarios

eddyashton:
Follow-up to #2823 and #2835.

This includes some further refresh of the `raft_driver` and `raft_test` stubs so that they test more of the real Raft behaviour (including ledger entries rather than empty AppendEntries, tracking the term and seqno of each entry), and some new scenarios enabled by those improvements that test the correct retention/dropping of ledger suffixes around elections.


tests/raft_scenarios/suffix_collision
```diff
@@ -0,0 +1,70 @@
...
+# Network heals
+connect,0,1
+connect,0,2
+
+# Node 0 is brought up-to-date
+periodic_all,10
+dispatch_all
+
+periodic_all,10
+dispatch_all
```
eddyashton, line 67: For the sake of completeness, here's the generated Mermaid traces showing the situation we were stuck in before #2835:

Node 0 hears about the new term and rolls back, but NACKs with a higher-than-matching index
image

Node 1 keeps sending AppendEntries that start too late, and Node 0 (correctly) keeps refusing them
image


Now that #2835 is in, we NACK with a matching index
image

And as a result we get an AppendEntries starting from `1.2` rather than `2.3`
image

achamayou (approved):

PR #2844, opened 2021-07-23
Record endorsed node certificate in KV store

### LINKED ISSUE #2554: Expose endorsed node certificate

jumaffre:
Resolves #2554 

**[Still very much WIP, in particular with regards to compatibility with 1.x LTS]**

Node endorsed certificates are now recorded in a new `public:ccf.gov.nodes.endorsed_certificates` table (rather than self-signed certificates in the existing nodes table). 

This has the following implications:
- New joiners pass their self-signed CSRs to the current primary, which records it in the store on join.
- Member proposal to trust a node also generates and records the new joiner's endorsed certificate. 
- New joiners should pick up their endorsed certificates from a new KV hook.
- 2.x nodes should still be able to verify node signatures from a 1.x ledger.

Still in progress:
- [ ] Change the moment TLS connections are endorsed on new joiners
- [ ] Compatibility: 
  - [ ] Join protocol (2.x node joining a 1.x service).
  - [x] 1.x Ledger 
  - [ ] Upgrade constitution in `lts_compatibility` test
- [ ] Unify recovery first and genesis transactions
- [ ] Changelog


src/node/nodes.h
```diff
@@ -48,15 +47,43 @@ namespace ccf
...
+
+    /**
+     * Fields below are deprecated
+     */
+
+    /** Deprecated as of 2.x.
+     * Node certificate. Only set for 1.x releases. Further releases record
+     * node identity in `public_key` field. Service-endorsed certificate is
+     * recorded in "public:ccf.nodes.endorsed_certificates" table */
+    std::optional<crypto::Pem> cert = std::nullopt;
```
jumaffre, line 71: Moving a previously required field to optional here. This means that it can be parsed from the historical ledger, but isn't serialised in the ledger for new node entries that don't use this field.

src/node/node_state.h
```diff
@@ -369,42 +375,37 @@ namespace ccf
...
+          setup_consensus(ServiceStatus::OPENING, false, endorsed_node_cert);
           setup_progress_tracker();
           setup_history();
 
           // Become the primary and force replication
           consensus->force_become_primary();
 
-          // Open member frontend for members to configure and open the
-          // network
-          open_frontend(ActorsType::members);
```
jumaffre, line 385: This is now done later, from the new hook.

src/node/node_signature_verify.h
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ccf/tx.h"
+#include "crypto/verifier.h"
+#include "node_signature_verify.h"
+
+namespace ccf
+{
+  static bool verify_node_signature(
```
jumaffre, line 11: New common place to verify node signatures, with backwards compatibility with 1.x ledger baked it.

doc/operations/start_network.rst
```diff
@@ -103,33 +103,33 @@ The following diagram summarises the steps that operators and members should fol
             Node 0-->>-Node 1: "Pending" state
         end
 
-        Operators->>+Node 1: Poll /node/state for "PartOfNetwork"
+        Operators->>+Node 1: Poll /node/state for "PartOfNetwork" (using self-signed certificate as CA)
```
achamayou, line 106: As identity, not CA?

src/js/wrap.cpp
```diff
@@ -489,6 +489,53 @@ namespace js
...
+    auto global_obj = JS_GetGlobalObject(ctx);
+    auto ccf = JS_GetPropertyStr(ctx, global_obj, "ccf");
+    auto node_ = JS_GetPropertyStr(ctx, ccf, "node");
+
+    auto node =
+      static_cast<ccf::AbstractNodeState*>(JS_GetOpaque(node_, node_class_id));
+
+    if (node == nullptr)
+    {
+      return JS_ThrowInternalError(ctx, "Node state is not set");
```
achamayou, line 519: Leaks what's allocated on lines 510, 522, 512.

src/js/wrap.cpp
```diff
@@ -489,6 +489,53 @@ namespace js
...
+    }
+
+    JS_FreeValue(ctx, node_);
+    JS_FreeValue(ctx, ccf);
+    JS_FreeValue(ctx, global_obj);
+
+    auto csr_cstr = JS_ToCString(ctx, argv[0]);
```
achamayou, line 526: Never freed.

doc/operations/start_network.rst
jumaffre: Hmm, `curl` only has the `--cacert` option, Python `requests` calls it the `REQUESTS_CA_BUNDLE`, hence the name here.

achamayou: Ok

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Added support for listening on multiple interfaces for incoming client RPCs, with individual session caps (#2628).
+- Service-endorsed node certificates are now recorded in a new `public:ccf.gov.nodes.endorsed_certificates` table, while existing `cert` field in the `public:ccf.gov.nodes.info` table is now deprecated (#2844).
+- Joining nodes now present service-endorsed certificate in client TLS sessions _after_ they have observed their own addition to the store, rathen than as soon as they have joined the service. Operators should monitor the initial progress of a new node using its self-signed certificate as TLS session certificate authority (#2844).
+- Updated `actions.js` constitution fragment (#2844).
```
achamayou, line 15: It's worth adding a note that picking up this change is mandatory for the previous two changes to take effect correctly.

src/js/wrap.cpp
```diff
@@ -489,6 +489,52 @@ namespace js
...
+
+    auto node =
+      static_cast<ccf::AbstractNodeState*>(JS_GetOpaque(node_, node_class_id));
+
+    if (node == nullptr)
+    {
+      return JS_ThrowInternalError(ctx, "Node state is not set");
+    }
+
+    auto csr_cstr = JS_ToCString(ctx, argv[0]);
```
achamayou, line 524: This needs to be wrapped in an auto_free, otherwise we'll leak the cstring when we throw on 527.

src/node/config.h
```diff
@@ -15,8 +15,13 @@ namespace ccf
     size_t recovery_threshold = 0;
 
     ConsensusType consensus = ConsensusType::CFT;
+
     std::optional<ReconfigurationType> reconfiguration_type = std::nullopt;
 
+    // If true, the service endorses the certificate of new trusted nodes, and
```
achamayou, line 21: I'm confused as to why this is optional and not just boolean.

achamayou (approved):

src/node/config.h
jumaffre: I started a discussion on this just now: https://github.com/microsoft/CCF/discussions/2916 I agree that having the field as non-optional is better but I think there are implications for ledger serialisation.

src/js/wrap.cpp
jumaffre: Yep. The pattern seems to be consistent across this file and I'm planning to address it in the separate PR (see #2910).

Issue #2847, opened 2021-07-26
Failure in `rotation_test_cft`


eddyashton:
The `rotation_test_cft` has started failing in the CI with a very odd error:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=29814&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=d37f6062-24ee-5eff-15b2-0eb74651c675&l=24705

```
67:   File "/mnt/build/2/s/tests/infra/checker.py", line 46, in check_can_progress
67:     c.wait_for_commit(r, timeout=timeout)
67:   File "/mnt/build/2/s/python/ccf/clients.py", line 672, in wait_for_commit
67:     raise ValueError("Response seqno and view should not be None")
67: ValueError: Response seqno and view should not be None
60/68 Test #67: rotation_test_cft .................***Failed   16.69 sec
```

```
67: 09:53:26.409 | INFO     | infra.network:wait_for_new_primary:891 - Waiting up to 16.0s for a new primary different from 0 (ccd4d5694d9a48b86ec88b8fec8c5dc5a146ce459220fe58713321caa264b8f3) to be elected...
67: 09:53:29.349 | INFO     | infra.network:wait_for_new_primary:918 - [1||] GET /node/network
67: 09:53:29.349 | INFO     | infra.network:wait_for_new_primary:918 - 200 @2.21 {"current_view":3,"primary_id":"af2b20a052f5240fc545d9259973c5ad7822a262d291fd7b65581701e5512812","service_certificate":"-----BEGIN CERTIFICATE-----\nMIIBtjCCAT2gAwIBAgIQZHhrgRuppbrh0mswKDkQdjAKBggqhkjOPQQDAzAWMRQw\nEgYDVQQDDAtDQ0YgTmV0d29yazAeFw0yMTAzMTEw + 559 chars
67: 09:53:29.349 | INFO     | infra.network:wait_for_new_primary:919 - New primary is 1 (af2b20a052f5240fc545d9259973c5ad7822a262d291fd7b65581701e5512812) in term 3
67: 09:53:29.529 | INFO     | infra.checker:check_can_progress:44 - [2|user0|] POST /app/log/private {'id': 42, 'msg': 'Hello world'}
67: 09:53:29.529 | INFO     | infra.checker:check_can_progress:44 - 404 {"error":{"code":"ResourceNotFound","message":"Unknown path: /log/127.211.183.156:45423."}}
```

[...]

PR #2849, opened 2021-07-27
Return compliant URL in `Location` header of `TEMPORARY_REDIRECT` responses

eddyashton:
Fix for #2847.

I misread the spec for the `Location` header when first implementing this `TEMPORARY_REDIRECT` response, and as a result it was junk. We returned `Location: 127.4.5.6:1234`, and the client treats this as an alternative relative path, so if it's a redirection for `/app/log/private` it produces `/app/log/127.4.5.6:1234`. We now correctly return a complete URL to the redirected resource, in the same way we do for `PERMANENT_REDIRECT`s in the `node_frontend`, including the same assumption that `pubhost:pubport` is useable by the client.

I'm surprised this wasn't seen before. I think we only hit this path when a forwarded request reaches a node who is no longer primary, and the recent `check_can_progress` changes made that more common. I can't think of a way to trigger that consistently - we'd need to partition a node from the new primary, but keep them in touch with the old primary, and send a request in the window _after_ the old primary knows it is not primary, but before this follower has timed out and triggered an election? Tricky.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-*cough*
+*cough!*
--- a/src/node/rpc/frontend.h
+++ b/src/node/rpc/frontend.h
@@ -135,9 +135,13 @@ namespace ccf
 
           if (info)
           {
-            ctx->set_response_header(
-              http::headers::LOCATION,
-              fmt::format("{}:{}", info->pubhost, info->pubport));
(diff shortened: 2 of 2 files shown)


eddyashton:
Merging despite the Daily failure, since it existed before and this should improve the pass rate.

PR #2850, opened 2021-07-28
Don't overwrite CTest results with `partitions_cft` test results

eddyashton:
I noticed that in our Azure Pipelines we're no longer getting accurate test analytics - the many recent `_bft` failures are not recorded:

https://dev.azure.com/MSRC-CCF/CCF/_test/analytics?definitionId=7&contextType=build

I believe this has been broken since we added the `partitions_cft` test in a separate step, because it overwrites the `Test.xml` file from the original `ctest` run. ~~This fix writes the second `ctest`'s results to a different file name, and publishes both (hopefully these are combined, otherwise we will have to find an alternative approach).~~

That option doesn't do what I thought, and I can't find a way to write the results xml file to a different name, or append to an existing one. Open to suggestions. For now, I've just removed the xml generation from the `partitions_test` step, so we invert the problem (get analytics for every _other_ test).


eddyashton:
Merging despite the CI failures - this does not introduce any new failures, and makes the existing failures easier to track.

achamayou:
As discussed, let's try to move sudo down so that we get a single consolidated command if we can.

PR #2851, opened 2021-07-29
Remove .js file from glob pattern when uploading JS apps

eddyashton:
Fixes `forum_app`, which has broken because of some dependency upgrade now using a `.js_commonjs-proxy` file (previously built but not used, so safely never uploaded). There's a risk that this uploads junk in future, if apps include more non-source things under their build folder, but seems a necessary tradeoff.


eddyashton:
Merging despite CI failure - this is in unrelated `recovery_test_bft`, which is already flaky. Will investigate it separately.

PR #2852, opened 2021-07-29
Raft committable suffix: efficient backtracking to find safe matching index

### LINKED ISSUE #2803: Incorrect suffix retained after election

eddyashton:
Resolves #2803.

Follow up to #2835, using a more accurate and crucially persistence-retaining method for finding the last index where 2 divergent logs disagree.

As described in #2803; sending an early index is potentially unsafe as it may cause a log to roll back too far and lose state it is responsible for committing (though later than its local knowledge of `commit_idx`). Discarding a matching prefix is non-trivial, since we do not know the term and index of an item inside an AppendEntries (which is what we wish to use for equality checking) until after we have deserialised it, and our deserialisation and application are closely coupled.

This PR includes 2 new unit tests (in `committable_suffix.cpp`) which demonstrate the risk here, and failed with the previous implementation. The first involves 5 nodes, where 2 are responsible for persisting a dead primary's committed suffix, but in resolving disagreement about their local attempts at extending the log, one was rolling back too far. The other is a 3-node network where 2 nodes build long suffixes of alternating terms, without advancing commit, and confirms both that the eventual stable primary is able to bring the other back in-sync to advance commit, and that the back-tracking logic to discover their matching index is efficient (bounded by the number of terms in the worst case, rather than the number of log entries).

This includes a refactoring of `raft/test/main.cpp`, to put these test cases in a separate file and move some of the shared test setup to `test_common.h`.

The interesting behaviour changes are in `raft.h` and `aft/impl/state.h`, centered on `find_highest_possible_match`. This is used both when constructing a NACK, and updating `matching_idx` to respond to that NACK, to backtrack through the best-possible estimates of matching indices until eventually the backup receives an AppendEntries with no conflicts, following immediately from their last matching index.


src/consensus/aft/raft.h
```diff
@@ -2424,33 +2457,47 @@ namespace aft
       {
         // Stale response, discard if success.
         // Otherwise reset sent_idx and try again.
-        LOG_DEBUG_FMT(
-          "Recv append entries response to {} from {}: stale term ({} != {})",
-          state->my_node_id,
-          from,
-          r.term,
-          state->current_view);
+        // NB: In NACKs this may contain the TxID of an estimated matching index
```
achamayou, line 2460:
```suggestion
        // NB: In NACKs this response may contain the TxID of an estimated matching index
```

src/consensus/aft/raft.h
```diff
@@ -2424,33 +2457,47 @@ namespace aft
...
+        // in the log, rather than the current term, so it is correct for it to
+        // be stale in this case.
```
achamayou, line 2462:
```suggestion
        // be older in this case.
```

src/consensus/aft/raft.h
```diff
@@ -2424,33 +2457,47 @@ namespace aft
...
+            state->my_node_id,
+            from,
+            r.term,
+            state->current_view);
           return;
         }
       }
       else if (r.last_log_idx < node->second.match_idx)
       {
         // Stale response, discard if success.
```
achamayou, line 2476:
```suggestion
        // Response about past indices, discard if success.
```

src/consensus/aft/test/committable_suffix.cpp
```diff
@@ -0,0 +1,860 @@
...
+    B: Rolls back to 1.1
+
+    B: [1.1]
+    C: [1.1] 1.2
+    D: [1.1]
+    E: [1.1]
+
+  - At this point a committed index (1.2) is no longer present on a majority of
+    nodes. While C is unlikely to advertise it (fancy election rules mean
+    they're waiting for commit at 1.2) and will continue to share it, its
```
achamayou, line 172: Not sure I understand that bit. If it was practically committed, it must have been committable (ie. signed) and so if C participates there's a good chance it will win. This is all moot because the election could resolve without C of course, as rightly pointed out in the next line.

src/consensus/aft/test/committable_suffix.cpp
```diff
@@ -0,0 +1,860 @@
...
+        else
+        {
+          ++it;
+        }
+      }
+    }
+
+    DOCTEST_REQUIRE(6 == dispatch_all(nodes, node_idA, channelsA->messages));
+
+    // NB: AppendEntriesResponses are not dispatched yet. So 1.4 is technically
```
achamayou, line 299: I think we ought to replace "technically committed" with "present on f+1 nodes", same with "True commit index", "Highest index present on f+1 nodes" etc, to distinguish that from commit index, which is what's advertised by the primary and eventually acknowledged by replicas.

src/consensus/aft/test/committable_suffix.cpp
```diff
@@ -0,0 +1,860 @@
...
+  nodes[node_id##N] = &r##N; \
+  auto channels##N = channel_stub_proxy(r##N);
+
+/**
+  Summary of the behaviour this test aims to explore (for implementation reasons
+  the indices in the actual test don't match those in this comment, but the
+  logical relationships are the same):
+
+  - Produce an initial state where A is primary in term 1 of a 5-node network,
+    has mixed success replicating its entries. Local commit index is marked by
```
achamayou, line 131:
```suggestion
    has mixed success replicating its entries. Local view of the commit index is marked by
```

We've used (and continue to use I believe) "local commit index" with very different semantics.

achamayou (approved):

src/consensus/aft/test/committable_suffix.cpp
eddyashton: I'll reword this. I was trying to say that the user-visible commit may not show the loss of persistence (ie - if C survives, then it _will_ share `1.2`, and that will be committed again, and the commit that a user could see won't move backwards). I think there may be space for an interesting distinction between the actual commit, and the user-visible commit, since the fancy election state lets a node know while have an inaccurate/lagging commit index? But as you say, that's all moot since there's still a fundamental persistence hole - _if_ C loses an election here, then we've lost persistence.

src/consensus/aft/test/committable_suffix.cpp
eddyashton:
```suggestion
    has mixed success replicating its entries. Each node's commit index is marked by
```

src/consensus/aft/test/committable_suffix.cpp
achamayou: Ah I see, advertise it to the user, I'd read that as advertise it in the election process. Yes, there is a distinction not just between "persisted by f+1 nodes" and "commit on leader", but also "commit as advertised by node to user".

src/consensus/aft/test/committable_suffix.cpp
```diff
@@ -0,0 +1,865 @@
...
+    C: [1.1] 1.2
+    D: [1.1]
+    E: [1.1]
+
+  - Node C doesn't hear from B for long enough, and becomes primary. Its
+    AppendEntries are regularly lost. Eventually a heartbeat ("I'm at 1.2")
+    reaches Node B, which responds with a NACK ("I don't accept that tail").
+    Node C responds to that NACK with a better AppendEntries, which _should_
+    bring Node B back to the committed level, and forward to wherever C is.
+    However, under the current implementation, this AppendEntries may
```
jumaffre, line 160: "under the current implementation" will be outdated once this is merged. Should we briefly mention what the first implementation was instead?

jumaffre (approved):

src/consensus/aft/test/committable_suffix.cpp
eddyashton: Good point, I've updated this comment.

```
However, if the AppendEntries returns an early-estimate of matching index
(as a previous implementation did), this AppendEntries may actually cause B
to roll back further than is safe, losing the committed state.
```

PR #2859, opened 2021-08-03
Fix `check-format.sh` script: Pass -Werror

eddyashton:
Without `-Werror`, this raises warnings but doesn't reformat. I think we want no warnings, to format all.


achamayou:
We definitely don't want warnings, this is an oversight. I hadn't realised this distinction even existed, it must be new in 10.

achamayou (approved):

jumaffre (approved):

PR #2865, opened 2021-08-04
Disable recovery_test_bft

achamayou:
Test is not stable, and is liable to change considerably when the identity work is complete. Will be revisited then.


eddyashton:
I believe this test _is_ stable after #2854, and the only failures are on branches that don't have that fix?

achamayou:
@eddyashton ok, let's get #2854 in shortly and find out.

PR #2871, opened 2021-08-04
Perf test in 1ES pool

achamayou:
Change the perf test to run a local two-node setup that fits in hosted 1ES pools without custom setup. While this is slightly further away from a realistic setup than the previous perf pool, it is:

1. much easier to maintain
2. fully dockerised, and therefore always in sync with the rest of the build dependency-wise
3. still suitable for regression-testing purposes

TPCC BFT failed because of a bug fixed by #2873.

### DIFF
--- a/.nodes.sh
+++ b/.nodes.sh
@@ -1,7 +1,3 @@
 #!/bin/bash
 
-HN=$(hostname)
-
-if [ "$HN" == "ccf-perf-uk-0" ]; then
-    echo "-n ssh://10.1.0.7 -n ssh://10.1.0.4 -n ssh://10.1.0.5"
-fi
+echo "-n local://localhost -n local://localhost"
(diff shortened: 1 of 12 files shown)


.nodes.sh
```diff
@@ -1,7 +1,3 @@
 #!/bin/bash
 
-HN=$(hostname)
-
-if [ "$HN" == "ccf-perf-uk-0" ]; then
-    echo "-n ssh://10.1.0.7 -n ssh://10.1.0.4 -n ssh://10.1.0.5"
-fi
\ No newline at end of file
+echo "-n local://localhost -n local://localhost"
```
eddyashton, line 3: This script was necessary when we had custom logic for certain nodes, but I think now it would be simpler to just have these `-n` arguments directly in the yaml, and remove this file?

eddyashton (approved):

.nodes.sh
achamayou: Sure, done.

PR #2876, opened 2021-08-05
Unify genesis and public recovery transaction execution

jumaffre:
Part of #2554 

We used to perform a call to `/gov/create` for the very first transaction but created a custom `Tx` in `node_state.h` for the very first post-public-recovery transaction. This PR unifies the two schemes by using the RPC call for recovery as well.

The main advantage is to reduce the number of places in the code base where we record nodes in the store. 

Note: I've also moved the create endpoint from `/gov/` to `/node/` since the `/gov/` frontend no longer needs to be opened before the post-recovery transaction is executed.


src/node/node_state.h
```diff
@@ -384,7 +384,7 @@ namespace ccf
           // network
           open_frontend(ActorsType::members);
 
-          if (!create_and_send_request())
+          if (!create_and_send_request(false))
```
achamayou, line 387:
```suggestion
          if (!create_and_send_request(false) /* Not a recovery */)
```

src/node/node_state.h
```diff
@@ -1049,18 +1018,10 @@ namespace ccf
...
 
-      // Sets itself as trusted
-      g.trust_node(self, network.ledger_secrets->get_latest(tx).first);
-
-#ifdef GET_QUOTE
-      g.trust_node_code_id(node_code_id);
-#endif
-
-      if (tx.commit() != kv::CommitResult::SUCCESS)
+      if (!create_and_send_request(true))
```
achamayou, line 1021:
```suggestion
      if (!create_and_send_request(true /* Recovery */))
```

src/node/node_state.h
```diff
@@ -1641,33 +1602,41 @@ namespace ccf
       open_frontend(ccf::ActorsType::users, &network.identity->cert);
     }
 
-    std::vector<uint8_t> serialize_create_request(const QuoteInfo& quote_info)
+    std::vector<uint8_t> serialize_create_request(bool recovery = false)
     {
       CreateNetworkNodeToNode::In create_params;
 
-      for (const auto& m_info : config.genesis.members_info)
+      if (!recovery)
```
achamayou, line 1609: Although this is used/not used for recovery, what it does is really "define_consortium" or "create_consortium", and ought to be named as a flag that way imo. Then in recovery, "create_consortium = false" because we inherit the consortium from the ledger instead.

src/node/node_state.h
```diff
@@ -1750,10 +1719,10 @@ namespace ccf
       return parse_create_response(response.value());
     }
 
-    bool create_and_send_request()
+    bool create_and_send_request(bool recovery = false)
```
achamayou, line 1722:
```suggestion
    bool create_and_send_init_request(bool recovery = false)
```

Or "genesis" if we like the name, otherwise it sounds much more generic than it really is.

achamayou (approved):
Very nice simplification!

jumaffre:
The `Check Daily Build status` GH Actions job seems stuck, but the Daily is green so merging this regardless
image

PR #2879, opened 2021-08-05
Fix reported `TxID` term during public recovery

jumaffre:
Noticed this term anomaly in the logs of backups in the `recovery_test_cft`:
image

This is because while deserialising public entries, the integrity of each ledger entry isn't verified, which means that the term (recorded in the AES-GCM IV) isn't set either. We now read and report the term even if we don't verify the integrity of the entry. 

Note that just like for the version, we rely on the following signature verification to verify the integrity of the ledger entry. If the following signature tells us that entry was corrupted, we may have _transiently_ reported the wrong `TxID`, but the store will be quickly rolled back at the previous verified signature.


eddyashton (approved):

achamayou:
@jumaffre but we would never report these potentially bogus TxID terms as committed, right?

PR #2885, opened 2021-08-09
Remove HTTP signature for self RPCs

jumaffre:
Nodes' self RPCs (genesis and JWT refresh) were signed, but with the wrong `keyId`. These signatures weren't actually verified so this PR removes the HTTP signature on those. Note that as discussed with @achamayou and @eddyashton, there is no apparent benefit in signing and storing these signatures currently (the primary node already emits signature transactions for blame), so we will hold off doing this for now.

Also:
- Move the existing `/gov/jwt_keys/refresh` endpoint to `/node`
- Guard the `/node/create` genesis endpoint that was made more permissive in #2876 with the node internal state, so that the endpoint can't be called once the service is open


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - CCF is now built with Clang 10. It is recommended that C++ applications upgrade to Clang 10 as well.
+- Internal `/gov/jwt_keys/refresh` endpoint has been moved to `/node/jwt_keys/refresh`.
```
jumaffre, line 13: This endpoint isn't listed in the OpenAPI spec yet it shows up in the metrics for a given frontend so I've listed it here just in case.

src/node/rpc/node_frontend.h
```diff
@@ -1078,6 +1096,108 @@ namespace ccf
...
+        try
+        {
+          parsed = body.get<SetJwtPublicSigningKeys>();
+        }
+        catch (const JsonParseError& e)
+        {
+          return make_error(
+            HTTP_STATUS_INTERNAL_SERVER_ERROR,
+            ccf::errors::InternalError,
+            "Unable to parse body.");
```
achamayou, line 1145: Including e.what() seems reasonable in this case.

achamayou (approved):

src/node/rpc/node_frontend.h
jumaffre: This is actually already caught by the generic frontend logic so removing this check here.

jumaffre:
Force merging this PR as the Daily failure is unrelated (see #2896).

PR #2886, opened 2021-08-10
Serialise `service_status` field in Join response

jumaffre:
The `service_status` field wasn't serialised in the `/node/join` response. Hopefully this doesn't break any BFT test but it might as I think all nodes join as `Follower` rather than `Learner` in BFT. 

Also updated the serialisation format for the `/node/join` response, using `std::optional` for the `NetworkInfo` struct which is only set when the node has been trusted.

### DIFF
--- a/src/node/rpc/node_call_types.h
+++ b/src/node/rpc/node_call_types.h
@@ -96,7 +96,6 @@ namespace ccf
       NodeStatus node_status;
       NodeId node_id; // Only used in BFT
 
-      // Only if the caller node is trusted
       struct NetworkInfo
       {
         bool public_only = false;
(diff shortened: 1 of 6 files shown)


achamayou (approved):

src/node/rpc/node_call_types.h
```diff
@@ -105,25 +104,31 @@ namespace ccf
...
+          ServiceStatus service_status) :
+          public_only(public_only),
+          last_recovered_signed_idx(last_recovered_signed_idx),
+          consensus_type(consensus_type),
+          ledger_secrets(ledger_secrets),
+          identity(identity),
+          service_status(service_status)
+        {}
+
+        bool operator==(const NetworkInfo& other) const = default;
```
achamayou, line 126: This won't work with C++17

jumaffre: Is the plan not to break compatibility with C++17 just yet?

jumaffre: As discussed, we won't break C++!7 compatibility just yet, until we know that existing CCF apps can upgrade to C++20 safely.

PR #2779, opened 2021-07-08
Cache repository data in LTS compatibility test

jumaffre:
~Add the [`GITHUB_COMPATIBILITY_TOKEN`](https://github.com/microsoft/CCF/blob/main/tests/infra/github.py#L20) envvar to the test part of our CI pipelines so that we avoid GitHub's rate limit errors in the LTS release test (e.g. https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=28520&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=cf4e34b3-24ee-5d97-0df9-ba9302d9fe90&l=20875).~

~I've already added the envvar to both our CI and Daily pipelines. Note that any PAT (with no specific access) works.~

For some reason, the solution above doesn't work as the tokens aren't passed properly by Azure DevOps to the LTS test (???).

Instead, we cache the results of the API calls to GitHub to avoid the low rate limit for unauthenticated requests.

PR #2888, opened 2021-08-10
Remove use of GithHub API for `lts_compatibility` test

jumaffre:
As we somehow cannot use GitHub token to prevent the GitHub API throttle (see #2779), this PR removes the use of the GitHub API and replaces it with plain old Git.

### DIFF
--- a/tests/infra/github.py
+++ b/tests/infra/github.py
@@ -92,10 +92,10 @@ class Repository:
     def __init__(self):
         self.g = git.cmd.Git()
         self.tags = [
-            tag.split("/")[-1]
+            tag.split("tags/")[-1]
             for tag in self.g.ls_remote(REMOTE_URL).split("\n")
-            if f"tags/{TAG_RELEASE_PREFIX}1" in tag
-        ]  # TODO: Remove 1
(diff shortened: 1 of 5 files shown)


tests/infra/github.py
```diff
@@ -74,27 +75,38 @@ def get_major_version_from_branch_name(branch_name):
     )
 
 
+def get_debian_package_url_from_tag_name(tag_name):
+    return f'{REMOTE_URL}/releases/download/{tag_name}/{tag_name.replace("-", "_")}{DEBIAN_PACKAGE_EXTENSION}'
+
+
+def has_release_for_tag_name(tag):
+    return requests.get(f"{REMOTE_URL}/releases/tag/{tag}").status_code == 200
```
achamayou, line 83: We could .head() on the release binary directly to check if it exists or not. It's nice if the release is there, but ultimately what we want is the .deb.

achamayou (approved):

tests/infra/github.py
jumaffre: Good point. This simplifies things slightly too!

PR #2889, opened 2021-08-11
Remove VM creation scripts

achamayou:
DC-series VM are now well supported by the standard VM creation interface, and oe-engine has fallen out of support, so this is removing references to the latter from documentation.


doc/build_apps/install_bin.rst
```diff
@@ -61,7 +61,7 @@ The CCF install notably contains:
 
 - The ``cchost`` binary required to spin up a CCF application
 - The ``cmake`` files required to build CCF applications
-- Azure and ``ansible`` scripts required to :doc:`create_vm` and :doc:`build_setup` (under ``getting_started/``)
+- The ``ansible`` playbooks required :doc:`build_setup` (under ``getting_started/``)
```
jumaffre, line 64:
```suggestion
- The ``ansible`` playbooks required for :doc:`build_setup` (under ``getting_started/``)
```
(since `:doc:build_setup` renders as "CCF Development Setup")

doc/build_apps/create_vm.rst
```diff
@@ -1,33 +1,11 @@
 Create Azure SGX VM
 ===================
 
-.. note:: These steps require an `Azure subscription <https://docs.microsoft.com/en-us/azure/billing/billing-create-subscription#create-a-subscription-in-the-azure-portal>`_.
-
-Creation
---------
-
-The install directory contains (under ``getting_started/``) the Azure and Ansible scripts required to create and setup an SGX-enabled development VM in Azure, with all necessary dependencies to build CCF itself and CCF applications.
+:term:`Azure Confidential Compute` (ACC) offers SGX-enabled `DC`_-series VMs, which can be deployed like other Azure SKUs.
```
jumaffre, line 4: We should at least mention the available regions for that SKU here.

jumaffre (approved):

doc/build_apps/create_vm.rst
achamayou: I don't really want to, because there are three SKUs (DC, DCv2 and DCv3), with different region availability, and they change quite often. If we find a good page, we'll link to it, but I'm reluctant to try to maintain an up to date list.

PR #2892, opened 2021-08-12
Switch BFT to hashed public key node ids

### LINKED ISSUE #1852: AFT: Support reconfiguration

wintersteiger:
Fixes #1852

### DIFF
--- a/src/node/rpc/node_call_types.h
+++ b/src/node/rpc/node_call_types.h
@@ -94,7 +94,7 @@ namespace ccf
     struct Out
     {
       NodeStatus node_status;
-      std::optional<NodeId> node_id; // Only used in BFT
+      std::optional<NodeId> node_id;
 
       struct NetworkInfo
       {
--- a/src/node/rpc/node_call_types.h
+++ b/src/node/rpc/node_call_types.h
@@ -94,7 +94,7 @@ namespace ccf
     struct Out
     {
       NodeStatus node_status;
-      NodeId node_id; // Only used in BFT
+      std::optional<NodeId> node_id; // Only used in BFT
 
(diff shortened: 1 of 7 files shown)


src/node/rpc/node_call_types.h
```diff
@@ -94,7 +94,7 @@ namespace ccf
     struct Out
     {
       NodeStatus node_status;
-      NodeId node_id; // Only used in BFT
+      std::optional<NodeId> node_id; // Only used in BFT
```
jumaffre, line 97:
```suggestion
      std::optional<NodeId> node_id;
```

src/node/rpc/node_frontend.h
```diff
@@ -611,6 +598,8 @@ namespace ccf
             if (primary_id.has_value() && !consensus->view_change_in_progress())
             {
               out.primary_id = primary_id.value();
+              LOG_DEBUG_FMT(
```
jumaffre, line 601: Delete?

wintersteiger:
> I believe the `values` tables can be removed altogether, without causing any issues of backwards compatibility either.

That's correct, but I might want to use it for the Byzantine identity work - I'll either move it to a new name or remove it altogether if I end up not needing it.

PR #2894, opened 2021-08-12
Report HTTP signature parsing error to client 

### LINKED ISSUE #2890: Expose HttpSignatureVerifier::parse() failure details in HTTP response, and use them in *SignatureAuthnIdentity
The signature parser will currently log as TRACE failure details of signature verification, such as whether particular headers are missing etc, but won't make them available to the caller.

We should expose them, and return them in the HTTP response, to help guide users (or members) trying to sign payloads but hitting formatting issues.

jumaffre:
Resolves #2890

### DIFF
--- a/src/http/authentication/sig_auth.h
+++ b/src/http/authentication/sig_auth.h
@@ -10,7 +10,7 @@ namespace ccf
 {
   namespace
   {
-    static std::optional<SignedReq> parse_signed_request(
+    static SignedReq parse_signed_request(
       const std::shared_ptr<enclave::RpcContext>& ctx)
     {
       return http::HttpSignatureVerifier::parse(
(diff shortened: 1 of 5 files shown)


src/http/authentication/sig_auth.h
```diff
@@ -72,37 +72,41 @@ namespace ccf
...
+      SignedReq signed_request;
+
+      try
+      {
+        signed_request = parse_signed_request(ctx);
+      }
+      catch (const std::exception& e)
+      {
+        error_reason = e.what();
+        return nullptr;
```
wintersteiger, line 84: Perhaps worth keeping the `std::optional` instead of `nullptr`s?

wintersteiger (approved):

src/http/authentication/sig_auth.h
jumaffre: As discussed, the function already returned a `nullptr` on error (the default view for the GitHub diff is confusing) so I'll leave this as it is.

jumaffre:
There seems to be a regression with `tpcc_sgx_cft`. Investigating.

jumaffre:
> There seems to be a regression with `tpcc_sgx_cft`. Investigating.

The regression was due to the fact what we threw an exception on every unsigned request for HTTP endpoints that require a signature. This isn't the case anymore as we return a `std::nullopt` for unsigned requests (i.e. no `Authorization` header).

src/http/http_sig.h
```diff
@@ -308,34 +307,31 @@ namespace http
         }
         else
         {
-          LOG_FAIL_FMT("Authorization parameter {} does not contain \"=\"", p);
-          return std::nullopt;
+          throw std::logic_error(fmt::format(
+            "authorization parameter '{}' does not contain \"=\"", p));
```
eddyashton, line 311:
```suggestion
            "Authorization parameter '{}' does not contain \"=\"", p));
```

eddyashton (approved):

PR #2897, opened 2021-08-13
Fix `test_node_replacement` failure in `full_test_suite`

### LINKED ISSUE #2896: `test_node_replacement` CI failures in `full_test_suite`
The bug has occurred quite a few times in the last few days, e.g. https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=30819&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=cf4e34b3-24ee-5d97-0df9-ba9302d9fe90&l=14021

jumaffre:
Fix #2896 

The `test_node_replacement`, when chained after `test_kill_primary` failed consistently as the consensus couldn't make progress. This is because the `at_least_n_nodes(3)` requirement wasn't precise enough. In this case, the network was already degraded when the test started (4 trusted nodes but 1 is stopped) so the backup suspension at the end of the test meant that they were only 2 nodes left for consensus, which is less than the quorum of 3.

Also added a `--dry-run` option to the test suite, and tweaked the local node id logic in the infra so that node ids aren't re-used when a node is retired.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Allo Allo
+Coin coin
--- a/tests/e2e_suite.py
+++ b/tests/e2e_suite.py
@@ -39,6 +39,9 @@ def mem_stats(network):
 def run(args):
     chosen_suite = []
 
+    if args.dry_run:
+        LOG.warning("--dry-run set. Test execution will be skipped")
+
     if not args.test_suite:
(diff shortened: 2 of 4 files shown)


jumaffre:
This is the fix for current failure on the Daily CI.

PR #2899, opened 2021-08-13
Add 2-tx reconfiguration RPCs/ORCs

wintersteiger:
Still a draft that needs cleanup, but the logic is there.


doc/audit/builtin_maps.rst
```diff
@@ -385,4 +385,9 @@ Evidence inserted in the ledger by a primary producing a snapshot to establish p
...
 Used to persist submitted shares during a recovery.
 
-While the contents themselves are encrypted, the table is public so as to be accessible by nodes bootstrapping a recovery service.
\ No newline at end of file
+While the contents themselves are encrypted, the table is public so as to be accessible by nodes bootstrapping a recovery service.
+
+``resharings``
+~~~~~~~~~~~~~~
+
+Internal information for identity resharing during 2-transaction reconfiguration.
```
achamayou, line 393: I think we would still like a basic description of the Key and Value if possible, for anyone who's working on the code base.

achamayou: As discussed, this is due further change very shortly, so not necessary right now.

achamayou (approved):

PR #2902, opened 2021-08-18
Default SAN to public RPC address

jumaffre:
If a `--san` isn't specified when `cchost` is started, the host in `--rpc-address` is automatically added as a SAN in the node certificate, which is useful to start a node with a minimal number of parameters. 

However, the SAN is mostly dealt with by clients which connect to a node via its _public_ RPC address. For example, a node could be running inside a container, binding to `--rpc-address X` but with `--public-rpc-address Y` as the external address accessible to clients on the host. 

This PR is just a very minor fix to make sure we default to the public RPC address as a SAN if no other `--san` are specified.


eddyashton (approved):

achamayou:
@jumaffre good catch! Can we add a test to capture this?

achamayou (approved):

tests/infra/crypto.py
```diff
@@ -258,6 +258,10 @@ def compute_cert_der_hash_hex_from_pem(pem: str):
     return cert.fingerprint(hashes.SHA256()).hex()
 
 
+def get_san_from_pem_cert(pem: str):
+    cert = load_pem_x509_certificate(pem.encode(), default_backend())
+    return cert.extensions.get_extension_for_class(x509.SubjectAlternativeName).value
+
```
jumaffre, line 264:
```suggestion


```

src/node/node_state.h
```diff
@@ -1548,8 +1548,8 @@ namespace ccf
         for (const auto& interface : config.node_info_network.rpc_interfaces)
         {
           sans.push_back(
-            {interface.rpc_address.hostname,
-             is_ip(interface.rpc_address.hostname)});
+            {interface.public_rpc_address.hostname,
+             is_ip(interface.public_rpc_address.hostname)});
```
achamayou, line 1552: Come to think of it, we should mention this in the CHANGELOG, it is user-visible.

Issue #1886, opened 2020-11-16
Update HTTP signature


jumaffre:
As we now support the `hs2019` signature scheme (see https://tools.ietf.org/html/draft-ietf-httpbis-message-signatures-00) where the signature hashing algorithm is derived from the key specified by `keyId`, we should remove the `ecdsa-sha256` scheme, which is deprecated according to the RFC draft.

This involves removing or patching the Python http signature package which only supports the `ecdsa-sha256` scheme.

eddyashton:
I think there are several more changes between the current `draft-ietf-httpbis` and whatever version of the `draft-cavage` document we initially implemented. An initial (incomplete) list of changes we should make for current compatibility:

- [ ] Use `Signature:` and `Signature-Input:` headers rather than `Authorization:`
- [ ] Normalize all headers as described in the spec

achamayou:
The RFC has been substantially updated with a new format:

```
   Signature-Input: sig1=("@request-target" "host" "date"
       "cache-control" "x-empty-header" "x-example"); keyid="test-key-a";
       alg="rsa-pss-sha512"; created=1402170695; expires=1402170995
   Signature: sig1=:K2qGT5srn2OGbOIDzQ6kYT+ruaycnDAAUpKv+ePFfD0RAxn/1BUe\
       Zx/Kdrq32DrfakQ6bPsvB9aqZqognNT6be4olHROIkeV879RrsrObury8L9SCEibe\
       oHyqU/yCjphSmEdd7WD+zrchK57quskKwRefy2iEC5S2uAH0EPyOZKWlvbKmKu5q4\
       CaB8X/I5/+HLZLGvDiezqi6/7p2Gngf5hwZ0lSdy39vyNMaaAT0tKo6nuVw0S1MVg\
       1Q7MpWYZs0soHjttq0uLIA3DIbQfLiIvK6/l0BdWTU7+2uQj7lBkQAsFZHoA96ZZg\
       FquQrXRlmYOh+Hx5D9fJkXcXe5tmAg==:
```

https://tools.ietf.org/html/draft-ietf-httpbis-message-signatures-03

achamayou:
Since clients, including the very popular https://github.com/joyent/node-http-signature seem not to track the drafts because they change too quickly, we will do the same and upgrade only once the final RFC is published.

PR #2912, opened 2021-08-24
requests -> httpx

achamayou:
This PR is removing requests and replacing it with httpx, to enable async use and HTTP/2 support in time.

TODO:

- [x] fix connections test
- [x] Update discussion about request signing format (#1886)


python/ccf/clients.py
```diff
@@ -8,26 +8,59 @@
...
-import requests
+import httpx
 from loguru import logger as LOG  # type: ignore
-from requests_http_signature import HTTPSignatureAuth  # type: ignore
 
 import ccf.commit
 from ccf.log_capture import flush_info
 
 
+class HttpSig(httpx.Auth):
```
achamayou, line 33: This is all we used from requests_http_signature, but compatible with httpx.

src/apps/js_generic/js_generic.cpp
```diff
@@ -149,11 +149,11 @@ namespace ccfapp
       crypto::Pem cert;
       if (is_member)
       {
-        result = get_user_cert_v1(endpoint_ctx.tx, id, cert);
```
achamayou, line 152: Incidental fix, was the wrong way around.

src/apps/js_generic/js_generic.cpp
```diff
@@ -329,80 +329,83 @@ namespace ccfapp
...
-        {
-          endpoint_ctx.rpc_ctx->set_response_header(
-            http::headers::CONTENT_TYPE,
-            http::headervalues::contenttype::OCTET_STREAM);
-          response_body =
-            std::vector<uint8_t>(array_buffer, array_buffer + buf_size);
-        }
-        else
+
+        if (!JS_IsUndefined(response_body_js))
```
achamayou, line 333: Allow responses with empty body, to enable (valid) user-defined 204.

tests/memberclient.py
```diff
@@ -40,39 +40,38 @@ def test_missing_signature_header(network, args):
...
 
 
-signature_regex = 'signature="([^"]*)",'
+signature_regex = 'signature="([^"]*)"'
 
 
 def missing_signature(request):
-    original = request.headers["Authorization"]
-    request.headers["Authorization"] = re.sub(signature_regex, "", original)
+    original = request.headers["authorization"]
```
eddyashton, line 54: Does httpx not use case-insensitive headers?

eddyashton (approved):

tests/memberclient.py
achamayou: It does, but since we use lower case everywhere for signing, it seemed weird not to here. I don't mind changing it back if you prefer.

eddyashton: No that's fine, just wanted to check.

Issue #2893, opened 2021-08-12
Add proposal action to refresh node certificate


jumaffre:
Follow-up from #2554 

It should now be easy enough to add a new proposal action to refresh the certificate of a node.

This is particularly useful when a node certificate is about to expire, but also after a code update from `1.x` to `2.x` in which the endorsed certificate of ` 2.x` node wasn't recorded by the `1.x` service.

jumaffre:
This should include:
- [ ] 1. A new "manual" proposal for members to renew the certificate of a single node, specifying a start date and a validity duration (either a start date, end date pair, or a start date and validity duration)
- [ ] 2. A new scheduled job to renew all expired nodes' certificates automatically. This should be scheduled on the primary node at a specific interval (passed in to `cchost`). The _untrusted_ host time should be used to check and set the validity period. On node start-up, operators should also specify whether they want to opt into automatic certificate renewal. 

In both cases, a service-wide certificate maximum validity period should be recorded in the store (to be checked on 1., and used by 2.). There should also be configurable values for how long before expiry a certificate should be renewed (so that there's no time during which the node certificate is expired).

jumaffre:
As discussed with @lynshi, the current implementation (manual proposals to update node certificate validity period) should be sufficient for 2.x. as it is convenient enough for operators and provide better auditability than with a self-refresh internal job.

Still to do before closing this:
- [ ] Change certificate refresh from local to global hook, which isn't as straightforward as I first thought.

Issue #59, opened 2019-05-13
Quote freshness for join protocol


achamayou:
See discussion below

achamayou:
This is connected to #178. It isn't enough to do this for TLS, we must also include the service identity in quotes.

To do so, the quotes will need to be over the service cert and a TBS, "To be signed", which is the cert of the node, endorsed by the service, but unsigned.
This TBS will be emitted by the service, for the joining node, to ensure quote freshness. Quotes over stale TBSs will be rejected.

Additional investigation regarding the exact implementation of the TBS with mbedtls is necessary before this feature can be implemented.

achamayou:
```
Cj: joiner cert
Cs: service cert
Cn: node cert, where node is part of the service

JOINER                            SERVICE
self-signed Cj ------- TLS -------> Cs Cn
            <--- TBS(Cj) ----------
            - quote(Cs, TBS(Cj)) -> [entry written to KV as PENDING]
                                    [VOTE]
                                    [ entry updated as TRUSTED, and identity as Cs Cj]
[ DONE ]    <--- secrets ----------
```

achamayou:
An additional requirement, connected to #47, is that Cs should contain alpha, the relevant enclave initialisation parameters.

jumaffre:
Splitting this issue. https://github.com/microsoft/CCF/issues/552 now deals with certificate endorsement while this ticket is about quote freshness.

achamayou:
```
      // TODO(#important,#TR): The "alpha" parameters, including the unique
      // service identifier, should also be included in the quote.
```

jumaffre:
Reviving this ticket after work and discussion about #2554 with @achamayou and @wintersteiger.

The solution described in previous comment is now outdated: it is no longer the node certificate that is quoted but the node's public key, since the latter is stable during the lifetime of the node but the certificate may be refreshed. 

[...]

PR #2924, opened 2021-08-27
Renew node certificate proposal

jumaffre:
Part of #2893 

~WIP~ Now ready for review

This PR adds a new proposal action `set_node_certificate_validity` (and associated `set_all_nodes_certificate_validity`) to renew the certificate of a node. 

The changes most concern:
- The OpenSSL functions to manipulate time (see `x509_time.h`)
- The certificate crypto API: caller can now specify a `valid_from` and `valid_to` string when creating a certificate
- New CLI arguments to `cchost` (passed to the enclave, with one recorded in the service configuration table)
- New JavaScript proposal actions 
- End-to-end testing. In particular, the infra remembers the validity period for a certificate so that the expected certificate validity period can be verified in the `full_test_suite`. 

TODO:
- [x] Changelog
- [x] Docs
- [x] Remove hardcoded node certificate validity period and use host time on node creation instead
- [x] Add test for service certificate 
- [x] Agree on strategy for initial validity period for joining nodes 
- [x] Add `validity_period_days` parameter to `transition_node_to_trusted` proposal
- [x] Tighten end-to-end tests
- [x] Figure out strategy for adding fields to KV table and backwards compatibility
- [ ] Local hook -> global hook to certificate refresh (follow-up PR, as requires extra refactoring)

### DIFF
--- a/js/ccf-app/src/consensus.ts
+++ b/js/ccf-app/src/consensus.ts
@@ -15,16 +15,22 @@ import { ccf } from "./global.js";
 /**
  * @inheritDoc CCFConsensus.getLastCommittedTxId;
  */
-export const getLastCommittedTxId = ccf.consensus.getLastCommittedTxId;
+export const getLastCommittedTxId = ccf.consensus.getLastCommittedTxId.bind(
+  ccf.consensus
+);
 
 /**
  * @inheritDoc CCFConsensus.getStatusForTxId;
  */
-export const getStatusForTxId = ccf.consensus.getStatusForTxId;
+export const getStatusForTxId = ccf.consensus.getStatusForTxId.bind(
+  ccf.consensus
+);
 
 /**
  * @inheritDoc CCFConsensus.getViewForSeqno;
  */
-export const getViewForSeqno = ccf.consensus.getViewForSeqno;
+export const getViewForSeqno = ccf.consensus.getViewForSeqno.bind(
+  ccf.consensus
+);
 
 export { TransactionStatus } from "./global";
--- a/js/ccf-app/src/historical.ts
+++ b/js/ccf-app/src/historical.ts
@@ -39,11 +39,13 @@ export const historicalState = ccf.historicalState;
 /**
  * @inheritDoc CCFHistorical.getStateRange
  */
-export const getStateRange = ccf.historical.getStateRange;
+export const getStateRange = ccf.historical.getStateRange.bind(ccf.historical);
 
 /**
  * @inheritDoc CCFHistorical.dropCachedStateRange
  */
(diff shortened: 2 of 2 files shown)


python/ccf/proposal_generator.py
```diff
@@ -325,6 +325,15 @@ def set_jwt_public_signing_keys(issuer: str, jwks_path: str, **kwargs):
     return build_proposal("set_jwt_public_signing_keys", args, **kwargs)
 
 
+@cli_proposal
+def renew_node_certificate(node_id: str, valid_from: str, valid_to: str, **kwargs):
+    return build_proposal(
+        "renew_node_certificate",
+        {"node_id": node_id, "valid_from": valid_from, "valid_to": valid_to},
```
jumaffre, line 332: I would like to revisit this and only have a `renew_node_certificates` proposal (with an optional `node_id` as argument). `valid_from` would be derived from the host untrusted time, and `valid_to` from the service-wide `node_cert_maximum_validity_period_days` configuration. I think this is much better from a usability perspective (similar to https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/). 

The only drawback is that members wouldn't be able to review the untrusted time before it is set in the node certificates, which may cause some availability issues. Automatically adding the untrusted time as a proposal argument inside CCF sounds appealing (so that voters can review it), but this would break the governance history (signed request and proposal don't match).

jumaffre: As discussed with @achamayou, it is preferable to stick to the manual `valid_to`/`valid_from` proposal arguments and avoid having the primary node implicitly use its untrusted time to determine the new validity of the node certificate. Otherwise, one malicious host could DOS the entire service (by setting the untrusted host time to far in the past/future). Additionally, the constitution may decide to skip member votes for that proposal, which means that no member could verify the validity of the host time, even if this one was included in the proposal.

A better future alternative would be to refresh the quote, and use the collateral time (which isn't extractable just yet, see #59) to deduce the certificate validity period. 

In any case, certificate refresh being such an infrequent event that the setting the validity period manually in the certificate is a fair trade-off.

python/ccf/proposal_generator.py
```diff
@@ -325,6 +325,15 @@ def set_jwt_public_signing_keys(issuer: str, jwks_path: str, **kwargs):
     return build_proposal("set_jwt_public_signing_keys", args, **kwargs)
 
 
+@cli_proposal
+def renew_node_certificate(node_id: str, valid_from: str, valid_to: str, **kwargs):
```
jumaffre, line 329: Should the proposal name be "trigger_node_certificate_renewal"?

achamayou: Yes!

PR #2927, opened 2021-08-27
Run some end to end tests concurrently

achamayou:
Run some tests concurrently, without messing up logging too much.

image

The cpp test is usually ~160s, and the js one ~120s, and this seems to run <170s without any instability, so it's a nice quick win imo.

governance_test is around 46/47s usually, and is now down to 17s.

TODO:

- [x] Logger wrapper
- [x] Set return code correctly when a test breaks
- [x] Pack more tests together
- [x] Investigate:

```

50:     └ <function service_startups at 0x7f106b4acca0>
50: 
50:   File "/__w/2/s/tests/membership.py", line 146, in service_startups
50:     network.start_and_join(args)
50:     │       │              └ Namespace(N=False, R=None, bft_view_change_timeout_ms=5000, binary_dir='/__w/2/s/build', ccf_version=None, client_connection_...
50:     │       └ <function Network.start_and_join at 0x7f106bbea5e0>
50:     └ <infra.network.Network object at 0x7f106a3d7c40>
50: 
50:   File "/__w/2/s/tests/infra/network.py", line 418, in start_and_join
50:     self.consortium.activate(self.find_random_node())
50:     │    │          │        │    └ <function Network.find_random_node at 0x7f106bbeb0d0>
50:     │    │          │        └ <infra.network.Network object at 0x7f106a3d7c40>
50:     │    │          └ <function Consortium.activate at 0x7f106bbe6790>
50:     │    └ <infra.consortium.Consortium object at 0x7f106b4338b0>
50:     └ <infra.network.Network object at 0x7f106a3d7c40>
50: 
50:   File "/__w/2/s/tests/infra/consortium.py", line 142, in activate
50:     m.ack(remote_node)
50:     │ │   └ <infra.node.Node object at 0x7f106a3b8bb0>
50:     │ └ <function Member.ack at 0x7f106d4aa280>
50:     └ <infra.member.Member object at 0x7f106b437760>
50: 
50:   File "/__w/2/s/tests/infra/member.py", line 176, in ack
50:     raise UnauthenticatedMember(
50:           └ <class 'infra.member.UnauthenticatedMember'>
50: 
50: infra.member.UnauthenticatedMember: Failed to ack member member1: 401
```

This is caused by the mechanism we use to deliberately corrupt signatures setting a class var and therefore affecting all threads signing requests during that time.


tests/e2e_logging.py
```diff
@@ -1331,13 +1344,23 @@ def run(args):
...
+    jsargs.package = "libjs_generic"
+    jsargs.nodes = infra.e2e_args.max_nodes(jsargs, f=0)
+    jsargs.initial_user_count = 4
+    jsargs.initial_member_count = 2
+    jsargs.label = f"js_{label}"
+
+    cppargs = copy.deepcopy(jsargs)
+    cppargs.js_app_bundle = None
+    cppargs.package = "liblogging"
+    jsargs.label = f"cpp_{label}"
```
eddyashton, line 1359:
```suggestion
    cppargs.label = f"cpp_{label}"
```

tests/e2e_logging.py
```diff
@@ -1331,13 +1344,23 @@ def run(args):
...
+
+    js = threading.Thread(name="JS", target=run, args=[jsargs])
+    cpp = threading.Thread(name="CPP", target=run, args=[cppargs])
+    js.start()
+    cpp.start()
+    js.join()
+    cpp.join()
```
eddyashton, line 1366: I think we should have CLI arguments to run just one of these, to simplify debugging.

achamayou: Yes, that's a good idea, and also to list all the available ones probably. I think we also want an easy mapping from the handle name to the label too, so we can find the directories easily.

So how about:

$TEST_NAME := name of test in CTest
$SUBTEST_NAME := name of Python thread running a portion of a test
$LABEL := $TEST_NAME + "_" + $SUBTEST_NAME

./test.py -N           list all sub-tests in test
./test.py -R $STR   run only sub-tests whose name includes $STR

That's very basic but consistent with CTest and easy to use for now.

eddyashton: Sounds good to me.

achamayou:
One awkward aspect of doing things this way is the CLI quickly accumulates a lot of options and switches specific to individual tests. I don't know what we can do to avoid that.

achamayou:
Unrelated tpcc_bft failure, the tests affected by this PR are green.

CMakeLists.txt
```diff
@@ -492,17 +492,6 @@ if(BUILD_TESTS)
       ${CCF_DIR}/src/runtime_config/default/apply.js
   )
 
-  # Storing signed governance operations
```
achamayou, line 495: Merged into single governance test

tests/infra/runner.py
```diff
@@ -200,3 +205,88 @@ def run(get_command, args):
...
+    def add(self, prefix, target, **args_overrides):
+        args_ = copy.deepcopy(self.args)
+        for k, v in args_overrides.items():
+            setattr(args_, k, v)
+        args_.label = f"{prefix}_{self.args.label}"
+        self.threads.append(threading.Thread(name=prefix, target=target, args=[args_]))
+
+    def run(self, max_concurrent=None):
+        config = {
+            "handlers": [
```
jumaffre, line 257: I wonder whether `thread.name` should be at the very start of the line so that it's easier to read and parse?

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -29,9 +39,281 @@ def run(args):
...
+    with infra.network.network(
+        args.nodes, args.binary_dir, args.debug_nodes, args.perf_nodes, pdb=args.pdb
+    ) as network:
+        network.start_and_join(args)
+        network = test_stack_size_limit(network, args)
+        network = test_heap_size_limit(network, args)
+
+
+@reqs.description("JWT authentication")
+def test_jwt_auth(network, args):
```
jumaffre, line 82: This and other tests have moved here but isn't JS-specific I don't think?

jumaffre (approved):

tests/infra/runner.py
achamayou: It seemed logical to group it with the module and function name, rather than put it on the other side of the timestamp, but I don't feel strongly either way, happy to move it left.

tests/js-custom-authorization/custom_authorization.py
achamayou: No, they're not, they're just a grouping of things that deal with auth/validation and can run on single node network, so they can be run at the same time without timeouts. I've deliberately not renamed anything, but we probably will want to. Input welcome!

achamayou: We can keep the files split, the reason I moved them is because they weren't usable/importable as modules where they were.

tests/infra/runner.py
jumaffre: I've just had a look at the CI output and it actually looks quite neat as it is!

PR #2931, opened 2021-09-01
Split identity

wintersteiger:
Details still very much in progress, but it would be great to get some feedback of the overall architecture and integration. 

High level plan: `splitid.h` will be CCF-independent so that we can publish it separately. `splitid_endpoints.h` and `splitid_context.h` are the CCF-specific glue.


src/node/splitid_context.h
```diff
@@ -0,0 +1,1053 @@
...
+      http::Request request(fmt::format(
+        "/{}/{}",
+        ccf::get_actor_prefix(ccf::ActorsType::nodes),
+        "splitid/register"));
+      request.set_header(
+        http::headers::CONTENT_TYPE, http::headervalues::contenttype::JSON);
+
+      auto body = serdes::pack(in, serdes::Pack::Text);
+      request.set_body(&body);
+      return node_client->make_request(request);
```
jumaffre, line 334: As discussed, this pattern is repeated in other functions in the file, so perhaps refactor?

src/node/splitid_endpoints.h
```diff
@@ -0,0 +1,643 @@
...
+      } \
+    }; \
+  endpoints \
+    ->make_endpoint( \
+      URL, \
+      HTTP_POST, \
+      json_adapter(T##_endpoint_fun), \
+      {std::make_shared<NodeCertAuthnPolicy>()}) \
+    .set_forwarding_required(endpoints::ForwardingRequired::Always) \
+    .set_openapi_hidden(true) \
```
eddyashton, line 525: If I'm reading correctly, then replacing this line with `.set_auto_schema<T::In, bool>()` should add useful auto-generated OpenAPI entries for these endpoints (at least saying "the request should be a `SamplingDealRPC::In`).

tests/splitid.py
```diff
@@ -0,0 +1,196 @@
...
+    return signatures
+
+
+def wait_for_identity(primary, retry_timeout):
+    have_identity = False
+    public_key = ""
+    with primary.client(timeout=60) as c:
+        timeout_before = time()
+        while time() < timeout_before + retry_timeout:
+            res = c.post("/node/splitid/get-identity", timeout=60)
```
jumaffre, line 96: Why 60 seconds here?

src/consensus/aft/raft.h
```diff
@@ -814,11 +813,17 @@ namespace aft
           (globally_committable ? " committable" : ""),
           hooks->size());
 
+        // Hooks may need to take the state lock if they trigger RPCs (via
+        // frontend::process), so we temporarily release it.
+        state->lock.unlock();
```
jumaffre, line 818: Releasing the lock here seems dangerous. `frontend::process` would eventually call into Raft again when the underlying transaction is passed back to the consensus for replication. This is something we've encountered before and I believe the best option is to call `frontend::process()` in an async task.

src/node/http_node_client.h
```diff
@@ -57,16 +106,45 @@ namespace ccf
...
-
-      if (rs != HTTP_STATUS_OK)
-      {
-        auto ser_res = ctx->serialise_response();
-        std::string str((char*)ser_res.data(), ser_res.size());
-        LOG_FAIL_FMT("Request failed: {}", str);
-      }
-
-      return rs == HTTP_STATUS_OK;
+      threading::retry_until(
```
jumaffre, line 109: This feels like this shouldn't be in the underlying HTTP client but instead as a wrapper around it.

src/node/http_node_client.h
```diff
@@ -57,16 +106,45 @@ namespace ccf
...
+              frontend->process(ctx);
+          }
+          return false;
+        },
+        std::chrono::milliseconds(100));
+
+      return ctx->get_response_status() == HTTP_STATUS_OK;
     }
 
     bool submit_orc(const NodeId& from, kv::ReconfigurationId rid) override
```
jumaffre, line 150: Same comment for the code below, which I didn't know was already there.

tests/splitid.py
```diff
@@ -0,0 +1,196 @@
...
+        new_node = network.create_node("local://localhost")
+        args.timeout = 60
+        network.join_node(new_node, args.package, args, from_snapshot=False)
+        network.trust_node(new_node, args)
+        with new_node.client() as c:
+            s = c.get("/node/state")
+            assert s.body.json()["node_id"] == new_node.node_id
+            assert (
+                s.body.json()["startup_seqno"] == 0
+            ), "Node started without snapshot but reports startup seqno != 0"
```
jumaffre, line 153:
```suggestion
```

tests/splitid.py
jumaffre: Should this be `c.get("/node/splitid/identity")` to be more RESTful?

src/node/splitid_endpoints.h
```diff
@@ -0,0 +1,643 @@
...
+      {
+        return make_error(
+          HTTP_STATUS_BAD_REQUEST, ccf::errors::InternalError, ex.what());
+      }
+    };
+
+    endpoints
+      ->make_endpoint(
+        "splitid/get-signature",
+        HTTP_POST,
```
eddyashton, line 385:
```suggestion
        "splitid/signature",
        HTTP_GET,
```

Should look at the other endpoints HTTP methods as well. I think that (mostly) the URLs are nouns? In which case we should use `GET`/`POST` as appropriate.

tests/splitid.py
```diff
@@ -0,0 +1,196 @@
...
+                else:
+                    cfg = rj["details"]["configs"][0]
+                    if new_node.node_id in rj["details"]["configs"][0]["nodes"]:
+                        LOG.info(f"active config: {cfg}")
+                        break
+                    else:
+                        LOG.info(f"configruation does not contain new node yet: {cfg}")
+                        sleep(1.0)
+
+        # Wait until resharing is done
```
eddyashton, line 173: Putting a comment here for reference, though it may not change anything in this PR: The next line doesn't actually wait at all (the earlier loop believes resharing is done, then the next line will return instantly since _an_ identity already exists to be returned, and always will be returned). Instead of this endpoint, we'd really like to confirm this by examining the network cert available from existing endpoints. We should put a custom claim in the x509, at least saying "this is _a_ shared identity", and ideally "this is shared between _these holders_" somehow, so that this reshare-due-to-reconfiguration can be confirmed.

wintersteiger:
Closing this for now, until we have a clear strategy for CFT identity.

PR #2933, opened 2021-09-02
Disable suites in CI, leave them in Daily

achamayou:
This is to save execution time, because suites only relatively rarely seem to catch errors that individual tests don't in CI.

They remain very useful though, because they've caught subtle problems more than once, so we do want to run them in Daily.

### DIFF
--- a/.azure-pipelines-templates/matrix.yml
+++ b/.azure-pipelines-templates/matrix.yml
@@ -29,9 +29,9 @@ parameters:
 
   test:
     NoSGX:
-      ctest_args: '-LE "benchmark|perf|tlstest|vegeta|partitions"'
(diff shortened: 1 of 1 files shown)


wintersteiger (approved):
I can see that the suites are added to the matrix, but I can't see how they are removed elsewhere (non-daily).

Found it, I just missed the `-LE`, all good.

achamayou:
@wintersteiger -LE stands for Label Exclude, so this change says "run all the tests, except those with these labels", and that now includes "suite", so it removes them from the CI.

The daily has its own matrix, called daily-matrix.yml. With even more templating, this could be avoided, but I do the templating and I claim we're templating enough as it is.

eddyashton (approved):

PR #2943, opened 2021-09-06
Fast sig e2e tests (take 2)

jumaffre:


jumaffre:
This should only be merged once #2958 is implemented, as the current snapshot logic causes some instability (see #2951).

PR #2947, opened 2021-09-06
Bump minimum cmake version to 3.16

### LINKED ISSUE #2946: Bump minimum CMake version to 3.16
CCF targets Ubuntu 20.04 now. This version ships CMake 3.16.3, while CCF in its build declares 3.11 as minimum supported version. Since that version is not tested in CI it makes sense to bump the minimum supported version to the one that is actually tested (3.16).

Bumping to 3.16 may change some CMake defaults, just something to watch out for.

letmaik:
Closes #2946.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache 2.0 License.
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.16)
 
 set(CCF_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 include(${CCF_DIR}/cmake/preproject.cmake)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache 2.0 License.
-cmake_minimum_required(VERSION 3.11)
```
achamayou, line 3: Please mention this in the changelog, since it is potentially user-visible (though unlikely to break).

achamayou (approved):

CMakeLists.txt
letmaik: I added "- Raised the minimum supported CMake version for building CCF to 3.16 (#2946).", specifically mentioning "building CCF", as the requirement doesn't propagate to installations directly. Though you could argue that some more complex hand-written CMake modules like `ccf_app.cmake` are shipped and we're free to use the features of 3.16 there as well. I'm slightly leaning towards removing "for building CCF", even though the requirement isn't enforced during `find_package(ccf)`.

achamayou: It does propagate indirectly because we may end up relying on 3.16 features, which may leak out via ccf_app.cmake indeed, so I think that's a good way to phrase it.

PR #2954, opened 2021-09-06
Use cmake helpers for `http_parser` library

letmaik:
Note: `POSITION_INDEPENDENT_CODE` is already set by the `add_enclave_library_c`/`add_host_library` helpers.


achamayou (approved):

achamayou:
@letmaik looking back at this, it's not quite right: the add_host_library() does set PIC, but also compile and link time flags that are C++ specific and unwanted in this case.

Could you revert the second part of the change?

PR #2955, opened 2021-09-07
Remove usage of CMake's `InstallRequiredSystemLibraries`

### LINKED ISSUE #2945: Clarify or remove usage of CMake's `InstallRequiredSystemLibraries`
https://github.com/microsoft/CCF/blob/3713c7c9d3178f6cf8c1e07894818829f562e114/cmake/cpack_settings.cmake#L4

I can't see how this actually affects anything, in particular in relation to CPack. The `.deb` package doesn't seem to contain anything not explicitly defined.

letmaik:
Closes #2945.


achamayou:
@letmaik please poke the daily_canary, so the daily runs and builds a release package, installs it and builds against it. I think that's the best way to test this change.

letmaik:
> @letmaik please poke the daily_canary, so the daily runs and builds a release package, installs it and builds against it. I think that's the best way to test this change.

Done, two unrelated test failures: `raft_test` and `reconfiguration_test_bft`.

achamayou (approved):

PR #2965, opened 2021-09-09
Combine adding users and opening

achamayou:
This is a small speedup/logging reduction on startup.


wintersteiger (approved):
LGTM. Thanks for adding this so quickly!

Issue #2958, opened 2021-09-07
Include evidence receipt in snapshot file


jumaffre:
The integrity/provenance of a snapshot has to currently be verified against a usually-short ledger suffix. For example, for a snapshot `snapshot_X_Y.committed_Z`, a new joiner has to verify that the evidence of the snasphot has a commit proof (i.e. a signature tx containing `commit_seqno` > `Y`) in the ledger suffix (i.e. before seqno `Z`). 

This is awkward for operators when starting up a node from a snapshot, or for auditors when verifying the integrity of a snapshot. Instead, we should include the receipt of the evidence directly in the snapshot file. 

Because snapshot files are currently not versioned, we will probably need to bump the version format for snapshot files in https://microsoft.github.io/CCF/main/audit/ledger.html#transaction-format to indicate that a snapshot also contains a receipt.

PR #2967, opened 2021-09-09
Add hooks for `kv::Set` and `kv::Value` and convert remaining service tables

jumaffre:
Fairly standalone change, which will be required for adding hooks on the signatures table (as part of #2958). 

The `kv::Set` and `kv::Value` (introduced in #2599) didn't yet support local and global commit hooks. This PR adds support for those and also convert the remaining service tables to using `kv::Value`.

### DIFF
--- a/src/kv/set.h
+++ b/src/kv/set.h
@@ -36,9 +36,49 @@ namespace kv
     using WriteOnlyHandle = kv::WriteableSetHandle<K, KSerialiser, Unit>;
     using Handle = kv::SetHandle<K, KSerialiser, Unit>;
 
-    using KeySerialiser = KSerialiser;
+    // Note: The type V of the value `std::optional<V>` does not matter here.
+    // The optional type is required to differentiate additions from deletions,
+    // and to provide a consistent interface with the more generic `TypedMap`.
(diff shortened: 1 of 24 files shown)


src/kv/set.h
```diff
@@ -36,9 +36,49 @@ namespace kv
     using WriteOnlyHandle = kv::WriteableSetHandle<K, KSerialiser, Unit>;
     using Handle = kv::SetHandle<K, KSerialiser, Unit>;
 
-    using KeySerialiser = KSerialiser;
+    // Note: The type V of the value `std::optional<V>` does not matter here.
+    // The optional type is required to differentiate additions from deletions,
+    // and to provide a consistent interface with the more generic `TypedMap`.
+    using Write = std::map<K, std::optional<kv::serialisers::SerialisedEntry>>;
```
jumaffre, line 42: The hook for `kv::Set` is a little awkward because we need to be able to differentiate additions from deletions (of a key). We could pass two separate lists to the hooks, but it would be inconsistent with the hook on `TypedMap`. Instead, we use a single map, with an optional value whose type `kv::serialisers::SerialisedEntry` doesn't really matter.

achamayou (approved):

PR #2975, opened 2021-09-10
Remove root from receipts

### LINKED ISSUE #2921: Remove root from receipts by default
Having the root in receipts is convenient for debugging, but counter-productive in terms of usability, since it makes it easier to implement incomplete/incorrect verification.

It should be off by default.

achamayou:
Fix #2921

### DIFF
--- a/include/ccf/receipt.h
+++ b/include/ccf/receipt.h
@@ -17,7 +17,7 @@ namespace ccf
     };
 
     std::string signature;
-    std::string root;
+    std::optional<std::string> root = std::nullopt;
     std::vector<Element> proof = {};
     std::string leaf;
     ccf::NodeId node_id;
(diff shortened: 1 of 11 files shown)


wintersteiger (approved):
I totally agree, the root doesn't need to be in there.

include/ccf/receipt.h
```diff
@@ -27,5 +27,6 @@ namespace ccf
   DECLARE_JSON_REQUIRED_FIELDS(Receipt::Element);
   DECLARE_JSON_OPTIONAL_FIELDS(Receipt::Element, left, right)
   DECLARE_JSON_TYPE(Receipt)
```
jumaffre, line 29:
```suggestion
  DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(Receipt)
```

jumaffre (approved):

Issue #2640, opened 2021-06-04
Distinguish UNKNOWN/PENDING from INVALID in historical queries wrapper


achamayou:
We have unfortunately not versioned `ccf::historical::adapter`, so I think we will:

1. rename `ccf::historical::adapter` to `ccf::historical::adapter_v1`
2. alias `ccf::historical::adapter` to `ccf::historical::adapter_v1`
3. add a `ccf::historical::adapter_v2` that returns different errors on UNKNOWN/PENDING and INVALID

achamayou:
After discussion with @letmaik, the proposal is that we return either:

- 404 with `TransactionPendingOrUnknown` and `No-Cache`
- 404 with `TransactionInvalid`

letmaik:
There is a further detail here, which is the cache-not-ready response:
https://github.com/microsoft/CCF/blob/ce877fcbb3451930fbb784dd8b054027d1f406da/include/ccf/historical_queries_adapter.h#L132-L146

Does it make sense to add this to the first 404 case?

achamayou:
@letmaik do you mean set a Retry-After?

letmaik:
No, it's just another case not captured in this issue but which clients have to consider.

achamayou:
I don't understand, what's the other case not captured by this issue? The snippet you posted is for transaction that are valid, but which we are still fetching. It's a 200.

This issue is about adding nuance to the "not committed (yet)" case, to differentiate "not committed yet, as far as we know" from "can never commit". Am I missing something?

letmaik:
Nevermind, 202 is fine for that case, even though I think the payload shouldn't be a plain-text string.

achamayou:
@letmaik happy to look at alternative payload proposals.

PR #2976, opened 2021-09-14
Version historical query adapter

achamayou:
Preliminary step for #2640.

At first I thought this was finally an occasion to use a function alias, but that doesn't support default arguments, so it wasn't.

### DIFF
--- a/src/endpoints/common_endpoint_registry.cpp
+++ b/src/endpoints/common_endpoint_registry.cpp
@@ -226,30 +226,8 @@ namespace ccf
 
     auto is_tx_committed =
       [this](ccf::View view, ccf::SeqNo seqno, std::string& error_reason) {
-        if (consensus == nullptr)
-        {
-          error_reason = "Node is not fully configured";
-          return false;
-        }
-
-        const auto tx_view = consensus->get_view(seqno);
-        const auto committed_seqno = consensus->get_committed_seqno();
-        const auto committed_view = consensus->get_view(committed_seqno);
-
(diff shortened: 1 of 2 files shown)


jumaffre (approved):

src/endpoints/common_endpoint_registry.cpp
```diff
@@ -226,30 +226,8 @@ namespace ccf
 
     auto is_tx_committed =
```
jumaffre, line 227: This function has to be defined by all apps using historical endpoints, even though it is very much framework wide. Could the `available` argument of `adapter()` default to `ccf::historical::is_tx_committed` instead (just like for `extractor`)?

achamayou: I wondered that, but for it to work, the adapter would need access to the consensus somehow instead of being a pure function.

eddyashton: That's right, this is a functor purely so it can capture `consensus`. Consensus could be a singleton, or otherwise globally-available, but we've avoided that so far.

eddyashton (approved):

src/endpoints/common_endpoint_registry.cpp
achamayou: Ok, let's wrap up #2640 first and think about this separately. I would prefer to have the adapter as a frontend mixin and avoid singletons in the code base as much as possible.

PR #2977, opened 2021-09-14
Historical adapter v2

achamayou:
Address #2640. This is essentially two changes:

1. Switch from 400 to 404, for historical transactions on requests that aren't committed (yet).
2. Give a different error reason for transaction ids that are invalid (forever) vs unknown/pending (so far, but may resolve).

TODO:

- [x] CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Added `foreach_key` and `foreach_value` to C++ KV API, to iterate without deserializing both entries when only one is used (#2918).
+- `ccf::historical::adapter_v2` now returns 404, with either `TransactionPendingOrUnknown` or `TransactionInvalid`, rather than 400 when a user performs a historical query for a transaction id that is not committed.
 
 ### Changed
 
@@ -18,7 +19,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated `actions.js` constitution fragment to record service-endorsed node certificate on the `transition_node_to_trusted` action. The constitution should be updated using the existing `set_constitution` proposal (#2844).
 - Improved performance for lookup of path-templated endpoints (#2918).
 - Raised the minimum supported CMake version for building CCF to 3.16 (#2946).
-- `ccf::historical::adapter_v2` now returns 404, with either `TransactionPendingOrUnknown` or `TransactionInvalid`, rather than 400 when a user performs a historical query for a transaction id that is not committed.
 
 ## [2.0.0-dev3]
 
--- a/include/ccf/historical_queries_adapter.h
+++ b/include/ccf/historical_queries_adapter.h
@@ -50,7 +50,7 @@ namespace ccf::historical
     return tx_id_opt;
   }
 
-  static inline bool is_tx_committed(
+  static inline bool is_tx_committed_v1(
     kv::Consensus* consensus,
     ccf::View view,
     ccf::SeqNo seqno,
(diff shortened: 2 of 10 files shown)


include/ccf/historical_queries_adapter.h
```diff
@@ -83,9 +83,148 @@ namespace ccf::historical
...
+              ccf::errors::TransactionPendingOrUnknown,
+              std::move(error_reason));
+            return;
+          case HistoricalTxStatus::Invalid:
+            args.rpc_ctx->set_error(
+              HTTP_STATUS_NOT_FOUND,
+              ccf::errors::TransactionInvalid,
+              std::move(error_reason));
+            return;
+          case HistoricalTxStatus::Valid:;
```
eddyashton, line 196:
```suggestion
          case HistoricalTxStatus::Error:
          {
            args.rpc_ctx->set_error(
              HTTP_STATUS_INTERNAL_SERVER_ERROR,
              ccf::errors::TransactionPendingOrUnknown,
              std::move(error_reason));
            return;
          }
          case HistoricalTxStatus::PendingOrUnknown:
          {
            // Set header No-Cache
            args.rpc_ctx->set_response_header(
              http::headers::CACHE_CONTROL, "no-cache");
            args.rpc_ctx->set_error(
              HTTP_STATUS_NOT_FOUND,
              ccf::errors::TransactionPendingOrUnknown,
              std::move(error_reason));
            return;
          }
          case HistoricalTxStatus::Invalid:
          {
            args.rpc_ctx->set_error(
              HTTP_STATUS_NOT_FOUND,
              ccf::errors::TransactionInvalid,
              std::move(error_reason));
            return;
          }
          case HistoricalTxStatus::Valid:
          {
            continue;
          }
```

include/ccf/historical_queries_adapter.h
```diff
@@ -83,9 +83,148 @@ namespace ccf::historical
...
+      ccf::TxID target_tx_id;
+      {
+        const auto tx_id_opt = extractor(args);
+        if (tx_id_opt.has_value())
+        {
+          target_tx_id = tx_id_opt.value();
+        }
+        else
+        {
+          return;
```
letmaik, line 163: I know it's the same as in v1, but... what happens exactly in this case?

letmaik (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated `actions.js` constitution fragment to record service-endorsed node certificate on the `transition_node_to_trusted` action. The constitution should be updated using the existing `set_constitution` proposal (#2844).
 - Improved performance for lookup of path-templated endpoints (#2918).
 - Raised the minimum supported CMake version for building CCF to 3.16 (#2946).
+- `ccf::historical::adapter_v2` now returns 404, with either `TransactionPendingOrUnknown` or `TransactionInvalid`, rather than 400 when a user performs a historical query for a transaction id that is not committed.
```
eddyashton, line 21: I think this should be under `Added` rather than `Changed`? But could be convinced.

include/ccf/historical_queries_adapter.h
```diff
@@ -83,9 +83,148 @@ namespace ccf::historical
...
+          target_tx_id.to_str()));
+        return;
+      }
+
+      // Call the provided handler
+      f(args, historical_state);
+    };
+  }
+
   static ccf::endpoints::EndpointFunction adapter_v1(
```
eddyashton, line 228: I think we'd like to remove this in 2.0, so we should deprecate it now if possible (`CCF_DEPRECATED("Will be removed in 2.0")` in the right place).

eddyashton (approved):

CHANGELOG.md
achamayou: Yes, will fix!

include/ccf/historical_queries_adapter.h
achamayou: Yes

include/ccf/historical_queries_adapter.h
achamayou: I mean, ok I guess, but that `continue`...

include/ccf/historical_queries_adapter.h
achamayou: That's a good question, whatever we do when we return from a handler without setting anything, which appears to be an empty 200?

achamayou: So no, what happens is the extractor has already set the return code and the error body.

Issue #2970, opened 2021-09-10
JS FFI plugin design


letmaik:
Part of https://github.com/microsoft/CCF/issues/2935.

https://github.com/microsoft/CCF/pull/2961 already contains initial steps towards a design.

There are multiple parts:
- Defining a plugin
- Building a plugin
- Consuming a plugin

Plugins need to be consumable within `js_generic` and custom (C++) apps, where the latter would benefit JS governance only.

To achieve that, I propose to add a new app-provided function `get_js_plugins` similar to `get_rpc_handler`. It returns a vector of JS FFI plugin structs (each defining the plugin `name`, `ccf_version`, quickjs `extend` function).

JS FFI plugins that are binary distributed need to make available two things:
- Static archives (enclave and virtual)
- Header file to retrieve the plugin definition for `get_js_plugins`

To support `js_generic`, new static libraries `js_generic.enclave.a`/`js_generic.virtual.a` containing `js_generic.cpp` are built and installed. The existing `js_generic` app is changed to have a single source file `plugins.cpp` with `get_js_plugins` only and linking against `js_generic.enclave.a`/`js_generic.virtual.a` and all JS plugin static libraries. This will come with all available plugins (e.g. the Open Enclave plugin). Users can choose to use this ready-made app or build a custom `js_generic` app by providing their own `plugins.cpp` and linking to the desired plugin static libraries.

Ideally, the plugin mechanism is used for every currently defined global, which would be `ccf` and `openenclave`. `ccf` contains a lot of things that depend on CCF context, while `openenclave` is independent. Supporting all use cases of `ccf` (KV access, special node operations, etc.) means that the signature of the plugin's `extend` function (called when a plugin is ready to be injected into a quickjs context) has to provide that context.

In https://github.com/microsoft/CCF/pull/2961, `extend` does not provide any CCF context: `std::function<void(JSContext* ctx)> extend;`
This would have to change to support `ccf`.

Feedback on the above proposal very welcome!

letmaik:
Done in https://github.com/microsoft/CCF/pull/2978.

PR #2978, opened 2021-09-14
JS FFI plugin building

letmaik:
Follow-up on https://github.com/microsoft/CCF/pull/2961. Implements https://github.com/microsoft/CCF/issues/2970. I had to change the plan slightly in regards to the planned `plugins.cpp` and instead provide an indirection to `get_rpc_handler`. Has to do with linking archives vs objects when resolving symbols. Also, I added `_base` to the new static libraries to avoid a naming conflict with the shared libraries `js_generic.*`.

The CMake code to consume an existing plugin is simple, but the CMake code to build one is still complex. I may improve on that in a future PR.

There are two new public headers:
- `js_plugin.h`: (moved from `src/js/plugin.h`) Defines the plugin interface.
- `js_openenclave_plugin.h`: The Open Enclave plugin definition, available when linking against `js_openenclave.{enclave,virtual}`.

I also renamed `js::` to `ccf::js` to keep the public API clean.


samples/apps/logging/logging.cpp
```diff
@@ -1198,11 +1198,16 @@ namespace loggingapp
...
     ccf::NetworkTables& nwt, ccfapp::AbstractNodeContext& context)
   {
     return make_shared<loggingapp::Logger>(nwt, context);
   }
-  // SNIPPET_END: rpc_handler
+
+  std::vector<ccf::js::FFIPlugin> get_js_plugins()
+  {
+    return {};
+  }
```
eddyashton, line 1211: I'd really like to avoid requiring this extra bit of boilerplate in every app, if possible. I think we can do that with a weak symbol?

src/js/wrap.cpp
```diff
@@ -19,1580 +18,1603 @@
 #include <quickjs/quickjs-exports.h>
 #include <quickjs/quickjs.h>
 
-namespace js
+namespace ccf
```
eddyashton, line 21: You can do `namespace ccf::js` to remove the indent diff here/elsewhere.

samples/apps/logging/logging.cpp
letmaik: I was just thinking the same, after fixing more apps...

src/js/wrap.cpp
letmaik: Thanks, did not know that!

samples/apps/logging/logging.cpp
letmaik: Done.

cmake/common.cmake
```diff
@@ -348,12 +348,80 @@ set(CCF_NETWORK_TEST_ARGS -l ${TEST_HOST_LOGGING_LEVEL} --worker-threads
                           ${WORKER_THREADS}
 )
 
+if("sgx" IN_LIST COMPILE_TARGETS)
```
jumaffre, line 351: Is the plan to ultimately move this to new cmake file located near the plugin code so that the plugin is self contained?

include/ccf/app_interface.h
```diff
@@ -27,5 +30,11 @@ namespace ccfapp
    */
   std::shared_ptr<ccf::RpcFrontend> get_rpc_handler(
     ccf::NetworkTables& network, AbstractNodeContext& context);
-  // SNIPPET_END: rpc_handler
+
+  /** To be implemented by the application to be registered by CCF.
```
jumaffre, line 34: I think this comment is a little confusing: C++ apps don't need to implement this and only the `js_generic` app (which is so far maintained by us) needs to do that.

jumaffre: Relatedly, `js_plugin.h` is included here but the app may be C++ only. Is there a way `get_js_plugins()` could be separated from the generic app interface?

letmaik: > I think this comment is a little confusing: C++ apps don't need to implement this and only the `js_generic` app (which is so far maintained by us) needs to do that.

C++ apps don't need to, but they can, and if they do then it affects JS governance.

cmake/common.cmake
letmaik: Yes.

include/ccf/app_interface.h
jumaffre: "affects JS governance": Good point, I'd missed that.

doc/build_apps/api.rst
```diff
@@ -5,6 +5,7 @@ A CCF application is composed of the following:
 
 - The :ref:`Application Entry Point <build_apps/api:Application Entry Point>` which registers the application in CCF.
 - A collection of :cpp:class:`endpoints <ccf::endpoints::Endpoint>` handling HTTP requests and grouped in a single :cpp:class:`registry <ccf::endpoints::EndpointRegistry>`. An :cpp:class:`endpoint <ccf::endpoints::Endpoint>` reads and writes to the key-value store via the :ref:`Key-Value Store API <build_apps/kv/api:Key-Value Store API>`.
+- An optional set of :ref:`JavaScript FFI Plugins <build_apps/api:JavaScript FFI Plugins>` that shall be registered to extend the built-in JavaScript API surface.
```
achamayou, line 8: shall -> can?

achamayou (approved):

jumaffre (approved):

PR #2988, opened 2021-09-15
js: turn crypto functions into plugin

letmaik:
Feedback needed:
- I added a new `add_ccf_plugin` function to `ccf_app.cmake` which under the hood is not really plugin specific but just produces static enclave/virtual libraries linked against the `ccf` base library. Does this make sense?
- I had to opt in to the plugin in all apps as it is generally needed in (JS) constitutions. It seems a bit boilerplate-y given that likely all apps need it, though on the other hand it allows customization. Opinions?

TODO:
- Changelog entry
- Change JS sample apps / constitutions to use `ccf.crypto.*` instead of `ccf.*`

### DIFF
--- a/src/apps/smallbank/app/smallbank.cpp
+++ b/src/apps/smallbank/app/smallbank.cpp
@@ -3,6 +3,7 @@
 #include "apps/smallbank/smallbank_serializer.h"
 #include "apps/utils/metrics_tracker.h"
 #include "ccf/app_interface.h"
+#include "ccf/js_crypto_plugin.h"
 #include "ccf/user_frontend.h"
 
 #include <charconv>
@@ -470,5 +471,10 @@ namespace ccfapp
     NetworkTables& nwt, AbstractNodeContext& context)
   {
     return make_shared<SmallBank>(*nwt.tables, context);
+  }
+
+  std::vector<ccf::js::FFIPlugin> get_js_plugins()
+  {
+    return {ccf::js::crypto_plugin};
   }
 }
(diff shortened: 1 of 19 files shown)


src/apps/smallbank/app/smallbank.cpp
```diff
@@ -471,4 +472,9 @@ namespace ccfapp
   {
     return make_shared<SmallBank>(*nwt.tables, context);
   }
+
+  std::vector<ccf::js::FFIPlugin> get_js_plugins()
+  {
+    return {ccf::js::crypto_plugin};
+  }
```
eddyashton, line 479: Can we put this in the weakly-linked default implementation, rather than repeating in every app?

letmaik: What about linking? Should the `ccf.{enclave,virtual}` targets always link to the plugin? If the plugin is not used (by overriding `get_js_plugins`, no code would be pulled in.

eddyashton: Hmm. Maybe this default implementation should be in a separate file, so all the apps that want to use it are defined in CMake as `SRCS logging.cpp default_plugins.cpp`, so it looks like less boilerplate in the code? Still a breaking change for any consumer though. I'm tempted to aim really hard to avoid the breaking change (by making the default weak implementation include + require-linking-of these default modules). Ideally we're very lucky with the behaviour of weakly-linked symbols, and we can actually avoid linking their dependencies if they're actually overridden? But I assume that's not the case.

letmaik: > But I assume that's not the case.

Why not? Objects from static archives are only pulled in if they're referenced, which is true since we're not using `--whole-archive`.

letmaik:
Not worth doing at this point.

PR #2989, opened 2021-09-16
Key-value doc fix

rengolin:
There seem to be a find&replace mismatch between example 1 and 2. I
think this is what the original intention was.


eddyashton (approved):
My bad, thanks for the spot!

achamayou (approved):

achamayou:
`stuint64_tring` is my new favourite type.

jumaffre (approved):
/azp run

PR #2991, opened 2021-09-16
Node certs in receipts

achamayou:
Add endorsed certificate of signing node to receipts, to facilitate verification.

This is implemented by writing the certificate in question to every signature transaction at the moment. Once full historical queries are implemented, it is very likely that we want this to be a lookup by node_id instead.

- [x] test receipts in lts_compatibility


src/node/node_state.h
```diff
@@ -372,6 +372,8 @@ namespace ccf
           setup_consensus(ServiceStatus::OPENING, false, endorsed_node_cert);
           setup_progress_tracker();
           setup_history();
+          if (endorsed_node_cert.has_value())
```
jumaffre, line 375: Can this be set in `setup_history()` instead?

achamayou: I don't think that works well for recovery, but I will double check, it's a good point.

src/node/node_state.h
```diff
@@ -561,6 +563,8 @@ namespace ccf
               n2n_channels_cert);
             setup_progress_tracker();
             setup_history();
+            if (endorsed_node_cert.has_value())
```
jumaffre, line 566: This is only for a 2.x node joining a 1.x service. It'd be nicer if it was in the `if` branch on line 539 the history isn't setup yet but a comment here would be useful?

tests/e2e_logging.py
```diff
@@ -656,8 +662,11 @@ def test_historical_receipts(network, args):
             r = first_receipt.json()["receipt"]
             root = ccf.receipt.root(r["leaf"], r["proof"])
             ccf.receipt.verify(root, r["signature"], primary_cert)
+            assert r["cert"], r
+            node_cert = load_pem_x509_certificate(r["cert"].encode(), default_backend())
+            ccf.receipt.check_endorsement(node_cert, network_cert)
```
jumaffre, line 667: Endorsement should be checked first?

tests/e2e_logging.py
```diff
@@ -645,6 +645,12 @@ def test_historical_receipts(network, args):
...
 
+    cert_path = os.path.join(primary.common_dir, "networkcert.pem")
+    with open(cert_path, encoding="utf-8") as c:
+        network_cert = load_pem_x509_certificate(
+            c.read().encode("ascii"), default_backend()
+        )
+
     TXS_COUNT = 5
     network.txs.issue(network, number_txs=5)
     for idx in range(1, TXS_COUNT + 1):
```
jumaffre, line 656: We've made significant changes around receipts/node certs since 1.x. I wonder if we could move the receipt verification to the more generic `logging_app.py` and verify the receipts of some (all?) transactions in the `lts_compatibility` test? For example, a 1.x ledger wouldn't have the node certificate in the receipt and the user would have to retrieve the certificate from elsewhere.

tests/e2e_logging.py
```diff
@@ -656,8 +662,11 @@ def test_historical_receipts(network, args):
             r = first_receipt.json()["receipt"]
```
jumaffre, line 662: It'd be great to add the receipt verification to `tutorial.py` and include it in https://microsoft.github.io/CCF/main/use_apps/verify_tx.html#transaction-receipts and/or https://microsoft.github.io/CCF/main/use_apps/python_tutorial.html (separate PR).

tests/e2e_logging.py
achamayou: I am very reluctant to do this because I think it would be unacceptably slow to run. I also don't like logging_app.py, I find it very opaque. I like to have the assertions inside the testcases directly, not offloaded to another module may or may not be checking what's right in the context of the test. That way lies flag madness in my experience.

Making the check work across 1.x and 2.x shouldn't be too difficult, it's just a matter of falling back to the old logic when the cert isn't in the receipt itself.

src/node/node_state.h
achamayou: As discussed, this isn't possible, but I've cleaned up the flow so the history is always created first, and the conditions for setting the endorsed cert are clearer.

src/node/node_state.h
achamayou: Yes, cleaned up.

tests/e2e_logging.py
achamayou: As discussed, I will update the lts test so we do some cross-upgrade checks.

jumaffre (approved):
I think this will need an entry in the changelog.

doc/use_apps/verify_tx.rst
```diff
@@ -76,15 +77,15 @@ Note that receipts over signature transactions are a special case, for example:
...
-Verifying a receipt is a two-phase process:
+Verifying a receipt is a three-phase process:
 
   - Combine ``leaf`` with the successive elements in ``proof`` to calculate the value of ``root``. See :py:func:`ccf.receipt.root` for a reference implementation.
-  - Verify ``signature`` over the ``root`` using the certificate of the node identified by ``node_id``. See :py:func:`ccf.receipt.verify` for a reference implementation.
-
-When a node certificate is first obtained, it is necessary to check it has been endorsed by the network identity.
\ No newline at end of file
+  - Verify ``signature`` over the ``root`` using the certificate of the node identified by ``node_id`` and ``cert``. See :py:func:`ccf.receipt.verify` for a reference implementation.
+  - Check that the certificate ``cert`` of ``node_id`` used to sign the receipt is endorsed by the CCF network. :py:func:`ccf.receipt.check_endorsement` for a reference implementation.
```
eddyashton, line 91:
```suggestion
  - Check that the certificate ``cert`` of ``node_id`` used to sign the receipt is endorsed by the CCF network. See :py:func:`ccf.receipt.check_endorsement` for a reference implementation.
```

eddyashton (approved):

PR #2998, opened 2021-09-17
Include receipt evidence in snapshot

jumaffre:
Part of #2958 

Very much WIP. 

Instead of having to verify the evidence of a snapshot in a ledger suffix, which complicated operations, the receipt of the evidence is now directory included in the snasphot file, making them self verifiable. 

More details to follow.


tests/reconfiguration.py
```diff
@@ -440,13 +440,6 @@ def run(args):
             test_add_node_from_snapshot(network, args)
             test_add_node_from_snapshot(network, args, from_backup=True)
             test_add_node_from_snapshot(network, args, copy_ledger_read_only=False)
```
jumaffre, line 442: We can now start a node from a snapshot without any ledger at all, whereas previously, a node would automatically reject snapshots whose evidence aren't covered by the ledger, are replay the entire ledger history instead.

python/ccf/ledger.py
```diff
@@ -595,14 +606,30 @@ def __init__(self, filename: str):
...
-    def commit_seqno(self):
-        try:
-            return int(self._filename.split("committed_")[1])
-        except IndexError:
-            # Snapshot is not yet committed
-            return None
+        entry_start_pos = super()._read_header()
+
+        # Snapshots embed evidence receipt since 2.x
+        if self.is_committed() and not self.is_snapshot_file_1_x():
```
jumaffre, line 613: We can now verify the integrity of a standalone snapshot from Python directly. Note that the top-level endorsement isn't verified. Should we add an argument here (and `read_ledger.py`) to specify the service CA to verify the full chain of trust?

src/host/test/ledger.cpp
```diff
@@ -1036,7 +1026,201 @@ TEST_CASE("Delete committed file from main directory")
   }
 }
 
-TEST_CASE("Find latest snapshot with corresponding ledger chunk")
+TEST_CASE("Generate and commit snapshots" * doctest::test_suite("snapshot"))
```
jumaffre, line 1029: These are better unit tests of the host snapshotting logic (handling of files, etc.). In particular, this tests that 1.x snapshots can still be read by the 2.x hosts, etc. (also verified by the end-to-end `lts_compatibility` test).

src/node/snapshot_serialise.h
```diff
@@ -0,0 +1,187 @@
...
+#include "kv/serialised_entry_format.h"
+#include "node/nodes.h"
+#include "node/service.h"
+#include "node/tx_receipt.h"
+
+#include <nlohmann/json.hpp>
+
+namespace ccf
+{
+  struct StartupSnapshotInfo
```
jumaffre, line 18: `StartupSnapshotInfo` used to be in `node_state.h` and I moved it to this new file which handles snapshot serialisation/deserialisation and receipt verification.

src/node/snapshot_serialise.h
```diff
@@ -0,0 +1,187 @@
...
+      store, snapshot, store->current_version(), evidence_seqno);
+  }
+
+  static std::vector<uint8_t> build_and_serialise_receipt(
+    const std::vector<uint8_t>& s,
+    const std::vector<uint8_t>& t,
+    const NodeId& node_id,
+    kv::Version seqno)
+  {
+    ccf::MerkleTreeHistory tree(t);
```
jumaffre, line 173: Note: The JSON receipt for the snapshot evidence appended to the end of the snapshot file once the evidence is committed.

src/node/snapshotter.h
```diff
@@ -217,6 +225,41 @@ namespace ccf
...
+      for (auto& pending_snapshot : pending_snapshots)
+      {
+        if (
+          pending_snapshot.evidence_idx < idx &&
+          !pending_snapshot.tree.has_value())
+        {
+          pending_snapshot.tree = tree;
+        }
+      }
+    }
```
jumaffre, line 261: Through these two functions that are called from a local hook on the signature and serialised tree tables, the snaphotter records the signature and proof for yet-to-be-committed snapshots.

src/node/snapshotter.h
```diff
@@ -201,12 +205,16 @@ namespace ccf
       next_snapshot_indices.push_back(last_snapshot_idx);
     }
 
+    // TODO: At first, it seemed that this could be merged with record_signature
+    // but record_signature is called from a hook and doesn't have a return
+    // value to tell the consensus to create a new ledger chunk...
     bool record_committable(consensus::Index idx)
```
jumaffre, line 211: I thought this function could trivially be merged with `record_signature()` but it's not since this one returns a `bool` value which indicates whether a new ledger chunk should be created. However, `record_signature()` is called from a generic hook which don't return anything.

python/ccf/ledger.py
achamayou: That would be nice, yes!

tests/infra/runner.py
```diff
@@ -280,6 +280,10 @@ def run(self, max_concurrent=None):
         if max_concurrent is None:
             max_concurrent = len(self.threads)
 
+        if max_concurrent == 0:
+            LOG.warning("No tests to run")
+            return
+
```
achamayou, line 286:
```suggestion
        assert max_concurrent
```

This is always an error, I see no legitimate use for running no tests at the same time. Alternatively, change the condition on :280 to 'if not max_concurrent' so that 0 becomes equivalent to None and means "all of them at once".

jumaffre: Good spot! I didn't realise this was still including in this PR.

python/ccf/ledger.py
jumaffre: Cool. I'll do that in a follow-up PR.

src/node/tx_receipt.h
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
```
jumaffre, line 3: Moving this out of `historical_queries_interface.h`.

.daily_canary
```diff
@@ -1 +1 @@
-Run the daily CI please.
\ No newline at end of file
+Quack Quack Quack
```
achamayou, line 1: If it quacks like a quanary...

doc/operations/ledger_snapshot.rst
```diff
@@ -65,11 +65,9 @@ Join/Recover From Snapshot
...
 
-To validate the snapshot a node is added from, the node first replays the transactions in the ledger following the snapshot until the proof that the snapshot was committed by the service to join is found. This process requires operators to copy the ledger suffix to the node's ledger directory. The validation procedure is generally quick and the node will automatically join the service once the snapshot has been validated. On recovery, the snapshot is automatically verified as part of the usual ledger recovery procedure.
+From 2.x releases, committed snapshot files embed the receipt of the evidence transaction. As such, nodes can join or recover a service from a standalone snapshot file. For 1.x releases, it is expected that operators also copy the ledger suffix containing the proof of commit of the evidence transaction to the node's ledger directory.
```
achamayou, line 68: May be worth clarifying that this only applies after -dev5 for pre-releases.

include/ccf/receipt.h
```diff
@@ -30,4 +31,25 @@ namespace ccf
...
+    for (auto const& element : receipt.proof)
+    {
+      if (element.left.has_value())
+      {
+        crypto::Sha256Hash left = element.left.value();
+        current = crypto::Sha256Hash(left, current);
+      }
+      else
+      {
+        assert(element.right.has_value());
```
achamayou, line 47: Perhaps assert above that right has no value when left does, if we are being defensive.

python/ccf/ledger.py
```diff
@@ -222,16 +225,25 @@ def get_seqno(self) -> int:
         return self._version
 
 
-def _byte_read_safe(file, num_of_bytes):
+def _byte_read_safe(file, num_of_bytes=None):
```
achamayou, line 228: It may be cleaner to default to 0, in which case the branch is simpler. None is a useful default for mutable types, but for immutable types, it's not necessarily better than a good sentinel.

src/enclave/interface.h
```diff
@@ -49,7 +49,8 @@ struct CCFConfig
 
   // Only if joining or recovering
   std::vector<uint8_t> startup_snapshot;
-  size_t startup_snapshot_evidence_seqno;
+  std::optional<size_t> startup_snapshot_evidence_seqno =
```
achamayou, line 52:
```suggestion
  std::optional<size_t> startup_snapshot_evidence_seqno_for_1_x =
```

src/host/snapshots.h
```diff
@@ -20,68 +21,83 @@ namespace asynchost
...
-      // Snapshot has no evidence idx
-      return std::nullopt;
+      throw std::logic_error(fmt::format(
+        "Snapshot file name {} does not contain snapshot seqno", file_name));
     }
 
-    auto evidence_proof_pos = file_name.find_last_of(snapshot_idx_delimiter);
-    if (evidence_proof_pos == std::string::npos)
+    // stol: unconverted suffix is ignored
+    return std::stol(file_name.substr(pos + 1));
```
achamayou, line 73: Same as above.

src/host/snapshots.h
```diff
@@ -20,68 +21,83 @@ namespace asynchost
...
+    if (evidence_idx_pos == std::string::npos)
     {
-      return std::nullopt;
+      throw std::logic_error(fmt::format(
+        "Snapshot file \"{}\" does not contain evidence index", file_name));
     }
 
-    return std::make_pair(evidence_idx, evidence_commit_idx);
+    // stol: unconverted suffix is ignored
+    return std::stol(file_name.substr(evidence_idx_pos + 1));
```
achamayou, line 100: Likewise.

src/node/hooks.h
```diff
@@ -119,4 +120,49 @@ namespace ccf
     }
   };
 
+  // Note: The SignaturesHook and SerialisedMerkleTreeHook are separate because
+  // the signature and the Merkle tree are recorded in distinct tables (for
+  // serialisation performance reason). However here, they are expected to
+  // always be called one after the other and for the same version.
```
achamayou, line 126: Perhaps clarify that this is to do with the sorting order of the table names?

src/node/node_state.h
```diff
@@ -198,7 +160,7 @@ namespace ccf
 #endif
     }
 
-    void initialise_startup_snapshot(bool recovery = false)
+    bool initialise_startup_snapshot(bool recovery = false)
```
achamayou, line 163: May be worth documenting what the boolean is, it's no intuitive that true indicates ledger verification is needed but false doesn't. An enum might be a helpful readability investment here.

src/node/rpc/test/proposal_id_test.cpp
```diff
@@ -120,14 +120,14 @@ class NullTxHistoryWithOverride : public ccf::NullTxHistory
...
-        crypto::Sha256Hash(std::to_string(version)),
+        crypto::Sha256Hash(CBuffer(std::to_string(version))),
         term_of_next_version};
     }
     else
     {
       return {
         {term_of_last_version, version},
-        crypto::Sha256Hash(std::to_string(version)),
+        crypto::Sha256Hash(CBuffer(std::to_string(version))),
```
achamayou, line 130: I think I would have added a const std::string& variant of the ctor rather than use CBuffer, but ymmv,

src/node/snapshot_serialise.h
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    bool requires_ledger_verification() const
+    {
+      // Snapshot evidence seqno is only set by the host for 1.x snapshots
+      // whose evidence need to be verified in the ledger suffix on startup
+      return evidence_seqno.has_value();
+    }
+
+    ~StartupSnapshotInfo()
```
achamayou, line 58: I'm somewhat confused by this. StartupSnapshotInfo doesn't take ownership of the vector but still clears it on destruction? It seems like it'd be a lot clearer for raw be at the end of a unique ptr that's moved in when this is constructed.

src/node/snapshot_serialise.h
```diff
@@ -0,0 +1,168 @@
...
+    {
+      raw.clear();
+      raw.shrink_to_fit();
+    }
+  };
+
+  static void deserialise_snapshot(
```
achamayou, line 65: Since this contains a deserialisation, perhaps a rename to snapshot_serdes or something would be helpful.

src/node/test/snapshotter.cpp
```diff
@@ -66,134 +103,100 @@ TEST_CASE("Regular snapshotting")
...
-  REQUIRE(snapshotter->record_committable(snapshot_tx_interval));
-  REQUIRE(snapshotter->record_committable(snapshot_tx_interval * 2));
-  REQUIRE(
-    snapshotter->record_committable(snapshot_tx_interval * interval_count));
-
   size_t commit_idx = 0;
-  INFO("Generate snapshot before first snapshot idx has no effect");
+  size_t snapshot_idx = snapshot_tx_interval;
+
+  INFO("Generate snapshot before interval as no effect");
```
achamayou, line 115:
```suggestion
  INFO("Generate snapshot before interval has no effect");
```

presumably?

python/ccf/ledger.py
```diff
@@ -595,14 +606,29 @@ def __init__(self, filename: str):
...
+        if self.is_committed() and not self.is_snapshot_file_1_x():
+            receipt_pos = entry_start_pos + self._header.size
+            receipt_bytes = _peek(self._file, num_bytes=None, pos=receipt_pos)
+
+            receipt = json.loads(receipt_bytes.decode("utf-8"))
+            root = ccf.receipt.root(receipt["leaf"], receipt["proof"])
+            node_cert = load_pem_x509_certificate(
+                receipt["cert"].encode(), default_backend()
+            )
+            ccf.receipt.verify(root, receipt["signature"], node_cert)
```
eddyashton, line 622: I may be missing something here, but it looks like we're claiming to verify this snapshot but not actually checking that the receipt is linked to the snapshot? Is there a link here that I'm missing, or would this pass if we embedded an arbitrary (valid!) receipt in the snapshot file? Is there somewhere else that the Python infra confirms that this receipt covers the evidence of this snapshot?

src/host/snapshots.h
```diff
@@ -20,68 +21,83 @@ namespace asynchost
...
+    return file_name.find(snapshot_committed_suffix) != std::string::npos;
+  }
+
+  std::optional<size_t> get_evidence_commit_idx_from_file_name(
+    const std::string& file_name)
+  {
+    // Only returns an evidence commit index for 1.x committed snapshots.
+    // 1.x committed snapshots file names are of the form:
+    // "snapshot_X_Y.committed_Z" while 2.x+ ones are of the form:
+    // "snapshot_X_Y.committed"
```
eddyashton, line 40: I think this is worth unit testing in isolation, and this can also test the `stoi` overrun mentioned below (eg `snapshot_1_2.committed_3helloimjunk`)

src/host/test/ledger.cpp
```diff
@@ -1036,7 +1026,201 @@ TEST_CASE("Delete committed file from main directory")
...
+  return std::nullopt;
+}
+
+TEST_CASE(
+  "Backwards compatibility with 1.x snapshots" *
+  doctest::test_suite("snapshot"))
+{
+  // To be removed as part of https://github.com/microsoft/CCF/issues/2981
+  fs::remove_all(ledger_dir);
+  fs::remove_all(snapshot_dir);
```
eddyashton, line 1130: This pattern of removing old dirs at the start of tests seems fragile. If we forget to do it in a new test case, it may work until the tests run in the wrong order. Can we instead ensure these are deleted with a RAII wrapper (ie `class FolderDeleter{ FolderDeleter(name); ~FolderDeleter() { fs::remove_all(name); } };`, `auto ledger_dir_deleter = std::make_shared<FolderDeleter>(ledger_dir);`)? If we want to keep the files around for manual inspection after a failed test, the local modification is just to not create that deleter.

src/node/snapshot_serialise.h
```diff
@@ -0,0 +1,168 @@
...
+      store, snapshot, hooks, view_history, public_only, evidence_seqno);
+    return std::make_unique<StartupSnapshotInfo>(
+      store, snapshot, store->current_version(), evidence_seqno);
+  }
+
+  static std::vector<uint8_t> build_and_serialise_receipt(
+    const std::vector<uint8_t>& s,
+    const std::vector<uint8_t>& t,
+    const NodeId& n,
+    const crypto::Pem& c,
```
eddyashton, line 155: Can we please give these fields better names?

python/ccf/ledger.py
jumaffre: Good point. This was raised but not documented in this ticket or elsewhere. This effectively requires public custom claims and will be addressed in a follow-up PR. See #3031

src/node/hooks.h
jumaffre: In this case, this isn't about the naming of the table but the fact that both tables are always written to at the same time.

src/node/snapshot_serialise.h
jumaffre: Yes. This pattern was really odd and should be fixed now.

src/node/rpc/test/proposal_id_test.cpp
jumaffre: I added that constructor in this PR, but it expects the string to be hex-encoded, which isn't the case here. Using a `version` for the hash is only seen in the context of this test, hence the hack here.

src/node/node_state.h
jumaffre: A lot of this code will be removed after 2.x so I've only added a comment to detail the `bool` here.

doc/operations/ledger_snapshot.rst
```diff
@@ -56,7 +56,7 @@ Snapshots are generated at regular intervals by the current primary node and sto
 
 To guarantee that the identity of the primary node that generated the snapshot can be verified offline, the SHA-256 digest of the snapshot (i.e. evidence) is recorded in the ``public:ccf.internal.snapshot_evidence`` table. The snapshot evidence will be signed by the primary node on the next signature transaction (see :ref:`operations/start_network:Signature Interval`).
 
-Committed snapshot files are named ``snapshot_<seqno>_<evidence_seqno>.commited_<evidence_commit_seqno>``, with ``<seqno>`` the sequence number of the state of the key-value store at which they were generated, ``<evidence_seqno>`` the sequence number at which the snapshot evidence was recorded and ``<evidence_commit_seqno>`` the sequence number at which the snapshot evidence was committed.
+Committed snapshot files are named ``snapshot_<seqno>_<evidence_seqno>.commited``, with ``<seqno>`` the sequence number of the state of the key-value store at which they were generated and ``<evidence_seqno>`` the sequence number at which the snapshot evidence was recorded.
```
achamayou, line 59:
```suggestion
Committed snapshot files are named ``snapshot_<seqno>_<evidence_seqno>.committed``, with ``<seqno>`` the sequence number of the state of the key-value store at which they were generated and ``<evidence_seqno>`` the sequence number at which the snapshot evidence was recorded.
```

PR #3000, opened 2021-09-20
Improve test documentation for contributors

### LINKED ISSUE #2996: Four test failures on clean build
**Describe the bug**
Running the tests on a clean build leads to test failures.
```
The following tests FAILED:
	 43 - modules_test (Failed)
	 51 - partitions_cft (Failed)
	 58 - vegeta_stress_cft (Failed)
	 59 - vegeta_stress_bft (Failed)
```

**To Reproduce**
```
$ cmake -DTARGET=virtual ..
$ ninja
$ TEST_ENCLAVE=virtual ./tests.sh --output-on-failure
```

**Environment information**
Ubuntu 20.04
Intel W-2155 (no SGX)

**Additional context**

Modules test: `No such file or directory: 'npm'`. Needs `npm` installed but the script doesn't install it. I installed `npm` and the test still fails: `Command '['npm', 'test']' returned non-zero exit status 1`. Running that by hand gives me the error: `no such file or directory, open '/home/rengolin/devel/msrc/CCF/build/package.json'`. I don't know enough about `npm` to even begin to guess.

Vegeta error: `No such file or directory: '/opt/vegeta/vegeta'`. Is this part of the install? I can't find such binary on the build dir after building CCF from source. The only package search result I found for "vegeta" on Ubuntu was a "tomato smashing game".

Partitions error: `iptc.ip4tc.IPTCError: can't initialize filter: b'Permission denied (you must be root)'`. [Seems like](https://github.com/ldx/python-iptables/issues/251) there's no work around. Could perhaps try to use sudo (and hope password is cached) if the uid isn't root?

rengolin:
The previous test doc didn't warn about root vs user tests and could
lead into new developers wondering about their configurations for too
long before looking into the pipeline configurations.

This updates the doc with commands that should always pass as well as
some notes on the approach.

Fixes #2996

### DIFF
--- a/doc/contribute/build_ccf.rst
+++ b/doc/contribute/build_ccf.rst
@@ -59,7 +59,19 @@ Tests can be started through the ``tests.sh`` wrapper for ``ctest``.
 .. code-block:: bash
 
     $ cd build
-    $ ./tests.sh
+    $ ./tests.sh -LE partitions
+
+The `partitions` test requires `root` access (to set up firewall rules), so it will not pass when using normal user accounts.
+To run the partitions test, please use `sudo`.
+
+.. code-block:: bash
+
+    $ sudo ./tests.sh -R partitions
+
+Be sure to run the non-root tests first, or the configuration will be created as root and break some non-root tests from altering it.
+
+.. note::
+   Use the `-j<threads>` option to speed up tests, unless you're running benchmarks, in which case it's best to run single-threaded.
 
 Although CCF's unit tests can be run through ``ctest`` directly, the end-to-end tests that start a network require some Python infrastructure. `tests.sh <https://github.com/microsoft/CCF/blob/main/tests/tests.sh>`_ will set up a virtual environment with these dependencies and activate it before running ``ctest`` (use ``-VV`` for verbose test output). Further runs will re-use that virtual environment.
 


eddyashton:
Thanks for raising this. I think it's actually better if new users have a simpler "run all the tests" command, rather than trying to document the additional options they always need. We can achieve that here by moving `partitions_test` to a separate CTest configuration, so it's enabled by an opt-in `-C partitions` rather than opt-out with `-LE` in many places. I'll open a PR with that approach for discussion.

rengolin:
I totally agree with you, but this looks to me like more than just renaming things. I think there will be changes in CMake, pipeline configuration, etc.

Of course, this isn't a big deal, so it should be fine to wait until it's fixed so that the docs are correct. But I don't feel comfortable changing the basic testing infrastructure of the project myself. :)

doc/contribute/build_ccf.rst
```diff
@@ -59,7 +59,19 @@ Tests can be started through the ``tests.sh`` wrapper for ``ctest``.
...
+To run the partitions test, please use `sudo`.
+
+.. code-block:: bash
+
+    $ sudo ./tests.sh -R partitions
+
+Be sure to run the non-root tests first, or the configuration will be created as root and break some non-root tests from altering it.
+
+.. note::
+   Use the `-j<threads>` option to speed up tests, unless you're running benchmarks, in which case it's best to run single-threaded.
```
achamayou, line 74: This definitely nitpicking, but I still take exception to the idea that we document features of 3party tooling. Documentation is expensive to produce and to maintain, I think it should cover all of CCF, and nothing but CCF.

doc/contribute/build_ccf.rst
```diff
@@ -59,7 +59,19 @@ Tests can be started through the ``tests.sh`` wrapper for ``ctest``.
 .. code-block:: bash
 
     $ cd build
-    $ ./tests.sh
+    $ ./tests.sh -LE partitions
+
+The `partitions` test requires `root` access (to set up firewall rules), so it will not pass when using normal user accounts.
```
achamayou, line 64: IMO a better way to do that is to have a check at the top of the partition test that prints this message when you try to run it. I expect nobody to find this piece of documentation before that warning.

rengolin: Well, docs don't need to be before the fact, they can also be reactive (`what did I do wrong?`).

But I prefer @eddyashton suggestion to not have that problem in the first place.

rengolin:
Dropping this since #3001 fixed the problem.

PR #3002, opened 2021-09-20
Tweak arg names and behaviour of ConcurrentRunner

eddyashton:
Giving these args long names for readability:

```
python ... --help
...
  -N, --show-only       List all sub-tests without executing (default: False)
  -R <string>, --regex <string>
                        Run sub-tests whose name includes this string (default: None)
```

<sub>(Sure it's not really a regex, but it's `-R` to match the CTest argument which _is_ a `--regex`, so I think that's what it's _called_ even if it's not what it _does_?)</sub>

Also printing just the thread name rather than the raw thread object, and filtering before printing to match CTest. Compare:

Old
```
python ... -R cpp -N
...
<Thread(js, initial)>
<Thread(cpp, initial)>
```

New
```
python ... -R cpp -N
...
cpp
```

### DIFF
--- a/tests/infra/runner.py
+++ b/tests/infra/runner.py
@@ -265,7 +265,7 @@ class ConcurrentRunner:
 
         if self.args.N:
             for thread in self.threads:
-                print(thread)
+                print(thread.name)
             return
 
         if self.args.R:
--- a/tests/infra/runner.py
+++ b/tests/infra/runner.py
@@ -234,11 +234,15 @@ class ConcurrentRunner:
         def add(parser):
             parser.add_argument(
                 "-N",
-                help="List all sub-tests",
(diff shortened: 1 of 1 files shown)


jumaffre (approved):

achamayou:
I liked the old printout :(, it was very clear what was printed, now it's not.

eddyashton:
> I liked the old printout :(, it was very clear what was printed, now it's not.

What do you think is missing that was in the old printout? If we want to decorate the sub-test names we can do so explicitly (eg, by indenting and numbering like CTest does), but it looks like the `<Thread(foo)>` decoration is just an implementation detail?

achamayou:
It tells me that these things are threads and they will run concurrently.

PR #3007, opened 2021-09-21
Fix issue with latest tag and release in progress

jumaffre:
During a release, the release tag is created first, but the release itself isn't created until the CI has successfully run. This means that we should ignore existing tags that haven't yet got a release (like we already do for the `lts_compatibility` test).

### DIFF
--- a/tests/infra/github.py
+++ b/tests/infra/github.py
@@ -112,8 +112,25 @@ class Repository:
             if "heads/release" in branch
         ]
 
+    def _filter_released_tags(self, tags):
+        # From a list of tags ordered by semver (latest first), filter out the ones
+        # that don't have a release yet
+        first_release_tag_idx = -1
+        for i, t in enumerate(tags):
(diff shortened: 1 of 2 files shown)


tests/infra/github.py
```diff
@@ -112,8 +112,25 @@ def __init__(self):
...
+            if not has_release_for_tag_name(t):
+                LOG.debug(f"No release available for tag {t}")
+                first_release_tag_idx = i
+            else:
+                break
+
+        return tags[first_release_tag_idx + 1 :]
+
     def get_latest_dev_tag(self):
-        return self.tags[-1]
```
achamayou, line 116: Is this called once per CI? If so, it sounds like another API throttling issue waiting to happen.

jumaffre: We no longer call the GitHub API (at all) but rely on the local git checkout instead.

PR #3008, opened 2021-09-21
Simplify objects initialisation in `node_state.h`

jumaffre:
As touched on #2991, most objects (e.g. `history`, `encryptor`) were initialised in multiple places in `node_state.h` because the node ID (`self`) wasn't known until the node had joined or recovered the service. This is no longer the case since node IDs are now derived from the node's public key, so attempting to simplify this.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Quack Quack Quack
+Run the daily CI please.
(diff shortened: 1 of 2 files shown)


achamayou (approved):

achamayou:
I think it would be tempting to add a way to check these are only called once.

eddyashton (approved):

Issue #2608, opened 2021-05-20
Python infra: add ability to spawn containers


jumaffre:
The Python infra can currently spawn nodes locally or a remote host (via ssh). However, we cannot currently spawn a node in a specific container (either locally or remotely). This has the following drawbacks:
1.  The end-to-end performance environment doesn't use Docker and is costly to maintain and upgrade.
2. We cannot use network partitioning/emulation tools (https://github.com/alexei-led/pumba) out of the box in the CI.
3. We cannot test compatibility across versions with their actual dependencies, only the ones available in the CI container at the time the compatibility test is run (see #2595).

We've dealt with 1. for a long time. For 2., we built #2553 to test for network partitions. However, 3. means that we won't exactly test for compatibility properly. So now has probably come the time to do this.

Possibly useful resources:
- https://docker-py.readthedocs.io/en/stable/containers.html
- https://itnext.io/docker-in-docker-521958d34efd

achamayou:
I don't think using iptables via Python like we do is a bad thing and something that could usefully be replaced by pumba. It's very deterministic and easy to use in tests, in combination with other tools, and it ought to work regardless of whether the nodes are started in containers or not.

Using pumba to test throttling or corruption is good, but is complementary to our current approach in my view.

jumaffre:
@achamayou Good point. I've re-phrased that bullet point.

jumaffre:
This was mostly done in #2900, even though some tweaks will be required to get it working for nodes started on remote machines. There is also an outstanding item left to get the `lts_compatibility` test working with this: the JWT issuer server should also be started inside a container, attached to the same network, so that CCF nodes can communicate with it for JWT issuer auto-refresh.

achamayou:
This is essentially complete for our current purposes, we will re-open targeted issues for further improvements as necessary.

PR #3009, opened 2021-09-21
Fix CodeQL job

jumaffre:
#2608 broke the CodeQL GitHub Actions job in which we run the `ccf-dev.yml` directly on the CI agent. However, `moby-buildx` is already installed there and the install of `docker-ce-cli` fails there. We now skip this install step if this is the case. 

Edit: Also fixes the broken container pipeline by making sure that docker CLI is installed of `ccf-app-ci` too.


.github/workflows/codeql-analysis.yml
```diff
@@ -8,6 +8,8 @@ name: "CodeQL"
 on:
   push:
     branches: [main]
+  pull_request:
+    branches: [main]
```
jumaffre, line 12: To revert before merge.

jumaffre:
```suggestion
```

eddyashton (approved):

PR #3018, opened 2021-09-22
Bump daily timeout

achamayou:
lts_compatibility was getting close to 300, and now exceeds it because receipt verification has become more expensive as a result of #2991.

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -1214,7 +1214,7 @@ def test_receipts(network, args):
 @reqs.description("Validate random receipts")
 @reqs.supports_methods("receipt", "log/private")
 @reqs.at_least_n_nodes(2)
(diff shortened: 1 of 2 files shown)


tests/e2e_logging.py
```diff
@@ -1214,7 +1214,7 @@ def test_receipts(network, args):
 @reqs.description("Validate random receipts")
 @reqs.supports_methods("receipt", "log/private")
 @reqs.at_least_n_nodes(2)
-def test_random_receipts(network, args, lts=False):
+def test_random_receipts(network, args, lts=True):
```
achamayou, line 1217: What this is doing is defaulting to _not_ checking network endorsement, so that the full_test_suite doesn't, and setting the logging test _only_ to do so, because it doesn't perform a recovery.

PR #3024, opened 2021-09-23
add small doc amendment + change run.sh in setup_vm

ThomasPluck:
Added a clear instruction to docs to upgrade the Azure CC VM to Ubuntu 20.04 using `do-release-upgrade`.

Changed `run.sh` to install `python3-pip` and then `ansible-base` via pip. This is because `ansible-base` can no longer be found in Ubuntu 20's `ppa:ansible/ansible` apt repository.

### DIFF
--- a/doc/build_apps/create_vm.rst
+++ b/doc/build_apps/create_vm.rst
@@ -4,8 +4,9 @@ Create Azure SGX VM
 :term:`Azure Confidential Compute` (ACC) offers SGX-enabled `DC`_-series VMs, which can be deployed like other Azure SKUs.
 Note that `DC`_ SKUs can only be configured as `Gen2`_ VMs, and so only `Gen2`_ compatible VM images can be used.
 
-.. note:: If you use `Visual Studio Code <https://code.visualstudio.com/>`_, you can install and set up the `Remote - SSH <https://code.visualstudio.com/docs/remote/ssh-tutorial>`_ extension to connect to your SGX-enabled VM.
+Your ACC VM must be deployed with Ubuntu 16.04 or 18.04 and then upgraded to Ubuntu 20.04 using `do-release-upgrade` before CCF can be installed.
 
+.. note:: If you use `Visual Studio Code <https://code.visualstudio.com/>`_, you can install and set up the `Remote - SSH <https://code.visualstudio.com/docs/remote/ssh-tutorial>`_ extension to connect to your SGX-enabled VM.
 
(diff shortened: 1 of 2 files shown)


doc/build_apps/create_vm.rst
```diff
@@ -4,8 +4,9 @@ Create Azure SGX VM
 :term:`Azure Confidential Compute` (ACC) offers SGX-enabled `DC`_-series VMs, which can be deployed like other Azure SKUs.
 Note that `DC`_ SKUs can only be configured as `Gen2`_ VMs, and so only `Gen2`_ compatible VM images can be used.
 
-.. note:: If you use `Visual Studio Code <https://code.visualstudio.com/>`_, you can install and set up the `Remote - SSH <https://code.visualstudio.com/docs/remote/ssh-tutorial>`_ extension to connect to your SGX-enabled VM.
+Your ACC VM must be deployed with Ubuntu 16.04 or 18.04 and then upgraded to Ubuntu 20.04 using `do-release-upgrade` before CCF can be installed.
```
achamayou, line 7: That's not correct, setting up a VM with Ubuntu 20.04 out of the box works fine, this is the setup everyone working on the project and the CI executors use.

getting_started/setup_vm/run.sh
```diff
@@ -8,7 +8,6 @@ set -ex
 # is broken on Ubuntu 20.04 with the default apt package.
 # See https://github.com/ansible/ansible/issues/68536 (fixed in ansible >= 2.10)
 sudo apt-get update
-sudo apt install software-properties-common
-sudo add-apt-repository -y --update ppa:ansible/ansible
-sudo apt install ansible-base -y
-ansible-playbook "$@"
\ No newline at end of file
+sudo apt install -y python3-pip
```
achamayou, line 11: This will not work on Ubuntu 20.04, which is our supported target platform.

achamayou:
Starting a 16 or 18.04 VM and upgrading is fine, but is not required.

ansible-base is in the 20.04 PPA, as far as I can tell: ansible-base is already the newest version (2.10.12-1ppa~focal).

doc/build_apps/create_vm.rst
ThomasPluck: This wasn't my experience whilst creating an Azure CC VM - I have the choice of either Ubuntu 16/18 (and that is all).

achamayou: Thank you for pointing out that this flow is unnecessarily restricted, I have reported it to the right person, and it will be addressed soon. Using the generic VM creation flow should work fine.

Issue #2971, opened 2021-09-10
Add a get_metrics_v1 interface


achamayou:
Some users may want to define their own metrics endpoint, for example with custom auth policies. To facilitate them doing so while retaining a stable API, we should define an explicit public interface for metrics in the form of `get_metrics_v1`.

PR #3025, opened 2021-09-24
Get metrics v1

achamayou:
Address #2971.

This is exposing a public and versioned way to get metrics, for the benefit of applications that want to customise or version themselves the output.


wintersteiger (approved):
LGTM!

eddyashton (approved):

PR #3027, opened 2021-09-28
V8 pipeline publishing artifact

rengolin:
Creating a script to build V8 and a pipeline config to run that,
publishing an artifact with the monolith version (ready for embedding).

The pipeline has two build options:
 * VERSION (mandatory): ex. 9.4.146.17
 * PUSH_ARTIFACT: true|false

We only need to publish the artifact once, but we can continue to build
making sure the build still works with the expected version, which it
won't if the version is updated upstream.

The version of the artifact needs to be changed due to Universal
Packages' constraints. It can only accept three sections, but with the
four from V8 and one from our builds, we'd have five.

Given that the first two sections in V8's version is the Chrome version
(ex. 9.4 -> 94), we join them. We also need to join the two last ones,
so that we can add a build ID.

The final versioning scheme is:
 * 9.4.146.17  ->  94.14617.$BuildId

This does create the problem where 14.617 and 146.17 yield the same
version on our side, but our builds of V8 will probably be few and far
between that we can worry about that when it stops us from progressing.

To download the latest 94.14617, use version "94.14617.*", or even the
latest 94, just use "94.*".

Example of a run publishing results: [#20210928.4](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=33424&view=logs&j=87e1acf2-5cf1-59b7-1e41-180861913346)
[Artifact published](https://dev.azure.com/MSRC-CCF/CCF/_packaging?_a=package&feed=V8&package=ccf-v8-94.14617.33424.tar.xz&protocolType=UPack&version=94.14617.33424)

Example of a run not publishing results: [#20210928.5](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=33425&view=logs&j=87e1acf2-5cf1-59b7-1e41-180861913346)

### DIFF
--- a/scripts/build-v8.sh
+++ b/scripts/build-v8.sh
@@ -5,7 +5,7 @@
 SYNTAX="build-v8.sh <version (ex. 9.4.146.17)> [publish (true|false)]"
 if [ "$1" == "" ]; then
   echo "ERROR: Missing expected argument 'version'"
-  echo $SYNTAX
+  echo "$SYNTAX"
   exit 1
 fi
 EXPECTED_VERSION="$1"
@@ -16,7 +16,7 @@ if [ "$2" != "" ]; then
     PUBLISH="$2"
   elif [ "$2" != "false" ]; then
     echo "ERROR: Publis can only be 'true' or 'false'"
-    echo $SYNTAX
+    echo "$SYNTAX"
     exit 1
   fi
 fi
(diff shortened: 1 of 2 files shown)


rengolin:
No idea what this is:
```
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
```

scripts/build-v8.sh
```diff
@@ -0,0 +1,98 @@
...
+  exit 1
+fi
+EXPECTED_VERSION="$1"
+MAJOR_VERSION="${1%.*.*}"
+PUBLISH=false
+if [ "$2" != "" ]; then
+  if [ "$2" == "true" ]; then
+    PUBLISH="$2"
+  elif [ "$2" != "false" ]; then
+    echo "ERROR: Publis can only be 'true' or 'false'"
```
achamayou, line 18:
```suggestion
    echo "ERROR: Publish can only be 'true' or 'false'"
```

achamayou (approved):

achamayou:
@rengolin could you run `./scripts/ci-checks.sh -f`? That should take care of the yaml formatting.

PR #3032, opened 2021-09-29
Add perf test of historical query fetching

eddyashton:
I've been running some local perf tests of the historical query performance, this adds it as an automated test for the CI. There are a few easy improvements to be had (increasing the page size), but I'll add those to this PR later to confirm that the CIMetrics entry works as expected.

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -692,14 +692,12 @@ def test_historical_query_range(network, args):
     first_seqno = None
     last_seqno = None
 
-    jwt = add_jwt(network)
-
     def get_all_entries(target_id, from_seqno=None, to_seqno=None):
         LOG.info(
             f"Getting historical entries{f' from {from_seqno}' if from_seqno is not None else ''}{f' to {last_seqno}' if to_seqno is not None else ''} for id {target_id}"
         )
         logs = []
-        with primary.client() as c:
+        with primary.client("user0") as c:
             timeout = 800
             start_time = time.time()
             end_time = start_time + timeout
@@ -710,7 +708,7 @@ def test_historical_query_range(network, args):
             if to_seqno is not None:
                 path += f"&to_seqno={to_seqno}"
             while time.time() < end_time:
-                r = c.get(path, log_capture=logs, headers={"authorization": f"Bearer {jwt}"})
+                r = c.get(path, log_capture=logs)
                 if r.status_code == http.HTTPStatus.OK:
                     j_body = r.body.json()
                     entries += j_body["entries"]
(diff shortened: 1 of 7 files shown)


tests/e2e_logging.py
```diff
@@ -46,6 +46,27 @@ def verify_receipt(receipt, network_cert, check_endorsement=True):
     ccf.receipt.verify(root, receipt["signature"], node_cert)
 
 
+def add_jwt(network):
```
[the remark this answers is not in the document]
eddyashton: This does use some of the existing infra, it's just moving the functionality that was previously in `test_multi_auth` into a standalone function so it can be re-used to add JWT-auth in other tests. I'm not actually using that now, but was briefly. I'll see if we can do the equivalent (create single token, don't need a server) with `jwt_issuer.py`.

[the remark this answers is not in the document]
eddyashton: Great catch - this entire thing can be replaced by 3 lines:

```python
jwt_issuer = infra.jwt_issuer.JwtIssuer()
jwt_issuer.register(network)
jwt = jwt_issuer.issue_jwt()
```

tests/historical_query_perf.py
```diff
@@ -0,0 +1,119 @@
...
+    id_c = 4
+
+    id_pattern = [id_a, id_a, id_a, id_b, id_b, id_c]
+
+    n_entries = 3001
+
+    primary, _ = network.find_primary()
+    with primary.client("user0") as c:
+        # Submit many transactions, overwriting the same IDs
+        msgs = dict()
```
achamayou, line 30:
```suggestion
        msgs = {}
```

tests/e2e_logging.py
```diff
@@ -674,6 +657,49 @@ def test_historical_receipts(network, args):
...
+    start_time = time.time()
+    end_time = start_time + timeout
+    entries = []
+    path = f"/app/log/private/historical/range?id={target_id}"
+    if from_seqno is not None:
+        path += f"&from_seqno={from_seqno}"
+    if to_seqno is not None:
+        path += f"&to_seqno={to_seqno}"
+    while time.time() < end_time:
+        r = client.get(path, log_capture=logs)
```
achamayou, line 675: Can we make this use JWT, so it's closer to what we generally are trying to replicate here?

achamayou (approved):

tests/e2e_logging.py
eddyashton: Yes. Since we already take the client as an argument, I've added a `common_headers` arg to `CCFClient` to have add the same headers to every request (ie - to use bearer auth), so we can create a JWT-auth'd client that's indistinguishable from a cert-authed at calling points like this.

PR #3041, opened 2021-10-01
Historical query perf improvements: Replace signature verification with GCM header check

eddyashton:
Signature verification is only required to confirm provenance. To confirm integrity, it is sufficient to validate the existing GCM header. Providence is not a meaningful property for nodes to confirm, instead they can return receipts which allow callers to check signatures + provenance if they wish.

An additional check which we didn't have to do previously but do now is explicitly confirming that each entry is a precursor to the current state, by checking its (integrity-protected) TxID matches the consensus' view history.

TODO:
- [x] The flow inside `historical_queries.h` is now over-complicated, since it only trusts each store once it has seen the following signature. This should be fixed.
- [x] We can go even faster by avoiding the signature fetch when the receipt is not requested. This requires some API change to indicate a request where the receipt is optional.

### DIFF
--- a/src/node/historical_queries.h
+++ b/src/node/historical_queries.h
@@ -100,8 +100,10 @@ namespace ccf::historical
       std::vector<StoreDetailsPtr> requested_stores;
       std::chrono::milliseconds time_to_expiry;
 
+      bool include_receipts;
+
       // Entries from outside the requested range (such as the next signature)
-      // may be needed to trust this range. They are stored here, distinct from
+      // may be needed to produce receipts. They are stored here, distinct from
       // user-requested stores.
       std::optional<std::pair<ccf::SeqNo, StoreDetailsPtr>>
         supporting_signature;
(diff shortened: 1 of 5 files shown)


achamayou:
@eddyashton both TODOs are useful, but are best done in separate PRs, because I suspect they will yield little benefit in the pathologically signature-dense cases we're primarily looking to improve at the moment.

eddyashton:
> @eddyashton both TODOs are useful, but are best done in separate PRs, because I suspect they will yield little benefit in the pathologically signature-dense cases we're primarily looking to improve at the moment.

@achamayou I've done both of these TODOs already. The API change was actually easy; we already have a distinction at the calling point of whether the caller was looking for a receipt or not - did they call `get_state` (includes receipt) or `get_store` (doesn't). So we just track that through the request's lifetime, and if we're not looking for receipts we don't continue with the signature fetching step.

jumaffre:
image
It looks like the new metric needs a `^` somewhere ;)

src/node/historical_queries.h
```diff
@@ -386,47 +342,6 @@ namespace ccf::historical
       }
     }
 
-    // Returns true if this is a valid signature that passes our verification
-    // checks
-    bool verify_signature(const StorePtr& sig_store, ccf::SeqNo sig_seqno)
```
jumaffre, line 391: Correct me if I'm wrong but I think we can actually delete retired nodes from the store now, as historical queries were the only reason we couldn't complete this work before (see #2474).

achamayou: That's correct, but best done in a separate PR.

achamayou (approved):

achamayou:
I think it'd be really good to add an end to end test along the following lines:

1. Start a three nodes network.
2. Isolate the primary.
3. Run transactions against both the old primary and the new, enough to produce a chunk on each side.
4. Copy the chunk from the old primary to one of the nodes in the partition
5. Attempt to run a historical query on that node in the range covered by the chunk, and confirm that it fails

PR #3042, opened 2021-10-01
Fetch and improve V8 build

rengolin:
Adding a new script to fetch the pipeline artifact for using V8 into
CCF.

Mistakes on artifact naming have made it hard to get the right package.
When implementing the fetch script I realised the problem and have now
fixed in all three places: build-v8, pipeline, fetch-v8

Also, build now fetches the tag "$VERSION" instead of relying the
branch-head to be on the same version and failing.

A future addition would be to pass less info (ex. 9.4) and get the
script to fetch the head and figure out the version on its own.

### DIFF
--- a/.azure-pipelines-v8.yml
+++ b/.azure-pipelines-v8.yml
@@ -18,9 +18,11 @@ jobs:
           sudo apt update
           sudo apt install -y pkg-config
         displayName: "Install dependencies"
+        continueOnError: false
 
       - script: scripts/build-v8.sh $(VERSION) $(PUSH_ARTIFACT)
         displayName: "Build V8"
+        continueOnError: false
 
       - script: |
           # Universal Packages versions only take three sections X.Y.Z.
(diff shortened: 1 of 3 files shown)


rengolin:
The [current CI loop](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=33666&view=logs&j=87e1acf2-5cf1-59b7-1e41-180861913346&t=be17585a-e119-57be-4c65-6490d81b7d4d) is broken because of the previous check for the version, which is fixed in this PR.

.azure-pipelines-v8.yml
```diff
@@ -18,9 +18,11 @@ jobs:
           sudo apt update
           sudo apt install -y pkg-config
         displayName: "Install dependencies"
+        continueOnError: false
```
achamayou, line 21: Isn't that the default?

achamayou (approved):

.azure-pipelines-v8.yml
rengolin: apparently not, or at least it was continuing when it should have failed and I put that to make it stop.

Could it be some global config that is overriding it?

achamayou: Very weird, https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#job alleges that it defaults to false. I can't find any information about an override, if there is, we're certainly not setting it on purpose.

PR #3044, opened 2021-10-01
JS: historical range queries

letmaik:
DRAFT!

So far, I adapted the ccf-app js package and the logging sample. If anyone wants to look at that before I implement the new bindings and test everything, please go ahead. Now's the time for API discussions.

@eddyashton @achamayou

### DIFF
--- a/js/ccf-app/src/consensus.ts
+++ b/js/ccf-app/src/consensus.ts
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+
+/**
+ * TODO change this
+ *
+ * @module
+ */
+
+import { ccf } from "./global.js";
+
+/**
+ * @inheritDoc CCF.consensus.getLastCommittedTxId;
+ */
+export const getLastCommittedTxId = ccf.consensus.getLastCommittedTxId;
+
+/**
+ * @inheritDoc CCF.consensus.getStatusForTxId;
+ */
+export const getStatusForTxId = ccf.consensus.getStatusForTxId;
+
+/**
+ * @inheritDoc CCF.consensus.getViewForSeqno;
+ */
+export const getViewForSeqno = ccf.consensus.getViewForSeqno;
+
+export { TransactionStatus } from "./global";
--- a/js/ccf-app/src/global.ts
+++ b/js/ccf-app/src/global.ts
@@ -39,6 +39,7 @@ export type JsonCompatible<T> = any;
 export interface KvMap {
   has(key: ArrayBuffer): boolean;
   get(key: ArrayBuffer): ArrayBuffer | undefined;
+  getVersionOfPreviousWrite(key: ArrayBuffer): number | undefined;
   set(key: ArrayBuffer, value: ArrayBuffer): KvMap;
   delete(key: ArrayBuffer): boolean;
   clear(): void;
@@ -102,7 +103,7 @@ export interface Receipt {
  */
 export interface HistoricalState {
   /**
-   * The ID of the transaction.
+   * The ID of the transaction, formatted as '<view>.<seqno>' string.
    */
   transactionId: string;
 
@@ -110,7 +111,21 @@ export interface HistoricalState {
    * The receipt for the historic transaction.
    */
   receipt: Receipt;
+
+  /**
+   * An object that provides access to the maps of the Key-Value Store
+   * associated with the historic transaction.
+   * Fields are map names and values are {@linkcode KvMap} objects.
+   */
+  kv: KvMaps;
 }
+
+export interface TransactionId {
+  view: number;
+  seqno: number;
+}
+
+export type TransactionStatus = "Committed" | "Invalid" | "Pending" | "Unknown";
 
 /**
  * [RSA-OAEP](https://datatracker.ietf.org/doc/html/rfc8017)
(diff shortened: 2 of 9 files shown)


js/ccf-app/src/global.ts
```diff
@@ -312,6 +347,40 @@ export interface CCF {
...
+      end_seqno: number,
+      seconds_until_expiry: number
+    ): HistoricalState[] | null;
+
+    /** Drop cached states for the given handle.
+     *
+     * May be used to free up space once a historical query has been resolved,
+     * more aggressively than waiting for the requests to expire.
+     */
+    dropCachedStateRange(handle: number): void;
```
letmaik, line 382: The C++ name is `drop_request`, however I find that very confusing and it also doesn't play well with re-exporting functions in the root module of the ccf-app npm package.

js/ccf-app/src/global.ts
```diff
@@ -102,16 +103,30 @@ export interface Receipt {
...
    * The receipt for the historic transaction.
    */
   receipt: Receipt;
+
+  /**
+   * An object that provides access to the maps of the Key-Value Store
+   * associated with the historic transaction.
+   * Fields are map names and values are {@linkcode KvMap} objects.
+   */
+  kv: KvMaps;
```
letmaik, line 120: Currently, this is unused for `"mode": "historical"` endpoints since the historical kv is already in `ccf.kv`. To satisfy the API it could be populated as well, though it would be redundant. But as at least the type doesn't have to be changed to be `| undefined`.

js/ccf-app/src/global.ts
```diff
@@ -301,6 +316,26 @@ export interface CCF {
...
+     */
+    getLastCommittedTxId(): TransactionId;
+
+    /**
+     * Get the status of a transaction by ID, provided as a view+seqno pair.
+     * This is a node-local property - while it will converge on all nodes in
+     * a healthy network, it is derived from distributed state rather than
+     * distributed itself.
+     */
+    getStatusForTxId(view: number, seqno: number): TransactionStatus;
```
letmaik, line 331: This could accept `TransactionId` as well.

js/ccf-app/src/global.ts
```diff
@@ -312,6 +347,40 @@ export interface CCF {
...
+     *  - `seconds_until_expiry` seconds elapse without calling this function
+     *  - This handle is used to request a different seqno or range
+     *
+     * The range is inclusive of both start_seqno and end_seqno. If a non-empty
+     * array is returned, it will always contain the full requested range; the
+     * array will be of length (end_seqno - start_seqno + 1).
+     */
+    getStateRange(
+      handle: number,
+      start_seqno: number,
```
letmaik, line 372: TODO camelCase

js/ccf-app/src/global.ts
```diff
@@ -301,6 +316,26 @@ export interface CCF {
...
+     * This is a node-local property - while it will converge on all nodes in
+     * a healthy network, it is derived from distributed state rather than
+     * distributed itself.
+     */
+    getStatusForTxId(view: number, seqno: number): TransactionStatus;
+
+    /**
+     * Get the view associated with a given seqno, to construct a valid TxID.
+     */
+    getViewForSeqno(seqno: number): number;
```
letmaik, line 336: TODO document behaviour for invalid seqnos

js/ccf-app/src/global.ts
```diff
@@ -312,6 +348,40 @@ export interface CCF {
...
+     * until it is deleted for one of the following reasons:
+     *  - A call to {@link CCF.historical.dropCachedStateRange}
+     *  - `seconds_until_expiry` seconds elapse without calling this function
+     *  - This handle is used to request a different seqno or range
+     *
+     * The range is inclusive of both start_seqno and end_seqno. If a non-empty
+     * array is returned, it will always contain the full requested range; the
+     * array will be of length (end_seqno - start_seqno + 1).
+     */
+    getStateRange(
```
letmaik, line 371: I wasn't able to figure out from the c++ code (`get_state_range_internal` in `historical_queries.h`) what happens when unknown seqnos are passed. It seems like the request is simply passed on to the host, so the result will always be empty since it will never finish? @eddyashton

eddyashton: The historical queries system doesn't check that the passed seqnos are known/committed. It could, but it's assumed this is better handled in the app. A request for a seqno which doesn't exist yet will be passed to the host, which will `consensus::ledger_no_entry` response message from the host, and the historical queries system deleting the state for this request (`handle_no_entry`).

letmaik: Right, but it also hides other error cases, like missing ledger files. In this case, being able to log the error and return a 500 or similar to the client would probably be better.

js/ccf-app/src/consensus.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+
+/**
+ * TODO change this
```
achamayou, line 5: ?

js/ccf-app/src/global.ts
achamayou: Depending on the circumstances, a 404 is perhaps more accurate.

js/ccf-app/src/historical.ts
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Apache 2.0 License.
 
 /**
+ * TODO change this
```
achamayou, line 5: ?

PR #3051, opened 2021-10-05
LTS compat test fix, pt.94

achamayou:


jumaffre (approved):

tests/infra/jwt_issuer.py
```diff
@@ -184,6 +184,7 @@ def wait_for_refresh(self, network, kid=TEST_JWT_KID, timeout=3):
                 stored_cert = r.body.json()[kid]
                 if self.cert_pem == stored_cert:
                     return
+                time.sleep(0.1)
```
eddyashton, line 187: Commenting for posterity: This loop was occasionally timing out, because the execution of line 182 above was taking an exceedingly long time (25s in the worst case, >5s in ~20% of runs). This was tracked to the logging of the response line, which took almost the entire time, _even when the call into `loguru` was replaced with a `print`_. We've been unable to isolate the cause of this - perhaps we're somehow hogging the main thread so it's eventually interrupted for a long time, perhaps the synchronisation of `jwks` in the `OpenIDServer` is incredibly slow - but this sensible sleep seems to have muted the issue.

eddyashton (approved):

PR #3057, opened 2021-10-06
Fix `full_test_suite` node replacement

jumaffre:
The `reconfiguration.test_node_replacement` failed with the seed `SHUFFLE_SUITE_SEED=1633506379` because the test didn't take into account nodes that were previously killed.

### DIFF
--- a/tests/reconfiguration.py
+++ b/tests/reconfiguration.py
@@ -258,9 +258,7 @@ def test_node_replacement(network, args):
     assert replacement_node.node_port == node_to_replace.node_port
     assert replacement_node.rpc_port == node_to_replace.rpc_port
 
-    allowed_to_suspend_count = infra.e2e_args.max_f(args, len(network.nodes)) - len(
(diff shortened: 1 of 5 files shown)


tests/reconfiguration.py
```diff
@@ -262,14 +257,19 @@ def test_node_replacement(network, args):
     assert replacement_node.rpc_host == node_to_replace.rpc_host
     assert replacement_node.node_port == node_to_replace.node_port
     assert replacement_node.rpc_port == node_to_replace.rpc_port
+
+    allowed_to_suspend_count = infra.e2e_args.max_f(args, len(network.nodes)) - len(
+        network.get_stopped_nodes()
+    )
```
jumaffre, line 263: This is the fix: only suspend n nodes based on the `f` of the service minus the number of nodes that are already killed/stopped without having been retired.

achamayou (approved):

PR #3062, opened 2021-10-06
Disable tx verification in LTS compatibility and suite

achamayou:
Following timeouts in this part of the test, turning this off while we investigate.

### DIFF
--- a/.threading_canary
+++ b/.threading_canary
@@ -1,1 +1,1 @@
-hmm
+poke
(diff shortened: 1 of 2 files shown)


tests/e2e_suite.py
```diff
@@ -166,7 +166,7 @@ def filter_fun(x):
         elapsed -= test_elapsed
 
     if not args.dry_run:
-        network.stop_all_nodes(verbose_verification=False)
+        network.stop_all_nodes(skip_verification=True)
```
eddyashton, line 169: PR says we're disabling this in LTS compatibility, but this line also disables in suite. Deliberate?

eddyashton (approved):

PR #3081, opened 2021-10-12
Adding YCM code-complete config

rengolin:
This overrides the need for compile_commands.json and avoids crashing clangd
on Vim. VS Code should update the compile options to match what's in
this file as well as disable reading from compile_commands.json.

I'm not sure this is acceptable, but it kinda works for me on Vim/Linux. I don't know how to save the IntelliSense config for particular projects in VS Code, but I'd be happy to include in this PR if anyone knows how to.


rengolin:
> Actually I was wrong, those particular settings would be expected to be in `.vscode/c_cpp_properties.json`, and all of us would want the same settings anyways, so we could add that too.

Do you have a copy we can include in this PR? And how would that integrate if users create a new project and overwrite the `.vscode` directory?

Issue #3075, opened 2021-10-11
Raft election timeout starts ticking too early


achamayou:
Raft starts counting down to election timeout as soon as the node becomes TRUSTED, but potentially quite a long time before channels are established and entries start replicating. This can result in pointless early election attempts.

Raft should instead wait until the transaction adding to the node to the configuration becomes committable locally, at which point it is certain that if it called an election, it would preserve its own addition. Running for election prior to this is only puts liveness at risk for no benefit.

achamayou:
This was first observed in the election_test (which has a relatively low 500ms timeout), and can be triggered by sufficiently lowering the election timeout, eg. to 100ms.

PR #3087, opened 2021-10-13
Joiners should not attempt to run for election in CFT before they've observed the first config they're in as committable

achamayou:
Fix for #3075.

Note: this does not solve the problem of elections happening shortly after join, rolling back the join. The joiner needs to observe commit of that transaction, and the operator needs to re-attempt if necessary.


achamayou:
@jumaffre I didn't think of that, but it's definitely a good idea, I'll do that.

PR #3097, opened 2021-10-15
Enable 2-tx reconfiguration with CFT

wintersteiger:
Mainly passing an additional option through the various parts of the infrastructure.

The new `reconfiguration_test_cft_2tx` should probably be disabled or moved to the long/daily tests for now.

### DIFF
--- a/.azure-pipelines-templates/matrix.yml
+++ b/.azure-pipelines-templates/matrix.yml
@@ -15,7 +15,7 @@ parameters:
 
   build:
     common:
-      cmake_args: '-DCMAKE_C_COMPILER_LAUNCHER="ccache" -DCMAKE_CXX_COMPILER_LAUNCHER="ccache"'
+      cmake_args: '-DCMAKE_C_COMPILER_LAUNCHER="ccache" -DCMAKE_CXX_COMPILER_LAUNCHER="ccache" -DENABLE_2TX_RECONFIG=ON'
     NoSGX:
       cmake_args: "-DCOMPILE_TARGETS=virtual"
     SGX:
@@ -25,7 +25,7 @@ parameters:
     perf:
       cmake_args: '-DBUILD_UNIT_TESTS=OFF -DBUILD_TPCC=ON -DDISTRIBUTE_PERF_TESTS="-n local://localhost -n local://localhost"'
     release:
-      cmake_args: "-DTLS_TEST=ON -DLONG_TESTS=ON -DENABLE_BFT=OFF"
+      cmake_args: "-DTLS_TEST=ON -DLONG_TESTS=ON -DENABLE_BFT=OFF -DENABLE_2TX_RECONFIG=OFF"
 
   test:
     NoSGX:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+- Added support for 2-transaction reconfiguration with CFT consensus (#3097), see [documentation](https://microsoft.github.io/CCF/main/overview/consensus/bft.html#two-transaction-reconfiguration).
+
 ### Changed
 
 - DNS resolution of client connections is now asynchronous.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added 
+
 - Added support for 2-transaction reconfiguration with CFT consensus (#3097), see [documentation](https://microsoft.github.io/CCF/main/overview/consensus/bft.html#two-transaction-reconfiguration).
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-### Added 
+### Added
 
 - Added support for 2-transaction reconfiguration with CFT consensus (#3097), see [documentation](https://microsoft.github.io/CCF/main/overview/consensus/bft.html#two-transaction-reconfiguration).
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -699,11 +699,15 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS ${ROTATION_TEST_ARGS}
   )
 
+  set(RECONFIG_TEST_ARGS --ccf-version ${CCF_VERSION})
+  if(ENABLE_2TX_RECONFIG)
+    list(APPEND RECONFIG_TEST_ARGS --include-2tx-reconfig)
+  endif()
   add_e2e_test(
     NAME reconfiguration_test_cft
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
     CONSENSUS cft
-    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION}
+    ADDITIONAL_ARGS ${RECONFIG_TEST_ARGS}  
   )
 
   add_e2e_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -624,16 +624,20 @@ if(BUILD_TESTS)
     CONSENSUS cft
   )
 
+  # Higher snapshot interval as the test currently assumes that no
+  # transactions are emitted while partitions are up. To be removed when
+  # https://github.com/microsoft/CCF/issues/2577 is implemented
+  set(PARTITIONS_TEST_ARGS --snapshot-tx-interval 10000)
+  if(ENABLE_2TX_RECONFIG)
+    list(APPEND PARTITIONS_TEST_ARGS --include-2tx-reconfig)
+  endif()
   add_e2e_test(
     NAME partitions_cft
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/partitions_test.py
     CONSENSUS cft
     LABEL partitions
-    CONFIGURATIONS partitions
-    # Higher snapshot interval as the test currently assumes that no
-    # transactions are emitted while partitions are up. To be removed when
-    # https://github.com/microsoft/CCF/issues/2577 is implemented
-    ADDITIONAL_ARGS --snapshot-tx-interval 10000
+    CONFIGURATIONS partitions    
+    ADDITIONAL_ARGS ${PARTITIONS_TEST_ARGS}
   )
 
   if(NOT SAN)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -711,6 +711,8 @@ if(BUILD_TESTS)
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
     CONSENSUS cft
     ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx
+    LABEL partitions
+    CONFIGURATIONS partitions
   )
 
   add_e2e_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -624,8 +624,8 @@ if(BUILD_TESTS)
     CONSENSUS cft
   )
 
-  # Higher snapshot interval as the test currently assumes that no
-  # transactions are emitted while partitions are up. To be removed when
+  # Higher snapshot interval as the test currently assumes that no transactions
+  # are emitted while partitions are up. To be removed when
   # https://github.com/microsoft/CCF/issues/2577 is implemented
   set(PARTITIONS_TEST_ARGS --snapshot-tx-interval 10000)
   if(ENABLE_2TX_RECONFIG)
@@ -636,7 +636,7 @@ if(BUILD_TESTS)
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/partitions_test.py
     CONSENSUS cft
     LABEL partitions
-    CONFIGURATIONS partitions    
+    CONFIGURATIONS partitions
     ADDITIONAL_ARGS ${PARTITIONS_TEST_ARGS}
   )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -711,9 +711,7 @@ if(ENABLE_2TX_RECONFIG)
     NAME reconfiguration_test_cft_2tx
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
     CONSENSUS cft
-    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx
-    LABEL partitions
-    CONFIGURATIONS partitions
+    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx    
   )
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,7 +520,7 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS
       --oe-binary
       ${OE_BINDIR}
-      --ledger-recovery-timeout
+      --ledger-read-timeout
       20
       --test-duration
       200
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -710,7 +710,7 @@ if(BUILD_TESTS)
     NAME reconfiguration_test_cft_2tx
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
     CONSENSUS cft
-    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx    
+    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx
     CONFIGURATIONS partitions
   )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -707,7 +707,7 @@ if(BUILD_TESTS)
     NAME reconfiguration_test_cft
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
     CONSENSUS cft
-    ADDITIONAL_ARGS ${RECONFIG_TEST_ARGS}  
+    ADDITIONAL_ARGS ${RECONFIG_TEST_ARGS}
   )
 
   add_e2e_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -710,8 +710,7 @@ if(BUILD_TESTS)
     NAME reconfiguration_test_cft_2tx
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
     CONSENSUS cft
-    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx
-    LABEL partitions
+    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx    
     CONFIGURATIONS partitions
   )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -706,13 +706,16 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS --ccf-version ${CCF_VERSION}
   )
 
+if(ENABLE_2TX_RECONFIG)
   add_e2e_test(
     NAME reconfiguration_test_cft_2tx
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
     CONSENSUS cft
     ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx
+    LABEL partitions
     CONFIGURATIONS partitions
   )
+endif()
 
   add_e2e_test(
     NAME election_test
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -706,15 +706,6 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS --ccf-version ${CCF_VERSION}
   )
 
-if(ENABLE_2TX_RECONFIG)
-  add_e2e_test(
-    NAME reconfiguration_test_cft_2tx
-    PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
-    CONSENSUS cft
-    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx    
-  )
-endif()
-
   add_e2e_test(
     NAME election_test
     PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/election.py
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,7 +520,7 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS
       --oe-binary
       ${OE_BINDIR}
-      --ledger-read-timeout
+      --ledger-recovery-timeout
       20
       --test-duration
       200
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -707,6 +707,13 @@ if(BUILD_TESTS)
   )
 
   add_e2e_test(
+    NAME reconfiguration_test_cft_2tx
+    PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/reconfiguration.py
+    CONSENSUS cft
+    ADDITIONAL_ARGS --ccf-version ${CCF_VERSION} --reconfiguration-type 2tx
+  )
+
+  add_e2e_test(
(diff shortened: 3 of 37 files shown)


jumaffre:
This will need a new entry in the changelog too.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+- Added support for 2-transaction reconfiguration with CFT consensus (#3097), see [documentation](https://microsoft.github.io/CCF/main/overview/consensus/bft.html#two-transaction-reconfiguration).
```
jumaffre, line 10:
```suggestion
### Added 

- Added support for 2-transaction reconfiguration with CFT consensus (#3097), see [documentation](https://microsoft.github.io/CCF/main/overview/consensus/bft.html#two-transaction-reconfiguration).
```

src/node/node_state.h
```diff
@@ -946,7 +953,12 @@ namespace ccf
         progress_tracker->set_node_id(self);
       }
 
-      setup_consensus(ServiceStatus::OPENING, true);
+      auto service_config = tx.ro(network.config)->get();
+      auto reconfiguration_type = ReconfigurationType::ONE_TRANSACTION;
+      if (service_config->reconfiguration_type.has_value())
+        reconfiguration_type = service_config->reconfiguration_type.value();
```
jumaffre, line 959: This could be making use of `.value_or()` to be more concise.

src/node/rpc/node_frontend.h
```diff
@@ -357,6 +364,12 @@ namespace ccf
             "No service is available to accept new node.");
         }
 
+        auto config = args.tx.ro(network.config);
+        auto service_config = config->get();
+        auto reconfiguration_type = ReconfigurationType::ONE_TRANSACTION;
+        if (service_config->reconfiguration_type.has_value())
+          reconfiguration_type = service_config->reconfiguration_type.value();
```
jumaffre, line 371: This seems to be called twice in the `/node/join` endpoint. Could we only call it once and pass the `reconfiguration_type` result around instead?

src/node/rpc/node_frontend.h
```diff
@@ -441,12 +455,14 @@ namespace ccf
           rep.node_status = node_status;
           if (
             node_status == NodeStatus::TRUSTED ||
-            node_status == NodeStatus::LEARNER)
+            node_status == NodeStatus::LEARNER ||
+            node_status == NodeStatus::RETIRING)
```
jumaffre, line 459: Do we really need this? AFAICS, this is for a new node that's waiting to be trusted by members but is retired in the meantime. So it might have been briefly trusted but never participated in the consensus (otherwise it would have already joined). So maybe joining nodes should never join if they're retiring?

src/node/rpc/node_frontend.h
```diff
@@ -1340,8 +1357,6 @@ namespace ccf
 
       auto orc_handler = [this](auto& args, const nlohmann::json& params) {
         const auto in = params.get<ObservedReconfigurationCommit::In>();
-        LOG_DEBUG_FMT(
-          "ORC for configuration #{} from {}", in.reconfiguration_id, in.from);
 
         if (consensus == nullptr)
         {
```
jumaffre, line 1362: I think the consensus checks underneath are already handled by the framework (see `frontend.h`).

src/node/rpc/node_frontend.h
```diff
@@ -1377,34 +1392,41 @@ namespace ccf
...
-            if (!node_info.has_value())
-            {
-              return make_error(
-                HTTP_STATUS_INTERNAL_SERVER_ERROR,
-                ccf::errors::InternalError,
-                fmt::format("Missing node information for {}", nid));
-            }
+            return make_error(
+              HTTP_STATUS_BAD_REQUEST,
+              ccf::errors::ResharingAlreadyCompleted,
```
jumaffre, line 1403: This isn't about resharing, is it?

tests/infra/consortium.py
```diff
@@ -293,13 +295,31 @@ def retire_node(self, remote_node, node_to_retire):
 
         with remote_node.client() as c:
             r = c.get(f"/node/network/nodes/{node_to_retire.node_id}")
-            # Until we have 2tx-promotion, we expect RETIRING but not RETIRED yet.
-            expected = (
-                NodeStatus.RETIRED.value
-                if self.consensus != "bft"
-                else NodeStatus.RETIRING.value
+            status = r.body.json()["status"]
```
jumaffre, line 298: From a UX perspective, it seems a little awkward to change the check for successfully retire node proposal logic depending on the type of reconfiguration. In all cases, the node should become `RETIRED` after a relatively short timeout, sometimes passing through the `RETIRING` state?

tests/infra/network.py
```diff
@@ -635,6 +637,8 @@ def join_node(
     def trust_node(self, node, args, valid_from=None, validity_period_days=None):
         primary, _ = self.find_primary()
         try:
+            # Note: the time it takes for a learner to catch up is proportional
+            # to the ledger size; we may want to factor that into the timeout.
```
jumaffre, line 641: There's already an argument for this, `--ledger-recovery-timeout`. Perhaps we want to rename it and use it here?

tests/reconfiguration.py
```diff
@@ -419,34 +428,66 @@ def test_learner_catches_up(network, args):
...
-    f = infra.e2e_args.max_f(args, len(nodes)) + 1
-    f_backups = backups[:f]
+    f_backups = backups[: network.get_f() + 1]
 
     new_node = network.create_node("local://localhost")
     network.join_node(new_node, args.package, args, from_snapshot=False)
-    network.trust_node(new_node, args)
 
-    # No way to keep the new node suspended in learner state?
+    with network.partitioner.partition(f_backups):
```
jumaffre, line 450: As discussed, can we move this to `partititons_test.py` as this requires `sudo`.

jumaffre: Also, I don't understand why the partition of f+1 backups happens so early. This could be the reason why `new_node` cannot join the service (no commit over its join transaction). I believe it would be more pertinent to only isolate `new_node` from the rest of the service before `trust_node(new_node)` and then check that progress on the service doesn't depend on this node (e.g. by partitioning some other backups). Then, when all partitions are lifted, the node should join, observe its addition, orc and become part of consensus.

src/consensus/aft/raft.h
```diff
@@ -538,11 +546,30 @@ namespace aft
...
+        {
+          if (conf2.find(nid) == conf2.end())
+          {
+            return false;
+          }
+        }
+        return true;
+      }
+      return false;
+    }
```
jumaffre, line 565: Since, the `hostname` and `port` for a given node should always stay the same, isn't this just `conf1 == conf2`?

src/consensus/aft/raft.h
```diff
@@ -615,11 +656,14 @@ namespace aft
         }
       }
 
-      uint32_t offset = get_bft_offset(conf);
-      configurations.push_back({idx, std::move(conf), offset, 0});
+      if (!same_node_ids(conf, configurations.back().nodes))
+      {
+        uint32_t offset = get_bft_offset(conf);
```
jumaffre, line 661: Isn't this path also used in CFT, when 2tx reconfiguration is enabled?

src/consensus/aft/orc_requests.h
```diff
@@ -0,0 +1,94 @@
...
+    std::shared_ptr<ccf::NodeClient> client,
+    const ccf::NodeId& from,
+    kv::ReconfigurationId rid,
+    bool delayed = false)
+  {
+    auto msg = std::make_unique<threading::Tmsg<AsyncORCTaskMsg>>(
+      orc_cb, client, from, rid);
+
+    threading::ThreadMessaging::thread_messaging.add_task_after(
+      std::move(msg), std::chrono::milliseconds(delayed ? 2000 : 0));
```
jumaffre, line 92: As discussed, if the primary is being retired, the delay should be a function of the election timeout so that ORC aren't submitted when we know there's little certainty that a new primary has been elected.

src/consensus/aft/orc_requests.h
```diff
@@ -0,0 +1,94 @@
...
+    ObservedReconfigurationCommit::In ps = {from, rid};
+
+    http::Request request(fmt::format(
+      "/{}/{}", ccf::get_actor_prefix(ccf::ActorsType::nodes), "orc"));
+    request.set_header(
+      http::headers::CONTENT_TYPE, http::headervalues::contenttype::JSON);
+
+    auto body = serdes::pack(ps, serdes::Pack::Text);
+    request.set_body(&body);
+    return client->make_request(request);
```
jumaffre, line 43: It took me a while to wrap my head around this: this is actually sending a request to itself and relying on the forwarding mechanism in the generic frontend to send the request to the new primary. I think this is OK. However, the logic in `http_node_client.h` does not actually wait for the response to the request (it will take a while to receive it from the forwardee) and I think we should wait for this asynchronously, and also wait for commit on this transaction.

src/consensus/aft/orc_requests.h
```diff
@@ -0,0 +1,94 @@
...
+    ccf::NodeId from;
+    kv::ReconfigurationId rid;
+    size_t retries;
+  };
+
+  inline void orc_cb(std::unique_ptr<threading::Tmsg<AsyncORCTaskMsg>> msg)
+  {
+    if (!submit_orc(msg->data.client, msg->data.from, msg->data.rid))
+    {
+      if (--msg->data.retries > 0)
```
jumaffre, line 69: Why is there a limit here? If the node never sends the ORC, will the node be stuck forever?

src/node/rpc/node_frontend.h
wintersteiger: I added this because it happens during the reconfiguration tests. A node attempts to join, but is retired before it hears about it and thus can't decide whether it should submit the join RPC. So, I think it's necessary, but maybe the behaviour should be the same as for `NodeStatus::PENDING`?

src/node/rpc/node_frontend.h
wintersteiger: I added that because it's also done in other handlers in this file, e.g. `is_primary` and `get_primary`. So, I concluded that I can't rely on the consensus being set at all time. Would this particular handler not be called at all if there's no consensus object?

src/node/rpc/node_frontend.h
wintersteiger: Nope, good catch, that's a copy&paste gone wrong :)

tests/infra/consortium.py
wintersteiger: Not in the test scripts - if there are bugs, they may not retire at all, so we need a check with timeout. Also, the other test scripts rely on the fact that this happens pretty much instantaneously and try to re-use the ip/port of the retiring node, so we have to make sure the node is actually gone before we continue (or add waiting loops in the corresponding tests).

src/consensus/aft/raft.h
wintersteiger: Yes, except there's no `operator==` for `NodeInfo`. I just wanted to compare the node ids, so I opted not to add an operator. Do you think it would be better to add one?

src/consensus/aft/raft.h
wintersteiger: Yes, except when 1tx reconfiguration is used `conf` is never the same twice. With 2-tx reconfig this happens when the only change to a configuration is a node transitioning from retiring to retired (because of the update to the nodes table).

src/consensus/aft/orc_requests.h
wintersteiger: Yes, it would be stuck. The limit is `SIZE_MAX` though. In the beginning I wasn't sure whether we need/want a limit, but you're right, we should keep trying forever.

src/consensus/aft/orc_requests.h
wintersteiger: In general that is true (and I've started adding code for that in the split identity PR), but for ORCs it's not necessary to wait. The "reply" is essential the primary changing the node status in the nodes table. I believe, if necessary, all nodes will hit the same reconfiguration commit check in `raft.h` again and re-schedule new ORCs.

tests/reconfiguration.py
wintersteiger: I added the `CONFIGURATION` option in cmake scripts, so this one also runs with `sudo`.

Not sure I understand. `trust_node` runs right after the partition is introduced, I just added a `check_does_not_progress` in between for good measure. What would you have done differently?

tests/infra/network.py
wintersteiger: That's great, thanks! Maybe `ledger-read-timeout`?

src/node/rpc/node_frontend.h
wintersteiger: Move checks to function `is_taking_part_in_acking`. Add comment to explain why?

src/node/rpc/node_frontend.h
jumaffre: I believe the two endpoints you mentioned are very much consensus-specific and execute outside consensus (`.set_execute_outside_consensus(ccf::endpoints::ExecuteOutsideConsensus::Locally)`), hence the extra checks. In the ORC case, these can be skipped.

tests/infra/consortium.py
jumaffre: OK. In this case, can the node already be in `retired` state in this RPC, even with a 2tx reconfiguration (i.e. if node was fast enough to retire itself)? It feels like the logic could be unified: wait for the node to be marked as `RETIRED` within a timeout. If we're in 1tx reconfiguration, the check should pass straight away.

src/consensus/aft/raft.h
jumaffre: `raft.h` is already chunky so defining the `operator==` for `NodeInfo` is probably simpler.

CMakeLists.txt
```diff
@@ -520,7 +520,7 @@ if(BUILD_TESTS)
     ADDITIONAL_ARGS
       --oe-binary
       ${OE_BINDIR}
-      --ledger-recovery-timeout
+      --ledger-read-timeout
```
achamayou, line 523: I am not sure this rename is good, but it's definitely incomplete (uses of the flag have not been updated), and it does not belong in this PR in any case.

src/consensus/aft/raft.h
jumaffre: As per `hooks.h`, I thought that retiring nodes were marked as `std::nullopt` in the configuration, whereas retired ones were explicitly passed in the `retirees` list. 

In any case, I'm surprised this calls into BFT code (`get_bft_offset`) even though this is on the CFT path.

src/consensus/aft/orc_requests.h
```diff
@@ -0,0 +1,85 @@
...
+    ccf::NodeId from;
+    kv::ReconfigurationId rid;
+  };
+
+  inline void orc_cb(std::unique_ptr<threading::Tmsg<AsyncORCTaskMsg>> msg)
+  {
+    if (!submit_orc(msg->data.client, msg->data.from, msg->data.rid))
+    {
+      threading::ThreadMessaging::thread_messaging.add_task_after(
+        std::move(msg), std::chrono::milliseconds(250));
```
achamayou, line 69: why 250?

src/consensus/aft/raft.h
```diff
@@ -593,6 +620,20 @@ namespace aft
           }
         }
 
+        if (!new_retirees.empty())
+        {
+          LOG_DEBUG_FMT(
+            "Configurations: new retirees: {{{}}}",
```
achamayou, line 626: Would suggest "retiring nodes" since "retiring" is a state name.

src/consensus/aft/raft.h
```diff
@@ -3197,6 +3247,39 @@ namespace aft
       return r;
     }
 
+    size_t num_retired(
+      const kv::Configuration& from, const kv::Configuration& to) const
+    {
+      size_t r = 0;
+      for (const auto& [id, _] : from.nodes)
+      {
+        auto rit = retirees.find(id);
```
achamayou, line 3256: s/retirees/retiring_nodes/

achamayou:
@wintersteiger this needs to either be disabled in release, or made safe when used against a 1tx network (ie. refuse to join with a helpful error) before merging.

src/consensus/aft/orc_requests.h
wintersteiger: Because it felt right! This is a re-schedule of a failed RPC (need to keep sending ORCs forever in general), but I didn't want to reschedule without any delay to avoid liveness issues.

src/consensus/aft/raft.h
wintersteiger: The offset is there to allow BFT to run with non-consecutive node IDs; after a view-change, the next primary is the one at view-change-number + offset in the list of nodes. This has nothing to do with this PR though.

src/node/rpc/node_frontend.h
wintersteiger: That makes a lot of sense, thanks for pointing that out!

src/consensus/aft/orc_requests.h
achamayou: Ok! Can we have a constexpr ORC_RPC_RETRY_INTERVAL_MS, and use that here? I think it'd be more readable.

src/consensus/aft/raft.h
achamayou: Ok, as discussed, I think we want to remove this, but you are absolutely right that it's nothing to do with this change.

.azure-pipelines-templates/matrix.yml
```diff
@@ -25,7 +25,7 @@ parameters:
     perf:
       cmake_args: '-DBUILD_UNIT_TESTS=OFF -DBUILD_TPCC=ON -DDISTRIBUTE_PERF_TESTS="-n local://localhost -n local://localhost"'
     release:
-      cmake_args: "-DTLS_TEST=ON -DLONG_TESTS=ON -DENABLE_BFT=OFF"
+      cmake_args: "-DTLS_TEST=ON -DLONG_TESTS=ON -DENABLE_BFT=OFF -DENABLE_2TX_RECONFIG=OFF"
```
wintersteiger, line 28: @achamayou Is this over-write of the new option in release mode fine here, or should I add `...=ON` to each of the other configurations?

achamayou: This is fine, imo.

achamayou (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+
```
jumaffre, line 11: The documentation should be moved out of the BFT page.

src/consensus/aft/raft.h
```diff
@@ -3333,18 +3425,26 @@ namespace aft
...
-              next->nodes.find(state->my_node_id) != next->nodes.end())
+              node_client && !is_learner() && !is_retired() &&
+              ((next->nodes.find(state->my_node_id) != next->nodes.end()) ||
+               (is_retiring() &&
+                next->nodes.find(state->my_node_id) == next->nodes.end())))
             {
-              LOG_TRACE_FMT(
-                "Configurations: not enough trusted nodes for configuration "
-                "#{} ({} out of {}); submitting ORC",
+              schedule_submit_orc(
```
jumaffre, line 3442: You mentioned in our discussion that we stop scheduling the ORC when we see our `RETIRED` state as committed. It seems that the only condition to stop scheduling the ORC is when an ORC has been forwarded to a node (`submit_orc()` returns true in `orc_requests.h`), which isn't enough to guarantee that the ORC has successfully been applied (the request may have been lost, or the transaction rolled back). What am I missing?

src/kv/test/stub_consensus.h
```diff
@@ -149,7 +149,8 @@ namespace kv::test
     void add_configuration(
       ccf::SeqNo seqno,
       const Configuration::Nodes& conf,
-      const std::unordered_set<NodeId>& learners = {}) override
+      const std::unordered_set<NodeId>& learners = {},
```
jumaffre, line 152: `learner_nodes` for consistency?

src/node/rpc/frontend.h
```diff
@@ -109,8 +109,8 @@ namespace ccf
           }
         }
         ctx->set_error(
-          HTTP_STATUS_INTERNAL_SERVER_ERROR,
-          ccf::errors::InternalError,
+          HTTP_STATUS_SERVICE_UNAVAILABLE,
+          ccf::errors::PrimaryNotFound,
```
jumaffre, line 113: Is it worth adding this to the changelog? And also checking that this is the return status we get from the ORC when the node is retiring?

src/consensus/aft/raft.h
achamayou: We do need to keep retrying until we've observed the ORC committed or our transition complete (retired or trusted), a node having accepted (to forward) the RPC isn't enough.

wintersteiger: It will keep scheduling new ORCs on every `commit()` _until_ it becomes `RETIRED` (which also means that the entire rest of the new configuration is ready to be applied, so no more ORCs needed). But, we don't need to know whether the ORC RPC call itself succeeded - we can't do anything to correct the error anyways, all we can do is retry. Am I missing something?

src/node/rpc/node_frontend.h
```diff
@@ -1377,34 +1386,41 @@ namespace ccf
             "Configurations: sufficient number of ORCs, updating nodes in "
```
jumaffre, line 1386: I'm a little worried about the call to `consensus->orc()` on line 1383. This endpoint writes to the KV and can thus be aborted or rolled back. However, the endpoint also has some side effects on the internal state of the consensus (`orc_sets`) which does not seem to be transactional in nature (e.g. I do not see any rollback logic for it). 

I think `orc_sets` should be recorded in the KV store instead (e.g. whether it's passed the quorum or not could be recorded as a `bool` in the `configuration` table), which then passed to the consensus via a hook (which might not be necessary as we already have a hook on the configuration table).

tests/infra/consortium.py
```diff
@@ -322,11 +317,17 @@ def trust_node(
...
+                remote_node, node_id, NodeStatus.LEARNER
+            ) and not self._check_node_exists(remote_node, node_id, NodeStatus.TRUSTED):
+                raise ValueError(
+                    f"Node {node_id} does not exist in state {NodeStatus.TRUSTED} or {NodeStatus.LEARNER}"
+                )
+        elif not self._check_node_exists(remote_node, node_id, NodeStatus.TRUSTED):
             raise ValueError(
-                f"Node {node_id} does not exist in state {NodeStatus.TRUSTED} or {NodeStatus.LEARNER}"
+                f"Node {node_id} does not exist in state {NodeStatus.TRUSTED}"
             )
```
jumaffre, line 331: AIUI, the test would pass if the node is fast enough to transition from `LEARNER` to `TRUSTED` before we check its state in the store. Wouldn't it be equivalent to `self.wait_for_node_to_exist_in_store(remote_node, node_to_retire.node_id, timeout, NodeStatus.TRUSTED)` for both 1tx and 2tx reconfiguration?

wintersteiger: Yes, I think that's right, I just didn't want to run the waiting loop here as it's fine if the node is still a learner.

src/node/rpc/frontend.h
jumaffre: As discussed, the ORC client doesn't have to check the response and should keep retrying until it's seen itself as `TRUSTED` from the consensus.

src/node/rpc/node_frontend.h
jumaffre: As discussed, this should be OK as long as i) nodes retry to send ORCs until they've seen their `TRUSTED`ness committed, ii) `orc_sets` (and other consensus state updated in `consensus->orc()` are flushed when a node becomes candidate. We should also try to add a test for this (end-to-end or unit test?) but it's OK to merge this PR as it is as long as the existing reconfiguration test is stable.

src/consensus/aft/raft.h
jumaffre: https://github.com/microsoft/CCF/pull/3097#discussion_r741986320

src/consensus/aft/raft.h
```diff
@@ -730,7 +730,16 @@ namespace aft
...
     // For more info about Observed Reconfiguration Commits see
-    // https://microsoft.github.io/CCF/main/overview/consensus/bft.html#two-transaction-reconfiguration
+    // https://microsoft.github.io/CCF/main/overview/consensus/2tx-reconfig.html
+    //
+    // Note that this call is not `const` and that it modifies `orc_sets`. This
+    // is safe, despite the fact that the primary may change or a
+    // reconfiguration may be (partially) rolled back, because the `orc_sets`
+    // are cleared upon entering/exiting the leader/follower replica states.
+    // This means that we never record spurious ORCs, while it is still
+    // guaranteed that we will eventuallky receive all of them, since all
```
achamayou, line 740:
```suggestion
    // guaranteed that we will eventually receive all of them, since all
```

src/node/rpc/node_frontend.h
wintersteiger: I created #3157 for this and assigned myself, so I'll resolve the conversation here.

jumaffre (approved):

PR #3110, opened 2021-10-20
Remove unused values table

achamayou:
`internal.values` takes a final bow.


achamayou:
@jumaffre indeed, I had missed that, but I am going to suggest that #2931 could and should have a specifically named table its purposes, which doesn't need to be present at all when replicated identity is used.

Issue #3142, opened 2021-10-27
Potential bug in processing of AppendEntries which cross view changes


eddyashton:
Here's a thing that can't happen currently with an honest host, but could be done by a malicious host, and may even happen accidentally as a side effect of #3124.

Node A produces the following ledger:
```
2.1 2.2 2.3 2.4 2.5 2.6
```

For simplicity, let's say that every even seqno is a signature.

Node A rolls back to 2.4, and produces new transactions in term 3:
```
2.1 2.2 2.3 2.4
3.5 3.6 3.7 3.8
```

Node A tries to catch up a new node B. They send an AppendEntries covering the range `2.1 -> 3.8`, believing that includes all of entries they hold.

However, the actual body of the AppendEntries is constructed on the host, so it is possible that it instead contains:
```
2.1 2.2 2.3 2.4 2.5 2.6
3.7 3.8
```

The important features are that 2.5 and 2.6 are valid entries (the receiving node will accept them individually, as their GcmHeader is valid), and that this looks like a valid `2.1 -> 3.8` range. This could be from a malicious host who held the suffix rather than truncating, or an unlucky asynchronous attempt to fetch the entries. The receiver _will_ notice an error on the next signature - the merkle root produced will not match. But it's not clear how the receiver should recover in this situation, especially if that signature arrives in a future entry (ie - they can't see anything wrong with an AppendEntries covering `2.1 -> 3.7`).

We need to confirm exactly what the receiving node's behaviour is here, and see if they're able to correctly rollback and recover if they later receive the correct AppendEntries.

Possible fixes:
- Do not allow a view change within an AppendEntries. The sender already chunks the range into multiple AEs, it can additionally choose to chunk on every view boundary.
- Send the full view history in the AE header, rather than just the `term_of_index`. This is dynamically sized, but in practice still bounded (at most 1 view change per entry, and at most N entries in an A) and a tiny fraction of the total AE size.

eddyashton:
Fixed in #3145.

Issue #1451, opened 2020-07-28
Raft catch-up logic is not efficient


jumaffre:
When a new node joins a network, it currently has to replay all transactions since the beginning of time (for a way to do this more effectively, see https://github.com/microsoft/CCF/milestone/12). However, the current catch up logic seems inefficient in many ways (with primary's node log evidence for each claim):
- The first Append Entries message sent to the new joiner is sent from the primary's `last_idx`, which will automatically be rejected by the backup:
```
2020-07-28T13:55:25.602905Z -0.554 0   [info ] ../src/node/channels.h:344           | node channel with 9 is now established
...
2020-07-28T13:55:25.664574Z -0.436 0   [debug] ../src/consensus/raft/raft.h:499     | Send append entries from 8 to 9: 155 to 154 (152)
2020-07-28T13:55:25.664684Z        100 [debug] ../src/host/node_connections.h:278   | send AE to node 9 [96]: 154, 154
2020-07-28T13:55:25.664762Z -0.437 0   [debug] ../src/consensus/raft/raft.h:785     | Recv append entries response to 8 from 9: failed
```
- Further Append Entries are sent one by one:
```
2020-07-28T13:55:25.664843Z -0.437 0   [debug] ../src/consensus/raft/raft.h:499     | Send append entries from 8 to 9: 1 to 2 (152)
2020-07-28T13:55:25.686221Z        100 [debug] ../src/host/node_connections.h:278   | send AE to node 9 [17255]: 2, 0
2020-07-28T13:55:25.687620Z -0.459 0   [debug] ../src/consensus/raft/raft.h:499     | Send append entries from 8 to 9: 3 to 3 (152)
2020-07-28T13:55:25.688096Z        100 [debug] ../src/host/node_connections.h:278   | send AE to node 9 [1088]: 3, 2
2020-07-28T13:55:25.688246Z -0.460 0   [debug] ../src/consensus/raft/raft.h:499     | Send append entries from 8 to 9: 4 to 4 (152)
2020-07-28T13:55:25.688913Z        100 [debug] ../src/host/node_connections.h:278   | send AE to node 9 [1675]: 4, 3
2020-07-28T13:55:25.689093Z -0.461 0   [debug] ../src/consensus/raft/raft.h:499     | Send append entries from 8 to 9: 5 to 5 (152)
2020-07-28T13:55:25.689470Z        100 [debug] ../src/host/node_connections.h:278   | send AE to node 9 [132]: 5, 4
...
[...]

PR #3145, opened 2021-10-29
Refuse AppendEntries from other branches

eddyashton:
Fix for #3142. Opening as a draft because it's late on a Friday, and I don't remember exactly what I hacked in to get here.

The issue, in short, is that a malicious host could construct a plausible AppendEntries body using entries from a dead branch (which should have been truncated, and will never be committed), which do not match the range that the enclave intended to describe in its brief AppendEntries header, but which will be accepted (at least partially) by another node, resulting in them reaching a deadend state (holding a suffix they believe is committed, which doesn't match the rest of the service's).

The fix has 2 parts:
- Assume that we know the expected `TxID` for every entry in an AppendEntries (aka - the value the sender intended to send us), and check that matches the entry we received (determined by the view and seqno in the GcmHeader IV). If they don't match, don't apply this entry and return a failure (including an AppendEntries NACK) straight away, without considering more of this message.
- To know the intended `TxID` without modifying the AppendEntries header format to include a dynamic view history, we limit each AppendEntries to contain entries from a single view. So as described in the comment, `term_of_idx` actually describes the view of all transactions in this message (`prev_term` may be different, in which case the first transaction in this message is the first transaction in the new `term_of_idx`).

I'm not sure how the latter plays with `lts_compatibility`, so this may need revisiting. Changing the format is likely more intrusive, but there's an issue in the current approach where the new receiving node is making assumptions from the header that a sender running old code didn't include. I _think_ these will just result in a mismatch ("I thought you told me all of these entries were in term 5, but this includes `2.10`?") that can be successfully resolved, since the sender will eventually send partial AEs which can be accepted? But need to confirm.

The other half of this PR is a unit test which confirms the behaviour, and the fix. This required a lot more plumbing than anticipated! The problem is that the serialisation and parsing of the ledger entries in our unit tests doesn't include the GcmHeader containing a TxID, so doesn't do the same checks that we expect to result in a recognised mismatch in the real nodes. We reproduce this by changing some APIs slightly, to explicitly pass a TxID from Raft (where it is known), for the sake of the tests, though it's already implicitly included in a real service. This is rubbish! Additionally, the abstractions around our ledger mocks and async consensus execution are getting extremely unwieldy, so a lot of this PR is just blind plumbing.


src/consensus/ledger_enclave.h
```diff
@@ -53,7 +62,9 @@ namespace consensus
       const uint8_t* data,
       size_t size,
       bool globally_committable,
-      bool force_chunk = false)
+      bool force_chunk,
+      kv::Term term,
+      kv::Version index)
```
jumaffre, line 67: Are these expected to be passed to the host?

eddyashton: No. They could be, but I don't think we have a purpose for them, and they're not needed at the moment. The argument is that Raft _knows_ the term and index it is appending; logically it is calling `ledger.append(term, index, data)`, so it should pass those fields in case they're important to the receiver. For the stub ledger used in the unit tests, this is important. For this real implementation it's not (currently). But maybe at some point in the future, it would make sense to do the GcmHeader construction here (still within enclave, but logically part of the ledger representation rather than the 'raw' body)?

jumaffre: I didn't see that they were used in the unit test logic, makes sense!

src/consensus/aft/raft.h
```diff
@@ -166,7 +166,7 @@ namespace aft
 
     size_t entry_size_not_limited = 0;
     size_t entry_count = 0;
-    Index entries_batch_size = 1;
+    Index entries_batch_size = 20;
```
eddyashton, line 169: This is another drive-by mitigation for #1451. Start with larger batches than 1, so that we have fewer AEs and faster catchup in normal operation.

achamayou: Possibly good to split in its own PR if we thought we wanted to measure the perf impact independently.

src/consensus/aft/raft.h
```diff
@@ -1444,20 +1450,39 @@ namespace aft
...
-        send_append_entries_range(to, start_idx, i);
-        start_idx = std::min(i + 1, state->last_idx);
-      }
+      auto calculate_end_index = [this](Index start) {
+        // Cap the end index in 2 ways:
+        // - Must contain no more than entries_batch_size entries
+        // - Must contain entries from a single term
+        auto max_end = state->last_idx;
+        const auto term_of_ae = state->view_history.view_at(start);
+        const auto end_of_term = state->view_history.end_of_view(term_of_ae);
```
achamayou, line 1467: I would suggest index_at_end_of_term instead.

src/consensus/aft/raft.h
```diff
@@ -1444,20 +1450,39 @@ namespace aft
...
-      for (Index i = end_idx; i < state->last_idx; i += entries_batch_size)
-      {
-        send_append_entries_range(to, start_idx, i);
-        start_idx = std::min(i + 1, state->last_idx);
-      }
+      auto calculate_end_index = [this](Index start) {
+        // Cap the end index in 2 ways:
+        // - Must contain no more than entries_batch_size entries
+        // - Must contain entries from a single term
+        auto max_end = state->last_idx;
```
achamayou, line 1465: max_index?

achamayou (approved):

PR #3154, opened 2021-11-02
Do not build doc pre-1.0.1x

achamayou:
The doc job is starting to take quite a long time, and so it's tempting not to maintain documentation for older LTS releases, particularly since the changes from one LTS to the next are very small.

I think we also only should build the doc for the latest -dev, but that's quite a bit more difficult to do with regexps sadly.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -198,8 +198,8 @@ breathe_default_project = "CCF"
 
 # Set up multiversion extension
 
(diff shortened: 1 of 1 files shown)


eddyashton (approved):

eddyashton:
For reference, looks like this reduces the `CCF GitHub Pages` job from ~35m to ~19m.

PR #3158, opened 2021-11-03
Tweak printing of legend for `ledger_viz`

eddyashton:
I keep misreading the key printed by `ledger_viz`.

image

At a glance, I find the word "Signature", and the colour right next to it is red, so the initial read is that Signatures are red. I think the problem is that the `: ` spacer introduces more distance between the key and its value, than the ` ` separator between entries.

This changes the separator so that my feeble brain can read it at a glance.

image

### DIFF
--- a/python/ccf/ledger_viz.py
+++ b/python/ccf/ledger_viz.py
@@ -64,9 +64,9 @@ class DefaultLiner(Liner):
 
     def help(self):
         print(
-            " ".join(
+            " | ".join(
                 [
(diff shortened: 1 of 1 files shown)


achamayou:
I think we ought to give each label its own line, that's how most legends work and it's unambiguous. It also encourages keeping the list short, which is vital for comprehension if not artistic merit.

achamayou (approved):

PR #3162, opened 2021-11-04
Move key_exchange to src/crypto

rengolin:
Even though this uses tls' key exchange, this has nothing to do with our
TLS handling and should be in src/crypto folder instead.

Like other crypto logic, this depends on mbedTLS and needs to be
migrated to OpenSSL sometime soon, but not now. The main priority is to
get TLS 1.3 operational (with OpenSSL) and moving this out helps reduce
the workload for that job.


jumaffre (approved):
Thank you!

achamayou (approved):

PR #3163, opened 2021-11-04
Fix #3102

### LINKED ISSUE #3102: Improve C++/TS docstring for getStatusForTxId
It's copied from the C++ docstring. Improving the C++ docstrings and copying to the JS ones is probably a separate PR. Can you open an issue?

_Originally posted by @letmaik in https://github.com/microsoft/CCF/pull/3044#discussion_r729654377_

achamayou:
fix #3102


include/ccf/base_endpoint_registry.h
```diff
@@ -83,8 +83,8 @@ namespace ccf
 
     /** Get the status of a transaction by ID, provided as a view+seqno pair.
      * This is a node-local property - while it will converge on all nodes in
-     * a healthy network, it is derived from distributed state rather than
-     * distributed itself.
+     * a healthy network, it is either derived from distributed state in BFT,
+     * or replicated out by the primary in CFT.
```
eddyashton, line 87: GitHub just lost a longer comment here.

I agree the original comment is a little unclear, but I think this is missing an important distinction between replicated (from the KV) and node-local (everything else) state.

"What is the value of key X in table Foo at TxID M.N?" => A question about replicated state, will get the same answer on ever node (if they know M.N)
"Is M.N committed?" => A derived node-local property, which may differ across nodes (even nodes which all _have_ M.N, and could answer the first question identically)

We have stronger consistency + serializability claims about the state that is directly replicated (we'll tell you a fact that exists at a specific TxID) than anything that is node-local. I know the main source of information for this bit of node-local state _is_ replicated by the primary, but I think the local thing is distinct, and derived (it's not the primary's last commit index, it's that combined with the last index we have locally, which means we may report a global commit index different from anything the primary ever claimed, because it advanced in different batches).

eddyashton: I guess what I want here is a strong warning that this API, and anything else that accesses consensus/node-local things, is _not_ transactional and may not do what the caller expects. But the replicated/derived language is a step removed from that. What we really want is a disclaimer that doing:

```
status1 = get_status_v1(ARGS);
time.sleep(5);
status2 = get_status_v1(ARGS);
assert(status1 == status2);
```

might fail.

achamayou: Is the new phrasing better?

eddyashton: Yes, I think the new phrasing covers all the important details!

include/ccf/base_endpoint_registry.h
```diff
@@ -82,9 +82,24 @@ namespace ccf
...
 
     /** Get the status of a transaction by ID, provided as a view+seqno pair.
-     * This is a node-local property - while it will converge on all nodes in
-     * a healthy network, it is derived from distributed state rather than
-     * distributed itself.
+     *
+     * Note that this value is the node's local understanding of the status
+     * of that transaction in the network at call time. For a given TxID, the
+     * initial status is always UNKNOWN, and eventually becomes COMMITTED or
+     * INVALID. See the documentation section entitled "Verifying Transactions"
```
letmaik, line 89:
```suggestion
     * INVALID. See the documentation section titled "Verifying Transactions"
```

include/ccf/base_endpoint_registry.h
```diff
@@ -82,9 +82,24 @@ namespace ccf
...
+     * for more detail.
+     *
+     *         UNKNOWN [Initial status]
+     *          v  ^
+     *        PENDING
+     *        v     v
+     *  COMMITTED INVALID [Final statuses]
```
letmaik, line 96: Are these transitions strictly true actually? For example, can't you jump from Unknown to Invalid directly? When would you transition back to Unknown once you're in Pending?

js/ccf-app/src/global.ts
```diff
@@ -253,8 +253,8 @@ export interface CCFConsensus {
   /**
    * Get the status of a transaction by ID, provided as a view+seqno pair.
    * This is a node-local property - while it will converge on all nodes in
-   * a healthy network, it is derived from distributed state rather than
-   * distributed itself.
+   * a healthy network, it is either derived from distributed state in BFT,
+   * or replicated out by the primary in CFT.
```
letmaik, line 257: Shouldn't this be a copy-paste of the full C++ docstring?

include/ccf/base_endpoint_registry.h
achamayou: 1. Not all states are necessarily observed, depending on when you call. '`A -> B -> C` can look like `A->C` if you don't call when it was `B`. So while it's possible to go straight from `UNKNOWN` to a final state, I think it's acceptable not to try and display it here.
2. Yes you can, see above.
3. Node (primary) creates 3.5 [PENDING]. Rollback occurs on node to 2.4. 3.5 is now [UNKNOWN], until N.5 is applied where N>3 (received from new primary).

js/ccf-app/src/global.ts
achamayou: Yes, it will be in a moment, since we agree on the phrasing now.

letmaik (approved):

eddyashton (approved):

PR #3169, opened 2021-11-05
Never rollback initial ledger secrets

jumaffre:
When a backup node joins a service, it is given the entire history of ledger secrets so far (part of the `/node/join` response). If an election is triggered when the node is catching up, the node will rollback its store, which will cause the ledger secrets to be rolled back too. This is incorrect since the backup will need the ledger secrets to decrypt further entries, once the catch up phase is complete.

For example, if a new node joined without a snapshot and received the ledger secrets at seqnos `{1, 10}`, caught up till `5` at which point an election was triggered (new joiner's store is rolled back at `5`, ledger secrets at now `{1}`), it wouldn't have been able to deserialise entries from `10` onwards.

We now protect the initial ledger secrets received in the `/node/join` response so that they cannot be rolled back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - DNS resolution of client connections is now asynchronous.
 
+### Fixed
+
+- Fixed issue with new node startup which could get stuck if an election was triggered while catching up (#3169).
+
 ## [2.0.0-dev5]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
-- Fixed issue with new node startup which could get stuck if an election was triggered while catching up (#3169).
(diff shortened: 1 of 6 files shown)


eddyashton (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - DNS resolution of client connections is now asynchronous.
 
+### Fixed
+
+- Fixed issue with new node startup which could get stuck if an election was triggered while catching up (#3169).
```
achamayou, line 20:
```suggestion
- Fixed issue with node join which could get stuck if an election was triggered while catching up (#3169).
```

src/node/ledger_secrets.h
```diff
@@ -25,6 +25,10 @@ namespace ccf
     std::mutex lock;
     LedgerSecretsMap ledger_secrets;
 
+    // Set once when the LedgerSecrets are initialised. This prevents a backup
+    // node to rollback not-yet-applicable ledger secrets when catching up.
```
achamayou, line 29:
```suggestion
    // node to rollback not-yet-applicable ledger secrets when catching up.
    // All rollback that would result in the removal of some of these secrets
    // would imply that the transaction that added the node itself was rolled back,
```

achamayou (approved):

PR #3172, opened 2021-11-05
Accept unknown nodes when receiving Raft RequestVote message

### LINKED ISSUE #3167: Nodes will not vote for new primaries in Raft if they do not know the requesting node
This looks like an unnecessary restriction. We believe it is safe to receive AppendEntries from an unknown node, since the header and contents are still useable (and often necessary to help this node catch up to their own joining point!). Presumably this is also true for a RequestVote.

This has an impact on liveness, since it means that a node that is catching up is _not_ a full member for consensus purposes, essentially they still count as a dead node, so the service doesn't have the fault tolerance that it may appear to have. So in the specific condition that a node is catching up, has not heard about their own join, and is counted as one of the necessary live nodes (eg - in a 3-node network), it's possible that a single node crash will kill the network.

We can add a test for this behaviour, and the fix (if safe) is simple - remove this early-out, and process RequestVotes from unknown nodes.

eddyashton:
Fixes #3167.

Adds a partition test exhibiting the failure, and removes the unnecessary early-out path.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -2800,6 +2800,13 @@ namespace aft
     {
       std::lock_guard<std::mutex> guard(state->lock);
 
+      // Do not check that from is a known node. It is possible to receive
+      // RequestVotes from nodes that this node doesn't yet know, just as it
+      // receives AppendEntries from those nodes. These should be obeyed just
+      // like any other RequestVote - it is possible that this node is needed to
+      // produce a primary in the new term, who will then help this node catch
+      // up.
+
       if (state->current_view > r.term)
       {
         // Reply false, since our term is later than the received term.
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -2800,17 +2800,6 @@ namespace aft
     {
       std::lock_guard<std::mutex> guard(state->lock);
 
-      // Ignore if we don't recognise the node.
-      auto node = nodes.find(from);
-      if (node == nodes.end())
-      {
-        LOG_FAIL_FMT(
-          "Recv request vote to {} from {}: unknown node",
-          state->my_node_id,
-          from);
-        return;
-      }
-
       if (state->current_view > r.term)
       {
         // Reply false, since our term is later than the received term.
--- a/tests/partitions_test.py
+++ b/tests/partitions_test.py
@@ -11,6 +11,7 @@ import pprint
 from ccf.tx_status import TxStatus
 import time
 import http
+import contextlib
 
 from loguru import logger as LOG
 
@@ -151,6 +152,7 @@ def test_isolate_and_reconnect_primary(network, args, **kwargs):
 
 
 @reqs.description("New joiner helps liveness")
+@reqs.exactly_n_nodes(3)
 def test_new_joiner_helps_liveness(network, args):
     primary, backups = network.find_nodes()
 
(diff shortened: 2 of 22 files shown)


tests/partitions_test.py
```diff
@@ -148,6 +150,56 @@ def test_isolate_and_reconnect_primary(network, args, **kwargs):
...
 
 
+@reqs.description("New joiner helps liveness")
+def test_new_joiner_helps_liveness(network, args):
+    primary, backups = network.find_nodes()
+
+    # Issue some transactions, so there is a ledger history that a new node must receive
+    network.txs.issue(network, number_txs=10)
+
+    # Remove a node, leaving the network frail
```
jumaffre, line 160: Does this test only makes sense if there are 3 nodes in the network to start with? If so, this could be expressed as a requirement.

tests/partitions_test.py
```diff
@@ -148,6 +150,56 @@ def test_isolate_and_reconnect_primary(network, args, **kwargs):
...
+    network.retire_node(primary, backups[-1])
+    backups[-1].stop()
+
+    primary, backups = network.find_nodes()
+
+    # Add a new node, but partition them before trusting them
+    new_node = network.create_node("local://localhost")
+    network.join_node(new_node, args.package, args, from_snapshot=False)
+    new_joiner_partition = [new_node]
+    new_joiner_rules = network.partitioner.partition(
```
jumaffre, line 170: Could you use the context manager variant for this? I believe it would make the test slightly clearer since there are two partitions going on and would also automatically delete the partition if an exception occurs.

tests/partitions_test.py
```diff
@@ -148,6 +150,56 @@ def test_isolate_and_reconnect_primary(network, args, **kwargs):
...
+
+    # Partition the primary, temporarily creating a minority service that cannot make progress
+    minority_partition = backups[len(backups) // 2 :] + new_joiner_partition
+    minority_rules = network.partitioner.partition(minority_partition)
+    time.sleep(
+        network.observed_election_duration
+    )  # Ensure this node has become primary
+    with backups[0].client("user0") as c:
+        r = c.post("/app/log/private", {"id": 42, "msg": "Hello world"})
+        assert r.status_code == http.HTTPStatus.SERVICE_UNAVAILABLE
```
jumaffre, line 186: Would `wait_for_new_primary()` wrapped in a `try: ... except TimeoutError:` be more consistent with that we do here?

src/consensus/aft/raft.h
```diff
@@ -2800,17 +2800,6 @@ namespace aft
     {
       std::lock_guard<std::mutex> guard(state->lock);
 
-      // Ignore if we don't recognise the node.
```
jumaffre, line 2803: Could we add a comment here, stating that it's OK to accept votes from unknown nodes?

tests/partitions_test.py
eddyashton: I'd tried something similar, but most of the existing infra helpers don't work here because they don't expect the node to be in this state. Specifically if we do anything before the timeout above, the request may timeout as its forwarded to an unreachable primary, and anything after returns this unusual error code. So I think it's nicer to be explicit (you can talk to this node, but it returns this error), rather than catching the exception thrown by something _else_ trying to talk to this node.

tests/partitions_test.py
eddyashton: We could use a context manager for one of these, but not both (easily), because they're opened and dropped in the same order, not nested. ie `create_partition(A); create_partition(B); drop_partition(B); drop_partition(A);` can be converted to context managers, but we need the exit stack thing for this `create_partition(A); create_partition(B); drop_partition(A); drop_partition(B);` pattern. I'll see if that works.

tests/partitions_test.py
```diff
@@ -163,41 +165,44 @@ def test_new_joiner_helps_liveness(network, args):
...
 
-    network.wait_for_primary_unanimity()
-    primary, _ = network.find_nodes()
-    network.wait_for_all_nodes_to_commit(primary=primary)
+        # Partition the primary, temporarily creating a minority service that cannot make progress
+        minority_partition = backups[len(backups) // 2 :] + new_joiner_partition
+        minority_rules = stack.enter_context(
+            network.partitioner.partition(minority_partition)
+        )
+        time.sleep(
```
achamayou, line 186: Please replace this with a periodic poll timing out at network.observed_election_duration, in many cases we probably don't need to (and therefore want to) wait this long. The comment is also inaccurate it seems? Nothing checks for a new primary afaict.

tests/partitions_test.py
eddyashton: Turns out this _is_ context stack-able, I didn't know you could still get the return values to manually drop. Cool.

tests/partitions_test.py
eddyashton: Added a comment explaining why this is a bit weird, and the sleep is the easiest solution.

achamayou (approved):

PR #3184, opened 2021-11-05
Fix retirement bugs in 2tx reconfig

wintersteiger:
This fixes a bug in the retirement logic for 2tx reconfiguration. A node would not revert to being a follower after it's change to `RETIRING` was rolled back. It does so with this fix, but it now triggers a long series of elections, sometimes ending up in a situation where it rejects append entries for a long time (possibly forever).

Also adds a function to wait for reconfiguration completion in tests to avoid tests interfering with each other.


wintersteiger:
All good. This is ready for review.

achamayou (approved):

PR #3186, opened 2021-11-08
Cherry-pick #3172 to 1.x LTS

eddyashton:
#3172

Note this doesn't include the test+infra changes from main, since the partitions test doesn't exist on 1.x.


achamayou (approved):

eddyashton:
@achamayou The `lts_compatibility` test was failing for a cascade of reasons which I believe were fundamentally based on incorrect assumptions that have since been removed on `main`, so I've also added the simplifications of #3046 to this PR.

I don't completely understand the sequence of events that was failing, but somehow we were stuck in an election cycle.

PR #3188, opened 2021-11-09
Update ringbuffer ledger API - fetch ranges rather than individual entries

eddyashton:
When serving range historical queries, we request a range of ledger entries from the host. Currently those are requested one-by-one, in individual async ringbuffer calls, even though the host-side ledger has a much more performant range-based API. This changes that so that all ledger requests from the enclave request a range, so we should get a significant perf speedup for historical range queries.


src/node/historical_queries.h
```diff
@@ -197,7 +218,7 @@ namespace ccf::historical
             const auto next_seqno = last_requested_seqno + 1;
             supporting_signature =
               std::make_pair(next_seqno, std::make_shared<StoreDetails>());
-            ret.insert(next_seqno);
+            ret.insert(std::make_pair(next_seqno, next_seqno));
```
achamayou, line 221:
```suggestion
            ret.emplace(next_seqno, next_seqno);
```

achamayou:
@eddyashton that doesn't seem to make such an obvious difference on the end to end historical perf test?

image

src/host/ledger.h
```diff
@@ -1107,28 +1114,37 @@ namespace asynchost
...
+                // NB: Even if status is cancelled (and entry is empty), we
+                // want to write this result back to the enclave
+                write_ledger_get_range_response(
+                  from_idx,
+                  to_idx,
+                  read_framed_entries(from_idx, to_idx),
+                  purpose);
+              };
+
+              work_handle->data = job;
```
achamayou, line 1147: work_handle presumably deletes the job at some point?

src/node/historical_queries.h
```diff
@@ -146,7 +146,7 @@ namespace ccf::historical
       // adjust to:
       //  0  1  2  3  4  5  6
       // we need to shift _and_ start fetching 0, 1, and 6.
-      std::set<ccf::SeqNo> adjust_range(
+      std::set<std::pair<ccf::SeqNo, ccf::SeqNo>> adjust_range(
```
achamayou, line 149: Maybe typedef an IndexRange? (using, but that doesn't verb well).

achamayou (approved):

src/host/ledger.h
eddyashton: Yes, this is deleted in `on_ledger_get_async_complete`. This is just renaming `data` to `job`, to avoid shadowing the existing `data` variable.

src/host/ledger.h
```diff
@@ -1017,7 +1021,8 @@ namespace asynchost
     {
       // Filled on construction
       Ledger* ledger;
-      size_t idx;
+      size_t from_idx;
+      size_t to_idx;
 
       // First argument is ledger entry (or nullopt if not found)
       // Second argument is uv status code, which may indicate a cancellation
```
jumaffre, line 1028: Should the `entry` member on line 1034 be renamed to `entries`?

PR #3189, opened 2021-11-09
Typedoc upgrade plus some doc fixes

achamayou:
Supersedes #3179

### DIFF
--- a/js/ccf-app/typedoc.json
+++ b/js/ccf-app/typedoc.json
@@ -13,8 +13,7 @@
   "out": "html",
   "theme": "default",
   "categorizeByGroup": false,
-  "validation":
-  {
+  "validation": {
     "invalidLink": true
   }
 }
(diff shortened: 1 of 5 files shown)


js/ccf-app/typedoc.json
```diff
@@ -11,7 +11,10 @@
     "src/polyfill.ts"
   ],
   "out": "html",
-  "theme": "./doc/theme",
+  "theme": "default",
```
achamayou, line 14: The old custom theme no longer works, and it's not entirely clear how the new mechanism is meant work either: https://github.com/TypeStrong/typedoc/blob/master/internal-docs/custom-themes.md

The version selector wasn't functional in the doc recently in any case, so we will revisit this when we find a good example for a custom theme.

eddyashton (approved):

js/ccf-app/typedoc.json
matteobruni: I've just found other bugs to the new typedoc custom themes, you can see the new issue here: https://github.com/TypeStrong/typedoc/issues/1786

achamayou: @matteobruni thank you for the pointer, much appreciated!

PR #3198, opened 2021-11-10
1tx to 2tx reconfiguration migration

wintersteiger:
Here's a sketch of how we could do 1tx to 2tx reconfiguration migration. Not necessarily intended to be merged, just a basis for discussion on how we want to do things. 

I think adding a hook on the service config table is a good idea in general, even if all we want to do is to reject any unwanted updates.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -719,6 +719,7 @@ namespace aft
       {
         details.acks[k] = v.match_idx;
       }
+      details.reconfiguration_type = reconfiguration_type;
       if (reconfiguration_type == ReconfigurationType::TWO_TRANSACTION)
       {
         details.learners = learner_nodes;
@@ -2529,6 +2530,16 @@ namespace aft
             node_info.second.hostname,
             node_info.second.port);
         }
+      }
+    }
+
+  public:
+    void update_parameters(kv::ConsensusParameters& params)
+    {
+      reconfiguration_type = params.reconfiguration_type;
+      if (reconfiguration_type == TWO_TRANSACTION && !node_client)
+      {
+        throw std::logic_error("missing node client");
       }
     }
   };
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -2536,14 +2536,9 @@ namespace aft
   public:
     void update_parameters(kv::ConsensusParameters& params)
     {
-      if (reconfiguration_type == params.reconfiguration_type)
-      {
-        return;
-      }
-      if (params.reconfiguration_type == TWO_TRANSACTION && !node_client)
-      {
-        throw std::logic_error("missing node client");
-      }
+      CCF_ASSERT_FMT(
+        params.reconfiguration_type != TWO_TRANSACTION || node_client,
+        "Bug; all enclaves that support 2tx reconfig must have node_clients");
       reconfiguration_type = params.reconfiguration_type;
     }
   };
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -2536,6 +2536,8 @@ namespace aft
   public:
     void update_parameters(kv::ConsensusParameters& params)
     {
+      // This should only be called when the state->lock is held, so we do not
+      // acquire the lock here.
       CCF_ASSERT_FMT(
         params.reconfiguration_type != TWO_TRANSACTION || node_client,
         "Bug; all enclaves that support 2tx reconfig must have node_clients");
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -2536,11 +2536,15 @@ namespace aft
   public:
     void update_parameters(kv::ConsensusParameters& params)
     {
-      reconfiguration_type = params.reconfiguration_type;
-      if (reconfiguration_type == TWO_TRANSACTION && !node_client)
+      if (reconfiguration_type == params.reconfiguration_type)
+      {
+        return;
+      }
+      if (params.reconfiguration_type == TWO_TRANSACTION && !node_client)
       {
         throw std::logic_error("missing node client");
       }
+      reconfiguration_type = params.reconfiguration_type;
     }
   };
 }
--- a/src/node/hooks.h
+++ b/src/node/hooks.h
@@ -172,7 +172,7 @@ namespace ccf
   inline void service_configuration_commit_hook(
     kv::Version version,
     const ccf::Configuration::Write& w,
-    std::shared_ptr<kv::Consensus> consensus)
+    const std::shared_ptr<kv::Consensus>& consensus)
   {
     LOG_DEBUG_FMT("Service configuration update hook");
     assert(w.has_value());
--- a/src/node/hooks.h
+++ b/src/node/hooks.h
@@ -3,6 +3,8 @@
 #pragma once
 
 #include "ds/logger.h"
+#include "enclave/reconfiguration_type.h"
+#include "node/config.h"
 #include "node/signatures.h"
 #include "node_info_network.h"
 
@@ -167,4 +169,28 @@ namespace ccf
     }
   };
 
+  class ServiceConfigurationUpdateHook : public kv::ConsensusHook
+  {
+    kv::Version version;
+    ServiceConfiguration new_service_config;
+
+  public:
+    ServiceConfigurationUpdateHook(
+      kv::Version version_, const Configuration::Write& w)
+    {
+      assert(w.has_value());
+      version = version_;
+      new_service_config = w.value();
+    }
+
+    void call(kv::ConfigurableConsensus* consensus) override
+    {
+      LOG_DEBUG_FMT("Service configuration update hook");
+      kv::ConsensusParameters cp;
+      cp.reconfiguration_type =
+        new_service_config.reconfiguration_type.value_or(
+          ReconfigurationType::ONE_TRANSACTION);
+      consensus->update_parameters(cp);
+    }
+  };
 }
--- a/src/node/hooks.h
+++ b/src/node/hooks.h
@@ -169,28 +169,17 @@ namespace ccf
     }
   };
 
-  class ServiceConfigurationUpdateHook : public kv::ConsensusHook
+  inline void service_configuration_commit_hook(
+    kv::Version version,
+    const ccf::Configuration::Write& w,
+    std::shared_ptr<kv::Consensus> consensus)
   {
-    kv::Version version;
-    ServiceConfiguration new_service_config;
-
-  public:
-    ServiceConfigurationUpdateHook(
-      kv::Version version_, const Configuration::Write& w)
-    {
-      assert(w.has_value());
-      version = version_;
-      new_service_config = w.value();
-    }
-
-    void call(kv::ConfigurableConsensus* consensus) override
-    {
-      LOG_DEBUG_FMT("Service configuration update hook");
-      kv::ConsensusParameters cp;
-      cp.reconfiguration_type =
-        new_service_config.reconfiguration_type.value_or(
-          ReconfigurationType::ONE_TRANSACTION);
-      consensus->update_parameters(cp);
-    }
-  };
+    LOG_DEBUG_FMT("Service configuration update hook");
+    assert(w.has_value());
+    auto new_service_config = w.value();
+    kv::ConsensusParameters cp;
+    cp.reconfiguration_type = new_service_config.reconfiguration_type.value_or(
+      ReconfigurationType::ONE_TRANSACTION);
+    consensus->update_parameters(cp);
+  }
 }
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -2025,10 +2025,10 @@ namespace ccf
 
       network.tables->set_global_hook(
         network.config.get_name(),
-        network.config.wrap_map_hook(
-          [](kv::Version version, const Configuration::Write& w)
-            -> kv::ConsensusHookPtr {
-            return std::make_unique<ServiceConfigurationUpdateHook>(version, w);
+        network.config.wrap_commit_hook(
+          [c = this->consensus](
+            kv::Version version, const Configuration::Write& w) {
+            service_configuration_commit_hook(version, w, c);
           }));
 
       setup_basic_hooks();
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -2023,6 +2023,14 @@ namespace ccf
             return std::make_unique<SerialisedMerkleTreeHook>(version, w);
           }));
 
+      network.tables->set_map_hook(
+        network.config.get_name(),
+        network.config.wrap_map_hook(
+          [](kv::Version version, const Configuration::Write& w)
+            -> kv::ConsensusHookPtr {
+            return std::make_unique<ServiceConfigurationUpdateHook>(version, w);
+          }));
+
       setup_basic_hooks();
     }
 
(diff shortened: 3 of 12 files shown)


tests/reconfiguration.py
```diff
@@ -596,6 +597,171 @@ def run_join_old_snapshot(args):
...
+        args.binary_dir,
+        args.debug_nodes,
+        args.perf_nodes,
+        pdb=args.pdb,
+        txs=txs,
+    ) as network:
+        network.start_and_join(args)
+        primary, _ = network.find_primary()
+
+        config_before = get_current_config(network)
```
jumaffre, line 665: In our end-to-end tests, we have found that it is better (as in faster and less flaky) to ask the live service rather than reading the ledger (especially when nodes are still writing to it). For nodes, you can use the existing `/node/network/nodes` endpoint. I don't know if we have an endpoint that returns the current service configuration but we could add it to the existing `/node/network` endpoint (or add a new `/node/service` endpoint)?

src/runtime_config/default/actions.js
```diff
@@ -1060,4 +1060,29 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "migrate_service_to_2tx_reconfig",
```
jumaffre, line 1064: Should this just be a generic `set_service_configuration` action that can have an arbitrary number of fields as arguments and would update those and only those? The existing `set_recovery_threshold` action could be removed to use this too.

src/node/node_state.h
```diff
@@ -2051,6 +2051,14 @@ namespace ccf
             return std::make_unique<SerialisedMerkleTreeHook>(version, w);
           }));
 
+      network.tables->set_map_hook(
```
jumaffre, line 2054: What if the transaction is rolled back?

wintersteiger: Good point, we would need to be careful not to apply this too early. I think this should live outside of `raft.h`, but I don't know how to check for global commit outside of that. Would a hook in `Store::global_hooks` give us that guarantee?

src/runtime_config/default/actions.js
wintersteiger: Yeah, I think in the long run that's the right thing to do, just offer one general way to perform config updates.

tests/reconfiguration.py
wintersteiger: Thanks, that's good to know! In this particular case I want to make sure that some things are correctly recorded on the ledger, e.g. find the transaction at which a learner becomes trusted, which I figured is best done by reading the ledger directly. I agree though, we should add an endpoint to get the service config.

src/node/node_state.h
jumaffre: Yes, a global hook would do the job as its effects cannot be rolled back.

tests/reconfiguration.py
achamayou: https://microsoft.github.io/CCF/main/operations/operator_rpc_api.html#get--config
https://microsoft.github.io/CCF/main/operations/operator_rpc_api.html#get--consensus

wintersteiger: That's the node/consensus config, but there is no endpoint for the service config (as in `src/node/config.h`), right?

achamayou: Indeed no, but to find out if a learner is now trusted, that should work. We can and should add an endpoint for the service config, I agree.

src/consensus/aft/raft.h
```diff
@@ -2531,5 +2532,14 @@ namespace aft
...
       }
     }
+
+  public:
+    void update_parameters(kv::ConsensusParameters& params)
+    {
+      CCF_ASSERT_FMT(
+        params.reconfiguration_type != TWO_TRANSACTION || node_client,
+        "Bug; all enclaves that support 2tx reconfig must have node_clients");
+      reconfiguration_type = params.reconfiguration_type;
```
jumaffre, line 2542: Since the consensus logic can be executed by multiple threads (at least when [`worker_threads` is > 0](https://microsoft.github.io/CCF/main/operations/configuration.html#worker-threads)), this feels like it needs to acquired the `state->lock` here?

src/node/hooks.h
```diff
@@ -167,4 +169,17 @@ namespace ccf
     }
   };
 
-}
\ No newline at end of file
+  inline void service_configuration_commit_hook(
+    kv::Version version,
+    const ccf::Configuration::Write& w,
+    std::shared_ptr<kv::Consensus> consensus)
```
jumaffre, line 175:
```suggestion
    const std::shared_ptr<kv::Consensus>& consensus)
```

src/runtime_config/default/actions.js
```diff
@@ -197,6 +197,84 @@ function setNodeCertificateValidityPeriod(
...
+  // First run all checks
+  checkReconfigurationType(config, new_config);
+  checkRecoveryThreshold(config, new_config);
+
+  // Then all updates
+  if (new_config.reconfiguration_type !== undefined) {
+    config.reconfiguration_type = new_config.reconfiguration_type;
+  }
+
+  let need_recovery_threshold_refresh = false;
```
jumaffre, line 258: Do we need this intermediate variable at all or could we call `ccf.node.triggerRecoverySharesRefresh();` directly on line `264` instead? `triggerRecoverySharesRefresh()` has no side effect apart from writing to the KV.

tests/reconfiguration.py
```diff
@@ -610,6 +611,170 @@ def run_join_old_snapshot(args):
...
+                            pending_at = i
+                        elif jv["status"] == "Learner":
+                            learner_at = i
+                        elif jv["status"] == "Trusted":
+                            trusted_at = i
+
+    return pending_at, learner_at, trusted_at
+
+
+def run_migration_tests(args):
```
jumaffre, line 662: Perhaps move all of this to a new file as the reconfiguration test file is already large?

tests/reconfiguration.py
```diff
@@ -610,6 +611,170 @@ def run_join_old_snapshot(args):
                 pass
 
 
+def get_current_config(network):
```
jumaffre, line 614: Could we use the new https://microsoft.github.io/CCF/main/operations/operator_rpc_api.html#get--service-configuration endpoint for this? 

Also, I think the `service-configuration` URI doesn't follow the existing convention we have for naming endpoints. I think `service/configuration` would be better.

jumaffre: Ah wait, this is about network configuration. I think my comment about using the dedicated endpoint still holds though.

tests/reconfiguration.py
```diff
@@ -610,6 +611,170 @@ def run_join_old_snapshot(args):
...
+            for n in network.nodes:
+                assert n.node_id in config_before["nodes"]
+
+        nodes_before = get_current_nodes_table(network)
+        for n in network.nodes:
+            assert n.node_id in nodes_before
+            assert nodes_before[n.node_id]["status"] == "Trusted"
+
+        # Check that the service config agrees that this is a 1tx network
+        tables, _ = network.get_latest_ledger_public_state()
```
jumaffre, line 691: Please use new endpoint here too.

tests/reconfiguration.py
```diff
@@ -610,6 +611,170 @@ def run_join_old_snapshot(args):
...
+        }
+        proposal = network.consortium.get_any_active_member().propose(
+            primary, proposal_body
+        )
+        network.consortium.vote_using_majority(
+            primary,
+            proposal,
+            {"ballot": "export function vote (proposal, proposer_id) { return true }"},
+            timeout=10,
+        )
```
jumaffre, line 716: This could be added to `consortium.py`, like for (most) other proposals.

src/runtime_config/default/actions.js
wintersteiger: I think we could, but it used to be done after the `.set`. If something goes wrong there, it would be called without the new value being recorded. I'm not sure whether something _can_ go wrong there though.

tests/reconfiguration.py
wintersteiger: Yes, this is the network config. At this point I want the list of nodes that's currently recorded on the ledger for later comparison. I don't want to test what any of the nodes think the config is, I want to test that things are recorded correctly on the ledger. 

(Sure, name change is fine)

src/consensus/aft/raft.h
wintersteiger: Good point, I agree.

wintersteiger: Actually, it deadlocks when I do that. The lock is apparently already help during a global hook, although I can't easily find where that's done.

tests/reconfiguration.py
wintersteiger: With the latest changes it shrunk a bit, I think it's fine to keep it there as I don't expect too many more lengthy migration tests will be added.

src/consensus/aft/raft.h
jumaffre: Good point. On the primary, it seems that lock is taken in `recv_append_entries_response()` which eventually calls `commit()` via `update_commit()`. We usually write a comment at the top of the function body when it's the case (see e.g. `get_term_history()`)

tests/reconfiguration.py
jumaffre: Because the latest state of the ledger (`get_latest_ledger_public_state()`) contains the same data than the KV store of the node, asking the node directly or reading the ledger should be equivalent. And as we found out the hard way, reading the ledger from the end-to-end tests can be flaky, it is preferable to send an RPC to the node to do so. In this case, AIUI, the `GET /node/consensus` should return the same value than what you're reading from the ledger?

tests/reconfiguration.py
```diff
@@ -610,6 +611,143 @@ def run_join_old_snapshot(args):
...
+                            trusted_at = i
+
+    return pending_at, learner_at, trusted_at
+
+
+def run_migration_tests(args):
+    if args.reconfiguration_type != "OneTransaction":
+        return
+
+    txs = app.LoggingTxs("user0")
```
jumaffre, line 666: `txs` is unused?

tests/reconfiguration.py
```diff
@@ -610,6 +611,143 @@ def run_join_old_snapshot(args):
...
+        # Add a new node and check that the reconfiguration is recorded correctly
+        tx_id_before = TxID(0, 0)
+        with primary.client() as c:
+            rj = c.get("/node/commit").body.json()
+            tx_id_before = TxID.from_str(rj["transaction_id"])
+
+        new_node = network.create_node("local://localhost")
+        network.join_node(new_node, args.package, args, from_snapshot=False)
+        network.trust_node(new_node, args)
+        network.wait_for_all_nodes_to_commit(primary)
```
jumaffre, line 720: This is already called in `trust_node()`.

tests/reconfiguration.py
```diff
@@ -610,6 +611,143 @@ def run_join_old_snapshot(args):
...
+            assert len(rj["details"]["learners"]) == 0
+
+        tx_id_after = TxID(0, 0)
+        with primary.client() as c:
+            rj = c.get("/node/commit").body.json()
+            tx_id_after = TxID.from_str(rj["transaction_id"])
+        pending_at, learner_at, trusted_at = get_joining_node_dates(
+            network, tx_id_before, tx_id_after, new_node.node_id
+        )
+        assert pending_at < learner_at < trusted_at
```
jumaffre, line 748: I feel like this tests more than just the migration but also the generic learner addition process. Is this done here because it is hard to observe that a new node was added as a learner (because nodes catch up fast!)?

src/consensus/aft/raft.h
wintersteiger: Great, I'll add a similar comment then!

tests/reconfiguration.py
wintersteiger: Yeah, I looked into that a little earlier today too. It basically gets the ledger from the primary anyways, so no big difference. Perhaps I should look into reading/auditing the ledger files after or at the end of the test to make sure it's all where it should be.

tests/reconfiguration.py
wintersteiger: No it isn't: `txs=txs,` in the call to `network(...)`

tests/reconfiguration.py
wintersteiger: Yes, this tests that the learner was in fact first added as a learner and then promoted to trusted/active, and that this happened between the two reconfiguration transactions. Since that's the basic 2tx guarantee, I wanted to make sure we test this. Like I said in the other comment above, maybe I should read the actual ledger files instead to make sure this is correctly auditable.

tests/reconfiguration.py
wintersteiger: I feel we should really do that validation on the ledger files after the test finishes. We can leave that for a separate PR, but is `read_ledger.py` as in the governance history test, the right way to access that?

jumaffre: Yes, reading the ledger once all the nodes are stopped is safer/better. You should be able to iterate over the ledger without `read_ledger.py`, by directly using the `ccf.ledger.Ledger` object (see https://microsoft.github.io/CCF/main/audit/python_library.html#tutorial).

src/runtime_config/default/actions.js
jumaffre: Because of https://github.com/microsoft/CCF/issues/1031, we used to tend to write proposal actions as if the proposal execution wasn't transactional. This is no longer the case and all transaction execution is now transactional so it doesn't really matter which order you call operations on the KV in.

wintersteiger: Great, then I'll simplify this!

tests/reconfiguration.py
```diff
@@ -610,6 +611,106 @@ def run_join_old_snapshot(args):
...
+            assert "reconfiguration_type" in rj["details"]
+            assert "learners" in rj["details"]
+            assert rj["details"]["reconfiguration_type"] == "TwoTransaction"
+            assert len(rj["details"]["learners"]) == 0
+
+        ledger_paths = primary.remote.ledger_paths()
+        learner_id = new_node.node_id
+
+        # Force ledger flush of all transactions so far.
+        network.get_latest_ledger_public_state()
```
wintersteiger, line 709: @jumaffre this is for the new checks on the ledger files. Is there a better/nicer way to force the ledger files to be flushed to disk? (It's necessary, I don't see the 2nd reconfig tx without it.)

jumaffre: Ah yes, this is because you passed `committed_only=True` when creating the `Ledger` instance on line `628` (which means only `.committed` ledger files are considered). Because you call `check_2tx_ledger()` when the network is stopped, all ledger files should be neatly closed and it should be OK to use `committed_only=False` and avoid having to call `network.get_latest_ledger_public_state()` here.

tests/reconfiguration.py
```diff
@@ -610,6 +611,106 @@ def run_join_old_snapshot(args):
...
+        pdb=args.pdb,
+        txs=txs,
+    ) as network:
+        network.start_and_join(args)
+        primary, _ = network.find_primary()
+
+        nodes_before = get_current_nodes_table(network)
+        for n in network.nodes:
+            assert n.node_id in nodes_before
+            assert nodes_before[n.node_id]["status"] == "Trusted"
```
jumaffre, line 671: Is this still needed?

jumaffre (approved):

tests/reconfiguration.py
wintersteiger: No, I'll remove that

tests/reconfiguration.py
wintersteiger: Thanks, that's great, looks much neater now!

PR #3200, opened 2021-11-10
Enable 2-tx reconfig in threading pipeline

wintersteiger:


wintersteiger:
I suspect the CI failure is expected? The 2tx reconfig test is fine, `e2e_logging_cft` failed during node shutdown.

eddyashton:
That CI failure has occurred in other runs on main, so isn't caused by this PR. Tracking it in #3202.

Please run this PR again, and see if we can get a green run before merging (the existing crash looks to be ~33% repro rate).

wintersteiger:
Great, thanks for checking @eddyashton! I did take a look but the information in the logs isn't enough for me to guess where the problem could be. Maybe add verbose logging for some CI runs to figure this out?

eddyashton (approved):

achamayou (approved):

PR #3205, opened 2021-11-11
Remove BFT timeouts options

jumaffre:
The BFT status interval was unused and we now use the Raft election timeout for BFT view changes as well.

### DIFF
--- a/tests/infra/remote.py
+++ b/tests/infra/remote.py
@@ -633,7 +633,6 @@ class CCFRemote(object):
         bin_path = os.path.join(".", os.path.basename(self.BIN))
         enclave_path = os.path.join(".", os.path.basename(lib_path))
 
-
         cmd = [
             bin_path,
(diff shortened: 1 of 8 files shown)


tests/infra/remote.py
```diff
@@ -651,7 +644,7 @@ def __init__(
             f"--snapshot-dir={self.snapshot_dir_name}",
             f"--node-cert-file={self.pem}",
             f"--host-log-level={host_log_level}",
-            election_timeout_arg,
+            f"--raft-election-timeout-ms={raft_election_timeout_ms}",
```
achamayou, line 647: I think we should take the config migration opportunity to rename this to "--cft-*".

achamayou (approved):

tests/infra/remote.py
jumaffre: Yep, it's on the list!

PR #3207, opened 2021-11-11
Atomic proposal application

### LINKED ISSUE #1031: Proposal completion is not atomic

achamayou:
Fix #1031.

Currently, on `apply()` failure (ie. JS exception), an unspecified amount of side-effects can have been performed already, and the details of the failure are recorded in the KV store. While this is convenient for debugging, this is very much non-atomic! Members who voted for `[a, b, c]` may only get `[a]`, or even `[a, some of b]` etc!

After this change, the entire transaction is rolled back in such cases. The submitted vote or proposal itself, as well as failure results are therefore _not_ persisted. The caller will receive error details.

After several rounds of discussion, recorded in #1031, this is thought to be the only safe way to implement atomic proposals.

Slightly more controversially perhaps, I have made the same change for `resolve()` failures. Because the side-effect situation is clean (votes and resolve are both read-only), recording the failure in the KV isn't problematic. I think I find it easier to think of all constitution code (as opposed to the user-provided ballot) as a single atomic unit, subject to failure or success? But I could be convinced otherwise.


tests/governance_js.py
```diff
@@ -507,7 +507,7 @@ def test_apply(network, args):
             "/gov/proposals",
             proposal(action("always_throw_in_apply")),
         )
-        assert r.status_code == 200, r.body.text()
+        assert r.status_code == 500, r.body.text()
```
jumaffre, line 510: Could we add a check that bundles 2+ actions together and check that if the last one fails, the proposal is still marked as `Failed`? I observed this with the bundling of the service open proposal.

src/node/rpc/member_frontend.h
```diff
@@ -985,19 +985,28 @@ namespace ccf
...
+        {
+          pi->put(
+            proposal_id,
+            {caller_identity.member_id,
+             rv.state,
+             {},
+             {},
+             std::nullopt,
+             rv.failure});
+          ctx.rpc_ctx->set_response_status(HTTP_STATUS_OK);
```
eddyashton, line 1008:
```suggestion
          ctx.rpc_ctx->set_response_status(HTTP_STATUS_INTERNAL_SERVER_ERROR);
          return;
        }
        else
        {
          pi->put(
            proposal_id,
            {caller_identity.member_id,
             rv.state,
             {},
             {},
             std::nullopt,
             rv.failure});
          ctx.rpc_ctx->set_response_status(HTTP_STATUS_OK);
          return;
```

We've been bitten before by code in endpoint handlers that looks like (/intends to) produce a final result, but doesn't return. Just adding explicit returns here for safety.

tests/governance_js.py
achamayou: Absolutely!

achamayou: I did, it didn't work, and I scratched my head for a bit trying to work out why. To save anyone else doing the same: the test constitution only applies the semantic that the proposal names suggest when they are by themselves, and defaults to voting otherwise.

achamayou:
As a result of digging into that membership test failure, I realised that while we return an HTTP server error, we also often return the current state of the proposal, which is sometimes made up (when submission failed altogether as a result of the error).

After discussion with @eddyashton, we agreed that this isn't a good idea, and that we should return OData-style responses, consistent with other errors rather than try emulate state-style responses we otherwise produce. Separately, the voting logic in infra needs fixing to deal with that.

tests/membership.py
```diff
@@ -184,7 +184,9 @@ def recovery_shares_scenario(args):
             test_remove_member(network, args, recovery_member=True)
             assert False, "Removing a recovery member should not be possible"
         except infra.proposal.ProposalNotAccepted as e:
-            assert e.proposal.state == infra.proposal.ProposalState.FAILED
+            # This is an apply() time failure, so the proposal remains Open
+            # since the last vote is effectively rolled back
```
jumaffre, line 188: "rolled back" is confusing to me here (it implies that the consensus rolled it back?). Would "discarded" be clearer?

tests/governance_js.py
```diff
@@ -507,22 +508,43 @@ def test_apply(network, args):
...
+        r = c.post(f"/gov/proposals/{proposal_id}/ballots", ballot_yes)
+        assert r.status_code == 200, r.body().text()
+
+        with node.client(None, "member1") as c:
+            r = c.post(f"/gov/proposals/{proposal_id}/ballots", ballot_yes)
+            assert r.body.json()["error"]["code"] == "InternalError", r.body.json()
+            assert (
+                r.body.json()["error"]["message"].split("\n")[0]
+                == "Failed to apply(): Error: Error message"
+            ), r.body.json()
```
jumaffre, line 537: Can we check that the proposal is reported to be in the `Open` state here?

eddyashton, line 537: Nit: I'm wary of precise asserts on error message format like this. I'd be tempted to say `assert(error_message.contains("Failed to apply()"))` and `assert(error_message.contains("Apply message"))` (and the same below), though in this case that's very similar, so not critical.

eddyashton (approved):

tests/membership.py
achamayou: Rolled back in the transaction/SQL sense, but we can call it discarded if you prefer?

tests/governance_js.py
achamayou: Can I ask you why you are wary of being precise here? I think I'd like to know if this changes (accidentally or not) in the future.

achamayou: We could, but I think we should not. That transactions returning a non 2xx aren't applied to the KV is a fundamental feature, checking that one value wasn't affected is neither a complete test for that, nor something we want to do every time we return a non-2xx.

eddyashton: I just think it's likely that at some point this does change, we introduce a newline after `apply():` for instance, and this assertion fails, and after staring at callstacks for a while we update the asserts to be `.contains()`. Making the change in advance gives us a little bit of flexibility, saving us time for that kind of incidental restructuring. We're asserting that the error message contains the things we think are important, but we're not strict in structure.

achamayou: I'd bet that day never comes, Now we'll miss changes that bork the line returns and we'll never know, but ok, I've done it.

PR #3208, opened 2021-11-11
Infra: Consider multiple read-only ledger directories

jumaffre:
Part of #2612 

Minor change to consider multiple read-only ledger directories as opposed to just one as CCF nodes accept an arbitrary number of them but the infra only coped with one so far.

### DIFF
--- a/tests/infra/network.py
+++ b/tests/infra/network.py
@@ -210,7 +210,7 @@ class Network:
         recovery=False,
         ledger_dir=None,
         copy_ledger_read_only=False,
-        read_only_ledger_dirs=[],
+        read_only_ledger_dirs=None,
         from_snapshot=False,
         snapshot_dir=None,
     ):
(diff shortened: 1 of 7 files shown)


tests/infra/network.py
```diff
@@ -542,52 +538,39 @@ def stop_all_nodes(self, skip_verification=False, verbose_verification=False):
...
-                        ) != infra.path.compute_file_checksum(
-                            os.path.join(committed_ledger_dir, ledger_file)
-                        ):
-                            raise Exception(
-                                f"Ledger file checksums between node {node_id} and most up-to-date node {most_up_to_date_node.node_id} did not match: {ledger_file}"
-                            )
+            longest_ledger_node = None
+            nodes_ledger = {}
 
+            longest_ledger_seqno = 0
```
jumaffre, line 546: Also simplify this clunky logic.

achamayou (approved):

PR #3209, opened 2021-11-11
JSON configuration for `cchost`

jumaffre:
Bulk of the work for #2612 

I am still expecting some tests to fail but the bulk of the work should be done and the PR ready for review.

Philosophy/notes:
- Group as many options as possible under common "themes".
- Re-use as many types as possible (e.g. `NodeNetworkInfo`), for consistency (e.g. storing configuration into the store).
- Mark as many fields as possible as "optional" so that we can have a minimal configuration file (see `minimal_config.json` sample). 
- We use Jinja + `config.jinja` template to create the node configuration file in the end-to-end tests.
- We still need to support the old way to pass arguments to `cchost` to be able to spin up LTS nodes (see `remote.py`).

TODO: 
- [x] Migration script from 1.x TOML to 2.x JSON
- [x] Cleanup: `interface.h`, `remote.py`
- [x] Configuration samples: full and minimal configuration samples
- [x] CFT timeout renames
- [x] IP/DNS parsing (+ remove `cli_helper.h`)
- [x] Recovery threshold
- [x] Fix sandbox in install
- [x] Add ability to read configuration file from sandbox
- [x] Fix containerised nodes `docker_wrap.sh`


src/host/main.cpp
```diff
@@ -79,10 +74,15 @@ int main(int argc, char** argv)
     return app.exit(e);
   }
 
+  LOG_INFO_FMT("Version: {}", ccf::ccf_version);
```
achamayou, line 77:
```suggestion
  LOG_INFO_FMT("CCF Version: {}", ccf::ccf_version);
```

src/host/main.cpp
```diff
@@ -79,10 +74,15 @@ int main(int argc, char** argv)
...
+
   std::string config_str = files::slurp_string(config_file_path);
   CCHostConfig config = nlohmann::json::parse(config_str);
   LOG_INFO_FMT("Configuration: {}", config_str);
 
+  LOG_FAIL_FMT(
```
achamayou, line 83: That seems oddly specific and yet context-free in the log?

jumaffre: This is now gone.

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
+{
```
jumaffre, line 1: This is the best place to look at what a configuration file looks like.

samples/config/minimal_config.json
```diff
@@ -0,0 +1,30 @@
+{
```
jumaffre, line 1: Minimal configuration sample, not to overwhelm newcomers with huge configuration files.

samples/config/minimal_config.json
```diff
@@ -0,0 +1,24 @@
...
+  "enclave_file": "lib_jsgeneric.enclave.so.signed",
+  "enclave_type": "release",
+  "network": {
+    "node_address": { "hostname": "127.0.0.1", "port": "0" },
+    "rpc_interfaces": [
+      { "rpc_address": { "hostname": "127.0.0.1", "port": "0" } }
+    ]
+  },
+  "start": {
+    "constitution_files": [],
```
jumaffre, line 11: We should probably set those.

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
...
+      "recovery_threshold": 0,
+      "max_allowed_node_cert_validity_days": 365
+    }
+  },
+  "join": {
+    "join_timer_ms": 1000,
+    "target_rpc_address": { "hostname": "127.0.0.1", "port": "0" }
+  },
+  "ledger": {
+    "ledger_dir": "ledger",
```
jumaffre, line 40: Here and elsewhere: should this simply be `"dir"` since it's already in the `"ledger"` object?

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
...
+  "tick_period_ms": 1,
+  "io_logging_threshold_ns": 10000000,
+  "client_connection_timeout_ms": 2000,
+  "node_client_interface": null,
+  "worker_threads": 0,
+  "memory": {
+    "circuit_size_shift": 22,
+    "max_msg_size_shift": 24,
+    "max_fragment_size_shift": 16
+  }
```
jumaffre, line 76: It would be nice to be able to express those as size directly. However, the numbers would be huge, so we should think of string sizes (e.g. used by CLI11) instead. In any case, this will be done in a separate PR.

doc/governance/common_member_operations.rst
```diff
@@ -15,7 +15,7 @@ Once the proposal successfully completes, the new node automatically becomes par
 Updating Code Version
 ---------------------
 
-For new nodes to be able to join the network, the version of the code they run (as specified by the ``--enclave-file``) should be first trusted by the consortium of members.
+For new nodes to be able to join the network, the version of the code they run (as specified by ``enclave_file``) should be first trusted by the consortium of members.
```
achamayou, line 18: We ought to have a page that documents each field in the config, and ``enclave_file`` would ideally link to the right entry there.

jumaffre: See `configuration.rst` for the new page. I haven't added links to all the fields on other pages, so I'll do it.

achamayou: What about writing a json schema and using https://sphinx-jsonschema.readthedocs.io/en/latest/directive.html#an-example?

achamayou: We would get links like :ref:`directive.rst#/definitions/person`, with the checking etc.

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
...
+    "snapshot_tx_interval": 10000
+  },
+  "logging": {
+    "host_log_level": "info",
+    "log_format_json": false
+  },
+  "consensus": {
+    "type": "CFT",
+    "timeout_ms": 100,
+    "election_timeout_ms": 4000
```
jumaffre, line 55: These were renamed from `raft_...`.

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
...
+  },
+  "intervals": {
+    "sig_tx_interval": 5000,
+    "sig_ms_interval": 1000,
```
jumaffre, line 59: Any objection to rename those to `sig_interval_tx/ms` to follow the existing `[parameter]-[unit]` convention?

src/enclave/rpc_sessions.h
```diff
@@ -22,6 +22,9 @@ namespace enclave
   using ServerEndpointImpl = http::HTTPServerEndpoint;
   using ClientEndpointImpl = http::HTTPClientEndpoint;
 
+  static constexpr size_t max_open_sessions_soft_default = 1000;
+  static constexpr size_t max_open_sessions_hard_default = 1010;
```
jumaffre, line 26: The defaults used to be set on the host if unspecified. This is now set here (enclave) if those have not been specified. 

Parallel questions: are these reasonable default values as they seem relatively low? See for example https://github.com/valyala/fasthttp#general-info-and-links which can apparently handle 1.5M connections,

tests/migrate_1.x_config.py
```diff
@@ -0,0 +1,176 @@
+import configparser
```
jumaffre, line 1: Script to migrate 1.x INI configuration to 2.x JSON format.

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+
+``enclave_file``
+~~~~~~~~~~~~~~~~
+
+Path to CCF enclave application.
+
+``enclave_type``
+~~~~~~~~~~~~~~~~
+
+Type of enclave application (either ``release``, ``debug`` or ``virtual``). Default value: ``release``.
```
achamayou, line 22: Every time I see several `foo_(a|b|..)`, I wonder if we shouldn't have a section `foo` with `foo.a`, `foo.b`. On the one hand, it seems cleaner and more future proof (can add fields without crowding the top level), but on the other hand, how much nesting is too much nesting? I genuinely do not know.

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+~~~~~~~~~~~
+
+The ``network`` section includes configuration for the interfaces a node listens on (both node-to-node and RPC).
+
+- ``node_address``: Address (hostname and port) to listen on for incoming node-to-node connections.
+
+- ``rpc_interfaces``: Addresses (hostname and port) to listen on for incoming client TLS connections.
+Each RPC address must contain:
+
+- The local RPC address (``rpc_address``) the node listens on.
```
achamayou, line 34: I wonder if we should go for `bind_address` and `advertised_address`?

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+.. note:: Common examples:
+
+    - A recovery member with member data: ``{"certificate_file": "member_cert.pem", "encryption_public_key_file": "member_enc_pubk.pem", "data_json_file": "member_data.json"}``
+    - A recovery member with no member data: ``{"certificate_file": "member_cert.pem", "encryption_public_key_file": "member_enc_pubk.pem"}``
+    - A non-recovery member with member data: ``{"certificate_file": "member_cert.pem", "data_json_file": "member_data.json"}``
+    - A non-recovery member with no member data: ``{"certificate_file": "member_cert.pem"}``
+
+- ``service_configuration``: Initial service configuration, including:
+
+    - ``max_allowed_node_cert_validity_days``: The maximum number of days allowed for node certificate validity period. Default value: ``365`` days. TODO: Rename
```
achamayou, line 88: rename?

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+Example:
+
+.. code-block:: json
+
+    "start": {
+        "constitution_files": ["actions.js", "validate.js", "resolve.js", "apply.js"],
+        "members": [
+            {"certificate_file": "member0_cert.pem", "data_json_file": null, "encryption_public_key_file": "member0_enc_pubk.pem"},
+            {"certificate_file": "member1_cert.pem", "data_json_file": "member1_data.json", "encryption_public_key_file": null},
+            {"certificate_file": "member2_cert.pem", "data_json_file": null, "encryption_public_key_file": "member2_enc_pubk.pem"}
```
achamayou, line 101: We can probably omit the keys for null values altogether?

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+
+- ``snapshots_dir``: Path to snapshot directory. Default value: ``snapshots``.
+- ``snapshot_tx_interval``: Minimum number of transactions between snapshots. Default value: ``10000``.
+
+``logging``
+~~~~~~~~~~~
+
+- ``host_log_level``: Logging level for the `untrusted host`. Default value: ``INFO``.
+
+.. note:: While it is possible to set the host log level at startup, it is not possible to change the log level of the enclave.
```
achamayou, line 148:
```suggestion
.. note:: While it is possible to set the host log level at startup, it is not possible to change the log level of the enclave without rebuilding it.
```

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+
+- ``log_format_json``: If ``true``, node logs will be formatted as JSON. Default value: ``false``.
```
achamayou, line 150: `logging.format` as `text|json` seems a little more flexible and concise?

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+
+``consensus``
+~~~~~~~~~~~~~
+
+- ``type``: Type of consensus protocol. Only ``CFT`` (Crash-Fault Tolerant) is currently supported in production. Default value: ``CFT``.
```
achamayou, line 155:
```suggestion
- ``type``: Type of consensus protocol. Only ``CFT`` (Crash-Fault Tolerant) is currently supported. Default value: ``CFT``.
```

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+- ``timeout_ms``: Hearbeat interval (ms) at which primary node sends messages to backup nodes to maintain primary-ship. This should be set to a significantly lower value than ``election_timeout_ms``. Default value: ``100`` ms.
```
achamayou, line 156: Heartbeats aren't quite the whole story, it's also how often the primary retries sending messages that haven't been acked.

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+- ``election_timeout_ms``: Timeout value (ms) after which backup node that have not received primary heartbeats will trigger a new election. Default timeout: ``4000`` ms.
```
achamayou, line 157:
```suggestion
- ``election_timeout_ms``: Timeout value (ms) after which backup node that have not received any message from the primary will trigger a new election. Default timeout: ``4000`` ms.
```

achamayou: Also probably worth mentioning that this should be set to be substantially larger than `timeout_ms`.

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+``intervals``
+~~~~~~~~~~~~~
+
+- ``sig_tx_interval``: Number of transactions after which a signature transaction is automatically generated. Default value: ``5000``.
+- ``sig_ms_interval``: Maximum duration (milliseconds) after which a signature transaction is automatically triggered. Default value: ``1000`` ms.
+
+.. note::
+    Transaction commit latency in a CCF network is primarily a function of signature frequency. A network emitting signatures more frequently will be able to commit transactions faster, but will spend a larger proportion of its execution resources creating and verifying signatures. Setting signature frequency is a trade-off between transaction latency and throughput.
+
+    The signature interval options specify the intervals at which the generation of signature transactions is `triggered`. However, because of the parallel execution of transactions, the actual intervals between signature transactions may be slightly larger.
```
achamayou, line 168: It's not just parallel execution, it's also queuing.

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+
+    "join": {
+        "join_timer_ms": 1000,
+        "target_rpc_address": {"hostname": "127.0.0.1", "port": "8080"}
+    }
+
+``ledger``
+~~~~~~~~~~
+
+- ``ledger_dir``: Path to main ledger directory. Default value: ``ledger``.
```
achamayou, line 133: `ledger` feels redundant here, perhaps `append_directory` and `read_only_directories` (and `chunk_size_bytes`?)

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
...
+- ``read_only_ledger_dirs``: Optional. Paths to read-only ledger directories. Note that only ``.committed`` files will be read from these directories. Default value: ``[]``.
+- ``ledger_chunk_bytes``: Minimum size (bytes) of the current ledger file after which a new ledger file (chunk) is created. Default value: ``5000000`` bytes.
+
+``snapshots``
+~~~~~~~~~~~~~
+
+- ``snapshots_dir``: Path to snapshot directory. Default value: ``snapshots``.
+- ``snapshot_tx_interval``: Minimum number of transactions between snapshots. Default value: ``10000``.
```
achamayou, line 141: Same as for ledger, these names feel redundant, and `interval_size_tx` would be somewhat more consistent with other options (unit at the end).

doc/operations/start_network.rst
```diff
@@ -13,41 +13,23 @@ To create a new CCF network, the first node of the network should be started wit
...
-
 The unique identifier of a CCF node is the hex-encoded string of the SHA-256 digest the public key contained in its identity certificate (e.g. ``50211327a77fc16dd2fba8fae5fffac3df909fceeb307cf804a4125ae2679007``). This unique identifier should be used by operators and members to refer to this node with CCF (for example, when :ref:`governance/common_member_operations:Trusting a New Node`).
 
-.. note:: The network certificate should be distributed to users and members to be used as the certificate authority (CA) when establishing a TLS connection with any of the nodes part of the CCF network. When using ``curl``, this is passed as the ``--cacert`` argument.
+CCF nodes can be started by using IP Addresses (both IPv4 and IPv6 are supported) or by specifying a fully qualified domain name. If an FQDN is used then a ``dNSName`` subject alternative name should be specified as part of the ``node_certificate.subject_alt_names`` configuration entry. Once a DNS has been setup it will be possible to connect to the node over TLS by using the node's domain name.
```
achamayou, line 26: We ought to have one end to end test that uses IPv6 if we make this explicit claim.

doc/operations/configuration.rst
```diff
@@ -0,0 +1,235 @@
+Configuration
```
eddyashton, line 1: Comment: We've leaned hard into using `configuration` to describe sets of active nodes, and the adding/removing node protocols. This was fine while we started a node with a set of "options", but is confusing if we start a node with a "configuration". But it's going to confuse somebody that `configuration.rst` has nothing to do with `reconfiguration.rst` (or currently, `1tx-reconfig.rst` and `2tx-reconfig.rst`). In particular, if we dump a portion of this configuration file inside the KV, and think some of these settings are updateable by governance, what is _that_ called?

To be clear, I think this new use is sensible and absolutely the right term, matching common uses of the configure/configuration elsewhere, and I'm not suggesting a rename.

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
+{
+  "enclave_file": "lib_jsgeneric.enclave.so.signed",
```
jumaffre, line 2: Use new "enclave" section.

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
...
+    ],
+    "service_configuration": {
+      "recovery_threshold": 0,
+      "max_allowed_node_cert_validity_days": 365
+    }
+  },
+  "join": {
+    "join_timer_ms": 1000,
+    "target_rpc_address": { "hostname": "127.0.0.1", "port": "0" }
+  },
```
jumaffre, line 38: Add start and join schemas.

samples/config/minimal_config.json
```diff
@@ -0,0 +1,29 @@
...
+      "validate.js",
+      "apply.js",
+      "resolve.js",
+      "actions.js"
+    ],
+    "members": [
+      {
+        "certificate_file": "member0_cert.pem",
+        "encryption_public_key_file": "member0_enc_pubk.pem",
+        "data_json_file": ""
```
jumaffre, line 21: Remove this.

samples/config/minimal_config.json
jumaffre: Minimal example for "start" and "join".

samples/config/minimal_config.json
```diff
@@ -0,0 +1,29 @@
+{
+  "enclave_file": "lib_jsgeneric.enclave.so.signed",
+  "enclave_type": "release",
+  "network": {
+    "node_address": { "hostname": "127.0.0.1", "port": "0" },
```
jumaffre, line 5: Merge "hostname" and "port".

samples/config/config.json
```diff
@@ -0,0 +1,77 @@
...
+  "intervals": {
+    "sig_tx_interval": 5000,
+    "sig_ms_interval": 1000,
+    "jwt_key_refresh_interval_s": 1800
+  },
+  "network_certificate_file": "networkcert.pem",
+  "node_certificate_file": "node.pem",
+  "node_pid_file": "node.pid",
+  "node_address_file": "node.node_address",
+  "rpc_addresses_file": "node.rpc_addresses",
```
jumaffre, line 66: These could be removed if we dump this information, along with the unique node ID, to syslog or similar.

jumaffre, line 66: These could be removed if we dump this information, along with the unique node ID, to syslog or similar.

doc/operations/configuration.rst
jumaffre: `chunk_size` as the unit will be in value string.

doc/overview/threading.rst
```diff
@@ -17,7 +17,10 @@ Implementation
 Configuration
 ~~~~~~~~~~~~~
 
-To enable multiple worker threads, pass the ``--worker-threads=`` flag along with the number of desired threads to ``cchost`` when starting a CCF node.
+.. warning:: Multiple enclave threads are expiremental for now. The ``worker_threads`` configuration option should be set to ``0`` in production environments.
```
achamayou, line 20:
```suggestion
.. warning:: Multiple enclave threads are experimental for now. The ``worker_threads`` configuration option should be set to ``0`` in production environments.
```

doc/overview/threading.rst
```diff
@@ -17,7 +17,10 @@ Implementation
...
+
+To enable multiple worker threads, the ``worker_threads`` configuration option can be used  along with the number of desired threads to ``cchost`` when starting a CCF node.
```
achamayou, line 22:
```suggestion
To enable multiple worker threads, the ``worker_threads`` configuration option can be used along with the number of desired threads to ``cchost`` when starting a CCF node.
```

tests/e2e_logging.py
```diff
@@ -1342,14 +1342,14 @@ def run(args):
...
 
-    cr.add(
-        "js",
-        run,
-        package="libjs_generic",
-        nodes=infra.e2e_args.max_nodes(cr.args, f=0),
-        initial_user_count=4,
-        initial_member_count=2,
-    )
+    # cr.add(
```
achamayou, line 1345: Don't forget ;)

tests/infra/e2e_args.py
```diff
@@ -362,6 +320,11 @@ def cli_args(add=lambda x: None, parser=None, accept_unknown=False):
         default="1tx",
         choices=("1tx", "2tx"),
     )
+    parser.add_argument(
+        "--config-file",
+        help="Absoliute path to node JSON configuration file",
```
achamayou, line 325:
```suggestion
        help="Absolute path to node JSON configuration file",
```

achamayou (approved):

doc/operations/start_network.rst
jumaffre: See #3215

doc/operations/configuration.rst
jumaffre: Agreed. As discussed, there isn't an obvious alternative name here ("membership" is already used for consortium members) - perhaps "servicing a network" for what we currently call node reconfiguration? I'm happy to revisit this later and upgrade our docs once #2612 is done.

samples/config/minimal_config.json
jumaffre: Will do in a follow-up PR: https://github.com/microsoft/CCF/issues/2612#issuecomment-973067527

samples/config/config.json
jumaffre: Will do in a follow-up PR: #2612 (comment)

samples/config/config.json
jumaffre: Follow-up in https://github.com/microsoft/CCF/issues/2612#issuecomment-973067527

doc/governance/common_member_operations.rst
jumaffre: Follow up in https://github.com/microsoft/CCF/issues/2612#issuecomment-973067527

doc/operations/configuration.rst
jumaffre: Done. Renamed to `bind_address` and `published_address`.

src/enclave/rpc_sessions.h
jumaffre: To be addressed by https://github.com/microsoft/CCF/issues/2612#issuecomment-973067527

PR #3210, opened 2021-11-12
Cleanup Raft implementation

eddyashton:
This PR removes some unused features and code from our Raft implementation.


src/consensus/aft/raft.h
```diff
@@ -1200,7 +908,11 @@ namespace aft
           }
         }
       }
-      else if (consensus_type != ConsensusType::BFT)
+      else if (
+        consensus_type != ConsensusType::BFT &&
+        replica_state != kv::ReplicaState::Retired &&
+        replica_state != kv::ReplicaState::Retiring &&
+        replica_state != kv::ReplicaState::Learner)
```
jumaffre, line 915: Is this a functional change for the CFT path?

src/node/rpc/frontend.h
```diff
@@ -547,131 +547,11 @@ namespace ccf
...
-            ctx->set_error(
-              HTTP_STATUS_INTERNAL_SERVER_ERROR,
-              ccf::errors::InternalError,
-              "Consensus is not yet ready.");
-            return ctx->serialise_response();
-          }
-        }
-      }
-
+      // NB: If we want to re-excute on backups, the original command could be
```
jumaffre, line 550:
```suggestion
      // NB: If we want to re-execute on backups, the original command could be
```

src/consensus/aft/raft.h
eddyashton: Huh, this looks like a bad merge, and is not deliberate. I'll take a look.

wintersteiger: Those conditions are unnecessary, because `can_endorse_primary` also checks them.

wintersteiger (approved):

src/consensus/aft/raft.h
eddyashton: Yeah this was a bad merge. Have restored this check to the `else if (consensus_type != ConsensusType::BFT)` that is on `main`.

achamayou (approved):

PR #3214, opened 2021-11-15
Update signed requests tests and documentation

### LINKED ISSUE #1886: Update HTTP signature

achamayou:
Close #1886 by addressing the remaining items (tests and doc).


doc/use_apps/issue_commands.rst
```diff
@@ -31,7 +31,8 @@ Signing
 -------
 
 In some situations CCF requires signed requests, for example for member votes.
-The signing scheme is compatible with the `IETF HTTP Signatures draft RFC <https://tools.ietf.org/html/draft-cavage-http-signatures-12>`_.
+The signing scheme is compatible with the `IETF HTTP Signatures draft RFC <https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-08>`_,
```
jumaffre, line 34: Since we only use/test `hs2019`, should we say that we are compatible with draft-12 only?

achamayou: Having reviewed the diffs, I do not think we are compatible with draft-12, specifically because of the following points:

1. `algorithm` became optional in draft-10: https://www.ietf.org/rfcdiff?url2=draft-cavage-http-signatures-10.txt, we do not treat it as such.
2. draft-11 introduced optional `(created)` and `(expired)`: https://www.ietf.org/rfcdiff?url2=draft-cavage-http-signatures-11.txt which we do not support

eddyashton (approved):

PR #3217, opened 2021-11-17
TLS Cleanup + OpenSSL Base64

rengolin:
In order to implement TLS in OpenSSL as a unit, I'm cleaning up all non-TLS-specific functionality that was in the `tls` directory, namely `error_string` and `base64`. Those are moving to `crypto` because they're used mostly there (`crypto` even had its own copy of `error_string`).

In addition, I'm throwing in the OpenSSL implementation of Base64 that I did as a research on how to move TLS to OpenSSL. Turns out I can't use the same template as `crypto` (that worked with `base64`) because all TLS classes (`Context`, `CA`, `Cert`, etc) are linked in a single implementation.

Furthermoire, OpenSSL's own structure and API is distinct enough from MbedTLS to make it hard to common them up at a higher level. For example, the use of `ssl` and `cfg` by MbedTLS doesn't correspond to the same API usage as `SSL` and `SSL_CTX` in OpenSSL.

Given the additional complexity, and the fact that I already had a `base64` implementation lying around, I decided to move them to `crypto` and work with only TLS structures as a whole in the TLS re-implementation (for another PR), which needed a cleanup first (this PR).

Also, it may be easier to review one commit at a time. I spent some good time splitting them into logical snippets.


rengolin:
Also, it may be easier to review one commit at a time. I spent some good time splitting them into logical snippets.

achamayou:
@rengolin can you add a picobench to compare the base64 impl for mbedtls and openssl? Something like https://github.com/microsoft/CCF/blob/main/src/crypto/test/digest_bench.cpp

rengolin:
Updated `crypto_test` with Base64 benchmarks:
image

src/crypto/openssl/base64.h
```diff
@@ -0,0 +1,120 @@
...
+      unsigned char output[max_size];
+      memset(output, '\0', max_size);
+
+      // Decode
+      int chunk_len = 0;
+      int rc = EVP_DecodeUpdate(ctx, output, &chunk_len, data, size);
+      if (rc < 0)
+      {
+        auto err = ERR_get_error();
+        char err_str[256];
```
achamayou, line 44: I would be happier if this was OPENSSL_MAX_ERROR_MSG_SIZE or something like that.

achamayou (approved):

src/crypto/openssl/base64.h
rengolin: Good point. The docs proposed 256, but using the right macro is definitely better.

PR #3218, opened 2021-11-17
Raft state refactoring

wintersteiger:
This adds the raft state refactoring into separate leadership and membership states. There are two points I'm not happy with that I'll point out in comments.


src/consensus/aft/raft.h
```diff
@@ -1174,12 +1199,14 @@ namespace aft
       }
 
       // Then check if those append entries extend past our retirement
-      if (
-        retirement_committable_idx.has_value() &&
-        r.idx > retirement_committable_idx && !is_retiring())
+      if (is_retired() && retirement_phase >= kv::RetirementPhase::Completed)
```
wintersteiger, line 1202: I believe this condition is too strong, but I got into trouble with the 1tx scheme without this. Shouldn't it be `>= ...::Signed`?

src/consensus/aft/raft.h
```diff
@@ -475,22 +481,29 @@ namespace aft
...
-        if (!new_learners.empty())
+        if (
+          !configurations.empty() &&
+          configurations.back().nodes.find(state->my_node_id) ==
+            configurations.back().nodes.end() &&
+          conf.find(state->my_node_id) == conf.end() &&
+          (is_retiring() ||
+           (is_retired() &&
+            retirement_phase == kv::RetirementPhase::Committed)))
+        {
```
wintersteiger, line 497: I believe this condition is correct, but it's super ugly. I was hoping that it would be the same as for 1tx, but it's not. At the time we transition from `RETIRING` to `RETIRED`, the node is indeed not in `configurations.back()` anymore.

src/consensus/aft/raft.h
```diff
@@ -171,13 +173,17 @@ namespace aft
...
+        kv::LeadershipState::Follower,
+      kv::MembershipState initial_membership_state_ =
+        kv::MembershipState::Active,
       ReconfigurationType reconfiguration_type_ =
         ReconfigurationType::ONE_TRANSACTION) :
       consensus_type(consensus_type_),
       store(std::move(store_)),
 
-      replica_state(initial_state_),
+      leadership_state(initial_leadership_state_),
```
achamayou, line 185: Shouldn't our initial leadership state be unset, and only be set once we transition to Active? I thought this was the whole reason it was optional.

src/consensus/aft/raft.h
achamayou: That is super ugly... :(

src/consensus/aft/raft.h
achamayou: No, we still want to ack the signature over our retirement (because we may hold up commit otherwise), just nothing (strictly) further.

src/consensus/aft/raft.h
wintersteiger: Yes, for 2tx that's the case (`node_state.h:1976`), but for 1tx I kept the previous default. You're right though, I'll change that and check nothing breaks.

src/consensus/aft/raft.h
wintersteiger: I think there may be a better way, if I pass the retiring states instead of the retired ones into `add_configuration`, but it may also create ugliness elsewhere :-/

wintersteiger: Gave this another try today and it's possibly, but it creates a bunch of complications in `commit`. The core of the problem is that for every actual reconfiguration I get two calls to `add_configuration` as there are two updates to the nodes table and I have to figure out which one's which afterwards. I think we should go ahead with the current version of the ugliness until I work out all of the details for the other one.

achamayou (approved):

doc/schemas/node_openapi.json
```diff
@@ -65,17 +65,23 @@
...
+          },
+          "retirement_phase": {
+            "$ref": "#/components/schemas/RetirementPhase"
           }
         },
         "required": [
           "configs",
           "acks",
-          "state"
+          "membership_state"
```
jumaffre, line 84: This looks like a breaking change, which should at least be mentioned in the changelog. Was the `state` field introduced after 1.x?

src/consensus/aft/raft.h
```diff
@@ -171,13 +173,17 @@ namespace aft
       std::chrono::milliseconds view_change_timeout_,
       size_t sig_tx_interval_ = 0,
       bool public_only_ = false,
-      kv::ReplicaState initial_state_ = kv::ReplicaState::Follower,
+      std::optional<kv::LeadershipState> initial_leadership_state_ =
+        std::nullopt,
```
jumaffre, line 177: Looking at `node_state.h`, this argument always seems to be set to `std::nullopt` anyway. Is it worth removing it?

doc/schemas/node_openapi.json
wintersteiger: @Amaury said so, but I haven't checked. I think the whole consensus details endpoint was introduced after 1.0 though.

src/consensus/aft/raft.h
wintersteiger: I kept it just in case, but I don't like it either. Easy enough to re-introduce if it ever becomes necessary, so I'll kill it.

doc/schemas/node_openapi.json
Amaury: Not the right Amaury I guess...

wintersteiger: Sorry about that! I wanted @achamayou :-)

PR #3219, opened 2021-11-18
More Raft cleanup

eddyashton:
Followup to #3210.

Removes tracking of conflict versions. This is an optimisation for parallel backup execution but is currently unused, so removing to simplify the KV conflict resolution and replication code.


src/kv/generic_serialise_wrapper.h
```diff
@@ -58,17 +57,16 @@ namespace kv
...
-      max_conflict_version(max_conflict_version_),
       is_snapshot(is_snapshot_),
       crypto_util(e)
     {
       set_current_domain(SecurityDomain::PUBLIC);
       serialise_internal(is_snapshot);
       serialise_internal(tx_id.version);
-      serialise_internal(max_conflict_version);
+      // TODO: Remove?
+      serialise_internal((Version)0u);
```
jumaffre, line 69: I'm assuming we still need this for backwards compatibility?

jumaffre (approved):

achamayou (approved):

Issue #3076, opened 2021-10-12
Add index to speed up sparse historical queries


eddyashton:
Historical range queries currently scan an entire range of ledger entries, fetching each from disk, through the ringbuffer, deserialising and verifying, and presenting to the application. In practice, the application is often interested in only some of these entries, those which affect specific keys. As those interesting entries become more sparse, this process becomes increasingly inefficient - we spend a lot of time processing entries that were not needed, and are instantly discarded by the application logic. Another way of phrasing this is that the performance of historical range queries is affected by the total size of the ledger, rather than the size of the target entries, which means it can degrade significantly in a way that is opaque to an individual user (_their_ response moved from 10 to 11 entries, yet got 1 million times slower!?)

We want to add an indexing mechanism to remove this limitation, so that only the desired entries are fetched through the pipeline and presented to the app. It is critical that this still retrieves _all_ relevant entries - a core advantage of doing historical queries in-enclave is the guarantee that this will not omit entries from the response.

Consider a ledger containing the following entries:
```
1: x=a
2: x=b
3: x=c
4: y=d
5: x=e
6: y=f
7: x=g
8: x=h
```

To do a historical query for "all the entries that wrote to key x" or "all the entries that wrote to key y", we must look at all 8 entries. However, if we built an index (in the KV, so that it is trusted and useable by all nodes) of all the previous writes (using `get_version_of_previous_write` whenever we write to a key), we could do much better:
```
1: x=a
2: x=b, x'=[1]
3: x=c, x'=[1,2]
4: y=d
5: x=e, x'=[1,2,3]
6: y=f, y'=[4]
7: x=g, x'=[1,2,3,5]
8: x=h, x'=[1,2,3,5,7]
```

If we maintained this index forever, we'd have a perfect history of every time each key was written to. In practice we only want this for some tables, but it probably makes sense to do it for every key _in_ that table, so can be implemented as a wrapper handle. But to maintain it forever, we have infinite memory growth, where the app is likely overwriting keys and using historical queries explicitly to avoid that memory growth.

So instead we need to bound how large that index can grow. I think the sensible solution is to chunk it, deleting the index when it grows too large and maintaining a single value in _another_ metatable, pointing at the last version where the index was present:
```
# Chunking at-most 2 entries in an index
1: x=a
2: x=b, x'=[1]
3: x=c, x'=[1,2]
4: y=d
5: x=e, x'=[], x''=3
6: y=f, y'=[4]
7: x=g, x'=[5]
[...]

PR #3221, opened 2021-11-18
Historical queries: Add support for fetching sets of seqnos

eddyashton:
TL;DR: You could previously say `get_state_at(5)`, or `get_state_range(5, 10)`, but you can now say `get_states_for({5, 6, 7, 100})`.

This is a necessary part of #3076, where we'll eventually have an index telling us the set of seqnos to fetch for a specific query, and want to do so efficiently (with a single handle!).

Naming is hard, so sometimes I call this a `set`, and a `sparse set` to indicate it's likely not a single contiguous range, but then I store it in a thing called a `ContiguousSet` so it's faster where it _is_ a single range, and then in the temporary sample endpoint it's called `multi`. Suggestions welcome.

This PR contains:
- A new container in `ds/contiguous_set.h`, and associated unit tests. It looks a bit like a `std::set`, but should only hold numbers, and densely packs descriptions of the sub-ranges it contains. This is to simplify the transition - the old `get_state_range` asks for _one_ of these, which we don't want to make unwieldily slow, so we represent the new set as a collection of these. It's likely that fetching ranges of entries continues to be a valuable speedup, so I expect to re-use this for the indexing.
- New methods on `HistoricalStateCache` for requesting sets of seqnos.
- Modified implementation of `HistoricalStateCache` to handle requests for non-contiguous sets. This got pretty hairy, as it breaks a lot of our existing assumptions about how to find a signature, but I believe it's also ironed out some edge cases from the existing range query API.
- Extended unit tests for historical queries. Part of this is just testing the new "sparse query" API in isolation, but I also rolled this into the `StateCache concurrent access` test which both stresses the thread-safety of accessing the `StateCache`, and simulates callers making repeated requests for different historical entries. The latter test has been refactored to avoid a bunch of code re-use, and the behaviour's improved slightly in that each worker now mixes different types of requests over the same handle. There's a handful of hard-coded cases that were previously failing, kept as regression tests.
- A sample endpoint in `logging.cpp` (`GET /log/private/historical/multi`), and e2e test that calls this. This is just to confirm that the API works as intended end-to-end, but isn't a useful example of how it should actually be exposed as a HTTP endpoint. I expect to remove this once the indexing is available, and the underlying API is stressed through existing calls rather than exposed directly.


src/node/historical_queries.h
```diff
@@ -17,6 +17,12 @@
 #include <memory>
 #include <set>
 
+#ifdef ENABLE_HISTORICAL_VERBOSE_LOGGING
+#  define HISTORICAL_LOG(...) LOG_INFO_FMT(__VA_ARGS__)
```
eddyashton, line 21: I kept adding and removing some verbose logging in this file. I don't think it's useful to keep at `TRACE`, but don't want to lose it entirely, so instead its disabled by this extra switch, intended to be turned on locally for debugging the unit tests. This is a great argument for custom/scoped loggers!

src/node/historical_queries.h
```diff
@@ -508,24 +590,15 @@ namespace ccf::historical
 
           if (request.include_receipts)
           {
-            const auto result = request.populate_receipts(seqno);
-            switch (result)
+            const auto next_seqno = request.populate_receipts(seqno);
```
eddyashton, line 593: This used to return an enum, because if it wanted you to fetch another entry (as a guess at a supporting signature), that was always the one after we'd currently looked at. That's no longer true!

Explanation: If we already have `{7, 8, 9}` (and maybe we already have receipts for them too), and we're now fetching `{4, 7, 8, 9}` (with receipts), and the only signature is at 11, then here's what happens:
- We ask for 4
- We get 4, validate it, but it's not a signature
- Looking for a receipt for 4, we check 7, 8, and 9 since we already have them, just in-case.
- None of those are a signature, so we fetch 5 as a supporting signature.
- We get 5, validate it, but it's not a signature.
- Still looking for a receipt for 4, our next best guess is 6
- We get 6, validate it, but it's not a signature.
- Still looking for a receipt for 4, our next best guess is _10_. In `populate_receipts`, we examined 6 (that we just got), and 7, 8, and 9 (that we already had), so returned `next_seqno = 10`.
- We get 10, validate it, but it's not a signature
- Still looking for a receipt for 4, our next best guess is 11.
- We get 11, validate it, it's a signature! Populate receipt for 4, and now we can answer the original state query.

Anyway all of this mess is only necessary because we're still linearly scanning for signatures. It should be possible to do this far more efficiently - either asking the host to return the next signature and letting it scan (potentially using ledger header bits), or indexing signatures internally,

src/node/history.h
```diff
@@ -55,7 +55,11 @@ namespace ccf
     COMPACT
   };
 
+#ifdef OVERRIDE_MAX_HISTORY_LEN
+  constexpr int MAX_HISTORY_LEN = OVERRIDE_MAX_HISTORY_LEN;
+#else
   constexpr int MAX_HISTORY_LEN = 1000;
+#endif
```
eddyashton, line 62: This is now overridden by the `historical_queries` unit test. The problem is that with the default value, when we only create a few dozen entries, _any_ signature contains a minitree that covers every single transaction. This gets in the way of testing that we're always finding a signature! Note that this history length actually introduces a really nasty subtlety in the historical queries - the receipt we return _may not be the canonically shortest receipt_ that you'd get previously, instead it's just _a_ receipt from a future signature transaction, but if we're unlucky (in the order we receive ledge entries) it may be 3 signatures in the future rather than the _next_ signature. I think we can avoid that if we're careful, but haven't confirmed that since I'm not sure if the canonical receipt concept is actually necessary.

tests/e2e_logging.py
```diff
@@ -794,6 +794,102 @@ def id_for(i):
...
+        n_entries = 100
+
+        for _ in range(n_entries):
+            network.txs.issue(
+                network,
+                repeat=True,
+                idx=idx,
+                wait_for_sync=False,
+                log_capture=[],
+                send_public=False,
```
eddyashton, line 819: I've added this `send_public=False` option because what I'd like to do here is submit a huge number of entries, and then check that fetching a few small sub-ranges is relatively fast, even if we do so in a single request. But if we interleave public and private entries (as we do by default, submitting one of each in each iteration), then we'll never get contiguous ranges of private seqnos to fetch.

samples/apps/logging/logging.cpp
```diff
@@ -1072,6 +1072,185 @@ namespace loggingapp
           ccf::endpoints::ExecuteOutsideConsensus::Locally)
         .install();
 
+      static constexpr auto get_historical_multi_path =
+        "/log/private/historical/multi";
```
achamayou, line 1076: I think `sparse` would be a little more accurate than `multi`, which could potentially include ranges, but naming etc.

src/ds/contiguous_set.h
```diff
@@ -0,0 +1,338 @@
...
+
+    // Define an iterator for accessing each contained element, rather than the
+    // ranges
+    template <typename RangeIt>
+    struct TIterator
+    {
+      RangeIt it;
+      size_t offset = 0;
+
+      // clang-format off
```
achamayou, line 30: why?

src/node/history.h
achamayou: Until we've worked out if canonical receipts are particularly interesting somehow, this is a good comment to add but not an actual issue.

tests/e2e_logging.py
achamayou: I continue to be skeptical of the value of the TxIssuer, and I worry that we're headed down the flag explosion route. It seems to me that it's simple enough to submit what we want, and check for commit of the last tx that it shouldn't need a wrapper.
If we namespace the keys with the test name, this ought to be absolutely fine.

achamayou (approved):

src/node/historical_queries.h
achamayou: Yes, that's definitely something we should look into after 2.0.

src/ds/contiguous_set.h
eddyashton: Was trying to keep this dense when there were a lot of methods. It's small enough now, so I've restored the standard formatting.

PR #3224, opened 2021-11-19
Refactor Unique_OBJ classes in openssl_wrappers

rengolin:
When adding a few more unique pointers to that class, I realised the
patterns were so close to each other that we could just do with a
template wrapper and simplify the code repetition.

This change is a non-functional but should make it easier to add new
unique pointers here without repeating too much.


rengolin:
Some considerations:

Now all `Unique_*` pointers have a `release` function. This was only the case for `BIGNUM`. If that was by design, then I should remove it from the templated version and move it back to `Unique_BIGNUM`. 

I haven't done so because this is a standard `unique_ptr` functionality and any user of `unique_ptr` has to worry when call that function. But if the main point of the wrappers was to hide that, then we should still hide it.

I don't mind either way.

Benchmark numbers shouldn't have changed at all, but:
* CHAMP (get) is worse than the recent average, but not than an older data point. Could be a fluke.
* tpcc_sgx_cft_mem is better than recent average, but not inconsistent with recent "betters".

I think this is all fine, let me know if you disagree.

src/crypto/openssl/openssl_wrappers.h
```diff
@@ -47,245 +54,184 @@ namespace crypto
...
         CHECKNULL(p.get());
       }
-      Unique_BIO(const void* buf, int len) :
-        p(BIO_new_mem_buf(buf, len), [](auto x) { BIO_free(x); })
+      /// C-tor with pointer created in base class
+      Unique_SSL_OBJECT(T* ptr, void (*dtor)(T*), bool check_null = true) :
+        p(ptr, dtor)
       {
-        CHECKNULL(p.get());
+        if (check_null)
```
wintersteiger, line 90: I don't understand the reason for this branch, shouldn't we always check that the underlying object isn't null?

wintersteiger (approved):

src/crypto/openssl/openssl_wrappers.h
rengolin: Not always. The cases where I override with `false` were the cases that didn't check in the past.

Special mention to `UniqueX509` which had a comment:
```
// p == nullptr is OK (e.g. wrong format)
```

Note, this is *only* for cases where the derived class provides the pointer, not on the case we construct the object in the templated base class.

wintersteiger: Right, that's correct, in those cases we do indeed not want that check!

achamayou (approved):

Issue #3225, opened 2021-11-19
Replace `proposal_generator` with templates


eddyashton:
We currently maintain a `proposal_generator` CLI written in Python. This is really part of the infra that is used to run test networks, with an auto-generated CLI layer on top which is largely untested. We believe this is unnecessary, as proposals are simple enough that they could instead be provided as templates, rather than generated from scripts. We should implement these as templates, and then remove the current CLI tool.

PR #3229, opened 2021-11-22
Add jinja templates for proposals and ballots

eddyashton:
Part of #3225, though still retaining the frontend CLI for now.

Rather than generating proposals and ballots in a Python script, generate them from Jinja templates. I believe these templates properly support multiple actions in a single proposal, although that's still relatively untested since it didn't work well with the previous generator.


eddyashton:
I think this is a useful step, but I'm not entirely sure how to close the final gap. The frontend we currently have is good for enforcing type safety, and discoverability of proposals. We'd like to make the canonical source of that `actions.js`, but perhaps that loses something and should be derived from some other spec that could more usefully be used to generate proposals?

And the other 'tricky' thing that the frontend currently does is convert files (and in the case of `set_js_app`, entire directories), into objects that we can then JSONify. We could do all of that through jq. The current approach uses that initial object representation to construct both a proposal and a ballot, but perhaps we should focus just on the proposal and instantiate the ballot template _from the proposal_. That should be an easy transformation, and then votes can be generated for any proposal (however it was created!), not just in-sync with the proposal generation.

achamayou (approved):

jumaffre (approved):

PR #3236, opened 2021-11-23
Isolate TlsClient's CA and Cert

rengolin:
TlsClient is a whole new implementation of a TLS client and reused
src/tls/{ca|cert}.h. With the migration to OpenSSL, those definitions
conflict when we set TLS_PROVIDER_IS_OPENSSL=True.

This is a stop-gap change to make sure we only work on one problem at a
time, in this case, tls_endpoint. Once that's out of the way, we also
need to migrate the tls_client to OpenSSL, which is a whole new thing,
as it uses sockets, so the API will likely change everywhere.


achamayou (approved):

eddyashton (approved):
LGTM!

Only suggestion is to move these copies from the `tls` namespace to something else (`client::tls`?) so it's clear they're separate, and then everything under the `tls` namespace is consistently backed by _either_ mbedtls _or_ openssl.

PR #3237, opened 2021-11-23
Update reconfiguration docs

wintersteiger:
This updates the reconfiguration documentation to the new state names/enums.


doc/overview/consensus/1tx-reconfig.rst
```diff
@@ -92,3 +92,34 @@ Instead, upon retiring from a network, retired leaders still respond to requests
...
+            Started-- RTX commits -->Ordered;
+            Ordered[Ordered: RI set]
+            Ordered-- Signature -->Signed;
+            Signed[Signed: RCI set]
+            Signed-- RCI commits -->Completed;            
+            Ordered-.->Started
+            Signed-.->Ordered
+        end
+
+Until the very last phase (``Completed``) is reached, a retiring leader will not actively attempt to change leadership, but elections may still happen as usual.
```
achamayou, line 122: This is a slightly odd sentence, because a leader never attempts to change leadership in Raft (or in our CFT). I think "will continue to act as leader" is better, although there is a slight caveat: a retiring leader having observed RCI on itself will not execute further transactions, but will continue to send append entries/heartbeats.

doc/overview/consensus/index.rst
```diff
@@ -28,52 +28,40 @@ By default CCF runs with CFT **and BFT is disabled on release versions**. To run
 Replica State Machine
 ---------------------
 
-Simplified
+Membership
 ~~~~~~~~~~
 
-Main states and transitions in CCF consensus. Note that while the implementation of the transitions differs between CFT and BFT, the states themselves do not.
+Any node of the network is always in one of four membership states. When using One-transaction reconfiguration, the ``Learner`` and
```
jumaffre, line 34: Could you point to the `/node` endpoint that returns this state here?

achamayou (approved):

doc/overview/consensus/index.rst
```diff
@@ -29,55 +29,52 @@ By default CCF runs with CFT **and BFT is disabled on release versions**. To run
...
 .. mermaid::
 
-    graph LR;
-        Follower-->FRI[Follower w/ RI];
-        FRI-->Follower;
-        FRI-->FRCI[Follower w/ RCI];
-        FRCI-->Follower;
-        FRCI-->Retired;
+    graph LR;        
+        Candidate-->Follower;
```
jumaffre, line 62: Leader to Candidate is not a valid transition. Follower to Follower and Candidate to Candidate are also valid transitions. See Raft paper:
image

wintersteiger: Great, thanks for the reference! We want exactly the same I think.

Issue #3088, opened 2021-10-13
KV: support for ordered maps


jumaffre:
The underlying implementation of the a key-value map currently uses a CHAMP, which is fast compared to the original Red-Black tree implementation (see benchmark results: https://github.com/microsoft/CCF/issues/2814#issuecomment-940099829), but does not provide ordered access to the element stored in it. This is an issue for determinist execution (see https://github.com/microsoft/CCF/issues/2814) and would also prevent us from implementing an efficient KV range API. Moreover, the performance difference between the two implementations do not seem to substantially affect the end-to-end performance of CCF (especially that the RB map doesn't seem to be as optimised as it could be):

<details><summary>Show cimetrics output (recorded here for posterity) </summary>
<p>

![image](https://user-images.githubusercontent.com/42961061/137143003-d64db697-05e0-484b-b6f4-4051f7d9808a.png)

</p>
</details>

We should replace the CHAMP with the RB map, and do the following (in this order): 
- [x] 1. Add comparison benchmark between `std::map` (and `std::unordered_map`) to the existing `map_bench`, so that we assess how much optimisation we can aim for for the RB map.
- [ ] 2. Optimise RB map implementation accordingly (in particular the calls to `left()` and `right()` which copy the root pointers).
- [x] 3. Add snapshotting capability for RB map (with backwards compatibility with snapshots created in 1.x). 
- [ ] 4. Improve `foreach()` for RB map, using iterators to be able to deal with committed state vs. local writes. 
- [ ] 5. Replace CHAMP with RB map in CCF. 
- [x] 6. Add `range(start_key, end_key)` API. 
- [ ] 7. Typed ranges (see https://github.com/microsoft/CCF/issues/4232) 
- [ ] 8. Support for `rb::map::remove()`

jeffa5:
Linking #4789 and #4788

achamayou:
- [ ] Revisit instances of foreach in built-in endpoints, particularly the replay protection code added in #4823 to see if they can't be optimised by switching to ordered maps when they become available.

achamayou:
The utility/cost calculus looks unlikely to change any time soon on this one, so I am closing this for now, and we can open something fresh if we think we want to go that route.

PR #3243, opened 2021-11-24
Make `RBMap` more like `CHAMP`

jumaffre:
Addresses point 3. in https://github.com/microsoft/CCF/issues/3088

This PR adds support for snapshotting for the `RBMap` implementation. 

Also: 
- Keep track of total number of nodes and total serialised size for `RBMap`. 
- Support for early exit in `foreach()`
- `map_test` now tests both map implementation


src/ds/champ_map.h
```diff
@@ -490,13 +456,18 @@ namespace champ
     {
       return root->foreach(0, std::forward<F>(f));
     }
+
+    std::unique_ptr<Snapshot<K, V, H>> make_snapshot() const
+    {
+      return std::make_unique<Snapshot<K, V, H>>(*this);
+    }
```
jumaffre, line 463: Slightly nicer interface to get a snapshot on a map (also available for the `RBMap`).

src/kv/change_set.h
```diff
@@ -4,18 +4,20 @@
...
 #include "ds/champ_map.h"
 #include "ds/hash.h"
+#include "ds/rb_map.h" // TODO: Remove include
 #include "kv/kv_types.h"
 
 #include <map>
 
 namespace kv
 {
+  // TODO: Revert to CHAMP for now
```
jumaffre, line 14: To revert before merging. I just want to make sure CCF works fine when using the `RBMap`.

eddyashton (approved):

src/node/rpc/member_frontend.h
```diff
@@ -808,7 +808,7 @@ namespace ccf
         JWTKeyMap kmap;
         keys->foreach([&kmap](const auto& kid, const auto& kpem) {
           kmap[kid] = crypto::cert_der_to_pem(kpem);
-          return false;
+          return true;
```
jumaffre, line 811: Full fix in https://github.com/microsoft/CCF/pull/3253

PR #3245, opened 2021-11-25
Rename tests from `_CFT` to `_cft`

jumaffre:
This was accidently changed in #3209


cmake/common.cmake
```diff
@@ -566,59 +567,6 @@ function(add_e2e_test)
   endif()
 endfunction()
 
-# Helper for building end-to-end function tests using the sandbox
```
jumaffre, line 569: Removed unused helped function.

eddyashton (approved):

PR #3246, opened 2021-11-25
Update V8 scripts

letmaik:
This is in preparation for further PRs that require pre-built V8 libraries.

Changes in the build script:
- Use host compiler (Clang 10) instead of vendored
- Debug/release builds
- ASAN option
- Experimental Open Enclave build (currently non-functional)


rengolin (approved):
LGTM, thanks for improving the scripts!

PR #3249, opened 2021-11-25
Fix 2tx partition test

jumaffre:
The existing `test_learner_does_not_take_part` end-to-end partition test did not consider the newly-added learner node when waiting for primary unanimity. And indeed, if it did, the test would fail since the `transition_node_to_trusted` proposal would be rolled back when the partition is lifted since we partitioned a majority of backups. 

To fix this, we now do two things: 
1. Wait for the addition of the node in state `Pending` to be committed on all nodes before creating the partition. 
2. Re-issue a `transition_node_to_trusted` proposal after the partition is lifted so that the original 3 nodes know about the the new learner and eventually catch it up. 

@eddyashton @wintersteiger


tests/partitions_test.py
```diff
@@ -227,34 +238,31 @@ def test_learner_does_not_take_part(network, args):
...
+        LOG.info("New joiner is not promoted to Trusted without f other backups")
         with new_node.client(self_signed_ok=True) as c:
             r = c.get("/node/network/nodes/self")
             assert r.body.json()["status"] == "Learner"
             r = c.get("/node/consensus")
             assert new_node.node_id in r.body.json()["details"]["learners"]
 
-    network.wait_for_primary_unanimity()
-    primary, _ = network.find_nodes()
+    LOG.info("Partition is lifted, wait for primary unaminity on original nodes")
```
wintersteiger, line 255:
```suggestion
    LOG.info("Partition is lifted, wait for primary unanimity on original nodes")
```

wintersteiger (approved):

tests/partitions_test.py
eddyashton, line 255:
```suggestion
    LOG.info("Partition is lifted, wait for primary unanimity on original nodes")
```

tests/partitions_test.py
```diff
@@ -227,34 +238,31 @@ def test_learner_does_not_take_part(network, args):
...
+    # Note: Because trusting the new node failed, the new node is not considered
+    # in the primary unanimity. Indeed, its transition to Trusted may have been rolled back.
+    primary = network.wait_for_primary_unanimity()
     network.wait_for_all_nodes_to_commit(primary=primary)
+
+    LOG.info("Trust new joiner again")
+    network.trust_node(new_node, args)
```
eddyashton, line 262: Just a thought: The fact that the majority partition makes commit progress is not completely set in stone (as discussed, only happens currently because of the timeout signature when first elected). If the partition time ever drops far enough, or we fix the timeout signature behaviour, then when the partition drops it is possible for 0 to win a future election and commit the original trust proposal, in which case this call will presumably fail? Probably worth adding an explicit `check_can_progress(f_backups[0])` while the partition is up.

eddyashton (approved):

tests/partitions_test.py
jumaffre: Good point. We should wait for a new partition primary to be elected before that too. I think that the `transition_node_to_trusted` proposal should be idempotent anyway, but it's better/clearer to add this here too.

wintersteiger: I think the proposal is idempotent as it checks `if (nodeInfo.status === "Pending") ... ` is there another check somewhere?

jumaffre: @wintersteiger Good spot. It definitely is idempotent then.

PR #3251, opened 2021-11-25
Add service config endpoint

wintersteiger:
Follow-up to https://github.com/microsoft/CCF/pull/3198#discussion_r756391952


achamayou (approved):

achamayou:
@wintersteiger since this is a new public API, could you add an entry in the changelog please?

As a rule of thumb, anything that results in an update to the openapi schema files very likely should have a changelog entry.

PR #3253, opened 2021-11-25
Fix `/gov/jwt_keys/all` endpoint which always returned one kid

jumaffre:
Also include JWT issuer name in returned object map. 

Note: There is no need to backport this fix to 1.x since the endpoint was introduced in 2.0.0-dev1


src/node/rpc/member_frontend.h
```diff
@@ -800,16 +808,24 @@ namespace ccf
...
+        keys->foreach(
+          [&kmap, &keys_to_issuer](const auto& kid, const auto& kpem) {
+            auto issuer = keys_to_issuer->get(kid);
+            if (!issuer.has_value())
+            {
+              throw std::logic_error(fmt::format("kid {} has no issuer", kid));
+            }
+            kmap.emplace(
+              kid, KeyIdInfo{issuer.value(), crypto::cert_der_to_pem(kpem)});
+            return true;
```
jumaffre, line 827: This is the fix: always return `true` to avoid early `foreach()` exit.

tests/jwt_test.py
```diff
@@ -28,6 +28,41 @@ def test_refresh_jwt_issuer(network, args):
...
+            network.consortium.set_jwt_public_signing_keys(
+                primary, issuer.name, jwks_fp.name
+            )
+
+    with primary.client(network.consortium.get_any_active_member().local_id) as c:
+        r = c.get("/gov/jwt_keys/all")
+        assert r.status_code == 200, r
+        info = r.body.json()
+        LOG.error(info)
+        for issuer in issuers:
```
eddyashton, line 59: Can we extend this to test multiple kids-per-issuer? I'm not sure why we're changing the API to return both if we're not testing that it is correctly supported + fetched.

So rather than having `issuers = [ ... ]`, we'd have `keys = { infra.jwt_issuer.JwtIssuer("issuer1"): [ "issuer1_kid1", "issuer2_kid2" ]... }`. I think that should be supported everywhere, though `create_jwks` might need to be modified to take a list of kids.

achamayou (approved):

tests/jwt_test.py
jumaffre: Good point. This is now done.

eddyashton (approved):

PR #3256, opened 2021-11-26
Set KV consensus post construction

eddyashton:
Drive by API simplification, pulled out of a larger branch.

We have several distinct constructors for `kv::Store`, that do totally different things, and are actually just used by unit tests. This one in particular is unnecessary - rather than calling `Store store(consensus);` we could just call `Store store; store.set_consensus(consensus);`. So we do in all the relevant unit tests, and remove an unnecessary constructor.


jumaffre (approved):

achamayou:
On the one hand, I like to see code removed, but on the other hand, setting stuff after constructions is always a little sad. This is also net +11, so, I don't know?

PR #3258, opened 2021-11-26
Experimental V8 app

letmaik:
Notes:
- Alternative to libjs_generic, strictly apps-only for now, not used in governance
- Opt-in V8 support: `-DENABLE_V8=ON`
- Bindings cover most of the JS app API, enough to run the logging sample; some crypto is missing and `ccf.host`, `ccf.rpc`
- Part of daily ci in the "NoSGX Instrumented" job (using V8 debug build)
- Part of PR CI in the debug and perf jobs (using V8 release build)


rengolin:
Curious as to why some of the benchmarks are consistently worse, if V8 isn't enabled by default. Though, some benchmarks didn't have numbers at all... Is that a known bug of the runner?

achamayou:
@rengolin discussed with @letmaik offline, but summarizing here for the record:

1. The `/memory` endpoint calls into snmalloc's internal stats, because enclave memory usage can't otherwise be measured on SGX. We've never used it in virtual, there's probably a stage where it's gated off based on platform, it needs looking into but otherwise explains why the `_mem` are all zero. In virtual, we can only measure the host+enclave consumption, which is perhaps ok depending how things are run on AMD.
2. The rest of the benchmark, which you may notice do not include `sgx` in the name are fresh new labels, without historical data, simply because we've never published them (virtual is just a debugging construct, its performance was not of concern until now).
3. The data structure micro-benchmarks (CHAMP, RB) were never run in enclave, and so aren't decorated with `sgx`.
4. The `Historical` benchmark really should follow the decoration convention on the other hand, I think it's an oversight. @eddyashton could fix it please?
5. Benchmarks a PR does not publish to (ie. `*_sgx_*` in this case) aren't printed. That's deliberate.

.azure-pipelines-templates/matrix.yml
```diff
@@ -60,6 +64,19 @@ jobs:
           artifact_name: "SGX_Perf"
           ctest_filter: "${{ parameters.test.perf.ctest_args }}"
 
+  # Performance (NoSGX)
+  # NOTE: CI doesn't support SGX and NoSGX perf tests at the same time.
```
achamayou, line 68: I don't think that's quite accurate, it's just that perf tests are run on an SGX build now. If the perf job built SGX + virtual + v8 app, there should be no difficulty in running all tests since they are decorated accordingly and won't collide.

What this would do is make the perf job take longer to build and run, but a possibility we've discussed if that's the case is to split that into two separate jobs, and a have a third dependent job do the plotting.

I think the main impact would be the sheer amount of metrics displayed in PRs, which would probably only be manageable with https://github.com/jumaffre/cimetrics/issues/75

scripts/v8/clang-wrap.sh
```diff
@@ -24,6 +24,12 @@ while (( "$#" )); do
       -Wmax-tokens) ;;
       # Prevents -stdlib=libc++ from working
       -no-canonical-prefixes) ;;
+      # Avoid debugging issues by making source paths absolute
```
achamayou, line 27: That will avoid debug issues for a local build, but will create them for someone using the .deb with a source checkout.

src/apps/js_v8/README.md
```diff
@@ -0,0 +1,40 @@
...
+The main problems are:
+
+- SGX doesn't allow execution on writable pages, which means JIT is out of the question.
+- V8 development is fast paced and focused on browser-like workloads, which may provide friction to upgrading versions (in order to keep up with security features).
+- Engine bring-up and configuration is way more complex than `quickjs`.
+- There may be calls from within V8 that violate enclave requirements.
+
+The answers to those problems are:
+
+- SGX is an old extension that is getting replaced with better alternatives from AMD, Arm and probably Intel, soon.
```
achamayou, line 21:
```suggestion
- While CCF currently uses SGX, other platforms that CCF intends to target in the near future such as AMD SEV-SNP do not have the same limitations.
```

src/apps/js_v8/README.md
```diff
@@ -0,0 +1,40 @@
...
+- V8 has a `jitless` mode, that does not compile anything. We're hoping its interpreter is faster than `quickjs`, including the initial bring-up.
+- Using the Virtual environment for current testing, and then AMD hardware for a hardware-backed enclave JIT execution is the plan.
```
achamayou, line 23:
```suggestion
- We can test JIT functionality in virtual mode, which is a good approximation for VM-based TEEs that we intend to target.
```

src/apps/js_v8/README.md
```diff
@@ -0,0 +1,40 @@
...
+- V8 versions tend to stick around much longer than other fast-paced projects (like LLVM), and the oldest stable release (9.4.146.\*) is still receiving updates.
+- This complexity is actually a great benefit. We can slice the bring-up into CCF bootstrap, endpoint bootstrap and execution, to only parse the script at the last part, while relying on V8 to cache previously-parsed scripts to speed things up even further.
+- We may be able to reduce the scope of the V8 execution to avoid enclave violations, or at least restrict in which kind of environment they execute on (ex. only AMD).
```
achamayou, line 26:
```suggestion
- We may be able to reduce the scope of the V8 execution to avoid enclave violations, or at least restrict in which kind of environment they execute on.
```

src/apps/js_v8/js_v8_base.cpp
```diff
@@ -0,0 +1,529 @@
...
+    ccfapp::AbstractNodeContext& node_context;
+    ::metrics::Tracker metrics_tracker;
+
+    void execute_request(
+      const ccf::endpoints::EndpointProperties& props,
+      ccf::endpoints::EndpointContext& endpoint_ctx)
+    {
+      if (props.mode == ccf::endpoints::Mode::Historical)
+      {
+        // Historical mode need a v2 adapter (why?)
```
achamayou, line 49: The v2 adapter has marginally more precise http return codes (https://github.com/microsoft/CCF/blob/623f403b817b4364cc2a98931642a951ff5b8247/CHANGELOG.md#added-3), in response to https://github.com/microsoft/CCF/issues/2640.

src/apps/js_v8/js_v8_base.cpp
```diff
@@ -0,0 +1,529 @@
...
+    /// Unpacks the request, load the JavaScript, executes the code
+    void do_execute_request(
+      const ccf::endpoints::EndpointProperties& props,
+      ccf::endpoints::EndpointContext& endpoint_ctx,
+      ccf::historical::StatePtr historical_state)
+    {
+      // Isolates are re-used across requests
+      thread_local V8Isolate isolate;
+
+      // Each request is executed in a new context
```
achamayou, line 81: We should justify why this is safe.

scripts/v8/clang-wrap.sh
letmaik: The debug info always contains absolute paths, this just changes the way they are stored. Before, it was a combination of comp_dir (compilation dir) + name (relative source path) and now name is absolute on its own. Both are fine and equally bad in theory, but the latter has the advantage that it is more compatible with some debuggers like lldb (or vscode-lldb?), at least from my experience. In any case, if you really want to debug V8 then a local build is still the best option as you get the matching source code as well, in the right location. The main point of having a debug variant in Azure Storage is to allow running V8 with all debug checks enabled, allow it to run together with a SAN-enabled build of CCF, and get stack traces and resolved symbols. Actual source-level debugging is not a goal.
If CCF intends to release debuggable .debs at some point then IMO this requires more thoughtful planning and testing to make sure it actually works as expected.

src/apps/js_v8/js_v8_base.cpp
```diff
@@ -0,0 +1,529 @@
...
+#include <vector>
+
+using namespace std;
+using namespace ccf;
+using namespace kv;
+
+namespace ccfapp
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wc99-extensions"
```
achamayou, line 31: I believe this was put in to deal with some quirks in quickjs's C code, we hopefully can remove it in this case?

src/apps/js_v8/kv_module_loader.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "ccf/tx.h"
+#include "node/entities.h"
+#include "node/modules.h"
+
+namespace ccf
+{
+  std::optional<std::string> v8_kv_module_load_callback(
+    const std::string& module_name, void* opaque)
+  {
+    auto tx = (kv::Tx*)opaque;
```
achamayou, line 15: I know we're not particularly good at this in the codebase, but I think we ought to try to stick to C++ casts as much as possible.

src/apps/js_v8/tmpl/HOWTO.md
```diff
@@ -0,0 +1,87 @@
+## How to write templates
+
```
achamayou, line 2: Thank you for writing this, it's super useful!

cmake/common.cmake
```diff
@@ -447,6 +447,126 @@ sign_app_library(
...
+    )
+    use_client_mbedtls(js_v8_base.virtual)
+    install(
+      TARGETS js_v8_base.virtual
+      EXPORT ccf
+      DESTINATION lib
+    )
+  endif()
+
+  option(ENABLE_V8_SGX
```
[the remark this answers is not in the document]
letmaik: This flag will go away, either once the SGX build is working or when the SGX build is abandoned. The purpose of the flag is to opt-in to the experimental failing SGX build. The currently "officially" supported target for v8 is only "virtual", and I did some minor extensions to CMake to handle the case where an app doesn't support all compile targets. 
Yes, the semantics of `COMPILE_TARGETS` are now "for each library, build the intersection of COMPILE_TARGETS and whatever targets the library supports".
This may not be the best solution, feel free to suggest something better.

.azure-pipelines-templates/matrix.yml
letmaik: Do you want to deal with this in this PR? I'd prefer to sort this out in a separate PR given the size of it.

achamayou: No, I'm very happy for it to be separate, I do agree it's a large PR.

cmake/common.cmake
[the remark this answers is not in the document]
achamayou: Why can't the single ENABLE_V8 flag enable the building of v8 for all COMPILE_TARGETS as appropriate? I think it's fine to mark it as experimental in all cases, I understand the sgx variant is much less functional than virtual, but neither of them is released/supported at this point.

scripts/v8/clang-wrap.sh
achamayou: Ok, I see.

> If CCF intends to release debuggable .debs at some point then IMO this requires more thoughtful planning and testing to make sure it actually works as expected.

Agreed, this entirely aspirational at this point.

src/apps/js_v8/js_v8_base.cpp
```diff
@@ -0,0 +1,529 @@
...
+        {
+          auto& path_op = ds::openapi::path_operation(
+            ds::openapi::path(document, key.uri_path),
+            http_verb.value(),
+            false);
+          if (!properties.openapi.empty())
+          {
+            for (const auto& [k, v] : properties.openapi.items())
+            {
+              LOG_INFO_FMT("Inserting field {}", k);
```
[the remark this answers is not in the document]
letmaik: This is copied 1:1 from `js_generic_base.cpp`. Should I make the change only here? Or do them together in a follow-up PR?

src/apps/js_v8/tmpl/kv_store.cpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  v8::Local<v8::ObjectTemplate> KVStore::create_template(v8::Isolate* isolate)
+  {
+    v8::EscapableHandleScope handle_scope(isolate);
+
+    v8::Local<v8::ObjectTemplate> tmpl = v8::ObjectTemplate::New(isolate);
+
+    // Field 0: TxContext
+    tmpl->SetInternalFieldCount(1);
```
achamayou, line 85: Can we keep a static array of fields and take the size of it, or something a little more robust?

src/apps/js_v8/tmpl/kv_store.cpp
```diff
@@ -0,0 +1,107 @@
...
+    v8::Local<v8::Context> context, TxContext& tx_ctx)
+  {
+    v8::Isolate* isolate = context->GetIsolate();
+    v8::EscapableHandleScope handle_scope(isolate);
+
+    v8::Local<v8::ObjectTemplate> tmpl =
+      get_cached_object_template<KVStore>(isolate);
+
+    v8::Local<v8::Object> result = tmpl->NewInstance(context).ToLocalChecked();
+    result->SetAlignedPointerInInternalField(0, (void*)&tx_ctx);
```
achamayou, line 102: We could also name the offsets `constexpr size_t TX_CTX_FIELD_OFFSET = 0` and use that to address it.

src/apps/js_v8/tmpl/receipt.cpp
```diff
@@ -0,0 +1,128 @@
...
+    v8::Local<v8::Name> name, const v8::PropertyCallbackInfo<v8::Value>& info)
+  {
+    v8::Isolate* isolate = info.GetIsolate();
+    v8::Local<v8::Context> context = isolate->GetCurrentContext();
+    ccf::Receipt* receipt = unwrap_receipt(info.Holder());
+
+    size_t size = receipt->proof.size();
+    std::vector<v8::Local<v8::Value>> elements;
+    elements.reserve(size);
+    uint32_t i = 0;
```
achamayou, line 66: is that used?

src/apps/js_v8/tmpl/receipt.cpp
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  v8::Local<v8::ObjectTemplate> Receipt::create_template(v8::Isolate* isolate)
+  {
+    v8::EscapableHandleScope handle_scope(isolate);
+
+    v8::Local<v8::ObjectTemplate> tmpl = v8::ObjectTemplate::New(isolate);
+
+    // Field 0: ccf::Receipt
+    tmpl->SetInternalFieldCount(1);
```
achamayou, line 94: Better if named, as mentioned above. Could also be wrapped into a call that takes an enum for safety.

src/apps/js_v8/tmpl/request_authn_identity.cpp
```diff
@@ -0,0 +1,408 @@
...
+  }
+
+  static void get_jwt_object(
+    v8::Local<v8::Name> name, const v8::PropertyCallbackInfo<v8::Value>& info)
+  {
+    ccf::JwtAuthnIdentity* jwt_ident = unwrap_jwt_authn_identity(info.Holder());
+    v8::Isolate* isolate = info.GetIsolate();
+    v8::Local<v8::Context> context = isolate->GetCurrentContext();
+
+    // Ideally, this should be a template as well.
```
achamayou, line 106: Is there a particular reason why it's not?

src/apps/js_v8/js_v8_base.cpp
letmaik: Done:
```cpp
      // Each request is executed in a new context.
      // A context is used in a browser to separate different
      // origins of a page, for example iframes.
      // No state is shared between contexts, except when
      // explicitly allowed through security tokens, which
      // are not used here.
```

src/apps/js_v8/js_v8_base.cpp
letmaik: I removed the comment, left-over from js_generic.

src/apps/js_v8/js_v8_base.cpp
letmaik: Removed.

src/apps/js_v8/tmpl/receipt.cpp
letmaik: Huh, no, it's gone.

cmake/common.cmake
```diff
@@ -447,6 +447,126 @@ sign_app_library(
 )
 # SNIPPET_END: JS generic application
 
+option(ENABLE_V8 "Enable building of the js_v8 app" OFF)
+if(ENABLE_V8)
```
jumaffre, line 451: Could we move all of this to a new `v8.camke` file?

src/apps/js_v8/tmpl/request_authn_identity.cpp
letmaik: No, I guess an extreme case of binding fatigue. I made it a template now.

getting_started/setup_vm/roles/ccf_build/vars/common.yml
```diff
@@ -25,6 +25,7 @@ debs:
   - iptables # partition test infra
   - libclang1-9 # required by doxygen
   - libclang-cpp9 # required by doxygen
+  - pkg-config # required by v8
```
jumaffre, line 28: Should the CI docker images be updated to pick this up?

cmake/common.cmake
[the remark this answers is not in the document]
letmaik: If we do that then we need a good way to exclude the sgx variant from tests. I haven't found one that actually works well.

src/apps/js_v8/tmpl/kv_store.cpp
letmaik: Yes, good point. The proper solution would be to use an IDL and generated code like browsers do to avoid all this manual code, but unfortunately the one that Chromium has is highly specialized for their use cases. I'll put an enum in each file which should help.

getting_started/setup_vm/roles/ccf_build/vars/common.yml
letmaik: Yep, then this can be removed:
https://github.com/microsoft/CCF/blob/dcff5f6e834f7ba19fbb3dd347ecaa33d1e13f89/.azure-pipelines-v8.yml#L38-L41

What's the right way to do this? Merge first, then update docker?

tests/js-modules/modules.py
```diff
@@ -53,6 +53,26 @@ def test_module_import(network, args):
     return network
 
 
+@reqs.description("Test dynamic module import")
+def test_dynamic_module_import(network, args):
+    if args.package != "libjs_v8":
+        LOG.warning("Skipping test_dynamic_endpoints, requires V8")
+        return network
```
jumaffre, line 60: You should be able to use the `@reqs.installed_package("libjs_v8")` decorator here, with the same effect.

src/apps/js_v8/v8_oe_stubs.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include <signal.h>
+#include <stdio.h>
+#include <sys/resource.h>
+
+// This file stubs the following functions used by V8 that
+// are not available in the Open Enclave standard libraries.
+// Stubbing them is ok because they are not used by V8, mostly
+// because we disable threading.
+
+/*
```
rengolin, line 17: ??

rengolin (commented):
A few comments on the build, but otherwise, looks ok to me.

I can only comment on the V8 parts that I know more or less, but the CCF internal logic still escapes me. Other people are commenting on those, so it should be fine.

getting_started/setup_vm/roles/ccf_build/vars/common.yml
jumaffre: Create the image locally and push it to `ccfciteam` on Docker Hub , update the image used by the CI in this PR, check that the CI is happy, merge.

src/apps/js_v8/tmpl/kv_store.cpp
letmaik: Done, I've added three new helper functions to set/get internal fields and set the field count using enum classes.

tests/js-modules/modules.py
letmaik: Done (both).

scripts/v8/v8-9.4.146.17.patch
```diff
@@ -164,14 +164,188 @@
...
++      has_sse_(true),
++      has_sse2_(true),
++      has_sse3_(false),
++      has_ssse3_(true),
++      has_sse41_(true),
++      has_sse42_(true),
++      is_atom_(false),
++      has_osxsave_(false),
++      has_avx_(true),
++      has_avx2_(true),
```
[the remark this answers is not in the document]
letmaik: If there's a concrete CPU feature that's not supported on all current / in-use SGX-supporting Intel CPUs, then yes, it should be `false`. I don't think SSE/AVX is going away any time soon?

src/apps/js_v8/v8_oe_stubs.cpp
letmaik: It's just a little dump so that anyone curious can see where those symbols are used in the code. Maybe it's useless?

cmake/common.cmake
```diff
@@ -447,6 +447,126 @@ sign_app_library(
 )
 # SNIPPET_END: JS generic application
 
+option(ENABLE_V8 "Enable building of the js_v8 app" OFF)
```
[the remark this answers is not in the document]
letmaik: Done.

src/apps/js_v8/v8_oe_stubs.cpp
rengolin: I normally do that with a short description on the top and a more edited version, but if people are fine with this, no worries.

scripts/v8/v8-9.4.146.17.patch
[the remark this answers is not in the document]
rengolin: Not sure what's available inside SGX enclaves, but whatever works...

.azure-pipelines-templates/matrix.yml
letmaik: Opened https://github.com/microsoft/CCF/issues/3302 for this.

tests/js-modules/modules.py
letmaik: Hm, I think this is not working. It ignores the condition if `--enforce-reqs` is not given, and if it's given then it fails the test.

cmake/common.cmake
```diff
@@ -447,6 +447,126 @@ sign_app_library(
...
+      TARGETS js_v8_base.virtual
+      EXPORT ccf
+      DESTINATION lib
+    )
+  endif()
+
+  option(ENABLE_V8_SGX
+         "Enable building of the SGX compile target of the js_v8 app" OFF
+  )
+  if(ENABLE_V8_SGX AND NOT "sgx" IN_LIST COMPILE_TARGETS)
```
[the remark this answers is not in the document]
letmaik: This is gone now.

getting_started/setup_vm/roles/ccf_build/vars/common.yml
letmaik: https://github.com/microsoft/CCF/issues/3303. I'm not going to do the manual build/push here. @achamayou said it shouldn't be hard, so we can use this PR as a test case.

src/apps/js_v8/tmpl/kv_store.cpp
letmaik: I couldn't figure out how to make it nicer with structs so that you get a compile error when you forget to set a field. I think though it's not a big deal and ok as it is.

cmake/common.cmake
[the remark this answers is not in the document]
letmaik: As discussed, and given that SGX seems to be working, I removed all this again.

src/apps/js_v8/tmpl/template.h
```diff
@@ -0,0 +1,58 @@
...
+  {
+    v8::Local<v8::Template> tmpl = get_cached_template<T>(isolate);
+    return tmpl.As<v8::ObjectTemplate>();
+  }
+
+  template <class T>
+  void set_internal_field_count(v8::Local<v8::ObjectTemplate> tmpl)
+  {
+    static_assert(std::is_enum_v<T>, "T must be an enum class");
+    tmpl->SetInternalFieldCount(sizeof(T));
```
achamayou, line 40: That's not right, or at least it's very different from what you were doing before. Mostly this will set the InternalFieldCount to 4, unless you have some quite large enums, since this is the number of bytes clang 10 uses by default for an enum class

Try:

```
#include <iostream>

enum class Foo
{
    Bar
};

int main()
{
    std::cout << sizeof(Foo) << std::endl;
}
```

letmaik: I went for an `END` marker member and replaced `set_internal_field` with a new `set_internal_fields` which uses a `std::array` with the static size templated based on `T::END`.

src/apps/js_v8/tmpl/kv_store.cpp
letmaik: Continued here: https://github.com/microsoft/CCF/pull/3258#discussion_r768080316
Marking as resolved.

src/apps/js_v8/js_v8_base.cpp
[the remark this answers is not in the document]
letmaik: > The latter, if you don't mind.

I opened https://github.com/microsoft/CCF/issues/3310 and will mark this as resolved.

tests/js-modules/modules.py
jumaffre: OK. Let me have a look at this. My first hunch is that we no longer need the `--enforce-reqs` argument anyway.

achamayou (approved):

tests/js-modules/modules.py
letmaik: Going to resolve this thread since it's not strictly related to this PR.

src/apps/js_v8/README.md
```diff
@@ -0,0 +1,40 @@
...
+- V8 versions tend to stick around much longer than other fast-paced projects (like LLVM), and the oldest stable release (9.4.146.\*) is still receiving updates.
+- This complexity is actually a great benefit. We can slice the bring-up into CCF bootstrap, endpoint bootstrap and execution, to only parse the script at the last part, while relying on V8 to cache previously-parsed scripts to speed things up even further.
+- We may be able to reduce the scope of the V8 execution to avoid enclave violations, or at least restrict in which kind of environment they execute on.
+
+## Compiling & Testing
+
+New CMake options:
+
+- `-DENABLE_V8=ON|OFF`: Enables building of the V8 app library, disabled by default.
+- `-DENABLE_V8_SGX=ON|OFF`: Enables building of the SGX target of the V8 app library, disabled by default.
```
jumaffre, line 33: Does this flag still exist?

jumaffre (approved):

src/apps/js_v8/README.md
letmaik: Nope, removed

Issue #561, opened 2019-11-14
Add unit test for TLS (SSL) stack


jumaffre:
At the moment, our TLS stack does not have any unit test and is only tested via end-to-end tests. We should add unit tests for it to test corner cases.

achamayou:
We have https://github.com/drwetter/testssl.sh now, and mbedtls has its own test suite run in Open Enclave CI. If there are specific scenarios we think are missed and warrant tests in our repo, let's open individual tickets for those.

rengolin:
#3270 implemented this issue, even if just a little. The `testssl.sh` test is still far more useful.

PR #3270, opened 2021-11-30
Simple TLS tests

rengolin:
We didn't have any tests checking the basic TLS functionality of the
src/tls implementation. This is an initial attempt at fixing that.

The idea is to have at least some basic checks to use when adding
OpenSSL functionality to implement TLS 1.3, so checks shouldn't depend
on mbedtls or openssl data structures.

There is only the handshake look that needs to know error values, but
this can be refactored once we bring OpenSSL into the picture.

This new test checks certificate creation, communication patterns,
including handshake, send and receive, close.


src/tls/test/main.cpp
```diff
@@ -0,0 +1,376 @@
...
+{
+  auto pipe = reinterpret_cast<TestPipe*>(ctx);
+  return pipe->recv(end, buf, len);
+}
+
+/// mbedtls debug call back.
+static void dbg_callback(
+  void*, int, const char* file, int line, const char* str)
+{
+  test_log.write(string(file) + ":" + to_string(line) + " " + str);
```
achamayou, line 87: We use fmtlib (https://github.com/fmtlib/fmt) elsewhere in CCF, it's quite good. This would be `fmt::format("{}:{} {}", file, line, str)` for example.

achamayou (approved):

src/tls/test/main.cpp
rengolin: Right, I was just writing quick logs to help debug, I'll change to use `fmt` and rebase.

src/tls/test/main.cpp
```diff
@@ -0,0 +1,378 @@
...
+#include <memory>
+#include <string>
+#include <sys/socket.h>
+#include <thread>
+
+using namespace std;
+using namespace crypto;
+using namespace tls;
+
+logger::ConsoleLogger test_log;
```
jumaffre, line 25: Any reason for not using the more convenient `LOG_..._FMT()` macros here?

src/tls/test/main.cpp
```diff
@@ -0,0 +1,378 @@
...
+  size_t response_length = strlen((const char*)response);
+
+  // Create a CA
+  auto ca = get_ca();
+
+  // Create bogus certificate
+  auto server_cert = get_dummy_cert(ca, "server");
+  auto client_cert = get_dummy_cert(ca, "client");
+
+  // Just testing communication channel, does not verify certificates.
```
jumaffre, line 368: This is the same comment than on line `321`, but for a different test. Should this be updated?

jumaffre (approved):
This PR addresses the long-closed issue https://github.com/microsoft/CCF/issues/561 👍

src/tls/test/main.cpp
```diff
@@ -0,0 +1,378 @@
...
+    test_log.write("Client handshake\n");
+    if (handshake(&client))
+      throw runtime_error("Client handshake error\n");
+  });
+
+  // Create a thread for the server handshake
+  thread server_thread([&server]() {
+    test_log.write("Server handshake\n");
+    if (handshake(&server))
+      throw runtime_error("Client handshake error\n");
```
eddyashton, line 232:
```suggestion
      throw runtime_error("Server handshake error\n");
```

src/tls/test/main.cpp
```diff
@@ -0,0 +1,378 @@
...
+  REQUIRE(strncmp((const char*)buf, (const char*)message, message_length) == 0);
+
+  // Send the response
+  test_log.write(fmt::format(
+    "Server sending message [{}]\n", string((const char*)response)));
+  written = server.write(response, response_length);
+  REQUIRE(written == response_length);
+
+  // Receive the response
+  read = client.read(buf, response_length);
```
eddyashton, line 273: Can we add a test that sends a long message, larger than the crypto implementation will process in a single read? I believe `mbedtls` has some internal buffer and will only give us (iirc) 4k on each call to read, so there are cases where we need to call `client.read` multiple times to retrieve everything from the corresponding `write`.

PR #3272, opened 2021-12-01
JSON configuration: size strings

jumaffre:
Part of #2612 

We now use size strings (e.g. `"400KB"`, `"123B"`, `"1000"`) for configuration entries such as the ledger chunk bytes or the ring buffer sizes, using CLI11's existing conversion logic. 

Also include some refactoring to avoid pulling CLI11 inside the enclave (proof: 
```
$ nm samples/apps/logging/liblogging.enclave.so.signed  | grep AsSizeValue || echo "not found"
not found
```
)

Next: time strings


achamayou:
I think we should list somewhere all the prefixes we support, and the values they correspond to.

achamayou:
+1 for the excellent check that cli11 isn't in enclave to be added to an end to end test in the CI.

src/common/configuration.h
```diff
@@ -0,0 +1,184 @@
...
+
+  bool operator==(const SizeString&) const = default;
+
+  inline operator size_t() const
+  {
+    return value;
+  }
+};
+
+// Note: from_json() is defined differently whether for host or in enclave
```
achamayou, line 53: It's a little confusing that this isn't next to from_json or that it doesn't explain why we do this (to avoid cli11 in enclave I'm guessing).

src/ds/cli_helper.h
```diff
@@ -129,4 +133,13 @@ namespace cli
 
     return option;
   }
+
+  inline static size_t convert_size_string(std::string input)
+  {
+    size_t ret = 0;
+    CLI::AsSizeValue(false)(input); // Parse both all values as multiple of 1024
```
achamayou, line 140:
```suggestion
    CLI::AsSizeValue(false)(input); // Parse all values as multiple of 1024
```

tests/schema.py
```diff
@@ -211,7 +211,7 @@ def add(parser):
         package="samples/apps/logging/liblogging",
         nodes=infra.e2e_args.min_nodes(cr.args, f=0),
         initial_user_count=1,
-        ledger_chunk_bytes=1,  # Chunk ledger at every signature transaction
+        ledger_chunk_bytes="1",  # Chunk ledger at every signature transaction
```
achamayou, line 214: Should that become ledger_chunk="1B"?

achamayou (approved):

tests/schema.py
jumaffre: This also works when omitting the unit suffix but I'll add it here for clarity.

src/enclave/enclave.h
```diff
@@ -24,6 +24,11 @@
 
 namespace enclave
 {
+  inline void from_json(const nlohmann::json& j, SizeString& str)
+  {
+    str = j.get<size_t>();
```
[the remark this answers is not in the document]
jumaffre: As discussed, this was the case because we didn't want to take a dependency on CLI11 in the enclave. I've re-implemented it so we no longer rely on CLI11 and can use the same JSON serder logic in both the host and the enclave.

tests/schema.py
achamayou: Is there any way to make not work? I think we'd decided it was better to be explicit as much as possible. Removing the unit from the names also seems good to do now if we can.

PR #3274, opened 2021-12-02
Fix new SAN issues from #3221

eddyashton:


achamayou (approved):

jumaffre (approved):

src/tls/test/main.cpp
```diff
@@ -204,6 +204,7 @@ void run_test_case(
   // There could be multiple communications between client/server while
   // doing the handshake, and they won't return an error until there's
   // nothing more to read/write, so we put them in separate threads.
+  LOG_INFO_FMT("Starting test");
```
eddyashton, line 207: Also needed to add this to the new `tls_test`. Our logger initialization isn't threadsafe, and we do it lazily on the first call attempt to log. We could add an explicit `loggers::initialize()` call, but we'd need to do it in every test that uses logging. Simpler instead to just make sure we log at least once on the main thread (doing the initialization), before starting any other threads which will try to log.

jumaffre:
Merging this as this is the fix for the broken Daily.

PR #3275, opened 2021-12-02
Convert bool is_snapshot to EntryType entry_type

achamayou:
Preliminary change to the ledger serialisation/deserialisation code to allow for a future EntryType::WriteSetWithClaims.

Needs to be backported to the LTS and released, and users upgrading to 2.0 will need to have applied this change to all their nodes before they can migrate.

### DIFF
--- a/src/kv/kv_types.h
+++ b/src/kv/kv_types.h
@@ -509,7 +509,6 @@ namespace kv
 
     virtual void enable_all_domains() {}
 
-    virtual void emit_signature() = 0;
     virtual ConsensusType type() = 0;
   };
 
--- a/src/kv/kv_types.h
+++ b/src/kv/kv_types.h
@@ -285,7 +285,7 @@ namespace kv
     WriteSet = 0,
     Snapshot = 1
   };
-  
+
   // EntryType must be backwards compatible with the older
   // bool is_snapshot field
   static_assert(sizeof(EntryType) == sizeof(bool));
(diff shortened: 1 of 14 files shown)


src/kv/kv_types.h
```diff
@@ -280,6 +280,16 @@ namespace kv
...
 
+  // EntryType must be backwards compatible with the older
+  // bool is_snapshot field
+  static_assert(sizeof(uint8_t) == sizeof(bool));
+
+  enum class EntryType : uint8_t
+  {
+    WriteSet = 0,
+    Snapshot = 1
+  };
```
jumaffre, line 291:
```suggestion
  enum class EntryType : uint8_t
  {
    WriteSet = 0,
    Snapshot = 1
  };
  
  // EntryType must be backwards compatible with the older
  // bool is_snapshot field
  static_assert(sizeof(EntryType) == sizeof(bool));
```

The diff is ugly but moving the assert underneath the enum declaration so we can assert on the size of `EntryType` and not `uint8_t`.

jumaffre (approved):

src/kv/kv_types.h
achamayou: That's ok, but we know sizeof(EntryType) == sizeof(uint8_t) because of line 287. What isn't standard and is compiler-dependent is sizeof(bool) however.

PR #3276, opened 2021-12-02
Add support for KV range for `kv::untyped::Map`

jumaffre:
Part of #3088 

This PR adds support for KV range API, only when the underlying KV state is implemented as a `rb::Map` (which is now enabled via the new `KV_RANGE_RB` cmake/compile-time flag, off by default). 

This new API is only available on the underlying `kv::untyped::Map` (and not its typed `kv::TypedMap` wrapper) since it is not necessarily useful as this point and specifying a custom key comparison operator is awkward (e.g. it would require extra deserialisation when committing a key).

As discussed with @eddyashton, this first implementation is not as efficient as it could be for the reasons listed in `untyped_map_handle.h`: mainly looping over the range twice and constructing an intermediate `std::map` to store the results in order. While the optimisation is important for performance and memory usage reasons, this is currently not urgent as this implementation gives the functionality described in #3088.


.azure-pipelines-templates/matrix.yml
```diff
@@ -15,7 +15,7 @@ parameters:
 
   build:
     common:
-      cmake_args: '-DCMAKE_C_COMPILER_LAUNCHER="ccache" -DCMAKE_CXX_COMPILER_LAUNCHER="ccache" -DENABLE_2TX_RECONFIG=ON'
+      cmake_args: '-DCMAKE_C_COMPILER_LAUNCHER="ccache" -DCMAKE_CXX_COMPILER_LAUNCHER="ccache" -DENABLE_2TX_RECONFIG=ON -DKV_STATE_RB=ON'
```
jumaffre, line 18: To revert before merging.

src/kv/test/kv_test.cpp
```diff
@@ -2644,4 +2646,179 @@ TEST_CASE("Reported TxID after commit")
...
+{
+  std::map<T, T> range;
+  auto f = [&range](const T& k, const T& v) { range.emplace(k, v); };
+  h->range(f, from, to);
+  return range;
+}
+
+TEST_CASE("Range")
+{
+  LOG_INFO_FMT("Running KV range tests");
```
jumaffre, line 2693: This is here to check that the test is actually running on the CI, rather than https://github.com/microsoft/CCF/pull/3274#discussion_r760977601.

src/kv/untyped_map_handle.h
```diff
@@ -82,6 +82,48 @@ namespace kv::untyped
       return &search->value;
     }
 
+    template <class F>
+    void foreach_state_and_writes(F&& f, bool always_consider_writes)
```
jumaffre, line 86: Refactoring this in another private method that lets us early exit once we've passed the end of the range, but still proceed to look at the writes (`always_consider_writes = true`). `always_consider_writes = false` for the existing `foreach` implementation.

src/kv/change_set.h
```diff
@@ -2,20 +2,31 @@
...
   template <typename V>
   using VersionV = map::VersionV<V>;
+#ifndef KV_STATE_RB
   template <typename K, typename V, typename H>
   using State = champ::Map<K, VersionV<V>, H>;
   template <typename K, typename V, typename H>
   using Snapshot = champ::Snapshot<K, VersionV<V>, H>;
+#else
+  template <typename K, typename V, typename H>
+  using State = rb::Map<K, VersionV<V>>;
```
achamayou, line 26: fwiw, you could save a few lines with single block where you do `using State = CCF_MAP_NS::Map<K, VersionV<V>>;` etc and a single `#define CCF_MAP_NS rb|champ" at lines 8 and 9 where you already have a the switch.

src/kv/untyped_map_handle.h
```diff
@@ -260,5 +266,53 @@ namespace kv::untyped
 
       return size_;
     }
+
+#ifdef KV_STATE_RB
```
eddyashton, line 270: I think this doesn't actually need to be gated by `KV_STATE_RB`, since it should also work with the champ map as currently written? The only change for the RB is that we can early out when we get past the range (which we can't do for the champ, since seeing one item past the range doesn't mean we won't see future entries within the range). This would be neater, as we can then use the API elsewhere without gating on this seemingly unrelated `KV_STATE_RB` define. Of course as soon as we start implementing the optimisations listed below the implementations will become even more divergent, but while we can keep the API stable it seems worth doing?

src/kv/test/kv_test.cpp
```diff
@@ -2644,4 +2646,179 @@ TEST_CASE("Reported TxID after commit")
...
+
+#ifdef KV_STATE_RB
+
+template <typename T>
+std::map<T, T> std_map_range(const std::map<T, T>& map, T from, T to)
+{
+  std::map<T, T> ret;
+  for (auto const& e : map)
+  {
+    if (e.first < from)
```
achamayou, line 2659: Wouldn't it be better to do an upper_bound(from) and iterate from there?

achamayou (approved):

src/kv/change_set.h
jumaffre: I thought about this too but because the `champ::Map` takes an extra template parameter `H` (that the `rb::Map` doesn't), it's a little awkward. I've simplified it as much as I think is possible though.

achamayou: Oops, indeed, I'd missed that!

eddyashton (approved):

PR #3280, opened 2021-12-03
Add initial implementation of Indexing system

eddyashton:
I believe this is just about green enough for CI testing.

I'll break out some of the unrelated changes into separate PRs so we can discuss the indexing in isolation.

TODO:
- [x] Remove the append entries intercepting hot path
- [x] Flip the index+consensus relationship (index asks consensus where commit is on each commit)
- [x] ~~Look at limiting to a `Strategy`'s considered range~~ (General form not needed yet. The simple case of having an upper cap on requested Txs is supported, so we can have lazy strategies that don't bloat the memory use of our perf tests)


eddyashton:
image

This PR is not ready for merging. I think I know what's causing this, will investigate next week.

src/node/node_state.h
```diff
@@ -1291,6 +1295,12 @@ namespace ccf
       }
 
       consensus->periodic(elapsed);
+
+      // if (sm.check(State::partOfNetwork))
+      // {
+      //   const auto tx_id = consensus->get_committed_txid();
+      //   indexer->update_strategies(elapsed, {tx_id.first, tx_id.second});
+      // }
```
eddyashton, line 1303: TODO: Revert this before merging. Temporarily disabling the indexing to confirm this removes the perf drop, since I can't reproduce it locally. Of course this will break a bunch of tests, so maybe we won't get any perf results from it...

eddyashton:
This PR is now ready for review.

This branch has too many commits, so linking directly to the ci-metrics comment: https://github.com/microsoft/CCF/pull/3280#issuecomment-985688193

I've resolved the memory bloat on `ls_sgx_cft_mem` while retaining the performance improvement in `hist_sgx_cft` by making the strategy in the logging app lazy (so it only constructs an index on-demand). This laziness is optional, so apps can build it eagerly if they wish, depending which strategy they use.

tests/e2e_logging.py
```diff
@@ -667,11 +667,15 @@ def test_historical_receipts_with_claims(network, args):
...
-def get_all_entries(client, target_id, from_seqno=None, to_seqno=None, timeout=5):
+def get_all_entries(
+    client, target_id, from_seqno=None, to_seqno=None, timeout=5, log_on_success=False
+):
     LOG.info(
         f"Getting historical entries{f' from {from_seqno}' if from_seqno is not None else ''}{f' to {to_seqno}' if to_seqno is not None else ''} for id {target_id}"
     )
     logs = []
+    if log_on_success:
+        logs = None
```
jumaffre, line 678:
```suggestion
    logs = None if log_on_success else []
```

include/ccf/indexing/strategy.h
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    // Receives every committed transaction, in-order
+    virtual void handle_committed_transaction(
+      const ccf::TxID& tx_id, const std::shared_ptr<kv::Store>& store) = 0;
+
+    virtual void tick() {}
+
+    // Returns highest tx ID for which this index should be populated, or
+    // nullopt if it wants all Txs. Allows indexes to be populate
```
jumaffre, line 33:
```suggestion
    // nullopt if it wants all Txs. Allows indexes to be populate
```

include/ccf/indexing/seqnos_by_key.h
```diff
@@ -0,0 +1,109 @@
...
+    {}
+
+    SeqnosByKey(const M& map) : SeqnosByKey(map.get_name()) {}
+
+    void handle_committed_transaction(
+      const ccf::TxID& tx_id, const StorePtr& store) override
+    {
+      // NB: Don't use M, instead get an untyped view over the map with the same
+      // name. This saves deserialisation here, where we work with the raw key.
+      auto tx = store->create_tx();
```
jumaffre, line 41:
```suggestion
      auto tx = store->create_read_only_tx();
```

include/ccf/indexing/seqnos_by_key.h
```diff
@@ -0,0 +1,109 @@
...
+    SeqNoCollection get_all_write_txs(const typename M::Key& key)
+    {
+      const auto serialised_key = M::KeySerialiser::to_serialised(key);
+      const auto it = seqnos_by_key.find(serialised_key);
+      if (it != seqnos_by_key.end())
+      {
+        return it->second;
+      }
+
+      return {};
```
jumaffre, line 65:
```suggestion
      return = SeqNoCollection();
```

For consistency with line :106

src/indexing/indexer.h
```diff
@@ -0,0 +1,132 @@
...
+            MAX_REQUESTABLE - uncommitted_entries.size(),
+            committed.seqno - first_requested);
+
+          SeqNoCollection seqnos(first_requested, additional);
+
+          auto stores = transaction_fetcher->fetch_transactions(seqnos);
+          for (auto& store : stores)
+          {
+            const auto tx_id_ = store->current_txid();
+            const ccf::TxID tx_id{tx_id_.term, tx_id_.version};
```
jumaffre, line 109: Why is this intermediate `TxID` object necessary here?

samples/apps/logging/logging.cpp
```diff
@@ -1023,11 +1033,48 @@ namespace loggingapp
...
+          ctx.rpc_ctx->set_response_header(
+            http::headers::RETRY_AFTER, retry_after_seconds);
+          ctx.rpc_ctx->set_response_header(
+            http::headers::CONTENT_TYPE, http::headervalues::contenttype::TEXT);
+          ctx.rpc_ctx->set_response_body(fmt::format(
+            "Still constructing index for private records on key {} - indexed "
+            "to {}/{}",
+            id,
+            indexed_txid.seqno,
+            to_seqno));
```
jumaffre, line 1053: Just double-checking my understanding here: we report the key (`id`) in the error message but the strategy does not actually look at the key and indexes all keys in the target KV map? And if it could only index some keys, `get_indexed_watermark()` would have to take `id` as an argument?

src/indexing/indexer.h
eddyashton: The first is a `kv::TxID`, the second is a `ccf::TxID`. These are still separate types! I've renamed the variables (and made the first type explicit) to clarify this. We should collide them at some point.

samples/apps/logging/logging.cpp
eddyashton: Yes, we're just reporting the `id` to be as verbose as possible. The indexing strategy is currently per-table, but we could write a strategy that only cared about a single key. Then the `id` would be provided where we construct the strategy (where we currently only give the table name).

include/ccf/indexing/seqnos_by_key.h
eddyashton: Sure, I've changed this for consistency. The instance on line :106 is actually required where this one is auto-derived, because there we're deliberately returning a _non-empty `optional`, containing an empty `SeqNoCollection`_ (where `return {}` would just be an empty `optional`).

include/ccf/indexing/strategy.h
```diff
@@ -30,7 +30,7 @@ namespace ccf::indexing
     virtual void tick() {}
 
     // Returns highest tx ID for which this index should be populated, or
-    // nullopt if it wants all Txs. Allows indexes to be populate
+    // nullopt if it wants all Txs. Allows indexes to be populatep
```
achamayou, line 33:
```suggestion
    // nullopt if it wants all Txs. Allows indexes to be populated
```

CMakeLists.txt
```diff
@@ -335,6 +335,11 @@ if(BUILD_TESTS)
       historical_queries_test PRIVATE http_parser.host sss.host
     )
 
+    add_unit_test(
+      indexing_test ${CMAKE_CURRENT_SOURCE_DIR}/src/indexing/test/indexing.cpp
+    )
+    target_link_libraries(indexing_test PRIVATE sss.host)
```
achamayou, line 341: Is there an unforeseen reason why the indexing test needs sss?

include/ccf/indexing/seqnos_by_key.h
```diff
@@ -0,0 +1,109 @@
...
+          to_it = seqnos.upper_bound(to);
+        }
+
+        SeqNoCollection sub_range(from_it, to_it);
+        return sub_range;
+      }
+
+      // In this case we have seen every tx in the requested range, but have not
+      // seen the target key at all
+      return SeqNoCollection();
```
achamayou, line 106: Unless I'm missing something, we don't capture here what we're returning about, it's either the full range or a nullopt if we were missing some of it, right? There's no way to say "here's 95% of the results, but the rest will require linear scanning".

src/indexing/indexer.h
```diff
@@ -0,0 +1,132 @@
...
+#include "indexing/transaction_fetcher_interface.h"
+
+#include <memory>
+#include <string>
+
+namespace ccf::indexing
+{
+  // This is responsible for managing a collection of strategies, and ensuring
+  // each has been given every transaction up to the commit point, in-order.
+  class Indexer : public IndexingStrategies
```
achamayou, line 16: I should know better than comment on naming at this point, but I do wonder if IndexBuilder rather than IndexingStrategy isn't the right descriptive and boring name here.

achamayou (approved):

CMakeLists.txt
eddyashton: Yes. To rekey, we create a `ccf::ShareManager`:
```
#include "node/share_manager.h"
->
`#include "secret_share.h"`
->
in function 'ccf::SecretSharing::split(...)' : undefined reference to 'sss_create_shares'
```

We could break this chain by moving some implementation from headers, but for now we have an unexpected `sss` dependency in several unit tests.

src/indexing/indexer.h
eddyashton: The usual response: I'm sure these names are eventually wrong, but I don't know what the right names are.

I prefer `Strategy` for the current unique greppableness, both literally and mentally. The final concept likely has some intricacies/specialisations that don't quite fit initial impressions given by  `Strategy` _or_ `Builder`, but I think `Strategy` is a more sturdy term (ie - easier to remember and distinguish) on which to hang all of those extra details than `Builder` (assuming here that `IndexBuilder` is either too unwieldy to pronounce every time so becomes `Builder`, or at least that it goes into a bucket in my head keyed by `Builder`, where it is fuzzily conflated with every other `*build*` function).

Naming is hard.

achamayou:
The Daily failures are all in the LTS tests and likely related to the proposal generator changes:

```
59: 18:51:00.260 | INFO     | infra.member:propose:136 - 400 {"error":{"code":"ProposalFailedToValidate","message":"Proposal failed to validate: transition_service_to_open at position 0 failed validation: Error: Proposal does not accept any argument, found \"[object Object]\"\n    at checkNone (public:ccf.gov.consti + 129 chars
59: 18:51:00.672 | INFO     | infra.jwt_issuer:stop:92 - OpenIdProviderServer stopped
```

PR #3281, opened 2021-12-03
Misc fixes, mostly typos

eddyashton:
Pulling out the trivial standalone fixes from #3280.


src/kv/untyped_map.h
```diff
@@ -290,7 +290,9 @@ namespace kv::untyped
       }
     };
 
-    // Public typedef for external consumption
+    // Public typedefs for external consumption
+    using ReadOnlyHandle = kv::untyped::MapHandle;
+    using WriteOnlyHandle = kv::untyped::MapHandle;
     using Handle = kv::untyped::MapHandle;
```
eddyashton, line 296: These typedefs are used by `tx.rw`, `tx.ro`, and `tx.wo`. Previously, since we only defined `Handle`, we could only call `tx.rw` on an untyped map. Now we can also call the others (though the returned type is something you can read-write with, so maybe you shouldn't be...

jumaffre (approved):

PR #3289, opened 2021-12-07
Fix recovery threshold update proposal action

jumaffre:
I believe the regression was introduced in #3198 because of my comment (https://github.com/microsoft/CCF/pull/3198#discussion_r760363198): even if the proposal execution is atomic, `triggerRecoverySharesRefresh()` needs to read the `kv::Tx`'s own writes to find out the latest value of the recovery threshold.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-squeak.
+cluck cluck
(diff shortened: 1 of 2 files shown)


eddyashton (approved):

wintersteiger:
Thanks for the quick fix! Also, good to know we still have to be careful about these things!

jumaffre:
> Thanks for the quick fix! Also, good to know we still have to be careful about these things!

@wintersteiger No worries. It's a little confusing because the `triggerRecoverySharesRefresh()` function does not take the (global) `Tx` object in JS, but the C++ underlying implementation uses the same `Tx` object to read the new recovery threshold and record the new recovery shares.

Issue #2565, opened 2021-05-06
Public custom claims


achamayou:
To improve the usability of receipts, CCF could offer an interface for handlers to attach public custom claims to a transaction (and therefore to receipts covering) in the ledger. The digests of these claims would be combined with the digest of the write set itself to form the leaf of the ledger Merkle Tree.

Receipt endpoints could then provide expansions of those claims, for example from KV tables.

API:

```
ctx.register_public_claim_digest(digest)
ctx.register_public_claim(data_to_digest)
```

or alternatively:

```
ctx.register_public_claim_digest(index, digest)
ctx.register_public_claim(index, data_to_digest)
```

Which would help with claim expansion but would necessitate a convention for missing/optional claims (all zeroes?).

Transactions including custom claims would make use of the flag (https://microsoft.github.io/CCF/main/audit/ledger.html#transaction-format), and add a new segment to ledger entries, distinct from the write set and _outside_ the integrity protection.

That segment must include:

1. A header with
  a. The number of claims
  b. Their nature (public/private)
3. A vector of the digests of the claims themselves

Leaves in the Ledger tree, which at the moment merely are digests of the write sets would become sub-Merkle Trees of `digest(write set), digest(header), vector of digests of claims`.

An endpoint attaching claims to entries it produces is expected to document them via OpenAPI using custom elements.

letmaik:
> Leaves in the Ledger tree, which at the moment merely are digests of the write sets would become sub-Merkle Trees

What's the advantage of using Merkle trees for that? I think this would complicate receipt validation quite a bit since you would have to build the tree again. Why wouldn't it be enough to compute the leaf as `digest(digest(write set) + digest(header) + digest(claim_1) + digest(claim_2) + ...)`?

achamayou:
@letmaik there is no advantage to nesting claims rather than a flat sequence, other than selective rebuilding. If we have multiple service claims, and they're digested first before being added to the claims, you can choose to treat them as opaque or not.

achamayou:
We have re-discussed this feature at length:

## Requirements

- CCF is agnostic about claims format, in line with KV and request/response content type support.
- Receipt verification must not require parsing the Write Set
  - Would break format flexibility (CCF ledger frame encoding is fixed)
  - Would require re-design of private table to bake in confidential claims support
- Full support for ledgers containing a mix of transactions with and without user claims

```
Claims Digest := Digest(User Claims)

Receipt := {
    Signature over root of Proof,
    Proof,
[...]

PR #3292, opened 2021-12-08
Application claims digest in ledger and receipts

achamayou:
Initial implementation of user claim digests in ledger, does not include commit nonce, and requires an additional compatibility change on LTS to work (ie. the ability to read, and correctly digest WriteSetWithClaims there).

- [x] LTS compatibility (#3346)
- [x] make header public, split implementation
- [x] Document ledger format change
- [x] Receipts API update
- [x] Claims C++ API
- [x] Claims JS API
- [x] Logging sample & documentation
- [x] Python verification sample
- [x] Split off Notes in issues (#2565, #3295)

TODO in a later PR:

- Expose claims to experimental v8 runtime (#3360)
- Make use of app claims to validate snapshots (#3348)


NOTES.md
```diff
@@ -0,0 +1,197 @@
+# User-defined claims
```
achamayou, line 1: I will move all of this content to GitHub discussion, as previously agreed. It can cross-link the relevant PR(s) where useful.

include/ccf/receipt.h
```diff
@@ -17,24 +17,75 @@ namespace ccf
...
+      {
+        crypto::Sha256Hash ws_dgst = components.write_set_digest.value();
+        crypto::Sha256Hash cl_dgst = components.claims_digest.value();
+        current = crypto::Sha256Hash(ws_dgst, cl_dgst);
+      }
+      else
+      {
+        throw std::logic_error(
+          "Cannot compute leaf if not both write_set_digest and claims_digest "
+          "are set");
```
jumaffre, line 81:
```suggestion
          "Cannot compute leaf if both write_set_digest and claims_digest "
          "are not set");
```

doc/build_apps/logging_cpp.rst
```diff
@@ -200,4 +200,29 @@ Historical state always contains a receipt. Users wishing to implement a receipt
...
+
+.. literalinclude:: ../../samples/apps/logging/logging.cpp
+    :language: cpp
+    :start-after: SNIPPET_START: set_claims_digest
+    :end-before: SNIPPET_END: set_claims_digest
+    :dedent:
+
+CCF will then record the digest of the transaction as the combined digest of the write set, plus this claims digest.
+
+Receipts for transactions that have set a claims digest expose a `leaf_components`, rather than an opaque `leaf`,
```
jumaffre, line 218:
```suggestion
Receipts for transactions that have set a claims digest expose a ``leaf_components``, rather than an opaque ``leaf``,
```

doc/build_apps/logging_cpp.rst
```diff
@@ -200,4 +200,29 @@ Historical state always contains a receipt. Users wishing to implement a receipt
...
+which means that a receipt endpoint can choose to reveal the claims and remove their digest from the receipt.
+
+.. literalinclude:: ../../samples/apps/logging/logging.cpp
+    :language: cpp
+    :start-after: SNIPPET_START: claims_digest_in_receipt
+    :end-before: SNIPPET_END: claims_digest_in_receipt
+    :dedent:
+
+A client consuming the output of this endpoint can then digest the claims themselves, combine the digest with the other leaf component
+(`write_set_digest`) to obtain the equivalent `leaf`.
```
jumaffre, line 228:
```suggestion
(``write_set_digest``) to obtain the equivalent ``leaf``.
```

doc/audit/ledger.rst
```diff
@@ -35,10 +35,12 @@ The following table describes the structure of a serialised transaction as it is
...
-|          | bool                                     | Is snapshot (``false`` for all ledger entries)                          |
+|          | :cpp:type:`kv::EntryType`                | One of WriteSet, Snapshot or WriteSetWithClaims                         |
 +          +------------------------------------------+-------------------------------------------------------------------------+
 |          | :cpp:type:`kv::Version`                  | Transaction version                                                     |
 +          +------------------------------------------+-------------------------------------------------------------------------+
+|          | :cpp:type:`crypto::Sha256Hash`           | User-defined claims digest, only when entry type is WriteSetWithClaims  |
++          +------------------------------------------+-------------------------------------------------------------------------+
 |          | :cpp:type:`kv::Version`                  | Indicates after which version can this ledger entry be executed while   |
 |          |                                          | maintaining linearizability                                             |
```
jumaffre, line 45: Unrelated to this PR, but is this still true? I believe this was removed in https://github.com/microsoft/CCF/pull/3219

samples/apps/logging/js/src/logging.js
```diff
@@ -61,6 +61,21 @@ export function get_historical_with_receipt(request) {
...
+  const id = get_id_from_query(parsedQuery);
+  // Forward-compatibility with 2.x
+  const kv = ccf.historicalState.kv || ccf.kv;
+  return get_record(kv["public:records"], id);
+}
+
+export function get_historical_public_with_receipt(request) {
+  const result = get_historical_public(request);
+  result.body.receipt = ccf.historicalState.receipt;
+  delete result.body.receipt.leaf_components.claims_digest;
```
jumaffre, line 75: I think this deserves a comment.

src/enclave/claims.h
```diff
@@ -0,0 +1,67 @@
...
+
+namespace ccf
+{
+  class ClaimsDigest
+  {
+  public:
+    using Digest = crypto::Sha256Hash;
+
+  private:
+    bool is_set = false;
```
jumaffre, line 16: Why not making `digest` a `std::optional` instead of having this extra `bool`?

src/enclave/claims.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+  };
+
+  static ClaimsDigest no_claims()
+  {
+    return ClaimsDigest();
+  }
+
+  static crypto::Sha256Hash entry_leaf(
+    const std::vector<uint8_t> write_set,
```
jumaffre, line 56:
```suggestion
    const std::vector<uint8_t>& write_set,
```

src/enclave/claims.h
```diff
@@ -0,0 +1,67 @@
...
+    const crypto::Sha256Hash& claims_digest)
+  {
+    crypto::Sha256Hash write_set_digest({write_set.data(), write_set.size()});
```
jumaffre, line 59:
```suggestion
    ClaimsDigest::Digest write_set_digest({write_set.data(), write_set.size()});
```

Here and in other places in this file.

src/enclave/claims.h
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: `enclave/` seems like an odd directory for this new file. Maybe `node/rpc` instead?

src/crypto/hash_provider.h
```diff
@@ -88,6 +98,12 @@ namespace crypto
     {
       return ds::to_hex(h);
     };
+
+    static inline Sha256Hash from_string(const std::string& str)
```
jumaffre, line 102: This may be prone to errors with the existing `Sha256Hash(const std::string& s)` constructor that expects an hex string. Maybe the existing constructor should be made like this function, as a `from_hex_string()` method?

samples/apps/logging/logging.cpp
```diff
@@ -298,6 +298,13 @@ namespace loggingapp
         // SNIPPET: public_table_access
         auto records_handle = ctx.tx.template rw<RecordsMap>(PUBLIC_RECORDS);
         records_handle->put(params["id"], in.msg);
+        // SNIPPET_START: set_claims_digest
+        if (in.record_claim)
+        {
+          ctx.rpc_ctx->set_claims_digest(
+            std::move(ccf::ClaimsDigest::Digest::from_string(in.msg)));
```
jumaffre, line 305: Is the `std::move()` required here since the result of `from_string()` will be a rvalue?

doc/build_apps/logging_cpp.rst
```diff
@@ -200,4 +200,29 @@ Historical state always contains a receipt. Users wishing to implement a receipt
...
     :start-after: SNIPPET_START: get_historical_with_receipt
     :end-before: SNIPPET_END: get_historical_with_receipt
-    :dedent:
\ No newline at end of file
+    :dedent:
+
+User-defined claims in receipts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A user wanting to tie transaction-specific values to a receipt can do so by attaching a claims digest to their transaction:
```
jumaffre, line 208:
```suggestion
A user wanting to tie request/transaction-specific values to a receipt can do so by attaching a claims digest to their transaction:
```

doc/build_apps/logging_cpp.rst
```diff
@@ -200,4 +200,29 @@ Historical state always contains a receipt. Users wishing to implement a receipt
     :language: cpp
     :start-after: SNIPPET_START: get_historical_with_receipt
     :end-before: SNIPPET_END: get_historical_with_receipt
-    :dedent:
\ No newline at end of file
+    :dedent:
+
+User-defined claims in receipts
```
jumaffre, line 205:
```suggestion
User-Defined Claims in Receipts
```
I think is the convention we've followed for documentation headings.

include/ccf/receipt.h
achamayou: In my (admittedly non-native) understanding, these sentences felt equivalent.

doc/audit/ledger.rst
achamayou: We've stopped populating it, but the entry is still there for compatibility.

samples/apps/logging/js/src/logging.js
achamayou: Agreed

doc/build_apps/logging_cpp.rst
```diff
@@ -200,4 +200,29 @@ Historical state always contains a receipt. Users wishing to implement a receipt
...
+.. literalinclude:: ../../samples/apps/logging/logging.cpp
+    :language: cpp
+    :start-after: SNIPPET_START: set_claims_digest
+    :end-before: SNIPPET_END: set_claims_digest
+    :dedent:
+
+CCF will then record the digest of the transaction as the combined digest of the write set, plus this claims digest.
+
+Receipts for transactions that have set a claims digest expose a `leaf_components`, rather than an opaque `leaf`,
+which means that a receipt endpoint can choose to reveal the claims and remove their digest from the receipt.
```
jumaffre, line 219: I think we need to explain why an endpoint would want to do this.

src/enclave/claims.h
achamayou: It started out that way, but it's very inflexible even now and in the future for extensions/crypto agility.

samples/apps/logging/logging.cpp
```diff
@@ -812,6 +819,64 @@ namespace loggingapp
...
+          const auto v = records_handle->get(id);
+
+          if (v.has_value())
+          {
+            LoggingGetReceipt::Out out;
+            out.msg = v.value();
+            assert(historical_state->receipt);
+            // SNIPPET_START: claims_digest_in_receipt
+            historical_state->receipt->describe(out.receipt);
+            out.receipt.leaf_components->claims_digest = std::nullopt;
```
jumaffre, line 855: Likewise, I think this needs a comment.

doc/build_apps/logging_cpp.rst
achamayou: I think this is "transaction" in the broader sense, including the response, the request, and anything read/written in the KV. Anything the transaction handler has access to qualifies.

src/js/wrap.cpp
```diff
@@ -678,6 +678,42 @@ namespace ccf::js
...
+      return JS_ThrowTypeError(ctx, "Argument must be an ArrayBuffer");
+
+    if (digest_size != ccf::ClaimsDigest::Digest::SIZE)
+      return JS_ThrowTypeError(
+        ctx, "Argument must be an ArrayBuffer of the right size");
+
+    ccf::ClaimsDigest::Digest::Representation r;
+    std::copy(digest, digest + digest_size, r.begin());
+    ccf::ClaimsDigest::Digest d;
+    d.set(std::move(r));
```
jumaffre, line 710: Couldn't we use the `Sha256Hash(const CBuffer& data)` constructor here instead?

src/crypto/hash_provider.h
achamayou: Yeah, it is, that's why it's split. I'll look for how many places need to change to remove the other ctor, but in general the digest API isn't super happy and deserves a cleanup pass.

src/enclave/claims.h
achamayou: Yeah I don't remember how it landed there. I think we want the header under include/ccf in fact, and the cpp somewhere else maybe node/rpc.

src/js/wrap.cpp
achamayou: I'd like to kill off CBuffer, so maybe a from_span() or something like that.

samples/apps/logging/logging.cpp
achamayou: Sure

tests/e2e_logging.py
```diff
@@ -41,7 +43,35 @@ def verify_receipt(receipt, network_cert, check_endorsement=True):
...
+    else:
+        assert "leaf_components" in receipt
+        write_set_digest = bytes.fromhex(receipt["leaf_components"]["write_set_digest"])
+        claims_digest = bytes.fromhex(receipt["leaf_components"]["claims_digest"])
+        leaf = sha256(write_set_digest + claims_digest).digest().hex()
+    root = ccf.receipt.root(leaf, receipt["proof"])
+    ccf.receipt.verify(root, receipt["signature"], node_cert)
+
+
+def verify_receipt_with_claims(receipt, network_cert, claims, check_endorsement=True):
```
jumaffre, line 57: This looks very similar to `verify_receipt()` above. Could we merge both with a new `has_claims=True` parameter or would this be too much of a flag fest?

doc/audit/ledger.rst
jumaffre: Yes, but the field is only still present to preserve backwards compatibility, so I think we should indicate that here.

include/ccf/receipt.h
```diff
@@ -17,24 +17,75 @@ namespace ccf
...
+      {
+        crypto::Sha256Hash ws_dgst = components.write_set_digest.value();
+        crypto::Sha256Hash cl_dgst = components.claims_digest.value();
+        current = crypto::Sha256Hash(ws_dgst, cl_dgst);
+      }
+      else
+      {
+        throw std::logic_error(
+          "Cannot compute leaf if both write_set_digest and claims_digest "
+          "are not set");
```
achamayou, line 81:
```suggestion
        throw std::logic_error(
          "Cannot compute leaf unless both write_set_digest and claims_digest "
          "are set");
```

doc/build_apps/logging_cpp.rst
achamayou: I've tried to explain this somewhat generically.

js/ccf-app/src/global.ts
```diff
@@ -242,6 +242,12 @@ export interface CCFRpc {
    * The default is `false`.
    */
   setApplyWrites(force: boolean): void;
+
+  /**
+   * Set a claims digest to be associated with the transaction if it succeeds. This
+   * digest can later be accessed from the receipt, and expanded into a full claim.
+   */
+  setClaimsDigest(digest: ArrayBuffer): void;
```
letmaik, line 250: The docstring should say what kind of hash this is, sha-256? Or at least how to figure it out if it's not mentioned here directly.

src/js/historical.cpp
```diff
@@ -23,8 +23,34 @@ namespace ccf::js
...
-      ctx, js_receipt, "leaf", JS_NewString(ctx, receipt_out.leaf.c_str()));
+    if (receipt_out.leaf.has_value())
+    {
+      JS_SetPropertyStr(
+        ctx, js_receipt, "leaf", JS_NewString(ctx, receipt_out.leaf->c_str()));
+    }
+    if (receipt_out.leaf_components.has_value())
+    {
+      auto leaf_components = JS_NewObject(ctx);
+      if (receipt_out.leaf_components->write_set_digest.has_value())
```
letmaik, line 34: When can this be missing?

src/js/historical.cpp
```diff
@@ -23,8 +23,34 @@ namespace ccf::js
...
+      {
+        JS_SetPropertyStr(
+          ctx,
+          leaf_components,
+          "write_set_digest",
+          JS_NewString(
+            ctx, receipt_out.leaf_components->write_set_digest->c_str()));
+      }
+      if (receipt_out.leaf_components->claims_digest.has_value())
```
letmaik, line 43: When can this be missing?

js/ccf-app/src/global.ts
achamayou: It's sha256 right now, I will update the docstring and see if I can add a ref to ccf.digest()

src/js/historical.cpp
achamayou: Right now never, but there's benefit in keeping a single schema/set of json wrappers for receipts, and some specific app handlers could conceivably want to provide the full write set as a claim, so it seems sensible to have all the leaves optional.

src/js/historical.cpp
achamayou: Same answer as above. We could have an InternalTxReceipt, strictly for interface between the C++ historical and JS historical ifaces I guess, where the leaf_components aren't optional, rather than reuse the same schema as for the user API, and then wrap it again for the C++ API.
It'd be quite a lot of extra wrappers to save two conditions though.

src/js/historical.cpp
letmaik: I don't understand, these are JS bindings, so the app is the js_generic app and it will always have those fields.

achamayou: After discussion, the agreement is to check for the following invariants before the objects are handed over to JS, so that TypeScript can model them accurately:

1. Either leaf or leaf_components must be present
2. All leaf_components must be present

There is no prejudice to a user wanting to mask or delete some components before serialisation.

tests/js-modules/modules.py
```diff
@@ -605,7 +605,10 @@ def run(args):
         network = test_bytecode_cache(network, args)
         network = test_app_bundle(network, args)
         network = test_dynamic_endpoints(network, args)
-        network = test_npm_app(network, args)
+        if "v8" not in args.package:
```
jumaffre, line 608: Do/should we have a ticket for this?

tests/lts_compatibility.py
```diff
@@ -236,6 +237,18 @@ def run_code_upgrade_from(
                 network.retire_node(primary, node)
                 if primary == node:
                     primary, _ = network.wait_for_new_primary(primary)
+                    spec = inspect.signature(network.txs.issue)
+                    if "record_claim" in spec.parameters:
```
jumaffre, line 241: Why are these two lines necessary?

tests/lts_compatibility.py
```diff
@@ -236,6 +237,18 @@ def run_code_upgrade_from(
...
+                        LOG.info("Run transaction with additional claim")
+                        txid = network.txs.issue(
+                            network, number_txs=1, record_claim=True
+                        )
+                        LOG.info(
+                            "Check receipts are fine, including transaction with claims"
+                        )
+                        test_random_receipts(
+                            network, args, lts=True, additional=[txid.seqno]
```
jumaffre, line 250: Could you rename `additional` to `additional_seqnos` or something similar? I (wrongly) first assumed that additional referred to additional claims in the receipt.

tests/lts_compatibility.py
```diff
@@ -236,6 +237,18 @@ def run_code_upgrade_from(
                 network.retire_node(primary, node)
                 if primary == node:
                     primary, _ = network.wait_for_new_primary(primary)
+                    spec = inspect.signature(network.txs.issue)
+                    if "record_claim" in spec.parameters:
+                        LOG.info("Run transaction with additional claim")
+                        txid = network.txs.issue(
+                            network, number_txs=1, record_claim=True
+                        )
```
jumaffre, line 245: As discussed, I don't think the JS app has been upgraded at this point (still using 1.x) so I believe the claim will not be recorded.

tests/lts_compatibility.py
achamayou: They're not anymore

tests/lts_compatibility.py
achamayou: Done

tests/lts_compatibility.py
achamayou: Absolutely, this and several more issues fixed as discussed

jumaffre (approved):

PR #3296, opened 2021-12-09
Remove HTTP redirect for already-forwarded requests

jumaffre:
We used to HTTP redirect requests that had already been forwarded, mostly to keep the forwarding logic simple (otherwise nodes would also have to know how to forward back responses on a node-to-node channel). This isn't great as deployments that rely on internal requests forwarding may not know how to HTTP redirect clients properly (e.g. if the service is behind a unique load balancer). 

Instead, we now raise an error in the unlikely event that this happens, rather than trying to support this in the `Forwarder` logic that may be [removed sooner and than later](https://github.com/microsoft/CCF/discussions/3190).

Also removed now-unused types in `forwarder.h`.


src/node/rpc/frontend.h
```diff
@@ -543,8 +520,8 @@ namespace ccf
         return ctx->serialise_response();
       }
 
-      // NB: If we want to re-execute on backups, the original command could be
-      // propagated from here
+      // NB: If we want to re-execute on backups, the original command could
+      // be propagated from here
```
eddyashton, line 524: These reflow changes look revertible - presumably a temporary indentation change?

eddyashton (approved):

achamayou (approved):

PR #3297, opened 2021-12-09
Remove `proposal_generator.py`

eddyashton:
This PR removes `proposal_generator.py`. Rather than piggybacking off our infra to provide type hints for proposal actions in the default constitution, we want to emphasise that these are generic JSON objects with a schema determined by the service's current constitution.

We provide a replacement script called `build_proposal.sh` to demonstrate construction of proposals using `jq`. This can generate every proposal that `proposal_generator.py` could (albeit with different arguments/prior mangling), and also proposals containing multiple actions and custom actions. However it does not include any type information, so this must be specified by the caller.

We also provide a `ballot_builder.py` script to provide a CLI wrapper around the Jinja ballot templates, and a `bundle_js_app.py` script to demonstrate packaging of JS apps for deployment to `js_generic`.

TODO:
- [ ] Check that the rendered docs display as expected
- [ ] Discuss `CHANGELOG` link
- [ ] Discuss creating a "migrating from 1.x to 2.x" document, with explicit "if you did that, now do this" guide


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Breaking change: Configuration for CCF node is now a JSON configuration file passed in to `cchost` via `--config /path/to/config/file/` CLI argument. Existing CLI arguments have been removed. The `migrate_1_x_config.py` script (included in `ccf` Python package) should be used to migrate existing `.ini` configuration files to `.json` format (#3209).
+- `ccf.proposal_generator` has been removed from the pip package. Sample tools to build proposals and ballots are now included in the pip package as `build_proposal.sh` and `ccf.ballot_builder`. See https://microsoft.github.io/CCF/main/governance/proposals.html#creating-a-proposal for a guide to using these new tools. TODO: Get a usefully future-proof link for this?
```
achamayou, line 17: Making the links usefully future-proof is up to us, we need to pick a name and stick to it. I don't think there's a magic bullet, although html redirect pages should give us ample flexibility when needed.

I am not convinced that `build_proposal.sh` belongs in the pip package, and I think I would prefer to see it move to samples/, but I can see arguments for both approaches.

doc/build_apps/auth/jwt.rst
```diff
@@ -52,23 +46,22 @@ The CA certificate is stored with a ``set_ca_cert`` proposal:
 
 .. code-block:: bash
 
-    $ python -m ccf.proposal_generator set_ca_cert jwt_ms cacert.pem
+    $ build_proposal.sh --action set_ca_cert name jwt_ms cert_bundle @pedro_cert.pem
```
achamayou, line 49: I'm not convinced `pedro_cert.pem` is clearer than `cacert.pem`.

achamayou: `microsoft_jwt_issuer` would be clearer than `jwt_ms` on the other hand.

doc/build_apps/auth/jwt.rst
```diff
@@ -52,23 +46,22 @@ The CA certificate is stored with a ``set_ca_cert`` proposal:
...
-      "auto_refresh": true,
-      "ca_cert_name": "jwt_ms"
-    }
-    $ python -m ccf.proposal_generator set_jwt_issuer issuer.json
+    $ build_proposal.sh \
+      --action set_jwt_issuer \
+        issuer "https://login.microsoftonline.com/common/v2.0" \
+        auto_refresh -b true \
+        key_filter all \
+        ca_cert_name jwt_ms
```
achamayou, line 60: Same as above.

doc/build_apps/js_app_bundle.rst
```diff
@@ -134,25 +129,35 @@ It validates the request body and returns the result of a mathematical operation
 Deployment
 ----------
 
-An app bundle can be wrapped into a governance proposal with the Python client for deployment:
+To deploy a new JS application, it must be embedded into a JSON governance proposal containing all of the new code and endpoint metadata. There is a tool in the CCF pip package to generate this object named ``bundle_js_app``.
```
achamayou, line 132: I think it's a "sample tool" but not necessarily how it must be done. The format of the proposal is standard and subject to release policy, the tool isn't and merely illustrates a possible usage.

doc/governance/proposals.rst
```diff
@@ -46,76 +46,130 @@ For transparency and auditability, all governance operations (including votes) a
...
-This can be used to create proposals for common operations like adding members and users, without writing any JSON.
-It also produces sample vote scripts, which validate that the executed proposed actions exactly match what is expected.
-These sample proposals and votes can be used as a syntax and API reference for producing more complex custom proposals.
-
-Assuming the CCF Python package has been installed in the current Python environment, the proposal generator can be invoked directly as ``ccf.proposal_generator``. With no further argument it will print help text, including the list of possible actions as subcommands:
+Proposals are JSON objects specifying a list of actions and associated arguments. The name of the action should match one of the actions available in your service (in ``actions.js``), and the expected types of the arguments are dictated by the validate and apply functions which handle that action.
+These proposal objects can be constructed by an tool or language which can produce JSON. The CCF pip package contains a sample of a tool to automate proposal creation, written in bash and wrapping the CLI tool ``jq``, named ``build_proposal.sh``:
```
achamayou, line 50:
```suggestion
These proposal objects can be constructed by any tool or language which can produce JSON. The CCF pip package contains a sample of a tool to automate proposal creation, written in bash and wrapping the CLI tool ``jq``, named ``build_proposal.sh``:
```

doc/governance/proposals.rst
```diff
@@ -46,76 +46,130 @@ For transparency and auditability, all governance operations (including votes) a
...
+      build_proposal.sh [--help | --action ACTION_NAME [[FLAGS] ARG_NAME ARG_VALUE...]...]
 
-.. code-block:: bash
+    This tool is a wrapper around jq, to simplify creation of CCF governance
+    proposals.
+    Specify a list of actions and associated args. A single flag per argument can be
+    used to indicate how the value should be parsed:
+      -s String (default)
+      -j JSON (including raw numbers)
+      -b Boolean (including case-insensitive parsing)
```
achamayou, line 65: If the only legal values are "true", "TRUE", "false" and "FALSE", we might as well list them here, and avoid any possible assumptions about 0, 1, yes, no etc (I know, and yet YAML does it).

doc/governance/proposals.rst
```diff
@@ -46,76 +46,130 @@ For transparency and auditability, all governance operations (including votes) a
...
+        {
+          "name": "upload_file",
+          "args": {
+            "contents": "This is a file.\nContaining multiple lines."
+          }
+        }
+      ]
+    }
+
+Ballots are JSON objects containing a JS script exporting a single ``vote`` function, which parse a proposal and return a boolean to indicate the submitter's assent (true to vote in favour, false to vote against). These votes may be hand-written to logically validate complex proposals, but for simple proposals it is often sufficient to do an equality check. Given a proposal object, we can generate a ballot which implements this quality check. Within the CCF pip package there are Jinja templates to generate these objects and scripts, and a Python script that demonstrates rendering these templates from a proposal, named ``ballot_builder``:
```
achamayou, line 97:
```suggestion
Ballots are JSON objects containing a JS script exporting a single ``vote`` function, which parse a proposal and return a boolean to indicate the submitter's conditional assent (true to vote in favour, false to vote against). These votes may be hand-written to logically validate complex proposals, but for simple proposals it is often sufficient to do an equality check. Given a proposal object, we can generate a ballot which implements this equality check. Within the CCF pip package there are sample Jinja templates to generate these objects and scripts, and a Python script that demonstrates rendering these templates from a proposal, named ``ballot_builder``:
```

doc/governance/proposals.rst
```diff
@@ -46,76 +46,130 @@ For transparency and auditability, all governance operations (including votes) a
...
-.. literalinclude:: ../../python/tutorial.py
-    :language: py
-    :start-after: SNIPPET: import_proposal_generator
-    :lines: 1
+    $ ballot_builder.py set_user_pedro.json > vote_for_pedro.json
 
-The proposal generation functions return dictionaries that can be submitted to a ``CCFClient``.
+    $ cat vote_for_pedro.json 
+    {
+      "ballot": "export function vote (rawProposal, proposerId) { let proposal = JSON.parse(rawProposal); if (!(\"actions\" in proposal)) { return false; } /* SNIP */ return true; }"
```
achamayou, line 141: SNIP?

doc/governance/proposals.rst
eddyashton: It also handles `True` and `False` (so that I could be lazy when calling this from Python), and by accident `tRuE` and all such dumb variations.

eddyashton: Actually saying that, I don't think that's needed anymore, since I dump most interesting things through `json.dumps` anyway. I'll see if I can just remove this `-b` type entirely

doc/build_apps/auth/jwt.rst
eddyashton: Good spot - `pedro_cert.pem` was the one I had around for testing, but I agree this has a more sensible name. Have renamed the issuer too.

doc/governance/proposals.rst
achamayou: Ok, that'd be even better.

tests/infra/consortium.py
```diff
@@ -7,21 +7,62 @@
...
 import ccf.ledger
 from infra.proposal import ProposalState
 
+from cryptography import x509
+import cryptography.hazmat.backends as crypto_backends
+
 from loguru import logger as LOG
 
 
+def generate_proposal(proposal_name, **kwargs):
```
achamayou, line 30: I am very confused by the need for this wrapper for build_proposal.sh. Why do we need it at all? It seems to that the benefit of using jinja templates should be to replace our instances of this with a very short call to their instantiation.

python/ccf/bundle_js_app.py
```diff
@@ -0,0 +1,55 @@
...
+import shutil
+import tempfile
+
+
+def read_modules(modules_path):
+    modules = []
+    for path in glob.glob(f"{modules_path}/**/*", recursive=True):
+        if not os.path.isfile(path):
+            continue
+        rel_module_name = os.path.relpath(path, modules_path)
```
achamayou, line 17: os.posixpath (as opposed to os.path) will do the right thing and won't need line 18

tests/infra/consortium.py
eddyashton: `build_proposal.sh` is not using a Jinja template. I should remove the Jinja template for constructing proposals. By the time you've constructed a structured input that you could pass to a template, you've already constructed a proposal. We can drop the invocation of `build_proposal.sh` here and just use Python.json to produce proposals. `ballot_builder` is just the a Python wrapper for instantiating a Jinja template.

eddyashton:
Closing, superceded by #3400.

Issue #3303, opened 2021-12-13
Build CI images in an automated pipeline


achamayou:
As pointed out by @letmaik, CI images are currently typically built (and named) by hand. They're typically `ccfciteam/ccf-ci:oe<OPEN_ENCLAVE_VERSION>`.

It would be nice to have an automated pipeline produce them instead, for example by having a .ci-trigger file containing a tag name, which would trigger a build and push on update, similar to [what's done for CCF releases]( https://github.com/microsoft/CCF/actions/workflows/containers.yml).

PR #3308, opened 2021-12-13
Add job to build CI containers

achamayou:
Attempt to solve #3303. Will trigger when a tag starting with `ccf_ci_image/$TAG` is pushed, and will create an image from this tag called `ccfciteam/ccf-ci:$TAG`.

Note: I know it says container**s**, and we only have the one CI container now, but I have a hunch this won't last and renaming GitHub CI flows is zero fun, since you can never delete the old ones.

Also note the deliberate absence of `:latest`, as opposed to releases, because I can't imagine ever using that happily.

### DIFF
--- a/.github/workflows/ci-containers.yml
+++ b/.github/workflows/ci-containers.yml
@@ -1,9 +1,9 @@
 name: "Build and Publish CI Containers"
 
-on:  
+on:
   push:
     tags:
-      - 'ccf_ci_image/*'
+      - "ccf_ci_image/*"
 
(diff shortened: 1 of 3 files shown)


jumaffre (approved):
Could you add a quick description of the expected workflow for a contributor [in the docs](https://microsoft.github.io/CCF/main/contribute/index.html)?

letmaik (approved):

achamayou:
@jumaffre very good idea, I've done that now!

PR #3313, opened 2021-12-14
Try test image

achamayou:
Produced by https://github.com/microsoft/CCF/pull/3308


achamayou:
Works fine, we just need #3315

PR #3316, opened 2021-12-14
Always run txs.verify() in e2e_logging test

letmaik:
At some point verification wasn't supported in js_generic due to missing receipt endpoints, but this is not the case anymore.


jumaffre (approved):

achamayou (approved):
Thank you for catching that!

PR #3321, opened 2021-12-15
Fix typedoc casing issue

### LINKED ISSUE #3311: Sphinx links to typedoc are broken
E.g. in https://microsoft.github.io/CCF/main/build_apps/js_app_bundle.html:

https://microsoft.github.io/CCF/main/js/ccf-app/interfaces/global.ccf.html

But also on https://microsoft.github.io/CCF/main/build_apps/js_app_ts.html

This seems to be an upper/lower-case issue where the real file names are mixed upper/lower case and the links are all-lowercase.

letmaik:
fixes #3311


letmaik:
@achamayou Right, so this is not going to work because of the change in how themes work between old and new version.

achamayou:
@letmaik ok, fine, I think we bit the bullet and fix the links for all ongoing versions by removing lower(), and that's it.

achamayou (approved):

PR #3325, opened 2021-12-15
Update dependencies: nlohmann/json, CLIUtils/CLI11, nodejs/llhttp

### LINKED ISSUE #3255: Dependencies update
Ahead of 2.0, we should make sure we use up to date versions of our dependencies, and upgrade:

nlohmann/json to 3.10.4
cliutils/cli11 to 2.1.2
nodejs/llhttp to 2.1.4

wintersteiger:
Fixes #3255


achamayou:
@wintersteiger historically, we have used this format: https://github.com/microsoft/CCF/pull/2983

One dependency per PR, specify to and from versions in PR title and branch name. This was suggested by @eddyashton, and it gives the following properties:

- CI Failures are easy to attribute to the relevant change
- Glancing at a file under 3rdparty in GitHub, it's trivial to see which version of the dependency we're using now, and what we used before.

The downside obviously is that it's a little more faff, but on the whole I think it's worth it.

PR #3332, opened 2021-12-16
Print signal a node was terminated by

### LINKED ISSUE #3331: Test runner does not print useful error if `cchost` was terminated by signal
Adding the following somewhere in the code and running the tests will not show the typical segfault error that would appear when running `cchost` directly in a shell:

```c
void* m = malloc(42);
free(m);
free(m);
```

The `err` file is empty. 

The reason is that `cchost` is not run inside a shell but directly, and hence the signal would have to be processed by the test infra and printed manually. See also https://stackoverflow.com/a/22253472.

letmaik:
New log output if a node was killed by a signal:
```
43: 11:27:26.261 | ERROR    | infra.remote:stop:495 - 127.6.188.245 exited with signal: Segmentation fault
```

Fixes #3331.

### DIFF
--- a/tests/infra/remote.py
+++ b/tests/infra/remote.py
@@ -489,6 +489,10 @@ class LocalRemote(CmdMixin):
         if self.proc:
             self.proc.terminate()
             self.proc.wait(10)
+            exit_code = self.proc.returncode
+            if exit_code is not None and exit_code < 0:
+                signal_str = signal.strsignal(-exit_code)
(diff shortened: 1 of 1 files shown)


jumaffre (approved):

wintersteiger:
> Fixes #3332.

Wrong number?

letmaik:
> > Fixes #3332.
> 
> Wrong number?

You're right, self-referencing the PR doesn't make sense :D

achamayou (approved):

PR #3333, opened 2021-12-16
Infra: cleanup `e2e_logging.py`

jumaffre:
Results from two recent discussions regarding `e2e_logging.py`:

- Common tests (i.e. unrelated to app) are moved to new `e2e_common_endpoints.py` (saving a few seconds of tests that do not need to run against JS and C++ logging app).
- Fix `--enforce-reqs` flag: Tests can now be skipped by using the `@reqs` decorator instead of having to explicitly branch on the type of app.


tests/e2e_suite.py
```diff
@@ -62,8 +62,8 @@ def run(args):
         random.shuffle(chosen_suite)
     s.validate_tests_signature(chosen_suite)
 
-    if args.enforce_reqs is False:
-        LOG.warning("Test requirements will be ignored")
+    # Throws TestRequirementsNotMet if test requirements are not met
+    args.enforce_reqs = True
```
letmaik, line 66: Out of curiosity, why is this useful for e2e_suite specifically? I'm just trying to think of `reqs` as the typical `skip-if` (well, the negation of it) in other test frameworks and they don't have a never-skip mode as far as I know.

achamayou: iirc suite we shuffle, and not all orders are supposed to work according to the reqs, which is fine. Outside suite, the expectation was the manually crafted test order should always work. If a req suddenly isn't met, we don't want to silently skip the test.

If this is correct, then I also don't like it much.

letmaik: I think there is a difference between "as a test, please give me these things" (fixtures in pytest) vs. "don't run me at all if x is true" (skipif in pytest). The former would be "i need a network with the js_v8 app loaded up" and the latter "skip this if ccf wasn't built with v8 support". Maybe thinking in those terms gets us to some good semantics.

achamayou: Sort of, but the twist on the suite reqs is that they need to bind very late. If your requirements is "there needs to be at least 3 live nodes in the network", you can't know that until just before you run. Maybe the previous test retires every node but one. Otherwise you need similar post-condition decorators everywhere to describe in which state each test leaves the network, so you can schedule based on requirements.

achamayou (approved):

tests/e2e_suite.py
jumaffre: To be clear, what I've aimed for is for tests in the suite to throw a `TestRequirementsNotMet` exception if the pre-conditions are not met, so that the suite runner can report the test as skipped in the overall suite report. In single tests (e.g. `e2e_logging_cft`), if its requirements are not met, the test will **not** throw but simply log a `WARNING` to say it was skipped because some pre-condition wasn't met (e.g. useful and concise when writing a new test that requires the X or Y endpoint, available in C++ not JS). 

In the end, the requirements are always enforced so the flag should definitely be renamed - the only difference being in how the test being skipped is reported. If we used the same runner for the suite than for "individual" tests (which is what we will be aiming for soon I'm sure), then the flag wouldn't need to exist.

letmaik (approved):

PR #3335, opened 2021-12-17
JS RAII refactoring

### LINKED ISSUE #2910: Use RAII wrappers for JS objects
It seems that we've stopped using those (defined in `js/wrap.h`) in some places. We should also look into simplifying the interface if possible.

wintersteiger:
Fixes #2910


wintersteiger:
This is mergeable, but more can be done, e.g. wrapped atom objects and property enums.

src/apps/js_generic/js_generic_base.cpp
```diff
@@ -399,35 +366,37 @@ namespace ccfapp
...
           {
-            auto prop_name = props[i].atom;
-            auto prop_name_cstr = ctx(JS_AtomToCString(ctx, prop_name));
-            auto prop_val =
-              ctx(JS_GetProperty(ctx, response_headers_js, prop_name));
-            auto prop_val_cstr = JS_ToCString(ctx, prop_val);
-            if (!prop_val_cstr)
+            auto prop_name = prop_enum[i].atom;
+            auto prop_name_str = ctx.to_str(prop_name);
+            if (!prop_name_str)
```
achamayou, line 392: I think for symmetry this should set_error like line 401 with a message like "Invalid endpoint function return value (header type).".

src/apps/js_generic/js_generic_base.cpp
```diff
@@ -436,12 +405,11 @@ namespace ccfapp
...
             }
             endpoint_ctx.rpc_ctx->set_response_header(
-              prop_name_cstr, prop_val_cstr);
-            JS_FreeCString(ctx, prop_val_cstr);
+              *prop_name_str, *prop_val_str);
           }
           for (uint32_t i = 0; i < prop_count; i++)
-            JS_FreeAtom(ctx, props[i].atom);
-          js_free(ctx, props);
+            JS_FreeAtom(ctx, prop_enum[i].atom);
```
achamayou, line 411: It would be good to have those in an RAII wrapper too, in case an exception gets thrown during the assembly of the header map.

src/js/wrap.h
```diff
@@ -57,6 +57,110 @@ namespace ccf::js
...
+    {
+      val = JS_DupValue(ctx, other.val);
+    }
+    JSWrappedValue(JSWrappedValue&& other) : ctx(other.ctx)
+    {
+      val = other.val;
+      other.val = JS_NULL;
+    }
+    ~JSWrappedValue()
+    {
```
achamayou, line 83: Can we add the extra logic for PropEnum and use the wrapper for it too?

src/js/wrap.h
```diff
@@ -57,6 +57,110 @@ namespace ccf::js
...
+      JS_SetPropertyStr(ctx, val, prop.c_str(), value);
+    }
+
+    void set(const std::string& prop, const JSValue& value) const
+    {
+      JS_SetPropertyStr(ctx, val, prop.c_str(), JS_DupValue(ctx, value));
+    }
+
+    JSValue steal()
```
achamayou, line 153: Maybe `take()` sounds a little more neutral? Naming is hard...

achamayou (approved):

PR #3336, opened 2021-12-17
(WIP) Re-use V8 contexts safely by freezing globals

letmaik:
Draft PR, mostly to get some perf numbers.
Please ignore for now. WIP.

TODO:
- npm app test: `/app/proto` test is broken
  - "proto.js:4: TypeError: Cannot read properties of undefined (reading 'Type')"
- npm app test: `/app/jwt` test is broken
  - `jwt_decode()` (from `jwt-decode` package): "Invalid token specified: undefined" and "Invalid token specified: r is not a function"
    Adding the following in a module breaks the module (small extract from `jwt_decode`):
    ```js
    function e(e) {}
    e.prototype = new Error
    e.prototype.name = "MyError";
    ```
    It's not clear why it would break, and there's no clear error related to freezing.
    Can be minimally reproduced through just `Object.freeze(Error.prototype)`.
    EDIT: See comment later on for explanation.


letmaik:
I think freezing all globals will in fact break many libraries as the semantics are changed too much. Simple example:
```
Object.freeze(Error.prototype);
let e = new Error
e.name = "Foo" // throws a TypeError, or silently ignored in non-strict mode
```

Normally, assigning to a property will only affect the top-level of the prototype chain, in the case above `e`. In particular, a new property is created in the top level. However, freezing the prototype changes the semantics such that any assignment leads to an error instead of creating a new property at the top-level. I'm not exactly sure why but it is fair to assume that such code is common practice.

letmaik:
Closing this since it doesn't seem like a reasonable way forward.

Issue #3312, opened 2021-12-14
Implement more bindings in `js_v8`


letmaik:
Follow-up from https://github.com/microsoft/CCF/pull/3258.

- [x] `ccf.rpc`
- [ ] `ccf.host`
- [x] `ccf.crypto`
- [x] `ccf.generateAesKey()`
- [x] `ccf.generateRsaKeyPair()`
- [x] `ccf.isValidX509CertBundle()`
- [x] `ccf.isValidX509CertChain()`
- [x] `ccf.wrapKey()`

eddyashton:
The fields added in #3498:

[ ] `request.path`
[ ] `request.url`
[ ] `request.route`
[ ] `request.query`
[ ] `request.method`
[ ] `request.hostname`

achamayou:
Not going to add substantial new user APIs before container re-design.

PR #3338, opened 2021-12-20
`js_v8`: crypto bindings

letmaik:
Or.. everything that was needed to run the npm app test (minus the subtest that exercises the Open Enclave plugin, because there's no plugin system for `js_v8` currently).

Part of https://github.com/microsoft/CCF/issues/3312.


jumaffre:
It looks like there's still an error with the new module test: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=39193&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=3aae4c9d-f0dc-5b97-1b0e-f86312395af9&l=6323

letmaik:
> It looks like there's still an error with the new module test

I fixed the immediate issue (disabling the OE test) but then the next issue cropped up: `Date.parse()` isn't working but is used in `jwt.ts` of the npm app test. V8 tries to invoke some unsupported POSIX function, which means the OE patch for V8 needs to be adapted. I'll disable this subtest for now and create an issue for it.

PR #3346, opened 2021-12-29
App claims read-only for 1.x

achamayou:
Read-only 1.x compatibility for #3292: LTS nodes can apply and verify transactions with application claims.

This allows a network containing a mix of 2.x and 1.x to replicate transactions with application claims, which means they can be enabled as soon as a 2.x node becomes primary, without a separate activation phase.

### DIFF
--- a/.azure-pipelines-gh-pages.yml
+++ b/.azure-pipelines-gh-pages.yml
@@ -13,7 +13,7 @@ pr:
 
 jobs:
   - job: build_and_publish_docs
-    container: ccfciteam/ccf-ci:oe0.17.2-clang-8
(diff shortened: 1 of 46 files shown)


jumaffre:
Could we add a check for this in the LTS compatibility test? As soon as the primary node becomes a 2.0 node (and 1.x nodes are still in the network), we should be able to issue a transaction with application claims and check that the service commits the transaction as expected.

achamayou:
@jumaffre yes, that check will be in the LTS test of 2.x though, after this is merged and released.

eddyashton (approved):

jumaffre (approved):

Issue #3155, opened 2021-11-03
Arbitrary ledger chunking and snapshot generation


jumaffre:
The current ledger chunking logic is triggered when a ledger file size has reached a specific limit or when a snapshot will be generated for a `seqno` (always on a signature). This requires some additional logic to keep the chunking in sync between all nodes.

Instead, to be able to support the generation of snapshots at arbitrary points (e.g. at the end of the recovery procedure), it would be preferable to have a specific flag in the unprotected header of each KV entry that the host or the enclave on the primary node can flip to signal the ledger chunking on other nodes. 

Tasks:
- [x] Have the primary node set a flag in the KV entry when a new ledger chunk is created, which backup nodes should obey.
- [x] Create mechanism to force a snapshot at the next signature. 

Note that this may require #3026 first.

jumaffre:
Notes from discussion with @wintersteiger.

Q. How to force a new ledger chunk?

- Should this be triggered via an operator/member RPC or process signal? What authorisation does it need (no DOS from arbitrary clients is acceptable)? Member RPC (expensive) or Operator RPC on specific interface?
- A new flag in ledger entry header should be set for the host to create a new chunk (so that ledger chunking is synced across nodes)
- At which ledger entry should this flat be set? All ledger chunks should end on a signature, so we would have to wait for the next signature. 
- Should we just wait for the next signature or force a new one? 

Q. How to force a new snapshot?

- Because nodes should starting from a snapshot should construct ledger chunks that are identical than the ones on the primary, - forcing a new snapshot should also force a new ledger chunk (but not the other way around).
- Similarly to forcing a new ledger chunk, how should the new snapshot be triggered? 
- Similarly to forcing a new ledger chunk, when should we force a new snapshot? Next signature as well?
- Would it be easier if snapshotter was in the kv::store? Not a blocker but would be nice (see #3026).

wintersteiger:
The second part was done in #3544

Issue #2981, opened 2021-09-15
Features to be removed after 3.x


jumaffre:
- [x] 1. Snapshot evidence verification with ledger suffix
Reason: snapshots will include receipt from 2.x and all snapshots used to start 3.x nodes have to be 2.x snapshots (see https://microsoft.github.io/CCF/main/overview/release_policy.html#operations-compatibility).
- [x] 2. `ccf::historical::adapter_v1`
- [ ] 3. Ability for the Python infra to spin up 1.x nodes (including `docker_wrap.sh` script). This is only required for the LTS ledger compatibility test but could be removed if all LTS tests published a ledger/snapshot (e.g. Azure Blob Storage) which is then used by other (i.e. newer) LTS compatibility tests for ledger compatibility.

achamayou:
@jumaffre I suggest rolling the remaining item into a "Features to be removed after 3.x" at this point.

PR #3348, opened 2021-12-30
Emit and verify snapshot digest as an application claim

achamayou:
Make use of #3292 to include snapshot digest in receipt. For the purposes of review, ignore all commits but the last: https://github.com/microsoft/CCF/pull/3348/commits/9ec3928e1ca30ec7362981ce34c382cd26ef1ef2


src/kv/committable_tx.h
```diff
@@ -80,9 +90,12 @@ namespace kv
      * @return transaction outcome
      */
     CommitResult commit(
+      const ccf::ClaimsDigest& claims = ccf::no_claims(),
       bool track_read_versions = false,
       std::function<std::tuple<Version, Version>(bool has_new_map)>
-        version_resolver = nullptr)
+        version_resolver = nullptr,
+      std::function<void(const std::vector<uint8_t>& write_set)>
+        write_set_observer = nullptr)
```
jumaffre, line 98: If we changed the snapshot generation logic to go through a new node-only endpoint to record the snapshot evidence (rather than calling `tx.commit()` directly), then this extra argument wouldn't be necessary as we would be using the generic mechanism to record the write set as a claim? 

I'm not saying it's necessarily a better solution, but it seems that we have already followed this pattern for data recorded by the current primary node directly (e.g. ORCs).

jumaffre:
> Subject to discussion, I think it would be good to backport the claims and receipt emission parts to 1.x, so that sufficiently recent LTS nodes can produce snapshots that 2.x can then consume.

@achamayou This is already the case: the old receipt verification code (using the ledger suffix to verify the authenticity of the evidence) is still present on `main` for this very purpose. As discussed, snapshot generation cannot currently be forced (see #3155) so even when upgrading to the latest 1.x LTS, it is not guaranteed that a snapshot-with-receipt-and-claims will be generated anyway. Note that we've planned to remove the ledger suffix snapshot verification post 2.0 (see https://github.com/microsoft/CCF/issues/2981).

achamayou:
@jumaffre yep, I'd completely missed that, and thought for some reason that we had the receipt-based validation logic in 1.x too. That's fine, it's easier that way.

python/ccf/ledger.py
```diff
@@ -671,7 +672,18 @@ def __init__(self, filename: str):
             receipt_bytes = _peek_all(self._file, pos=receipt_pos)
 
             receipt = json.loads(receipt_bytes.decode("utf-8"))
-            root = ccf.receipt.root(receipt["leaf"], receipt["proof"])
+            if "leaf" in receipt:
+                leaf = receipt["leaf"]
```
jumaffre, line 676: I'm assuming that this path is only used since we've added receipts in snapshots (#2998) but hasn't been included in an official (i.e. non-dev) release just yet. Is my understanding correct?

jumaffre (approved):

python/ccf/ledger.py
achamayou: Yes, this is for compatibility with pre-claims snapshots, and so I think we should remove it from 2.0 final. What do you think?

jumaffre: Removing it sounds like a good idea!

PR #3355, opened 2022-01-05
Update to doxygen 1.9.3

jumaffre:
Since the latest doxygen update, the CodeQL job is failing (https://github.com/microsoft/CCF/runs/4712658927?check_suite_focus=true) as the Doxygen 1.9.1 binary is no longer available on their server. 

I cannot find a "latest" symlink but they seem to keep the latest two versions online and they only update twice a year so updating this version once a year doesn't seem too bad?


achamayou (approved):

achamayou:
This is mildly annoying, if 1.9.N+1 is released and contains breaking changes while we're on 1.9.N, we wouldn't want to have the forced upgrade as the only option.

If this becomes an issue, I think we'll want to create a feed like we did for v8.

PR #3356, opened 2022-01-05
Add tls test past max buffer

rengolin:
Maximum buffer of both mbedTLS and OpenSSL is 16k, so the existing
"large message" test wasn't stressing that. This commit adds a large
enough buffer as well as a handler to update the buffer pointer, like
TLS's real users do.


src/tls/test/main.cpp
```diff
@@ -375,6 +396,36 @@ TEST_CASE("large message")
...
+    message.size(),
+    move(server_cert),
+    move(client_cert),
+    true);
+}
+
+TEST_CASE("very large message")
+{
+  // Uninitialised on purpose, we don't care what's in here
+  size_t len = 16149;
```
achamayou, line 416: That's less than 16k?

rengolin: yes, but the base64 will be more.

rengolin: though, I could just set that to 16k...

achamayou: Ah indeed, still kind of a weird number to pick but ok.

achamayou (approved):

PR #3359, opened 2022-01-06
Clean up doc structure

achamayou:
Move a couple of sections to Contribute, and stop cross-linking them, spotted by @rengolin, thank you!


doc/build_apps/index.rst
```diff
@@ -9,27 +9,13 @@ These endpoints can read or mutate the state of a unique :ref:`build_apps/kv/ind
 
 .. panels::
 
-    :fa:`cloud` :doc:`create_vm`
-    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-    Create a SGX-enabled Azure Virtual Machine.
```
jumaffre, line 15: I would have said that these pages should remain under `build_apps/` instead of `contribute/` as I'd hope we have more app developers than contributors?

achamayou: So the build setup is definitely the build setup for CCF itself, not for apps, which is covered by https://microsoft.github.io/CCF/main/build_apps/install_bin.html

For the VM creation, it's not necessary to build apps as discussed (JS), and even for native apps, it's only really useful for testing performance and making sure you haven't introduced instructions that are illegal in enclave. So moving out of the typical build apps flow doesn't seem unreasonable. Also, I don't think we need to keep a whole page for this, because the DC SKU is now completely vanilla.

jumaffre: Agreed for the `create_vm` page. I think a link to `contribute/build_setup` from `build_apps/install_bin` would be nice though.

jumaffre (approved):

eddyashton (approved):

rengolin (approved):

doc/build_apps/index.rst
achamayou: Done

PR #3361, opened 2022-01-06
Implement TLS in OpenSSL, support TLS 1.3

rengolin:
This is a large commit, implementing the TLS layer in OpenSSL,
together with mbedTLS, and is part of a transition to eliminate
mbedTLS from the codebase.

The src/crypto part is done already and this is a similar effort
on src/tls. The last part is src/clients (used for testing) and
will be done as a future patch, with extensive testing separate
than this work.

This commit has a temporary structure, to allow a compiler flag
to switch between OpenSSL and mbedTLS. Unlike src/crypto, we want
the transition to be as short lived as possible, so we didn't try
to create common classes, just an include trick to use the right
library.

The main changes, in order from most generic to most specific, are:

1.  The CMake flag `TLS_PROVIDER_IS_MBEDTLS` was introduced, turned
    OFF by default, to build CCF with mbedTLS instead of OpenSSL. This
    is for testing, in case we find issues with the new implementation.
    This is in addition to `CRYPTO_PROVIDER_IS_MBEDTLS` so we can set
    each flag independently.
2.  The OpenSSL library (`libssl`) that we use is from the same place
    we pick the Crypto library (`libcrypto`). Either the system or
    OpenEnclave, decided elsewhere. No new libraries are needed.
3.  Some of the existing mbedTLS headers were moved to src/clients to
    isolate that migration from the server side. So far, it only
    needed `ca.h` and `cert.h`, which are self-conatined. Those
    will be removed with the mbedTLS cleanup.
4.  A number of `#ifdef TLS_PROVIDER_IS_MBEDTLS` were added where
    there was no other way to common up implementation. Most notably
    in the previous TLS headers (`src/tls/*.h`) that are now just
    wrappers to the actual headers (`src/tls/{mbedtls,openssl}/*.h`).
5.  All TLS errors on both mbedTLS and OpenSSL have been #defined
    to a common set of `TLS_ERR_` errors, to allow for common error
    handlind, including `WANT_READ` and `WANT_WRITE`. To keep the
    mbedTLS way, OpenSSL has a hack to return the negative
    counterparts of the error codes, so that we can treat them as
    errors, like mbedTLS.
6.  To keep the current CCF implementation intact, we had to keep
    the callback mechanism in OpenSSL. This isn't the most natural
    way of handling BIOs but it works. Both `tls_endpoint` and
    `tls_test` have a similar implementation, handling asynchronous
    I/O. This should disappear later (probably well after removing
    mbedTLS from the codebase), once we manage to get BIOs directly
    to the users, from the OpenSSL implementation.
7.  A number of mbedTLS functionality, most notably in `context.h`,
    has been adapted to allow a common implementation between the
    two libraries, for example `peer_cert()` and `get_verify_error`.
8.  A copy of `crypto::error_string()` is present in the TLS side
    so that we can more easily control each (crypto/tls) with the
    CMake flags.
9.  A number of log messages were added or changed to convey more
    information to help diagnose TLS issues. Not all development
    log messages survived into the production patch, but those that
    did can still convey most of the information and help at least
    zero in on where you need to add more logs later.
10. Removed two unused methods in `Context`: `available_bytes` and
    `set_require_auth`. Those were never called and had no easy
    equivalent in OpenSSL. Added exceptions for some unused parameters
    in OpenSSL's `CA` and `Cert` to make sure we never use them.
    If those never hit on any further testing or benchmarking, then
    we're safe to remove them when we clean up mbedTLS.
11. Added a number of `FIXME` comments, highlighting the assumptions
    from reading the mbedTLS code and what the tests have shown it
    does/needs. Some of them may be wrong, and if so, we need to
    fix them before we start with the mbedTLS cleanup.
12. Added a very large `get_verify_error` in `OpenSSL::Context`
    with all errors the `verify_result` can throw. This wasn't
    really necessary but was added as a debugging tool. Once we're
    happy none of those weird corner cases happen on our code, we
    should really remove all unused ones and common up in a
    `case default:`, where developers can add specific ones later
    for debugging purposes again.
13. Updated the `tls_test` to handle multi-part messages (past 16K),
    supporting large messages, like `tls_endpoint`. Also truncated
    the large messages to make it easier to read the test logs.
14. Changed the e2e_logging.py large message range from 2^10
    (instead of 2^14), to help developers see some messages pass
    before the 16K threshold, which helped me realise what the
    problem was.
15. Added a long README document with all core issues with the port,
    and discussions on how to clean up in the future.


rengolin:
All tests pass on my machine, on virtual. I'm really hoping we don't have enclave specific errors, or it'd be a pain to investigate... :(

rengolin:
Good, the only issue in the sanitizers seems to be on the tls_test, which looks easy to fix. I'll do that tomorrow.

samples/apps/logging/logging.cpp
```diff
@@ -12,6 +12,10 @@
 #include "ccf/http_query.h"
 #include "ccf/user_frontend.h"
 #include "ccf/version.h"
+// FIXME: The header below is used for a single check of the certificate and
```
jumaffre, line 15: FYI, the CI will fail if it finds a `TODO` in the code (we want to rely on GitHub issues to track those items). If you're used to using `FIXME`s, maybe you could add it to the `check-todo.sh` script that is run by the CI?

rengolin: I'm deliberately using FIXMEs to avoid the TODO failures. Those are meant as warnings to the reviewers and I intend to remove them before the final merge, leaving just a comment or a fix. But I don't want CI to fail "just for this", as the `ci-checks` script stops at first failure.

rengolin: Ok, most of the FIXMEs were unnecessary and just comments for the future. The only FIXME that perhaps needs to be done now is the entropy source, which I still haven't figured out how to do that.

rengolin: Added a FIXME check on `check-todo.sh`.

eddyashton: FWIW, my usual pattern for this is to add TODOs, but comment out the `check-todo.sh` call in `ci-checks.sh` so that the other useful checks are run. Then I just have to remember to uncomment the `check-todo.sh` line, which can be soft-enforced with a comment on the PR (can't merge with unresolved comments).

src/enclave/tls_endpoint.h
```diff
@@ -568,9 +561,136 @@ namespace enclave
...
+        // Pipe object
+        void* ctx = (BIO_get_callback_arg(b));
+        int put = send_callback(ctx, (const uint8_t*)argp, len);
+
+        // WANTS_WRITE
+        if (put == TLS_WRITING)
+        {
+          LOG_TRACE_FMT("TLS Endpoint::send_cb() : WANTS_WRITE");
+          *processed = 0;
+          return -1;
```
eddyashton, line 603: What is the -1 here? Is that equivalent to `return WANTS_WRITE;`?

rengolin: Not always, but in this case, yes. `-1` just means `error`, but in this case the error can only be `WANTS_WRITE`, thus the log line.

This is one of the complications of using callbacks in OpenSSL. The callback has to be super careful to return the expected value as well as update the `processed` value or it breaks the rest of the read/write logic in `BIO_read_ex` and `BIO_write_ex`. This is why there's a complex manipulation of return values on both send and receive callbacks.

samples/apps/logging/logging.cpp
rengolin: Just realised the entropy source is an mbedTLS thing, so no point in keeping it for OpenSSL, which already has its own PRNG implementation which I'm guessing is more than enough for our purposes.

rengolin:
The main differences between the previous TLS Report and now are:
* TLS 1.3 vs 1.2
* The cipher list. Before, we were restricting the list and now I haven't done that. It's not complicated to do that but I wanted to know what ciphers we should offer.

This is what the client chooses:
* OpenSSL: TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
* MbedTLS: ECDHE-ECDSA-AES128-GCM-SHA256, 521 bit ECDH (P-521)

OpenSSL does have the same as MbedTLS and we can offer that if needed.

The other problems are:
* Secure Client-Initiated Renegotiation VULNERABLE (NOT ok), DoS threat (6 attempts)
* LUCKY13 (CVE-2013-0169), experimental potentially VULNERABLE, uses cipher block chaining (CBC) ciphers with TLS. Check patches.

All other `Not OK` items are the same on both sides.

Could be that choosing a different cipher would help?

rengolin:
Ok, I have fixed the TLS report issues by forcing the same cipher as before and disabling renegotiation to avoid DoS attacks. Now the TLS report is strictly better (with TLS 1.3 support).

rengolin:
Actually, there's one more thing: certificate validation. MbedTLS was bad because it was >365, and OpenSSL is bad because it's one day (so tests pass, but production will fail).

What is the expected expiration timeframe, ie, how often do we renew certificates?

rengolin:
All tests pass, including daily! Tomorrow the daily check will be green and the PR can be merged!

rengolin:
Also, #3366 fixes some `tlstest` issues that can be merged before or after this one. If before, then we can see the golden diff being applied to the commit in its full glory, which will be cool.

rengolin:
All green now, YAY!

Also, it may be luck, but `hist_sgx_cft` looks consistently above standard deviation...

samples/apps/logging/logging.cpp
```diff
@@ -12,6 +12,10 @@
...
+// could be done using OpenSSL. For now, we keep it as is, but we should make
+// the change once we deprecate, and remove, mbedTLS.
+#include "crypto/mbedtls/mbedtls_wrappers.h"
```
achamayou, line 18: I would suggest grouping in a ticket all the stuff that needs cleaning up when mbedTLS goes, especially what won't easily be grepped.

scripts/check-todo.sh
```diff
@@ -9,11 +9,18 @@ if [ "$#" -eq 0 ]; then
   exit 1
 fi
 
-TODOS=$(git ls-files "$@" | xargs grep -n TODO)
+REDLIST="TODO FIXME"
```
achamayou, line 12: DENYLIST is colour-neutral

src/crypto/openssl/openssl_wrappers.h
```diff
@@ -100,6 +101,11 @@ namespace crypto
       {
         return p.get();
       }
+      /// Reset pointer, free old if ayn
```
achamayou, line 104:
```suggestion
      /// Reset pointer, free old if any
```

src/tls/openssl/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Certificate validation can be complex to handle if you can accept connections
+with certificates or not, and if they come, when and how to validate.
+
+MbedTLS is a lot more lenient on checks. For example, CAs are not tested for
+validity of actually signing other certificates, while OpenSSL has extensive
+checks, which can fail functionality that was previously passing.
+
+For this reason, a number of extra checks in the OpenSSL side were disabled.
+Once we get rid ob MbedTLS we should revisit those checks again and improve
```
achamayou, line 32:
```suggestion
Once we get rid of MbedTLS we should revisit those checks again and improve
```

achamayou (approved):
LGTM!

samples/apps/logging/logging.cpp
rengolin: They're all in the README file. I have removed all the FIXME's for now and left the comments.

This specific comment will be noted when I try to remove the header, so not going to be forgotten.

rengolin (commented):
I've pushed all in one instead of committing through Github to avoid multiple triggers of the tests, which are expensive.

PR #3362, opened 2022-01-07
Docs: Reference endpoints from generated OpenAPI spec 

jumaffre:
Use automated reference to endpoint using `http:VERB:/endpoint` rather than manually referencing the endpoint. 

Note that the frontend prefix (e.g. `/app`) is not included in the path (but is usually included in the curl example underneath the reference, and always is present at the top of the OpenAPI spec documentation page) and the Sphinx extension doesn't seem to let users configure this.


eddyashton (approved):

jumaffre:
Despite the main CI being red for unrelated reasons, I'm merging this as the GitHub Pages job is green and the CI is doc-only.

PR #3363, opened 2022-01-07
Service certificate validity period and renewal

### LINKED ISSUE #3090: Service certificate renewal
Just like for node certificates (see https://github.com/microsoft/CCF/issues/2893), it should be possible for members to refresh the service certificate before this one reaches expiration.

jumaffre:
Resolves #3090 

This PR replaces the hardcoded validity period for the service/network certificate which is now variable, as follows:
- It is originally valid from the `cchost` system host time at startup to a date specified by operators in the JSON config file (defaults to 1 day to force operators/members to renew the server certificate early) (`start` and `recover` node only). 
- The service certificate can then be updated with a new `set_service_certificate_validity` proposal, very similarly to the existing `set_node_certificate_validity` proposal. 

Note:
- I've completely removed the hardcoded validity dates in `key_pair.cpp` (so that we never accidently create a certificate with an hardcoded date). This means that unit tests now have to specify the validity period for certificates they create. 
- I'd wrongly assumed that the renewed service certificate would have to be passed to `rpc_sessions.h` to be presented in the TLS handshake with clients. This is actually not necessary as the service certificate (root CA) is not presented in the handshake as it is assumed that the client already has it in its local trusted store.


doc/operations/certificates.rst
```diff
@@ -43,4 +45,36 @@ The procedure that operators and members should follow is summarised in the foll
...
+    axisFormat  %d/%m
+    todayMarker off
+
+    section Members
+    Service Open By Members + set_service_certificate_validity :milestone, 01-01/15:00, 0d
+    Members must renew certs before expiry (set_service_certificate_validity)              :crit, 01-05/15:00, 1d
+
+    section Service <br> Certificate
+    Initial Validity Period (24h default): done, 01-01/00:00, 1d
+    Post Service Open Validity Period    : 01-01/15:00, 5d
```
jumaffre, line 79: image

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- Service/network certificate validity period is no longer hardcoded and can instead be set by operators and renewed by members:
```
jumaffre, line 12: I wonder if this is an opportunity to rename network certificate to service certificate? Or at least choose one and stick to it and it seems that we have referred to each one interchangeably.

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
...
+  - The new `network_certificate_initial_validity_days` (defaults to 1 day) configuration entry lets operators set the initial validity period for the network certificate (valid from the current system time).
```
achamayou, line 13: 1 day seems like a very short default, I think we should default to 365 days instead.

achamayou (approved):

CHANGELOG.md
jumaffre: The logic behind this is the same as for the node certificate: a short initial validity period (1 day) forces operators to renew the certificate early (and hopefully automate the process!) and avoid learning about it in a year's time when the certificate unexpectedly expires. Is there any specific reason  you think this logic shouldn't apply to the service certificate as well?

CHANGELOG.md
jumaffre: Done. This certificate should always be called service certificate from now on (rendering this PR rather large).

PR #3366, opened 2022-01-08
Update TLS test to make it more robust

### LINKED ISSUE #3365: SSL test (TLS Report) should fail if different
The SSL test passes if it runs successfully, but ignores the contents of the TLS report. In #3361, the report was different, and should have been flagged broken, but didn't.

We could add a check against a golden file to make sure we only change when want (ex. by changing the golden file on the commit).

The test publishes multiple files (HTML, json) we just need to find the most stable one and compare it.

rengolin:
Removes output files before running to allow it to run multiple times without failing.

Compares it to a golden file (`tls_report.html`) to make sure we're not changing the output without notice.

If merged before #3361, we'll need to add the changes to the report in that PR (which is nice).

Otherwise, we'll need to change this PR's golden file after it's merged.

Fixes #3365

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Trigger daily
+Trigger daily for tlstest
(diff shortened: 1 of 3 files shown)


rengolin:
Turns out, the TLS Report has random output (certificates and hashes), so can't be done with a simple golden file. I'll close this and try it again later.

achamayou:
It's a good idea though, finding a way to wildcard the bits that are expected to change would be nice.

PR #3375, opened 2022-01-10
Use python -m mypy on ci-checkss

rengolin:
Calling `mypy` directly works in CI but not on my machine. Using
`python -m <module>` works everywhere, so this should be a no-op that
can now work on my machine. :)


achamayou:
Why doesn't it work? Is there something wrong with your venv somehow?

rengolin:
Right, I don't use venv for this, just call the script to check stuff. I mean, everything else works and this could too if it didn't rely on the command being on the path.

achamayou:
I am super confused, the script sources a venv at line 59, and make sure that it contains various packages including mypy, which means they are available in $PATH, unless something is horribly broken.

You are presumably not running the script somehow, and instead copy-pasting individual lines out of it? I don't mind this individual change, it's fairly innocuous, but you understand that we can't have it as a silent requirement that the script must still work if some arbitrary critical lines are removed right?

rengolin:
I'm just running the script... `./scripts/ci-checks.sh`. Now I saw the `venv` and yeah, I can't understand why it doesn't work on my machine. This is a CCF container on my machine, so nothing wild.

I'll investigate it tomorrow to see if I can find out what's weird about it.

achamayou:
Ok, like I said, I have no problem with the change itself, but it's very strange that it's needed/helpful if you are just running the script.

jumaffre (approved):

rengolin:
Closing it for now until I can figure out what really happened.

PR #3376, opened 2022-01-10
Add more ciphers and restrict elliptic curves in OpenSSL

rengolin:
These are the approved ciphers and curves that we can support. The TLS
1.2 cipher remains the same for compatibility, but the TLS 1.3 cipher
suites is updated to include stronger ciphers.

Comparing the TLS Report, now that I know what to look for, this is now
on par with the previous report (same bit-rate, server-order, restricted
curves), but strictly better with TLS 1.3 and more, stronger approved
ciphers in 1.3.

### DIFF
--- a/src/tls/openssl/context.h
+++ b/src/tls/openssl/context.h
@@ -37,11 +37,13 @@ namespace tls
       // Disable renegotiation to avoid DoS
       SSL_CTX_set_options(
         cfg,
-        SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION |
+        SSL_OP_CIPHER_SERVER_PREFERENCE |
+          SSL_OP_NO_SESSION_RESUMPTION_ON_RENEGOTIATION |
           SSL_OP_NO_RENEGOTIATION);
       SSL_set_options(
         ssl,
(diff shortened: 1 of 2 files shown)


achamayou (approved):

src/tls/openssl/context.h
```diff
@@ -37,20 +37,34 @@ namespace tls
...
+        "TLS_AES_128_GCM_SHA256");
+      SSL_set_ciphersuites(
+        ssl,
+        "TLS_AES_256_GCM_SHA384:"
+        "TLS_CHACHA20_POLY1305_SHA256:"
+        "TLS_AES_128_GCM_SHA256");
+
+      // Restrict the curves to approved ones
+      SSL_CTX_set1_curves_list(cfg, "P-521:P-384");
+      SSL_set1_curves_list(ssl, "P-521:P-384");
```
jumaffre, line 67: Is there an authoritative list to determine this? Perhaps https://docs.microsoft.com/en-us/windows/win32/secauthn/tls-cipher-suites-in-windows-server-2022 or https://csrc.nist.gov/publications/detail/sp/800-52/rev-2/final?

Issue #3107, opened 2021-10-20
Document code update process from 1.x to 2.x


jumaffre:
The `lts_compatibility` test had to be tweaked significantly to be able to carry out code updates from 1.x to 2.x releases (see for example https://github.com/microsoft/CCF/pull/2924/files#r732778659). 

We should document the upgrade process in details so that operators can update services and get the broadest set of features as a result.

jumaffre:
This should also include documenting the new `cchost` configuration migration script (see #2612).

jumaffre:
We should also aggregate the changelog from all `2.0.0-devX` releases so far and include it in the release notes for 2.0.0.

PR #3383, opened 2022-01-11
Documentation: code upgrade procedure

jumaffre:
Part of #3107 (not closing #3107 to remind us to review this before cutting 2.0)

This PR adds documentation for the generic code upgrade procedure as well as specificity of the upgrade from 1.x to 2.0 as tested by the LTS compatibility test.


doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
...
+
+    graph TB;
+        classDef Primary stroke-width:4px
+
+        subgraph Service
+            Node0((Node 0))
+            Node1((Node 1))
+            class Node1 Primary
+            Node2((Node 2))
+        end
```
jumaffre, line 35: image

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
...
+
+            subgraph New Nodes
+                Node3((Node 3))
+                Node4((Node 4))
+                Node5((Node 5))
+                class Node3 NewNode
+                class Node4 NewNode
+                class Node5 NewNode
+            end
+        end
```
jumaffre, line 63: image

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
...
```
jumaffre, line 99: image

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
...
+            subgraph New Nodes
+                Node3((Node 3))
+                Node4((Node 4))
+                class Node4 Primary
+                Node5((Node 5))
+                class Node3 NewNode
+                class Node4 NewNode
+                class Node5 NewNode
+            end
+        end
```
jumaffre, line 129: image

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
...
```
jumaffre, line 160: image

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
...
+
+        subgraph Service
+            Node3((Node 3))
+            Node4((Node 4))
+            class Node4 Primary
+            Node5((Node 5))
+            class Node3 NewNode
+            class Node4 NewNode
+            class Node5 NewNode
+        end
```
jumaffre, line 178: image

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
+Code Upgrade
+============
+
+.. note:: Refer to :doc:`/operations/code_upgrade_1x` for specific guidelines on how to upgrade a 1.x CCF service to 2.0.
+
+This page describes how operators/members can upgrade a live CCF service to a new version with minimal downtime.
```
jumaffre, line 6: I'm not sure how to correctly convey the "minimal downtime" part. Is it really "no downtime at all" since the service is always available for read-only transactions during the code upgrade procedure, but write transactions cannot be processed while the election is happening (which can also happen in a normal election process)? I'm leaning towards "no downtime at all".

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,191 @@
...
+
+.. tip::
+
+    - Note that there is no need to run the code upgrade procedure detailed on this page if `only` the JavaScript/TypeScript application needs updating (see :ref:`JavaScript/TypeScript bundle deployment procedure <build_apps/js_app_bundle:Deployment>`).
+    - If more than a majority of nodes have failed, the disaster recovery procedure should be run by operators instead (see :doc:`/operations/recovery`).
+
+.. note:: CCF guarantees specific live compatibility across different LTS versions. See :ref:`overview/release_policy:Operations compatibility` for more details.
+
+Procedure
+---------
```
jumaffre, line 21: The code upgrade procedure described here is very much the slowest possible one (add new nodes one by one, retired old ones one by one). We know we should be able to support an atomic proposal to retire all old nodes and add the new ones in one go. However, we don't have a test for this so recommending the step-by-step approach seems like the most sensible option for now.

doc/governance/common_member_operations.rst
```diff
@@ -6,12 +6,20 @@ Trusting a New Node
...
 Once the proposal successfully completes, the new node automatically becomes part of the network.
 
 .. note:: Once trusted, it may take some time for the new node to update its ledger and replay the transactions run on the network before it joined (from the beginning of time, or from the snapshot it started from).
 
+Removing an Existing Node
+-------------------------
+
+A node that is already part of the service can safely be retired using the ``remove_node`` proposal.
+
+.. note:: If the now-retired node was the primary node, once the proposal successfully completes, a new primary node may have to be elected.
```
achamayou, line 20:
```suggestion
.. note:: If the now-retired node was the primary node, once the proposal successfully completes, a new primary node will have to be elected.
```

doc/operations/code_upgrade.rst
achamayou: I think "minimal downtime" is fair, there is some write downtime due to the inevitable election, but it's no worse than what can normally happen outside an upgrade if f=1.

doc/operations/code_upgrade.rst
achamayou: There's an intermediate variant where nodes are added as a batch, and removed as a batch but in separate transactions. Agree that we shouldn't describe something we don't have a CI test for.

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,189 @@
...
+                Node4((Node 4))
+                Node5((Node 5))
+                class Node3 NewNode
+                class Node4 NewNode
+                class Node5 NewNode
+            end
+        end
+
+
+3. The old code version (i.e. the code version run by nodes 0, 1 and 2) can be removed using the ``remove_node_code`` proposal action.
```
achamayou, line 66: Doing the code version swap atomically should be fine, except that it would need to coordinate with the automated node re-provisioning service: if old nodes need auto-replacement during the transition, that can only succeed if the old code id is removed.

Depending how the auto-provisioning works, keeping the old code id in the table until quite late may be necessary.

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,189 @@
...
+                Node3((Node 3))
+                Node4((Node 4))
+                Node5((Node 5))
+                class Node3 NewNode
+                class Node4 NewNode
+                class Node5 NewNode
+            end
+        end
+
+- ``Node 1`` (primary) is retired, 4 nodes remaining, ``f = 1``. ``Node 4`` becomes primary after election phase (during which service cannot temporarily process requests that mutate the state of the key-value store):
```
achamayou, line 101: We should make it clear that's true for any election, not specific to a code update.

doc/operations/code_upgrade.rst
```diff
@@ -0,0 +1,189 @@
...
+                Node4((Node 4))
+                class Node4 Primary
+                Node5((Node 5))
+                class Node3 NewNode
+                class Node4 NewNode
+                class Node5 NewNode
+            end
+        end
+
+5. Once all old nodes ``0``, ``1`` and ``2`` have been retired, operators can safely stop them:
```
achamayou, line 162: It's necessary to learn from the service that retirement has been committed to take that step, the meaning of "retired" needs clarification here.

doc/operations/code_upgrade_1x.rst
```diff
@@ -0,0 +1,44 @@
...
+    $ /opt/ccf/bin/cchost --config 2_x_config.json --check
+    2022-01-01T10:00:00.000000Z        100 [info ] ../src/host/main.cpp:78              | CCF version: ccf-2.0.0
+    2022-01-01T10:00:00.000000Z        100 [info ] ../src/host/main.cpp:95              | Configuration file successfully verified
+
+Upgrade to Latest 1.x LTS before 2.0
+------------------------------------
+
+During the code upgrade process, the service will temporarily be made of nodes from different CCF versions. This means that breaking changes introduced in 2.0 (e.g. ledger format for application claims) introduce incompatibility with the ledger format that 1.x nodes can apply.
+
+To prevent this, operators should first complete the :ref:`code upgrade procedure <operations/code_upgrade:Procedure>` from the current 1.x version the service is running on (e.g. ``1.0.10``) to a version of the 1.x LTS as least as high as `1.0.16 <https://github.com/microsoft/CCF/releases/tag/ccf-1.0.16>`_. Once this is done, the service can be safely upgraded to 2.0.
```
achamayou, line 37: To the latest LTS, that's the release compatibility policy, we should avoid hardcoding a version here.

achamayou (approved):

doc/operations/code_upgrade.rst
jumaffre: Yes, good point. I've created https://github.com/microsoft/CCF/issues/3390 to cover this in the future.

doc/operations/code_upgrade.rst
jumaffre: Good point. I'd moved this further down than the order in which we do it in the end-to-end test but it's safer to move it to the latest stage.

PR #3394, opened 2022-01-12
Remove mbedTLS

### LINKED ISSUE #3385: Clean up MBedTLS from code base
Now that both TLS and Crypto are implemented in OpenSSL, we should get rid of the MBedTLS dependency.

These are the tasks that need done:
1. Remove MBedTLS TLS implementation. This is the easiest of them all, just remove `src/tls/mbedtls` and move the OpenSSL code into `/src/tls` and disable the CMake flag.
2. Move `tls_client` to OpenSSL. This is slightly harder as the client itself has some MBedTLS logic that we need to mimic, in addition to using OpenSSL's `CA` and `Cert` classes.
3. Remove MBedTLS Crypto implementation. This should be similar to the TLS side, but there are common classes in /src/crypto` that have a different API as either MBedTLS and OpenSSL. We'd need to move the OpenSSL logic inside them, rather than just moving files.
4. Remove any `mbedtls/*.h` include and move the logic to the remaining `crypto` or `tls` implementations.

wintersteiger:
Fixes #3385

Based on @rengolin's excellent work on getting TLS working with OpenSSL, this now removes the dependency on mbedTLS. 

Still open:

- [x] tls_client (`src/clients`) needs to be ported to OpenSSL
- [x] investigate `d2i_PUBKEY` vs `d2i_PublicKey` (see also https://github.com/openssl/openssl/issues/16989). These functions producing different DER encodings are likely the reason for the LTS compatibility test failures


src/crypto/ecdsa.h
```diff
@@ -10,6 +10,13 @@
 
 namespace crypto
 {
+  // Emulating from mbedtls/ecp.h
+  // Used in raft_types.h
+  // Renato: Can we grow the array at use time?
```
achamayou, line 15: What array?

src/crypto/entropy.h
```diff
@@ -35,8 +35,8 @@ namespace crypto
     Entropy() = default;
     virtual ~Entropy() = default;
 
-    virtual void* get_data() = 0;
-    virtual rng_func_t get_rng() = 0;
+    // virtual void* get_data() = 0;
+    // virtual rng_func_t get_rng() = 0;
```
achamayou, line 39: TBD?

src/crypto/entropy.h
```diff
@@ -328,15 +328,15 @@ namespace crypto
...
     }
 
-    rng_func_t get_rng() override
-    {
-      return &rng;
-    }
+    // rng_func_t get_rng() override
+    // {
+    //   return &rng;
+    // }
```
achamayou, line 334: TBD?

src/crypto/entropy.h
```diff
@@ -328,15 +328,15 @@ namespace crypto
...
 
-    void* get_data() override
-    {
-      return this;
-    }
+    // void* get_data() override
```
achamayou, line 336: TBD?

src/crypto/key_exchange.h
```diff
@@ -4,199 +4,118 @@
...
+    std::vector<uint8_t> get_own_key_share() const
     {
-      free_ctx();
+      if (!own_key)
+      {
+        throw std::runtime_error("missing node key");
+      }
 
-      OPENSSL_cleanse(key_share.data(), key_share.size());
+      // For backwards compatability we need to keep the format we used with
```
achamayou, line 51:
```suggestion
      // For backwards compatibility we need to keep the format we used with
```

src/crypto/key_exchange.h
```diff
@@ -4,199 +4,118 @@
...
-    }
-
-    void load_peer_key_share(CBuffer ks)
-    {
-      if (!ctx)
+      if (ks.size() == 0)
       {
-        throw std::runtime_error(
-          "Missing key exchange context when loading peer key share");
+        throw std::runtime_error("Missing peer key share");
```
achamayou, line 81: I don't care either way, but we shouldn't have half our exceptions descriptions capitalised and the other half not, it looks sloppy.

src/tls/ca.h
```diff
@@ -2,8 +2,73 @@
...
+        X509* res = PEM_read_bio_X509(certBio, NULL, NULL, NULL);
+        BIO_free(certBio);
+        if (!res)
+        {
+          auto err_str = crypto::OpenSSL::error_string(ERR_get_error());
+          LOG_FAIL_FMT("CA::ctor: Could not parse CA: {}", err_str);
+          throw std::logic_error("Could not parse CA: " + err_str);
+        }
+        // The PEM_read function above checks the validity of the certificate,
+        // but not if it's a CA or cn be used as such. This is what MbedTLS
```
achamayou, line 44:
```suggestion
        // but not if it's a CA or can be used as such. This is what MbedTLS
```

src/tls/ca.h
```diff
@@ -2,8 +2,73 @@
...
+        // To cater to that usage, we should create a generic helper in crypto
+        // to do the certificate check and add X509_check_ca() here to be more
+        // robust on our verification.
+        tmp_ca.reset(res);
+      }
+
+      if (crl_.n > 0)
+      {
+        // We don't seem to be using CRL anywhere in CCF, so we should
+        // really remove this option once MbedTLS is gone.
```
achamayou, line 57: Now?

src/tls/ca.h
```diff
@@ -2,8 +2,73 @@
...
+        LOG_FAIL_FMT("CA::ctor: Using CRL in OpenSSL CA");
+        throw std::logic_error("Using CRL in OpenSSL CA");
+      }
+
+      ca = std::move(tmp_ca);
+      crl = std::move(tmp_crl);
+    }
+
+    ~CA() {}
```
achamayou, line 66: = default?

src/clients/cert.h
```diff
@@ -120,28 +97,17 @@ namespace client::tls
...
-
-        case tls_auth_optional:
-        {
-          // Peer certificate is checked but handshake continues even if
-          // verification fails
-          return MBEDTLS_SSL_VERIFY_OPTIONAL;
+          return SSL_VERIFY_NONE;
         }
 
+        case tls_auth_optional: // Note: OpenSSL doesn't seem to support
```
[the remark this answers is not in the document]
wintersteiger: Optional means that it will continue even if the certificate checks fail, I don't think we exercise that anywhere.

src/tls/cert.h
```diff
@@ -2,8 +2,177 @@
...
+
+        has_own_cert = true;
+      }
+
+      own_cert = std::move(tmp_cert);
+      own_pkey = std::move(tmp_pkey);
+
+      if (pw.n)
+      {
+        // We don't seem to be using PW anywhere in CCF, so we should
```
achamayou, line 93: TBD?

src/tls/cert.h
```diff
@@ -2,8 +2,177 @@
...
+        // really remove this option once MbedTLS is gone.
+        LOG_FAIL_FMT("Cert::ctor: Unused password");
+        throw std::logic_error(
+          "Unused password: " + std::string((const char*)pw.p));
+      }
+    }
+
+    ~Cert() {}
```
achamayou, line 101: = default?

src/crypto/ecdsa.h
wintersteiger: I think @rengolin referred to the `std::array<uint8_t, crypto::ECDSA_MAX_LEN> sig;` in `SignedAppendEntriesResponse`. It seems we don't use this struct at all though. Maybe remove it? Perhaps better to turn the array into a vector?

src/crypto/entropy.h
wintersteiger: Ah! Thanks, that can all go now.

src/tls/context.h
```diff
@@ -2,8 +2,518 @@
...
+      SSL_set1_curves_list(ssl, "P-521:P-384");
+
+      // Initialise connection
+      if (client)
+        SSL_set_connect_state(ssl);
+      else
+        SSL_set_accept_state(ssl);
+    }
+
+    virtual ~Context() {}
```
achamayou, line 80: = default

src/tls/context.h
```diff
@@ -2,8 +2,518 @@
...
+
+      // Set cipher for TLS 1.3 (same as above)
+      SSL_CTX_set_ciphersuites(
+        cfg,
+        "TLS_AES_256_GCM_SHA384:"
+        "TLS_AES_128_GCM_SHA256");
+      SSL_set_ciphersuites(
+        ssl,
+        "TLS_AES_256_GCM_SHA384:"
+        "TLS_AES_128_GCM_SHA256");
```
achamayou, line 67: So we no longer support the restrict cipherlist flag in cmake, which is fine, can you delete it too please (and its mention in the doc)?

src/crypto/key_exchange.h
wintersteiger: I completely agree, it seems to be totally random throughout the codebase though. I'll fix my brain to stick to lower-case for now.

achamayou (approved):

src/clients/cert.h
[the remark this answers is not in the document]
achamayou: Ok, I see. Can we just delete the case then?

src/crypto/ecdsa.h
achamayou: If we don't use it, let's remove it!

src/tls/ca.h
wintersteiger: Did we ever use CRLs in the past? I added an implementation in `src/clients/ca.h`, but I'm happy to remove it from both places if you think it's the right thing to do.

src/tls/cert.h
wintersteiger: Will we ever use password-protected certs? If not, I agree, we can kill this branch

achamayou: No, but I'm not sure how things would work if we did?

wintersteiger: Yeah, too much complication and the code is easy to add back if we ever decide to go down that path.

src/tls/tls.h
```diff
@@ -17,27 +15,25 @@
...
+#define TLS_WRITING -SSL_WRITING
+#define TLS_ERR_WANT_READ -SSL_ERROR_WANT_READ
+#define TLS_ERR_WANT_WRITE -SSL_ERROR_WANT_WRITE
+#define TLS_ERR_CONN_CLOSE_NOTIFY -SSL_ERROR_ZERO_RETURN
+#define TLS_ERR_NEED_CERT -SSL_ERROR_WANT_X509_LOOKUP
 // No counterpart in OpenSSL
-#  define TLS_ERR_CONN_RESET INT_MIN
-#  define TLS_ERR_PEER_VERIFY INT_MIN + 1
-#  define TLS_ERR_X509_VERIFY INT_MIN + 2
+#define TLS_ERR_CONN_RESET INT_MIN
```
rengolin, line 25: We don't need to treat those errors anymore, so they could just be removed from the switch/cases.

rengolin (commented):
Looking awesome, thanks!

Some inline comments.

src/clients/cert.h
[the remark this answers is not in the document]
wintersteiger: Yeah, I'll do that.

src/crypto/base64.cpp
```diff
@@ -1,18 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the Apache 2.0 License.
 
-#include "mbedtls/base64.h"
+#include "openssl/base64.h"
```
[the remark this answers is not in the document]
wintersteiger: Yeah, I agree we should do that at some point

src/crypto/entropy.cpp
```diff
@@ -16,6 +16,6 @@ namespace crypto
       return intel_drng_ptr;
     }
 
-    return std::make_shared<Entropy_mbedTLS>();
+    return std::make_shared<Entropy_OpenSSL>();
```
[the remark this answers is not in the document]
wintersteiger: Would what be possible?

src/clients/cert.h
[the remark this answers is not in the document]
rengolin: We don't, that was an mbed thing. We can even delete the whole enum and just have `auth` / `no auth`.

src/crypto/ecdsa.h
rengolin: Ah, then removing would be better, yes.

src/crypto/entropy.cpp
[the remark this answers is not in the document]
rengolin: Sorry, rewording. MBedTLS would pass the entropy source to the configuration, but I'm not doing that at the moment in OpenSSL because I couldn't find a way to do that in the same way mbed does. so the current openssl uses the standard entropy source, which I guess it's good enough by reading the OpenSSL docs.

src/clients/ca.h
```diff
@@ -2,17 +2,14 @@
...
 #include "crypto/pem.h"
 #include "ds/buffer.h"
 
 #include <exception>
 
-// This is a copy of src/tls/ca.h
-// Moving tls to OpenSSL means this different client implementation needs
-// to be isolated while we change tls_endpoint. Once that's done, we can
-// come back here and refactor this too.
+using namespace crypto;
```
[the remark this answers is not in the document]
wintersteiger: I didn't know you copied this over, I thought it's just old stuff. I'll give it a try, although I prefer the one currently in src/clients because it uses the unique wrapper objects.

src/crypto/ecdsa.h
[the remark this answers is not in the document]
wintersteiger: Yes, we will kill the whole struct though.

src/crypto/entropy.cpp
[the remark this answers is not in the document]
wintersteiger: Yes, that's right, the default initialization is fine for OpenSSL. We still have an entropy.h for random number generation in other places though (e.g. ledger secrets) and that is fixed to using the Intel DRNG where supported.

src/clients/ca.h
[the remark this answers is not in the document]
wintersteiger: All done already, there's a small change for a CI failure if I overlooked something.

src/tls/tls.h
wintersteiger: They are still used in a number of places. Are you sure we can remove all those switch/cases? I'm inclined to leave this to you :-)

getting_started/setup_vm/roles/ccf_build/tasks/install.yml
```diff
@@ -34,41 +34,7 @@
...
-    make
-  args:
-    chdir: "{{ workspace }}/mbedtls-{{ mbedtls_dir }}/build"
-
-- name: Install mbedtls
-  command: make install
-  args:
-    chdir: "{{ workspace }}/mbedtls-{{ mbedtls_dir }}/build"
-  become: true
+# TODO: Find OpenSSL?
```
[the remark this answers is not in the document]
wintersteiger: `common.yml` already lists `libssl-dev`, are there any others we should add?

src/tls/tls.h
wintersteiger: Awesome, thanks!

rengolin: #3418

rengolin (approved):

PR #3396, opened 2022-01-13
QUIC OpenSSL library in CCF

rengolin:
Creates a new QUICTLS build into an Azure Universal Package repository,
and adds a way to retrieve it from the CCF build, making sure we include
and link to the right objects in the pre-built QUIC version of OpenSSL
for the enclave side.

It should not change the host side (still using the system's OpenSSL),
but we also need to guarantee that the versions are compatible (ie. have
the same version and backports, ex. 1.1.1f or 1.1.1o, etc).

Because Universal Packages version is strct, we can't have the final
patch letter in the end, so we make sure the packages uploaded there are
a match with the currently supported version of the Linux distro we use.

If we decide to support multiple distros in the future, we can move that
to a blob storage and do the name resolution at the Yaml/CMake level on
both building and importing sides.

MAJOR ISSUE: We can't get the 1.1.1f version that is the same as OE
because the quictls-openssl repo deletes branches too aggressively, all
we have access is from `h` onward. We're using `o` in hope it's backward
compatible but that too will soon disappear. We should NOT use this for
anything serious before we can solve this problem.


rengolin:
@sylvanc

.azure-pipelines-quictls.yml
```diff
@@ -0,0 +1,61 @@
...
+    type: string
+    default: "o"
+  - name: PUSH_ARTIFACT
+    displayName: Publish to package feed
+    type: boolean
+    default: true
+
+jobs:
+  - job: build_quictls
+    container: ccfciteam/ccf-ci:oe0.17.5-2
```
achamayou, line 20:
```suggestion
    container: ccfmsrc.azurecr.io/ccf/ci/sgx:oe-0.17.7-1
```

achamayou (approved):

PR #3399, opened 2022-01-13
Create onboarding page

achamayou:
Plus various documentation fixes. The goal of this initial version of the onboarding is to create a linear flow for someone coming to CCF to learn about the code base, and to branch out into the component or components that are of interest to them.

image
image
image


doc/contribute/onboarding.rst
```diff
@@ -0,0 +1,102 @@
...
+==========
+
+This page aims to be a list of resources that should be helpful if you want to start contributing to CCF.
+
+Introduction
+------------
+
+Start by reading :doc:`/overview/what_is_ccf`.
+
+If you encounter any terms or acronyms that you don't know, check the :doc:`/overview/glossary`. If the word you are looking for isn't defined, create an `issue <https://github.com/microsoft/CCF/issues/new/choose>`_.
```
jumaffre, line 11: Even though this page is more informal than the rest of the documentation, I'd avoid using abbreviations (e.g. "don't")?

doc/contribute/onboarding.rst
```diff
@@ -0,0 +1,102 @@
...
+
+Create an SGX VM if necessary
+-----------------------------
+
+If you intend to make changes that need to work on Intel SGX, you will want an :doc:`SGX VM </contribute/create_vm>` to check you didn't introduce instructions that are illegal in enclave, and to evaluate the performance impact of your change.
+
```
jumaffre, line 17: I feel like this should be phrased the other way around to lower the barrier to entry: unless the changes are large (e.g. bringing in another library) or require touching SGX-specific features (e.g. attestation), developing on non-SGX machine is OK and the CI will run of SGX-specific hardware anyway. Otherwise, you can create an SGX-VM?

doc/contribute/onboarding.rst
```diff
@@ -0,0 +1,102 @@
...
+:doc:`/build_apps/logging_cpp` documents the sample native app under ``/samples/apps/logging`` in the repo.
+
+The logging application is simple, but exercises most features of the framework, and is extensively used in the end to end tests.
+
+To run a locally built copy of this application in a sandbox, see :doc:`/build_apps/run_app`. The package name will be ``samples/apps/logging/liblogging``:
+
+.. code-block:: bash
+
+    ~/CCF/build$ ../tests/sandbox/sandbox.sh -p samples/apps/logging/liblogging
```
jumaffre, line 34: Maybe link to here to avoid duplication https://microsoft.github.io/CCF/main/build_apps/run_app.html?highlight=sandbox?

achamayou: There is already a link on line 30, but I deliberately added this invocation because it's not quite the same in two important ways:

1. it's done for a local build, not install of CCF
2. it points at the non-obviously located liblogging package, which is now three levels deep in a subdirectory because of CMake

doc/contribute/onboarding.rst
achamayou: For the time being, SGX is the only target with security properties, and it's relatively easy as we've seen to create incompatibilities even when not touching SGX-specific features. Once there are other supported TEEs, we can revisit this.

doc/contribute/onboarding.rst
achamayou: Ok, addressed, there's a few other places where we've been doing this.

doc/contribute/onboarding.rst
jumaffre: I'd missed that, thanks.

eddyashton (approved):

PR #3400, opened 2022-01-13
Remove `proposal_generator.py`

eddyashton:
This is a replacement for #3297.

Rather than trying to maintain a generic tool for formatting data into JSON, we instead provide sample JSON objects in our documentation and assume that the construction and formatting of these can be done in each client. This should make it clear that the default constitution actions are not privileged, and custom actions can be called from the same client code - the controlling factor is the schema defined by the action code itself.

Our own infra now directly produces JSON proposals, rather than going via `proposal_generator`, meaning the correct arg names need to be known a little earlier. These proposals are still dumped to files so they can be examined later. The associated ballots are now simple `return true`s.

We retain a few samples, but no longer distribute these in the Python package. Specifically:
- We keep a "construct a careful ballot for this proposal" tool, in the form of the `ballot.json.jinja` template. This is tested from a new test case in `governance.py`, and mentioned in the docs, but not used elsewhere by our infra.
- For JS app packaging, we have a short `build_bundle.js` that can be run as part of `npm build`, and is tested as part of our `npm-app` deployment. For most of our JS apps we bundle directly from the source directories in Python (`set_js_app_from_dir` vs `set_js_app_from_json`), as we construct the proposal, but this sample decouples that and moves it into the JS build step , where it could form the basis of a shared deployment for arbitrary governance clients/infra.


doc/build_apps/js_app_bundle.rst
```diff
@@ -134,25 +134,196 @@ It validates the request body and returns the result of a mathematical operation
...
 ----------
 
-An app bundle can be wrapped into a governance proposal with the Python client for deployment:
+An app bundle must be wrapped into a JSON object for submission as a ``set_js_app`` proposal to deploy the application code onto a CCF service.
+For instance a proposal which deploys the example app above would look like:
 
-.. code-block:: bash
+.. code-block:: json
+
+    {
```
eddyashton, line 142: This block is huge, and largely a repetition of the `app.json` we already paste above. But it's a near-minimal example, and I think it's better to be verbose than to trim some sections?

doc/governance/proposals.rst
```diff
@@ -46,52 +46,200 @@ For transparency and auditability, all governance operations (including votes) a
...
 
-    $ python -m ccf.proposal_generator transition_node_to_trusted 6d566123a899afaea977c5fc0f7a2a9fef33f2946fbc4abefbc3e10ee597343f 211019154318Z
-    SUCCESS | Writing proposal to ./trust_node_proposal.json
-    SUCCESS | Wrote vote to ./trust_node_vote_for.json
+.. code-block:: json
 
-    $ cat trust_node_proposal.json
-    {"actions": [{"name": "transition_node_to_trusted", "args": {"node_id": "6d566123a899afaea977c5fc0f7a2a9fef33f2946fbc4abefbc3e10ee597343f", "valid_from": "211019154318Z"}}]}
+    {
+      "ballot": "export function vote (rawProposal, proposerId)\n
```
eddyashton, line 117: Another docs styling choice - I've added real new-lines next to the embedded escaped new-lines. This matches how these string are pretty-printed in JS, and I think is far more readable, at the cost of not being copy-pastable as a valid JSON object.

tests/infra/consortium.py
```diff
@@ -535,9 +647,9 @@ def set_node_certificate_validity(
     ):
         proposal_body, careful_vote = self.make_proposal(
             "set_node_certificate_validity",
-            node_to_renew.node_id,
-            valid_from,
-            validity_period_days,
+            node_id=node_to_renew.node_id,
+            valid_from=valid_from,
+            validity_period_days=validity_period_days,
```
eddyashton, line 652: Just to make sure this change is clear:

Previously this passed some `*args` through a looked-up-by-name function to `proposal_generator`, where they were type-checked/converted to the correct types, for embedding into JSON. Now the names and types defined here are directly inserted into the JSON object, so the values here need to match the expectations in the corresponding JS action.

tests/npm-app/build_bundle.js
```diff
@@ -0,0 +1,49 @@
+import { readdirSync, statSync, readFileSync, writeFileSync } from "fs";
```
eddyashton, line 1: @letmaik - Could you please scan this file, and let me know if it does anything egregious? It's taking a CCF JS app `src` directory, and converting that into a `bundle` object ready to be inserted into `set_js_app` proposal. My JS is very rusty, so all of the library functions are the first thing that worked, please let me know if there's obvious alternatives.

doc/governance/proposals.rst
```diff
@@ -46,52 +46,200 @@ For transparency and auditability, all governance operations (including votes) a
...
 -------------------
 
-For custom proposals with multiple actions and precise conditional requirements you will need to write the proposal script by hand.
-For simple proposals there is a helper script in the CCF Python package - ``proposal_generator.py``.
-This can be used to create proposals for common operations like adding members and users, without writing any JSON.
-It also produces sample vote scripts, which validate that the executed proposed actions exactly match what is expected.
-These sample proposals and votes can be used as a syntax and API reference for producing more complex custom proposals.
+A proposal's body contains a JSON object with a list of desired actions.
+The actions are identified by name, matching a function from the constitution which should be called to verify and apply this action.
+Each action may associated arguments.
```
jumaffre, line 51:
```suggestion
Each action may have associated arguments.
```

tests/governance.py
```diff
@@ -123,6 +124,52 @@ def test_user(network, args, verify=True):
     return network
 
 
+@reqs.description("Validate sample Jinja templates")
+@reqs.supports_methods("log/private")
+def test_jinja_templates(network, args, verify=True):
+    # Note: This test should not be chained in the test suite as it creates
+    # a new user
```
jumaffre, line 131: Why is this an issue?

tests/test_python_cli.sh
```diff
@@ -10,24 +10,3 @@ set -ex
 keygenerator.sh --help
```
jumaffre, line 10: It looks like we can get rid of this file too.

tests/governance.py
eddyashton: Looks like it's not! I copied this from the function above, but since we don't use a custom `LoggingTxs` object, and remove the user afterwards, it feels like that shouldn't be a problem for the suite. However, it does rely on a custom arg which doesn't feel worth adding to the suite? So I'll remove the comment, but not chain it.

tests/npm-app/build_bundle.js
jumaffre: Wouldn't it be nice to ship this script as part of the `ccf-app` npm package? Then we could document its usage rather than explaining to users how to include the content of `app.json` in the `set_js_app` proposal?

tests/test_python_cli.sh
jumaffre: Apart from `--help`, we already call it to create members/users' keys/certs in the CI so I think it's OK to delete this file.

tests/npm-app/build_bundle.js
```diff
@@ -0,0 +1,50 @@
...
+import { join } from "path";
+
+const args = process.argv.slice(2);
+
+const getAllFiles = function (dirPath, arrayOfFiles) {
```
letmaik, line 6: no need for lambdas here

tests/npm-app/build_bundle.js
```diff
@@ -0,0 +1,50 @@
...
+  var files = readdirSync(dirPath);
```
letmaik, line 7:
```suggestion
  const files = readdirSync(dirPath);
```

tests/npm-app/build_bundle.js
```diff
@@ -0,0 +1,50 @@
...
+
+  arrayOfFiles = arrayOfFiles || [];
+
+  files.forEach(function (file) {
```
letmaik, line 11: Use a normal `for(.. of ..)`

tests/npm-app/build_bundle.js
```diff
@@ -0,0 +1,50 @@
...
+const removePrefix = function (s, prefix) {
+  return s.substr(prefix.length);
+};
+
+const rootDir = args[0];
+
+const metadataPath = join(rootDir, "app.json");
+const metadata = JSON.parse(readFileSync(metadataPath, "utf-8"));
+
+const srcDir = join(rootDir, "src", "/");
```
letmaik, line 32: The final `/`, also in combination with join, seems really strange.

tests/npm-app/build_bundle.js
achamayou: I don't think we want to increase the number of CLI tools we ship and support, or the number of languages we do it in. I think we especially want to avoid doing that in JS, considering the dependency climate in the npm world. I suggest we use the ccf-app package strictly to ship TypeScript interfaces, unless there's material user demand to the contrary that we think we have the resources to provide.

achamayou (approved):

tests/npm-app/build_bundle.js
eddyashton: Thanks for the feedback, I've incorporated those suggestions.

This trailing `/` is actually because we want to strip it in `removePrefix`. I've added a comment to clarify that, and removed it from the join.

tests/npm-app/build_bundle.js
eddyashton: As discussed - we want to make a distinction between our fully-maintained API and helpful samples. This is, for now, fully in the latter camp.

jumaffre (approved):

PR #3401, opened 2022-01-13
Move constitutions to samples directory

eddyashton:
This seems to be a more natural home for these files.


CHANGELOG.md
```diff
@@ -249,7 +249,7 @@ Browse our [documentation](https://microsoft.github.io/CCF/main/index.html) to g
 
 ### Changed
 
-- The default constitution no longer contains `set_service_principal` or `remove_service_principal` since they are not used by the core framework. Instead any apps which wish to use these tables should add them to their own constitution. A [sample implementation](https://github.com/microsoft/CCF/tree/main/src/runtime_config/test/service_principals/actions.js) is available, and used in the CI tests.
+- The default constitution no longer contains `set_service_principal` or `remove_service_principal` since they are not used by the core framework. Instead any apps which wish to use these tables should add them to their own constitution. A [sample implementation](https://github.com/microsoft/CCF/tree/main/samples/constitutions/test/service_principals/actions.js) is available, and used in the CI tests.
```
achamayou, line 252:
```suggestion
- The default sample constitution no longer contains `set_service_principal` or `remove_service_principal` since they are not used by the core framework. Instead any apps which wish to use these tables should add them to their own constitution. A [sample implementation](https://github.com/microsoft/CCF/tree/main/samples/constitutions/test/service_principals/actions.js) is available, and used in the CI tests.
```

CMakeLists.txt
```diff
@@ -208,10 +208,10 @@ install(
 )
 
 install(PROGRAMS tests/sandbox/sandbox.sh DESTINATION bin)
-install(FILES src/runtime_config/default/actions.js DESTINATION bin)
-install(FILES src/runtime_config/default/validate.js DESTINATION bin)
-install(FILES src/runtime_config/sandbox/resolve.js DESTINATION bin)
-install(FILES src/runtime_config/default/apply.js DESTINATION bin)
+install(FILES samples/constitutions/default/actions.js DESTINATION bin)
```
eddyashton, line 211: `samples/constitution` or `samples/constitutions`? I went with the latter because:
- Consistency with the other folders under `samples`, which are mostly pluralised
- The subfolders are mostly distinct, named constitutions

samples/constitutions/sandbox/sandbox/resolve.js
```diff
@@ -0,0 +1,7 @@
+export function resolve(proposal, proposerId, votes) {
+  // The constitution in a real CCF application deployment would at least
+  // count votes and compare to a threshold of members, but in the sandbox sample,
```
achamayou, line 3:
```suggestion
  // count votes and compare to a threshold of active members, but in the sandbox sample,
```

eddyashton: This file shouldn't exist (it's a new `sandbox/sandbox` file, I don't know how I did that), but I'll make this change in the correct copy.

CHANGELOG.md
eddyashton: I think tweaking old CHANGELOG entries isn't worth doing. Updating the link so it still works is right on the cusp, but tweaking the wording feels like a sin.

PR #3404, opened 2022-01-14
Fix `lts_compatibility` breaks in Daily CI

eddyashton:


jumaffre (approved):

tests/infra/consortium.py
```diff
@@ -590,10 +590,7 @@ def transition_service_to_open(self, remote_node):
             if r.body.json()["state"] == infra.node.State.PART_OF_NETWORK.value:
                 is_recovery = False
 
-        proposal_body, careful_vote = self.make_proposal(
-            "transition_service_to_open",
-            args=None if self.consensus == "CFT" else {"nonce": str(uuid.uuid4())},
-        )
+        proposal_body, careful_vote = self.make_proposal("transition_service_to_open")
```
eddyashton, line 593: If we want to pass this, it should be `self.make_proposal("transition_service_to_open", nonce = str(uuid.uuid64()))`. But the logic in `make_proposal` is currently "if there are any kwargs, create an `args` object, and populate it with everything that's non-None". So it's hard to make that optionally an `args` object, or no `args` field at all. Rather than forcing that, I've just removed this brach, since `self.consensus == "CFT"` is always true.

PR #3407, opened 2022-01-14
Add an Edit link in docs footer

eddyashton:
Our docs were missing an "Edit on GitHub" button. I had a look for an off-the-shelf extension, but there don't seem to be any with clean config or that work directly with the Furo theme.

Instead I'm just inserting it manually into the footer, via a `<script>` tag that modifies the default footer.

If we want to fiddle with the template further we could put it elsewhere on the page, but it seems most naturally to be next to `Show Source`, which we have no control over with this theme.

Looks like this (hovering to show link destination):
image

### DIFF
--- a/doc/_templates/page.html
+++ b/doc/_templates/page.html
@@ -1,0 +1,11 @@
+{#-
+  This extends page.html from the Furo template to add an Edit on GitHub button to the footer.
+-#}
+{% extends "!page.html" %}
+{% block footer %}
+  {{ super() }}
+  <script>
+    var relatedInfo = document.querySelector(".related-information");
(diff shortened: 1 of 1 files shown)


achamayou (approved):

achamayou:
The only mild problem with this is that it goes to main even when the release is based on a release branch (eg. release/1.x). I don't think that's particularly easy to address though.

eddyashton:
> The only mild problem with this is that it goes to main even when the release is based on a release branch (eg. release/1.x). I don't think that's particularly easy to address though.

I think that should be possible (we have a configuration variable with the current version, so we just need to map that to a branch name). But if we have releases for 1.5 and 2.1, do we want the docs for 1.3 to have an `Edit me` button? Even if that led to the `release/1.x` branch, you may be editing a more recent version. Maybe we should be removing the `Edit` button for everything but `main`?

achamayou:
@eddyashton there could still be a typo or something worth fixing on a release branch, so I'd be inclined to stay that for all `N.0.*`, the Edit button should edit the head of `release/N.x`. But we could equally disable it for all but main.

PR #3408, opened 2022-01-14
Fix `test_join_straddling_primary_replacement`

jumaffre:
Another `reconfiguration.py` fix. I incorrectly changed the logic of the `test_join_straddling_primary_replacement` test a while back when refactoring infra code (see https://github.com/microsoft/CCF/pull/2595/files#diff-db3e477cfd37d006e74550f29fc15823d8650c8a8da4bcc27734fb2deef5b4db): the new node shouldn't be transition to trusted until the next proposal.


eddyashton (approved):

achamayou (approved):

jumaffre:
As discussed with @wintersteiger, this fix raises an issue with the 2tx reconfiguration scheme.

The extra `network.trust_node()` hid the two-actions-in-one-proposal created below, which meant the change to the network configuration table looked like this:
```
table "public:ccf.gov.nodes.network.configurations" (3 writes):
      <u64: 0>:
        {
          "nodes": [],
          "rid": 6
        }
      <u64: 5>:
        {
          "nodes": [
            "eb0f1a5a91fa4cdc594a562b69c11f126e37a9bf2e3a1ce696890525356bce5d",
            "19e97c31bc4bd254a7c6a10b63c4ee4ee08b3f8e5468b8f330a0684b124c38ce",
            "de559970c5eae6452a437e3665725cae4922becb1c3ad5714868b665bca86443",
            "b1150bf5f977cea8129557f3983d7d55471cde8acd4d5a06420dd723e2f02653",
            "e44b79c118f63a688f3f54fb9739c1b0d89bc00df7afe697b8dc8e1b4739affb"
          ],
          "rid": 5
        }
      <u64: 6>:
        {
          "nodes": [
            "e44b79c118f63a688f3f54fb9739c1b0d89bc00df7afe697b8dc8e1b4739affb",
            "b1150bf5f977cea8129557f3983d7d55471cde8acd4d5a06420dd723e2f02653",
            "de559970c5eae6452a437e3665725cae4922becb1c3ad5714868b665bca86443",
            "eb0f1a5a91fa4cdc594a562b69c11f126e37a9bf2e3a1ce696890525356bce5d"
          ],
          "rid": 6
        }
```

However, the consensus hooks for this table and the reconfiguration logic in the consensus expects only single changes to this table for the scheme to work correctly. We believe the best scheme is to re-work the schema of the `ublic:ccf.gov.nodes.network.configurations` table and have a single value of all recorded nodes in the current configuration, i.e. `0 -> {"nodes": [...], "rid": x}`. That way, a single transaction can update the configuration without having to think of its local writes vs. the committed state.

PR #3409, opened 2022-01-14
Remove retired nodes from store

### LINKED ISSUE #2320: Remove retired nodes from the KV
Follow-up from #1401 

Deleting the primary node from the store on retirement isn't as straightforward as it first seemed: backup nodes won't be able to verify the signature that follows (which node signed this signature transaction?), and won't be able to commit the retirement of the primary node. 

Instead, the primary node shouldn't be deleted straight away but be marked as `Retired` and then deleted later, once it is safe to do so (the node will no longer emit signature transactions).

The cleanest way to trigger the deletion of the retired primary probably is when a new node becomes primary: the new node can look up all nodes that are marked as `Retired` and actually deletes them from the store. Note that this probably can't be done synchronously in Raft's `become_leader()` because of consensus locking but should be scheduled asynchronously at that point instead.

Note that it is safe to delete backups on retirement straight away as they will not issue signature transactions.

jumaffre:
Fix #2320 

This PR adds automatic cleanup of `Retired` nodes from the store/ledger, which reduces the memory pressure in the enclave for long-lived services (note: a single node entry currently uses about ~20KB memory). 

This is achieved by triggering a new internal endpoint (`DELETE /node/network/nodes/retired`) when a new consensus configuration is committed (from `raft.h`). The changes required for this are in the end pretty small. In particular, the Python `ccf.ledger` module can now cope with node deletions. 

~**Warning: Node deletion is not supported in 1.x (see `hooks.h`) so this PR will have to be merge once #3412 is merged and released.**~ 1[.0.17 has now been released](https://github.com/microsoft/CCF/releases/tag/ccf-1.0.17).

TODO: 
- [x] 1tx reconfiguration
- [x] 2tx reconfiguration


tests/lts_compatibility.py
```diff
@@ -111,6 +111,14 @@ def test_new_service(
 
     test_all_nodes_cert_renewal(network, args)
 
+    with primary.client() as c:
+        r = c.get("/node/network/nodes")
+        nodes = r.body.json()["nodes"]
+        LOG.error(nodes)
+
+    # TODO: Check that all nodes are eventually removed from the store!
+    # Waiting on https://github.com/microsoft/CCF/pull/3412 and 1.x release
```
jumaffre, line 120: Waiting on 1.0.17 release.

achamayou (approved):

PR #3416, opened 2022-01-18
Clarify version dependency between ccf-app and npm-app

eddyashton:


doc/build_apps/js_app_bundle.rst
```diff
@@ -13,12 +13,7 @@ This page documents the components of a bundle and the JavaScript API available
     `TypeScript <https://www.typescriptlang.org/>`_.
     CCF provides an example app built with these tools.
     They involve a `build` step that generates an app bundle suitable for CCF.
-    See the following for more details:
-
-    .. toctree::
-      :maxdepth: 1
-
-      js_app_ts
+    See the :ref:`TypeScript Application <build_apps/js_app_ts>` section for more details
```
eddyashton, line 16: This additional `toctree` caused `TypeScript Application` to be referenced twice:

image

achamayou: I'm not convinced of the benefits of one more level of depth here, I think two levels of toctree is enough.

doc/build_apps/js_app_ts.rst
```diff
@@ -57,8 +58,8 @@ Dependencies
 
 The sample uses several runtime and development packages (see ``package.json``).
 One of them is the :typedoc:package:`ccf-app` package.
-This package is referenced locally using ``file:``.
-You should replace this with a reference to a recently published version (adjust the version number accordingly), to match the version of CCF you are running:
+This package references the current branch's version of the ``ccf-app`` package using ``file:``.
+To test against a published version you should adjust the version number accordingly::
```
achamayou, line 62: Extra colon?

achamayou:
For Python code, Sphinx can insert a source code link, and multiversion does seem to handle that correctly now: https://github.com/Holzhaus/sphinx-multiversion/issues/40

Getting the links to point to the right place seems possible with a template (https://holzhaus.github.io/sphinx-multiversion/master/templates.html#version-banners), but that's not particularly nice to use from rst. I don't know how easy it would be extend sphinx-multiversion with a :srcpath:`dir/file.cpp` directive, but it might be an interesting way to go.

achamayou (approved):

eddyashton:
> For Python code, Sphinx can insert a source code link, and multiversion does seem to handle that correctly now: [Holzhaus/sphinx-multiversion#40](https://github.com/Holzhaus/sphinx-multiversion/issues/40)
> 
> Getting the links to point to the right place seems possible with a template (https://holzhaus.github.io/sphinx-multiversion/master/templates.html#version-banners), but that's not particularly nice to use from rst. I don't know how easy it would be to extend sphinx-multiversion with a :srcpath:`dir/file.cpp` directive, but it might be an interesting way to go.

Yeah I considered templates, but I'm not sure if we can have a template that generates RST that becomes HTML, or if we can retain a template through the RST->HTML conversion so it can be picked up by the templating later.

Anyway, @jumaffre has found an extension that does basically the latter approach, so we'll use that.

doc/build_apps/js_app_bundle.rst
eddyashton: Sure, I'll raise both of these to go under `Build Apps`.

Issue #857, opened 2020-02-18
Enable error logging during initial create_node ECALL


eddyashton:
We currently make 2 ECALLs: `enclave_create_node` and `enclave_run`. This makes the code in `enclave_create_node` special, it runs before the host is pumping the outbound ringbuffer so its log messages are ignored on failure.

We should combine these, with a single `enclave_run` ECALL which does creation as its first step. The outputs from the current `create` (node cert etc) can be written over the ringbuffer. This way any errors during processing of the `Create` RPC will be easier to track, as they will fail through similar paths to a standard RPC.

achamayou:
@eddyashton having had a quick look now, I think the merge may be more trouble than it's worth, because what the split allows is a single-threaded setup phase for the node, before spawning all the threads that enclave_run().

A single merged create_node_and_run() would need to race for a setup lock, block while the winner sets up etc.

I think we could realise much if not all the benefit of the merge just by consuming log messages from the ringbuffer in the event of a failed create_node()?

eddyashton:
@achamayou Since the 'first' thread that calls `enclave_run` already takes a very different path, I think it would still be helpful to collapse that into a single ECALL? When the node is sufficiently 'created' that the host wants to create additional worker threads, those worker threads can either reuse the same ECALL and skip all of the creation, or use a still-separate ECALL. The latter seems nice, to more clearly distinguish the main thread from the workers, outside of "which managed to mark themselves as a worker first". That probably still requires a blocking loop somewhere, where the main enclave thread waits for the workers to initialise themselves before it proceeds to anything that assumes `ThreadMessaging` is initialised, but that doesn't sound too onerous?

The main goal is definitely consuming log messages/common error reporting. But I think any other approach to that would also be complex - to pump the ringbuffer when `create_node` fails, we need to re-do a chunk of the setup that we would do in the successful case, for a while, and then exit?

achamayou:
I don't think I understand the simplification we would get from this merge. Once the RB is setup, we can and do log, so the dispatcher/workers distinction doesn't seem super useful?

We do have some initial setup though, which is more easily done single threaded in `create_node`, if we just make sure we set up the RB early, and set a failure RC otherwise, we can on failure go pump messages out rather than continue with the setup?

PR #3419, opened 2022-01-18
Simplify `logger.h`

eddyashton:
This was a refactor I did before Christmas as a stepping stone to #857 - opening as a PR now to avoid it rotting.

The main thing is that it removes 175 lines from `logger.h`. It looks like a large diff there, but is largely just re-ordering existing functions.

- This reduces the number of static global objects. Some of the code for time management/offset calculation/writing to the ringbuffer has moved out of `logger.h`, and is now scoped to an appropriate calling point.
- This makes the edge-case behaviour more well-defined. If no loggers have been configured yet, the logging macros will no longer potentially produce access violations (dependent on a collection of macros) - instead we'll simply drop the log. The default is to have no loggers, so the unit tests will no longer output generally useless logging lines (unless we explicitly tell them to).
- Each logger instance is an `AbstractLogger`, and there are 3 implementations of that; write text to stdout; write JSON to stdout; write to the ringbuffer. If we ever want to log to multiple files, it's obvious where that plugs in (and that code wouldn't need to reason about any potentially default-constructed loggers).


src/ds/logger.h
```diff
@@ -38,67 +29,115 @@ namespace logger
...
 
-  static constexpr long int ns_per_s = 1'000'000'000;
+  static constexpr const char* LevelNames[] = {
+#ifdef VERBOSE_LOGGING
+    "trace",
+    "debug",
+#endif
+    "info",
+    "fail",
+    "fatal"};
```
jumaffre, line 39: Probably not in the scope of this PR but mentioning it here as related. For releases, this enum only contains 3 values (info, fail and fatal), which is as expected for the enclave but a little too strict on the host as the log level isn't as sensitive and can be set at launch (see https://microsoft.github.io/CCF/main/operations/configuration.html#logging). For example, setting the host log level to "debug" is currently not possible for CCF releases. 

I had a quick go at trying to define different enums for the host and the enclave but it got messy quickly. Would the changes in this PR make this change easier?

eddyashton: I think we could achieve that just by changing the guards on `trace` and `debug` from `#ifdef VERBOSE_LOGGING` to `#if defined(VERBOSE_LOGGING) or not defined(INSIDE_ENCLAVE)`, so they're always available on the host. But is there much value in this? Do we have debug logging on the host that we'd like to make available in a release? As I write this, I think "probably", so I'll see if it really is that esy.

eddyashton: Yeah this should be possible, modulo some adjustments to make sure the enum values match up between host and enclave, but it's failing without a clear error so I'm going to skip it for now.

src/host/handle_ring_buffer.h
```diff
@@ -48,13 +48,38 @@ namespace asynchost
...
-            thread_id,
-            msg,
-            log_time_us_count);
+          logger::LogLine ll(
+            log_level, file_name.c_str(), line_number, thread_id);
+          ll.msg = msg;
+
+          // Represent offset as a real (counting seconds) to handle both small
+          // negative _and_ positive numbers. Since the system clock used is not
+          // monotonic, the offset we calculate could go in either direction,
```
jumaffre, line 57: What's the reason we don't we use `steady_clock` again?

src/perf_client/scenario_perf_client.cpp
```diff
@@ -130,6 +130,9 @@ class ScenarioPerfClient : public Base
 
 int main(int argc, char** argv)
 {
+  logger::config::loggers().emplace_back(
+    std::make_unique<logger::TextConsoleLogger>());
```
jumaffre, line 134: Could we have a helper `logger::config::easy_init()` or something similar for a simpler API to be used in tests?

jumaffre (approved):

src/host/handle_ring_buffer.h
eddyashton: I'm not sure. Perhaps something to do with converting to printable C types? "It is the only C++ clock that has the ability to map its time points to C-style time." But I don't see a strong requirement for it.

src/perf_client/scenario_perf_client.cpp
eddyashton: Sure, I'll add that.

cmake/ccf_app.cmake
```diff
@@ -186,7 +186,11 @@ function(add_ccf_app name)
     )
 
     if(NOT SAN)
-      target_link_options(${virt_name} PRIVATE LINKER:--no-undefined)
+      target_link_options(
+        ${virt_name}
+        PRIVATE
+        LINKER:--no-undefined,--undefined=enclave_create_node,--undefined=enclave_run
```
eddyashton, line 192: These symbols are defined in `enclave/main.cpp`, which gets built into `libccf.virtual.a` and `libccf.enclave.a`, which then get linked to by CCF apps into libraries like `liblogging.virtual.so`. Something in this branch, though I don't know what, changed the behaviour of these symbols, and they were no longer exported from the combined `.so`. In classic fashion, I'm now not sure why this _ever_ worked without this line, which tells the linker to explicitly re-export those symbols from the shared object, even though they're not used _by_ the shared object. It works now, afaict.

PR #3421, opened 2022-01-18
Remove `public:ccf.gov.nodes.network.configuration` KV table and simplify reconfiguration hook logic

### LINKED ISSUE #3410: Network configuration table does not support multiple updates within same Tx
Identified here: https://github.com/microsoft/CCF/pull/3408#issuecomment-1013319372

The existing `public:ccf.gov.nodes.network.configurations` table schema should be modified to easily support multiple updates from the same `kv::Tx`. The new proposed scheme would be to have a single value of all recorded nodes in the current configuration, i.e. `0 -> {"nodes": [...], "rid": x}`. That way, a single transaction can update the configuration without having to think of its local writes vs. the committed state.

jumaffre:
Resolves #3410 
Unblocks #3408 

This PR removes the existing `public:ccf.gov.nodes.network.configuration` table and uses the existing hook on the `public:ccf.gov.nodes.info` table to handle all the reconfiguration logic in the consensus.


src/consensus/aft/raft.h
```diff
@@ -517,12 +515,28 @@ namespace aft
...
         }
+
+        if (orc_sets.find(idx) == orc_sets.end())
+        {
+          orc_sets[idx] = {};
+        }
+
+        if (resharing_tracker)
+        {
+          kv::NetworkConfiguration netconfig = {idx, {}}; // TODO: Fix
```
wintersteiger, line 526: The resharing tracker needs to know the exactly list of nodes in the next configuration (the nodes in that config need to agree on a new distribution of key shares and it can't include nodes from earlier or later configurations). Can we still get that by running over `configurations` and match on the `rid`?

src/consensus/aft/raft.h
```diff
@@ -614,8 +629,13 @@ namespace aft
     // guaranteed that we will eventually receive all of them, since all
     // nodes keep re-submitting ORCs until they are able to switch to the next
     // pending configuration.
-    bool orc(kv::ReconfigurationId rid, const ccf::NodeId& node_id) override
+
+    std::optional<kv::Configuration::Nodes> orc(
+      kv::ReconfigurationId rid, const ccf::NodeId& node_id) override
     {
+      // TODO: What about lock?
```
wintersteiger, line 636: This is called from the endpoint and if I remember correctly, the lock is already held there, at least when we're in BFT mode. But then, BFT may be going away anyways...

src/consensus/aft/raft.h
```diff
@@ -626,34 +646,44 @@ namespace aft
...
+          else
+          {
+            oit->second.insert(node_id);
+            LOG_DEBUG_FMT(
+              "Configurations: have {} ORCs out of {} for configuration #{}",
+              oit->second.size(),
+              ncnodes.size(),
+              rid);
+
+            // Note: Learners in the next configuration become trusted when
```
wintersteiger, line 668: Correct, good idea to spell that out here.

src/node/hooks.h
```diff
@@ -94,33 +94,31 @@ namespace ccf
...
       // because the hooks are sorted by table name.
       assert(
-        std::string(Tables::NODES) < std::string(Tables::NODES_CONFIGURATIONS));
+        std::string(Tables::NODES) < std::string(Tables::NODES_CONFIGURATION));
 
-      for (const auto& nc : configs)
-      {
-        consensus->reconfigure(version, nc);
-      }
+      // TODO: Not called anymore!
```
wintersteiger, line 120: If this all works, I guess we should remove the hook code altogether?

src/consensus/aft/raft.h
jumaffre: I believe it is possible but haven't explored this yet as I'm waiting for the CI to run on the current changes first.

src/consensus/aft/raft.h
jumaffre: Good point. It looks like it works with the lock in CFT mode so let's see.

achamayou (approved):

PR #3422, opened 2022-01-18
tlstest.sh runs multiple times, compares to golden

### LINKED ISSUE #3365: SSL test (TLS Report) should fail if different

rengolin:
The tlstest.sh script fails if the tls_report.* files exist. This commit
removes the files, if they exist, before running the tests, to avoid
spurious failures during multiple re-runs of all tests on a local
machine.

Also adding golden file to tlstest to compare to output, filtering the
unstable output (IP address, hashes, comments, etc).

This makes sure we actually check the output and make sure there aren't
any changes, or if there are, they need to be copied to the new golden
file and show in the diff of the commit.

Fixes #3365


tests/tls_report.log
```diff
@@ -0,0 +1,182 @@
+
+
+ Service detected:       HTTP
+
+
+[1m[4m Testing protocols [m[4mvia sockets except NPN+ALPN [m
```
eddyashton, line 6: Can we strip out these non-printing characters and colour codes? It doesn't matter for golden file comparison, but does make the file less human-readable. I'm sure I had a regex for finding these colour codes once, but I can't find it now.

rengolin: I intentionally didn't, because when they mismatch, the output comes out coloured too. :)

For human readable output, just `cat tls_report.log`.

But I can try to do that on both sides, so the diff still passes, if that's the preference...

eddyashton: Ah-ha, found it:
`escape_sequence_regex = re.compile(r"\033\[(\d+(?:;\d+)*)m")`

tests/tlstest.py
```diff
@@ -6,17 +6,71 @@
...
+            continue
+        filtered.append(line)
+
+    # Diff stable output on both files
+    success = True
+    for line in difflib.unified_diff(
+        golden, filtered, fromfile="Golden", tofile="Output", lineterm=""
+    ):
+        print(line)
+        success = False
```
eddyashton, line 50: How do we produce a new golden file? It would be nice if, on failure, we dumped the (filtered) output to disk, as a candidate to be dropped in place of the golden file. This is roughly what we do for our other golden file tests.

rengolin: Right, good point. Before the grep, it was just a copy, now, it's a filtered version. I'll print into a file if different.

tests/tls_report.log
eddyashton: Ah cool, I assumed the diff tool would apply its own colouring rather than retaining these colour codes! Happy to keep it for that feature in that case.

tests/tlstest.py
```diff
@@ -6,17 +6,71 @@
...
+    with open("tls_report.log", encoding="UTF-8") as o:
+        output = o.readlines()
+
+    # Golden file is clean of random output (IP address, hashes)
+    # New log needs to clean of random output
+    reCOMMENT = re.compile(r"^#")
+    reIPADDR = re.compile(r"\d+\.\d+\.\d+\.\d+:?\d*")
+    reFIELDS = re.compile(r"(Serial|Fingerprints|subjectAltName|Certificate Validity)")
+    reHASH = re.compile(r"SHA\d+ [0-9A-F]{20,}")
+    filtered = list()
```
achamayou, line 32:
```suggestion
    filtered = []
```
Always use literals when you can, they're shorter, faster, and can't be shadowed.

tests/tlstest.py
```diff
@@ -6,17 +6,71 @@
...
+import os
+import difflib
+import re
+
+
+def compare_golden():
+    script_path = os.path.realpath(__file__)
+    script_dir = os.path.dirname(script_path)
+    golden_file = os.path.join(script_dir, "tls_report.log")
+    print("Comparing output to golden file " + golden_file)
```
achamayou, line 18:
```suggestion
    print(f"Comparing output to golden file: {golden_file}")
```

achamayou (approved):

rengolin:
Not quite. Both the HTML and test reports have inconsistent line breaks (extensions) depending on the terminal width. I understand that for the log but not for the HMTL, but I guess they're using the same string.

I might have to compare the CSV or JSON outputs instead and filter differently. Flashbacks from my early days of parsing Apache and sendmail logs with Perl...

PR #3428, opened 2022-01-19
Remove more mbedTLS leftovers

wintersteiger:
AFAICT these formatters are not used anymore and nothing depends on the output format of the logging app responses.


achamayou (approved):
Well spotted!

eddyashton (approved):

rengolin:
ah, right, I was about to do that and didn't know if they were really used or not.

I'm also doing the error codes and cleaning up some comments, and will create a PR after yours.

Issue #3429, opened 2022-01-19
Re-work TLS to match OpenSSL's style


rengolin:
After #3361 and #3394, there's only an OpenSSL implementation of TLS in CCF. However, because we have MbedTLS as the previous implementation, and we wanted to play safe on the migration, we emulated their style with OpenSSL to match.

Now that MbedTLS is completely removed from the tree, we can finally look at the TLS implementation (including the TLS endpoint) and adapt it to match the OpenSSL style, and consequently reduce the complexity of the code, increasing manitainability.

Things to look out for:
 * The callback was the only way MbedTLS could operate, but OpenSSL can use BIOs, so we should change the endpoint's handlers to match.
 * OpenSSL returns success and error code separately, so we should remove the "negative return hack".
 * Wants read/write and other non-errors are being used in the return value and we need to check it on every read. We should just use the return value of the read/write and check for status/errors on the caller.
 * Make sure we still need the certificate to be null-terminated. If not, there's a bunch of logic we can clean up.

I'm sure there are more, these are just the bigger ones.

achamayou:
Closed by #8117

PR #3430, opened 2022-01-19
Remove the final MbedTLS references from the code

rengolin:
This commit cleans up the final references to MbedTLS on the code,
making sure what's left is consistent with OpenSSL.

The only remaining references to "MbedTLS" are the ones that we
need to worry about when we work on #3429.

### DIFF
--- a/src/crypto/pem.h
+++ b/src/crypto/pem.h
@@ -13,8 +13,7 @@
 
 namespace crypto
 {
-  // Convenience class ensuring null termination of PEM-encoded certificates as
-  // required by mbedTLS
+  // Convenience class ensuring null termination of PEM-encoded certificates
   class Pem
   {
     std::string s;
--- a/src/crypto/test/crypto.cpp
+++ b/src/crypto/test/crypto.cpp
@@ -378,7 +378,6 @@ void run_csr(bool corrupt_csr = false)
 
   std::vector<SubjectAltName> subject_alternative_names;
 
-  // mbedtls doesn't support parsing SAN from CSR
   if constexpr (std::is_same_v<T, KeyPair_OpenSSL>)
   {
     subject_alternative_names.push_back({"email:my-other-name", false});
--- a/src/crypto/test/crypto.cpp
+++ b/src/crypto/test/crypto.cpp
@@ -378,19 +378,11 @@ void run_csr(bool corrupt_csr = false)
 
   std::vector<SubjectAltName> subject_alternative_names;
 
-  if constexpr (std::is_same_v<T, KeyPair_OpenSSL>)
-  {
-    subject_alternative_names.push_back({"email:my-other-name", false});
-    subject_alternative_names.push_back({"www.microsoft.com", false});
-    subject_alternative_names.push_back({"192.168.0.1", true});
-    valid_from = "20210311000000Z";
-    valid_to = "20230611235959Z";
-  }
-  else
-  {
-    valid_from = "20210311000000";
-    valid_to = "20230611235959";
-  }
+  subject_alternative_names.push_back({"email:my-other-name", false});
+  subject_alternative_names.push_back({"www.microsoft.com", false});
+  subject_alternative_names.push_back({"192.168.0.1", true});
+  valid_from = "20210311000000Z";
+  valid_to = "20230611235959Z";
 
   auto csr = kpm.create_csr(subject_name, subject_alternative_names);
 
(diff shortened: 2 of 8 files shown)


src/crypto/pem.h
```diff
@@ -13,8 +13,7 @@
 
 namespace crypto
 {
-  // Convenience class ensuring null termination of PEM-encoded certificates as
-  // required by mbedTLS
+  // Convenience class ensuring null termination of PEM-encoded certificates
```
achamayou, line 16: Do we still need to null terminate certs now? If not, we might as well get rid of this logic?

achamayou: I guess we probably do for compatibility with the current LTS.

tests/infra/path.py
```diff
@@ -61,7 +61,7 @@ def cert_bytes(cert_file_name):
                 chars.append(c)
             else:
                 break
-        # mbedtls demands null-terminated certs
+        # null-terminated certs, just in case
```
achamayou, line 64: For compatibility, more likely.

achamayou (approved):

src/crypto/test/crypto.cpp
```diff
@@ -378,7 +378,6 @@ void run_csr(bool corrupt_csr = false)
 
   std::vector<SubjectAltName> subject_alternative_names;
 
-  // mbedtls doesn't support parsing SAN from CSR
   if constexpr (std::is_same_v<T, KeyPair_OpenSSL>)
```
jumaffre, line 381: The `else` branch should be deletable now I believe.

jumaffre: And there's no longer the need for the loop in `TEST_CASE("Create sign and verify certificates")`.

src/crypto/pem.h
rengolin: Probably not, but since this is a cleanup PR, I'd rather leave that to #3429. I'll add a comment on that issue.

tests/infra/path.py
rengolin: Yeah, same answer as above.

src/crypto/test/crypto.cpp
rengolin: Of course! Thanks!

src/tls/cert.h
```diff
@@ -60,9 +60,8 @@ namespace tls
       {
         // Peer hostname is only checked against peer certificate (SAN
         // extension) if it is set. This lets us connect to peers that present
-        // certificates with IPAddress in SAN field (mbedtls does not parse
-        // IPAddress in SAN field). This is OK since we check for peer CA
-        // endorsement.
+        // certificates with IPAddress in SAN field. This is OK since we check
```
jumaffre, line 63: Oh, that's an interesting one! I've created https://github.com/microsoft/CCF/pull/3431 to add further verification in the join protocol TLS handshake.

src/enclave/tls_endpoint.h
```diff
@@ -554,9 +552,9 @@ namespace enclave
 
     // These callbacks below are complex, using the callbacks above and
     // manipulating OpenSSL's BIO objects accordingly. This is just so we can
-    // emulate MbedTLS.
-    // Once we get rid of MbedTLS we can move the callbacks above to handle BIOs
-    // directly and hopefully remove the complexity below.
+    // emulate what MbedTLS used to do.
+    // Now that we have removed it from the code, we can move the callbacks
+    // above to handle BIOs directly and hopefully remove the complexity below.
```
jumaffre, line 557: Does this deserve its own GitHub issue, or a link to an existing one?

src/crypto/test/crypto.cpp
rengolin: > And there's no longer the need for the loop in `TEST_CASE("Create sign and verify certificates")`.

I'm not sure I agree. This is related to corrupting the CSR and checking for the throw, not any previous mbedTLS behaviour.

src/tls/cert.h
rengolin: Should this comment change in that PR, too?

src/enclave/tls_endpoint.h
rengolin: This is #3429, should we add that to the comment, too?

jumaffre: Yes, that'd be super useful!

src/tls/cert.h
jumaffre: I think the comment as you've re-phrased it still holds.

jumaffre (approved):

wintersteiger (approved):

PR #288, opened 2019-08-07
Connections test

achamayou:
Test for #264:

1. Cap max fds
2. Spawn many more concurrent connections than there are free fds on the process
3. Disconnect
4. Establish connections again

I've run this quite a few times without problem so far, I'm going to:

1. Add some RPCs after stages 1 and 4
2. Add asserts

Thoughts on how this can be improved are welcome.

PR #3433, opened 2022-01-19
Handle UBSAN failures in connections test

wintersteiger:
Relates to https://github.com/microsoft/CCF/pull/288

Here are my notes on the UBSAN failures in the connections tests. It turns out that these failures are legit and that they happen during normal execution, not during cleanup before exit as we initially thought. The problem is that `fmt::format` indirectly requires free file descriptors, of which there are none left when the `fopen` in `ledger.h` fails. It's perfectly happy if I free up some FDs manually, or when I use `snprintf` instead of `fmt::format`. 

I strace'd it and found this:

```
...
write(1, "2022-01-19T15:20:52.580255Z -0.0"..., 190) = 190
openat(AT_FDCWD, "0.ledger/ledger_523", O_RDWR|O_CREAT|O_TRUNC, 0666) = -1 EMFILE (Too many open files)
pipe(0x7ffc8e99fb68)                    = -1 EMFILE (Too many open files)
pipe(0x7ffc8e99fb58)                    = -1 EMFILE (Too many open files)
ioctl(2, TCGETS, 0x7ffc8e99fa30)        = -1 ENOTTY (Inappropriate ioctl for device)
... UBSAN complaint output ...
+++ exited with 1 +++
```

Note the two calls to `pipe`, which do not happen when I use `snprintf` or when I don't try to print `file_path` in the format call.

While playing around with this, I also triggered a similar problem in `snapshots.h`, but only twice all afternoon. I'm not 100% sure that it's exactly the same problem because I haven't gotten lucky enough to trigger the problem while strace'ing. 

I wasn't able to find out what exactly makes those two calls to `pipe`. It's most likely somewhere in the system library (not the STL, but possibly in format.h, need to take another look). Sadly, I think this may not be a cheap fix, as we can't just randomly close file descriptors every time we want to `fmt::format` something...


wintersteiger:
I managed to get an strace of the second occurrence in `snapshots.h`: identical pattern, 2 calls to `pipe()` during format.

achamayou:
This is very worrying, thank you for looking into it! I don't understand how fmt::format() can require fds, and still work in enclave, I must be missing something.

achamayou:
Wild unsubstantiated guess: fmt::format() only needs fds when it formats timestamps, because it looks at tzinfo and that results in it opening an Olson DB file.

achamayou:
Alternate wild guess: strerror() can result in a file access because of localisation.

wintersteiger:
Update: It's UBSAN itself that makes those calls to `pipe` ([here](https://github.com/llvm/llvm-project/blob/c0f9592daae3be17e5fd1528a8f6067cb0c3bd91/compiler-rt/lib/sanitizer_common/sanitizer_posix_libcdep.cpp#L276)). I submitted a report and patch to LLVM ([here](https://reviews.llvm.org/D117793)). If/when that gets merged these problems will go away, but until then I'm looking for a better way to suppress these reports. `ubsan.blacklist` is giving me a bit of trouble; I can't figure out the right expressions to match on.

wintersteiger:
This can now be merged. The connections test now has it's own suppressions for just those few locations where it would get into trouble when no file descriptors are left.

tests/connections.supp
```diff
@@ -0,0 +1,11 @@
+#############################################################################
+
+# Sanitizer runtime suppressions. The undefined behavior sanitizer (specifically
+# the vptr check) produces spurious reports when we run out of file descriptors.
```
achamayou, line 4: Can we include a link to the llvm issue (https://reviews.llvm.org/D117793)?

achamayou (approved):

PR #3434, opened 2022-01-20
Connection log

rengolin:
This doesn't do much but helps with #1486 before we add metrics.

### DIFF
--- a/src/host/tcp.h
+++ b/src/host/tcp.h
@@ -579,7 +579,7 @@ namespace asynchost
 
       if (!DNS::resolve(host, service, this, on_resolved, async))
       {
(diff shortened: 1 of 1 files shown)


src/host/tcp.h
```diff
@@ -577,6 +579,7 @@ namespace asynchost
 
       if (!DNS::resolve(host, service, this, on_resolved, async))
       {
+        LOG_DEBUG_FMT("Bind to '{}' failed", host);
```
achamayou, line 582: That's not trying to bind though?

achamayou (approved):

jumaffre (approved):

PR #3435, opened 2022-01-20
Fix a couple of throws

achamayou:
Drive-by PR while doing something else.


eddyashton (approved):

src/crypto/openssl/public_key.cpp
```diff
@@ -37,7 +37,7 @@ namespace crypto
     key = d2i_PUBKEY_bio(buf, &key);
     if (!key)
     {
-      throw new std::runtime_error("Could not read DER");
+      throw std::runtime_error("Could not read DER");
```
letmaik, line 40: Isn't there some linting for this kind of thing?

Issue #3372, opened 2022-01-10
Remove all but ledger and receipt utility from `ccf` Python package


jumaffre:
Since CCF has been using standard client-server protocols (TLS, HTTP) for a while now, `clients.py` and other client helper functions can be moved back to `tests/infra`. The only remaining modules should be the ones that inspect the ledger or receipts.

PR #3441, opened 2022-01-21
Cleanup `ccf` Python package dependencies

jumaffre:
These should have been removed in #3372


tests/requirements.txt
```diff
@@ -13,3 +13,4 @@ docker
 better_exceptions
 pyasn1
 Jinja2
+httpx
```
achamayou, line 16: I think we probably ought to keep the pin, for stability: httpx == 0.21.*

achamayou (approved):

tests/requirements.txt
jumaffre: While this wasn't the case for our released Python package, I think it's OK if our tests break because an httpx upgrade broke compatibility?

achamayou: I would prefer if they didn't, and dependabot did the update on a PR instead. httpx is still pre-1.x, and a 0.N+1 can and does break (0.20 with allow_redirect for example).

On the LTS branch, I think it's definitely right to pin, and even on main I would prefer if we did.

jumaffre, line 16:
```suggestion
httpx == 0.21.*
```

PR #3442, opened 2022-01-24
Add 1tx to 2tx reconfiguration migration to LTS test

jumaffre:
The migration from 1tx to 2tx is now tested in the LTS test so that we can migrate existing service to the 2tx reconfiguration scheme later on.


wintersteiger (approved):
Nice! I thought it would take a bunch more effort to get this working with the LTS.

PR #3444, opened 2022-01-24
Add system for offloading index to disk

eddyashton:
This PR adds a new system for offloading index storage (or, really, arbitrary enclave data) to disk, and fetching it asynchronously over the ringbuffer. This means we can index the full history of a service without producing infinite memory growth.

First, a note on naming: I've gone back and forth on a bunch of names for this, and haven't found anything that sticks. In initial discussions it was a set of "scratch files", I was initially implementing it as "blob storage", and really it's just a specific kind of "cache" (disk-backed at the moment, but could be host memory - a slow, explicit simulacrum of the actual enclave memory cache). It is currently called LFS, aka Large File Storage, because that seemed appropriate during a recent renaming pass. I'd like to make it clear on the enclave side that this is a way of offloading things out of enclave memory, but be vague about what the host does with it (maybe we put it in a DB for faster lookup, rather than managing a collection of files ourselves?). There's a host-side and an enclave-side component for reading their respective ringbuffer messages, currently called `asynchost::LFSFileHandler` and `ccf::indexing::EnclaveLFSAccess` respectively. If anyone has suggestions for better names, I'm willing to discuss, but it may best just to throw our hands up and say "it's uniquely greppable and _we_ know what it means", for now. Relatedly it's all under `indexing/`/`ccf::indexing::` for now, but should be easy to move out if/when we find a more general use for it.

Anyway, a high-level summary:
- Some new ringbuffer messages for storing/fetching named files.
- An in-enclave interface that sends those messages.
  - It also encrypts and integrity protects any data that is written, and verifies + decrypts anything that is read.
  - The encryption key is entirely node-local, and generated uniquely for this system rather than derived from anything shared. This means the files will only work for this node, this run, and can't be shared between nodes. I think that's sufficient for now, and if we convince ourselves it's safe in future it should be easy to derive a shared key so we can share indexes.
  - Everything is written encrypted, because that's a safe default for now, where any indexing metadata could be leaking confidential information to the host. We could add the option for merely protected but unencrypted content, if we have a use and convince ourselves it's safe.
  - This can report missing or corrupted files, and any caller needs to handle that appropriately (since even a merely-unsober host may tamper with the files).
- A host-side manager that reads those messages and responds.
  - Currently stores as separate files in a local folder.
  - Again, room for growth, but a simple default for now.
- A new indexing strategy, `SeqNosByKey_Bucketed`, which uses this LFS to offload old results, and keeps only a small cache in-memory
  - Various modifications to the indexing strategies/APIs to support this. Pulled out some common functionality, and moved the old strategy to `SeqNosByKey_Memory`. Would like to keep it for testing, and build a collection of sample strategies to discuss their tradeoffs.
- Unit tests for the new stuff

Probably some other structural pieces I've forgotten about, and will try to highlight inline.


include/ccf/indexing/strategies/seqnos_by_key_in_memory.h
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
eddyashton, line 1: This is mostly `seqnos_by_key.h`, just moved here and with some of the index-construction factored out to `VisitEachEntryInMap` (so both this and `_Bucketed` implement `visit_entry`, rather than similar copies of `handle_committed_transaction`).

include/ccf/indexing/strategy.h
```diff
@@ -29,13 +29,10 @@ namespace ccf::indexing
...
-    // nullopt if it wants all Txs. Allows indexes to be populated
-    // lazily on-demand.
-    virtual std::optional<ccf::TxID> highest_requested()
-    {
-      return std::nullopt;
-    }
+    // Returns next tx for which this index should be populated, or
+    // nullopt if it wants none. Allows indexes to be populated
+    // lazily on-demand, or out-of-order, or reset
+    virtual std::optional<ccf::SeqNo> next_requested() = 0;
```
eddyashton, line 35: If the `_Bucketed` strategy asks for a file (which it believes it wrote, and encrypted), and that file is missing, or corrupted, what should it do? We could be strict, view this as a malicious host, and fail hard, but I think that makes us too susceptible to accidental file mangling. The current behaviour is to happily say "no harm, no foul", and begin re-indexing from scratch. To re-index, we need to change the API between the indexer and Strategies - rather than assuming they're done with a Tx once they've seen it once, we ask each for the SeqNo they'd like next. This lets them reset to 0 to re-index.

include/ccf/indexing/strategies/seqnos_by_key_bucketed.h
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    Range get_range_for(ccf::SeqNo seqno) const
+    {
+      const auto begin = (seqno / seqnos_per_bucket) * seqnos_per_bucket;
+      const auto end = begin + seqnos_per_bucket;
+      return {begin, end};
+    }
+
+    void visit_entry(
```
eddyashton, line 123: This is the write-side of the new `Strategy`. Each time the `Indexer` gives it an entry (a write to the target table), we:
- Find the bucket to store this result in. This is one-bucket-per-key, and a caller-configured range of seqnos in each bucket.
- If we have a current bucket for this key, but from an old range, then dump that bucket to disk
  - That involves find a unique name for the bucket, serialising the contents, and handing it off to the LFS system for encryption.
  - Then building a new empty bucket for the correct range.
- Record the given seqno in this bucket.

include/ccf/indexing/strategies/seqnos_by_key_bucketed.h
```diff
@@ -0,0 +1,373 @@
...
+
+                // NB: This could probably be more precise about what is
+                // re-indexed. Technically only need to build an index to build
+                // the current query, and that applies to re-indexing too. But
+                // for safety, and consistency with the simple indexing
+                // strategies currently used, this re-indexes everything from
+                // the start of time.
+                current_txid = {};
+                old_results.clear();
+                current_results.clear();
```
eddyashton, line 258: Related to the comment on `strategy.h`, this is where we begin to re-index if we were given a corrupt file. Note that as the comment says, we're pessimistic and eager here; we re-index everything from 0, where we could technically (probably) just re-index to answer the current query (lazy re-indexing). But avoiding such complications for now, and accepting resulting inefficiency.

src/node/rpc/test/node_stub.h
```diff
@@ -253,44 +253,10 @@ namespace ccf
     {
       throw std::logic_error("Not implemented");
     }
-  };
-
-  class StubRecoverableNodeState : public StubNodeState
```
eddyashton, line 258: These 2 classes were unused, so I've removed them.

src/indexing/enclave_lfs_access.h
```diff
@@ -0,0 +1,282 @@
...
+          }
+          else
+          {
+            LOG_TRACE_FMT(
+              "Ignoring not_found message for unrequested key {}", obfuscated);
+          }
+        });
+    }
+
+    static inline LFSKey obfuscate_key(const LFSKey& key)
```
eddyashton, line 218: I realised that as well as encryption, we maybe need some kind of key obfuscation? We should discuss exactly what kind of information leakage to the host with these index files is acceptable/inevitable (additional things we could add - a salt to this hash, and padding to the encryption).

src/indexing/enclave_lfs_access.h
```diff
@@ -0,0 +1,282 @@
...
+      return false;
+    }
+  }
+
+  class EnclaveLFSAccess : public AbstractLFSAccess
+  {
+  protected:
+    using PendingResult = std::weak_ptr<FetchResult>;
+
+    std::unordered_map<LFSKey, PendingResult> pending;
```
eddyashton, line 96: To minimise memory copies and memory use by this API, it stores results in a caller-managed `FetchResult`. So to drop a request, the caller simply drops their handle (and then our weak ptr lock fails). And there's no 2-step check-for-result + read-contents process, instead the caller has the result and contents together (modulo some locking, which I haven't implemented yet).

src/host/lfs_file_handler.h
```diff
@@ -0,0 +1,81 @@
...
+            RINGBUFFER_WRITE_MESSAGE(
+              ccf::indexing::LFSMsg::response, writer, key, blob);
+          }
+          else
+          {
+            LOG_TRACE_FMT("File {} not found", target_path);
+            RINGBUFFER_WRITE_MESSAGE(
+              ccf::indexing::LFSMsg::not_found, writer, key);
+          }
+        });
```
eddyashton, line 78: This is the host-side management of these blobs. Just reading and writing files. As I said, super simple for now, but hopefully generic enough to do something smarter/faster as and when it's required.

src/indexing/test/lfs.cpp
```diff
@@ -0,0 +1,376 @@
...
+{
+  auto corrupted(original);
+  REQUIRE(i < corrupted.size());
+  corrupted[i]++;
+  std::ofstream f(p, std::ios::trunc | std::ios::binary);
+  f.write((char const*)corrupted.data(), corrupted.size());
+  f.close();
+}
+
+TEST_CASE("Basic cache" * doctest::test_suite("lfs"))
```
eddyashton, line 36: First unit test, prodding the LFS APIs directly. Still a bunch of boiler-plate to create test ringbuffers. Then write some contents, read some contents, confirm that the integrity+encryption recognises corruption.

src/indexing/test/lfs.cpp
```diff
@@ -0,0 +1,376 @@
...
+      enclave_bp.read_all(inbound_reader);
+
+      REQUIRE(result->fetch_result == ccf::indexing::FetchResult::Corrupt);
+      REQUIRE(result->contents != blob_b);
+    }
+  }
+#endif
+}
+
+TEST_CASE("Integrated cache" * doctest::test_suite("lfs"))
```
eddyashton, line 126: Second unit test, using the new `Strategy` that uses the LFS to offload entries. Again a lot of boiler-plate, since this also has the boiler-plate of the indexing system.

eddyashton:
This PR is now ready for review.

include/ccf/indexing/strategies/seqnos_by_key_bucketed.h
```diff
@@ -0,0 +1,373 @@
...
+namespace ccf::indexing::strategies
+{
+  // Stores only a subset of results in-memory, on-demand, and dumps the
+  // remainder to disk. The size of the per-key buckets which will be retained,
+  // and the number of buckets which may be held in-memory, are configurable
+  template <typename M>
+  class SeqnosByKey_Bucketed : public VisitEachEntryInMap
+  {
+  protected:
+    const size_t seqnos_per_bucket = 100;
```
achamayou, line 18: That seems like something that would be good to expose via configuration.

src/host/lfs_file_handler.h
```diff
@@ -0,0 +1,81 @@
...
+#include "ccf/indexing/lfs_types.h"
+#include "ds/messaging.h"
+
+#include <filesystem>
+
+namespace asynchost
+{
+  struct LFSFileHandler
+  {
+    const std::filesystem::path root_dir = ".ccf_lfs";
```
achamayou, line 14: We should probably consider naming the index files .index for clarity. Other uses of LFS may use different suffixes to indicate what they contain, but naming them all after the mechanism doesn't seem ideal.

src/host/lfs_file_handler.h
achamayou: Definitely of interest for eventual KV-swapping.

include/ccf/indexing/strategies/seqnos_by_key_bucketed.h
```diff
@@ -0,0 +1,373 @@
...
+        "{}: {} -> {} for {}", get_name(), range.first, range.second, hex_key);
+    }
+
+    void store_to_disk(
+      const kv::untyped::SerialisedEntry& k,
+      const Range& range,
+      SeqNoCollection&& seqnos)
+    {
+      const BucketKey bucket_key{k, range};
+      const auto blob_key = get_blob_name(bucket_key);
```
achamayou, line 112: Are we sure that we never re-use a bucket_key?

include/ccf/indexing/strategies/seqnos_by_key_bucketed.h
eddyashton: I'll remove the default value here since it's unused. This is actually set in the constructor, so it is configurable by the app-developer. We don't currently have a mechanism to expose app-specific configuration in the per-node configuration, if that's what you mean (but would be a sensible thing to add, if we think heterogeneous/non-compile time node-configuration is necessary).

src/host/lfs_file_handler.h
eddyashton: Sure, different suffixes/dirs are sensible. I think we want them under a single dir so they can be easily cleaned (if the operator finds some problem with them, they're taking up too much space/wants to symlink them). But I'll call that `.index/` for now.

eddyashton: (And of course this dir is something that should eventually be configurable! But I don't think there's any need for that yet)

src/indexing/enclave_lfs_access.h
```diff
@@ -0,0 +1,282 @@
...
+#include "ds/hex.h"
+#include "ds/messaging.h"
+
+#include <optional>
+#include <set>
+#include <unordered_map>
+
+// Uncomment to disable encryption and obfuscation, writing cache content
+// directly unencrypted to host disk
+// #define PLAINTEXT_CACHE
```
jumaffre, line 18: Can we add another check to make sure this isn't set when built in the enclave?

src/indexing/enclave_lfs_access.h
```diff
@@ -0,0 +1,282 @@
...
+
+namespace ccf::indexing
+{
+  static inline std::vector<uint8_t> get_iv(const LFSKey& key)
+  {
+    auto h = crypto::SHA256((const uint8_t*)key.data(), key.size());
+    h.resize(crypto::GCM_SIZE_IV);
+    return h;
```
jumaffre, line 26: If the bucket `key` is re-used to encrypt multiple content (which I believe it is?), then the AES GCM IV will be re-used, compromising the encryption key. I think it'd be better to use a monotonic counter for the IV instead.

src/indexing/enclave_lfs_access.h
jumaffre: In terms of naming, I think this is more than just obfuscation as we hash the `key` here?

src/indexing/enclave_lfs_access.h
eddyashton: So the same key shouldn't be reused for different content (it'll only be reused to rewrite the same bucket, which should contain the same contents). But I keep thinking the IV needs to be independently derivable and of course it doesn't - stamping a monotonic counter makes more sense, I'll do that.

src/indexing/enclave_lfs_access.h
eddyashton: I've added this, so I'm not sure how useful it will be. If we ever want to test the enclave build with this debugging feature, we'll just comment out the lines, and have the same check on PRs that we don't check that in.

doc/architecture/indexing.rst
```diff
@@ -0,0 +1,81 @@
...
+
+CCF supports the creation of endpoints which process historical state.
+Historical transactions are asynchronously fetched from the ledger, verified and decrypted, and presented to the application endpoint for processing.
+These queries can also operate on a range of historical transactions, and build an aggregated result.
+The framework handles all details of fetching and validating these entries, to present a simple API to the application programmer.
+
+Fetching each transaction on-demand is too slow for some use cases, especially for historical queries which may be called regularly.
+To improve the performance of this kind of query, the framework provides a generic indexing system.
+Applications can define and install a strategy for processing historical queries.
+Each strategy will be given the raw contents of historical transactions shortly after they are committed, and can build its own index ahead of a user-query.
```
jumaffre, line 12:
```suggestion
Each strategy will be given the raw contents of transactions shortly after they are committed, and can build its own index ahead of a user-query.
```

src/indexing/enclave_lfs_access.h
eddyashton: I said monotonic, but actually implemented random. This introduced an issue caught by one of the unit tests, where a malicious host could substitute `file_b` for `file_a` - both are encrypted and self-consistent, we just have completely the wrong contents. To avoid _that_, I've prefixed the plaintext with the file name before encrypting, so we still have a cheap check for "is this what I asked for" on correctly decrypted contents. We could also do that by keeping part of the IV for a hash of the filename, or using per-file HKDF derived keys (I'm leaning towards the latter, just need to work out the details).

src/indexing/enclave_lfs_access.h
achamayou: Let's keep this public-only until we've reviewed the full implications on confidentiality, just digesting the key + range sounds vulnerable to dictionary attacks.

achamayou (approved):

PR #3445, opened 2022-01-24
JWT metrics

### LINKED ISSUE #1916: JWT key auto-refresh: include connection failures in metrics
Follow-up from https://github.com/microsoft/CCF/pull/1908.

[Expose connection failures](https://github.com/microsoft/CCF/issues/1486#issuecomment-731081498) to enclave to be able to record key refresh errors in those situations. This includes TLS errors like SAN mismatch.

rengolin:
Testing if this makes sense..

This creates an endpoint `jwt_metrics` that captures the number of attempts from each key in `refresh_jwt_keys` and the number of successes from the total calls to endpoint `jwt_keys/refresh` (which is also done per key, per issuer).

The way I'm getting the endpoint doesn't seem kosher, so happy to change to a more principled way.

CHANGELOG updated, with a link to a new section on the docs, however, I don't know how to link to the JSON generated file in the API docs from the JWT doc page.

Fixes #1916


rengolin:
The CI checker gives me:
```
OpenAPI
doc/schemas/app_openapi.json is valid
doc/schemas/gov_openapi.json is valid
Token "JWTMetrics" does not exist.
```

I'm not sure how to fix that.

eddyashton:
> The CI checker gives me:
> 
> ```
> OpenAPI
> doc/schemas/app_openapi.json is valid
> doc/schemas/gov_openapi.json is valid
> Token "JWTMetrics" does not exist.
> ```
> 
> I'm not sure how to fix that.
> 
> Note: for some reason, this does not fail the script...

The OpenAPI documents are auto-generated. The `schema.py` e2e test pulls those auto-generated documents, and checks that they match the ones that are checked in (or, if they're a legitimate evolution of the document, writes the updated document to disk for you to check in). The CI check just confirms those documents are valid OpenAPI. Not sure why the missing element doesn't fail. Anyway, to fix:

1) Bump the `openapi_info.document_version = ` version `node_frontend.h`.
2) Run `schema_test_cft`. This will still fail, but the error is different (previously saying "docs don't match and versions are the same", now saying "docs didn't match but version incremented, so I've overwritten").
3) Check in the modified `node_openapi.json`, which should include the version bump, the new endpoint, and any components required by that new endpoint.
4) CI checks and `schema_test_cft` should now pass.

rengolin:
> 1. Bump the `openapi_info.document_version = ` version in `node_frontend.h`.
> 2. Run `schema_test_cft`. This will still fail, but the error is different (previously saying "docs don't match and versions are the same", now saying "docs didn't match but version incremented, so I've overwritten").
> 3. Check in the modified `node_openapi.json`, which should include the version bump, the new endpoint, and any components required by that new endpoint.
> 4. CI checks and `schema_test_cft` should now pass.

Thanks! That worked. I bumped the build version (2.10.1) instead of the minor version (2.11.0), is that the right change?

eddyashton:
> Thanks! That worked. I bumped the build version (2.10.1) instead of the minor version (2.11.0), is that the right change?

No, this should bump the minor version for consistency with our other changes - adding an endpoint is a minor version. This is a bigger change than semver requires, but is the pattern we've gone with so far.

src/node/rpc/node_frontend.h
```diff
@@ -1194,6 +1203,29 @@ namespace ccf
...
         .install();
 
+      auto jwt_metrics = [this](auto&, nlohmann::json&&) {
+        JWTMetrics m;
+        // Attempts are recorded by the key refresh code itself, registering
+        // before each call to each issuer's keys
+        m.attempts = context.get_node_state().get_jwt_attempts();
+        // Success is marked by the fact that the key succeeded and called
+        // our internal "jwt_keys/refresh" endpoint.
+        auto e = fully_qualified_endpoints["/jwt_keys/refresh"][HTTP_POST];
```
rengolin, line 1213: Is there a better way of getting the endpoint, here?

eddyashton: No, this looks like the sensible way of getting that data to me.

rengolin:
> No, this should bump the minor version for consistency with our other changes - adding an endpoint is a minor version. This is a bigger change than semver requires, but is the pattern we've gone with so far.

Ok, bumped to 2.11.

rengolin:
Ready for review

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Added `set_claims_digest()` API to `RpcContext`, see [documentation](https://microsoft.github.io/CCF/main/build_apps/logging_cpp.html#user-defined-claims-in-receipts) on how to use it to attach application-defined claims to transaction receipts.
+- Added a `POST /jwt_metrics` endpoint to monitor attempts and successes of key refresh for each issuer. See [documentation](https://microsoft.github.io/CCF/main/build_apps/auth/jwt.html#extracting-jwt-metrics) on how to use it.
```
achamayou, line 29:
```suggestion
- Added a `GET /jwt_metrics` endpoint to monitor attempts and successes of key refresh for each issuer. See [documentation](https://microsoft.github.io/CCF/main/build_apps/auth/jwt.html#extracting-jwt-metrics) on how to use it.
```

surely?

doc/build_apps/auth/jwt.rst
```diff
@@ -180,3 +180,11 @@ Any attempt to add a certificate with mismatching claims in a ``set_jwt_public_s
...
 Some IdPs, like MAA, advertise a mix of SGX and non-SGX signing certificates.
 In this case, ``key_filter`` must be set to ``sgx`` such that only those certificates are stored which contain SGX evidence.
+
+Extracting JWT metrics
+----------------------
+
+CCF tracks the number of JWT queries that were attempted and succeeded. This can be used to identify errors, for example when the number of attempts doesn't match the number of successes.
+Each attempt, for each key on each issuer, is tracked as an attempt, and eventually a success, if the update completes.
+
+To query those numbers, use the RPC API on node endpoint `/jwt_metrics` as described [here](https://microsoft.github.io/CCF/main/operations/operator_rpc_api.html#get--jwt_metrics).
```
achamayou, line 190: There should be a link to get--node/metrics also, where the endpoint metrics are, stating that the success stats can be found there for endpoint "/jwt_keys/refresh", because it's not necessarily obvious to the user.

achamayou: Ok, I see, nevermind :)

src/node/rpc/node_frontend.h
```diff
@@ -1183,6 +1192,29 @@ namespace ccf
...
+        auto metric = get_metrics_for_endpoint(e);
+        m.successes = metric.calls;
```
achamayou, line 1204:
```suggestion
        m.successes = metric.calls - (metric.failures + metric.errors);
```

Just because the endpoint was called doesn't mean it succeeded.

achamayou (approved):

doc/build_apps/auth/jwt.rst
letmaik, line 190: rst, not md :)

src/node/jwt_key_auto_refresh.h
```diff
@@ -274,6 +276,9 @@ namespace ccf
       jwt_issuers->foreach([this, &ca_cert_bundles](
                              const JwtIssuer& issuer,
                              const JwtIssuerMetadata& metadata) {
+        // Increment attempts
+        attempts++;
```
letmaik, line 280: Shouldn't this happen only when auto_refresh is enabled for the issuer? Otherwise there will definitely be a mismatch between attempts and successes which may be misleading if multiple issuers are used where only some of them have auto_refresh enabled.

doc/build_apps/auth/jwt.rst
rengolin: ugh! sorry!!

doc/build_apps/auth/jwt.rst
```diff
@@ -180,3 +180,11 @@ Any attempt to add a certificate with mismatching claims in a ``set_jwt_public_s
 
 Some IdPs, like MAA, advertise a mix of SGX and non-SGX signing certificates.
 In this case, ``key_filter`` must be set to ``sgx`` such that only those certificates are stored which contain SGX evidence.
+
+Extracting JWT metrics
+----------------------
+
+CCF tracks the number of JWT queries that were attempted and succeeded. This can be used to identify errors, for example when the number of attempts doesn't match the number of successes.
```
letmaik, line 187: I think this paragraph is very confusing. It's not clear what a JWT query is. Is it a client doing a request to a service endpoint with a JWT? This should explicitly mention auto refresh.

doc/build_apps/auth/jwt.rst
```diff
@@ -180,3 +180,11 @@ Any attempt to add a certificate with mismatching claims in a ``set_jwt_public_s
...
+Each attempt, for each key on each issuer, is tracked as an attempt, and eventually a success, if the update completes.
```
letmaik, line 188: This is per-issuer only, not per-key.

src/node/jwt_key_auto_refresh.h
rengolin: Good point. I'll submit a new PR.

doc/build_apps/auth/jwt.rst
rengolin: Ah, correct! I interpreted the foreach loop the wrong way round, thanks!

doc/build_apps/auth/jwt.rst
rengolin: I tried to rewrite, will go on the next PR.

rengolin:
Response to @letmaik comments post-merge in #3451

PR #3446, opened 2022-01-25
JSON schema for `cchost` configuration file

### LINKED ISSUE #3354: JSON schema for `cchost` configuration
Follow-up from https://github.com/microsoft/CCF/issues/2612

We should add a JSON schema for the new `cchost` JSON configuration file so that operators can easily create new configuration files and we can automatically generate the docs currently available at https://microsoft.github.io/CCF/main/operations/configuration.html.

jumaffre:
Resolves #3354 

This PR adds schema validation for the new JSON configuration file for `cchost`. The main goal is to provide the operator with meaningful error messages when the configuration file is not valid.

Sample schema validation error:

```bash
$ cchost --config ../samples/config/start_config.json --check
2022-01-25T10:51:13.248512Z        100 [info ] ../src/host/main.cpp:80              | CCF version: ccf-2.0.0-dev7-110-gb015f7c1b
2022-01-25T10:51:13.252585Z        100 [fail ] ../src/host/main.cpp:105             | Error validating JSON schema for configuration file ../samples/config/start_config.json: 
Error #0:
  context: <root>[command]
  desc: Missing required property 'start'.
Error #1:
  context: <root>[command]
  desc: Failed to validate against a conditional schema set by if-then-else constraints.
Error #2:
  context: <root>[command]
  desc: Failed to validate against child schema #0.
Error #3:
  context: <root>
  desc: Failed to validate against schema associated with property name 'command'


```

Notes:
- This PR re-introduces `valijson` as an internal dependency (host only) to validate the schema. 
- The reference schema (`doc/schemas/cchost_config.json`) is baked in the `cchost` binary. 
- The documentation is now automatically generated from `cchost_config.json` (i.e. one source of truth). I looked at https://sphinx-jsonschema.readthedocs.io/en/latest/ first but the output table was way too wide and definitely not user friendly so I wrote a simple `generate_config_rst.py` script to generate the documentation page (output is very similar to existing https://microsoft.github.io/CCF/main/operations/configuration.html). 

TODO:
- [x] Min/max values


CMakeLists.txt
```diff
@@ -51,6 +51,14 @@ configure_file(
 )
 install(FILES ${CCF_DIR}/include/ccf/version.h DESTINATION include/ccf)
 
+# TODO: Not very elegant...
```
jumaffre, line 54: This is to bake `cchost_config.json` as a single string variable in the `cchost` binary. I can't think of anything more minimal than this, but it's not very elegant. Suggestions welcome!

eddyashton: Can we use a raw string literal, rather than manual escaping? See eg `j_record_public_in` in `logging_schema.h` for an example. So `config_schema.h.in` would contain something like `host_config_schema = R"!!!(@HOST_CONFIG_SCHEMA@)!!!";`

src/host/json_schema.h
```diff
@@ -0,0 +1,49 @@
...
+        std::string error_ctx;
+        for (auto const& c : error.context)
+        {
+          error_ctx += c;
+        }
+        validation_error_msg += fmt::format(
+          "\nError #{}:\n  context: {}\n  desc: {}",
+          error_num,
+          error.description,
+          error_ctx);
```
eddyashton, line 43:
```suggestion
        validation_error_msg += fmt::format(
          "\nError #{}:\n  context: {}\n  desc: {}",
          error_num,
          error_ctx,
          error.description);
```

src/host/json_schema.h
```diff
@@ -0,0 +1,49 @@
...
+        ++error_num;
+      }
+      throw std::logic_error(validation_error_msg);
```
eddyashton, line 46: It seems weird to convert this structured thing to an exception, to go 1 step up the stack. Can we just return an optional string error message?

doc/schemas/cchost_config.json
```diff
@@ -0,0 +1,422 @@
...
+          "max_msg_size": {
+            "type": "string",
+            "default": "16MB",
+            "description": "Maximum size (size string) for a message sent over the ringbuffer, as a power of 2. Messages may be split into multiple fragments, but this limits the total size of the sum of those fragments"
+          },
+          "max_fragment_size": {
+            "type": "string",
+            "default": "64KB",
+            "description": "Maximum size (size string) of individual ringbuffer message fragments, as a power of 2. Messages larger than this will be split into multiple fragments"
+          }
```
eddyashton, line 414: It's odd that these descriptions say "as a power of 2". I think that's an artifact from when we were passing 8 and turning that into 2^8. Perhaps should say "must be a power of 2" (if that's still a requirement)?

jumaffre: Good point. I've re-phrased this and removed the note where the requirement is not applicable (i.e. max msg/fragment size).

doc/host_config_schema/cchost_config.json
```diff
@@ -0,0 +1,412 @@
...
+              "type": "string",
+              "description": "Local address the node binds to and listens on"
+            }
+          },
+          "description": "Address (host:port) to listen on for incoming node-to-node connections (e.g. internal consensus messages)",
+          "required": ["bind_address"]
+        },
+        "rpc_interfaces": {
+          "type": "object",
+          "additionalProperties": {
```
letmaik, line 40: I haven't seen this style before. Why not move the inner bits out?

doc/host_config_schema/cchost_config.json
```diff
@@ -0,0 +1,412 @@
...
+      "description": "This section includes configuration for the interfaces a node listens on (for both client and node-to-node communications)",
+      "required": ["node_to_node_interface", "rpc_interfaces"]
+    },
+    "command": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "enum": ["start", "join", "recover"],
+          "default": "start",
```
letmaik, line 77: "type" is listed in "required", so a default wouldn't make sense.

doc/host_config_schema/cchost_config.json
```diff
@@ -0,0 +1,412 @@
...
+                      "required": ["certificate_file"]
+                    },
+                    "description": "List of initial consortium members files, including identity certificates, public encryption keys and member data files"
+                  },
+                  "service_configuration": {
+                    "type": "object",
+                    "properties": {
+                      "recovery_threshold": {
+                        "type": "integer",
+                        "default": 0,
```
letmaik, line 135: Same here, default doesn't make sense for required fields.

doc/host_config_schema/cchost_config.json
```diff
@@ -0,0 +1,412 @@
...
+          "type": "string",
+          "default": "CN=CCF Node",
+          "description": "Subject name to include in node certificate"
+        },
+        "subject_alt_names": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of ``iPAddress:`` or ``dNSName:`` strings to include as Subject Alternative Names (SAN) in node certificates. If none is set, the node certificate will automatically include the value of the main RPC interface 'published_rpc_address'"
```
letmaik, line 219:
```suggestion
          "description": "List of ``iPAddress:`` or ``dNSName:`` strings to include as Subject Alternative Names (SAN) in node certificates. If none are set, the node certificate will automatically include the value of the main RPC interface 'published_rpc_address'"
```

doc/host_config_schema/cchost_config.json
```diff
@@ -0,0 +1,412 @@
...
+      "description": "Address to bind to for node-to-node client connections. If unspecified, this is automatically assigned by the OS. This option is particularly useful for testing purposes (e.g. establishing network partitions between nodes)"
+    },
+    "client_connection_timeout": {
+      "type": "string",
+      "default": "2000ms",
+      "description": "Maximum duration after which unestablished client connections will be marked as timed out and either re-established or discarded"
+    },
+    "worker_threads": {
+      "type": "integer",
+      "default": "0",
```
letmaik, line 386:
```suggestion
      "default": 0,
```

src/host/json_schema.h
```diff
@@ -0,0 +1,53 @@
...
+      size_t error_num = 0;
+      while (results.popError(error))
+      {
+        std::string error_ctx;
+        for (auto const& c : error.context)
+        {
+          error_ctx += c;
+        }
+        validation_error_msg += fmt::format(
+          "\nError #{}:\n  context: {}\n  desc: {}",
```
letmaik, line 42:
```suggestion
          "\nError #{}:\n  context: {}\n  message: {}",
```
context and message have the same number of characters, makes it a bit nicer visually.

doc/generate_config_rst.py
```diff
@@ -0,0 +1,145 @@
...
+    def add_line(self, text):
+        self._add_lines([text])
+
+    def render(self):
+        return "\n".join(self._lines)
+
+
+def print_attributes(entry):
+    def stringify_output(entry, s):
+        s = f'"{s}"' if entry["type"] == "string" else s
```
achamayou, line 37: Not repr()?

doc/host_config_schema/cchost_config.json
jumaffre: I followed this approach: https://stackoverflow.com/a/27375654 This is because there can be any number of RPC interfaces (none might be OK too?).

letmaik: Gotcha, misunderstanding.

doc/conf.py
```diff
@@ -394,3 +400,12 @@ def setup(app):
     breathe_projects["CCF"] = str(doc_dir / breathe_projects["CCF"])
     if not os.environ.get("SKIP_DOXYGEN"):
         subprocess.run(["doxygen"], cwd=root_dir, check=True)
+
+    # configuration generator
+    input_file_path = doc_dir / "host_config_schema/cchost_config.json"
+    output_file_path = doc_dir / "operations/generated_config.rst"
+
+    if os.path.exists(input_file_path):
+        generate_config_rst.generate_configuration_docs(
```
eddyashton, line 409: Is there a reason we can't implement this as roles/directives (like the typedoc roles above), so that we could reference/link to specific subsections of the config, or paste sections of it rather than the whole doc? Non-critical - obviously generating a single doc works for now. Just wondering if that's something we could do in future.

eddyashton (approved):

doc/conf.py
achamayou: Agreed it would be nice, but non-essential for a first pass.

doc/generate_config_rst.py
```diff
@@ -0,0 +1,146 @@
...
+def print_attributes(entry):
+    def stringify_output(s):
+        return f"``{json.dumps(s)}``"
+
+    desc = ""
+    if "description" in entry:
+        desc += entry["description"]
+    if "enum" in entry:
+        desc += (
+            f' (values: {", ".join(map(lambda s: stringify_output(s), entry["enum"]))})'
```
achamayou, line 44:
```suggestion
            f' (values: {", ".join(stringify_output(s) for s in entry["enum"])})'
```

is clearer in my opinion.

doc/conf.py
jumaffre: Yes, I think this would be nice too. I was aiming for parity with what the manually-written configuration doc currently looked like, but we can definitely improve the rendering script in the future.

PR #3449, opened 2022-01-25
Fix TLS certificate verification

### LINKED ISSUE #3432: Remove tls::Auth in ca.h
MbedTLS had a optional auth while OpenSSL it's either on or off. Furthermore, there's no difference between `auth_default` and `auth_required`, so the enum `Auth` is now outdated and can be replaced with a boolean flag.

wintersteiger:
Fixes one of the two CodeQL security reports reported recently (both are spurious). Also fixes #3432.

### DIFF
--- a/src/enclave/rpc_sessions.h
+++ b/src/enclave/rpc_sessions.h
@@ -191,9 +191,10 @@ namespace enclave
 
       // Caller authentication is done by each frontend by looking up
       // the caller's certificate in the relevant store table. The caller
-      // certificate does not have to be signed by a known CA (nullptr).
-      cert =
-        std::make_shared<tls::Cert>(nullptr, cert_, pk, std::nullopt, false);
+      // certificate does not have to be signed by a known CA (nullptr) and
+      // verification is not required here.
+      cert = std::make_shared<tls::Cert>(
+        nullptr, cert_, pk, std::nullopt, /*auth_required ==*/false);
     }
 
     void accept(tls::ConnID id, const ListenInterfaceID& listen_interface_id)
--- a/src/enclave/rpc_sessions.h
+++ b/src/enclave/rpc_sessions.h
@@ -192,7 +192,8 @@ namespace enclave
       // Caller authentication is done by each frontend by looking up
       // the caller's certificate in the relevant store table. The caller
       // certificate does not have to be signed by a known CA (nullptr).
-      cert = std::make_shared<tls::Cert>(nullptr, cert_, pk, false);
+      cert =
+        std::make_shared<tls::Cert>(nullptr, cert_, pk, std::nullopt, false);
     }
 
     void accept(tls::ConnID id, const ListenInterfaceID& listen_interface_id)
--- a/src/enclave/rpc_sessions.h
+++ b/src/enclave/rpc_sessions.h
@@ -192,7 +192,7 @@ namespace enclave
       // Caller authentication is done by each frontend by looking up
       // the caller's certificate in the relevant store table. The caller
       // certificate does not have to be signed by a known CA (nullptr).
-      cert = std::make_shared<tls::Cert>(nullptr, cert_, pk, tls::auth_default);
+      cert = std::make_shared<tls::Cert>(nullptr, cert_, pk, false);
     }
 
     void accept(tls::ConnID id, const ListenInterfaceID& listen_interface_id)
--- a/src/node/jwt_key_auto_refresh.h
+++ b/src/node/jwt_key_auto_refresh.h
@@ -245,7 +245,7 @@ namespace ccf
       auto jwks_url_port = !jwks_url.port.empty() ? jwks_url.port : "443";
 
       auto ca_cert = std::make_shared<tls::Cert>(
-        ca, std::nullopt, std::nullopt, true, jwks_url.host);
+        ca, std::nullopt, std::nullopt, /*auth_required=*/true, jwks_url.host);
 
       LOG_DEBUG_FMT(
         "JWT key auto-refresh: Requesting JWKS at https://{}:{}{}",
--- a/src/node/jwt_key_auto_refresh.h
+++ b/src/node/jwt_key_auto_refresh.h
@@ -245,7 +245,7 @@ namespace ccf
       auto jwks_url_port = !jwks_url.port.empty() ? jwks_url.port : "443";
 
       auto ca_cert = std::make_shared<tls::Cert>(
-        ca, std::nullopt, std::nullopt, tls::auth_required, jwks_url.host);
+        ca, std::nullopt, std::nullopt, true, jwks_url.host);
 
       LOG_DEBUG_FMT(
         "JWT key auto-refresh: Requesting JWKS at https://{}:{}{}",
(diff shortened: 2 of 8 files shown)


src/tls/cert.h
```diff
@@ -70,9 +70,10 @@ namespace tls
         peer_ca->use(ssl_ctx);
       }
 
-      if (auth != auth_default)
+      if (auth_required)
       {
-        SSL_CTX_set_verify(ssl_ctx, authmode(auth), NULL);
+        SSL_CTX_set_verify(
```
wintersteiger, line 75: This should be applied to `ssl` as well as it inherited the verification options from `ssl_ctx` when it was created, but hasn't been updated since. This leads to certification validation errors because of host name mismatches. I believe this was also a hidden problem with mbedTLS on the server side, where `mbedtls_ssl_conf_authmode` was not called when `Auth::auth_default` was enabled.

src/enclave/rpc_sessions.h
```diff
@@ -192,7 +192,7 @@ namespace enclave
       // Caller authentication is done by each frontend by looking up
       // the caller's certificate in the relevant store table. The caller
       // certificate does not have to be signed by a known CA (nullptr).
-      cert = std::make_shared<tls::Cert>(nullptr, cert_, pk, tls::auth_default);
+      cert = std::make_shared<tls::Cert>(nullptr, cert_, pk, false);
```
rengolin, line 195: Nit: I'd make `false` the default and not need the extra arg here.

src/node/jwt_key_auto_refresh.h
```diff
@@ -245,7 +245,7 @@ namespace ccf
       auto jwks_url_port = !jwks_url.port.empty() ? jwks_url.port : "443";
 
       auto ca_cert = std::make_shared<tls::Cert>(
-        ca, std::nullopt, std::nullopt, tls::auth_required, jwks_url.host);
+        ca, std::nullopt, std::nullopt, true, jwks_url.host);
```
rengolin, line 248:
```suggestion
        ca, std::nullopt, std::nullopt, /*auth_required=*/true, jwks_url.host);
```

src/node/jwt_key_auto_refresh.h
```diff
@@ -310,11 +310,7 @@ namespace ccf
 
         auto ca = std::make_shared<tls::CA>(ca_cert_bundle_pem.value());
         auto ca_cert = std::make_shared<tls::Cert>(
-          ca,
-          std::nullopt,
-          std::nullopt,
-          tls::auth_required,
-          metadata_url.host);
+          ca, std::nullopt, std::nullopt, true, metadata_url.host);
```
rengolin, line 313:
```suggestion
          ca, std::nullopt, std::nullopt, /*auth_required=*/true, metadata_url.host);
```

src/node/node_state.h
```diff
@@ -416,7 +416,7 @@ namespace ccf
         network_ca,
         self_signed_node_cert,
         node_sign_kp->private_key_pem(),
-        tls::Auth::auth_required,
+        true,
```
rengolin, line 419:
```suggestion
        /*auth_required=*/true,
```

src/tls/cert.h
rengolin: There was an issue with adding that to the SSL object, I can't remember what it was. It could have been just that I was using the wrong flags, or a conflict in using the old `Auth` structure, but it'd be good to know what was wrong and how to do it right. It can be for a different path, if it's too convoluted.

rengolin (commented):
Good to see that horribly large `get_verify_error` gone. :)

src/enclave/rpc_sessions.h
rengolin: Or perhaps better, make `true` the default and only override when you don't really care...

wintersteiger: Yeah, that was my plan too, but since the previous default wasn't `auth_required`, I'm getting a few verification errors that I need to sort out first.

rengolin: Right, this may be related to the errors I was seeing back then. would be good to understand what's the underlying cause.

src/tls/cert.h
wintersteiger: The `ssl` object is created just after `ssl_ctx`, with the default options and it won't be updated with the new settings via `SSL_CTX_set_verify`, so either we need to create it later, or duplicate the setting here (actually, just the `ssl` object would fine too I think). There's a problem with the peer host name not matching the name/SAN in the certificate, but I think that's just another matter of setting the right options.

rengolin: This is another thing we need to cleanup, to only create the SSL object at the last moment, so we only configure the context, not the session. But it was done that way to emulate mbedTLS.

wintersteiger: Yeah, I think so too. Alternatively, don't configure `ssl_ctx`, but consistently always `ssl`.

src/enclave/rpc_sessions.h
wintersteiger: Done, `auth_required` is now the default.

src/enclave/rpc_sessions.h
```diff
@@ -192,7 +192,8 @@ namespace enclave
       // Caller authentication is done by each frontend by looking up
       // the caller's certificate in the relevant store table. The caller
       // certificate does not have to be signed by a known CA (nullptr).
-      cert = std::make_shared<tls::Cert>(nullptr, cert_, pk, tls::auth_default);
+      cert =
+        std::make_shared<tls::Cert>(nullptr, cert_, pk, std::nullopt, false);
```
wintersteiger, line 196: @jumaffre: I think you were the last one to touch this piece of code. This used to require `auth_optional`, which does not exist anymore. Do you have any concerns about the certificate validation being disabled/ignored here? (I think it's fine, just a sanity check.)

src/tls/cert.h
rengolin: In our use case, it doesn't matter because we always create both on a new connection. The easy solution is to ignore CTX and configure only SSL. This would still keep the format as we have today.

But we can also have a more persistent CTX and only create a new SSL for each connection. In that case, we'd need to split `Context` from `Connection` and only have `send` and `recv` on the latter.

I don't know what's the overhead of setting up, but I'd guess creating a new SSL from an existing (and already configured) CTX would be almost a copy construction.

The downside is that, if we have different types of connections (with and without verification, different certificates), then the persistent CTX needs to be a list of them, not just one, and we'd need a way to choose between them.

rengolin (approved):

src/tls/cert.h
wintersteiger: Moved to  #3461 so I can resolve the conversation here.

src/tls/cert.h
```diff
@@ -62,17 +62,23 @@ namespace tls
         // extension) if it is set. This lets us connect to peers that present
         // certificates with IPAddress in SAN field. This is OK since we check
         // for peer CA endorsement.
-        SSL_set1_host(ssl, peer_hostname->c_str());
+        SSL_set_tlsext_host_name(ssl, peer_hostname->c_str());
```
jumaffre, line 65: I think "This lets us connect to peers that present certificates with IPAddress in SAN field." only applied with `mbedtls` so this can go.

src/enclave/rpc_sessions.h
jumaffre: LGTM. Perhaps mention why `auth_required` is `false` too in the comment above?

src/tls/cert.h
wintersteiger: Good point, I'll remove that.

rengolin:
@jumaffre @eddyashton @achamayou Would be nice to get this merged as I want to use this as a base for some cleanups due to future QUIC work (ex. cleaning up DTLS and some other unused stuff that we kept in hope QUIC would use).

eddyashton (approved):

Issue #2875, opened 2021-08-05
Allow for different server TLS endorsement on different RPC interface


jumaffre:
Follow-up from #2521 and #2554 

We should also allow for different TLS endorsement on different RPC interfaces, typically self-signed node cert for operators (`/node/` route) and endorsed cert for users/members (`/app/` and `/gov/` routes). This simplifies the process for operators who would no longer need to switch from self-signed CA to service-endorsed CA when a new joining node has caught up.

achamayou:
To start with, endorsed/un-endorsed may be enough of a distinction, since the un-endorsed/node-only interface is unlikely to be used for anything other than debugging a starting node.

PR #3450, opened 2022-01-25
Network: per-interface endorsement type

letmaik:
Closes https://github.com/microsoft/CCF/issues/2875.


src/enclave/rpc_sessions.h
```diff
@@ -217,7 +243,18 @@ namespace enclave
 
       auto& per_listen_interface = it->second;
 
-      if (
+      if (!certs.contains(listen_interface_id))
```
letmaik, line 246: This is a change in behaviour for the "network" endorsement type, which would be the default. Previously, the selfsigned cert would be used and later replaced, but now the interface only accepts sessions when the right endorsement becomes available. I think this makes sense, but worth pointing out.

src/node/node_info_network.h
```diff
@@ -9,6 +9,15 @@
 
 namespace ccf
 {
+  enum class EndorsementType
+  {
+    Node,
```
letmaik, line 14: I initially named those "Self" and "Service" but then renamed them to match the existing terminology.

src/node/node_info_network.h
```diff
@@ -37,12 +46,15 @@ namespace ccf
       std::optional<size_t> max_open_sessions_soft = std::nullopt;
       std::optional<size_t> max_open_sessions_hard = std::nullopt;
 
+      EndorsementType endorsement_type = EndorsementType::Network;
```
jumaffre, line 49: As this is a new field in a KV table (that automatically ends up in the ledger), could you declare the field as `std::optional` so that it is clear when accessing it that it may not be present (i.e. for historical ledgers). We've followed this approach so far, and should document it under https://microsoft.github.io/CCF/main/contribute/index.html.

src/enclave/rpc_sessions.h
```diff
@@ -155,6 +157,8 @@ namespace enclave
         li.max_open_sessions_hard = interface.max_open_sessions_hard.value_or(
           max_open_sessions_hard_default);
 
+        li.endorsement_type = interface.endorsement_type;
+
         LOG_INFO_FMT(
           "Setting max open sessions on interface \"{}\" ({}) to [{}, "
           "{}]",
```
jumaffre, line 164: Perhaps log endorsement type too?

src/enclave/rpc_sessions.h
```diff
@@ -217,7 +243,18 @@ namespace enclave
 
       auto& per_listen_interface = it->second;
 
-      if (
+      if (!certs.contains(listen_interface_id))
+      {
+        LOG_INFO_FMT(
```
jumaffre, line 248: `LOG_DEBUG_FMT`?

jumaffre (commented):
LGTM! This will need some sort of end-to-end tests ([this option](https://github.com/microsoft/CCF/blob/main/tests/infra/node.py#L504) will be useful) and a changelog entry.

src/node/node_info_network.h
```diff
@@ -9,6 +9,15 @@
...
 namespace ccf
 {
+  enum class EndorsementType
+  {
+    NODE,
+    NETWORK
+  };
+  DECLARE_JSON_ENUM(
+    EndorsementType,
+    {{EndorsementType::NODE, "Node"}, {EndorsementType::NETWORK, "Network"}});
```
jumaffre, line 19: Looking at this again, I think `Service` would be better than `Network` as we've tried to move to `Service` for other things (e.g. service certificate).

letmaik: I'll wait for others to upvote or chime in and then make a wholesale replacement. In any case, do you mean changing only the JSON strings, or also the C++ type names?

doc/host_config_schema/cchost_config.json
```diff
@@ -58,6 +58,12 @@
                 "type": "integer",
                 "default": 1010,
                 "description": "The maximum number of active client sessions on that interface after which clients sessions will be terminated, before the TLS handshake is complete. Note that its value must be greater than the value of ``max_open_sessions_soft``"
+              },
+              "endorsement_type": {
```
letmaik, line 62: An alternative would be `"endorsement"` with a single member `"type"`. I didn't do that for simplicitly, but that may be short-sighted.

src/node/node_info_network.h
jumaffre: Only the JSON string so that it's consistent for users/operators.

doc/host_config_schema/cchost_config.json
achamayou: Yes!

letmaik: Offline consensus: `"endorsement"` as object with single field (for now) `"authority"` being an enum of values `"Node"`, `"Service"`.

src/node/node_info_network.h
letmaik: After more discussion, I'll rename it to `Service` including the C++ types.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added `set_claims_digest()` API to `RpcContext`, see [documentation](https://microsoft.github.io/CCF/main/build_apps/logging_cpp.html#user-defined-claims-in-receipts) on how to use it to attach application-defined claims to transaction receipts.
 - Added a `GET /jwt_metrics` endpoint to monitor attempts and successes of key refresh for each issuer. See [documentation](https://microsoft.github.io/CCF/main/build_apps/auth/jwt.html#extracting-jwt-metrics) on how to use it.
 
+### Added
```
achamayou, line 15: That needs to be moved out now that dev8 is cut.

samples/config/start_config.json
```diff
@@ -16,7 +16,10 @@
         "bind_address": "127.0.0.1:8082",
         "published_address": "ccf.dummy.com:12346",
         "max_open_sessions_soft": 1000,
-        "max_open_sessions_hard": 1010
+        "max_open_sessions_hard": 1010,
+        "endorsement": {
+          "authority": "Node"
```
achamayou, line 21: @jumaffre do we also capitalise enum values elsewhere in the config? If not, I'd rather have the config look one way than arbitrarily enforce consistency with the C++ type on a one-off basis.

achamayou: Lower case: enclave.type, command.type, node_certificate.curve_id, logging.host_level, logging.format
Camel case: reconfiguration_type
All caps, but going away: consensus_type

letmaik: To avoid a mix of casing between enums exposed in OpenAPI and the config, esp. in cases where they are used in both, we decided to align with OpenAPI and rename all config enum values to be camel case. Therefore, this PR won't change since it uses that convention already.
@jumaffre

src/node/node_info_network.h
```diff
@@ -137,4 +161,33 @@ namespace ccf
...
+    switch (authority)
+    {
+      case (ccf::Authority::NODE):
+      {
+        return format_to(ctx.out(), "Node");
+      }
+      case (ccf::Authority::SERVICE):
+      {
+        return format_to(ctx.out(), "Service");
+      }
```
achamayou, line 189: A default throw is probably prudent here.

tests/reconfiguration.py
```diff
@@ -69,11 +69,28 @@ def wait_for_reconfiguration_to_complete(network, timeout=10):
 
 @reqs.description("Adding a valid node without snapshot")
 def test_add_node(network, args):
-    new_node = network.create_node("local://localhost")
+    # Note: 127.0.0.1 is supplied explicitly to avoid having differently
+    # assigned IPs for the interfaces, something which the test infra doesn't
+    # support widely yet.
+    operator_rpc_interface = "operator_rpc_interface"
+    host = "127.0.0.1"
```
achamayou, line 76: We really want to avoid hardcoding interfaces that way, because it makes running concurrent tests flakier than they need to be. Can you use `net.expand_localhost()` instead?

src/node/node_info_network.h
letmaik: It's not needed. With new enum members you would a compile error: `node_info_network.h:180:13: error: enumeration value 'FOO' not handled in switch [-Werror,-Wswitch]`

tests/reconfiguration.py
letmaik: Done.

achamayou (approved):

PR #3452, opened 2022-01-25
Fix Python docs for historical versions

jumaffre:
I've noticed that the documentation on historical versions of our docs was broken:
image

This is because:
1. We re-generate the docs for all (including historical `-dev` and LTS releases) every time the GH pages CI job runs on main (*)
2. The Python client API was recently removed on `main` (see #3372), along with its dependencies

We now import the historical version of the `ccf` package when building the versioned docs (with `sphinx-multiversion`) and install the corresponding historical dependencies. (We could try to install the historical dependencies on the fly, from the historical checkout, but this requires running `cmake` to generate `version.py`, but the auto-checkout is in a temporary directory but `cmake` requires a specific directory name to deduce the version, etc. - I think it's simpler to keep a file with the historical dependencies around).

(*) This can probably be optimised/simplified after 2.0 so that the documentation is only built once per release. @achamayou @eddyashton

### DIFF
--- a/.azure-pipelines-gh-pages.yml
+++ b/.azure-pipelines-gh-pages.yml
@@ -23,7 +23,7 @@ jobs:
           git checkout -b main $(git rev-parse HEAD)
         displayName: Prepare repo
 
-      # Used to generate setup.py
+      # Used to generate version.py
       - template: .azure-pipelines-templates/cmake.yml
         parameters:
           cmake_args: ""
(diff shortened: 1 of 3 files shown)


.azure-pipelines-gh-pages.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           git checkout -b main $(git rev-parse HEAD)
         displayName: Prepare repo
 
-      # Used to generate setup.py
+      # Used to generate version.py
```
jumaffre, line 26: See https://github.com/microsoft/CCF/pull/3452

achamayou (approved):

PR #3453, opened 2022-01-25
Added ability to disable OE usage

ihaller:
This is the starting point of a CCF variant which does not use any OE bits.
It only allows virtual mode and for now there is not alternative verifier in place for attestation reports.
This will enable exploration around porting CCF to other platforms without needing to first port OE.

### DIFF
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -1,13 +1,24 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the Apache 2.0 License.
 
-set(ALLOWED_TARGETS "sgx;virtual")
+if (NOT DISABLE_OE)
+  set(ALLOWED_TARGETS "sgx;virtual")
+else()
+  set(ALLOWED_TARGETS "virtual")
+endif()
 
 set(COMPILE_TARGETS
-    "sgx;virtual"
+    ${ALLOWED_TARGETS}
     CACHE
       STRING
       "List of target compilation platforms. Choose from: ${ALLOWED_TARGETS}"
+)
+
+set(DISABLE_OE
+    False
+    CACHE
+      BOOL
+      "Disable the requirements for OE, but also to ability to verify OE attesation reports."
 )
 
 set(IS_VALID_TARGET "FALSE")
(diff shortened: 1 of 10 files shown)


cmake/ccf_app.cmake
```diff
@@ -1,15 +1,26 @@
...
     CACHE
       STRING
       "List of target compilation platforms. Choose from: ${ALLOWED_TARGETS}"
 )
 
+set(DISABLE_OE
+    False
+    CACHE
+      BOOL
+      "Disable the requirements for OE, but also to ability to verify OE attesation reports."
```
achamayou, line 21:
```suggestion
      "Disable the requirement for OE, but also to ability to verify OE attestation reports."
```

cmake/common.cmake
```diff
@@ -358,24 +378,28 @@ if("sgx" IN_LIST COMPILE_TARGETS)
...
-  set_property(
-    TARGET js_openenclave.virtual PROPERTY POSITION_INDEPENDENT_CODE ON
-  )
-  install(
-    TARGETS js_openenclave.virtual
-    EXPORT ccf
-    DESTINATION lib
-  )
+if("virtual" IN_LIST COMPILE_TARGETS AND NOT DISABLE_OE)
+  if (NOT DISABLE_OE)
```
achamayou, line 384: redundant with check above?

src/apps/js_generic/js_generic.cpp
```diff
@@ -14,7 +14,11 @@ namespace ccfapp
 
   std::vector<ccf::js::FFIPlugin> get_js_plugins()
   {
+#if !defined(DISABLE_OE)
     return {ccf::js::openenclave_plugin};
```
achamayou, line 18: The rationale for the existence of the js oe plugin was the need to verify attestation in JS apps, which was once a requirement for ACS. It would be worth checking if that's still the case.

achamayou (approved):

cmake/ccf_app.cmake
ihaller: Thanks!

cmake/common.cmake
ihaller: Yes, I was oscillating between the two and ended up forgetting to remove one. Good catch!

src/apps/js_generic/js_generic.cpp
ihaller: Thanks, I will try to understand the situation with the plugins.

achamayou: Coincidentally, that requirement has re-emerged for another internal project, so it definitely must stay.

achamayou:
@ihaller could you please run `./scripts/ci-checks.sh -f` to reformat your changes?

achamayou:
The PR looks good to me otherwise, please feel free to click "Ready for review" whenever convenient.

raghavthind2005 (approved):

achamayou:
@ihaller I am going to close this for now, and we will look at this again when we are ready to.

Issue #3092, opened 2021-10-15
js: expose more details in `Request` object (url, path, etc.)


letmaik:
(from https://github.com/microsoft/CCF/pull/3044#discussion_r729008654)

Following the [express library](https://expressjs.com/en/4x/api.html#req):

- `req.url` - URL present in the actual HTTP request, e.g. `/app/sample?a=b`
- `req.path` - just the path part of the URL, e.g. `/app/sample`
- `req.hostname` - hostname derived from the Host HTTP header, e.g. `foo.com`
- `req.method` - HTTP method, e.g. `POST`

caleteeter:
I need the same, specifically the complete path for the request for a security module/validation that we have been working on.  I will take a look at adding this.

eddyashton:
In addition to the suggestions above, we should probably expose both the path provided by the caller, and the (potentially templated) path we used for dispatch.

So if the app installs `POST /app/foo/{id}`, and that's processing a call to `POST foo.com/app/foo/bob?count=42` then we should provide (at least):
```
req.url = "/app/foo/bob?count=42"
req.path = "/app/foo/bob"
req.dispatch_path = "/app/foo/{id}" # Not attached to the name, but want to expose this value
req.hostname = "foo.com"
req.method = "POST"
```
The prefixing of `/app` is a little messy at the moment - it's not specified at install time, and is stripped from some of the values in `RpcContext`. I think we should standardise towards including it everywhere, even if that introduces additional inconsistencies at the moment.

rengolin:
@caleteeter, I'm looking at #1916 now, so if you start working on this one before I have time to look at it, let me know (or assign it to yourself).

caleteeter:
@rengolin great, I have already started working on adding the full path, which I have done in CCF, but working on the changes for ccf-app (for JS).  Need to add tests here as well.

caleteeter:
I have tested the changes adding the request path to the request object in tag v1.0.17 with success.

PR #3454, opened 2022-01-25
Updated js_generic and the ts package to include the request path in the request object.

caleteeter:
The current js_generic implementation does not expose the current request path in the request object that is presented via the TS package.  The issue is mentioned in #3092.  This PR has the changes for both the core js_generic code and the TS package.


achamayou:
@caleteeter thank you for making this improvement! Could you run `./scripts/ci-checks.sh -f`, that will make formatting adjustments so the code passes the format and linting checks in the CI.

achamayou (approved):

Issue #3405, opened 2022-01-14
Test compatibility between `N.x` and `N-1.x` 


jumaffre:
Once we release 2.x, the existing `lts_compatibility` test should also check that any release on the `2.x` branch is compatible with the latest version of the `1.x` LTS (we currently only check compatibility between `N.x` and `N.x-1`).

Issue #2699, opened 2021-06-18
LTS compatibility test does not cope with release tags


jumaffre:
When creating a new release from a release branch, the Azure DevOps environment does not know about the local branch (rightly so) but only the tag name (e.g. `ccf-1.0.4`). As we use the local branch name for many things (e.g. detect the major version number or detect whether we are on a release branch), this is effectively not right for release jobs.

The current workaround for the `1.x` branch is simply to hardcode the version number (see https://github.com/microsoft/CCF/pull/2698).

achamayou:
That needs fixing for 2.x one way or another.

PR #3457, opened 2022-01-26
Fix `lts_compatibility` test in preparation for 2.0 release

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/2699
Resolves https://github.com/microsoft/CCF/issues/3405

The LTS compatibility test now:
- Handles release tags (by associating the release tag to the appropriate release branch, e.g. `ccf-2.0.1` -> `release/2.x`)
- Tests the compatibility between `N+1.x` and `N.y` when `N+1.x` is released, e.g. we now test the compatibility between `2.0.1` and `1.0.17` when `2.0.1` is released (whereas we only used to check the compatibility between `2.0.1` and `2.0.0`). This adds a new line to the compatibility report which now looks like this (example for `main` branch):
```json
 {
  "version": "ccf-2.0.0-dev7-79-g21b1c0516",
  "live compatibility": {
    "with previous LTS": "1.0.17",    << This is where the compatibility between 2.0.1 and 1.0.17 would be displayed
    "with same LTS": null,  << This is where the compatibility between 2.0.1 and 2.0.0 would be displayed
    "with following LTS": null
  }
}
```

Since the LTS version logic has become easy to break, I've also added at test at the bottom of `github.py` (also run in the CI).


tests/lts_compatibility.py
```diff
@@ -538,28 +547,41 @@ def add(parser):
...
         {"with latest": latest_lts_version}
     )
+
+    # Compatibility with latest LTS on the same release branch
+    # (e.g. when releasing 2.0.1, check compatibility with existing 2.0.0)
+    latest_lts_version = run_live_compatibility_with_latest(
+        args, repo, env.branch, this_release_branch_only=True
+    )
+    compatibility_report["live compatibility"].update(
+        {"with latest (same LTS)": latest_lts_version}
```
eddyashton, line 566: I think of the LTS as 1.x or 2.x, rather than specific builds. So, if renaming is free, I'd be tempted to call these builds `with previous LTS` and `with current LTS`, rather than `with latest`.

jumaffre: Good point. I went for `with previous LTS`, `with same LTS` (which I think is less ambiguous than "current", which sounds like the most recent LTS) and `with following LTS`.

wintersteiger (approved):
LGTM!

tests/infra/github.py
```diff
@@ -291,3 +323,105 @@ def install_next_lts_for_branch(self, branch):
...
+            ["release/1.x", "release/2.x"],
+            "release/2.x",
+            {"latest": "ccf-1.0.1", "latest_same_lts": "ccf-2.0.0", "next": None},
+        ),
+        # Patch to 1.x
+        make_test_vector(
+            ["ccf-1.0.0", "ccf-1.0.1", "ccf-2.0.0", "ccf-1.0.2"],
+            ["release/1.x", "release/2.x"],
+            "ccf-1.0.3",
+            {"latest": None, "latest_same_lts": "ccf-1.0.2", "next": "ccf-2.0.0"},
```
eddyashton, line 373: It looks like this is the only test case that sets `next`, so there are none which set both `latest` and `next`. Can we add one to confirm that that kind of sandwich release will work correctly (ie - testing `2.0.x` once `3.0.0` is released, and still recognising `1.0.y` as latest)?

eddyashton (approved):

tests/infra/github.py
```diff
@@ -216,47 +240,59 @@ def install_release(self, tag):
         LOG.info("Unpacking debian package...")
         shutil.rmtree(install_directory, ignore_errors=True)
         install_cmd = ["dpkg-deb", "-R", download_path, install_directory]
-        assert infra.proc.ccall(*install_cmd).returncode == 0, "Installation failed"
+        assert (
+            subprocess.run(install_cmd, check=True).returncode == 0
```
achamayou, line 244: check=True already implies returncode == 0 and throws if not: https://docs.python.org/3/library/subprocess.html#subprocess.run, so the assert is unreachable.

achamayou (approved):

tests/infra/github.py
jumaffre: Good point. This is now checked too.

PR #3460, opened 2022-01-26
Commit evidence in ledger

### LINKED ISSUE #3295: Commit evidence in receipts
# Commit receipts

Terminology: **Execution Receipt** vs **Commit Receipt**

`Commit Receipt := Execution Receipt + Evidence of Commit`

# How can a user trust a receipt is for a committed state?

### The contents of the signature Tx stay in enclave until commit?

- Memory usage
- Complexity for the host, breaks send append entries as range

**No**

### The contents of the signature tx are temporarily encrypted when written out, and decrypted and re-written in place on commit.

- Require new SendAppendEntriesSigned, with a key tagging along
- Size/serialisation complexity

**No**

### - A later signature

`TxID <- Sig1 = Sig([..TxID..]) <- Sig2([commit >= Sig1])`

- Latency for producing receipt
- Can never get a receipt over the last transaction.
- Must be careful not to lose uncommitted proof during recovery, seems fine.

**No**

`TxID <- Sig(commit >= TxID)`? Doesn't bound commit latency. Lose provenance. **Also No**

### - Another signature

- On demand `Sig(commit >= TxID)`, not in the ledger: expensive, cache (efficient via view history for old values)? Stable across catastrophic recoveries.
- Batch - Off ledger - different tradeoff, extra latency, less execution cost - On ledger - done by primary, least execution cost, even more latency
  Note: on ledger not necessary for recovery, because the members decide where they resume.

**Can work, but expensive**

### - A Nonce/Secret

Cheap to produce, include digest in receipt (committment), release in the fullness of time on commit.

```
TxDigest := Digest(Write Set) + Digest(User claims) + Digest(Commit Nonce/Secret [+ TxID])
Commit Nonce/Secret := Digest(Ledger Secret[TxID], TxID)
```

Note: ledger alone doesn't tell you what's committed. Need a receipt (or a snapshot).
Doesn't seem like a big problem, if service is live, can ask for receipt. If not, members decide, persistence is meaningless and provenance can still be checked.

If `Digest(Nonce)` in ledger, then:

- anything committable can be recovered.
- possible to emit quasi receipt (with only digest and not nonce) -> not really a problem, verifiers have to stick to algorithm

Else:

- can't give quasi receipt
- but can't do public recovery or verify offline
  => Digest(Nonce) _must_ be in the ledger, or persisted somewhere

**Seems best, simple and low overhead**

#### What goes in the nonce?

- `Commit Nonce/Secret := HMAC(Ledger Secret[TxID], TxID)`

Granular, unique per-Tx.

- `Commit Nonce/Secret := HMAC(Ledger Secret[TxID@Previous signature], TxID@Previous signature)`

Same nonce for all transactions between `Sig` and `next(Sig)` => fewer nonces if we decide to store them.

But more difficult for a node to derive the nonce, must find signature that immediately precedes transaction. Unless we tag each Tx with the TxID of the previous signature. Or tag the TxID in the next signature, but then there's no provenance.

Intuition of difficulties around rekey, but no obvious counter-example.

#### Can the nonce go in the ledger?

To make sure the ledger alone is enough to produce commit receipts (assuming transparent user claims!), we could store the nonces in either their own transaction, or a signature, from time to time.

It must not be a signature, signatures can only contain a signature!

If we store the nonces, having them per signature rather than per transaction may be attractive.

But storing the nonces in a separate Tx generates a constant stream of transactions, the "Treadmill problem".

If we batch, we must batch per signature because:

- we can't commit between signatures! anything more precise is wasteful of space.
- not batching every signature potentially delays commit reveal/receipt.

# TxID in receipt

## How can a user trust a receipt is for a specific TxID?

1. Bind at the leaf: include the TxID in a user separable way inside the TxDigest (ie. not in the WriteSet) **Yes**
2. Bind at the signature: execution receipt could sign over (root + TxID), path offset to get TxID, ONLY for canonical receipt, otherwise need view history. **No**

# Preferred direction summary

## Top-level

```
User Claims Digest := Digest(User Claims)
TxDigest := Digest(Service Claims Digest + User Claims Digest)

Service Claims Digest A := Digest(Digest(Write Set), Digest(Commit Nonce), Digest(TxID))
Service Claims Digest B := Digest(Digest(Write Set), Digest(Commit Nonce + TxID))
Service Claims Digest C := Digest(Digest(Write Set), Digest(Commit Nonce), TxID))

or

TxDigest := Digest(Digest(Write Set), Digest(User claims), Digest(Commit Nonce/Secret + TxID))
```

## Commit evidence

```
Commit Nonce/Secret := Digest(Ledger Secret[TxID], TxID)
```

Easy to derive, requires no storage.

achamayou:
Ready for early review, with some minor TODOs left. Performance will be investigated and improved in later PRs.

- [x] Fix unit tests that don't set an encryptor
- [x] Fix JS FFI
- [x] Fix reconfiguration test
- [x] Documentation
- [x] Compatibility patch on 1.x #3474
- [x] Disable historical_receipt_test for js_v8
- [x] Fix historical_queries_test
- [x] Clean up last commit that fixes problems caught by the LTS test

Fixes #3295


python/ccf/ledger.py
```diff
@@ -23,6 +23,8 @@
 import ccf.receipt
 from hashlib import sha256
 
+from pkg_resources import EntryPoint
```
jumaffre, line 26: Unused?

src/crypto/test/bench.cpp
```diff
@@ -372,3 +392,10 @@ namespace Base64_bench
   auto openssl_base64_5000 = base64_bench<5000>;
   PICOBENCH(openssl_base64_5000).PICO_HASH_SUFFIX();
 }
+
+PICOBENCH_SUITE("hmac");
+namespace Base64_bench
```
jumaffre, line 397:
```suggestion
namespace HMAC_bench
```

src/crypto/test/crypto.cpp
```diff
@@ -594,3 +595,10 @@ TEST_CASE("x509 time")
     }
   }
 }
+
+TEST_CASE("hmac")
+{
+  std::vector<uint8_t> key(0, 32);
+  std::vector<uint8_t> data(0, 64);
+  auto r = crypto::hmac(MDType::SHA256, key, data);
```
jumaffre, line 603: Maybe check that calling the function twice with the same inputs return the same result, and than calling with slightly different arguments return a different output?

src/node/rpc/claims.h
```diff
@@ -14,14 +14,31 @@ namespace ccf
...
+      commit_evidence_digest,
+      crypto::Sha256Hash(
+        write_set_digest, claims_digest, commit_evidence_digest));
+    return crypto::Sha256Hash(
+      write_set_digest, claims_digest, commit_evidence_digest);
+  }
+
+  static crypto::Sha256Hash entry_leaf(
+    const std::vector<uint8_t>& write_set,
+    const crypto::Sha256Hash& commit_evidence_digest)
```
jumaffre, line 34: Would it be simpler to have only one `entry_leaf()` function with an optional `claims_digest` argument?

src/kv/committable_tx.h
```diff
@@ -44,14 +46,27 @@ namespace kv
       }
 
       auto e = store->get_encryptor();
-      auto entry_type = claims_digest.empty() ? EntryType::WriteSet :
-                                                EntryType::WriteSetWithClaims;
+      auto commit_nonce = e->get_commit_nonce({commit_view, version});
+      commit_evidence = fmt::format(
+        "ce:{}.{}:{}", commit_view, version, ds::to_hex(commit_nonce));
```
jumaffre, line 51: Since the receipt is very much JSON already, would it be nicer if the comment evidence was in the format `{"tx_id": "2.3", "commit_nonce": "<some hash>"}`?

src/kv/committable_tx.h
```diff
@@ -44,14 +46,27 @@ namespace kv
       }
 
       auto e = store->get_encryptor();
-      auto entry_type = claims_digest.empty() ? EntryType::WriteSet :
-                                                EntryType::WriteSetWithClaims;
+      auto commit_nonce = e->get_commit_nonce({commit_view, version});
```
jumaffre, line 49: As discussed, this now required the encryptor to be set on the store, even when serialising public maps only (which has been useful for some unit tests). Should we only set this if the encryptor exists, which would make the commit evidence optional, which may complicate some other tests/checks.

achamayou: That's a good question, and I don't know. I've been adding encryptors everywhere for now, because it seems easier than making the tests subtle, and less likely to go wrong, but I'm open to ideas.

src/kv/committable_tx.h
achamayou: The receipt aren't JSON, we only have a JSON sample but some applications want to use binary formats. This format is I think a good half-way house between binary formats and JSON, but I'm open to suggestions that are also compatible with both.

src/crypto/test/crypto.cpp
achamayou: Yes, done.

python/ccf/ledger.py
achamayou: Yes, thank you

src/crypto/test/bench.cpp
achamayou: Done

src/node/rpc/claims.h
achamayou: Yes, but then it means either a ref to an optional, which we have to construct when there's no claims, or switching to a ptr. Meh?

jumaffre: `const std::optional<ClaimsDigest::Digest>& claims_digest = std::nullopt` should be OK, no?

src/kv/committable_tx.h
jumaffre: Good point, I'd missed that the formatting of the receipt was up to the application. I guess we could install a formatting lambda on the store to specify the serialisation format - but not for this PR.

src/kv/committable_tx.h
jumaffre: Yeah, we might as well bite the bullet now and make the encryptor mandatory on the store OR we can set a default commit once (i.e. `00000...00`) if the encryptor isn't set?

src/node/rpc/claims.h
achamayou: Sure but then we have to construct a std::nullopt even when there are no claims.

jumaffre: Not if `claims_digest` is the last argument to the function. See e.g. https://github.com/microsoft/CCF/blob/f6f4bcecd9e397ef69fc14780d5b6892a1f961bc/src/consensus/aft/raft.h#L1429

src/kv/committable_tx.h
achamayou: I've gone with setting the null encryptor where necessary for two reasons as discussed:

1. I don't think it's a very good default for a normal user.
2. It's easier to catch the exception where it's not set than deal later on with unexpectedly incorrect digests.

src/kv/kv_types.h
```diff
@@ -513,15 +535,18 @@ namespace kv
...
 
   public:
     MovePendingTx(
       std::vector<uint8_t>&& data_,
       ccf::ClaimsDigest&& claims_digest_,
+      crypto::Sha256Hash&& commit_evidence_digest_,
       ConsensusHookPtrs&& hooks_) :
       data(std::move(data_)),
       claims_digest(claims_digest_),
+      commit_evidence_digest(commit_evidence_digest_),
```
jumaffre, line 549: Should this use `std::move()` too?

include/ccf/receipt.h
```diff
@@ -20,13 +20,16 @@ namespace ccf
     struct LeafComponents
     {
       std::optional<std::string> write_set_digest = std::nullopt;
+      std::optional<std::string> commit_evidence;
```
eddyashton, line 23:
```suggestion
      std::optional<std::string> commit_evidence = std::nullopt;
```

For consistency.

include/ccf/receipt.h
```diff
@@ -68,13 +71,16 @@ namespace ccf
...
           components.write_set_digest.value());
+        auto ce_dgst =
+          crypto::Sha256Hash::from_string(components.commit_evidence.value());
         auto cl_dgst =
           crypto::Sha256Hash::from_hex_string(components.claims_digest.value());
-        current = crypto::Sha256Hash(ws_dgst, cl_dgst);
+        current = crypto::Sha256Hash(ws_dgst, ce_dgst, cl_dgst);
       }
       else
       {
```
eddyashton, line 86: Can't make a suggestion, but the error below should mention all 3 elements, or just "Cannot compute leaf unless all leaf elements are set".

src/kv/raw_serialise.h
```diff
@@ -210,7 +210,7 @@ namespace kv
       else if constexpr (std::is_same_v<T, kv::EntryType>)
       {
         uint8_t entry_type = read_entry<uint8_t>();
-        if (entry_type > 2)
+        if (entry_type > 4)
```
eddyashton, line 213: This is an obscure error next time we add an EntryType, would be nice to handle it automatically.

```suggestion
        if (entry_type > kv::EntryType::MAX)
```
and add `MAX = WriteSetWithCommitEvidenceAndClaims` to the enum definition.

src/kv/test/kv_test.cpp
```diff
@@ -2464,6 +2512,8 @@ TEST_CASE("Store clear")
 TEST_CASE("Reported TxID after commit")
 {
   kv::Store kv_store;
+  auto encryptor = std::make_shared<kv::NullTxEncryptor>();
+  kv_store.set_encryptor(encryptor);
```
eddyashton, line 2516: Not needed since you've added these lines. But next time we make something mandatory on the `Store`, or the boiler plate changes, we should probably just replace all of these with a `kv::test::Store` whose constructor sets null encryptors, stub consensuses, etc etc.

src/node/test/snapshot.cpp
```diff
@@ -74,14 +77,17 @@ TEST_CASE("Snapshot with merkle tree" * doctest::test_suite("snapshot"))
     ccf::MerkleTreeHistory target_tree(tree.value());
     REQUIRE(source_root_before_signature == target_tree.get_root());
 
-    target_tree.append(crypto::Sha256Hash(serialised_signature));
+    target_tree.append(ccf::entry_leaf(
+      serialised_signature, crypto::Sha256Hash::from_string("ce:0.4:")));
```
eddyashton, line 81: This looked like a bug to me, but it turns out that `from_string` is _not_ `from_hex_string`. I suggest we rename this function to `of_string`, or just make it a constructor for `Sha256Hash`. So then `from_` is consistently "load an existing hash", and construction is "hash this argument(s)".

src/kv/kv_types.h
achamayou: Done

include/ccf/receipt.h
achamayou: Done

src/kv/raw_serialise.h
achamayou: Done

src/kv/test/kv_test.cpp
achamayou: I did contemplate doing that in Store, as mentioned above, but at least for this particular task it wouldn't have paid off: setting a default null encryptor produces painfully surprising digests rather than an early exception. Having a `make_test_store()` that does this explicitly when wanted is definitely a good step, I agree.

src/node/rpc/claims.h
achamayou: I'm not convinced this buys us meaningful simplicity, we only get to de-duplicate one line: the write set digest computation, only to dispatch to the two paths in a runtime branch.

src/node/test/snapshot.cpp
achamayou: Yes, the current digest interface isn't particularly nice, I agree. I'm tempted to go for an extra constructor, since all the other ones digest their inputs, whereas the from_* are conversions.

If I could start from scratch, I think I'd be tempted to make both named.

doc/architecture/ledger.rst
```diff
@@ -35,11 +35,11 @@ The following table describes the structure of a serialised transaction as it is
...
 |          | uint64_t                                 | Length of serialised public domain                                      |
 +----------+------------------------------------------+-------------------------------------------------------------------------+
-|          | :cpp:type:`kv::EntryType`                | One of WriteSet, Snapshot or WriteSetWithClaims                         |
+|          | :cpp:type:`kv::EntryType`                | Snapshot, or a WriteSet variant                                         |
 +          +------------------------------------------+-------------------------------------------------------------------------+
 |          | :cpp:type:`kv::Version`                  | Transaction version                                                     |
 +          +------------------------------------------+-------------------------------------------------------------------------+
-|          | :cpp:type:`crypto::Sha256Hash`           | User-defined claims digest, only when entry type is WriteSetWithClaims  |
+|          | :cpp:type:`crypto::Sha256Hash`           | User-defined claims digest, when entry type is WriteSetWith.*Claims     |
```
jumaffre, line 42: We should document the new commit evidence digest here.

achamayou: Indeed, thank you for spotting it.

jumaffre (approved):

tests/e2e_logging.py
```diff
@@ -1336,8 +1354,8 @@ def additional_interfaces(local_node_id):
         if args.package == "samples/apps/logging/liblogging":
             network = test_receipts(network, args)
             network = test_historical_query_sparse(network, args)
-        network = test_historical_receipts(network, args)
         if "v8" not in args.package:
+            network = test_historical_receipts(network, args)
```
eddyashton, line 1358: Do you know why this test no longer works in v8?

eddyashton (approved):

tests/e2e_logging.py
achamayou: Yes, it's because it now requires leaf_components to be exposed so the commit_evidence is available. I will do that next, but it's not a requirement for 2.0.

src/node/historical_queries.h
```diff
@@ -105,6 +105,30 @@ namespace ccf::historical
...
+          return e->get_commit_nonce(
+            {transaction_id.view, transaction_id.seqno}, true);
+        }
+        else
+        {
+          throw std::logic_error("Store pointer not set");
+        }
+      }
+
+      std::string get_commit_evidence()
```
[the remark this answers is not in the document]
achamayou: Yes, thank you.

PR #3464, opened 2022-01-27
Include frontend prefix in generated OpenAPI documents

### LINKED ISSUE #3364: Include frontend prefix in OpenAPI endpoint spec
Each frontend is currently identified by a single URI prefix (`/gov/`, `/node`/, `/app/`) under which all corresponding endpoints are registered. Those are reflected in the OpenAPI spec for that frontend in the `servers` object, e.g.:
https://github.com/microsoft/CCF/blob/45fbadc02c44205034e8b55050de154fb6d8e345/doc/schemas/gov_openapi.json#L880-L883

This isn't ideal as the complete URI for each endpoint isn't directly listed in the OpenAPI spec, which means that the frontend prefix isn't displayed when referencing an endpoint in the Sphinx documentation, e.g. `/node/` missing in `/quotes/self`:
image

A fairly minor change would be to include the frontend prefix in each endpoint definition in the OpenAPI spec (and remove the use of `servers`). This would have consequences for the `app.json` file of JS applications which would have to either include the `/app/` prefix or have CCF automatically include it by default.

eddyashton:
Fixes #3364.

This is a low-impact approach, inserting the prefix only at OpenAPI generation but expecting it is omitted at definition and dispatch time. This means the OpenAPI is accurate, but there's no knock-on changes that must be implemented by the user.

Separately, I do think we should include this prefix everywhere, for consistency, and believe this is necessary to mimic existing service APIs (ie - those which want to handle `/foo` rather than `/app/foo`). But that also requires removing the frontend dispatch, so ignoring for now.

Most of this PR is just updating docs to correctly link to the now-prefixed OpenAPI element.


tests/schema.py
```diff
@@ -30,7 +30,9 @@ def run(args):
     documents_valid = True
     all_methods = []
 
-    def fetch_schema(client, prefix):
+    def fetch_schema(client, prefix, file_prefix = None):
```
eddyashton, line 33: Sorry these changes are unrelated, and part of an abandoned change. I noticed we don't fetch the API for any JS apps, so was adding that here. But it turns out they're generally _not_ valid OpenAPI yet. So we can look at implementing that separately.

jumaffre (approved):

rengolin (approved):
I got bitten by this every time I tried to use the endpoints after a week or two from previous (my brain's own cache invalidation time). I just followed the pattern on my own docs to match.

PR #3465, opened 2022-01-27
Change docs CSS to make text in JSON samples visible

eddyashton:
I use the dark theme on our docs, and noticed that strings in highlighted sections (such as JSON strings) have a weirdly light background, which makes them hard to read. This tweaks the CSS for those highlights, hopefully without breaking anything else.

Before:
image

After:
image

### DIFF
--- a/doc/_static/css/custom.css
+++ b/doc/_static/css/custom.css
@@ -5,3 +5,5 @@
 img {
   margin: .3rem !important;
 }
+
+body[data-theme="dark"] .highlight .s2 { background-color: #303030 } /* Literal.String.Double */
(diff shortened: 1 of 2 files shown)


eddyashton:
> @eddyashton can we try a different pygments theme instead that works for both kinds of backgrounds?

Sure! How do we do that?

achamayou:
@eddyashton by setting https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-pygments_style

eddyashton:
Found it. First one I tried (zenburn) seems nice, and handles dark theme highlights:

image

image

achamayou:
The copy-paste button looks weird, and I think I know why, but it has nothing to do with this.

achamayou (approved):

PR #3476, opened 2022-01-28
Combine `cchost` and `cchost.virtual`

eddyashton:
~~Opening as a draft for further discussion.~~

It's been bothering me for a while that we have a separate `cchost.virtual` binary. It makes manually starting nodes harder, as you need to change the binary (as well as the enclave image and type, this all seems redundant). I realised we can do that by reworking the build definitions. Rather than reusing the `VIRTUAL_ENCLAVE` definition for creating the virtual images, and taking hard branches on that, we can write `cchost` with an additive set of enclave types, defined by `CCHOST_SUPPORTS_[VIRTUAL/SGX]` defines.

This is largely juggling macros and definition points so that the code compiles with this approach. I think the result is a big useability win for any users/operators trying manual network starts - we can give precise errors on what types are supported/provided.


src/enclave/virtual_enclave.h
```diff
@@ -2,27 +2,28 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+#ifndef CCHOST_SUPPORTS_VIRTUAL
```
eddyashton, line 5: Renamed this file. It was `ccf_v.h` to indicate it was a redefinition of `ccf_u.h`/`ccf_t.h`, but actually that restriction was unnecessary. While doing this, I've removed a bunch of unused definitions from this file.

src/host/enclave.h
```diff
@@ -46,49 +78,86 @@ namespace host
...
-     * @param flags Flags passed to oe_create_enclave. eg OE_ENCLAVE_FLAG_DEBUG,
-     * OE_ENCLAVE_FLAG_SIMULATE. Alternatively, ENCLAVE_FLAG_VIRTUAL will not
-     * use OE at all, instead loading a shared library directly
+     * @param type Type of enclave to load, influencing what flags should be
+     * passed to OE, or whether to dlload a virtual enclave
      */
-    Enclave(const std::string& path, uint32_t flags) :
-      is_virtual_enclave(false),
-      e(nullptr)
+    Enclave(const std::string& path, EnclaveType type)
```
eddyashton, line 96: So this function does now have more awkward macro-enabled blocks, but the result is a consistent, user-optimised error reporting flow:

- Does the file exist? If not, throw an error
- Is this a type of file that this build of `cchost` can handle? If not, throw an error
- Does this file have the expected suffix? If not, throw an error
- Otherwise try to load it, and report if _that_ fails

Previously steps 2 and 3 were swapped, so you could (if you really tried) get a "have you tried *.enclave.so.signed?" message, spend a while building/finding that image, and then be told by exactly the same `cchost` "actually, I don't want any files like that".

src/host/configuration.h
```diff
@@ -13,14 +13,14 @@ namespace host
...
+    SGX_RELEASE,
+    SGX_DEBUG,
+    VIRTUAL,
   };
   DECLARE_JSON_ENUM(
     EnclaveType,
-    {{EnclaveType::RELEASE, "release"},
-     {EnclaveType::DEBUG, "debug"},
+    {{EnclaveType::SGX_RELEASE, "release"},
+     {EnclaveType::SGX_DEBUG, "debug"},
```
eddyashton, line 23: This PR is also baby steps towards clean multi-platform support. I wonder if we should be explicit now and rename `"release"` to `"sgx_release"`? But I think it's an unnecessary breaking change, and may prove unnecessary, depending what those platforms look like.

achamayou: Not all platforms will use an executable + .so, so I don't think we should worry about this too much for now. It's also easier and more compatible to create "new_platform_release" than rename existing types.

achamayou:
The idea sounds good, but the performance impact is very surprising.

eddyashton:
> The idea sounds good, but the performance impact is very surprising.

Agreed, that is extremely unexpected. There must be some unintended change in the CMake, presumably around snmalloc usage. I'll take a look on Monday.

tests/infra/path.py
```diff
@@ -39,10 +39,7 @@ def build_lib_path(lib_name, enclave_type=None, library_dir="."):
         return os.path.join(library_dir, os.path.normpath(f"{lib_name}{ext}"))
 
 
-def build_bin_path(bin_name, enclave_type=None, binary_dir="."):
-    if enclave_type == "virtual":
-        bin_name = f"{bin_name}.virtual"
-
+def build_bin_path(bin_name, binary_dir="."):
```
jumaffre, line 42: I believe this is what breaks the LTS compatibility test for `NoSGX`: 1.x nodes still need to start with `cchost.virtual`.

jumaffre (commented):
I don't think this will work for release version of CCF that are built only with `-DCOMPILE_TARGETS=sgx` (see `matrix.yml`) as the resulting `cchost` executable will not be able to run virtual enclaves. Should `cchost` always be able to run virtual enclaves, no matter the value of `COMPILE_TARGETS`?

cmake/ccf_app.cmake
```diff
@@ -41,9 +41,12 @@ set(OE_TARGET_ENCLAVE_AND_STD openenclave::oeenclave openenclave::oelibcxx
...
 )
 # These oe libraries must be linked in specific order
-set(OE_TARGET_ENCLAVE_CORE_LIBS openenclave::oeenclave openenclave::oesnmalloc
-                                openenclave::oecore openenclave::oesyscall
+set(OE_TARGET_ENCLAVE_CORE_LIBS openenclave::oeenclave openenclave::oecore
+                                openenclave::oesyscall
 )
+if(USE_SNMALLOC)
+  list(APPEND OE_TARGET_ENCLAVE_CORE_LIBS openenclave::oesnmalloc)
+endif()
```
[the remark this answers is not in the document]
eddyashton: Yup, that's what the performance regression was. Reverted this change.

eddyashton:
> I don't think this will work for release version of CCF that are built only with `-DCOMPILE_TARGETS=sgx` (see `matrix.yml`) as the resulting `cchost` executable will not be able to run virtual enclaves. Should `cchost` always be able to run virtual enclaves, no matter the value of `COMPILE_TARGETS`?
> 
> I think that touching the daily canary might confirm this?

That's deliberate, and I think safe - if you build with `-DCOMPILE_TARGETS=sgx`, you will not build `.virtual.so` enclaves, and will build a `cchost` which cannot launch virtual enclaves. I believe we only set this for explicitly `NoSGX` or `SGX` targets, after which we only run platforms-specific tests? But I'll touch the canary to confirm that!

tests/infra/path.py
eddyashton: I assumed so, but unable to repro locally. Will investigate when I get time.

achamayou (approved):

achamayou (approved):

Issue #2726, opened 2021-06-24
BFT Reconfiguration: Switch AFT to new config on local commit of UpToDate quorum (rewrite all nodes as Trusted)


ashamis:
Upon a promotion RPC, the primary checks whether there are enough up-to-date LEARNER and TRUSTED nodes in the next configuration, it promotes all LEARNERs to TRUSTED, on which the switch to the next configuration depends.

Questions: Unclear what happens when there are no learners (e.g. only retirements, see also #2731)

achamayou:
Done in #2899 (the solution is: all nodes send ORCs, not just learners).

PR #3477, opened 2022-01-31
Format with Black v22.1.0

eddyashton:
A new version of Black was [just released](https://black.readthedocs.io/en/stable/change_log.html#id1). It contains exactly one breaking change for our code:

> Remove spaces around power operators if both operands are simple (#2726)

I think we want to stick with the latest version (especially since it's finally non-Beta!), but this kind of behaviour change is extremely annoying (and will likely break on our next `release/1.x` build...).

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -156,7 +156,7 @@ def test_large_messages(network, args):
             # pass but not others, and finding where does it fail).
             log_id = 40
             for p in range(10, 20) if args.consensus == "CFT" else range(10, 13):
-                long_msg = "X" * (2 ** p)
+                long_msg = "X" * (2**p)
                 check_commit(
                     c.post("/app/log/private", {"id": log_id, "msg": long_msg}),
                     result=True,


jumaffre (approved):

eddyashton:
From Black's stability policy:

```
Stability Policy
----------------

The following policy applies for the *Black* code style, in non pre-release
versions of *Black*:

- The same code, formatted with the same options, will produce the same
  output for all releases in a given calendar year.

  This means projects can safely use `black ~= 22.0` without worrying about
  major formatting changes disrupting their project in 2022. We may still
  fix bugs where *Black* crashes on some code, and make other improvements
  that do not affect formatting.

- The first release in a new calendar year *may* contain formatting changes,
  although these will be minimised as much as possible. This is to allow for
  improved formatting enabled by newer Python language syntax as well as due
  to improvements in the formatting logic.
```

So we could pin to a year, or accept that this kind of break will hit us "roughly now" every year.

FWIW, I think this formatting change is bonkers. The hugging policy is a heuristic, only loosely approximating operator precedence, based on a few claims that it makes (a very specific instance of) maths code (_slightly_) easier to read. That seems like a weird trade-off for an opinionated, uncompromising, deterministic formatter to make! Surely just opening the door for a never-ending parade of "why isn't _this_ operand considered simple?" issues! I'll get off my soapbox now.

achamayou (approved):

achamayou:
@eddyashton fwiw, I'm also not a fan of the hugging policy. The value of black for us (or at least for me), isn't in the format style itself but instead in the fact that there's only one of it, without options, and that the tool runs quickly and reliably for both checking and formatting. Put another way, the value of the tool is the lack of formatting PRs and review comments about formatting.

If that changes over time, and too many changes are introduced, we can either decide to pin or switch formatters.

PR #3479, opened 2022-01-31
Fix JSON configuration RPC public address

jumaffre:
Spotted by @letmaik 

These weren't in sync with the C++ code.

### DIFF
--- a/doc/host_config_schema/cchost_config.json
+++ b/doc/host_config_schema/cchost_config.json
@@ -44,7 +44,7 @@
                 "type": "string",
                 "description": "Local address the node binds to and listens on"
               },
-              "published_rpc_address": {
+              "published_address": {
                 "type": "string",
                 "default": "Value of 'bind_address'",
                 "description": "The published RPC address advertised to clients"
@@ -218,7 +218,7 @@
           "items": {
             "type": "string"
           },
-          "description": "List of ``iPAddress:`` or ``dNSName:`` strings to include as Subject Alternative Names (SAN) in node certificates. If none are set, the node certificate will automatically include the value of the main RPC interface 'published_rpc_address'"
+          "description": "List of ``iPAddress:`` or ``dNSName:`` strings to include as Subject Alternative Names (SAN) in node certificates. If none are set, the node certificate will automatically include the value of the main RPC interface 'published_address'"
         },
         "curve_id": {
           "type": "string",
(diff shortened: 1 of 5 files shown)


eddyashton (approved):

eddyashton:
Can we add a test that runs `cchost --check` with our sample configs, to make sure they remain schema compliant?

jumaffre:
> Can we add a test that runs `cchost --check` with our sample configs, to make sure they remain schema compliant?

There's one already: https://github.com/microsoft/CCF/blob/main/tests/e2e_operations.py#L119

eddyashton:
> > Can we add a test that runs `cchost --check` with our sample configs, to make sure they remain schema compliant?
> 
> There's one already: https://github.com/microsoft/CCF/blob/main/tests/e2e_operations.py#L119
> 
> In this case, the validation still worked because we don't set the `additionalProperties: false` field to the objects. I've tried to set it but it doesn't have any effect, which is intriguing. I'm investigating this now but I think we can merge this PR in the meantime.

Ah, no, I think the reason this still passed is that the schema was _also_ still using the old field name. Can we add a stage to `--check`, where after running valijson to make sure it matches the schema, we try to parse it as our expected C++ type, to make sure that schema matches the code?

jumaffre:
> > > Can we add a test that runs `cchost --check` with our sample configs, to make sure they remain schema compliant?
> > 
> > 
> > There's one already: https://github.com/microsoft/CCF/blob/main/tests/e2e_operations.py#L119
> > In this case, the validation still worked because we don't set the `additionalProperties: false` field to the objects. I've tried to set it but it doesn't have any effect, which is intriguing. I'm investigating this now but I think we can merge this PR in the meantime.
> 
> Ah, no, I think the reason this still passed is that the schema was _also_ still using the old field name. Can we add a stage to `--check`, where after running valijson to make sure it matches the schema, we try to parse it as our expected C++ type, to make sure that schema matches the code?

We do this too on `--check` (before the schema validation even): https://github.com/microsoft/CCF/blob/09ffcdf066c7eb0e20a573da799794798fe5bbd2/src/host/main.cpp#L130-L152

In this case, the schema validation was successful too because `published_address` is marked as optional in the `nlohmann::json` schema.

eddyashton:
> 
> We do this too on `--check` (before the schema validation even):
> 
> In this case, the schema validation was successful too because `published_address` is marked as optional in the `nlohmann::json` schema.

Ahhh, my bad. Cool! Ship it!

PR #3482, opened 2022-02-01
Remove failing install test

eddyashton:
This final test now fails due to schema validation, rather than the network failing to start up. It's no longer a necessary test, so removing it.


eddyashton:
Removing this "release means run extra tests" arg entirely from `test_install.sh` - it's already rarely tested, I don't think it's safe to have even rarer cases.

achamayou:
@eddyashton agreed, all release steps should be tested at least in Daily if not in CI, otherwise we can't be confident they work when we need to cut a release or a patch quickly.

achamayou (approved):

jumaffre:
> @eddyashton agreed, all release steps should be tested at least in Daily if not in CI, otherwise we can't be confident they work when we need to cut a release or a patch quickly.

This is tested in the Daily CI as part of the "SGX Release" job. The goal was to test that BFT wasn't allowed for release builds. I believe we still want such a check, even though the error is now so early (JSON config schema) that it's a bit misleading.

jumaffre (approved):

jumaffre (approved):

PR #3484, opened 2022-02-01
Longer validity range for certificates in unit tests

jumaffre:
It should fix the instability of the `tls_test` in the CI.

### DIFF
--- a/.azure-pipelines-templates/test.yml
+++ b/.azure-pipelines-templates/test.yml
@@ -9,7 +9,7 @@ steps:
       set -ex
       ASAN_SYMBOLIZERS=(/usr/bin/llvm-symbolizer-*)
       export ASAN_SYMBOLIZER_PATH=$(realpath ${ASAN_SYMBOLIZERS[0]})
-      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test -R tls_test --repeat-until-fail 1000
+      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
     env:
       METRICS_MONGO_CONNECTION: $(METRICS_MONGO_CONNECTION)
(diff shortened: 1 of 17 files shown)


.azure-pipelines-templates/test.yml
```diff
@@ -9,7 +9,7 @@ steps:
       set -ex
       ASAN_SYMBOLIZERS=(/usr/bin/llvm-symbolizer-*)
       export ASAN_SYMBOLIZER_PATH=$(realpath ${ASAN_SYMBOLIZERS[0]})
-      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
+      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test -R tls_test --repeat-until-fail 1000
```
jumaffre, line 12: Revert before merging.

src/node/test/channels.cpp
```diff
@@ -60,9 +60,9 @@ static crypto::Pem generate_self_signed_cert(
   const crypto::KeyPairPtr& kp, const std::string& name)
 {
   constexpr size_t certificate_validity_period_days = 365;
-  auto valid_from =
-    crypto::OpenSSL::to_x509_time_string(std::chrono::system_clock::to_time_t(
-      std::chrono::system_clock::now())); // now
+  auto valid_from = crypto::OpenSSL::to_x509_time_string(
+    std::chrono::system_clock::to_time_t(std::chrono::floor<std::chrono::weeks>(
+      std::chrono::system_clock::now())));
```
achamayou, line 65:
```suggestion
      std::chrono::system_clock::now() - 24h)));
```

achamayou (approved):

src/node/test/channels.cpp
jumaffre: Good catch!

PR #3497, opened 2022-02-02
Fix for async DNS resolution request cleanup

### LINKED ISSUE #3491: libuv DNS resolve may access deleted object
Seen in CI: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=40968&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=ae32e15b-0191-5f39-7410-be7e80165871&l=42798

```
65: ../src/host/tcp.h:592:41: runtime error: member call on address 0x61500000e900 which does not point to an object of type 'asynchost::TCPImpl'
65: 0x61500000e900: note: object has invalid vptr
65:  45 00 00 62  20 00 00 38 00 00 00 00  00 e9 00 00 50 61 00 00  60 f2 57 b9 a0 7f 00 00  0c 00 00 00
65:               ^~~~~~~~~~~~~~~~~~~~~~~
65:               invalid vptr
65:     #0 0xd2d333 in asynchost::TCPImpl::on_resolved(uv_getaddrinfo_s*, int, addrinfo*) /__w/1/s/build/../src/host/tcp.h:592:41
65:     #1 0x7fa0b955ba84 in uv__work_done (/lib/x86_64-linux-gnu/libuv.so.1+0xca84)
65:     #2 0x7fa0b955eea7  (/lib/x86_64-linux-gnu/libuv.so.1+0xfea7)
65:     #3 0x7fa0b956fb7f in uv__io_poll (/lib/x86_64-linux-gnu/libuv.so.1+0x20b7f)
65:     #4 0x7fa0b955f84b in uv_run (/lib/x86_64-linux-gnu/libuv.so.1+0x1084b)
65:     #5 0x4f13ff in main /__w/1/s/build/../src/host/main.cpp:527:5
65:     #6 0x7fa0b8e7b0b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
65:     #7 0x42521d in _start (/__w/1/s/build/cchost+0x42521d)
65: 
65: SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/host/tcp.h:592:41 in 
```

This looks like a legitimate bug - if we do an async resolve, the `TCP` object may have been deleted by the time `on_resolved` gets called.

I've created a branch which consistently repros this, by introducing an artificial delay before completing async resolves, so we more regularly (in the reconfig test) hit this: https://github.com/eddyashton/CCF/tree/uv_tcp_resolve_async_leak

I also tried a fix there, which was to record the existence of an async resolve in the `TCP`, and try to cancel it when the `TCP` is destroyed. That doesn't work, has exactly the same failure behaviour. I believe this is because the cancel doesn't happen 'fast enough' (ie - the resolve is already in-flight, so doesn't see/report its cancellation).

wintersteiger:
Here is an overly cautious fix for the cleanup of DNS resolution request objects. We probably don't stricly need `pending_resolve_requests` to be global, but I believe the lock on it is necessary. The key change here is that we run over all pending resolution requests in `~TCPImpl` and we set all references to `this` to `nullptr`, so that we can skip the callback when it fires later. 

Fixes #3491.

### DIFF
--- a/src/host/dns.h
+++ b/src/host/dns.h
@@ -4,10 +4,14 @@
 
 #include "../ds/logger.h"
 
+#include <unordered_set>
 #include <uv.h>
 
 namespace asynchost
 {
+  static std::unordered_set<uv_getaddrinfo_t*> pending_resolve_requests;
+  static std::mutex pending_resolve_requests_mtx;
+
   class DNS
   {
   public:
@@ -31,6 +35,11 @@ namespace asynchost
 
       if (async)
       {
+        {
+          std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
+          pending_resolve_requests.insert(resolver);
+        }
+
         if (
           (rc = uv_getaddrinfo(
              uv_default_loop(),
@@ -46,6 +55,10 @@ namespace asynchost
             host,
             service,
             uv_strerror(rc));
+          {
+            std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
+            pending_resolve_requests.erase(resolver);
+          }
           delete resolver;
           return false;
         }
--- a/src/host/tcp.h
+++ b/src/host/tcp.h
@@ -148,6 +148,18 @@ namespace asynchost
 
     ~TCPImpl()
     {
+      {
+        std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
+        for (auto& req : pending_resolve_requests)
+        {
+          // The UV request objects can stay, but if there are any references
+          // to `this` left, we need to remove them.
+          if (req->data == this)
+          {
+            req->data = nullptr;
+          }
+        }
+      }
       if (addr_base != nullptr)
       {
         uv_freeaddrinfo(addr_base);
@@ -589,7 +601,15 @@ namespace asynchost
 
     static void on_resolved(uv_getaddrinfo_t* req, int rc, struct addrinfo*)
     {
-      static_cast<TCPImpl*>(req->data)->on_resolved(req, rc);
+      {
+        std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
+        pending_resolve_requests.erase(req);
+      }
+
+      if (req->data)
+      {
+        static_cast<TCPImpl*>(req->data)->on_resolved(req, rc);
+      }
     }
 
     void on_resolved(uv_getaddrinfo_t* req, int rc)
--- a/src/host/tcp.h
+++ b/src/host/tcp.h
@@ -601,10 +601,8 @@ namespace asynchost
 
     static void on_resolved(uv_getaddrinfo_t* req, int rc, struct addrinfo*)
     {
-      {
-        std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
-        pending_resolve_requests.erase(req);
-      }
+      std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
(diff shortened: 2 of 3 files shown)


eddyashton (approved):

src/host/tcp.h
```diff
@@ -589,7 +601,15 @@ namespace asynchost
...
-      static_cast<TCPImpl*>(req->data)->on_resolved(req, rc);
+      {
+        std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
+        pending_resolve_requests.erase(req);
+      }
+
+      if (req->data)
+      {
+        static_cast<TCPImpl*>(req->data)->on_resolved(req, rc);
+      }
```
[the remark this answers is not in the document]
wintersteiger: Hmmm. I think you're right about that, we should extend that. In the spirit of being overly cautious I'll make that change, just in case the usage pattern of `TCPImpl` changes.

wintersteiger:
Performance: `ls_js_jwt_sgx_cft` is 2% down which looks to be a little outside of the noise. I don't think that's a real concern though.

src/host/tcp.h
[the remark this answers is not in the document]
eddyashton: There's a risk that this could lead to recursive locking, if `on_resolve_failed` re-attempts DNS resolution. It doesn't look like any behaviour does at the moment, but I think there's a risk either way.

src/host/dns.h
```diff
@@ -46,6 +55,10 @@ namespace asynchost
             host,
             service,
             uv_strerror(rc));
+          {
+            std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
+            pending_resolve_requests.erase(resolver);
+          }
```
jumaffre, line 61: Should we only append to `pending_resolve_requests` if `uv_getaddrinfo()` is successful?

src/host/tcp.h
```diff
@@ -589,7 +601,13 @@ namespace asynchost
 
     static void on_resolved(uv_getaddrinfo_t* req, int rc, struct addrinfo*)
     {
-      static_cast<TCPImpl*>(req->data)->on_resolved(req, rc);
+      std::unique_lock<std::mutex> guard(pending_resolve_requests_mtx);
+      pending_resolve_requests.erase(req);
```
jumaffre, line 605: Should the lock be released after this line?

wintersteiger: No! See previous (resolved) comment by Eddy. In theory (but not in the current usage pattern) it may become possible that `~TCPImpl` runs between check and use of `req->data`.

src/host/dns.h
wintersteiger: I'm not sure about that. Is it guaranteed that the callback will not be run if `uv_getaddrinfo(...) < 0`? If so, then we can do the insert afterwards, because we won't need `resolver->data`.

jumaffre: http://docs.libuv.org/en/v1.x/dns.html#api

> If successful, the callback will get called sometime in the future with the lookup result

Which I believe implies that the callback won't be called if the call isn't successful?

wintersteiger: Hmm. I think I may have been wrong about this. If we were to add it after a successful call to `uv_getaddrinfo`, then `~TCPImpl` and the callback could be executed (in both orders) before we add it to the set, potentially allowing an invalid pointer to slip through.

PR #3498, opened 2022-02-02
Add more request fields to JS API

### LINKED ISSUE #3092: js: expose more details in `Request` object (url, path, etc.)

eddyashton:
Resolves #3092.

Exposing URL components as described in the issue/new docs.

@letmaik: I haven't implemented this for V8. I started to, but noticed all the other tests in `custom_authorization.py` are not currently running with V8, and fail when they do. What's the expected API parity here?

### DIFF
--- a/js/ccf-app/src/endpoints.ts
+++ b/js/ccf-app/src/endpoints.ts
@@ -45,18 +45,50 @@ export interface Request<T extends JsonCompatible<T> = any> {
 
   /**
    * An object mapping URL path parameter names to their values.
+   * 
+   * GET /app/person/bob?fields=all, matched to /app/person/{name} => {"name": "bob"}
    */
   params: { [key: string]: string };
 
   /**
-   * The query string of the requested URL.
+   * The full original requested URL.
+   * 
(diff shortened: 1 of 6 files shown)


letmaik:
> @letmaik: I haven't implemented this for V8. I started to, but noticed all the other tests in `custom_authorization.py` are not currently running with V8, and fail when they do. What's the expected API parity here?

It's not critical if it's not done immediately (since it's experimental still), but I often find it easier to modify both pieces at the same time. At the very least, we should track any remaining work in https://github.com/microsoft/CCF/issues/3312.

jumaffre (approved):
This deserves a new entry in `CHANGELOG.md`.

PR #3500, opened 2022-02-02
Add receipt leaf components to Typescript definition

### LINKED ISSUE #3397: Add leaf components field to TypeScript `Receipt` type
Left-over from https://github.com/microsoft/CCF/pull/3292.

`js/ccf-app/src/global.ts` line 93

Also, I just noticed that `nodeId` may be incorrect, at least the C++ code says `node_id`.

eddyashton:
Resolves #3397.

### DIFF
--- a/src/js/historical.cpp
+++ b/src/js/historical.cpp
@@ -69,7 +69,7 @@ namespace ccf::js
     JS_SetPropertyStr(
       ctx,
       js_receipt,
-      "nodeId",
+      "node_id",
       JS_NewString(ctx, receipt_out.node_id.value().c_str()));
     auto proof = JS_NewArray(ctx);
     uint32_t i = 0;
--- a/js/ccf-app/src/global.ts
+++ b/js/ccf-app/src/global.ts
@@ -66,6 +66,23 @@ export interface ProofElement {
   right?: string;
 }
 
+export interface LeafComponents {
+  /**
+   * Hex-encoded hash of transaction's write set.
+  */
+  write_set_digest: string;
+
+  /**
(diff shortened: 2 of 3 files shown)


src/js/historical.cpp
```diff
@@ -69,7 +69,7 @@ namespace ccf::js
     JS_SetPropertyStr(
       ctx,
       js_receipt,
-      "nodeId",
+      "node_id",
```
jumaffre, line 72: I'm surprised no (end-to-end?) test broke because of this?

eddyashton: Yeah we don't have great coverage of the JS/TS API, especially the recent Receipt changes. I'd like to expand that, but in a separate PRs.

eddyashton: (Of course another way of saying that is "this PR is completely untested, but _looks ok_"...)

jumaffre: It may be worth tracking this in a new (or existing) GH issue?

eddyashton: #3503

achamayou: I don't think the qjs coverage of receipts is any less than for C++, it's only v8 that's missing some bits. The node_id was once used before the PEM was in the receipts, to look up files, but isn't anymore.

It doesn't really make a lot of sense to have it _and_ the PEM, the thought was that we'd have full state historical queries, and make signatures more compact and do the lookup on receipt creation (or even leave it to the user), but it's not clear that this is the right tradeoff. We have a discussion scheduled about receipts, I think we'll want to cover that then.

PR #3504, opened 2022-02-02
Extend checks in `extract-release-notes.py`, and run in `ci-checks.sh`

eddyashton:
We do PRs like #3499 to clean up `CHANGELOG.md` pretty often. We already have a script which scans for a release notes section for the current version. I've made it also look for the definition of the link that is acting as the section's title.

I also removed the `out_path` argument and resort to shell redirection, because I think it's a sensible route for simple tools like this. I'm tempted to re-add the arg parser so we can tell it to scan for arbitrary versions, but that seems unnecessary.

This is of course a stupidly unnecessary thing to touch mid-release, but I did it anyway.


achamayou:
It's tempting to take it one step further and:

1. hook it up to ci-checks.sh, should you PR go in if it makes the CHANGELOG dirty?
2. Implement -f, if the link is missing, it's pretty easy to generate it?

eddyashton:
> It's tempting to take it one step further and:
> 
> 1. hook it up to ci-checks.sh, should you PR go in if it makes the CHANGELOG dirty?
> 2. Implement -f, if the link is missing, it's pretty easy to generate it?

Running it in `ci-checks.sh` would be quite different, since there's not an obvious version name we're looking for at that point. Are we checking that every version title has a matching link, or that there is an unreleased section, or...?

achamayou:
> Running it in `ci-checks.sh` would be quite different, since there's not an obvious version name we're looking for at that point. Are we checking that every version title has a matching link, or that there is an unreleased section, or...?

I'm tempted to say: yes?

jumaffre (approved):
It seems that we only run this script as part of the release job, which when it breaks, causes a whole re-tag, PR, push, etc. cycle. Should we run it as part of `ci-checks.sh` too?

eddyashton:
Following discussion in this PR, I've made further modifications to the script. It now scans the entire CHANGELOG file, and checks that each version has an associated link definition. This runs as part of `ci-checks.sh`, with an optional `-f` flag to insert the assumed link.

To fulfil the original check, it also takes flags to specify target versions or derive that from git, and asserts that version is present. This is still run on release tags. In practice, this is probably still "too late" - the point where we know we want a `3.4.5` section in `CHANGELOG.md`, and can assert/fail a test if it's not present, is only once we've decided to tag and release `3.4.5`. Of course we can _manually_ check for it before then, but we could do that before...

tests/infra/github.py
```diff
@@ -163,8 +163,8 @@ def get_latest_dev_tag(self):
         return self._filter_released_tags(dev_tags)[0]
 
     def get_tags_for_major_version(self, major_version=None):
-        version_re = f"{major_version}\." if major_version else ""
-        tag_re = f"^{TAG_RELEASE_PREFIX}{version_re}([.\d+]+)(-rc.*|)$"
+        version_re = rf"{major_version}\." if major_version else ""
```
eddyashton, line 166: @jumaffre: These changes are just for the linting issues introduced in #3512, that are also solved by #3515 - I'll take those once that's merged.

PR #3506, opened 2022-02-03
Add some explicit base64 tests

eddyashton:
Exploring this failure from the CI:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=41170&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=3aae4c9d-f0dc-5b97-1b0e-f86312395af9&l=9690

Added tests for null strings, mis-formatted strings, etc. Can't find anything that will make the _encode_ throw.


eddyashton:
Converting this to a draft, going to add some more debug checks in the implementation

rengolin (approved):

eddyashton:
Went down a rabbit hole on this one. I was trying to work out if the error came from an earlier call that we hadn't checked the OpenSSL error code for. So I tried to add more consistent checking of OpenSSL error code (as well as/alongside checks of the return value). This exposed a few cases where we were leaking earlier error codes (but not the error that started this whole hunt), but also a bunch of cases where we make a call that produces an OpenSSL error code that we actually want to consume/clear/ignore (ie - OpenSSL couldn't parse a PEM? We handle that, clear the error!). Working out what set the error code (where the only debug information you have is that the _next_ call to OpenSSL was, probably, _this_) is an extremely frustrating snark hunt. You can track my progress on [this branch](https://github.com/eddyashton/CCF/tree/more_openssl_error_code_checks).

Anyway, that's all a sideshow to the original issue, as it never got close to revealing a base64 encode call that sets an OpenSSL error code. So I'm reverting this PR to get its original benefit - a few more test cases that rule out some simple hypotheses.

wintersteiger:
👍 good stuff, we definitely want those new tests and improved error checks.

Since the original failure indicated something ASN1-related (and therefore X509-related) during a base64 call, i.e. something completely unrelated, might it be possible that there's a way for an uninitialized error code variable to make it through to the checks?

achamayou (approved):

PR #3507, opened 2022-02-03
Upgraded fmt to 8.1.1

ihaller:
Constexpr and consteval changes in clang 12+ requires upgrading fmt to 8.1.1.
This does bring some changes as fmt::format assumes compile-time constant format string and verfication:
1) No constexpr formatter for scoped enums. Need to provide it.
2) Format string for format_to must be constant string.
3) Cannot have `{` or `}` as printed characters in format string.
4) Formatters need to be defined before first usage. 
5) FMT_STRING fails to build in some contexts involving lambdas. But the benefit for FMT_STRING is not clear to me for clang 12+ as it forces compile-time constant format strings even without it (at least this is my limited understanding).

Also added a fix for a newly added warning triggering on parsing code.

### DIFF
--- a/3rdparty/exported/fmt/args.h
+++ b/3rdparty/exported/fmt/args.h
@@ -143,6 +143,8 @@ class dynamic_format_arg_store
   }
 
  public:
+  constexpr dynamic_format_arg_store() = default;
+
   /**
     \rst
     Adds an argument into the dynamic store for later passing to a formatting
--- a/3rdparty/exported/fmt/chrono.h
+++ b/3rdparty/exported/fmt/chrono.h
@@ -11,12 +11,28 @@
 #include <algorithm>
 #include <chrono>
 #include <ctime>
+#include <iterator>
 #include <locale>
-#include <sstream>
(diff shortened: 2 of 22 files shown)


ihaller:
I don't mind splitting the clang 12 changes into a separate PR, but kept it together in case somebody wants to try it locally.

An alternative to disabling the warnings is to run the default initializer for T, which should be optimized away as the value is unused, but might result in extra code for debug builds.

achamayou:
@ihaller it seems that pragmas for the new warnings need to be gated by a clang version check:

```
../src/ds/json_schema.h:69:34: error: unknown warning group '-Wuninitialized-const-reference', ignored [-Werror,-Wunknown-warning-option]
#pragma clang diagnostic ignored "-Wuninitialized-const-reference"
```

ihaller:
Thanks for pointing out the warning gating! According to clang release notes the warning first shows up in 11.0 so I added that as the version limit.

achamayou (approved):

PR #3512, opened 2022-02-04
Various fixes for `lts_compatibility` test

### LINKED ISSUE #3494: Expand `github.py` test vectors
The test vectors in `github.py` test a lot of normal operation flow, but are missing a few edge cases (for example, the existence of only `-rc` tags on a branch, which prompted #3492).

Here's some things I'd like to add:
- More tests of badly named branches, tags, etc, both expected variations and unexpected.
- Assertions on the previous, same, and next LTS tags for every branch in each vector. Currently we specify several branches, but only test the results of one, we should make this more thorough.
- Potentially a progression, rather than a restatement of constants. ie, rather than repeating the tags `["ccf-1.0.0"], ["ccf-1.0.0", "ccf-1.0.1"]`, writing this as something like `mock_env.add_tag("ccf-1.0.1")`. That would make the test definition procedural rather than declarative, but I think would make it more readable.

jumaffre:
Resolves #3494

Fixes a number of issues with the `lts_compatibility` that became apparent during the release of `ccf-2.0.0-rc0`:
- Consider release candidate tags (`-rcX`)
- Ignore development tags (`-devX`)
- No longer ignore release tags if the corresponding release branch (`release/N.x`) doesn't exit
- Recovery test only considers releases prior to local branch (rather than also including future versions)

The overall logic should now be simpler and the set of unit tests has also been expanded.


tests/infra/github.py
```diff
@@ -78,8 +78,11 @@ def sanitise_branch_name(branch_name):
         # For simplification, assume that dev tags are only released from main branch
         LOG.debug(f"Considering dev tag {branch_name} as {MAIN_BRANCH_NAME} branch")
         return MAIN_BRANCH_NAME
-    if is_release_tag(branch_name):
-        equivalent_release_branch = f'{BRANCH_RELEASE_PREFIX}{branch_name.split(TAG_RELEASE_PREFIX)[1].split(".")[0]}.x'
+    elif is_release_tag(branch_name):
+        tag_major_version = int(branch_name.split(TAG_RELEASE_PREFIX)[1].split(".")[0])
```
achamayou, line 82: Not a huge amount of utility, but probably still more readable than lots of splits and index lookups: https://pypi.org/project/semver/

tests/infra/github.py
```diff
@@ -160,64 +162,53 @@ def get_latest_dev_tag(self):
...
-        # Branches are ordered based on major version, with ordering based on newest_first
-        return sorted(
-            self.release_branches,
-            key=get_major_version_from_release_branch_name,
-            reverse=newest_first,
+    def get_tags_for_major_version(self, major_version=None):
+        version_re = f"{major_version}\." if major_version else ""
+        tag_re = f"^{TAG_RELEASE_PREFIX}{version_re}([.\d+]+)(-rc.*|)$"
+        tags = sorted(
+            [tag for tag in self.tags if re.match(tag_re, tag)],
```
achamayou, line 169:
```suggestion
            tag for tag in self.tags if re.match(tag_re, tag),
```
Intermediary list isn't necessary.

tests/infra/github.py
```diff
@@ -328,150 +305,155 @@ def install_latest_lts_for_branch(self, branch, this_release_branch_only):
...
 
 
 if __name__ == "__main__":
     # Run this to test
     class MockGitEnv:
-        def __init__(self, tags, release_branches):
-            self.tags = tags
-            self.release_branches = release_branches
+        def __init__(self, tags=None, local_branch=None):
+            self.tags = set(tags) if tags else set()
```
achamayou, line 319:
```suggestion
            self.tags = set(tags or ())
```

achamayou (approved):

tests/infra/github.py
eddyashton: We have a `versionifier.py` that tries to parse to the setuptools version type. eg

```
    v = to_python_version("ccf-1.2.3-dev3-5-deadbeef")
    assert v.release == (1, 2, 3)
    assert v.dev == 3
    assert v.local == "5.deadbeef"
 ```

It's a bit weird (was written purely to generate the version number for `setup.py`), but could be re-used here to lean on existing version parsers?

Comment (jumaffre) [reply tests/infra/github.py:82]: Good spot. There's already a `get_version_from_tag_name()` function a few lines up that extracts the major version more cleanly and that I should have been using..

File: tests/infra/github.py (hunk 2 above)
Comment (jumaffre) [reply tests/infra/github.py:169]: Python tells me that I need one:
```bash
32: Test command: /usr/bin/unbuffer "python3" "/home/jumaffre/git/CCF/tests/infra/github.py"
32: Test timeout computed to be: 10000000
32:   File "/home/jumaffre/git/CCF/tests/infra/github.py", line 169
32:     tag for tag in self.tags if re.match(tag_re, tag),
32:     ^
32: SyntaxError: Generator expression must be parenthesized
```

eddyashton:
```suggestion
            (tag for tag in self.tags if re.match(tag_re, tag)),
```

Should be fine?

achamayou: Sorry, that's my fault, I checked the generator expression worked in sorted() but as a sole argument, when more than one argument is used you do need `()`. But you do avoid the construction of the intermediary list still.

PR #3513, opened 2022-02-04
Include Doxygen html pages in docs

eddyashton:
The Doxygen pages can include some very helpful dependency diagrams. I haven't found a better way to include them into the built docs other than copying them verbatim as `html_extra_path`. There is, for instance, no way to reference these diagrams with Breathe. Annoyingly, this means the xml Doxygen output is _also_ part of our docs, because it doesn't seem possible to output these to separate sub directories/convince Sphinx to rename/partially copy.


getting_started/setup_vm/roles/ccf_build/vars/common.yml
```diff
@@ -23,9 +23,10 @@ debs:
   - curl # client test infra
   - shellcheck # bash lint
   - iptables # partition test infra
-  - libclang1-9 # required by doxygen
-  - libclang-cpp9 # required by doxygen
+  - libclang1-9 # required to build doxygen
+  - libclang-cpp9 # required to build doxygen
   - pkg-config # required by v8
+  - graphviz # required to run doxygen
```
achamayou, line 29: I seem to remember we saved a good chunk of time when we ran doxygen for fewer files and XML only. Have we checked how much time the new change adds? I'm not saying we don't do it, just do we know?

eddyashton: Sure.

On main:
```
$ rm -rf doxygen
$ time doxygen
real    0m1.625s
user    0m1.429s
sys     0m0.197s
$ time doxygen
real    0m1.695s
user    0m1.400s
sys     0m0.281s
```

On this branch:
```
$ rm -rf doxygen/
$ time doxygen
real    0m7.575s
user    0m17.656s
sys     0m9.538s
$ time doxygen
real    0m3.395s
user    0m2.677s
sys     0m0.585s
```

achamayou: Ok, so with ~20 releases, this adds 16 * 20 = 5 minutes to the doc job, taking it from 30 to 35 minutes, plus whatever time the extra files being checked into git take.

I suggest we drop the doc for all 2.0.0-dev releases, that should more than offset this change.

eddyashton: I assume that the `sphinx-multiversion` build will use the relevant `Doxyfile` at each version it builds, so this extra work is only done on future tags? I'll try to confirm that in this PR before merging.

achamayou: I believe so, but we would still eventually get to that.

achamayou (approved):

jumaffre (approved):

PR #3518, opened 2022-02-04
Move service table definitions to new `service/tables` folder

jumaffre:
Codebase hygiene: Move table definitions to a new `service/tables` folder, rather than being under `node/`.


src/service/network_tables.h
```diff
@@ -1,33 +1,35 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the Apache 2.0 License.
 #pragma once
-#include "backup_signatures.h"
-#include "cert_bundles.h"
-#include "client_signatures.h"
-#include "code_id.h"
-#include "config.h"
+
```
eddyashton, line 4: This reminded me that I had a plan ages ago to remove `network_tables.h`. Not a high priority, but I've at least finally written it down [here](https://github.com/microsoft/CCF/issues/3521).

jumaffre: Good point. I originally intended to move the table names inside each table type file but I thought it should be left to another PR, which is now covered by #3521 so all is good!

PR #3519, opened 2022-02-04
Add governance action to force ledger chunks at following signature

wintersteiger:
The first half of #3155.

Adds a governance action to force a ledger chunk at the following signature and tests.


src/kv/store.h
```diff
@@ -101,6 +101,9 @@ namespace kv
     // If true, use historical ledger secrets to deserialise entries
     const bool is_historical = false;
 
+    // Ledger entry header flags
+    uint8_t flags = 0;
```
jumaffre, line 105: This needs to be cleared on rollback.

src/host/ledger.h
```diff
@@ -899,6 +901,19 @@ namespace asynchost
     size_t write_entry(
       const uint8_t* data, size_t size, bool committable, bool force_chunk)
     {
+      auto header = serialized::peek<kv::SerialisedEntryHeader>(data, size);
```
jumaffre, line 904: Could you add a unit test in `ledger.cpp` to cover this new behaviour?

tests/e2e_operations.py
```diff
@@ -149,8 +150,77 @@ def run_configuration_file_checks(args):
         assert rc == 0, f"Failed to run tutorial script: {rc}"
 
 
+@reqs.description("Forced ledger chunk")
+def test_forced_ledger_chunk(args):
```
jumaffre, line 154: Could you add this to the test suite too (`test_suite.py`)?

src/kv/store.h
wintersteiger: See line 615; I was conservative and roll back only the chunking flag instead of all of them.

tests/e2e_operations.py
```diff
@@ -149,8 +150,77 @@ def run_configuration_file_checks(args):
...
+            primary, proposal_body
+        )
+        proposal = network.consortium.vote_using_majority(
+            primary,
+            proposal,
+            careful_vote,
+            wait_for_global_commit=True,
+            timeout=3,
+        )
+        assert proposal.state == ProposalState.ACCEPTED
```
jumaffre, line 184: `vote_using_majority()` should already throw if the proposal is not accepted.

tests/e2e_operations.py
```diff
@@ -149,8 +150,77 @@ def run_configuration_file_checks(args):
...
+        )
+        proposal = network.consortium.get_any_active_member().propose(
+            primary, proposal_body
+        )
+        proposal = network.consortium.vote_using_majority(
+            primary,
+            proposal,
+            careful_vote,
+            wait_for_global_commit=True,
+            timeout=3,
```
jumaffre, line 182:
```suggestion
```

tests/e2e_operations.py
```diff
@@ -149,8 +150,77 @@ def run_configuration_file_checks(args):
...
+        )
+        assert proposal.state == ProposalState.ACCEPTED
+
+        # Issue some more transactions
+        network.txs.issue(network, number_txs=args.sig_tx_interval * 2)
+
+        ledger_dirs = primary.remote.ledger_paths()
+        network.stop_all_nodes(skip_verification=False)
```
jumaffre, line 190: Stopping all the nodes in a `test_` means that the test cannot be included in the suite. Could you use a variant of `network.get_latest_ledger_public_state()` instead?

tests/e2e_operations.py
```diff
@@ -149,8 +150,77 @@ def run_configuration_file_checks(args):
...
+                tables = pd.get_tables()
+                if (
+                    pd.get_seqno() >= proposal.seqno
+                    and next_signature is None
+                    and ccf.ledger.SIGNATURE_TX_TABLE_NAME in tables
+                ):
+                    next_signature = pd.get_seqno()
+            if first <= proposal.seqno and proposal.seqno <= last:
+                assert last == next_signature
+                assert next_signature - proposal.seqno < args.sig_tx_interval
```
jumaffre, line 212: Because signatures can also be triggered by a timeout, I feel like this check may not always be right?

tests/e2e_operations.py
```diff
@@ -149,8 +150,77 @@ def run_configuration_file_checks(args):
...
+                    last = pd.get_seqno()
+                tables = pd.get_tables()
+                if (
+                    pd.get_seqno() >= proposal.seqno
+                    and next_signature is None
+                    and ccf.ledger.SIGNATURE_TX_TABLE_NAME in tables
+                ):
+                    next_signature = pd.get_seqno()
+            if first <= proposal.seqno and proposal.seqno <= last:
+                assert last == next_signature
```
jumaffre, line 211: Could you add a new `LOG.info` line specifying the `seqno` of the proposal, the `seqno` of the signature and the name of the ledger chunk? This would be useful when looking at the test output.

src/kv/store.h
jumaffre: Thanks, I'd missed that!

src/kv/committable_tx.h
```diff
@@ -320,6 +325,11 @@ namespace kv
     {
       root_at_read_version = r;
     }
+
+    void set_store_flags(uint8_t flags)
+    {
+      store->set_flags(flags);
+    }
```
jumaffre, line 332: This calls makes the setting of this flag non-transactional. If the transaction is aborted before it commits, the flag will have been set on the store but the transaction will be not have been recorded in the global order. 

Instead, a flag should be set on this transaction and be propagated to the store when it commits (see call to `apply_changes()`).

src/kv/store.h
```diff
@@ -1205,5 +1209,15 @@ namespace kv
       // is incremented.
       return ReservedTx(this, term_of_last_version, tx_id);
     }
+
+    virtual void set_flags(uint8_t flags) override
```
jumaffre, line 1213: This will probably need a lock as the `store` is accessed by multiple transactions, unless this is called from a location where the lock is already taken, in which case `_unsafe` should be appended to the function name.

python/ccf/read_ledger.py
```diff
@@ -79,8 +79,9 @@ def counted_string(l, name):
 def dump_entry(entry, table_filter, tables_format_rules):
     public_transaction = entry.get_public_domain()
     public_tables = public_transaction.get_tables()
+    flags = entry.get_transaction_header().flags
     LOG.success(
-        f"{indent(2)}seqno {public_transaction.get_seqno()} ({counted_string(public_tables, 'public table')}) [{len(entry.get_raw_tx())} bytes]"
+        f"{indent(2)}seqno {public_transaction.get_seqno()} ({counted_string(public_tables, 'public table')}) [{len(entry.get_raw_tx())} bytes, flags={hex(flags)}]"
```
jumaffre, line 84: Maybe don't print them if the flags are all zeros?

tests/e2e_operations.py
wintersteiger: Yes, I totally agree. The only reason I do this is because the default signature interval here is quite large, I wanted to have a small one so that it doesn't take forever.

tests/e2e_operations.py
wintersteiger: I think it's always right, but it would be possible to see a chunk that was not triggered by the chunking flag. Do you have an idea how to test for this more reliably?

src/kv/test/kv_test.cpp
```diff
@@ -2863,4 +2864,79 @@ TEST_CASE("Range")
...
+    // Header flag is set in the last entry
+    const uint8_t* entry_data = data.data();
+    size_t entry_data_size = data.size();
+    auto header =
+      serialized::peek<kv::SerialisedEntryHeader>(entry_data, entry_data_size);
+    REQUIRE((header.flags & kv::EntryFlags::FORCE_LEDGER_CHUNK) != 0);
+  }
+
+  // Ledger chunk flag is not set in the store anymore
+  REQUIRE(
```
jumaffre, line 2939: Can we check that the flag is reset on rollback too? (but not on a no-op rollback)

src/kv/committable_tx.h
wintersteiger: Hmm, I assumed the store would be rolled back properly if the transaction fails. In  that case, where does the rollback logic go? Is that even necessary when the flag is in the transaction object?

jumaffre: If the transaction fails, the store is not rolled back (the rollback mechanism is used on a different code path, e.g. when a node becomes candidate/follower). For example, a transaction could fail if has conflicted with another one, or if the endpoint of this transaction throws an error. 

The only way to get this working transactionally is to set the flag on the store when we're sure that the transaction has been committed, i.e. "around" `apply_changes`.

wintersteiger: Ah! Got it, we need flags on both.

src/host/ledger.h
```diff
@@ -899,6 +901,19 @@ namespace asynchost
...
+
+      bool force_chunk_in_header =
+        header.flags & kv::EntryFlags::FORCE_LEDGER_CHUNK;
+
+      if (force_chunk_in_header)
```
jumaffre, line 909: Should this throw if `force_chunk_in_header` but `committable` is `false` since it would break the invariants that all ledger chunks should end on a committable entry (i.e. signature)?

tests/e2e_operations.py
jumaffre: You can also wait for `sig_ms_interval` (which is set at 100ms for the e2e test by default) to have a new signature with an existing network.

src/host/ledger.h
wintersteiger: Good point, yes it should!

tests/e2e_operations.py
jumaffre: I think it's OK if the chunking sometimes end up being the regular chunking rather than the forced one implementing in this PR. This is why including the test in the test suite should help and make sure this is tested in a range of different scenarios (the test suite also runs randomly in the Daily CI).

src/kv/committable_tx.h
```diff
@@ -160,6 +166,13 @@ namespace kv
         committed = true;
         version = c.value();
 
+        if (flags & AbstractStore::Flags::LEDGER_CHUNK_AT_NEXT_SIGNATURE)
+        {
+          store->set_flags(
+            store->get_flags() |
+            AbstractStore::Flags::LEDGER_CHUNK_AT_NEXT_SIGNATURE);
+        }
```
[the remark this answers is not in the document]
wintersteiger: So, you would prefer the flags `enum` to be an `enum class` and have `set_flags` take a single flag instead of a `uint8_t`?

tests/e2e_operations.py
wintersteiger: Right, that's an ok time to just wait for it, that should work.

tests/e2e_operations.py
wintersteiger: Sure, sounds good. I could perhaps also test for the signature timeout not being exceeded here, but that would be rather involved I think.

src/kv/committable_tx.h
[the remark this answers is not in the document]
jumaffre: Yes, exactly. The call to `set_flags()` shouldn't need to call `get_flags()` and then `OR` the flag it wants to set, but instead say `set_flags(AbstractStore::Flags::LEDGER_CHUNK_AT_NEXT_SIGNATURE)`.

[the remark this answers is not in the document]
wintersteiger: (First part fine, second part done)

tests/suite/test_suite.py
```diff
@@ -126,6 +127,8 @@
     # test the most complete ledgers
     governance_history.test_ledger_is_readable,
     governance_history.test_tables_doc,
+    # e2e_operations:
+    e2e_operations.test_forced_ledger_chunk,
```
jumaffre, line 131: Could you move this up as the previous `governance_history` tests should be run last.

samples/constitutions/default/actions.js
```diff
@@ -1122,4 +1122,13 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "request_ledger_chunk",
```
jumaffre, line 1126: This will require a new entry in the changelog.

tests/e2e_operations.py
```diff
@@ -53,6 +54,62 @@ def test_parse_snapshot_file(network, args):
...
+            ):
+                next_signature = pd.get_seqno()
+        if first <= proposal.seqno and proposal.seqno <= last:
+            LOG.info(
+                f"Found ledger chunk {chunk.filename()} with chunking proposal @{proposal.seqno} and signature @{next_signature}"
+            )
+            assert last == next_signature
+            assert next_signature - proposal.seqno < args.sig_tx_interval
+
+
```
jumaffre, line 112:
```suggestion
    return network
```

jumaffre (approved):

PR #3522, opened 2022-02-08
Document ledger invariants

jumaffre:
These invariants were not documented anywhere, even though we try very hard/have tests to verify those, mostly for auditor and operator convenience.

### DIFF
--- a/doc/operations/ledger_snapshot.rst
+++ b/doc/operations/ledger_snapshot.rst
@@ -42,8 +42,6 @@ The listing below is an example of what a ledger directory may look like:
 
 .. note:: On startup, a CCF node started with existing ledger files may suffix some of the file names with ``.corrupted`` if the ledger file cannot be parsed, depending on the sequence number the node will join from.
 
-
-
 Snapshots
 ---------
 
--- a/doc/operations/ledger_snapshot.rst
+++ b/doc/operations/ledger_snapshot.rst
@@ -97,7 +97,7 @@ Invariants
 - Committed ledger files start and end at the same ``seqno``.
 - Committed ledger files with the same name are byte-identical.
 
-3. Snapshots are always taken at the ``seqno`` of a signature transaction.
+3. Snapshots are always generated for the ``seqno`` of a signature transaction (but not all signature transactions trigger the generation of snapshot).
 
 4. The generation of a snapshot triggers the creation of a new ledger file. This is a corollary of 2. and 3., since new nodes should be able to join from a snapshot only and generate further ledger files that are the same as on the other nodes.
 
--- a/doc/operations/ledger_snapshot.rst
+++ b/doc/operations/ledger_snapshot.rst
@@ -42,6 +42,8 @@ The listing below is an example of what a ledger directory may look like:
 
 .. note:: On startup, a CCF node started with existing ledger files may suffix some of the file names with ``.corrupted`` if the ledger file cannot be parsed, depending on the sequence number the node will join from.
 
+
+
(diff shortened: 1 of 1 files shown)


doc/operations/ledger_snapshot.rst
```diff
@@ -86,3 +86,15 @@ A low value for ``ledger.chunk_size`` means that smaller ledger files are genera
 Similarly, a low value for ``snapshots.tx_count`` means that snapshots are generated often and that join/recovery time will be short, at the cost of additional workload on the primary node for snapshot generation.
 
 .. tip:: Uncommitted ledger files (which are likely to contain committed transactions) can also be used on join/recovery, as long as they are copied to the node's ``ledger.directory`` directory.
+
+Invariants
+----------
+
+1. To facilitate audit and verification of the integrity of the ledger, individual ledger files always end on a signature transaction.
+
+2. For operator convenience, all committed ledger files (``.committed`` suffix) are the same on all up-to-date nodes.
```
eddyashton, line 95: I think this could be slightly clearer/stronger. ie all committed transactions are present in exactly 1 `.committed` file (no duplicates), `.committed` files start and end at the same seqnos on every node (consistent chunking), and files with the same name on different nodes are _identical_ (no node-specific headers/contents in ledger). All caveated by "on the nodes where these files are present", or "all up-to-date nodes", or some other way of describing this.

jumaffre: Good points. Wrt "all committed transactions are present in exactly 1 .committed file", I think this is bit confusing as 
since some committed transactions will be stored in non-committed chunks (explained at the bottom of this section (https://microsoft.github.io/CCF/main/operations/ledger_snapshot.html#file-layout). I've added the other points as sub-points in the `2.` invariant.

doc/operations/ledger_snapshot.rst
```diff
@@ -86,3 +86,18 @@ A low value for ``ledger.chunk_size`` means that smaller ledger files are genera
...
+----------
+
+1. To facilitate audit and verification of the integrity of the ledger, individual ledger files always end on a signature transaction.
+
+2. For operator convenience, all committed ledger files (``.committed`` suffix) are the same on all up-to-date nodes. More precisely, among up-to-date nodes:
+
+- Committed ledger files start and end at the same ``seqno``.
+- Committed ledger files with the same name are byte-identical.
+
+3. Snapshots are always taken at the ``seqno`` of a signature transaction.
```
wintersteiger, line 100: Maybe rephrase to "snapshots end at a signature", to make it clear that not every signature produces a snapshot?

doc/operations/ledger_snapshot.rst
```diff
@@ -86,3 +86,18 @@ A low value for ``ledger.chunk_size`` means that smaller ledger files are genera
 Similarly, a low value for ``snapshots.tx_count`` means that snapshots are generated often and that join/recovery time will be short, at the cost of additional workload on the primary node for snapshot generation.
 
 .. tip:: Uncommitted ledger files (which are likely to contain committed transactions) can also be used on join/recovery, as long as they are copied to the node's ``ledger.directory`` directory.
```
wintersteiger, line 88: Should we start adjusting that to exclude temporary recovery files?

wintersteiger (approved):

doc/operations/ledger_snapshot.rst
jumaffre: Yes, as soon as the feature is on `main`, we should reflect this here.

Issue #3517, opened 2022-02-04
Clarify dependencies


eddyashton:
The build-time dependency graph between components in CCF has grown messier over time. We originally had `README.md`s in some `src/` folders indicating their intended dependencies, but these have rotted over time.

We should re-evaluate what dependencies we have and can enforce, to keep the code base navigable.

A few high-level goals to start with:
- Public headers should not include any private headers (#2428).
- 3rd party/STL libs are generally useable anywhere/visible everywhere (for instance `nlohmann/json` is used throughout the codebase, including in core containers), with a few exceptions: We should limit visibility of anything platform-specific (including SGX and host-only libraries).
- `ds/` should not depend on anything else. This is a place for generic containers and shims that may be used by multiple other components.
- We should aim to have a set of pure components which depend only on `ds/`. For instance `crypto/`, `tls/`, and `http/` could be written like this - generic protocol implementations and helpers, stripped of platform-integration features.
- Where we have tightly-coupled systems like `kv/` and `consensus/`, we should try to document this coupling to ensure it is required.

eddyashton:
The repo currently contains 3 READMEs which claim the following dependencies:

```
ds/README.md
# Data Structures

This directory contains data structures and utilities used by the rest of the code. No files
in this directory should include files from any other directory.

kv/README.md
# Transactional Key Value Store

This directory implements a transactional key value store. The only directory
it should depend on is `ds`.

crypto/README.md
# Crypto

This directory implements several cryptography helper types. The only directory
it should depend on is `ds`.
```

None of these dependencies are currently enforced. The actual dependencies between the folders under `src/` look like this:
![image](https://user-images.githubusercontent.com/6000239/153004744-5694a0d0-4c31-4dd5-a791-d97d47251a64.png)

Dotted lines show mutual/circular dependencies. Initial thoughts:
- We're aware of the mutual dependency between `kv/` and `consensus/`, and believe it's extremely hard to break
- We should try to break the dependency from `ds/` to `crypto/`
- A surprising number of things rely directly on `crypto/`, I wonder if it's worth splitting this
- This shows we do already have a few 'clean' components at the _bottom_ of the tree, which are only consuming other libraries (`clients/`, `host/`, `apps/`), we should enforce that distinction while we have it

achamayou:
Updated from `main` at [`488f8da`](https://github.com/microsoft/CCF/commit/488f8da0e339a4ac1752b0f27a3b34792ec47ed4) (2026-07-10).

[...]

PR #3523, opened 2022-02-08
Add a script to automatically graph dependencies from includes

eddyashton:
As part of #3517, I created a new `ci-checks` script that scans our `#include`s to render a mermaid diagram from them. I'm proposing checking that in so we know whenever we change those dependencies.

The other half of that script/this PR is trying to remove local includes so that this kind of analysis is easier. So, for instance, `src/node/backup_signatures.h` says `#include "node/service_map.h"` rather than `#include "service_map.h"`. That's more boiler-plate renaming whenever we _move_ files, but I think worth it for the consistency. I've not included that change here yet, because it's a 2.6k line diff over 176 files - think it should be agreed on before it obfuscates this PR.


eddyashton:
Closing this PR for now, will re-open in future if I have time to explore it further.

PR #3526, opened 2022-02-09
Fix new SAN issues in base64 test

eddyashton:
New base64 tests introduced yesterday spotted a regular leak on error paths, luckily caught by sanitizers. This fixes those, we should probably do a scan for similar paths in the new OpenSSL code.

Also fixing an uninitialized var from another of yesterday's PRs.

### DIFF
--- a/src/crypto/openssl/base64.h
+++ b/src/crypto/openssl/base64.h
@@ -27,7 +27,7 @@ namespace crypto
         return {};
 
       // Initialise the encode context
-      auto ctx = EVP_ENCODE_CTX_new();
+      OpenSSL::Unique_EVP_ENCODE_CTX ctx;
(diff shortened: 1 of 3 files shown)


jumaffre (approved):

wintersteiger:
Shouldn't this rather be done via a wrapper object to automatically clean up the stack?

eddyashton:
> Shouldn't this rather be done via a wrapper object to automatically clean up the stack?

Good call! I was going to say "yes, we should do this eventually", but then realised it's already available with the `OpenSSL::Unique_SSL_OBJECT` pattern. I've added a new instance for this context class, and it's now free'd automatically.

wintersteiger (approved):
Sweet, thanks for the quick fix, this looks much better!

PR #3527, opened 2022-02-09
Only build documentation for `main` and `release/N.x` branches

jumaffre:
Stop building documentation for release tags but instead only build documentation for the `HEAD` of `release/` branches as well as `main`. This will keep the documentation build time constant across LTS patches.


doc/conf.py
```diff
@@ -204,8 +204,10 @@
 
 # Set up multiversion extension
 
-# Build tags from ccf-1.0.1x
-smv_tag_whitelist = r"^ccf-(1\.\d+\.1\d+|2.*)$"
+whitelist_1_x = r"1\.\d+\.1\d+"  # from ccf-1.0.1x
+whitelist_2_x = r"2([.\d+]+)(-rc.*|)"  # all 2.x but no -dev
+whitelist_others = r"([3-9]|\d{2,}).*"  # all others
+smv_tag_whitelist = rf"^ccf-({whitelist_1_x}|{whitelist_2_x}|{whitelist_others})$"
```
eddyashton, line 210: Should we hack in some unit tests for these regexes? ie add a few a lines below like `assert re.match(smv_tag_whitelist, "ccf-2.0.0")`? I guess we don't really want to run them on every run of `conf.py`, since the errors will be buried in a call to `sphinx`...

eddyashton (approved):

doc/conf.py
achamayou: I don't think so because I believe we want to go to per-branch doc very soon, but if not, then I would say yes.

jumaffre: I've added some `asserts` below as you suggested. I think it's OK if it fails in the invocation of `sphinx` as 1) we'd see the error anyway (see below) and 2) we don't expect these to change very often.

```
Running Sphinx v4.4.0

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/home/jumaffre/git/CCF/env/lib/python3.8/site-packages/sphinx/config.py", line 340, in eval_config_file
    exec(code, namespace)
  File "/home/jumaffre/git/CCF/doc/conf.py", line 214, in <module>
    assert False
AssertionError

```

PR #3530, opened 2022-02-09
Re-enable ledger consistency verification

jumaffre:
The ledger consistency verification run at the end of all end-to-end tests seems to be skipped for most cases. I want to make sure this is still working, and measure any eventual impact on the total duration of our tests.


achamayou (approved):

jumaffre:
> @jumaffre how much time does this add to the CI?

It is within the noise for the jobs except for the performance tests, so I will increase the chunking threshold for these tests.

`main`:
image

this PR:
image
image

PR #3532, opened 2022-02-10
TLS and QUIC internal architecture documents

rengolin:
These are very rough texts with some updates from my recent investigation.

They should form the base for two following works: #3429 and QUIC.


doc/architecture/index.rst
```diff
@@ -68,4 +82,4 @@ Architecture
     ledger
     raft_tla
     node_to_node
-    indexing
\ No newline at end of file
+    indexing
```
eddyashton, line 85:
```suggestion
    indexing
    tls_internals
    quic_internals
```

This `toctree` directive produces the navigation tree on the left, so if new files aren't explicitly added they won't be easily discoverable.

doc/architecture/index.rst
```diff
@@ -57,6 +57,20 @@ Architecture
...
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    Internal implementation of TLS communication.
+
+    ---
+
+    :fa:`address-book` :doc:`quic_internals`
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    Internal implementation of QUIC communication.
```
eddyashton, line 72: Up for discussion, but I'd be tempted to add an "(in-progress)" disclaimer to this?

We don't currently have a great distinction in the architecture docs between descriptions of what _is_ implemented, and what we _plan_ to implement, but I think as we expand we should make that distinction!

rengolin:
Ping. If possible, I'd like to merge this just to get over with it and continue other docs and code separately?

eddyashton (approved):

Issue #2428, opened 2021-04-08
Private headers should not be needed to build CCF apps


eddyashton:
Longer term follow-up to #697. We currently implement much of the project in headers, and so our public headers include private headers which must be installed to build CCF apps. We should break these dependency chains to make a hard distinction between publicly supported APIs and implementation details. This will likely involve moving more implementations to `.cpp` files, and separating the project into individually built libraries.

achamayou:
`~/CCF/include/ccf$ find . -type f | xargs grep "#include \""  | grep -v 'ccf/' | cut -d ' ' -f 2 | sort -u`

```
./tx_id.h:#include "ds/hash.h"
./tx_id.h:#include "ds/json.h"
./claims_digest.h:#include "crypto/hash_provider.h"
./indexing/strategy.h:#include "kv/store.h"
./indexing/strategies/visit_each_entry_in_map.h:#include "indexing/indexing_types.h"
./indexing/strategies/seqnos_by_key_bucketed.h:#include "ds/lru.h"
./indexing/lfs_types.h:#include "ds/ring_buffer_types.h"
./base_endpoint_registry.h:#include "enclave/node_context.h"
./endpoint_registry.h:#include "ds/ccf_deprecated.h"
./endpoint_registry.h:#include "ds/json_schema.h"
./endpoint_registry.h:#include "ds/openapi.h"
./endpoint_registry.h:#include "http/http_consts.h"
./endpoint_registry.h:#include "node/endpoint_metrics.h"
./endpoint_registry.h:#include "node/rpc/serialization.h"
./endpoint_context.h:#include "enclave/rpc_context.h"
./endpoint_context.h:#include "http/authentication/authentication_types.h"
./endpoint.h:#include "ds/json.h"
./endpoint.h:#include "ds/openapi.h"
./endpoint.h:#include "kv/serialise_entry_blit.h"
./endpoint.h:#include "node/service_map.h"
./http_query.h:#include "ds/nonstd.h"
./tx.h:#include "crypto/hash.h"
./tx.h:#include "ds/ccf_assert.h"
./tx.h:#include "kv/kv_types.h"
./tx.h:#include "kv/untyped_map.h"
./historical_queries_interface.h:#include "consensus/ledger_enclave_types.h"
./historical_queries_interface.h:#include "ds/contiguous_set.h"
./historical_queries_interface.h:#include "kv/store.h"
./historical_queries_interface.h:#include "node/history.h"
[...]

PR #3533, opened 2022-02-10
Move some headers to public, remove some unnecessary includes

eddyashton:
Part of #2428.

This catches much of the low-hanging fruit:
- We think apps can use `LOG_INFO_FMT` and `DECLARE_JSON_TYPE`, so `logger.h` and `json.h` are moved to public includes
- Forward includes and a manual scan remove a few unnecessary includes
- The `jsonhandler` helper function implementations move to `.cpp`, removing includes of (private) implementation details

Still more cleanup to do, but I think it needs to be split into multiple PRs at this point. FWIW, this PR reduces the number of private includes from our public headers from 45 to 25.


include/ccf/tx.h
```diff
@@ -171,21 +141,64 @@ namespace kv
...
+      }
+      return handles;
+    }
+
     template <class THandle>
     THandle* get_handle_by_name(const std::string& map_name)
     {
-      auto search = all_changes.find(map_name);
-      if (search != all_changes.end())
+      auto possible_handles = get_possible_handles(map_name);
```
eddyashton, line 161: This refactor in `tx.h` is small steps towards it a public-only header. The problem is that while the interface we want to give users is conceptually simple (a `tx` object you can call `rw` on to get a handle, and a handle you can `get`/`put` on), the implementation is heavily templated so harder to hide. This pushes the templates up the stack slightly, so we could at this point hide most of the implementation details. But not (yet) the change set itself. I think that will be a whole PR to itself.

achamayou (approved):

PR #3534, opened 2022-02-10
Fix for `read_ledger.py` for snapshot parsing

jumaffre:
This was broken because untested in the CI. We now test this in `governance_history`.

### DIFF
--- a/tests/governance.py
+++ b/tests/governance.py
@@ -472,9 +472,6 @@ if __name__ == "__main__":
         governance_history.run,
         package="samples/apps/logging/liblogging",
         nodes=infra.e2e_args.max_nodes(cr.args, f=0),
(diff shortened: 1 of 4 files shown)


tests/governance.py
```diff
@@ -472,9 +472,6 @@ def add(parser):
         governance_history.run,
         package="samples/apps/logging/liblogging",
         nodes=infra.e2e_args.max_nodes(cr.args, f=0),
-        # Higher snapshot interval as snapshots trigger new ledger chunks, which
-        # may result in latest chunk being partially written
-        snapshot_tx_interval=10000,
```
jumaffre, line 477: I believe this was set because of old limitations of the Python infra which didn't cope well with ledger chunks that were in progress.

wintersteiger (approved):
LGTM!

PR #3535, opened 2022-02-10
Add option to skip verification ledger in `ccf.Ledger` Python module

jumaffre:
Verifying the integrity of a ledger when parsing it is useful by default, but we sometimes want to display the content of a single ledger file without having to parse all previous ledger files. This PR adds a new option to skip the integrity verification (also available via `read_ledger.py`). 

I understand there may be more subtle ways to do this so opening as a draft PR for discussion.

Also contains changes from #3534

### DIFF
--- a/python/ccf/ledger.py
+++ b/python/ccf/ledger.py
@@ -792,10 +792,12 @@ class Ledger:
     _current_chunk: LedgerChunk
     _ledger_validator: LedgerValidator
 
-    def _reset_iterators(self):
+    def _reset_iterators(self, insecure_skip_verification: bool = False):
         self._fileindex = -1
         # Initialize LedgerValidator instance which will be passed to LedgerChunks.
-        self._ledger_validator = LedgerValidator()
+        self._ledger_validator = (
+            LedgerValidator() if not insecure_skip_verification else None
+        )
 
     @classmethod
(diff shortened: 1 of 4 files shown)


jumaffre:
> o an integrity check from the last snapshot we can find, but allowing skipping with the right flag may also be useful of course.

Yep, that would be a good improvement for the script too, even though I haven't needed that feature just yet.

python/ccf/ledger.py
```diff
@@ -944,9 +951,13 @@ def last_verified_txid(self) -> TxID:
...
-        return TxID(
-            self._ledger_validator.last_verified_view,
-            self._ledger_validator.last_verified_seqno,
+        return (
+            TxID(
+                self._ledger_validator.last_verified_view,
+                self._ledger_validator.last_verified_seqno,
+            )
+            if self._ledger_validator
+            else TxID(0, 0)
```
eddyashton, line 960: I think it'd be better to return `None` here.

eddyashton: Or raise, I don't know which is more Pythonic.

jumaffre: I went for `None`.

eddyashton (approved):

PR #3536, opened 2022-02-10
Split authentication policies into public definition, private implementation

eddyashton:
A relatively standalone part of #2428.

This moves the existing auth policies from `http/authentication` to `endpoints/authentication`, and splits out the private implementation from the public definition. This gives each a clean public interface.

In practice, this exposes that the current publicly available types don't make it easy for apps to implement their own auth policies! But we will deal with that separately, and likely get most of the win from moving service table definitions to public.


include/ccf/endpoints/authentication/cert_auth.h
```diff
@@ -0,0 +1,86 @@
...
+      const override
+    {
+      return get_cert_based_security_schema();
+    }
+  };
+
+  struct NodeCertAuthnIdentity : public AuthnIdentity
+  {
+    ccf::NodeId node_id;
+  };
```
eddyashton, line 70: Diff doesn't deal well with moving this file _and_ splitting it in half. This is mostly `src/http/authentication/cert_auth.h` with some of the function implementations stripped out. This is the only semantic change - this struct used to contain a `ccf::NodeInfo node_info;`. I think that's a type we _do_ want to expose publicly eventually, but I also think it's presence here was accidental and inconsistent (no other policies bundle extra metadata here, they expect it to be looked up independently).

cmake/common.cmake
```diff
@@ -162,6 +162,11 @@ set(CCF_ENDPOINTS_SOURCES
     ${CCF_DIR}/src/endpoints/endpoint_registry.cpp
     ${CCF_DIR}/src/endpoints/base_endpoint_registry.cpp
     ${CCF_DIR}/src/endpoints/common_endpoint_registry.cpp
+    ${CCF_DIR}/src/endpoints/authentication/authentication_types.cpp
```
achamayou, line 165: For cpp files, where the collision risk is low, suffixes like _auth may or may not be necessary?

achamayou (approved):

Issue #1652, opened 2020-09-24
Failed public recovery entries will block further recoveries


jumaffre:
If a recovery fails (i.e. recovery shares are not successfully submitted), the ledger will contain a suffix of entries that are not covered by any recovery shares. This is because we don't issue new recovery shares on the new network until the old shares have been submitted (since the new ledger secrets encrypt the previous ones). If an operator retries a recovery with that ledger, the suffix will happily be deserialised (we only check the integrity of entries via the Merkle Tree at this point) but the recovery procedure will eventually fail.

Instead, when we recover a ledger, we should discard entries that belong to a new service but are not covered by recovery shares. For convenience of operators handling ledger files, we should also start a new ledger chunk when recovering a service, and possibly mark the "orphan" ledger chunks so that they are ignored by the new service.

Note that adding a new end-to-end test for this should be pretty trivial.

achamayou:
In case it's any help to someone reading this, the reason the recovery procedure will eventually fail is that integrity checks of replicated entries are performed with the applicable ledger secret.

PR #3539, opened 2022-02-11
Create a new ledger chunk when starting recovery process

jumaffre:
Part of #1652 

We now create a new ledger chunk when starting the recovery process. This means that operators can neatly distinguish ledger files that belong to different service epochs. 

To do so, the service open transaction sets a new flag on the `kv::Tx` to force a new ledger chunk **before** this transaction is replicated to the ledger.


src/host/test/ledger.cpp
```diff
@@ -1440,4 +1440,14 @@ TEST_CASE("Chunking according to entry header flag")
...
     REQUIRE(number_of_files_in_ledger_dir() == 3);
   }
+
+  INFO("Write an entry with the ledger chunking before header flag enabled");
+  {
+    auto ledger_files_count = number_of_files_in_ledger_dir();
+    entry_submitter.write(
+      is_committable, false, kv::EntryFlags::FORCE_LEDGER_CHUNK_BEFORE);
+
+    // Forcing a new chunk before created a new chunk to store this entry
```
wintersteiger, line 1450:
```suggestion
    // Forcing a new chunk before creating a new chunk to store this entry
```

src/node/node_state.h
```diff
@@ -130,7 +130,7 @@ namespace ccf
     kv::Version recovery_v;
     crypto::Sha256Hash recovery_root;
     std::vector<kv::Version> view_history;
-    consensus::Index last_recovered_signed_idx = 1;
+    consensus::Index last_recovered_signed_idx = 0;
```
wintersteiger, line 133: This change looks suspicious to me, because `last_recovered_signed_idx` isn't mentioned anywhere else in this changeset, but it's probably fine :-)

tests/recovery.py
```diff
@@ -182,8 +183,30 @@ def run(args):
...
                 node.verify_certificate_validity_period()
 
+            primary, _ = network.find_primary()
+
             LOG.success("Recovery complete on all nodes")
 
+    # Verify that a new ledger chunk was created at the start of each recovery
+    ledger = ccf.ledger.Ledger(primary.remote.ledger_paths(), committed_only=False)
+    for chunk in ledger:
+        first_tx_in_chunk = True
```
wintersteiger, line 193: Perhaps we should add functions to get first and last seqno from the `LedgerChunk`. That would make this a little easier to read, I think.

wintersteiger (approved):
LGTM!

src/node/node_state.h
jumaffre: Yes, I changed this in the abandoned PR #3525 as I used `last_recovered_signed_idx` but not in this PR. However, `last_recovered_signed_idx` should be initialised at `0` and not `1` and this shouldn't have any impact on the recovery procedure.

src/node/rpc/node_frontend.h
```diff
@@ -1275,6 +1275,9 @@ namespace ccf
           in.certificate_signing_request,
           in.public_key};
 
+        auto tx_ = static_cast<kv::CommittableTx*>(&ctx.tx);
+        tx_->set_flag(kv::AbstractStore::Flag::LEDGER_CHUNK_BEFORE_THIS_TX);
```
achamayou, line 1279: Why do we need to cast?

src/kv/serialised_entry_format.h
```diff
@@ -9,10 +9,12 @@
...
 {
   static constexpr auto entry_format_v1 = 1;
+  using SerialisedEntryFlags = uint8_t;
 
-  enum EntryFlags : uint8_t
+  enum EntryFlags : SerialisedEntryFlags
   {
-    FORCE_LEDGER_CHUNK = 0x01
+    FORCE_LEDGER_CHUNK_AFTER = 0x01,
+    FORCE_LEDGER_CHUNK_BEFORE = 0x02
```
achamayou, line 17: What's the difference between this and Flag in kv_types.h? Why do we have both?

achamayou (approved):

src/node/rpc/node_frontend.h
jumaffre: We don't want to expose the `set_flag()` API to the `kv::Tx` used by application endpoints, hence the cast to the internal `CommittableTx` type here.

eddyashton: We should check that `tx_` is not nullptr after the static cast.

tests/recovery.py
jumaffre: Done.

Issue #3026, opened 2021-09-27
Snapshotter can be taken out of consensus


jumaffre:
As pointed by @achamayou, Now that snapshotting is a stable feature of CCF, it is evident that the `Snapshotter` object could be taken out of the consensus module and become more tightly integrated with the `kv::Store`.

jumaffre:
Thanks to @wintersteiger for doing this work in #3737.

PR #3544, opened 2022-02-11
Add governance action to request snapshot

wintersteiger:
This adds a governance action to request a snapshot at the following signature.


src/node/snapshotter.h
```diff
@@ -23,6 +23,13 @@ namespace ccf
   public:
     static constexpr auto max_tx_interval = std::numeric_limits<size_t>::max();
 
+    enum class Flag : uint8_t
+    {
+      FORCE_SNAPSHOT_AT_NEXT_SIGNATURE = 0x01
+    };
+
+    kv::Version forced_snapshot_after_idx = -1;
```
wintersteiger, line 31: I don't like this extra variable at all, but it's needed because `commit()` would force snapshots too early sometimes (at the signature just before the proposal is accepted, instead of at the next one). That index could probably be put into `next_snapshot_indices` somehow, with a flag to force it immediately instead of checking the snapshot interval.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - The new `endorsement` configuration entry lets operators set the desired TLS certificate endorsement, either service-endorsed or node-endorsed (self-signed), for each network RPC interface of a node, defaulting to service-endorsed (#2875).
 - A new governance action `request_ledger_chunk` to request the creation of a ledger chunk at the next signature (#3519).
+- A new governance action `request_snapshot` to request the creation of a snapshot at the next signature (#3544).
```
jumaffre, line 14: We should probably rename those `trigger_` rather than `request_` to be consistent with existing proposal names.

jumaffre, line 14: Could you also add some brief documentation for these proposals in https://microsoft.github.io/CCF/main/operations/ledger_snapshot.html please?

tests/e2e_operations.py
```diff
@@ -87,31 +108,65 @@ def test_forced_ledger_chunk(network, args):
     # Check that there is indeed a ledger chunk that ends at the
     # first signature after proposal.completed_seqno
     ledger = ccf.ledger.Ledger(ledger_dirs)
```
jumaffre, line 110: This ignores uncommitted ledger files by default. How do we guarantee that the proposal completion is in a committed chunk here?

tests/e2e_operations.py
```diff
@@ -87,31 +108,65 @@ def test_forced_ledger_chunk(network, args):
...
+    )
+
+    proposal = network.consortium.vote_using_majority(
+        primary,
+        proposal,
+        careful_vote,
+    )
+
+    # Issue some more transactions
+    network.txs.issue(network, number_txs=13)
```
jumaffre, line 143: 13?

jumaffre: Same on line `100`.

tests/e2e_operations.py
```diff
@@ -87,31 +108,65 @@ def test_forced_ledger_chunk(network, args):
...
+
+    # Wait for the signature interval to ensure we see at least one signature
+    time.sleep(args.sig_ms_interval / 1000)
```
jumaffre, line 146: `txs.issue()` already waits for a signature to be emitted (since it waits for commit of the transactions) so this shouldn't be necessary here?

tests/e2e_operations.py
```diff
@@ -87,31 +108,65 @@ def test_forced_ledger_chunk(network, args):
...
+
+    ledger_dirs = primary.remote.ledger_paths()
+    snapshots_dir = network.get_committed_snapshots(primary)
+    network.check_ledger_files_identical()
```
jumaffre, line 150: I don't think this is necessary anymore since this is checked on network shutdown?

tests/e2e_operations.py
wintersteiger: Yes, a small prime number :-)

src/node/snapshotter.h
```diff
@@ -327,11 +336,21 @@ namespace ccf
         idx,
         next_snapshot_indices.front());
 
-      auto snapshot_idx = next_snapshot_indices.front();
-      if (snapshot_idx - last_snapshot_idx >= snapshot_tx_interval)
+      bool forced_snapshot = idx > forced_snapshot_after_idx &&
```
jumaffre, line 339: I don't see `forced_snapshot_after_idx` being set to anything apart from `-1` here?

src/node/node_state.h
```diff
@@ -1188,6 +1188,14 @@ namespace ccf
       tx2->set_flag(kv::AbstractStore::Flag::LEDGER_CHUNK_AT_NEXT_SIGNATURE);
     }
 
+    void request_snapshot(kv::Tx& tx) override
+    {
+      snapshotter->forced_snapshot_after_idx =
+        network.tables->current_version();
+      snapshotter->set_flag_unsafe(
+        ccf::Snapshotter::Flag::FORCE_SNAPSHOT_AT_NEXT_SIGNATURE);
```
jumaffre, line 1196: This is a side effect that is not tied to a specific `seqno` so it feels unsafe (how is it rolled back, etc.?). I would expect this to set a flag on `tx` which is later passed on the `store` and the store passes it to the snapshotter (through `raft` or better, #3026).

wintersteiger: It is tied to the tx index via `forced_snapshot_after_idx`, which can be rolled back there if we want to keep that variable. I didn't see a way for the store to pass this on to the snapshotter (it doesn't have a reference to the snapshotter or raft, but that may change after #3026 is done).

doc/operations/ledger_snapshot.rst
```diff
@@ -40,7 +40,10 @@ The listing below is an example of what a ledger directory may look like:
     -rw-rw-r-- 1 user user 1.1M Jan 31 14:00 ledger_92502-97520.committed
     -rw-rw-r-- 1 user user 553K Jan 31 14:00 ledger_97521 # File still in progress
 
-.. note:: On startup, a CCF node started with existing ledger files may suffix some of the file names with ``.corrupted`` if the ledger file cannot be parsed, depending on the sequence number the node will join from.
+.. note:: On startup, a CCF node started with existing ledger files may suffix some of the file names with ``.corrupted`` if the ledger file cannot be parsed, depending on 
+the sequence number the node will join from.
+
+.. note:: A new ledger chunk can also be created by the ``trigger_ledger_chunk`` governance action, which will produce a new chunk at the following signature transaction.
```
jumaffre, line 46:
```suggestion
.. note:: A new ledger chunk can also be created by the ``trigger_ledger_chunk`` governance action, which will automatically produce a new chunk at the following signature transaction.
```

src/kv/committable_tx.h
```diff
@@ -395,8 +409,11 @@ namespace kv
       // This is a signature and, if the ledger chunking flag is enabled, we
```
jumaffre, line 409: The comment needs updating to reflect the fact that we trigger a new ledger chunk on the snapshot too.

src/node/snapshotter.h
```diff
@@ -63,14 +63,23 @@ namespace ccf
...
     // Used to suspend snapshot generation during public recovery
     bool snapshot_generation_enabled = true;
 
     // Indices at which a snapshot will be next generated
-    std::deque<consensus::Index> next_snapshot_indices;
+    typedef struct
+    {
+      consensus::Index idx;
+      bool forced;
+    } snapshot_indices_entry;
```
jumaffre, line 80:
```suggestion
    struct SnapshotEntries
    {
      consensus::Index idx;
      bool forced;
    };
```

The comment above will need updating too.

tests/e2e_operations.py
```diff
@@ -78,39 +98,66 @@ def test_forced_ledger_chunk(network, args):
...
+    assert next_signature - proposal.completed_seqno < args.sig_tx_interval
     return network
 
 
+@reqs.description("Forced snapshot")
+def test_forced_snapshot(network, args):
+    primary, _ = network.find_primary()
+
+    # Submit some dummy transactions
+    network.txs.issue(network, number_txs=7)
```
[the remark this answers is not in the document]
wintersteiger: The numbers themselves don't matter, but I wanted some more unrelated transaction on the ledger just in case. And it did help find problems with other tests...

tests/e2e_operations.py
```diff
@@ -54,6 +53,27 @@ def test_parse_snapshot_file(network, args):
...
 
+def find_ledger_chunk(ledger, seqno):
+    for chunk in ledger:
+        first = last = next_signature = None
+        for tx in chunk:
+            pd = tx.get_public_domain()
+            if first is None:
+                first = pd.get_seqno()
+            else:
+                last = pd.get_seqno()
```
jumaffre, line 64:
```suggestion
            last = pd.get_seqno()
```

tests/e2e_operations.py
```diff
@@ -54,6 +53,27 @@ def test_parse_snapshot_file(network, args):
     return network
 
 
+def find_ledger_chunk(ledger, seqno):
```
jumaffre, line 56:
```suggestion
def find_ledger_chunk_for_seqno(ledger, seqno):
```

tests/infra/logging_app.py
```diff
@@ -318,3 +335,24 @@ def get_receipt(self, node, idx, seqno, view, timeout=3, domain="private"):
...
+        txs = self.priv if priv else self.pub
+        txs.pop(log_id)
+
+    def request(self, log_id, priv=False, log_capture=None):
+        primary, _ = self.network.find_primary(log_capture=log_capture)
+        with primary.client(self.user) as c:
+            table = "private" if priv else "public"
+            return c.get(
+                f"/app/log/{table}?id={log_id}", headers=self._get_headers_base()
+            )
```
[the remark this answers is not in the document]
wintersteiger: Not possible without further changes - I need the returned failure message to check that's indeed a failure

src/kv/committable_tx.h
```diff
@@ -23,8 +33,11 @@ namespace kv
 
     kv::TxHistory::RequestID req_id;
 
-    SerialisedEntryFlags flags = 0;
+  public:
```
jumaffre, line 36: Should this be removed?

src/kv/committable_tx.h
```diff
@@ -398,12 +422,19 @@ namespace kv
       std::string commit_evidence;
 
       // This is a signature and, if the ledger chunking flag is enabled, we
-      // want the host to create a chunk when it sees this entry.
+      // want the host to create a chunk when it sees this entry. On backups, we
+      // also trigger a ledger chunk when the primary creates a snapshot (which
+      // also triggers a new ledger chunk).
```
jumaffre, line 427: I think the comment on backups applies to the primary node too. In fact, it's probably to leave the details about primary/backup nodes out of the KV store implementation.

src/node/snapshotter.h
```diff
@@ -248,14 +258,23 @@ namespace ccf
...
+
+      bool force = store->flag_enabled(
+        kv::AbstractStore::Flag::SNAPSHOT_AT_NEXT_SIGNATURE);
 
-      if ((idx - next_snapshot_indices.back()) >= snapshot_tx_interval)
+      if (force)
       {
-        next_snapshot_indices.push_back(idx);
+        store->unset_flag(kv::AbstractStore::Flag::SNAPSHOT_AT_NEXT_SIGNATURE);
+        schedule_snapshot(idx);
```
jumaffre, line 272: Scheduling the snapshot at this point is too early since the signature may still be rolled back. This is why we only create snapshots when the snapshotted index cannot be rolled back (i.e. on commit).

wintersteiger: So that would be in `record_signature` then?

src/node/snapshotter.h
```diff
@@ -238,7 +248,7 @@ namespace ccf
       last_snapshot_idx = idx;
 
       next_snapshot_indices.clear();
-      next_snapshot_indices.push_back(last_snapshot_idx);
+      next_snapshot_indices.push_back({last_snapshot_idx, false});
     }
 
     bool record_committable(consensus::Index idx)
```
jumaffre, line 254: The force snapshot flow should also be tested by the snapshotter unit test (`snapshotter.cpp`).

src/node/snapshotter.h
jumaffre: I believe what you had previously should work with some minor tweaks:
1. Primary snapshotter pushes `{idx, true}` in `next_snapshot_indices` when a snapshot is forced (`record_committable()` doesn't even need to return `true` as the ledger chunk will be forced by the entry header anyway).
2. On `commit(commit_idx)`, if `commit_idx >= next_snapshot_indices.back().idx`, then a snapshot can be scheduled because it's forced (even though the `snapshot_tx_interval` hasn't passed).
3. The backup node doesn't need to know a snapshot was created at this index, since the ledger chunk will have been automatically created because of the ledger entry header.

In other words, we decouple the force snapshot generation from the regular interval generation, which is fine.

src/kv/committable_tx.h
jumaffre: I meant `public:`. It seems that these members are only accessed inside `CommittableTx`?

src/consensus/aft/raft_types.h
```diff
@@ -93,6 +95,25 @@ namespace aft
...
+    }
+
+    void unset_flag(kv::AbstractStore::Flag f) override
+    {
+      auto p = x.lock();
+      if (p)
+      {
+        return p->unset_flag(f);
+      }
+    }
```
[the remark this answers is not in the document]
wintersteiger: The `Adaptor` derives from `Store`, so it needs an implementation of the pure virtuals. I think it's better to define them instead of providing a dummy/empty implementation in the base class. I have no idea what kind of concept the `Adaptor` is supposed to adapt though, so I really can't tell whether these functions are out of place here.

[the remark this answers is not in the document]
jumaffre: `aft::Store` (not `kv::Store`) is defined a few lines above the `Adaptor`. It is the limited interface from Raft to the `kv::Store`. Since Raft doesn't make used of these functions, I believe you don't need to define them in `aft::Store` at all?

src/node/snapshotter.h
wintersteiger: Isn't that exactly what I had before? The problem was that the snapshot interval on the backups would get out of sync. I forget where exactly that was, but even after fixing the `last_snapshot_idx` to not be updated for forced snapshots, it still got out of sync.

src/node/snapshotter.h
wintersteiger: Done!

src/node/snapshotter.h
wintersteiger: OK, I think I got it now. It took some more precision for the cases in which snapshot interval and forced index coincide. Waiting for the CI now and if it passes this is ready to be merged.

src/kv/committable_tx.h
```diff
@@ -23,8 +33,10 @@ namespace kv
 
     kv::TxHistory::RequestID req_id;
 
-    SerialisedEntryFlags flags = 0;
+    TxFlags flags = 0;
+    SerialisedEntryFlags entry_flags = 0;
 
+  protected:
```
jumaffre, line 39:
```suggestion
```

src/node/snapshotter.h
```diff
@@ -248,16 +258,34 @@ namespace ccf
...
-        next_snapshot_indices.back());
+        next_snapshot_indices.back().idx);
+
+      bool forced = store->flag_enabled(
+        kv::AbstractStore::Flag::SNAPSHOT_AT_NEXT_SIGNATURE);
+
+      consensus::Index last_unforced_idx = last_snapshot_idx;
+      for (size_t i = next_snapshot_indices.size() - 1;
+           i < next_snapshot_indices.size();
+           i--)
```
jumaffre, line 272: Could we make use of reverse iterators here instead of accessing `next_snapshot_indices` by index?

wintersteiger: Sure, np.

src/node/snapshotter.h
```diff
@@ -248,16 +258,34 @@ namespace ccf
...
         idx,
-        next_snapshot_indices.back());
+        next_snapshot_indices.back().idx);
+
+      bool forced = store->flag_enabled(
+        kv::AbstractStore::Flag::SNAPSHOT_AT_NEXT_SIGNATURE);
+
+      consensus::Index last_unforced_idx = last_snapshot_idx;
+      for (size_t i = next_snapshot_indices.size() - 1;
+           i < next_snapshot_indices.size();
```
jumaffre, line 271: This condition seems weird if we iterate through the list backwards, no?

wintersteiger: Yeah, but `!= SIZE_MAX` is equally weird...

src/node/snapshotter.h
```diff
@@ -248,16 +258,34 @@ namespace ccf
...
+      }
 
-      if ((idx - next_snapshot_indices.back()) >= snapshot_tx_interval)
+      auto due = (idx - last_unforced_idx) >= snapshot_tx_interval;
+      if (due || forced)
       {
-        next_snapshot_indices.push_back(idx);
-        LOG_TRACE_FMT("Recorded {} as snapshot index", idx);
-        return true;
+        next_snapshot_indices.push_back({idx, !due});
```
jumaffre, line 284:
```suggestion
        next_snapshot_indices.push_back({idx, forced});
```

wintersteiger: No, that's incorrect when `due && forced`. If the two coincide, we can ignore the forced one.

PR #3547, opened 2022-02-14
Ignore historical service open in global hook

jumaffre:
Noticed this while working on #1652 

We should ignore historical service opens (i.e. from a previous, pre-recovery, service) in the global hook for the `public:ccf.gov.service.info` table.

### DIFF
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -1830,16 +1830,26 @@ namespace ccf
 
       network.tables->set_global_hook(
         network.service.get_name(),
-        network.service.wrap_commit_hook(
-          [this](kv::Version hook_version, const Service::Write& w) {
-            if (!w.has_value())
-            {
-              throw std::logic_error("Unexpected deletion in service value");
-            }
+        network.service.wrap_commit_hook([this](
+                                           kv::Version hook_version,
+                                           const Service::Write& w) {
+          if (!w.has_value())
+          {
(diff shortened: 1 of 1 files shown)


src/node/node_state.h
```diff
@@ -1830,16 +1830,26 @@ namespace ccf
...
+          }
 
-            network.identity->set_certificate(w->cert);
-            open_user_frontend();
-          }));
+          // Service open on historical service has no effect
+          auto hook_pubk_pem =
+            crypto::public_key_pem_from_cert(crypto::cert_pem_to_der(w->cert));
+          auto current_pubk_pem =
+            crypto::make_key_pair(network.identity->priv_key)->public_key_pem();
```
eddyashton, line 1845: Trying to convince myself this is safe. When is `network.identity->priv_key` set? Is it definitely set before this hook executes, to the "current" identity? Or is it possible that it happens later, or temporarily has an earlier iteration's identity?

Feels like we could log something useful at line 1848 to display progress/make it clear this has been skipped (including the version and the public cert of the old identity?)

jumaffre: `network.identity` is always initialised when `NodeState` is `create`d for start and recover nodes, or when a node has successfully joined - i.e. in all cases, before the node has executed any transactions when the consensus is active. 

I've added a log line when the hook logic is skipped for historical service open.

eddyashton (approved):

Issue #3540, opened 2022-02-11
Adding Users - Error: No signing certificate found in arguments (--signing-cert)


DavidZhongSyd:
**Describe the bug**

**First question:**
In https://microsoft.github.io/CCF/main/governance/open_network.html?highlight=add%20user

cat **set_user.json** whereas in scurl.sh, it use @**add_user.json**. Is it a typo in file name?

```
$ cat set_user.json
{
    "actions": [
        {
            "name": "set_user",
            "args": {
                "cert": "-----BEGIN CERTIFICATE-----\nMIIBs...<SNIP>...yR\n-----END CERTIFICATE-----\n"
            }
        }
    ]
}

$ scurl.sh https://<ccf-node-address>/gov/proposals --cacert service_cert --key member0_privk --cert member0_cert --data-binary @add_user.json -H "content-type: application/json"
{
    "ballot_count": 0,
    "proposal_id": "f665047e3d1eb184a7b7921944a8ab543cfff117aab5b6358dc87f9e70278253",
    "proposer_id": "2af6cb6c0af07818186f7ef7151061174c3cb74b4a4c30a04a434f0c2b00a8c0",
    "state": "Open"
}
```

**Second question:**
I experienced difficulties when add new user.
Under ~/CCF/python/utils, I use command `./keygenerator.sh --name user0 --gen-enc-key` generate the following certificate files for user0: 
```
root@3af2a582a0d9:~/CCF/python/utils# ls -l user0*
-rw-r--r-- 1 root root  648 Feb 11 07:30 user0_cert.pem
-rw------- 1 root root 1679 Feb 11 07:30 user0_enc_privk.pem
-rw-r--r-- 1 root root  451 Feb 11 07:30 user0_enc_pubk.pem
-rw------- 1 root root  359 Feb 11 07:30 user0_privk.pem

root@3af2a582a0d9:~/CCF/python/utils# cat user0_enc_pubk.pem
-----BEGIN PUBLIC KEY-----
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAvV1EjPw7Allz6hfYSBM9
zicp84rgpNGAmw++3TGPL5vaXNxeGDW+cul5jJ9CmM3Z711mB/GV2FUP1s3/3Pam
l7MxJBv+bbv5UiutMT9XPkY9LAwoEZkoa4y2sySNvA+cqceE6a4LgQ9krRLtu0Ik
[...]

PR #3548, opened 2022-02-15
Update adding_member.rst

achamayou:
See #3540


doc/governance/adding_member.rst
```diff
@@ -55,7 +55,7 @@ First, the new member should update and retrieve the latest state digest via the
 
 .. code-block:: bash
 
-    $ curl https://<ccf-node-address>/gov/ack/update_state_digest  --cacert service_cert.pem --key new_member_privk.pem --cert new_member_cert.pem
+    $ curl -X POST https://<ccf-node-address>/gov/ack/update_state_digest --cacert service_cert.pem --key new_member_privk.pem --cert new_member_cert.pem
```
eddyashton, line 58: Good spot! While here, we should also update the endpoint names above and below this to link to the OpenAPI docs (ie `:http:GET:`/gov/ack/update_state_digest` rather than ``/gov/ack/update_state_digest``).

And as a very minor thing, I'd prefer the url to be the first argument, because our `scurl` arg-parsing isn't quite as permissive as `curl`, so if this changed from `curl` to `scurl` for some reason it'd break.

```suggestion
    $ curl https://<ccf-node-address>/gov/ack/update_state_digest -X POST --cacert service_cert.pem --key new_member_privk.pem --cert new_member_cert.pem
```

eddyashton (approved):

achamayou:
@eddyashton thank you for fixing that!

PR #3552, opened 2022-02-15
Remove public `get_node_state()` API

eddyashton:
Explanation from #2428:

> We currently expose this to apps, even though it mostly contains methods explicitly for the node or gov frontends that we don't want apps to call. We should split this up into smaller interfaces, and pass each only where it is needed. This will probably mean a breaking change removing functionality from node_context.h, but this functionality should be unused. Relatedly I'm tempted to rewrite AbstractNodeContext as a more generic interface container/service discovery point, which would make it clearer that we can (and should!) shard the systems we register with it into their minimal atoms.

What this means in practice is that instead of giving a `ccf::AbstractNodeState&` to all frontends (including methods that apps definitely shouldn't be calling!), we pass some smaller, more-focussed interfaces to the node and gov frontends to do the things they need. This involves some extremely hard naming, suggestions welcome.

The actual implementation of these new interfaces is currently, for minimal impact, just redirecting to the existing `NodeState` monolith, but hopefully the separation of interfaces is an entry point to how this _could_ be split up.

Opening as a draft for now, because I'd like to take a run at the last part of that: rather than a mix of access/ownership patterns, and increasingly messy constructors for the node and gov frontends, we could replace (without breaking changes!) the existing `AbstractNodeContext` with a more generic container/access point for subsystems, some private and some public. This isn't a necessary part of this change, but very related - if it's small enough to be glued on the end, I'll include it in this PR.

_EDIT - This last bit is now done. It may be standalone readable in this commit: https://github.com/microsoft/CCF/pull/3552/commits/b4aaba9e9fe074a8b25f2a21bfcad66bb989beb0. This approach has major benefits in decoupling these subsystems, and making it trivial to make them public/private as required. For instance, it lets us also make the entire `LFSAccess` subsystem private. It is used by apps, if they use the Bucketed indexing strategy, but we don't really want to give them direct access. Since the fact it's used is entirely private (in the private implementation of that strategy), we can make the entire subsystem (even its abstract interface type) private and unavailable to apps._

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-ping
+No no he's not dead, he's, he's resting.
--- a/src/host/tcp.h
+++ b/src/host/tcp.h
@@ -608,6 +608,12 @@ namespace asynchost
       {
         static_cast<TCPImpl*>(req->data)->on_resolved(req, rc);
       }
+      else
+      {
+        // The TCPImpl that submitted the request has been destroyed, but we
+        // need to clean up the request object.
+        delete req;
+      }
     }
 
     void on_resolved(uv_getaddrinfo_t* req, int rc)
--- a/tests/infra/network.py
+++ b/tests/infra/network.py
@@ -134,6 +134,7 @@ class Network:
         if existing_network is None:
             self.consortium = None
             self.users = []
+            self.hosts = hosts
             self.next_node_id = 0
             self.txs = txs
             self.jwt_issuer = jwt_issuer
@@ -143,10 +144,10 @@ class Network:
             self.next_node_id = existing_network.next_node_id
             self.txs = existing_network.txs
             self.jwt_issuer = existing_network.jwt_issuer
+            self.hosts = [node.host for node in existing_network.nodes]
 
         self.ignoring_shutdown_errors = False
         self.nodes = []
-        self.hosts = hosts
         self.status = ServiceStatus.CLOSED
         self.binary_dir = binary_dir
         self.library_dir = library_dir
@@ -176,7 +177,7 @@ class Network:
         except FileNotFoundError:
             pass
 
-        for host in hosts:
+        for host in self.hosts:
             self.create_node(host, version=self.version)
 
     def _get_next_local_node_id(self):


include/ccf/node_context.h
```diff
@@ -3,19 +3,21 @@
 #pragma once
 
 #include "ccf/historical_queries_interface.h"
+#include "ccf/host_processes_interface.h"
 #include "ccf/indexing/indexer_interface.h"
 #include "ccf/indexing/lfs_interface.h"
-#include "node/rpc/node_interface.h"
```
eddyashton, line 8: Here's the goal - 406 additions and 130 deletions for the privilege of deleting _exactly this line_.

include/ccf/node_context.h
```diff
@@ -3,19 +3,21 @@
...
   {
     virtual ~AbstractNodeContext() = default;
 
+    virtual ccf::NodeId get_node_id() const = 0;
     virtual ccf::historical::AbstractStateCache& get_historical_state() = 0;
-    virtual ccf::AbstractNodeState& get_node_state() = 0;
     virtual ccf::indexing::IndexingStrategies& get_indexing_strategies() = 0;
     virtual ccf::indexing::AbstractLFSAccess& get_lfs_access() = 0;
+
+    virtual ccf::AbstractHostProcesses* get_host_processes() = 0;
```
eddyashton, line 21: Since we want to be able to launch host processes from JS generic, we currently need to make this a public thing that _any_ app could do. My generic-container solution would avoid this - JS generic is privately implemented, so may make use of framework-private subsystems that aren't exposed to any other app. I think that's an interesting win, though not something we should do _too_ often.

src/enclave/enclave.h
```diff
@@ -146,29 +153,33 @@ namespace enclave
...
+      context->host_processes = std::make_unique<ccf::HostProcesses>(*node);
 
       LOG_TRACE_FMT("Creating RPC actors / ffi");
+      node_operation = std::make_unique<ccf::NodeOperation>(*node);
+      gov_effects = std::make_unique<ccf::GovernanceEffects>(*node);
+
       rpc_map->register_frontend<ccf::ActorsType::members>(
         std::make_unique<ccf::MemberRpcFrontend>(
-          network, *context, share_manager));
+          network, *context, share_manager, *gov_effects, *node_operation));
```
eddyashton, line 175: This section highlights the core functional change here. `MemberRpcFrontend` used to take an `AbstractNodeContext`, on which it could call things like `context.get_node_state().is_reading_private_ledger()` and `context.get_node_state().rekey_ledger()`. But that same `AbstractNodeContext` was also exposed to app frontends, for the few bits of functionality they also needed (in practice I think that's just "getting the current node's ID"). This means apps could call those functions too, where the former is _maybe_ safe and the latter is _definitely not!_ So we remove that functionality from the shared `AbstractNodeContext`, which no longer has a `get_node_state()`, and instead move those methods to subsystem-grouped interfaces like `GovernanceEffects` or `NodeOperation`, and pass those only to the frontends under our control.

eddyashton: And an addendum to this with recent changes: In the initial version of this PR we had to add something like `*gov_effects, *node_operation` to every constructor of a `MemberRpcFrontend` (including tests that probably don't use them at all!). Now those are _maybe_ available from the existing `context` argument - the constructor doesn't bloat, the dependency on this subsystem can be moved closer to where it's actually needed rather than being widely intrusive.

achamayou (approved):

PR #3555, opened 2022-02-15
Fix check order in enclave_create_node

achamayou:


eddyashton (approved):
For clarity: Looks like this has only been broken for a short time. The check correctly preceded use until the cast was moved earlier in #3413. So this has never been broken in a 1.x release.

PR #3557, opened 2022-02-15
Fix problems with static log IDs in tests.

wintersteiger:
Fixes a number of problems with static log IDs in tests. This surfaced while working on #3544, which pushed the number of transactions past 40 and after the re-enabling of TX verification in #3530, which now checks the presence and contents of transactions. This fails in various ways when transactions are deleted or overwritten in random order in the `full_test_suite`.


achamayou (approved):
I'll look at namespacing shortly.

PR #3561, opened 2022-02-16
Convert `build_bundle.js` to use CommonJS

eddyashton:
The `import` syntax was only recently supported in `node` (v12), so does not work in the node that is available in our Docker containers (v10). To avoid compile errors, I've rewritten it to use the older `require` syntax, so it should work across more `node` versions.

### DIFF
--- a/tests/npm-app/build_bundle.js
+++ b/tests/npm-app/build_bundle.js
@@ -1,15 +1,15 @@
-import { readdirSync, statSync, readFileSync, writeFileSync } from "fs";
-import { join } from "path";
+const fs = require("fs");
+const path = require("path");
 
 const args = process.argv.slice(2);
 
 const getAllFiles = function (dirPath, arrayOfFiles) {
   arrayOfFiles = arrayOfFiles || [];
 
-  const files = readdirSync(dirPath);
+  const files = fs.readdirSync(dirPath);
   for (const file of files) {
-    const filePath = join(dirPath, file);
-    if (statSync(filePath).isDirectory()) {
+    const filePath = path.join(dirPath, file);
+    if (fs.statSync(filePath).isDirectory()) {
       arrayOfFiles = getAllFiles(filePath, arrayOfFiles);
     } else {
       arrayOfFiles.push(filePath);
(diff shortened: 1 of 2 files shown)


jumaffre:
Would it be better to add a later version of node to our `ccf-app-ci` container image instead? Just like we do for `ccf-ci`. If we're expecting developers to build JS apps using the `ccf-app-ci` image, I guess we should use the latest type of includes there too?

eddyashton:
> Would it be better to add a later version of node to our `ccf-app-ci` container image instead? Just like we do for `ccf-ci`. If we're expecting developers to build JS apps using the `ccf-app-ci` image, I guess we should use the latest type of includes there too?

I think we should do that as well/separately, but I think this fix is independently useful - even if people aren't using our Docker image, this tiny util script should be useable in most npm environments.

jumaffre:
> > Would it be better to add a later version of node to our `ccf-app-ci` container image instead? Just like we do for `ccf-ci`. If we're expecting developers to build JS apps using the `ccf-app-ci` image, I guess we should use the latest type of includes there too?
> 
> I think we should do that as well/separately, but I think this fix is independently useful - even if people aren't using our Docker image, this tiny util script should be useable in most npm environments.

I agree for the util script. I'm also thinking of the consequences of setting `"type": "commonjs"` in `package.json`. The sample there is TypeScript anyway, so maybe that's fine?

eddyashton:
> > > Would it be better to add a later version of node to our `ccf-app-ci` container image instead? Just like we do for `ccf-ci`. If we're expecting developers to build JS apps using the `ccf-app-ci` image, I guess we should use the latest type of includes there too?
> > 
> > 
> > I think we should do that as well/separately, but I think this fix is independently useful - even if people aren't using our Docker image, this tiny util script should be useable in most npm environments.
> 
> I agree for the util script. I'm also thinking of the consequences of setting `"type": "commonjs"` in `package.json`. The sample there is TypeScript anyway, so maybe that's fine?

Yeah I've reverted this change and replaced it with a version dependency on node >=14. I think trying to support v10 is a waste of time (this was only adding support for a single util script, the remainder of our samples still had cascading errors). There's still the question of whether we should be installing 14 in our `app-ci` images, but I think not - at the moment it's used for CI of C++ apps, which don't want an unnecessary node dependency. If we think `app-dev` includes JS `app-dev`, then we should install 'our' version of node there.

jumaffre (approved):

PR #3563, opened 2022-02-16
Create temporary ledger chunks while recovery is in progress

### LINKED ISSUE #1652: Failed public recovery entries will block further recoveries

jumaffre:
Resolves #1652 

This PR introduces temporary the `.recovery` suffix for ledger chunks created while the recovery is in progress. These `.recovery` files are ignored by subsequent nodes on start-up, which means that a recovery can now be automatically performed after a previous attempt failed. 

This done by:
- Marking new ledger chunks as `.recovery` from the point the historical ledger is truncated (i.e. start of recovery) until the last recovery share is submitted. 
- These `.recovery.` can of course be committed while the recovery is in progress, i.e. `ledger_x-y.committed.recovery`.
- When the recovery is complete, the `.recovery` suffix is removed from all recovery ledger chunks.
- Nodes ignore or delete `.recovery` ledger files on startup. In other words, they ignore previously failed recovery attempts and re-initiate a recovery as if the failed recovery never happened.

How is this tested?
- Unit test in `ledger.cpp`
- A new end-to-end test that checks that a service shutdown before the last recovery share is submitted can be recovered by a new service

TODO:
- [x] Further ledger unit tests
- [x] Add test to suite
- [x] Documentation
- [x] Inline comments in PR
- [x] Fix `ledger_open` call without argument
- [x] Service open as well as recovery


python/ccf/ledger.py
```diff
@@ -830,8 +833,17 @@ def __init__(
         ledger_files: List[str] = []
         for directory in directories:
             for path in os.listdir(directory):
-                if committed_only and not path.endswith(COMMITTED_FILE_SUFFIX):
+                sanitised_path = path
+                if path.endswith(RECOVERY_FILE_SUFFIX):
```
wintersteiger, line 837: `path.removesuffix(RECOVERY_FILE_SUFFIX)`?

src/host/ledger.h
```diff
@@ -871,16 +931,73 @@ namespace asynchost
...
+      for (auto& f : files)
+      {
+        if (f->is_recovery())
+        {
+          f->open();
+        }
+      }
+
+      // Then, open committed files
+      for (auto const& f : fs::directory_iterator(ledger_dir))
```
wintersteiger, line 961: The files in this loop aren't necessarily seen in order of seqnos, so if something throws we may end up with a gap in the committed chunks. I don't know whether that would be handled correctly elsewhere though.

src/host/ledger.h
```diff
@@ -947,7 +1064,12 @@ namespace asynchost
 
       if (require_new_file)
       {
-        files.push_back(std::make_shared<LedgerFile>(ledger_dir, last_idx + 1));
+        size_t start_idx = last_idx + 1;
+        bool is_recovery = recovery_start_idx.has_value() &&
+          start_idx > recovery_start_idx.value();
```
wintersteiger, line 1069: `.value_or(0)`?

src/host/test/ledger.cpp
```diff
@@ -92,12 +92,16 @@ size_t number_of_files_in_ledger_dir()
...
+size_t number_of_committed_files_in_ledger_dir(bool allow_recovery = false)
 {
   size_t committed_file_count = 0;
   for (auto const& f : fs::directory_iterator(ledger_dir))
   {
-    if (is_ledger_file_committed(f.path().string()))
+    auto file_name = f.path().string();
+    if (
+      (allow_recovery && is_ledger_file_name_recovery(file_name) &&
+       file_name.find(ledger_committed_suffix) != std::string::npos) ||
```
wintersteiger, line 103: `.ends_with`?

tests/recovery.py
```diff
@@ -159,26 +238,25 @@ def run(args):
...
             # with and without snapshots
-            if i % 2 == 0:
+            if i % recoveries_count == 0:
                 if args.consensus != "BFT":
-                    recovered_network = test_share_resilience(
-                        network, args, from_snapshot=True
-                    )
-                else:
-                    recovered_network = network
+                    network = test_share_resilience(network, args, from_snapshot=True)
```
wintersteiger, line 251: AFAICT the BFT branch isn't actually executed in our tests, is it?

python/ccf/ledger.py
jumaffre: I wrote it like that at first but unfortunately `removesuffix` was introduced in Python 3.9.0 and we're still on 3.8.

src/host/ledger.h
jumaffre: This is setting the `is_recovery` boolean to true if `recovery_start_idx` is set and the value of `start_idx` is greater than `recovery_start_idx.value()`. I think what you're proposing is setting `is_recovery` to the value of `start_idx`, which isn't the same thing. What am I missing?

src/host/test/ledger.cpp
jumaffre: I tried this at first but that doesn't work because a committed and recovery ledger file will still end in `.recovery`, e.g. `ledger_1-2.committed.recovery`.

tests/recovery.py
jumaffre: I don't think it currently is for this test. We have this check sprinkled in a few places and should probably clean it up at some point but I'll leave that to a separate PR.

src/host/ledger.h
wintersteiger: No, that's not what I meant - I'm just proposing the shorter `bool is_recovery = start_idx > recovery_start_idx.value_or(start_idx);`

src/host/test/ledger.cpp
wintersteiger: Ooh, I see, they could have both suffixes, good point!

python/ccf/ledger.py
wintersteiger: Darn; I'm usually a version behind, not ahead :-)

src/host/ledger.h
jumaffre: Good point. As discussed, we now keep recovery committed files in `files` so they can be open in a stable order when the recovery is complete.

src/host/ledger.h
```diff
@@ -132,14 +142,23 @@ namespace asynchost
     bool completed = false;
     bool committed = false;
 
+    bool recovery = false;
+
   public:
     // Used when creating a new (empty) ledger file
-    LedgerFile(const std::string& dir, size_t start_idx) :
+    LedgerFile(const fs::path& dir, size_t start_idx, bool recovery = false) :
```
jumaffre, line 149: A `LedgerFile` can now be marked as `recovery` when created (this is done by the higher-level `Ledger` object). The file name will automatically be suffixed with `.recovery` (even when the file is committed, i.e. `ledger_x-y.committed.recovery`) and ignored/deleted by other nodes on startup. Recovery ledger files are automatically renamed (i.e. `.recovery` suffix dropped) once the service is open.

src/host/ledger.h
```diff
@@ -871,16 +933,61 @@ namespace asynchost
...
+      // recovery ledger chunks are renamed as they can now be recovered.
+      // Note: this operation cannot be rolled back.
+      LOG_INFO_FMT("Ledger complete recovery");
+
+      for (auto it = files.begin(); it != files.end();)
+      {
+        auto& f = *it;
+        if (f->is_recovery())
+        {
+          f->open();
```
wintersteiger, line 963: Super-corner-case: If there are multiple `.recovery` files in `files` and some renamings fail, we may still end up with a gap in the ledger (`open()` ignores errors). This could lead to gaps, or to missing suffixes, e.g. only the first chunk is renamed, but subsequent ones aren't. Perhaps make `open` check for errors and abort the entire recovery if one of the renamings fails?

wintersteiger (approved):

src/host/ledger.h
achamayou: We should not abort the recovery because one of the nodes is failing its file rename. All our file operations are best effort (host is under control of the attacker) and we hope that at least f+1 nodes are able to perform them correctly.

achamayou (approved):

PR #3566, opened 2022-02-17
Infra split network start and open

jumaffre:
Very minor change to the Python infra: rename `network.start_and_join` to `network.start_and_open` and split to two separate functions: `start` that deals with starting nodes (i.e. operations) and `open` to open the service to users (i.e. governance).

### DIFF
--- a/tests/infra/network.py
+++ b/tests/infra/network.py
@@ -380,7 +380,7 @@ class Network:
             infra.proc.ccall(*cmd).returncode == 0
         ), f"Could not symlink {self.KEY_GEN} to {self.common_dir}"
 
-    def start_and_join(self, args, **kwargs):
+    def start(self, args, **kwargs):
         """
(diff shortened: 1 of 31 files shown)


tests/infra/network.py
```diff
@@ -507,6 +513,7 @@ def recover(self, args):
         Recovers a CCF network previously started in recovery mode.
         :param args: command line arguments to configure the CCF nodes.
         """
+        self.consortium.activate(self.find_random_node())
```
wintersteiger, line 516: Seems like an unrelated change to me?

wintersteiger (approved):

tests/infra/network.py
jumaffre: Good spot. Yes, this slipped through. I'd like to leave this in though as we didn't test this before: "ack"tivating members during the recovery process.

PR #3568, opened 2022-02-17
Fix docs build

eddyashton:
Mistake in #3564, which tries to copy a folder which doesn't exist in old versions/branches.


doc/conf.py
```diff
@@ -411,7 +411,8 @@ def setup(app):
         subprocess.run(["doxygen"], cwd=root_dir, check=True)
         doxygen_html_src = str(root_dir / "doxygen/html")
         doxygen_html_dest = str(out_dir / "doxygen")
-        subprocess.run(["cp", "-r", doxygen_html_src, doxygen_html_dest], check=True)
+        # check=False => it's fine if the source dir doesn't exist
+        subprocess.run(["cp", "-r", doxygen_html_src, doxygen_html_dest], check=False)
```
achamayou, line 415: That doesn't seem ideal because we'll also ignore errors like running out of disk or the dest doesn't exist? Checking os.exists() && os.isdir() would be a little better imo.

eddyashton: Sure. I've added a check that the path exists and is a dir, and restored `check=True`.

jumaffre (approved):
FYI, if you wanted to test that this works before merging, you could temporarily comment out the condition for the `Sphinx Multi-version` step in `.azure-pipelines-gh-pages.yml`.

eddyashton:
> FYI, if you wanted to test that this works before merging, you could temporarily comment out the condition for the `Sphinx Multi-version` step in `.azure-pipelines-gh-pages.yml`.

I've just been running `sphinx-multiversion` locally to repro the failure/confirm the fix. I think it should have roughly the same failure cases as the CI job? And both may be slightly different than the first post-merge run, because they all build `main` rather than the local branch, which means they only test building a branch with all the new behaviour after the merge has completed.

PR #3572, opened 2022-02-18
Update GcmHeader types

eddyashton:
Side effect of #3569 was introducing a new public-header-includes-private chain from `symmetric_key.h`. This was because `GcmHeader` was templated on the `IV_SIZE`, so the implementation details need to be public. I've refactored that so `GcmHeader` is no longer templated (but still fixed-size by construction), and derived a supporting type for the users that still want statically known IV sizes.

Additionally I've simplified this public interface a little, moving the Tx-encryption specifics (interpreting the IV as a seqno + term + snapshot-bit) out of the public definition, and closer to where it's used.

### DIFF
--- a/src/node/encryptor.h
+++ b/src/node/encryptor.h
@@ -8,6 +8,45 @@
 
 namespace ccf
 {
-  using NodeEncryptor =
-    kv::TxEncryptor<ccf::LedgerSecrets, crypto::GcmHeader<crypto::GCM_SIZE_IV>>;
+  struct TxGcmHeader : public crypto::GcmHeader<crypto::GCM_SIZE_IV>
+  {
+    using Base = crypto::GcmHeader<crypto::GCM_SIZE_IV>;
+    using Base::Base;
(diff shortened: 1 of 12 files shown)


src/node/encryptor.h
```diff
@@ -8,6 +8,44 @@
...
+    uint64_t get_term() const
+    {
+      return *reinterpret_cast<const uint32_t*>(iv.data() + IV_DELIMITER);
+    }
+
+    void set_iv_is_snapshot()
+    {
+      // Set very last bit in IV
+      iv.back() |= (1 << ((sizeof(uint8_t) * 8) - 1));
+    }
```
eddyashton, line 47: NB: this is a different signature than the previous `void set_iv_snapshot(bool is_snapshot)`. I noticed the implementation wouldn't actually unset the bit if `set_iv_snapshot(false)` was ever called. Since that's not needed, I changed the signature to avoid future confusion.

PR #3575, opened 2022-02-18
Remove CBuffer/Buffer in favour of std::span

### LINKED ISSUE #1198: Replace CBuffer with span<uint8_t>
We could use https://github.com/martinmoene/span-lite to tide us over until https://en.cppreference.com/w/cpp/container/span

achamayou:
fix #1198

This is substantially uglier than it could be because Open Enclave ships a relatively old libcxx. A number of explicit constructor calls and static casts are therefore necessary, but can go away once we upgrade.


src/ds/rb_map.h
```diff
@@ -300,15 +300,15 @@ namespace rb
       return map.get_serialized_size();
     }
 
-    CBuffer& get_serialized_buffer()
+    std::span<const uint8_t>& get_serialized_buffer()
     {
       return serialized_buffer;
     }
```
eddyashton, line 306: I'm confused by this API, both here and in `champ_map.h`, but it looks like they're unused. Can you remove `get_serialized_buffer()` and the `serialized_buffer` field from both?

And then, although unrelated to this PR, in a "while we're here": it seems we're relying on the fact that the `data` argument passed to `serialize()` has a (non-visible) size at least as large as `get_serialized_size()`. That's weird and dangerous. `serialize` should take a `std::span<uint8_t>` to write into, confirm that the size is large enough before it does anything, and return the size written. And in practice, it looks like all the callers would be happy with `std::vector<uint8_t> serialize()`, so we should provide that method instead/as well.

src/tls/ca.h
```diff
@@ -18,11 +17,11 @@ namespace tls
     Unique_X509 ca;
 
   public:
-    CA(CBuffer ca_ = nullb)
+    CA(std::span<const uint8_t> ca_ = {})
```
eddyashton, line 20: Since all the callers have this as a c-string or similar, can we move the awkward cast inside this constructor (eg, make this take a `std::span<const char> ca_`, or `char const* ca_, size_t ca_len`)?

achamayou: There's a test one that doesn't, easily changed, but also the config, so it's not quite as nice as that.

src/ds/rb_map.h
achamayou: Sure, I'll rip out the stuff that's not needed. This was done quite mechanically, so I'd missed it, thanks for pointing it out.

Agreed about this weirdness, will fix separately.

src/node/channels.h
```diff
@@ -236,8 +239,8 @@ namespace ccf
         return false;
       }
 
-      auto ret =
-        recv_key->decrypt(header.get_iv(), header.tag, cipher, aad, plain.p);
+      auto ret = recv_key->decrypt(
```
achamayou, line 242: Passing a single pointer for plain.data() and having the size be implicit is dodgy, and having a single call do both is not great either. We should split it and make the size explicit.

src/crypto/openssl/rsa_key_pair.cpp
```diff
@@ -29,10 +29,11 @@ namespace crypto
     RSAPublicKey_OpenSSL(std::move(k))
   {}
 
-  RSAKeyPair_OpenSSL::RSAKeyPair_OpenSSL(const Pem& pem, CBuffer pw)
+  RSAKeyPair_OpenSSL::RSAKeyPair_OpenSSL(
+    const Pem& pem, std::span<const uint8_t> pw)
   {
     Unique_BIO mem(pem);
-    key = PEM_read_bio_PrivateKey(mem, NULL, NULL, (void*)pw.p);
+    key = PEM_read_bio_PrivateKey(mem, NULL, NULL, (void*)pw.data());
```
achamayou, line 36: Does this rely on data() being null-terminated? We really should make use of size() here.

eddyashton (approved):

PR #3576, opened 2022-02-21
Infra: start same number of nodes when recovering network

jumaffre:
Minor change to the Python infra that's worth its own PR.

On recovery, we now start the same number of nodes that made up the original service when it was stopped, rather than the number of nodes the original service started with.


wintersteiger (approved):
Does this perhaps mask bugs that depend on the number of nodes being different at recovery time? (Not sure that's important, but it feels like we would want to test that too.)

jumaffre:
> Does this perhaps mask bugs that depend on the number of nodes being different at recovery time? (Not sure that's important, but it feels like we would want to test that too.)

I think that if anything, it should test more edge cases in the test suites. Before, the "entropy" of the network (number of nodes, possibly in different state) created by the suite would be reset by a recovery test, whereas now we carry it over to the new service.

wintersteiger:
> I think that if anything, it should test more edge cases in the test suites. Before, the "entropy" of the network (number of nodes, possibly in different state) created by the suite would be reset by a recovery test, whereas now we carry it over to the new service.

Oh yes, that's right. I was just thinking about the special case of bugs that depend on the numbers being different, but there are probably much fewer of those than what the additional "entropy" catches.

tests/infra/network.py
```diff
@@ -143,10 +144,12 @@ def __init__(
             self.next_node_id = existing_network.next_node_id
             self.txs = existing_network.txs
             self.jwt_issuer = existing_network.jwt_issuer
+            self.hosts = []
+            for n in existing_network.nodes:
+                self.hosts.append(n.host)
```
achamayou, line 149:
```suggestion
            self.hosts = [node.host for node in existing_network.nodes]
```

achamayou (approved):

PR #3578, opened 2022-02-21
Separate public KV headers from implementation details

eddyashton:
Part of #2428.

This pulls the parts of the KV which should be visible to apps (abstract `Tx`s, `Map`s, `MapHandle`s etc) into public headers and removes their visible dependency on the implementation details (ie - the actual implementation of an `untyped::ChangeSet`, or a `kv::Store`).

There's a few different approaches for different aspects of this:
- Pulling a few useful definitions from the monolithic `kv_types.h` to their own public header (eg `ccf/kv/version.h`)
- Moving implementations to `.cpp`, built in a new `ccf_kv` library (eg `tx.cpp`, `historical_queries_adapter.cpp`)
- Implementing with a private implementation pattern to remove visibility of implementation types (eg `Tx::PrivateImpl`)

I've also moved a few definitions of existing governance tables to public includes, but paused this as I think it should go in a separate PR (some tables currently called `public:ccf.gov` perhaps shouldn't be? And if we're making any semantic changes like that, they should be auditable independently of this monster...)


src/kv/untyped_change_set.h
```diff
@@ -0,0 +1,179 @@
...
+    SnapshotChangeSet(SnapshotChangeSet&) = delete;
+
+    bool has_writes() const override
+    {
+      return true;
+    }
+  };
+}
+
+namespace map
```
eddyashton, line 101: I was surprised by the existence of this namespace, which defines some templates for calculating the size of serialized values in `champ_map.h`/`rb_map.h`. I think it should be rolled into the existing `serialized::`, and documented as a requirement for any types using those maps. But that's out-of-scope here.

src/kv/tx.cpp
```diff
@@ -0,0 +1,146 @@
...
+  {
+    auto store = pimpl->store;
+
+    all_changes.clear();
+    root_at_read_version.reset();
+    pimpl.reset();
+
+    pimpl = std::make_unique<PrivateImpl>();
+    pimpl->store = store;
+  }
```
eddyashton, line 135: I convinced myself that this needed to be a `reset()` again rather than the move assignment we'd switched to? But now I think I was wrong. I'll take another look.

src/kv/tx_pimpl.h
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ccf/tx.h"
+
+namespace kv
+{
+  struct BaseTx::PrivateImpl
```
eddyashton, line 9: Now that it's all done, it's clear that this pimpl pattern may be unnecessary; we already have an abstract `Tx`, implemented by a concrete `CommittableTx`, which could hold these fields? But that's easy enough to change to, and has the same effect of breaking the public dependency.

src/node/rpc/test/frontend_test_infra.h
```diff
@@ -127,21 +127,6 @@ auto frontend_process(
...
-{
-  const auto getter = create_request(
-    nullptr,
-    fmt::format("proposals/{}/ballots/{}", proposal_id, voter),
-    HTTP_GET);
-
-  return parse_response_body<Script>(
-    frontend_process(frontend, getter, caller));
-}
-
```
eddyashton, line 144: This was unused, and with it gone the old `Script` type is also unused. Both gone.

src/kv/untyped_change_set.h
```diff
@@ -0,0 +1,179 @@
...
+#pragma once
+
+#include "ccf/byte_vector.h"
+#include "ccf/kv/hooks.h"
+#include "ccf/kv/untyped.h"
+#include "ccf/kv/version_v.h"
+#include "kv/kv_types.h"
+
+// #include "ds/hash.h"
+// #include "kv/kv_types.h"
```
jumaffre, line 12: To remove?

src/kv/tx.cpp
eddyashton: Yeah this was unnecessary, so I've restored the default move-assignment operator (safer, as it doesn't rely on us updating this definition any time we add fields).

src/kv/tx_pimpl.h
eddyashton: Had a look at this while restoring the move-assignment operator, thinking maybe only `CommittableTx` needs to be move-assignable? But it turns out we do instantiate other types in the chain, notably `ReadOnlyTx`. This needs to have all of the same handle-management logic, so we need to implement these in `BaseTx` (or thereabouts), so a PrivateImpl makes sense.

include/ccf/kv/version_v.h
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+namespace kv
```
achamayou, line 5: That doesn't seem like it should be public?

src/ds/test/map_test.cpp
```diff
@@ -21,6 +21,14 @@ struct CollisionHash
 using K = uint64_t;
 using V = uint64_t;
 
+namespace map
+{
+  inline size_t get_size(uint64_t n)
```
achamayou, line 26: What's n?

src/ds/map_serializers.h
```diff
@@ -2,65 +2,20 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
-#include "ccf/byte_vector.h"
 #include "ccf/ccf_assert.h"
-#include "serialized.h"
+#include "ccf/ds/buffer.h"
```
achamayou, line 6:
```suggestion
```

include/ccf/kv/version_v.h
eddyashton: Good spot! The only bit of these we need was already in `version.h`, so this has returned to the implementation.

src/ds/test/map_test.cpp
eddyashton: Any `uint64_t`? These are the free functions used by champ/RB maps to work out how big their serialized values are.
```suggestion
  inline size_t get_size(uint64_t)
```

jumaffre (approved):

PR #3581, opened 2022-02-22
Scoped logging transactions in tests

wintersteiger:
@achamayou is this roughly what you imagined for scoped logging transactions or did you want separate pub/priv tables per test as well?


achamayou:
@wintersteiger sort of, but I think we should also prefix the keys to avoid collisions.

wintersteiger:
This one's taking shape, so I'm opening it up for review, although I fully expect that some of you may want changes to how things are done. 

I'm not sure about the `identity` argument passed to the `scoped_txs` decorator. In some test cases there are multiple users (e.g., `test_anonymous_caller`) and I don't know whether any one of them should work or whether we have to track multiple `LoggingTxs`, one for each. But, perhaps for a first version the single identity is good enough.

tests/e2e_logging.py
```diff
@@ -159,27 +159,20 @@ def send_bad_raw_content(content):
...
-            log_id = network.txs.find_max_log_id() + 1
-            for p in range(10, 20) if args.consensus == "CFT" else range(10, 13):
-                long_msg = "X" * (2**p)
-                check_commit(
-                    c.post("/app/log/private", {"id": log_id, "msg": long_msg}),
-                    result=True,
-                )
-                check(c.get(f"/app/log/private?id={log_id}"), result={"msg": long_msg})
-                log_id += 1
+    # TLS libraries usually have 16K intrernal buffers, so we start at
```
eddyashton, line 166:
```suggestion
    # TLS libraries usually have 16K internal buffers, so we start at
```

tests/e2e_logging.py
```diff
@@ -1347,33 +1348,33 @@ def additional_interfaces(local_node_id):
...
+    #     initial_user_count=4,
+    #     initial_member_count=2,
+    # )
+
+    # cr.add(
+    #     "common",
+    #     e2e_common_endpoints.run,
+    #     package="samples/apps/logging/liblogging",
+    #     nodes=infra.e2e_args.max_nodes(cr.args, f=0),
+    # )
```
eddyashton, line 1378: To be restored?

tests/infra/logging_app.py
```diff
@@ -336,25 +365,97 @@ def get_receipt(self, node, idx, seqno, view, timeout=3, domain="private"):
...
+                    r = c.get(
+                        f"/app/log/fresh_scope?scope={func.__name__}", headers=headers
+                    )
+                    if r.status_code == http.HTTPStatus.OK:
+                        return r.body.json()["scope"]
+                    elif r.status_code == http.HTTPStatus.NOT_FOUND:
+                        attempts -= 1
+                        time.sleep(0.1)
+
+            raise ValueError("fresh scope request failed")
```
eddyashton, line 457: Shouldn't `func.__name__` give each call a suitably unique scope name already? Why do we have to additionally call `fresh_scope`?

tests/infra/logging_app.py
```diff
@@ -336,25 +365,97 @@ def get_receipt(self, node, idx, seqno, view, timeout=3, domain="private"):
...
+                network.txs = LoggingTxs(identity, scope=scope)
+                network.txs.network = network
+                r = func(*args, **kwargs)
+            else:
+                r = func(*args, **dict(kwargs, scope=scope))
+
+            if network:
+                if verify:
+                    network.txs.verify(network=network)
+                network.txs = previous_txs
```
eddyashton, line 442: Rather than reconstructing/restoring the entire `LoggingTxs` object, can we just do something like `network.txs.scope = scope` (or push/pop from a scope stack)?

tests/infra/logging_app.py
eddyashton: If we can avoid calling `GET /fresh_scope`, and reconstructing `LoggingTxs`, then we don't need to pass `identity`, and the decorators become `@scoped_txs()` rather than `@scoped_txs("user0")`. I think that's more readable (I assumed `"user0"` was the scope prefix!), and I'd lean towards giving `identity` a default argument of `"user0"` rather than passing it repeatedly.

samples/apps/logging/logging.cpp
```diff
@@ -1486,6 +1530,41 @@ namespace loggingapp
...
+      };
+
+      make_read_only_endpoint(
+        "/log/fresh_scope",
+        HTTP_GET,
+        ccf::json_read_only_adapter(fresh_scope),
+        auth_policies)
+        .add_query_parameter<std::string>("scope")
+        .set_auto_schema<void, LoggingFreshScope::Out>()
+        .install();
```
eddyashton, line 1567: Since we don't record the fresh scope in the KV, it's possible to re-use it if we don't have a `POST` in between:

```
scope1 = GET /app/log/fresh_scope?scope=foo # scope1 == foo_1
scope2 = GET /app/log/fresh_scope?scope=foo # scope2 == foo_2

# Now these conflict, despite using fresh_scope!
POST /app/log/private?scope={scope1}
POST /app/log/private?scope={scope2}
```

I think this is a bad pattern to include in the sample app. We should either add a new table to store the last `i` assigned to a fresh scope (so `fresh_scope` includes a read and write, and reliably gives each caller a unique response), or roll it into the first `POST` request (so we find the first table with no writes, then write to it _and_ tell you in the response how to write to this in future). Alternatively, if possible it'd be preferable to make the scopes entirely client determined - the caller is responsible for finding a unique string name for their scope, and the server doesn't help order multiple callers if they're using the same name.

samples/apps/logging/logging.cpp
```diff
@@ -175,6 +178,37 @@ namespace loggingapp
...
+
+    static std::string public_records(auto& ctx)
+    {
+      return public_records(get_scope(ctx));
+    }
+
+    static std::string private_records(const std::optional<std::string>& scope)
+    {
+      return scope ? std::string(PRIVATE_RECORDS) + "-" + *scope :
+                     PRIVATE_RECORDS;
```
eddyashton, line 203:
```suggestion
      return scope.has_value() ? fmt::format("{}-{}", PRIVATE_RECORDS, *scope) :
                     PRIVATE_RECORDS;
```

and similar elsewhere. `has_value()` avoids errors if we remove the `optional` wrapper in future (becomes a compile error we need to think about, rather than silently using another conversion to bool). `fmt` construction is just to be consistent with most of our other string formatting.

eddyashton (approved):

tests/e2e_logging.py
wintersteiger: How did that make it through!?! Sorry about that :-)

tests/infra/logging_app.py
wintersteiger: The function name is not enough, e.g. the full test suite runs some tests multiple times and we want separate scopes for each of them. Post-hoc I found that we could possible do that from Python by checking that the size of the table is `0` instead of adding the `fresh_scope` endpoint. I don't have a preference for one over the other though.

For the user, I don't know whether `user0` is mostly/always the expected argument, but it was pretty much the choice for all the tests I remember changing.

tests/infra/logging_app.py
wintersteiger: Sure, no problem. Might require some changes to the TX verification code, but that shouldn't be too hard to do.

samples/apps/logging/logging.cpp
wintersteiger: Hmm. TBH, I think this is fine. A scope that ends up not being used can be re-used. If the client wants to make sure that a particular scope is reserved to it, it would have to write something to it. Like I mentioned further above, we could move this completely into Python scripts and just check for table size == 0 when we require a fresh scope (and we can keep scoped counters to get the next one immediately too).

samples/apps/logging/js/src/logging.js
```diff
@@ -15,6 +15,20 @@ function get_id_from_query(parsedQuery) {
   return ccf.strToBuf(parsedQuery.id);
 }
 
+function get_scope(parsedQuery) {
+  return parsedQuery.scope === undefined ? undefined : parsedQuery.scope;
```
achamayou, line 19: I'm super confused by this, how is that not equivalent to parsedQuery.scope?

samples/apps/logging/js/src/logging.js
```diff
@@ -15,6 +15,20 @@ function get_id_from_query(parsedQuery) {
...
+}
+
+function public_records(store, scope = undefined) {
```
achamayou, line 22: Isn't undefined the default default?

tests/infra/logging_app.py
```diff
@@ -109,7 +114,7 @@ def issue(
             f"Applying {number_txs} logging txs to node {remote_node.local_node_id}"
         )
 
-        with remote_node.client(self.user) as c:
+        with remote_node.client(user if user else self.user) as c:
```
achamayou, line 117:
```suggestion
        with remote_node.client(user or self.user) as c:
```

tests/infra/logging_app.py
```diff
@@ -336,25 +365,97 @@ def get_receipt(self, node, idx, seqno, view, timeout=3, domain="private"):
                 f"Unable to retrieve entry at TxID {view}.{seqno} (idx:{idx}) on node {node.local_node_id} after {timeout}s"
             )
 
-    def delete(self, log_id, priv=False, log_capture=None):
+    def delete(self, log_id, priv=False, log_capture=None, user=None):
         primary, _ = self.network.find_primary(log_capture=log_capture)
         check = infra.checker.Checker()
-        with primary.client(self.user) as c:
+        with primary.client(user if user else self.user) as c:
```
achamayou, line 371:
```suggestion
        with primary.client(user or self.user) as c:
```

tests/infra/logging_app.py
```diff
@@ -336,25 +365,97 @@ def get_receipt(self, node, idx, seqno, view, timeout=3, domain="private"):
...
             if priv:
                 self.priv.pop(log_id)
             else:
                 self.pub.pop(log_id)
 
-    def request(self, log_id, priv=False, log_capture=None):
+    def request(self, log_id, priv=False, log_capture=None, user=None):
         primary, _ = self.network.find_primary(log_capture=log_capture)
-        with primary.client(self.user) as c:
+        with primary.client(user if user else self.user) as c:
```
achamayou, line 384:
```suggestion
        with primary.client(user or self.user) as c:
```

tests/infra/logging_app.py
```diff
@@ -336,25 +365,97 @@ def get_receipt(self, node, idx, seqno, view, timeout=3, domain="private"):
...
+                url += "&scope=" + self.scope
             return c.get(
-                f"/app/log/{table}?id={log_id}", headers=self._get_headers_base()
+                url,
+                headers=self._get_headers_base(),
+            )
+
+    def post_raw_text(self, log_id, msg, log_capture=None, user=None):
+        primary, _ = self.network.find_primary(log_capture=log_capture)
+        with primary.client(user if user else self.user) as c:
```
achamayou, line 396:
```suggestion
        with primary.client(user or self.user) as c:
```

tests/infra/logging_app.py
```diff
@@ -336,25 +365,97 @@ def get_receipt(self, node, idx, seqno, view, timeout=3, domain="private"):
...
+            if isinstance(args[0], infra.network.Network):
+                network = args[0]
+                if hasattr(network, "txs"):
+                    previous_txs = network.txs
+                if hasattr(network, "jwt_issuer") and network.jwt_issuer is not None:
+                    headers = infra.jwt_issuer.make_bearer_header(
+                        network.jwt_issuer.issue_jwt()
+                    )
+                node = network.find_random_node()
+            elif isinstance(args[0], infra.node.Node):
```
achamayou, line 425: Can we just name the first arg network and regularise all usage?

tests/infra/logging_app.py
achamayou: +1 for removing fresh_scope and defaulting user0.

achamayou (approved):

samples/apps/logging/js/src/logging.js
wintersteiger: Possible - lack of JS knowledge, I'll look it up.

tests/infra/logging_app.py
wintersteiger: No, `check_can_progress` doesn't have a network, only a node. The use-case there is a bit different though, I might be able to work around it.

samples/apps/logging/js/src/logging.js
wintersteiger: I remember having a reason for it, but I agree, it should be totally unnecessary. I think I was just confused about undefined/null/`===`.

samples/apps/logging/js/src/logging.js
achamayou: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

PR #3586, opened 2022-02-22
Fix recovery start instructions

wintersteiger:
Lost instructions when we switched to config files.

### DIFF
--- a/doc/operations/recovery.rst
+++ b/doc/operations/recovery.rst
@@ -16,7 +16,7 @@ The recovery procedure consists of two phases:
 Establishing a Recovered Public Network
 ---------------------------------------
 
-To initiate the first phase of the recovery procedure, one or several nodes should be started with the ``Recover`` command in the `cchost` config file:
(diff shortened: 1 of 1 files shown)


doc/operations/recovery.rst
```diff
@@ -16,10 +16,16 @@ The recovery procedure consists of two phases:
...
-To initiate the first phase of the recovery procedure, one or several nodes should be started with the ``recover`` option:
+To initiate the first phase of the recovery procedure, one or several nodes should be started with the ``Recover`` command in the `cchost` config file:
 
 .. code-block:: bash
 
+    $ cat /path/to/config/file
+      ...
+      "command": {
+        "type": "Recover",
+        "start":
```
eddyashton, line 27: Good spot! One issue with this sample:

```suggestion
        "recover":
```

We duplicate the command, as both `command.type` and the name of a field containing command-specific params - see `recover_config.json` for a sample.

wintersteiger: Aha! Yes, that wasn't obvious from the actual file. I'll add the whole `recover` element and perhaps add a link to the sample config.

eddyashton (approved):

PR #3588, opened 2022-02-23
CI: add debug job for multi-thread pipeline

letmaik:


.multi-thread.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       env:
         container: sgx
         pool: 1es-dcv2-focal
-      cmake_args: "-DCOMPILE_TARGETS=sgx -DWORKER_THREADS=4 -DENABLE_2TX_RECONFIG=ON -DLONG_TESTS=ON"
```
achamayou: Enabling Debug seems good, removing the long tests less so?

letmaik: The SGX job never ran long tests, but I thought they should be enabled, that's why I initially added it. I did hit the 60min job limit though, after which I removed it again... Should I increase the limit?

letmaik:
Closing this as it's not the right approach.

Issue #1858, opened 2020-11-05
Snapshots may be larger than ring buffer maximum message size


jumaffre:
As a CCF service processes more transactions which write to different tables at different keys, snapshots may grow without bounds. Currently, when a snapshot is generated, it is written to the outbound ring buffer in one go. It might be split into fragments if it is too large but if it actually is larger than the maximum message size (`--max-msg-size` option to `cchost`), the CCF node will crash, e.g. (with `16777216` the default maximum message size):

```
2020-11-04 15:22:18.596 | ERROR    | infra.remote:log_errors:86 - Contents of /home/lyshi/repos/CCF/tests/workspace/recovery_0/err:
terminating with uncaught exception of type std::logic_error: Requested a write of 16827539 bytes, max allowed is 16777216
```

To solve this, we could just bump the default maximum message size but this just pushes the problem further. The solution to this should work for any snapshot size, however large it is. Instead, we could either:
- Split the snapshot into multiple chunks before writing it to the ring buffer.
- Build a enclave<->host large message system which lets the enclave directly write to host memory, bypassing the ring buffer altogether. This would involve allocating a sufficiently large buffer on the host to store the snapshot. It seems that this solution should only apply to snapshots as other ring buffer writers already chunk messages into smaller bits (e.g. TLS). 

**Notes:**
- This issue should not affect services with small key-value store footprint, e.g. services that write to a few tables and/or a few keys. 
- To reproduce this issue:
```bash
$ TEST_ENCLAVE=virtual python recovery.py --gov-script CCF/tests/gov.lua --use-snapshot --snapshot-tx-interval 1000 --msgs-per-recovery 1000000
```

eddyashton:
Splitting the snapshot message into multiple chunks so that it never exceeds the maximum message size is the simpler solution from an implementation perspective, but still requires the enclave thread to be blocked for a long time writing these chunks, and likely saturates the host thread that is reading those chunks for a long time. Since we cap the amount of data that the host can read each UV tick, this processing may cause a noticeable delay on any other work in the service.

[...]

PR #3603, opened 2022-02-25
Do not write snapshot out if it is too big for ring buffer

jumaffre:
Temporary workaround for https://github.com/microsoft/CCF/issues/1858 

If a snapshot is too large for the ring buffer, we now report an error and do not write it out, instead of causing a node crash.


src/ds/ring_buffer_types.h
```diff
@@ -96,6 +96,11 @@ namespace ringbuffer
 
     virtual WriteMarker write_bytes(
       const WriteMarker& marker, const uint8_t* bytes, size_t size) = 0;
+
+    virtual std::optional<size_t> get_max_message_size()
+    {
+      return std::nullopt;
+    }
```
eddyashton, line 103: I don't think we need an optional here; there's a sensible value for every implementation of `AbstractWriter`, so this could just be `virtual size_t get_max_message_size() = 0;`?

Then `ringbuffer::Writer` would override it to return `Const:max_size()` (enforced in prepare), and `NonBlockingWriter` would override it to return `underlying_writer->get_max_message_size()`.

achamayou (approved):

src/ds/ring_buffer_types.h
jumaffre: Good point, I'd missed that we checked for the maximum size in the vanilla ring buffer implementation.

PR #3604, opened 2022-02-25
Remove duplicated log messages in Raft

### LINKED ISSUE #3545: Remove spurious `Removed raft node...` message when replaying historical transactions

jumaffre:
Resolves #3545 

The `INFO` line "`Added raft node...`" used to be printed too many times in the node log. In particular, when a node was catching up, it was printed many times for the same node because the it was also guarded by a `||` check with whether the node-to-node channel was created, which isn't until the nodes have sent messages to each other. We now only print this message once when the node is added to the set of known `nodes`.


src/consensus/aft/raft.h
```diff
@@ -1250,8 +1250,6 @@ namespace aft
           continue;
         }
 
-        LOG_DEBUG_FMT("Replicating on follower {}: {}", state->my_node_id, i);
```
jumaffre, line 1253: This was printed for each entry in an AE at the start of the batch, rather than being printed when each entry is individually executed.

src/consensus/aft/raft.h
```diff
@@ -1298,16 +1296,16 @@ namespace aft
       for (auto& ae : append_entries)
       {
         auto& [ds, i] = ae;
-        state->last_idx = i;
```
jumaffre, line 1301: Simplified the logic slightly here: only bump `last_idx` when the entry is successfully applied.

src/consensus/aft/raft.h
```diff
@@ -2313,15 +2311,6 @@ namespace aft
       return state->cft_watermark_idx;
     }
 
-    const Configuration::Nodes& get_last_configuration_nodes()
```
jumaffre, line 2316: This was unused.

src/consensus/aft/raft.h
```diff
@@ -2482,21 +2471,24 @@ namespace aft
...
 
           // A new node is sent only future entries initially. If it does not
           // have prior data, it will communicate that back to the leader.
           auto index = state->last_idx + 1;
           nodes.try_emplace(node_info.first, node_info.second, index, 0);
 
-          channels->associate_node_address(
-            node_info.first, node_info.second.hostname, node_info.second.port);
-
           if (leadership_state == kv::LeadershipState::Leader)
```
jumaffre, line 2492: The duplicated log line is on line `2497`.

achamayou (approved):

PR #3607, opened 2022-02-28
Document ways to build old CCF versions in container

jumaffre:
This is useful when building old versions of CCF locally, without having to use a different dev VM altogether.

### DIFF
--- a/doc/contribute/build_ccf.rst
+++ b/doc/contribute/build_ccf.rst
@@ -83,9 +83,9 @@ Building older versions of CCF may require a different toolchain than the one us
     $ git checkout ccf-1.0.15 # e.g. building CCF 1.0.15
     $ docker run -ti --device /dev/sgx_enclave:/dev/sgx_enclave --device /dev/sgx_provision:/dev/sgx_provision -v $LOCAL_CCF_CHECKOUT_PATH:/CCF ccfciteam/ccf-ci:$CCF_CI_IMAGE_TAG bash
     # container started, following lines are in container
-    $ root@:/# cd CCF/
-    $ root@:/# mkdir build_docker && cd build_docker
-    $ root@:/# cmake -GNinja .. && ninja
+     $ cd CCF/
+     $ mkdir build_docker && cd build_docker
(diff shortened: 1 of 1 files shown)


doc/contribute/build_ccf.rst
```diff
@@ -70,17 +70,36 @@ Although CCF's unit tests can be run through ``ctest`` directly, the end-to-end
...
+
+    $ export CCF_CI_IMAGE_TAG="oe0.17.2-clang-8" # e.g. building CCF 1.0.15
+    $ export LOCAL_CCF_CHECKOUT_PATH=/path/to/local/ccf/checkout
+    $ cd $LOCAL_CCF_CHECKOUT_PATH
+    $ git checkout ccf-1.0.15 # e.g. building CCF 1.0.15
+    $ docker run -ti --device /dev/sgx_enclave:/dev/sgx_enclave --device /dev/sgx_provision:/dev/sgx_provision -v $LOCAL_CCF_CHECKOUT_PATH:/CCF ccfciteam/ccf-ci:$CCF_CI_IMAGE_TAG bash
+    # container started, following lines are in container
+    $ root@:/# cd CCF/
+    $ root@:/# mkdir build_docker && cd build_docker
+    $ root@:/# cmake -GNinja .. && ninja
```
eddyashton, line 88: I know these are probably copied verbatim, but it's confusing to mix use of `#` as comment delimiter and command prompt. It's highlighted as a comment in this GitHub diff, and probably will be within the rendered code-block.

```suggestion
    # container started, following lines are in container
     $ cd CCF/
     $ mkdir build_docker && cd build_docker
     $ cmake -GNinja .. && ninja
```

eddyashton (approved):

doc/contribute/build_ccf.rst
jumaffre: Yeah, I wasn't sure what would be clearer and went for the `/#` because it didn't actually render as a comment in Sphinx. I think the "container started..." comment above is sufficient so the `$` is clearer you're right.

achamayou (approved):

PR #3608, opened 2022-02-28
Move definition of public service tables to public headers

eddyashton:
Another part of #2428.

Moves the definition of various governance tables, and their supporting types, to public includes. Also moves the relevant serialisers nearer to their point of use, rather than being defined centrally/in a non-obvious header. The only awkward one is the blit serialiser for a `Pem`, which I don't think should be lumped in for all consumers of `Pem`, but needs to be visible to several table definitions, so gets its own header.


achamayou:
Unless I'm missing something this is user-visible and so should have a changelog entry?

achamayou (approved):

jumaffre (approved):

PR #3610, opened 2022-02-28
Remove dependency on `TxReceipt` from public API

eddyashton:
Another part of #2428.

Applications previously needed to include `node/receipt.h`, to be able to call `TxReceipt::describe()`, to produce the `ccf::Receipt` type we want them to use. I've replaced this with a free function `ccf::describe_receipt()`, so that `TxReceipt` (and `TxReceiptPtr`) can remain opaque handles, and the internal types used to store hashes, or to represent our merkle trees, do not need to be exposed.


include/ccf/historical_queries_interface.h
```diff
@@ -13,8 +13,15 @@
 
 namespace ccf
 {
+  // This is an opaque, incomplete type, but can be summarised by the functions
```
achamayou, line 16:
```suggestion
  // This is an opaque, incomplete type, but can be summarised to a JSON-serialisable form by the functions
```

achamayou (approved):

Issue #3549, opened 2022-02-15
Examine statefulness of `/gov/recovery_share` endpoint


eddyashton:
While looking at something else, I noticed that `initiate_private_recovery()`, which is currently called from `/gov/recovery_share`, is unsafe to call from within a transaction, since it makes stateful changes to the local node that will cause issues if the transaction is re-executed due to conflict. It expects a current state machine, then advances the state machine, so on first execution it will have triggered a state machine advance (and possibly other side effects), and on re-execution will throw an exception - returning an HTTP error and applying an incongruous set of writes.

Any stateful side effects should happen within a hook, not transaction execution.

Relatedly, I'd like to confirm that the `share_manager` calls from `MemberRpcFrontend` do not have side effects by removing their dependency on the `network_tables` member, making them pure static functions, and thus more obviously transactionally safe.

PR #3611, opened 2022-02-28
Recovery global hook

jumaffre:
Fix https://github.com/microsoft/CCF/issues/3549

The recovery procedure used to modify the local state of the node (e.g. initiating the private recovery) from code paths that could be rolled back (local hooks) or re-executed (`POST /gov/recovery_share` endpoint). Instead, we now rely on global hooks whenever possible: 
- On the primary node, a new global hook on the `encrypted_submitted_shares` table which, once all the recovery shares have been submitted, triggers the private recovery.
- On the backup nodes, the local hook on the `encrypted_ledger_secrets` table has been split so that the live ledger rekey is still done on a local hook, but the private recovery is now triggered from a global hook on that table.


src/node/node_state.h
```diff
@@ -1072,6 +1073,10 @@ namespace ccf
         setup_one_off_secret_hook();
         auto tx = network.tables->create_tx();
 
+        // Clear recovery shares that were submitted to initiate the recovery
+        // procedure
+        share_manager.clear_submitted_recovery_shares(tx);
```
jumaffre, line 1078: We used to clear the shares in the `POST /gov/recovery_share` endpoint, once the recovery was initiated. This wasn't ideal as it meant the encrypted recovery share was never recorded in the ledger. Instead, we now clear the submitted recovery shares once the private recovery is complete.

src/node/node_state.h
```diff
@@ -1764,7 +1751,12 @@ namespace ccf
         network.secrets.wrap_map_hook(
           [this](kv::Version hook_version, const Secrets::Write& w)
             -> kv::ConsensusHookPtr {
-            LedgerSecretsMap restored_ledger_secrets;
+            // Used to rekey the ledger on a live service
+            if (is_part_of_public_network())
```
achamayou, line 1755: It would probably make more sense to express this as !is_part_of_network()?

src/node/share_manager.h
```diff
@@ -415,7 +417,7 @@ namespace ccf
         // This seems potentially dangerous, in that it requires that RVO
```
achamayou, line 417: Comment no longer applies?

src/node/share_manager.h
```diff
@@ -425,8 +427,6 @@ namespace ccf
         latest_ls = s.first->second;
       }
 
-      recovery_ledger_secrets.clear();
```
achamayou, line 428: Why is this no longer cleared?

achamayou (approved):

PR #3612, opened 2022-02-28
More detailed startup logging

achamayou:
Addresses #3605

PR #3614, opened 2022-02-28
Log early messages from a separate thread

eddyashton:
Companion to #3612, and follow-up to #3413.

If the `create_node` ECALL blocks, we currently get very little debugging information. This adds a separate thread flushing the outbound ringbuffer while that ECALL executes, so we will get more precise `LOG_` output indicating the point where the code blocked.

### DIFF
--- a/src/host/main.cpp
+++ b/src/host/main.cpp
@@ -460,17 +460,12 @@ int main(int argc, char** argv)
     LOG_INFO_FMT("Initialising enclave: enclave_create_node");
     std::atomic<bool> ecall_completed = false;
     auto flush_outbound = [&]() {
-      while (true)
+      do
       {
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
 
         bp.read_all(circuit.read_from_inside());
-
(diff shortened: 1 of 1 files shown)


src/host/main.cpp
```diff
@@ -458,6 +458,21 @@ int main(int argc, char** argv)
     }
 
     LOG_INFO_FMT("Initialising enclave: enclave_create_node");
+    std::atomic<bool> ecall_completed = false;
+    auto flush_outbound = [&]() {
+      while (true)
```
achamayou, line 463: while (!ecall_completed) ? Always a bit nervous when I see a while(true);

eddyashton: Sure, I'll re-write this as a `do-while` (I'd prefer that to a `while`, so that we clearly try to call `read_all` once _after_ `ecall_completed` has been set. We can't perfectly sync, but that's fine (anything that gets left on the boundary will get flushed eventually). This pattern just does its best to flush as early as possible.

PR #3624, opened 2022-03-02
Add previous and next service identities to transition_service_to_open after recovery

wintersteiger:
These changes add support for previous and next service identities during recovery. The previous identity is used to verify that the snapshot a recovery node started from is endorsed by that identity and the new one is used to check that the recovered service indeed uses the expected new identity. Both are signed by members as part of the `transition_service_to_open` governance action.


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The C++ types used to define public governance tables are now exposed in public headers. Any C++ applications reading these tables should update their include paths (ie - `#include "service/tables/nodes.h"` => `#include "ccf/service/tables/nodes.h"`) (#3608).
 - `TxReceipt::describe()` has been replaced with `ccf::describe_receipt()`. Includes of the private `node/tx_receipt.h` from C++ applications should be removed (#3610).
 - Python `ccf.read_ledger` and `ccf.ledger_viz` tools now accept paths to individual ledger chunks, to avoid parsing the entire ledger.
+- The `transition_service_to_open` governance proposal now requires the service identity as an argument to ensure the correct service is started. During recovery, it further requires the previous service identity to ensure the right service is recovered (#3624).
```
achamayou, line 19: Please move to an -rc4 section, -rc3 is already out.

include/ccf/service/tables/service.h
```diff
@@ -12,14 +12,16 @@ namespace ccf
     OPENING = 1,
     OPEN = 2,
     WAITING_FOR_RECOVERY_SHARES = 3,
-    CLOSED = 4 // For now, unused
+    RECOVERING = 4,
+    CLOSED = 5 // For now, unused
```
achamayou, line 16: I think we should axe CLOSED, it's been there for years and there's still no credible plan to use it for anything. It can be added when there is.

achamayou (approved):

PR #3632, opened 2022-03-04
Distinguish public `ReadOnlyStore` from private `Store`

eddyashton:
We don't want to expose the full `kv::Store` type to apps. The only interface they need is that used in historical queries, to create read-only txs and query the current TxID. I've split this into a separate type, called `kv::ReadOnlyStore`.

I've gone back and forth on whether this simpler type gets a new name, or if _it_ gets to be called `kv::Store` (in theory, avoiding a breaking change for consumers), and the current "full-fat" implementation gets some new name. I've gone for this approach because I haven't found a sensible name for the other approach (`kv::ConcreteStore`? `kv::RealStore`? `kv::StoreImpl`? these are all rubbish), and I'm optimistic this will not require any changes in apps in practice (assuming they use auto, or references to the field in a `historical::State` which has kept its names and core methods).

I've also done a pass on the docs as part of this, initially removing references to `kv::Store` (if it's no longer a type that apps should use, I don't think it's an worthwhile concept to document for app developers - instead we refer in slightly more general terms to the "KV" as a whole). While doing this, I refreshed a bunch of the KV How-To - it was previously changed during big transitions (privacy domain becoming part of the map name, map creation becoming implicit, handles becoming first class concepts), so was worded to distinguish the new behaviour from the old. I think it's clearer to forget the old behaviour, and describe instead precisely what the current semantisc are.


achamayou:
@eddyashton this really needs a CHANGELOG entry, as it's very user visible. I also wonder to what extent some of these types would benefit from being abbreviated, eg. `ReadOnlyStorePtr` to `ROStorePtr` etc, but I suppose that's quite subjective.

achamayou (approved):

PR #3633, opened 2022-03-07
Fix memory leaks and cycles

wintersteiger:
This fixes `shared_ptr` cycles via `RPCMap` and `RPCResponder` causing a bunch of memory to be leaked. Also fixes a direct leak of the `Enclave` object during shutdown. We use `std::weak_ptr` very sparingly so I wouldn't be surprised if there are a bunch of other cycles elsewhere.

### DIFF
--- a/src/enclave/main.cpp
+++ b/src/enclave/main.cpp
@@ -217,7 +217,7 @@ extern "C"
     }
 #endif
 
-    enclave::Enclave* enclave;
+    enclave::Enclave* enclave = nullptr;
 
     try
     {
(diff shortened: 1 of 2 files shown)


src/enclave/main.cpp
```diff
@@ -249,15 +249,25 @@ extern "C"
...
+        node_cert,
+        node_cert_size,
+        node_cert_len,
+        service_cert,
+        service_cert_size,
+        service_cert_len);
+    }
+    catch (...)
+    {
+      delete enclave;
```
eddyashton, line 268: It looks like we never delete this object in the normal operation path, after we've stored it in `static std::atomic<enclave::Enclave*> e;`. Should we also be deleting it on line 304?

wintersteiger: I tried that at some point but it didn't make a difference in the memory leaks I saw. I've mostly been looking at unsuccessful enclave creation though. I will add it anyways, just to be on the safe side.

eddyashton (approved):

src/enclave/main.cpp
```diff
@@ -291,24 +301,25 @@ extern "C"
...
       }
       else
       {
-        auto s = e.load()->run_worker();
+        s = e.load()->run_worker();
         num_complete_threads.fetch_add(1);
-        return s;
       }
+      delete e.load();
+      return s;
```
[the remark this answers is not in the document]
wintersteiger: Ah! Good point, too hasty with my commit :)

PR #3634, opened 2022-03-07
Extend testing of invalid HTTP requests and other HTTP protocols

eddyashton:
Confirming the behaviour of our HTTP parser with some more malformed requests. I've found and fixed one instance where a JSON-reencode could turn into a HTTP 500 Internal Server Error, but all result in a valid HTTP response to the client.

Edit: I've also added tests of various other protocols supported by curl. Specifically explicitly requesting HTTP/1.0, HTTP/1.1, HTTP/2, HTTP/3, and WebSockets. The first 2 succeed, WebSockets and HTTP/2 (with upgrade negotiation) succeed (by ignoring the upgrade request, and falling back to HTTP/1), HTTP/2 (without negotiation, _assuming_ support) results in a parse error, and HTTP/3 is currently unsupported by curl (but we're ready for it!). This is partly about confirming our current behaviour in response to those (ie - these tests will change as we add support!), but mostly about ensuring we are _robust_ to these requests.


src/http/test/http_test.cpp
```diff
@@ -102,6 +102,32 @@ DOCTEST_TEST_CASE("Parsing error")
   DOCTEST_CHECK(sp.received.empty());
 }
 
+DOCTEST_TEST_CASE("Parsing fuzzing")
```
wintersteiger, line 105: Perhaps worth adding a test that adds/removes selected headers at random?

wintersteiger (approved):

include/ccf/rpc_context.h
```diff
@@ -116,7 +116,10 @@ namespace enclave
     {
       nlohmann::json body = ccf::ODataErrorResponse{
         ccf::ODataError{std::move(error.code), std::move(error.msg)}};
-      const auto s = body.dump();
+      // Set error_handler to replace, to avoid throwing if the error message
+      // contains non-UTF8 characters. Other args are default values
+      const auto s =
+        body.dump(-1, ' ', false, nlohmann::json::error_handler_t::replace);
```
achamayou, line 122: We ought to have a test in the logging sample where we return a non UTF-8 message, to check that this does in fact do what we think it does.

src/http/test/http_test.cpp
```diff
@@ -102,6 +102,32 @@ DOCTEST_TEST_CASE("Parsing error")
...
+{
+  std::vector<uint8_t> r;
+
+  http::SimpleRequestProcessor sp;
+  http::RequestParser p(sp);
+
+  const auto orig_req = http::build_post_request(r);
```
achamayou, line 112: We could parametrise on the verb here, since some of them do have special handling, eg. PRI.

include/ccf/rpc_context.h
eddyashton: This was tested in some of the e2e_logging variations - anything that returns a 404 Not Found rather than a 400 Bad Request includes a re-encoding of a non-UTF8 message (and before this change, converting to a 500 as this line exploded).

But I've made that more clear, with an explicit new test in `e2e_logging`:
```
    # Tests non-UTF8 encoding in OData
    send_bad_raw_content(b"POST /node/\xff HTTP/2.0\r\n\r\n")
```

src/http/test/http_test.cpp
eddyashton: This now runs for all verbs that llhttp defines.

PR #3642, opened 2022-03-09
Multiple expected curl error messages

wintersteiger:
Older versions of curl produce different expected error messages, so no we accept one of multiple.

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -251,12 +251,7 @@ def test_protocols(network, args):
         else:
             assert res.returncode != 0, res.returncode
             err = res.stderr.decode()
-            failed = False
-            for err_i in expected_errors:
-                if err_i in err:
-                    failed = True
-                    break
-            assert failed, "none of the expected error messages were found"
+            assert any(expected_error in err for expected_error in expected_errors), err
 
     # Valid transactions are still accepted
     network.txs.issue(
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -227,16 +227,22 @@ def test_protocols(network, args):
     expected_response_body = body[: body.rfind("\n")]
 
     # Test additional HTTP versions with curl
-    for (protocol, expected_error) in (
+    for (protocol, expected_errors) in (
         ("", None),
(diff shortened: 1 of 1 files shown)


tests/e2e_logging.py
```diff
@@ -245,7 +251,12 @@ def test_protocols(network, args):
         else:
             assert res.returncode != 0, res.returncode
             err = res.stderr.decode()
-            assert expected_error in err, err
+            failed = False
+            for err_i in expected_errors:
+                if err_i in err:
+                    failed = True
+                    break
+            assert failed, "none of the expected error messages were found"
```
eddyashton, line 259: This doesn't print the original `err` message if none of the expected errors are found, which will make it harder to debug. It's possible to find `expected_errors` from the callstack + inspecting this source, but not the runtime value of `err`. We should log that on failure, and can do so with a list comprehension assertion:

```suggestion
            assert any(expected_error in err for expected_error in expected_errors), err
```

tests/e2e_logging.py
```diff
@@ -227,16 +227,22 @@ def test_protocols(network, args):
...
+            "--http3",
+            [
+                "the installed libcurl version doesn't support this",
+                "option --http3: is unknown",
+            ],
+        ),
     ):
         res = infra.proc.ccall("curl", protocol, *common_options)
-        if expected_error is None:
+        if expected_errors is None:
```
achamayou, line 245: imho, [] is a reasonable way to express "no expected errors", and then this can be "if expected_errors:" but that's clearly nitpicking.

achamayou: or set()

achamayou (approved):

tests/e2e_logging.py
wintersteiger: Yeah, that's much better, thanks! 

@achamayou do you mean a set intersection non-empty check? Would that be shorter?

tests/e2e_logging.py
wintersteiger: Yeah, I think so too, just didn't want the diff to be larger than necessary, will change.

PR #3643, opened 2022-03-10
Add ALPN extension to CCF servers, advertising HTTP/1.1

eddyashton:
While looking for standard tests of protocol support, I realised that we're not currently implementing the ALPN extension as part of the TLS handshake, which can be used to negotiate the following application protocol. This is visible in the golden TLS report. I've added support for this though OpenSSL, so CCF servers now indicate during the TLS handshake that they expect HTTP/1.1 in the following traffic. This will need extending with future protocol support, but for now should return more consistent + detailed errors to clients expecting anything different.


CMakeLists.txt
```diff
@@ -661,12 +661,16 @@ if(BUILD_TESTS)
...
+      OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/testssl/testssl.sh
       COMMAND
-        test -d testssl || git clone https://github.com/drwetter/testssl.sh
+        rm -rf ${CMAKE_CURRENT_BINARY_DIR}/testssl && git clone --depth 1
+        https://github.com/drwetter/testssl.sh
         ${CMAKE_CURRENT_BINARY_DIR}/testssl
     )
+    add_custom_target(
+      testssl ALL DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/testssl/testssl.sh
+    )
```
eddyashton, line 673: The beahaviour is the same (there is a `testssl` target which clones the repo to get a local `testssl/testssl.sh` file). But with the additional custom target indirection, we only build that when we need to. Previously, whenever `TLS_TEST` was on, we'd build the `testssl` target every time, and rebuild additional targets whenever it was turned on or off.

tests/e2e_logging.py
```diff
@@ -206,51 +206,64 @@ def test_protocols(network, args):
...
-        ("--http2-prior-knowledge", ["Error in the HTTP2 framing layer"]),
-        (
-            "--http3",
-            [
+    expected_response_body, status_code, http_version = parse_result_out(res)
+    assert status_code == "200", status_code
+    assert http_version == "1.1", http_version
+
+    # Test additional protocols with curl
+    for protocol, expected_result in {
```
eddyashton, line 233: These test the same things, but now also confirm that the resulting HTTP version after negotiation is `1.1`, regardless of what we requested of curl. Additionally I swapped the "no args" and "try websockets" cases, so the former is now used to get the golden result.

wintersteiger (approved):

achamayou (approved):

PR #3647, opened 2022-03-10
Fix LTS recovery tests

wintersteiger:


tests/infra/consortium.py
```diff
@@ -389,12 +389,12 @@ def add_users_and_transition_service_to_open(self, remote_node, users):
...
-            {
-                "name": "transition_service_to_open",
-                "args": {"next_service_identity": self.get_service_identity()},
-            }
-        )
+
+        args = {}
+        if not remote_node.version_before("ccf-2.0.0-rc4"):
+            args = {"next_service_identity": self.get_service_identity()}
+        proposal["actions"].append({"name": "transition_service_to_open", **args})
```
achamayou, line 396:
```suggestion
        proposal["actions"].append({"name": "transition_service_to_open", "args": args})
```

tests/infra/node.py
```diff
@@ -617,6 +627,13 @@ def check_log_for_error_message(self, msg):
                         return True
         return False
 
+    def version_before(self, version):
+        rc = version_rc(version)
```
eddyashton, line 631:
```suggestion
        rc = version_rc(version)
        if rc is None or self.version is None:
            return False
```

tests/infra/consortium.py
wintersteiger: This was correct - in case `args` is empty I want no "args" entry at all, because the old governance checks that there are no arguments, as opposed to an empty args argument.

eddyashton: Ah, then we need to insert the `"args": ` level inside the `args` variable.

`**args` in the original code expands to `"next_service_identity": ...`, but in this object we need `"args": {"next_service_identity": ...}`

achamayou (approved):

PR #3662, opened 2022-03-15
Add free-form `node_data` field

### LINKED ISSUE #3644: Add last_written to each node in /node/network/nodes
- [ ] When listing nodes for operations purposes, it's useful to find out when they were last modified. We could add that relatively easily by tacking on the return value of get_version_of_previous_write()
- [ ] Also add a user_data, similar to user and members so operators can easily associate information such a pod name with a node in the list

eddyashton:
Resolves #3644.

`node_data` can be passed at cchost startup time*, or via a `set_node_data` proposal. The objects returned in `/network/nodes` also include `last_written`, as a hint to when each node was added.

_\* The config argument is a top-level field named `node_data_json_file`. It's a path to a file for consistency with `member_data`. We could inline these now that we have a JSON config, rather than requiring them to be dumped to a separate file, but it's a more awkward change when migrating from 1.x to 2.x_

### TODO
- [x] CHANGELOG
- [x] Documented config schema

### DIFF
--- a/include/ccf/service/node_info.h
+++ b/include/ccf/service/node_info.h
@@ -56,6 +56,10 @@ namespace ccf
     /// Public key
     std::optional<crypto::Pem> public_key = std::nullopt;
 
+    /// Free-form user data, can be used to store operator correlation
+    /// IDs/labels for the node for example
+    nlohmann::json node_data = nullptr;
+
     /**
      * Fields below are deprecated
      */
@@ -75,7 +79,8 @@ namespace ccf
     ledger_secret_seqno,
     code_digest,
     certificate_signing_request,
-    public_key);
+    public_key,
+    node_data);
 }
 
 FMT_BEGIN_NAMESPACE
--- a/src/host/configuration.h
+++ b/src/host/configuration.h
@@ -60,6 +60,9 @@ namespace host
     ds::TimeString slow_io_logging_threshold = {"10ms"};
     std::optional<std::string> node_client_interface = std::nullopt;
     ds::TimeString client_connection_timeout = {"2000ms"};
+    // TODO: This is a path to a file, for consistency with member data. But
+    // those should probably both be inlined now that we have JSON configs
+    std::optional<std::string> node_data_json_file = std::nullopt;
 
     struct OutputFiles
     {
@@ -210,6 +213,7 @@ namespace host
     slow_io_logging_threshold,
     node_client_interface,
     client_connection_timeout,
+    node_data_json_file,
     output_files,
     ledger,
     snapshots,
--- a/src/host/configuration.h
+++ b/src/host/configuration.h
@@ -60,8 +60,6 @@ namespace host
     ds::TimeString slow_io_logging_threshold = {"10ms"};
     std::optional<std::string> node_client_interface = std::nullopt;
     ds::TimeString client_connection_timeout = {"2000ms"};
-    // TODO: This is a path to a file, for consistency with member data. But
-    // those should probably both be inlined now that we have JSON configs
     std::optional<std::string> node_data_json_file = std::nullopt;
 
     struct OutputFiles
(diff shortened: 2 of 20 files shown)


src/host/configuration.h
```diff
@@ -210,6 +211,7 @@ namespace host
     slow_io_logging_threshold,
     node_client_interface,
     client_connection_timeout,
+    node_data_json_file,
```
achamayou, line 214: Can we update cchost_config.json with a description for this?

achamayou: See other comment

src/node/rpc/node_frontend.h
```diff
@@ -876,7 +877,13 @@ namespace ccf
...
           }
 
-          out.nodes.push_back({nid, ni.status, is_primary, ni.rpc_interfaces});
+          out.nodes.push_back(
+            {nid,
+             ni.status,
+             is_primary,
+             ni.rpc_interfaces,
+             ni.node_data,
+             nodes->get_version_of_previous_write(nid).value_or(0)});
```
jumaffre, line 886: This is expected to always have a value, no?

include/ccf/service/node_info.h
```diff
@@ -56,6 +56,10 @@ namespace ccf
     /// Public key
     std::optional<crypto::Pem> public_key = std::nullopt;
 
+    /// Free-form user data, can be used to store operator correlation
+    /// IDs/labels for the node for example
+    nlohmann::json node_data = nullptr;
```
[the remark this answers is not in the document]
achamayou: Yes, that's necessary for ledger compatibility, good catch.

src/node/rpc/node_frontend.h
achamayou: Not if the key is missing, iirc?

achamayou (approved):

achamayou:
@eddyashton doxygen's not happy:

```
/__w/1/s/src/node/rpc/serialization.h:76: error: Found ';' while parsing initializer list! (doxygen could be confused by a macro call without semicolon) (warning treated as error, aborting now)
```

src/node/rpc/serialization.h
```diff
@@ -72,7 +72,8 @@ namespace ccf
     public_encryption_key,
     code_digest,
     node_info_network)
-  DECLARE_JSON_OPTIONAL_FIELDS(CreateNetworkNodeToNode::In, genesis_info)
+  DECLARE_JSON_OPTIONAL_FIELDS(
+    CreateNetworkNodeToNode::In, genesis_info, node_data);
```
achamayou, line 76:
```suggestion
    CreateNetworkNodeToNode::In, genesis_info, node_data)
```

include/ccf/service/node_info.h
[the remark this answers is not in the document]
eddyashton: That was discussed here: https://github.com/microsoft/CCF/discussions/2916

We said:
> The new field will either have to be defined as i) std::optional or ii) _a non-std::optional type, with a sensible default value_.

In general, we need to make new fields `OPTIONAL` in the serialisation macro (to deserialise old entries where it is not present), but only need `std::optional` where we need a special value to distinguish cases where it is unset. I understand why we need to do that for the fields above (we expect all new nodes to have a `public_key`, and take different execution paths when that is the case).

But for this `node_data` field, we a) already have a sensible distinct default value (JSON `null`), and b) can treat old and new nodes identically. We never need to distinguish "an old node without `node_data` (because it didn't exist)" from "a new node without `node_data` (because it wasn't set)", and in fact have a third case of "an old node _with_ `node_data` (because it was set on a new peer!)".

So I think this is fine, and fits in our previously discussed restrictions.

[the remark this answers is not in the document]
achamayou: Ok, this is consistent with what we do for members and users too (using json `null` as the default I mean).

PR #3663, opened 2022-03-15
Add documentation for identity parameters of transition_service_to_open

### LINKED ISSUE #3082: Add optional `--defunct-network-cert` option to `cchost` for recovery
As discussed with @msftsettiy, @sidmore and @lynshi

When recovering a service (i.e. ledger + snapshots), it would be nice to let operators specify the certificate of the defunct network certificate (new optional `--defunct-network-cert` option to `cchost`). Assuming that the ledger/snapshots and defunct service certificate are retrievable from different locations, this would mostly be about avoiding confusion when managing multiple ledger instances, by making sure that the operators have confidence that the service they're recovering is the correct instance. 

In particular, this would be useful to verify the endorsement of the recovery snapshot receipt early on, or verify that the service certificate matches the one recorded in the [`public:ccf.gov.service.info`](https://microsoft.github.io/CCF/main/audit/builtin_maps.html#service-info) table.

The option should be optional in case it is not possible for operators to retrieve the defunct service certificate before initiating the recovery procedure.

wintersteiger:
Follow-up to #3624. Fixes #3082.


doc/governance/accept_recovery.rst
```diff
@@ -48,6 +51,8 @@ A member proposes to recover the network and other members can vote on the propo
 
 Once the proposal to recover the network has passed under the rules of the :term:`Constitution`, the recovered service is ready for members to submit their recovery shares.
 
+Note that the ``transition_service_to_open`` proposal takes two parameters: the previous and the next service identity. This is to ensure that the correct network is recovered and to facilitate auditing, as well as to avoid forks. The previous service identity is used to validate the snapshot the recovery node is started from; CCF will refuse to start from a snapshot where the signing node certificate is not endorsed by the previous service identity. Since both identities are recorded on the ledger with the proposal, it is always clear at which point the identity changed.
```
eddyashton, line 54: We should clarify that the previous identity is optional. Or more precisely, must _not_ be provided for new services, but _must_ be provided for recoveries.

doc/governance/accept_recovery.rst
```diff
@@ -17,7 +17,10 @@ A member proposes to recover the network and other members can vote on the propo
         "actions": [
             {
                 "name": "transition_service_to_open",
-                "args": null
+                "args": {
+                    "previous_service_identity": "-----BEGIN CERTIFICATE-----\nMIIBuDCCAT6gAwIBAgIRANWm4xJICc6i4sir+jRXE2gwCgYIKoZIzj0EAwMwFjEU\nMBIGA1UEAwwLQ0NGIE5ldHdvcmswHhcNMjIwMzExMTcwNTEyWhcNMjIwMzEyMTcw\nNTExWjAWMRQwEgYDVQQDDAtDQ0YgTmV0d29yazB2MBAGByqGSM49AgEGBSuBBAAi\nA2IABOyCL4ZOG0mu7fLpciVWcDHFp1dOVr1osONVgG/fhjjZryR/HS5xIc20d96L\nN4yl6qbtoEGE1r1juQB44xoEKOox7OLRD2S0N1/T/DfdCIdgyv5rAVIFCMZVtxGA\nsg6I26NQME4wDAYDVR0TBAUwAwEB/zAdBgNVHQ4EFgQUAz9Pfzi3sEN2w6KuncV2\n0wFXDC0wHwYDVR0jBBgwFoAUAz9Pfzi3sEN2w6KuncV20wFXDC0wCgYIKoZIzj0E\nAwMDaAAwZQIxAKM+T5Lvv4/2nKn8ZL87DkKiBwaGh1kLmrM/0xLhlQYgRp13iqw8\ndt/Zm+/dLCZe/AIwBrgsP5YM2TZ/AAHgC50H8+DKd0k/DfVIy28qhMb/6jr1bCMp\nf0CN7wvG22F59hDa\n-----END CERTIFICATE-----\n\u0000",
```
eddyashton, line 21: Our other cert-pem-in-proposal instances don't include the trailing null byte. Is that required for this identity? If not, please remove it for consistency.

wintersteiger: That's a good question! It shouldn't be required (post mbedTLS), but I'll run some tests.

wintersteiger: Yep, all fine without the null bytes. I added a few lines of code to actively remove them when reading a certificate file, but I'm not sure we should keep that, because it looks more complicated than just not worrying about them. They are in the last commit and I'm happy to revert that if you agree.

doc/governance/accept_recovery.rst
```diff
@@ -48,6 +51,10 @@ A member proposes to recover the network and other members can vote on the propo
 
 Once the proposal to recover the network has passed under the rules of the :term:`Constitution`, the recovered service is ready for members to submit their recovery shares.
 
+Note that the ``transition_service_to_open`` proposal takes two parameters: the previous and the next service identity. This is to ensure that the correct network is recovered and to facilitate auditing, as well as to avoid forks. The previous service identity is used to validate the snapshot the recovery node is started from; CCF will refuse to start from a snapshot where the signing node certificate is not endorsed by the previous service identity. Since both identities are recorded on the ledger with the proposal, it is always clear at which point the identity changed.
```
jumaffre, line 54: Could you mention that the "service identity" is the service x509 certificate (PEM)?

doc/governance/accept_recovery.rst
```diff
@@ -17,7 +17,10 @@ A member proposes to recover the network and other members can vote on the propo
         "actions": [
             {
                 "name": "transition_service_to_open",
-                "args": null
+                "args": {
```
jumaffre, line 20: I think it's better to not include a full, valid, certificate as it's likely to be pasted as is by users for their own service. Perhaps `"-----BEGIN CERTIFICATE-----\n ...  \n-----END CERTIFICATE-----\n"` is sufficient here?

doc/host_config_schema/cchost_config.json
```diff
@@ -207,8 +207,13 @@
                     "default": 1,
                     "description": "Initial validity period (days) for service certificate",
                     "minimum": 1
+                  },
+                  "previous_service_identity_file": {
+                    "type": "string",
+                    "description": "Path to the file that contains the previous network certificate (PEM)"
```
jumaffre, line 213:
```suggestion
                    "description": "Path to the previous service certificate (PEM) file"
```

samples/config/recover_config.json
```diff
@@ -24,7 +24,8 @@
     "type": "Recover",
     "service_certificate_file": "service_cert.pem",
     "recover": {
-      "initial_service_certificate_validity_days": 1
+      "initial_service_certificate_validity_days": 1,
+      "previous_service_identity_file": "/tmp/previous_service_cert.pem"
```
jumaffre, line 28:
```suggestion
      "previous_service_identity_file": "previous_service_cert.pem"
```

To be consistent with other paths?

src/host/main.cpp
```diff
@@ -434,6 +434,11 @@ int main(int argc, char** argv)
       }
       LOG_INFO_FMT("Reading previous service identity from {}", idf);
       startup_config.recover.previous_service_identity = files::slurp(idf);
+      while (startup_config.recover.previous_service_identity->back() == 0x00)
+      {
+        // Trailing null byte not required, but tolerated.
```
jumaffre, line 439: Is that because of 1.x service certificates?

wintersteiger: See response to Eddy's comment further up. I just added that to make sure we accept certs without the null byte, but I'm thinking about removing that commit as it looks more complicated to users, when they should just not worry about it.

doc/governance/accept_recovery.rst
wintersteiger: Sure, np.

doc/governance/accept_recovery.rst
wintersteiger: I added the complete cert because that's what's done in other places, e.g. `proposals.rst`, not sure if we want this here or not.

achamayou: It's fine to do this, we can't remove every sample value in the doc for fear it'll get copy-pasted. Having realistic examples makes the doc much easier to read.

Besides, there's no way that copy pasted proposal will succeed, so it won't do any harm if they copy-paste this particular sample verbatim.

doc/governance/open_network.rst
```diff
@@ -130,7 +130,9 @@ Once users are added to the opening network, members should create a proposal to
         "actions": [
             {
                 "name": "transition_service_to_open",
-                "args": null
+                "args": {                 
+                    "next_service_identity": "-----BEGIN CERTIFICATE-----\nMIIBezCCASGgAwIBAgIRAOVHYf9qhvjzdoIw3fPHp5YwCgYIKoZIzj0EAwIwFjEU\nMBIGA1UEAwwLQ0NGIE5ldHdvcmswHhcNMjIwMzExMTcwNTQzWhcNMjIwMzEyMTcw\nNTQyWjAWMRQwEgYDVQQDDAtDQ0YgTmV0d29yazBZMBMGByqGSM49AgEGCCqGSM49\nAwEHA0IABBZXMHCrjfBeO+FHqDG8Szjzc4lQC8KmvTX8Il0ZERXH/mjLZ7Dc52rX\nnilD1ghdRDWXiKMQWT9RPvm4tefWHD6jUDBOMAwGA1UdEwQFMAMBAf8wHQYDVR0O\nBBYEFCUmm9u05D0/IFupggFW5VgVlUSyMB8GA1UdIwQYMBaAFCUmm9u05D0/IFup\nggFW5VgVlUSyMAoGCCqGSM49BAMCA0gAMEUCIQCy6WoeLtTUD8GRIOM+oRNe/lTj\nRrrry+0AxZgxBU1oSwIgJmyrTfT90re+rzAkF9uiqoL44TVWkQf1t3cZrgVFYK8=\n-----END CERTIFICATE-----\n"
```
jumaffre, line 134: Could `next_service_identity` be renamed to `service_identity` as this is a little confusing for the very first service open?

src/host/main.cpp
```diff
@@ -440,6 +440,11 @@ int main(int argc, char** argv)
       }
       LOG_INFO_FMT("Reading previous service identity from {}", idf);
       startup_config.recover.previous_service_identity = files::slurp(idf);
+      while (startup_config.recover.previous_service_identity->back() == 0x00)
```
jumaffre, line 443: Could these be removed before this PR is merged?

jumaffre (approved):

doc/governance/open_network.rst
wintersteiger: Yeah, I think so too, I'll rename it.

src/host/main.cpp
wintersteiger: Yes, I want to remove those too. @eddyashton last chance if you want to keep them, but I don't think you do.

eddyashton: Sorry I missed this - yeah I think those should be removed, thanks.

eddyashton (approved):

samples/constitutions/default/actions.js
```diff
@@ -575,11 +575,11 @@ const actions = new Map([
     new Action(
       function (args) {
         checkType(
-          args.next_service_identity,
+          args.service_identity,
           "string",
-          "next service identity (PEM certificate)"
+          "service identity (PEM certificate)"
```
achamayou, line 580: I am not sure removing "next" is doing that much for clarity, it seemed nicely symmetrical with "previous"?

Issue #3654, opened 2022-03-14
Make a neater public-facing `RpcContext`


eddyashton:
This type is currently under `enclave::`, but should really be under `ccf::`. It also includes some methods and fields which do not need to be exposed to apps, and should be removed.

eddyashton:
Also needs to be properly Doxygen'd and included in the docs.

Issue #3580, opened 2022-02-22
Nest all public CCF code under `ccf` namespace


eddyashton:
We currently define CCF types in various namespaces, along historic divisions in the codebase. This introduces some confusion for consumers in finding the correct name for our types, and may lead to naming conflicts in C++ projects with other large dependencies.

Ideally we should move everything under `ccf::`, nested as appropriate (eg `ccf::kv::`, `ccf::crypto::`).

For an approximation of the namespaces exposed by our public headers:
```
$ find . -type f | xargs grep -h -e "namespace" | sort -u

namespace ccf
namespace ccf::endpoints
namespace ccf::historical
namespace ccf::indexing
namespace ccf::indexing::strategies
namespace ccf::js
namespace ccfapp
namespace crypto
namespace ds
namespace enclave
namespace http
namespace kv
namespace kv::serialisers
namespace logger
namespace nonstd
namespace serdes
namespace siphash
namespace std
namespace threading
```

achamayou:
```
$ find . -type f | xargs grep -h -e "^namespace" | sort -u | grep -v "namespace ccf::"
namespace ccf
namespace ccfapp
namespace crypto
namespace http
namespace http2
namespace kv
namespace kv::serialisers
namespace kv::untyped
namespace logger
namespace nonstd
namespace serdes
namespace std
```

achamayou:
Still left to do:

```
namespace ccfapp
namespace kv
namespace kv::serialisers
namespace kv::untyped
namespace logger
```

achamayou:
Done in 5.0

PR #3664, opened 2022-03-16
Remove `enclave` namespace

eddyashton:
As part of #3654, we want to rename `enclave::RpcContext` to `ccf::RpcContext`. We could do this purely for the public types, and keep some internal types under `enclave::`. But I think `enclave::` is an unhelpful namespace name that we intend to abolish in future, as part of #3580, so I've taken the opportunity to remove it wholesale now. This makes the diff far _simpler_; it's just a blind rename, rather than a per-instance rename to recognise a new namespace boundary! That's all this PR does, and any more interesting semantic changes will be PR'd separately.


eddyashton:
> @eddyashton this needs a changelog entry, since it's user-visible.

Ah I had a CHANGELOG, just forgot to push it. Will add that now.

Issue #3596, opened 2022-02-23
Operator documentation update


achamayou:
Document how an operator can check network health, and understand if a DR is necessary

achamayou:
Adding or modifying network endpoints may be necessary to make monitoring easier.

jumaffre:
The issue here is about detecting `f + 1` failures in a network of `2f + 1` nodes (e.g. detecting `2` failures in a network of `3` nodes - `f = 1`). 

By failure here we mean non-temporary failures, i.e. those that will not be automatically resolved by the network within a reasonable amount of time (with "reasonable" still to be defined).

Temporary network failures include:
1. Follower(s) failure, i.e. one or up to `f` follower node(s) are unresponsive or have crashed. In this case, the network can still process and commit write transactions and the follower node(s) should be replaced by the operator. 
2. Leader change, i.e. a new node is being elected as primary node (maybe because of a temporary partition, or the leader node crashed). The service will be able to elect a new primary node and resume when the election process is successful, which I think we can assume to be complete by 3 * [`election_timeout`](https://microsoft.github.io/CCF/main/operations/configuration.html#consensus) (we really need real-life measurements to tune this factor).
3. Network partition, where the primary node is in the minority partition. This is equivalent to 2. Note that the isolated old primary will still be considered as primary for all the nodes in the minority partition, even though this minority partition will not be able to commit new transactions.

An important requirement for the process of detecting disaster failures is that it should be non-intrusive, i.e. the operator shouldn't have to submit new transactions that will be recorded in the ledger to find out if the service can still commit new transactions.

--

To detect non-temporary network failures, the operator should regularly issue a `GET` request to all the nodes in the network to find out the consensus state:
- Finding out from each node which node is currently primary tells the operator whether the network is in a stable consensus state: `1` leader node and at least `f` followers in the same term (other nodes can be candidate) tells us the network can commit new transactions.
- It is OK to observe a majority `f+1` (or more) of nodes in the candidate state (reported term isn't important), which happens during a leader change (2.). This state shouldn't last for more than 3 * `election_timeout`. If the cluster remains in that state after this duration, the recovery procedure should be started.
- When the primary node is part of a minority partition (3.), the nodes in the majority partition will become candidates and eventually elect a new leader in a new term. The old primary (and other nodes in the minority partition) will still consider it as their primary, while the majority partition will consider the new primary as their primary. This is OK since the majority partition will still be able to commit new transactions. 

[...]

PR #3666, opened 2022-03-17
Add more fields to `GET /node/consensus` response

jumaffre:
This is so operators can determine the triplet `{Raft state, current view, primary id}` by querying only one endpoint (see #3596).


src/consensus/aft/impl/state.h
```diff
@@ -141,7 +141,6 @@ namespace aft
     State(const ccf::NodeId& my_node_id_) : my_node_id(my_node_id_) {}
 
     std::mutex lock;
-    std::map<ccf::NodeId, std::shared_ptr<Replica>> configuration;
```
jumaffre, line 144: This was unused.

eddyashton (approved):

jumaffre:
Merging this as the Daily failure is unrelated and has already been fixed.

PR #3672, opened 2022-03-18
Fix Raft partition test

### LINKED ISSUE #3670: Partitions test failure
Partitions test break: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=43320&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=cc6f52ed-d79a-5851-ff49-59b4954ad736

```
59: Traceback (most recent call last):
59:   File "/__w/1/s/tests/partitions_test.py", line 351, in <module>
59:     run(args)
59:   File "/__w/1/s/tests/partitions_test.py", line 331, in run
59:     test_isolate_primary_from_one_backup(network, args)
59:   File "/__w/1/s/tests/suite/test_requirements.py", line 20, in wrapper
59:     return func(*args, **kwargs)
59:   File "/__w/1/s/tests/partitions_test.py", line 117, in test_isolate_primary_from_one_backup
59:     assert (
59: AssertionError: Consensus view 8 should not changed after partition is dropped: no 11
```

eddyashton:
Resolves #3670.

We don't know what changed recently that made this failure more common, but it appears the test was only passing because of lucky timing. We now wait for the node we expect to be primary to be elected.


CMakeLists.txt
```diff
@@ -689,11 +689,18 @@ if(BUILD_TESTS)
...
+      # Higher snapshot interval as the test currently assumes that no
+      # transactions
+      # are emitted while partitions are up. To be removed when
+      # https://github.com/microsoft/CCF/issues/2577 is implemented
+      --snapshot-tx-interval
+      10000
+      # Partitions test waits for a lot of elections, so reduce the timeout to
+      # complete faster
+      --election-timeout-ms
+      2000
```
eddyashton, line 702: This might be an overly-optimistic change, but would be a nice time saving. Let's see what the CI thinks.

jumaffre: Could we change the default value in `e2e_args.py` so that it applies to all the tests?

tests/partitions_test.py
```diff
@@ -87,8 +87,10 @@ def test_partition_majority(network, args):
 
 
 @reqs.description("Isolate primary from one backup")
+@reqs.exactly_n_nodes(3)
 def test_isolate_primary_from_one_backup(network, args):
-    primary, backups = network.find_nodes()
+    p, backups = network.find_nodes()
+    b_0, b_1 = backups
```
jumaffre, line 93: Could we call these `p_0`, `b_1` and `b_2` so that it matches the `local_node_id`s in the infra?

eddyashton: It probably doesn't, since previous tests have caused multiple partitions and elections. These names are completely local to the test - happy to change them to `a, b, c` if that's easier, just want to know precisely which node we're talking about.

CMakeLists.txt
eddyashton: I'd rather not. I assume something obscure will fail, for not much speed-up benefit.

achamayou (approved):

tests/partitions_test.py
jumaffre: Ah yes, these tests are chained, even on the non-suite runners.

jumaffre (approved):

CMakeLists.txt
eddyashton: In fact even this change is causing the Debug build to occasionally timeout, so I've removed it.

PR #3676, opened 2022-03-18
Recovery detection sample and infra

jumaffre:
This PR writes the (infra) code corresponding to what is documented in https://github.com/microsoft/CCF/issues/3596#issuecomment-1049771595 

This introduces the infra `NetworkHealthWatcher` which can be started in separate thread to monitor the health of a network. If a majority of nodes are unavailable for `DEFAULT_ELECTION_FACTOR * election_timeout`, then the service is considered as dead and will need to be recovered via the disaster recovery procedure.


tests/infra/health_watcher.py
```diff
@@ -0,0 +1,157 @@
...
+def get_primary(node, client_node_timeout_s=3, verbose=True):
+    """
+    Returns the primary reported by a given node, and in which view or
+    None if the given node is unreachable.
+    """
+    try:
+        with node.client() as c:
+            logs = None if verbose else []
+            r = c.get(
+                "/node/network", timeout=client_node_timeout_s, log_capture=logs
```
jumaffre, line 39: This will become `/node/consensus` once https://github.com/microsoft/CCF/pull/3666 is merged, so we also get the Raft state, and can give more precise state on the service health.

lynshi:
Thanks for this; the example is helpful for how we can enhance our monitoring.

Our current scheme is to have a watcher for each node that monitors its local commit relative to the network commit, and marks the node as unhealthy when that node is not progressing as it should (i.e. local commit < network commit AND local commit has not increased in X time). Is this sort kind of monitoring of node progress useful in addition to the logic here or can it be omitted?

achamayou:
@lynshi we've discussed this internally, and while monitoring this is definitely good, taking immediate automatic action is less clearly helpful:

1. If this is caused by network isolation, the best step to take is to wait until the network is back. Replacing the node is risk with no upside in that case.
2. If this is caused by a bug, it needs to be one that leaves the node able to answer RPCs, but unable to receive append entries, and is not deterministically caused by a transaction. In that case a replacement makes sense.

On the whole, replacing (ie. retiring and trusting a successor) is an unmitigated win for crashed processed. For nodes that are otherwise healthy but do not make progress, the best bet is to wait for the network to heal. An aggressive replacement is probably harmful on average.

doc/operations/recovery.rst
```diff
@@ -3,6 +3,13 @@ Disaster Recovery
 
 For unexpected reasons, a significant number [#crash]_ of CCF nodes may become unavailable. In this catastrophic scenario, operators and members can recover transactions that were committed on the crashed service by starting a new network.
 
+The disaster recovery procedure is costly (e.g. the service identity certificate will need to be re-distributed to clients) and should only be staged once operators are confident that the service will not heal by itself. In other words, the recovery procedure should only be staged once a majority of nodes do not consistently report one of them as their primary node.
```
achamayou, line 6: There's also the downtime

tests/infra/health_watcher.py
```diff
@@ -0,0 +1,157 @@
...
+DEFAULT_ELECTION_FACTOR = 1
+
+
+class HealthState(Enum):
+    stable = auto()  # Network can commit new transactions
+    unavailable = auto()  # The primary or majority of nodes are unreachable
+    election = auto()  # An election is in progress
+
+
+def get_primary(node, client_node_timeout_s=3, verbose=True):
```
achamayou, line 30:
```suggestion
def get_primary(node, client_node_timeout_s=DEFAULT_CLIENT_NODE_TIMEOUT_S , verbose=True):
```

tests/infra/health_watcher.py
```diff
@@ -0,0 +1,157 @@
...
+    majority = (len(nodes) + 1) // 2
+
+    primaries = {}
+    for node in nodes:
+        primaries[node.node_id] = get_primary_fn(
+            node, client_node_timeout_s, verbose=verbose
+        )
+    assert len(primaries) == len(nodes)
+
+    # Count how many primary nodes are reported by all nodes in the network.
```
achamayou, line 61:
```suggestion
    # Count how many (primary nodes, views) are reported by all nodes in the network.
```

PR #3671, opened 2022-03-18
Parsing errors as metrics

jumaffre:
Resolve #3651 

Instead of writing a new line to the node's log, we now report parsing error under the interface-generic `GET /node/metrics` endpoint. 

Sample:
```
GET /node/metrics
{"sessions":{"active":2,"interfaces":{"first_interface":{"active":0,"errors":{"parsing":100},"hard_cap":1010,"peak":3,"soft_cap":1000}}}
```

This wasn't as straightforward as catching the exception from the parent object `RPCSessions` since the execution is forked off to a worker thread at this point. Instead, we pass a `shared_ptr` to `RPCSessions` to each `HTTPEndpoint` that can report parsing errors.

PR #3678, opened 2022-03-18
Make fewer calls to `/node/metrics`

eddyashton:
#3671 made 2 changes to `e2e_logging_cft`, which pushed it over the time limit in the `Instrumented` build:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=43374&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=ae32e15b-0191-5f39-7410-be7e80165871&l=39384

The changes were to include the illegal traffic test in existing runners, rather than their own dedicated runners, and to call `/node/metrics` twice for every corrupt HTTP request. We make a lot of corrupt HTTP calls, so that adds up! I think we can get the time back under the bound by making fewer `/node/metrics` calls, but ~~if not I'll restore the dedicated runners~~ _I was wrong about that, so I've restored the dedicated runners._


jumaffre:
Good spot! Thanks for fixing this. 

If we're so close to the CI timeout for the instrumented build, should we increase this timeout for this specific job?

jumaffre (approved):

eddyashton:
> Good spot! Thanks for fixing this.
> 
> If we're so close to the CI timeout for the instrumented build, should we increase this timeout for this specific job?

I don't think we are particularly close, I think this was a multiplier on the runtime. Previous runs finished in [~270s](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=43187&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=ae32e15b-0191-5f39-7410-be7e80165871&l=36373). And if we were close, I think we should get back under the boundary by splitting into more runners, rather than just increasing the run time.

jumaffre:
> > Good spot! Thanks for fixing this.
> > If we're so close to the CI timeout for the instrumented build, should we increase this timeout for this specific job?
> 
> I don't think we are particularly close, I think this was a multiplier on the runtime. Previous runs finished in [~270s](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=43187&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=ae32e15b-0191-5f39-7410-be7e80165871&l=36373). And if we were close, I think we should get back under the boundary by splitting into more runners, rather than just increasing the run time.

OK, this is for the particular `e2e_logging_cft` test, rather than the whole `ctest` run (which I think is also close to its timeout).

Issue #3343, opened 2021-12-21
Support for receipts post-recovery


jumaffre:
We do not currently return the full root of trust to the current service certificate when returning receipts for pre-recovery transactions. Instead, we should:

1. [ ] Endorse the `N`th service identity with the `N+1`th when performing a disaster recovery and record this endorsement in a new KV table (either as an ever-growing table or better, by only recording the latest service endorsement). 
2. [ ] Fetch these endorsements from the ledger when retrieving historical receipts and include them in the receipts
3. [ ] Update receipt verification sample and remove `check_endorsement` flag in infra receipt verification function:
https://github.com/microsoft/CCF/blob/3f67461a324e9e1142d11f70627a573bf2091fb0/tests/e2e_logging.py#L37

achamayou:
There are a few possibilities with different benefits and tradeoffs:

Pair-wise endorsement
-------------------------

At each recovery, `SI[N] - endorses -> SI[N-1]`, this endorsement is stored in a table.

When a user requests a receipt endorsed by `SI[P]`, they are given the list of `SI[I] -> SI[I-1]` from `P` to `N` where `N` is the current service identity.

Benefits:

1. Constant time to endorse on recovery
2. Easy audit (pair-wise chain)

Drawbacks:

1. Space usage in KV is linear in the number of recoveries `R`

achamayou:
Upon further discussion, it becomes clear that the on-demand scheme do not benefit from writing their endorsements to the ledger, on the contrary.

It is better in that scheme to maintain a per-node in-memory cache, which can be primarily done by the backups, serving historical queries, leaving the primary free to handle writes. This removes the quadratic space impact (and indeed any impact) on the ledger size.

PR #3679, opened 2022-03-18
Add service identity endorsements to historical receipts

wintersteiger:
Implements #3343


include/ccf/crypto/key_pair.h
```diff
@@ -60,7 +61,8 @@ namespace crypto
       const Pem& signing_request,
       const std::string& valid_from,
       const std::string& valid_to,
-      bool ca = false) const = 0;
+      bool ca = false,
+      bool signed_by_issuer = false) const = 0;
```
achamayou, line 65: `Signer signed_by = Signer::SUBJECT/Signer::ISSUER` may be clearer and potentially more extensible?

src/node/historical_queries_adapter.cpp
```diff
@@ -189,13 +196,112 @@ namespace ccf::historical
...
+
+    // Iterate until we find the most recent write to the service info that
+    // precedes the target seqno.
+    std::optional<ServiceInfo> hservice_info = service->get();
+    SeqNo i = -1;
+    do
+    {
+      // If no previous service info version available, e.g. after recovery of
+      // a 1.x service, we scan linearly.
+      i = hservice_info->previous_service_identity_version.value_or(i - 1);
```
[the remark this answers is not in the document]
achamayou: I think we may just throw an error at this point, and not add any additional endorsements. Having the successor endorse the node without having checked its attestation/code_id seems strange and is going to make auditing potentially quite indirect.

tests/e2e_logging.py
```diff
@@ -36,15 +36,63 @@
...
 
 
+# def show_cert(name, cert):
+#     from OpenSSL.crypto import dump_certificate, FILETYPE_TEXT
+
+#     dc = dump_certificate(FILETYPE_TEXT, cert).decode("unicode_escape")
+#     LOG.info(f"{name} cert: {dc}")
+
+
+# def verify_endorsements_openssl(service_cert, receipt):
```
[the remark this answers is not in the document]
achamayou: Fine to keep it if it works.

achamayou (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - New `GET /node/consensus` endpoint now also returns primary node ID and current view (#3666).
 - The `enclave::` namespace has been removed, and all types which were under it are now under `ccf::`. This will affect any apps using `enclave::RpcContext`, which should be replaced with `ccf::RpcContext` (#3664).
 - HTTP parsing errors are now recorded per-interface and returned by `GET /node/metrics` (#3671).
+- Receipts now come with service endorsements of previous service identities after recoveries (#3679). See `verify_receipt` in `e2e_logging.py` for an example of how to verify the resulting certificate chain.
```
jumaffre, line 16: Could we add a note in the docs about this, and explain how the receipts will differ for pre-recovery receipts?

include/ccf/crypto/key_pair.h
jumaffre: Why would the issuer ever sign the CSR?

include/ccf/receipt.h
```diff
@@ -51,6 +54,8 @@ namespace ccf
     std::optional<std::string> leaf = std::nullopt;
     /// Leaf components in transactions emitted by 2.x networks.
     std::optional<LeafComponents> leaf_components = std::nullopt;
+
```
jumaffre, line 57:
```suggestion
/// Service certificate (PEM) endorsing the node certificate (cert). Unset if the service never experienced disaster recovery.
```

Or something like that?

include/ccf/service/tables/service.h
```diff
@@ -27,13 +28,13 @@ namespace ccf
     /// x.509 Service Certificate, as a PEM string
     crypto::Pem cert;
     /// Status of the service
-    ServiceStatus status;
-    /// Previous service identity, before the last recovery
-    std::optional<crypto::Pem> previous_service_identity = std::nullopt;
+    ServiceStatus status = ServiceStatus::OPENING;
+    /// Version of previous service identity (before the last recovery)
+    std::optional<kv::Version> previous_service_identity_version = std::nullopt;
```
jumaffre, line 33:
```suggestion
    std::optional<kv::Version> previous_service_identity_seqno = std::nullopt;
```

For consistency with the rest of the API.

samples/apps/logging/logging.cpp
```diff
@@ -843,7 +844,11 @@ namespace loggingapp
         "/log/private/historical",
         HTTP_GET,
         ccf::historical::adapter_v2(
-          get_historical, context.get_historical_state(), is_tx_committed),
+          get_historical,
+          context.get_historical_state(),
+          *context.get_subsystem<ccf::NetworkIdentitySubsystemInterface>()
+             .get(),
+          is_tx_committed),
```
jumaffre, line 851: This seems like a breaking change in the historical queries API that should be clearly documented in the changelog.

src/node/historical_queries_adapter.cpp
```diff
@@ -189,13 +196,112 @@ namespace ccf::historical
     return HistoricalTxStatus::Valid;
   }
 
+  std::optional<ServiceInfo> find_previous_service_identity(
+    auto& ctx,
+    ccf::historical::StatePtr& state,
+    AbstractStateCache& state_cache)
+  {
+    SeqNo target_seqno = state->transaction_id.seqno;
+    LOG_DEBUG_FMT("target seqno: {}", target_seqno);
```
jumaffre, line 205: Remove these debug lines, or make them `LOG_TRACE_FMT` maybe?

src/node/historical_queries_adapter.cpp
```diff
@@ -189,13 +196,112 @@ namespace ccf::historical
...
+
+    // We start at the previous write to the latest (current) service info.
+    auto service = ctx.tx.template ro<Service>(Tables::SERVICE);
+    LOG_DEBUG_FMT("current identity: {}", service->get()->cert.str());
+
+    // Iterate until we find the most recent write to the service info that
+    // precedes the target seqno.
+    std::optional<ServiceInfo> hservice_info = service->get();
```
jumaffre, line 213:
```suggestion
    auto hservice_info = service->get();
```

src/node/historical_queries_adapter.cpp
```diff
@@ -189,13 +196,112 @@ namespace ccf::historical
...
+          auto endorsement = create_endorsed_cert(
+            hpubkey,
+            ReplicatedNetworkIdentity::subject_name,
+            {},
+            crypto::OpenSSL::to_x509_time_string(
+              crypto::OpenSSL::to_time_t(NULL)),
+            1 /* days valid */,
+            network_identity->priv_key,
+            network_identity->cert,
+            true);
```
jumaffre, line 287: Does this mean the historical service certificate that's included in the receipt contains the same public key as the original service certificate, but with a different validity period (and maybe other attributes)?

include/ccf/crypto/key_pair.h
wintersteiger: The issuer signs the CSR on behalf of the previous service, which isn't available anymore. It's a somewhat usual thing to do, but in our case we could get around it by immediately issuing a CSR in the recovery transaction, so that a future service identity can then sign it after the next recovery. I don't think that's necessary, but it's definitely an option, especially if we do not support old ledgers.

src/service/genesis_gen.h
```diff
@@ -292,7 +292,8 @@ namespace ccf
       auto service = tx.rw(tables.service);
       service->put(
         {service_cert,
-         recovering ? ServiceStatus::RECOVERING : ServiceStatus::OPENING});
+         recovering ? ServiceStatus::RECOVERING : ServiceStatus::OPENING,
```
jumaffre, line 295: I wasn't aware of the new `RECOVERING` status but I know some things (e.g. `ledger_viz.py`) rely on the `OPENING` status to detect a new service/recovery. I'd like to keep the recovery and service opening sequence as similar as possible because the two flows are already too complex (especially the latter). 

I can see the `RECOVERING` state only used to check that we require a `previous_service_identity` on the `transition_service_to_open` proposal. I feel like we could check this with the new `previous_service_identity_version` field in the service info table (i.e. if `std::nullopt`, then it's the first service open, otherwise it's a recovery).

tests/recovery.py
```diff
@@ -9,13 +9,69 @@
...
+import random
 from infra.runner import ConcurrentRunner
 from distutils.dir_util import copy_tree
 from infra.consortium import slurp_file
+from e2e_logging import verify_receipt
 
 from loguru import logger as LOG
 
 
+def split_all_ledger_files_in_dir(input_dir, output_dir):
```
jumaffre, line 21: This was moved to another file in a separate PR - merge conflict?

tests/recovery.py
```diff
@@ -9,13 +9,69 @@
...
+        ref_msg = network.txs.priv[idx][-1]
+        ref_msg["idx"] = idx
+    r = network.txs.get_receipt(
+        primary,
+        ref_msg["idx"],
+        ref_msg["seqno"],
+        ref_msg["view"],
+    )
+    verify_receipt(r.json()["receipt"], network.cert)
+    return ref_msg
```
jumaffre, line 72: I would have expected that the `logging_txs` already handles receipt verification?

samples/apps/logging/logging.cpp
wintersteiger: The subsystem was Eddy's suggestion and I liked it because it means we don't have to give access to the private key to the application unnecessarily. I chose the `*get` syntax just to make it fit into the same interface pattern as the historical state, which is also a subsystem, but we can e.g. move it into a helper function similar to `get_historical_state`

src/node/historical_queries_adapter.cpp
wintersteiger: Good point, most of those shouldn't be necessary anymore.

samples/apps/logging/logging.cpp
jumaffre: The subsystem is a good idea, but the fact that the historical queries subsystem requires the network identity shouldn't be a worry for the app developers. I think it would be sufficient to pass the network identity to the `ccf::historical::StateCache` the same way that we do for the `network.ledger_secrets` (see `enclave/enclave.h`) so that this is hidden from the app completly.

src/service/genesis_gen.h
wintersteiger: That's a good observation and I think you're right that the previous version should be enough of an indicator. On the side I'm always thinking about recovering a 2.x service from a 1.x ledger, and in that scenario, it wouldn't work, but maybe we just don't want to support that.

tests/recovery.py
wintersteiger: Yep, bad merge it seems.

tests/recovery.py
wintersteiger: Only for logging-app specific transactions with an `idx`. This here is a general receipt request for any transaction, which didn't exist yet (to my surprise).

src/service/genesis_gen.h
jumaffre: Good point. I think it's OK to not support that for historical 1.x ledgers.

src/node/historical_queries_adapter.cpp
jumaffre: > I'm very flexible about the validity period though - it's not particularly useful in any case, because all of this is about historical state and therefore "expired by definition".

Is it better to use the same validity period (and other fields such as CN) for the new certificate then?

tests/recovery.py
wintersteiger: Wait, my comment doesn't make sense for this code location (I thought this was about `get_receipt` in `node.py`, but apparently my reading skills aren't up to scratch today). In any case, `get_and_verify_historical_receipt` just makes sure there is at least one historical transaction for which we can verify the receipt later, including the logging-app specific `idx` and `msg`. It keeps track of which transaction that is in `ref_msg`.

include/ccf/historical_queries_adapter.h
```diff
@@ -52,6 +53,7 @@ namespace ccf::historical
   ccf::endpoints::EndpointFunction adapter_v2(
     const HandleHistoricalQuery& f,
     AbstractStateCache& state_cache,
+    NetworkIdentitySubsystemInterface& network_identity_subsystem,
     const CheckHistoricalTxStatus& available,
     const TxIDExtractor& extractor = txid_from_header);
```
eddyashton, line 58: We shouldn't add a new argument and still calls this `_v2`. Instead we should add a `_v3`, which probably takes an `AbstractNodeContext` rather than a `StateCache` _or_ a `IdentitySubsystem`, and gets both (+ whatever it needs in future, maybe) from the `AbstractNodeContext`. For the implementations, we keep as much common as we can, and we deprecate `_v2` if that's too hard/it can no longer be provided.

wintersteiger: That's a very good point, will add that immediately.

samples/apps/logging/logging.cpp
eddyashton: `*ctx.get<>().get()` is unnecessary. Either the callee can take a `shared_ptr` (remove the `*` and `.get()`), or if it needs to take `&`, then we can still remove the final `.get()` (`shared_ptr` implements `operator *` to do the same thing, so we don't need to extract the raw pointer and deref it).

In terms of adding helpers to `get_historical_state`, we should be sure we're only doing that for definitely public subsystems. It's a case of "if the app should use it, here's a helper function to get it". Other possibilities are "it's gettable by the app, but not often enough/definitively enough that we want to give it a helper", and "it's a private subsystem which can _not_ get a publicly visible helper method".

wintersteiger: With the switch to `adapter_v3` this now takes an `AbstractNodeContext` and it gets the subsystems internally. I think this is the right structure. In future we may want to expose more of the network identity, e.g. the possibility of signing something on behalf of the application, while not providing direct access to the private key, so I believe it's fine to have an abstract interface for this. However, if you don't agree, I'm also happy to remove that helper and make it internal only.

src/service/genesis_gen.h
wintersteiger: Actually, halfway through making those changes, I realized that this is already in RC4. Is it still okay to remove this state, or do the same arguments as for the renaming of arguments apply?

wintersteiger: I'm leaving this in; that can be a separate PR if we decide to remove that state.

samples/apps/logging/logging.cpp
wintersteiger: Changed my mind again. I kept the subsystem, but made it private for now, but we can easily make it public again by moving the interface header into `include/ccf` if we want to.

src/service/genesis_gen.h
jumaffre: Could you create an issue for this as this is important for some of our tooling?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - New `GET /node/consensus` endpoint now also returns primary node ID and current view (#3666).
 - The `enclave::` namespace has been removed, and all types which were under it are now under `ccf::`. This will affect any apps using `enclave::RpcContext`, which should be replaced with `ccf::RpcContext` (#3664).
 - HTTP parsing errors are now recorded per-interface and returned by `GET /node/metrics` (#3671).
+- Receipts now come with service endorsements of previous service identities after recoveries (#3679). See `verify_receipt` in `e2e_logging.py` for an example of how to verify the resulting certificate chain. This also applies to ccf::historical::adapter_v3 which replaces ccf::historical::adapter_v2, which did not produce these endorsements.
```
achamayou, line 16: The phrasing makes it sound like this applies to both v2 and v3, but I'm guessing only v3 has that feature?

achamayou, line 16:
```suggestion
- Receipts now come with service endorsements of previous service identities after recoveries (#3679). See `verify_receipt` in `e2e_logging.py` for an example of how to verify the resulting certificate chain. This functionality is introduced in `ccf::historical::adapter_v3`.
```

python/ccf/receipt.py
```diff
@@ -51,3 +51,14 @@ def check_endorsement(endorsee: Certificate, endorser: Certificate):
     endorser_pk.verify(
         endorsee.signature, digest, ec.ECDSA(utils.Prehashed(digest_algo))
     )
+
+
+def check_endorsements(
+    node_cert: Certificate, service_cert: Certificate, endorsements: List[Certificate]
+):
+    cert_i = node_cert
+    if endorsements:
```
achamayou, line 60: That check isn't necessary with `endorsements` not being optional (as declared), but if you want to be defensive and make it optional, you could `for endorsement in endorsements or []:`.

wintersteiger: Yeah, just being defensive, I added the `or []`

tests/e2e_logging.py
```diff
@@ -1382,16 +1428,23 @@ def test_random_receipts(
...
                         # Always a signature receipt
                         assert receipt["proof"] == [], receipt
                     break
                 elif rc.status_code == http.HTTPStatus.ACCEPTED:
                     time.sleep(0.5)
+                elif (
+                    lts
+                    and rc.status_code == http.HTTPStatus.INTERNAL_SERVER_ERROR
+                    and "is in a pre-2.0 part of the ledger"
+                    in rc.body.json()["error"]["message"]
```
wintersteiger, line 1445: @achamayou The endpoint currently throws exceptions when it can't find old service identities and when node certificates are missing, so I catch them here in the test scripts. This may break behavior which is only noticed after an upgrade to 2.0. Alternatively, we could either just ignore the errors and deliver empty endorsements (pretty much 1.x behaviour). In that case I would like to add some kind of warning message though. Do we have warning headers for HTTP replies or something alike?

python/ccf/receipt.py
achamayou: Ok, fine, I think it's just a little weird, because there are two level of validation that make sense to me:

1. Assume the stuff passed in is fine, the contract is what's stated in the docstring or the type definition.
2. Be defensive, and runtime check our assumptions, in which case, the right thing to do here is check for `iterable` before we continue.

Instead, this kind of a weird in-the-middle where we say "Assume it's fine, unless it coerces to False, in which case we assume it's an empty iterable, but we don't check or handle anything else at runtime", which is weirdly specific? If `endorsements` was `Optional[List[..]]`, then sure, that's 1.

tests/e2e_logging.py
achamayou: I think it's fine not to return endorsements for pre-2.0 transactions, and it needn't be an error. We very rarely retroactively add features, there's not a particularly strong reason to expect this one to be different?

We could set http headers, but there's no standard for warning, and it's very safe to assume it would be ignored.

I think it's particularly ok in the case of receipts, because there's a strong assumption they're checked by the client quite quickly (waiting until audit time to do so is extremely obviously flawed).

PR #3687, opened 2022-03-22
Move split ledger test

wintersteiger:
When running the recovery tests more than 3 times with ledger splitting enabled, it frequently happened that the ledger files on disc went out of sync. We didn't see this in CI runs because we only ever ran 3 or 4 recoveries with splitting enabled for some of them. That test is now moved to e2e_operations, where we check that the split ledger can be read successfully and the recovery tests run without splitting ledgers.

### DIFF
--- a/tests/e2e_operations.py
+++ b/tests/e2e_operations.py
@@ -199,7 +199,9 @@ def test_split_ledger_on_stopped_network(primary, args):
         )
 
     # Check that the split ledger can be read successfully
-    ccf.ledger.Ledger([current_ledger_dir] + committed_ledger_dirs, committed_only=False)
+    ccf.ledger.Ledger(
+        [current_ledger_dir] + committed_ledger_dirs, committed_only=False
+    )
 
 
 def run_file_operations(args):
--- a/tests/e2e_operations.py
+++ b/tests/e2e_operations.py
@@ -199,7 +199,7 @@ def test_split_ledger_on_stopped_network(primary, args):
         )
 
     # Check that the split ledger can be read successfully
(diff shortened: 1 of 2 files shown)


tests/e2e_operations.py
```diff
@@ -149,6 +149,59 @@ def test_forced_snapshot(network, args):
...
+
+    current_ledger_dir, committed_ledger_dirs = primary.get_ledger()
+    split_all_ledger_files_in_dir(current_ledger_dir, current_ledger_dir)
+    if committed_ledger_dirs:
+        split_all_ledger_files_in_dir(
+            committed_ledger_dirs[0], committed_ledger_dirs[0]
+        )
+
+    # Check that the split ledger can be read successfully
+    ccf.ledger.Ledger([current_ledger_dir] + committed_ledger_dirs, False)
```
jumaffre, line 202:
```suggestion
    ccf.ledger.Ledger([current_ledger_dir] + committed_ledger_dirs, committed_only=False)
```

tests/recovery.py
```diff
@@ -489,9 +441,7 @@ def run(args):
                     network, args, from_snapshot=False
                 )
             else:
-                network = test_recover_service(
-                    network, args, from_snapshot=False, split_ledger=True
-                )
+                network = test_recover_service(network, args, from_snapshot=False)
 
             for node in network.get_joined_nodes():
                 node.verify_certificate_validity_period()
```
jumaffre, line 447: Can you also remove the comment on line `496` and the test-specific arguments that were only set because of this?

wintersteiger: Of course, thanks for pointing that out!

PR #3689, opened 2022-03-23
Another run at partition stability

eddyashton:
Correction to #3672.

The large comment is correct, but the code doesn't match this. Calling `wait_for_new_primary` in a loop doesn't work when `b_0` is fully capable of winning multiple elections in a row - the view increments, but the primary ID never changes. I thought about adding a `wait_for_view_increment`, but I think the existing `wait_for_primary_unanimity` does what we need.

I'm running this in a loop locally, and no issues so far, but I did the same with the previous approach. I'll increase the loop iterations.


tests/partitions_test.py
```diff
@@ -114,18 +114,13 @@ def test_isolate_primary_from_one_backup(network, args):
...
-        new_primary, new_view = network.wait_for_new_primary(
-            old_primary, nodes=backups, timeout_multiplier=6
-        )
-        if new_primary == b_1:
-            break
-        else:
-            old_primary = new_primary
-
+    new_primary = network.wait_for_primary_unanimity(
+        min_view=initial_txid.view, timeout_multiplier=30
```
achamayou, line 118: 30 seems very large?

achamayou (approved):

tests/partitions_test.py
eddyashton: It's lower than the 10*6 we potentially had last time. In practice, we often converge much faster, but I have seen ~60s waits in local runs. But if we hit the unlucky timing where 2 very-close elections cancel each other out (why we use 6 for a 'normal' timeout), this still lets us do 5 such elections in a row.

Issue #3677, opened 2022-03-18
OpenSSL bad error handling


jumaffre:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=43347&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=3aae4c9d-f0dc-5b97-1b0e-f86312395af9&l=15373

```
OSSL: Could not encode update to base64 string: error:0D08303A:lib(13):func(131):reason(58)
51: terminating with uncaught exception of type std::exception: std::exception
```

jumaffre:
Linking to @eddyashton's investigation branch: https://github.com/eddyashton/CCF/commits/more_openssl_error_code_checks

rengolin-ms:
It was also mentioned here: #3368

jumaffre:
@rengolin-ms Do you happen to have a link to another instance of a similar failure but for another test in the CI?

eddyashton:
@jumaffre The one at the bottom of #3368 is `tpcc_sgx_cft`, and includes _a_ OSSL error. I _think_ the earlier `e2e_logging_cft` failures in that test are also OSSL, but we don't have access to those logs.

rengolin:
I've rebased @eddyashton branch on current master here: https://github.com/rengolin/CCF/tree/more_openssl_error_code_checks

achamayou:
Not observed in quite some time, will re-open if necessary.

PR #3690, opened 2022-03-23
Improve base64 error handling

rengolin:
This is an attempt to understand #3677, or at least be certain that, if we
still get errors, it cannot be a leak from an earlier bad call.


src/crypto/openssl/base64.h
```diff
@@ -83,8 +91,7 @@ namespace crypto
       auto err = ERR_get_error();
       if (err != 0)
       {
-        char err_str[256];
-        ERR_error_string(err, err_str);
+        auto err_str = OpenSSL::error_string(ERR_get_error());
         throw std::logic_error(fmt::format(
           "OSSL: Could not encode update to base64 string: {}", err_str));
```
eddyashton, line 96: While we're here, we should make this exception more verbose, to help identify the issue if it repros only on the CI. ie - print `size`, `max_size`, `chunk_len`, maybe `data`, etc.

achamayou (approved):

PR #3694, opened 2022-03-24
Define clock_gettime rather than gettimeofday

achamayou:
Tested against https://github.com/letmaik/ccf-repro-1


achamayou:
Ok, so while that links, we get the OCall...

Issue #3685, opened 2022-03-21
Cannot build C++ app with t_cose library: multiple definition of `gettimeofday'


letmaik:
**Describe the bug**

> /opt/oe_lvi/ld: /opt/ccf/lib/libquickjs.enclave.a(stub_time.c.o): in function `gettimeofday':
/__w/1/s/build/../src/enclave/stub_time.c:7: multiple definition of `gettimeofday'; /opt/openenclave/lib/openenclave/enclave/liboelibc.a(gettimeofday.c.o):/source/openenclave/3rdparty/musl/musl/src/time/gettimeofday.c:6: first defined here

**To Reproduce**
Use `ccfciteam/ccf-app-ci:2.0.0-rc4` as base image and try to build a CCF app.

**Expected behavior**
It should build without adding further flags that would allow multiple definitions.

**Additional context**
The full invocation:
[...]

PR #3695, opened 2022-03-24
Tweak stub to avoid symbol conflict

achamayou:
Another stab at a solution for #3685 

- [x] Links correctly in https://github.com/letmaik/ccf-repro-1
- [x] No symbols for `gettimeofday` or `localtime_r` in `libquickjs.enclave.a`


letmaik (approved):

3rdparty/exported/quickjs/quickjs.c
```diff
@@ -48,6 +48,8 @@
 #include "libbf.h"
 #endif
 
+#include "quickjs-time.h"
```
jumaffre, line 51: Does this modify the actual QuickJS source code? This seems like something we'll forget to do when we next upgrade QuickJS?

achamayou: Yes, and yes, but there are only so many solutions to this problem and we don't update quickjs often, since it's not released often.

eddyashton (approved):

PR #3699, opened 2022-03-24
Add jobs to docs banner

eddyashton:
We're currently hiring for CCF, so we should advertise that to anyone using the docs.


achamayou (approved):

jumaffre (approved):

jumaffre:
We could also remove the 2.0 announcement to keep the banner minimal, and only put it back on `2.0.0` final is out.

PR #3579, opened 2022-02-22
Fix DNS resolution memory leak

wintersteiger:
With hindsight, this was a completely obvious memory leak...

Fixes #3553

PR #3702, opened 2022-03-25
Attempt at fixing the same DNS::resolve leak as before

### LINKED ISSUE #3693: Leak in committable suffix test
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=43595&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=ae32e15b-0191-5f39-7410-be7e80165871

```
47: =================================================================
47: ==6941==ERROR: LeakSanitizer: detected memory leaks
47: 
47: Direct leak of 64 byte(s) in 1 object(s) allocated from:
47:     #0 0x49da7d in malloc (/__w/1/s/build/cchost+0x49da7d)
47:     #1 0x7f26e7562d79  (/lib/x86_64-linux-gnu/libc.so.6+0x106d79)
47:     #2 0x7f26e75650d8 in getaddrinfo (/lib/x86_64-linux-gnu/libc.so.6+0x1090d8)
47:     #3 0x44cdca in getaddrinfo (/__w/1/s/build/cchost+0x44cdca)
47:     #4 0x7f26e7b6d5fc  (/lib/x86_64-linux-gnu/libuv.so.1+0x165fc)
47: 
47: SUMMARY: AddressSanitizer: 64 byte(s) leaked in 1 allocation(s).
```

Almost certainly related to #3553 and #3579.

rengolin:
Maybe this is the same thing that happened in #3579?

Does that fix #3693?

### DIFF
--- a/src/host/tcp.h
+++ b/src/host/tcp.h
@@ -222,7 +222,16 @@ namespace asynchost
     static void on_client_resolved(
       uv_getaddrinfo_t* req, int rc, struct addrinfo*)
     {
-      static_cast<TCPImpl*>(req->data)->on_client_resolved(req, rc);
+      if (req->data)
+      {
+        static_cast<TCPImpl*>(req->data)->on_client_resolved(req, rc);
+      }
+      else
+      {
+        // The TCPImpl that submitted the request has been destroyed, but we
+        // need to clean up the request object.
+        delete req;
+      }
     }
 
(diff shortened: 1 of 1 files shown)


wintersteiger:
I don't think that's a problem, because the client side always runs synchronously, so by the time it returns from `on_client_resolved` the request is already handled. It may still be a good idea to keep that code path though, just in case someone ever wants an asynchronous client.

PR #3707, opened 2022-03-25
Add generic load testing to end-to-end infra

jumaffre:
Something that had been on my mind for a while now: we have existing 1) end-to-end test suites that test all the interactions users/operators/members can have with CCF and 2) stress testing of a stable service. This PR introduces a new `ServiceLoad` class that hammers the service at a given rate while the rest of the existing test suite is being performed. This should help us reproduce errors that are seen on a live deployments while the service is under load, and hopefully catch others early, before release.

To do this, we use ~`vegeta`~ `locust` (https://github.com/locustio/locust), which is started in a separate thread. When a network event happens (e.g. primary change, a node is added or killed), we restart `locust` with the new target RPC addresses. At each restart, we stitch the `.csv` output files together so that at the end of the test run, we end up with a single `.csv` file which can be rendered by `matplotlib` (latency and errors over time, with indications on when network events happened). 

Output image, to be found in `workspace/test_common/load_results.png`:

image

TODO:
- [x] Cleanup
- [x] More precise event reporting
- [x] Add title to rendered diagram
- [x] Add HTTP error code to graph
- [x] More precise timing of network events if possible
- [ ] Plugin to existing test suites (see #3740)
- [x] Fix issue when waiting for snapshots

Note that the load is currently limited to 100 requests/second for now (not to overwhelm nodes when recovering the ledger/joining without snapshots) but I'm planning to increase this in the future.


wintersteiger:
This is awesome! Not only is it great to catch performance bugs, but also helps me to generally understand better how the system behaves under load.

tests/infra/network.py
```diff
@@ -1178,31 +1201,29 @@ def wait_for_snapshot_committed_for(self, seqno, timeout=3):
         return self.wait_for_commit_proof(primary, snapshot_evidence_seqno, timeout)
 
     def get_committed_snapshots(self, node):
-        # Wait for all available snapshot files to be committed before
-        # copying snapshot directory, so that we always use the latest snapshot
+        # Wait for the snapshot including target_seqno to be committed before
```
jumaffre, line 1204: This logic assumed that no other transaction were running concurrently. Obviously, this isn't true anymore.

rengolin:
Why are there `start`s without `stop`s? Having `stop[12]` without `start[12]` doesn't help knowing to which start it belongs to.

rengolin:
The `req/s` axis has a different rate than the `error/s` one, making it hard to compare. I imagine they're independent and dynamically set by the tool, but it'd be nice to fix them beforehand with `max(req/s, error/s)` for both.

rengolin:
Finally, is the bottom X axis wall clock time? Perhaps better if that was "time since start" instead? It's easier to read duration times that way.

jumaffre:
> Finally, is the bottom X axis wall clock time? Perhaps better if that was "time since start" instead? It's easier to read duration times that way.

I thought about this too but I think it's better to stick to clock time since the existing tests output the current time, so that we can easily correlate the performance results with the test exection.

jumaffre:
> Why are there `start`s without `stop`s? Having `stop[12]` without `start[12]` doesn't help knowing to which start it belongs to.

Good point. `start` is to indicate that the load client has started, whereas `stop p[12]` is to indicate that the primary node `12` was stopped. I rename `start` to `run` so that this is clearer.

tests/infra/network.py
```diff
@@ -142,13 +146,23 @@ def __init__(
             self.next_node_id = 0
             self.txs = txs
             self.jwt_issuer = jwt_issuer
+            self.service_load = (
+                infra.service_load.ServiceLoad(self) if with_load else None
```
achamayou, line 150: It'd be more flexible to pass an optional ServiceLoad than a boolean flag to the network.

tests/infra/service_load.py
```diff
@@ -0,0 +1,258 @@
...
+                    LOG.warning(
+                        "Network configuration has changed, restarting service load client"
+                    )
+                    event = ""
+                    if primary not in new_nodes:
+                        event = f"stop p[{primary.local_node_id}]"
+                    elif new_primary != primary:
+                        event = f"elect p[{primary.local_node_id}] -> p[{new_primary.local_node_id}]"
+                    else:
+                        added = list(set(new_nodes) - set(known_nodes))
```
achamayou, line 242: Why list()? set() is coercible to bool and iterable, no?

achamayou (approved):

tests/infra/network.py
jumaffre: As discussed, this is now done.

PR #3720, opened 2022-03-29
Better Python Ledger iterators

wintersteiger:
Mainly to allow multiple iterations over a ledger.

### DIFF
--- a/python/ccf/ledger.py
+++ b/python/ccf/ledger.py
@@ -849,7 +849,7 @@ class Ledger:
 
     _filenames: list
     _insecure_skip_verification: bool = False
-    _latest_iterator: LedgerIterator = None
+    _latest_iterator: Optional[LedgerIterator] = None
 
     def __init__(
         self,
@@ -990,7 +990,7 @@ class Ledger:
 
         :return int: Number of verified signature transactions.
         """
-        return self._latest_iterator.signature_count()
+        return self._latest_iterator.signature_count() if self._latest_iterator else 0
 
     def last_verified_txid(self) -> Optional[TxID]:
         """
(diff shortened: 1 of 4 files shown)


python/ccf/ledger.py
```diff
@@ -803,6 +803,43 @@ def get_seqnos(self):
...
+    _filenames: list
+    _fileindex: int
+    _current_chunk: LedgerChunk
+    _ledger_validator: Optional[LedgerValidator] = None
+
+    def __init__(
+        self, filenames: list, ledger_validator: Optional[LedgerValidator] = None
+    ):
+        self._filenames = filenames
+        self._fileindex = -1
```
achamayou, line 816: Init at line 808?

python/ccf/ledger.py
```diff
@@ -803,6 +803,43 @@ def get_seqnos(self):
...
+        self._fileindex += 1
+        if len(self._filenames) > self._fileindex:
+            self._current_chunk = LedgerChunk(
+                self._filenames[self._fileindex], self._ledger_validator
+            )
+            return self._current_chunk
+        else:
+            raise StopIteration
+
+    def signature_count(self) -> int:
```
achamayou, line 829: This isn't great, it's not possible to distinguish a ledger that legitimately has no signatures, versus one that does but the validator wasn't set. I think we should remove this call and force the caller to deal with by exposing the validator, or raise an exception.

achamayou: Or possibly make it optional like last_verified_txid below, but avoid overlap between "no value" and 0.

achamayou (approved):

jumaffre (approved):

PR #3722, opened 2022-03-30
Trigger snapshot at end of recovery

### LINKED ISSUE #3665: Force snapshot on recovery completion
As discussed with @wintersteiger, now that we have #3624, it is possible that a recovered service (`N+1`, that started from a snapshot emitted by the previous service `N`) does not process a sufficient number of transactions of generate a new snapshot. If the service `N+1` crashes, then the new service `N+2`, that starts from the latest generated snapshot emitted by service `N`, will not be able to verify the snapshot if the `previous_service_cert_file` specified by the operator on startup if the one of the latest service, i.e. `N+1`.

To reduce the possibility of this happening, we should force the generation of a snapshot when the private recovery is complete. This would mean that the service `N+2` would start from the latest generated snapshot, created by service `N+1`.

In fact, we should probably also force the generation of a snapshot when the initial service is open.

wintersteiger:
Fixes #3665

### DIFF
--- a/src/node/node_state.h
+++ b/src/node/node_state.h
@@ -1152,16 +1152,16 @@ namespace ccf
           throw std::logic_error("Service could not be opened");
         }
 
+        // Trigger a snapshot (at next signature) to ensure we have a working
+        // snapshot signed by the current (now new) service identity, in case we
+        // need to recover soon again.
+        trigger_snapshot(tx);
+
         if (tx.commit() != kv::CommitResult::SUCCESS)
         {
           throw std::logic_error(
             "Could not commit transaction when finishing network recovery");
         }
-
-        // Trigger a snapshot (at next signature) to ensure we have a working
-        // snapshot signed by the current (now new) service identity, in case we
-        // need to recover soon again.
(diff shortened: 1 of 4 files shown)


achamayou (approved):

achamayou:
Although the change looks reasonable at first glance, it seems to have a massive impact on memory consumption!

src/node/node_state.h
```diff
@@ -1372,6 +1377,7 @@ namespace ccf
 
         GenesisGenerator g(network, tx);
         g.open_service();
+        trigger_snapshot(tx);
```
eddyashton, line 1380: I'm guessing this is the perf hit. This adds a snapshot on every service open, and the perf tests probably didn't produce any snapshot previously.

wintersteiger: Yeah, that's a good point! I would have expected that impact to be less severe, but you may be right.

wintersteiger:
Not that the logic is correct, it looks like the performance also improved to a level where I'm happy with it.

Issue #3692, opened 2022-03-23
Document and test behavior for historical queries when old ledger files are not yet available


lynshi:
**Is your feature request related to a problem? Please describe.**
Currently it is possible (and often desirable) to start a node from snapshot without old ledger files present. Ledger files will eventually be copied to the node's read-only ledger directory, but this allows nodes to be added faster for strengthening consensus and/or servicing requests.

It is possible for a new node to receive historical queries before the old ledger files have been copied. We believe the behavior will be that CCF will drop the query, and subsequent requests will be processed as if they were the first (i.e. as if a new historical query was just started, instead of there being one currently running).

**Describe the solution you'd like**
As discussed with @jumaffre, the behavior in such a situation should be well documented and tested.

PR #3728, opened 2022-03-31
Initial end-to-end tests simulating host tampering, impact on historical queries

eddyashton:
Part of #3692.

The current tests include truncation and removal of ledger files. I have a test that corrupts ledger files, but this fails unexpectedly - I'll investigate that separately.

I haven't yet included tests that corrupt index files. The tricky thing here is that we need to create a huge volume of transactions before the index is flushed to disk, rather than kept in-memory. I've reduced the sample app's indexing thresholds so the files are at least written, but they're not relied on until we fill all buckets. This may already be a perf hit that needs to be removed, and I think to support this properly we'll need to make the index thresholds adjustable per-node, which relies on some other work (app-specific configuration).


tests/tamper_robustness.py
```diff
@@ -0,0 +1,274 @@
...
+def modify_file(src_path):
+    with tempfile.TemporaryDirectory(dir=os.path.dirname(src_path)) as tmp_dir_name:
+        dst_path = os.path.join(tmp_dir_name, os.path.basename(src_path))
+        LOG.warning(f"Temporarily modifying file {src_path} (backed up at {dst_path})")
+        shutil.copy(src_path, dst_path)
+        with open(src_path, "rb+") as f:
+            content = list(f.read())
+            size = len(content)
+            for idx in (size // 3, size // 2, int(size // 1.5)):
+                content[idx] //= 2
```
wintersteiger, line 129: 0 stays 0? If the files are big enough it wouldn't make much of a difference I guess. Maybe `content[idx] ^ 0b01010101`?

wintersteiger (commented):
LGTM!

jumaffre (approved):
The new test seems to take ~90 secs. Would it be worth running it in parallel with `recovery.run_corrupted_ledger` so that it doesn't impact the total test duration and also groups these two similar tests in the same file?

raghavthind2005 (approved):

eddyashton:
Closing this PR as it has rotted, and the key work is covered by #4149.

PR #3731, opened 2022-04-01
Restore public binary `TxReceipt` type

eddyashton:
There's 2 design choices in here that warrant further discussion. I'll describe in comments for increased visibility.


src/node/history.h
```diff
@@ -235,14 +236,30 @@ namespace ccf
...
+            ccf::TxReceiptPathStep::Left :
+            ccf::TxReceiptPathStep::Right;
+          const auto hash = crypto::Sha256Hash::from_span(
+            {node.hash.bytes, sizeof(node.hash.bytes)});
+          ret.push_back({direction, hash});
+        }
+      }
+
+      return ret;
     }
```
eddyashton, line 263: `TxReceipt` previously contained types from `merklecpp`. I don't think we want to expose those in the public API, and instead want to reuse our existing `Sha256Hash` type. That means we need to convert from the library representation to the public CCF API representation at some point (done here, on demand, in every call), and separately still have another conversion to a JSON-friendly format for those that want it. That's a little confusing, and potentially inefficient (if we `get_` far more than any other call, we should convert earlier). But I think it's the right choice - we don't want an additional hash type, and a duplicated `Path` type is cheap.

include/ccf/historical_queries_interface.h
```diff
@@ -7,17 +7,13 @@
...
 
 namespace ccf
 {
-  // This is an opaque, incomplete type, but can be summarised to a
-  // JSON-serialisable form by the functions below
-  struct TxReceipt;
-  using TxReceiptPtr = std::shared_ptr<TxReceipt>;
-
   ccf::Receipt describe_receipt(
     const TxReceipt& receipt, bool include_root = false);
```
eddyashton, line 18: Naming is hard, as this function signature makes clear. I think we should rename either `ccf::Receipt` or `ccf::TxReceipt`. The former is harder to rename because it's already public. Maybe the latter should be `BinaryReceipt`, or `RawReceipt`, or something, to highlight it contains directly useable hash byte arrays, rather than a hex/b64 encoding?

src/node/tx_receipt.h
```diff
@@ -3,15 +3,15 @@
 #pragma once
 
 #include "ccf/receipt.h"
-#include "node/history.h"
+#include "ccf/tx_receipt_path.h"
 
 namespace ccf
 {
   struct TxReceipt
```
wintersteiger, line 10: Is it necessary that we keep this class/file? Looks like a copy of the public one to me.

include/ccf/historical_queries_interface.h
wintersteiger: I think `ccf::Receipt` should be the binary type name for a C++ API, but since it's been public already that's hard to do :-/ SerializedReceipt, B64Receipt, StringReceipt?

wintersteiger (approved):

src/node/tx_receipt.h
eddyashton: Good spot, this should have been deleted.

include/ccf/tx_receipt_path.h
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ccf/crypto/sha256_hash.h"
+
+#include <vector>
+
+namespace ccf
+{
+  struct TxReceiptPathStep
```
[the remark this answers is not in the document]
eddyashton: Nice. I'd thought the `PathStep` needed to be separately visible, but I think that was because of an include order stemming from the should-have-been deleted copy noted below. I've now removed this file, and rolled this inside `TxReceipt`.

include/ccf/historical_queries_interface.h
eddyashton: I've gone back and forth on this, and I'm going to rename the current `ccf::Receipt` type. It seems clear that the old names are the wrong way round, and I'd rather bite the bullet early rather than later.

wintersteiger: +1 that's what I would do too, I hope it won't trigger too much confusion!

raghavthind2005 (approved):

eddyashton:
Closing, superceded by #3793.

PR #3732, opened 2022-04-01
Account for timezone in certificate validity period

jumaffre:
When calculating the validity period for node/service certificates, we did not account for the local time zone when converting the `ASN1_TIME` to `time_t` (using `std::mktime`). This resulted in the sandbox throwing an exception when verifying the validity period for the service certificate, e.g. using the `Asia/Singapore` timezone:

```
$ sandbox.sh ...
ValueError: Validity period for service certificate is not as expected: valid to "2022-04-01 17:52:20" but expected "2022-04-02 09:52:20"
```

This is now fixed (passing the `TZ=Asia/Singapore` envvar to the node):
```
$ sandbox.sh -p ...
...

$ TZ=Asia/Singapore date
Fri Apr  1 21:16:24 +08 2022

$ openssl x509 -in ./workspace/sandbox_common/0.pem -text -noout
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            a3:e1:54:41:0c:aa:a6:2e:03:4a:ca:85:e2:bd:db:ad
        Signature Algorithm: ecdsa-with-SHA384
        Issuer: CN = CCF Node
        Validity
            Not Before: Apr  1 13:16:04 2022 GMT
            Not After : Apr  2 13:16:03 2022 GMT
```

This is now fixed as we account for the local time zone when converting the `ASN1_TIME`. I also moved the validity period check out of the generic `network.py`. 

Also removed the dependency on `crypto` from `cchost`.


src/crypto/openssl/x509_time.h
```diff
@@ -36,7 +36,8 @@ namespace crypto::OpenSSL
   {
     tm tm_time;
     CHECK1(ASN1_TIME_to_tm(time, &tm_time));
-    return std::mktime(&tm_time);
+    // Adjust for timezone (timezone global defined in time.h)
+    return std::mktime(&tm_time) - timezone;
```
achamayou, line 40: That does not seem right? I think we want to take UTC inputs, and it's the validation that should change instead.

wintersteiger: I wonder whether we shouldn't just keep an `ASN1_TIME` for those times and use `ASN1_TIME_print` to convert to strings to avoid `std::mktime`. There's also `ASN1_UTCTIME_adj` which wouldn't require us to make any timezone adjustments I think.

jumaffre: As discussed, going for `timegm()` for now. There must be a way to do this with pure C++ standard library function calls to avoid the conversion to `time_t` required by the `ASN1_TIME_adj` function.

wintersteiger (approved):

PR #3736, opened 2022-04-04
Pin `testssl` to v3.0.7

eddyashton:
We were previously pulling latest, which meant using the default `3.1dev` branch. As a result we get unstable changes like the re-ordering which broke the Daily Build today.

Instead we pin to a specific `v3.0.7`. We should remember to check this dependency when updating our other dependencies, and take `v3.1` when it is released.


CMakeLists.txt
```diff
@@ -668,7 +668,7 @@ if(BUILD_TESTS)
       OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/testssl/testssl.sh
       COMMAND
         rm -rf ${CMAKE_CURRENT_BINARY_DIR}/testssl && git clone --depth 1
-        https://github.com/drwetter/testssl.sh
+        --branch v3.0.7 --single-branch https://github.com/drwetter/testssl.sh
```
jumaffre, line 671: Should we simply download `https://github.com/drwetter/testssl.sh/releases/latest` (in a similar vein than what we do for our non-pinned Python/Node deps) and fix it when it occasionally breaks? For a dependency that's aimed to test the security of the TLS layer, having the very latest version always seems like a win to me.

eddyashton: We could, but that's a release (/redirect link), rather than a clonable git branch. So we could follow the `/latest` link to find the release page, and either parse that for a tag to clone, or download the release binary and use that. But I think those are both larger changes, so I'll punt on them for now.

Issue #3613, opened 2022-02-28
Use ledger entry flags to synchronise ledger chunks on all nodes


jumaffre:
Now that we have the ability to set flags on a ledger entry to trigger the generation of a new ledger chunk on the host-side, we should use this feature to trigger the generation of a new ledger chunk on the host when a new snapshot is created, rather than assuming that all nodes have been started with the same `snapshots.tx_count` and their ledger chunking sync'ed up that way.

This would allow services recovered with a different snapshot `tx_count` to maintain the ledger chunk consistency on all the nodes for the historical ledger.

Note that this will probably require #3026 (moving the `Snapshotter` to the `kv::store`) first.

PR #3737, opened 2022-04-04
Move snapshotter out of consensus

wintersteiger:
Implements #3026


src/kv/snapshotter.h
```diff
@@ -7,16 +7,18 @@
 #include "consensus/ledger_enclave_types.h"
 #include "ds/thread_messaging.h"
 #include "kv/kv_types.h"
+#include "kv/snapshot_serdes.h"
```
jumaffre, line 10: I think it'd be better to leave the snapshotter out of `kv/`. The kv component only needs to know about the `AbstractSnapshotter` type, which is implemented by the `ccf::Snapshotter`.

src/consensus/aft/raft_types.h
```diff
@@ -92,6 +93,16 @@ namespace aft
       }
       return nullptr;
     }
+
+    virtual bool record_committable(Index idx) override
```
jumaffre, line 97: This isn't necessary. Since the `kv::Store` has now access to the snapshotter directly, it can record a committable index when it `commit`s a `globally_committable` entry.

wintersteiger: There are two calls (1 leader, 1 backup) to `record_committable` left in `raft.h` and both of them are used to determine whether a ledger chunk hast to be forced, i.e. this goes into `ledger->put_entry`, so the ledger would have to have a reference to the store or snapshotter. I think I know how to do this from the store before getting into raft, but I might break some stuff before it starts working again :)

src/kv/kv_types.h
```diff
@@ -680,6 +684,8 @@ namespace kv
     // Thus, a large rollback is one which did not result from the map creating
     // issue. https://github.com/microsoft/CCF/issues/2799
     virtual bool should_rollback_to_last_committed() = 0;
+
+    bool force_ledger_chunk = false;
```
wintersteiger, line 688: @jumaffre I hate having to modify `AbstractChangeSet`, but somehow I need this one bit of information to be carried from `ds->apply()` to `ledger->put_entry(... ds->force_ledger_chunk ... )` in `raft.h`. The only alternative I can think of is to add a flag to `aft::Store`, but that's not much of a difference. Do you have a better idea?

jumaffre: As discussed, this will go anyway once https://github.com/microsoft/CCF/issues/3613 is implemented.

jumaffre: As discussed, this will go anyway once https://github.com/microsoft/CCF/issues/3613 is implemented.

src/node/node_state.h
```diff
@@ -898,7 +901,10 @@ namespace ccf
             // Inform snapshotter of all signature entries so that this node can
             // continue generating snapshots at the correct interval once the
             // recovery is complete
-            snapshotter->record_committable(last_recovered_idx);
+            if (snapshotter->record_committable(last_recovered_idx))
+            {
```
jumaffre, line 905: This isn't necessary here and will be remove altogether once https://github.com/microsoft/CCF/issues/3613 is implemented. The `GenesisGenerator` on line `857` may be removable too as unused?

src/node/node_state.h
```diff
@@ -2265,20 +2271,30 @@ namespace ccf
...
           }));
 
-      network.tables->set_map_hook(
+      // Note: The Signatures hook and SerialisedMerkleTree hook are separate
+      // because the signature and the Merkle tree are recorded in distinct
+      // tables (for serialisation performance reasons). However here, they are
+      // expected to always be called together and for the same version as they
+      // are always written by each signature transaction.
+
+      network.tables->set_global_hook(
```
jumaffre, line 2280: This should be `set_map_hook()` and return `return kv::ConsensusHookPtr(nullptr);`.

src/node/snapshotter.h
```diff
@@ -191,7 +193,7 @@ namespace ccf
 
       for (auto it = pending_snapshots.begin(); it != pending_snapshots.end();)
       {
-        if (idx > it->evidence_idx)
+        if (idx > it->evidence_idx && it->sig.has_value())
```
jumaffre, line 196:
```suggestion
        if (idx > it->evidence_idx)
```

This can go once the hook become local commit hooks (`set_map_hook`).

jumaffre (approved):

Issue #2468, opened 2021-04-14
js: Date.now() returns 0


letmaik:
Trying to get insecure host time in JS via `Date.now()` returns `0`.

I'm not sure why this is happening. quickjs uses `gettimeofday()` which is listed as supported in the [OE docs](https://github.com/openenclave/openenclave/blob/master/docs/LibcSupport.md) (time is part of the [default OCALLs](https://github.com/openenclave/openenclave/blob/52b065d0b8fb652d8c919a47a99411aa46ab9f6d/include/openenclave/internal/calls.h#L86)).

EDIT: quickjs also uses `localtime_r()`
EDIT2: For this to work reliably, time would have to be "deterministic", meaning it is constant within a transaction and stored, such that re-doing a transaction (in case of conflict or similar) uses the same time.

eddyashton:
You need to opt in to get this OCALL from OE, and we don't (because we don't want surprising OCALLs in the middle of our code). We provide a stub `gettimeofday()` in `stub_time.c`. We do have time from the host (`enclave_time.h`) that we _could_ use here, but converting that to standard time is not yet done. And in general we don't want to pretend we have trusted time, and its one of the (many) bits of a 'standard' programming environment that we generally want to tell people "your in an enclave - this doesn't do what you expect".

letmaik:
After thinking about it a bit more I think exposing a separate function is ok. Any library that validates time always accepts a specific time as well (e.g. jsrsasign). Date/time libraries like Moment.js also accept a custom time to be parsed and then further modified. `ccf.getUntrustedTime()` returning a UNIX timestamp seems good enough.

eddyashton:
I think enforcing a single time value to reduce non-determinism is the wrong approach.

Apps in both C++ and JS can already access non-deterministic things in their execution (the currently executing node ID, calling `random()`, surely many others). These will break in BFT, but I think the fix for that has to be (a) an explanation that you _can't_ put these directly into the KV in a Byzantine configuration, with (b) some way to capture those non-deterministic values and access them in re-execution. But even when we can capture them, I think hiding those non-deterministic values behind existing APIs is going to be more work than its worth (with difficulties like that noted above of no context object, requiring an implementation that reads from some thread-local pre-stored value?).

[...]

PR #3739, opened 2022-04-05
Improve handling of X509 times

wintersteiger:
This simplifies the way we handle x509 certificate expiry times. The idea is to do as much as possible in C++ chrono and to convert to/from OpenSSL types only when we are about to generate/verify a certificate. Everything is normalized to `std::chrono::system_clock`, which is in UTC.


src/node/historical_queries_adapter.cpp
```diff
@@ -298,7 +298,7 @@ namespace ccf::historical
               hpubkey,
               ReplicatedNetworkIdentity::subject_name,
               {},
-              ds::to_x509_time_string(crypto::OpenSSL::to_time_t(NULL)),
+              ds::to_x509_time_string(std::chrono::system_clock::now()),
```
jumaffre, line 301: What does this return in an SGX enclave?

wintersteiger: The current time, which I believe it gets from the host. I haven't checked that, but I figured if `time(NULL)` works, this should too. I'll take a closer look later today.

jumaffre: https://github.com/microsoft/CCF/issues/2468 may give you some more context on OCALLs and time. Maybe we could simply extract the validity period from the historical certificate and use it in the newly-endorsed certificate?

wintersteiger: Yeah, that's a good idea, we really shouldn't depend on the host time anywhere. I would rather take the times from the node cert though, because that's the endorser.

tests/reconfiguration.py
```diff
@@ -400,9 +400,7 @@ def test_join_straddling_primary_replacement(network, args):
                 "name": "transition_node_to_trusted",
                 "args": {
                     "node_id": new_node.node_id,
-                    "valid_from": str(
-                        infra.crypto.datetime_to_X509time(datetime.now())
-                    ),
+                    "valid_from": str(datetime.now()),
```
jumaffre, line 403: This is awesome! As we now accept multiple time formats, could you update https://microsoft.github.io/CCF/main/governance/common_member_operations.html?highlight=asn1#renewing-node-certificate to mention this and add an entry to the changelog?

doc/governance/common_member_operations.rst
```diff
@@ -153,7 +153,7 @@ Renewing Node Certificate
 
 To renew the soon-to-be-expired certificate of a node, members should issue a ``set_node_certificate_validity`` proposal, specifying the date at which the validity period of the renewed certificate should start (``valid_from``), as well as its validity period in days (``validity_period_days`` -- optional).
 
-- The ``valid_from`` date argument should be a ASN1 UTCTime string, i.e. ``"YYMMDDhhmmssZ"``.
+- The ``valid_from`` date/time argument accepts time points in ASN.1 UTCTime format (``"YYMMDDhhmmssZ"``) or in the Python-default ISO 8601 format (``"YYYY-MM-DD HH:MM:SS.ssssss+HH:MM"``), with optional fractional seconds and timezone offset. For details see :ccf_repo:`src/ds/x509_time_fmt.h`.
```
jumaffre, line 156: Is the ISO 8601 format that is now supported Python specific? If so, it's probably better to omit that detail here. Also, I think the link to `x509_time_fmt.h` isn't necessary is we list all supported formats here?

tests/lts_compatibility.py
```diff
@@ -93,6 +95,9 @@ def test_new_service(
     nodes_to_cycle = network.get_joined_nodes() if cycle_existing_nodes else []
     nodes_to_add_count = len(nodes_to_cycle) if cycle_existing_nodes else 1
 
+    # Pre-2.0 nodes require X509 time format
+    valid_from = str(infra.crypto.datetime_to_X509time(datetime.datetime.now()))
```
jumaffre, line 99: This is surprising because the dynamic certificate validity range feature was added in 2.0?

jumaffre (approved):

tests/lts_compatibility.py
wintersteiger: All those changes were prompted by test failures, so it's definitely required, but it may have been a 2.0-rcX network

doc/governance/common_member_operations.rst
wintersteiger: No, ISO 8601 is an ISO standard and it's not one specific format, but a class of them, and Python chooses one of them as their default (e.g. `T` after date, before time is omitted, timezone offset is optional...). I'll remove `Python-default` here. 

We don't list all formats here as there are too many optional elements, so I figured I'd just add a link to the code.

tests/lts_compatibility.py
wintersteiger: Also! The 2.0 checks are also run on 1.x networks in the LTS tests, so they need to support both.

src/ds/x509_time_fmt.h
```diff
@@ -3,16 +3,93 @@
...
+    }
+
+    // Then there are formats that strptime doesn't support...
+    std::vector<std::pair<const char*, int>> more_formats = {
+      // Note: longest format to match first
+      {"%04u-%02u-%02u %02u:%02u:%f %d:%02u", 8},
+      {"%04u-%02u-%02uT%02u:%02u:%f %d:%02u", 8},
+      {"%04u-%02u-%02u %02u:%02u:%f %03d %02u", 8},
+      {"%04u-%02u-%02uT%02u:%02u:%f", 6},
+      {"%04u-%02u-%02u %02u:%02u:%f", 6}};
```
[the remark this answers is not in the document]
wintersteiger: Sure, np.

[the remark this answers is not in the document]
wintersteiger: Actually, that's already covered by `"%04u-%02u-%02uT%02u:%02u:%f"`. The `sscanf` patterns succeed if there is extraneous text at the end of the string. I don't like that too much either, but I don't there's an easy way to enforce reading all of the string.

tests/lts_compatibility.py
jumaffre: Oh I see, `valid_from` is automatically set as a ISO 8601 time string inside `network.trust_node()` if unset. Perhaps the comment should say "Test ASN1 UTC time string explicitly as default is ISO 8601"?

As discussed the LTS compatibility check should primarily check the compatibility with the latest 1.x so we should fix this.

eddyashton (approved):

PR #3741, opened 2022-04-05
Fix and automatically check for broken links in doc job

jumaffre:


wintersteiger (approved):
Wow, that's a few more broken links than I expected! Definitely a good idea to check those automatically.

.azure-pipelines-gh-pages.yml
```diff
@@ -28,6 +28,16 @@ jobs:
...
           cmake_args: ""
 
+      - script: |
+          set -ex
+          python3.8 -m venv env
+          source env/bin/activate
+          pip install wheel
+          pip install -U -r doc/requirements.txt
+          pip install -U -e ./python
+          sphinx-build -b linkcheck doc build/html
```
eddyashton, line 38: Can we include this in the existing single-version build, rather than as a separate step?

eddyashton (approved):

.azure-pipelines-gh-pages.yml
jumaffre: As discussed, I couldn't find a way to do this in one step because `linkcheck` is a different builder altogether.

Issue #3738, opened 2022-04-05
Accelerate recovery process by reading ledger entries range


jumaffre:
When recovering the private/public ledger, we read entries in the ledger one by one. Instead, we could read entries in batch of say, `1000` (operator configurable?) to accelerate the recovery process. 

Current implementation:
https://github.com/microsoft/CCF/blob/caf6a7c6e63ee7fc0751a234cdf15d09f9556292/src/node/node_state.h#L2253-L2261

This would require the host ledger implementation to be changed slightly to return a smaller range when reaching the end of the ledger.

PR #3746, opened 2022-04-06
Decrease recovery time by ~60% by recovering ledger in batches

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/3738

Instead of recovering entries one from the ledger one by one (and thus having the ring buffer overhead for each individual entry), we now recover the entries in batch of `1000`. 

This reduces the time of recovery by ~60% according to my local tests:
```bash
# Live service loaded with vegeta (sandbox)

$ curl https://127.1:8000/node/commit -k
{"transaction_id":"2.5467"}

# Crash service
$ cp -r ./workspace/sandbox_0/0.ledger .

# Before: ~ 15sec for recovery
$ sandbox.sh --recover --ledger-dir 0.ledger ...
[16:17:39.471] Starting 1 CCF node...
[16:17:39.471] Recovering network from:
[16:17:39.471]  - Ledger: 0.ledger/
[16:17:39.471] No available snapshot to recover from. Entire transaction history will be replayed.
[16:17:54.165] Started CCF network with the following nodes:
[16:17:54.165]   Node [0] = https://127.0.0.1:8000

# Now: ~5secs for recovery
$ sandbox.sh --recover --ledger-dir 0.ledger ...
[16:15:12.112] Starting 1 CCF node...
[16:15:12.113] Recovering network from:
[16:15:12.113]  - Common directory: /home/jumaffre/git/CCF/build/workspace/sandbox_common
[16:15:12.113]  - Ledger: 0.ledger/
[16:15:12.113] No available snapshot to recover from. Entire transaction history will be replayed.
[16:15:17.366] Started CCF network with the following nodes:
[16:15:17.367]   Node [0] = https://127.0.0.1:8000
```

This introduces a slight subtlety on the ledger host code: as the enclave now asks for batches of `1000`, the host should be able to return as many entries as it knows if it's asked to read past its `last_idx`, hence the new `strict` mode (default to `true`, unless for recovery).

### DIFF
--- a/src/consensus/ledger_enclave.h
+++ b/src/consensus/ledger_enclave.h
@@ -15,6 +15,37 @@ namespace consensus
   public:
     static constexpr size_t FRAME_SIZE = sizeof(uint32_t);
 
+    /**
+     * Retrieve a single entry, advancing offset to the next entry.
+     *
+     * @param data Serialised entries
+     * @param size Size of overall serialised entries
+     *
+     * @return Raw entry as a vector
+     */
+    static std::vector<uint8_t> get_entry(const uint8_t*& data, size_t& size)
+    {
+      auto header = serialized::peek<kv::SerialisedEntryHeader>(data, size);
+      size_t entry_size = kv::serialised_entry_header_size + header.size;
+      std::vector<uint8_t> entry(data, data + entry_size);
+      serialized::skip(data, size, entry_size);
+      return entry;
+    }
+
+    /**
+     * Skip a single entry, when backup.
+     *
+     * Does not write any entry to the legder.
+     *
+     * @param data Serialised entries
+     * @param size Size of overall serialised entries
+     */
+    static void skip_entry(const uint8_t*& data, size_t& size)
+    {
+      auto header = serialized::read<kv::SerialisedEntryHeader>(data, size);
+      serialized::skip(data, size, header.size);
+    }
+
   private:
     ringbuffer::WriterPtr to_host;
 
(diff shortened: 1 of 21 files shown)


src/node/node_state.h
```diff
@@ -1045,47 +1072,61 @@ namespace ccf
...
     //
-    void recover_private_ledger_entry(const std::vector<uint8_t>& ledger_entry)
+    void recover_private_ledger_entries(const std::vector<uint8_t>& entries)
     {
       std::lock_guard<std::mutex> guard(lock);
       if (!sm.check(NodeStartupState::readingPrivateLedger))
       {
         LOG_FAIL_FMT(
-          "Node is state {} cannot recover private ledger entry", sm.value());
+          "Node is state {} cannot recover private ledger entries", sm.value());
```
eddyashton, line 1081:
```suggestion
          "Node in state {} cannot recover private ledger entries", sm.value());
```

eddyashton (approved):

src/consensus/ledger_enclave.h
```diff
@@ -15,6 +15,23 @@ namespace consensus
...
 
+    /**
+     * Retrieve a single entry, advancing offset to the next entry.
+     *
+     * @param data Serialised entries
+     * @param size Size of overall serialised entries
+     *
+     * @return Raw entry as a vector
+     */
+    static std::vector<uint8_t> get_entry(const uint8_t*& data, size_t& size)
```
achamayou, line 26: @jumaffre could this be a std::span to reduce copies?

PR #3748, opened 2022-04-07
Fix fetching of sparse entries during indexing

### LINKED ISSUE #3744: Indexing response is incorrect about sparse entries
Potentially the cause of #3703.

I noticed this while trying to extend a unit test. If the indexing system tries to fetch a bucket that doesn't exist (for example, there were not writes to table `foo` before seqno 5,000, and we ask for a bucket containing the writes to `foo` between 500 and 1,000), the host can only tell it this file doesn't exist, and the enclave assumes this is a malicious host and begins a re-index.

For simple cases, it maybe possible to resolve this by storing first-writes (and last-writes?) in the indexing system. We already store per-key state. However I think this still needs tombstone hole-filling bucket writes, to distinguish known-empty ranges from deleted files. If we're doing that at all, I think we can do it from the start-of-time. And at the same time, we can (and should) combine writes from multiple keys into a single bucket.

eddyashton:
Resolves #3744.

This fixes a bug where a key which was rarely written to, and missing entirely from some range-based buckets, would be reported by the indexing strategy as a missing file. Instead we write empty buckets for empty ranges, and recognise never-written keys as well.

This new unit test hit a few other bugs in the indexing system, around edge cases/off-by-one errors in the handling of maximum requestable ranges. I believe they're all fixed, and the new test includes a random component to try and improve coverage of these cases.


src/indexing/strategies/seqnos_by_key_bucketed.cpp
```diff
@@ -124,6 +124,21 @@ namespace ccf::indexing::strategies
...
+      {
+        LOG_TRACE_FMT(
+          "Storing empty bucket for range [{}, {}) for key {:02x}",
+          begin.first,
+          begin.second,
+          fmt::join(k, ""));
+        store_to_disk(k, begin, {});
+        begin = get_range_for(begin.second);
+      }
+    }
```
eddyashton, line 140: This is the main fix. The way this strategy works is to divide the history of seqnos into equal-sized ranges, and create `(range, kv-key)` keyed _buckets_ of results, which have a predictable name so can be easily fetched. For instance, we might write index chunks like the following, if the bucket-size was 10 (where the right-hand side content is "every time this key was written to"):

```
([0, 10), "hello") => {4, 7, 9}
([10, 20), "hello") => {10, 13, 14, 17}
([10, 20), "saluton") => {10, 11, 15}
([40, 50), "saluton") => {42}
```

This function makes sure that, in the case above, where we didn't receive any writes to key `"saluton"` before seqno 10, or between seqnos 20 and 40, we still produce appropriate empty chunks when we're aware of them:
```
([0, 10), "saluton") => {}
([20, 30), "saluton") => {}
([30, 40), "saluton") => {}
```

jumaffre (approved):

PR #3750, opened 2022-04-07
Host ledger fixes and improvements

jumaffre:
This PR fixes a bug in the ledger code (host-side) and improves the management of undesirable files when starting from a snapshot. In particular:

- Bug fix: Committing past the `last_idx` has now no effect, whereas it would previously result in a segmentation fault if multiple ledger files were in flight.
- Improvement: we no longer delete uncommitted ledger files that are past the start-up snapshot `idx`, but instead mark them as `.ignored`. This is less aggressive and leaves the responsibility of deleting ledger files to the operator entirely. 

Note that this is the first bug found by using the generic service load infra while running our existing test suites (see #3707).


src/host/ledger.h
```diff
@@ -826,12 +849,15 @@ namespace asynchost
         {
           auto file_name = f.path().filename();
 
-          if (is_ledger_file_name_recovery(file_name))
+          if (
+            is_ledger_file_name_recovery(file_name) ||
+            is_ledger_file_name_ignored(file_name))
```
wintersteiger, line 854: Perhaps combine the two so that `is_ledger_file_name_ignored` also returns true when `is_ledger_file_name_recovery`?

wintersteiger (approved):

PR #3751, opened 2022-04-08
Describe text objects as `text/plain` in generated OpenAPI

eddyashton:
We previously marked all auto-generated OpenAPI request/response bodies as `application/json`, on the assumption they were using JSON serializers. This is obviously not true for plain-text responses, leading to inaccurate OpenAPI elements (and a follow-on warning during docs build).

We would ideally only insert the `application/json` content type if we knew it was using a JSON serializer, but that would need a type trait from the macro that doesn't currently exist, and it's not clear what we do in the fallback cases (we need a sensible default value, since the content type of each response is required).

### DIFF
--- a/doc/schemas/app_openapi.json
+++ b/doc/schemas/app_openapi.json
@@ -322,7 +322,7 @@
   "info": {
     "description": "This CCF sample app implements a simple logging application, securely recording messages at client-specified IDs. It demonstrates most of the features available to CCF apps.",
     "title": "CCF Sample Logging App",
-    "version": "1.9.1"
+    "version": "1.9.2"
   },
   "openapi": "3.0.0",
   "paths": {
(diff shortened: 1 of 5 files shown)


jumaffre (approved):

doc/schemas/app_openapi.json
```diff
@@ -990,7 +990,7 @@
         "responses": {
           "200": {
             "content": {
-              "application/json": {
+              "text/plain": {
```
wintersteiger, line 993: I tried this yesterday and Sphinx would not generate example responses anymore. If we want to have those, it may be worth another look to make sure it works.

wintersteiger (approved):

doc/schemas/app_openapi.json
eddyashton: I think that's fine. The auto generated example responses are nice-to-have, but they're not particularly helpful. At some point we'll add explicit examples.

Issue #3752, opened 2022-04-08
Failure in `recovery_test_suite`


eddyashton:
Debugging what caused this failure in the Instrumented job in the Daily Build:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=44172&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=ae32e15b-0191-5f39-7410-be7e80165871&l=17484

`recovery.test_recover_service` failed as the 13th step in a `recovery_test_suite` job.

Comparing against the log from a [good build](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=44113&view=logs&j=383d248c-4494-5797-d98f-6cef5140601e&t=ae32e15b-0191-5f39-7410-be7e80165871&l=17559). The difference is that in the failing run the first node instantly reported it was `PartOfNetwork`, then took a long time to open its frontend (and the second node took so long to open its frontend that the test timed out). In a passing test, we instead spend a while polling `/node/state` and seeing `"state": "ReadingPrivateLedger"`. When this eventually completes, the frontend opens quickly thereafter. This suggests a timing change caused by #3746.

jumaffre:
As discussed and investigated, the issue is that the new batching logic introduced by #3746 means that the enclave doesn't process any `tick`s from the host while the batch of ledger entries is being deserialised. When the entire batch has finally been deserialised, the enclave receives a new tick which may push it over the Raft election timeout and cause the node to become candidate. While we're expecting the election to eventually converge and a new primary node to be elected, this is largely untested and I believe it wouldn't work in some cases (e.g. if the private recovery completes while all the nodes are all candidates).

jumaffre:
@eddyashton I suggest we close this issue as this was an intermittent CI failure fixed by #3754 and the interesting follow-up work has been recorded in #3755.

PR #3756, opened 2022-04-08
Increase timeout for app to open after recovery

wintersteiger:
I've seen `recovery.test_recover_service` time out in `_wait_for_app_open` a few times locally but also in the CI (e.g. one of today's daily runs), usually when this test is execute multiple times (like in the test suite). I think the timeout needs to be increased because of the additional time taken for snapshot generation after recovery, which we added recently, but `args.ledger_recovery_timeout` may perhaps be too much.


eddyashton:
I think this is the same issue described in #3752 and fixed by #3754.

wintersteiger:
Yep, agreed, closing this for now!

Issue #3757, opened 2022-04-08
Enabling debug or trace with non-unsafe releases hangs


caleteeter:
**Describe the bug**
Inability to enable debug or trace logging for non-unsafe packages.

**To Reproduce**
Create a new app-dev prepped VM, start the sandbox (sandbox.sh) passing a package (e.g., $HOME/CCF/tests/js-authentication).  When starting with -l debug  or -l trace, the startup hangs and exception thrown.

![image](https://user-images.githubusercontent.com/8452367/162447919-5b05fa17-8097-4cce-9f9c-15acdb57b622.png)

**Expected behavior**
Expected to get enhanced logging.

**Environment information**
Used the latest 2.0.0-rc6 package (https://github.com/microsoft/CCF/releases/download/ccf-2.0.0-rc6/ccf_2.0.0_rc6_amd64.deb)

jumaffre:
@caleteeter Thanks for raising this bug. This is an issue that has recently surfaced within the team as well (https://github.com/microsoft/CCF/issues/3753). 

With release builds (i.e. non-unsafe), it is currently expected that the log level on the host cannot be set to a value lower than `info` (i.e. only `info`, `error` and `fail` are allowed). This is a limitation that we hope to address by the first non release-candidate `2.0.0` release. Note that you will be able to set a more verbose logging level on the host for unsafe builds, as you've probably already found out.

In the meantime, we've made the `sandbox.sh` more verbose and quicker to react to early failures such as this one (see https://github.com/microsoft/CCF/pull/3759). This will be available from `2.0.0-rc7`.

caleteeter:
@jumaffre thanks for confirming this, it would be good to have this listed in the main doc site, as its a bit confusing that I can set log levels that don't work for non-unsafe builds.

Additionally, when I run unsafe builds (with sandbox) using -v and -l debug, I see the messages (including debug) as the ccfhost starts but I don't see the log stream I expect when calling governance functions for instance.  I am debugging the JWT key refresh which should happen after some proposals, but not sure why I can't see the lower levels logs for this.  This should be logged as part of node state changes.

PR #3759, opened 2022-04-08
Quicker error reporting when using `sandbox.sh`

jumaffre:
After issues such as #3757, we now report an error early (5secs vs. 60secs previously) when a node does not start up successfully when using `sandbox.sh`. 

For example, setting the host log level to `trace` for a release build:
```
$ sandbox.sh -l trace --js-app-bundle ~/git/CCF/tests/js-authentication/ 
Setting up Python environment...
Python environment successfully setup
[15:59:30.004] Virtual mode enabled
[15:59:30.005] Starting 1 CCF node...
[15:59:35.126] Error! Some nodes ran into issues:
[15:59:35.126] - Node [0]:
terminating with uncaught exception of type std::logic_error: No enclave file found at ./libjs_generic.virtual.so

[15:59:35.126] Please raise a bug if the issue is unexpected: https://github.com/microsoft/CCF/issues/new?assignees=&labels=bug&template=bug_report.md&title=Unexpected%20error%20when%20running%20sandbox%20script
```


tests/infra/network.py
```diff
@@ -72,8 +72,10 @@ class StartupSnapshotIsOld(Exception):
     pass
 
 
-class NodeShutdownError(Exception):
-    pass
+class NetworkShutdownError(Exception):
+    def __init__(self, msg, errors=None):
+        self.errors = errors
+        super().__init__(msg)
```
achamayou, line 78: Seems odd to initialize the superclass last rather than first?

achamayou:
@jumaffre this is a good idea, but we could raise much quicker than the newly introduced timeout, as soon as the subprocess returns. Waiting on the PEM is reasonable if the process is still going, but not if it's already terminated.

tests/infra/remote.py
```diff
@@ -884,8 +889,8 @@ def suspend(self):
     def resume(self):
         self.remote.resume()
 
-    def get_startup_files(self, dst_path, timeout=FILE_TIMEOUT):
-        self.remote.get(self.pem, dst_path, timeout=timeout)
+    def get_startup_files(self, dst_path, timeout=FILE_TIMEOUT_S):
+        self.remote.get(self.pem, dst_path, timeout=0.1)
```
jumaffre, line 893:
```suggestion
        self.remote.get(self.pem, dst_path, timeout=REMOTE_STARTUP_TIMEOUT_S)
```

jumaffre:
> @jumaffre this is a good idea, but we could raise much quicker than the newly introduced timeout, as soon as the subprocess returns. Waiting on the PEM is reasonable if the process is still going, but not if it's already terminated.

I thought about this too and I don't think we want to introduce an additional stall when we start the node process to check for successful startup as this would introduce a significant delay to all node launches, even successful ones. Hence adding the check around an existing operation, with a much shorter timeout and retry logic so that it doesn't impact the run time when the node startups successfully, while reporting startup failures quickly.

achamayou:
@jumaffre I don't understand, I think we can check for the return code in the same loop that we check for the PEM? That shouldn't add any delay?

achamayou (approved):

Issue #3743, opened 2022-04-06
Node-endorsed certificates are never renewed


jumaffre:
The new `set_node_certificate_validity` proposal action renews the endorsed node certificate but does not renew the self-signed node certificate. This means that RPC interfaces that use the self-signed node certificate are never refreshed and clients will eventually get validation errors when the certificate expires. 

After discussion with @eddyashton, to fix this, we have multiple options:
1. Bundle the self-signed certificate renewal with the existing `set_node_certificate_validity` proposal action. 
2. Decouple the renewal of the two certificates, to cater for more flexible deployment scenarios and be more flexible if/when we introduce different types of endorsements.
3. Have the self-signed certificate automatically renew (using the host's time) since it is only meant to be used by operators.

Since this is to be fixed for 2.x, I'm leaning towards 1. but happy to discuss this further.

lynshi:
I think 2 is a bit more extensible. In fact, maybe `set_node_certificate_validity` should take an argument for the name of the RPC interface(s) for which to refresh the certificate.

As of now we understand that there should be one "primary" RPC interface and one for operators, but (correct me if I'm wrong) this isn't necessarily prescribed in the documentation and it's implied a network can have as many interfaces as desired. If this is the case, then the proposals should be as flexible as possible as well, otherwise why not default to only two interfaces (one node-endorsed, one service-endorsed) by design?

lynshi:
Relatedly, suppose someone defines 2 service-endorsed interfaces and 2 node-endorsed interfaces. Would the certificates presented by each type of interface be the same (e.g. would the 2 service-endorsed interfaces present the same certificate)?

jumaffre:
@lynshi Many thanks for your input on this. 

One thing to note is that the node certificate isn't only used for the TLS RPC interface. It's also recorded in the ledger (https://microsoft.github.io/CCF/main/audit/builtin_maps.html#nodes-endorsed-certificates) and the corresponding private key is used to sign the Merkle root and the signature is also recorded in the ledger (https://microsoft.github.io/CCF/main/audit/builtin_maps.html#_CPPv4N3ccf13NodeSignature3sigE). Consequently, the endorsed node certificate is also included in every transaction receipt (see https://microsoft.github.io/CCF/main/use_apps/verify_tx.html#write-receipts). 

[...]

PR #3767, opened 2022-04-13
Also renew self-signed node certificate

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/3743

We now also renew the self-signed node certificate when issuing the `set_node_certificate_validity` proposal. Note that the first approach described in #3743 isn't in fact implementable (see https://github.com/microsoft/CCF/issues/3743#issuecomment-1098194676) so this is implemented a much simpler way, although slightly less flexible: The self-signed certificate is automatically renewed from the (global) commit hook on the existing `ccf.nodes.endorsed_certificates` KV map, using the same validity period as the endorsed certificate. This in turn updates the certificates presented by [RPC interfaces whose authority is `Node`](https://microsoft.github.io/CCF/main/operations/configuration.html#endorsement). 

Also added a new `GET /node/self_signed_certificate` endpoint command to return the self-signed certificate of the target node so that operators can retrieve the self-signed node certificate after renewing it.

TODO:
- [x] Update docs


tests/governance.py
```diff
@@ -388,6 +393,10 @@ def test_each_node_cert_renewal(network, args):
     validity_period_allowed = args.maximum_node_certificate_validity_days - 1
     validity_period_forbidden = args.maximum_node_certificate_validity_days + 1
 
+    # Nodes should be started with at least 2 RPC interfaces to
+    # test endorsed and self-signed node certificate refresh
+    assert len(primary.host.rpc_interfaces) > 1
```
jumaffre, line 398:
```suggestion
```

tests/governance.py
```diff
@@ -388,6 +393,7 @@ def test_each_node_cert_renewal(network, args):
     validity_period_allowed = args.maximum_node_certificate_validity_days - 1
     validity_period_forbidden = args.maximum_node_certificate_validity_days + 1
 
+
```
jumaffre, line 396:
```suggestion
```

CHANGELOG.md
```diff
@@ -10,11 +10,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Fixed issue with incorrect node and service certificate validity period when starting node in non-GMT timezone (#3732).
+- Fixed issue with self-signed node certificates that are now renewed when the `set_node_certificate_validity` proposal is issued (#3767).
```
achamayou, line 13:
```suggestion
- Fixed issue with self-signed node certificates that are now renewed when the `set_node_certificate_validity` proposal is applied (#3767).
```

achamayou (approved):

PR #3768, opened 2022-04-14
Synchronise ledger chunks via entry header flags

wintersteiger:
Implements #3613

### DIFF
--- a/src/kv/committable_tx.h
+++ b/src/kv/committable_tx.h
@@ -427,13 +427,7 @@ namespace kv
       // This is a signature and, if the ledger chunking or snapshot flags are
       // enabled, we want the host to create a chunk when it sees this entry.
       // version_lock held by Store::commit
-      bool force_ledger_chunk =
-        pimpl->store->flag_enabled_unsafe(
-          AbstractStore::Flag::LEDGER_CHUNK_AT_NEXT_SIGNATURE) ||
-        pimpl->store->flag_enabled_unsafe(
-          AbstractStore::Flag::SNAPSHOT_AT_NEXT_SIGNATURE);
-
-      if (force_ledger_chunk)
+      if (pimpl->store->must_force_ledger_chunk_unsafe(version))
       {
         entry_flags |= EntryFlags::FORCE_LEDGER_CHUNK_AFTER;
         LOG_DEBUG_FMT(
--- a/src/kv/store.h
+++ b/src/kv/store.h
@@ -12,7 +12,6 @@
 #include "kv/untyped_map.h"
 #include "kv_serialiser.h"
 #include "kv_types.h"
-#include "service/tables/signatures.h"
 
 #define FMT_HEADER_ONLY
 #include <fmt/format.h>
@@ -51,9 +50,8 @@ namespace kv
     Version last_committable = 0;
     Version rollback_count = 0;
 
-    std::
-      unordered_map<Version, std::tuple<std::unique_ptr<PendingTx>, bool, bool>>
-        pending_txs;
+    std::unordered_map<Version, std::tuple<std::unique_ptr<PendingTx>, bool>>
+      pending_txs;
 
   public:
     void clear()
@@ -113,8 +111,7 @@ namespace kv
       Version v,
       Term term,
       const MapCollection& new_maps,
-      kv::ConsensusHookPtrs& hooks,
-      bool& force_ledger_chunk) override
+      kv::ConsensusHookPtrs& hooks) override
     {
       auto c = apply_changes(
         changes,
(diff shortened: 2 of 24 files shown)


src/node/node_state.h
```diff
@@ -900,7 +900,7 @@ namespace ccf
             // Inform snapshotter of all signature entries so that this node can
             // continue generating snapshots at the correct interval once the
             // recovery is complete
-            snapshotter->record_committable(last_recovered_idx);
+            snapshotter->record_committable_unsafe(last_recovered_idx);
             snapshotter->commit(last_recovered_idx, false);
```
jumaffre, line 904: I believe this path was to maintain the ledger file size invariant during the recovery, which I think can now go?

tests/lts_compatibility.py
```diff
@@ -501,10 +503,25 @@ def run_ledger_compatibility_since_first(args, local_branch, use_snapshot):
                         version,
                     )
 
+                # We accept ledger chunk file differences during upgrades
+                # from 2.x-rc6 to 2.x-rc7 and from 1.x to 2.x-rc7
```
jumaffre, line 507: I think we should explain why too, and perhaps link to #3613 so that we can clean this up once 2.0 is released.

src/node/snapshotter.h
```diff
@@ -262,7 +262,7 @@ namespace ccf
       next_snapshot_indices.push_back({last_snapshot_idx, false, true});
     }
 
-    bool record_committable(consensus::Index idx) override
+    bool record_committable_unsafe(consensus::Index idx) override
```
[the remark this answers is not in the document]
wintersteiger: Good question! I think I did that because it accesses the store in an unsafe manner, but the store is now also the only client, so I'll undo that.

src/node/node_state.h
wintersteiger: Yep, I think so too. Just waiting for the CI to confirm that this doesn't break anything.

tests/lts_compatibility.py
```diff
@@ -501,10 +503,28 @@ def run_ledger_compatibility_since_first(args, local_branch, use_snapshot):
                         version,
                     )
 
+                # We accept ledger chunk file differences during upgrades
+                # from 2.x-rc6 to 2.x-rc7 and from 1.x to 2.x-rc7. This
+                # is necessary because we broke the ledger compatibility between
+                # those versions (see https://github.com/microsoft/ccf/issues/3613).
```
jumaffre, line 509: I think this is too strong of a statement. The ledger compatibility isn't broken between these versions, it's only that the ledger files may not be chunked at the same interval between these nodes since 2.x nodes require a ledger entry flag to create a new chunk, and this flag isn't present on 1.x ledgers.

src/kv/committable_tx.h
```diff
@@ -427,13 +427,7 @@ namespace kv
...
       // enabled, we want the host to create a chunk when it sees this entry.
       // version_lock held by Store::commit
-      bool force_ledger_chunk =
-        pimpl->store->flag_enabled_unsafe(
-          AbstractStore::Flag::LEDGER_CHUNK_AT_NEXT_SIGNATURE) ||
-        pimpl->store->flag_enabled_unsafe(
-          AbstractStore::Flag::SNAPSHOT_AT_NEXT_SIGNATURE);
-
-      if (force_ledger_chunk)
+      if (pimpl->store->must_force_ledger_chunk_unsafe(version))
```
jumaffre, line 430: I think this deserves a unit test in `kv_test.cpp`.

src/kv/store.h
```diff
@@ -1058,6 +1042,26 @@ namespace kv
...
     }
 
+    bool must_force_ledger_chunk(Version version) override
+    {
+      std::lock_guard<std::mutex> vguard(version_lock);
+      return must_force_ledger_chunk_unsafe(version);
+    }
+
+    bool must_force_ledger_chunk_unsafe(Version version) override
+    {
```
jumaffre, line 1052: I think this needs a comment to specify that `version` has to be of a committable entry (i.e. signature).

wintersteiger: Yes, I agree. Is there a way to automatically check that?

tests/lts_compatibility.py
wintersteiger: Oh yes that's what I meant to say, rephrasing.

src/kv/committable_tx.h
wintersteiger: Done!

src/kv/store.h
wintersteiger: Comment added, but I think it would be nice if we could check that automatically. In theory we could save the txid of the reserved TX when we created it, but it feels like that might trigger a bunch of other work to make sure there aren't multiple of them in flight, etc.

jumaffre: Alternatively, the `force_ledger_chunk` flag could be passed in the constructor of the `ReservedTx` when we call `store->create_reserved_tx(tx_id)` since we could ask the snapshotter early whether a the transaction will require the creation of a new ledger chunk.

wintersteiger: Hmm. Could that potentially create problems when the transaction doesn't get committed?

wintersteiger:
azp /run

jumaffre (approved):

PR #3769, opened 2022-04-14
Keep track of time of backup latest ACK on primary

### LINKED ISSUE #3688: Report time of last backup ACK in `GET /node/consensus` endpoint
When monitoring the state of a service (see #3676), it is possible that the primary node is able to send heartbeats to backup nodes but the backup nodes cannot send an ack back to the primary (in other words, one way partition between the primary and a majority of backups). It would then seem that the service is healthy, but in reality the service wouldn't be able to commit new transactions. 

To help operators detect such failure case, we should add some details on the last time an ACK was received from each backup.

Note that this is effectively the first step towards the CheckQuorum Raft extension described in #2577.

jumaffre:
Resolves #3688 

The primary node keeps track of the time since last ACK from each backup and report it in the `GET /node/consensus` endpoint. This is useful to detect one-way partitions between the primary and backups, and a useful building block towards `CheckQuorum`.

This is reflected in the `health_watcher.py` sample script and tested by the existing `partitions_test.py` end-to-end test.

TODO:
- [x] Changelog
- [x] Implement in `health_watcher.py` sample
- [x] Add tests in `partitions_test.py`


src/consensus/aft/raft.h
```diff
@@ -44,6 +44,9 @@ namespace aft
       // the highest matching index with the node that was confirmed
```
jumaffre, line 44: The functional changes are in this file. The rest of the changes are only for testing purposes or for updating the API.

tests/election.py
```diff
@@ -35,6 +35,19 @@ def test_kill_primary_no_reqs(network, args):
             r = c.get("/node/network")
             c.wait_for_commit(r)
 
+            # Also verify that reported last ack time are as expected
```
jumaffre, line 38: When a new primary node has been elected, we check that the backup nodes report `0` as the time of latest backup ACK (backups don't exchange append entries between each other!) and the primary reports sensible ACK times for each remaining backups.

tests/infra/health_watcher.py
```diff
@@ -83,6 +85,19 @@ def get_network_health(network, get_primary_fn, client_node_timeout_s=3, verbose
         # about to lose its primaryship
         return HealthState.election
 
+    # The current primary is one-way partitioned from f backups, i.e. cannot
+    # receive acks from a majority of backups
+    # Note: This can be removed once the CheckQuorum extension is implemented
+    # as the partitioned primary node will automatically step down.
+    delayed_acks = [
```
jumaffre, line 92: From a single operator's point-of-view, a network needs to be recovered if the primary doesn't receive any ACKs from a majority of backups for a long time, which would be the sign of one-way partitions between the primary and backups.

wintersteiger:
azp /run

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Added
 
 - Added new `GET /node/self_signed_certificate` endpoint to retrieve the self-signed certificate of the target node (#3767).
+- Primary node now also reports time at which the ack from each backup node was last received (`GET /node/consensus` endpoint). This can be used by operators to detect one-way partitions between the primary and backup nodes (#3769).
```
achamayou, line 22: Needs moving out to a new block

achamayou (approved):

CHANGELOG.md
jumaffre: Good spot!

raghavthind2005 (approved):

PR #3776, opened 2022-04-20
Better LTS compatibility checks when committing to `main`

jumaffre:
This is expected to fail until https://github.com/microsoft/CCF/pull/3775 is merged.

When committing to `main`, the `lts_compatibility` test only checked for live compatibility with the latest release candidate (i.e. `ccf-2.0.0-rc6`), but not with the latest previous final release (i.e. `ccf-1.0.19`). The compatibility with `ccf-1.0.19` was only checked when the next release tag (i.e. `ccf-2.0.0-rc7`) was cut:

```bash
# when committing to main, lts_compatibility test outputs:
{
  "version": "ccf-2.0.0-rc6-27-gea38c18f",
  "live compatibility": {
    "with previous LTS": "ccf-2.0.0-rc6",
    "with same LTS": null
  }
}
```

This now checks for compatibility with both `ccf-2.0.0-rc6` and `ccf-1.0.19`, helping us catch compatibility issues early:

```bash
# when committing to main, lts_compatibility test now outputs:
 {
  "version": null,
  "live compatibility": {
    "with previous LTS": "ccf-1.0.19",
    "with same LTS": "ccf-2.0.0-rc6"
  }
}
```

PR #3777, opened 2022-04-21
Reduce number of verified receipts in LTS compatibility test

jumaffre:
#3776 seems to have taken the `lts_compatibility` over its allowed timeout so this PR reduces the number of receipts we sample in this test.


wintersteiger (approved):

raghavthind2005 (approved):

eddyashton (approved):

jumaffre:
Also bumped the timeout value for the instrumented build as we were still dangerously close to the previous 10-minute limit.

PR #3780, opened 2022-04-21
Reset `last_committable` on rollback

jumaffre:
~We did not keep track of the last committable/signature seqno (i.e. `last_committable` in `Store`) on backup nodes. This had the following consequences:~
~- A freshly-elected primary node would often create a new signature as soon as it was elected.~
~- For long ledgers, the `Recover` primary node would trigger a signature **inline** (i.e. triggered by the transaction interval rather than time) with the first recovery transaction which could cause deadlocks if hooks set the `NodeState` lock.~

The above will be addressed by another PR that makes signature generation asynchronous when triggered by the transaction count since the last signature.

We now reset `last_committable` on rollback so that a signature is always triggered when a node becomes primary.

Also renamed `commit_gap()` to `committable_gap()`.


src/kv/store.h
```diff
@@ -131,6 +131,11 @@ namespace kv
         version = v;
         last_replicated = version;
         term_of_last_version = term;
+
+        if (changes.find(ccf::Tables::SIGNATURES) != changes.end())
```
eddyashton, line 135: I don't love this separate evaluation of "is this globally committable". In particular it means that in unit tests, we will end up with a different sequence of `last_committable` values on backups than on primaries. I'm not sure what a better solution would be.

jumaffre: Yes, I agree that this isn't ideal. I feel like that there's no quick win here and that this need to be part of a bigger refactoring beyond the scope of this PR?

achamayou (approved):

PR #3782, opened 2022-04-22
Make tx interval signature generation asynchronous

jumaffre:
Having signatures created synchronously from the frontend execution may cause dead locks in some cases where the triggered hooks also synchronously triggered may acquire locks that were already taken by the frontend caller in the first place. 

Specially, this would arise when recovering a ledger longer than the [signature tx count](https://microsoft.github.io/CCF/main/operations/configuration.html#ledger-signatures): the very first post-public recovery transaction would trigger a new signature that would cause a synchronous (global) commit hook that would try to acquire the already-acquired `ccf:NodeState` lock.

### DIFF
--- a/src/node/history.h
+++ b/src/node/history.h
@@ -720,6 +720,8 @@ namespace ccf
 
     std::mutex signature_lock;
 
+    bool forced_signature = false;
+
     void try_emit_signature() override
     {
       std::unique_lock<std::mutex> mguard(signature_lock, std::defer_lock);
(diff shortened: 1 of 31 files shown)


src/node/history.h
```diff
@@ -731,7 +731,20 @@ namespace ccf
       if (store.commit_gap() >= sig_tx_interval)
       {
         mguard.unlock();
-        emit_signature();
+
+        // Note: emit signature asynchronously to avoid possible deadlocks in
+        // case the signature synchronously triggers commit hooks that acquire
+        // already-taken locks (typically with the /node/create transaction
+        // triggered from the node state).
+        auto emit_sig_msg = std::make_unique<threading::Tmsg<EmitSigMsg>>(
```
jumaffre, line 739: Because of the current fairness strategy between incoming ring buffer messages and thread messages, this currently means that in the very worst case, a signature may be scheduled 256 `seqno`s after the actual `tx_count` signature interval.

https://github.com/microsoft/CCF/blob/84461e142c56a6d969faefaec15920dc95867036/src/enclave/enclave.h#L414

There's a way to schedule the signature at its exact interval by modifying the fairness logic. I'll push the change in a separate commit (but same PR) so we can have a look at it separately.

jumaffre: See https://github.com/microsoft/CCF/pull/3782/commits/54cfc533ec14fb31810c1d0d40795401cf646612

jumaffre:
This change was deemed to complex and risky to implement just before release 2.0.0. Instead, the issue around synchronous signature generation causing a deadlock was addressed by #3780 and #3791.

PR #3783, opened 2022-04-22
Always emit custom claims

### LINKED ISSUE #3606: Custom claims to be produced for all transactions
This is to avoid leaf nodes being shorter than intermediary nodes in the tree going forward. Must happen by 2.0 final.

This is to avoid attacks such as the one described in https://bitslog.com/2018/06/09/leaf-node-weakness-in-bitcoin-merkle-tree-design/, where a leaf node could be re-interpreted as an intermediary node in a taller tree.

achamayou:
Fix #3606.

With this change, new transactions always contain a claims digest, regardless of whether the application logic specified one. Transactions that do not contain one are _still_ accepted, for compatibility with older ledgers.


doc/use_apps/verify_tx.rst
```diff
@@ -58,39 +58,28 @@ This means that the request may return ``202 Accepted`` at first, with a suggest
...
+    'node_id': 'c0718b67b986748d074a293201fa2f9bc3a29e07d204ccc24167afdc59bdf0c3',
+    'proof': [{'left': '26e7edf1065e23a387b8d6fd3743b4885aae1919c4c347d73b792e3e77b57695'},
+            {'left': '76c79df9146479c0aebdd51f228e1d304d60f382f61571a729f2fba7d93ce2a7'},
+            {'left': '6b54d109e3f474f9f546d5d462de40440ca335b133d3899c228532fcc535e732'},
+            {'left': 'bf3e529174209df204342224fc39ffee68189c30db03e53ba1d4baa053e02471'},
+            {'left': '740dcdb82afe5f2b311056c96bd6afcc1b571e1cdb8e918bfe24eda48213eb1a'},
+            {'left': '05efd90dfd3bf86a3efa185c6a869342cb0d9af4dd92256d37dc826b86e99684'}],
+    'signature': 'MGYCMQCI1yY5xhGNFXyKjO6mFhxZCjAsuTsquO1agIzKkxh/Uf9Em+P7AwnxuQ2uC7WU++QCMQCSzpE9k8xkSku5BlkUelDPBHxLuokLIpKAyn73HXtXiz+zXOVWIZkYftUEoQ31pG0='}
 
 `cert` contains the certificate of the signing node, endorsed by the service identity. `node_id` is the node's ID inside CCF, a digest of its public key.
```
jumaffre, line 85: I can't comment on the specific line, but on `109`, I believe "when applicable" can now be removed?

jumaffre: Also, on `113`, the link to `check_endorsements` is broken and should be `:py:func:ccf.receipt.check_endorsements` instead.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - When using the `sandbox.sh` script, always wait for `/app` frontend to be open on all nodes before marking the service as open (#3779).
+- Every leaf in the Merkle Tree now includes a claims digest (#3606).
```
jumaffre, line 13: Should this be under `### Changed`? I wonder if this changelog entry should be phrased in terms of user-visible change here, i.e. receipts?

jumaffre (approved):

jumaffre:
Could we run the daily pipeline on this change too so that it tests for compatibility with previous version during a recovery?

eddyashton (approved):
LGTM!

doc/use_apps/verify_tx.rst
achamayou: > I can't comment on the specific line, but on 109, I believe "when applicable" can now be removed?

I nearly did this, but did not because it won't for transactions executed before this change that did not have user-set claims.

CHANGELOG.md
achamayou: I was thinking of it in terms of fixing the security issue, rather than the user-visible change, but you're right we can list both.

PR #3784, opened 2022-04-22
Restore ability to launch infra nodes with debugger attached

eddyashton:
The recent changes to add early-crash detection broke the `-d` argument to sandbox/start_network/etc. Specifically, `check_done` attempts to access `self.proc`, which only exists in the `self.remote.start()` path, but not if the node was started manually with a debugger.

I considered moving the `Detect whether node started up successfully` block, so it wouldn't run at all on the debugger-attached nodes. I think this approach, even though it's an ugly change, gives nice behaviour - if debug-attached nodes crash on start-up, the infra recognises this quickly from missing startup files, rather than ignoring it entirely.

### DIFF
--- a/tests/infra/remote.py
+++ b/tests/infra/remote.py
@@ -543,7 +543,7 @@ class LocalRemote(CmdMixin):
         return f"cd {self.root} && {DBG} --args {cmd}"
 
     def check_done(self):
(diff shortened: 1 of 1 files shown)


jumaffre (approved):
Thanks for fixing this!

PR #3787, opened 2022-04-25
UDP enclave echo server

rengolin:
This is an end-to-end UDP echo server, which receives UDP packets from
multiple clients at the same time, passes them to the enclave with a
message type that is handled by a QUICEchoEndpoint (which just echoes
back the messages to the client) and send the message back through the
ring buffer to the host, which then send to the correct client.

The UDP packets, alongside their socket addresses, are passed to the
enclave as quic_inbound rung buffer messages and received back as
quic_outbound messages. For now, those are encoded as {long,long} to
avoid size issues in the serialization, but eventually, they'll be
broken down to their components {short,char[14]}.

UDP interfaces are created like their TCP counterparts, via a new
field in RPCInterface called 'protocol', with default to "tcp". If
the field is set to "udp", the host will listen on the UDP version
of RPCConnection and use the 'quic_*' messages, which connects to the
new QUICEchoEndpoint.

Also:
 * Renaming "service" to "port"
 * Refactoring TCPBehaviour into SocketBehaviour<ConnType>
 * Factor next_id out to share amongst all RPC connections
 * Use local ConnID in rpc_connections, same as tcp/udp
 * Factor some common logic into socket.h
 * Factor pending writes/reads into ds/pending_io.h

Issues:
 * There is an implicit 'quic_start' message but not a 'quic_close' one,
   which can be a problem for many connections, over time. It's not
   clear to me yet that this will create message mismatches, but we have
   to make sure it doesn't on the next step.
 * The current RPCConnections is parametrized by TCPImpl/UDPImpl, which
   is handy for now but creates the implicit structural dependency
   between the two parameters. We need a common interface to derive
   from, so that the two structures can actually be checked at compile
   time.
 * To implement the step above, we need to fix 'proxy_ptr' and
   'close_ptr' to allow pointer checks (virtual inheritance). I tried
   implementing that first but it creates ripples throughout the code.
   We may never fix this, but then things like the auto-increment will
   have to continue being external, amongst other things.

Next Steps:

 1. To implement all services through unencrypted UDP, just like TCP, so
    we can test the longecity and stability of the UDP channel. We can
    ignore all TLS errors, and just make sure requests and responses are
    passed to the right services and back to the right clients. This
    will also need a UDP client, so that nodes can RPC each other in
    UDP.

 2. To implement QUICEndpoint using ngtcp2 and HTTP3Endpoint using nghttp3
    and test using proper TLS certificates and make sure the services are
    still available in the same way.


doc/host_config_schema/cchost_config.json
```diff
@@ -49,6 +49,11 @@
                 "default": "Value of 'bind_address'",
                 "description": "The published RPC address advertised to clients"
               },
+              "protocol": {
+                "type": "string",
+                "default": "(empty)",
+                "description": "The RPC protocol used (udp | tcp)"
+              },
```
jumaffre, line 56: I think this should be an enum (see for example `"endorsement"` on line `62`) that defaults to TCP.

jumaffre: I also wonder whether the protocol field should be more opinionated with regards to the application layer protocol too, i.e. `http/1`, `http/2` or `http/3`?

src/enclave/rpc_sessions.h
```diff
@@ -420,14 +442,40 @@ namespace ccf
...
           auto [id] = ringbuffer::read_message<tls::tls_close>(data, size);
           remove_session(id);
         });
+
+      DISPATCHER_SET_MESSAGE_HANDLER(
+        disp, quic::quic_start, [this](const uint8_t* data, size_t size) {
+          auto [new_id, listen_interface_name] =
+            ringbuffer::read_message<quic::quic_start>(data, size);
+          accept(new_id, listen_interface_name, true);
+          // UDP sessions are never removed because there is no connection
```
jumaffre, line 459: Does this imply infinite memory growth inside the enclave then?

tests/e2e_logging.py
```diff
@@ -1467,6 +1467,54 @@ def test_rekey(network, args):
...
+def run_udp_tests(args):
+    # Register secondary interface as an UDP socket on first node
+    primary = args.nodes[0].rpc_interfaces["primary_rpc_interface"]
+    udp_interface = infra.interfaces.make_secondary_interface()
+    for interface in udp_interface.values():
+        interface.host = primary.host
+        interface.port = primary.port
+        interface.public_host = primary.public_host
+        interface.public_port = primary.public_port
+        interface.transport = "udp"
```
jumaffre, line 1501: Is there a strong reason to have the UDP interface listen on the same host:port as the main TCP interface? If not, I think it'd be simpler to do:
```python
primary.rpc_interfaces.update(infra.interfaces.make_secondary_interface(protocol="udp"))
```

doc/host_config_schema/cchost_config.json
rengolin: Right now, this is at best experimental. It's not clear to me what's the best granularity (http, quic, udp, tcp) or if we want a parseable descriptor (like the "local,tcp" hack in the string representation). I personally prefer to leave as a string until we're sure what we want and how we want.

I'm also not sure about being too opinionated. HTTP 1 through 2 can really only work with TCP (+TLS) and HTTP 3 can only work with QUIC and QUIC can only work with UDP (no TLS). And for all purposes, HTTP 2 is an extension of 1.1 which is an extension of 1, so we end up with only two choices:
 * HTTP {1, 1.1, 2} + TCP + TLS or
 * HTTP 3 + QUIC

Whether we call it QUIC or UDP or HTTP 3 doesn't make much of a difference, I think.

src/enclave/rpc_sessions.h
rengolin: Yes, because `remove_session` isn't called. This is the weakest point of the current experiment, and one of the things I want to discuss in a meeting.

One way to get around this is to call accept only once (ex. with a static flag that, once set, ignores new connections). This is a hack, but works quickly and easily. This is not a long term solution, but works because we only have an echo server for now.

Because UDP has no connection, we can continue that pattern (with a better implementation) for the foreseeable future. UDP packets don't start/stop connections, they always have src/dst addresses and the connection is always open, so a single HTTP 3 Endpoint is what we want, really.

Another way is to detect QUIC "streams". Those are handled by the enclave side and know when a stream has finished its work, but it has to tell the host to remove the connection. So, calling `accept` would be a reaction to a host event (new address not in the open list), while calling `remove_session` would be a reaction to an enclave event (QUIC telling us the stream is closing).

Because these don't come from the same side, we'll need some plumbing. Furthermore, there are many ways in which a QUIC stream may die, including not sending another packet. If it turns out is common for clients to not bother ACKing packets, servers need to be smarter to close the streams, and that may require even more plumbing.

So, the original design is to have a single channel between host(udp) and enclave(quic), and let the QUIC layer handle sessions, streams and memory lifetime.

tests/e2e_logging.py
rengolin: There was a reason, but not anymore. These ports are all `0` anyway, so they're different ones at runtime.

I did the `make_secondary_interface` change you proposed earlier, but I realised I still wanted to control other things (like host), but with the new registration process, even that's not needed any more.

I'll try again with the simpler call.

doc/host_config_schema/cchost_config.json
jumaffre: Well summarised. We may also want to distinguish http from httpS (in case we "add" support for this in the future?). I think there's an argument for only allowing HTTP/2 too (e.g. when apps are gRPC only and we don't want to support the upgrade from 1.1 to 2).

rengolin: I thought we only wanted to support TLS protected streams, be it HTTPS or QUIC (ie, no HTTP).

About gRPC, seems it makes use of core ideas in HTTP/2 that aren't in HTTP/1.1, so we'll have to differentiate that, too. In a way, this maps nicely to `{ http1, http2, http3 }` better than `{ tcp, udp }`.

But for now, since even calling it QUIC is a misnomer, we can leave it as UDP because that's all it is. We can easily change the naming later or create an enum or an integer `{ 1, 2, 3 }`, once we have an HTTP/3 endpoint.

src/host/main.cpp
```diff
@@ -294,15 +298,44 @@ int main(int argc, char** argv)
...
     rpc.register_message_handlers(bp.get_dispatcher());
 
+    // This is a temporary solution to keep UDP RPC handlers in the same
+    // way as the TCP ones without having to parametrize per connection,
+    // which is not yet possible, due to UDP and TCP not being derived
+    // from the same abstract class.
+    asynchost::RPCConnections<asynchost::UDP> rpc_udp(
+      writer_factory, idGen, config.client_connection_timeout);
+    rpc_udp.register_quic_message_handlers(bp.get_dispatcher());
+
```
eddyashton, line 314: Why do these share an `idGen`, rather than having their own? As long as we only ever use the ID within protocol-specific messages (ie, either `tls::` or `quic::`), it's fine if ID 5 is reused.

src/host/rpc_connections.h
```diff
@@ -46,50 +138,89 @@ namespace asynchost
...
       {}
 
-      void on_listening(
-        const std::string& host, const std::string& service) override
+      void on_accept(ConnType& peer) override
       {
-        LOG_INFO_FMT("Listening for RPCs on {}:{}", host, service);
-      }
+        // UDP connections don't register peers
+        if (getConnTypeID<ConnType>() == CONN_UDP)
```
eddyashton, line 171: If you make `getConnTypeID` constexpr, this can be `if constexpr (...)`, and perhaps even `if constexpr (std::is_same_v<ConnType, asynchost::TCP>())` (if the `TCP` and `UDP` types are safely visible here, and the enum's not used for anything other than this type of switch).

src/enclave/rpc_sessions.h
eddyashton: I wonder if it's worth creating "sessions" for UDP at all at this point? We don't want the same session limit behaviour, especially if we're not cleaning up (IIUC, we're creating a _new_ session for every incoming packet, so the session counts are really packet counts?).

In that model, we'd remove `quic_start` entirely, and in `quic_inbound` we'd dispatch to a single QuicHandler/QuicEndpoint instance. Internally, once that's done some QUIC parsing, it might be able to tell us about meaningful sessions/streams/connections that we want to count, or limit. But I don't think it's helpful to pair those to TLS connections before we know that?

(Note: I think it's fine to leave this as-is for this PR, and explore this separately in future)

tests/e2e_logging.py
```diff
@@ -1467,6 +1467,47 @@ def test_rekey(network, args):
...
+
+    text = recv[0]
+    LOG.info(f"Testing UDP echo server received {text}")
+    assert text == test_string
+
+
+def run_udp_tests(args):
+    # Register secondary interface as an UDP socket on first node
+    udp_interface = infra.interfaces.make_secondary_interface("udp")
+    args.nodes[0].rpc_interfaces.update(udp_interface)
```
eddyashton, line 1495: We should update this on all nodes, not just node 0 (assumption that `args.nodes[0]` constructs `primary` found later isn't safe).

tests/infra/interfaces.py
```diff
@@ -91,10 +99,10 @@ def from_json(json):
         return interface
 
 
-def make_secondary_interface():
+def make_secondary_interface(transport="tcp"):
```
rengolin, line 102: Make name a parameter and use instead of SECONDARY_RPC_INTERFACE

src/enclave/rpc_sessions.h
rengolin: At the very least, make sure we only create the endpoint once

tests/e2e_logging.py
rengolin: Right, this is left-over from a time when I was creating the nodes by hand. Now I should just use `get_primary` like the other tests (as I did initially).

eddyashton: You can't use `get_primary` here - the network doesn't exist yet (just the `args` that specify how it should be created), and the primary may change during execution. Instead we want to make sure every node gets a UDP interface, by updating the spec for every node:

```suggestion
    for node in args.nodes:
        node.rpc_interfaces.update(udp_interface)
```

src/enclave/rpc_sessions.h
rengolin: Couple of points here:
* We need to create one UDP session per listening interface. If there are more than one UDP interface listening (different ports, IPs, etc), we need one `ConnID` for each and the ring-buffer needs to differentiate between them.
* The whole communication between host and enclave is based on `ConnID` and interface names and refactoring that now would be a prohibitive cost for the time being.
* In the end, the memory doesn't grow, because the `start` behaviour is only called once per UDP server listen, not once per client connection, so the right number of sessions/endpoints are created.

src/host/main.cpp
rengolin: I've factored that out after getting errors in the log saying "A connection with this ID already exists". I didn't did too deeply, though.

It may be conceptually fine for different protocols to have their own ID space, but the sessions' IDs are bundled somewhere in the middle, and I'm guessing on the enclave side that doesn't differentiate the protocol in `accept(id, ...)`.

src/host/rpc_connections.h
rengolin: Done.

tests/e2e_logging.py
rengolin: Done

tests/infra/interfaces.py
rengolin: Done

doc/host_config_schema/cchost_config.json
```diff
@@ -51,6 +51,11 @@
                 "default": "Value of 'bind_address'",
                 "description": "The published RPC address advertised to clients"
               },
+              "protocol": {
+                "type": "string",
+                "default": "(empty)",
```
jumaffre, line 56:
```suggestion
                "default": "tcp",
```

Issue #3788, opened 2022-04-25
Deadlock when committing start of recovery transaction


jumaffre:
The recent addition of the following line in the (global) commit hook of the node endorsed certificate table causes a dead-lock on the `Recovery` node when recovering a large ledger:
https://github.com/microsoft/CCF/blob/556214627d7ad025df16afac8bd4ffe44ea4c632/src/node/node_state.h#L2089

This is because the commit hook is executed synchronously with the execution of the start of recovery request/transaction. The hook acquires the lock on the node state but this one is already taken. Note that this only happens with ledger longer than the [ledger_signatures.tx_count](https://microsoft.github.io/CCF/main/operations/configuration.html#ledger-signatures) configuration entry.

The solution is to make the execution of this request asynchronous , in line with existing node-issued requests (e.g. retired nodes cleanup, ORC, etc.).

PR #3791, opened 2022-04-26
Asynchronous execution of start of recovery transaction 

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/3788

The start of recovery transaction, automatically emitted by the `Recover` node once the public ledger is recovered used to be synchronous, which may trigger the generation of a signature transaction via the `tx_count` logic, and result in a deadlock if a global commit hook acquired the node's lock.

This is now fixed by scheduling the execution of this transaction asynchronously, as well as only starting the timer to regularly generate signatures once this transaction is successfully committed.


src/node/node_state.h
```diff
@@ -984,6 +986,13 @@ namespace ccf
       verify_snapshot_end_unsafe();
     }
 
+    void advance_part_of_public_network()
+    {
+      std::lock_guard<std::mutex> guard(lock);
+      history->start_signature_emit_timer();
```
achamayou, line 992: Missing an expect (ReadingPublicLedger, I'm guessing, but please check)

eddyashton: Added this, and local runs passed. Will let the CI stress the full test suite before merging.

eddyashton (approved):

PR #3804, opened 2022-05-03
Fix LTS ledger compatibility chunks

jumaffre:
Since releasing [2.0.0-rc8](https://github.com/microsoft/CCF/releases/tag/ccf-2.0.0-rc8), the `lts_compatibility` test is failing because of the ledger chunks inconsistency introduced in https://github.com/microsoft/CCF/pull/3768 when upgrading from 1.x to 2.x-rc8+. I've simplified the logic so that we accept any ledger incompatibility when upgrading from 1.x to 2.x-rc8+.

### DIFF
--- a/tests/lts_compatibility.py
+++ b/tests/lts_compatibility.py
@@ -425,8 +425,6 @@ def run_ledger_compatibility_since_first(args, local_branch, use_snapshot):
     )
     with jwt_issuer.start_openid_server():
         txs = app.LoggingTxs(jwt_issuer=jwt_issuer)
-        previous_major_version = None
-        previous_version_past_2_rc7 = False
         for idx, (_, lts_release) in enumerate(lts_releases.items()):
             if lts_release:
                 version, install_path = repo.install_release(lts_release)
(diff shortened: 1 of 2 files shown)


tests/lts_compatibility.py
```diff
@@ -504,20 +503,18 @@ def run_ledger_compatibility_since_first(args, local_branch, use_snapshot):
...
+                )
                 is_ledger_chunk_breaking = (
-                    not previous_major_version
-                    or previous_major_version == 1
-                    or not previous_version_past_2_rc7
-                ) and current_version_past_2_rc7
-                previous_major_version = primary.major_version
-                previous_version_past_2_rc7 = current_version_past_2_rc7
+                    has_pre_2_rc7_ledger and current_version_past_2_rc7
+                )
```
wintersteiger, line 517: There might be a problem when the version is equal to RC7, but I think that won't be triggered by any of the CI runs (and we don't care once 2.0 is released).

wintersteiger (approved):

achamayou (approved):

PR #3805, opened 2022-05-03
Pin `httpx` to 0.22

jumaffre:
This currently causes the following security alert: https://github.com/microsoft/CCF/security/dependabot/3

image

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -13,6 +13,6 @@ docker
 better_exceptions
 pyasn1
 Jinja2
-httpx == 0.21.*
+httpx
 locust
 pyOpenSSL


tests/requirements.txt
```diff
@@ -13,6 +13,6 @@ docker
 better_exceptions
 pyasn1
 Jinja2
-httpx == 0.21.*
+httpx
```
achamayou, line 16: I am reluctant to let go of this, because httpx does break between minor releases (which is their prerogative under semver). I would rather we pinned to 0.22.*

achamayou, line 16:
```suggestion
httpx == 0.22.*
```

Issue #3808, opened 2022-05-04
Reconfiguration test break


achamayou:
CI: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=44959&view=logs&j=e982d093-d12a-5a2d-1a1d-0c6d38b8823f&t=8be5454d-0d8f-51e4-7372-58df022e9ebc

```
62: 2022-05-04 03:41:35.688 | ERROR    | {1tx_reconfig} infra.runner:log_exception:211 - Failure in 1tx_reconfig: ValueError('Response seqno and view should not be None')
62: Traceback (most recent call last):
62: 
62:   File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
62:     self.run()
62:     └ <Thread(1tx_reconfig, started 140040401565440)>
62:   File "/usr/lib/python3.8/threading.py", line 870, in run
62:     self._target(*self._args, **self._kwargs)
62:     │             │             └ <Thread(1tx_reconfig, started 140040401565440)>
62:     │             └ <Thread(1tx_reconfig, started 140040401565440)>
62:     └ <Thread(1tx_reconfig, started 140040401565440)>
62:   File "/__w/1/s/tests/reconfiguration.py", line 571, in run
62:     test_retire_primary(network, args)
62:     │                   │        └ Namespace(binary_dir='/__w/1/s/build', ccf_version='ccf-2.0.0-rc8-1-ge3f677a2', common_read_only_ledger_dir=None, config_file=No...
62:     │                   └ <infra.network.Network object at 0x7f5da9936820>
62:     └ <function test_retire_primary at 0x7f5da9bfb160>
62:   File "/__w/1/s/tests/suite/test_requirements.py", line 20, in wrapper
62:     return func(*args, **kwargs)
62:            │     │       └ {}
62:            │     └ (<infra.network.Network object at 0x7f5da9936820>, Namespace(binary_dir='/__w/1/s/build', ccf_version='ccf-2.0.0-rc8-1-ge3f677a2...
62:            └ <function test_retire_primary at 0x7f5da9bfb280>
62:   File "/__w/1/s/tests/suite/test_requirements.py", line 48, in wrapper
62:     return func(network, args, *nargs, **kwargs)
62:            │    │        │      │        └ {}
62:            │    │        │      └ ()
[...]

PR #3810, opened 2022-05-04
Raft: Do not grant vote if leader is known

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/3808

We currently grant a vote for a candidate in term `T`, even if we know the (other) primary in `T` but didn't vote for them. This could happen if the primary in `T` was elected by other nodes and we found out about their primary-ship through an AppendEntries message but never had the chance to grant a vote for them. 

We now check that the primary is unknown before granting a vote. 

TODO: 
- [x] Unit test


src/consensus/aft/raft.h
```diff
@@ -1614,6 +1614,19 @@ namespace aft
         become_aware_of_new_term(r.term);
       }
 
+      if (leader_id.has_value())
+      {
+        // Reply false, since we already know the leader.
```
achamayou, line 1619:
```suggestion
        // Reply false, since we already know the leader in this term
```

achamayou (approved):

PR #3816, opened 2022-05-05
Stricter JSON configuration schema for `cchost` 

### LINKED ISSUE #3813: cchost configuration check ignores extra fields
**Describe the bug**
`cchost --check -c $configFile` allows users to verify the configuration file and exit. Presumably, the same schema check is done at runtime when `cchost` is used to run a CCF app. This configuration check does not throw an error if unexpected or extra fields are provided.

**To Reproduce**
Create a minimal `config.json` as in the sample [`start_config.json`](https://github.com/microsoft/CCF/blob/main/samples/config/start_config.json). Add a field like the following:
```json
"snapshot": {
  "directory": "/mnt/snapshots",
  "tx_count": 10000
}
```
Note that the correct field is `snapshots`, plural. `cchost --check -c config.json` outputs
```bash
{"h_ts":"2022-05-04T17:12:47.912147Z","thread_id":"100","level":"info","file":"../src/host/main.cpp","number":"113","msg":"CCF version: ccf-2.0.0-rc8\n"}
{"h_ts":"2022-05-04T17:12:47.912196Z","thread_id":"100","level":"info","file":"../src/host/main.cpp","number":"117","msg":"Configuration file successfully verified\n"}
```
and `cchost -c config.json` starts a CCF node normally.

**Expected behavior**
`cchost` should throw an exception when it encounters an unknown field.

**Environment information**
CCF 2.0rc8

**Additional context**
By accepting unknown fields as part of the configuration, typos in the configuration file can result in hard-to-detect misconfiguration.

jumaffre:
Resolves #3813 

We did not throw an error when an operator started a node with extra entries in the JSON configuration file. This was because we parsed the configuration file into the `host::CCHostConfig` configuration type first (`nlohmann::json(config_str)`), which removed the extra entries. We now do this with `nlohmann::json::parse(config_str)`, which we feed directly to `valijson`.

I also updated the documentation generator to accommodate this and fixed a couple of inconsistencies in some of the sample configuration files.

### DIFF
--- a/doc/generate_config_rst.py
+++ b/doc/generate_config_rst.py
@@ -88,9 +88,7 @@ def print_object(output, obj, depth=0, required_entries=None, additional_desc=No
                 )
                 # Strict schema with no extra fields allowed https://github.com/microsoft/CCF/issues/3813
                 assert (
-                    "allOf" in v
-                    or "additionalProperties" in v
-                    and v["additionalProperties"] is False
+                    "allOf" in v or v.get("additionalProperties") == False
                 ), f"AdditionalProperties not set to false in {k}:{v}"
             if "additionalProperties" in v:
                 if isinstance(v["additionalProperties"], dict):
@@ -135,7 +133,7 @@ def generate_configuration_docs(input_file_path, output_file_path):
         output, j["properties"], required_entries=j["required"], depth=START_DEPTH
     )
     assert (
-        "additionalProperties" in j and j["additionalProperties"] is False
+        j.get("additionalProperties") == False
     ), f"AdditionalProperties not set to false in top level schema"
 
     out = output.render()
--- a/doc/generate_config_rst.py
+++ b/doc/generate_config_rst.py
@@ -86,6 +86,10 @@ def print_object(output, obj, depth=0, required_entries=None, additional_desc=No
                 print_object(
                     output, v["properties"], depth=depth + 1, required_entries=reqs
                 )
+                # Strict schema with no extra fields allowed https://github.com/microsoft/CCF/issues/3813
+                assert (
+                    "additionalProperties" in v and v["additionalProperties"] is False
+                ), f"AdditionalProperties not set to false in {k}"
             if "additionalProperties" in v:
                 if isinstance(v["additionalProperties"], dict):
                     print_object(
@@ -128,6 +132,9 @@ def generate_configuration_docs(input_file_path, output_file_path):
     print_object(
         output, j["properties"], required_entries=j["required"], depth=START_DEPTH
     )
+    assert (
+        "additionalProperties" in j and j["additionalProperties"] is False
+    ), f"AdditionalProperties not set to false in top level schema"
 
     out = output.render()
     # Only update output file if the file will be modified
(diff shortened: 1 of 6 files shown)


doc/generate_config_rst.py
```diff
@@ -61,6 +61,8 @@ def print_entry(output, entry, name, required=False, depth=0):
 
 
 def has_subobjs(obj):
+    if not isinstance(obj, dict):
```
achamayou, line 64:
```suggestion
    if not isinstance(obj, collections.abc.Mapping):
```

is more generic, although not in any way that matters here.

doc/generate_config_rst.py
```diff
@@ -84,13 +86,20 @@ def print_object(output, obj, depth=0, required_entries=None, additional_desc=No
                 print_object(
                     output, v["properties"], depth=depth + 1, required_entries=reqs
                 )
+                # Strict schema with no extra fields allowed https://github.com/microsoft/CCF/issues/3813
+                assert (
+                    "allOf" in v
+                    or "additionalProperties" in v
+                    and v["additionalProperties"] is False
```
achamayou, line 93:
```suggestion
                    v.get("additionalProperties") == False
```

is terser, and does not rely on precedence.

doc/generate_config_rst.py
```diff
@@ -121,6 +134,9 @@ def generate_configuration_docs(input_file_path, output_file_path):
     print_object(
         output, j["properties"], required_entries=j["required"], depth=START_DEPTH
     )
+    assert (
+        "additionalProperties" in j and j["additionalProperties"] is False
```
achamayou, line 138:
```suggestion
        j.get("additionalProperties") == False
```

achamayou (approved):

src/host/main.cpp
```diff
@@ -101,6 +91,17 @@ int main(int argc, char** argv)
       schema_error_msg.value()));
   }
 
+  host::CCHostConfig config = {};
+  try
+  {
+    config = nlohmann::json::parse(config_str);
```
eddyashton, line 97:
```suggestion
    config = config_json;
```

Avoid a double parse?

doc/host_config_schema/cchost_config.json
```diff
@@ -250,10 +265,11 @@
           "minimum": 1
         }
       },
-      "description": "This section includes configuration for the node x509 identity certificate"
+      "description": "This section includes configuration for the node x509 identity certificate",
+      "additionalProperties": false
     },
     "node_data_json_file": {
-      "type": "string",
+      "type": ["string", "null"],
```
eddyashton, line 272: I don't understand the purpose of these changes. Do we want/expect `"type": null` entries in the JSON config? I would have thought these were optional fields, but if present they have to be string?

jumaffre: We set `node_data_json_file` to `null` in the Python infra because of the [Jinja template we currently use](https://github.com/microsoft/CCF/blob/main/tests/config.jinja#L17). We could modify the template to not include the field when it's `null` but I thought allowing a `null` value for all optional configuration entries that refer to files give a little more flexibility.

eddyashton: In general, I think treating both missing keys and `"key": null` as valid is an unnecessary complication, and not flexibility that we want. It makes manually type-checking and parsing the doc more complicated. That said, maybe it's the right choice in this instance, as it makes the generated configs that deliberately omit this value more explicit.

PR #3826, opened 2022-05-10
Remove duplicate lines

eddyashton:
These lines are repeated above. Minor cleanup, no functional impact.


achamayou (approved):
Should be applied to 2.0

Issue #3830, opened 2022-05-11
UDP echo server test contains a use-after-free


eddyashton:
This is consistently triggered in the instrumented test. The sanitizer identifies a heap-use-after-free, the node crashes, but the test times out.

Example here:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=45285&view=logs&j=62a51552-835c-5bc7-8b94-1c4e1e3fa3fa&t=e9f61859-18a9-578f-c07c-850acd97df57&l=37045

From a local repro, the ASan error is:

```
=================================================================
==1073411==ERROR: AddressSanitizer: heap-use-after-free on address 0x6020000c7490 at pc 0x00000049ce0a bp 0x7fad401f8a90 sp 0x7fad401f8258
READ of size 16 at 0x6020000c7490 thread T2
    #0 0x49ce09 in __asan_memcpy (/data/src/1.CCF/build.inst/cchost+0x49ce09)
    #1 0x7fad443a612a in ringbuffer::Writer::write_bytes(std::__1::optional<unsigned long> const&, unsigned char const*, unsigned long) /data/src/1.CCF/build.inst/../src/ds/ring_buffer.h:342:9
    #2 0x7fad443b8ec2 in oversized::Writer::write_bytes(std::__1::optional<unsigned long> const&, unsigned char const*, unsigned long) /data/src/1.CCF/build.inst/../src/ds/oversized.h:259:35
    #3 0x7fad4432cee0 in auto bool ringbuffer::AbstractWriter::write_multiple<ringbuffer::MessageSerializers<1614725988u>, long&, short&, std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >&, serializer::ByteRange>(unsigned int, bool, long&, short&, std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >&, serializer::ByteRange&&)::'lambda'(auto const&)::operator()<std::__1::shared_ptr<serializer::MemoryRegionSection> >(auto const&) const /data/src/1.CCF/build.inst/../src/ds/ring_buffer_types.h:123:16
[...]

PR #3831, opened 2022-05-11
Fix ownership issue in UDP echo test

eddyashton:
Currently broken, tracking in #3830. ~~Disabling to fix the Daily build.~~

Also converting the failure into a sensible timeout.

EDIT: Found and the actual bug, just variable shadowing.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Piou Piou Piou
+Pilau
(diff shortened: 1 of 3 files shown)


eddyashton:
This fixes a consistent Daily failure, and I believe the failure on this run (in `rotation_test_cft`) is unrelated, so I'm force-merging.

rengolin:
d'oh!

Issue #3770, opened 2022-04-14
JWT auto key refresh not returning success or failure


caleteeter:
**Describe the bug**
When enabling JWT auto key refresh feature using Azure AD as the IDP, the request is formed but the handler is never called for either success or failures.  Enabling verbose and debug logging nets the following results:

![log](https://user-images.githubusercontent.com/8452367/163472464-2e0e0439-64a0-43f9-8157-7d49ceba6206.png)

The logs show the request being formed.  I added some additional logging and found that I can log up to this point

https://github.com/microsoft/CCF/blob/1777feed00cd5c894caceb65958e38e2553d409a/src/node/jwt_key_auto_refresh.h#L336

I never see the handler called (for the response)

https://github.com/microsoft/CCF/blob/1777feed00cd5c894caceb65958e38e2553d409a/src/node/jwt_key_auto_refresh.h#L191

I have tested this both AAD and OKTA.  Additionally, I have added incorrect urls for the IDP and no errors are shown in the logs as expected.

**To Reproduce**
1) Build ccf 2.0.0-rc6, start the tests app (js-authentication) with the following options:
  ```
  sudo /opt/ccf/bin/sandbox.sh --jwt-key-refresh-interval-s 10 -l debug --js-app-bundle $HOME/CCF/tests/js-authentication
  ```
2) Tail the log in /workspace/sandbox_0/out

3) Submit proposals to add ca_cert an jwt_issuer

```
{
    "actions": [
        {
            "name": "set_jwt_issuer",
            "args": {
                "issuer": "https://login.microsoftonline.com/common/v2.0",
                "key_filter": "all",
                "ca_cert_bundle_name": "jwt_ms",
                "auto_refresh": true
            }
        }
    ]
}
```

Observe the logs as shown above every 10 seconds, but no response.

**Expected behavior**
Expect the handler to be called after the http client returns the content from the IDP.

**Environment information**
- CCF version 2.0.0-rc6
- Ubuntu 20.04LTS

achamayou:
When either the Digicert root, or both the root and the MSFT cert are in the ca_store, we see:

`Pre-verify failed (20) at depth 1: unable to get local issuer certificate`

When only the MSFT cert is in store, we see:

`Pre-verify failed (2) at depth 1: unable to get issuer certificate`

I've tried with and without terminating line returns, but that has no impact.

achamayou:
CLI verification of the chain works, starting with a negative test to confirm local CA store isn't used:

```
[...]

PR #3840, opened 2022-05-16
Fix counting of authentication exceptions in endpoint metrics

eddyashton:
Discovered while following up on something from #3770:

> - Don't increment successes in /jwt_metrics when a manual /jwt_keys/refresh returns 500.

The reason that was happening is 2-fold.
1. First, exceptions thrown by an authentication policy result are caught in `http_endpoint.h`  rather than the main loop in `frontend.h`, and don't get counted in endpoint metrics. I've fixed that by extending the scope of the generic `try-catch`, to cover all of the code once the `endpoint` is known.
2. Second, the node cert authentication policy was throwing an OpenSSL error when called anonymously (eg, by `curl https://127.0.0.1:8000/node/jwt_keys/refresh -k -X POST`). I've converted this to a nicer error working within the authn policy chain system. A missing cert means you can't use `[User|Member|Node]CertAuthn`, where previously we'd calculate the empty digest and that _could_ have been accepted by governance, presumably resulting in weirdness later.

### DIFF
--- a/src/node/rpc/frontend.h
+++ b/src/node/rpc/frontend.h
@@ -363,7 +363,7 @@ namespace ccf
             update_metrics(ctx, endpoint);
             return ctx->serialise_response();
           }
-          catch (JsonParseError& e)
+          catch (const JsonParseError& e)
           {
             ctx->set_error(
               HTTP_STATUS_BAD_REQUEST,
--- a/src/node/rpc/frontend.h
+++ b/src/node/rpc/frontend.h
@@ -182,227 +182,235 @@ namespace ccf
       // are not counted against any particular endpoint.
       endpoints.increment_metrics_calls(endpoint);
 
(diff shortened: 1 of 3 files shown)


jumaffre (approved):

src/node/rpc/frontend.h
```diff
@@ -182,227 +182,235 @@ namespace ccf
...
-        {
+          catch (const kv::CompactedVersionConflict& e)
+          {
+            // The executing transaction failed because of a conflicting
+            // compaction. Reset and retry
+            LOG_DEBUG_FMT(
+              "Transaction execution conflicted with compaction: {}", e.what());
+            continue;
+          }
+          catch (RpcException& e)
```
achamayou, line 360: Why not const?

achamayou (approved):

PR #3860, opened 2022-05-19
Correct RC version detection in Python infra

eddyashton:
The `lts_compatibility` test is currently failing on `release/2.x`, because this `version_rc` function fails to parse versions like `ccf-2.0.0-1-g51f57be9a`. Now it recognized that they're not a release candidate.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Rice nice
+Nice rice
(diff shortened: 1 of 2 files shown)


wintersteiger (approved):
Many thanks!

Issue #2577, opened 2021-05-10
Raft extensions for omission faults


jumaffre:
Our current implementation of CFT consensus still suffers from some limitations and liveness issues around omission faults (i.e. dropped messages). This is especially the case if one or several nodes are partitioned out of the service, as demonstrated by the end-to-end tests introduced in #2553: a single partitioned backup will automatically become primary if it was partitioned for >= `election_timeout`. 

Note that this is only true when no new write transactions are processed by the current leader. Otherwise, the partitioned node wouldn't be able to win an election as its last known `seqno` would be behind.

The following two extensions should help mitigate this family of issues:

### 1. PreVote 

Each potential candidate should first check that a quorum of nodes would accept this node as the primary should it become candidate. Only then the node should transition to a candidate state and request votes from other nodes.

Other nodes should respond to `PreVote` messages as if it was a real election, but don't need to keep track of which nodes they have granted their `PreVote`. It is only when a quorum of nodes have responded positively to the `PreVote` round that the node can become candidate.

### 2. Leader stickiness/CheckQuorum

The goal here is to make sure that a primary stays primary for as long as possible, i.e. doesn't step down because one node _only_ started an election.

Nodes should grant their `PreVote`/`Vote`s if they haven't heard from a primary within their election timeout. As I understand it, this implies that a node should only grant `PreVote`/`Vote`s when it is already in the new "campaign" (is this a good name for it?) or candidate state.

Moreover, a primary should actively step down (i.e. become a follower in the same term of its primary-ness) if it hasn't heard `AppendEntries` responses from a majority of backups within the election timeout.

Note that this also impacts the "sunny day" election scenario as the first half of the nodes whose election timeout expires wouldn't manage to get a quorum of nodes (because these ones still known about the current leader and haven't yet timed out). This is also a positive change as this would basically average out the election timeout of the service over a quorum of nodes rather than be set by the nodes with the smallest election timeout.

**Sources:**
- https://decentralizedthoughts.github.io/2020-12-12-raft-liveness-full-omission/
- https://www.openlife.cc/sites/default/files/4-modifications-for-Raft-consensus.pdf

heidihoward:
This short workshop paper might be useful: https://dl.acm.org/doi/pdf/10.1145/3447851.3458739

heidihoward:
[...]

PR #3865, opened 2022-05-19
CheckQuorum Raft extension

jumaffre:
Introduces the Raft CheckQuorum extension described in #2577 

The primary node now automatically steps down as a backup, in the same term, when it hasn't heard back from a majority of backup nodes within an election timeout.

This is tested by the existing partitions test (specifically `test_isolate_and_reconnect_primary`) as well as new `check_quorum` test scenario.


doc/host_config_schema/cchost_config.json
```diff
@@ -202,7 +202,7 @@
                   "retry_timeout": {
                     "type": "string",
                     "default": "1000ms",
-                    "description": "Interval (time string) at which the node sends join requests to the existing network"
+                    "description": "Interval (time string) at which the node sends join requests to the existing service. This should be less than the value of 'consensus.election_timeout' set on the primary node of the existing service to join."
```
jumaffre, line 205: This is just documentation to avoid the primary node detecting new joiners as inactive in between the time they have been trusted by the consortium and the time they actually join the network ([more details here](https://microsoft.github.io/CCF/main/architecture/consensus/1tx-reconfig.html)). The default values, 5s for the election timeout and 1s for the join retry, are sensible.

src/consensus/aft/raft.h
```diff
@@ -275,17 +275,6 @@ namespace aft
       return membership_state == kv::MembershipState::RetirementInitiated;
     }
 
-    ccf::NodeId get_primary(ccf::View view)
```
jumaffre, line 278: I deleted some unused code here.

src/consensus/aft/raft.h
```diff
@@ -883,9 +872,27 @@ namespace aft
...
+        {
+          // CheckQuorum: The primary automatically steps down if it has not
+          // heard back from a majority of backups during an election timeout.
+          LOG_INFO_FMT(
+            "Stepping down as follower {}: No ack received from a majority {} "
+            "of backups in last {}",
+            state->my_node_id,
+            backup_ack_timeout_count,
+            election_timeout);
+          become_follower();
```
jumaffre, line 895: Note that this calls `become_follower()` rather than `become_aware_of_new_term(term)`. This is because this is a weaker transition than finding out about a new term. In particular, the stepping down node still keeps track of who it voted for in this term.

heidihoward:
Might want to add a note about CheckQuorum to the docs here: https://microsoft.github.io/CCF/main/architecture/consensus/index.html#cft-consensus-protocol

heidihoward:
I'm so sorry, I accidentally marked this as ready for review. Not sure how to undo :(

heidihoward:
A leader who has stepped down might receive AppendEntries/RequestVote responses for the current term. Both ``recv_request_vote_response`` and ``recv_append_entries_response`` seem like they will ignore the messages in this case, however, it seems that ``recv_append_entries_response`` does not log the ignored messages to debug unlike ``recv_request_vote_response`` does. It might be helpful to add that

src/consensus/aft/raft.h
```diff
@@ -883,9 +872,34 @@ namespace aft
...
+          {
+            LOG_FAIL_FMT(
+              "For node {}, last_ack_timeout: {}",
+              node.first,
+              node.second.last_ack_timeout);
+            backup_ack_timeout_count++;
+          }
+        }
+
+        if (backup_ack_timeout_count >= get_quorum(nodes.size()))
```
heidihoward, line 889: Is there anything stopping someone from deploying an even number of nodes? In theory, this should be n-q+1 not q, where n is number of nodes and q is the quorum size.

jumaffre: Deploying an even number of nodes is supported by CCF, even though it's preferable to deploy an odd number of nodes. In `raft.h`, `nodes` only keeps track of other nodes in the service and doesn't include ourselves. This means that on the primary node, `nodes` contains all the backup nodes. When a majority of them (or more) times out, this means that it's likely that they're electing their own new primary and that we should step down.

Examples:
- 2-node network, q=2, q(`nodes`)=1, when 1 backup times out, we step down
- 3-node network, q=2, q(`nodes`)=2, when 2 backups time out, we step down
- 4-node network, q=3, q(`nodes`)=2, when 2 backups time out, we step down
- 5-node network, q=3, q(`nodes`)=3, when 3 backups time out, we step down

jumaffre:
> A leader who has stepped down might receive AppendEntries/RequestVote responses for the current term. Both `recv_request_vote_response` and `recv_append_entries_response` seem like they will ignore the messages in this case, however, it seems that `recv_append_entries_response` does not log the ignored messages to debug unlike `recv_request_vote_response` does. It might be helpful to add that

Good point. This is now done.

tests/infra/e2e_args.py
```diff
@@ -218,7 +218,7 @@ def cli_args(add=lambda x: None, parser=None, accept_unknown=False):
         "--join-timer-s",
```
jumaffre, line 218: This is to avoid marking a backup as "timed out" when it's been trusted but hasn't yet polled the primary again to join the consensus. This was set to a high value here historically, because the quote verification was expensive on the primary. I believe that recent changes to the DCAP client means that we cache quotes + collaterals a lot more efficiently now so this shouldn't be an issue anymore.

tests/infra/network.py
```diff
@@ -736,22 +736,20 @@ def trust_node(
         try:
             if self.status is ServiceStatus.OPEN:
                 valid_from = valid_from or datetime.now()
-                timeout = ceil(args.join_timer_s * 2)
+                # Note: Timeout is function of the ledger size here since
```
jumaffre, line 739: Changing some of the timeout logic here that was actually silently affected by previous changes (e.g. only opening the RPC interface when the service open transaction has been committed, rather than as soon as we've joined).

doc/architecture/consensus/index.rst
```diff
@@ -20,6 +20,18 @@ CFT parameters can be configured when starting up a network (see :doc:`here </op
 - ``consensus.message_timeout`` is the Raft heartbeat timeout. The Raft leader sends heartbeats to its followers at regular intervals defined by this timeout. This should be set to a significantly lower value than ``consensus.election_timeout``.
 - ``consensus.election_timeout`` is the Raft election timeout. If a follower does not receive any heartbeat from the leader after this timeout, the follower triggers a new election.
 
+Extensions for Omission Faults
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
achamayou, line 24: A link to a definition of omission fault would be a good thing to have in this paragraph.

achamayou (approved):

tests/partitions_test.py
```diff
@@ -192,6 +200,7 @@ def test_isolate_and_reconnect_primary(network, args, **kwargs):
             # arbitrarily allow 3 time periods to avoid being too brittle when
             # raft timeouts line up badly.
             c.wait_for_commit(new_tx_resp, timeout=(network.election_duration * 4))
+            r = c.get("/node/consensus").body.json()["details"]
```
jumaffre, line 203:
```suggestion
```

src/consensus/aft/raft.h
jumaffre: As discussed, we're keeping the same logic for a 2-node network, even if the primary knows that the partitioned backup cannot elect a new primary on its own. This is mostly for consistency logic for all network configuration cases.

jumaffre:
Merging this PR as the current daily failure is unrelated (GitHub was down while the LTS compatibility test tried to download the latest release).

PR #3866, opened 2022-05-19
Log exceptions thrown during enclave init

eddyashton:
We convert errors thrown during enclave creation into an appropriate error code, but lose any specific message associated with the exception. This PR now logs those at TRACE level, so they're available in Unsafe builds (to discuss - is it safe to log these all the time?).

What this means in practice is that if the app tries to (for instance) create an indexing strategy over a private table, it would previously crash on startup with the opaque:

```
An error occurred when creating CCF node: EnclaveInitFailed
```

But that's now preceded by a more informative log message:
```
2022-05-19T14:22:36.489171Z        0   [trace] ../src/enclave/enclave.h:144         | Creating RPC actors / ffi
2022-05-19T14:22:36.589253Z        0   [trace] ../src/enclave/main.cpp:251          | exception during enclave init: This Strategy (SeqnosByKey records) is currently only implemented for public tables, so cannot be used for 'records'
2022-05-19T14:22:36.589364Z        100 [fail ] ../src/host/main.cpp:549             | An error occurred when creating CCF node: EnclaveInitFailed
```


jumaffre (approved):

src/enclave/main.cpp
```diff
@@ -231,20 +231,27 @@ extern "C"
...
     }
-    catch (const ccf::ccf_openssl_rdrand_init_error&)
+    catch (const ccf::ccf_openssl_rdrand_init_error& e)
     {
+      LOG_TRACE_FMT(
+        "ccf_openssl_rdrand_init_error during enclave init: {}", e.what());
       return CreateNodeStatus::OpenSSLRDRANDInitFailed;
     }
-    catch (const std::exception&)
+    catch (const std::exception& e)
```
achamayou, line 252: Add a comment explaining why e.what() can't contain anything that would impact confidentiality or integrity at this point.

PR #3867, opened 2022-05-19
Add published_address to node-to-node interface configuration

wintersteiger:
This adds a `published_address` to the configuration of the node-interface. The reason to add this is that my nodes run in different virtual networks, where they may have the same `bind_address`, or when the `bind_address` is not an internet-reachable address, e.g. `0.0.0.0:0`.

In my specific use-case, I have two nodes on Azure VMs in different vnets. One of them on `10.1.0.4` and the other on `172.18.0.4` and they do not themselves know what their actual IPs are. I can make them bind to `0.0.0.0`, but then that address is put into the nodes table and consequently nodes can't connect to each other when they use the address from the nodes table. Thus, I now put `published_address` into the nodes table, which defaults to the `bind_address` if not specified.

Many thanks to @jumaffre for figuring out how to solve this problem!

### DIFF
--- a/doc/host_config_schema/cchost_config.json
+++ b/doc/host_config_schema/cchost_config.json
@@ -35,7 +35,7 @@
             "published_address": {
               "type": "string",
               "default": "Value of 'bind_address'",
-              "description": "The published RPC address advertised to other nodes"
+              "description": "The published node address advertised to other nodes"
             }
           },
           "description": "Address (host:port) to listen on for incoming node-to-node connections (e.g. internal consensus messages)",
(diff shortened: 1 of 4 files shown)


doc/host_config_schema/cchost_config.json
```diff
@@ -31,6 +31,11 @@
             "bind_address": {
               "type": "string",
               "description": "Local address the node binds to and listens on"
+            },
+            "published_address": {
+              "type": "string",
+              "default": "Value of 'bind_address'",
+              "description": "The published RPC address advertised to other nodes"
```
eddyashton, line 38:
```suggestion
              "description": "The published node address advertised to other nodes"
```

PR #3870, opened 2022-05-20
Combine docs on running sandbox.sh demo app

### LINKED ISSUE #3863: Sandbox examples in docs switch between CCF installed under ``/opt`` & compiled from source
https://microsoft.github.io/CCF/main/build_apps/install_bin.html suggests installing CCF using the debian package.

The examples in https://microsoft.github.io/CCF/main/build_apps/run_app.html use the sandbox from ``opt/ccf/bin`` but the next page https://microsoft.github.io/CCF/main/build_apps/demo.html uses the sandbox from ``CCF/tests/sandbox.sh``.

eddyashton:
Resolves #3863.

As @jumaffre pointed out, `demo.rst` and `run_app.rst` have a very similar focus. Rather than trying to unpick them for separate audiences, I've combined them into a single doc, focussed on how `sandbox.sh` is used to run either installed apps, or apps built against the installed CCF. I think I've saved all of the interesting content from `demo.rst`. I've also moved the `Debugging` section, which actually talks about building a debuggable enclave, to `build_app`.


doc/build_apps/run_app.rst
```diff
@@ -1,20 +1,35 @@
 Running CCF Applications
```
jumaffre, line 1: Should this be more specific? e.g. "Running CCF Applications in Sandbox"? Otherwise it suggests that this is a generic way to deploy CCF applications.

jumaffre (approved):

doc/build_apps/run_app.rst
eddyashton: I've tried to clarify that with a comment below:

> ``sandbox.sh`` is a demonstration tool, and not intended for use in production deployments.

PR #3873, opened 2022-05-20
Add sample of value-based indexing strategy

eddyashton:
Expanding samples of how historical queries/indexes can be used, with a strategy defined entirely in app-space. There's a few more things to do here, though they might not all make this PR:

- [ ] Add tests for this new endpoint
- [ ] Add docs referencing this app-space strategy.
- [ ] Factor out the common bits of historical endpoints. Now that we have several different range/sparse historical queries, we have a lot of duplicated code. We should factor this out, and see if it makes sense to create another `adapter` for historical queries.
- [ ] See if we can build an intermediate data structure for dumping data from an index like this to LFS. I don't think they should all access it directly, but I don't yet see the API for doing so generically.

### DIFF
--- a/samples/apps/logging/index_by_value.h
+++ b/samples/apps/logging/index_by_value.h
@@ -30,7 +30,7 @@ namespace loggingapp
       const ccf::TxID tx_id;
       const ccf::ByteVector key;
       const ccf::ByteVector value;
-    }
+    };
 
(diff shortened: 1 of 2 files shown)


eddyashton:
Closing this PR as not an immediate priority. Will revisit when we add more samples.

PR #3877, opened 2022-05-20
Add ACME client for globally endorsed TLS certificates

wintersteiger:
This adds an ACME client that allows us to obtain globally endorsed TLS certificates via services like Let's Encrypt.


doc/host_config_schema/cchost_config.json
```diff
@@ -467,6 +467,48 @@
...
       "description": "This section includes configuration for the host-enclave ring-buffer memory (modify with care!)",
       "additionalProperties": false
+    },
+    "acme_client_config": {
+      "type": "object",
+      "properties": {
+        "ca_cert": {
+          "type": "string",
+          "default": "",
+          "description": "Certificate of the ACME server we will connect to (to avoid certificate verification failures)"
```
achamayou, line 477: PEM?

doc/host_config_schema/cchost_config.json
```diff
@@ -467,6 +467,48 @@
...
+          "type": "string",
+          "default": "",
+          "description": "DNS name of the service we represent"
+        },
+        "contact": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "Contact address(es)"
```
achamayou, line 494: Email?

doc/host_config_schema/cchost_config.json
```diff
@@ -467,6 +467,48 @@
...
+        },
+        "terms_of_service_agreed": {
+          "type": "boolean",
+          "default": false,
+          "description": "Indication that the user/operator is aware of the latest terms and conditions"
```
achamayou, line 499:
```suggestion
          "description": "Indication that the user/operator is aware of the latest terms and conditions for the CA"
```

src/crypto/openssl/verifier.cpp
```diff
@@ -180,4 +180,15 @@ namespace crypto
...
+
+  size_t Verifier_OpenSSL::remaining_seconds() const
+  {
+    auto [from, to] = validity_period();
+    auto tp_from = ds::time_point_from_string(from);
+    auto tp_to = ds::time_point_from_string(to);
+    auto now = std::chrono::system_clock::now();
+    return std::chrono::duration_cast<std::chrono::seconds>(tp_to - now)
+             .count() +
+      1;
```
achamayou, line 192: Why +1?

wintersteiger: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.5

x509 times are inclusive, i.e. if from/to are set to the same time, the validity period is 1 second.

doc/host_config_schema/cchost_config.json
wintersteiger: It's more complicated than that, I'll expand.

src/node/acme_client.h
```diff
@@ -0,0 +1,1213 @@
...
+    {
+      using namespace std::chrono_literals;
+      using namespace std::chrono;
+
+      bool ok = true;
+      system_clock::duration delta(0);
+
+      if (last_request && !override_time)
+      {
+        // Let's encrypt recommends this retry strategy in their integration
```
jumaffre, line 82: Add link to integration guide here?

doc/host_config_schema/cchost_config.json
```diff
@@ -467,6 +484,65 @@
...
+              }
+            },
+            "description": "ACME Configuration",
+            "additionalProperties": false
+          },
+          "description": "ACME Configurations"
+        },
+        "challenge_server_interface": {
+          "type": "string",
+          "default": "0.0.0.0:80",
```
jumaffre, line 540: I don't think we should impose a default here. However, we should mention in the description that this interface should be accessible publicly via port 80.

include/ccf/crypto/verifier.h
```diff
@@ -202,6 +202,13 @@ namespace crypto
     /** The validity period of the certificate */
     virtual std::pair<std::string, std::string> validity_period() const = 0;
 
+    /** The number of seconds of the validity period of the
+     * certificate remaining */
+    virtual size_t remaining_seconds() const = 0;
```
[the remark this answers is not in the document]
wintersteiger: Good point, I'll pull that out.

src/tls/ca.h
```diff
@@ -14,18 +14,31 @@ namespace tls
   class CA
   {
   private:
-    Unique_X509 ca;
+    std::vector<Unique_X509> cas;
+    bool partial_ok = false;
```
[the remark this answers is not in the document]
wintersteiger: For testing with intermediate CAs, e.g. the Let's Encrypt staging environment.

tests/acme_endorsement.py
```diff
@@ -0,0 +1,338 @@
...
+    ca = get_without_cert_check("https://" + mgmt_address + "/roots/0")
+    intermediate = get_without_cert_check(
+        "https://" + mgmt_address + "/intermediates/0"
+    )
+    return ca, intermediate
+
+
+@reqs.description("Test against a local pebble CA")
+def run_pebble(args):
+    binary_url = "https://github.com/letsencrypt/pebble/releases/latest/download/pebble_linux-amd64"
```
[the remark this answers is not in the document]
wintersteiger: I will look into this later today, but I'm not familiar with docker at all. Would that go into `docker/ccfci` then?

include/ccf/service/tables/service.h
```diff
@@ -31,10 +31,13 @@ namespace ccf
     ServiceStatus status = ServiceStatus::OPENING;
     /// Version of previous service identity (before the last recovery)
     std::optional<kv::Version> previous_service_identity_version = std::nullopt;
+    /// ACME interface certificates
+    std::optional<std::map<std::string, crypto::Pem>> acme_certificates;
```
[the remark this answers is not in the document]
wintersteiger: Good point, I'll move that into a new table.

src/host/acme_challenge_server.h
```diff
@@ -0,0 +1,283 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
wintersteiger: TBH, I like this being on the host side and the extra dependency doesn't bother me. This can all be completely untrusted, so can stay out of the TCB. 

The ACME-endorsed interface currently serves encrypted requests until it's ACME-endorsed, but with the network-endorsed certificate. I'm not sure it's a good idea to add a non-encrypted interface even part time, because users may easily break security guarantees without being aware of it.

doc/host_config_schema/cchost_config.json
```diff
@@ -467,6 +484,65 @@
       },
       "description": "This section includes configuration for the host-enclave ring-buffer memory (modify with care!)",
       "additionalProperties": false
+    },
+    "acme": {
```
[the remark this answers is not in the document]
wintersteiger: Hmmm. I'll move it to `network.rpc_interfaces.acme_configurations`, I think that's the right scope for it.

doc/host_config_schema/cchost_config.json
wintersteiger: This is just a default value, which will work for most systems I think. Note that the port 80 restriction only applies to Let's Encrypt's services, but not in general. E.g. the pebble CA for testing allows us to use other ports so that we don't need system permissions for port 80. So, we can give a default, but we can't throw a hard error if someone choses a different port.

doc/host_config_schema/cchost_config.json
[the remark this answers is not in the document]
wintersteiger: No, I changed my mind, in theory non-rpc interfaces also could use ACME, so I'll move it to `network.acme_configurations`.

tests/acme_endorsement.py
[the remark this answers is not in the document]
wintersteiger: That all seems to be working now, thanks for the detailed instructions!

doc/operations/certificates.rst
```diff
@@ -78,3 +78,48 @@ The procedure that operators and members should follow is summarised in the foll
...
+    The various options are as follows:
+
+    .. code-block:: json
+
+        "network": {
+            "rpc_interfaces": { 
+                ... ,
+                "acme_endorsed_interface": { ... , "acme_configuration": "my-acme-cfg" }
+            },
+            "acme": {
```
jumaffre, line 103: I feel like this could be under `rpc_interfaces` as we have no plan to do this for node-to-node interfaces.

tests/acme_endorsement.py
```diff
@@ -0,0 +1,341 @@
...
+
+    # This requires a DNS name by which the network is reachable (which we don't have in the CI), e.g. your dev VM name.
+    # On the interface of that name, we also need port 80 to be reachable from the internet for the challenge responses (see Network on the Azure portal panel of your VM).
+    # The Let's Encrypt staging environment is described here: https://letsencrypt.org/docs/staging-environment/ (we need the CA certs for the staging environment, which are not globally endorsed).
+    # Further, to connect to Let's Encrypt, we also need their public root cert, which can be found at https://letsencrypt.org/certificates/
+    # (Clients won't need this as they usually have ISRG Root X1 installed, but our enclaves don't.)
+
+    service_dns_name = "acc-cwinter.uksouth.cloudapp.azure.com"
+
+    ca_certs = [
```
jumaffre, line 302: Is there a way to fetch dynamically instead of hardcoding them in the test?

wintersteiger: That would be nice, but I don't know of a way. Azure VMs are usually in private subnets, so reverse DNS wouldn't work and they are also configured sometimes, e.g. my dev VM says my `dnsdomainname` is `g33j0yzbu3berhjiliyh1p440h.zx.internal.cloudapp.net`

doc/operations/certificates.rst
wintersteiger: The C++ class for node-to-node and RPC interfaces is the same, we would have to split that as well then.

doc/host_config_schema/cchost_config.json
```diff
@@ -76,18 +76,81 @@
...
-                    "enum": ["Node", "Service"],
+                    "enum": ["Node", "Service", "ACME"],
                     "default": "Service",
                     "description": "The type of endorsement for the TLS certificate used in client sessions. If the endorsement is not available, client sessions will be terminated, before the TLS handshake is complete. 'Node' means self-signed, 'Service' means service-endorsed."
                   }
                 },
                 "required": ["authority"],
                 "additionalProperties": false
+              },
+              "acme_configuration": {
```
letmaik, line 87: Shouldn't that be inside `"endorsement"`?

wintersteiger: That's just an enum, but in theory we could promote it to enum+name, yes.

doc/operations/certificates.rst
```diff
@@ -78,3 +78,49 @@ The procedure that operators and members should follow is summarised in the foll
...
+
+1. Get a globally reachable DNS name for your CCF network, e.g. ``my-ccf.example.com``, which resolves to the address of at least one node in the network. Multiple nodes or a load balancer address are fine too.
+
+2. ACME `http-01 <https://letsencrypt.org/docs/challenge-types/>`_ challenges require a challenge server to be reachable on port 80 (non-negotiable). 
+To be able to bind to that port, the ``cchost`` binary may need to be given special permission, e.g. by running ``sudo setcap CAP_NET_BIND_SERVICE=+eip cchost``. Alternatively, port 80 can be redirected to a non-privileged port that ``chost`` may bind to without special permission.
```
letmaik, line 91:
```suggestion
To be able to bind to that port, the ``cchost`` binary may need to be given special permission, e.g. by running ``sudo setcap CAP_NET_BIND_SERVICE=+eip cchost``. Alternatively, port 80 can be redirected to a non-privileged port that ``cchost`` may bind to without special permission.
```

doc/operations/certificates.rst
```diff
@@ -78,3 +78,49 @@ The procedure that operators and members should follow is summarised in the foll
...
+
+3. Each interface defined in the ``chost`` configuration file can be given the name of an ACME configuration to use. The settings of each ACME configuration are defined in ``network.acme``. Note that this information is required by *all* nodes as they might have to renew the certificate(s) later.
```
letmaik, line 93:
```suggestion
3. Each interface defined in the ``cchost`` configuration file can be given the name of an ACME configuration to use. The settings of each ACME configuration are defined in ``network.acme``. Note that this information is required by *all* nodes as they might have to renew the certificate(s) later.
```

doc/operations/certificates.rst
```diff
@@ -78,3 +78,49 @@ The procedure that operators and members should follow is summarised in the foll
...
+                    "contact": ["mailto:john@example.com"],
+                    "terms_of_service_agreed": true,
+                    "challenge_type": "http-01",
+                    "challenge_server_interface": "0.0.0.0:80"
+                }
+            }
+        }
+
+
+    - ``ca_certs``: CCF will need to establish https connections with the CA, but does not come with root certificates by default and therefore will fail to establish connections. This setting is populated with on or more such certificates; e.g. for Let's Encrypt this would be their ISRG Root X1 certificate (see `here <https://letsencrypt.org/certificates/>`_) in PEM format.
```
letmaik, line 118:
```suggestion
    - ``ca_certs``: CCF will need to establish https connections with the CA, but does not come with root certificates by default and therefore will fail to establish connections. This setting is populated with one or more such certificates; e.g. for Let's Encrypt this would be their ISRG Root X1 certificate (see `here <https://letsencrypt.org/certificates/>`_) in PEM format.
```

include/ccf/service/tables/acme_certificates.h
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ccf/crypto/pem.h"
+#include "ccf/ds/json.h"
+#include "ccf/service/map.h"
+
+namespace ccf
+{
+  // Maps each interface names to a certificate
```
letmaik, line 11:
```suggestion
  // Maps each interface name to a certificate
```

doc/host_config_schema/cchost_config.json
letmaik: `"endorsement"` is an object with a single property `"authority"` (the enum). The idea was that any endorsement-related details go in as additional fields.

src/node/acme_client.h
```diff
@@ -0,0 +1,1198 @@
...
+        const std::string& header_b64,
+        const std::string& payload_b64,
+        crypto::KeyPair& signer)
+      {
+        auto msg = header_b64 + "." + payload_b64;
+        auto sig = signer.sign(s2v(msg));
+        convert_signature_to_ieee_p1363(sig);
+        auto sig_b64 = crypto::b64url_from_raw(sig);
+
+        (*this)[header_is_protected ? "protected" : "header"] = header_b64;
```
letmaik, line 481: If the header is unprotected then it's not base64-encoded. Is there a use case for this anyway?

src/node/acme_client.h
```diff
@@ -0,0 +1,1198 @@
...
+        (*this)["payload"] = payload_b64;
+        (*this)["signature"] = sig_b64;
+      }
+    };
+
+    class JWK : public nlohmann::json::object_t
```
letmaik, line 487: Why not model this and JWS like the other strongly-typed JSON classes?

doc/host_config_schema/cchost_config.json
wintersteiger: Oh yes, that's right, I'll move it there.

src/node/acme_client.h
wintersteiger: Not sure what you mean by strongly-typed, but the abstraction just makes sense.

include/ccf/service/acme_client_config.h
```diff
@@ -0,0 +1,51 @@
...
+
+    // Indication that the user/operator is aware of the latest terms and
+    // conditions for the CA
+    bool terms_of_service_agreed = false;
+
+    // Type of the ACME challenge (currently only http-01 supported)
+    std::string challenge_type = "http-01";
+
+    // Validity range (Note: not supported by Let's Encrypt)
+    std::optional<std::string> not_before;
```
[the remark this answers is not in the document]
wintersteiger: Good catch, thanks!

src/node/acme_client.h
wintersteiger: No, I was wrong, that's for unsecured JWS, not for unprotected headers. Killing it.

tests/acme_endorsement.py
```diff
@@ -0,0 +1,341 @@
...
+@reqs.description("Test against Let's Encrypt's staging environment")
+def run_lets_encrypt(args):
+
+    # This requires a DNS name by which the network is reachable (which we don't have in the CI), e.g. your dev VM name.
+    # On the interface of that name, we also need port 80 to be reachable from the internet for the challenge responses (see Network on the Azure portal panel of your VM).
+    # The Let's Encrypt staging environment is described here: https://letsencrypt.org/docs/staging-environment/ (we need the CA certs for the staging environment, which are not globally endorsed).
+    # Further, to connect to Let's Encrypt, we also need their public root cert, which can be found at https://letsencrypt.org/certificates/
+    # (Clients won't need this as they usually have ISRG Root X1 installed, but our enclaves don't.)
+
+    service_dns_name = "acc-cwinter.uksouth.cloudapp.azure.com"
```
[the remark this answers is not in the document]
jumaffre: I think it's better to avoid writing the name of your personal VM in this repo. Because this test needs some manual setup anyway, I'd rather have a comment here saying what this variable should be set to.

src/node/acme_client.h
wintersteiger: (That's not what I would call strongly typed, but that's beside the point.) Yes, I'm sure that would work, but it's an additional copy of the data structure, because you first have to create the C++ struct and then create the json object. In the JWK/JWS case, I will never ever need the C++ struct, only the json.

letmaik: I'll delegate to the others for this, maybe I'm missing something. @eddyashton @jumaffre

doc/operations/certificates.rst
jumaffre: You can decide which fields are allowed in the configuration schema, but still have the same structure for both (see https://microsoft.github.io/CCF/main/operations/configuration.html#network).

tests/acme_endorsement.py
jumaffre: Isn't it as likely to break when the certificates are refreshed? Unless these certificates have more than one year (at least) validity period.

src/node/acme_client.h
jumaffre: I think it's preferable to stick to what we already do for all serialisable structures and use the `DECLARE_JSON_...()` macros for this new type too?

tests/acme_endorsement.py
wintersteiger: Yes, all of this will break when Let's Encrypt changes anything, which they may well do long before those certificates expire. This would likely also change the names of and links to certificate files. This is the staging environment, so no hard guarantees, I just put in there what works today.

src/node/acme_client.h
wintersteiger: I don't need a serializable structure here, I just need the json, but if you think it helps I can add one anyways.

doc/operations/certificates.rst
wintersteiger: The `acme_configuration` has since moved into the `endorsement`. Is there a way to express that some field of a member must be absent?

src/node/acme_client.h
wintersteiger: Ah! I think both of you missed the fact that `class JWS` and `class JWK` have no fields, i.e. nothing to serialize. They are quite literally just json-constructors. If it feels more natural to you, I can also put them into functions that return `json` for instance.

src/node/acme_client.h
```diff
@@ -0,0 +1,1183 @@
...
+        (*this)["x"] = x;
+        (*this)["y"] = y;
+        if (alg)
+          (*this)["alg"] = *alg;
+        if (use)
+          (*this)["use"] = *use;
+        if (kid)
+          (*this)["kid"] = *kid;
+      }
+      virtual ~JWK() {}
```
achamayou, line 495:
```suggestion
      virtual ~JWK() = default;
```

src/node/acme_client.h
achamayou: That also seems fine to me, and so does a possible free function. If the other facilities offered by the macro (deserialisation and schema generation) aren't needed, I don't see why they must be used.

doc/operations/certificates.rst
wintersteiger: Just saw that those are copies in the schema instead of a common base definitions, so yes, `endorsement` is already guaranteed to be absent for node-to-node interfaces.

src/enclave/rpc_sessions.h
```diff
@@ -29,8 +29,8 @@ namespace ccf
 
   static constexpr size_t max_open_sessions_soft_default = 1000;
   static constexpr size_t max_open_sessions_hard_default = 1010;
-  static constexpr ccf::Endorsement endorsement_default =
-    ccf::Endorsement{ccf::Authority::SERVICE};
+  static const ccf::Endorsement endorsement_default = {
+    ccf::Authority::SERVICE, std::nullopt};
```
achamayou, line 33: What's the nullopt for here?

wintersteiger: The `acme_configuration`, which is optional, i.e. only present when the auhority is `ACME`.

src/host/acme_challenge_server.h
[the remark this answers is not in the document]
achamayou: This being out of TCB is limited to the current SGX situation, and the in-enclave flow benefits from various DoS protection mechanisms, which may be good to have on what must be a public http interface.

src/host/acme_challenge_server.h
```diff
@@ -0,0 +1,283 @@
...
+      ringbuffer::WriterPtr to_enclave) :
+      asynchost::SocketBehaviour<asynchost::TCP>("", ""),
+      parser(sp),
+      socket(socket),
+      lock(lock),
+      prepared_responses(prepared_responses),
+      to_enclave(to_enclave)
+    {}
+
+    virtual ~ClientBehaviour() {}
```
achamayou, line 49:
```suggestion
    virtual ~ClientBehaviour() = default;
```

src/host/acme_challenge_server.h
```diff
@@ -0,0 +1,283 @@
...
+
+    void reply(http::Response& r, const std::string& body)
+    {
+      if (!socket.is_null())
+      {
+        std::vector<uint8_t> vody(body.begin(), body.end());
+        r.set_body(vody.data(), vody.size());
```
achamayou, line 56: Feels like we should have a set_body(const std::string&) instead here.

src/host/acme_challenge_server.h
```diff
@@ -0,0 +1,283 @@
...
+        auto bytes = r.build_response();
+        socket->write(bytes.size(), bytes.data());
+      }
+    }
+
+    virtual void on_read(size_t len, uint8_t*& incoming, sockaddr sa) override
+    {
+      std::string body((char*)incoming, len);
```
achamayou, line 64: I may be missing something here, but it looks like we don't mutate body, so do we really need the copy?

src/host/acme_challenge_server.h
```diff
@@ -0,0 +1,283 @@
...
+                "ACME: challenge response for token '{}' provided to {}",
+                token,
+                socket->get_peer_name());
+            }
+          }
+        }
+      }
+      catch (const std::exception& ex)
+      {
+        http::Response r(HTTP_STATUS_BAD_REQUEST);
```
achamayou, line 145: I think we definitely want to log ex.what() here.

src/host/acme_challenge_server.h
```diff
@@ -0,0 +1,283 @@
...
+      disp,
+      ACMEMessage::acme_challenge_server_stop,
+      [this](const uint8_t* data, size_t size) {
+        listener = nullptr;
+        prepared_responses.clear();
+        LOG_DEBUG_FMT("ACME: challenge server stopped");
+      });
+  }
+
+  virtual ~ACMEChallengeServer() {}
```
achamayou, line 269:
```suggestion
  virtual ~ACMEChallengeServer() = default;
```

tests/acme_endorsement.py
```diff
@@ -0,0 +1,341 @@
...
+            config_filename,
+            "-dnsserver",
+            dns_address,
+        ],
+        stdout=out,
+        stderr=err,
+        close_fds=True,
+        env=env,
+    )
+    time.sleep(2)
```
achamayou, line 158: That's potentially brittle, it'd be better to wait for a condition, like a line printed to stdout, or the OS knowing something is listening on the socket for example.

tests/acme_endorsement.py
```diff
@@ -0,0 +1,341 @@
...
+    ]
+    data = str(
+        json.dumps(
+            {
+                "host": network_name,
+                "addresses": endorsed_hosts,
+            }
+        )
+    ).encode("utf-8")
+    urllib.request.urlopen("http://" + dns_mgmt_address + "/add-a", data=data)
```
achamayou, line 135:
```suggestion
    urllib.request.urlopen(f"http://{dns_mgmt_address}/add-a", data=data)
```

This is really a matter of preference, but I find f-strings clearer and more compact? ymmv!

tests/acme_endorsement.py
```diff
@@ -0,0 +1,341 @@
...
+            "",
+            "-management",
+            mgmt_address,
+        ],
+        stdout=out,
+        stderr=err,
+        close_fds=True,
+        env=env,
+    )
+    time.sleep(1)
```
achamayou, line 119: That seems brittle too, for the same reasons as for pebble.

tests/acme_endorsement.py
```diff
@@ -0,0 +1,341 @@
...
+            if time.time() > end_time:
+                raise TimeoutError(
+                    f"Not all nodes had the correct ACME-endorsed TLS certificate installed after {timeout} seconds"
+                )
+
+            num_ok = 0
+            for node in network.nodes:
+                iface = node.host.rpc_interfaces["acme_endorsed_interface"]
+                try:
+                    context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
```
achamayou, line 59: Why not 1.3?

wintersteiger: Never occurred to me to try that, I'll switch it and see whether anything breaks.

src/host/acme_challenge_server.h
wintersteiger: I just need that string when printing an error message, so I moved it further down. The reason I wanted a string is that I can't rely on `incoming` being zero-terminated.

tests/acme_endorsement.py
wintersteiger: Looks like that's not supported by some of the Python things. It just hangs when I try to use `PROTOCOL_TLSv1_3` without providing any information. Interestingly, it also just hangs when I use `PROTOCOL_TLSv1_5`... However, pylint does report that they both are not available.

doc/host_config_schema/cchost_config.json
```diff
@@ -76,9 +76,14 @@
                 "properties": {
                   "authority": {
                     "type": "string",
-                    "enum": ["Node", "Service"],
+                    "enum": ["Node", "Service", "ACME", "NONE"],
                     "default": "Service",
-                    "description": "The type of endorsement for the TLS certificate used in client sessions. If the endorsement is not available, client sessions will be terminated, before the TLS handshake is complete. 'Node' means self-signed, 'Service' means service-endorsed."
+                    "description": "The type of endorsement for the TLS certificate used in client sessions. If the endorsement is not available, client sessions will be terminated, before the TLS handshake is complete. 'Node' means self-signed, 'Service' means service-endorsed, 'ACME' means an ACME-capable CA, 'NONE' means no authority (for http)."
```
jumaffre, line 81: I think the `NONE` needs to be worded much strongly as it removes confidentiality for client requests.

include/ccf/actors.h
```diff
@@ -31,6 +32,10 @@ namespace ccf
       {
         return "node";
       }
+      case ActorsType::acme_challenge:
```
jumaffre, line 35: We may decide to add more endpoints under `.well_known/` in the future, so perhaps the actors type should be `well_known` to cater for future use cases?

doc/host_config_schema/cchost_config.json
wintersteiger: Good point, will do.

include/ccf/actors.h
wintersteiger: Yeah, I think so too. That's why I would have liked a two-level actor prefix.

wintersteiger:
I have no stripped the last few commits for the in-enclave frontend-based challenge server, which will go into a separate PR. This one is done and ready to merge once the CI gives us a green.

tests/infra/interfaces.py
```diff
@@ -96,6 +106,8 @@ def from_json(json):
         )
         if "endorsement" in json:
             interface.endorsement = Endorsement.from_json(json["endorsement"])
+        if "acme_configuration" in json:
+            interface.acme_configuration = json.get("acme_configuration")
```
achamayou, line 110: get() returns None if the key is missing, so you probably don't need the condition here (it's used above because getitem will raise if the key is missing).

tests/acme_endorsement.py
achamayou: Very odd, I thought it was just a thin layer on top of openssl. Anyway, quite orthogonal to the feature being added :)

jumaffre (approved):

jumaffre:
I think this'll need an entry in the changelog 💯

PR #3880, opened 2022-05-23
Add endpoint to get previous service identity

### LINKED ISSUE #3879: Add an endpoint that returns the previous service identity
In 2.x, the default constitution requires the previous service identity to be passed during the open network proposal. For security, this should be retrieved externally. But for simple cases, we could return this from the current service. If we can return all previous service identities, then trust in this service can be used to trust previously issued receipts (without explicit endorsement chains).

eddyashton:
Adds a `/node/network/previous_identity` endpoint. ~~This only works in a node started as `Recover`, and reads the identity-of-the-previous-service passed to this node on startup. That's simpler than writing it to the KV to be shared to all nodes, and reasoning about the edge cases that introduces (nodes having the wrong value while they're catching up).~~

EDIT - This is now written to the KV in the recovery-genesis transaction, so can be read by all nodes. Also tested in e2e recovery tests.

Resolves #3879.


src/node/rpc/network_identity_subsystem.h
```diff
@@ -26,5 +26,11 @@ namespace ccf
     {
       return network_identity;
     }
+
+    virtual std::optional<std::vector<uint8_t>> get_previous_identity_cert()
```
jumaffre, line 30: Do we still need to call into the node state as the certificate is stored in the KV?

src/node/rpc/node_frontend.h
```diff
@@ -821,6 +838,34 @@ namespace ccf
...
+        else
+        {
+          return make_error(
+            HTTP_STATUS_NOT_FOUND,
+            ccf::errors::ResourceNotFound,
+            "This service is not a recovery of a previous service.");
+        }
+      };
+      make_read_only_endpoint(
+        "/network/previous_identity",
```
jumaffre, line 860: Should this be `/service/previous_identity`? Or maybe, should this be folded into the existing `GET /node/network` endpoint?

jumaffre (approved):

src/node/rpc/network_identity_subsystem.h
eddyashton: Yes - this is where we get the value from that is then written to the KV.

src/node/rpc/node_frontend.h
eddyashton: Yes to the first! I thought we had multiple endpoints already nested under `/node/network`, but looking again they're mostly variations on `/node/network` itself. So this should probably be `/node/service/previous_identity`, similar to `/node/service/configuration`.

src/node/rpc/network_identity_subsystem.h
jumaffre: Why not just read it from the [`service.info`](https://microsoft.github.io/CCF/main/audit/builtin_maps.html#_CPPv4N3ccf11ServiceInfo4certE) table directly?

eddyashton:
@jumaffre and I noticed some other issues around this, so I'm fixing those as well before merging.

src/service/tables/previous_service_identity.h
```diff
@@ -0,0 +1,20 @@
...
+#include <string>
+#include <vector>
+
+namespace ccf
+{
+  using PreviousServiceIdentity = ServiceValue<crypto::Pem>;
+
+  namespace Tables
+  {
+    static constexpr auto PREVIOUS_SERVICE_IDENTITY =
```
jumaffre, line 17: This new table should be documented in `builtin_maps.rst`. Then, I remembered we had a test to check that we document all tables but the CI seems happy. I guess that's because the table is only created on recovery?

jumaffre (approved):

src/service/tables/previous_service_identity.h
achamayou: The happens at the end of the testsuite, and so should include a recovery?

achamayou: https://github.com/microsoft/CCF/blob/main/tests/suite/test_suite.py#L133

eddyashton: Yes, but only in the Daily (which wasn't previusl run on this PR, but is now failing).

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - The node-to-node interface configuration now supports a `published_address` to enable networks with nodes running in different (virtual) subnets (#3867).
+- Added a `/node/service/previous_identity` endpoint (#3880).
```
achamayou, line 21:
```suggestion
- Added a `/node/service/previous_identity` endpoint (#3880), which can be used during a recovery to look up the identity of the service before the catastrophic failure.
```

Issue #733, opened 2020-01-28
Fail fast on wrong network certificate when joining


iqsarv:
Description
`cchost` application is going on a loop when `join` command is supplied with wrong network cert

Steps to reproduce the behavior.
1. Start a new network
2. Join to the network with same IP:port but a different network certificate

Expected behavior
Join should fail fast rather than on a infinite loop

Environment information
Commit on master 316d57a80b915a3ff98a8cb5d82bdf46ce9ce06b

Additional context

```
./cchost --enclave-file=libloggingenc.so.signed         --consensus raft         --enclave-type=debug         --ledger-file=ccf4q.ledger         --node-cert-file=node_cert.pem         --quote-file=quote.bin         --host-log-level=info         --raft-election-timeout-ms=100000         --raft-timeout-ms=500         --node-address=10.240.0.68:3100         --rpc-address=10.240.0.68:3200 join --network-cert-file=network_cert.pem --target-rpc-address 10.240.0.69:3200
2020-01-28T10:17:54.540007Z        [info ] /tmp/CCF/src/host/main.cpp:376       | Creating new node - joining existing network at 10.240.0.69:3200
2020-01-28T10:17:54.783241Z        [info ] /tmp/CCF/src/host/main.cpp:399       | Created new node
2020-01-28T10:17:54.783486Z -0.243 [info ] /tmp/CCF/src/node/nodestate.h:1089   | Node TLS connections now accepted
2020-01-28T10:17:57.736523Z -0.155 [info ] F/src/enclave/framedtlsendpoint.h:37 | Received additional data for 9223372036854775811, ignoring due to status 3
2020-01-28T10:18:13.510969Z -0.842 [info ] F/src/enclave/framedtlsendpoint.h:37 | Received additional data for 9223372036854775826, ignoring due to status 3
2020-01-28T10:18:13.510990Z -0.842 [info ] F/src/enclave/framedtlsendpoint.h:37 | Received additional data for 9223372036854775826, ignoring due to status 3
2020-01-28T10:18:13.510993Z -0.842 [info ] F/src/enclave/framedtlsendpoint.h:37 | Received additional data for 9223372036854775826, ignoring due to status 3
2020-01-28T10:18:14.543082Z -0.879 [info ] F/src/enclave/framedtlsendpoint.h:37 | Received additional data for 9223372036854775827, ignoring due to status 3

```

achamayou:
The loop is there to retry in case one of two failures occurs:

1. temporary network failure
[...]

PR #3895, opened 2022-05-31
Shutdown new joining node early if target service certificate is invalid

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/733

When the target service certificate supplied to a new node is wrong (e.g. misconfigured by the operator), the node automatically shuts down with the following error message: `Early error when joining existing network at 127.147.212.20:34215: TLS -1 invalid cert on handshake: unable to get local issuer certificate [error:FFFFFFFF80000000:lib(128):func(0):reason(0)]. Shutting down node gracefully...`

### DIFF
--- a/src/enclave/client_endpoint.h
+++ b/src/enclave/client_endpoint.h
@@ -7,6 +7,9 @@
 
 namespace ccf
 {
+  // TODO: Remove duplication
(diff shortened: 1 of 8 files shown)


wintersteiger (approved):
LGTM!

PR #3917, opened 2022-06-10
Infra: make ledger file copy more resilient

jumaffre:
When the Python infra copies ledger files (e.g. when starting a new node), it is possible that a ledger file gets committed while the infra is copying the file, resulting in rare errors in the CI (e.g. [here](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=46315&view=logs&j=e982d093-d12a-5a2d-1a1d-0c6d38b8823f&t=8be5454d-0d8f-51e4-7372-58df022e9ebc&l=18006)). 

We now retry copying files if the initial copy fails (and up to 5 times), like we already do with snapshot files.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -315,13 +315,6 @@ max-line-length=100
 # Maximum number of lines in a module.
 max-module-lines=1000
 
-# List of optional constructs for which whitespace checking is disabled. `dict-
-# separator` is used to allow tabulation in dicts, etc.: {1  : 1,\n222: 2}.
-# `trailing-comma` allows a space between comma and closing bracket: (a, ).
-# `empty-line` allows space-only lines.
-no-space-check=trailing-comma,
-               dict-separator
-
 # Allow the body of a class to be on the same line as the declaration if body
 # contains single statement.
 single-line-class-stmt=no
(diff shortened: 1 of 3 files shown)


.pylintrc
```diff
@@ -315,13 +315,6 @@ max-line-length=100
 # Maximum number of lines in a module.
 max-module-lines=1000
 
-# List of optional constructs for which whitespace checking is disabled. `dict-
-# separator` is used to allow tabulation in dicts, etc.: {1  : 1,\n222: 2}.
-# `trailing-comma` allows a space between comma and closing bracket: (a, ).
-# `empty-line` allows space-only lines.
-no-space-check=trailing-comma,
```
jumaffre, line 322: This option was removed in [Pylint 2.6.0](https://github.com/PyCQA/pylint/releases/tag/pylint-2.6.0).

wintersteiger (approved):
LGTM, although I'm not sure what would happen to the ledger file being committed, i.e. if we could loose commits. It's probably fine though, because this only happens during startup/recovery.

eddyashton (approved):

PR #3918, opened 2022-06-10
Ensure `crypto::Pem`s contain PEM encoded contents

eddyashton:
Follow up to #3916, hoping to avoid similar issues in other places.

The real fix here is that when a `crypto::Pem` is constructed with contents, it confirms they look a bit like a PEM. Maybe that check should be more precise, but I think it's sufficient for now. This broke a bunch of unit tests, which used various kinds of empty PEMs, but now generate plausible test identities.

I'd like to remove the default constructor for PEMs as well, so we disallowed empty PEMs as well, at compile time. Unfortunately this is tricky to do alongside JSON serialization - we _can_ convert to `crypto::Pem` even when it is not default-constructible, but we'd need to do the same thing for all of our PoD JSON-serialized structs which hold `crypto::Pem`s (or make those `std::optional` or similar). That's tricky and messy for no gain, so I've abandoned it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -432,12 +432,12 @@ if(BUILD_TESTS)
     )
 
     add_unit_test(
-      node_frontend_test
+      node_frontend_test ${CMAKE_CURRENT_SOURCE_DIR}/src/js/wrap.cpp
       ${CMAKE_CURRENT_SOURCE_DIR}/src/node/rpc/test/node_frontend_test.cpp
     )
     target_link_libraries(
       node_frontend_test PRIVATE ${CMAKE_THREAD_LIBS_INIT} http_parser.host
-                                 sss.host ccf_endpoints.host ccf_kv.host
+                                 sss.host ccf_endpoints.host ccf_kv.host quickjs.host
     )
 
     add_unit_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -436,8 +436,9 @@ if(BUILD_TESTS)
       ${CMAKE_CURRENT_SOURCE_DIR}/src/node/rpc/test/node_frontend_test.cpp
     )
     target_link_libraries(
-      node_frontend_test PRIVATE ${CMAKE_THREAD_LIBS_INIT} http_parser.host
-                                 sss.host ccf_endpoints.host ccf_kv.host quickjs.host
+      node_frontend_test
+      PRIVATE ${CMAKE_THREAD_LIBS_INIT} http_parser.host sss.host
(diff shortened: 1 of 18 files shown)


CMakeLists.txt
```diff
@@ -432,12 +432,13 @@ if(BUILD_TESTS)
     )
 
     add_unit_test(
-      node_frontend_test
+      node_frontend_test ${CMAKE_CURRENT_SOURCE_DIR}/src/js/wrap.cpp
```
wintersteiger, line 435: Probably fine, but not obvious to me why this change is needed.

wintersteiger (approved):
Thanks for looking into this, this will catch problems much earlier now!

CMakeLists.txt
eddyashton: It's because `node_frontend_test.cpp` now includes `frontend_test_infra.h` -> `member_frontend.h` -> `wrap.h`. We could build this separately if we refactor our libraries, or break this dependency chain (I don't know if the common test stuff actually _needs_ the member frontend's JS support?) Including this to match similar other unit tests.

include/ccf/crypto/pem.h
```diff
@@ -16,14 +16,25 @@ namespace crypto
...
   {
+  private:
     std::string s;
 
+    void check_pem_format()
+    {
+      if (!s.starts_with("-----"))
+      {
+        throw std::runtime_error(
+          fmt::format("PEM constructed with non-PEM data: {}", s));
```
achamayou, line 27: That doesn't seem to be a very thorough check that the data is in fact PEM-encoded?

eddyashton: It's not, but it's good enough to catch all of our current failure cases (empty strings, DERs). I looked at OpenSSL's `PEM_read_bio`, which would do more thorough validation, but would (afaict) still pass things that might fail when we actually try to use them (ie - as certs, or keys) later.

achamayou: Let's gate that on NDEBUG then?

wintersteiger: I think we want this in release/NDEBUG builds too as it catches simple errors earlier. Without that, users may get cryptic OpenSSL messages that they can't make sense of (e.g. "ASN.1 content too short", without information about where that ASN.1 came from and why it's too short).

achamayou: But that's at the expense of running this string comparison on every single Pem object construction, which seems really arbitrary and expensive. If we want to validate input, like when parsing JSON, we should do it all the way, not like this.

PR #3935, opened 2022-06-14
Describe forwarding behaviour in generated OpenAPI

### LINKED ISSUE #3882: Generated OpenAPI should describe whether endpoints are forwarded or not
We have a mix of forward-always, forward-sometimes, and never-forward endpoints, and it's not clear to users what consistency guarantees they should expect. We should include this as an optional extension to the OpenAPI spec. We should also revisit what behaviour we choose for each endpoint.

eddyashton:
Resolves #3882.


doc/schemas/app_openapi.json
```diff
@@ -238,12 +238,26 @@
         "scheme": "signature",
         "type": "http"
       }
+    },
+    "x-ccf-forwarding": {
+      "always": {
+        "description": "If this request is made to a backup, it will be forwarded to a primary node for execution. Should be used for operations which may produce writes.",
```
jumaffre, line 244:
```suggestion
        "description": "If this request is made to a backup node, it will be forwarded to the primary node for execution. Should be used for operations which may produce writes.",
```

doc/schemas/app_openapi.json
```diff
@@ -238,12 +238,26 @@
...
+        "value": "always"
+      },
+      "never": {
+        "description": "This call will never be forwarded, and is always executed on the node it was sent to. If this attempts to write on a backup, this will fail. This will be executed by the receiving node even if earlier requests on the same session were forwarded, potentially breaking session consistency. This should be used for operations which want to read node-local state rather than the latest replicated state, for instance for all historical queries.",
+        "value": "never"
+      },
+      "sometimes": {
+        "description": "If this request is made to a backup, it may be forwarded to a primary node for execution. Specifically, if this request is sent as part of a session which was already forwarded, then it will also be forwarded. This is the default value, and should be used for most read-only operations.",
```
jumaffre, line 252:
```suggestion
        "description": "If this request is made to a backup node, it may be forwarded to the primary node for execution. Specifically, if this request is sent as part of a session which was already forwarded, then it will also be forwarded. This is the default value, and should be used for most read-only operations.",
```

doc/schemas/app_openapi.json
jumaffre: "Should be used for operations which may produce writes." reads like something I'd expect to read in the documentation for building apps, but not in the OpenAPI spec for a given endpoint, once the app has been built?

eddyashton: Good point. I've moved these sections to a comment on the enum (which should be in the docs), and removed it from the OpenAPI.

jumaffre (approved):

heidihoward:
It might be useful to define in the OpenAPI what happens regarding forwarding when the primary changes. Are requests (in the same session) now sent to the new primary or are they dropped? If it is the former, then it might be possible for a client to observe a transaction move from the committed state back to pending.

achamayou:
@heidihoward this is tracked as #3952

PR #3941, opened 2022-06-15
Restrict maximum size of client HTTP request

### LINKED ISSUE #3912: Support setting maximum HTTP message size in config
Currently, sending an HTTP message that exceeds enclave memory will cause an out of memory crash before reaching the endpoint handler function.

A new config option should be added that restricts the maximum HTTP message size (possibly this is just for the payload, and the header size is a compile-time constant). I don't think this has to be per-endpoint but rather a global option.

If the maximum size is exceeded then the node should just close the connection. Responding early with a 413 Payload Too Large is generally poorly supported by clients, since they try to finish sending first, before reading the response. I would reserve 413 for soft limits enforced by endpoint handlers and treat the node config option as a hard limit that simply drops the connection.

jumaffre:
Resolves #3912 

This PR adds configuration parameters (set by the operator on node startup) to cap the maximum size of client HTTP request a) body, 2) header size and 3) headers count. When this occurs we send back a [413](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) (a) or [431](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/431) and automatically close the client session.


include/ccf/http_configuration.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <optional>
+
+namespace http
+{
+  struct ParserConfiguration
+  {
+    ds::SizeString max_body_size = {"1MB"};
+    ds::SizeString max_header_size = {"16KB"};
+    size_t max_headers_count = 256;
```
jumaffre, line 16: The default values for these entries are:
- `max_body_size`: `1MB` is what we've tested in the CI for a very long time now.
- `max_header_size`: NodeJS default value.
- `max_headers_count`: roughly guided by the two other values, and would leave 256 headers with a size of 4KB each (2KB for each key and value) to end up with a total headers size the size of the total body size.

achamayou: These values and the fact they can be set should be touched upon in the operator doc.

jumaffre: Documentation for these is automatically generated [here](https://microsoft.github.io/CCF/main/operations/configuration.html). Do you mean something more specific?

PR #3945, opened 2022-06-16
ACME challenge frontend

wintersteiger:
This removes the host-side ACME challenge server and adds an in-enclave one based on the existing frontend infrastructure.


src/node/ccf_acme_client.h
```diff
@@ -31,21 +39,133 @@ namespace ccf
     }
   }
 
+  class ACMERpcEndpoints : public CommonEndpointRegistry
```
jumaffre, line 42: Could you move the server implementation to another file to keep this separate from the client logic?

src/node/ccf_acme_client.h
```diff
@@ -31,21 +39,133 @@ namespace ccf
...
+  public:
+    ACMERpcEndpoints(
+      NetworkState& network, ccfapp::AbstractNodeContext& context) :
+      CommonEndpointRegistry(get_actor_prefix(ActorsType::well_known), context)
+    {
+      auto handler = [this](auto& ctx) {
+        http_status response_status = HTTP_STATUS_INTERNAL_SERVER_ERROR;
+        std::string response_body;
+
+        try
```
jumaffre, line 53: This `try ... catch` logic should already be generically handled by the HTTP server frontend.

src/node/ccf_acme_client.h
```diff
@@ -31,21 +39,133 @@ namespace ccf
...
+          response_status = HTTP_STATUS_INTERNAL_SERVER_ERROR;
+          response_body = ex.what();
+        }
+
+        ctx.rpc_ctx->set_response_status(response_status);
+        ctx.rpc_ctx->set_response_body(std::move(response_body));
+      };
+
+      make_endpoint(
+        "/acme-challenge/{token}", HTTP_GET, handler, no_auth_required)
```
jumaffre, line 100: Is the endpoint supposed to return a JSON object? If so, the `json_adapter` could be used here.

src/enclave/rpc_sessions.h
```diff
@@ -354,7 +357,12 @@ namespace ccf
         }
         else
         {
-          auto ctx = std::make_unique<tls::Server>(certs[listen_interface_id]);
+          std::unique_ptr<tls::Context> ctx;
+          if (
+            per_listen_interface.endorsement.authority == Authority::UNSECURED)
+            ctx = std::make_unique<nontls::PlaintextServer>();
```
jumaffre, line 363: Should we restrict which endpoint URI clients can trigger via this interface? Otherwise, they could be a couple of typos away (http:// instead of https:// and forgetting to type a port number, defaulting to `80`) from submitting their private payload to a plaintext endpoint?

achamayou: Yes, as discussed, we should. There is also an integrity issue with requests that are JWT authenticated, and virtually all responses other than receipts.

src/node/ccf_acme_client.h
wintersteiger: No, it's supposed to return just a string (`response_body = token + "." + tit->second;`)

tests/acme_endorsement.py
```diff
@@ -345,12 +351,24 @@ def run_lets_encrypt(args):
...
+            node.rpc_interfaces["acme_challenge_server_if"] = challenge_server_interface
+
     wait_for_certificates(args, service_dns_name, ca_certs)
 
 
 if __name__ == "__main__":
     args = infra.e2e_args.cli_args()
     args.package = "samples/apps/logging/liblogging"
-    args.nodes = infra.e2e_args.max_nodes(args, f=1)
+    args.nodes = infra.e2e_args.min_nodes(args, f=1)
```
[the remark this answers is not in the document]
wintersteiger: What's the relationship between those two line numbers? I just changed this to min so we don't start more nodes than necessary.

src/enclave/rpc_sessions.h
wintersteiger: Proposal coming up.

wintersteiger: See this commit (which is part of this PR): https://github.com/microsoft/CCF/pull/3945/commits/1c3cb5a208d390078c5a14126183426a327dbc2c

That's the simplest way I could think of, except I'm sure there is a way to get the local copy of the RPC interfaces instead of reading it transactionally. I'm not too familiar with the frontend infrastructure, so this may well contain errors/bugs/unintended consequences.

doc/host_config_schema/cchost_config.json
```diff
@@ -76,9 +76,9 @@
                 "properties": {
                   "authority": {
                     "type": "string",
-                    "enum": ["Node", "Service", "ACME"],
+                    "enum": ["Node", "Service", "ACME", "Unsecured"],
```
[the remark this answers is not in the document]
wintersteiger: Yeah, I agree, that would make sense. The `authority`/`endorsement` fields never felt quite right to me anyways.

src/node/rpc/frontend.h
```diff
@@ -178,6 +178,38 @@ namespace ccf
...
+          }
+          if (ifit->second.endorsement->authority == Authority::UNSECURED)
+          {
+            ctx->set_response_status(HTTP_STATUS_SERVICE_UNAVAILABLE);
+            return ctx->serialise_response();
+          }
+        }
+        else
+        {
+          // internal or forwarded, OK?
```
wintersteiger, line 209: I'm not sure whether we should accept RPCs that are forwarded to an unsecured interface.

src/enclave/rpc_sessions.h
```diff
@@ -354,7 +357,12 @@ namespace ccf
...
+          else
```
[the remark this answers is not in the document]
wintersteiger: Done!

achamayou (approved):

doc/host_config_schema/cchost_config.json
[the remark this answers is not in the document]
wintersteiger: @achamayou what do you think, is that something we can change at this point in time?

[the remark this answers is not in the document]
achamayou: I don't think so. http/https is insufficient to capture the level of detail we want here, so I don't know what benefit we'd get from this.

wintersteiger:
Done; this now uses the node config subsystem and I expect everything, including the perf tests, to pass.

jumaffre (approved):

PR #3953, opened 2022-06-17
Remove use of `SIG_IGN` for `SIGPIPE`

eddyashton:
~~Exploring a potential fix for the segfaults that have been introduced since OE 0.18.0. I believe they're caused by a combination of OE now trying to forward signals to the enclave, then calling the fallback/original handler, and not correctly reading the `SIG_IGN` we've set here.~~

~~This probably isn't the actual fix, but want to pump it through the CI for some validation.~~

Temporary workaround for a bug in OE: https://github.com/openenclave/openenclave/issues/4542

Rather than using `SIG_IGN`, we register an actual empty signal handler for `SIGPIPE`.


wintersteiger (commented):
LGTM. I agree with @jumaffre, the simpler `signal(sig, func)` would look better.

achamayou (approved):

jumaffre (approved):

Issue #3602, opened 2022-02-25
Document the meaning of writeset and commit evidence


letmaik:
Those should be documented since they are exposed in receipts.

achamayou:
- [ ] Add Write Set entry to the glossary pointing to https://microsoft.github.io/CCF/main/architecture/ledger.html?highlight=writeset#transaction-format
- [ ] Add Commit Evidence to the glossary and include a short summary of its function (details under https://github.com/microsoft/CCF/issues/3295)

Also:

- [ ] Receipt architecture page, motivating the high-level goals, linking to Audit and Use Apps pages
- [ ] Use Apps: cover receipts as service signatures with optional user claims
- [ ] Add detail about cert, identity of signer
- [ ] Make explicit the requirement to have the service identity
- [ ] Add a commented sample use of custom claims in the JS logging app, showing how to expose the log message directly in the receipt
- [ ] Audit: cover receipts as audit elements, potentially to be combined with a ledger, move to first position
- [ ] Add a commented sample use of a receipt with a ledger or online service to detect and blame a malicious catastrophic recovery
- [ ] Add a commented sample use of a receipt + ledger to check for inclusion

achamayou:
Started under https://github.com/achamayou/CCF/tree/receipt_doc_changes, will tick off as PR'ed.

PR #3954, opened 2022-06-17
Receipt documentation

achamayou:
WIP towards #3602

### DIFF
--- a/doc/operations/certificates.rst
+++ b/doc/operations/certificates.rst
@@ -98,8 +98,14 @@ To be able to bind to that port, the ``cchost`` binary may need to be given spec
 
         "network": {
             "rpc_interfaces": { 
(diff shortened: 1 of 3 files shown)


doc/operations/certificates.rst
```diff
@@ -98,8 +98,14 @@ To be able to bind to that port, the ``cchost`` binary may need to be given spec
```
eddyashton, line 98: Can't comment on the _actual_ line (1 above), but should we change this from `json` to `python`, so we can include comments, snips, multi-line strings?

eddyashton (approved):

doc/operations/certificates.rst
achamayou: Done

Issue #3950, opened 2022-06-17
A newly elected leader must not advance the commit index until a transaction in the new term commits


achamayou:
It is not safe for a newly elected leader to advance the commit index in terms prior to the one it has been elected in. They must instead wait until at least one transaction has been emitted on the new term, and observe its commit.

This is to make sure that any node that could win a subsequent election has the same suffix, and cannot potentially cause a rollback of previously committed transactions.

In practice, this should happen quickly, because a newly elected leader starts their term by sending a signature. The current logic does not however correctly wait for this, and will treat past signatures as committable.

Another way to think about this is that a newly elected node should only ever commit its own signatures in the current term, never those emitted in a previous term.

eddyashton:
This is described in section 5.4.2 and figure 8 of [the Raft paper](https://raft.github.io/raft.pdf).

achamayou:
Closing in favour #3971.

Issue #3948, opened 2022-06-17
Leader election must wait for quorum in each active config, rather than across all active nodes


achamayou:
Follow up to https://github.com/microsoft/CCF/issues/3837

The current logic in CCF incorrectly uses a quorum across all active nodes: https://github.com/microsoft/CCF/blob/main/src/consensus/aft/raft.h#L1980

This can lead to multiple primaries being elected in certain circumstances, and although only one of them would be able to make progress, the other could cause nodes with valid logs to truncate them, affecting persistence.

heidihoward:
Just to elaborate further.  

**Problem** 

Today CCF handles [reconfiguration](https://microsoft.github.io/CCF/main/architecture/consensus/1tx-reconfig.html) as follows: 

The current leader adds a reconfiguration txn to the ledger. This txn (and subsequent ones) are sent to all servers in both the old and new configurations. Once the next signature txn has been replicated by the current leader to a majority in both the old and new configurations, then the reconfiguration txn is considered committed. The leader switches to the new configuration.  Once the followers learn that the reconfiguration txn is committed, they also switch to the new configuration. Nodes that are not included in the new configuration will enter the “retired” state. 

This is the happy path where the leader who proposes the reconfiguration txn also commits it. The tricky bit is what happens if there are leader elections during this time? There are two options: 
* If the leader did not add a signature txn after the reconfiguration txn then the reconfiguration txn will be discarded during an election. 
* If a follower becomes a candidate with a signed but uncommitted (aka committable) reconfiguration txn in its ledger, then it currently needs a **majority quorum across both the old and new configurations** to become a leader. 

The 2nd option is a problem as the new leader might not have all previously committed txns in its ledger as it might not have votes from a majority quorum in the old configuration (see example discussed in https://github.com/microsoft/CCF/issues/3837). There can also be multiple leaders with the same term (for the same reason). 

This is not a problem if only one server is added/removed and there is only one reconfiguration in-flight at a time. 

**Fix** 

The proposed fix is to switch from the current logic where a candidate requires votes from a **majority across the active configurations** to requiring votes from a **majority in each active configuration**. I believe that the same change also needs to be made to the CheckQuorum logic.

Issue #3837, opened 2022-05-13
Invariant ``MoreUpToDateCorrectInv`` violated in TLA+ specification


heidihoward:
**Describe the bug**
TLC found a violation of the ``MoreUpToDateCorrectInv`` invariant when checking the current TLA+ specification in https://github.com/microsoft/CCF/blob/main/tla/raft_spec/ccfraft.tla.

Note that this is different TLA+ specification (and thus a different issue) from #3828.

**To Reproduce**
Use TLC to check the current spec in https://github.com/microsoft/CCF/blob/main/tla/raft_spec/MCraft.tla using a term limit of 3. As you can see from the output below, TLC took >2 days to find this bug (after checking 37 billon states) on a well-resourced machine so I'll need to be patient.

**Expected behavior**
TLC should have completed checking without finding a safety violation.

**Environment information**
Host: [Azure standard_HB120rs_v3](https://docs.microsoft.com/en-us/azure/virtual-machines/hbv3-series)
OS: Ubuntu 20.04 LTS
TLC: TLC2 Version 2.17 of 02 February 2022

**Additional context**
The final output & counter example provided by TLC is copied below:
```
State 27: <ClientRequest line 351, col 5 to line 362, col 80 of module ccfraft>
/\ ReconfigurationCount = 1
/\ committedLogDecrease = FALSE
/\ messages = { [ mtype |-> RequestVoteRequest,
    mterm |-> 2,
    mlastLogTerm |-> 0,
    mlastLogIndex |-> 0,
    msource |-> NodeTwo,
    mdest |-> NodeThree ] }
/\ clientRequests = 3
/\ matchIndex = ( NodeOne :>
      ( NodeOne :> 0 @@
        NodeTwo :> 0 @@
        NodeThree :> 0 @@
        NodeFour :> 0 @@
        NodeFive :> 0 ) @@
  NodeTwo :>
      ( NodeOne :> 2 @@
        NodeTwo :> 0 @@
        NodeThree :> 0 @@
        NodeFour :> 0 @@
        NodeFive :> 0 ) @@
  NodeThree :>
      ( NodeOne :> 0 @@
        NodeTwo :> 0 @@
        NodeThree :> 0 @@
        NodeFour :> 0 @@
        NodeFive :> 0 ) @@
  NodeFour :>
      ( NodeOne :> 0 @@
        NodeTwo :> 0 @@
        NodeThree :> 0 @@
        NodeFour :> 0 @@
[...]

PR #3965, opened 2022-06-21
Fixing TLA+ specifications

heidihoward:
~~Currently, the function of this PR is to improve visibility of the ongoing work on the [TLA+ specifications](https://github.com/microsoft/CCF/tree/main/tla). It is not yet ready to be reviewed/merged.~~

This PR updates the current TLA+ specifications and will bring them closer to the current codebase of [CCF's consensus protocol](https://github.com/microsoft/CCF/tree/main/src/consensus). 

This includes the following:

**Deduplicating the two specifications**
- [x] Merge the two specifications into one. Currently, there is [one with reconfiguration](https://github.com/microsoft/CCF/tree/main/tla/raft_spec) and [one without](https://github.com/microsoft/CCF/tree/main/tla/reduced_raft_spec), however they have diverged in other ways and are difficult to maintain/test. We can instead use different TLC configurations to check the same specification both with and without reconfiguration. Merging the specifications will involve starting with the [reconfiguration specification](https://github.com/microsoft/CCF/tree/main/tla/raft_spec) and copying changes over from the other specification before deletion.

**Checking the existing TLC+ specification in TLC**
- [x] Document and patch any bugs
  - [x] The specs were using ``LastLogTerm``in leader election logic & ``MoreUpToDateCorrectInv`` invariant instead of ``LastCommittableTerm``.
  - [x] ``CommittedLogDecrease`` logic did not allow the same log entry to be committed twice by different leaders. This is in fact safe as long as the log entries are the same.
  - [x] A leader should not update its ``commitIndex`` until it has replicated a signature in its current term - see https://github.com/microsoft/CCF/issues/3950. This is not an issue in the TLA+ atm as there are currently no heartbeat ``AppendEntries`` but it should be patched anyway
  - [x] TLA+ specs and CCF codebase was using quorums across configurations instead of in each configuration for leader election - see https://github.com/microsoft/CCF/issues/3837 and https://github.com/microsoft/CCF/issues/3948

**Reduce the state space and make it quicker to model check**
- [x] Investigate enabling symmetry by default
- [x] Replace arbitrary reconfigurations with a list of reconfigurations to check - also useful to stop a node being removed and added back again

**Add more invariants to verify correctness of the specification**
- [x] Add ``TypeOK`` invariant to check type safety
- [x] Add debugging invariants which are designed to produce counter-examples to check that useful states are being reached

**Make the TLA+ specification more similar to the current codebase**
- [x] Update NACK handling to reflect the CCF logic for determine which log entries the leader should sent next
- [x] Implement CheckQuourm - add an action allowing a leader to choose to step down in the current term - see https://github.com/microsoft/CCF/pull/3865

**General tidying up**
- [x] Minor reformatting, fixing typos, adding more documentation
- [x] Consolidate docs into a single page [here](https://microsoft.github.io/CCF/main/architecture/raft_tla.html) and update docs to reflect the fact that there is now only one core specification with multiple configurations and to reflect the latest version of TLC used to check the specs
- [x] Remove dummy values from signature transactions


heidihoward:
https://github.com/microsoft/CCF/discussions/4264 documents future changes we might want to make to the TLA+ specs but are beyond scope for this PR

achamayou (approved):

PR #3969, opened 2022-06-21
Experiment with election fix

achamayou:
First look at #3950.

This is changing the signature and post-election logic to:

1. never commit anything prior to the election index, included.
2. emit a signature as soon as the node becomes leader.

The end to end tests seem to pass, but there are failures in the raft scenario and unit tests that need investigation.


achamayou:
Closing in favour of #3971

PR #3971, opened 2022-06-22
New Raft leaders must only commit entries from the current term

eddyashton:
See #3950. Took the same approach as #3969 (`committable_indices.clear()`) but want to make sure all related fixes end in whichever version we settle on.


src/consensus/aft/raft.h
```diff
@@ -1796,6 +1789,13 @@ namespace aft
       // and has a genesis or recovery tx as the last transaction
       election_index = last_committable_index();
 
+      // A newly elected leader must not advance the commit index until a
+      // transaction in the new term commits. We achieve this by clearing our
+      // list committable indices - so nothing from a previous term is now
+      // considered committable. Instead this new primary will shortly produce
+      // their own signature, which _will_ be considered committable.
+      committable_indices.clear();
```
eddyashton, line 1797: This is the fix, to prevent previous entries being considered committable when this node becomes primary.

src/consensus/aft/raft.h
```diff
@@ -1344,7 +1334,10 @@ namespace aft
               }
               else
               {
-                state->view_history.update(prev_lci + 1, ds->get_term());
+                // NB: This is only safe as long as AppendEntries only contain a
+                // single term. If they cover multiple terms, then we need to
+                // know our previous signature locally.
+                state->view_history.update(r.prev_idx + 1, ds->get_term());
```
eddyashton, line 1340: `Multi-term divergence` committable suffix test found the following buggy situation with the old code:

```
B: 1.1, 1.2, 1.3, 1.4, 7.5, 7.6, 7.7, 7.8, 8.9, 9.10
A: 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 3.7, 5.8
B->A: AE (1.4, 7.6], term=7, prev_term=1
```

Node A has cleared their `committable_indices`, so no longer considers the shared entry 1.4 as committable. So when deserialising this signature, the LCI is incorrect, and can't be used to work out where this signature's term starts from.

As long as we don't have multi-term AEs, we can use `r.prev_idx` instead - the LCI (and a persistent history of signatures) is only necessary when the AE might end in a different term.

src/consensus/aft/raft.h
```diff
@@ -380,20 +380,11 @@ namespace aft
...
-      }
-      else
-      {
-        return std::nullopt;
-      }
+      kv::Consensus::SignableTxIndices r;
+      r.version = state->last_idx;
+      r.term = get_term_internal(r.version);
+      r.previous_version = std::max(last_committable_index(), election_index);
+      return r;
```
eddyashton, line 387: Probably also should remove the optional wrapper, now this is always filled. This was used to prevent a signature being emitted in the short period after an election where the new primary doesn't know commit, to avoid signatures that say commit moved backwards. That's no longer possible - the new primary _must_ produce a new signature, and that's the first thing they _could_ commit. `previous_version` is used to truncate the merkle tree, so needs to be reasonably fresh - it's possible this is still insufficient.

tests/raft_scenarios/fancy_election.1
```diff
@@ -61,7 +61,11 @@ dispatch_one,1
 # Node 2 responds in support
 dispatch_one,2
 
+# Node 1 appends a new committable entry to confirm its primaryship
+replicate,2,ImTheCaptainNow
+
 # Node 1 sends an initial AppendEntries probe
+periodic_one,1,10
```
eddyashton, line 68: The change here is semantically the same as in `scenarios_gen.py`. Scenarios that previously reached state sync now need an additional write in the new term and an additional round-trip to confirm commit of this.

src/consensus/aft/raft.h
achamayou: That sounds about right, but I am tempted to add a check to read_ledger to confirm the trees we produce always cover the right interval. `previous_version` is only used to select the range for the serialisation, the compaction is done on commit.

It's worth noting that the first signature of every term will now report that the commit has reached the end of the previous term, which is not necessarily true. We know we need to remove those fields in any case.

achamayou: Will add this check in a separate PR, as discussed.

src/kv/kv_types.h
```diff
@@ -467,6 +467,14 @@ namespace kv
     virtual bool is_backup() = 0;
     virtual bool can_replicate() = 0;
 
+    enum SignatureDisposition
```
achamayou, line 470:
```suggestion
    enum class SignatureDisposition
```

Come to think of it, it'd be nice to have a linter thing that makes sure we only have enum classes.

src/consensus/aft/raft.h
```diff
@@ -758,6 +776,7 @@ namespace aft
             become_retired(index, kv::RetirementPhase::Signed);
           }
           committable_indices.push_back(index);
+          is_new_primary = false;
```
achamayou, line 779: That seems a little brittle, I *think* we can't get a batch with two signatures inside at the moment, but it would be cleaner to reset on `get_signature_disposition()` to make sure we only force the one signature?

eddyashton: So I think `get_signature_disposition()` is too early, as it risks resetting but then not producing a signature because of a later early-out. But I agree this is a little late, and potentially means we produce many signatures. I'll add an explicit reset of this once we think we've actually emitted a signature.

eddyashton: Added this inside `get_signable_txids`, to avoid taking the lock twice, with comments around the reset and calling point to make the expectations clear. Considered `get_signable_txids_and_reset_should_sign`, but that makes it a little _too_ obvious this is a weird thing to do. Staring again, this bool could exist entirely in `history.h`, avoiding locking concerns and localising its impact, but I think I prefer the explicit naming of this in-consensus enum approach.

src/consensus/aft/raft.h
```diff
@@ -104,6 +104,13 @@ namespace aft
     // entries, the initial index will not advance until this node acks.
     bool is_new_follower = false;
 
+    // When this node becomes primary, they should produce a new signature in
+    // the current view. This signature is the first thing they may commit, as
+    // they cannot confirm commit of anything from a previous view (Raft paper
+    // section 5.4.2). This bool is true from the point this node becomes
+    // primary, until it is explicitly by a call to get_signable_txid()
```
achamayou, line 111:
```suggestion
    // primary, until it is explicitly reset by a call to get_signable_txid()
```

src/consensus/aft/raft.h
achamayou: That seems like a good compromise, I agree.

achamayou (approved):

tests/infra/consortium.py
```diff
@@ -711,10 +711,10 @@ def check_for_service(self, remote_node, status, recovery_count=None):
             current_status = r["service_status"]
             current_cert = r["service_certificate"]
             # Note: to change once this is backported to 2.x
-            if remote_node.version_after("ccf-2.0.3"):
-                current_recovery_count = r["recovery_count"]
-            else:
-                assert "recovery_count" not in r
+            # if remote_node.version_after("ccf-2.0.3"):
```
jumaffre, line 714: See https://github.com/microsoft/CCF/pull/3994 for fix.

src/kv/store.h
```diff
@@ -638,7 +638,6 @@ namespace kv
 
         version = tx_id.version;
         last_replicated = tx_id.version;
-        last_committable = 0;
```
jumaffre, line 641: The comment on lines `50-51` needs updating.

tests/lts_compatibility.py
```diff
@@ -120,6 +120,7 @@ def test_new_service(
         network.retire_node(primary, node)
         if primary == node:
             primary, _ = network.wait_for_new_primary(primary)
+            network.txs.issue(network, number_txs=1, repeat=True)
```
jumaffre, line 123: I think this deserves a comment.

tests/infra/consortium.py
achamayou: Yes, waiting on that to merge.

src/kv/store.h
achamayou: Good catch, thank you!

tests/lts_compatibility.py
achamayou: Agreed, done.

src/consensus/aft/raft.h
lemmy: This change not only has the desired effect that a new primary cannot advance its commit index until a transaction in the new term commits, but also that the leader may truncate its log if an update term message is received before the leader produces its own signature (see https://github.com/microsoft/CCF/issues/5674#issuecomment-1769868427).

Issue #2784, opened 2021-07-09
Add second snapshot directory option to `cchost`


jumaffre:
Since only the primary node generates snapshots but all nodes will want to make use of snapshots for historical queries (once #1705 is done) and to determine the longest ledger in a recovery scenario, operators should be able to give each node a shared volume for committed snapshot files. 

I can see two options for this:
1. snapshot-dir/read-only-snapshot-dir for consistency with the existing ledger-dir/read-only-ledger-dir scheme for the ledger files, leaving it up to the operator to copy committed files from one to the other.
2. Have `cchost` automatically copy committed snapshot files from one directory to the other (i.e. snapshot-dir/committed-snapshot-dir naming scheme). 

I prefer option 1. for consistency and because operators will have already built a similar monitor/copy process for committed ledger files. It also avoid having the single-threaded `cchost` copy potentially large snapshot files (unless we make use to [libuv's thread pool](http://docs.libuv.org/en/v1.x/fs.html)). I believe this is also @msftsettiy's preferred option.

msftsettiy:
I agree with option 1.

PR #3973, opened 2022-06-23
Add support for second read-only snapshots directory

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/2784

This PR adds support for a second read-only snasphot directory (in the same manner as for the ledger read-only directories) that could/should be shared between nodes to share committed snapshot files. 

TODO: 
- [x] Changelog
- [x] Documentation


tests/reconfiguration.py
```diff
@@ -212,11 +212,7 @@ def test_add_node_from_snapshot(network, args, copy_ledger=True, from_backup=Fal
     # is at least one historical entry to verify.
     network.txs.issue(network, number_txs=1)
     idx, historical_entry = network.txs.get_last_tx(priv=True)
-    for _ in range(1, args.snapshot_tx_interval):
-        network.txs.issue(network, number_txs=1, repeat=True)
-        last_tx = network.txs.get_last_tx(priv=True)
-        if network.wait_for_snapshot_committed_for(seqno=last_tx[1]["seqno"]):
```
jumaffre, line 218: Removed the `wait_for_snapshot_committed_for` function which was duplicated in `get_committed_snapshots`.

achamayou (approved):

PR #3975, opened 2022-06-23
Recovery benchmark: Check that all entries have been recovered

jumaffre:
This is adding a quick sanity check to make sure that all entries have been recovered when using the `recovery_benchark.sh` script.


PallabPaul (approved):
Thanks for this Julien!

achamayou (approved):

PR #3981, opened 2022-06-24
Add Troubleshooting page to documentation

darracott:
The "Node Output" page has been relabelled as "Troubleshooting" in the documentation and CLI commands for troubleshooting have been added to it. 
Also added a note to the "Updating documentation" documentation to help developers build documentation pages locally.


doc/contribute/build_ccf.rst
```diff
@@ -102,4 +102,6 @@ or if there are no Doxygen changes
 
 .. code-block:: bash
 
-    $ SKIP_DOXYGEN=ON ./livehtml.sh
\ No newline at end of file
+    $ SKIP_DOXYGEN=ON ./livehtml.sh
+
+.. note:: If you haven't already built CCF then to build local documentation you may need to temporarily rename ``python/version.py.in`` to ``python/version.py`` and edit it to contain a valid version.
```
jumaffre, line 107: Good spot! 

The `version.py.in` file is an input file for `cmake` which automatically converts it to `version.py` [here](https://github.com/microsoft/CCF/blob/main/CMakeLists.txt#L49-L51) so that the version of the CCF Python package is automatically deduced from the git commit (which is done [here](https://github.com/microsoft/CCF/blob/main/cmake/version.cmake) if you're curious), so that we don't have to manually update the version every time we want to release a new version of CCF (which happens [often enough](https://github.com/microsoft/CCF/releases)!)

We've always (wrongly) assumed that people build the docs after having built CCF (see https://microsoft.github.io/CCF/main/contribute/build_ccf.html) and thus have a `version.py` script automatically generated under `python/`. In fact, this is what our CI does [here](https://github.com/microsoft/CCF/blob/main/.azure-pipelines-gh-pages.yml#L26-L29). 

In this case, because we've written `livehtml.sh` too, I think it's preferable to have the script generate `version.py` via `cmake` if this one doesn't already exist. Something like:

```bash
mkdir -p build_docs
cd build_docs
cmake .. # Generates version.py
cd ..
rm -rf build_docs
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Generated OpenAPI now describes whether each endpoint is forwarded (#3935).
+- The "Node Output" page has been relabelled as "Troubleshooting" in the documentation and CLI commands for troubleshooting have been added to it.
```
jumaffre, line 17: We don't normally write documentation changes to the changelog because it's already very verbose. However, we should find a way to flag these somehow. If you have any suggestions, they are welcome!

doc/operations/troubleshooting.rst
```diff
@@ -0,0 +1,81 @@
...
+
+.. note:: CCF may be deployed with a loadbalancer which may cache the node which last responded to a query from an IP address. Until the cache clears, any subsequent queries from that IP address will be handled by the same node. As an example, if the cache clears after one minute, then in order to get a response from a different node, an operator must wait one minute between queries.  
+
+Below are descriptions of CLI commands and how they are useful for diagnosing CCF issues:
+
+**“What node is handling my requests?”**
+
+.. code-block:: bash 
+
+    curl -k https://example-ccf-domain.com/node/network/nodes/self -i
```
jumaffre, line 18: We usually discourage users/members/operators to use `curl -k`, which means that the service identity is not verified, although we use it in practice so that we don't have to find the service certificate.

As you can see [here](https://microsoft.github.io/CCF/main/use_apps/issue_commands.html#issuing-commands), we recommend users to specify the `--cacert` option, which points to the service certificate.

jumaffre (approved):
That's an awesome first contribution!

doc/operations/troubleshooting.rst
```diff
@@ -0,0 +1,81 @@
+Troubleshooting CCF
+===================
+
+This page contains troubleshooting tips for CCF.
+
+Tips for interacting with CCF to diagnose issues
+------------------------------------------------
+.. note:: In the examples below this documentation uses ``example-ccf-domain.com`` as an example CCF domain, you will need to replace that with your own CCF domain when using these commands. 
+
+.. note:: CCF may be deployed with a loadbalancer which may cache the node which last responded to a query from an IP address. Until the cache clears, any subsequent queries from that IP address will be handled by the same node. As an example, if the cache clears after one minute, then in order to get a response from a different node, an operator must wait one minute between queries.
```
achamayou, line 10: This is correct, but:

1. it's independent from CCF, and completely dependent on whether and how a load balancer is used
3. "will be handled by the same node" is superficially correct, but can easily be interpreted too strongly: nodes can and will forward queries to another nodes depending on endpoint forwarding policies and session state. So while a limited amount of handling will always happen on that same node, the actual execution may not.

doc/operations/troubleshooting.rst
achamayou: +1 for avoiding -k as much as possible in the documentation, particularly outside a strict operation context with local IP addresses.

doc/operations/troubleshooting.rst
```diff
@@ -0,0 +1,81 @@
...
+**“What nodes are part of the current network?”**
+
+.. code-block:: bash
+
+    curl -k https://example-ccf-domain.com/node/network/nodes
+
+This will show information for all nodes in the network. In a healthy network all nodes will show ``“status” = “Trusted”``, and one node only will show ``“primary” = true``. This is the healthy state of the network. 
+Around upgrades/restarts/migrations nodes will transition through unhealthy states temporarily. If the network remains in an unhealthly state for a long time, this indicates there is an issue. 
+
+You can obtain this information for a single node by appending its ID to this endpoint. E.g. ``https://<your-domain>/node/network/nodes/<ID>``, where ``<ID>`` can be obtained from the ``/node/network/nodes/self`` endpoint described above. Take note of the ``node_data`` field which contains useful correlation IDs. E.g. ``node 0``, ``pod name``.
```
achamayou, line 39: When referring to endpoints, using the notation :http:GET:`/node/version` will create a link to the corresponding OpenAPI schema entry.

doc/contribute/build_ccf.rst
darracott: Thanks, good idea! I've now updated `livehtml.sh` to generate `version.py` if it doesn't already exist, and I've removed the note I had added to `doc/contribute/build_ccf.rst` about it.

CHANGELOG.md
darracott: How does adding a ​`doc/DOC-CHANGELOG.md`​ where docs changes can be tracked sound? 

I could be convinced that we have versioned docs already and so the old versions of docs act as a history. However, I do think there's not much overhead to maintaining a docs changelog and it will be useful to users who are already familiar with a version of CCF. It will towards those users towards updated or new areas of documentation that they should pay attention to when interaction with a different version of CCF.

darracott: I've taken the `### Documentation` suggestion

doc/operations/troubleshooting.rst
darracott: Thanks I've removed the `-k` and informed readers, in the note at the top of the page, they will need to add authentication parameters to the curl commands, as in the example [here](https://microsoft.github.io/CCF/main/use_apps/issue_commands.html#issuing-commands)

doc/operations/troubleshooting.rst
darracott: 1. Agreed and hopefully that is clear in the wording.
2. I've clarified the wording to avoid misinterpretation. 
On 1) I think it is useful to have this note as users without a loadbalancer can easily ignore it and those with a loadbalancer may benefit. I can be convinced to drop this note altogether, but otherwise I'll leave it as it is now.

doc/operations/troubleshooting.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Below are descriptions of CLI commands and how they are useful for diagnosing CCF issues:
+
+**“What node is handling my requests?”**
+
+.. code-block:: bash 
+
+    curl -k https://example-ccf-domain.com/node/network/nodes/self -i
+
+This is useful to identify which node is handling queries. The node ID can be found in the Location header. Please read the note above about loadbalancers.
```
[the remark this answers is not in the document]
darracott: Thanks, I've added example output!

doc/operations/troubleshooting.rst
darracott: Oh cool! I've now made use of this throughout

doc/operations/troubleshooting.rst
```diff
@@ -0,0 +1,81 @@
...
+
+    curl -k https://example-ccf-domain.com/node/consensus
+
+This has a few bits of data that might help us diagnose a partitioned/faulty network. In particular, most of the time there should be a single entry in the ``configs`` list. During an upgrade/restart/migration, there may be multiple values. If multiple values persist for a long time, it suggests something went wrong during the reconfiguration.
+
+**“Is the CCF network stable?”**
+
+.. code-block:: bash
+
+    curl -k https://example-ccf-domain.com/node/commit
```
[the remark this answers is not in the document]
darracott: Thanks, done

doc/operations/troubleshooting.rst
```diff
@@ -0,0 +1,89 @@
+Troubleshooting CCF
+===================
+
+This page contains troubleshooting tips for CCF.
+
+Tips for interacting with CCF to diagnose issues
+------------------------------------------------
+.. note:: In the examples below this documentation uses ``example-ccf-domain.com`` as an example CCF domain, you will need to replace that with your own CCF domain when using these commands. You will also need to add authentication parameters such as ``--cacert`` to the curl commands, see :doc:`Issuing commands </use_apps/issue_commands>` for an example.
+
+.. note:: CCF may be deployed with a loadbalancer which may cache the node which last responded to a query from an IP address. Until the cache clears, the loadbalancer will direct any subsequent queries from that IP address to the same node. As an example, if the cache clears after one minute, then in order to get a response from a different node, an operator must wait one minute between queries.
```
achamayou, line 10:
```suggestion
.. note:: CCF may be deployed with a load balancer which may cache the node which last responded to a query from an IP address. Until the cache clears, the load balancer will direct any subsequent queries from that IP address to the same node. As an example, if the cache clears after one minute, then in order to get a response from a different node, an operator must wait one minute between queries.  
```

achamayou (approved):
Thank you for adding this page!

achamayou:
@darracott please resolve conflicts, and the PR will auto-merge after a round of CI.

PR #3986, opened 2022-06-27
Limit size of recovered ledger entries batch

jumaffre:
The batching of recovered ledger entries introduced in #3746 causes a bug when the ledger entries to recover are too large. For example, if ledger entries are `2MB` each, the recovered batch of `100` entries will be `200MB` and will violate the current limit of the size of ring buffer messages ([`max_msg_size`](https://microsoft.github.io/CCF/main/operations/configuration.html#memory)).

To address this, the host now returns however many entries (crudely) can fit within a single ring buffer message and sends this to the enclave for deserialising. 

TODO: 
- [x] Unit test
- [x] Changelog

### DIFF
--- a/src/host/ledger.h
+++ b/src/host/ledger.h
@@ -345,25 +345,51 @@ namespace asynchost
       return get_last_idx();
     }
 
-    size_t entries_size(size_t from, size_t to) const
+    std::pair<bool, size_t> entries_size(
+      size_t from,
+      size_t to,
+      std::optional<size_t> max_size = std::nullopt) const
     {
       if ((from < start_idx) || (to < from) || (to > get_last_idx()))
       {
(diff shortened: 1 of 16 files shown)


src/host/ledger.h
```diff
@@ -345,25 +345,52 @@ namespace asynchost
...
 
-      if (to == get_last_idx())
-      {
-        return total_len - positions.at(from - start_idx);
-      }
-      else
+      bool has_cut = false;
+      size_t size = 0;
+
+      // If max_size is set, return entries that fit within in (best effort).
```
achamayou, line 361:
```suggestion
      // If max_size is set, return entries that fit within it (best effort).
```

achamayou (approved):

Issue #1713, opened 2020-10-06
Expose to the operator clearly when retired nodes can be shut down


achamayou:
A node being removed may observe the commit of its removal before its successor network does. In some circumstances, when the amount of nodes being removed is high enough to prevent the newly elected primary from establishing consensus over the committable suffix it inherits (as implemented in #1641), terminating those removed nodes too early could effectively cause consensus to become permanently unable to progress, requiring a catastrophic recovery.

If instead, nodes observing their retirement being committed redirect user requests to the new network from that point onwards, then an operator can simply wait for the transaction status of the removal transaction to become COMMITTED, from any node in either configuration, at which point the removed node(s) can be safely terminated.

This can be implemented by adding to the removal hook code that extracts the new configuration, and sending redirects to one of those nodes in response to all requests once the node is in state RETIRED.

achamayou:
Following discussion with @heidihoward and @eddyashton, it is clear that this approach will not work.

Instead, it is necessary for a primary having executed a reconfiguration that removes one or more nodes to:

1. wait for the reconfiguration to commit
2. wait for the commit index to make progress again, so as to make sure that any node who would win an election from there on would have observed the reconfiguration being committed, and would no longer rely on the old configuration to win their election

The consensus logic should keep track of this, and show an additional attribute on retired nodes indicating whether they can safely be removed.

While typing this, I realise this directly conflicts with the current primary step down logic described under https://microsoft.github.io/CCF/main/architecture/consensus/1tx-reconfig.html#retirement-details, which disables transaction emission after the first transaction following a reconfiguration.

heidihoward:
Just to elaborate further,

It is necessary for these two steps to happen in series.  The primary must wait for the reconfiguration to be committed and thus for its local commit index to be greater than the reconfiguration transaction ID. Then the primary must commit a subsequent transaction such that any future leaders in the new configuration will have a commit index greater than the reconfiguration transaction ID. The primary then knows that it is ok to shut down removed nodes and can communicate this knowledge to the operator.

Aside: If I recall correctly (I might not be), there was some discussion of whether an operator could look at the ledger and determine whether it was safe to shut down a node by checking for two committed signatures in the same term after the reconfiguration. On reflection, I do not think this approach would work as its not clear if the signatures where committed by the primary of the term in question or by a subsequent primary. I think there needs to be some explicit information from the primary to say that its is ok to shut down the old configuration.

achamayou:
[...]

PR #4008, opened 2022-07-01
Safe node removal

achamayou:
Fix for #1713

- [x] Add a `retired_committed` node flag in the KV. Nodes are transitioned to this flag being set by the primary observing that their `RETIRED` state for them has been committed.
- [x] Add a `GET /node/network/removable_nodes` endpoint to query `RETIRED && retired_committed` nodes - this allows the operator to find out what nodes can be safely shut down
- [x] Add an endpoint to flush `RETIRED && retired_committed` nodes from the KV to avoid KV growth

We may choose to go with a "fresh read" for step 2 instead, removing the need for step 3 and allowing us to elide the `retired_committed` step and go straight to removal. This involves turning `GET /node/network/removable_nodes` into a write, for which operators would need to not forget to check `tx/status`. This seems unergonomic right now.

Once we have blocking transactions though, we probably want to revisit this.

- [x] Update operator documentation to explain safe removal flow.


achamayou:
Some notes for Monday:

Interestingly, the LTS fails with:

```
Exception in virtual void aft::Aft<consensus::LedgerEnclave>::recv_message(const ccf::NodeId &, const uint8_t *, size_t) [LedgerProxy = consensus::LedgerEnclave]
```

Occuring in a 2.0.4 node, which suggests that a new node is sending messages the old nodes are unable to cope with. I was initially suspicious that this may be because the nodes are removed from the table too early, and signature verification fails, but removing the node removal call reveals that is not the case!

include/ccf/ds/openapi.h
```diff
@@ -421,6 +421,11 @@ namespace ds
       const nlohmann::json& param)
     {
       auto& params = parameters(path(document, uri));
+      for (auto& p : params)
+      {
+        if (p["name"] == param["name"])
```
achamayou, line 426: Without this, path parameters get added once per HTTP verb, causing duplicates (eg. `GET` and `DELETE` on the same url). This is "first definition wins", which I think is ok because path parameters are always otherwise identical/fully determined by their name.

include/ccf/service/node_info.h
```diff
@@ -69,6 +69,13 @@ namespace ccf
      * node identity in `public_key` field. Service-endorsed certificate is
      * recorded in "public:ccf.nodes.endorsed_certificates" table */
     std::optional<crypto::Pem> cert = std::nullopt;
+
+    /** Commit state for Retired state
+     *
+     * Introduced during 2.x (2.0.5), and so optional for backward
+     * compatibility.
+     */
+    bool retired_committed = false;
```
jumaffre, line 78: I thought we once said (probably not documented!) that all added fields going forward should be `std::optional`?

src/node/rpc/node_frontend.h
```diff
@@ -958,6 +964,75 @@ namespace ccf
...
+          node->retired_committed)
+        {
+          nodes->remove(node_id);
+        }
+
+        return make_success(true);
+      };
+
+      make_endpoint(
+        "/network/nodes/{node_id}",
```
jumaffre, line 1029: This looks like once can remove any node, but this is really about retired nodes. Maybe `GET /network/nodes/retired/{node_id}` instead?

src/node/rpc/node_frontend.h
```diff
@@ -958,6 +964,75 @@ namespace ccf
           "status", ccf::endpoints::OptionalParameter)
         .install();
 
+      auto get_removable_nodes = [this](auto& args, nlohmann::json&&) {
+        GetNodes::Out out;
+
+        auto nodes = args.tx.ro(this->network.nodes);
+        nodes->foreach(
+          [this, &out, nodes](const NodeId& node_id, const NodeInfo& ni) {
+            auto node = nodes->get_globally_committed(node_id);
```
jumaffre, line 973: Perhaps add a comment that this endpoint should be modified with caution as it checks for globally committed state and may break serialisation?

jumaffre:
This needs a changelog entry and a new section in the operation docs [here](https://microsoft.github.io/CCF/main/operations/code_upgrade.html) and [here](https://microsoft.github.io/CCF/main/governance/common_member_operations.html#removing-an-existing-node).

tests/infra/consortium.py
```diff
@@ -324,14 +325,43 @@ def get_proposal(self, remote_node, proposal_id):
...
+            r = c.get(f"/node/network/nodes/{node_to_retire.node_id}")
+            pending = r.body.json()["status"] == infra.node.State.PENDING.value
         LOG.info(f"Retiring node {node_to_retire.local_node_id}")
         proposal_body, careful_vote = self.make_proposal(
             "remove_node",
             node_id=node_to_retire.node_id,
         )
         proposal = self.get_any_active_member().propose(remote_node, proposal_body)
         self.vote_using_majority(remote_node, proposal, careful_vote)
+        if remote_node == node_to_retire:
```
jumaffre, line 340: Could the following be done under `infra.network.retire_node()` instead to separate concern between the execution of the proposal (members) and the cleanup of retired nodes (operators)?

tests/infra/consortium.py
```diff
@@ -324,14 +325,43 @@ def get_proposal(self, remote_node, proposal_id):
...
+                        r = c.get("/node/network/removable_nodes").body.json()
+                        if node_to_retire.node_id in {n["node_id"] for n in r["nodes"]}:
+                            check_commit = infra.checker.Checker(c)
+                            r = c.delete(
+                                f"/node/network/nodes/{node_to_retire.node_id}"
+                            )
+                            check_commit(r)
+                            break
+                        else:
+                            r = c.get(
```
jumaffre, line 357: `r` is unused here?

include/ccf/service/node_info.h
jumaffre, line 78: We should also add an assertion in `ledger.py` to make sure no signatures are recorded for a node where this is set.

achamayou: It's not strictly speaking necessary if we provide a default value, which we do here, and if that default value is logically correct in the cases where the serialised form does not have the field present (which is the case here). Check with @eddyashton. But in general, yes.

src/node/rpc/node_frontend.h
achamayou: It seemed more resty to `DELETE` the same path that you `GET`? One of the attributes is indeed "Retired" for this to work, but that's just a piece of validation?

I don't feel super strongly about this, and I hope to remove this eventually, so I could be swayed either way. @eddyashton @MahatiC?

src/node/rpc/node_frontend.h
achamayou: I mean, sure, but then I feel this is true of many node frontend endpoints, chief among which `join`.

This isn't so much breaking serialization as it is adding a read dependency on something in addition to the KV state (the replication status). The transaction will still be coherent with the total order if it happens, but may not take place depending on extra-non-KV-state (same as trying to use any endpoint before open!).

achamayou:
> This needs a changelog entry and a new section in the operation docs [here](https://microsoft.github.io/CCF/main/operations/code_upgrade.html) and [here](https://microsoft.github.io/CCF/main/governance/common_member_operations.html#removing-an-existing-node).

Yes, I've updated those.

tests/infra/consortium.py
achamayou: Yes, probably, I will take a look.

include/ccf/service/node_info.h
achamayou: We should really enforce a single signature after `RETIRED`, and this is weaker than that, but it doesn't hurt to check in addition.

tests/infra/consortium.py
achamayou: Mostly yes, but if we end up raising at :364, it's nice to know what the last state was (ie. what we're stuck on), and that's why this line is there at all.

include/ccf/service/node_info.h
jumaffre: I vaguely remember a discussion following https://github.com/microsoft/CCF/discussions/2916 where we said we wanted `std::optional` for new fields to make sure that it was clear for programmers to know that the value may not exist in the ledger. I'm not too fussed either way, but I think we should strive for consistency and we've gone the `std::optional` way so far.

src/node/rpc/node_frontend.h
eddyashton: I think this still warrants a comment. It's a near-unique use of `get_globally_committed` (the only other being the frontend open check, which is at least far enough from any other KV operations in code-distance that it avoids a lot of potential confusion. We've convinced ourselves this non-transactional dependency is sound (as you outlined), but we should include that in a comment in the code for posterity + discoverability. I don't think `join` or the other node frontend endpoints do anything like this with non-transactional dependencies?

include/ccf/service/node_info.h
jumaffre: Yes, the assertion is currently too loose [there](https://github.com/microsoft/CCF/blob/0943d8c4a23985bc1fafd50032594c6f5c1405a1/python/ccf/ledger.py#L521-L528).

src/node/rpc/node_frontend.h
```diff
@@ -958,6 +964,75 @@ namespace ccf
...
+        }
+
+        auto nodes = args.tx.rw(this->network.nodes);
+
+        auto node = nodes->get_globally_committed(node_id);
+        if (
+          node.has_value() && node->status == ccf::NodeStatus::RETIRED &&
+          node->retired_committed)
+        {
+          nodes->remove(node_id);
```
jumaffre, line 1022: We need to remove the endorsed certificate too.

include/ccf/service/node_info.h
achamayou: From that discussion:

> The new field will either have to be defined as i) std::optional or ii) a non-std::optional type, with a sensible default value.

This falls under ii) as far I can tell? I would argue false is a sensible default here? I'm fine changing this, I just don't see much of a benefit (consistency is dubious, most of our fields aren't optional, users already have to deal with the fact there's a mix).

src/node/rpc/node_frontend.h
achamayou: `join` does redirect things no other endpoint does, many of the built-ins end up being fairly special, and I think we're on average better off with explicit naming and clear code than large comments that don't necessarily stay up to date. But I also believe in code reviews, so I'll add a comment :)

src/node/rpc/node_frontend.h
achamayou: Yes, thank you!

include/ccf/service/node_info.h
achamayou: Ok, so after discussion, I will:

1. Keep this as it is for now
2. Look separately at changing serialisation to always serialise non-std::nullopt values (#4074)

src/node/rpc/node_frontend.h
achamayou: Will stay with the current path naming, as discussed with @eddyashton and @jumaffre.

src/node/rpc/node_frontend.h
achamayou: @jumaffre @eddyashton I've re-read all three uses carefully, and concluded one of them was unnecessary. I've commented the other two.

tests/infra/consortium.py
achamayou: It largely works, the only thing is the need to return/pass whether the node was pending before we retired it, but that seems fine.

jumaffre (approved):

eddyashton (approved):

PR #4011, opened 2022-07-04
Docs: Fix TLA+ link

jumaffre:
The previous link is now dead so the doc is now pointing to the TLA+ website index instead.


jumaffre:
The `lts_compatibility` test seems unstable for unrelated reasons. Merging regardless.

Issue #3985, opened 2022-06-27
Prevent nodes added during a rolled back reconfiguration from disrupting the service


heidihoward:
It is my understanding that newly added nodes begin ticking (see: https://github.com/microsoft/CCF/blob/src/consensus/aft/raft.h#L1305) and thus can become a candidate when their reconfiguration txn becomes committable (followed by a signature txn). If the reconfiguration is then rolled back, then the newly added nodes will timeout, repeatedly trying to become a leader, and thus disrupt the active configuration.

One possible solution is to require newly added nodes (nodes who have not yet seen the reconfiguration txn which added them be committed) to check the status of the reconfiguration txn before starting an election. If the reconfiguration txn is invalid, then they can safety shut down.

This is low priority as it is hypothetical and could only impact liveness if a reconfiguration is unsuccessful. This could also be fixed by PreVote ([#2577](https://github.com/microsoft/CCF/issues/2577)).

jumaffre:
I believe this is effectively the same issue as described in #3250.

PR #4018, opened 2022-07-08
Raft: count election quorums in all active configurations

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/3948

To find out whether a node can be elected as primary, nodes now count quorums in all active configurations, rather than in the union of all nodes across configurations. For example, in the `C_0 = [0, 1, 2]` network, if node `3` is added (`C_1 = [2, 3, 4]`) and an election occurs while the reconfiguration is in flight, the new primary node will require votes from at least 2 nodes in `C_0` (e.g. `1` and `2`) and 2 nodes in `C_1` (e.g. `3` and `4`). This ultimately prevents a node from becoming primary if it will not be able to commit new transactions.

TODO:
- [x] Fix partition test
- [x] 2tx scheme
- [x] Unit test
- [x] Changelog


tests/infra/consortium.py
```diff
@@ -333,24 +333,35 @@ def retire_node(self, remote_node, node_to_retire, timeout=10):
...
+        wait_for_commit=True,
+        timeout=3,
     ):
-        proposal_body, careful_vote = self.make_proposal(
-            "transition_node_to_trusted",
-            node_id=node_id,
-            valid_from=valid_from,
-            validity_period_days=validity_period_days,
-        )
+        proposal_body = {"actions": []}
```
jumaffre, line 345: This proposal can trust multiple node at once, whereas the infra could only trust one single node per proposal previously.

tests/infra/network.py
```diff
@@ -1098,22 +1099,12 @@ def wait_for_new_primary(
         error = TimeoutError
         logs = []
 
-        backup = self.find_any_backup(old_primary)
```
jumaffre, line 1101: BFT-specific code was removed here.

tests/infra/node.py
```diff
@@ -446,25 +449,29 @@ def is_stopped(self):
...
-                rep = nc.get("/node/commit")
-                assert (
-                    rep.status_code == 200
-                ), f"An error occurred after node {self.local_node_id} joined the network: {rep.body}"
-                self.network_state = infra.node.NodeNetworkState.joined
-        except infra.clients.CCFConnectionException as e:
-            raise TimeoutError(
-                f"Node {self.local_node_id} failed to join the network"
-            ) from e
+        start_time = time.time()
```
jumaffre, line 457: This logic is different now that we check that a node has joined via the node-endorsed RPC interface.

src/consensus/aft/raft.h
```diff
@@ -453,7 +469,7 @@ namespace aft
       const std::unordered_set<ccf::NodeId>& new_learner_nodes = {},
       const std::unordered_set<ccf::NodeId>& new_retired_nodes = {}) override
     {
-      LOG_DEBUG_FMT("Configurations: add {{{}}}", conf);
+      LOG_DEBUG_FMT("Configurations: add new conf at {}: {{{}}}", idx, conf);
```
achamayou, line 472:
```suggestion
      LOG_DEBUG_FMT("Configurations: add new configuration at {}: {{{}}}", idx, conf);
```

src/consensus/aft/raft.h
```diff
@@ -121,7 +128,11 @@ namespace aft
 
     // Configurations
     std::list<Configuration> configurations;
-    std::unordered_map<ccf::NodeId, NodeState> nodes;
+    // Union of other nodes (i.e. all nodes but us) in all active
+    // configurations. This should be used for diagnostic or broadcasting
+    // messages but _not_ for counting quorums, which should be done for all
+    // active configurations.
```
achamayou, line 134:
```suggestion
    // messages but _not_ for counting quorums, which should be done for each
    // active configuration.
```

src/consensus/aft/raft.h
```diff
@@ -1999,30 +2041,41 @@ namespace aft
 
     void add_vote_for_me(const ccf::NodeId& from)
     {
-      size_t quorum = -1;
-
-      if (reconfiguration_type == ReconfigurationType::TWO_TRANSACTION)
+      // Add vote for from node in all configuration where it is present
```
achamayou, line 2044:
```suggestion
      // Add vote for from node in each configuration where it is present
```

src/consensus/aft/raft.h
```diff
@@ -2236,14 +2295,17 @@ namespace aft
...
+              "resharing "
               "not completed yet.");
             break;
           }
         }
 
         if (reconfiguration_type == ReconfigurationType::ONE_TRANSACTION)
         {
+          LOG_DEBUG_FMT(
+            "Configurations: discard committed conf at {}", conf->idx);
```
achamayou, line 2308:
```suggestion
            "Configurations: discard committed configuration at {}", conf->idx);
```

src/consensus/aft/raft.h
```diff
@@ -2438,6 +2500,8 @@ namespace aft
 
       while (!configurations.empty() && (configurations.back().idx > idx))
       {
+        LOG_DEBUG_FMT(
+          "Configurations: rollback conf at {}", configurations.back().idx);
```
achamayou, line 2504:
```suggestion
          "Configurations: rollback configuration at {}", configurations.back().idx);
```

tests/infra/consortium.py
```diff
@@ -271,7 +271,7 @@ def get_member_by_service_id(self, service_id):
         )
 
     def vote_using_majority(
-        self, remote_node, proposal, ballot, wait_for_global_commit=True, timeout=3
+        self, remote_node, proposal, ballot, wait_for_commit=True, timeout=3
```
achamayou, line 274: There are dozens of references to "global_commit" in test code, if we're going to clean it up (which is fine), I would rather have it done in a separate PR, rather than have bits done here and there.

tests/infra/network.py
achamayou: Same as above, that would be better in its own PR, imo.

achamayou (approved):

tests/partitions_test.py
```diff
@@ -275,6 +276,92 @@ def test_new_joiner_helps_liveness(network, args):
...
+def test_election_reconfiguration(network, args):
+    # Test for issue described in https://github.com/microsoft/CCF/issues/3948
+    # Note: this test makes use of node-endorsed secondary RPC interface since
+    # new nodes never observe commit of their configuration and thus never
+    # open their service-endorsed primary RPC interface.
+    primary, backups = network.find_nodes()
+
+    LOG.info("Join new nodes without trusting them just yet")
+    new_nodes = []
+    # Start N+1 new nodes to make sure they cannot elect of them as a primary
```
eddyashton, line 290:
```suggestion
    # Start N+1 new nodes to make sure they cannot elect any of them as a primary
```

This comment is missing at least a word, but I think could be reworded further for clarity (I don't understand it).

tests/partitions_test.py
```diff
@@ -275,6 +276,92 @@ def test_new_joiner_helps_liveness(network, args):
...
+
+        LOG.info(
+            "Make sure that new nodes cannot elect a primary node among themselves"
+        )
+        try:
+            network.wait_for_new_primary(
+                primary,
+                nodes=new_nodes,
+                interface_name=infra.interfaces.SECONDARY_RPC_INTERFACE,
+                timeout_multiplier=1,
```
eddyashton, line 338:
```suggestion
                timeout_multiplier=3,
```

Just a guess. A lot of our elections require multiple rounds of election time outs before they pass in good, expected situations. If we want to assert that no primary _can_ be elected, I think we need to wait for multiple elections.

tests/partitions_test.py
```diff
@@ -275,6 +276,92 @@ def test_new_joiner_helps_liveness(network, args):
...
+        except infra.network.PrimaryNotFound:
+            LOG.info(
+                "As expected, new primary could not be elected as old configuration could not make progress"
+            )
+        else:
+            assert False, "No new primary should be elected while partition is up"
+
+        LOG.info("Stop all new nodes")
+        for node in new_nodes:
+            node.stop()
```
eddyashton, line 349: Why do we do this so early? In 'real operations', we can't guarantee that they will be killed so early, so what happens if they're still alive when the partition is dropped?

tests/raft_scenarios/election_while_reconfiguration
```diff
@@ -0,0 +1,74 @@
...
+dispatch_one,0 # 0 sends vote response to 2
+assert_is_candidate,2
+
+dispatch_one,1 # 1 sends vote response to 2
+
+# Eventually, 2 has received votes from a majority of nodes in _both_ configurations
+state_all
+assert_is_primary,2
+
+
```
eddyashton, line 74: Minor: I think we should have an `assert_state_sync` at the end of all the scenarios (preceded by however many connections + roundtrips are required to reach agreement), to confirm that we're able to reach a fully agreed stable state.

tests/partitions_test.py
jumaffre: Good point. I wanted the test to run quickly as we're expecting no new primary to be elected but we should be consistent and wait for a reasonable amount of time here too.

src/consensus/aft/raft.h
```diff
@@ -889,32 +906,58 @@ namespace aft
...
           }
         }
 
-        if (backup_ack_timeout_count >= get_quorum(nodes.size()))
+        if (!has_quorum_of_backups)
         {
           // CheckQuorum: The primary automatically steps down if it has not
-          // heard back from a majority of backups during an election timeout.
+          // heard back from a majority of backups in _all_ active
+          // configurations during an election timeout.
```
eddyashton, line 956: This comment describes the opposite of the comment above, unless my close-reading skills are fried:

```
// If primary has quorum of active backups in any configuration, it
// should remain primary
```
```
// CheckQuorum: The primary automatically steps down if it has not
// heard back from a majority of backups in _all_ active
// configurations during an election timeout.
```

Are we 'supported' (ie - unwilling to step down as primary) by a quorum in a single configuration, or requiring a quorum in all configurations?

src/consensus/aft/raft.h
```diff
@@ -1999,30 +2042,41 @@ namespace aft
...
     {
-      size_t quorum = -1;
-
-      if (reconfiguration_type == ReconfigurationType::TWO_TRANSACTION)
+      // Add vote for from node in each configuration where it is present
+      for (auto const& conf : configurations)
       {
-        const auto& cfg = configurations.front();
+        auto const& nodes = conf.nodes;
+        votes_for_me[conf.idx].quorum = get_quorum(nodes.size());
```
eddyashton, line 2049: Why are we calculating this every time?

I think this means we're constructing `votes_for_me` on-the-fly as each vote arrives? I would have expected us to populate this when we become candidate/trigger an election - at _that_ index, we work out and store the quorum in each configuration, and create buckets for the votes to go in. I haven't thought about what it means for configurations to change mid-election.

Since we iterate through configurations here, `votes_for_me` is always big-enough by line 2067, but only just in time.

src/consensus/aft/raft.h
```diff
@@ -2036,7 +2090,7 @@ namespace aft
       auto new_commit_cft_idx = std::numeric_limits<Index>::max();
 
       // Obtain CFT watermarks
-      for (auto& c : configurations)
+      for (auto const& c : configurations)
```
eddyashton, line 2093: I was going to suggest `auto const&` for consistency with the other code, but it turns out we do it this way round in loads of places!

`auto const&` -> 39 results in 16 files
`const auto&` -> 448 results in 112 files

So ignore for now, but we should standardisze that in the enforced coding standard if we can.

src/consensus/aft/raft.h
```diff
@@ -2236,14 +2296,17 @@ namespace aft
             !resharing_tracker->have_resharing_result_for(rr.value(), idx))
           {
             LOG_TRACE_FMT(
-              "Configurations: not switching to next configuration, resharing "
+              "Configurations: not switching to next configuration, "
+              "resharing "
```
eddyashton, line 2300:
```suggestion
              "Configurations: not switching to next configuration, resharing "
```

Looks like a temporarily indented reflow that we can revert.

tests/partitions_test.py
jumaffre: That's because we don't want the new nodes to interfere (i.e. bump the term) with the remaining original two backup nodes for the remainder of the test. This will eventually be fixed by PreVote and has been discussed in https://github.com/microsoft/CCF/issues/3985.

tests/raft_scenarios/election_while_reconfiguration
```diff
@@ -0,0 +1,85 @@
...
+dispatch_one,0 # 0 sends vote response to 2
+assert_is_candidate,2
+
+dispatch_one,1 # 1 sends vote response to 2
+
+# Eventually, 2 has received votes from a majority of nodes in _both_ configurations
+state_all
+assert_is_primary,2
+
+# Node 1 appends a new committable entry to confirm its primaryship
```
eddyashton, line 74:
```suggestion
# Node 2 appends a new committable entry to confirm its primaryship
```

Or maybe just "New primary appends a new...". The argument here is the term, but in this instance the new primary is 2, not 1.

src/consensus/aft/raft.h
achamayou: I'd rather we switched to "each" everywhere possible, rather than "all", which is somewhat more vague.

jumaffre: I've renamed this for clarity.

src/consensus/aft/raft.h
jumaffre: Fixed. We now clear and set `votes_for_me` when becoming candidate or becoming aware of a new term.

src/consensus/aft/raft.h
```diff
@@ -1833,18 +1833,28 @@ namespace aft
       timeout_elapsed = std::chrono::milliseconds(distrib(rand));
     }
 
+    void reset_votes_for_me()
+    {
+      for (auto const& conf : configurations)
```
eddyashton, line 1838:
```suggestion
      votes_for_me.clear();
      for (auto const& conf : configurations)
```

Else we're also looking for quorum in extremely old/dead configurations?

jumaffre: Thank you!

eddyashton (approved):

jumaffre:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|release/2.x|An unhandled error occurred. Please see the logs for details|

### Manual backport
```
backport --pr 4018
```

### Questions ?

jumaffre:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|release/2.x|An unhandled error occurred. Please see the logs for details|

### Manual backport
```
backport --pr 4018
```

### Questions ?

Issue #4019, opened 2022-07-08
Bundle script for TS based applications fails on Windows OS.


caleteeter:
**Describe the bug**
When bundling the Typescript sample application (/tests/npm-app) on a Windows system, the resulting bundled application fails when proposing to the network.  The issue is caused by the file path referenced in the name of the modules section of the bundle.  The `\\` should in fact be `/`.

example:
<img width="584" alt="image" src="https://user-images.githubusercontent.com/8452367/178066789-9221e71b-988c-4643-b98e-88007850aa12.png">

**To Reproduce**
On a Windows system, clone the repo, run `npm run build` and include the bundled output from dist/ in a proposal for the js app. 

**Expected behavior**
Deployment of the bundle should complete if built on Windows or Linux based system.

**Environment information**
CCF version 2.0.0

**Additional context**

jumaffre:
Fixed in #4020.

PR #4020, opened 2022-07-08
Fixed bundling bug for Windows based systems.

caleteeter:
Fixed issue #4019


achamayou:
I think I would prefer an explicit call to a "convert to Unix path" function, implemented like:

```
const path = require("path");

// ...

const definitelyPosix = somePathString.split(path.sep).join(path.posix.sep);
```

(from https://stackoverflow.com/questions/53799385/how-can-i-convert-a-windows-path-to-posix-path-using-node-path)

achamayou (approved):

Issue #4013, opened 2022-07-05
Separate/tag application logging from framework logging


achamayou:
Some users have expressed interest in being able to differentiate application-specific logging from framework-level logging.

This is easier to do in js_generic than for C++ applications, but something to investigate.

achamayou:
We may want to look at logging libraries such as https://github.com/odygrd/quill, on platforms that aren't as restricted as SGX.

eddyashton:
Application code from both JS and C++ is now tagged with `[app]`.

PR #4024, opened 2022-07-11
Add support for tagged loggers and app-specific logging macros

eddyashton:
Implements #4013.

_Draft for further discussion._

Primarily a re-factoring of the macros in `logger.h` to support an additional tag field, and allow applications to extend with their own macros + tags.

So all an app developer has to write is:

```
#define EDDYS_LOGGER CCF_LOG_FMT(INFO, "!Eddy!")
...

EDDYS_LOGGER("x is {}", x);
```

To see logs like:
```
2022-07-11T15:55:20.573507Z        100 [info ] ../src/host/snapshots.h:301          | Committing snapshot file "snapshot_31_33.committed" [1363]
2022-07-11T15:55:20.783049Z -0.001 0   [!Eddy!] samples/apps/logging/logging.cpp:423 | x is 42
2022-07-11T15:55:22.628467Z        100 [info ] ../src/host/sig_term.h:24            | SIGTERM: Shutting down enclave gracefully...
```

2 big open questions:
- Where do we put the tag? I like using the existing place where we have the logging level, as I think we should avoid making our lines even longer. ~~But in the current implementation that completely overwrites the level. Is that ok, or should we keep both?~~ _Best of both worlds - we allow the level to be overwritten, but by convention keep the level and an additional tag_
- Do we need to hide the existing/old macros? Even if we do this, I don't think we can guarantee that application code doesn't collide with our internal logging (ie - they could define a matching tag). So I'm tempted to just keep the existing macros visible, and strongly recommend that apps namespace their own loggers if they want to grep for their own output. On the flipside, there's very few uses of these macros in our own public headers, so they're cheap to deprecate, and this might let us give them better names + start tagging logs from our own systems.

And a few smaller ones:
- Do we want a default macro with some generic tag like `app`? ~~I've added this at the moment, but don't know if its valuable. Would much rather app developers see from the start that they can namespace their own logs. But I guess they can't from JS, unless we add custom logging functions.~~ _Yes, for ease of use, and at each level! These are now added, with an `[app]` tag, and used from JS_
- If we put the tags here, should we fix the formatting length so that the `|` column remains aligned? ~~I think this requires multiple `fmt` calls, so is a lot more expensive, but it does look much nicer.~~ _Turns out this isn't too expensive, so I've done the fancy fixed-length formatting_
- Do we want to start tagging logging from our internal systems for easier debugging?
- A few JS improvements listed in the TODOs below.

TODO:
- [x] Clean up code TODOs
- [x] Documentation
- [x] Properly tag all text coming from application JS
- - ~~Display JS line numbers rather an C++ line numbers?~~ _Turns out this isn't possible in QuickJS. The options are to throw an error and manually parse the string stacktrace it contains, or a major hack of the internals to find the debug info. I say no to both_
- - [x] Add `console.info`, `console.debug` etc?
- [x] Testing


jumaffre (approved):
How does it work with JSON-formatted logs? Is the tag added to the `"msg"` field or is it a separate field altogether? We may also want to add a note to this in https://microsoft.github.io/CCF/main/operations/troubleshooting.html#node-output.

eddyashton:
> How does it work with JSON-formatted logs? Is the tag added to the `"msg"` field or is it a separate field altogether?

It's a separate field (`"tag"`):
```
{"h_ts":"2022-07-14T13:17:35.271818Z","thread_id":"100","level":"info","tag":"[info ][app]","file":"../src/ds/test/logger.cpp","number":"136","msg":"BBB"}
```

> We may also want to add a note to this in https://microsoft.github.io/CCF/main/operations/troubleshooting.html#node-output.

I've added a cross-link to `logging.rst` from here.

PR #4027, opened 2022-07-12
Move `COMPILE_LIBCXX` so that it is set for apps

### LINKED ISSUE #4026: `COMPILE_LIBCXX` is not defined for apps
When building "standalone" apps (i.e. relying on CCF being installed under `/opt/ccf`), if the [`add_host_library`](https://github.com/microsoft/CCF/blob/b749361f59e18e20ef8a2c38a190229ba2f2b92b/cmake/ccf_app.cmake#L241) helper function is used, the `COMPILE_LIBCXX` variable won't have been set to `-stdlib=libc++` because this one in set in `common.cmake`. 

The temporary fix is to simply add the following line:
```cmake
add_host_library(my_lib ${my_srcs})
target_include_directories(my_lib PUBLIC ${my_includes})
target_compile_options(my_lib PUBLIC -stdlib=libc++)  # < this line
```

jumaffre:
Resolves #4026

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-It's a new dawn, it's a new canary
+It's a new dawn, it's a new canary.
--- a/cmake/ccf_app.cmake
+++ b/cmake/ccf_app.cmake
@@ -4,10 +4,10 @@
 set(ALLOWED_TARGETS "sgx;virtual")
 
 set(COMPILE_TARGETS
-  "sgx;virtual"
-  CACHE
-  STRING
-  "List of target compilation platforms. Choose from: ${ALLOWED_TARGETS}"
+    "sgx;virtual"
+    CACHE
+      STRING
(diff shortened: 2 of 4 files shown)


achamayou:
I am confused by this, why do we not also expose LINK_LIBCXX? I doesn't seem like apps should need one without the other?

jumaffre:
> I am confused by this, why do we not also expose LINK_LIBCXX? I doesn't seem like apps should need one without the other?

Yes, good point. I've moved it too. This wasn't an issue in practice since linking the final application library also had those flags as it's linked against the CCF static libraries (e.g. `libccf`) that was linked with the `LINK_LIBCXX` flags.

achamayou (approved):

achamayou:
Looks like cmake format decided to reformat everything? That's unfortunate.

PR #4039, opened 2022-07-15
Deprecate old logging macros from application code

eddyashton:
Follow up to #4024, second part of #4013.

This deprecates `LOG_INFO_FMT` (through a trick I'm extremely proud of, with an identically-named-but-non-conflicting enum value), so it will no longer be used from application code (replaced by `CCF_APP_INFO`). However we still use this throughout the framework code. To avoid touching every file to provide a separate definition, we define a magic preproc def in `CMake` so that the framework code _doesn't_ consider these macros deprecated. And then we make sure to disable that def for the sample apps, so the macros is correctly deprecated there. Minimal changes, maximum impact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Deprecated
+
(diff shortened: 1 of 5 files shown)


jumaffre (approved):
Should the logging C++ app make use of these new macros to make sure they work in a standalone app and also showcase their usage?

eddyashton:
> Should the logging C++ app make use of these new macros to make sure they work in a standalone app and also showcase their usage?

Huh, absolutely! I had added some uses in the logging app in the last PR, but must have stripped them out. I'll go add them now.

PR #4042, opened 2022-07-18
Ring buffer hardening

eddyashton:
Fixes 2 issues with the ringbuffer code:
- No bounds check on read means a malicious writer could cause a `Reader` to read out-of-bounds, and clear beyond the end of the ringbuffer. This PR implements `check_access()` and uses it in `Reader` to confirm safe access.
- Fix ringbuffer calculations when offsets overflow. Theoretical exploit (requires writing a total of 2^64 bytes), but now executes as expected as normal when this happens.

Includes unit tests for both.


src/ds/ring_buffer.h
```diff
@@ -348,17 +380,29 @@ namespace ringbuffer
     }
 
   private:
-    void write64(size_t index, uint64_t value)
+    // We use this to detect whether the head is ahead of the tail. In real
+    // operation they should be close to each, relative to the total range of a
+    // uin64_t. To handle wrap-around (ie - when a has overflowed past the max
```
achamayou, line 385:
```suggestion
    // uint64_t. To handle wrap-around (ie - when a has overflowed past the max
```

achamayou (approved):

achamayou (approved):

PR #4047, opened 2022-07-19
Drop tasks in `ThreadMessaging` destructor

eddyashton:
I don't think we should be calling this externally, as it implies an owning/responsible system. Instead we can cleanup when the static instance is destroyed to avoid leaks.

### DIFF
--- a/src/ds/thread_messaging.h
+++ b/src/ds/thread_messaging.h
@@ -225,6 +225,11 @@ namespace threading
       tasks(num_threads)
     {}
 
+    ~ThreadMessaging()
+    {
+      drop_tasks();
+    }
+
     // Drop all pending tasks, this is only ever to be used
     // on shutdown, to avoid leaks, and after all thread but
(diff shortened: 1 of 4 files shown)


src/ds/thread_messaging.h
```diff
@@ -225,6 +225,11 @@ namespace threading
...
     {}
 
+    ~ThreadMessaging()
+    {
+      drop_tasks();
+    }
+
     // Drop all pending tasks, this is only ever to be used
     // on shutdown, to avoid leaks, and after all thread but
     // the main one have been shut down.
```
jumaffre, line 235: Also make `drop_tasks` private?

jumaffre (approved):

src/ds/thread_messaging.h
eddyashton: Ha, good spot! I'd misread that as using the static instance, not a local one.

PR #4049, opened 2022-07-19
Raft scenario for election while reconfiguration is in flight

jumaffre:
Part of #3948 

This PR adds a new raft scenario for when an election occurs during a reconfiguration. To do so, we start a 3-node network and add 3 nodes and retire 1 node at once. A new primary is elected when this one has received 4 votes in total (N = 6, quorum is 4). 

This PR also contains some changes to the raft driver to support node retirement and to set the initial configuration correctly. 

This is the bug described in #3948 and the actual fix will be in #4018


jumaffre:
I will fold this in https://github.com/microsoft/CCF/pull/4018.

PR #4050, opened 2022-07-19
Created CCF PAL

ihaller:
- In the current design, there are a number of places where platform specific (OpenEnclave, virtual) functionality is needed.
- This PR attempts to pull together all of those platform specific functionalities into a platform abstraction layer.
- This work does not cover quote generation as that is a separate design work.
- It also does not cover attestation verification since it makes sense even for the virtual platform to be able to verify attestation reports from an OpenEnclave node.
- It pulls together functionality from oe_shim, oe_init (the part related to attestation initialization), mutex and ccf_memcpy.


include/ccf/endpoint_registry.h
```diff
@@ -3,7 +3,7 @@
 #pragma once
 
 #include "ccf/ds/json_schema.h"
-#include "ccf/ds/mutex.h"
```
eddyashton, line 6: It looks like this new file isn't included in the PR?

include/ccf/endpoint_registry.h
```diff
@@ -127,7 +127,7 @@ namespace ccf::endpoints
       std::map<RESTVerb, std::shared_ptr<PathTemplatedEndpoint>>>
       templated_endpoints;
 
-    ccf::Mutex metrics_lock;
+    ccf::Pal::Mutex metrics_lock;
```
eddyashton, line 130: The ccf coding style is for lower-cased namespaces - can you please rename this to `ccf::pal`?

include/ccf/ds/pal.h
```diff
@@ -0,0 +1,204 @@
...
+#endif
+
+/**
+ * This file implements a platform abstraction layer to enable platforms, such
+ * as OpenEnclave to offer custom implementations for certain functionalities.
+ * By centralizing the platform-specific code to one file, we can avoid exposing
+ * platform-specific types to the rest of the code and have a good overview of
+ * all the functionality that is custom to a given platform. The platform
+ * abstraction layer can also be used in code shared between the host and the
+ * enclave as there is a host implementation for it is well.
```
achamayou, line 29:
```suggestion
 * enclave as there is a host implementation for it as well.
```

include/ccf/ds/pal.h
```diff
@@ -0,0 +1,204 @@
...
+      return true;
+    }
+  };
+
+  using Pal = HostPal;
+
+#else
+  class OEPal
+  {
+    class MutexImpl
```
achamayou, line 87: Can you add a comment to clarify this exists because of https://github.com/openenclave/openenclave/issues/4555, and can be removed once a release of Open Enclave containing the fix is published?

achamayou (approved):

include/ccf/ds/pal.h
ihaller: Thanks, updated!

include/ccf/ds/pal.h
ihaller: Thanks, updated!

ihaller:
To keep header includes follow the current structure, I needed to move `ccf_exception.h` to `include\ds`.
The alternative solution to this is to split the Pal into a public `Pal` (including mutex and memory management) and a private `PalInternal` to handle platform internal functionality such as initialization.
As a further alternative, once the Mutex functionality can go away, the whole `Pal` could go into `src\ds`.

eddyashton (approved):

PR #4051, opened 2022-07-19
Update js_app_ts.rst

DarqueWarrior:
Updated the import statement to use fully qualified name.

It now matches the code in the _tests_ folder.


achamayou (approved):
Thank you @DarqueWarrior!

Issue #3868, opened 2022-05-19
Duplication of logging API in docs


heidihoward:
There seems to be two copies of the API for the example logging application in the docs:
1. https://microsoft.github.io/CCF/main/build_apps/logging_rpc_api.html
2. https://microsoft.github.io/CCF/main/use_apps/rpc_api.html

I think they are the same (except that 2 of the 3 built-in endpoints that are listed are different). 

I'd suggest removing the one in the build apps section and replacing the sample built-in endpoints with a pointer to the list of complete built-in endpoints: https://microsoft.github.io/CCF/main/operations/operator_rpc_api.html

PR #4057, opened 2022-07-21
Cleanly separate example RPC API with built-in endpoints (#3868)

DomAyre:
Tried to have the example API in the build_apps section, and then the built-in /app endpoints in a separate page under use_apps, might be a better way to structure this so open to input.


doc/build_apps/example_rpc_api.rst
```diff
@@ -6,4 +6,5 @@ Example app RPC API
    :include:
       /app/log/.*
       /app/api
```
achamayou, line 8: /app/api is also built-in, effectively

doc/use_apps/rpc_api.rst
```diff
@@ -2,12 +2,7 @@ RPC API
 =======
 
 Application RPCs are exposed under ``/app`` and require authentication with a valid user identity.
-This is a sample of what an application endpoint may look like, as generated by the :ref:`example C++ logging app <build_apps/example_cpp>`.
-
-.. openapi:: ../schemas/app_openapi.json
-   :examples:
-   :include:
-      /app/log/.*
+You can read what application endpoints may look like, as generated by the :ref:`example C++ logging app <build_apps/example_cpp>`, in :doc:`/build_apps/example_rpc_api`.
```
achamayou, line 5: Is app_openapi.json expanded there?

doc/build_apps/example_rpc_api.rst
DomAyre: Ahh okay, in which case I think we should include `/app/api` in `use_apps/rpc_api.rst` and only expand `api/log/` here

doc/use_apps/rpc_api.rst
DomAyre: Yes, my thinking is that the built-in parts are expanded here, and then the parts which are more of an example (aka `/app/log/`) are expanded in `example_rpc_api.rst`

achamayou: Ok!

achamayou (approved):

eddyashton (approved):

PR #4078, opened 2022-07-29
Don't run CI on doc-only changes

achamayou:
Shamelessly inspired from https://github.com/microsoft/verona/blob/master/devops/ci.yml

This is somewhat awkward with our CI files, because we:

1. Do a lot at template expansion time
2. Create a bunch of jobs during that time
3. But the variable docOnly is only set a runtime, after all the expansion has happened and the jobs been created

So we need to add the dependency and condition in a lot of places, and not forget any:

```
dependsOn: ConfigureCIRun
condition: eq(dependencies.configure.outputs['setVarStep.docOnly'], 'false')
```

Another thing that's not as bad but still annoying is that running the step to decide if we need the docOnly or not adds about 30 seconds to the CI.


achamayou:
Ok, so this seems to work now (tested with a small mod to make it ignore yml changes too):

image

.azure-pipelines.yml
```diff
@@ -43,6 +43,25 @@ variables:
...
+  - job: configure
+    displayName: "Configure CI"
+    pool:
+      vmImage: "ubuntu-20.04"
+    steps:
+      - checkout: self
+        clean: true
+      - script: |
+          echo "Determine if any code has changed."
+          if git diff --ignore-submodules=dirty --quiet origin/main -- ':!docs' ':!*.md' ':!*.txt' ':!*.yml'; then
```
achamayou, line 55: This is hardcoding `origin/main`, but we want the target branch most of the time (which could be release/*). I'm looking for a way to do that.

achamayou: $SYSTEM_PULLREQUEST_TARGETBRANCH

eddyashton:
Will we get a green badge on PRs which only make doc changes? Or will they end up with a skipped (red) mark against a required check?

eddyashton (approved):

achamayou:
@eddyashton I believe we will, that's what the .yml change was testing. The CI job still "runs" but only runs one sub-job that checks that it doesn't need to run anything else.

Issue #4074, opened 2022-07-27
Always serialise newly added fields to the ledger when they aren't logically null


achamayou:
Follow-up from #2916 and discussion with @eddyashton and @jumaffre. Optional fields are currently not serialised when set to the type's default value.

We want to explore serialising them in the case when they do match that default value, so long as it's not std::nullopt. The rationale is that this transfers some information from the code to the ledger, which facilitates audit (at least to an extent).

achamayou:
Investigated, and ultimately closed. Details on the draft PR.

PR #4082, opened 2022-07-29
Serialise default values

achamayou:
Looking at #4074 

TODO:

- [ ] Extend unit tests accordingly
- [ ] Publish at least one end-of-test ledger size to cimetrics, probably the reconfiguration test


achamayou:
Serialising all values except nullopts leads us to serialise null for nlohman::json fields, such as member_data. @eddyashton and I discussed adding those as an exception too, but this seems unsatisfactory for the following reason:

When a set_member_data proposal explicitly sets the field to null, as a json value, the KV write would instead be an object without the key. A very surprising result for a write that does not logically involve a transform (ie. json to json).

I am no longer convinced this change is helpful with this many subtleties and exceptions. In particular, given all the default values we have now are also default json values, I do not know that we move the auditability cursor significantly by doing this.

The one case we've identified that could be improved is adding new enums, where the default enum value does not coincide with the default json value. The better way to handle that is to split serialisation from deserialisation and handle them separately. @eddyashton has some preliminary code for this, which can be picked up when we meet that situation.

Issue #4068, opened 2022-07-26
Support for AMD SEV-SNP quotes in CCF


jumaffre:
To provide support for AMD SEV-SNP in ACI, we should add support for retrieval and verification of SEV-SNP attestation reports so that an existing network can verify the measurement of a new joining node.

This includes:
- [x] 1. Add test to make sure virtual node cannot join SGX/SEV-SNP service #4083
- [x] 2. Move quote retrieval/verification to new `ccf::Pal` abstraction layer #4083
- [x] 3. Retrieval of AMD SEV-SNP report on new node
- [x] 4. Retrieval of endorsement certificates ([VECK, ASK, ARK](https://www.amd.com/system/files/TechDocs/57230.pdf)), either from the Azure cache or AMD directly (via new [configuration](https://microsoft.github.io/CCF/main/operations/configuration.html) parameter)
- [x] 5. Parsing and verification of report on primary node (using AMD root of trust and certificate chain)
- [x] 6. Verification of collaterals against AMD root key
- [x] 7. Update existing `code_ids` table to cater for both SGX and SEV-SNP attestations.

Risk: ~It is not yet clear how to retrieve and verify the endorsement certificates for SEV-SNP attestation reports.~ This should be done via the dedicated Azure caching server or AMD server.

jumaffre:
As discussed with @DomAyre, and with regards to point 7. above, it appears necessary to record the type of enclave measurement along with the actual measurement itself. Otherwise, it may be possible (depending on whether we support mixed network - see #4072), for members to allow code `foo` to run (assuming AMD SEV-SNP) but a completely different node from a different platform (e.g. SGX) with the same code `foo` to still be able to join the service. While it should be in practice impossible for a malicious user to craft a specific code measurement, separating the namespaces for code ids of different platforms would prevent a compromised platform (e.g. root attestation key of platform A has been leaked) to the confidentiality guarantees of an existing service on a different, non-compromised, platform B.

This will require changing the schema for the `code_ids` table, recording the type of platform along with the code status in the value. Because of backwards compatibility, a new 3.x node should still be able to read old values (the LTS compatibility test should catch this).

PR #4083, opened 2022-07-29
Move attestation report generation/verification to `Pal`

jumaffre:
Part of https://github.com/microsoft/CCF/issues/4068 (items 1. & 2.).

The attestation report generation/verification functions are moved to `ccf::Pal` so that it's easier to add support for SEV-SNP attestation generation/verification in the future.

In particular, this PR:
- Removes the historical `GET_QUOTE` compile-time flag and mocks more of the attestation code (i.e. support for new `ccf::QuoteFormat` enums for virtual and SEV-SNP). 
- Moves some files to public includes so that they can be accessed from the public `ccf::Pal`. 

TODO: 
- [x] Add test for submitting fake client `POST /node/join` 
- [x] Fix attestation types


DomAyre (approved):

include/ccf/base_endpoint_registry.h
```diff
@@ -3,10 +3,10 @@
 #pragma once
 
 #include "ccf/crypto/pem.h"
+#include "ccf/ds/quote_info.h"
```
achamayou, line 6: The move to ds is a little surprising, given that we were trying to keep ds dependency-free, as I understand it.

include/ccf/ds/attestation_types.h
```diff
@@ -2,14 +2,26 @@
...
 #  include <openenclave/attestation/sgx/evidence.h>
 #  include <openenclave/attestation/verifier.h>
 
+#endif
+
+#include <array>
+
 namespace ccf
 {
+  static constexpr size_t attestation_report_data_size = 32;
```
achamayou, line 17: I don't think we want to assume these are necessarily the same across all platforms, they could vary, and we may want to take advantage of that.

achamayou (approved):

PR #4088, opened 2022-08-02
Support `secp256r1` for Server Temp Key

eddyashton:
Fixes a regression introduced in the migration from mbedtls to OpenSSL. Specifically, when using TLS 1.2, the selection of the curve used for the initial temporary key exchange appears to depend on both `--curve-id` and the list specified here. We need to support P-256 in both to communicate with environments which do not provide P-384 or P-521.

I'm calling this Server Temp Key, as that's the name given to it by `openssl s_client -msg`. Note this has no tests at the moment, but has manually been tested by adding various `-tls1_2 -groups "P-256"` arguments to `openssl s_client`.


jumaffre:
> Fixes a regression introduced in the migration from mbedtls to OpenSSL. Specifically, when using TLS 1.2, the selection of the curve used for the initial temporary key exchange appears to depend on both `--curve-id` and the list specified here. We need to support P-256 in both to communicate with environments which do not provide P-384 or P-521.

Some background reading for further justification:

[TLS 1.2:](https://datatracker.ietf.org/doc/html/rfc8422#section-5.3)
>  If the client has used a Supported Elliptic Curves Extension, the public key in the server's  certificate MUST respect the client's choice of elliptic curves.  A server that cannot satisfy this requirement MUST NOT choose an ECC cipher suite in its ServerHello message.)

whereas [TLS 1.3](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.7) is more flexible:
> This extension was also used to negotiate ECDSA curves. Signature algorithms are now negotiated independently.

jumaffre (approved):

PR #4093, opened 2022-08-03
Populate all error details on authentication failure

### LINKED ISSUE #4044: Misleading authentication errors when multiple auth methods are allowed
**Describe the bug**
If a CCF endpoint is configured with multiple authentication options, a misleading error may be returned if the user fails authentication. Specifically, if the user intended to use Auth Method A on an endpoint that has methods A and B (in that order), when A fails, the user may get an error message indicating that Auth Method B failed, which is misleading because it wasn't intended to be used.

**To Reproduce**
Configure a JS app with the following auth policies on an endpoint.
```
"authn_policies": [
  "user_cert",
  "jwt"
],
```

Attempt to connect to the endpoint with an invalid or unexpected user certificate. The returned error is

```json
{
  "error": {
    "code": "InvalidAuthenticationInfo",
    "message": "Missing authorization header"
  }
}
```

**Expected behavior**
While the reported error is technically true, an error for the intended auth policy would be more illuminating for the user. If it's not possible to detect the intended policy, an error message containing errors from all the configured auth policies would be helpful as well.

**Environment information**
CCF 2.0.4

**Additional context**
Reported by @caleteeter

MahatiC:
Resolves #4044 

When an application is configured with multiple auth options, respond with auth specific error details on failure.


include/ccf/odata_error.h
```diff
@@ -7,14 +7,28 @@
...
+  };
+
+  DECLARE_JSON_TYPE(ODataErrorDetails);
+  DECLARE_JSON_REQUIRED_FIELDS(ODataErrorDetails, auth_policy, code, message);
+
   struct ODataError
   {
     std::string code;
     std::string message;
+    std::vector<ODataErrorDetails> moredetails = {};
```
eddyashton, line 26: This field should be called `details`, rather than `moredetails`, to match the field defined by the [OData spec](http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#_Toc38457792). We can insert any other fields we want (as we do with the `auth_policy` in `ODataErrorDetails`), but in this case we're implementing a standard/expected field.

src/node/rpc/frontend.h
```diff
@@ -258,20 +258,36 @@ namespace ccf
...
+                 ccf::errors::InvalidAuthenticationInfo,
+                 auth_error_reason});
+            }
           }
 
           if (identity == nullptr)
           {
-            // If none were accepted, let the last set an error
+            // If none were accepted, let the last set the response header
             endpoint->authn_policies.back()->set_unauthenticated_error(
```
eddyashton, line 282: I think we should go to each of the implementations of `set_unauthenticated_error`, and remove the calls to `ctx->set_error`. They construct an OData object for their own error and insert that into the body, but that's then overwritten by the call to `ctx->set_error` below. We should remove those calls that get overwritten to avoid confusion.

include/ccf/endpoints/authentication/authentication_types.h
```diff
@@ -29,6 +29,8 @@ namespace ccf
   public:
     virtual ~AuthnPolicy() = default;
 
+    static constexpr auto SECURITY_SCHEME_NAME = "";
```
eddyashton, line 32:
```suggestion
    static constexpr auto SECURITY_SCHEME_NAME = "BaseAuthPolicy";
```

We should insert a default value here, so we can spot if it ever leaks through.

samples/apps/logging/logging.cpp
```diff
@@ -52,6 +52,7 @@ namespace loggingapp
   class CustomAuthPolicy : public ccf::AuthnPolicy
   {
   public:
+    static constexpr auto SECURITY_SCHEME_NAME = "";
```
eddyashton, line 55:
```suggestion
    static constexpr auto SECURITY_SCHEME_NAME = "CustomAuthPolicy";
```

Since `SECURITY_SCHEME_NAME` is now part of the required API, we should use a sensible value for this sample policy.

src/node/rpc/frontend.h
```diff
@@ -258,20 +258,36 @@ namespace ccf
...
             identity = policy->authenticate(tx, ctx, auth_error_reason);
             if (identity != nullptr)
             {
               break;
             }
+            else
+            {
+              // Collate error details
+              errordetails.push_back(
+                {policy->SECURITY_SCHEME_NAME,
```
eddyashton, line 273: I don't think this will actually work? The type of `policy` here is `std::shared_ptr<AuthPolicy>`, so `policy->SECURITY_SCHEME_NAME` is always looking up the static member in `AuthPolicy`, not the derived classes. You'll need a `virtual std::string get_scheme_name()` or similar to correctly override this.

Quick demo here:
https://coliru.stacked-crooked.com/a/51a2c0da21e1ee7c

MahatiC: Ah let me check. I did initially have a `get_scheme_name`, but thought this would work.

include/ccf/odata_error.h
MahatiC: Done

include/ccf/endpoints/authentication/authentication_types.h
MahatiC: Done

samples/apps/logging/logging.cpp
MahatiC: Done

src/node/rpc/frontend.h
MahatiC: Good catch, there is an explicit method to return the name now.

eddyashton (approved):
LGTM!

jumaffre (approved):

src/node/rpc/frontend.h
```diff
@@ -258,20 +258,36 @@ namespace ccf
         if (!endpoint->authn_policies.empty())
         {
           std::string auth_error_reason;
+          std::vector<ccf::ODataErrorDetails> errordetails;
```
achamayou, line 261:
```suggestion
          std::vector<ccf::ODataErrorDetails> error_details;
```

src/node/rpc/frontend.h
```diff
@@ -258,20 +258,36 @@ namespace ccf
...
           {
             identity = policy->authenticate(tx, ctx, auth_error_reason);
             if (identity != nullptr)
             {
               break;
             }
+            else
+            {
+              // Collate error details
+              errordetails.push_back(
```
achamayou, line 272:
```suggestion
              error_details.push_back(
```

src/node/rpc/frontend.h
```diff
@@ -258,20 +258,36 @@ namespace ccf
...
+            // If none were accepted, let the last set the response header
             endpoint->authn_policies.back()->set_unauthenticated_error(
               ctx, std::move(auth_error_reason));
+            // Return collated error details for the auth policies declared
+            // in the request
+            ctx->set_error(
+              HTTP_STATUS_UNAUTHORIZED,
+              ccf::errors::InvalidAuthenticationInfo,
+              "Invalid info",
+              errordetails);
```
achamayou, line 290:
```suggestion
              error_details);
```

achamayou (approved):

PR #4098, opened 2022-08-04
[release/2.x] Support new quote formats in 2.x

jumaffre:
Because of the new `Insecure_Virtual` quote format type introduced in 3.x (see https://github.com/microsoft/CCF/pull/4083), 2.x nodes should be able to parse this format when the service is being upgraded from 2.x to 3.x. This PR allows parsing of these formats on the 2.x release branch, with end-to-end tests to test the expected behaviour.

Note that 2.x **virtual** nodes still produce empty quotes with the `OE_sgx_v1` format, for consistency with current 2.x releases.


src/node/rpc/node_frontend.h
```diff
@@ -139,19 +139,18 @@ namespace ccf
       {
         case QuoteVerificationResult::Failed:
           return std::make_pair(
-            HTTP_STATUS_INTERNAL_SERVER_ERROR, "Quote could not be verified");
+            HTTP_STATUS_UNAUTHORIZED, "Quote could not be verified");
```
jumaffre, line 142: Since the quote is supplied by the client on join, the error status when the quote is invalid should be client rather than server errors.

eddyashton: Not sure if this is true of all these errors. If the local verification toolchain blows up, that's still a `SERVER_ERROR`? But I agree for the known failure codes.

jumaffre: Good point. I'll change the generic error to a `SERVER_ERROR`.

DomAyre (approved):
Looks okay to me aside from the failing CI

eddyashton (approved):

PR #4102, opened 2022-08-04
Re-execute auth policies on conflict

eddyashton:
Includes #4101, so keeping as a draft until that has merged. This will also need to be manually merged with #4093, since they touch the same lines of auth code.

The auth policies were previously executed once, at the start of `process_command()`, and were not re-executed on Tx conflicts, meaning they could break serializability. The fix is simple (moving a block of code in `frontend.h`), but it's non-trivial to test. I added a `TestManualConflictsFrontend` that uses `condition_variable` wait points to insert arbitrary execution (including committing other transactions) during an endpoint's execution. I've also added a bunch of other unit tests with this, to confirm operations have the expected dependencies. The relevant unit test, testing the auth policy issue that we're actually looking at, is the last one ("Removed caller ident post-read").


MahatiC (approved):
Lgtm pending rebase

PR #4106, opened 2022-08-05
Support SEV-SNP attestation in CCF

DomAyre:
Contributes to https://github.com/microsoft/CCF/issues/4068

Key Decisions:
- We perform a runtime check in virtual builds in PAL to see if we support SEV-SNP
- We strictly don't allow any kind of mixed networks for now
- We make a breaking change to the code_ids table, to state the platform the id came from, this is to distinguish potentially identical ids generated by different attestations (either by very remote chance or more likely by an attack on a compromised platform)

Todo:
- [x] Refactor attestation generation/verification into the PAL layer (covered by https://github.com/microsoft/CCF/pull/4083)
- [x] Generate and store the raw quote
- [x] Parse the quote to extract the code id and report data
- [x] Extract and verify the quote endorsements
  - [ ] Authenticate when getting endorsements
  - [ ] Allow user to specify where to get endorsements from
  - [x] Get the endorsements in PAL


include/ccf/ds/attestation_types.h
```diff
@@ -0,0 +1,191 @@
...
+    uint64_t raw;
+  };
+
+  struct signature {
+    uint8_t r[72];
+    uint8_t s[72];
+    uint8_t reserved[512-144];
+  };
+
+  struct attestation_report {
```
jumaffre, line 106: It would be nice to keep the link to the AMD document as a comment, along with the corresponding table number.

include/ccf/ds/attestation_types.h
```diff
@@ -0,0 +1,191 @@
...
+    uint8_t       measurement[attestation_measurement_size];  /* 0x090 */
+    uint8_t       host_data[32];                              /* 0x0C0 */
+    uint8_t       id_key_digest[48];                          /* 0x0E0 */
+    uint8_t       author_key_digest[48];                      /* 0x110 */
+    uint8_t       report_id[32];                              /* 0x140 */
+    uint8_t       report_id_ma[32];                           /* 0x160 */
+    union tcb_version reported_tcb;                           /* 0x180 */
+    uint8_t       reserved1[24];                              /* 0x188 */
+    uint8_t       chip_id[64];                                /* 0x1A0 */
+    //   uint8_t       reserved2[192];                        /* 0x1E0 */
```
jumaffre, line 128: Remove?

include/ccf/ds/attestation_types.h
```diff
@@ -0,0 +1,191 @@
...
+
+  static constexpr oe_uuid_t oe_quote_format = {OE_FORMAT_UUID_SGX_ECDSA};
+  static constexpr auto sgx_report_data_claim_name = OE_CLAIM_SGX_REPORT_DATA;
+
+#else
+  static constexpr size_t attestation_measurement_size = 48;
+  using attestation_measurement =
+    std::array<uint8_t, attestation_measurement_size>;
+
+  union tcb_version {
```
jumaffre, line 89: It would be nice to move the AMD-specific types to a new file.

include/ccf/ds/pal.h
```diff
@@ -81,6 +94,106 @@ namespace ccf
...
+
+      else if (quote_info.format == QuoteFormat::amd_sev_snp_v1) {
+        if (!is_sev_snp) {
+          throw std::logic_error(
+            "Cannot verify SEV-SNP quote if node is virtual");
+        }
+        attestation_report quote = *reinterpret_cast<const attestation_report*>(quote_info.quote.data());
+        std::copy(
+          std::begin(quote.report_data),
+          std::begin(quote.report_data) + attestation_report_data_size,
```
jumaffre, line 174: This needs fixing. We shouldn't truncate the measurement on AMD but use the full 64 bytes. I guess `attestation_measurement` has to become a `std::vector<uint8_t>` and we should check its size based on the type of platform type later.

DomAyre: This is the report data rather than the measurement, the report data if I understand correctly is provided by us and is a 32 byte hash of the nodes public key? In which case I don't think there is any issues with truncating down to 32 bytes right?

The measurement uses all 48 bytes for AMD as of this commit: https://github.com/microsoft/CCF/pull/4106/commits/48b6eabca9ca5801916b4aab826de5913345d573

include/ccf/ds/attestation_types.h
DomAyre: I've added a link to the ABI spec as well as and references to tables where I thought it was appropriate.

include/ccf/ds/attestation_types.h
DomAyre: Yeah sounds good

include/ccf/ds/attestation_types.h
DomAyre: Done a full split into attestation_sgx.h and attestation_sev_snp.h

include/ccf/ds/attestation_sev_snp.h
```diff
@@ -0,0 +1,151 @@
...
+  };
+
+  struct signature {
+    uint8_t r[72];
+    uint8_t s[72];
+    uint8_t reserved[512-144];
+  };
+
+  // Table. 105
+  enum class signature_algo_t : uint32_t
```
eddyashton, line 56: Our coding standard is for `CamelCaseTypes`, so this (for example) should be `SignatureAlgo`. These types also shouldn't take names in the `ccf::` namespace, they should probably be `ccf::amd`, or `ccf::pal::sevsnp`, or something along those lines.

include/ccf/ds/attestation_sev_snp.h
```diff
@@ -0,0 +1,151 @@
...
+  union tcb_version {
+    struct _tcb_version {
+      uint8_t boot_loader;
+      uint8_t tee;
+      uint8_t reserved[4];
+      uint8_t snp;
+      uint8_t microcode;
+    };
+    uint64_t raw;
+  };
```
eddyashton, line 47: I like anonymous structs, but I think the union here is unnecessary if we're only reading from `.raw`. I think this should be:

```
struct TcbVersion { uint8_t boot_loader; ... }
static_assert(sizeof(TcbVersion) == sizeof(uint64_t), "Can't cast TcbVersion to uint64_t");
```

Then anywhere you want to read this memory as a uint64, you instead do an explicit memory cast (`*(uint64_t*)(&tcb_version)`).

doc/audit/builtin_maps.rst
```diff
@@ -128,7 +128,7 @@ Versions of the code allowed to join the current network.
 
 **Value** Represented as JSON.
 
-.. doxygenenum:: ccf::CodeStatus
+.. doxygenstruct:: ccf::CodeInfo
```
achamayou, line 131: I think it's the first time we do this (change the value of a type, as opposed to add a new field). Stating that the value here is always a CodeInfo represented as JSON is a stretch, the old ones are are a CodeStatus.

I think we can do one of two things:

1. Leave the CodeStatus type in, and document one before the version, and the other after (literally)
2. Remove the old type, but document it inline here, with the version at which it was removed.

Someone writing an script that audits the ledger is (possibly) going to want to know about this.

include/ccf/pal/attestation_sev_snp.h
```diff
@@ -0,0 +1,167 @@
...
+
+    // Table. 105
+    enum class SignatureAlgorithm : uint32_t
+    {
+      invalid = 0,
+      ecdsa_p384_sha384 = 1
+    };
+
+    // Table 21
+    struct Attestation
```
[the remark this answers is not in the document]
DomAyre: I'm not very familiar with this, from my reading it seems like the decision is about is a saving memory at potentially the cost of performance, is that right? If so, I'm not sure if we're especially memory constrained in the same way we would be for SGX? 

Could you explain for my relatively novice level understanding why that might be a trade-off we want?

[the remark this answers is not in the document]
wintersteiger: Tight packing means that the compiler is not allowed to pad any of the space between the fields. This does save a little bit of memory sometimes, but it is more important for structs the come out of the hardware, i.e. the hardware won't put any padding between the fields, so your compiler shouldn't either, if you're using the struct as a direct representation of the what the hardware produces. Without that, you'd need some sort of translation from packed to unpacked.

In this particular case, I don't think it's strictly necessary, but I think it would be a good idea to do it anyway.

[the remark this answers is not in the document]
achamayou: +1 for being explicit. @DomAyre there isn't really a tradeoff here, it's what we get in the attestation, we need to make sure we cast to the right offsets.

doc/audit/builtin_maps.rst
```diff
@@ -128,7 +128,17 @@ Versions of the code allowed to join the current network.
...
-.. doxygenenum:: ccf::CodeStatus
+**Pre 3.0.0-dev2**
+
+``enum class ccf::CodeStatus``
+
+*Values:*
+
+``enumerator ALLOWED_TO_JOIN``
+
+**After 3.0.0-dev2**
```
achamayou, line 139: Best to have that at first I think.

include/ccf/crypto/pem.h
```diff
@@ -134,6 +134,23 @@ namespace crypto
     return "Pem";
   }
 
+  static std::vector<crypto::Pem> split_x509_cert_bundle(
```
achamayou, line 137: Unit test please :)

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+      snp::AttestationReq req = {};
+      snp::AttestationResp resp = {};
+
+      // Arbitrary report data
+      memcpy(req.report_data, report_data.data(), attestation_report_data_size);
+
+      snp::GuestRequest payload = {
+        .req_msg_type = snp::MSG_REPORT_REQ,
+        .rsp_msg_type = snp::MSG_REPORT_RSP,
+        .msg_version = 1,
```
achamayou, line 58: why is that 1?

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+        throw std::logic_error(
+          "Failed to issue ioctl SEV_SNP_GUEST_MSG_REPORT");
+      }
+
+      auto quote = &resp.report;
+      auto quote_bytes = reinterpret_cast<uint8_t*>(&resp.report);
+      node_quote_info.quote.assign(quote_bytes, quote_bytes + resp.report_size);
+
+      client::RpcTlsClient client{
+        "americas.test.acccache.azure.net", // TODO: Make Configurable
```
achamayou, line 79: Needs a ticket

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+        "443",
+        nullptr,
+        std::make_shared<tls::Cert>(
+          nullptr, // TODO: Use auth
```
achamayou, line 83: Needs a ticket, contact the owners and find out

include/ccf/pal/attestation.h
achamayou: Default to AMD

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+          "Cannot verify SEV-SNP quote if node is virtual");
+      }
+
+      auto quote =
+        *reinterpret_cast<const snp::Attestation*>(quote_info.quote.data());
+
+      std::copy(
+        std::begin(quote.report_data),
+        std::begin(quote.report_data) + attestation_report_data_size,
+        // quote supports 64 bytes but our code only uses 32
```
achamayou, line 143: expand

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+
+      auto root_cert_verifier = crypto::make_verifier(root_certificate);
+
+      if (
+        root_cert_verifier->public_key_pem().str() !=
+        snp::amd_milan_root_signing_public_key)
+      {
+        throw std::logic_error(
+          "The root of trust public key for this attestation was not the "
+          "expected one");
```
achamayou, line 164: add fingerprint of the unexpected root of trust

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+
+      if (quote.signature_algo != snp::SignatureAlgorithm::ecdsa_p384_sha384)
+      {
+        throw std::logic_error(fmt::format(
+          "Unsupported signature algorithm: {} (supported: {})",
+          quote.signature_algo,
+          snp::SignatureAlgorithm::ecdsa_p384_sha384));
+      }
+
+      // Make ASN1 der signature
```
achamayou, line 191:
```suggestion
      // Make ASN1 DER signature
```

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+      auto quote_signature = crypto::ecdsa_sig_from_r_s(
+        quote.signature.r,
+        sizeof(quote.signature.r),
+        quote.signature.s,
+        sizeof(quote.signature.s),
+        false /* little endian */
+      );
+
+      std::vector<uint8_t> quote_without_signature = {
```
achamayou, line 200: Could maybe be a span?

include/ccf/pal/attestation.h
```diff
@@ -0,0 +1,373 @@
...
+      node_quote_info.format = QuoteFormat::insecure_virtual;
+    }
+
+    else
+    {
+      node_quote_info.format = QuoteFormat::amd_sev_snp_v1;
+      int fd = open(snp::DEVICE, O_RDWR | O_CLOEXEC);
+      if (fd < 0)
+      {
+        throw std::logic_error("Failed to open \"/dev/sev\"");
```
achamayou, line 46: DEVICE

achamayou (approved):

doc/audit/builtin_maps.rst
DomAyre: Added a note below stating old version

include/ccf/crypto/pem.h
DomAyre: Split off into https://github.com/microsoft/CCF/issues/4142

include/ccf/pal/attestation.h
DomAyre: Added a link to docs above

include/ccf/pal/attestation.h
DomAyre: Looks like using the AMD URL returns a certificate that when parsed as PEM is slightly different to the one from Azure. I don't know if there's an issue with our/openssl's parsing or if the certificate is genuinely different, but for the sake of getting something that works merged, I propose we add the AMD server as default as part of https://github.com/microsoft/CCF/issues/4144

achamayou: Sounds good to me.

include/ccf/pal/attestation.h
achamayou: Stapled to #4144

include/ccf/ds/attestation_sev_snp.h
achamayou: @eddyashton the union is about the same to define, and slightly shorter and neater at call sites though? I don't feel strongly about this.

eddyashton: If we actually use the union, it's technically undefined behaviour (reading from any union member other than the last written). It's handled sensibly by all known compilers, but still a risk (and might be flagged by smarter tooling in future). With all of these, I think the purpose of the struct should be to expose/decompose the internal fields, and the fact they're efficiently packed/serializable is separate, not intrinsic. If the call site gets too ugly, we can wrap it in a method (maybe all of these structs deserve something like a `std::range<uint8_t, N> raw() const` member?)

MahatiC: I also think it's easier to read this way (with union), plus I don't see a downside of using the union here?

include/ccf/pal/attestation_sev_snp.h
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    // Table. 105
+    enum class SignatureAlgorithm : uint32_t
+    {
+      invalid = 0,
+      ecdsa_p384_sha384 = 1
+    };
+
+#  pragma pack(push, 1)
```
eddyashton, line 67: I think this packing argument applies to all of the structs defined in this class, so we should pack all of them? ie - if there's any place where we try and cast raw bytes to one of these structs, we need to make sure they're appropriately packed.

include/ccf/crypto/pem.h
eddyashton, line 137: I think we could do this with `nonstd::split` (defined in `ds/nonstd.h`)? Would then need an additional parsing of each as `Pem`s. No urgent though - makes sense to leave this until the unit test in #4142, and simplify the implementation if it still passes that test.

include/ccf/ds/attestation_sev_snp.h
achamayou: @eddyashton huh, indeed, it's only allowed in C99, but explicitly UB in C++ since '11. Very unfortunate.

include/ccf/pal/attestation.h
```diff
@@ -2,22 +2,108 @@
...
+
+      auto params = nlohmann::json::object();
+      params["api-version"] = "2020-10-15-preview";
+
+      auto response = client.get(
+        fmt::format(
+          "/SevSnpVM/certificates/{}/{}",
+          fmt::format("{:02x}", fmt::join(quote->chip_id, "")),
+          fmt::format("{:0x}", *(uint64_t*)(&quote->reported_tcb))),
+        params);
```
eddyashton, line 96: This client code was previously only used in untrusted clients, so this is the first thing that brings it into the TCB. The underlying crypto bits are shared with the already trusted implementation, so probably fine, but we need to be aware of this change. This is also a blocking call making a network request, with no timeout. We may need to add a timeout, or convert this entire `generate_quote` call to something async to handle implementations where it is inherently async. Fine for this first-pass PR, but I think we should follow up with another examination/hardening pass.

include/ccf/pal/attestation.h
```diff
@@ -26,31 +112,129 @@ namespace ccf::pal
...
+      std::span quote_without_signature{
+        quote_info.quote.data(),
+        quote_info.quote.size() - sizeof(quote.signature)};
+      if (!chip_cert_verifier->verify(quote_without_signature, quote_signature))
+      {
+        throw std::logic_error(
+          "Chip certificate (VCEK) did not sign this attestation");
+      }
+    }
+
```
eddyashton, line 208:
```suggestion
```

include/ccf/pal/attestation.h
```diff
@@ -26,31 +112,129 @@ namespace ccf::pal
...
-        "Cannot verify real attestation report on virtual build");
+      if (is_sev_snp)
+      {
+        throw std::logic_error(
+          "Cannot verify virtual quote if node is SEV-SNP");
+      }
+      unique_id = {};
+      report_data = {};
+    }
+
```
eddyashton, line 127:
```suggestion
```

include/ccf/pal/attestation.h
```diff
@@ -2,22 +2,108 @@
...
-    node_quote_info.format = QuoteFormat::insecure_virtual;
+    auto is_sev_snp = access(snp::DEVICE, F_OK) == 0;
+
+    // If there is no SEV-SNP device, assume we are using insecure virtual
+    // quotes
+    if (!is_sev_snp)
+    {
+      node_quote_info.format = QuoteFormat::insecure_virtual;
+    }
+
```
eddyashton, line 39:
```suggestion
```

eddyashton (commented):
Coding style nit: we don't add blank lines before `else`, only (sometimes, for readability) before `if`.

eddyashton (approved):
A few suggestions for cleanup/followup, but nothing critical. LGTM!

include/ccf/pal/attestation.h
DomAyre: Sounds like we can do it alongside making this whole thing configurable?

achamayou: Side note: that doesn't fundamentally need to be trusted, it could be done on the host side, before starting the enclave. In Parma, there is no such distinction, but there will be in Monza, and so could be on that side there.

include/ccf/crypto/pem.h
DomAyre: Added a note to https://github.com/microsoft/CCF/issues/4142

include/ccf/pal/attestation_sev_snp.h
DomAyre: I've wrapped most of the structs, AttestationReq and GuestRequest  we construct ourselves so left those

include/ccf/ds/attestation_sev_snp.h
DomAyre: If everyone is happy with version I have I will resolve this thread

wintersteiger: TBH, I think a class with a private `uint8_t bs[8];` and accessor functions like `uint8_t tee() const { return bs[1]; }` would look even neater to me :-)

src/enclave/rpc_sessions.h
```diff
@@ -3,6 +3,7 @@
 #pragma once
 
 #include "ccf/ds/logger.h"
+#include "ccf/pal/locking.h"
```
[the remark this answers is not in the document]
DomAyre: Leaving for now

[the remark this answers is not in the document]
achamayou: Let's keep it that way for now, further renaming can happen in other PRs.

PR #4117, opened 2022-08-11
Fix failures in multi-threaded indexing unit tests

### LINKED ISSUE #4100: Investigate failures in multi-threaded indexing unit tests
Added in #4075, temporarily disabled in #4099.

Seems to only fail on IceLake executors.

eddyashton:
Resolves #4100.

~~Temporary Draft while I throw the CI at it~~

I believe I've fixed the issues in both tests that were leading to failures (aggressive backoff sometimes backing off too far, and unlucky timing ending a necessary tick before the polls had even started executing).

### DIFF
--- a/src/indexing/test/indexing.cpp
+++ b/src/indexing/test/indexing.cpp
@@ -458,8 +458,7 @@ const auto max_multithread_run_time = 10s;
 
 // Uses the real classes, and access + update them concurrently
 TEST_CASE(
-  "multi-threaded indexing - in memory" * doctest::test_suite("indexing") *
-  doctest::may_fail(true))
+  "multi-threaded indexing - in memory" * doctest::test_suite("indexing"))
 {
   auto kv_store_p = std::make_shared<kv::Store>();
   auto& kv_store = *kv_store_p;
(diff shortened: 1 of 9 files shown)


achamayou (approved):

src/indexing/test/indexing.cpp
```diff
@@ -833,16 +833,14 @@ TEST_CASE(
...
 
-      std::chrono::milliseconds sleep_time(10);
       while (!results.has_value())
       {
         // May be contesting for limited cached buckets with other users of this
-        // index (no handle for unique claims). Back-off exponentially, with
-        // random variation, to break deadlock
+        // index (no handle for unique claims). Uniform random sleep to avoid
+        // deadlock.
+        const auto sleep_time = std::chrono::milliseconds(rand() % 100);
```
MahatiC, line 841: Question: If it's multi-threaded, and if the expected behavior is for multiple users to be able to share the cached buckets, shouldn't we be handling the deadlock in the code than the test?

MahatiC, line 841: Question: If it's multi-threaded, and if the expected behavior is for multiple users to be able to share the cached buckets, shouldn't we be handling the deadlock in the code than the test?

eddyashton: It's not a deadlock, just contention (that might block each other for so long that the test fails). This contention is inevitable - if we're executing multiple requests in parallel, the polling APIs we provide lock internally, but when they're doing async fetches its possible for them to pre-empt each other. Dealing with that contention while still providing liveness is left to the apps (to add back-offs, leases, rate limits, whatever is appropriate). It's another case where, with more memory, we'd likely have a friendlier caching layer on top (ie - implementing handles or leases as helpers), but those are only mitigations - for arbitrary workloads, there will always come a point where the app has to decide how to allocate limited resources.

MahatiC (approved):
CI seems happy!

Issue #4040, opened 2022-07-15
Dependency updates July '22


achamayou:
- [ ] CLI11
- [ ] fmtlib
- [ ] llhttp

achamayou:
p1 because of llhttp

PR #4122, opened 2022-08-12
Upgrade fmtlib from 8.1.1 to 9.0.0

achamayou:
Second step of #4040


src/kv/kv_types.h
```diff
@@ -797,8 +797,8 @@ struct formatter<kv::Configuration::Nodes>
   }
 
   template <typename FormatContext>
-  auto format(const kv::Configuration::Nodes& nodes, FormatContext& ctx)
-    -> decltype(ctx.out())
+  auto format(const kv::Configuration::Nodes& nodes, FormatContext& ctx) const
+    -> decltype(ctx.out()) const
```
eddyashton, line 801:
```suggestion
  auto format(const kv::Configuration::Nodes& nodes, FormatContext& ctx) const
    -> decltype(ctx.out())
```

Second const alters the return type, not the method.

eddyashton (approved):

PR #4124, opened 2022-08-12
Upgrade nlohmann::json from 3.10.2 to 3.11.2

achamayou:
This is deliberately not back ported to 2.x (yet), because the release is too recent at this point.


DomAyre:
Do we expect reviewers to verify that the hash matches the new 3rd party code?

achamayou:
@DomAyre not quite, the hash is the git sha for the release we use, in the nlohmann/json repo. So there's one more indirection. I think it would be nice if the PR reviewers did check it, I try to :)

eddyashton (approved):

PR #4125, opened 2022-08-12
Enable perf regression tests for virtual builds

### LINKED ISSUE #3302: Performance tests in CI for virtual build
From https://github.com/microsoft/CCF/pull/3258#discussion_r767517496.

@achamayou said:
> perf tests are run on an SGX build now. If the perf job built SGX + virtual + v8 app, there should be no difficulty in running all tests since they are decorated accordingly and won't collide.
>
> What this would do is make the perf job take longer to build and run, but a possibility we've discussed if that's the case is to split that into two separate jobs, and a have a third dependent job do the plotting.
>
> I think the main impact would be the sheer amount of metrics displayed in PRs, which would probably only be manageable with jumaffre/cimetrics#75

achamayou:
Close #3302. Now that rendering is split, we can also push metrics from non-perf end-to-end tests just by adding the right job dependency, for example the total ledger size in reconfiguration_test.


cmake/common.cmake
```diff
@@ -602,6 +602,8 @@ function(add_perf_test)
   set(TESTS_SUFFIX "")
   if("sgx" IN_LIST COMPILE_TARGETS)
     set(TESTS_SUFFIX "${TESTS_SUFFIX}_sgx")
+  else()
+    set(TESTS_SUFFIX "${TESTS_SUFFIX}_virtual")
```
eddyashton, line 606: This looks like a coarse distinction. Can we instead have both `ls_sgx_cft` and `ls_virtual_cft` if we built both (choosing between them by adding a `--enclave-type` arg to the python script)?

achamayou: done

eddyashton (approved):

achamayou:
Ah the micro-benchmarks aren't decorated with the architecture. We need to either do that and split them, or only run them once. Thoughts @eddyashton?

eddyashton:
> Ah the micro-benchmarks aren't decorated with the architecture. We need to either do that and split them, or only run them once. Thoughts @eddyashton?

Ah. Probably run them only once, with some combination of `-L` and `-LE`?

PR #4126, opened 2022-08-15
Add a method that decodes path parameters

### LINKED ISSUE #4112: Getting decoded request path parameter
We talked about this in a previous email thread but since it didn't come out in 2.0.5 we might've lost track of it, so creating an issue just in case (I also didn't find an existing issue)! As discussed, the current code
* URL decodes queries
* URL decodes fragments

It does not
* URL decode the path before dispatch
  * So "/foo bar" and "/foo%20bar" are two different endpoints
* URL decode the path params after dispatch

This is believed to be intentional because of https://github.com/microsoft/CCF/blob/main/src/http/test/http_test.cpp#L393.

While normalizing paths for dispatch could be a topic for future discussion, a more immediately useful method would be one like `get_decoded_request_path_parms()`, which returns URL-decoded path parameters (unlike the existing `get_path_param()`, which will return the path parameter, encoded).

Currently the application must explicitly do the decoding after fetching path parameters.

MahatiC:
Resolves #4112


MahatiC:
Would like to take a second look at this, so marked it as draft. But comments are welcome.

src/endpoints/endpoint_registry.cpp
```diff
@@ -298,14 +299,18 @@ namespace ccf::endpoints
...
       for (auto& [original_method, verb_endpoints] : templated_endpoints)
       {
         auto templated_endpoints_for_verb =
           verb_endpoints.find(rpc_ctx.get_request_verb());
         if (templated_endpoints_for_verb != verb_endpoints.end())
         {
           auto& endpoint = templated_endpoints_for_verb->second;
-          if (std::regex_match(method, match, endpoint->spec.template_regex))
+          if (
+            std::regex_match(method, match, endpoint->spec.template_regex) &&
```
achamayou, line 311: I don't think we can do something this simple here, because URL equivalence is tricky. This is detailed under https://www.rfc-editor.org/rfc/rfc3986#section-6

For example "/foo/b%%2far" should match "/foo/{id}" with decoded id "b/ar", but this will instead look for a "/foo/b/ar" endpoint.

We can look at improving dispatch, but I suggest that's a separate matter from decoding path params, as agreed on the ticket.

MahatiC: Isn't `/` a reserved character for path parameters?

achamayou: It is, but "%2f" isn't, and decoding before dispatch makes them indistinguishable to the dispatch.

achamayou (approved):
LGTM, but since this is a user-facing change could you please update the changelog?

MahatiC:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|release/2.x|Conflict resolution was aborted by the user|

### Manual backport
```
backport --pr 4126
```

### Questions ?

PR #4127, opened 2022-08-15
Mitigations for potential stale reads from Legacy xAPIC

eddyashton:
Vulnerability described here:
https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/stale-data-read-from-xapic.html
https://www.intel.com/content/www/us/en/security-center/advisory/intel-sa-00657.html

The solution is to require 8-byte aligned (start address _and size_) reads when the SGX enclave reads from host memory. We have relatively few code paths that do this, so this is primarily validation of the arguments around the initial ECALL (and manual padding by the host for the variably sized JSON config argument). The tricky instance is the ringbuffer, which would like to hand off a host memory range to each handler to deserialize at will. To avoid those handlers doing unaligned reads, the `ringbuffer::Reader` now contains a path which copies the (overaligned) message memory range into enclave memory before passing to the handlers. This introduces a significant extra cost, so is only done on vulnerable platforms, as detected by a run-time call to `cpuid`. There's also a compile-time switch to force enable this where necessary.


include/ccf/ds/pal.h
```diff
@@ -48,6 +49,47 @@ namespace ccf
...
+
+  static void cpuid(CpuidInfo* info, uint64_t leaf, uint64_t subleaf)
+  {
+    asm volatile(
+      "cpuid"
+      : "=a"(info->eax), "=b"(info->ebx), "=c"(info->ecx), "=d"(info->edx)
+      : "a"(leaf), "c"(subleaf));
+  }
+
+  static bool is_intel_cpu()
```
achamayou, line 68: While the PAL is definitely the right place for this in principle, I'm not sure we want to expose all of this in our public API?

eddyashton: I think that's true of pretty much everything in the `PAL` - it should move to private implementation at some point, but we don't want to shard the implementation.

Weighing this up with the _current_ PAL, I think we should give up on the class and expose these (already static, stateless) functions to a `pal` namespace. That way we can easily add implementations in several places (ie - some files only need `pal/attestation.h`), rather than trying to maintain a single monolothic PAL, and in turn move most of those implementations to private.

achamayou: Agreed. It's fine to go this route for now in the interest of time.

achamayou (approved):

achamayou:
@eddyashton if you add the following under parameters in matrix.yml:

```
    SGXIceLake:
      container: sgx
      pool: 1es-dcdv3-focal
```

and switch `target` to `SGXIceLake` for the perf job on line 62, you should be able to get perf runs there. A separate PR will need to be open with just these changes though, to have an accurate comparison.

eddyashton:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|release/2.x|Conflict resolution was aborted by the user|

### Manual backport
```
backport --pr 4127
```

### Questions ?

eddyashton:
Manually backporting this, as it includes changes in the non-backported PAL.

PR #4138, opened 2022-08-16
Test AMD pool

achamayou:
This is running the virtual perf tests on Standard D8ads v5 (AMD), rather than Standard D8ds v5 (Intel).


achamayou:
Switch from D4ads to D8ads

/azp run

PR #4145, opened 2022-08-17
Permanently delete items from the KV

### LINKED ISSUE #1992: Permanently delete items in the KV champ maps

eddyashton:
Resolves #1992.

Until now we've retained tombstones for everything that was ever present in the KV, with negative versions indicating deletions. This instead removes those entries for real. Note that in `ChangeSet`s and the like during computation, we still track deletions with `nullopt`s, but when we finally apply that to a `state` (a `ChampMap`), we truly forget about it.

`DeleteableVersion` and `is_deleted()` are gone, ~~and we technically get another bit of representable `Version`s~~ _but we don't get another bit of representable Versions, because we need to maintain compatibility with old representations_.

The semantic impact of this is most obvious on snapshots, as shown in our `kv_snapshot` unit test. Previously, they would record every key which had ever existed, including deletions. Now they don't contain anything that was deleted before they are created.


src/kv/untyped_map.h
```diff
@@ -225,7 +223,7 @@ namespace kv::untyped
             if (search.has_value())
             {
               changes = true;
-              state = state.put(it->first, VersionV{-v, v_, {}});
+              state = state.remove(it->first);
```
achamayou, line 226: I am concerned that removing the deletion from the state results in it being gone from the roll, and affects conflict resolution. What am I missing?

achamayou: Ok, so mostly for myself, this seems to be fine, because a read dependency on a key missing is not inherently versioned. So long as the key is not currently there, there is no conflict, even if the key was added and removed since.

achamayou (approved):

eddyashton:
Repeating a big comment from offline here, largely a rewording of the discussion in #1992 but hopefully convincing that this is safe:

Is there any difference in the app-facing API between a "never-existed" key and a "deleted at the version you're reading" key?

If there is then we need to treat them differently during conflict resolution (and as such, remember them forever).

In CCF there's not, so we _can_ really delete them.

This means during conflict resolution, we can be considering a read of a very old version, before the key ever existed.

Consider a transaction T, which during execution took a branch on non-existence, and wrote "key K doesn't exist" somewhere else.

In the interim, another transaction came along and wrote K, and committed, then another-other transaction deleted K, and committed.

Now our original transaction T comes to commit, and sees no conflict (because the head of the roll has returned to a state where it says "key K is at NoVersion", matching the read we did at "NoVersion"), and is allowed to commit.

The result is equivalent to what this transaction would have done if it had executed entirely after the deletion.

eddyashton:
From the last attempt at this, in #1781:
> Do we correctly do read-your-own-write dependence on keys we deleted ourselves, whether or not they previously existed?

I'll add a unit test of this before merging.

include/ccf/kv/map_handle.h
```diff
@@ -229,16 +229,13 @@ namespace kv
...
      * @param key Key to be removed
-     *
-     * @return true if the key had a value previously
      */
-    bool remove(const K& key)
+    void remove(const K& key)
     {
-      return write_handle.remove(KSerialiser::to_serialised(key));
+      write_handle.remove(KSerialiser::to_serialised(key));
     }
```
eddyashton, line 239: @achamayou @MahatiC @jumaffre: This is a late-breaking, breaking-change - removing the return value from calls to `remove()`. The problem is that if we return a bool indicating whether this previously had a value, we may be introducing an implicit read dependency. ie, an application could do:

```
auto handle_foo = kv.rw(foo);
auto handle_bar = kv.rw(bar);
if (handle_foo->remove("A"))
{
  handle_bar->put("result", "Successfully removed A");
}
else
{
  handle_bar->put("result", "Failed to remove A");
}
```

For this to serialise correctly, we'd need to record and conflict a read dependency on all remove ops as well. Instead, I think it's simpler to remove this largely-unused return value. That makes `remove()` a pure-write, consistent with `put()` (which has no return value), and where apps actually _want_ a dependency like above to branch on, they can explicitly introduce it with a call to `has()`.

Another way to look at this is that it makes it possible to create pure-delete ops, which may be 're-ordered' like other pure-write ops. With a return value, there's a class of operations that the app-developer intended to be re-orderable as pure writes, but we track as reads and introduce unnecessarily conflicts.

achamayou: That seems fine for 3.x (given appropriate mention in release notes), but definitely not for backporting to 2.x.

eddyashton: Agreed. I've added a `CHANGELOG` entry, and wasn't intending to backport this to 2.x.

However, this has made me wonder how this interacts over multiple versions with ledgers that contain tombstones... I'll go add another test.

MahatiC: > it's simpler to remove this largely-unused return value. 

Yup, that seems like a good idea and a clean way to handle this

src/kv/test/kv_snapshot.cpp
```diff
@@ -165,6 +165,83 @@ TEST_CASE("Simple snapshot" * doctest::test_suite("snapshot"))
   }
 }
 
+TEST_CASE("Old snapshots" * doctest::test_suite("snapshot"))
```
eddyashton, line 168: Added this test for a simpler repro of the incompatibility that was failing the `lts_compat` test, that I alluded to in [this comment](https://github.com/microsoft/CCF/pull/4145/files#r949289405).

Basically the first version of this PR was too aggressive, and lost all knowledge of tombstone entries in serialized snapshots. This meant we couldn't parse snapshots from old nodes, which is a major compatibility break. This test checks we can still parse both kinds of snapshots, and we do that by retaining a few "is this Version negative" checks.

PR #4148, opened 2022-08-18
Refactor PAL into a namespace with separate header files

DomAyre:
Changed PAL from a class into a namespace
Separate header files with functionality organised into semantic groups
Simplifies imports as most PAL users only need the Mutex abstraction


include/ccf/pal/mem.h
```diff
@@ -0,0 +1,167 @@
...
+
+  struct CpuidInfo
+  {
+    uint64_t eax;
+    uint64_t ebx;
+    uint64_t ecx;
+    uint64_t edx;
+  };
+
+  static void cpuid(CpuidInfo* info, uint64_t leaf, uint64_t subleaf)
```
eddyashton, line 37: I don't think these `cpuid` related types/functions should be in `mem.h`, but instead in `pal/cpuid.h` or `pal/hardware_info.h`.

achamayou (approved):

include/ccf/pal/mem.h
DomAyre: Yeah, I can see that, I think there's also an argument about if getting information about your platform should even be in the abstraction layer since we're not really abstracting it? But that would probably complicate things and I can see that it might not be worth that?

eddyashton: I think the next step of the refactor would be to move some of the implementation details to `.cpp` so they're not exposed in the public PAL. Then the user-facing functions like `require_alignment_for_untrusted_reads` (and _maybe_ `is_intel_cpu`) would remain here, but the fact we're actually calling `cpuid` and how we handle that wouldn't be part of the public API.

eddyashton: But that's a "maybe one day", not worth doing right now.

achamayou: cpuid is at least partially emulated on at least SGX, and it's not inconceivable it will be on other TEEs too. It's also useful to have a degree of centralised control over this, for targeted manipulation rather than abstraction purposes.

DomAyre: Moved to hardware_info.h for now

Issue #3872, opened 2022-05-20
Support `make_read_only_endpoint` to be used with historical query adapter


letmaik:
https://github.com/microsoft/CCF/blob/1430f4dc48184ad0002176e2bc8557600299ee9f/samples/apps/logging/logging.cpp#L844-L848

I'm assuming the reason why the historical query adapter is written around `EndpointContext` instead of `ReadOnlyEndpointContext` is to give more flexibility in case something has to written to the KV when a historical query was made. Even though a rare use case, it may exist.

From a dev perspective though it seems odd that this means I have to disable the static guarantees that come from using `make_read_only_endpoint`. It would be nice if the adapter somehow supported read-only endpoint functions as well.

achamayou:
@letmaik that's a good point, we could have `read_only_adapter_v1` and `write_only_adapter_v1` I suppose?

PR #4149, opened 2022-08-18
Test behaviour of historical queries with missing files

### LINKED ISSUE #3692: Document and test behavior for historical queries when old ledger files are not yet available

achamayou:
Fixes #3692. The current behaviour is that we return 202, since we do not trust the host: an integrity-checked entry is found, anything else is the absence of an entry, and we have no way to check if it is final or not.

Blocked on fixing #3872 - see #4162


tests/infra/node.py
```diff
@@ -511,6 +515,28 @@ def get_ledger(self):
...
+    def hide_committed_ledger_chunk(self):
+        ledger_dir = self.remote.ledger_paths()[0]
+        ledger_chunks = list(pathlib.Path(ledger_dir).glob("*.committed"))
+
+        def interval(path):
+            start, end = path.stem.rsplit("_")[-1].split("-")
+            return int(start), int(end)
+
+        # Pick a random chunk, near the start of the ledger, to avoid the cache
+        (chunk_to_hide,) = random.sample(ledger_chunks[:5], 1)
```
achamayou, line 527: This is fairly nasty, I think the right change to make is to spin up a new node *without* access to historical files instead, but:

1. it's not quite the same thing (gap vs no files up to that point)
2. annoyingly, we only do reconfig with quotes enabled, not in virtual mode. We don't do any reconfig currently in e2e_logging.

We could:

1. move this to reconfig, but that's annoying because having lots of ledger files is useful, and only the logging test really does that.
2. keep this here, but then we add reconfig to logging tests.

achamayou:
After some investigation, I have realised that because of #3872, and of https://github.com/microsoft/CCF/blob/main/src/endpoints/endpoint_registry.cpp#L148, "/app/receipt" is always forwarded!

So the only reason the test passed locally most of the time was because the timeout was too low, and build (debug with verbose logging) was too slow! This is completely unaffected by the swapping out vs starting new node choice.

So there seem to be a few more choices here:

1. set forwarding to Sometimes on /app/receipt

That seems like the minimal change to get the test passing, and it's arguably a good thing most of the time in terms of scaling. But it's also worse in bad cases, where the entry is available on the primary!

2. Fix #3872 and update most, if not all historical endpoints to be read only

This is better, because it's consistent across endpoints, and reads generally scale better. It's still worse in the worst case scenario though.

3. Implement logic that decides, after a number of retries/time out on historical queries, to forward out. This is a bit awkward, because while choosing the primary for writes/session consistency makes sense, that's not the case for ledger gaps! The primary could have an even sparser history than we do.

So I'm leaning towards 2. at this point, but I will sleep on it first, and I'd like input from @eddyashton, @MahatiC and @DomAyre.

eddyashton:
> 2. Fix [Support `make_read_only_endpoint` to be used with historical query adapter #3872](https://github.com/microsoft/CCF/issues/3872) and update most, if not all historical endpoints to be read only
> 
> This is better, because it's consistent across endpoints, and reads generally scale better. It's still worse in the worst case scenario though.

I'm strongly in favour of 2. Historical queries explicitly rely on node-local state, the caller is expected to repeat their query on the same node for any response, and should also get faster responses for _related queries_ from the same node. It should be possible to run them all on a backup, if you choose to.

I don't follow the comment about worst cases. I think each node, backup or primary, is uniformly likely to have ledger holes. So if you don't forward, you may get stuck on a node that has a hole, when you could have forwarded to a node without (but you could be forwarded and also have a hole). And if you do forward, you may forward from a node without a hole to a primary with a hole (and again, maybe both have a hole, and the forwarding complexity, with an additional heuristic, didn't gain you anything).

MahatiC:
> 2. Fix https://github.com/microsoft/CCF/issues/3872 and update most, if not all historical endpoints to be read only

Question - Wouldn't we still need to set forwarding to `sometimes` even when there is a read_only endpoint? If not, what happens when the there is a ledger hole on the current node?

tests/e2e_logging.py
```diff
@@ -774,6 +774,35 @@ def test_historical_query(network, args):
...
+        args,
+        copy_ledger=False,
+        target_node=network.find_any_backup(),
+        from_snapshot=True,
+    )
+    network.trust_node(new_node, args)
+
+    LOG.info("Check historical queries return ACCEPTED")
+    with new_node.client("user0") as c:
+        for seqno in [2, 50, 100]:
```
eddyashton, line 796: This test won't fit nicely into the suites, because of these hardcoded seqnos and assumption they're in term 2.

I think we can use `/node/commit` + `/node/tx` to sample valid TxIDs, and ask for those. And while we're here we could confirm the node can still distinguish future, unknown TxIDs from known-but-in-a-hole.

(After dicsussion, this is deliberately _not_ in the suite, so not needed now)

achamayou:
> > 2. Fix [Support `make_read_only_endpoint` to be used with historical query adapter #3872](https://github.com/microsoft/CCF/issues/3872) and update most, if not all historical endpoints to be read only
> > 
> > This is better, because it's consistent across endpoints, and reads generally scale better. It's still worse in the worst case scenario though.
> 
> I'm strongly in favour of 2. Historical queries explicitly rely on node-local state, the caller is expected to repeat their query on the same node for any response, and should also get faster responses for _related queries_ from the same node. It should be possible to run them all on a backup, if you choose to.
> 
> I don't follow the comment about worst cases. I think each node, backup or primary, is uniformly likely to have ledger holes. So if you don't forward, you may get stuck on a node that has a hole, when you could have forwarded to a node without (but you could be forwarded and also have a hole). And if you do forward, you may forward from a node without a hole to a primary with a hole (and again, maybe both have a hole, and the forwarding complexity, with an additional heuristic, didn't gain you anything).

I'm also in favour of 2. and going to do that.

As discussed, the comment about worst case wasn't assuming uniform holes, instead it was made with the initial gap before backfill in mind. Anyway, we still want to scale historical reads, and there are other ways to mitigate this backfill period as discussed (such as waiting for it to happen before adding the node to the SLB).

achamayou:
> > 2. Fix [Support `make_read_only_endpoint` to be used with historical query adapter #3872](https://github.com/microsoft/CCF/issues/3872) and update most, if not all historical endpoints to be read only
> 
> Question - Wouldn't we still need to set forwarding to `sometimes` even when there is a read_only endpoint? If not, what happens when the there is a ledger hole on the current node?

Absolutely, and that's what make_read_only_endpoint does. It only helps with session consistency though, _not_ with holes in the current node.

tests/e2e_logging.py
achamayou: It won't indeed, and needs several refinements at least:

1. retire the node that's been added
2. determine the view dynamically
3. find out the snapshot txid, and ask for transactions in that range, rather than assume 100+

With that said, I think this basic test in e2e_logging is already reliable, and helped us spot the issue with forwarding.

achamayou: All done now.

tests/infra/node.py
```diff
@@ -511,6 +515,31 @@ def get_ledger(self):
 
         return current_ledger_dir, [committed_ledger_dir]
 
+    def most_recent_read_only_snapshot_seqno(self):
+        return self.remote.most_recent_read_only_snapshot_seqno()
+
+    def hide_committed_ledger_chunk(self, index):
```
achamayou, line 521: will remove

tests/infra/node.py
```diff
@@ -511,6 +515,31 @@ def get_ledger(self):
...
+
+        # Pick a random chunk, near the start of the ledger, to avoid the cache
+        (chunk_to_hide,) = random.sample(ledger_chunks[:5], 1)
+        shutil.move(
+            chunk_to_hide, os.path.join(tempfile.gettempdir(), chunk_to_hide.name)
+        )
+
+        return chunk_to_hide, interval(chunk_to_hide)
+
+    def unhide_committed_ledger_chunk(self, previously_hidden_chunk):
```
achamayou, line 537: will remove

src/endpoints/common_endpoint_registry.cpp
```diff
@@ -241,6 +241,7 @@ namespace ccf
       ccf::historical::adapter_v3(
         get_receipt, context, is_tx_committed, txid_from_query_string),
       no_auth_required)
+      .set_forwarding_required(endpoints::ForwardingRequired::Sometimes)
```
achamayou, line 244: Should be removed, along with schema impact, before merging.

eddyashton:
- [ ] Add documentation describing this behaviour

doc/operations/ledger_snapshot.rst
```diff
@@ -79,7 +79,9 @@ It is important to note that new nodes cannot resume from a snapshot and join a
...
+Nodes that started from a snapshot can still process historical queries if the historical ledger files (i.e. the ledger files preceding the snapshot) are made accessible to the node via the ``ledger.read_only_directories`` option to ``cchost``. Although the read-only ledger directory must be specified to the node on start-up, the historical ledger file contents can be copied to this directory `after` the node is started (see :ref:`operations/data_persistence:Data Persistence`).
+
+Before these ledger files are present the node will be functional, participating in consensus and able to accept new transactions, but historical queries targeting the missing entries will permanently stall. Calls to the historical query APIs will return loading responses, as these APIs do not currently distinguish between temporarily missing and permanently missing files. It is the responsibility of the operator to ensure that the ledger files visible to all nodes are complete, including back-filling missing entries when required.
```
MahatiC, line 84:
```suggestion
Before these ledger files are present the node will be functional, participating in consensus and able to accept new transactions, but historical queries targeting the missing entries will permanently stall. Calls to the historical query APIs will return loading responses, as these APIs do not currently distinguish between temporarily missing and permanently missing files. It is the responsibility of the operator to ensure that the ledger files visible to all nodes are complete, including back-filling missing files when required.
```

MahatiC: Just to be consistent.

MahatiC (approved):

tests/e2e_logging.py
```diff
@@ -774,6 +774,67 @@ def test_historical_query(network, args):
...
+        copy_ledger=False,
+        target_node=network.find_any_backup(),
+        from_snapshot=True,
+    )
+    network.trust_node(new_node, args)
+
+    # Pick some sequence numbers before the snapshot the new node started from, and for which
+    # the new node does not have corresponding ledger chunks
+    missing_seqnos = random.sample(
+        range(0, new_node.most_recent_read_only_snapshot_seqno()), 5
```
jumaffre, line 798: The `GET /node/state` endpoint returns `startup_seqno`, which is the seqno of the snapshot the node started from (`0` if no snapshot). I think this could be used here instead.

jumaffre (commented):
I think that this case is already covered by `test_add_node_from_snapshot` in `reconfiguration.py`, although maybe with less precision around the error codes/messages. Is it worth tweaking the existing test instead?

tests/e2e_logging.py
eddyashton: Nice, I've done this and removed the redundant functions.

eddyashton:
> I think that this case is already covered by `test_add_node_from_snapshot` in `reconfiguration.py`, although maybe with less precision around the error codes/messages. Is it worth tweaking the existing test instead?

True, I hadn't seen how similar those were. I've rolled these additional tests into that, and they now run there `if not copy_ledger:`.

PR #4151, opened 2022-08-18
Update docs for activating a member

lynshi:
Minor changes after some customer interactions.
1. Use a JSON file to save and send the request digest (eliminates manual copying for convenience).
2. `--data-binary '{"state_digest": <...>}'` seems to work inconsistently, maybe because the content length is slightly different based on the system...?
3. Add note on how to verify activation of the member.


achamayou (approved):
@lynshi thank you, using a file is probably a little tidier and more realistic. Spelling out how to read members is also a good idea.

eddyashton (approved):

PR #4162, opened 2022-08-19
Add read_only_adapter_v3

### LINKED ISSUE #3872: Support `make_read_only_endpoint` to be used with historical query adapter

achamayou:
Resolve #3872 

- [x] add read_only_adapter_v3
- [x] convert all relevant endpoints to use it


eddyashton:
I've done the second part of this (converting existing historical endpoints to use `read_only_adapter_v3`). We don't do this for historical queries in JS, because we find out that they're historical too late (ie - every JS request is may-write as far as dispatch is concerned, only later do we decide it was read-only/historical).

src/node/historical_queries_adapter.cpp
```diff
@@ -397,18 +398,19 @@ namespace ccf::historical
     return true;
   }
 
-  ccf::endpoints::EndpointFunction adapter_v3(
-    const HandleHistoricalQuery& f,
+  template <class HQY, class EFN, class CTX, class EXT>
```
MahatiC, line 401: Perhaps add a small comment indicating what these template abbreviations are? Something like: `EFN = EndpointFunction, CTX=EndpointContext, EXT=Extractor` ? (Not sure what `HQY` is)

MahatiC: Better yet, name the template parameters as `EndpointFunctionT`, `EndpointContextT`, `ExtractorT` etc ending them with letter `T` to indicate they are template parameters but they also the names give context as to what they represent.

MahatiC: Better yet, name the template parameters as EndpointFunctionT, EndpointContextT, ExtractorT etc ending them with letter T to indicate they are template parameters but they also the names give context as to what they represent.

eddyashton: Good suggestion - I've renamed these template parameters for clarity.

MahatiC (approved):

eddyashton:
Admin-merging to skip unrelated Daily failure

PR #4167, opened 2022-08-24
Improved OpenSSL keygen usage

ihaller:
The current code is functionally correct, but the is a minor API misuse that makes it brittle against OpenSSL changes.
- According to OpenSSL documentation (https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_keygen_init.html), error is signaled on these APIs with 0 or lower return value. Missing to detect an error here results in a very difficult to understand error later during usage.
- `paramgen` and `keygen` are separate operations, changing from one to the other (by calling the init of the second one) could in fact erase previously set parameters (such as the curve id). This does not happen for EC in OpenSSL 1.1.1, but will in OpenSSL 3.x.
- `EVP_PKEY_CTX_set_ec_paramgen_curve_nid` and `EVP_PKEY_CTX_set_ec_param_enc` are slightly poorly named as they apply both to paramgen and keygen operations.
- If people preferred that we still do full parameter generation, then the correct way to do it is to use the `paramgen` operation to generate a set of parameters that the `keygen` operation uses as input (https://wiki.openssl.org/index.php/Elliptic_Curve_Diffie_Hellman). Happy to write up that variant if that is feedback.
- As far as I am aware (s3_lib within OpenSSL uses this pattern, also the documentation at https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_CTX_set_ec_paramgen_curve_nid.html alludes to this), when using a named curve, it is not necessary to go through the parameter generation process and the `keygen` can be used directly with the curve name, but I did not find an official online write-up for this.


eddyashton (approved):
LGTM!

@ihaller TYVM for the detailed write up.

PR #4168, opened 2022-08-24
Fix lts_compatibility test

MahatiC:
... by reverting openapi_schema changes so that it can test with old nodes
successfully.


MahatiC:
Force merging this to fix daily build.

PR #4169, opened 2022-08-24
Add protobuf definition of interface between CCF and executors

### LINKED ISSUE #4153: Multi-Container: Define interface between CCF node and Executor
Eventually this will be gRPC. As an initial implementation, we may build a similar interface communicating HTTP.

The initial interface is about making sure we cover all of the necessary concepts, and coming up with common terminology. We may make additional changes before the final release.

As a minimum, this needs to cover:
- [Executor -> CCF] Registration of a new executor instance, with a fresh attestation
- [CCF -> Executor] Contents of a client request
- [Executor -> CCF] KV get
- [Executor -> CCF] KV put
- [Executor -> CCF] KV remove
- [Executor -> CCF] KV has
- [Executor -> CCF] Response to be returned to client
 
Additional features to consider, in increasing difficulty:
- [Executor -> CCF] KV foreach
- [Executor -> CCF] Close/Deregistration of executor instance
- [Executor <-> CCF] Error reporting, timeouts
- [Executor -> CCF] Fetching historical state

eddyashton:
Resolves #4153.

Naming is still up in the air - I've called this `remote_executor` for now.


MahatiC:
> Naming is still up in the air - I've called this `remote_executor` for now.

`remote_app` maybe? I'll write here when I think of some more.

src/apps/external_executor/protobuf/executor_registration.proto
```diff
@@ -0,0 +1,89 @@
+syntax = "proto3";
+
+package ccf;
+
+option optimize_for = LITE_RUNTIME;
```
jumaffre, line 5: Do we have a rough idea of the performance impact of this setting?

jumaffre (approved):

src/apps/external_executor/protobuf/executor_registration.proto
eddyashton: None at all. Not particularly attached to it, will likely replace it at least on the KV interface at some point.

Issue #4096, opened 2022-08-03
Limit access to privileged tables in app space


eddyashton:
We currently have a distinction between different kind of internal CCF governance tables, and application tables (applications can read governance but not write it, even governance can't write internal system tables like signatures), but we make no distinction between application tables.

The historical view on this was that they were all in the same trust domain, so any attempt at sandboxing could be circumvented (ie - if you're writing a native app, you can always poke around in memory and find secret state you weren't supposed to see). This isn't true in scripting runtimes, for example in JS we can intercept all KV accesses and reliably prevent writes to governance tables. There is a request to offer the same protection in application tables, so we can support large apps with multiple privilege rings.

I think this would be implemented in the form of a per-endpoint allowlist/blocklist of tables, that would be enforced by any sandboxing layer (ie - JS interpreter, remote BL dispatcher, etc). It may also be interesting to offer a mid-transaction kill-switch (ie "from now on, don't let me access _these_ tables"), to support offloading work to less privileged code within transaction execution.

achamayou:
This is effectively #6240, except for the mid-transaction kill switch. We can revisit that if a use case for it arises.

PR #4174, opened 2022-08-25
Governance actions for executor code ids

### LINKED ISSUE #4154: Multi-Container: Governance actions for controlling Executor Code IDs
Similar to the existing code ID tables controlling acceptable CCF nodes, we need governance for controlling acceptable executor nodes. We should consider how we make this extensible (for eg adding per-code ID restrictions on which URIs are served, or restricting which tables are accessible).

MahatiC:
Resolves #4154 

Currently this change adds code for the governance of executor code ids

TODO:
~~- [ ] Add quote verification~~
~~- [ ] Add a test~~
~~- [ ] Extend the metadata for executor_code_id table~~

P.S. pushing this as a draft for early review comments

The TODOs either can be a separate PR or depend on AMD work which is WIP


MahatiC:
CI failure is due to the schema change (will update schema later)

MahatiC:
After discussions with Eddy, most of the changes in this PR will be moved into the app itself (and will be self-contained) once we have https://github.com/microsoft/CCF/pull/4173 working

include/ccf/service/tables/executor_code_id.h
```diff
@@ -0,0 +1,43 @@
...
+    {
+      code_info.status = j;
+      code_info.platform = QuoteFormat::amd_sev_snp_v1;
+    }
+    else
+    {
+      code_info.status = j["status"];
+      code_info.platform = j["platform"];
+    }
+  }
```
jumaffre, line 42: We should be able to use the `DECLARE_JSON...` helper macros here instead as we don't need to deal with backwards compatibility here as this is a new ledger table.

samples/constitutions/default/actions.js
```diff
@@ -914,6 +914,19 @@ const actions = new Map([
...
   ],
+  [
+    "add_executor_node_code",
+    new Action(
+      function (args) {
+        checkType(args.executor_code_id, "string", "executor_code_id");
+      },
+      function (args) {
+        const codeId = ccf.strToBuf(args.executor_code_id);
+        const ALLOWED = ccf.jsonCompatibleToBuf("AllowedToJoin");
```
jumaffre, line 925: `AllowedToJoin` makes sense but it is the same enum used for nodes to join the CCF service. For clarify, I think we should create a new `CodeStatus` enum and use `AllowedToExecute` or something similar to avoid ambiguity.

src/service/genesis_gen.h
```diff
@@ -420,6 +420,15 @@ namespace ccf
...
     }
 
+    void trust_executor_code_id(
+      const CodeDigest& executor_code_id, const QuoteFormat& platform)
+    {
+      auto executor_codeid = tx.rw(tables.executor_code_ids);
+      executor_codeid->put(
+        executor_code_id,
+        {.status = CodeStatus::ALLOWED_TO_JOIN, .platform = platform});
+    }
```
jumaffre, line 430: This shouldn't be necessary as all modifications to this new table should be done in JS, via governance.

include/ccf/service/tables/executor_code_id.h
```diff
@@ -0,0 +1,43 @@
...
+#include "ccf/service/code_digest.h"
+#include "ccf/service/map.h"
+
+namespace ccf
+{
+  struct ExecutorCodeInfo
+  {
+    CodeStatus status;
+    QuoteFormat platform;
+    // FIXLater: Add more metadata that includes allowed URIS for each executor
```
achamayou, line 15: What are the allowed URIs for each executor? I'm not sure why we want to have them, considering we expect the executors to be clients and to connect to us.

Meta: we try to not to have TODO/FIXME etc comments in the CCF project, and instead we prefer to file tickets, and TODO is one template for them. We didn't always do it that way, but we observed that TODO comments were not always getting addressed or prioritized in a timely manner.

src/service/genesis_gen.h
MahatiC: Done

samples/constitutions/default/actions.js
MahatiC: Done

include/ccf/service/tables/executor_code_id.h
MahatiC: Done

include/ccf/service/tables/executor_code_id.h
eddyashton: @achamayou There's 2 kinds of metadata I think we'll want to set here - what URIs is this CodeID allowed to handle, and what tables is this CodeID allowed to access. When the CodeID completely describes the executor's behaviour (ie - it's compiled and has no dynamically loaded code), these are unnecessary but maybe useful (you don't _need_ to do a full code audit, if you trust this new executor is limited to only accessing its "own" tables). It's possible we'll also want to support executors that dynamically load their code in the same style as `js_generic`, in which case the CodeID tells us nothing about their behaviour. I think that limiting the tables and URIs the executor can own is then a useful restriction.

Simpler answer: If everything becomes a remote-executor app, then this is a nice place to implement #4096.

MahatiC: The `executor` here refers to the business logic containers. The allowed URIs will let BL containers register themselves for specific URIs.

> we try to not to have TODO/FIXME etc comments in the CCF project, and instead we prefer to file tickets,

Ah got it! Though the TODO here was more of a note to the reviewers since this is an evolving change for the demo. Removed it though, will probably put it into a ticket.

samples/constitutions/default/actions.js
MahatiC: I reverted this change for a couple of reasons a) This means the API schema will change and we don't really want this in 3.x and b) Even though the enum is same, since it sits in a different table it's distinguishable

src/apps/external_executor/executor_code_id.h
```diff
@@ -0,0 +1,46 @@
...
+{
+  ccf::CodeStatus status;
+  ccf::QuoteFormat platform;
+};
+
+DECLARE_JSON_TYPE(ExecutorCodeInfo);
+DECLARE_JSON_REQUIRED_FIELDS(ExecutorCodeInfo, status, platform);
+
+using ExecutorCodeIDs = ccf::ServiceMap<ccf::CodeDigest, ExecutorCodeInfo>;
+namespace Tables
```
eddyashton, line 42: This should be `namespace tables` (lowercase), and the namespace could be removed entirely.

The existing `ccf::Tables` namespace breaks this convention (namespace names are lowercase), because it was originally a single class with a bunch of static members, and we wanted to avoid a breaking change when that was sharded into a namespace.

eddyashton (approved):

src/apps/external_executor/executor_code_id.h
MahatiC: Ah okay. Yeah I did think of removing the namespace itself, done now. Thanks.

samples/constitutions/default/actions.js
eddyashton: I think we can go one step further, and usually a totally custom enum/string for this value (since, as you say, it exists in a separate table). So creating an `ExecutorCodeStatus` enum, with a single value `AllowedToExecute`, only visible to this application, and using that in place of `ccf::CodeStatus`. That means if `CodeStatus` ever gets new values (eg - `AllowedToJoinAsLearner` or something), we don't ever accidentally consider that a valid status for an Executor.

MahatiC: Sounds reasonable, done.

eddyashton (approved):
LGTM!

achamayou (approved):

PR #4181, opened 2022-08-30
Add `libprotobuf` (v21.5) as 3rdparty

jumaffre:
This PR adds `libprotobuf` as an internal third-party dependency. 

Releases are taken from here (see [protoc-21.5-linux-x86_64.zip](https://github.com/protocolbuffers/protobuf/releases) asset) and the extracted source tree is trimmed to remove unwanted files (e.g. `protoc`-specific code, which is installed separately - see #4178) are removed.

Note that `libprotobuf` is always built (~20 secs to build both enclave and virtual variants in total) but not linked in to CCF (it will be in the external executor app).


cmake/protobuf.cmake
```diff
@@ -0,0 +1,36 @@
...
+get_target_property(LIBPROTOBUF_INCLUDE_DIRS libprotobuf INCLUDE_DIRECTORIES)
+
+# enclave
+add_enclave_library(protobuf.enclave ${LIBPROTOBUF_SOURCES})
+target_include_directories(protobuf.enclave PUBLIC ${LIBPROTOBUF_INCLUDE_DIRS})
+target_compile_options(
+  protobuf.enclave PUBLIC "-Wno-deprecated-enum-enum-conversion"
+) # Remove warnings in generated_message_tctable_impl.h
+target_compile_options(protobuf.enclave PUBLIC "-Wno-invalid-noreturn"
+)# https://github.com/protocolbuffers/protobuf/issues/9817
```
eddyashton, line 27: Any reason these are specified separately, rather than in a single call? Would also be good to combine the settings for both `.virtual` and `.enclave` so they don't drift (eg `for target in (protobuf.enclave, protobuf.virtual): target_compile_options($target ...)`)

jumaffre: Good point. I've refactored this.

eddyashton (approved):

PR #4182, opened 2022-08-31
Send a timeout error to clients for forwarded requests

### LINKED ISSUE #486: CCF node should reply with a timeout error if a request takes too long to execute
In the case of `Write` requests executed directly on the leader or `Read` on any node, RPC requests execute and return synchronously to the caller. However, in the case of forwarded `Write` requests (or when using PBFT), requests could take a long time to execute or, in a failure case, not execute at all (e.g. if the forwardee node dies). 

As of now, the session will hang until the client decides to close it. To address this, we should make sure that a RPC session opened for too long in the enclave times out and returns a error message (e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) to the caller.

This could make use of https://github.com/microsoft/CCF/blob/master/src/node/timer.h.

eddyashton:
Resolves #486.

If a node forwards a request to another node, it was previously fire-and-forget - we _may_ get a corresponding response, but if we hear nothing we do nothing, and don't tell the client what we did. This PR adds a timeout, so if we do not hear a response within a fixed deadline (currently 3s for all requests), we will return a HTTP error to the client.

This was slightly trickier than expected because of the combination of pipelining support and forwarding. We previously had no way to track a specific forwarded request, but we need this so that when we forward multiple pipelined requests, we report the correct result/error for each of these. I'm now generating this as a per-node ID. Note this is still technically incomplete from a session consistency standpoint - if we forward A, then B, then C, and the response for B is dropped, we may return to the client `[A_response, C_response, B_timeout]`. I've also not worked out how we support this breaking change to the `ForwardedHeader` struct across LTS releases.

- [x] Work out `lts_compat`.


eddyashton:
I have now updated this to use explicit new message types (`forwarded_[cmd|response]_v2`) rather than adding a field to the existing message. This depends on #4185 being merged and backported first, which can _understand_ those messages but will never produce them. Then this change can go onto main, which will also _produce_ those messages, and includes a test that timed out forwarding results in a HTTP error.

achamayou (approved):

Issue #4160, opened 2022-08-19
Multi-Container: Replace HTTP with gRPC


eddyashton:
I believe at the moment it will be simpler to implement a working multi-container prototype using the existing HTTP transport, with the entire system as a CCF app. Eventually, we want to replace that with gRPC transport, and benefit from more flexible (async) execution. This may be integrated earlier than I expect, but I'm currently treating this as a single porting step once some demo implementation exists.

eddyashton:
Thought about this a little more, and realised it may be a relatively small change.

We're planning for the initial demo to be HTTP/1.1, with a JSON payload. When we say "replace with gRPC", that's really just HTTP/2, with a protobuf payload. We can use the same URLs for both, making the switch-over easier.

The additional piece beyond that, is how/whether we can provide better support for gRPC features than we can for a purely call-and-response based HTTP/1.1 system. ie - providing streamed responses, and long-lived blocking execution of RPC calls (where we currently require polling).

eddyashton:
Dumping some more thoughts and code here so I don't lose them.

The step from HTTP/1.1 to HTTP/2 is actually done already too - for demo purposes we can start a CCF service with `--http2`, and then use HTTP/2 clients. (When ALPN/merged endpoints from #3342 are done, we won't even need to manually turn this on - it will always be supported).

So the open question is "can we/should we build a shim to get a gRPC client talking JSON rather than protobuf?". I've built a hacky little thing that handles the serialization layer cheaply*, but it looks like we'd still need an interceptor to deal with the headers (because if CCF thinks it's speaking JSON, it inserts a "content-type: application/json" header in the response, and then the gRPC client blows up: `E0825 09:32:26.304294001   35396 hpack_parser.cc:1234]       Error parsing metadata: error=invalid value key=content-type value=application/json`).

In summary, I think we should just bite the bullet and go for full gRPC + protobuf-body support for the demo, as there's not that much work left to do here. I think we'll still be missing an async-response API, but that's fine for now.

<details>
  <summary>* Here's a bit of Python hackery that will dynamically generate a thing that looks like a `FooStub` from a `Foo` service definition, but using Python serialization:</summary>

```
from google.protobuf.json_format import MessageToJson, Parse

# Generates a class that looks like a Stub, but uses JSON serialization
def json_stub_class(module, service_name):
    service_descriptor = module.DESCRIPTOR.services_by_name[service_name]
    name = f"{service_name}Stub_JSON"

    def to_bytes(fn):
[...]

PR #4183, opened 2022-08-31
Add basic gRPC support to external executor app

jumaffre:
Part of #4160 

This PR adds gRPC support to the external executor app and implements the `Get` and `Put` endpoints defined in `kv.proto`. 

TODO: 
- [x] Generate Python source files from proto files automatically


src/apps/external_executor/grpc.h
```diff
@@ -0,0 +1,143 @@
...
+// Licensed under the MIT license.
+
+#include "ccf/odata_error.h"
+#include "ds/serialized.h"
+
+#include <arpa/inet.h>
+#include <variant>
+#include <vector>
+
+namespace grpc
```
eddyashton, line 11:
```suggestion
namespace ccf::grpc
```

src/apps/external_executor/grpc.h
```diff
@@ -0,0 +1,143 @@
...
+
+  MessageLength read_message_frame(const uint8_t*& data, size_t& size)
+  {
+    auto compressed_flag = serialized::read<CompressedFlag>(data, size);
+    if (compressed_flag > 1)
+    {
+      throw std::logic_error(fmt::format(
+        "gRPC compressed flag has unexpected value {}", compressed_flag));
+    }
+    return ntohl(serialized::read<MessageLength>(data, size));
```
eddyashton, line 27: We're not storing the `compressed_flag` anywhere, so I think we're actually expecting that it's always `0`? If it is ever `1`, then the contents are encoded based on the `Message-Encoding` header, and I don't think we handle that case anywhere? Fine if we don't for now, but we should make this error more precise (`>1` => unexpected value, `!=0` => unhandled value).

src/apps/external_executor/protobuf/build.sh
```diff
@@ -2,28 +2,27 @@
...
+GENERATED_DIR=${2}
 
 if [ ! -f "env/bin/activate" ]
     then
         python3.8 -m venv env
 fi
 
 source env/bin/activate
-pip install -U -r "${THIS_DIR}/requirements.txt"
+pip install -U -r "${THIS_DIR}/requirements.txt" > /dev/null
```
eddyashton, line 19:
```suggestion
pip install -q -U -r "${THIS_DIR}/requirements.txt"
```

I think this is similar but safer.

src/apps/external_executor/grpc.h
jumaffre: The intent was to throw if the value is unexpected or compression is enabled but I clearly can't write a comparison operator.  Fixed now (`>= 1`).

src/apps/external_executor/external_executor.cpp
```diff
@@ -44,32 +30,49 @@ namespace externalexecutor
 
       install_kv_service();
 
-      auto do_echo = [this](ccf::endpoints::EndpointContext& ctx) {
-        CCF_APP_INFO("ECHO HANDLER BEGIN");
+      auto put = [this](
```
eddyashton, line 33: Please move these to inside `install_kv_service()`.

src/apps/external_executor/grpc.h
```diff
@@ -0,0 +1,143 @@
...
+  static constexpr size_t message_frame_length =
+    sizeof(CompressedFlag) + sizeof(MessageLength);
+
+  MessageLength read_message_frame(const uint8_t*& data, size_t& size)
+  {
+    auto compressed_flag = serialized::read<CompressedFlag>(data, size);
+    if (compressed_flag >= 1)
+    {
+      throw std::logic_error(fmt::format(
+        "gRPC compressed flag has unexpected value {}", compressed_flag));
```
eddyashton, line 25: Still think this flag should be clearer so we have an idea to fix when it's hit - "... unexpected value {} - currently only support unencoded gRPC payloads".

src/apps/external_executor/grpc.h
```diff
@@ -0,0 +1,143 @@
...
+  In get_grpc_payload(const std::shared_ptr<ccf::RpcContext>& ctx)
+  {
+    auto& request_body = ctx->get_request_body();
+    auto request_content_type =
+      ctx->get_request_header(http::headers::CONTENT_TYPE);
+
+    auto data = request_body.data();
+    auto size = request_body.size();
+
+    // Only remove gRPC frame is content type is gRPC so that it is possible to
```
eddyashton, line 47:
```suggestion
    // Only remove gRPC frame if content type is gRPC so that it is possible to
```

src/apps/external_executor/grpc.h
```diff
@@ -0,0 +1,143 @@
...
+      ctx->get_request_header(http::headers::CONTENT_TYPE);
+    if (request_content_type == http::headervalues::contenttype::GRPC)
+    {
+      r_size += grpc::message_frame_length;
+      r.resize(r_size);
+      r_data = r.data();
+      grpc::write_message_frame(r_data, r_size, resp.ByteSizeLong());
+      ctx->set_response_header(
+        http::headers::CONTENT_TYPE, http::headervalues::contenttype::GRPC);
+    }
```
eddyashton, line 90: I'd expect a `415 Unsupported Media Type` error if this header is not `application/grpc`, as we do in the JSON handler, but I can't see that anywhere. If we want to maintain a debug path where we can send raw protobuf by hand, it should probably insist on _no_ content type, or some other dedicated debug content type.

jumaffre: I'm no longer sure keeping the non-gRPC (i.e. raw protobuf) path is actually helpful for debugging now that unary gRPC works so I've removed it and now mandates that the `content-type` is gRPC (`415` is returned otherwise).

MahatiC (approved):

PR #4184, opened 2022-09-01
Rename `sgxinfo.sh` to `platform_info.sh` and cater for AMD SEV-SNP

jumaffre:
This simple script now displays if the current platform is SEV-SNP enabled and returns early if that's the case.


DomAyre (approved):

eddyashton (approved):

achamayou:
@jumaffre I'm not sure about this, most of the output is very SGX specific, and this is going to break whoever depended on the old name. It seems preferable to create an `snpinfo.sh` and to leave `sgxinfo.sh` as it, both for new and existing users. Am I missing something?

PR #4185, opened 2022-09-01
Parse new versions of forwarded request

eddyashton:
Subset of #4182 required for compatibility. This knows how to parse and use the new message types, but will never produce them itself. This should be backported to 2.x.

Related to #486.


src/node/rpc/forwarder.h
```diff
@@ -80,24 +146,47 @@ namespace ccf
       }
       serialized::write(data_, size_, raw_request.data(), raw_request.size());
 
-      ForwardedHeader msg = {
-        ForwardedMsg::forwarded_cmd, rpc_ctx->frame_format()};
+#if false
```
achamayou, line 149: `?>_>`

eddyashton: Minimising the diff between this PR, and the one that actually enables the behaviour

MahatiC: Seems to be a small diff anyway? Either way, we probably shouldn't resort to `if false` :)

eddyashton: Sure, I've removed this block.

src/node/rpc/forwarder.h
```diff
@@ -243,19 +323,84 @@ namespace ccf
                 return;
```
MahatiC, line 323: Can we refactor out the error handling part so that the switch case for `ForwardedMsg::forwarded_cmd_v1:` and `ForwardedMsg::forwarded_cmd_v2:` is better handled? Something like this[0]? (P.S. the diff example is not properly formatted/syntaxed)

[0]https://gist.github.com/MahatiC/2c29ab357ea3f52cf7f1703f4dd87486

eddyashton: We could do something like this. The common function would have to return the calculated `fwd_handler`, with nullptr for errors.

I think this leads to more duplicated code though - both cases do the same setup and call+handling of this new helper function, and the blocks leading to `send_forwarded_response` end up further apart so it's less obvious that they're extremely similar. What's the advantage to this refactor?

MahatiC: > The common function would have to return the calculated `fwd_handler`, with nullptr for errors.

Correct, or `fwd_handler` can be passed as a reference for `handle_errors` (and this function will populate it) and if it's a nullptr, `handle_errors`  will return false and then we exit (or return) from the original call point

> I think this leads to more duplicated code though

The code is small and it differentiates cleanly that we're calling two different versions of the forwarded command

> What's the advantage to this refactor?

Using multiple `if else` blocks for the same purpose that the switch case was intended for defeats the point of a switch case and makes it less readable

eddyashton: Sure, I've made this change so we're no longer combining switch cases.

MahatiC (approved):
Looks good, thank you!

achamayou (approved):

PR #4196, opened 2022-09-05
Add `snpinfo.sh` script

jumaffre:
Revert some of the changes done in #4184 so that we have two different scripts, one for SGX and one for SEV-SNP.

### DIFF
--- a/.azure-pipelines-templates/cmake.yml
+++ b/.azure-pipelines-templates/cmake.yml
@@ -4,15 +4,6 @@ parameters:
 steps:
   - script: |
       set -ex
-      samples/scripts/platform_info.sh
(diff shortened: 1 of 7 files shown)


achamayou (approved):

DomAyre:
Our CI is still calling platform_info.sh, I'm not personally against having this script which just checks platform and calls the appropriate script?

achamayou:
@DomAyre the CI is already specialised based on the environment, so we could also conditionally run the right script for the right environment.

jumaffre:
I've reorganised the CI YAML files so that we call the right info script based on the target (and only run the magic `sgx_prv` group commands on SGX).

PR #4198, opened 2022-09-06
Support for `.proto` files with `clang-format`

jumaffre:
It turns out that `clang-format` can format `.proto` files so use that, with the `Microsoft` style.


src/apps/external_executor/echo.proto
```diff
@@ -1,16 +0,0 @@
-syntax = "proto3";
```
jumaffre, line 1: This had been unused since #4183

.clang-format
```diff
@@ -128,4 +128,7 @@ StatementMacros:
 TabWidth:        2
 UseCRLF:         false
 UseTab:          Never
+---
+Language: Proto
+BasedOnStyle: Microsoft
```
eddyashton, line 133:
```suggestion
BasedOnStyle: Microsoft
TabWidth: 2
```

For consistency?

jumaffre: `IndentWidth: 2` did the trick!

eddyashton (approved):

MahatiC (approved):

PR #4201, opened 2022-09-07
Fix SAN failure in Daily build

eddyashton:
The Instrumented build has detected a container overflow since #4183, which added the external executor test.

We can't work out what's different about this build/app/run, or what's unusual about this bit of code that causes the error. I'm leaning towards it being a false positive, but can't be sure.

The fix here is to refactor the code to avoid the alert entirely.

### DIFF
--- a/.daily_canary
+++ b/.daily_canary
@@ -1,1 +1,1 @@
-Canary.
+Canary!
--- a/src/tls/ca.h
+++ b/src/tls/ca.h
@@ -17,29 +17,34 @@ namespace tls
     std::vector<Unique_X509> cas;
     bool partial_ok = false;
 
-  public:
-    CA(const std::string& ca_ = "", bool partial_ok = false) :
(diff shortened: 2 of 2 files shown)


achamayou (approved):

MahatiC (approved):
Have we found out where in the code `CA` constructor is being called by default values? (I couldn't find any)

eddyashton:
> Have we found out where in the code `CA` constructor is being called by default values? (I couldn't find any)
> 
> Edit: I think the answer is no, just looked at your edited description

I don't think there any instances calling the constructors with the default values, as they would be ambiguous (is `CA c;` supposed to call the `std::string` or `std::vector` constructor?). That's why I've also removed the default values.

MahatiC:
> (is `CA c;` supposed to call the `std::string` or `std::vector` constructor?)

Without a default constructor explicitly specified, it will result in an [error](https://onlinegdb.com/u3ynaqeTD)

PR #4202, opened 2022-09-07
Support for gRPC errors in external executor app

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/4160

This is the last piece missing to get a fully-fledge unary gRPC app. Similarly to JSON over HTTP/1, endpoints can now call `make_success()` and `make_error()` to return to the client. We also use the [rich error model](https://www.grpc.io/docs/guides/error/#richer-error-model) supported by most gRPC implementations to return meaningful errors to the client (similarly to [OData ](https://docs.microsoft.com/en-us/odata/) when using JSON). 

This PR does not add support for generic gRPC errors at the framework level. For example, if a gRPC request fails authentication, the gRPC will still receive a OData/JSON error rather than gRPC. This will require more work that I will ticket now.


src/apps/external_executor/external_executor.cpp
```diff
@@ -28,32 +28,35 @@ namespace externalexecutor
...
+            GRPC_STATUS_NOT_FOUND,
+            fmt::format("Key {} does not exist", payload.key()));
         }
 
         ccf::KVValue r;
         r.set_value(value.value());
 
-        ctx.rpc_ctx->set_response_status(HTTP_STATUS_OK);
-        return r;
+        return ccf::grpc::make_success(r);
```
jumaffre, line 59: This was the goal of the PR: Have a generic way to return payloads and errors to the client, similarly to JSON over HTTP/1.

src/apps/external_executor/external_executor.cpp
```diff
@@ -28,32 +28,35 @@ namespace externalexecutor
...
+        -> ccf::grpc::GrpcAdapterResponse<ccf::KVValue> {
         auto records_handle = ctx.tx.template ro<Map>(payload.table());
         auto value = records_handle->get(payload.key());
         if (!value.has_value())
         {
-          ctx.rpc_ctx->set_response_status(HTTP_STATUS_NOT_FOUND);
-          return ccf::KVValue(); // Handle errors
+          // Note: no need to specify `make_error<ccf::KVValue>` here as lambda
+          // returns `-> ccf::grpc::GrpcAdapterResponse<ccf::KVValue>`
+          return ccf::grpc::make_error(
```
jumaffre, line 51: As discussed with @eddyashton, we want to avoid having to specify `make_error<ccf::KVValue>` here as the error doesn't contain any `ccf::KVValue` although the handler is strongly typed (`ccf::grpc_read_only_adapter<ccf::KVKey, ccf::KVValue>(get)`). The hint on line `44` allows us to do that (along with some `std::variant>` shenanigans in `grpc.h`).

src/apps/external_executor/grpc_status.h
```diff
@@ -0,0 +1,87 @@
...
+
+#include "status.pb.h"
+
+#include <optional>
+#include <string>
+
+// Mapping to HTTP errors as per
+// https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto
+#define GRPC_STATUS_MAP(XX) \
+  XX(0, OK, "Ok", 200) \
```
jumaffre, line 13: The last value corresponds to the equivalent HTTP status but is not yet used. We can use it when we add generic support for gRPC errors at the framework level.

src/apps/external_executor/grpc_status.h
```diff
@@ -0,0 +1,87 @@
...
+  {
+    return static_cast<int32_t>(status);
+  }
+
+  Status make_grpc_status(
+    enum grpc_status status,
+    const std::optional<std::string>& msg = std::nullopt,
+    const std::optional<std::string>& details = std::nullopt)
+  {
+    // Note: details are not currently supported
```
jumaffre, line 64: The details are added to the `Status` object but not yet returned to the client. The gRPC Python client does not seem to support these so I'll try with another one to find out how to serialise them in the HTTP/2 trailers.

src/apps/external_executor/grpc_status.h
```diff
@@ -0,0 +1,87 @@
...
+  XX(5, NOT_FOUND, "Not Found", 404) \
+  XX(6, ALREADY_EXISTS, "Already Exists", 409) \
+  XX(7, PERMISSION_DENIED, "Permission Denied", 403) \
+  XX(8, RESOURCE_EXHAUSTED, "Resource Exhausted", 429) \
+  XX(9, FAILED_PRECONDITION, "Failed Precondition", 400) \
+  XX(10, ABORTED, "Aborted", 409) \
+  XX(11, OUT_OF_RANGE, "Out Of Range", 400) \
+  XX(12, UNIMPLEMENTED, "Unimplemented", 501) \
+  XX(13, INTERNAL, "Internal", 500) \
+  XX(14, Unavailable, "Unavailable", 503) \
```
achamayou, line 27:
```suggestion
  XX(14, UNAVAILABLE, "Unavailable", 503) \
```
probably?

achamayou (approved):

src/apps/external_executor/grpc.h
```diff
@@ -79,65 +137,60 @@ namespace ccf::grpc
...
   {
-    size_t r_size = resp.ByteSizeLong();
-    std::vector<uint8_t> r(r_size);
-
-    auto r_data = r.data();
-
-    auto request_content_type =
-      ctx->get_request_header(http::headers::CONTENT_TYPE);
-    if (request_content_type == http::headervalues::contenttype::GRPC)
+    auto r_ = std::get_if<SuccessResponse<Out>>(&r);
```
eddyashton, line 143: `r_` is not a great name, and reusing it for different types will bite us. Can we have this as `auto success_response = ...`, and line 170 as `auto error_reponse = ...`?

jumaffre: Indeed. I've renamed these.

eddyashton (approved):

MahatiC (approved):

PR #4204, opened 2022-09-08
Expose HMAC through the Public C++ Crypto API

andpiccione:
This PR exposes the HMAC cryptographic function through the Public C++ Crypto API by moving the corresponding `hmac.h` definition to `include/ccf`. Public docs have also been updated to include the addition of the function.


achamayou (approved):
Thank you!

Issue #4157, opened 2022-08-19
Multi-Container: Build a long-lived Tx context


eddyashton:
To actually implement the server-side API defined in #4153, CCF needs to persists a `Tx` across multiple incoming requests. This is awkward if we think of it as an app in the current HTTP model, so may require pushing some implementation up the stack. That in turn may require separating executor traffic from client traffic very early.

MahatiC:
@eddyashton looks like we can close this as we have tickets for other follow-ups?

eddyashton:
Yes, this should have been closed with #4218.

PR #4205, opened 2022-09-08
Refactor frontend to remove return-based response, and retry endpoint lookup on conflict

eddyashton:
Steps towards #4157. While doing this, I spotted another bug that's basically the same as #4102, so that's fixed here.

Details:
We previously relied on the return value from a call to `RpcFrontend::process()`, to work out whether a request was forwarded or not. If it's not, we assumed we could serialise the response immediately, and return that to be sent onwards. That's not true for HTTP/2, and makes it harder to use the forwarding for other things - we're restricted by both "every path in `process()` needs to return a serialised response", and "code which can't request an early-return can't indicate that they have forwarded". Luckily this is a pretty simple transformation. We already insert the details used to construct a response into the `RpcContext`, and were just calling `serialise_response()` on that in the return paths. Now we delay that by one step, and call it outside at the top of the stack, _after_ a call to `process()`, only on the HTTP/1 path.

The knock-on from this is updates to a bunch of test cases that want to call `process()` and then parse a serialised response - they now need 2 lines to do that rather than 1.

With that done, it becomes easier to move chunks of `process()/process_command()` around, so that's what I've done - factoring the standalone tasks into separate functions. This is a big diff, but the only functional change is that _everything_ relying on the `Tx` now happens inside the retry loop, so is retried on conflict (including `find_endpoint`, which _could_ change if the JS app has updated in the interim!). I'd like to make that even more explicit by constructing the `Tx` inside `process_command()`, rather than outside (which is a recipe for doing things that won't get re-executed on conflict), but need to work out how that interacts with `set_root_on_proposals()` - will do that in a future cleanup.


include/ccf/endpoint_registry.h
```diff
@@ -130,8 +130,8 @@ namespace ccf::endpoints
     ccf::pal::Mutex metrics_lock;
     std::map<std::string, std::map<std::string, Metrics>> metrics;
 
-    EndpointRegistry::Metrics& get_metrics_for_endpoint(
-      const EndpointDefinitionPtr& e);
+    EndpointRegistry::Metrics& get_metrics_for_request(
```
jumaffre, line 133: Is this an API change?

jumaffre (approved):

include/ccf/endpoint_registry.h
eddyashton: Ach, good spot! Hadn't realised this was public. This change is necessary so that we can increment the number of calls once, before endpoint lookup. I'll add a note to the changelog

eddyashton: I've mentioned this in the changelog, but the better fix would be to remove this (and more importantly the `increment_metrics_*` calls) from the public API. Tracked in #4207.

PR #4218, opened 2022-09-12
Move ownership of Tx to `unique_ptr`, and demonstrate extending ownership in `external_executor`

eddyashton:
Completes #4157.

The scary changes here that need careful audit are those in `frontend.h`, where we allow some extremely weird new behaviour by apps. Then `external_executor` demonstrates a simple useage of that, managing a single long-lived `Tx` object. In future, this `Tx` will be stolen/extended from a _user request_, rather than a `StartTx` call - this is just testing that the concept works as expected.

@jumaffre: I've also tweaked the `grpc` wrapper to remove the custom `EmptyResponse` struct, and instead use `google::protobuf::Empty`. This is more discoverable, specifically when trying to work out how to spell the return type/error type of something by looking at the `.proto` spec (we consistently use the C++ types generated from this file), but is maybe awkward for other services with their own empty placeholder type. Let me know what you think.


jumaffre:
> @jumaffre: I've also tweaked the grpc wrapper to remove the custom EmptyResponse struct, and instead use google::protobuf::Empty. This is more discoverable, specifically when trying to work out how to spell the return type/error type of something by looking at the .proto spec (we consistently use the C++ types generated from this file), but is maybe awkward for other services with their own empty placeholder type. Let me know what you think.

Nice! That's much better indeed.

src/apps/external_executor/external_executor.cpp
```diff
@@ -47,12 +48,99 @@ namespace externalexecutor
         .install();
     }
 
+    // Note: As a temporary solution for testing, this app stores a single Tx,
+    // stolen from a StartTx RPC rather than a real client request
```
jumaffre, line 52: It looks like the endpoint actually steals the client request, from `ctx`?

eddyashton: Terminology is fuzzy here, but `StartTx` is (eventually) from an executor, not a 'real client'. Currently, as a PoC, we're storing a single `Tx`, and stealing that from `StartTx`. Eventually, we'll be storing many `Tx`s, each stolen from a user request to any other endpoint (eg - a call to `POST /students/{id}`)

jumaffre (approved):

achamayou (approved):

Issue #4212, opened 2022-09-09
Allow making proposals service-specific, and illustrate it in a sample


achamayou:
It should be possible to include a pure validation-based action in a proposal, which would check the service identity matches a user-specified fingerprint. This would allow binding proposals to a particular service instance, where desired.

```
{
  "actions": [
    {
      "name": "assert_service_identity",
      "args": {
        "fingerprint": "ba9faac9683f7854c2cf0a97f57e63c260bf8d06f8183772c5655093c0af6e19"
      }
    }
  ]
}
```

Performing this check requires giving `validate()` access to the Key Value store, at the time the proposal is filed. This isn't enabled at the moment, but I can't think of any downsides to doing this. @eddyashton @jumaffre, your opinions are particularly welcome.

eddyashton:
Completed in #4220.

PR #4220, opened 2022-09-12
Add assert_service_identity

achamayou:
Solve #4212 by adding an `assert_service_identity()` action, which is a pure validation action, in a governance proposal. This action ensures that the proposal can only be applied on a specific service.

Interestingly, more subtle variants of that proposal may be desirable, where instead of an identity check, the validation checks that the service itself has been endorsed by another entity. This allows expressing "perform these governance actions on all services endorsed by this identity", as opposed to "on this service alone".

A minor change to the governance frontend was necessary to expose the store read-only to the `validate()` call in the constitution.


src/node/rpc/member_frontend.h
```diff
@@ -899,8 +899,9 @@ namespace ccf
         js::Runtime rt;
         js::Context context(rt, js::TxAccess::GOV_RO);
         rt.add_ccf_classdefs();
+        js::TxContext txctx{&ctx.tx};
         js::populate_global(
-          nullptr,
+          &txctx,
```
eddyashton, line 904: Writing this here so we have a record on the PR:

I was expecting we'd need to add more changes here, so that validation is only run once, when a proposal is first submitted. If it is run multiple times, and transactionally against new states of the KV, it becomes a lot easier to get validation errors during a later execution (ie - assert there are <=3 members, this is true when the proposal is submitted but not when the 2nd vote arrives), which is ugly as we can't report/handle these late validation errors nicely.

It looks like that change is unneeded, because the validation is _already_ run only once, when the proposal is created. We call `resolve_proposal()` in both `POST /proposals` and `POST /proposals/{id}/ballots`, but evaluate the `"validate"` function only in `POST /proposals`. @achamayou can you confirm that understanding?

eddyashton (approved):

tests/infra/consortium.py
```diff
@@ -473,6 +473,15 @@ def set_constitution(self, remote_node, constitution_paths):
         proposal = self.get_any_active_member().propose(remote_node, proposal_body)
         return self.vote_using_majority(remote_node, proposal, careful_vote)
 
+    def assert_service_identity(self, remote_node, service_cert_path):
+        service_cert_pem = slurp_file(service_cert_path)
+        proposal_body, careful_vote = self.make_proposal(
+            "assert_service_identity", service_identity=service_cert_pem
+        )
+        print(proposal_body)
```
jumaffre, line 481: Remove this line?

achamayou: Yes, thank you!

achamayou, line 481:
```suggestion
```

jumaffre (approved):

PR #4222, opened 2022-09-12
Fix operator resolve.js to not throw on undefined member data

lynshi:
In the sample `resolve.js` for operators, `isOperator` can throw an exception if a member is defined without member data as member data field access was unconditional (i.e. `info.member_data.is_operator` without checking if `info.member_data` is defined). Further, `isRecoveryMember` checks if the member has a public encryption key, but the public encryption key is not a field in member data.


eddyashton (approved):

achamayou (approved):
@lynshi thank you for fixing those, I think we need to expand our test coverage of the constitution samples.

PR #4224, opened 2022-09-13
Reduce number of nodes active in `reconfiguration_test_cft`

eddyashton:
We've had intermittent but regular failures in `reconfiguration_test_cft`, with a variety of callstacks/error messages. I believe the root cause of that is running too many nodes in parallel, and have included some other fixes along the way - comments inline.

Opening as a draft, with some changes to the ADO yml to stress this single test - must revert those before merging.
- [x] Revert CI changes.


tests/reconfiguration.py
```diff
@@ -723,6 +723,7 @@ def run_all(args):
         test_issue_fake_join(network, args)
 
         if args.consensus != "BFT":
+            test_add_as_many_pending_nodes(network, args)
```
eddyashton, line 726: This is the real fix, I think. Where we were previously running this, we'd added enough nodes to reach a 7-node network. This test case then adds 7 more, meaning the test box is maintaining 14 nodes at once. Where 2tx reconfig is enabled, and this test is executed by both in parallel, we could end up with 28 nodes on a single box.

That leads to each node running much slower than our timeouts can handle, producing spurious TLS handshake failures, Raft elections, and eventual timeouts in the top-level testing asserts.

tests/infra/network.py
```diff
@@ -1260,6 +1260,12 @@ def wait_for_snapshots_to_be_committed(src_dir, list_src_dir_func, timeout=20):
                         )
                         return True
 
+                if time.time() > end_time:
+                    LOG.error(
+                        f"Could not find committed snapshot for seqno {target_seqno} after {timeout:.2f}s in {src_dir}: {list_src_dir_func(src_dir)}"
+                    )
+                    return False
+
```
eddyashton, line 1268: Restructuring this while loop to check the condition here, rather than at the start. The problem with the other form is that when the node is running slow, the calls to `update_state_digest` and `wait_for_commit` can take an extremely long time, and once they finish we don't actually check if a snapshot has been produced in the interim. This form ensures we always checking for a snapshot _after_ manually inducing one, even if that induction takes a long time and only permits us a single iteration.

tests/infra/clients.py
```diff
@@ -498,7 +498,7 @@ def request(
                 timeout=timeout,
                 content=request_body,
             )
-        except httpx.ReadTimeout as exc:
+        except httpx.TimeoutException as exc:
```
eddyashton, line 501: A maybe-risky change. See [here](https://www.python-httpx.org/exceptions/#the-exception-hierarchy) for httpx's exception hierarchy. This change means that other kinds of httpx timeout will also be converted to `TimeoutError`s for our infra to deal with. The motivation was seeing `ConnectTimeout`s when a TLS handshake failed (presumably because of a slow node?), which triggered an early test failure despite passing through multiple parts of the stack which were explicitly repeating/robust to delays.

eddyashton:
Evidence that this fixes `reconfiguration_test_cft` are the green runs running that test in a loop:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=49662&view=results
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=49661&view=results

Eagle-eyed viewers may notice those builds aren't actually 100% green, and suggest I'm trying to pull a fast one. But the failures are from pipeline steps that expected other tests (eg perf tests, tlstest) to run and produce artifacts, so are a safe side-effect of the yml hacks. The important tests, running `reconfiguration_test_cft` in a loop, passed in every build configuration.

jumaffre (approved):

tests/infra/clients.py
DomAyre: So what is the consequence of raising them as a timeout exception instead of a runtime error? Do we gracefully handle timeout exceptions somewhere else?

eddyashton: Yes - `TimeoutError` (our type) is caught and handled differently in many places in the code.

PR #4225, opened 2022-09-13
Fix for occasional failures in `recovery_test_cft`

eddyashton:
Example failure in CI here:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=49618&view=logs&j=b450cd14-f66d-5f5f-6465-8f7fbd24a6b4&s=96ac2280-8cb4-5df5-99de-dd2da759617d&t=e296d40c-72fd-5e58-bd34-e9658c399ec0&l=10129

The test fails when the `test_recover_service_truncated_ledger` function has completely truncated a file, and we later try to parse the resulting `Ledger`. The total truncation is an accident, when we fail to find a signature transaction. I've not been able to reproduce this consistently but believe it's a plausible race condition. We're looking only at a node's `.current` ledger directory; that may contain a single new chunk, with no signatures. The simple fix is to pause and retry when that happens, with a strong assumption that this is coming from a temporary race condition. I've also refactored the way we calculate/specify truncation points, and produced an earlier assertion that they're non-`None`.

- [x] Revert CI changes


jumaffre (approved):

tests/recovery.py
```diff
@@ -499,14 +481,46 @@ def run_corrupted_ledger(args):
...
+                for tx in chunk:
+                    if verbose:
+                        LOG.info(f"Considering tx {tx.get_tx_digest()}")
+                    yield chunk, tx
+
+        def corrupt_first_tx(ledger, verbose=False):
+            LOG.info("Finding first tx to corrupt")
+            for chunk, tx in all_txs(ledger, verbose):
+                return chunk.filename(), get_middle_tx_offset(tx)
+            return None, None
```
DomAyre, line 502: Rather than returning none and having the check would it be simpler to raise an exception here?

eddyashton:
Marking this for backport to 2.x, as we're seeing fairly frequent `recovery_test_cft` failures there which I _think_ match this.

Issue #4144, opened 2022-08-17
Allow user to configure how SEV-SNP endorsements are fetched


DomAyre:
_Originally posted by @achamayou in https://github.com/microsoft/CCF/pull/4106#discussion_r947720759_

TODO (gathered from comments below):
- [x] 1. Use in-enclave HTTPS client (#4226)
- [x] 2. Increase robustness of endorsement fetching (e.g. if request times out) (#4277)
- [x] 3. Add support for AMD endpoint (note: requires contacting 2 separate endpoints - see https://www.amd.com/system/files/TechDocs/57230.pdf) (#4277)
- [x] 4. Add node configuration entry to select between AMD and Azure endpoint (#4277)
- [ ] #4637
- [ ] ~6. Cache endorsement certificates~. As discussed with @lynshi, Azure DCAP client caching is unlikely to be used in production with current SGX deployments so we will not prioritise this yet for SEV-SNP, until proven otherwise).
- [ ] 7. Allow for retrieval of endorsements certificates and CRLs via THIM agent, made available to ACI via environment variables.

achamayou:
We want to default to the public AMD endpoint, because it will work anywhere, and allow the Azure endpoint to be configured where appropriate.

achamayou:
We also need to look at https://github.com/microsoft/CCF/pull/4106#discussion_r947721064 when we revisit the endorsement endpoints.

DomAyre:
We might need to swap out the client::RpcTlsClient with the enclave HTTP client so that it works with Monza

DomAyre:
We also want to be robust to failures fetching the endorsements (specifically too many requests)

achamayou:
@DomAyre the bulk of this issue is complete, can you close it and roll the remaining CRL/collateral fetching items into a freshly named 4.x ticket please?

PR #4226, opened 2022-09-13
Use HTTP enclave client to fetch SEV SNP endorsement certificates

jumaffre:
Addresses 1. in #4144 

We now use the HTTP enclave client (rather than the generic client use by the performance test, which makes direct use of sockets) so that all the communications between the CCF node and the Azure/AMD server go through the ring buffer. 

This has a substantial knock-on effect on the node state as this client is asynchronous by nature and a bunch of node initialisation code has been moved to a slightly later state (i.e. when the endorsement certificates have been fetched). 

Note that most of the HTTP fetching code is abstracted to the `ccf::Pal` layer as a single `RetrieveEndorsementCallback` callback is passed to the `generate_quote()` function (no effect on SGX and pure virtual).

TODO:
- [x] Cleanup callback


src/node/quote_endorsements_client.h
```diff
@@ -0,0 +1,65 @@
...
+          if (status != HTTP_STATUS_OK)
+          {
+            LOG_FAIL_FMT(
+              "Error fetching endorsements for attestation report: {}", status);
+          }
+
+          cb(std::move(data));
+        },
+        [](const std::string& error_msg) {
+          // TLS errors should be handled here
```
DomAyre, line 51: Does this need doing in this PR?

DomAyre (approved):

src/node/quote_endorsements_client.h
jumaffre: I'm planning to do this as part of 2. in #4144 in a follow-up PR.

PR #4228, opened 2022-09-13
Use ccache to speed up Daily build

eddyashton:
We specify these `COMPILER_LAUNCHER` args in `matrix.yml`, but not `daily-matrix.yml`. The result is that although we mounted a useful shared drive for cached build results for Daily builds, we never actually used it.


MahatiC (approved):
Nice catch!

PR #4230, opened 2022-09-14
Fix reconfiguration test

DomAyre:
Since fetching attestations involves a network request for endorsements on SNP platforms, it is more likely that an election is triggered while we block for the response.

The reconfiguration tests check trusted nodes exactly match pre-post a new node joining, however the extra election can mean the 'primary' field can have a different value if a different node was elected.

So we check only the trusted node-ids match


tests/reconfiguration.py
```diff
@@ -380,7 +380,16 @@ def test_node_filter(network, args):
         def get_nodes(status):
             r = c.get(f"/node/network/nodes?status={status}")
             nodes = r.body.json()["nodes"]
-            return sorted(nodes, key=lambda node: node["node_id"])
+            return sorted(
+                [
+                    {
+                        "node_id": node["node_id"],
```
DomAyre, line 386: I think this is sufficient fields to include but if anyone knows better, please let me know

eddyashton: I think we should keep all the metadata apart from `primary` in this object - we don't expect anything else to change. We could do that as a dict comprehension, but probably simpler to just explicitly delete this field above

tests/reconfiguration.py
```diff
@@ -380,7 +380,16 @@ def test_node_filter(network, args):
...
+                        "status": node["status"],
+                    }
+                    for node in nodes
+                ],
+                key=lambda node: node["node_id"],
+            )
```
eddyashton, line 392:
```suggestion
            # Primary may change during operation, so do not check for primary equality
            for node in nodes:
              del node['primary']
            return sorted(nodes, key=lambda node: node["node_id"])
```

eddyashton (commented):
We should also add logging just before and just after the calls to `get_quote`, to more easily identify when it is slow (and how slow).

jumaffre (approved):

PR #4234, opened 2022-09-14
Document version = seqno in glossary

jeffa5:
Looking at some bits in the KV API it mentions the `Version` (https://microsoft.github.io/CCF/main/build_apps/kv/api.html) but I couldn't find it in the glossary so this should help.


doc/overview/glossary.rst
```diff
@@ -85,7 +85,7 @@ Glossary
     `Transport Layer Security <https://en.wikipedia.org/wiki/Transport_Layer_Security>`_ is an IETF cryptographic protocol standard designed to secure communications between a client and a server over a computer network.
 
   Transaction ID
-    Unique transaction identifier in CCF, composed of a View and a Sequence Number separated by a period. Sequence Numbers start from 1, and are contiguous. Views are monotonic. E.g. The transaction ID ``2.15`` indicates the View is ``2`` and the Sequence Number is ``15``.  
+    Unique transaction identifier in CCF, composed of a View and a Sequence Number separated by a period. Sequence Numbers start from 1, and are contiguous. Views are monotonic. E.g. The transaction ID ``2.15`` indicates the View is ``2`` and the Sequence Number is ``15``. Sequence Numbers are also referred to as ``Version``s.
```
jumaffre, line 88: Thank you!

```suggestion
    Unique transaction identifier in CCF, composed of a View and a Sequence Number separated by a period. Sequence Numbers start from 1, and are contiguous. Views are monotonic. E.g. The transaction ID ``2.15`` indicates the View is ``2`` and the Sequence Number is ``15``. Sequence Numbers are also referred to as :cpp:type:`kv::Version`s.
```

I think should link directly to (https://microsoft.github.io/CCF/main/build_apps/kv/api.html#_CPPv4N2kv7VersionE) - to be test locally (see [instructions](https://microsoft.github.io/CCF/main/contribute/build_ccf.html#update-the-documentation)).

achamayou: May be good to add "in the context of the Key-Value store"

eddyashton (approved):

Issue #4188, opened 2022-09-02
Verify SEV-SNP attestation security policy


DomAyre:
Currently, we take the "measurement" from the SEV SNP attestation report and treat it identically to MRENCLAVE from SGX. This does some verification but isn't sufficient for full confidence.

Measurement is a digest of the UVM given to SEV SNP, host_data is a hash of the security policy applied within that container which verifies the code being run.

We should verify the policy at least matches a known good policy, this can be done in several ways:

- We have a new table in the KV which stores the unhashed policy, when a new node joins, we check the hash of the policy matches the hash on one of our known good policies.
- We verify the policy first, then hash it and treat the hash as a code_id

The former is likely better since when we eventually want to check for equivalence between non-identical policies we may need the unhashed good policy to do so.

jumaffre:
Retrieving the security policy from the CCF node can be done via the `SECURITY_POLICY` environment variable, that's only available to pid `1` in the ACI container. This should be fine for a real deployment where `cchost` is the command specified as `command` in the container deployment (to be tested). However, for the CI, we currently do not pass any environment variable to the CCF node (so we'd have to explicitly pass this in) and we could for now hardcode the well-known `"allow_all:true"` (`eyJhbGxvd19hbGwiOnRydWUsImNvbnRhaW5lcnMiOnsibGVuZ3RoIjowLCJlbGVtZW50cyI6bnVsbH19`) security policy in the Python infra and use that.

achamayou:
@jumaffre we could also default to fetching the starting node's container, grab its security policy digest, and default to that while leaving the actual policy contents blank.

We can log the fact that it's not set, and suggest setting the (otherwise undiscoverable) environment variable.

I'd rather not have too much test code/sandbox that sets allow_all, because we run the risk that it will be copy-pasted in a real deployments. Defaulting to the actual security policy of the first container seems safer in that respect, at the expense of a small amount of usability for a potential auditor.

jumaffre:
@achamayou @DomAyre and I also talked about this but it'd be nice that a node always checks that, when writing a new key to the security policy table, the key is equal to the hash of the value.

This is very much a CI/dev environment only issue. We could also have the `command` dumps the value of the `SECURITY_POLICY` in a file that can then be read by the Python infra and passed on to each node as an envvar at startup, just like for a real deployment.

achamayou:
[...]

Issue #4193, opened 2022-09-05
Move from identity check of SNP measurement to signer validation


DomAyre:
We currently verify that a measurement, which is a digest of the UVM, matches exactly a known good measurement.

Since UVM updates can happen outside of the users control, this is a fairly brittle system, without telling us too much about the trustworthiness of the node being attested. It would be better to instead verify that the measurement is signed by a trusted entity (via a trusted chain or otherwise). This would give us confidence that the UVM is in a known good state in a way that is also auditable.

We will need to allow the user to configure the root of trust, which likely means a new table to record roots of trust (could initially hardcode the MS one)

Issue #4070, opened 2022-07-26
CI setup for SEV-SNP enabled ACI


jumaffre:
Once #4068 is complete, we should create an ACI setup to test the SEV-SNP variant of CCF in the CI. 

- [x] 1. Add SEV-SNP ACI CI agent (https://github.com/microsoft/CCF/pull/4171)
- [ ] #4619
- [ ] #4621
- [ ] #4620
- [ ] #4622

Risk: We may want to de-risk this early to make sure it is possible to the spin-up a SEV-SNP enabled ACI container with the existing version of the `azure-mgmt-containerinstance`.

DomAyre:
Basic CI running as of https://github.com/microsoft/CCF/pull/4171

Additional tests for functionality should probably wait for/be done as part of changes to the attestation verification in https://github.com/microsoft/CCF/issues/4188 and https://github.com/microsoft/CCF/issues/4193

achamayou:
@DomAyre the pool/jobs are only in the Daily build, but not in the CI, is that correct?

achamayou:
@DomAyre can we close this, and roll the remaining items into a freshly named ticket?

PR #4235, opened 2022-09-14
Use host_data for SNP verification

DomAyre:
Addresses https://github.com/microsoft/CCF/issues/4188


src/node/quote.cpp
```diff
@@ -57,6 +57,57 @@ namespace ccf
...
+    }
+
+    return digest;
+  }
+
+  QuoteVerificationResult verify_security_policy_against_store(
+    kv::ReadOnlyTx& tx,
+    const QuoteInfo& quote_info)
+  {
+    if (quote_info.format == QuoteFormat::amd_sev_snp_v1) {
```
achamayou, line 93: I'm not sure I understand how this works, it seems to be called with the quote_info.format is virtual, but then checks for sev_snp? Defaulting to verified doesn't seem ideal.

src/service/genesis_gen.h
```diff
@@ -420,6 +420,13 @@ namespace ccf
         {.status = CodeStatus::ALLOWED_TO_JOIN, .platform = platform});
     }
 
+    void trust_node_security_policy(
+      const RawPolicy& raw_security_policy, const DigestedPolicy& host_data)
+    {
```
achamayou, line 425: This needs to be exposed via the JS bindings, and then to governance through an action. We need a test, and a documentation sample illustrating how to use it.

include/ccf/node/startup_config.h
```diff
@@ -64,6 +65,8 @@ struct StartupConfig : CCFConfig
 
   nlohmann::json node_data = nullptr;
 
+  RawPolicy security_policy;
```
jumaffre, line 68: Should this be `std::optional` as this would only bet set on SEV-SNP?

include/ccf/service/tables/security_policies.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ccf/service/map.h"
+
+using DigestedPolicy = std::array<uint8_t, 32>;
```
jumaffre, line 7: Should this be a `crypto::Sha256Hash` directly?

src/host/main.cpp
```diff
@@ -403,6 +404,16 @@ int main(int argc, char** argv)
     startup_config.worker_threads = config.worker_threads;
     startup_config.node_certificate = config.node_certificate;
 
+    // Get the nodes security policy via environment variable
+    if (access(ccf::pal::snp::DEVICE, F_OK) == 0) {
+      auto policy = std::getenv("SECURITY_POLICY");
+      if (policy == NULL) {
```
jumaffre, line 410:
```suggestion
      if (policy == nullptr) {
```

As more C++-y.

src/node/quote.cpp
```diff
@@ -57,6 +57,57 @@ namespace ccf
...
+      std::copy(
+        std::begin(quote.host_data),
+        std::end(quote.host_data),
+        digest.begin());
+    }
+    catch (const std::exception& e)
+    {
+      LOG_FAIL_FMT("Failed to verify attestation report: {}", e.what());
+      return std::nullopt;
+    }
```
jumaffre, line 84: Isn't this a critical error that shouldn't be caught here?

tests/infra/remote.py
```diff
@@ -881,6 +881,10 @@ def __init__(
         env = {}
         if enclave_type == "virtual":
             env["UBSAN_OPTIONS"] = "print_stacktrace=1"
+            security_policy_key = "SECURITY_POLICY"
+            if security_policy_key not in os.environ:
+                raise ValueError(f"Security policy environment variable {security_policy_key} is not set")
+            env["SECURITY_POLICY"] = os.environ[security_policy_key]
```
jumaffre, line 887: `os.environ.get(security_policy_key)` so that it doesn't throw if it's unset?

DomAyre: Yeah this was pre the discussion about allowing it to be unset

src/node/quote.cpp
DomAyre: I was just following the pattern used by get_code_id for now since I want to change this eventually anyway, but I could fix both now?

include/ccf/node/startup_config.h
DomAyre: Yeah makes sense, wasn't sure how optional is serialised and figured an empty string would be simpler (But I don't think the current version of the code actually does this)

include/ccf/service/tables/security_policies.h
DomAyre: Yeah assuming it is serialised nicely?

src/node/quote.cpp
jumaffre: Following existing pattern is preferable - we can always change this later.

tests/expired_service/common/actions.js
```diff
@@ -908,6 +908,36 @@ const actions = new Map([
       }
     ),
   ],
+  [
```
jumaffre, line 911: No need to update this file as this is test data only (we should probably move it under a new `testdata/` folder, but definitely out of scope for this PR!).

include/ccf/service/tables/security_policies.h
```diff
@@ -0,0 +1,18 @@
...
+using RawPolicy = std::string;
+
+namespace ccf
+{
+  using SecurityPolicies = ServiceMap<DigestedPolicy, RawPolicy>;
+  namespace Tables
+  {
+    static constexpr auto SECURITY_POLICIES =
+      "public:ccf.gov.security_policies";
```
jumaffre, line 16: I think this should be namespaced under `node`, i.e. `public:ccf.gov.nodes.security_policies` as this is about nodes (see https://microsoft.github.io/CCF/main/audit/builtin_maps.html#nodes-code-ids). 

Also, should this be `public:ccf.gov.nodes.host_data` (more generic) as the security policy is just how this field is interpreted in ACI deployments?

src/node/node_state.h
```diff
@@ -372,6 +372,28 @@ namespace ccf
...
+
+        if (!config.security_policy.has_value()) {
+          LOG_INFO_FMT("Security Policy environment variable not set, skipping check against digest");
+        }
+        else {
+          LOG_INFO_FMT("TEST: Policy found = {}", config.security_policy.value());
+          auto digest = crypto::Sha256Hash(config.security_policy.value());
+          if (digest.h != quoted_digest) {
+            LOG_INFO_FMT("TEST: Policy failed");
+            throw std::logic_error("Raw security policy doesn't match digest provided in attestation");
```
jumaffre, line 392: We should print both so that the operator can find out what has gone wrong.

src/node/rpc/node_frontend.h
```diff
@@ -1515,6 +1516,10 @@ namespace ccf
           in.public_key};
         g.add_node(in.node_id, node_info);
         g.trust_node_code_id(in.code_digest, in.quote_info.format);
+        if (in.quote_info.format == QuoteFormat::amd_sev_snp_v1){
+          auto digest = EnclaveAttestationProvider::get_security_policy_digest(in.quote_info).value();
```
jumaffre, line 1520: Note that it matters much for performance, but you could simply re-hash `in.security_policy` instead of re-extracting from the quote here.

src/service/genesis_gen.h
```diff
@@ -420,6 +420,18 @@ namespace ccf
...
+    void trust_node_security_policy(
+      const std::optional<RawPolicy>& raw_security_policy, const DigestedPolicy& host_data)
+    {
+      auto security_policies = tx.rw(tables.security_policies);
+      if (raw_security_policy.has_value()){
+        LOG_INFO_FMT("Trusting node with policy {}", raw_security_policy.value());
+        security_policies->put(host_data, raw_security_policy.value());}
+      else{
+        LOG_INFO_FMT("Trusting node with unset policy");
+        security_policies->put(host_data, "");}
```
jumaffre, line 432: It would be nice to have a new C++ type for the empty security policy case, rather than `""`.

tests/infra/is_snp.py
```diff
@@ -4,3 +4,5 @@
 import os
 
 IS_SNP = os.path.exists("/dev/sev")
+DEFAULT_SNP_SECURITY_POLICY_DIGEST = "ab6ccfea6e8c9c6233b1a4631cfd6440ecf910694b34d7e92e0236738758831e"
+DEFAULT_SNP_SECURITY_POLICY = "{\"allow_all\":true,\"containers\":{\"length\":0,\"elements\":null}}"
```
jumaffre, line 8: I think we need a ticket and a TODO here to say that it'd be better to pick these up from the CI container directly (out of scope for this PR).

tests/code_update.py
```diff
@@ -51,6 +51,180 @@ def test_verify_quotes(network, args):
...
+    )
+
+    primary, others = network.find_nodes()
+    prev_nodes = [primary, *others]
+    new_node = network.create_node("local://localhost")
+    network.join_node(new_node, replacement_package, args, timeout=3)
+    network.trust_node(new_node, args)
+    primary, others = network.find_nodes()
+    nodes = [primary, *others]
+    assert len(nodes) == len(prev_nodes) + 1, "New node not added to network"
```
jumaffre, line 74: If `network.trust_node(new_node, args)` succeeds (i.e. doesn't throw), then you know that the new has successfully joined the network so need to check for this here.

tests/code_update.py
```diff
@@ -51,6 +51,180 @@ def test_verify_quotes(network, args):
 
     return network
 
+@reqs.description("Node with correct security policy joins successfully")
+def test_add_node_with_correct_security_policy(network, args):
```
jumaffre, line 55: I think this is the "sunny day" scenario that's already tested in the CI when a new node joins successfully on SNP (e.g. `reconfiguration_test_cft`), so this only tests the new `GET /gov/security_policy` endpoint, which I think can be folded into the existing `test_update_all_nodes` test.

tests/code_update.py
```diff
@@ -51,6 +51,180 @@ def test_verify_quotes(network, args):
...
+    assert policies == expected, [(a, b) for a, b in zip(policies, expected)]
+
+@reqs.description("Node with no security policy set but good digest matching ledger joins successfully")
+def test_add_node_with_no_raw_security_policy_match_ledger(network, args):
+
+    if not IS_SNP:
+        LOG.warning("Skipping test_add_node_with_correct_security_policy with non SNP platform")
+        return network
+
+    replacement_package = (
```
jumaffre, line 97: I don't think we need to use the replacement package here and simply `join_node(new_node, args.package)` for conciseness?

DomAyre: Haha yeah I just did this, great minds

tests/code_update.py
```diff
@@ -51,6 +51,180 @@ def test_verify_quotes(network, args):
...
+    )
+
+    primary, others = network.find_nodes()
+    prev_nodes = [primary, *others]
+
+    LOG.info(f"Removing security policy set by node 0 so that a new joiner is seen as an unmatching policy")
+    network.consortium.retire_security_policy(primary, DEFAULT_SNP_SECURITY_POLICY_DIGEST)
+
+    new_node = network.create_node("local://localhost")
+    network.join_node(new_node, replacement_package, args, timeout=3)
```
jumaffre, line 222: I would expect this call to timeout in this case?

include/ccf/service/tables/security_policies.h
DomAyre: I think this could go either way, I agree we should be more generic, but maybe the table cares about security policies, since that determines how we verify it (also includes the raw policy not just host data). And then the platform specific thing is that on ACI deployments, we get the digest of the security policy by reading the host_data?

tests/code_update.py
```diff
@@ -51,6 +51,180 @@ def test_verify_quotes(network, args):
...
+        "samples/apps/logging/liblogging"
+        if args.package == "libjs_generic"
+        else "libjs_generic"
+    )
+
+    primary, others = network.find_nodes()
+    prev_nodes = [primary, *others]
+    new_node = network.create_node("local://localhost")
+    try:
+        network.join_node(new_node, replacement_package, args, timeout=3, env={"SECURITY_POLICY": "eyJhbGxvd19hbGwiOmZhbHNlLCJjb250YWluZXJzIjp7Imxlbmd0aCI6MCwiZWxlbWVudHMiOm51bGx9fQ=="})
```
jumaffre, line 193:
```suggestion
        network.join_node(new_node, replacement_package, args, timeout=3, env={"SECURITY_POLICY": "invalid_security_policy"})
```

So that it's obvious to the reader what the test is doing.

src/node/rpc/node_frontend.h
DomAyre: Yeah I was hoping to have the refactor done where we would just read the appropriate field rather than re-verifying, I'll change this

DomAyre: Oh wait, we can't do this if we want to allow user to not set the raw security policy

src/service/genesis_gen.h
DomAyre: Would it still serialise to the empty string?

tests/code_update.py
DomAyre: That's fair although maybe that test then becomes a bit unwieldly? Since joining is tested in the setup maybe I keep this test but just check the policy is what we expect?

tests/code_update.py
DomAyre: Unfortunately in it's current state it's joining successfully, I suspect something to do with the table not being up to date everywhere when we retire a policy?

tests/code_update.py
DomAyre: If we do that, joining fails during converting to a string rather than the check to ensure the raw policy matches the digest, so we wouldn't have that coverage of that check. Of course, if the base64 doesn't parse then it also shouldn't match so I can see the argument not to worry about that?

tests/code_update.py
DomAyre: Okay removed some of the checks

tests/code_update.py
jumaffre: Good point. Would this work if we did `b64encode(b"invalid_security_policy").decode()` (valid base64 string but clear to the reader what the test does)?

DomAyre: Yeah that's a great idea

src/service/genesis_gen.h
jumaffre: Or it could be `static constexpr default_empty_security_policy = "";`.

tests/code_update.py
```diff
@@ -52,6 +53,91 @@ def test_verify_quotes(network, args):
     return network
 
 
+@reqs.description("Test that the security policies table is correctly populated")
+@reqs.snp_only()
+def test_add_node_with_correct_security_policy(network, args):
+
+    LOG.info(f"Checking governance table security policies")
```
jumaffre, line 60: FYI, the `@reqs.description` message is printed when the test starts so no need for an extra log line here.

tests/code_update.py
jumaffre: Yes this is unexpected. AIUI, the call to `join_node` should time out here?

include/ccf/service/tables/security_policies.h
DomAyre: It is now serialised nicely

tests/code_update.py
DomAyre: Yeah I think fixing the governance stuff while changing digest to a SHA256 fixed this so now it does time out as expected

src/service/genesis_gen.h
DomAyre: Added some docs, testing bad security policies has some coverage of the actions since they rely on us removing the default policy and re-adding it later. I think that's sufficient if you agree?

doc/operations/code_upgrade.rst
```diff
@@ -179,7 +179,7 @@ Procedure
 - Members should be use the ``set_constitution`` proposal action to update the constitution scripts.
 - See :ref:`bundle deployment procedure <build_apps/js_app_bundle:Deployment>` to update the JavaScript/TypeScript application.
 
-6. Finally, once the code upgrade process has been successful, the old code version (i.e. the code version run by nodes 0, 1 and 2) can be removed using the ``remove_node_code`` proposal action.
+6. Finally, once the code upgrade process has been successful, the old code version (i.e. the code version run by nodes 0, 1 and 2) can be removed using the ``remove_node_code`` or ``remove_security_policy`` proposal action.
```
jumaffre, line 182:
```suggestion
6. Finally, once the code upgrade process has been successful, the old code version (i.e. the code version run by nodes 0, 1 and 2) can be removed using the ``remove_node_code`` or ``remove_security_policy`` proposal actions.
```

src/node/rpc/node_frontend.h
```diff
@@ -149,6 +150,13 @@ namespace ccf
           return std::make_pair(
             HTTP_STATUS_UNAUTHORIZED,
             "Quote report data does not contain node's public key hash");
+        case QuoteVerificationResult::FailedSecurityPolicyDigestNotFound:
+          return std::make_pair(
+            HTTP_STATUS_UNAUTHORIZED,
+            "Quote doesn't contain a security policy digest");
```
jumaffre, line 156:
```suggestion
            "Quote does not contain a security policy digest");
```

src/node/rpc/node_frontend.h
```diff
@@ -1515,6 +1523,14 @@ namespace ccf
           in.public_key};
         g.add_node(in.node_id, node_info);
         g.trust_node_code_id(in.code_digest, in.quote_info.format);
+        if (in.quote_info.format == QuoteFormat::amd_sev_snp_v1)
+        {
+          auto digest = EnclaveAttestationProvider::get_security_policy_digest(
+                          in.quote_info)
+                          .value();
+          LOG_INFO_FMT("Trusting digest {}", digest.hex_str());
```
jumaffre, line 1531: Remove log here?

samples/constitutions/default/actions.js
```diff
@@ -929,6 +929,43 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "add_security_policy",
+    new Action(
+      function (args) {
+        checkType(args.raw_security_policy, "string", "raw_security_policy");
+        checkType(
+          args.digested_security_policy,
```
jumaffre, line 938: Maybe `security_policy_hash` or `security_policy_digest` as "digested" sounds a little odd?

samples/constitutions/default/actions.js
```diff
@@ -929,6 +929,43 @@ const actions = new Map([
...
+          "string",
+          "digested_security_policy"
+        );
+      },
+      function (args, proposalId) {
+        const digest = ccf.strToBuf(args.digested_security_policy);
+        const raw = ccf.jsonCompatibleToBuf(args.raw_security_policy);
+        ccf.kv["public:ccf.gov.nodes.security_policies"].set(digest, raw);
```
jumaffre, line 946: We need to enforce that `digest` is the digest of `raw` here.

src/service/genesis_gen.h
```diff
@@ -420,6 +420,24 @@ namespace ccf
...
+    void trust_node_security_policy(
+      const std::optional<RawPolicy>& raw_security_policy,
+      const DigestedPolicy& host_data)
+    {
+      auto security_policies = tx.rw(tables.security_policies);
+      if (raw_security_policy.has_value())
+      {
+        LOG_INFO_FMT(
+          "Trusting node with policy {}", raw_security_policy.value());
+        security_policies->put(host_data, raw_security_policy.value());
```
jumaffre, line 432: Same as for the constitution, we need to enforce that `host_data` is the digest of `raw_security_policy` here.

tests/infra/is_snp.py
```diff
@@ -4,3 +4,10 @@
 import os
 
 IS_SNP = os.path.exists("/dev/sev")
+# TODO: Get these from the container
```
jumaffre, line 7: FYI, I've added item 5. in https://github.com/microsoft/CCF/issues/4070 to cover this.

tests/code_update.py
```diff
@@ -52,6 +57,126 @@ def test_verify_quotes(network, args):
...
+    network.consortium.add_new_security_policy(
+        primary, "", DEFAULT_SNP_SECURITY_POLICY_DIGEST
+    )
+
+    new_node = network.create_node("local://localhost")
+    network.join_node(new_node, args.package, args, timeout=3, env=dict())
+    network.trust_node(new_node, args)
+
+    LOG.info(f"Checking node joined with a blank security policy")
+    # TODO: Do this
```
jumaffre, line 102: Remove?

tests/code_update.py
```diff
@@ -52,6 +57,126 @@ def test_verify_quotes(network, args):
...
+            "raw": DEFAULT_SNP_SECURITY_POLICY,
+        }
+    ]
+    expected.sort(key=lambda x: x["digest"])
+
+    assert policies == expected, [(a, b) for a, b in zip(policies, expected)]
+
+
+@reqs.description(
+    "Node with no security policy set but good digest matching ledger joins successfully"
```
jumaffre, line 81: I think that here and elsewhere in this file, we should avoid saying "matching ledger" but use un/authorised instead as this test doesn't actually check the ledger files.

tests/code_update.py
```diff
@@ -52,6 +57,126 @@ def test_verify_quotes(network, args):
...
+        network.join_node(
+            new_node,
+            args.package,
+            args,
+            timeout=3,
+            env={"SECURITY_POLICY": b64encode(b"invalid_security_policy").decode()},
+        )
+        network.trust_node(new_node, args)
+    except Exception:
+        ...
```
jumaffre, line 143: I think this should be a `TimeoutError` here? Also, checking the length of the `nodes` below could be replaced by a simple `else` statement to the `try catch` that will be triggered if no exception is raised.

tests/code_update.py
```diff
@@ -52,6 +57,126 @@ def test_verify_quotes(network, args):
...
+    LOG.info(
+        f"Removing security policy set by node 0 so that a new joiner is seen as an unmatching policy"
+    )
+    network.consortium.retire_security_policy(
+        primary, DEFAULT_SNP_SECURITY_POLICY_DIGEST
+    )
+
+    new_node = network.create_node("local://localhost")
+    try:
+        network.join_node(new_node, args.package, args, timeout=3)
```
jumaffre, line 166: Here and for other negative tests, could you `new_node.stop()` at the end of the test to make sure we don't have too many nodes running by the end of the suite?

jumaffre (approved):
Minor comments - LGTM!

tests/code_update.py
DomAyre: The tricky part with authorised/unauthorised is either way the policy is authorised (since we authorise the digest). I've had a go at phrasing it better

tests/infra/is_snp.py
DomAyre: Sounds great

src/service/genesis_gen.h
DomAyre: We currently do a check on node start up, my understanding is that this isn't exposed publicly so would the check on node startup give us enough confidence not to need the check here? Or is it for defence in depth?

jumaffre: I think this is more of a test for us, as this is hard to unit test, in case we ever break this.

src/service/genesis_gen.h
DomAyre: I've had a slight change of heart here, I had done this as of a68507926995c6e0ef747055e7dd0e8b1387c56b but that was an `snp::` field which then wouldn't compile for SGX, I also only really used it here so I don't think it's strictly necessary?

DomAyre: Fixing SGX builds allowed me to bring this back

samples/constitutions/default/actions.js
```diff
@@ -929,6 +943,56 @@ const actions = new Map([
...
+        const digest = ccf.strToBuf(args.security_policy_digest);
+        const raw = ccf.jsonCompatibleToBuf(args.security_policy_raw);
+
+        const redigested_raw = ccf.bufToStr(
+          ccf.digest("SHA-256", ccf.strToBuf(args.security_policy_raw))
+        );
+        const quoted_digest = ccf.bufToStr(
+          hexStrToBuf(args.security_policy_digest)
+        );
+        if (args.security_policy_raw != "" && redigested_raw != quoted_digest) {
```
achamayou, line 967: I seems clearer to have this condition earlier, and to only do the redigest if the string isn't empty?

doc/audit/builtin_maps.rst
```diff
@@ -143,6 +143,15 @@ Versions of the code allowed to join the current network.
...
 .. note:: On versions ``3.0.0-dev1`` and below, value was an enum with one possible value ``ALLOWED_TO_JOIN``
 
+``nodes.security_policies``
+~~~~~~~~~~~~~~~~~~
+
+Security policies that nodes can run and join the current network.
+
+**Key** Security Policy Digest: Hash of the security policy string which is reported by the SNP attestation.
+
+**Value** Raw Security Digest: The raw string of the security policy
```
achamayou, line 153: We need a link to something that explains what a security policy is.

doc/audit/builtin_maps.rst
```diff
@@ -143,6 +143,15 @@ Versions of the code allowed to join the current network.
 
 .. note:: On versions ``3.0.0-dev1`` and below, value was an enum with one possible value ``ALLOWED_TO_JOIN``
 
+``nodes.security_policies``
+~~~~~~~~~~~~~~~~~~
+
+Security policies that nodes can run and join the current network.
+
+**Key** Security Policy Digest: Hash of the security policy string which is reported by the SNP attestation.
```
achamayou, line 151: This should state which attestation fields the security policy digest is reported in.

doc/governance/common_member_operations.rst
```diff
@@ -27,7 +27,13 @@ Updating Code Version
...
 For new nodes to be able to join the network, the version of the code they run (as specified by ``enclave.file``) should be first trusted by the consortium of members.
 
-If the version of the code being executed needs to be updated (for example, to support additional endpoints), members can create an ``add_node_code`` proposal, specifying the new code version.
+If the version of the code being executed needs to be updated (for example, to support additional endpoints), members can use the following endpoints depending on the platform:
+
+- SGX
+Create an ``add_node_code`` proposal, specifying the new code version.
+
+- SNP
+If the new code requires a new security policy, create an ``add_security_policy`` proposal, specifying both the raw policy string and the digest of the policy, (SHA-256 of the raw policy)
```
achamayou, line 36: I am less and less convinced by the re-use of the code id concept. In C-ACI usage, this code is is the UVM digest, it is totally independent of the CCF enclave code. Only the security policy specifies what code the node is running. In fact, we will validate this UVM identity using signatures, precisely because it can and will change without the application code changing at all.

samples/constitutions/default/actions.js
DomAyre: Yeah makes sense

doc/governance/common_member_operations.rst
DomAyre: Yes I agree, I think the plan is as part of https://github.com/microsoft/CCF/issues/4193 we verify signatures instead of the identity check then this issue goes away.

We could have a separate, clearly named table in the meantime (something like uvm_measurment) but probably out of scope for this PR?

doc/audit/builtin_maps.rst
DomAyre: This is a bit similar to a discussion I had with Julien about maybe calling it host_data rather than security policy, I think it would make sense to keep the table documentation separate from the specific way in which it's attained in different platforms? As far as I understand, other platforms may have different usages of these fields right? 

I could mention we get it from host_data for the SNPCC case?

doc/audit/builtin_maps.rst
DomAyre: I spoke to Ken and he said there isn't any public documentation yet, I can keep an eye on it and add it when there are some?

achamayou: @DomAyre sounds good! I think a note above this saying this is experimental would be good.

doc/audit/builtin_maps.rst
achamayou: Yes, that would be good, and yes it can differ on other platforms and we'll need to document that if/when it happens.

achamayou (approved):

Issue #4155, opened 2022-08-19
Multi-Container: Implement the registration protocol


eddyashton:
It should be possible to separate registration from the rest of execution, so that registration can be done in a common sidecar and handed over to the 'real' executor.

The flow here is roughly:
- A new executor creates a fresh public-private key-pair
- The executor contacts CCF's `/register_executor` endpoint over (anonymous-client) TLS, presenting their attestation A and public key P
- The CCF nodes checks the attestation A, and checks the contained code ID is trusted (#4154)
- The CCF node stores the public key P as accepted for its executor auth policy, and responds with an OK
- The executor uses the private key (possibly transmitted from the side car to the 'real' executor) to establish a TLS connection with CCF
- That TLS connection authorises access to the API defined in #4153

MahatiC:
Revisit:

- [ ] Explore if we need certificate endorsements for Executors
- [ ] Explore if we need to use attested TLS

achamayou:
@MahatiC @eddyashton is this complete? is this only missing an end to end test using ACI + attestation container?

MahatiC:
@achamayou Yes, an end to end test will complete this -- that's already being tracked here https://github.com/microsoft/CCF/issues/4799. So I'm closing this.

PR #4237, opened 2022-09-15
Add registration protocol for the executor

MahatiC:
Part of #4155


src/apps/external_executor/external_executor.cpp
```diff
@@ -17,11 +21,14 @@
 namespace externalexecutor
 {
   using Map = kv::Map<std::string, std::string>;
+  using ExecutorNodeId = ccf::EntityId<ccf::NodeIdFormatter>;
```
jumaffre, line 24: `ccf::NodeId`?

src/apps/external_executor/external_executor.cpp
```diff
@@ -45,6 +52,35 @@ namespace externalexecutor
         ccf::no_auth_required)
         .set_auto_schema<void, GetExecutorCode::Out>()
         .install();
+
+      // create an endpoint to register the executor
+      auto register_executor = [this](auto& ctx, ccf::NewExecutor&& payload)
+        -> ccf::grpc::GrpcAdapterResponse<ccf::RegistrationResult> {
+        // generate and store executor node id locally
+        crypto::Pem executor_public_key(payload.cert());
```
jumaffre, line 60: I think we should use the public key from the `ctx` object (I'm assuming we're doing mutual TLS here?). See existing example in `node_frontend.h` around line `440` and the use of `rpc_ctx->get_session_context()->caller_cert`.

src/apps/external_executor/external_executor.cpp
```diff
@@ -45,6 +52,35 @@ namespace externalexecutor
...
+        auto pubk_der = crypto::cert_pem_to_der(executor_public_key);
+        ExecutorNodeId executor_node_id =
+          ccf::compute_node_id_from_pubk_der(pubk_der);
+
+        ExecutorNodeInfo executor_info = {
```
jumaffre, line 65: We should check that the hash of the public key is recorded as report data in the attestation.

src/apps/external_executor/external_executor.cpp
MahatiC: Yeah, it's the same. It's just a bit of future proofing -- in case we decide to use a custom `ExecutorNodeIdFormatter` instead of `NodeIdFormatter`

src/apps/external_executor/external_executor.cpp
MahatiC: This is intentionally retrieved from the body. For a couple of reasons as discussed with @eddyashton 

1. We avoid the need to obtain a private key from the `Executor`
2. If we go the route of offloading registration to a side car, we don't end up passing a private key this way (although there is still a chance that the side car could go rogue on its own without a private key as well)

eddyashton: To clarify: we're _not_ doing mutual TLS auth on the register endpoint - it's possible and expected that the registration is done over an anonymous TLS channel.

Another benefit of doing it this way that I hadn't quite twigged yesterday, is that the auth policy for registration is extremely awkward the other way round. You need to come with _a_ TLS cert, but not one we know anything about yet? In fact, if it is known, is that an auth error? Is a mismatch between the TLS cert and the attestation an auth error? I think it's simpler to reason about this cert alongside the attestation purely from the (transport-agnostic!) payload.

jumaffre: We still need to check for ownership of the private key corresponding to the public key, otherwise an attacker could submit a copy the attestation of the valid node (assumed public), along with the valid node's public key and successfully register that way. We could guarantee ownership of the public key with a signature using the corresponding private key and submit that signature in the payload but this doesn't guarantee freshness like TLS does.

eddyashton: Why do we need to check ownership at this stage? That allows an attacker to register a valid executor, but not to substitute their own key in any way - the result is still that the executor is registered, with the executor's public key. When they use that for TLS to actually interact (ie - make any of the calls to the `KV` service), they prove ownership of the corresponding private key at that point.

src/apps/external_executor/external_executor.cpp
eddyashton: How difficult is it to create an attestation with specific report data? As in, is this more difficult/more awkward to do in arbitrary languages than getting a 'dumb' attestation?

(I agree we need to do this regardless, just thinking about the implication - if it's tricky it's more likely we build a sidecar/binary/call-this solution)

src/apps/external_executor/external_executor.cpp
jumaffre: OK, that makes sense but e.g. if the external executor gets recorded in the ledger (I don't think it is in the current plan?), then it may at least confusing to an auditor?

src/apps/external_executor/external_executor.cpp
jumaffre: For SEV-SNP, we need to make an `ioctl` call so probably OK for most low-level languages but awkward from something like JS? For SGX, we need to call into Open Enclave so even more awkward. It seems to me that this needs to be a separate binary, either running in the same container as the BL or a side-car as you said.

src/apps/external_executor/external_executor.cpp
MahatiC: Yes in the current plan, executor details are local to the node and won't be in the ledger.

src/apps/external_executor/protobuf/executor_registration.proto
```diff
@@ -61,28 +61,10 @@ message NewExecutor
...
-
-    // The given public cert is already registered on this node for a
-    // distinct code ID. Note that calls should be idempotent, so repeated
-    // registration with identical requests should return ACCEPTED
-    DUPLICATE_CERT = 3;
-
-    // Any error not covered above
-    REJECTED = 4;
-  }
-  Outcome outcome = 1;
```
MahatiC, line 85: These are covered by gRPC error responses `ccf::grpc::make_error`

src/apps/external_executor/external_executor.cpp
achamayou: @jumaffre even if we did, that would be indistinguishable from an executor that had registered successfully, but then never handled any requests, perhaps because it crashed in between. This is true both in terms of audit and impact.

src/apps/external_executor/executor_code_id.h
```diff
@@ -3,10 +3,50 @@
...
 #include "ccf/service/code_digest.h"
 #include "ccf/service/map.h"
 #include "ccf/service/tables/code_id.h"
+#include "endpoints/grpc_status.h"
+#include "executor_registration.pb.h"
+
+struct ExecutorNodeInfo
+{
+  crypto::Pem public_key;
+  ccf::Attestation attest;
```
jumaffre, line 16:
```suggestion
  ccf::Attestation attestation;
```

src/apps/external_executor/external_executor.cpp
```diff
@@ -45,6 +52,45 @@ namespace externalexecutor
...
+          executor_public_key, payload.attestation()};
+
+        ExecutorNodeIDs[executor_node_id] = executor_info;
```
jumaffre, line 68: We should verify the quote before registering the executor node.

src/apps/external_executor/external_executor.cpp
```diff
@@ -45,6 +52,45 @@ namespace externalexecutor
...
+        -> ccf::grpc::GrpcAdapterResponse<ccf::RegistrationResult> {
+        // verify quote
+        ccf::CodeDigest code_digest;
+        ccf::QuoteVerificationResult verify_result = verify_executor_quote(
+          ctx.tx, payload.attestation(), payload.cert(), code_digest);
+
+        if (verify_result != ccf::QuoteVerificationResult::Verified)
+        {
+          const auto [code, message] = verification_error(verify_result);
+          return ccf::grpc::make_error(GRPC_STATUS_UNAUTHENTICATED, message);
```
eddyashton, line 67:
```suggestion
          return ccf::grpc::make_error(code, message);
```

src/apps/external_executor/protobuf/executor_registration.proto
eddyashton: If we're going this route, I think we should remove the `outcome` field entirely. It's not helpful to give them an additional string to parse, it should either be an enum with limited values, or a description for humans (the latter is already provided by `details`). We should probably insert the `executor_id` in here as well - the Executor never _needs_ to know this ID we generated for them, but it may help correlate things in debugging logs.

Separately, I'm not sure if switching from in-body error messages to gRPC errors is a good idea, as there's a risk of the generated code being awkward for clients to use. In Python for instance, this will result in a bunch of try-catches around registration calls. I think it's helpful to distinguish fundamental errors (you didn't create the right kind of channel, you didn't format your body correctly, you didn't send the correct argument type) which are gRPC errors, from application-space errors (basically as soon as you've given us a parseable `NewExecutor` message) which can be reported in application-space response bodies. This is maybe an unnecessary distinction - we don't do it for HTTP, where both most of our apps and the framework report failures by the status code. But we should work out what's the right pattern for gRPC.

eddyashton: Ok, ignore the second part.

After a little searching, it looks like standard gRPC practice is to use errors at the gRPC level (ie - `grpc-status` and associated details) rather than in app-space. This gets converted to client side exceptions in languages that don't mind exceptions, and manual failure-checking in things like C++. That seems backwards to me - you can't do the same for streamed responses, so you have a weird step-change in behaviour when you switch from unary response (might throw) to streamed response (check the body of each response for errors, a throw means a channel-level failure). But I don't make the precedent.

achamayou (approved):

PR #4249, opened 2022-09-16
Add option to automatically shut down `sandbox.sh`

jumaffre:
Minor addition to the existing `sandbox.sh` script: it is sometimes useful to be able to start the script and automatically shut down after the service is successfully open, for example when app developers run this script as part of their CI.

### DIFF
--- a/tests/start_network.py
+++ b/tests/start_network.py
@@ -113,6 +113,16 @@ def run(args):
                     )
                 )
 
+            if args.auto_shutdown:
+                if args.auto_shutdown_delay_s > 0:
+                    LOG.info(
+                        f"Waiting {args.auto_shutdown_delay_s}s before automatic shutdown..."
+                    )
+                    time.sleep(args.auto_shutdown_delay_s)
+
+                LOG.info("Automatically shut down network after successful opening")
+                return
+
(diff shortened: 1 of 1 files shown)


tests/start_network.py
```diff
@@ -113,6 +113,16 @@ def pad_node_id(nid):
...
 
+            if args.auto_shutdown:
+                if args.auto_shutdown_delay_s > 0:
+                    LOG.info(
+                        f"Waiting {args.auto_shutdown_delay_s}s before automatic shutdown..."
+                    )
+                    time.sleep(args.auto_shutdown_delay_s)
+
+                LOG.info("Automatically shut down network after successful opening")
+                return
```
eddyashton, line 124: Why here rather than at line 140? If anyone ever runs this by hand, it would be nice to keep the logging similar (and still allow `Ctrl+C` to terminate the service earlier than the auto shutdown?)

jumaffre: Yes you're right. I started with the no delay case first, but this doesn't make sense with the delayed shutdown case. Fixed now.

MahatiC (approved):

eddyashton (approved):

PR #4251, opened 2022-09-16
Refactor attestation code into platform specific files

DomAyre:


DomAyre:
Currently a minimal refactor, still populates QuoteInfo and uses existing structures to handle it, haven't removed any of the old code yet

src/pal/attestation.cpp
```diff
@@ -0,0 +1,97 @@
...
+#if !defined(INSIDE_ENCLAVE) || defined(VIRTUAL_ENCLAVE)
+#  include "pal/attestation_snp.cpp"
+#  include "pal/attestation_virtual.cpp"
+#else
+#  include "pal/attestation_sgx.cpp"
+#endif
+
+namespace ccf::pal::attestation {
+
+#if !defined(INSIDE_ENCLAVE) || defined(VIRTUAL_ENCLAVE)
```
jumaffre, line 13: Would it be possible to reduce the number of times we have this compile-time pattern in the code base and use this instead, e.g.:
```c++
#if !defined(INSIDE_ENCLAVE) || defined(VIRTUAL_ENCLAVE)
  using AttestationImpl = SnpAttestation;
#else
  using AttestationImpl = SgxAttestation;
#endif

ReportData make_report_data(crypto::Sha256Hash& node_public_key_hash) 
{
    AttestationImpl::make_report_data(node_public_key_hash);
}
```

src/pal/attestation_snp.cpp
```diff
@@ -0,0 +1,373 @@
...
+            CCF_APP_FAIL("Payload error: {}", payload.error);
+            throw std::logic_error(
+            "Failed to issue ioctl SEV_SNP_GUEST_MSG_REPORT2");
+        }
+
+        auto quote = &resp.report;
+        auto quote_bytes = reinterpret_cast<uint8_t*>(&resp.report);
+        attestation.quote.raw.assign(quote_bytes, quote_bytes + resp.report_size);
+
+        client::RpcTlsClient client{
```
jumaffre, line 227: This will conflict with #4226

DomAyre:
Not a priority, so closing for now

PR #4255, opened 2022-09-21
Remove support for 1.x snapshots

jumaffre:
Addresses 1. in https://github.com/microsoft/CCF/issues/2981

Now that 2.x releases are cut from their own `release/2.x` branch, we no longer need to support 1.x snapshots on `main`. These snapshots didn't embed a receipt because these weren't available at the time snapshots were introduced, but they do now and we do not expect operators to start 3.x services from a 1.x snapshots. 

Note: Also contains changes from #4226


achamayou:
@jumaffre can we split out the changes from #4226? They make reviewing this PR to check it only changes the snapshot handling a bit difficult.

src/node/snapshot_serdes.h
```diff
@@ -71,62 +50,56 @@ namespace ccf
     auto tx_hdr = serialized::peek<kv::SerialisedEntryHeader>(data, size);
     auto store_snapshot_size = sizeof(kv::SerialisedEntryHeader) + tx_hdr.size;
 
-    // Snapshots without a snapshot evidence seqno specified by the host should
-    // be self-verifiable with embedded receipt
-    if (!evidence_seqno.has_value())
+    auto receipt_data = data + store_snapshot_size;
+    auto receipt_size = size - store_snapshot_size;
+
```
achamayou, line 55: An early check here, with an error message if the receipt_size is empty to point out that this is unsupported snapshot may be good.

jumaffre:
> @jumaffre can we split out the changes from #4226? They make reviewing this PR to check it only changes the snapshot handling a bit difficult.

Yes. The goal was to have #4226 merged first before opening this PR for review. Hopefully #4226 gets a green CI run soon and this PR gets easier to review/merge.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - `set_user` action in sample constitutions correctly handles `user_data` (#4229).
 
+### Removed
+
+- Snapshots generated by 1.x services can no longer be used to join from or recover a new service (#4255).
```
achamayou, line 16: Let's add a note to make it clear that this means a 1.x ledger must effectively upgrade to 2.x first and produce at least one snapshot, before going to 3.x. Or that it is at the very least strongly recommended, unless the ledger is quite short.

achamayou (approved):

include/ccf/node_startup_state.h
```diff
@@ -14,8 +14,7 @@ namespace ccf
     partOfPublicNetwork,
     partOfNetwork,
     readingPublicLedger,
-    readingPrivateLedger,
-    verifyingSnapshot
+    readingPrivateLedger
```
achamayou, line 17: That enum is never serialized to the ledger, right?

src/kv/kv_types.h
```diff
@@ -354,7 +354,6 @@ namespace kv
     PASS_BACKUP_SIGNATURE_SEND_ACK = 4,
     PASS_NONCES = 5,
     PASS_NEW_VIEW = 6,
-    PASS_SNAPSHOT_EVIDENCE = 7,
```
achamayou, line 357: Is that gap correctly handled as unknown status? We don't want it falling through accidentally.

include/ccf/node_startup_state.h
jumaffre: It isn't. It's only reported by one of the `/node` operator endpoints.

achamayou: Great, I thought this was the case, just wanted to confirm.

src/kv/kv_types.h
achamayou: Great, thank you.

achamayou (approved):

PR #4257, opened 2022-09-22
Initial implementation of KV service, basic gRPC sample app

eddyashton:
Implements most of the current RPCs from the `KV.proto` file, and extends the e2e test with an example of a standalone executor script (currently run in a separate thread, not in a container).

The big feature still missing is forwarding the response (sent from the Executor to `EndTx`) to the calling client. Currently their response is immediate and uninteresting. I'll add HTTP/2 forwarding support in a separate PR.


tests/external_executor/external_executor.py
```diff
@@ -50,18 +53,7 @@ def test_put_get(network, args):
...
-    def require_missing(tx, table, key):
-        try:
-            tx.Get(KV.KVKey(table=table, key=key))
-        except grpc.RpcError as e:
-            assert e.code() == grpc.StatusCode.NOT_FOUND  # pylint: disable=no-member
-            assert (
-                e.details() == f"Key {key.decode()} does not exist"
-            )  # pylint: disable=no-member
-        else:
-            assert False, f"Getting unknown key {key} should raise an error"
```
jumaffre, line 62: I think this is still a valuable test?

src/apps/external_executor/protobuf/http.proto
```diff
@@ -0,0 +1,74 @@
+syntax = "proto3";
```
jumaffre, line 1: Why is this file required?

eddyashton: So that the `EndTx` gRPC message can describe a HTTP response.

MahatiC (approved):

jumaffre (approved):

PR #4258, opened 2022-09-22
Support for simple (synchronously produced) streaming gRPC messages

eddyashton:
The important changes are in `grpc.h`, everything else is just a sample/test code. That's currently a new service implemented in `external_executor.cpp` - we should probably break that out into a standalone gRPC test/sample app, as `external_executor` gets more real.

Adds a way to implement gRPC RPCs which send and receive _streams_ of messages, rather than simply unary messages. We represent those in the CCF code as a `std::vector<T>` rather than a `T`, with the same expectations on `T` (ie - it should be a type generated from a `.proto` file). The actual encoding is simple - each message gets its own 5-byte header, and there's no shared "this is a stream" header.

The messages we actually send in the test service has a `oneof` message type to confirm we're handling variant-typed messages correctly, and know how to parse + produce them on either end.


jumaffre (approved):
Is my understanding correct that the client sends `N` gRPC messages, over the same HTTP/2 request, that are processed all at once by the handler which constructs a single response payload made up of `N` gRPC responses? If so, that's neat!

eddyashton:
> Is my understanding correct that the client sends `N` gRPC messages, over the same HTTP/2 request, that are processed all at once by the handler which constructs a single response payload made up of `N` gRPC responses? If so, that's neat!

That's correct, with the caveat that the `N`'s don't have to match - it's also possible to return more or fewer responses, even unary or empty responses.

PR #4266, opened 2022-09-26
Also install `.inc` files

jumaffre:
This is required for the new `libprotobuf` dependency.

### DIFF
--- a/.azure-pipelines-templates/configure.yml
+++ b/.azure-pipelines-templates/configure.yml
@@ -8,7 +8,7 @@ jobs:
         clean: true
       - script: |
           echo "Determine if any code has changed."
-          if git diff --ignore-submodules=dirty --quiet origin/${SYSTEM_PULLREQUEST_TARGETBRANCH:-origin/main} -- ':!docs' ':!*.md' ':!*.txt'; then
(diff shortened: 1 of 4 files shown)


.azure-pipelines-templates/configure.yml
```diff
@@ -8,7 +8,7 @@ jobs:
         clean: true
       - script: |
           echo "Determine if any code has changed."
-          if git diff --ignore-submodules=dirty --quiet origin/${SYSTEM_PULLREQUEST_TARGETBRANCH:-origin/main} -- ':!docs' ':!*.md' ':!*.txt'; then
+          if git diff --ignore-submodules=dirty --quiet origin/${SYSTEM_PULLREQUEST_TARGETBRANCH:-origin/main} -- ':!doc' ':!*.md' ':!*.rst'; then
```
jumaffre, line 11: Unrelated to the aim of this PR but I noticed that the CI wasn't running on this change because I only modified `.txt` and `md` files. This should be fixed now.

achamayou: The txt exemption is because of the third party notices, perhaps we should exclude that explicitly then, given that CMakeLists.txt should certainly trigger a CI run. The .rst check seems redundant with doc?

jumaffre: Most of our `.txt` files are `CMakeLists.txt` so I think it's OK to run the full CI for all `.txt` files (keeps the exception simple). I'll remove the `.rst` exception though.

achamayou (approved):

achamayou (approved):

PR #4271, opened 2022-09-26
Add ccf.generateEcdsaKeyPair

achamayou:
We support creating RSA and AES keys in the JS API, but not ECDSA. This is adding that for the curves we currently support: secp256r1 and secp384r1.

TODO:

- [x] polyfill
- [x] testcase
- [x] documentation

### DIFF
--- a/js/ccf-app/src/crypto.ts
+++ b/js/ccf-app/src/crypto.ts
@@ -28,6 +28,11 @@ export const generateAesKey = ccf.generateAesKey;
 export const generateRsaKeyPair = ccf.generateRsaKeyPair;
 
 /**
+ * @inheritDoc CCF.generateEcdsaKeyPair
+ */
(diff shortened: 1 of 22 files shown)


eddyashton (approved):

js/ccf-app/src/crypto.ts
```diff
@@ -27,6 +27,11 @@ export const generateAesKey = ccf.generateAesKey;
  */
 export const generateRsaKeyPair = ccf.generateRsaKeyPair;
 
+/**
```
jumaffre, line 30: This change requires a new entry in the changelog.

jumaffre (approved):

js/ccf-app/src/crypto.ts
achamayou: So it does.

PR #4276, opened 2022-09-27
Always enable time-based signature after first tx

jumaffre:
Follow-up from #4226 

~Draft for now as I'd like a few successful CI runs first, especially on SNP as the startup sequence is very async as we need to fetch endorsement certificates first.~

Now that the node startup sequence is even more async, it was possible that we started emitting time-based signatures before the first transaction was executed, which would result in a node crash as the node certificate needs to be first set before issuing signatures.


eddyashton (approved):

eddyashton:
Admin-merging - remaining failures are unrelated and pre-existing, and this reliably fixes some of the existing failures.

PR #4280, opened 2022-09-28
Expose validation failure details

achamayou:
Include the stack trace in the error message when `validate()` throws.

Before:

```
{"error":
  {"code":"InternalError",
    "message":"Failed to execute validation"}}
```

After:

```
{"error":
  {"code":"InternalError",
    "message":"Failed to execute validation: SyntaxError: expecting property name     at <input>:1\n    at parse (native)\n    at validate (public:ccf.gov.constitution[0]:1368)\n"}}
```


MahatiC (approved):

src/node/rpc/member_frontend.h
```diff
@@ -925,11 +925,14 @@ namespace ccf
 
         if (JS_IsException(val))
         {
-          js::js_dump_error(context);
+          auto [reason, trace] = js_error_message(context);
```
eddyashton, line 928: This is no longer printing the error to stdout. I think we still want that behaviour, as well as returning it to the caller, or is its removal deliberate?

achamayou: It's in line with what we do in the same circumstances for other failing governance calls (resolve, ballots and apply), I think this one had been missed. We could decide to print out to stdout as well, but I'm not sure this is justifiable because there are two main sources for these errors: bad user input, and bad governance code. In either case, not something the operator needs to, or can do very much about.

We persist errors in the KV for ballots, since they're naturally scoped to the lifetime of the proposal, but not for resolve or apply. I don't think we'd want to do that for validate, since we'd face the same lifetime/cleanup issue.

Perhaps there is a case for a separate governance log, of interest to the consortium? Or just a [gov] tag, and we start logging to that systematically (there is a slight annoyance with the fact that the errors often contain a multi-line intepreter trace, but that's not an issue with JSON logging).

eddyashton (approved):

PR #4288, opened 2022-09-30
Implement open ranges on the untyped map

### LINKED ISSUE #4238: Open ranges on the untyped map

There isn't an easy way to perform open range queries that can get all items less than a key or all items greater than or equal to a key with the current range interface:

https://github.com/microsoft/CCF/blob/d34518da64e0012796b855fa8443189495d18c28/include/ccf/kv/untyped_map_handle.h#L69-L71

A way to make either end open, with both being open meaning return every item.

Maybe:

```cpp
void range(
    const ElementVisitor& fn, 
    const KeyType& from, 
    const std::optional<std::reference_wrapper<KeyType>> to);
```

The `from` parameter should be open when the keytype is empty (e.g. an empty vector) but the `to` parameter needs an explicit option to specify its openness.

Making a big key and using that as the upper bound but it could always miss cases.

**Additional context**

https://github.com/microsoft/ccf-kvs/issues/23

jeffa5:
Fixes #4238


include/ccf/kv/untyped_map_handle.h
```diff
@@ -67,6 +67,8 @@ namespace kv::untyped
 
     // Sub-range variant of foreach. Visits the range of keys in [from, to).
     void range(
-      const ElementVisitor& fn, const KeyType& from, const KeyType& to);
+      const ElementVisitor& fn,
+      const std::optional<KeyType> from,
+      const std::optional<KeyType> to);
```
jumaffre, line 72:
```suggestion
      const std::optional<KeyType>& from,
      const std::optional<KeyType>& to = std::nullopt);
```

src/kv/untyped_map_handle.cpp
```diff
@@ -190,8 +190,8 @@ namespace kv::untyped
 
   void MapHandle::range(
     const MapHandle::ElementVisitor& f,
-    const MapHandle::KeyType& from,
-    const MapHandle::KeyType& to)
+    const std::optional<MapHandle::KeyType> from,
+    const std::optional<MapHandle::KeyType> to)
```
jumaffre, line 194:
```suggestion
    const std::optional<MapHandle::KeyType>& from,
    const std::optional<MapHandle::KeyType>& to = std::nullopt)
```

include/ccf/kv/untyped_map_handle.h
eddyashton: I think that if we have default `nullopt` for either, we should have default `nullopt` for both?

src/kv/untyped_map_handle.cpp
```diff
@@ -206,7 +206,9 @@ namespace kv::untyped
     // std::lower_bound()/std::upper_bound() and loop over it, interleaves
     // with the local writes.
 
-    if (from == to || to < from)
+    if (
+      from.has_value() && to.has_value() &&
+      (from.value() == to.value() || to.value() < from.value()))
```
jumaffre, line 211: Not sure why the condition was like this in the first place, but I think this line should be equivalent to `from.value() >= to.value() `? Also, what if `from = std::nullopt` but `to` has a value?

include/ccf/kv/untyped_map_handle.h
jeffa5: I think it could be confusing to read something like `range(map, "key1")` as the `key1` doesn't indicate whether the range starts or ends there whereas `range(map, "key1", std::nullopt)` gives a better idea of where the `key1` is used. Equally `range(map)` seems like some strangeness and something that should just compile down to `foreach` but won't and has an extra overhead.

src/kv/untyped_map_handle.cpp
jeffa5: Fixing the condition, not sure why it was like that. If `from = std::nullopt` and `to` has a value then this is an open ended range so we iterate from the first value up to (and excluding) `to`.

jeffa5: I don't think the condition bit will want to work nicely with `>=` so I'll leave it for another PR. Works fine currently still anyway.

include/ccf/kv/untyped_map_handle.h
achamayou: This is a perfect illustration of why named arguments are an excellent language feature.

achamayou (approved):

Issue #4256, opened 2022-09-21
Ability to have a function called once the transaction is committed to modify the response


jeffa5:
**Is your feature request related to a problem? Please describe.**

Currently there is no way to incorporate the transaction ID in the response since the commit only happens after we return.

**Describe the solution you'd like**

The ability to update the response with the transaction id, perhaps through another callback that gets registered.

Ideally this would not need to deserialise the response body again, instead using the existing one returned from the endpoint to avoid the extra overhead.

This could also tie in how the transaction id currently gets set as a header.

**Describe alternatives you've considered**

Guessing the transaction id which is unreliable.

eddyashton:
Rough plan for this is split into 2 sections.

#### Adding dev-specified execution after commit
- Modify `ccf::endpoints::Endpoint` to contain a second function instance. As well as the current `func` (which is roughly `void (RpcContext, AuthnIdentity, Tx)`), we add something like a `locally_committed_func` (taking just `(TxID, RpcContext)`? Perhaps `AuthnIdentity` too, just in case, but definitely _not_ `Tx`).
- Move the [current post-commit behaviour](https://github.com/microsoft/CCF/blob/main/src/node/rpc/frontend.h#L467-L471) (which just sets the `TxID` into a header value) into a function, and use that for the default value of `locally_committed_func`. This should probably be invoked by the same pattern of normal execution, like we do on line 410 (`endpoints.execute_endpoint(endpoint, args)`) - call a virtual function on the `Endpoints` object, whose default implementation just casts to a concrete `Endpoint` and invokes the functor. This will make it easier to support the same behaviour in JS generic/other applications.
- Add a new method to `ccf::endpoints::EndpointRegistry`, like `make_endpoint_function_with_commit_handler(method, verb, func, locally_committed_func, auth_policies)`, which maybe does some type-fiddling and stores the new `locally_committed_func`.
- Add a sample instance/test case doing something more interesting (eg - inserting the `TxID` into the response body, rather than a header)
- Ensure that error-handling around this new functor is sound. This lets app-writers throw or do other dangerous things later than they previously could, and specifically later than the `Tx` is committed. If their new functor explodes, we should catch that and produce some standard error response including a description of the resulting `TxID`.

#### Persisting some response state between the functors
- The above can use the current `RpcContext`, and assume that the new functors either set a header, or deserialise the body, insert their own fields, and overwrite with the new serialised body (ie - they work with the existing `set_response_body(vec)`).
[...]

PR #4291, opened 2022-09-30
Post commit execution: part 1

jeffa5:
This implements the first half of post commit execution logic, as described [here](https://github.com/microsoft/CCF/issues/4256#issuecomment-1259491788).

This allows users to execute logic (non-transaction related) after the transaction has (locally) committed. This allows custom handling of the txid header for example. This follows a similar pattern to the typical endpoint function handling but has extra safety about the failures from the post-commit function call since we can't abort the transaction. We currently respond with a 500 error and try to describe the failure message which may not be ideal.

The sample app implements safe and unsafe variants of each endpoint for a counter (increment, decrement, get value). The `_exception` paths trigger an exception in the post-commit logic and so return a 500 error. An interesting scenario to run with this is to trigger an `increment_exception` and then `value` to see that the increment still took place and committed despite the 500 error.


src/node/rpc/frontend.h
```diff
@@ -464,11 +464,33 @@ namespace ccf
...
+                  endpoints.execute_endpoint_locally_committed(
+                    endpoint, args, tx_id.value());
+                }
+                catch (std::exception& e)
+                {
+                  // run default handler to set transaction id in header
+                  ccf::endpoints::default_locally_committed_func(
+                    tx_id.value(), args);
+                  ctx->set_error(
+                    HTTP_STATUS_INTERNAL_SERVER_ERROR,
```
jeffa5, line 478: I've defaulted to a 500 for these errors as they are an internal error. I wonder if we could more subtly indicate to the user that something failed so that they can still see a success from the main logic and we just set the txid in the header. Maybe we could also set an error message in a header and otherwise leave the response alone?

e.g. `x-ms-ccf-locally-committed-error`: e.what()`

jeffa5: Currently this would work since the locally_commited function can't get data from the main function without sending it through the response. However, with part 2 of the work we'll allow them to pass arbitrary data which may mean the response isn't filled at all, potentially emitting a far less useful body. Not sure if that would still be useful as we wouldn't be too intrusive.

Mainly from a client perspective I'd be interested in seeing the success response code to know the transaction succeeded even if I can't parse the body properly, rather than a complete failure message that actually did something.

eddyashton: I think if the `committed_func` throws, we want to convert that into a HTTP error so its loud and obvious - ideally we'd enforce that at it doesn't throw at compile-time, but the next-best we can do is send a strong signal to the developer that it has thrown and they should fix that. Adding a header only in failing instances is likely to be missed (and potentially leave the rest of the response unparseable).

I see the appeal in trying to embed the original response inside this error, so a dedicated client _can_ still work out what happened, but we can't do that in general while maintaining the odata response body. I think this approach is fine - if it throws we discard the original response, and return a 500. Note that the app developer can still return something different if they wish - they can override `execute_endpoint_locally_committed`, have _that_ run a try-catch block, and return the original response/some encoding of it in the request, if they wish.

jeffa5:
> We should add an end-to-end test for this new app, showing that the response contains the TxID, and checking what happens when the new functor throws.

I've had a go at adding some basic tests that check both types of endpoints, that the tx id gets set to the header in the case of exception in the post-commit handler and that the error message gets returned as the json body with 500 error code.

src/endpoints/endpoint_registry.cpp
```diff
@@ -132,6 +132,16 @@ namespace ccf::endpoints
     return metrics[method][verb];
   }
 
+  void default_locally_committed_func(
+    CommandEndpointContext& ctx, const TxID& tx_id)
+  {
+    // Only transactions that acquired one or more map handles
+    // have a TxID, while others (e.g. unauthenticated commands)
+    // don't. Also, only report a TxID if the consensus is set, as
+    // the consensus is required to verify that a TxID is valid.
```
eddyashton, line 141: I think this comment should stay on its original location in `frontend.h`, since its describing the check _before_ `execute_endpoint_locally_committed` is called rather than the behaviour of this default implementation.

include/ccf/endpoint_registry.h
```diff
@@ -170,6 +173,30 @@ namespace ccf::endpoints
...
 
+    /** Create a new endpoint with a commit handler.
+     *
+     * Like make_endpoint but takes a functor to call once the transaction has
+     * been committed, but before consensus has completed.
+     *
+     * @param l Functor which will be invoked after the transaction has been
+     * assigned an ID.
+     */
+    virtual Endpoint make_endpoint_with_commit_handler(
```
eddyashton, line 184:
```suggestion
    virtual Endpoint make_endpoint_with_local_commit_handler(
```

I think these methods should all include `local_commit` in the name, rather than `commit`, to avoid confusion. One day we may find better terminology for local commit, as we've largely erased it from the public API, but I don't think we have a better alternative name here.

eddyashton (approved):

include/ccf/endpoint_registry.h
jeffa5: That does make it clearer, wasn't sure on the status of erasing the local/global from the API

CMakeLists.txt
```diff
@@ -728,6 +728,14 @@ if(BUILD_TESTS)
     )
   endif()
 
+  add_e2e_test(
+    NAME e2e_localcommithandler
+    PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/e2e_localcommithandler.py
+    CONSENSUS cft
+    CONTAINER_NODES TRUE
+    ADDITIONAL_ARGS -p "samples/apps/localcommithandler/liblocalcommithandler"
+  )
```
jumaffre, line 737: Is there a specific reason why this is a separate app/end-to-end test rather than fold it in the existing logging C++ app that demonstrates most (all?) C++ API calls?

tests/suite/test_suite.py
```diff
@@ -86,6 +87,9 @@
     e2e_logging.test_view_history,
     e2e_logging.test_tx_statuses,
     e2e_logging.test_random_receipts,
+    # e2e_localcommithandler
+    e2e_localcommithandler.test_safe,
+    e2e_localcommithandler.test_unsafe,
```
jumaffre, line 92: I think this is OK not to include in the full test suite as this is a new stateless (i.e. does not impact recovery, ledger, etc.).

CMakeLists.txt
jeffa5: No specific reason besides that I wasn't wanting to clog up the logging demo but happy to merge it if that's still desired.

jumaffre: My understanding is that the logging app is already clogged up as it demonstrates the entire C++ API rather than a minimal getting started sample. @eddyashton, what do you think?

eddyashton: Yeah let's roll this into the logging sample app for now.

jeffa5: Had a go at adding a new endpoint.

samples/apps/logging/logging.cpp
```diff
@@ -287,6 +314,76 @@ namespace loggingapp
...
+            std::move(error_reason));
+        }
+
+        LoggingPut::Out resp;
+        resp.success = true;
+        LoggingPut::Intermediate interm;
+        interm.out = resp;
+        if (fail == "true") {
+          interm.fail = true;
+        }else {
```
eddyashton, line 372: @jeffa5 Can you run clang-format on this code? `scripts/ci-checks.sh -f` will run all of the linters and formatters, and auto-fix what it can.

tests/suite/test_suite.py
```diff
@@ -2,6 +2,7 @@
 # Licensed under the Apache 2.0 License.
 
 import e2e_logging
+import e2e_localcommithandler
```
eddyashton, line 5: I think this can be removed?

jeffa5:
Seeing CI fail on `schema_test_cft` maybe due to https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=50637&view=logs&j=f3457e4b-102d-5a4f-f801-dc119e4853a8&t=99421a5f-0ead-59f3-a851-fa9ba80114c2&l=27969 or https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=50637&view=logs&j=f3457e4b-102d-5a4f-f801-dc119e4853a8&t=99421a5f-0ead-59f3-a851-fa9ba80114c2&l=27986 ? Though I'm not sure why these would be failing

eddyashton:
> Seeing CI fail on `schema_test_cft` maybe due to https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=50637&view=logs&j=f3457e4b-102d-5a4f-f801-dc119e4853a8&t=99421a5f-0ead-59f3-a851-fa9ba80114c2&l=27969 or https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=50637&view=logs&j=f3457e4b-102d-5a4f-f801-dc119e4853a8&t=99421a5f-0ead-59f3-a851-fa9ba80114c2&l=27986 ? Though I'm not sure why these would be failing

These will be failing because you've added an endpoint to `logging.cpp`, which ends up in `app_openapi.json`, and part of the schema test is trying to ensure we mark revisions to that file.

The fix:
- Bump `openapi_info.document_version` in `logging.cpp` (increasing minor version is fine for this change).
- Rebuild and run `schema_test_cft` locally. This will modify `app_openapi.json`. (The test will also fail this first time when it modified the file, but succeed on subsequent runs).
- Check in the modified `app_openapi.json`, which will include the version bump and describe the new endpoint.

eddyashton (approved):

PR #4296, opened 2022-09-30
Post commit execution: part 2

### LINKED ISSUE #4256: Ability to have a function called once the transaction is committed to modify the response

jeffa5:
Based on #4291 but can't target it as it is from my fork.

Fixes #4256


src/node/rpc/rpc_context_impl.h
```diff
@@ -16,6 +16,8 @@ namespace ccf
   protected:
     std::shared_ptr<SessionContext> session;
 
+    void* response_user_data;
```
eddyashton, line 19: The problem with a `void*` is that we can't delete this when the `RpcContext` goes out of scope (eg - the handler calls `set_response_user_data`, then throws). If we instead store this as a `std::shared_ptr<void>`, we get both lifetime management, and a per-instance deleter func, "for free". Example here: https://coliru.stacked-crooked.com/a/c03618118aad2854

samples/apps/local_commit_handler/localcommithandler.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+        auto current_value = counter_handle->get("counter").value_or(0);
+
+        Response resp;
+        resp.value = current_value;
+
+        ctx.rpc_ctx->set_response_status(HTTP_STATUS_OK);
+        ctx.rpc_ctx->set_response_header(
+          http::headers::CONTENT_TYPE, http::headervalues::contenttype::JSON);
+        ctx.rpc_ctx->set_response_user_data(static_cast<Response*>(&resp));
```
eddyashton, line 122: This won't transfer ownership of `resp`, which is just a local variable which will be deleted  at line 123. The `void*` stored in `rpc_ctx` is then pointing at junk - part of the stack that previously held `resp`.

This should instead be an owning pointer type so that we can extend the lifetime to the future call, and transfer ownership with this call. Either a `unique_ptr` or `shared_ptr`.

src/node/rpc/rpc_context_impl.h
jeffa5: Yep, that makes sense. I've updated it to be a `shared_ptr<void>`. Unfortunately we can't seem to have templating on the get/set functions, trying it I got linker errors presumably since the setter and getter can't be `virtual` anymore.

samples/apps/local_commit_handler/localcommithandler.cpp
jeffa5: Is there any best practice for `unique_ptr` vs `shared_ptr` here given that they can still just get the value back straight away?

Currently I've gone with a `unique_ptr` so that the rpc context actually owns it.

jeffa5:
This currently doesn't build as I get some errors to do with `invalid application of 'sizeof' to an incomplete type 'void'` which seems like it might be coming from the fact that we take a `unique_ptr<void>`. I'd appreciate some help getting this working if possible.

samples/apps/local_commit_handler/localcommithandler.cpp
eddyashton: No strong reason either way. I think while we expect single-ownership, it's simpler to make it a `unique_ptr`. Then iff we find a reason for it to be a `shared_ptr` before ownership is transferred, we can make it a `shared_ptr`.

jeffa5:
Updated to use a `shared_ptr` internally. Let me know if its too much `shared_ptr` now

samples/apps/logging/logging.cpp
```diff
@@ -318,10 +318,8 @@ namespace loggingapp
         static constexpr auto CCF_TX_ID = "x-ms-ccf-transaction-id";
         ctx.rpc_ctx->set_response_header(CCF_TX_ID, tx_id.to_str());
 
-        const nlohmann::json body_j =
-          nlohmann::json::parse(ctx.rpc_ctx->get_response_body());
-
-        auto interm = body_j.get<LoggingPut::Intermediate>();
+        auto interm =
+          *static_cast<LoggingPut::Intermediate*>(ctx.rpc_ctx->get_user_data());
```
jeffa5, line 322: @eddyashton Is there a nicer way to put this?

eddyashton: There's nothing much nicer - we've lost any type safety by casting to `void*`, so we need to `static_cast` _somewhere_. We could hide this if we were doing it a lot, but I think it's fine to be this obvious in a sample. We could also build a polymorphic type hierarchy for the values you're allowed to store as `user_data`, inheriting from a common type so we can `dynamic_cast` back and guarantee we've got the expected type, but I don't think that's worth it.

However, I think the return type should be `Intermediate*` rather than `Intermediate` (ie - remove the `*` before the `static_cast`, replace `interm.` with `interm->`). That way we can return a slightly clearer error if it ends up being null (when some path in the handler _doesn't_ call `set_user_data()`), which is a good thing to include in the sample implementation.

jumaffre (approved):
While I don't think it's adding this new feature to the documentation, could you add an entry to `CHANGELOG.md` so that users are aware of it in the next release?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - New "attestation" section in node JSON configuration to specify remote endpoint required to retrieve the endorsement certificates for SEV-SNP attestation report (#4277).
+- The `ccf::RpcContext` now contains functionality for storing user data with `set_user_data` and retrieving it with `get_user_data` (#4291).
```
achamayou, line 13: I'm tempted to suggest renaming to `set/get_ephemeral_user_data()` to make it abundantly clear that the persistence of this data is in no way guaranteed. But I may be worrying too much.

achamayou (approved):
Thank you for making this change, this opens interesting possibilities!

CHANGELOG.md
jeffa5: I'd think being on the rpc context would indicate that it won't stay around afterwards. I'm not too sure on which other name it should have either though. Is there a way to mark it experimental so that the naming could change later without being breaking?

achamayou: I've just looked at the set_* calls on the RpcContext, and I think you're absolutely right.

Issue #4292, opened 2022-09-30
OpenAPI prefixes app to all endpoints


jeffa5:
**Describe the bug**

The OpenAPI JSON available at `/api` from a running app prefixes all of the paths with `/app`. I would expect them not to after https://github.com/microsoft/CCF/issues/3893.

**To Reproduce**

```sh
# build ccf normally
cd build
../tests/sandbox/sandbox.sh -p samples/apps/logging/liblogging.virtual.so

# in another terminal window, from the build directory
curl --cacert workspace/sandbox_common/service_cert.pem --key workspace/sandbox_common/user0_privk.pem --cert workspace/sandbox_common/user0_cert.pem https://127.0.0.1:8000/api | jq '.paths | keys'
```

Observe how all of the paths are prefixed with `/app`.

**Expected behavior**

Endpoints that are registered without the `/app` prefix shouldn't have it applied in the OpenAPI output.

**Environment information**

CCF commit `a221771673f0eddccfff071ff63d79ec25bcc251`

jeffa5:
Fixed on main now.

PR #4298, opened 2022-10-03
Remove `app` prefix from API docs

MahatiC:
This is an attempt to work around the `lts_compatibility` test (see https://github.com/microsoft/CCF/pull/4168) and may end up resolving https://github.com/microsoft/CCF/issues/4292

### DIFF
--- a/include/ccf/endpoint.h
+++ b/include/ccf/endpoint.h
@@ -111,7 +111,7 @@ namespace ccf::endpoints
     /// Full URI path to endpoint, including method prefix
     URI full_uri_path;
 
-    /// URI path for the API documentation, that reflects the latest changes
+    /// URI path for the API documentation
     URI api_uri_path;
 
     EndpointProperties properties;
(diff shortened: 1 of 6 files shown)


include/ccf/endpoint.h
```diff
@@ -111,6 +111,9 @@ namespace ccf::endpoints
     /// Full URI path to endpoint, including method prefix
     URI full_uri_path;
 
+    /// URI path for the API documentation, that reflects the latest changes
+    URI api_uri_path;
```
achamayou, line 115: I'm not entirely clear on what "the latest changes" means in that context. Is this effectively the api uri path without the frontend/actor prefix?

MahatiC: Yes, that's correct. But that bit is unnecessary (and unclear) in the comment. Will remove.

include/ccf/endpoint.h
```diff
@@ -111,6 +111,9 @@ namespace ccf::endpoints
     /// Full URI path to endpoint, including method prefix
     URI full_uri_path;
 
+    /// URI path for the API documentation, that reflects the latest changes
```
MahatiC, line 114:
```suggestion
    /// URI path for the API documentation
```

achamayou (approved):

Issue #3298, opened 2021-12-10
Design notes: V8


letmaik:
*This is an issue for keeping a log of design notes for integrating V8 into CCF. New notes after the initial comment will become new comments. Everyone should feel free to comment along the way.*

### What is this about?
Integrating V8 into CCF, initially as additional app like `js_generic`, eventually integrated into the core.

### Why V8?
Battle-tested (browsers, server runtimes) and feature-rich (e.g. Wasm support).

### Is V8 faster than QuickJS?
According to https://bellard.org/quickjs/bench.html, yes. This is expected since QuickJS does not have a JIT compiler.
Our own experiments within virtual mode CCF in https://github.com/microsoft/CCF/pull/3258/ show that V8 is slower at the moment. This is likely related to the fact that typical benchmarks (edit: like in https://bellard.org/quickjs/bench.html) re-use both a single JS context and the loaded scripts/modules. In those conditions, the JIT kicks in, especially when the benchmark doesn't just run for a few microseconds. In CCF each request gets a fresh JS context for security reasons, which means there is overhead in context creation and also in re-loading/evaluating modules. The overheads especially from context creation seem to be lower in QuickJS, making it faster overall. Even if context creation was equally fast in V8 though, we wouldn't benefit from the JIT since jitted code is not re-used across contexts.
We have some ideas on safely re-using a context in V8 which would avoid the performance issues described above. This will be explored in the near future.

### Does it work in SGX / OE?
Probably, but from experiments we know V8 requires some patches (so far: cpuid, time, semaphores, mmap, dynamic queries for page size) to deal with limitations of OE. So far, we have not completed this work and OE is not supported yet.

### Can it run in virtual mode for now?
Yes, virtual mode is working and has been implemented as an app like `js_generic` (QuickJS), currently called `js_v8`.

### Does governance run through V8?
Not yet. This would be premature and not possible generally given that OE support is not working yet. 

### Can apps still be run with `js_generic`?
Yes, `js_v8` is strictly experimental and optional at this point. `js_v8` will likely be included in CCF releases as additional (virtual) enclave image with a disclaimer that it is for experimentation only.

### Should V8 eventually replace QuickJS?
Yes. Maintaining two JS runtimes as (build/runtime) options requires too much maintenance effort.

### V8 is huge, does this make build times long?
[...]

PR #4307, opened 2022-10-05
Disable remaining V8-app builds in CI

achamayou:
Part of the v8 removal, motivated by https://github.com/microsoft/CCF/issues/3298#issuecomment-1268060991


MahatiC (approved):
Is the plan also to remove v8-app from the repo? If so, maybe we should consider archiving it?

achamayou:
@MahatiC yes it is. It will live on in git history.

PR #4309, opened 2022-10-05
Remove 'Enclave' naming for AttestationProvider

MahatiC:
... as it now supports attestation for both SGX and SEV-SNP


include/ccf/node/quote.h
```diff
@@ -19,12 +19,12 @@ namespace ccf
     FailedInvalidQuotedPublicKey,
   };
 
-  class EnclaveAttestationProvider
```
eddyashton, line 22: This is part of the public API (under `include/ccf`), so renaming is a breaking change that needs a CHANGELOG entry. I think we should clarify the comment to make the behaviour across platforms clearer, but don't think it's worth a breaking change in the name.

MahatiC: hmm, for now I could revert the name change and just put a comment. But I think given it's a public API, it's worth making it clearer at some point. @achamayou thoughts?

achamayou: I had always thought of an SEV-SNP VM as being an enclave, but looking at https://www.amd.com/system/files/TechDocs/SEV-SNP-strengthening-vm-isolation-with-integrity-protection-and-more.pdf again, you're right that AMD doesn't call them enclaves.

I don't mind the change, but as Eddy says, we need at least a changelog entry because this is user-visible. A slightly softer variant would be to define the old name, deprecate it, and put a comment that says it will go away in 4.x. We've done this before, for example in #4295.

eddyashton: +1 for renaming but keeping a deprecated instance of the old name to soften the breaking change. Should be as simple as adding something like `CCF_DEPRECATED(reason, will be removed in 4.x) using EnclaveAttestationProvider = AttestationProvider;`

MahatiC: Thanks for your inputs, I made the change

PR #4310, opened 2022-10-05
Remove 1.x configuration migration script

jumaffre:
No longer needed on `main` (3.x) branch.


samples/config/minimal_config.json
```diff
@@ -8,9 +8,7 @@
     "rpc_interfaces": {
       "interface_name": {
         "bind_address": "127.0.0.1:8080",
-        "published_address": "ccf.dummy.com:12345",
-        "max_open_sessions_soft": 1000,
-        "max_open_sessions_hard": 1010
```
jumaffre, line 13: Unrelated to the primary aim of this PR but these aren't necessary by default so removing them from the minimal configuration.

achamayou (approved):

PR #4311, opened 2022-10-05
Fix CodeQL complaints about scanf

wintersteiger:


wintersteiger:
I confirmed offline that this makes CodeQL happy, so I expect this to automatically remove the code scanning alerts.

eddyashton (approved):

PR #4313, opened 2022-10-06
Fix implementation of `Math.random()` in JS interpreter

eddyashton:
Our stubs for QuickJS's `gettimeofday` functions weren't clearing the passed value, so were relying on uninitialized data.

Beyond that, we don't actually want to rely on `gettimeofday` to seed the random state, which is the implementation of `Math.random()` that QuickJS uses. Instead we overwrite this function entirely with our own implementation, which uses our existing in-enclave `Entropy`.

Includes an end-to-end test that tries to ensure we don't see the same sequence of numbers from this, across calls or network restarts.

### DIFF
--- a/tests/js-custom-authorization/custom_authorization.py
+++ b/tests/js-custom-authorization/custom_authorization.py
@@ -379,6 +379,34 @@ def run_content_types(args):
 
 
 @reqs.description("Test request object API")
+def test_random_api(args):
+    # Test that Math.random() does not repeat values:
+    # - on multiple invocations within a single request
+    # - on multiple requests
+    # - on network restarts
+
+    seen_values = set()
+
+    def assert_fresh(n):
+        assert n not in seen_values, f"{n} has already been returned"
+        seen_values.add(n)
+
+    n_repeats = 3
+    for _ in range(n_repeats):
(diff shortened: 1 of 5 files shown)


tests/js-custom-authorization/custom_authorization.py
```diff
@@ -378,6 +378,34 @@ def run_content_types(args):
...
+    def assert_fresh(n):
+        assert n not in seen_values, f"{n} has already been returned"
+        seen_values.add(n)
+
+    n_repeats = 3
+    for _ in range(n_repeats):
+        with infra.network.network(
+            args.nodes, args.binary_dir, args.debug_nodes, args.perf_nodes, pdb=args.pdb
+        ) as network:
+            network.start_and_open(args)
```
jumaffre, line 399: Restarting an network seems expensive - could we simply run it on different nodes in the initial service instead?

eddyashton: This is expensive, as I want to ensure we're as pessimistic as possible. The test still only takes ~13s, so I think this is fine? Testing across multiple nodes is also a good idea though - I'll add that.

Issue #4219, opened 2022-09-12
Multi-Container: Forwarding from within a handler


eddyashton:
Follow-up to #4157, as this missed the remaining tricky bit: Once we've stolen the `Tx` so it could be used by future requests, how do we: `(a)` mark this user request as pending, `(b)` commit (locally) the final `Tx`, and `(c)` return the HTTP response to the calling user.

`a` should be pretty easy, now that we have `response_is_pending` on the RPC context. `b` and `c` need some thought. They're different from the current forwarding because there we use the entire existing execution stack on the executing node, and don't have any additional processing on the forwarder. For external execution, we need to decide whether `EndTx` commits, or something else does afterwards, and how we reuse the existing commit (+retry!) logic to build the serialised response (ie - inserts the `TxID` header) for the caller.

I've been trying to draw some diagrams to explain this, abusing the `gitGraph` in Mermaid. Not sure if they're useful, but here they are:

Normal CCF transactions are assigned their own independent `Tx`, and use that for all KV interaction. Once the handler has completed, the framework tries to apply (commit) this `Tx` to the local KV, and then constructs a combined response.

![normal_tx_flow](https://user-images.githubusercontent.com/6000239/189667259-648be8c8-3783-48d9-ab82-049ecb6edc84.png)

With the `external_executor` model it's more complicated. When a user request comes in, its `Tx` object is persisted, to be used for executing multiple future RPC calls. We've worked out the first half of that, which is how to steal and persist a `Tx`, and understand how that will be looked up for the future RPC calls. The remaining question is how and where we eventually call `commit()` on that `Tx` object, and paste the result into a HTTP response (combined with the response described by the call to `EndTx`), and transmit that HTTP response to the original caller.

The following 2 diagrams show 2 places that might be done. A simple approach is to do it inside `EndTx` - explicitly calling `tx->commit()` and replicating some of the logic from the `Frontend`, and then calling something like `forward_response->respond_async()` to actually respond to the client.
![tx_ownership_a](https://user-images.githubusercontent.com/6000239/189669405-71e86de3-682b-435c-91a6-1ddfa689813f.png)

Another approach, which is how we'd design this if we had asynchrony in the enclave, is to consider the original request paused/suspended rather than forwarded, and have a way to join/reconnect to it inside the handler (with the remaining block of the handler just fetching the pending response and `Tx`, and using them to replace its own), so that commit and response construction were handled by existing code.
[...]

PR #4316, opened 2022-10-06
[Draft, do-not-merge] Send response from within application code

eddyashton:
**_Extremely DRAFT, just opening for an easy diff view while I break out some of the standalone changes into separate PRs. This will likely remain a monster, but hopefully (eventually) an explicable one._**

Implementation of #4219.

Headlines:
- Loads of renames in the `http::` namespace to clarify concepts, and avoid reuse of `Endpoint`. The main one is that `Endpoint` becomes `Session`
- HTTP/2 implementation squeezed closer to HTTP/1.1, by separating a `Parser` object from the `Session`, and moving all the callbacks to a separate file
- Making `tls::Session` a member-of rather than base-class-of the `HTTPSession` and `HTTP2Session`. Simplifies the class hierarchy, though it remains complex, and opens the door for a plain-text IO layer for all sessions
- Broader use of `std::span<const uint8_t>`. I think this removes a bunch of unnecessary copies as we switch between moved vectors and raw pointers, but it enforces a single eventual copy. I dunno, I think its simpler to read regardless

The core difficulty, which I hadn't grokked before I was deep into the weeds, is the need for additional info in lookup -between a HTTP/1.1 session which can be looked up or referenced by a single `ConnID`, and a HTTP/2 session which sometimes needs an additional `StreamId`. In particular to respond to an actual request, we need a `StreamId`, but for early parsing errors/parity with HTTP/1.1, we may not have one. I think the right approach here is to break the concepts of a lifetime-managed TLS socket, from a writeable/readable request/response stream. This is 1-to-1 in HTTP/1.1, but 1-to-many in HTTP/2. This is partially done in this PR, with multiple `HTTPResponder`s per `HTTP2ServerSession`, but this is incomplete.


jumaffre (commented):
Could you touch the `.daily_canary` as HTTP/2 tests are mostly run in the daily currently?

tests/external_executor/external_executor.py
```diff
@@ -166,9 +172,9 @@ def test_simple_executor(network, credentials, args):
 
     with executor_thread(WikiCacherExecutor(primary, credentials)):
         with primary.client() as c:
-            c.post("/not/a/real/endpoint")
-            c.post("/update_cache/Earth")
-            c.get("/article_description/Earth")
+            c.post("/not/a/real/endpoint", timeout=2)
+            c.post("/update_cache/Earth", timeout=2)
+            c.get("/article_description/Earth", timeout=2)
```
MahatiC, line 177: Not sure I understand why this timeout is added?

MahatiC (commented):
Most are rename suggestions. For some context, when variable names are single/double lettered and (for example) are declared in line 1, one can quickly forget what that represents in line 20 or so. Sometimes `F12` in VSCode also becomes quite tricky to do it on one or two lettered variables.

tests/external_executor/external_executor.py
eddyashton: Good spot! These were added as part of debugging, but are unneeded now - I've removed them.

src/apps/external_executor/external_executor.cpp
```diff
@@ -35,11 +38,17 @@ namespace externalexecutor
...
-    // Note: As a temporary solution for testing, this app stores a single Tx,
-    // stolen from a StartTx RPC rather than a real client request
-    std::unique_ptr<kv::CommittableTx> active_tx = nullptr;
+    struct PendingRequest
+    {
+      std::unique_ptr<kv::CommittableTx> tx = nullptr;
+      externalexecutor::protobuf::RequestDescription request_description;
+      std::shared_ptr<http::HTTPResponder> responder;
+    };
+    std::queue<PendingRequest> pending_requests;
```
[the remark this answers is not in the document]
eddyashton: I've opted not to rename this one, as I think the fact this is a queue is just an implementation detail. As far as most of the code is concerned it's some kind of container for adding and fetching `PendingRequest`s. The actual type and details may change in future.

eddyashton:
> Most are rename suggestions. For some context, when variable names are single/double lettered and (for example) are declared in line 1, one can quickly forget what that represents in line 20 or so. Sometimes `F12` in VSCode also becomes quite tricky to do it on one or two lettered variables.
> 
> Otherwise, nice work on this change!

@MahatiC Thanks, I've renamed the short variables as suggested :)

src/apps/external_executor/external_executor.cpp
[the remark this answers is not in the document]
MahatiC: Sure, makes sense. The only reason I think I suggested a rename here is because elsewhere I suggested `pending_request` and was trying to differentiate this one more explicitly.

tests/external_executor/external_executor.py
```diff
@@ -38,10 +38,16 @@
...
 
 @contextlib.contextmanager
-def wrap_tx(stub):
-    stub.StartTx(Empty())
-    yield stub
-    stub.EndTx(KV.ResponseDescription())
+def wrap_tx(stub, primary):
+    with primary.client(connection_timeout=0.1) as c:
+        try:
+            c.get("/placeholder", timeout=0.1)
```
MahatiC, line 44: Was this for debugging? Don't think there is an endpoint like this?

eddyashton: This call is still necessary. The executor can't do any KV `put`s or `get`s unless the node it's speaking to thinks that executor is currently processing a request. So to retain the existing tests where we probe the KV API directly (ie - outside of trying to implement a _real_ executor), we need to send a dummy request to the node (this one), such that the `StartTx()` we call after returns a real `RequestDescription`.

This is another temporary situation - this low level test may not be needed shortly, or may be moved to a separate test that's just testing gRPC access on another app. I'll add a comment in the code explaining this.

achamayou (approved):

MahatiC (approved):

Issue #4267, opened 2022-09-26
How to detect deletes in the index building


jeffa5:
**Is your feature request related to a problem? Please describe.**

When building an index with the [handle_committed_transaction](https://microsoft.github.io/CCF/main/build_apps/api.html#_CPPv4N3ccf8indexing8Strategy28handle_committed_transactionERKN3ccf4TxIDERKN2kv16ReadOnlyStorePtrE) we can iterate through the writes in that transaction but I can't see a way to detect what was deleted. Without this information we can't keep our data up to date accurately.

**Describe the solution you'd like**

A method or field on the `StorePtr` that returns an iterator or list of the keys that were deleted in this transaction.

**Describe alternatives you've considered**

Changing the application to model deletes with a write instead of performing an actual delete. This has the downside of not being so explicit and adds extra logic to ignore the tombstones in the main request handling logic.

jumaffre:
This was probably introduced recently, in #4145, as it was possible to identify a deleted key (negative version), hence I think this is a bug? I'm not sure how we can fix this unless we re-introduce the old tombstone behaviour or implement https://github.com/microsoft/CCF/issues/3197 so that we can detect whether a key was deleted in a specific indexing interval? @eddyashton will be able to answer.

eddyashton:
We are still tracking deletes in the per-transaction data that we're using, but we throw that away when we deserialise to a `kv::Store`. I think the real outcome here is that re-using the `kv::Store` and `kv::Tx` types for historical queries and indexes which are only seeing the results of a single transaction isn't actually helpful. We thought it was useful to provide a consistent API, but it's actually misleading (suggests we're operating over a projected KV state rather than a single transaction's write set), and conflates different use cases (as this issue notes, deliberately abstracting deletions vs wanting to know they happened).

So we should probably be providing something like the hook API for both of these systems. We can build a stubby `ReadOnlyTx` on top of that to maintain the current API.

jeffa5:
Having looked through the code a bit I think I'd plan to add this through the following changes:

- Create a new `TxDiff` type that is similar to a `ReadOnlyTx` but wraps an `untyped::ChangeSet` directly (or perhaps just the `untyped::Writes` from within that, if that is all we need)
[...]

PR #4335, opened 2022-10-10
Add user hooks for global commit

jeffa5:
This exposes the global commit hook that already existed (called in `post_commit`) on the maps. This also makes the following changes:
-  Add a new subsystem `UserHooks` to register the hooks on maps
- Add an example in an endpoint on the logging app
- Distinguish user and system hooks. This prevents users overriding hooks that perform integral operations for CCF

Happy for other names, potentially `HookRegistry` if we don't want the `User` part?

Relates to https://github.com/microsoft/CCF/issues/4267


achamayou (changes requested):
I am not comfortable with this change as is, because:

1. Consensus in the team around hooks so far has been that they should not be exposed to users, since they are so easily mis-used. This view has been reinforced after the public Indexer API was added. Could you explain what purpose this change serves in your application that the Indexer does not satisfy?
2. The regression tests indicate that this change in its current form has a severe performance impact.

jeffa5:
This is trying to solve this issue: https://github.com/microsoft/CCF/issues/4267 through hooks rather than reworking the indexing logic and API.

I agree that it seems the indexing API can also easily be misused as it doesn't propagate deletes and conflates concepts with Txs and maps etc.

Re performance: I've just updated the `post_compact` method to only iterate through `commit_deltas` when there is a hook set to execute.

PR #4336, opened 2022-10-10
Add endpoint to display all open proposals

MahatiC:
Resolves https://github.com/microsoft/CCF/issues/4333


src/node/rpc/member_frontend.h
```diff
@@ -1024,6 +1024,33 @@ namespace ccf
         .set_auto_schema<jsgov::Proposal, jsgov::ProposalInfoSummary>()
         .install();
 
+      using ProposalID = ccf::MemberId;
+      using AllOpenProposals = std::map<ProposalID, jsgov::ProposalInfo>;
```
eddyashton, line 1028:
```suggestion
      using AllOpenProposals = std::map<ProposalId, jsgov::ProposalInfo>;
```

This type already exists. Reusing `MemberId` results in an inaccurate OpenAPI (claiming the map is `MemberId_to_ProposalInfo`). If we want more precise OpenAPI describing a `ProposalId` (ie - constraining it to a fixed length hex string, rather than an arbitrary string), we should replace the existing typedefs of `ProposalId` to use `EntityId` or similar.

src/node/rpc/member_frontend.h
```diff
@@ -1024,6 +1024,33 @@ namespace ccf
...
+          auto proposal_info = ctx.tx.ro<ccf::jsgov::ProposalInfoMap>(
+            jsgov::Tables::PROPOSALS_INFO);
+          AllOpenProposals response;
+          proposal_info->foreach(
+            [&response](
+              const ProposalId& pid, const ccf::jsgov::ProposalInfo& pinfo) {
+              if (pinfo.state == ProposalState::OPEN)
+              {
+                response[pid] = pinfo;
+              }
```
eddyashton, line 1040: Why are we only including `OPEN` proposals here?

src/node/rpc/member_frontend.h
MahatiC: Yeah, that is the only reason I marked it as Draft. I should have a left a comment here. My bad.

> we should replace the existing typedefs of ProposalId to use EntityId or similar.

I was planning on doing this as well

src/node/rpc/member_frontend.h
eddyashton: @achamayou Can we clarify the ask here? I was assuming we want an endpoint to show _all_ proposals. The problem with the existing `/proposals/{id}` endpoint is that if you forget a proposal ID, it's essentially lost and unfindable. That seems just as true for proposals which are rejected or failed as those which are open. So we should have _a_ way to return all proposals, with maybe a filter to return only `Open` proposals if needed (but I think that's easier to do on the client side).

achamayou: The initial intention was to do this for open proposals, precisely for the reason you have highlighted. We could also display _all_ proposals, but the result wouldn't necessarily be what a user expects. This is because we automatically clean up rejected or failed proposals on the next proposal being posted, to avoid infinite growth.

So the result would always either be open proposals, or open proposals plus the most recent failed/rejected proposal. To me, that does not feel particularly intuitive, but I'm happy to get other perspectives.

I think there is a separate use case for historically querying past proposals and their status, but that's out of scope for this ticket.

achamayou (approved):

eddyashton (approved):

PR #4343, opened 2022-10-13
Add trusted authority role sample constitution 

DomAyre:


tests/infra/network.py
```diff
@@ -433,13 +433,16 @@ def start(self, args, **kwargs):
         mc = max(1, args.initial_member_count)
         initial_members_info = []
         for i in range(mc):
+            member_data = None
+            if i < args.initial_trusted_authority_count:
+                member_data = {"is_trusted_authority": True}
+            elif i < args.initial_trusted_authority_count + args.initial_operator_count:
+                member_data = {"is_operator": True}
```
achamayou, line 440: Do we have a check somewhere that `mc > (initial_trusted_authority_count + initial_operator_count)` ?

achamayou:
Not sure about "trusted_authority" as a name for these members, it's very generic. In some sense every member is a TA, since we accept signed statements from them. I think "operator_provisioning_authority" would more clearly convey that this member is specifically allow to (re-)provision operators.

samples/constitutions/trusted_authority/resolve.js
```diff
@@ -0,0 +1,81 @@
...
+  // Operators cannot be recovery members.
+  if (isRecoveryMember(memberId)) {
+    return false;
+  }
+  return getMemberInfo(memberId).member_data?.is_operator ?? false;
+}
+
+// Defines which of the members are trusted authorities.
+function isTrustedAuthority(memberId) {
+  return getMemberInfo(memberId).member_data?.is_trusted_authority ?? false;
```
achamayou, line 26: Operator provisioning authorities also should not be recovery members (see line 18).

samples/constitutions/trusted_authority/resolve.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Defines actions that can be passed with sole trusted authority input.
+function canTrustedAuthorityPass(action) {
+  // Some actions can always be called by trusted authorities.
+  const allowedTrustedAuthorityActions = ["trust_node", "retire_node"];
```
achamayou, line 32: That's not right, a OPA cannot trust or retire nodes, it can only set/remove operator members.

tests/infra/network.py
DomAyre: We do as of https://github.com/microsoft/CCF/pull/4343/commits/f2cae12a91725da39d1a4d0d60ea6bb7fe45f6ef

DomAyre:
@achamayou I've renamed trusted authorities to operator provisioner https://github.com/microsoft/CCF/pull/4343/commits/1dbd4d2f5329149f890c6e328c2fe735a0434857

samples/constitutions/trusted_authority/resolve.js
```diff
@@ -0,0 +1,81 @@
...
+  if (allowedTrustedAuthorityActions.includes(action.name)) {
+    return true;
+  }
+  // Trusted authorities can add or retire operators.
+  return (
+    {
+      set_member_data: () => action.args["member_data"]?.is_operator ?? false,
```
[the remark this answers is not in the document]
DomAyre: https://github.com/microsoft/CCF/pull/4343/commits/c2965ec512245b4ab465b9b1de5a87e38bfa66ca

tests/governance_js.py
```diff
@@ -432,6 +432,41 @@ def test_operator_proposals_and_votes(network, args):
...
+        r = c.post("/gov/proposals", set_operator)
+        assert r.status_code == 200, r.body.text()
+        assert r.body.json()["state"] == "Accepted", r.body.json()
+
+    # Reset the member to an operator for other tests
+    network.consortium.set_member_data(
+        node,
+        operator_provisioner.service_id,
+        member_data={"is_operator": True},
+    )
```
achamayou, line 467: Can we have a negative tests where the operator provisioner tries to pass something that's not adding (or removing, which is missing) an operator, and confirm that it doesn't pass automatically?

samples/constitutions/trusted_authority/resolve.js
```diff
@@ -0,0 +1,78 @@
...
+  );
+}
+
+export function resolve(proposal, proposerId, votes) {
+  const actions = JSON.parse(proposal)["actions"];
+
+  // If the node is an operator provisioner, strictly enforce what proposals it can
+  // make
+  if (isOperatorProvisioner(proposer_id)) {
+    return actions.every(canOperatorProvisionerPass) ? "Accepted" : "Rejected";
```
achamayou, line 49: It could be tempting to be paranoid about this, and check that the proposal contains exactly one of each.

I wonder in particular what would happen with:

```
set_member foo
remove_member bar
remove_member foo
```

In a system with an operator bar to start with. I suspect the net result is a system without an operator, which doesn't seem ideal!

DomAyre: Are we thinking that the provisioner is only for the disaster recovery? Would it be responsible for provisioning the very first operator (in which case I guess we would need to allow either just an add and maybe just a remove?)

We could certainly enforce provisioners can only do one in, one out actions, but if not, I guess we can't prevent a provisioner removing a member it didn't set it?

jumaffre (commented):
Could this new sample be briefly discuss in https://microsoft.github.io/CCF/main/governance/constitution.html please?

samples/constitutions/test/resolve.js
```diff
@@ -1,3 +1,43 @@
...
+function isOperatorProvisioner(memberId) {
+  return getMemberInfo(memberId).member_data?.is_operator_provisioner ?? false;
+}
+
+// Defines actions that can be passed with sole operator provisioner input.
+function canOperatorProvisionerPass(action) {
+  // Operator provisioners can add or retire operators.
+  return (
+    {
+      set_member_data: () => action.args["member_data"]?.is_operator ?? false,
```
jumaffre, line 34: This is different from the other `resolve.js` file - is that on purpose?

tests/governance_js.py
DomAyre: Added the negative test and expanded to include coverage of `remove_member` in https://github.com/microsoft/CCF/pull/4343/commits/5223abe0d0d4449667145610fae7762f9daf688b

samples/constitutions/test/resolve.js
DomAyre: I originally thought it was okay to include but it wasn't in the end, so removed in https://github.com/microsoft/CCF/pull/4343/commits/5223abe0d0d4449667145610fae7762f9daf688b

DomAyre:
@jumaffre Done as part of https://github.com/microsoft/CCF/pull/4343/commits/de0eacec1e8bdcc619ff626c653dc2db0ea986ad

samples/constitutions/trusted_authority/resolve.js
achamayou: The initial provisioning is done through config, and effectively signed off by the service open proposal.

PR #4344, opened 2022-10-13
Jsruntime memory limit

MahatiC:
Part of #1459


samples/constitutions/default/actions.js
```diff
@@ -762,6 +762,25 @@ const actions = new Map([
...
+        checkType(args.max_heap_size, "integer", "max_heap_size");
+        checkType(args.max_stack_size, "integer", "max_stack_size");
+      },
+      function (args) {
+        const js_engine_map = ccf.kv["public:ccf.gov.jsengine"];
+        const heap_size = args.max_heap_size;
+        const stack_size = args.max_stack_size;
+        const heap_name = ccf.strToBuf("max_heap_size");
+        const stack_name = ccf.strToBuf("max_stack_size");
+        js_engine_map.set(heap_name, ccf.jsonCompatibleToBuf(heap_size));
```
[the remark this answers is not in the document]
MahatiC: This looks ok to me

MahatiC:
I tested this in a hacky way using logging in cpp code and a small test in python infra that submits this new proposal. It does reflect the new values between runtimes after submitting the proposal. I could perhaps write a function to expose these values to python infra, but I'm not sure if it's worth it. If anyone feels strongly about this, please let me know.

samples/constitutions/default/actions.js
```diff
@@ -762,6 +762,25 @@ const actions = new Map([
       }
     ),
   ],
+  [
+    "set_js_engine_options",
+    new Action(
+      function (args) {
+        checkType(args.max_heap_size, "integer", "max_heap_size");
+        checkType(args.max_stack_size, "integer", "max_stack_size");
```
achamayou, line 770: A convention we've adopted in the configuration is to suffix the unit name, eg. _days, _count. We do use _size in places where we allow strings and interpret the suffix instead, eg. max_msg_size: "4MB".

Here I would suggest suggest either max_heap_bytes, int, or max_heap_size, string, allowing "4MB" etc.

achamayou (approved):

src/js/wrap.cpp
```diff
@@ -1700,6 +1732,12 @@ namespace ccf::js
         ctx, js_is_valid_x509_cert_chain, "isValidX509CertChain", 2));
     JS_SetPropertyStr(
       ctx, ccf, "pemToId", JS_NewCFunction(ctx, js_pem_to_id, "pemToId", 1));
+    JS_SetPropertyStr(
+      ctx,
+      ccf,
+      "updateJSruntimememory",
```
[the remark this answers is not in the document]
MahatiC: Done

src/js/wrap.cpp
```diff
@@ -31,6 +31,8 @@ namespace ccf::js
 
   using KVMap = kv::untyped::Map;
 
+  size_t Runtime::max_stack_size_ = 1024 * 1024;
+  size_t Runtime::max_heap_size_ = 100 * 1024 * 1024;
```
jumaffre, line 35: Could we define the default values as `constexpr` in this file?

samples/constitutions/default/actions.js
```diff
@@ -762,6 +762,25 @@ const actions = new Map([
...
+        checkType(args.max_stack_bytes, "integer", "max_stack_bytes");
+      },
+      function (args) {
+        const js_engine_map = ccf.kv["public:ccf.gov.jsengine"];
+        const heap_bytes = args.max_heap_bytes;
+        const stack_bytes = args.max_stack_bytes;
+        const heap_name = ccf.strToBuf("max_heap_bytes");
+        const stack_name = ccf.strToBuf("max_stack_bytes");
+        js_engine_map.set(heap_name, ccf.jsonCompatibleToBuf(heap_bytes));
+        js_engine_map.set(stack_name, ccf.jsonCompatibleToBuf(stack_bytes));
```
jumaffre, line 779: We write these values to the new `jsengine` KV map but this isn't actually enforced by the fact that these are written to the map but via a different call to `updateJSRuntimeMemory` below, which is a little odd. 

Usually, values written to the KV that have side effects on the node are picked up via a KV hook (in this case, global). Otherwise, this means that this value is only picked up by the node executing the proposal (here, the primary node).

src/js/wrap.cpp
```diff
@@ -1277,6 +1279,36 @@ namespace ccf::js
...
+    if (JS_ToInt64(ctx, &stack_size, argv[1]) < 0)
+    {
+      return JS_EXCEPTION;
+    }
+
+    size_t heap = (size_t)heap_size;
+    size_t stack = (size_t)stack_size;
+
+    Runtime::max_heap_size_ = heap;
+    Runtime::max_stack_size_ = stack;
```
jumaffre, line 1307: This doesn't seem thread safe? If the proposal is executed by a governance proposal in thread `t1`, a new JS runtime may be concurrently created by thread `t2`.

samples/constitutions/default/actions.js
```diff
@@ -762,6 +762,25 @@ const actions = new Map([
...
   ],
+  [
+    "set_js_engine_options",
+    new Action(
+      function (args) {
+        checkType(args.max_heap_bytes, "integer", "max_heap_bytes");
+        checkType(args.max_stack_bytes, "integer", "max_stack_bytes");
+      },
+      function (args) {
+        const js_engine_map = ccf.kv["public:ccf.gov.jsengine"];
```
jumaffre, line 773: If this PR creates a new KV map (which I think it does?), it should be documented here: https://microsoft.github.io/CCF/main/audit/builtin_maps.html

samples/constitutions/default/actions.js
achamayou: The simplest thing to do would be to read them from the KV at every Runtime instantiation. We already do that for the code/bytecode, we could do that cheaply for those values too.

Hooks-based things are hard to make correct and transactional, and I don't think we especially want to wait for a global hook on this.

MahatiC: I just noticed this. (I should have better GitHub notifications?!)

You're right, this change indeed effects only the current executing node. I also need to update the js_generic code as well.

MahatiC: Done

src/js/wrap.cpp
MahatiC: Done

samples/constitutions/default/actions.js
MahatiC: Done

include/ccf/service/tables/jsengine.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
+
+#include "ccf/service/map.h"
+
+namespace ccf
+{
+  using JSEngine = ServiceMap<std::string, size_t>;
```
jumaffre, line 9: I think this should be a `ServiceValue` table (i.e. a table with only one key), similar to the service configuration (`config.h`).

MahatiC: It stores two keys though?

MahatiC: @jumaffre I notice that in tables like `config.h`, we do store multiple key values in a struct form. Is there a convention to this? As in when to store them as separate keys vs storing them all in a single struct form?

achamayou: There isn't a convention, it's down to two things:

1. Are the values the same type. If so, there is no downside to storing them under separate keys.
2. Are they ever/often updated separately? If they are, they really shouldn't be in the same value, that will conflict, and require a whole struct parse/serialisation every time for a single key.

For infrequently updates values, 2. isn't a factor, and future extension may put 1. at risk. But then, it's always possible to create another table for the new values.

MahatiC: @achamayou Thank you! Given what you said, since both values are of same type, and also that there is a possibility that only one of them is updated (although in the current proposal, it is enforced that both values must be submitted, the proposal itself can contain an update to only one value) imo, it looks ok to use `ServiceMap` in this case. 

But what is the downside if both values are of different types though? Does the KV map expect storing/parsing only similar type of values?

jumaffre (approved):

doc/audit/builtin_maps.rst
```diff
@@ -257,6 +257,18 @@ JavaScript engine version of the module cache, accessible by JavaScript endpoint
 
 **Value** QuickJS version as a string.
 
+``jsengine``
+~~~~~~~~~~~~
+QuickJS runtime memory, accessible by Javascript endpoint function.
+
+**Key** Sentinel value 0, represented as a little-endian 64-bit unsigned integer.
+
+**Value**
```
achamayou, line 266:
```suggestion
**Value** Represented as JSON
```

doc/audit/builtin_maps.rst
```diff
@@ -257,6 +257,18 @@ JavaScript engine version of the module cache, accessible by JavaScript endpoint
 
 **Value** QuickJS version as a string.
 
+``jsengine``
```
achamayou, line 260: We seem to use the term "runtime" everywhere else (doc, type names), can this perhaps be "runtime" too, or is there a specific reason it's "engine" instead?

achamayou: Most tables names are underscore separated, we do abbreviate, but we don't concatenate words afaict, so "js_runtime" would be more consistent:

image

achamayou: Or even "js_runtime_options"

tests/js-modules/modules.py
```diff
@@ -139,6 +139,25 @@ def test_bytecode_cache(network, args):
...
+        primary, max_heap_bytes=50 * 1024 * 1024, max_stack_bytes=1024 * 512
+    )
+    with primary.client("user0") as c:
+        r = c.get("/node/js_metrics")
+        body = r.body.json()
+        assert body["max_heap_size"] == 50 * 1024 * 1024
+        assert body["max_stack_size"] == 1024 * 512
+    # reset the heap and stack sizes to default values
+    network.consortium.set_js_engine_options(
+        primary, max_heap_bytes=100 * 1024 * 1024, max_stack_bytes=1024 * 1024
```
achamayou, line 156: If we don't have one already, it'd be good to have a test checking what happens when we exceed those limits, also to confirm that this is triggered around the amounts we expect.

MahatiC: Sure, I created a TODO for this https://github.com/microsoft/CCF/issues/4371

CHANGELOG.md
```diff
@@ -20,11 +20,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Added a new proposal `set_js_runtime_options` that accepts `max_heap_bytes` and `max_stack_bytes` for QuickJS runtime.
```
achamayou, line 23:
```suggestion
- Added a new proposal action `set_js_runtime_options` that accepts `max_heap_bytes` and `max_stack_bytes` for QuickJS runtime.
```

doc/audit/builtin_maps.rst
```diff
@@ -257,6 +257,18 @@ JavaScript engine version of the module cache, accessible by JavaScript endpoint
 
 **Value** QuickJS version as a string.
 
+``js_runtime_options``
+~~~~~~~~~~~~
```
jumaffre, line 261:
```suggestion
``js_runtime_options``
~~~~~~~~~~~~~~~~~~~~~~
```

I believe one of the tests in the Daily complains if this isn't aligned.

jumaffre (approved):

PR #4345, opened 2022-10-13
Speculative load hardening

wintersteiger:
This is a test to see whether the overall performance is impacted negatively by enabling `-x86-speculative-load-hardening`. 

I _believe_ we _should_ enable this for all enclaves when LVI mitigations are enabled. Open-enclave adds it automatically for anything that depends on `oecore` ([here](https://github.com/openenclave/openenclave/blob/95062ffefdbcbd6d9369ea9b0ba0ad4f86768654/cmake/openenclave-config.cmake.in#L156)), but apparently not all our sample/app/test enclaves do depend on that. (But my attempts to add ravl did add that dependency.)

I understand very little of load hardening though, so please correct me if I'm wrong and we don't want that flag on our enclaves :-)


wintersteiger:
Got it. Open Enclave applies those options only for the `openenclave::oecore` target, but not for `openenclave::oecore-lvi-cfg`. I'm not sure whether this is intended or not, maybe the LVI mitigations also take care of SPECTRE, so those flags aren't necessary. I'll ask over at OE.

wintersteiger:
https://github.com/openenclave/openenclave/issues/4641

cmake/ccf_app.cmake
```diff
@@ -55,6 +55,7 @@ endif()
 function(add_lvi_mitigations name)
   if(LVI_MITIGATIONS)
     apply_lvi_mitigation(${name})
+    target_link_libraries(${name} PRIVATE openenclave::oecore)
```
achamayou, line 58:
```suggestion
    # Necessary to make sure Spectre mitigations are applied until
    # https://github.com/openenclave/openenclave/issues/4641 is fixed
    target_link_libraries(${name} PRIVATE openenclave::oecore)
```

achamayou (approved):

PR #4350, opened 2022-10-14
Test

MahatiC:


MahatiC:
@eddyashton @jumaffre So the daily passed with this change. Is it reasonable to assume this as an indicator to merge or am I missing something else?

jumaffre:
> @eddyashton @jumaffre So the daily passed with this change. Is it reasonable to assume this as an indicator to merge or am I missing something else?

I realise that this branch was behind `main` when this first ran so let's make sure it runs now that's updated.

eddyashton:
> @eddyashton @jumaffre So the daily passed with this change. Is it reasonable to assume this as an indicator to merge or am I missing something else?

If the daily runs and passes, then it's safe to merge.

The only thing this doesn't cover is that some of the tests _in_ the `lts_compatibility` job are still being skipped because of the OpenAPI prefix change:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=51265&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=a2a8c6c1-1b13-59a7-596c-97d4f22a6857&l=40608
> Test requirements not met, skipping "test_random_receipts": Missing required methods: /log/private

The test requirement decorator runs correctly, but believes its found a service that doesn't support this endpoint. I think we want to ensure this test runs, either with a new decorator requirement ("require either this method OR this method"), or just disabling the requirements checks for `lts_compatibility`.

MahatiC:
@eddyashton  I agree, and until there's a way for `lts_compatibility` test to run safely, it looks like the OpenAPI schema changes for `/app` will have to revert. I will push a change for it.

PR #4351, opened 2022-10-14
Return errors when session consistency would be broken

### LINKED ISSUE #3952: Session consistency does not hold across elections
If a user submits a stream of transactions across an election, they may observe an inconsistent session history as one or more transactions are rolled back.

We can mitigate this by closing user TLS sessions when an election is observed.

eddyashton:
Resolves #3952.

This currently implements a softer variant of what was previously discussed. Rather than killing sessions, we return HTTP errors. Even after returning such an error, we will not kill the session - the user may ask us multiple times over the same session and get repeated errors, or even ask us a pure command (non-transactional and thus not inconsistent) and get a real response back.

I've added an end-to-end test that I _think_ covers everything, but I'm considering a stochastic for broader coverage: spammy clients confirming they see a consistently ratcheting TxID (or this new error), while we load the service/cause elections/kill nodes.


src/http/http_endpoint.h
```diff
@@ -174,6 +183,46 @@ namespace http
...
+    void record_response_txid(std::span<const uint8_t> raw_response) override
+    {
+      // To avoid a full HTTP parse, search for the desired header directly
+      std::string_view response(
+        (char const*)raw_response.data(), raw_response.size());
+      std::string_view target(http::headers::CCF_TX_ID);
+      auto header_begin = std::search(
+        response.begin(), response.end(), target.begin(), target.end());
+      auto header_name_end = std::find(header_begin, response.end(), ':');
+      auto header_value_end = std::find(header_name_end, response.end(), '\r');
```
achamayou, line 195: That seems unfortunate at best, and possibly insecure at worst if it's possible to add CCF_TX_ID inline as a header value, followed by a bogus value. We should at least look for `"\r{}", CCF_TX_ID`, I think.

tests/partitions_test.py
```diff
@@ -561,6 +709,7 @@ def add(parser):
     args = infra.e2e_args.cli_args(add)
     args.nodes = infra.e2e_args.min_nodes(args, f=1)
     args.package = "samples/apps/logging/liblogging"
+    args.snapshot_tx_interval = 20
 
     run(args)
-    run_2tx_reconfig_tests(args)
+    # run_2tx_reconfig_tests(args)
```
achamayou, line 715: Presumably needs to be put back?

tests/partitions_test.py
```diff
@@ -538,14 +685,15 @@ def run(args):
...
 
-        test_invalid_partitions(network, args)
-        test_partition_majority(network, args)
-        test_isolate_primary_from_one_backup(network, args)
-        test_new_joiner_helps_liveness(network, args)
-        for n in range(5):
-            test_isolate_and_reconnect_primary(network, args, iteration=n)
-        test_election_reconfiguration(network, args)
-        test_forwarding_timeout(network, args)
+        # test_invalid_partitions(network, args)
```
achamayou, line 688: Same as below?

achamayou:
I'm a bit skeptical about the tradeoffs involved in going above and beyond when we know for sure there has been an election, to still attempt to accurately report status. It seems to me that immediately returning an error and shutting down the connection is cleaner to implement and to reason about in terms of availability.

A batching client cannot avoid having to implement a backtracking procedure, as far as I can tell anyway.

achamayou:
Perhaps something we could do in that situation is respond with errors to all further requests, generically at first, and then with the last committed transaction id in the last term the session wrote to.

```
POST -> COMMITTED 5.16
POST -> PENDING 5.17
*** ELECTION ***
POST -> ERROR
POST -> ERROR
POST -> ERROR, LAST COMMIT WAS 5.16 - NOW ON 6.21
CLOSE CONNECTION
```

achamayou:
Summary of discussion with @heidihoward yesterday: the tradeoff for trying to do better than just dropping the connection is unclear, particularly in an environment where many client libraries use connection pools and will not expose sessions directly. It is unlikely that a client will write logic that correctly handles these responses.

eddyashton:
> I'm a bit skeptical about the tradeoffs involved in going above and beyond when we know for sure there has been an election, to still attempt to accurately report status. It seems to me that immediately returning an error and shutting down the connection is cleaner to implement and to reason about in terms of availability.
> 
> A batching client cannot avoid having to implement a backtracking procedure, as far as I can tell anyway.

> Summary of discussion with @heidihoward yesterday: the tradeoff for trying to do better than just dropping the connection is unclear, particularly in an environment where many client libraries use connection pools and will not expose sessions directly. It is unlikely that a client will write logic that correctly handles these responses.

My view is that this approach was easier to implement than closing the session (application code returning an application error, rather than affecting session lifetimes), that it makes little difference to clients which don't handle this behaviour (is your session closed aggressively or do you get permanent errors? Could you or do you want to handle one of these under the hood?), and that it's much nicer in the case where there's a human-in-the-loop request flow (harmless elections are invisible and allow you to proceed, you get a readable error if your state was lost that helps you backtrack).

achamayou:
@eddyashton in the case of an automatic client with a connection pool, keeping the connection open but returning errors for every subsequent request is clearly suboptimal, even if it's for a while until we can provide information about where the rollback took place. It results in a much higher rate of failure going forward than closing the connection.

eddyashton:
Parking this PR for now, and will try a simpler approach of aggressively killing sessions on election.

PR #4352, opened 2022-10-14
Index deletes

### LINKED ISSUE #4267: How to detect deletes in the index building

jeffa5:
- Adds a new `TxDiff` type that is similar to a `Tx` in that it takes changesets and a map name but produces a `MapDiff` instead of a `MapHandle`.
- This `MapDiff` works directly off the changeset given (which represents just the changes in a single transaction) and so has some simpler code related
- There is both a typed and untyped `MapDiff` interface.
- It also exposes deletes in the interface, e.g. in `foreach`, `get` and a new `is_deleted` function.
- Rather than adding a new function to the `indexing::Strategy` class this leverages making the transaction from the `ReadOnlyStore` class. This means indexes can continue to work as they did before but now the same function can be used to get a diff and execute different logic, preventing breakage

The performance impact on clients not using this extra logic is expected to be near-zero: it sets writes on the changeset and on the commits of a map. These are already constructed and so only expected to be a small overhead
- When building the `TxDiff` we set a new flag on the calls to build it (`keep_writes`)
- When doing historical queries we also set this flag so that commits when building up the map store the writes
- When building the map we always store the commit if this flag is set
- When building the changeset for the diff this flag causes us to copy the writes out from the commit and into the changeset

Fixes #4267


jeffa5:
Note: I expect the metrics for memory usage to increase as the logging sample now maintains an additional index so that we can properly test this new behaviour.

src/kv/untyped_map.h
```diff
@@ -259,7 +261,7 @@ namespace kv::untyped
           }
         }
 
-        if (changes)
+        if (changes || keep_all_writes)
```
eddyashton, line 264: I don't understand this `keep_all_writes` flag. If there were any writes or removals of existing keys, then `changes` would already be set to `true` by lines 249 or 258. Why do we want to insert anything into the roll if neither of those cases has happend?

jeffa5: From my understanding, when creating an untyped map for an index update the map doesn't contain the full state up to the current point, it is intended only to store the diff of what this transaction did. In this situation there might be a remove operation that occurred but won't cause the `changes = true` section as it is seen to delete an already missing key. This instead forces the logic to track the case when a delete on a missing key occurs.

With this intuition I'll move the check up earlier to just set the changes on that case.

include/ccf/historical_queries_interface.h
```diff
@@ -79,6 +79,14 @@ namespace ccf::historical
     virtual void set_default_expiry_duration(
       ExpiryDuration seconds_until_expiry) = 0;
 
+    /** EXPERIMENTAL: Set the tracking of deletes on missing keys for historical
+     * queries.
+     *
+     * This is experimental but setting this to true ensures that the `tx_diff`
+     * available in index handlers can observe deleted values.
+     */
```
achamayou, line 87: Is it safe to call after indexing has begun? If yes, does it do what a user might expect (start tracking deletes from there)? A mention of this in the docstring would be nice.

include/ccf/kv/map_diff.h
```diff
@@ -0,0 +1,159 @@
...
+     */
+    bool is_deleted(const K& key)
+    {
+      return map_diff.is_deleted(KSerialiser::to_serialised(key));
+    }
+
+    /** Iterate over all entries in the map.
+     *
+     * The passed functor should have the signature
+     * `bool(const K& k, const std::optional<V>& v)`.
```
achamayou, line 72: Perhaps worth mentioning that a nullopt indicates a key being deleted in that diff?

src/kv/untyped_map.h
```diff
@@ -256,6 +260,12 @@ namespace kv::untyped
               changes = true;
               state = state.remove(it->first);
             }
+            else if (track_deletes_on_missing_keys)
```
achamayou, line 263: This is a surprising choice, is it necessary for etcd behaviour?

src/kv/untyped_map.h
```diff
@@ -809,10 +821,16 @@ namespace kv::untyped
...
+          kv::untyped::Write writes;
+          if (track_deletes_on_missing_keys)
+          {
+            writes = current->writes;
+          }
           changes = std::make_unique<untyped::ChangeSet>(
             roll.rollback_counter,
             current->state,
             roll.commits->get_head()->state,
+            writes,
```
achamayou, line 833: Unless I am missing something, this augments the changeset with all the writes, which mostly duplicate current->state except for deletes. Is that correct? If so, and if this is done for expediency, can we have a ticket that tracks that it's possible to optimise this in the future by adding only deletes here?

achamayou (approved):

src/kv/untyped_map.h
jeffa5: This is needed as when we rebuild the map for just the diff we won't have any keys and so need to track the deletes to retain their writes in the commit.

src/kv/untyped_map.h
jeffa5: That sounds correct, yes. I'm sure there is some extra duplication going on, will make a ticket.

jeffa5: #4483

src/kv/untyped_map.h
achamayou: Ah ok.

PR #4353, opened 2022-10-14
Allow applications to be built with sanitizers enabled.

plietar:
In order for this to work, we need a few things:
- Install the ubsan.suppressions file
- Change the add_san function to point to the correct location, based on whether it is called on a CCF library or downstream app.
- Persist the value of the SAN option in ccf-config.cmake, so it gets picked up when importing CCF in downstream projects.


wintersteiger (approved):
LGTM!

achamayou (approved):

PR #4358, opened 2022-10-17
Only build `-DCOMPILE_TARGETS=sgx` for SGX Daily jobs

eddyashton:
This should reduce Daily build times slightly.

### DIFF
--- a/tests/test_install_build.sh
+++ b/tests/test_install_build.sh
@@ -5,5 +5,5 @@ set -ex
 
 mkdir -p build_against_install
 cd build_against_install
-CC=$(command -v clang-10) CXX=$(command -v clang++-10) cmake -GNinja "$1" ../samples/apps/logging/
+CC=$(command -v clang-10) CXX=$(command -v clang++-10) cmake -GNinja "$@" ../samples/apps/logging/
 ninja


achamayou (approved):

achamayou:
@eddyashton this breaks the unsafe build, which expects both a release and a virtual build.

DomAyre (approved):

DomAyre:
@achamayou Could you explain exactly how the unsafe build expects SGX and virtual? I've had a look and the UNSAFE_VERSION flag which I assume is what will result in this expectation, but I can't find the connection to expecting both builds?

achamayou:
@DomAyre they both run test_install_build.sh, which attempts to build a sample in both SGX and virtual modes, see output of https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=51526&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=d3314ab3-4689-5901-7382-e30098648b5c

DomAyre:
@achamayou Ahh I'd forgotten no `COMPILE_TARGETS` == both compile targets

.azure-pipelines-templates/install_deb.yml
```diff
@@ -35,11 +35,21 @@ steps:
       set -ex
       unsafe_build=$(grep unsafe /tmp/install_prefix) || true
       if [[ -n "$unsafe_build" ]]; then
-        ./tests/test_install_build.sh -DUSE_UNSAFE_VERSION=ON
+        ./tests/test_install_build.sh -DUSE_UNSAFE_VERSION=ON -DCOMPILE_TARGETS=sgx
       else
-        ./tests/test_install_build.sh
+        ./tests/test_install_build.sh -DCOMPILE_TARGETS=sgx
```
achamayou, line 40: While this proximally solves the issue, it is also the case that the current release .deb contains both sgx and virtual binaries, and so testing both of them is a feature! This is removing the test we have that the virtual artifacts we ship are usable, which is not what we want.

eddyashton: @achamayou The step below builds the app in virtual, which tests the virtual artifacts.

jumaffre (approved):

.azure-pipelines-templates/install_deb.yml
achamayou: I'd somehow not seen those (or perhaps they've been added since?). Either way, this is probably fine for now, despite not being completely equivalent to the previous situation nor testing what happens for users who don't pass a COMPILE_TARGETS at all.

PR #4367, opened 2022-10-18
Change `COMPILE_TARGETS` to take a single target, including SNP

DomAyre:


doc/contribute/build_ccf.rst
```diff
@@ -46,7 +46,7 @@ The most common build switches include:
 
 * **BUILD_TESTS**: Boolean. Build all tests for CCF. Default to ON.
 * **SAN**: Boolean. Build unit tests with Address and Undefined behaviour sanitizers enabled. Default to OFF.
-* **COMPILE_TARGETS**: String. List of target compilation platforms. Defaults to ``sgx;virtual``, which builds both "virtual" enclaves and actual SGX enclaves.
+* **COMPILE_TARGET**: String. Target compilation platform. Defaults to ``sgx``.
```
eddyashton, line 49:
```suggestion
* **COMPILE_TARGET**: String. Target compilation platform. Defaults to ``sgx``. Supported values are ``sgx``, ``snp``, or ``virtual``.
```

CMakeLists.txt
```diff
@@ -125,15 +125,67 @@ if("sgx" IN_LIST COMPILE_TARGETS)
   )
 
   add_dependencies(ccf ccf.enclave)
-endif()
 
-if("virtual" IN_LIST COMPILE_TARGETS)
+elseif(COMPILE_TARGET STREQUAL "snp")
```
jumaffre, line 129: I think we should add a comment here to say that the SNP build uses virtual for now (i.e. `.host` linked libraries), but that this is likely to change in the near future as SNP will require a separate toolchain, compile flags, etc.

CMakeLists.txt
```diff
@@ -125,15 +125,67 @@ if("sgx" IN_LIST COMPILE_TARGETS)
...
+
+  # virtual version
+  add_library(ccf.virtual STATIC ${CCF_IMPL_SOURCE})
```
achamayou, line 132: Maybe not in this PR, but we want to produce distinct artefacts, very likely with a different toolchain and flags.

cmake/common.cmake
```diff
@@ -262,11 +262,13 @@ target_include_directories(cchost PRIVATE ${CCF_GENERATED_DIR})
...
-if("sgx" IN_LIST COMPILE_TARGETS)
-  target_compile_definitions(cchost PUBLIC CCHOST_SUPPORTS_SGX)
-endif()
-if("virtual" IN_LIST COMPILE_TARGETS)
-  target_compile_definitions(cchost PUBLIC CCHOST_SUPPORTS_VIRTUAL)
+if(COMPILE_TARGET STREQUAL "sgx")
+  target_compile_definitions(cchost PUBLIC PLATFORM_SGX)
+elseif(COMPILE_TARGET STREQUAL "snp")
+  target_compile_definitions(cchost PUBLIC PLATFORM_SNP)
+  target_include_directories(cchost PRIVATE ${OE_INCLUDEDIR})
```
achamayou, line 269: Do we still need OE_INCLUDEDIR here?

CMakeLists.txt
DomAyre: https://github.com/microsoft/CCF/pull/4367/commits/6b89de802bee3a8468af44636caeb4065ff3906b

CMakeLists.txt
DomAyre: Agreed, as per [Juliens suggestion](https://github.com/microsoft/CCF/pull/4367#pullrequestreview-1154913030) I've added a comment to that effect

cmake/common.cmake
DomAyre: Probably not, but I wanted to start with a minimal change which maintains the current behaviour just with the new options

achamayou (approved):

PR #4359, opened 2022-10-17
PEM to JWK converters

jumaffre:
Resolves #4321 

This PR adds C++ and JS/TS API to convert a public/private EC/RSA key from PEM to JSON Web Key (JWK). Note that this uses the Python [`JWCrypto`](https://pypi.org/project/jwcrypto/) package to compare the results to a well-known implementation.

TODO: 
- [x] TS polyfill

PR #4372, opened 2022-10-18
Move all JS crypto functions to `ccf.crypto` namespace

jumaffre:
Noticed this as part of #4359. All the JS-related functions are now moved the `ccf.crypto` namespace (rather than `ccf`). We still keep the existing functions under `ccf`, but mark them as deprecated in the TypeScript doc.


letmaik (approved):
You should also update `js/ccf-app/src/crypto.ts` to use the functions in `ccf.crypto`.

achamayou (approved):

MahatiC (approved):

Issue #4362, opened 2022-10-17
Container job needs upgrade


achamayou:
Build ContainersNode.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkoutShow

Build ContainersThe `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #4381, opened 2022-10-21
Container images release improvements for 3.0.0

jumaffre:
Also resolves https://github.com/microsoft/CCF/issues/4362

Improve Docker container image release process in line of upcoming 3.0.0 release.

Will be merged after https://github.com/microsoft/CCF/pull/4367.

- We now publish containers for both SGX, SNP and Virtual
- `app_dev` and `app_run` dockerfiles have been cleaned up and now cater for all 3 platform types
- Ansible playbooks have been updated to only install SGX dependencies on SGX containers
~- The containers GitHub Action that now run on `main` too (without publishing the containers) so that we know early when/if we've broken this pipeline~
- The CCF debian package now includes the compile target as part of its name (version field), e.g. `ccf_3.0.0-sgx~dev6-69-gf8b673678`

TODO:
- [x] Update ansible
- [x] Template Dockerfiles 
- [x] Templated Github Actions `containers.yml`
- [x] Reuse `--build-arg=platform=sgx` for ansible if possible

### DIFF
--- a/.azure-pipelines-templates/matrix.yml
+++ b/.azure-pipelines-templates/matrix.yml
@@ -11,7 +11,7 @@ parameters:
       pool: 1es-dv4-focal
     SGX:
       container: sgx
-      pool: 1es-dcv2-focal
+      pool: 1es-dcdv3-focal
     SNPCC:
       pool: sev-snp-pool
 


docker/app_dev
```diff
@@ -1,33 +1,34 @@
...
+FROM ubuntu:20.04 AS base-virtual
 
-ENV UBUNTU=focal
-ENV PSW_VERSION=2.17.100
-RUN if [ -z "$PSW_VERSION" ]; then echo "Please set PSW_VERSION (e.g. 2.11)." >&2; exit 1; fi
+# Final dev image
+FROM base-${platform} AS final
 
-RUN apt-get update && apt-get install -y wget gnupg
+ARG platform=sgx
```
DomAyre, line 22: Do we need this a second time?

jumaffre: Unfortunately yes as the build arg needs to be repeated for all stages (it took me a while to find out it was necessary!).

DomAyre (approved):

PR #4382, opened 2022-10-21
add missing headers required when including files for certain options

fotiskoun:
Importing these files throw an error during compilation because of the following missing headers.


achamayou (approved):
Thank you!

achamayou:
Causes an LTS failure by virtue of the branch name starting with "ccf-", which lts_compatibility mistakes for a CCF version, one for which it can't find the corresponding release artefacts. CI looks good aside from that.

PR #4386, opened 2022-10-24
Use `ccf_repo` directive when linking to files

eddyashton:
Another fix for the 2.x docs.

#4383/#4385 make `ccf_repo` directive point to the current branch, but we have several hardcoded links to files in the docs that don't use this directive. When those files move, the docs on old versions will break.

This PR finds instances of those links in the docs, and replaces them with uses of `ccf_repo`.


jumaffre (approved):
We should probably enforce the use of the `ccf_repo` directive at some point. I can't find a way to configure Sphinx to forbid the use of a string, so we could perhaps include a `grep` for `github.com/microsoft/CCF/blob` in the `ci-checks.sh` script?

eddyashton:
> We should probably enforce the use of the `ccf_repo` directive at some point. I can't find a way to configure Sphinx to forbid the use of a string, so we could perhaps include a `grep` for `github.com/microsoft/CCF/blob` in the `ci-checks.sh` script?

Maybe. I was wondering about going the other way actually - run a transform of the entire docs after Sphinx has finished, which finds any `microsoft/CCF/main` or `microsoft/CCF/blob/main`, and rewrites those to point to the current version. A classic "if we can find it, we can fix it".

Issue #4384, opened 2022-10-23
Generate JWK test failure


achamayou:
```
47:     network = test_npm_app(network, args)
47:     │         │            │        └ Namespace(binary_dir='/__w/4/s/build', ccf_version=None, common_read_only_ledger_dir=None, config_file=None, consensus='CFT', co...
47:     │         │            └ <infra.network.Network object at 0x7fe8922013d0>
47:     │         └ <function test_npm_app at 0x7fe8921f5a60>
47:     └ <infra.network.Network object at 0x7fe8922013d0>
47:   File "/__w/4/s/tests/suite/test_requirements.py", line 21, in wrapper
47:     return func(*args, **kwargs)
47:            │     │       └ {}
47:            │     └ (<infra.network.Network object at 0x7fe8922013d0>, Namespace(binary_dir='/__w/4/s/build', ccf_version=None, common_read_only_led...
47:            └ <function test_npm_app at 0x7fe8921f59d0>
47:   File "/__w/4/s/tests/js-modules/modules.py", line 706, in test_npm_app
47:     generate_and_verify_jwk(c)
47:     │                       └ <infra.clients.CCFClient object at 0x7fe89208eb20>
47:     └ <function generate_and_verify_jwk at 0x7fe8921e9e50>
47:   File "/__w/4/s/tests/js-modules/modules.py", line 59, in generate_and_verify_jwk
47:     assert body == ref_priv_jwk, f"{body} != {ref_priv_jwk}"
47:            │       │                │         └ {'kty': 'EC', 'crv': 'P-256', 'x': '5dblDkbv4WEFMhEYLKOfao0aToTUWgxPRZzKX6FbYd4', 'y': 'SbD3SvBUdnvzqikcxzL8NjvUxvdjdiWMw1T7iocj...
47:            │       │                └ {'crv': 'P-256', 'd': 'HOOBF16lgs1MdDvY4CZF9fQYry_dklMCuUE0JKXMYw', 'kid': '5nPlZiazKVmZutIuBrrNidbLbsxcofEPOOcz0q6WnXI', 'kty':...
47:            │       └ {'kty': 'EC', 'crv': 'P-256', 'x': '5dblDkbv4WEFMhEYLKOfao0aToTUWgxPRZzKX6FbYd4', 'y': 'SbD3SvBUdnvzqikcxzL8NjvUxvdjdiWMw1T7iocj...
[...]

PR #4388, opened 2022-10-24
Fix EC JWK private key size calculation

jumaffre:
Fix https://github.com/microsoft/CCF/issues/4384

The size of the EC private key shouldn't be calculated from the big number as per https://www.openssl.org/docs/man1.0.2/man3/BN_num_bytes.html: 

> Some have tried using BN_num_bits() on individual numbers in RSA keys, DH keys and DSA keys, and found that they don't always come up with the number of bits they expected (something like 512, 1024, 2048, ...). This is because generating a number with some specific number of bits doesn't always set the highest bits, thereby making the number of significant bits a little lower. If you want to know the "key size" of such a key, either use functions like RSA_size(), DH_size() and DSA_size(), or use BN_num_bytes() and multiply with 8 (although there's no real guarantee that will match the "key size", just a lot more probability).

This would only occur rarely, when the first byte of the big number was `0x00`.

There's no easy way to test this against a well-known JWK implementation I don't think but I've confirmed the fix locally by running the specific `modules_test` successfully 100 times.


eddyashton:
> This would only occur rarely, when the first byte of the big number was `0x00`.
> I've confirmed the fix locally by running the specific `modules_test` successfully 100 times.

I'd expect a null first byte approximately 1-in-256 runs, so 100 runs is only soft assurance. Assuming this is relatively fast, can you give it another few hundred iterations? Or perhaps throw a loud message somewhere once we have generated a key with a null first byte (when `BN_num_bytes` doesn't equal the new size?), and confirm that test run still passes?

wintersteiger (approved):
LGTM!

I would have gone through

```
EC_GROUP *g = EC_KEY_get0_group(ec);
int bits = EC_GROUP_order_bits(g);
```

But, I think `EVP_PKEY_bits` is the right way to do it (it's also in 3.0, while some parts of `EC_GROUP*` are not).

wintersteiger:
+1 to test!

jumaffre:
> > This would only occur rarely, when the first byte of the big number was `0x00`.
> > I've confirmed the fix locally by running the specific `modules_test` successfully 100 times.
> 
> I'd expect a null first byte approximately 1-in-256 runs, so 100 runs is only soft assurance. Assuming this is relatively fast, can you give it another few hundred iterations? Or perhaps throw a loud message somewhere once we have generated a key with a null first byte (when `BN_num_bytes` doesn't equal the new size?), and confirm that test run still passes?
> 
> EDIT - Or even stronger - can you generate this key in a loop until we get one with a null initial byte, confirm that's the cause of the initial failure, confirm the new calc fixes it?

The `modules_test` generates 3 EC keys, hence the value of 100 as a loose threshold over the 256 mark. I've run this 500 times over lunch (so 1,500 keys) and the test passed successfully. The issue was identified and fixed exactly as you describe though (check the size of the private key, log an error if it's not a number we expect, fix the issue and make sure it succeeds when the size of the private key isn't as expected) but this scaffolding isn't something I intend to leave in the codebase.

eddyashton (approved):

PR #4389, opened 2022-10-24
Update maximum header count test

eddyashton:
Noticed this while looking at other session-killing errors for session consistency. Will explain the changes in comments.


src/http/http_parser.h
```diff
@@ -308,7 +308,7 @@ namespace http
         complete_header();
         auto const& max_headers_count =
           configuration.max_headers_count.value_or(default_max_headers_count);
-        if (headers.size() > max_headers_count)
```
eddyashton, line 311: This is fixing an off-by-one error in the enforcement of this cap. This is executed each time a new header is identified, and the code above has just popped the previous header if any exists. So if the cap is 256, and `headers.size()` is 256 (ie - we're processing the 257th), we should return an error for _this_ header, not the next one.

tests/e2e_common_endpoints.py
```diff
@@ -266,23 +270,27 @@ def run_large_message_test(
...
                 c,
-                args.max_http_headers_count - extra_headers_count,
+                args.max_http_headers_count,
                 http.HTTPStatus.REQUEST_HEADER_FIELDS_TOO_LARGE,
                 "RequestHeaderTooLarge",
                 "request_header_too_large",
-                len(headers),
+                len(headers) + extra_headers_count,
                 headers=headers,
             )
```
eddyashton, line 295: The count and logic here were wrong. `requests` may have added 2 headers, but `httpx` adds 5. We were saying something like "the cap is 254, and we're sending 255 headers", which always went down the "this should be an error" path, and _would_ always return an error for a different reason - because we were really sending 259 (>256) headers.

The new call is "the cap is 256, and this request will contain `s` headers (even though `headers` itself may contain fewer)".

tests/infra/clients.py
```diff
@@ -393,6 +403,13 @@ def request(
     def close(self):
         pass
 
+    def extra_headers_count():
+        # curl inserts the following headers in every request
+        #  host: <address>
+        #  user-agent: curl/<version>
+        #  accept: */*
+        return 3
```
eddyashton, line 411: I don't really like hardcoding this, but can't see a cheap better solution. The actual number of auto-inserted headers depends on whether we pass a body or not, this number is only true for empty bodies like this test uses.

src/http/http_parser.h
jumaffre: Is this right? We've just called `complete_header()` that calls `headers.emplace()` (i.e. `headers` size is incremented). So if the cap is 256, and `headers.size()` has just gone up to 256, then we don't want to throw just yet (the [configuration entry](https://microsoft.github.io/CCF/main/operations/configuration.html#http-configuration) is called `max_headers_count`, which implies inclusion of max value). What do you mean by "the code above"?

eddyashton: We've just emplaced the _previous_ header (copied out of `partial_parsed_header`). If we're here, and `headers.size()` is 256, then we just emplaced the 256, and we're getting a callback because the parser has found a 257th. The happy path with exactly 256 headers doesn't come here for the 257th, instead it calls `headers_complete()`. So we accept 256, but not 257.

jumaffre: Ah! Well spotted and thanks for fixing it.

MahatiC (approved):

PR #4391, opened 2022-10-24
Add eddsa with curve25519 to the JS crypto API

takuro-sato:
- Add `ccf.crypto.generateEddsaKeyPair()` with Curve25519 support
- Support Curve25519 for `ccf.crypto.verifySignature()`


include/ccf/crypto/eddsa_key_pair.h
```diff
@@ -0,0 +1,69 @@
...
+#include "ccf/crypto/san.h"
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace crypto
+{
+  class EdDSAKeyPair
```
eddyashton, line 20: Maybe I'm missing something - why do we need this type, distinct from `crypto::KeyPair`?

It looks like the code is:
```
auto kp_a = crypto::make_key_pair(crypto::CurveID::SECP384R1); // Can pass any value except CURVE25519
auto kp_b = crypto::make_eddsa_key_pair(crypto::CurveID::CURVE25519); // Must pass CURVE25519, any other value is an error
```
But after that, you could use `kp_a` and `kp_b` interchangeably (ie - call `sign`, `verify` on them, the only difference being a few more public overloads on the `KeyPair` type). So it'd be simpler if they were the same type externally. Afaict the implementations are close enough that should be possible.

takuro-sato: There are several reasons behind that:
- `KeyPair` includes a parameter of type `MDType` that EdDSA doesn't use. 
```cpp
virtual std::vector<uint8_t> sign(
      std::span<const uint8_t> d, MDType md_type = {}) const = 0;
```
- There are too many abstract member functions to implement. Most of them would not be used.
- Consistency with `RSAPublicKey`. It's separate class/interface from `Keypair`

eddyashton (approved):

Issue #3875, opened 2022-05-20
Move away from http request signing


achamayou:
Governance currently uses http request signing to authenticate governance commands. This causes two main problems:

1. the draft keeps changing, causing churn (#3011, #1866 etc)
2. client support is (very) poor

So we would like to investigate alternatives, in particular the possibility of using https://cose-wg.github.io/cose-spec/#rfc.section.4.2, and setting the result in a CCF-specific header.

The expected benefits are:

1. stability, since COSE is published
2. shorter client code, thanks to COSE libraries, independent from the HTTP layer

We want to validate 2. by writing some sample client code.

achamayou:
There are two main top-level possibilities to decouple signing from transport:

1. Include the signature in the message body itself, using a format like [COSE Sign1](https://pycose.readthedocs.io/en/latest/cose/messages/sign1message.html)
2. Include the signature as a HTTP header, for example as a [JWT](https://jwt.io/introduction) or a [CWT](https://datatracker.ietf.org/doc/html/rfc8392)

The benefit of the first approach is complete transport independence. Transports that do not support headers would still work, and programming contexts that do not allow clients to easily control header setting and formatting aren't problematic. Because the body can encoded arbitrarily, this is also an efficient choice.

The second approach inevitably requires base64 encoding, which is somewhat wasteful, and the use of the Authorization header in HTTP. JWTs are however more widespread than COSE, and library support is more widely available.

achamayou:
In the case of governance, the signed payload is used repeatedly across transactions, as the proposal or the vote is being evaluated. For that reason, it is extracted and stored in its own table.

In approach 1. where the request body contains the payload inline, this potentially leads to duplication in the KV store: the proposal would be stored once, parsed out, and again, in the middle of a COSE Sign1 frame, for offline audit purposes.

To alleviate this problem, CCF could make use of detached content, as described in https://cose-wg.github.io/cose-spec/#rfc.section.2: the headers and signatures could be store in the evidence table, away from the proposal itself. A verifier would treat the proposal as detached content.

Library support for detached content in COSE seems limited today, from a quick survey:

- COSE-JAVA: [yes](https://github.com/cose-wg/COSE-JAVA/blob/0385f31bfcfb2793abb9d6a6560acf7f9a405269/src/main/java/COSE/Message.java#L227)
[...]

PR #4392, opened 2022-10-24
Enable COSE Sign1 authentication for governance endpoints

achamayou:
Second half of #3875, with tests:

- [x] `POST /gov/ack`
- [x] `POST /gov/proposals`
- [x] `POST /gov/proposals/{proposal_id}/ballots`
- [x] `POST /gov/proposals/{proposal_id}/withdraw`

The change to `clients.py` to use COSE Sign1 is a bit of a bodge, but gets us a fair bit of coverage quickly. I will continue to clean up the implementation and the tests, but I think the interface is good to go for rc0.


src/service/tables/governance_history.h
```diff
@@ -13,4 +13,9 @@ namespace ccf
   {
     static constexpr auto GOV_HISTORY = "public:ccf.gov.history";
   }
+  using COSEGovernanceHistory = ServiceMap<MemberId, std::vector<uint8_t>>;
+  namespace Tables
+  {
+    static constexpr auto COSE_GOV_HISTORY = "public:ccf.gov.cose_history";
```
jumaffre, line 19: I believe this table needs to be documented and we should also have an end-to-end audit test for this (`governance_history.py`).

src/node/rpc/member_frontend.h
```diff
@@ -414,82 +431,172 @@ namespace ccf
...
+
+        return false;
+      }
+
+      if (must_be_active && !check_member_active(ctx.tx, member_id.value()))
+      {
+        ctx.rpc_ctx->set_error(
+          HTTP_STATUS_FORBIDDEN,
+          ccf::errors::AuthorizationFailed,
+          "Member is not active.");
```
jumaffre, line 470:
```suggestion
          fmt::format("Member {} is not active.", member_id.value()));
```

src/node/rpc/member_frontend.h
```diff
@@ -499,39 +606,59 @@ namespace ccf
...
               fmt::format("Error issuing new recovery shares: {}", e.what()));
+            return;
           }
         }
-        return make_success();
+        ctx.rpc_ctx->set_response_status(HTTP_STATUS_NO_CONTENT);
+        return;
       };
-      make_endpoint("/ack", HTTP_POST, json_adapter(ack), member_sig_only)
+      make_endpoint("/ack", HTTP_POST, ack, member_sig_only)
```
jumaffre, line 619: It's not clear to me why the `json_adapter` wrapper has been removed here as the payload is still JSON?

src/node/rpc/member_frontend.h
```diff
@@ -630,15 +808,16 @@ namespace ccf
...
             HTTP_STATUS_FORBIDDEN,
             errors::NodeAlreadyRecovering,
             "Node is already recovering private ledger.");
+          return;
         }
 
-        const auto in = params.get<SubmitRecoveryShare::In>();
-        auto raw_recovery_share = crypto::raw_from_b64(in.share);
-        OPENSSL_cleanse(const_cast<char*>(in.share.data()), in.share.size());
+        std::string share = params["share"];
```
jumaffre, line 818: This seems quite loose compared to what we had before. As per my other comment, if the endpoint was still wrapped with `json_adapter()`, we could keep this as it was.

src/service/tables/governance_history.h
achamayou: Yes!

src/node/rpc/member_frontend.h
achamayou: Because the payload isn't JSON, the payload is now `application/cose`, with JSON contents.

src/node/rpc/member_frontend.h
achamayou: The wrapper can't be used (see above), but we could still disallow additional fields, I agree.

doc/schemas/gov_openapi.json
```diff
@@ -463,6 +463,11 @@
       }
     },
     "securitySchemes": {
+      "member_cose_sign1": {
+        "description": "Request payload must be a COSE Sign1 document, with expected protected headers.Signer must be a member identity registered with this service.",
```
DomAyre, line 467:
```suggestion
        "description": "Request payload must be a COSE Sign1 document, with expected protected headers. Signer must be a member identity registered with this service.",
```

tests/governance.py
```diff
@@ -627,6 +627,165 @@ def test_cose_auth(network, args):
...
+        [
+            jinja2.FileSystemLoader(args.jinja_templates_path),
+            jinja2.FileSystemLoader(os.path.dirname(new_user.cert_path)),
+        ]
+    )
+    template_env = jinja2.Environment(
+        loader=template_loader, undefined=jinja2.StrictUndefined
+    )
+    proposal_template = template_env.get_template("set_user_proposal.json.jinja")
+    proposal_body = proposal_template.render(cert=os.path.basename(new_user.cert_path))
```
eddyashton, line 685: To avoid introducing this Jinja arg dependency on all the suites that want to run this test, we could use `consortium.make_proposal` here. I think that always tries to dump the proposals to file rather than returning the bodies, but we can separate out the method to do the first and use it here.

tests/governance.py
```diff
@@ -654,7 +813,11 @@ def gov(args):
         test_all_nodes_cert_renewal(network, args)
         test_service_cert_renewal(network, args)
         test_service_cert_renewal_extended(network, args)
-        test_cose_auth(network, args)
+        if args.authenticate_session != "COSE":
```
eddyashton, line 816:
```suggestion
        if args.authenticate_session == "COSE":
```

Unless I'm missing something...

tests/governance.py
```diff
@@ -689,6 +852,15 @@ def add(parser):
...
     infra.log_capture.COLORS = False
 
+    cr.add(
+        "session_coseauth",
+        gov,
+        package="samples/apps/logging/liblogging",
+        nodes=infra.e2e_args.max_nodes(cr.args, f=0),
+        initial_user_count=3,
+        authenticate_session="COSE",
+    )
```
eddyashton, line 862: When this instance runs, the `authenticate_session="COSE"` arg means that _all_ of the governance auth happens with COSE signatures, right? So what are the manual `test_cose_*` functions doing differently?

src/node/rpc/member_frontend.h
```diff
@@ -754,6 +944,33 @@ namespace ccf
...
+        if (sig_auth_id.has_value())
+        {
+          request_digest = sig_auth_id->request_digest;
+        }
+        if (cose_auth_id.has_value())
+        {
+          // This isn't right, instead the digest of the COSE Sign1
+          // TBS should be used here.
+          request_digest = crypto::sha256(
+            {cose_auth_id->envelope.begin(), cose_auth_id->envelope.end()});
```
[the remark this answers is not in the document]
eddyashton: Thanks, that make sense.

tests/governance.py
achamayou: Agreed, but I think governance, governance_js and governance_history are due for a proper merge & cleanup, so I will focus on that.

tests/governance.py
achamayou: They are sent over TLS authenticated sessions, strangely. But I will refactor this completely, because it's very odd and disjoint in its current form.

tests/governance.py
eddyashton: Ah ha! Ta. That clears up the comment below as well. So I think the current tests are:

`sessions_auth`
- client-authenticated TLS
- using HTTP signatures for governance,
- with additional tests that COSE signatures are accepted
`sessions_noauth`
- anonymous TLS
- using HTTP signatures for governance
- with additional tests that COSE signatures are accepted
`sessions_coseauth`
- client-authenticated TLS (?)
- using COSE signatures for governance
- skipping additional COSE tests

That's good coverage for now, but we should refactor to simplify this at some point.

eddyashton (approved):

DomAyre (approved):

tests/governance.py
MahatiC: I had this same question, thanks for clearing it up

Issue #4390, opened 2022-10-24
`partitions_cft` failing with "Multiple primaries" error


eddyashton:
We've been seeing this error intermittently for a while.

I believe there's a few symptoms, and we sometimes see _no_ primaries, and sometimes only get a primary reported from a single node. The error message is misleading and should be clarified.

Instance of the latter is here:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=51993&view=logs&j=f3457e4b-102d-5a4f-f801-dc119e4853a8&t=ec4c947d-98d7-5162-4c2f-20c4a5e8f908&l=1210

```
55: Traceback (most recent call last):
55:   File "/__w/1/s/tests/partitions_test.py", line 565, in <module>
55:     run(args)
55:   File "/__w/1/s/tests/partitions_test.py", line 547, in run
55:     test_election_reconfiguration(network, args)
55:   File "/__w/1/s/tests/suite/test_requirements.py", line 21, in wrapper
55:     return func(*args, **kwargs)
55:   File "/__w/1/s/tests/suite/test_requirements.py", line 49, in wrapper
55:     return func(network, args, *nargs, **kwargs)
55:   File "/__w/1/s/tests/partitions_test.py", line 355, in test_election_reconfiguration
55:     network.wait_for_primary_unanimity(nodes=backups)
55:   File "/__w/1/s/tests/infra/network.py", line 1241, in wait_for_primary_unanimity
55:     assert all_good, f"Multiple primaries: {primaries}"
55: AssertionError: Multiple primaries: [<infra.node.Node object at 0x7f41d7d17430>]
```

```
55: 14:56:17.543 | INFO     | infra.partitions:drop:60 - Dropping rules "[2],[3],[0,4,5,6,7]"
55: 14:56:17.713 | INFO     | __main__:test_election_reconfiguration:352 - As partition is lifted, check that isolated original backups elect primary
55: 14:56:17.713 | INFO     | infra.network:wait_for_primary_unanimity:1214 - Waiting up to 15.0s for all nodes to agree on the primary
55: 14:56:36.226 | INFO     | infra.network:wait_for_primary_unanimity:1234 - [2||] GET /node/network
[...]

PR #4393, opened 2022-10-24
Fix for `partitions_test_cft` - wait for commit sync before partitioning

eddyashton:
Fix for #4390.

The issue is that we add some nodes and then immediately partition, leading to potential differentiation between the live backups (some of them may have heard about the new joiners, while other haven't). This results in more failed elections (we have 2 nodes trying to call them, but only one can win), which occasionally last longer than our timeout.

This raises an interesting space for optimising our wait timeouts for elections. The number of election cycles we need to wait actually depends on the fraction of eligible candidates (who are up-to-date enough to win) as a proportion of the total active nodes (who are randomly calling elections). Our default timeouts assume (I think) that all or nearly all nodes are eligible. In the case where we have only a single eligible node, we need longer timeouts to reach the universe where they call an election first, scaling extremely poorly as we have larger node sets.


achamayou (approved):

achamayou:
@eddyashton very nice catch!

I think we should look at our timeouts as an upper bound, the downside of keeping them too high is just additional delay on failure cases, which we think is bearable because our aspiration is that they're infrequent!

achamayou:
Marking for auto-backport, because this potentially affects the reliability of the release pipeline there, and is probably fairly conflict-free.

PR #4396, opened 2022-10-25
Add QuickJS runtime execution cap

MahatiC:
Part of https://github.com/microsoft/CCF/issues/1459

This uses untrusted host time to calculate the execution time. I explored using the trusted time, but found it tricky to do this dynamically with QuickJS runtime options. 

- [x] Update built-in maps doc and change log entry


src/js/wrap.cpp
```diff
@@ -90,17 +115,32 @@ namespace ccf::js
     }
     size_t stack_size = 1024 * 1024;
     size_t heap_size = 100 * 1024 * 1024;
+    std::chrono::seconds default_max_execution_time{240};
```
MahatiC, line 118: This default value is still up in the air

include/ccf/service/tables/jsengine.h
```diff
@@ -12,11 +12,13 @@ namespace ccf
     size_t max_heap_bytes;
     /// @brief stack size for QuickJS runtime
     size_t max_stack_bytes;
+    /// @brief max execution time for QuickJS
+    uint64_t max_execution_time;
```
jumaffre, line 16: As for the other values, we should include the unit in the name.

src/js/wrap.cpp
```diff
@@ -90,17 +115,32 @@ namespace ccf::js
...
     {
       heap_size = js_runtime_options.value().max_heap_bytes;
       stack_size = js_runtime_options.value().max_stack_bytes;
+      default_max_execution_time =
+        std::chrono::seconds{js_runtime_options.value().max_execution_time};
     }
 
     JS_SetMaxStackSize(rt, stack_size);
     JS_SetMemoryLimit(rt, heap_size);
+    UntrustedHostTime* host_time = new UntrustedHostTime();
```
jumaffre, line 133: I don't think this pointer is freed anywhere here?

src/js/wrap.cpp
```diff
@@ -90,17 +115,32 @@ namespace ccf::js
...
+    const auto curr_time = ccf::get_enclave_time();
+    host_time->start_time = curr_time;
+    host_time->max_execution_time = default_max_execution_time;
+    JS_SetInterruptHandler(rt, js_custom_interrupt_handler, host_time);
+  }
+
+  Runtime::~Runtime()
+  {
+    JS_FreeRuntime(rt);
+    JS_SetInterruptHandler(rt, NULL, NULL);
```
eddyashton, line 143:
```suggestion
    JS_SetInterruptHandler(rt, NULL, NULL);
    JS_FreeRuntime(rt);
```

I don't think that resetting the interrupt handler should be necessary. But if we're going to, we must do that _before_ freeing the runtime.

src/js/wrap.cpp
eddyashton: I think we need finer granularity than seconds (probably milliseconds), and that the default should be `O(10s)` as an absolute max for now, and I think that 100ms might be more sensible. Since we're still single threaded, a single endpoint executing for this time is blocking the node from doing any other work. This includes making it essentially dead as far as consensus is concerned.

src/js/wrap.cpp
```diff
@@ -81,6 +82,30 @@ namespace ccf::js
...
+  {
+    UntrustedHostTime* time = reinterpret_cast<UntrustedHostTime*>(opaque);
+    auto now = ccf::get_enclave_time();
+    auto elapsed_time = now - time->start_time;
+    auto elapsed_seconds =
+      std::chrono::duration_cast<std::chrono::seconds>(elapsed_time);
+    if (elapsed_seconds.count() >= time->max_execution_time.count())
+    {
+      LOG_DEBUG_FMT("JS execution has timed out");
+      return 1;
```
eddyashton, line 101: When we return 1 here, what happens next? Does the stack return to the original call to `JS_Call` with some error value? We need to eventually convert this into a HTTP error that's reported to the caller.

src/js/wrap.cpp
MahatiC: Yes for the latter! As for not resetting the interrupt handler, it does look like it shouldn't have an impact. But given that it will set the (now) deleted raw memory (`host_time`) to `NULL` (preventing future double deletes, though it's unlikely to happen in this case) it seems fine to keep it.

src/js/wrap.cpp
MahatiC: It issues a `SIGTERM`. So I think will need a handler for that.

src/js/wrap.cpp
MahatiC: It was just being set to `NULL`, it is now deleted too

src/js/wrap.cpp
MahatiC: >  This includes making it essentially dead as far as consensus is concerned.

That's a good point. I will test this with lower values

src/js/wrap.cpp
eddyashton: I don't think that will work - if it tries to issue a `SIGTERM` within the enclave, it'll need to OCALL. And that'll only be handled by the host - it won't immediately terminate this JS execution. But I can't see any signal handling/sending code in `quickjs.c`, so I don't think this is how it's reporting interrupts.

include/ccf/service/tables/jsengine.h
MahatiC: Done

src/js/wrap.cpp
MahatiC: `100ms` is too low, the npm app test itself is failing. I've updated the default to `1s`

src/js/wrap.cpp
achamayou: +1 for the cautious reset, it doesn't hurt.

src/node/rpc/node_frontend.h
```diff
@@ -1394,10 +1396,13 @@ namespace ccf
         auto js_engine_options = js_engine_map->get();
         m.max_stack_size = 1024 * 1024;
         m.max_heap_size = 100 * 1024 * 1024;
+        m.max_execution_time = 240;
```
jumaffre, line 1399: These defaults should be defined somewhere else I think, with the same values as `wrap.cpp:110-112`.

MahatiC: Do you mean they shouldn't be hardcoded for the `js_metrics` response? (although yes this value should be updated from 240 to 1000 to reflect latest testing). IMO

jumaffre: I think there should be a single source of truth for the default values for those, otherwise as you said, they could drift and we could end up reporting incorrect values in the endpoint. Perhaps they could move to `wrap.h` that we could include from `node_frontend.h` too?

MahatiC: Yup that makes sense

MahatiC:
I added a test that calls one of the npm app test's endpoints and it times-out if the execution value is on the lower side. But I'm still not able to assess as to what kind of CCF workloads would take longer time to execute and what is the ideal upper cap that will not abort the request. To be on the safer side, we could just leave the upper cap to a value that is slightly lower than the `election_timeout` and then leave it upto the users to set the execution value through the `set_js_runtime_options` proposal.

Some notes about QuickJS:

- we don't really need to know the interval at which QuickJS calls the interrupt handler since we're aiming for a maximum execution cap. And since we have a test that determines that the interrupt handler is indeed getting called (for example at `250ms`, we can safely say any value above that will work)

- The interrupt handler check isn't always based on a counter value in QuickJS, it calls on specific opcodes like `OP_if_false`, `OP_if_true8` etc and also in `IsInstanceOf` calls

MahatiC:
I added logging in the `context` destructor to measure the execution time roughly. I then used the `e2e_logging_cft` with the js app. More than half the operations indeed took only `1ms`. But some notable operations that exceeded it considerably are:

- historical range query (consistently took around `135ms`)
- proposal submission and acceptance (`40ms`)
- adding the bytecode to kv (`24ms`)

CHANGELOG.md
```diff
@@ -98,6 +98,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - The "Node Output" page has been relabelled as "Troubleshooting" in the documentation and CLI commands for troubleshooting have been added to it.
 
+### Changed
+
+- The endpoint `GET /node/js_metrics` now also returns the QuickJS runtime execution time
```
achamayou, line 103:
```suggestion
- The endpoint `GET /node/js_metrics` now also returns the QuickJS runtime maximum execution time
```

src/js/wrap.cpp
```diff
@@ -81,28 +82,73 @@ namespace ccf::js
...
+        js_runtime_options.value().max_execution_time_ms};
     }
 
     JS_SetMaxStackSize(rt, stack_size);
     JS_SetMemoryLimit(rt, heap_size);
   }
 
+  Runtime::~Runtime()
+  {
+    JS_FreeRuntime(rt);
```
achamayou, line 149: Did we not free the runtime before? It's very surprising this wasn't caught by the performance tests, considering we create a new runtime every request.

achamayou: I see, this was moved.

src/js/wrap.h
```diff
@@ -238,22 +247,28 @@ namespace ccf::js
   class Context
   {
     JSContext* ctx;
+    JSRuntime* js_run_time;
 
   public:
     const TxAccess access;
+    UntrustedHostTime* host_time;
```
achamayou, line 254: Why is this a raw pointer?

achamayou (approved):

achamayou:
@MahatiC let's set the default to 1s and document it.

src/js/wrap.h
```diff
@@ -39,6 +39,12 @@ namespace ccf::js
   extern JSClassDef node_class_def;
   extern JSClassDef network_class_def;
 
+  const std::chrono::milliseconds default_max_execution_time{200};
```
achamayou, line 42:
```suggestion
  const std::chrono::milliseconds default_max_execution_time{1000};
```

src/js/wrap.h
```diff
@@ -39,6 +39,12 @@ namespace ccf::js
...
+  const size_t default_stack_size = 1024 * 1024;
+  const size_t default_heap_size = 100 * 1024 * 1024;
+  static std::chrono::milliseconds max_execution_time =
```
eddyashton, line 45: Why is this static?

samples/constitutions/default/actions.js
```diff
@@ -768,11 +768,10 @@ const actions = new Map([
       function (args) {
         checkType(args.max_heap_bytes, "integer", "max_heap_bytes");
         checkType(args.max_stack_bytes, "integer", "max_stack_bytes");
+        checkType(args.max_stack_bytes, "integer", "max_execution_time_ms");
```
eddyashton, line 771:
```suggestion
        checkType(args.max_execution_time_ms, "integer", "max_execution_time_ms");
```

tests/js-modules/modules.py
```diff
@@ -766,6 +773,66 @@ def test_npm_app(network, args):
...
+        assert r.status_code == http.HTTPStatus.INTERNAL_SERVER_ERROR, r.status_code
+        body = r.body.json()
+        assert body["error"]["message"] == "Operation took too long to complete."
+
+        # reset the execution time
+        network.consortium.set_js_runtime_options(
+            primary,
+            max_heap_bytes=50 * 1024 * 1024,
+            max_stack_bytes=1024 * 512,
+            max_execution_time=1000,
```
eddyashton, line 826: Rather than repeating the defaults here, we should fetch the current caps from `/js_metrics` before line 810, and then reset to those value here.

tests/js-modules/modules.py
```diff
@@ -766,6 +773,66 @@ def test_npm_app(network, args):
...
+        user_id = "user0"
+        jwt = infra.crypto.create_jwt({"sub": user_id}, jwt_key_priv_pem, jwt_kid)
+
+        # set JS execution time to a lower value which will timeout this
+        # endpoint execution
+        network.consortium.set_js_runtime_options(
+            primary,
+            max_heap_bytes=50 * 1024 * 1024,
+            max_stack_bytes=1024 * 512,
+            max_execution_time=0.5,
```
eddyashton, line 814:
```suggestion
            max_execution_time=1,
```

Setting this to `0.5ms` should be disallowed by the action validation.

tests/infra/consortium.py
```diff
@@ -528,11 +528,14 @@ def set_js_app_from_json(
...
 
-    def set_js_runtime_options(self, remote_node, max_heap_bytes, max_stack_bytes):
+    def set_js_runtime_options(
+        self, remote_node, max_heap_bytes, max_stack_bytes, max_execution_time
+    ):
         proposal_body, careful_vote = self.make_proposal(
             "set_js_runtime_options",
             max_heap_bytes=max_heap_bytes,
             max_stack_bytes=max_stack_bytes,
+            max_execution_time_ms=max_execution_time,
```
eddyashton, line 538: I think we should call the Python parameter `max_execution_time_ms` as well. It's confusing to have a subtle rename for just this field here, and elsewhere Python time args are usually in seconds.

src/js/wrap.cpp
```diff
@@ -81,28 +82,73 @@ namespace ccf::js
...
+    size_t heap_size = default_heap_size;
+
     const auto jsengine = tx->ro<ccf::JSEngine>(ccf::Tables::JSENGINE);
     const std::optional<JSRuntimeOptions> js_runtime_options = jsengine->get();
 
     if (js_runtime_options.has_value())
     {
       heap_size = js_runtime_options.value().max_heap_bytes;
       stack_size = js_runtime_options.value().max_stack_bytes;
+      max_execution_time = std::chrono::milliseconds{
```
achamayou, line 139: This needs to be scoped to the Runtime instance.

src/js/wrap.h
MahatiC: Done

src/js/wrap.h
MahatiC: It is no longer static

samples/constitutions/default/actions.js
MahatiC: Done

tests/infra/consortium.py
MahatiC: Done

tests/js-modules/modules.py
MahatiC: Makes sense, done

src/js/wrap.cpp
MahatiC: Done

src/js/wrap.cpp
```diff
@@ -81,28 +82,73 @@ namespace ccf::js
...
+  static int js_custom_interrupt_handler(JSRuntime* rt, void* opaque)
+  {
+    UntrustedHostTime* time = reinterpret_cast<UntrustedHostTime*>(opaque);
+    auto now = ccf::get_enclave_time();
+    auto elapsed_time = now - time->start_time;
+    auto elapsed_ms =
+      std::chrono::duration_cast<std::chrono::milliseconds>(elapsed_time);
+    if (elapsed_ms.count() >= time->max_execution_time.count())
+    {
+      LOG_INFO_FMT("JS execution has timed out");
```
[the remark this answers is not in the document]
MahatiC: Yeah, that's a good suggestion. Done.

src/js/wrap.h
MahatiC: This needs to be a pointer. If we simply pass an address instead of a pointer, it will end up being a local variable and will go out of scope in the `Context::call` function

eddyashton: We need to pass an address that will still be alive when the interrupt handler is called, but that doesn't need to be a raw pointer. We could use a `std::unique_ptr` to avoid remembering to manually call `delete`, or just use a direct struct member here (`UntrustedHostTime host_time;` rather than `UntrustedHostTime*;`). It'll be default constructed when `Context` is constructed, and automatically deleted in `~Context()`. When calling `JS_SetInterruptHandler`, we pass `&host_time`, and know that its lifetime spans the time where the interrupt handler could be called.

MahatiC: > just use a direct struct member here (`UntrustedHostTime host_time;` rather than `UntrustedHostTime*;`). 

It's data will be assigned in `Context::call` and if it's a struct member instead of a pointer, that assignment will not work. (Because we can't assign class member variables in a member function).

> We could use a std::unique_ptr to avoid remembering to manually call delete

`std::unique_ptr` can't be copied into the `void *`. Maybe I could use `std::shared_ptr<void>` type. I will check that.

eddyashton: > `std::unique_ptr` can't be copied into the `void *`. Maybe I could use `std::shared_ptr<void>` type. I will check that.

If `host_time` is declared as `std::unique_ptr<UntrustedHostTime> host_time;`, then we can call `host_time.get()` to return the raw `UntrustedHostTime*` (convertible to `void*`, `JS_SetInterruptHandler(..., host_time.get())` should work).

> It's data will be assigned in `Context::call` and if it's a struct member instead of a pointer, that assignment will not work. (Because we can't assign class member variables in a member function).

I don't follow this bit. If this was `UntrustedHostTime host_time;` (NB - non-const), then `host_time.start_time = curr_time;` in `Context::call(...)` should work fine?

CHANGELOG.md
```diff
@@ -98,6 +98,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - The "Node Output" page has been relabelled as "Troubleshooting" in the documentation and CLI commands for troubleshooting have been added to it.
 
+### Changed
+
+- The endpoint `GET /node/js_metrics` now also returns the QuickJS runtime maximum execution time
```
eddyashton, line 103: This should be highlighted under `Operations` as a new feature. We should say both that there _are_ JS execution caps (for memory sizes and execution time) that can be controlled by governance, and that the current caps can be fetched from `/js_metrics`.

eddyashton (commented):
LGTM!

Think this should be highlighted further up the CHANGELOG, but otherwise all good.

src/js/wrap.h
```diff
@@ -238,22 +245,28 @@ namespace ccf::js
   class Context
   {
     JSContext* ctx;
+    JSRuntime* js_run_time;
```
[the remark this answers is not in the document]
MahatiC: Done

CHANGELOG.md
MahatiC: I have pushed this up and under `Operations`. I also reworded to highlight that all three options can be changed/fetched although the previous PR talks about memory options as well. Happy to have the CHANGELOG `suggestion` to alter it if someone feels otherwise.

achamayou (approved):

PR #4412, opened 2022-10-26
Fix `sandbox.sh` to work on multiple platforms

jumaffre:
Another temporary fix for [the issue ](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=52346&view=logs&j=77890be3-fbd3-52a1-2c40-e5bc14cbbdef&t=34adc969-9b89-58c3-d8cf-816b63b5bdb3)raised when running test `test_install.sh` script against a non-SGX install: the script failed because we explicitly passed `-e release` to `sandbox.sh`, which automatically defaulted to using the `.enclave.so.signed` variant of the enclave library, which isn't available to virtual and SNP installs. 

Instead, we provide a way to identify the target platform (new `PLATFORM` file under `share/`) and deduce the default `--enclave-type` from this. 

There will probably be a neater way to do this when the packaging of non-SGX platforms is solidified.


jumaffre:
Merging this to unblock 3.0.0-rc0 release, as the changes in this PR are not tested until a release is cut.

PR #4413, opened 2022-10-27
Fix `test_install_build.sh` 

jumaffre:
Now that we have multiple packages (sgx, snp and virtual), we need to pass the appropriate `-DCOMPILE_TARGET` depending on the type of install.

There may be a better fix here where cmake's `add_ccf_app()` automatically sets the `COMPILE_TARGET` based on the install type, but I think we can clean this after `3.0.0-rc0`.


.azure-pipelines-templates/install_deb.yml
```diff
@@ -31,15 +34,38 @@ steps:
...
+        displayName: Test building a sample against installed CCF (non-SGX)
+
+  - ${{ if eq(parameters.target, 'SNPCC') }}:
+      - script: |
+          set -ex
+          unsafe_build=$(grep unsafe /tmp/install_prefix) || true
+          if [[ -n "$unsafe_build" ]]; then
+            ./tests/test_install_build.sh -DUSE_UNSAFE_VERSION=ON -DCOMPILE_TARGET=virtual
+          else
+            ./tests/test_install_build.sh -DCOMPILE_TARGET=virtual
```
DomAyre, line 66: What's the reason for virtual rather than snp? They're the same for now but when we diverge them it'll just work if it's snp?

jumaffre: Good spot. I've fixed this now.

DomAyre (approved):

PR #4415, opened 2022-10-27
[Draft] Removing AppendEntries batching from Raft

eddyashton:
Exploring a blanket removal of this feature.

We currently have no way to configure these parameters, and no metrics on their perf impact. Opening this PR to see what this change does to the perf numbers, and if those show minimal effect we can discuss the _safety_ of this change.


eddyashton:
Some comments on this PR, to revisit next week:
- It looks like there's no obvious mega perf impact of this change, hurrah!
- It looks like this PR spanned some fundamental change in the perf infra, so the graphs show very noisy results, boo!
- I think the original batching logic is really covering 2 things; _splitting a large AE into manageable chunks_, and _deciding to eagerly send AEs_

We should separate out the last part. I think the latter is a feature worth keeping, but we should be eagerly (ie - inline during replicate) sending AEs for _committable_ entries, and nothing else. For the former, I think the dynamic heuristic approach we have at the moment has risky pessimal cases and is poorly measured - we should either implement a simpler approach based on known tx counts ("remove it entirely" is this, but maybe "only send 1000 txs at a time" is reasonable too?), or find a way to accurately measure all historical tx sizes (likely an in-enclave cache paired with some communication protocol with the host).

achamayou:
The splitting into manageable chunks is definitely worth having, just to avoid memory exhaustion in the enclave. If a node falls behind quite a bit, and we need to send a large AE, we will need to hold it completely in the backup's enclave to check the GCM tag and do the integrity check. If that exceeds available memory there, the backup will crash.

The delay until batch size is reached has no functional benefit, it's a pure throughput vs latency optimisation.

eddyashton:
Closing this PR for now as no time to look at it properly. Tracking in #4544.

PR #4416, opened 2022-10-27
Run SNP release job on virtual

achamayou:


.azure-pipelines-templates/install_deb.yml
```diff
@@ -22,17 +22,18 @@ steps:
     workingDirectory: build
     displayName: Install CCF Debian package
 
-  - script: |
-      set -ex
-      cat /tmp/install_prefix | xargs -i bash -c "PYTHON_PACKAGE_PATH=../python ./test_install.sh {}"
-    workingDirectory: build
-    displayName: Test installed CCF
+  - ${{ if not(eq(parameters.target, 'SNPCC')) }}:
```
jumaffre, line 25: I feel like we don't need to skip this if we move to virtual executors for SNP?

achamayou: We do, because the attestation won't verify.

jumaffre (approved):

PR #4418, opened 2022-10-27
[Draft] Kill sessions on potential consistency loss

eddyashton:
Draft for now, as this will fail `lts_compatibility` until the parsing of the new message types has been backported.

Detailed discussion in #4401. This version locks each session to transactions in a single `View`. Any request which executes outside of that view gets a HTTP response followed by a session termination. That logic is also present on forwarded messages - the backup declares the View they think is valid, the primary uses this in their execution, and the primary responds with a boolean `true` if they think the session should be closed.

To decide
- Should we standardise on _always_ terminating a session after a `5xx` response? If we're adding any meaningful session closures (like this one), maybe we should add more.
- Do we need a more thorough test? I'm still thinking about a stress test that just checks for monotonic consistency in parallel with a bunch of elections.


src/node/rpc/frontend.h
```diff
@@ -273,6 +306,13 @@ namespace ccf
         return;
       }
 
+      // Before attempting to forward, make sure we're in the same View as we
+      // previously thought we were.
+      if (!check_session_consistency(ctx))
```
eddyashton, line 311: There's a subtle hole here - we check the view we're executing in now, and then the primary we're forwarding to below. Technically there could be an election between these steps, and we'd forward our request to the wrong primary. This is fine - that primary will report a consistency break and tell us to close the session, rather than us doing so ourselves. But I wonder if, long-term, all of this consensus state should be snapped atomically once-per-transaction, rather than allowing it to vary during transaction execution.

eddyashton:
Closing this PR, as it is superseded by #4595 and #4611.

PR #4423, opened 2022-10-28
Make `consensus.message_timeout` configurable

eddyashton:


jumaffre (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- `sandbox.sh` now accepts a `--consensus-update-timeout-ms` to modify the `consensus.message_timeout` value in each node's configuration. This can be used to alter multi-node commit latency.
+
```
MahatiC, line 13: Why do we need this? I though we didn't want to expand `sandbox.sh` with more configurable options?

eddyashton: True! The intent is to expose it so we can configure it in our e2e tests so we can run some latency benchmarks, and for consistency with other values that end up in the node config (this `message_timeout` should really always be available when `election_timeout` is, they affect similar things). This is only visible in `sandbox.sh` as an incidental side effect, but I thought that addition was worth calling out in the CHANGELOG.

MahatiC (approved):

PR #4427, opened 2022-10-28
Add commit latency test, and fix bugs causing commit latency multipliers

eddyashton:
I can add the test and the fixes separately if people would prefer.

The test is a little WIP, and should probably be producing a single number to send to CIMetrics, but I haven't done that yet.

I'll explain the fixes inline.

The headline is that for a 3-node network, with `message_timeout` reduced to `10ms`, we've gone from this:

```
Commit latency with 256ms sig interval (20 entries)
 -   min = 280.04ms
 -   max = 778.25ms
 -  mean = 732.51ms
 - stdev = 106.87ms
Mean commit latency / sig_interval = 2.86
```

To this:

```
Commit latency with 256ms sig interval (20 entries)
 -   min = 104.83ms
 -   max = 220.17ms
 -  mean = 206.65ms
 - stdev =  24.28ms
Mean commit latency / sig_interval = 0.81
```

For a 1-node network, these improvements scale pretty well towards very low sig intervals (I don't have the before numbers for this, but they were similarly bad):

```
Commit latency with 1ms sig interval (20 entries)
 -   min = 1.27ms
 -   max = 4.76ms
 -  mean = 2.61ms
 - stdev = 0.77ms
Mean commit latency / sig_interval = 2.61
Commit latency with 2ms sig interval (20 entries)
 -   min = 2.01ms
 -   max = 3.79ms
 -  mean = 2.83ms
 - stdev = 0.62ms
Mean commit latency / sig_interval = 1.42
Commit latency with 4ms sig interval (20 entries)
 -   min = 2.01ms
 -   max = 4.03ms
 -  mean = 3.16ms
 - stdev = 0.68ms
Mean commit latency / sig_interval = 0.79
Commit latency with 8ms sig interval (20 entries)
 -   min =  2.05ms
 -   max = 45.75ms
 -  mean =  5.98ms
 - stdev =  9.49ms
Mean commit latency / sig_interval = 0.75
```

### DIFF
--- a/src/enclave/enclave.h
+++ b/src/enclave/enclave.h
@@ -300,7 +300,7 @@ namespace ccf
                 time_now - last_tick_time);
             if (elapsed_ms.count() > 0)
             {
-              last_tick_time = time_now;
+              last_tick_time += elapsed_ms;
 
               node->tick(elapsed_ms);
               historical_state_cache->tick(elapsed_ms);
--- a/src/node/history.h
+++ b/src/node/history.h
@@ -553,7 +553,7 @@ namespace ccf
                   if (
                     self->store.committable_gap() > 0 &&
                     time > time_of_last_signature &&
-                    (time - time_of_last_signature) > sig_ms_interval)
+                    (time - time_of_last_signature) >= sig_ms_interval)
                   {
                     should_emit_signature = true;
                   }
--- a/src/node/history.h
+++ b/src/node/history.h
@@ -524,20 +524,10 @@ namespace ccf
           std::unique_lock<ccf::pal::Mutex> mguard(
             self->signature_lock, std::defer_lock);
 
-          const int64_t sig_ms_interval = self->sig_ms_interval;
-          int64_t delta_time_to_next_sig = sig_ms_interval;
           bool should_emit_signature = false;
 
           if (mguard.try_lock())
           {
-            // NOTE: time is set on every thread via a thread message
-            //       time_of_last_signature is a atomic that can be set by any
-            //       thread
-            auto time = threading::ThreadMessaging::thread_messaging
-                          .get_current_time_offset()
-                          .count();
-            auto time_of_last_signature = self->time_of_last_signature.count();
-
             auto consensus = self->store.get_consensus();
             if (consensus != nullptr)
             {
@@ -550,10 +540,7 @@ namespace ccf
                 }
                 case kv::Consensus::SignatureDisposition::CAN_SIGN:
                 {
-                  if (
-                    self->store.committable_gap() > 0 &&
-                    time > time_of_last_signature &&
-                    (time - time_of_last_signature) >= sig_ms_interval)
+                  if (self->store.committable_gap() > 0)
                   {
                     should_emit_signature = true;
                   }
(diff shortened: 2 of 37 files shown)


src/enclave/enclave.h
```diff
@@ -300,7 +300,7 @@ namespace ccf
                 time_now - last_tick_time);
             if (elapsed_ms.count() > 0)
             {
-              last_tick_time = time_now;
+              last_tick_time += elapsed_ms;
```
eddyashton, line 303: Tiny change, big effect. Accumulating errors!

`last_tick_time` and `time_now` are in microseconds. But we want to avoid ticking anything more often than every 1ms, so we wait for the `elapsed_ms` to go `> 0` before ticking. Assigning `time_now` means we're losing any overflow from the rounding done on line 298. If we got a 1.5ms, we tick these systems saying 1ms elapsed. Then a second 1.5ms comes in, and we tick exactly 1ms again. 3ms total real-time gone, only 2ms ticked to all of the listening systems.

The fix is to 'save' that overflow by not assigning it to `last_tick_time` - we only advance it by the `elapsed_ms` we're reporting to the other systems, so it stays in-sync with those counts.

The effect of this was that all of the tasks triggered by the timers below (including `thread_messaging`, which is _everything else_), were ticking slow. So a task that was supposed to happen in 200ms, would only happen after ~240ms of real-time.

src/node/history.h
```diff
@@ -550,10 +540,7 @@ namespace ccf
                 }
                 case kv::Consensus::SignatureDisposition::CAN_SIGN:
                 {
-                  if (
-                    self->store.committable_gap() > 0 &&
-                    time > time_of_last_signature &&
-                    (time - time_of_last_signature) > sig_ms_interval)
```
eddyashton, line 556: The minimal change here would be to change this `>` to a `>=`. With a strict inequality, we came in here exactly when we were supposed to (as far as our timers were concerned), but didn't trigger a signature this time! So we said "sign every 100ms", after 100ms we came here with a commit gap and said "the time gap is 100ms, but I need > 100ms, I'll come back next time". And next time is 100ms, so we double the effective timer.

After making that, I stared at the use of these timers and decided it's all unnecessary. We want this to run at a flat "every 100ms". If we could update it to _delay_ when a different path created a signature, we would, but we can't. So there's no reason to correct or update the inter-callback delay - we don't get accidentally woken up early, and if we _did_, we might as well try to sign _now_, and come back in 100ms. The previous logic was, afaict, trying to say that if we'd said 100ms, and somehow got executed after 20ms, we should only wait 80ms to check again? Sounds wrong to me, an unnecessary.

tests/commit_latency.py
```diff
@@ -0,0 +1,130 @@
...
+        64,
+        128,
+        256,
+    ):
+        all_stats[sig_interval] = measure_commit_latency(
+            args, sig_interval=sig_interval
+        )
+
+    for sig_interval, stats in all_stats.items():
+        factor = stats.mean() / sig_interval
```
achamayou, line 116: Hook up to cmake + cimetrics?

achamayou (approved):

Issue #4419, opened 2022-10-27
`3.0.0-rc0` follow-up TODOs


jumaffre:
Follow-up from the recent `3.0.0-rc0` release, the following points should be addressed to get the CI/release pipeline in a good state:
- [x] Revert main SGX pool to CoffeeLake https://github.com/microsoft/CCF/pull/4408 
- [x] Fix containers release job https://github.com/microsoft/CCF/pull/4417
- [x] Re-enable tests on SNPCC for release job https://github.com/microsoft/CCF/pull/4422
- [x] Fix LTS compatibility test (not until `release/3.x` branch is created)
- [x] Include platform (sgx, snp, virtual) in the cmake project name https://github.com/microsoft/CCF/pull/4421
- [x] Fix and re-enable COSE test in full test suite 
- [x] Fix `vegeta` stress test on SNPCC
- [x] Re-enable all pipelines (Daily, stress, etc.)
- [x] Re-enable/re-create disabled/flaky SNPCC executors
- [x] Aggregate release notes before `rc1` #4424 
- [x] Remove extra `.host` library in SGX package #4428 
- [x] Update docs and samples #4467 
- [x] Re-add open-enclave to non-SGX docker images #4431

PR #4428, opened 2022-10-28
Only install `.host.a` libraries on non-SGX platforms

jumaffre:
Part of #4419 

While we still need to build virtual/host variant of the libraries to be able to build unit tests for the SGX compile target, we do not need to install these in the SGX Debian package. This should reduce the size of the SGX Debian package for 3.0.0.

### DIFF
--- a/cmake/common.cmake
+++ b/cmake/common.cmake
@@ -44,7 +44,6 @@ if(USE_NULL_ENCRYPTOR)
 endif()
 
 option(SAN "Enable Address and Undefined Behavior Sanitizers" OFF)
-option(DISABLE_QUOTE_VERIFICATION "Disable quote verification" OFF)
 option(BUILD_END_TO_END_TESTS "Build end to end tests" ON)
 option(COVERAGE "Enable coverage mapping" OFF)
 option(SHUFFLE_SUITE "Shuffle end to end test suite" OFF)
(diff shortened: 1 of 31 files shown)


cmake/common.cmake
```diff
@@ -312,13 +312,16 @@ if(COMPILE_TARGET STREQUAL "sgx")
...
 endif()
+
 add_library(http_parser.host "${HTTP_PARSER_SOURCES}")
 set_property(TARGET http_parser.host PROPERTY POSITION_INDEPENDENT_CODE ON)
-install(
-  TARGETS http_parser.host
-  EXPORT ccf
-  DESTINATION lib
-)
+if(NOT COMPILE_TARGET STREQUAL "sgx")
```
DomAyre, line 318: Would it be more precise to state which platforms do need this rather than which ones don't?

achamayou: I'd also rather have a INSTALL_HOST_LIBRARIES, set in a single place if we can.

cmake/common.cmake
```diff
@@ -336,12 +339,14 @@ if(COMPILE_TARGET STREQUAL "sgx")
...
 add_host_library(ccf_kv.host "${CCF_KV_SOURCES}")
 add_san(ccf_kv.host)
-add_warning_checks(ccf_kv.host)
-install(
-  TARGETS ccf_kv.host
-  EXPORT ccf
-  DESTINATION lib
-)
+if(INSTALL_VIRTUAL_LIBRARIES)
+  add_warning_checks(ccf_kv.host)
```
jumaffre, line 343:
```suggestion
add_warning_checks(ccf_kv.host)
if(INSTALL_VIRTUAL_LIBRARIES)
```

achamayou (approved):

plietar:
Having these libraries around was useful to write unit tests in downstream applications.
At the very least, we would want to have `ccfcrypto`, `t_cose` and `qcbor`.

`ccf_endpoints` would be nice but I can probably work around it with some refactoring (it isn't actually run by the tests, but I have files where part of the file is tested, and other parts depend on it).

Issue #848, opened 2020-02-17
Generic performance testing driver


achamayou:
Follow up from #741 and associated offline discussions, this is what we think the performance tooling should look like for CCF:

A simple C++ driver, taking a sequence of pre-generated frames to send over TLS, from a file. Those can and likely will be generated from a script. The driver should read back response frames, and write them out to a mapped file, interleaved with rdtsc-sourced timestamp.
It should also write out send timestamps for indices in the input sequence.

All re-assembly, parsing, and correlation should happen offline, in a script, for flexibility. Any number of these drivers can be started concurrently.

This should help keep the tooling costs down, while getting maximum accuracy and the flexibility we need to obtain global commit and receipt latencies. The surrounding scripts will probably be Python, but could conceivably be written in any language.

achamayou:
https://github.com/WillAllmeida/performance_client

achamayou:
Link #4563 #4555 #4509 #4491 #4489 #4482 #4458 #4429

achamayou:
We will continue to improve Piccolo, but I think this is effectively complete now. Thank you @fotiskoun!

fotiskoun:
I am glad I was able to contribute!!!

PR #4429, opened 2022-10-28
PR of the perf tool generator component

fotiskoun:
This is the generator component of the performance tool. It includes a file containing the functions for the construction of the parquet output, a file for command line execution, and a sample file using the functions for the creation of more complex scenarios.


perf-system/generator/generator.py
```diff
@@ -0,0 +1,80 @@
...
+
+import pandas as pd  # type: ignore
+
+# pylint: disable=import-error
+import fastparquet as fp  # type: ignore
+
+REQUEST_CONTENT_TYPE = "content-type: application/json"
+REQUEST_LENGTH_TEXT = "content-length: "
+
+df = pd.DataFrame(columns=["messageID", "request"])
```
achamayou, line 15: Why is this global? It seems like fill_df could just create and return it?

perf-system/generator/generator.py
```diff
@@ -0,0 +1,80 @@
...
+def fill_df(
+    host: str, req_path: str, req_type: str, req_verb: str, req_iters: int, data: str
+):
+    """
+    Creates a dataframe with the data
+    required for the requests
+    """
+    # entering the private file paths as metadata in the start of parquet
+
+    print("Starting generation of requests")
```
achamayou, line 27: We use loguru elsewhere in the infra and tests, I would suggest doing the same. It does a bunch of nice things for free: timestamps, colours, levels etc.

perf-system/generator/generator.py
```diff
@@ -0,0 +1,80 @@
...
+    ]
+
+
+def create_parquet(parquet_filename: str):
+    """
+    Takes the dataframe data and stores them
+    in a parquet file in the current directory
+    """
+    print("Start writing requests to " + parquet_filename)
+    fp.write(parquet_filename, df)
```
achamayou, line 79: This seems short enough that it doesn't need a wrapper. It also seems to accept a full path, not just a file name.

perf-system/generator/generate_packages.py
```diff
@@ -0,0 +1,71 @@
...
+        help="A string with the data to be sent with a request",
+        type=str,
+    )
+
+    args = parser.parse_args()
+
+    generator.fill_df(
+        args.host or arg_host,
+        args.path or arg_path,
+        args.type or arg_type,
```
MahatiC, line 61: This doesn't compile because `type` argument is not defined above

MahatiC (commented):
Thank you for this change. This looks like a static generator of request logs, is this intended to be plugged into some other system? Or perhaps be used inline with the code so that requests can be logged? If it's the latter, we already have existing logging system that does the same.

perf-system/generator/generator.py
achamayou: I suggest the following interface:

```
class Messages:
  def __init__(): # create the DF, set the columns
    pass

  def append(host, path, content_type, verb, data, iterations=1): # append rows to the DF
    pass # return the df, PRE-concat to allow manipulating request ids for example

  def to_parquet_file(path): # write to disk
    pass
```

That way a user can do:

```
msgs = Messages()
inputs = msgs.append(..., POST, ...)
for i, r in input.iterrows():
   msgs.append(..., GET, ...)
msgs.to_parquet_file(...)
```

Alternatively, provide a free function generate_request() with the same input as append, minus the iterations, and show in samples how to use it directly with pandas.

achamayou:
@MahatiC this is the generator part of #848

heidihoward:
@MahatiC just to elaborate on @achamayou's response.

This RP contains the first of three components (generator, submitter & analyzer) of a new load testing tool for CCF that @fotiskoun is building. The plans for this tool are described in https://github.com/microsoft/CCF/discussions/2324 & https://github.com/microsoft/CCF/issues/848. @fotiskoun has been working so far on a branch of CCF https://github.com/fotiskoun/CCF/tree/testing-performance-system and this is the first of a few PRs to merge his efforts with CCF mainline. He has split the contributions into smaller RP for easier review, but this does mean that this first PR is not directly useful on its own. The generator library & cli produces parquet files that will be the input to the submitter component.

MahatiC:
@heidihoward Thank you for the background and details!

perf-system/generator/generator.py
fotiskoun: Thank you for your comment. The interface looks good and I will work towards that.
I am just not sure what you mean by ` PRE-concat to allow manipulating request ids for example` in the append function

heidihoward:
@MahatiC No worries! 

Please let us know how we can better document this tool. For instance, we could create a page in the docs if that helps?

perf-system/generator/generate_packages.py
```diff
@@ -0,0 +1,71 @@
...
+    arg_parquet_filename = "requests.parquet"
+    arg_data = '{"id": 1, "msg": "Send message with id 1"}'
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-hs",
+        "--host",
+        help="The host to submit the request. Default `https://127.0.0.1:8000`",
+        type=str,
+    )
```
eddyashton, line 27: Repeating the default value risks it getting out-of-sync (need to update it in 2 places, keep the type consistent, keep the formatting consistent each place you define a default). You can instead pass a default directly to this `add_argument` call, and then use `formatter_class=argparse.ArgumentDefaultsHelpFormatter` so that it is displayed in the help output:

```suggestion
    parser = argparse.ArgumentParser(description="Generator for perf workloads", formatter_class=argparse.ArgumentDefaultsHelpFormatter)
    parser.add_argument(
        "-hs",
        "--host",
        help="The host to submit the request.",
        default="https://127.0.0.1:8000",
        type=str,
    )
```

Then when you want to use the parsed-argument-or-the-default, you can just use `args.host` rather than `args.host or default_host`

fotiskoun: Great suggestion @eddyashton. I missed the default argument on argparse lib.

perf-system/generator/generate_packages.py
fotiskoun: I missed that somehow... Thank you @MahatiC, it will be fixed in the next commit.

perf-system/generator/generator.py
fotiskoun: You were right, I replaced the naming

perf-system/generator/generator.py
fotiskoun: I don't think it makes sense to use any logging in this component since I replaced `fill_df`. Although, I will replace my generator component logs with loguru,

MahatiC:
> For instance, we could create a page in the docs if that helps?

@heidihoward I think once we have all the components, we could then have a doc page explaining the perf system.

perf-system/generator/generate_packages.py
```diff
@@ -0,0 +1,91 @@
...
+        "--path_to_parquet",
+        help="Path to the parquet file to store the\
+            generated requests",
+        default="./requests.parquet",
+        type=str,
+    )
+    parser.add_argument(
+        "-ct",
+        "--content_type",
+        help="he Content-Type representation header is used\
```
achamayou, line 60:
```suggestion
        help="The Content-Type representation header is used\

```

achamayou (approved):

PR #4431, opened 2022-10-31
Use Open Enclave hostverify package for virtual and SNP builds

jumaffre:
Part of #4419 

This PR adds the Open Enclave lightweight hostverify package for virtual and SNP builds as this is still required to verify SGX attestation reports.


achamayou (approved):

getting_started/setup_vm/roles/ccf_build/vars/common.yml
```diff
@@ -33,6 +33,6 @@ debs:
 docker_debs:
   - docker-ce-cli
 
-doxygen_ver: "1.9.4"
+doxygen_ver: "1.9.5"
```
jumaffre, line 36: A new version of doxygen was released concurrently with this PR and the old 1.9.4 install tarball is no longer available.

PR #4432, opened 2022-10-31
Download Doxygen from SourceForge to keep v1.9.4

eddyashton:
CodeQL is failing because Doxygen v1.9.5 is out, and the previous v1.9.4 we were using is no longer present.

We already had to fix this in #3823. Rather than bumping the version, I think we should just download from SourceForge, which should host old versions for longer.

### DIFF
--- a/.azure-pipelines-templates/install_deb.yml
+++ b/.azure-pipelines-templates/install_deb.yml
@@ -68,11 +68,11 @@ steps:
           fi
         displayName: Test building a sample against installed CCF (SNP)
 
-  # Note: Do not publish pipeline artefact for SGXIceLake to avoid 
+  # Note: Do not publish pipeline artefact for SGXIceLake to avoid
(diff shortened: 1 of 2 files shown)


.azure-pipelines-templates/install_deb.yml
```diff
@@ -68,11 +68,11 @@ steps:
...
-    - task: PublishPipelineArtifact@1
-      inputs:
-        path: $(Build.ArtifactStagingDirectory)/$(pkgname)
-        artifact: $(pkgname)
-        displayName: "Publish CCF Debian Package"
+      - task: PublishPipelineArtifact@1
+        inputs:
+          path: $(Build.ArtifactStagingDirectory)/$(pkgname)
+          artifact: $(pkgname)
+          displayName: "Publish CCF Debian Package"
```
eddyashton, line 78: Formatter didn't like this from #4430. Formatting here.

jumaffre (approved):

PR #4411, opened 2022-10-26
Another subset of small improvements of TLA+ specification listed in #4264

lemmy:

PR #4437, opened 2022-10-31
No need to intersect with config_server if agree_server is defined in… … terms of config_server.

lemmy:
Depends on https://github.com/microsoft/CCF/pull/4411


heidihoward (approved):
Thanks!

PR #4438, opened 2022-10-31
Rename currentConfigurations to configurations and model value as `[ Servers -> [ Nat -> SUBSET Servers ] ]` instead of `[ Servers -> Seq(<<Nat, SUBSET Server>>)`.

lemmy:
Depends on https://github.com/microsoft/CCF/pull/4437


heidihoward (approved):
lgtm!

PR #4439, opened 2022-10-31
Remove unused variable votesSent.

lemmy:
Depends on https://github.com/microsoft/CCF/pull/4438


heidihoward (approved):
lgtm, thanks!

lemmy:
FWIW: I was curious how the ccfraft spec ended up with the (unused) variable `votesSent`.  This variable is [called `votesResponded` in Ongardie's original Raft spec](https://github.com/ongardie/raft.tla/blob/974fff7236545912c035ff8041582864449d0ffe/raft.tla#L66-L69).  There, the variable [prevents a candidate from requesting a vote](https://github.com/ongardie/raft.tla/blob/974fff7236545912c035ff8041582864449d0ffe/raft.tla#L191) iff the recipient [has already responded](https://github.com/ongardie/raft.tla/blob/974fff7236545912c035ff8041582864449d0ffe/raft.tla#L310-L311).  The variable `votesResponded` is *not* persistent in the [`Restart`](https://github.com/ongardie/raft.tla/blob/974fff7236545912c035ff8041582864449d0ffe/raft.tla#L168) and [`Timeout`](https://github.com/ongardie/raft.tla/blob/974fff7236545912c035ff8041582864449d0ffe/raft.tla#L183) actions, and, thus, can very likely be considered an optimization that doesn't affect safety (it should be checked how preventing multiple (redundant) vote requests affects liveness).
@jinlmsft renamed the variable to `votesSent`, re-defined its value from `[Server -> SUBSET Server]` to `[Server -> BOOLEAN]`, and [removed the conjunct](https://github.com/ongardie/raft.tla/pull/4/commits/bae0bc1d57eb8af16e6cb61afc78a8aeba91dd34#diff-bb3d38bb39f3fe7a00caad9b3342885e40e0d8bd96eb74897bf9ffe137a83355L191) that prevents a server to re-request another response.  I checked (simulated) the original spec *without* `votesResponded` for a couple of minutes up to diameter 12, and no safety violation (multiple leaders) was found.


Related:
https://github.com/pingcap/tla-plus/pull/3#issuecomment-361812051

PR #4442, opened 2022-11-01
TLA changes do not require CI runs, yet.

achamayou:


achamayou:
Should unblock things like #4439, #4438, #4437, #4436 etc.

MahatiC (approved):

MahatiC:
This actually came up in today's stand-up!

eddyashton (approved):

heidihoward (approved):
Thanks! This will be very helpful when it comes to merging in TLA+ PRs quickly

lemmy:
Motivated by https://github.com/microsoft/CCF/issues/4472, where would I start integrating TLA+ simulation with a time-budget into the [build/PR workflow](https://github.com/microsoft/CCF/tree/main/.github/workflows)?

achamayou:
@lemmy adding a new workflow would be fantastic if you have time. If you don't, but you can drop a few pointers in a issue, we can pick it up. If it needs beefier kit than what's available by default, we can set up a hosted 1es pool.

PR #4447, opened 2022-11-01
Build Docker release images in one bash script

jumaffre:
Part of #4419 

This PR adds a single `build_release_images.sh` bash script to build all release Docker images in one script that is now run in the Daily and release pipelines (Azure DevOps) to mitigate the flakiness of the post-release containers build and publish GitHub Action.


jumaffre:
Closing this PR as won't do as we're going for parallel building of containers: https://github.com/microsoft/CCF/pull/4587

Issue #4434, opened 2022-10-31
Do not publish PyPi and npm package until release is complete


jumaffre:
We should not publish the `ccf` PyPi and `ccf-app` npm packages until the release is complete, to avoid having to manually delete them if the release fails at a later stage (e.g. container job failure).

To do so, as the Python wheel and npm archive are included as assets for each release, we could expand the current container GitHub Actions and publish them to PyPi/npm once the containers have been built successfully.

We should probably also build the containers in Azure DevOps (in Daily and release) so that we verify that they work correctly before publishing a new release.

PR #4453, opened 2022-11-01
Split script pkg publishing

achamayou:
Addresses #4434


.github/workflows/npm.yml
```diff
@@ -0,0 +1,33 @@
+name: "Build and Publish NPM package"
```
jumaffre, line 1: Should these be part of a **single** post-release public Action that publishes artefacts in order (easiest to revert first, i.e. push containers to MCR last) instead of parallel jobs? Otherwise, the npm package could be successfully published but building the docker containers could still fail (although mitigated by #4447).

achamayou: I discussed this with @eddyashton yesterday, I'd started doing it like that at first, but:

1. It's never going to be atomic. We can decide what goes before or after, but we cannot prevent half a release going out, and if it happens there is no easy answer. Neither MCR, NPM or PyPi can be rolled back either.
2. Splitting allows separate retries, without having to write stuff around the previous steps to make them idempotent.

achamayou: To be clear, I could still easily be convinced, there's just this mild retry-based reason to split that I can see.

jumaffre: Good points. I think we can remove PyPi packages manually via the web UI (not great though) and I imagine it's also possible for NPM but having individual retries is much nicer I agree.

achamayou: There are limited circumstances where unpublishing from NPM is possible: https://docs.npmjs.com/unpublishing-packages-from-the-registry, but we cannot rely on them always being true I think.

.github/workflows/npm.yml
```diff
@@ -0,0 +1,33 @@
...
+
+on:
+  release:
+    types: [published]
+
+env:
+  ACR_REGISTRY: ccfmsrc.azurecr.io
+  ACR_TOKEN_NAME: app-push-token
+  DOCKER_BUILDKIT: 1 # https://docs.docker.com/develop/develop-images/build_enhancements/
```
jumaffre, line 10:
```suggestion
```

jumaffre, line 10:
```suggestion
env:
```

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,36 @@
+name: "Build and Publish PyPi package"
+
+on:
+  release:
+    types: [published]
+
+env:
+  ACR_REGISTRY: ccfmsrc.azurecr.io
+  ACR_TOKEN_NAME: app-push-token
+  DOCKER_BUILDKIT: 1 # https://docs.docker.com/develop/develop-images/build_enhancements/
```
jumaffre, line 10:
```suggestion
```

jumaffre (approved):

PR #4454, opened 2022-11-02
Add sign api to js crypt api

takuro-sato:
Add `ccf.crypto.sign()`


js/ccf-app/src/global.ts
```diff
@@ -235,7 +235,15 @@ export interface EcdsaParams {
   hash: DigestAlgorithm;
 }
 
-export type SigningAlgorithm = RsaPkcsParams | EcdsaParams;
+/**
+ * EdDSA signature algorithm parameters.
+ */
+export interface EddsaParams {
+  name: "EdDSA";
+  hash: null;
```
takuro-sato, line 243: `hash: null` looks awkward, but it's necessary to keep the current design. If I remove it, codes like the following get a TS error.
```typescript
function foo(algorithm: SigningAlgorithm ) {
  console.log(algorithm.hash); // no property `hash` error
}
``` 

If we are happy to change the public interfaces including existing one, maybe we can do something like in future:
```typescript
export type AlgorithmName = "RSASSA-PKCS1-v1_5" | "ECDSA" | "EdDSA";

export interface SigningAlgorithm {
  name: AlgorithmName
  hash?: DigestAlgorithm
}
```

achamayou: We can change the public API in 3.0, since it's a major release.

achamayou (approved):

src/js/crypto.cpp
```diff
@@ -503,6 +503,122 @@ namespace ccf::js
...
+      {
+        std::vector<uint8_t> contents(data, data + data_size);
+        crypto::Pem key_pem(key);
+        auto key_pair = crypto::make_eddsa_key_pair(key_pem);
+        auto sig = key_pair->sign(contents);
+        return JS_NewArrayBufferCopy(ctx, sig.data(), sig.size());
+      }
+      catch (const std::exception& ex)
+      {
+        auto e = JS_ThrowRangeError(ctx, "%s", ex.what());
```
eddyashton, line 559:
```suggestion
        auto e = JS_ThrowInternalError(ctx, "%s", ex.what());
```

eddyashton (approved):

js/ccf-app/src/global.ts
```diff
@@ -307,6 +315,21 @@ export interface JsonWebKeyRSAPrivate extends JsonWebKeyRSAPublic {
...
+   * @param algorithm Signing algorithm and parameters
+   * @param key A PEM-encoded private key
+   * @param data Data that was signed
+   * @throws Will throw an error if the key is not compatible with the
+   *  signing algorithm or if an unknown algorithm is used.
+   */
+  sign(
+    algorithm: SigningAlgorithm,
+    key: string,
+    data: ArrayBuffer
```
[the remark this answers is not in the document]
takuro-sato: `digest(algorithm: "SHA-256", data: ArrayBuffer): ArrayBuffer` and `verifySignature(algorithm: SigningAlgorithm, key: string, signature: ArrayBuffer, data: ArrayBuffer): boolean` have `data` parameter as well. Do we want to make the same change to them?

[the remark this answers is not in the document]
achamayou: @takuro-sato yes let's do that there too.

js/ccf-app/src/global.ts
takuro-sato: I made the change in [770bd07](https://github.com/microsoft/CCF/pull/4454/commits/770bd07f860ceb6e5b2c75fb15b8a3535de685fa).
Please check it just in case. Other than the change on `SigningAlgorithm` itself, the important change would be removing `RsaPkcsParams` and `EcdsaParams` from the public API.

js/ccf-app/src/global.ts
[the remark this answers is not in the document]
takuro-sato: Done in [0631415](https://github.com/microsoft/CCF/pull/4454/commits/0631415a20a40a6b1701c2b57b6e4ce0029fd1f3)

PR #4456, opened 2022-11-02
Support multiple concurrent external executors

eddyashton:
Previously the application stored a single active `PendingRequest` that was being processed. Now it stores one _per Executor_, so multiple Executors may have multiple transactions in-progress at the same time.


src/apps/external_executor/external_executor.cpp
```diff
@@ -150,36 +187,33 @@ namespace externalexecutor
...
-        if (!active_request.has_value())
+        const auto executor_id = get_caller_executor_id(ctx);
+        const auto it = active_requests.find(executor_id);
+        if (it == active_requests.end())
         {
-          return ccf::grpc::make_error(
-            GRPC_STATUS_FAILED_PRECONDITION,
-            "Not managing an active transaction - this should be called after "
-            "a successful call to StartTx");
+          return out_of_order_error;
```
MahatiC, line 216: Replace this (line 212-216) with `find_active_request` call like below in line 305?

eddyashton: I do the lookup manually here, so that we have the iterator to later call `erase`.

MahatiC (approved):

PR #4457, opened 2022-11-02
Tentative autoCancel fix

achamayou:
Implementing https://learn.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#ive-got-a-conditional-step-that-runs-even-when-a-job-is-canceled-how-do-i-manage-to-cancel-all-jobs-at-once

### DIFF
--- a/.azure-pipelines-templates/cimetrics.yml
+++ b/.azure-pipelines-templates/cimetrics.yml
@@ -2,7 +2,7 @@ jobs:
   - job: cimetrics_render
     displayName: "CI Metrics Render"
     dependsOn: ${{ parameters.depends_on }}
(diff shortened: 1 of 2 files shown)


MahatiC (approved):
Nice! I guess the only way to check this works is by merging it?

achamayou:
@MahatiC I'm hoping that it works on the PR and that I can test it by pushing another commit just now :) Let's see!

achamayou:
It does seem to work. The jobs apparently get canceled with extreme prejudice, but do free up their slot in the pool queue, and that's the main thing we want:

image

PR #4461, opened 2022-11-03
Publish new per target platform CI docker images

jumaffre:
This is necessary as we now depend on `open-enclave-hostverify` to install the CCF Debian packages for virtual and SNP builds. This will also facilitate the use of different toolchains per platform in the future. 

The Azure DCAP client also needs to be included in the virtual and SNP images so that CCF nodes can verify quotes.

### DIFF
--- a/getting_started/setup_vm/ccf-dev.yml
+++ b/getting_started/setup_vm/ccf-dev.yml
@@ -3,15 +3,23 @@
     - import_role:
         name: intel
         tasks_from: sgx-psw.yml
+      when: platform == "sgx"
     - import_role:
         name: intel
         tasks_from: sgx-group.yml
+      when: platform == "sgx"
     - import_role:
         name: az_dcap
         tasks_from: install.yml
+      when: platform == "sgx"
     - import_role:
         name: openenclave
         tasks_from: binary_install.yml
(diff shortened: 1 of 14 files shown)


getting_started/setup_vm/ccf-dev.yml
```diff
@@ -3,15 +3,23 @@
...
         name: az_dcap
         tasks_from: install.yml
+      when: platform == "sgx"
     - import_role:
         name: openenclave
         tasks_from: binary_install.yml
+      when: platform == "sgx"
+    - import_role:
+        name: openenclave
+        tasks_from: install_host_verify.yml
```
eddyashton, line 21: Looks like this file isn't included in this PR yet?

jumaffre: It should have been added to `main` in https://github.com/microsoft/CCF/pull/4431/files#diff-e7e6a99610d72802a104fb456bb1b7f13c2f67a3353bfdd27854b1c4b0c8c53b

eddyashton: Ah you're right, I just didn't see it in my old branch.

achamayou (approved):

PR #4470, opened 2022-11-03
Remember node and log index instead of having TLC hang on to a copy of a prefix of some node's log.

lemmy:
refactoring


heidihoward:
One of the ideas that `committedLog` and `committedLogConflict` were trying inelegantly to capture is that different nodes (or even the same node) might try to commit conflicting log entries (different values at the same index) at different times. On quick inspection, I'm not sure if this refactoring still captures this, I'll do a more thorough investigation on next week.

lemmy:
> One of the ideas that `committedLog` and `committedLogConflict` were trying inelegantly to capture is that different nodes (or even the same node) might try to commit conflicting log entries (different values at the same index) at different times. On quick inspection, I'm not sure if this refactoring still captures this, I'll do a more thorough investigation on next week.


```tla
CommittedLogProperty ==
   [][ \A i \in 1..cl.index: log[cl.node][i] = log[cl'.node][i] ]_vars
```

heidihoward:
My mistake, this approach does still capture the same properties.

heidihoward (approved):

Issue #4348, opened 2022-10-14
Multi-Container: Implement logging app as a sample Executor


eddyashton:
We have implementations of the logging app in C++ and JS. We should also add one as an external Executor, probably in Python as that's what we have easy support for.

This will highlight what we need for feature parity. We can do this in multiple stages as we add gRPC support for more features (ie - get the request details in for basic request first, think about historical queries and indexing later).

eddyashton:
We have a basic version of the logging API implemented as an external executor (see #4477), but not full feature parity.

Closing as not under active development, as work on external executors is suspended.

PR #4477, opened 2022-11-04
Initial implementation of Logging as external executor

eddyashton:
First part of #4348.

To actually test this, I think it will be much simpler if we have ALPN negotiation (so we can have interfaces serving both HTTP/1.1 and HTTP/2), so I'll look at that next.

### DIFF
--- a/src/apps/external_executor/external_executor.cpp
+++ b/src/apps/external_executor/external_executor.cpp
@@ -395,6 +395,44 @@ namespace externalexecutor
         {executor_auth_policy})
         .install();
 
+      auto get_all = [this](
+                       ccf::endpoints::ReadOnlyEndpointContext& ctx,
+                       externalexecutor::protobuf::KVTable&& payload)
+        -> ccf::grpc::GrpcAdapterResponse<
+          std::vector<externalexecutor::protobuf::KVKeyValue>> {
+        // TODO: Rebase
+        if (!active_request.has_value())
+        {
+          return ccf::grpc::make_error(
+            GRPC_STATUS_FAILED_PRECONDITION,
+            "Not managing an active transaction - this should be called "
+            "after a successful call to StartTx and before EndTx");
+        }
+
+        auto handle = active_request->tx->ro<Map>(payload.table());
+        std::vector<externalexecutor::protobuf::KVKeyValue> results;
+
+        handle->foreach([&results](const auto& k, const auto& v) {
+          externalexecutor::protobuf::KVKeyValue& result =
+            results.emplace_back();
+          result.set_key(k);
+          result.set_value(v);
+
+          return true;
+        });
+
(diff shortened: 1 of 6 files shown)


MahatiC (approved):

tests/external_executor/executors/logging_app.py
```diff
@@ -0,0 +1,100 @@
...
+                table=table,
+                key=body["id"].to_bytes(8, "big"),
+                value=body["msg"].encode("utf-8"),
+            )
+        )
+        response.status_code = HTTP.HttpStatusCode.OK
+        header = response.headers.add()
+        header.field = "content-type"
+        header.value = "application/json"
+        response.body = json.dumps(True).encode("utf-8")
```
jumaffre, line 40: Beyond the scope of this PR: are there any plans to look at how this would integrate with existing HTTP frameworks (e.g. Flask) and if so, provide a sample app?

jumaffre (approved):

tests/external_executor/executors/logging_app.py
eddyashton: No concrete plans, but I think it's something we'll look at when building more sample apps.

tests/external_executor/executors/logging_app.py
```diff
@@ -0,0 +1,100 @@
...
+    def run_loop(self, terminate_event):
+        target_uri = f"{self.ccf_node.get_public_rpc_host()}:{self.ccf_node.get_public_rpc_port()}"
+        with grpc.secure_channel(
+            target=target_uri,
+            credentials=self.credentials,
+        ) as channel:
+            stub = Service.KVStub(channel)
+
+            while not (terminate_event.is_set()):
+                request_description_opt = stub.StartTx(Empty())
```
takuro-sato, line 71: This is just a question rather than request to change.
It looks like it pulls requests very actively. Is it effective enough for us? Or are we going to make some changes like making the executer a http server?

eddyashton: The eventual plan (may be possible after #4537), is to make `StartTx` a streaming RPC. Since we don't currently support streaming, we get a similar result by polling in a loop. The result should be similar to making the executor a server (gets _sent_ requests to execute, as soon as they arrive, rather than having to manually ask for them), without requiring that the executors are reachable as a server (ie - have some unique network name).

We could sleep or backoff in this polling loop, trading off responsiveness/latency for reducing load. If we need to keep the polling long-term, I think we'll do that, but since its a temporary inefficiency I think we're happy to keep this sample simple (if inefficient).

takuro-sato: Thanks. I understood it. Making `StartTx` sounds nice also because it only requires gRPC which we already have.

PR #4482, opened 2022-11-04
Remove empty line in generator, breaking the command line run

fotiskoun:
While making the suggested changes on the generator-pr, I missed an empty line on the arg parsing options, which very cunningly escaped my tests. As a result, when one wants to generate requests using command line options, the script would break.  This PR resolves this issue.

### DIFF
--- a/perf-system/generator/generate_packages.py
+++ b/perf-system/generator/generate_packages.py
@@ -58,9 +58,7 @@ def main():
         "-ct",
         "--content_type",
         help="The Content-Type representation header is used\
-
-            to indicate the original media type of the resource.\
-            Default `application-json`",
+            to indicate the original media type of the resource.",
         default="application-json",
         type=str,
     )


heidihoward:
How did this sneak its way past the CI checks?

achamayou (approved):

achamayou:
@heidihoward the generator does to appear to be run by the CI yet.

heidihoward:
I thought there might be something to check the syntax of python files. 

Either way this directory does seems to be missing from the python format and linting checks https://github.com/microsoft/CCF/blob/main/scripts/ci-checks.sh

achamayou:
@heidihoward pylint scans python/ (the ccf package, published to pypi) and tests/ (our test infra). I think we should move perf-system under tests/, since it is testing infrastructure.

heidihoward:
Agreed. @fotiskoun can you move `perf-system/` under `tests/` as @achamayou suggested. The python files will then be checked by black and pylint as part of the CI.

fotiskoun:
I have added on the ci-checks (where applicable) the perf-system directory and I was about to add the change on my next PR about the submitter. If people prefer instead to move the perf system under test and leave the ci-checks as is, I can do that and remove my changes on the ci-checks.

heidihoward:
@fotiskoun Yes, I think inside the test directory seems like a good home for this tool (unless anyone thinks otherwise).

PR #4491, opened 2022-11-07
Perf tool cpp submitter

fotiskoun:
The second component out of 3, of the performance system tool. After the generator PR, submitted a few days ago, which generates requests to be submitted on a CCF node, the submitter component, included in this PR, is responsible for opening a TLS connection, submitting requests produced by the generator component, and receiving the responses.
The submitter component is implemented in C++ (`submit.cpp`), reads a parquet file containing the requests using the libarrow and libparquet libraries and opens a new TLS connection with the use of certificate files specified by the user. The paths to the certificate files, paths to read and write parquet files, the server-address and the option to submit pipelined HTTP/1.1 requests are passed as arguments and handled by `handle_arguments.hpp`.
After opening a TLS connection using `create_connection()` which is a variation of the existing `perf_client.h/create_connection()`, the submitter sends requests either waiting for a response after each request or sending a number of requests to the CCF (specified from `max_block_write` variable) and wait for a response when reaching this threshold. 
For the requests' response, a new function `read_raw_response()` is implemented under `rpc_tls_client.h` that returns the response as sent by the server. 
To avoid reading from parquet and writing to the parquet files while submitting data, `parquet_data.hpp` helps add the values required to vectors and then stores the responses and timestamps using `storeParquetResults()`.
In order to compile the submitter component, inside `common.cmake` was added a new compilation target called `submit` containing and linking the libraries required.


cmake/common.cmake
```diff
@@ -748,3 +748,49 @@ function(add_picobench name)
...
+
+# Perf tool executable
+set(SUBMITTER_DIR ${CCF_DIR}/tests/perf-system/submitter)
+
+add_executable(
+  submit ${SUBMITTER_DIR}/submit.cpp ${SUBMITTER_DIR}/handle_arguments.hpp
+         ${SUBMITTER_DIR}/parquet_data.hpp
+)
+
+add_library(stdcxxhttp_parser.host "${HTTP_PARSER_SOURCES}")
```
eddyashton, line 760: I added a similar comment on the ansible PR, but I think it got lost:

Do we still need stdcxx? Do we still need to rebuild this and attempt to link with a different std lib implementation? If we're building within the CCF codebase, we should be able to use the same compiler toolchain as the other host libraries.

cmake/common.cmake
```diff
@@ -748,3 +748,49 @@ function(add_picobench name)
...
+  target_link_libraries(
+    submit PRIVATE stdcxxhttp_parser.host stdcxxccfcrypto.host arrow parquet
+  )
+
+else()
+  target_link_libraries(
+    submit PRIVATE stdcxxhttp_parser.host stdcxxccfcrypto.host arrow parquet
+                   c++fs
+  )
+endif()
```
eddyashton, line 796: Repeated calls to `target_link_libraries` will append libraries. So it would be simpler to have a single call outside the conditional listing the common libraries, then:

```
if(NOT CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 9)
  target_link_libraries(submit PRIVATE c++fs)
endif()
```

cmake/common.cmake
```diff
@@ -748,3 +748,49 @@ function(add_picobench name)
...
   set_property(TEST ${name} PROPERTY LABELS benchmark)
 endfunction()
+
+# Perf tool executable
+set(SUBMITTER_DIR ${CCF_DIR}/tests/perf-system/submitter)
+
+add_executable(
+  submit ${SUBMITTER_DIR}/submit.cpp ${SUBMITTER_DIR}/handle_arguments.hpp
+         ${SUBMITTER_DIR}/parquet_data.hpp
+)
```
eddyashton, line 758: I don't think this should live in `common.cmake`. Ideally it should go in `tests/.../submitter/CMakeLists.txt`, and then that should be included from the root `CMakeLists.txt` or similar.

tests/perf-system/submitter/handle_arguments.hpp
```diff
@@ -0,0 +1,90 @@
...
+      .add_option(
+        "--generator-filepath",
+        generator_filepath,
+        "Path to parquet file with the generated requests to be submitted.")
+      ->capture_default_str();
+    app
+      .add_flag("--pipeline", isPipeline, "Enable HTTP/1.1 pipelining option.")
+      ->capture_default_str();
+  }
+  void argument_assigner(int argc, char** argv);
```
eddyashton, line 87: This appears to be both unimplemented and uncalled - can we remove it?

tests/perf-system/submitter/handle_arguments.hpp
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
eddyashton, line 1: A few coding style changes to match the CCF coding style:
- File should be `handle_arguments.h` rather than `.hpp`
- `#ifndef HANDLE_ARGUMENTS_H` guards should be replaced with a single `#pragma once`
- Functions we define should be `snake_cased` rather than `camelCased` (ie - `read_parquet_file` rather than `readParquetFile`)

cmake/common.cmake
fotiskoun: Sorry for not addressing your comment on the previous PR.
So, the problem is that libarrow-dev and libparquet-dev that contain the header files for arrow and parquet are compiled against libstdcpp. In the beginning, when trying to compile my program, depending on arrow and parquet, with libcpp I got errors coming form libarrow. Discussing the compiling issues with @achamayou, he thought that we should best avoid rebuilding arrow using libcpp and instead, since there are no security issues for the perf-tool, build crypto and http_parser against libstdcpp.
@achamayou would probably be able to explain the issue even better in case I missed or misunderstood something.

cmake/common.cmake
fotiskoun: Yes, great suggestion. I made my statement looking the following [line](https://github.com/microsoft/CCF/blob/94f53a0f8bd4035aa3bf541f4756e9f3538b3a62/cmake/common.cmake#L296) which does something similar and could benefit from the same suggestion. Happy to make the change in the line above, if people want me to.

tests/perf-system/submitter/handle_arguments.hpp
fotiskoun: Sure, I removed the implementation but missed the call. Thanks

cmake/common.cmake
eddyashton: Ah ha! Yes, please fix that instance as well :)

cmake/common.cmake
eddyashton: Thanks for the summary, @achamayou explained the same. I had completely missed this dependency, but it makes sense, so the std library variants stay!

cmake/common.cmake
fotiskoun: I made a new CMakeLists.txt inside submitter directory and I included it in  CMakeLists.txt as `include(${CCF_DIR}/tests/perf-system/submitter/CMakeLists.txt)`.
If you think it looks good, please feel free to resolve the conversation. Thanks!

tests/perf-system/submitter/handle_arguments.h
```diff
@@ -0,0 +1,85 @@
...
+{
+public:
+  std::string label; //< Default set in constructor
+  std::string pid_file; //< Default set in constructor
+
+  std::string cert;
+  std::string key;
+  std::string rootCa;
+  std::string server_address = "127.0.0.1:8000";
+  std::string send_filepath = "../tests/perf-system/submitter/cpp_send.parquet";
```
achamayou, line 21: The following three default values are... very specific? I'm not sure I like the idea of having default output paths in particular, it seems like it makes it a bit too easy to accidentally overwrite some results?

tests/perf-system/submitter/handle_arguments.h
```diff
@@ -0,0 +1,85 @@
...
+  std::string response_filepath =
+    "../tests/perf-system/submitter/cpp_respond.parquet";
+  std::string generator_filepath =
+    "../tests/perf-system/generator/requests.parquet";
+  bool pipeline = false;
```
achamayou, line 26: This would be better as max_inflight_requests, where =0 is pipeline=false, an amount can be specified, and -1 means no backpressure.

tests/perf-system/submitter/handle_arguments.h
```diff
@@ -0,0 +1,85 @@
...
+
+  ArgumentParser(
+    const std::string& default_label,
+    const std::string& default_pid_file,
+    CLI::App& app) :
+    label(default_label),
+    pid_file(fmt::format("{}.pid", default_pid_file))
```
achamayou, line 33: What do we use the pid file for? That's generally useful for a long running server/daemon we detach, but I don't expect we will ever use the submitter like that.

tests/perf-system/submitter/submit.cpp
```diff
@@ -0,0 +1,284 @@
...
+  arrow::fs::LocalFileSystem file_system;
+  std::shared_ptr<arrow::io::RandomAccessFile> input =
+    file_system.OpenInputFile(generator_filepath).ValueOrDie();
+
+  // Open Parquet file reader
+  std::unique_ptr<parquet::arrow::FileReader> arrow_reader;
+  st = parquet::arrow::OpenFile(input, pool, &arrow_reader);
+  if (!st.ok())
+  {
+    LOG_INFO_FMT("Couldn't found generator file");
```
achamayou, line 40:
```suggestion
    LOG_INFO_FMT("Couldn't find generator file");

```

achamayou: Also, should we continue after that?

tests/perf-system/submitter/submit.cpp
```diff
@@ -0,0 +1,284 @@
...
+  else
+  {
+    LOG_INFO_FMT("Opened generator file");
+  }
+
+  std::shared_ptr<::arrow::ChunkedArray> column;
+
+  ::arrow::Status column1Status = arrow_reader->ReadColumn(1, &column);
+  std::shared_ptr<arrow::StringArray> col1Vals =
+    std::dynamic_pointer_cast<arrow::StringArray>(column->chunk(
```
achamayou, line 64: Are those casts necessary?

tests/perf-system/submitter/submit.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+  // Keep only the host and port removing any https:// characters
+  std::string separator = "//";
+  auto exists_index = server_address.find(separator);
+  if (exists_index != std::string::npos)
+  {
+    server_address = server_address.substr(exists_index + separator.length());
+  }
+
+  int max_block_write = 1000; // Threshold for maximum pending writes
```
achamayou, line 213: This is what the replacement for pipeline should set.

achamayou (approved):
@fotiskoun lgtm, left a few minor comments.

tests/perf-system/submitter/handle_arguments.h
fotiskoun: You are talking about the three filepaths, correct? I think you are right, I will remove them.

tests/perf-system/submitter/handle_arguments.h
fotiskoun: I took a decision at this point about `max_inflight_requests !=0`. Instead of using one loop and overcomplicating the if statement of when to read based on the `max_inflight_requests <0` or `max_inflight_requests>0, I am making this comparison before the loop to avoid overcomplicating once the submission has started.

tests/perf-system/submitter/submit.cpp
fotiskoun: No we shouldn't!

tests/perf-system/submitter/submit.cpp
fotiskoun: Unfortunately, afaik yes.

achamayou (approved):

achamayou:
@fotiskoun can you run `./scripts/ci-checks.sh -f` on your branch to fix the formatting, and push again please?

eddyashton (approved):

Issue #4299, opened 2022-10-03
Verify additional fields in SEV-SNP attestation report


jumaffre:
As per [Table 21](https://www.amd.com/system/files/TechDocs/56860.pdf), we should also verify additional fields in the attestation report for SEV-SNP. In particular, the `GUEST_POLICY` field (table 8) and the debug, migrate and SMT flags should be checked in accordance to member governance. 

We should dig this up a little more to find out which fields are critical and how to expose this nicely to operators/members.

achamayou:
I think we should contemplate moving this to p0 and making it a blocker for 3.0.

DomAyre:
- We should check SVN version (in the COSE Sign 1 doc not the attestation report, check it's newer, possibly according to gov)
- Does SMT present some side channel issues? (also in platform info, whether it is actually enabled)
- Reported/launch/committed TCB? (Could check minimum version, we care about reported)
- Should we handle live updates? (No notification mechanism)
- Chip ID (prevent nodes on the same box)

PR #4492, opened 2022-11-07
Track guest policy fields in SNP attestation

DomAyre:
Currently just expands the field and enforces some fields with hard coded expectations
Addresses #4299


include/ccf/pal/attestation_sev_snp.h
```diff
@@ -75,13 +75,28 @@ QPHfbkH0CyPfhl1jWhJFZasCAwEAAQ==
...
+    struct GuestPolicy
+    {
+      uint8_t abi_minor;
+      uint8_t abi_major;
+      uint8_t smt : 1;
+      uint8_t reserved : 1;
+      uint8_t migrate_ma : 1;
+      uint8_t debug : 1;
+      uint8_t single_socket : 1;
+      uint64_t reserved2 : 43;
```
jumaffre, line 89:
```suggestion
      uint8_t reserved2[344];
```

For consistency with the other structs.

include/ccf/pal/attestation_sev_snp.h
```diff
@@ -75,13 +75,28 @@ QPHfbkH0CyPfhl1jWhJFZasCAwEAAQ==
...
+#pragma pack(pop)
+
 #pragma pack(push, 1)
     // Table 21
     struct Attestation
     {
       uint32_t version; /* 0x000 */
       uint32_t guest_svn; /* 0x004 */
-      uint64_t policy; /* 0x008 */
+      struct GuestPolicy policy; /* 0x008 */
```
jumaffre, line 99:
```suggestion
      struct GuestPolicy guest_policy; /* 0x008 */
```

include/ccf/pal/attestation.h
```diff
@@ -197,6 +197,16 @@ namespace ccf::pal
         throw std::logic_error(
           "Chip certificate (VCEK) did not sign this attestation");
       }
+
+      if (quote.policy.debug == 1)
+      {
+        throw std::logic_error("Debugging should not be enabled");
```
jumaffre, line 203:
```suggestion
        throw std::logic_error("SNP attestation report guest policy debugging should not be enabled");
```

jumaffre (commented):
Could you touch the daily canary file so that this runs on SNP executors too?

include/ccf/pal/attestation_sev_snp.h
DomAyre: Are these the same thing? My understanding is the current is a bit field which is 43 bits long whereas the suggestion is an array of 64 bit values making it 2752 bits long?

include/ccf/pal/attestation_sev_snp.h
DomAyre: I'm happy to do this, the only risk is then we diverge from the AMD docs?

include/ccf/pal/attestation_sev_snp.h
jumaffre: Ah yes, you're right. I didn't read this right. The equivalent would be `uint8_t reserved2[43]`.

include/ccf/pal/attestation_sev_snp.h
jumaffre: Absolutely, let's not deviate from the official docs. I was just trying to avoid confusion with the `host_data` field that's the hash of a security policy.

include/ccf/pal/attestation_sev_snp.h
DomAyre: Isn't that still an array of 43 uint8_t's (344 bits) as opposed to a uint64_t restricted to 43 bits?

jumaffre: Forget what I said!

include/ccf/pal/attestation.h
```diff
@@ -197,6 +197,18 @@ namespace ccf::pal
         throw std::logic_error(
           "Chip certificate (VCEK) did not sign this attestation");
       }
+
+      if (quote.policy.debug == 1)
+      {
+        throw std::logic_error(
+          "SNP attestation report guest policy debugging should not be "
```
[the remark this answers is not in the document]
DomAyre: https://github.com/microsoft/CCF/pull/4492/commits/75ddf6d7969570e2ad70c92d8d74bbae64801fba

achamayou (approved):

PR #4506, opened 2022-11-09
Add ccf_cose_sign1 CLI tool

### LINKED ISSUE #3875: Move away from http request signing

achamayou:
Replace the first half of scurl.sh, in a way that allows piping directly into curl (the second half of scurl.sh!).

- [x] Update CurlClient to use ccf_cose_sign1 | curl
- [x] Update documentation samples that use scurl to use ccf_cose_sign1
- [x] Update governance history test to verify cose sign1 payloads

Closes #3875


python/setup.py
```diff
@@ -48,5 +48,6 @@
         path.join(UTILITIES_PATH, "submit_recovery_share.sh"),
         path.join(UTILITIES_PATH, "verify_quote.sh"),
     ],
+    entry_points={"console_scripts": ["ccf_cose_sign1 = ccf.cose:sign_cli"]},
```
takuro-sato, line 51: Do we want to have test for `ccf_cose_sign1`? I think it would be helpful as example usage not only preventing degradation.

tests/infra/signing.py
```diff
@@ -1,121 +1,12 @@
...
-        cose_key = from_cryptography_eckey_obj(key)
-    else:
-        raise NotImplementedError("unsupported key type")
-    msg.key = cose_key
-
-    return msg.encode()
+from pycose.messages import Sign1Message  # type: ignore
 
 
 def get_cert_key_type(cert_pem: str) -> str:
```
takuro-sato, line 12: What is the reason to leave some functions here?

python/setup.py
achamayou: Yes we do!

tests/infra/signing.py
achamayou: They're not needed for the tool or for signing, they're only used for tests, so I'd rather not release them.

takuro-sato: It looks like actually they are not used now. So if we don't have a plan to use it again in near future, it might be better to remove the file.

python/setup.py
achamayou: I'm updating the CurlClient to have a path that uses ccf_cose_sign1, that'll make it easier to generate the doc and test the tool end to end.

tests/infra/signing.py
achamayou: I think that's a mistake on my part, the governance audit test should require them.

python/setup.py
achamayou: @takuro-sato done now, this is tested by the sandbox/test_install* scripts, rather than the main tests themselves. We use to run some of them with curl client too, but they're very slow and not that interesting.

python/setup.py
```diff
@@ -48,5 +48,6 @@
         path.join(UTILITIES_PATH, "submit_recovery_share.sh"),
         path.join(UTILITIES_PATH, "verify_quote.sh"),
```
jumaffre, line 49: Should the new utility be included in `scripts` so that it's available in the path when the `ccf` Python package is installed?

python/ccf/cose.py
```diff
@@ -118,81 +124,74 @@ def create_cose_sign1(
...
+        help="Path to signing key, PEM-encoded",
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
+        "--signing-cert",
+        help="Path to signing key, PEM-encoded",
+        type=str,
+        required=True,
+    )
```
jumaffre, line 157: What is the certificate used for in the COSE Sign1 signature?

jumaffre (approved):
It'd be great to add some documentation on this [here](https://microsoft.github.io/CCF/main/governance/proposals.html) and add a new changelog entry too.

python/setup.py
takuro-sato: I feel like I am missing something, but I can't find any file sandbox/test_install* that has been changed. Did you forget to upload the change?

achamayou: They have not, but what they do is they use the sandbox, which now sets CURL_CLIENT=1 and CURL_CLIENT_USE_COSE=1, and the CurlClient has been updated to use ccf_cose_sign1 when those are set. Another way to see that in action is to run the sandbox.

tests/infra/signing.py
achamayou: Done now.

takuro-sato (approved):

PR #4507, opened 2022-11-09
Remove SNPCC from daily and have separate file based trigger

DomAyre:


DomAyre:
Looks like this needs to merge before I can create the azure pipelines for SNP, unless I create a branch on the MS repo?

.azure_pipelines_snpcc.yml
```diff
@@ -0,0 +1,23 @@
+pr:
```
jumaffre, line 1: I think we'll want to trigger this when pushing to `main` as well?

```
trigger:
  branches:
    include:
      - main


```

jumaffre (approved):

PR #4509, opened 2022-11-09
add files for the analyzer component

fotiskoun:
The 3rd and last component of the performance testing tool. After requests are submitted by the second component and the responses and timestamps are written in parquet files, the Analyzer component is responsible for reading these files and produce some reasonable metrics. 
There are 3 files included in this component. The two satellite files are either call from the command line a `default-analysis` or manipulate some of the functions provided by the 3rd file to create a small analysis.
The main file `analyze.py` contains the functions that make the analysis. Almost all functions are under the class which needs to be instantiated and its object will call all the different functions, except `default_analysis()` which lives out of the class and is called from the `analyze_packages.py` as a default analysis from the command line.
`.gitignore` is edited to prevent committing the images produced.


tests/perf-system/analyzer/analyzer.py
```diff
@@ -0,0 +1,284 @@
...
+        # will need to handle the re-submission response from submitter when decided
+        return (
+            df_responses.iloc[req_id]["receiveTime"] - df_sends.iloc[req_id]["sendTime"]
+        )
+
+    def check_success(self, df_responses: pd.DataFrame, req_id: int) -> int:
+        req_resp = df_responses.iloc[req_id]["rawResponse"].split("\n")
+        status_list = req_resp[0].split(" ")
+        # if we get a full statues and says ok increase the successful
+        if len(status_list) > 1 and status_list[1][:3] == "200":
```
achamayou, line 45: Best to check for "2\d\d", empty success will be 204 for example, rather than 200.

tests/perf-system/analyzer/analyzer.py
```diff
@@ -0,0 +1,289 @@
...
+        )
+        return generic_output_table
+
+    def latencies_table(
+        self, df_sends: pd.DataFrame, df_responses: pd.DataFrame
+    ) -> PrettyTable:
+        ms_time_spent_sum = self.total_time_in_sec(df_sends, df_responses) * SEC_MS
+        latency_output_table = PrettyTable()
+        latency_output_table.field_names = [
+            "Latency (ms)",
```
heidihoward, line 109: Maybe update label to highlight that this is the medium (or 50% percentile)

tests/perf-system/analyzer/analyzer.py
```diff
@@ -0,0 +1,289 @@
...
+        plt.figure(figsize=(15, 15), dpi=80)
+
+    def plot_throughput_per_block(
+        self, df_responses: pd.DataFrame, time_block: float
+    ) -> None:
+        """
+        It splits the dataset in buckets of time_block seconds difference
+        and will plot the throughput for each bucket
+        """
+        # sort the latencies as it make sense to get the throughput
```
heidihoward, line 165:
```suggestion
        # sort the latencies as it makes sense to get the throughput

```

tests/perf-system/analyzer/analyzer.py
```diff
@@ -0,0 +1,290 @@
...
+            df_responses.iloc[req_id]["receiveTime"] - df_sends.iloc[req_id]["sendTime"]
+        )
+
+    def check_success(self, df_responses: pd.DataFrame, req_id: int) -> int:
+        req_resp = df_responses.iloc[req_id]["rawResponse"].split("\n")
+        status_list = req_resp[0].split(" ")
+        # if we get a full statues and says ok increase the successful
+        if (
+            len(status_list) > 1
+            and status_list[1][:3] >= "200"
```
achamayou, line 47: This probably mostly works in practice, but doesn't seem ideal, since "3" for example would be accepted. I would be better to enforce something like "2\d\d" I think.

tests/perf-system/analyzer/analyzer.py
```diff
@@ -0,0 +1,290 @@
...
+        plt.plot(block_latency, throughput_per_block)
+        plt.ylabel("Throughput(req/s)")
+        plt.xlabel("time(ms)")
+        plt.savefig("throughput_across_time.png")
+
+    def plot_latency_distribution(self, ms_separator: float, highest_vals=15):
+        """
+        Starting from minimum latency with ms_separator
+        step split the ms latency list in buckets
+        and plots the highest_vals top buckets
```
achamayou, line 200: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.hist.html does this reasonably well?

tests/perf-system/analyzer/analyzer.py
```diff
@@ -0,0 +1,290 @@
...
+    LOG.info(f"The request type sent is {analysis.get_req_type(df_responses)}")
+
+    print(
+        analysis.time_success_throughput_table(
+            df_sends, df_responses, successful_percent
+        )
+    )
+    print(analysis.latencies_table(df_sends, df_responses))
+
+    x = ["-"] * 20
```
achamayou, line 283:
```suggestion
    x = "-" * 20
```
and then you can save on all the joins ;)

tests/perf-system/analyzer/analyzer.py
fotiskoun: I totally missed that! Thanks for catching it. I could probably add a check for the `len()` of the string, but I think it makes sense to import regex at this point, which I was trying to avoid.

eddyashton: Much simpler than importing regex is to try and parse this as an int, and confirm that int is between 200 and 299. If its not parseable, or not in the correct range, that's an error.

tests/perf-system/analyzer/analyzer.py
fotiskoun: The problem I had with `pd.hist()` and other built in histogram functions is: 
1) It was complicated to remove 0 element bars
2) I wanted the bars to split in the step that the user would give e.g. every 0.1s and have bars [9.0, 9.1), [9.1, 9.2) e.t.c. Whereas with the built-in functions you can only specify how many bars/hists you want. That means that even after finding out how many bars you would have with the specific "step" (e.g. for a min_latency=9.067 and max_latency=10 with step 0.1 you would have 10 bars), the library will start the first bar with the first latency so the bars will be [9.067. 9.167), [9.167, 9.267) and so on, which I didn't want.

If that doesn't make sense and you want me to use pandas built-in please let me know.

achamayou: @fotiskoun no that's fine. I think you would do that by rounding your latencies in a separate column, counting the uniques and setting the number of bins to that, but if you have something that works, that's fine too.

MahatiC:
@fotiskoun If it isn't done yet, since this is the last PR, would you mind adding a quick doc entry on the usage of this performance tool? Separate PR is fine too.

tests/perf-system/analyzer/analyzer.py
achamayou: I can't resist pointing out that this comes in as a string, so while you will need an extra `import re`, you can use a single `re.compile("^2\d\d$").match()` as a condition. The conversion to int on the other hand may throw, and that's at least another two lines to handle.

Having said that, I am of course fine with anything that's correct ^_^.

achamayou (approved):

tests/perf-system/analyzer/analyzer.py
fotiskoun: I used re which is builtin `re.search("^2[0-9][0-9]$", status_list[1])`. I was thinking as well the parsing @eddyashton but I was wondering if the parsing and comparing would be faster than doing a `re.search()`. I am happy to change it to parsing if you think it is better.

fotiskoun: `\d` and `[0-9]` produce the same result in our case (if I am not very wrong, their difference is very tiny which would never occur in our case).

fotiskoun:
@MahatiC thanks for the suggestion. I have along a `README.md`, if that's what you mean. I was thinking to add it as a separate PR when all the components are submitted and I would have a version with all the components that are on CCF.

MahatiC:
@fotiskoun I was hoping to have it as an entry over here https://microsoft.github.io/CCF/main/index.html, with instructions on how to use this performance tool? This will mean creating a new .rst file (titled `Performance` maybe) and then adding that to the index.rst. You can then verify doc changes by running `./livehtml.sh` and navigating to localhost (VSCode will do the automatic port forwarding and you can view the doc changes locally!)

fotiskoun:
Thank you @MahatiC, sounds good. I will create one and I add it in a future PR.

tests/perf-system/analyzer/analyzer.py
achamayou: @fotiskoun https://docs.python.org/3/library/re.html

```
\d

For Unicode (str) patterns:
Matches any Unicode decimal digit (that is, any character in Unicode character category [Nd]). This includes [0-9], and also many other digit characters. If the [ASCII](https://docs.python.org/3/library/re.html#re.ASCII) flag is used only [0-9] is matched.

For 8-bit (bytes) patterns:
Matches any decimal digit; this is equivalent to [0-9].
```

Issue #4512, opened 2022-11-09
ci-checks.sh -f fails with write permission error


ross-p-smith:
**Describe the bug**
Before submitting a PR you can run `./scripts/ci-checks.sh -f` to fix any linting/formatting issues. However when running this you receive an error and it does not fix the formatting errors.

**To Reproduce**
- Add an extra line in README.md markdown file
- run `./scripts/ci-checks.sh -f` 
- Look in the output

```
[error] Unable to write file: README.md
[error] EACCES: permission denied, open 'README.md'
```

**Expected behavior**
The script should fix simple linting/formatting issues

**Environment information**
This is a permissions devcontainer issue 

**Additional context**
This is really low priority!

achamayou:
@ross-p-smith is this something that happens only in devcontainers? I'm confused as to why the files mounted there would not be writeable, or README.md would be different from the rest of the files in the repo?

ci-checks.sh will not attempt to elevate, it only ever runs as the current user, so this sounds like a devcontainer set up issue? We don't do anything special permission-wise there, afaict.

achamayou:
Cannot reproduce.

PR #4513, opened 2022-11-09
Format documents on save

ross-p-smith:
Related to #4512 

Really small change and more than happy for you to just ignore this! :-)

In my devcontainer I cannot get the ci_checks to fix my files. So I used a plugin. Changes are

1. Changing the name of DevContainer as it's not a sample
2. You should not define context as you have an inline image tag and context is only used if you are using a Dockerfile
3. Added the Prettier extension into the devcontainer as that is what you are using for your linting checks.
4. Tell vscode that the default document formatter is prettier
5. Then tell vscode to format the documents when you save them.

To show this working, rebuild your dev container with these settings and then add some bad document formatting (e.g. extra line breaks). You should be able to Ctrl+S (save) the file and it will format it for you.

This will also work for js/json/etc files

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,17 @@
 {
-  "name": "Sample Development Environment for CCF",
-  "context": "..",
+  "name": "CCF Development Environment",
   "image": "ccfmsrc.azurecr.io/ccf/ci:oe-0.18.2-virtual",
   "runArgs": [],
-  "extensions": ["ms-vscode.cpptools", "ms-python.python"]
+  "extensions": [
+    "ms-vscode.cpptools",
+    "vsls-contrib.codetour",
+    "esbenp.prettier-vscode"
+  ],
+  "settings": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "editor.formatOnSave": true,
+    "editor.codeActionsOnSave": {
+      "source.fixAll": true
+    }
+  }
 }


.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,17 @@
 {
-  "name": "Sample Development Environment for CCF",
-  "context": "..",
+  "name": "CCF Development Environment",
   "image": "ccfmsrc.azurecr.io/ccf/ci:oe-0.18.2-virtual",
   "runArgs": [],
-  "extensions": ["ms-vscode.cpptools", "ms-python.python"]
```
DomAyre, line 6: It's probably worth keeping the python extension

ross-p-smith: eek! I copied the wrong extension list! Thanks ever so much!!!

.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,17 @@
...
+  "extensions": [
+    "ms-vscode.cpptools",
+    "ms-python.python",
+    "esbenp.prettier-vscode"
```
achamayou, line 8: I am a little worried about this, and I hope I'm wrong - here is why: we've had problems in the past caused by vscode not running the same version of a tool that we use in ci-checks.sh (mostly with cmake format, not with prettier). We pick up new versions directly from npm, whereas the plugin might lag behind a bit. When this happens, vscode produces code that doesn't pass CI checks.

Anyway, I'm happy to try this and see how it goes, so @ross-p-smith please close this and merge when you've read it.

achamayou (approved):

.devcontainer/devcontainer.json
ross-p-smith: Yes - I can see the problem you outline. Ultimately, your CI will always run ci-checks so this will be the 'right' version. Adding this is just going to help those developers who use the vscode "Format" feature. But yes I see your point! :-)

I also won't be able to merge as I am not a maintainer - so one of your team will have to.

PR #4514, opened 2022-11-10
Move ring buffer bound check early

jumaffre:
Because the enclave logger makes use of the ring buffer, we cannot log an error before we've checked that the ring buffer bounds are not in enclave memory so this PR moves the `is_outside_enclave` check early.

### DIFF
--- a/src/enclave/main.cpp
+++ b/src/enclave/main.cpp
@@ -110,6 +110,14 @@ extern "C"
     auto writer_factory = std::make_unique<oversized::WriterFactory>(
       *basic_writer_factory, ec.writer_config);
 
+    if (!ccf::pal::is_outside_enclave(
+          ec.from_enclave_buffer_start, ec.from_enclave_buffer_size))
+    {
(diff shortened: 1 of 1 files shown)


src/enclave/main.cpp
```diff
@@ -172,13 +180,6 @@ extern "C"
...
       }
 
-      if (!ccf::pal::is_outside_enclave(
-            ec.from_enclave_buffer_start, ec.from_enclave_buffer_size))
-      {
-        LOG_FAIL_FMT("Memory outside enclave: from_enclave buffer start");
-        return CreateNodeStatus::MemoryNotOutsideEnclave;
-      }
-
       if (!ccf::pal::is_outside_enclave(
```
eddyashton, line 183: I think we need to move all of these checks earlier, include the offsets.

jumaffre: Yes of course - thank you!

DomAyre (approved):

MahatiC (approved):

achamayou (approved):

PR #4524, opened 2022-11-10
Eddsa pem to jwk

takuro-sato:


include/ccf/crypto/jwk.h
```diff
@@ -96,4 +118,19 @@ namespace crypto
   };
   DECLARE_JSON_TYPE_WITH_BASE(JsonWebKeyRSAPrivate, JsonWebKeyRSAPublic);
   DECLARE_JSON_REQUIRED_FIELDS(JsonWebKeyRSAPrivate, d, p, q, dp, dq, qi);
+
+  struct JsonWebKeyEdDSAPublic : JsonWebKey
```
takuro-sato, line 122: Both EdDSA and ECDH has the same kty (key type) `OKP` and the same jwk format.
So, we have a few options about how to define OKP JWK structs:
1. Use the current implementation. When we want to ECDH support, we can add `JsonWebKeyECDHPublic` separately (with exact same contents.) I think it makes sense because they are for different key algorithms even though the JWK format is shared.
2. Define `JsonWebKeyOKPPublic` and share it for both EdDSA and ECDH. It also makes sense because they use the same JWK format. There is a downside for CCF developers (not for users). Our code sometimes chosoe functions to call based on type like https://github.com/microsoft/CCF/blob/7683f3e93b8f7eb9de1763362f5fb8059d3e018a/src/js/crypto.cpp#L342 . If a type is related to multiple algorithms (EdDSA and ECDH here), the code would be less simple because we would need a code to check the algorithm.
3. Make `JsonWebKeyOKPPublic` a parent class for classes for `JsonWebKeyEdDSAPublic ` and `JsonWebKeyECDHPublic`. I can't imagine a case where it is useful though.


Please let me know what you think.

js/ccf-app/src/polyfill.ts
```diff
@@ -381,40 +380,98 @@ class CCFPolyfill implements CCF {
...
+        format: "jwk",
+      });
+      return {
+        d: jwk.d as string,
+        crv: jwk.crv as string,
+        x: jwk.x as string,
+        y: jwk.y as string,
+        kty: jwk.kty as string,
+        kid: kid,
+      };
```
jumaffre, line 411: I'm a little worried about this, as we manually construct the object rather than just using `jwk` as created on `:401`. Is there a reason for this? I'm thinking that if `jwk` had an extra field `foo` in it, we wouldn't return it here.

js/ccf-app/test/polyfill.test.ts
```diff
@@ -569,4 +571,25 @@ describe("polyfill", function () {
       assert.equal(foo.get(key_buf), undefined);
     });
   });
+  describe("isValidX509CertChain", function (this) {
```
jumaffre, line 574: Is this moved down because of the NodeJS crypto bug that you found? If so, it may be worth adding a comment here.

src/crypto/openssl/eddsa_key_pair.cpp
```diff
@@ -71,4 +71,22 @@ namespace crypto
...
+    const std::optional<std::string>& kid) const
+  {
+    JsonWebKeyEdDSAPrivate jwk = {
+      EdDSAPublicKey_OpenSSL::public_key_jwk_eddsa(kid)};
+
+    std::vector<uint8_t> raw_priv(EVP_PKEY_size(key));
+    size_t raw_priv_len = raw_priv.size();
+    EVP_PKEY_get_raw_private_key(key, raw_priv.data(), &raw_priv_len);
+    raw_priv.resize(raw_priv_len);
+    jwk.d = b64url_from_raw(raw_priv, false);
```
jumaffre, line 89:
```suggestion
    jwk.d = b64url_from_raw(raw_priv, false /* with_padding */);
```

tests/js-modules/modules.py
```diff
@@ -94,6 +94,29 @@ def generate_and_verify_jwk(client):
...
+
+    # Public
+    ref_pub_jwk = jwk.JWK.from_pem(pub_pem.encode()).export(as_dict=True)
+    r = client.post(
+        "/app/pubEddsaPemToJwk", body={"pem": pub_pem, "kid": ref_pub_jwk["kid"]}
+    )
+    body = r.body.json()
+    assert r.status_code == http.HTTPStatus.OK
+    assert body["kty"] == "OKP"
+    assert body == ref_pub_jwk, f"{body} != {ref_pub_jwk}"
```
jumaffre, line 118: As a quick sanity check, could you run this test in a loop for at least 1000 times locally? This is because we had a surprising bug with the initial JWK conversion (see https://github.com/microsoft/CCF/pull/4388). Looking at the OpenSSL functions you're calling to do the conversion, I believe it'll be absolutely fine but just checking to avoid a spurious CI failure down the line would be great.

jumaffre (approved):

include/ccf/crypto/jwk.h
```diff
@@ -61,6 +65,24 @@ namespace crypto
...
+    JsonWebKeyEdDSACurve, {{JsonWebKeyEdDSACurve::ED25519, "Ed25519"}});
+
+  static JsonWebKeyEdDSACurve curve_id_to_jwk_eddsa_curve(CurveID curve_id)
+  {
+    switch (curve_id)
+    {
+      case CurveID::CURVE25519:
+        return JsonWebKeyEdDSACurve::ED25519;
+      default:
+        throw std::logic_error(fmt::format("Unknown curve {}", curve_id));
```
jumaffre, line 82:
```suggestion
        throw std::logic_error(fmt::format("Unknown EdDSA curve {}", curve_id));
```

js/ccf-app/src/polyfill.ts
takuro-sato: If we are developing a general crypto library, we shouldn't miss `foo`, but it's polyfill which should simulate the CCF behaviour. Considering non-polyfill response is strongly typed by C++ struct and can't handle `foo`, I think it's OK.

tests/js-modules/modules.py
takuro-sato: It's an interesting bug, but fortunately the test passed locally 1000 times!

js/ccf-app/src/polyfill.ts
jumaffre: As long as our implementation is tested separately against a well-known library (that's what the Python test is here for), then I think we're good.

include/ccf/crypto/jwk.h
jumaffre: I think 1. is fine.

jumaffre:
@takuro-sato Could you add an entry to the changelog as well please?

PR #4533, opened 2022-11-11
Add workflow for TLA+ checking

DomAyre:


.github/workflows/tlaplus.yml
```diff
@@ -0,0 +1,15 @@
+name: "TLA+ Spec Testing"
+
+on: [push, pull_request]
```
achamayou, line 3: Can we filter and run this only on pull requests that touch a *.tla file, or this .yml file? The vast majority of pushes or pull requests will not contain a relevant change.

achamayou: Using https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths, probably.

lemmy: You would definitely want to include changes to `*.cfg` as well.   However, it is probably best to trigger a build on any change to the `tla/` folder.

.github/workflows/tlaplus.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+
+  model-checker:
+    name: Run TLC Model Checker Against CCF Raft Spec
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - run: python ./tla/install_deps.py
+      - run: (cd ./tla && tlc.sh MCccfraft.tla)
```
lemmy, line 14: The run section should also include the second model, which takes about 35s on my local machine.  Assuming a sensible `tlc` alias: 

```
tlc -config MCccfraft.cfg MCccfraftWithReconfig.tla
```

.github/workflows/tlaplus.yml
heidihoward: Could we also enable on `workflow_dispatch` so we have a button to run on demand (for instance, if there's a new release of tlc/community modules and we want to see the change of checking time).

I have implemented a TLA+ CI here if that's any help: https://github.com/fpaxos/fpaxos-tlaplus/blob/main/.github/workflows/tlatools.yml

lemmy: Another inspiration (doesn't use https://github.com/pmer/tla-bin.git, which IIRC doesn't pass JVM properties):

https://github.com/tlaplus-workshops/ewd998/blob/main/.github/workflows/main.yml

.github/workflows/tlaplus.yml
heidihoward: It might be better to use the `tlc` alias from `bashrc` instead of the `tlc.sh` wrapper. They should both work but they pass slightly different configuration parameters to java and I assume the `tlc` alias ones are better (because they were written by @lemmy instead of by me :-))

.github/workflows/tlaplus.yml
```diff
@@ -0,0 +1,15 @@
+name: "TLA+ Spec Testing"
+
+on: [push, pull_request]
+
+jobs:
+
+  model-checker:
+    name: Run TLC Model Checker Against CCF Raft Spec
+    runs-on: ubuntu-latest
```
heidihoward, line 9: Maybe we could use a larger runner: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners

lemmy: It is also fairly easy to hook in a spare but beefy machine from the lab if budget is an issue.  For example, the TLA+ performance tests use a bare metal instance, and it's a plain Ubuntu with just the GitHub runner installed.

heidihoward: Let's stick with the basic machine for now to get this merged and we can scale up later as we extend the CI

heidihoward (approved):

achamayou (approved):

PR #4534, opened 2022-11-11
Forwarded request's actor fix

MahatiC:


tests/infra/logging_app.py
```diff
@@ -104,6 +104,7 @@ def issue(
         msg=None,
         user=None,
         url_suffix=None,
+        private_url=None,
```
jumaffre, line 107: I think this should be called `url`, or in fact `uri` in this case as this doesn't include the full address?

tests/infra/logging_app.py
```diff
@@ -138,6 +139,7 @@ def issue(
                     if self.scope is not None:
                         args["scope"] = self.scope
                     url = "/app/log/private"
+                    url = private_url if private_url else url
```
jumaffre, line 142:
```suggestion
                    url = url or "/app/log/private"
```

jumaffre (approved):

tests/infra/logging_app.py
MahatiC: Just went with the existing convention in the current test i.e. `url`

tests/infra/logging_app.py
MahatiC: I believe this suggestion is applicable only if I change the param name to `url` from `private_url`? (Other than replacing the if else semantic)

jumaffre: Yes, this is just a "Python can do it in one line" comment :)

MahatiC: Gotcha, will remember this for next time :)

DomAyre (approved):

eddyashton (approved):

Issue #4325, opened 2022-10-07
Support for gRPC server streaming


jumaffre:
As part of #4034, we need to add support for server -> client streaming. Tasks identified so far:

- [x] 1. Create `Stream` object that can be created from `EndpointContext` by application. The `Stream` (shared pointer) can be stored by the app to send data to the client asynchronously (API to be defined, but inspired by https://grpc.io/docs/languages/cpp/basics/#streaming-rpcs).
- [x] 2. Create gRPC streaming endpoint wrapper that does not submit the `NGHTTP2_DATA_FLAG_EOF` and `NGHTTP2_DATA_FLAG_NO_END_STREAM` flags when the endpoint returns. Instead it is up to server to close the `Stream` when they're done.
- [x] 3. Add wrappers to serialise gRPC payloads to pass in `Stream` object when writing back to the client.
- [x] 4. Handle stream lifetime (e.g. what if the underlying connection is closed by the client?)
- [ ] 5. Add close callback so that detached streams can be cleaned up as early as possible when underlying session is closed.

jumaffre:
With regard to programming model, I believe we want to support the following modes:
1. **Unary**: a request is executed synchronously by a single endpoint. This is the current programming model.
```mermaid
stateDiagram-v2
direction LR
Uninitialised --> Closing: endpoint returns sync (sync)
```

2. **Synchronous**: The endpoint synchronously stream data back to the client, until there's nothing else to stream back and the endpoint returns as per 1. This will facilitate bidirectional streaming once client-side is implemented. 
```mermaid
stateDiagram-v2
direction LR
Uninitialised --> Streaming: stream to client (sync)
Streaming --> Closing: endpoint returns (sync)
```

Note: To close the stream, the application should simply return from the endpoint handler.

3. **Detached**: a `Stream` can be detached from the `EndpointContext` and stored by the application. The endpoint returns synchronously as per 1., but the application can later stream data back to the client (e.g. when a key is updated in the store), as long as the connection is still open. 
```mermaid
stateDiagram-v2
direction LR
Uninitialised --> AboutToStream: endpoint returns (sync)
AboutToStream --> Streaming:  stream to client (async)
Streaming --> Closing: stream is closed by app (async)
```

Note: The detachable stream should be explicitly closed by the app here.

Issue #3342, opened 2021-12-21
Support for HTTP/2


jumaffre:
CCF currently only supports HTTP/1.1. We should add support for additional client-server protocols to improve performance/allow for a wider range of apps and languages. 

_Note: Updated to reflect past HTTP/2 work_

TODO:
- [x] Experimental support (#4010 
- [x] `h2spec` compliance (#4757)
- [x] Configuration (#4787, #4790) 
- [ ] Merge HTTP/1.1 and HTTP/2 endpoints
- [ ] Support for Application-Layer Protocol Negotiation (ALPN) (https://datatracker.ietf.org/doc/html/rfc7540#section-3.3)
- [ ] Forwarding (possibly superseded by https://github.com/microsoft/CCF/issues/4670)

Useful links:
- RFC: https://datatracker.ietf.org/doc/html/rfc7540 
- https://http2-explained.haxx.se/
- https://nghttp2.org/documentation/

eddyashton:
To support gRPC streamed requests, we need to allow a handler to view incoming data before the request is complete, probably as each DATA frame arrives. Similarly for long-lived streamed responses, we need to be able to send the start of a response (headers and perhaps initial DATA frames), without immediately terminating the response, and allow some other component to send further responses + eventually terminate the response. The only way I can see to support this, is to move incoming request concatenation much deeper in the parsing stack, in fact inside each handler after dispatch. Then we add a variant handler which doesn't do this gluing, and sees each chunk/frame as it arrives. Big API changes!

jumaffre:
[`h2spec`](https://github.com/summerwind/h2spec) test failures as per `02e6cceaa` on `main` below. 140/146 tests are passing successfully so we should be able to hook this into the CI very soon.

```
Failures: 

Generic tests for HTTP/2 server
  3. Frame Definitions
    3.8. GOAWAY
      using source address 127.0.0.1:58954
      × 1: Sends a GOAWAY frame
        -> The endpoint MUST accept GOAWAY frame.
           Expected: Connection closed
                     PING Frame (length:8, flags:0x01, stream_id:0, opaque_data:h2spec)
             Actual: Timeout

Hypertext Transfer Protocol Version 2 (HTTP/2)
  4. HTTP Frames
    4.2. Frame Size
      using source address 127.0.0.1:59212
      × 3: Sends a large size HEADERS frame that exceeds the SETTINGS_MAX_FRAME_SIZE
[...]

PR #4537, opened 2022-11-11
gRPC server streaming

jumaffre:
Addresses #4325 

This is now ready for review. 

This PR implements server-side gRPC streaming, either by streaming synchronously from within the endpoint handler, or by detaching the stream and streaming asynchronously, e.g. from within the indexer.

Notable changes:
- HTTP/2 support for server-side streaming, by returning `NGHTTP2_ERR_DEFERRED` from within `read_outgoing_callback`.
- New `ccf::grpc::Stream` and `ccf::grpc::DetachedStream` objects accessible from within the application via new `grpc_command_unary_stream_adapter` endpoint wrapper.

Still to be done (separate PR):
- Provide close callback to `DetachedStream` so that app can perform cleanup when the client session is closed asynchronously.

TODO:
- [x] API: Create separate detachable stream that can be closed (!= inline streaming)
- [x] Streaming endpoint should return nothing (i.e. `make_success()`). 
- [x] Fix for when synchronous stream returns no message
- [x] Pub-sub endpoint
- [x] Attach `lookup_responder` to `RpcContext`. 
- [x] HTTP/2 layer cleanup
- [x] Fix segfault when detached stream responds when client has closed stream
- [x] Split stream close
- [x] Support for no data streamed in inline streaming endpoint


src/apps/external_executor/external_executor.cpp
```diff
@@ -466,6 +513,75 @@ namespace externalexecutor
...
+          ccf::grpc::make_stream<externalexecutor::protobuf::KVKeyValue>(
+            ctx.rpc_ctx, responder_lookup);
+
+        for (int i = 0; i < 5; i++)
+        {
+          bool close_stream = i == 4;
+          externalexecutor::protobuf::KVKeyValue kv = {};
+          kv.set_key("my_key");
+          kv.set_value(fmt::format("my_value: {}", close_stream));
+          stream->stream_msg(kv, close_stream);
```
eddyashton, line 568: Passing a `close_stream` bool to each call seems cumbersome. Preferably we'd just have a final `stream->close_stream()` call, and maybe even call that implicitly from the `Stream`'s destructor if we're feeling RAIIdical.

src/apps/external_executor/external_executor.cpp
```diff
@@ -466,6 +513,75 @@ namespace externalexecutor
...
+        }
+
+        // TODO: Fix return value here, which should return success but nothing
+        return ccf::grpc::make_success(
+          ccf::grpc::Stream<externalexecutor::protobuf::KVKeyValue>{});
```
eddyashton, line 573: I think something like `return ccf::grpc::pending();` would actually be clearer than `ccf::grpc::make_success();`. But it may also be better to avoid these return-based wrappers for streaming endpoints? So `grpc_adapter` gives us 2 features; type-safety, and a transformation to stack-constrained API (here's your params, return your result). We want to keep the former, but the latter can be dropped for these if it makes sense.

src/apps/external_executor/external_executor.cpp
```diff
@@ -161,6 +162,52 @@ namespace externalexecutor
...
+      bool close_stream)
+    {
+      externalexecutor::protobuf::KVKeyValue kv;
+      kv.set_key("my_key");
+      kv.set_value(fmt::format("my_value: {}", close_stream));
+
+      stream->stream_msg(kv, close_stream);
+    }
+
+    static void async_send_stream_data(
```
eddyashton, line 188: I think we should add a pub/sub endpoint (ie - one streaming endpoint which gets another message contains every time `/foo` is called) as a sample, in this PR. The underlying implementation is also tested by this endpoint, but that would prove out the `Stream` ownership model/reachability a little more (this sample still triggers everything on the `Stream` from a single user call - albeit transitively via the delayed thread message - and we want to confirm we can also write responses triggered from other endpoints, or hooks, or indexes, or...)

eddyashton (commented):
Looks great. Have added some very early comments on the API, let's discuss further tomorrow.

include/ccf/rpc_context.h
```diff
@@ -140,6 +140,10 @@ namespace ccf
     {
       set_response_trailer(name, std::to_string(n));
     }
+    virtual void set_response_trailer(const http::HeaderKeyValue& kv)
+    {
+      set_response_trailer(kv.first, kv.second);
+    }
```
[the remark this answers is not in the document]
MahatiC: I chatted with @jumaffre and it will be eventually supported

src/apps/external_executor/external_executor.cpp
jumaffre: I've added a `/Pub` and `/Sub` endpoints. I'll leave the `/Stream` endpoint in this PR until just before this PR is merged just to demonstrate multiple examples.

src/apps/external_executor/external_executor.cpp
jumaffre: Done. There's now a separate `close()` function, called from within the destructor too.

src/apps/external_executor/external_executor.cpp
jumaffre: Done. The adapter is now much cleaner, and creates the `Stream` for the handler.

src/apps/external_executor/external_executor.cpp
```diff
@@ -159,6 +161,66 @@ namespace externalexecutor
...
+      size_t call_count = 0;
+
+      AsyncMsg(
+        ccf::grpc::DetachedStreamPtr<externalexecutor::protobuf::KVKeyValue>&&
+          stream_,
+        size_t call_count_ = 0) :
+        stream(std::move(stream_)),
+        call_count(call_count_)
+      {}
+    };
```
jumaffre, line 177: Remove async task endpoint before merging this PR.

include/ccf/rpc_context.h
```diff
@@ -140,6 +140,10 @@ namespace ccf
...
```
eddyashton, line 146: Do we need this overload? We don't have it for `set_response_header`, and it seems weird to make those overload sets different.

src/apps/external_executor/external_executor.cpp
```diff
@@ -264,7 +326,7 @@ namespace externalexecutor
                   (http_status)payload.status_code(),
                   std::move(headers),
                   std::move(trailers),
-                  {(const uint8_t*)body_s.data(), body_s.size()}))
+                  {body_s.begin(), body_s.end()}))
```
eddyashton, line 329: IUIC, this is now constructing a `std::vector` here, rather than a span pointing at the existing payload bytes? That's OK for now because the `std::vector` is moved directly into a buffer in the implementation (so there would already have been a copy there, we've just moved it here). But if we're able to copy directly to the target memory range in this `send_response` call, then we'd prefer to copy _from_ the original byes (with a `span`) than this intermediate `vector`.

I think we should prefer spans in general, and only do copies to an owning `std::vector` internally/as late as possible. That makes it possible to eventually remove intermediate copies entirely, whereas `std::vector`s in the API suggest a requirement for these intermediate copies.

src/apps/external_executor/external_executor.cpp
```diff
@@ -641,19 +701,117 @@ namespace externalexecutor
...
+          return ccf::grpc::make_error(
+            GRPC_STATUS_NOT_FOUND,
+            fmt::format(
+              "Updates for key {} in table {} has no subscriber",
+              payload.key(),
+              payload.table()));
+        }
+
+        return ccf::grpc::make_success();
+      };
```
eddyashton, line 783: This sample isn't actually interacting with the KV at all, so its odd that it uses `KVKeyValue`. I think it's a dangerous sample, because it _looks_ like a event-based notification of KV modifications, but this is executing entirely before commit/consensus.

I think these Pub/Sub endpoints should use new message types, with a simpler key indexing into the `subscribed_keys` table. If we want to build a sample of pub-sub event notification of changes to the KV, I think we need to do that in an indexer (and think about/comment on the reliability guarantees).

src/endpoints/grpc/grpc.h
```diff
@@ -84,6 +33,13 @@ namespace ccf::grpc
           http::headervalues::contenttype::GRPC));
     }
 
+    // Set response header here rather than in set_grpc_response as data stream
+    // may be sent to client before endpoint returns
+    for (auto const& h : default_response_headers)
+    {
+      ctx->set_response_header(h.first, h.second);
+    }
+
```
eddyashton, line 42: I don't think that's safe - what if the endpoint wants to overwrite these default headers? I think this is saying they might be able to, if they call `set_response_header` and overwrite this _before_ their endpoint calls something which triggers a data stream send? If that's the API, we need to make those send calls extremely obvious, and/or make header modification _after_ that send return an error.

tests/external_executor/external_executor.py
```diff
@@ -411,16 +411,89 @@ def compare_op_results(stub, n_ops):
...
+        # Note: There may not be any subscriber yet, so retry until there is one
+        while True:
+            try:
+                s.Pub(KV.KVKeyValue(table=my_table, key=my_key, value=my_value))
+                break
+            except grpc.RpcError:
+                LOG.debug(f"Waiting for subscriber for key {key}")
+            time.sleep(0.1)
+
+        t.join(timeout=3)
```
MahatiC, line 471: why do we need this timeout?

src/endpoints/grpc/grpc.h
jumaffre: Good spot. This was only moved there before I added the `ccf::grpc::Stream` object so I've moved this back to `set_grpc_response`.

tests/external_executor/external_executor.py
```diff
@@ -461,11 +534,12 @@ def run(args):
...
-        network = test_kv(network, args)
-        network = test_simple_executor(network, args)
-        network = test_parallel_executors(network, args)
-        network = test_streaming(network, args)
+        test_executor_registration(network, args)
+        test_kv(network, args)
+        test_simple_executor(network, args)
+        test_parallel_executors(network, args)
+        test_async_streaming(network, args)
+        test_streaming(network, args)
```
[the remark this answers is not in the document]
jumaffre: Merge conflict gone wrong it seems. I have reverted.

tests/external_executor/external_executor.py
jumaffre: I've removed the timeout.

include/ccf/rpc_context.h
jumaffre: I added it so that we can call `ctx->set_response_trailer(make_status_trailer(response.status.code()));` as it has to be called in two different places (unary and stream trailers), and I'd forgotten to call `std::to_string(response.status.code())` in one of them, which compiled fine, but caused weird errors at runtime. I'm open to another way of doing this.

src/apps/external_executor/external_executor.cpp
jumaffre: Done.

src/apps/external_executor/external_executor.cpp
jumaffre: Good point. I've changed this back to a `std::span`, which is now also used for streaming data too.

src/apps/external_executor/protobuf/misc.proto
```diff
@@ -1,7 +1,6 @@
 syntax = "proto3";
 
 import "google/protobuf/empty.proto";
-import "kv.proto";
 
 package temp;
```
MahatiC, line 5: If this is intended to be an example, maybe we can rename this to `example`

jumaffre: Yes. We will move the gRPC-specific endpoints to a new app altogether but I'll leave this to a further PR.

src/apps/external_executor/external_executor.cpp
```diff
@@ -681,18 +688,81 @@ namespace externalexecutor
...
+        std::unique_lock<ccf::pal::Mutex> guard(subscribed_events_lock);
+
+        auto search = subscribed_events.find(event.SerializeAsString());
+        if (search != subscribed_events.end())
+        {
+          if (!search->second->stream_msg(payload))
+          {
+            // Manual cleanup of closed streams. We should have a close
+            // callback for detached streams to cleanup resources when
+            // required instead
```
achamayou, line 744: Does that need a ticket?

src/http/http2_types.h
```diff
@@ -19,12 +20,60 @@ namespace http2
...
+  };
+
+  class DataSource
+  {
+    // Utility class to consume data from underlying data vector in chunks from
+    // nghttp2_data_source_read_callback
+  private:
+    std::vector<uint8_t> data;
+
+    std::span<const uint8_t> span;
```
achamayou, line 39: Not sure why we have the span here? That looks like the ownership is with data, we can return a fresh span when necessary?

achamayou (approved):

src/apps/external_executor/external_executor.cpp
jumaffre: This is item 5. in https://github.com/microsoft/CCF/issues/4325.

src/http/http2_utils.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <nghttp2/nghttp2.h>
+#include <optional>
+
+namespace http2
+{
+  // Functions to create HTTP2 headers
+  static nghttp2_nv make_nv(const uint8_t* key, const uint8_t* value)
+  {
+    // Note: Investigate no copy flags here
```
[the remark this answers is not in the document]
jumaffre: This is item 8. in https://github.com/microsoft/CCF/issues/3342.

src/http/http_rpc_context.h
```diff
@@ -12,6 +12,28 @@
 
 namespace http
 {
+  inline std::vector<uint8_t> error(ccf::ErrorDetails&& error)
+  {
+    nlohmann::json body = ccf::ODataErrorResponse{
+      ccf::ODataError{std::move(error.code), std::move(error.msg)}};
+    const auto s = body.dump();
+
+    std::vector<uint8_t> data(s.begin(), s.end());
```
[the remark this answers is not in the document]
jumaffre: Yes. I think this work is part of the wider "avoid copies whenever possible" in the HTTP/2 stack (see https://github.com/microsoft/CCF/issues/3342).

src/http/http2_types.h
eddyashton: For posterity:

This struct is used to store data we're trying to `send` through http2. When http2 calls us back to copy this data, its target buffer has a capped size. So if we're trying to send a 10MB `data`, we may get called multiple times and told to copy in just 4K at a time. From our perspective, this all happens under a single call we made, but http2 insists on chunking it per-frame or something. When that happens, instead of truncating the start of `data` each time (with potential reallocation/realignment costs) to indicate how much we've written, or maintaining a separate `offset` counting how many bytes we've written, we thought it was neater to use a `span` - it always points at a subrange (and in fact suffix) of `data`, but is cheap to trim the start off of as we advance.

achamayou: @jumaffre can we put what @eddyashton wrote in the comment above? :)

PR #4543, opened 2022-11-14
[Draft - In development] Dynamically deploy ACI's in CI

DomAyre:


DomAyre:
Addresses #4620

scripts/arm_aci.py
```diff
@@ -0,0 +1,199 @@
...
+        type=str,
+    )
+
+    args = parser.parse_args()
+
+    return Deployment(
+        properties=DeploymentProperties(
+            mode=DeploymentMode.INCREMENTAL,
+            parameters={},
+            template={
```
jumaffre, line 110: This JSON schema should probably become a `.jinja` file eventually, so that it's easier to modify, etc. in isolation.

scripts/arm_aci.py
```diff
@@ -0,0 +1,199 @@
...
+                    {
+                        "type": "Microsoft.ContainerInstance/containerGroups",
+                        "apiVersion": "2022-04-01-preview",
+                        "name": f"{args.deployment_name}-{i}",
+                        "location": "westeurope",
+                        "properties": {
+                            "sku": "Standard",
+                            "confidentialComputeProperties": {
+                                "isolationType": "SevSnp",
+                                "ccePolicy": "eyJhbGxvd19hbGwiOnRydWUsImNvbnRhaW5lcnMiOnsibGVuZ3RoIjowLCJlbGVtZW50cyI6bnVsbH19",
```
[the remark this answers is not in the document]
DomAyre: Yeah agreed, ideally with the ability to specify the policy in the non-b64 format then we can do a conversion

scripts/arm_aci.py
DomAyre: Yeah sounds good, I reckon these two comments could be rolled into a "make arm deployments more flexible" issue

PR #4548, opened 2022-11-14
Unskip tests

MahatiC:
Currently a few tests are skipped because the OpenAPI doc isn't updated. This should fix that.

However, this fix should be other way around i.e. update doc and then adjust the `supported_methods` accordingly. But until then, this fix should enable the few tests that are being skipped.


achamayou (approved):

MahatiC:
I found that tests that were previously skipped due the `app` changes are now being run in CI. There are a few (2-3 endpoints) tests that are being skipped for the `js` app - I haven't dug into it but they seem unrelated to the `app` API changes since the same endpoints are working for the `cpp` app.

PR #4549, opened 2022-11-14
Python client + infra: Distinguish exceptions during connection from later IO errors

eddyashton:
This is pulled out of #4418 as a standalone fix.

There are 2 commits on this PR showing the steps - a seemingly innocuous refactor that exposes an underlying flaw (test wasn't testing what we thought it was), and a fix.


tests/e2e_common_endpoints.py
```diff
@@ -180,119 +179,116 @@ def run_large_message_test(
...
-        except infra.clients.CCFConnectionException:
-            # In some cases, the client ends up writing to the now-closed socket first
-            # before reading the server error, resulting in a connection error
-            assert length > threshold
-            assert get_main_interface_errors()[metrics_name] == before_errors_count + 1
-        else:
-            if length > threshold:
-                assert r.status_code == expected_status.value
-                assert r.body.json()["error"]["code"] == expected_code
+        with primary.client("user0") as client:
```
eddyashton, line 182: Diff here is clearer if you ignore whitespace. We're now creating a client in each call to `run_large_message_test`, rather than reusing 1 client for all the tests.

The problem with reusing a client, is that once we've sent a single too-large message, all future attempts (whether legal or also too-large) get an error because _the session they're trying to use was killed earlier_. Since we test with increasing message sizes, this failure looks like one we expected, but happens for the wrong reasons.

Creating separate clients reveals a different bug. When CCF says the request is too big and kills the connection, this was previously reported as a `CCFConnectionException`. That was only supposed to be thrown for errors during _connection establishment_, but was now thrown for much later errors. We have a little loop in `clients.py` which retries `CCFConnectionException`s, so we actually submit _many_ too-large requests (silently re-creating sessions internally, because the underlying session is created when we first call something, not when this client object is created). An assertion in this function fails on the largest body test, because we expected to see an increment of 1 in `request_payload_too_large`, but actually see an increment of 18.

The subsequent fix is in `clients.py` - distinguish errors during connection from later read/write errors, based on what httpx tells us.

tests/e2e_common_endpoints.py
```diff
@@ -209,23 +210,12 @@ def run_large_message_test(
...
-            args.max_http_body_size,
-            args.max_http_body_size + 1,
-            args.max_http_body_size * 2,
-            args.max_http_body_size * 200,
-        ]
-    )
+    def get_sizes(n):
+        ns = [n // 2, n - 10, n - 1, n, n + 1, n + 10, n * 2, n * 20]
+        random.shuffle(ns)
+        return ns
```
eddyashton, line 216: One more fix to try and rule out this kind of order-dependent issue - use the same (relative) sizes for each cap, and test them in a random order.

jumaffre (approved):

MahatiC (approved):
Thanks for the diff explanation!

PR #4553, opened 2022-11-15
Add common_dir for member certs

jeffa5:
Fixes the daily LTS issue


DomAyre:
Looks good to me, worth touching .daily_canary to confirm it fixes the issue?

DomAyre:
I ran locally with and without the change and can confirm it fixes it (at least on virtual build but that shouldn't matter)

DomAyre (approved):

PR #4563, opened 2022-11-16
Plot commits and posts in the analyzer

fotiskoun:
This comes as an addition to the analyzer component, implementing the a plot of commits and posts. In order for this plot to produce correct results, it is dependent, for now, on the structures of the requests produced from the generator component. So, the generator should produce requests like: N posts 1 commit and in the end end add multiple commits in order to get the final commit stage.
This PR also removes some figure settings on the analyzer plots and two prints that are no longer needed.


tests/perf-system/analyzer/analyzer.py
```diff
@@ -130,6 +130,57 @@ def customize_table(self, fields_list: List[str], values_list: List[List]):
...
 
+    def plot_commits(self, df_responses: pd.DataFrame, df_generator: pd.DataFrame):
+        """
+        For submitted requests of the type: N posts 1 commit, this function will
+        plot the number of posted versus committed messages
+        """
+        init_post_id = 0
+        init_time = 0
+        custom_header = "x-ms-ccf-transaction-id"
+        for head in df_responses.iloc[0]["rawResponse"].split("\n"):
```
achamayou, line 141: I would suggest using https://docs.python.org/3/library/http.client.html#http.client.parse_headers

tests/perf-system/analyzer/analyzer.py
```diff
@@ -130,6 +130,57 @@ def customize_table(self, fields_list: List[str], values_list: List[List]):
...
+                    len(post_ids) > 1
+                    and post_ids[-1] == post_ids[-2]
+                    and post_ids[-1] == committed_ids[-1]
+                ):
+                    break
+
+                res_headers = df_responses.iloc[row]["rawResponse"].split("\n")
+                for h in res_headers:
+                    if custom_header in h.split(":")[0]:
+                        # get the latest post_id from the committed message
```
achamayou, line 166: That seems extremely specific to one kind of request/generated payloads. It would be good to be explicit about this, and perhaps make it pluggable.

tests/perf-system/analyzer/analyzer.py
```diff
@@ -130,6 +130,57 @@ def customize_table(self, fields_list: List[str], values_list: List[List]):
...
+                            int(h.split(":")[1][:-1].split(".")[1]) - init_post_id
+                        )
+                committed_ids.append(
+                    int(res_headers[-1].split(":")[1][3:-2]) - init_post_id
+                )
+                time_units.append(
+                    (float(df_responses.iloc[row]["receiveTime"]) - init_time)
+                )
+        plt.figure()
+        plt.scatter(time_units, post_ids, label="Posts", marker="o")
```
achamayou, line 177: Txids, those can be PUT, DELETE etc, anything that causes a write will create a Tx and be assigned a Txid.

tests/perf-system/analyzer/analyzer.py
achamayou, line 166:
```suggestion
                        # get the latest tx_id from the committed message

```

tests/perf-system/analyzer/analyzer.py
fotiskoun: I used parse email_parser and json to parse better. thanks for the suggestion

tests/perf-system/analyzer/analyzer.py
fotiskoun: You are right! I replaced `post` to tx

fotiskoun:
@microsoft-github-policy-service agree [company=""]

fotiskoun:
@microsoft-github-policy-service agree

achamayou (approved):

PR #4564, opened 2022-11-16
Update t_cose to version 1.1.

plietar:
This adds support for RSA and EdDSA signature verification. While not useful for CCF (at least in the short term), this allows applications that link to CCF to use those algorithms. Unfortunately now that CCF includes t_cose, applications aren't free to link in their own version anymore.

https://github.com/laurencelundblade/t_cose/releases/tag/v1.1

These were the commands used to update it:
```
cd 3rdparty/exported
rm -r t_cose
curl -L https://github.com/laurencelundblade/t_cose/archive/refs/tags/v1.1.tar.gz | tar xz
mv t_cose-1.1 t_cose
```


achamayou:
@plietar thank you for the update!

The headers that are under inc/ need to move to t_cose, as they were in the previous version. This is so that the installed and local paths have the same suffix for includes.

3rdparty/exported/t_cose/inc/t_cose/t_cose_common.h
```diff
@@ -13,6 +13,7 @@
 #define __T_COSE_COMMON_H__
 
 #include <stdint.h>
+#include <stdbool.h>
```
achamayou, line 16: Can you move the content of inc/t_cose back to t_cose/t_cose?

achamayou (approved):

3rdparty/exported/t_cose/inc/t_cose/t_cose_common.h
achamayou: As discussed, this may not be necessary now that it's inc/t_cose, let's see if the Daily passes and if it does we're fine.

PR #4567, opened 2022-11-16
Fix installing virtual and sgx together

jeffa5:
When switching between building SGX and Virtual targets we need both
ccf_virtual and ccf_sgx installed, these don't have any conflicts and
install to different locations but the virtual package doesn't like the
full open-enclave that the sgx requires.

This fix makes it possible to install both ccf_virtual and ccf_sgx
packages by letting the virtual package tolerate open-enclave instead of
open-enclave-hostverify.


jumaffre:
Could you touch the `.daily_canary` file as the Daily pipeline tests a few more things that may be impact by this change?

achamayou:
> the virtual package doesn't like the full open-enclave that the sgx requires

What do you mean by that? Is there a problem even if open-enclave-hostverify is installed? If so, that seems like that something we should report to Open Enclave. If not, it seems preferable that the dependencies remain separate.

jeffa5:
> > the virtual package doesn't like the full open-enclave that the sgx requires
> 
> What do you mean by that? Is there a problem even if open-enclave-hostverify is installed? If so, that seems like that something we should report to Open Enclave. If not, it seems preferable that the dependencies remain separate.

With `open-enclave-hostverify` installed (version 0.18.4) then trying to install `open-enclave` (also version 0.18.4) I get the following:

```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following NEW packages will be installed:
  open-enclave
0 upgraded, 1 newly installed, 0 to remove and 4 not upgraded.
Need to get 0 B/33.3 MB of archives.
After this operation, 125 MB of additional disk space will be used.
Selecting previously unselected package open-enclave.
(Reading database ... 188817 files and directories currently installed.)
Preparing to unpack .../open-enclave_0.18.4_amd64.deb ...
Unpacking open-enclave (0.18.4) ...
dpkg: error processing archive /var/cache/apt/archives/open-enclave_0.18.4_amd64.deb (--unpack):
 trying to overwrite '/opt/openenclave/bin/oeverify', which is also in package open-enclave-hostverify 0.18.4
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 /var/cache/apt/archives/open-enclave_0.18.4_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

achamayou:
@jeffa5 I see, they're mutually exclusive. That's unfortunate.

achamayou (approved):

PR #4580, opened 2022-11-17
View history

### LINKED ISSUE #4455: Offline TxID committed checking

When we have a transaction id of something we're waiting to be committed, and the latest committed transaction id we can't easily confirm the status of the transaction id locally, we need to ask the node to confirm it.

e.g. given a transaction id of `3.5` and a committed transaction id of `3.6` we can't tell whether our transaction id was committed or dropped since we don't know if `3.6` is the first txid in term `3` or whether the term started earlier (it could be `2.5` that was committed).

A function available to c++ code that exposes the highest seqno a node has seen in a given raft term, something like

```c++
ApiResult get_last_committed_txid_for_view_v1(ccf::View& view, ccf::SeqNo& seqno);
```

Just asking nodes for the status of a transaction id but this includes round trips and doesn't enable offline checking.

jeffa5:
Fixes #4455

This adds a new function to the C++ API `get_view_history_v1` which can be used to get the transaction id starting each view.
It optionally takes a `since` parameter for speedily catching up on things, e.g. from a client that knew some history.

This also wires up the new function to an addition in the `/app/commit` API endpoint so that all CCF users can request the view history.

This logic aims to enable a larger opportunity for check whether transactions have been committed without the client calling the `/app/tx` endpoint which could cause more work for the CCF nodes.


include/ccf/base_endpoint_registry.h
```diff
@@ -85,6 +85,17 @@ namespace ccf
     BaseEndpointRegistry(
       const std::string& method_prefix_, ccfapp::AbstractNodeContext& context_);
 
+    /** Get the history of the raft view changes.
```
jumaffre, line 88:
```suggestion
    /** Get the history of the consensus view changes.
```

src/endpoints/base_endpoint_registry.cpp
```diff
@@ -18,6 +18,36 @@ namespace ccf
...
+    {
+      if (consensus != nullptr)
+      {
+        if (since < 1)
+        {
+          // views start at 1
+          return ApiResult::InvalidArgs;
+        }
+        const auto view_history = consensus->get_view_history_since(since);
+        for (uint64_t i = 0; i < view_history.size(); i++)
```
jumaffre, line 34:
```suggestion
        for (ccf::View = 0; i < view_history.size(); i++)
```

doc/schemas/node_openapi.json
```diff
@@ -987,6 +996,24 @@
...
+            "in": "query",
+            "name": "view_history",
+            "required": true,
+            "schema": {
+              "$ref": "#/components/schemas/boolean"
+            }
+          },
+          {
+            "in": "query",
+            "name": "view_history_since",
```
achamayou, line 1010: Perhaps arguably better called "view_history_since_seqno" or "from_seqno", since it takes a single integer, which could equally be a seqno or a view?

achamayou: Hah, it was view, and "view_history_since_view" does seem a bit silly.

src/endpoints/common_endpoint_registry.cpp
```diff
@@ -68,29 +70,91 @@ namespace ccf
...
+        }
+      }
+
+      return make_success(out);
     };
     make_command_endpoint(
       "/commit", HTTP_GET, json_command_adapter(get_commit), no_auth_required)
       .set_auto_schema<GetCommit>()
+      .add_query_parameter<bool>(view_history_param_key)
+      .add_query_parameter<ccf::View>(view_history_since_param_key)
```
jumaffre, line 157: I think these should be optional not to break compatibility with previous versions?

src/endpoints/base_endpoint_registry.cpp
```diff
@@ -18,6 +18,36 @@ namespace ccf
...
+  ApiResult BaseEndpointRegistry::get_view_history_v1(
+    std::vector<ccf::TxID>& history, ccf::View since)
+  {
+    try
+    {
+      if (consensus != nullptr)
+      {
+        if (since < 1)
+        {
+          // views start at 1
```
achamayou, line 30: Actually, CCF views start at 2, for historical reasons. But we might fix that one day.

tests/e2e_logging.py
```diff
@@ -1588,6 +1588,61 @@ def test_committed_index(network, args):
     }
 
 
+@reqs.description("Test the commit endpoints view_history feature")
+def test_commit_view_history(network, args):
```
jumaffre, line 1592: It seems that this test would make more sense in `election_test.py`? And possible worth included in the `test_suite.py` too?

tests/e2e_logging.py
```diff
@@ -1588,6 +1588,61 @@ def test_committed_index(network, args):
...
+        assert type(res.body.json()["view_history"]) == list
+        view_history = res.body.json()["view_history"]
+
+        # view_history should override the view_history_since
+        res = c.get("/app/commit?view_history=true&view_history_since=2")
+        assert res.status_code == http.HTTPStatus.OK
+        assert res.body.json() == {"transaction_id": txid, "view_history": view_history}
+
+        # ask for since the current view
+        res = c.get("/app/commit?view_history_since=2")
```
jumaffre, line 1619: The current view is returned by the `GET /node/network` endpoint.

jumaffre (commented):
Could also include two new changelog entries, once for the existing `commit` endpoint that now returns the view history, and one for the new C++ API?

tests/e2e_logging.py
```diff
@@ -1588,6 +1588,61 @@ def test_committed_index(network, args):
...
+    remote_node, _ = network.find_primary()
+    with remote_node.client() as c:
+        # the original endpoint should still work
+        res = c.get("/app/commit")
+        assert res.status_code == http.HTTPStatus.OK
+        assert "view_history" not in res.body.json()
+        txid = res.body.json()["transaction_id"]
+
+        # non-true view_history should still work
+        res = c.get("/app/commit?view_history=nottrue")
```
achamayou, line 1602: This seems surprising, it would perhaps be more user friendly to return a 400 that says "Invalid value, view_history must be one of true, false".

tests/e2e_logging.py
```diff
@@ -1588,6 +1588,61 @@ def test_committed_index(network, args):
...
+        assert res.status_code == http.HTTPStatus.BAD_REQUEST
+        assert res.body.json() == {
+            "error": {
+                "code": "InvalidQueryParameterValue",
+                "message": "Error code: InvalidArgs",
+            }
+        }
+
+        # ask for something that doesn't exist yet
+        res = c.get("/app/commit?view_history_since=100")
```
achamayou, line 1641: Arguably 404?

achamayou (approved):

src/endpoints/common_endpoint_registry.cpp
jeffa5: Sneaky default params.

src/endpoints/base_endpoint_registry.cpp
jeffa5: Seemingly so from the user perspective but I think the internal consensus code differs: https://github.com/jeffa5/CCF/blob/13f39e69ce18705fc3b4216b64f42de5ad0bbf31/src/consensus/aft/impl/state.h#L20

I've gone with starting from term 1 (which in practice works out nicely with other endpoints too)

tests/e2e_logging.py
jeffa5: Moved, seems it can be a bit flaky now, any ideas as to how I should make it more robust to the leader changing in-between requests.

tests/e2e_logging.py
jeffa5: Changed to be `true` or `false` and return 400 for others.

tests/e2e_logging.py
jeffa5: Yep, changed future views to return 404.

tests/e2e_logging.py
jeffa5: Seems it is when the transaction id changes (just seqno not view) so will just ignore txid in these tests.

tests/election.py
```diff
@@ -57,6 +58,77 @@ def test_kill_primary(network, args):
     return test_kill_primary_no_reqs(network, args)
 
 
+@reqs.description("Test the commit endpoints view_history feature")
+def test_commit_view_history(network, args):
+    remote_node, _ = network.find_primary()
+    with remote_node.client() as c:
+        # the original endpoint should still work
```
jumaffre, line 65:
```suggestion
        # Endpoint works with no query parameter
```

tests/election.py
```diff
@@ -57,6 +58,77 @@ def test_kill_primary(network, args):
...
+        res = c.get("/app/commit")
+        assert res.status_code == http.HTTPStatus.OK
+        assert "view_history" not in res.body.json()
+
+        # non-true view_history should still work
```
jumaffre, line 70:
```suggestion
        # Invalid query parameter
```

tests/election.py
```diff
@@ -57,6 +58,77 @@ def test_kill_primary(network, args):
...
+                "message": "Invalid value for view_history, must be one of [true, false] when present",
+            }
+        }
+
+        # true view_history should list all history
+        res = c.get("/app/commit?view_history=true")
+        assert res.status_code == http.HTTPStatus.OK
+        assert "view_history" in res.body.json()
+        assert type(res.body.json()["view_history"]) == list
+        view_history = res.body.json()["view_history"]
```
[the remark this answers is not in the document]
jeffa5: I've added a check in the `test_kill_primary_no_reqs` test that the view is longer when the new primary comes up and that it the old view history is a prefix.

tests/suite/test_suite.py
```diff
@@ -62,11 +62,13 @@
     e2e_logging.test_rekey,
```
jumaffre, line 62: You may want to touch `,daily_canary` to trigger this suite.

jumaffre (approved):

tests/election.py
```diff
@@ -57,6 +82,77 @@ def test_kill_primary(network, args):
...
+                "message": "Invalid value for view_history_since, must be in range [1, current_term]",
+            }
+        }
+
+        # view_history should override the view_history_since
+        res = c.get(f"/app/commit?view_history=true&view_history_since={current_view}")
+        assert res.status_code == http.HTTPStatus.OK
+        assert res.body.json()["view_history"] == view_history
+
+
```
jumaffre, line 155:
```suggestion
return network
```

PR #4583, opened 2022-11-18
Validate supported endpoints and dispatch

MahatiC:
... to the appropriate Executor.

This needs a bit more work
- [x]  parallel executors test and a more efficient dispatch

### DIFF
--- a/src/apps/external_executor/external_executor.cpp
+++ b/src/apps/external_executor/external_executor.cpp
@@ -715,11 +715,15 @@ namespace externalexecutor
       std::string method = endpoint_ctx.rpc_ctx->get_request_verb().c_str();
       std::string uri = endpoint_ctx.rpc_ctx->get_request_path();
 
-      if (supported_uris.find(method + uri) == supported_uris.end())
+      auto it = supported_uris.find(method + uri);
+
+      if (it == supported_uris.end())
       {
-        throw std::logic_error("Only registered endpoints are supported");
+        throw std::logic_error(
+          "Only registered endpoints are supported. No executor was found");
       }
-      auto executor_id = supported_uris[method + uri].get_executor_id();
+      auto executor_id = it->second.get_executor_id();
+
       return executor_id;
     }
 
--- a/src/apps/external_executor/external_executor.cpp
+++ b/src/apps/external_executor/external_executor.cpp
@@ -709,20 +709,14 @@ namespace externalexecutor
       return std::make_shared<ExternallyExecutedEndpoint>();
     }
 
-    ExecutorId validate_supported_endpoints(
-      ccf::endpoints::EndpointContext& endpoint_ctx)
+    std::optional<ExecutorId> validate_supported_endpoints(
+      std::string method, std::string uri)
     {
-      std::string method = endpoint_ctx.rpc_ctx->get_request_verb().c_str();
-      std::string uri = endpoint_ctx.rpc_ctx->get_request_path();
-
       auto it = supported_uris.find(method + uri);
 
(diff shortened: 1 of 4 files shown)


src/apps/external_executor/external_executor.cpp
```diff
@@ -689,14 +728,29 @@ namespace externalexecutor
...
+      ccf::endpoints::EndpointContext& endpoint_ctx)
+    {
+      std::string method = endpoint_ctx.rpc_ctx->get_request_verb().c_str();
+      std::string uri = endpoint_ctx.rpc_ctx->get_request_path();
+
+      if (supported_uris.find(method + uri) == supported_uris.end())
+      {
+        throw std::logic_error("Only registered endpoints are supported");
+      }
+      auto executor_id = supported_uris[method + uri].get_executor_id();
```
eddyashton, line 741: We should store the iterator here to avoid doing the lookup twice.
```suggestion
      auto it = supported_uris.find(method + uri);
      if (it == supported_uris.end())
      {
        throw std::logic_error("Only registered endpoints are supported");
      }
      auto executor_id = it->second.get_executor_id();
```

May be `it.second->get` at the end there, I forget which way round that goes.

Also, should report the unsupported path in the error, something like "Could not find any executors supporting {method} {uri}".

tests/external_executor/external_executor.py
```diff
@@ -421,10 +450,50 @@ def compare_op_results(stub, n_ops):
...
+            assert r.body.text() == content
+
+    return network
+
+
 def test_logging_executor(network, args):
     primary, _ = network.find_primary()
 
-    credentials = register_new_executor(primary, network)
+    supported_endpoints = {("POST", "/app/log/public"), ("GET", "/app/log/public")}
```
eddyashton, line 493: This should be stored in `LoggingExecutor`, something like `LoggingExecutor.supported_endpoints`? The equivalent for `WikiCacher` would need to be a function (taking the list of topics to construct the concrete endpoints), until we support slug templates.

tests/external_executor/external_executor.py
```diff
@@ -261,12 +270,21 @@ def test_kv(network, args):
...
+    credentials = register_new_executor(
+        primary, network, supported_endpoints=supported_endpoints
+    )
 
     with executor_thread(WikiCacherExecutor(primary, credentials)):
         with primary.client() as c:
             r = c.post("/not/a/real/endpoint")
-            assert r.status_code == http.HTTPStatus.NOT_FOUND
+            body = r.body.json()
+            assert r.status_code == http.HTTPStatus.INTERNAL_SERVER_ERROR
```
eddyashton, line 286: I don't think we should lose this 404, and convert it to a 500. I guess that's coming from the `logic_error` thrown in `validate_supported_endpoints`?

If we instead returned a `std::optional` from that, we could treat nullopt (=> no active executor can handle this request) into a 404.

MahatiC: 500 is being thrown by the test infra (as it encounters the `logic_error`).

Sure, we can return probably a nullopt and convert it in the test, but I'm not sure how that will translate to Executors? It seems better to throw 500 with a clear error message like it is now?

eddyashton: We can still return a descriptive error in a 404. Since this is the error that will be returned when users mis-spell a uri, I think it should definitely be a 404 rather than a 500? 500s should generally indicate something that's definitely not the caller's fault, where some internal invariant is broken and we have nothing more specific to report - I don't think that's the case here.

eddyashton (approved):

PR #4584, opened 2022-11-18
Adding a maintenance page to the doc

shokouedamsr:
Creating a maintenance page under the overview section to clarify what work needs doing after everything it set up and running. This is to cover the full dev lifecycle


doc/overview/maintenance
```diff
@@ -0,0 +1,16 @@
+Maintenance
+===========
```
shokouedamsr, line 2:
```suggestion
### Maintenance
```

doc/overview/maintenance
```diff
@@ -0,0 +1,16 @@
...
+
+In order to keep the CCF network in a healthy state, there are two kinds of maintenance needed; privilaged and un-privilaged. 
+
+Priviliged: Governance
+-----------------------
```
shokouedamsr, line 7:
```suggestion
**Privileged: Governance**
```

doc/overview/maintenance
```diff
@@ -0,0 +1,16 @@
+Maintenance
+===========
+
+In order to keep the CCF network in a healthy state, there are two kinds of maintenance needed; privilaged and un-privilaged.
```
shokouedamsr, line 4:
```suggestion
In order to keep the CCF network in a healthy state, there are two kinds of maintenance needed; privileged and un-privilaged. 
```

doc/overview/maintenance
```diff
@@ -0,0 +1,16 @@
+### Maintenance
```
achamayou, line 1: This needs to be an RST file (see https://www.writethedocs.org/guide/writing/reStructuredText/), with a filename ending in ".rst", and the index.rst file in the same directory needs to be edited to add it to the TOC and add a new tile leading to it.

achamayou: @shokouedamsr thank you for updating the format of the file! For this new file to be reachable in the documentation, there must a be an entry in the table of contents in the overview/index.rst (so that it appears in the navigation), and a tile linking to it (for the main page). See https://github.com/microsoft/CCF/blob/main/doc/overview/index.rst, and use one of the other files, such as `governance.rst` as an example.

shokouedamsr: Thanks @achamayou . I created another PR #4645 to update the overview page.

achamayou (approved):

PR #4585, opened 2022-11-18
Install python scripts using entry_points

jeffa5:
This ensures that they will be runnable with just the filename, not requiring something like `python3 $(which read_ledger.py)`.

c.f. https://docs.python.org/3/distutils/setupscript.html#installing-scripts


jeffa5:
Note, installing the python package with `pip install -e ./python` will do other things which are different from `pip install ./python` which this is fixing.

jumaffre (approved):
Thank you!

jeffa5:
After chatting with @achamayou I'm going to update it to use entry_points

python/ccf/split_ledger.py
```diff
@@ -168,12 +168,15 @@ def run(args_):
...
     LOG.remove()
     LOG.add(
         sys.stdout,
         format="<level>{message}</level>",
     )
 
     run(sys.argv[1:])
+
+
+if __name__ == "__main__":
```
achamayou, line 181: This is unnecessary when using entry_points.

python/ccf/read_ledger.py
```diff
@@ -244,3 +243,7 @@ def run(
         args.digests_only,
     ):
         sys.exit(1)
+
+
+if __name__ == "__main__":
```
achamayou, line 248: This is unnecessary when using entry_points.

MahatiC (approved):

achamayou (approved):

PR #4587, opened 2022-11-18
Make container build job concurrent

achamayou:
The current job is quite repetitive, and takes nearly an hour. This is attempting to make it shorter and quicker.

It does get rid of the LTS tag for now, while I try to work out how to put it back.

Edit: tested on branch, seems to work ok: https://github.com/microsoft/CCF/actions/runs/3499137408

Do we still need the LTS tag on 4.x?


eddyashton (approved):

achamayou:
Moving to draft, until I've tested this with the advice @jumaffre provided.

.github/workflows/containers.yml
```diff
@@ -11,82 +12,25 @@ env:
 
 jobs:
   build_and_publish:
-    name: "Build and publish containers for all platforms"
+    name: "Build & push"
```
achamayou, line 15: Temporary for testing, do not push.

.github/workflows/containers.yml
```diff
@@ -11,8 +11,16 @@ env:
...
```
achamayou, line 14: Make shorter on purpose, so that GitHub appended matrix parameters don't get truncated in UI.

jumaffre (approved):

Issue #4591, opened 2022-11-18
Run release pipeline daily


jumaffre:
We should be able to run the release pipeline for `main` and active LTS release branches daily by tagging the head of each branch with `ccf-latest-main` or `ccf-latest-N.x`. 

Tasks:
- [ ] Fix LTS compatibility test
- [ ] Build docker images in this job (without publishing)
- [ ] Create separate release pipeline in Azure DevOps

PR #4592, opened 2022-11-18
LTS compatibility: cope with `daily-` tags

jumaffre:
Part of https://github.com/microsoft/CCF/issues/4591

As we want to be able to regularly release and tag the head of `main` and active `release/N.x` branches, this PR fixes the LTS compatibility test to be able to cope and parse this new tag meaningfully.


achamayou:
@jumaffre I would prefer if they weren't prefixed with ccf-*, because that's what we have the tag protection rule on. Can they just be latest-*?

eddyashton:
I'd prefer if we avoided using `latest` in this name as well. I think `latest` is generally an alias for a real release, if it exists. Can we instead call this `daily` or `nightly`, to make it more obvious that it's created automatically and regularly (and as a result, less stability guarantees).

jumaffre:
> I'd prefer if we avoided using `latest` in this name as well. I think `latest` is generally an alias for a real release, if it exists. Can we instead call this `daily` or `nightly`, to make it more obvious that it's created automatically and regularly (and as a result, less stability guarantees).

Good point. I've renamed it to `daily-`.

achamayou (approved):

eddyashton (approved):

PR #4593, opened 2022-11-18
Add piccolo to ccf

fotiskoun:
Integrate piccolo to the CCF ci to run tests


tests/infra/piccolo_driver.py
```diff
@@ -0,0 +1,315 @@
...
+
+def my_configure_remote_client(args, client_id, client_host, node, command_args):
+    if client_host == "localhost":
+        client_host = infra.net.expand_localhost()
+        remote_impl = infra.remote.LocalRemote
+    else:
+        remote_impl = infra.remote.SSHRemote
+    try:
+        remote_client = infra.remote_client.CCFRemoteClient(
+            "client_" + str(client_id),
```
achamayou, line 60:
```suggestion
            f"client_{client_id}",
```

tests/infra/piccolo_driver.py
```diff
@@ -0,0 +1,315 @@
...
+def run(get_command, args):
+    if args.fixed_seed:
+        seed(getpass.getuser())
+
+    hosts = args.nodes
+    if not hosts:
+        hosts = ["local://localhost"] * minimum_number_of_local_nodes(args)
+
+    args.initial_user_count = 3
+    args.sig_ms_interval = 1000  # Set to cchost default value
```
achamayou, line 88:
```suggestion
    args.sig_ms_interval = 100
```

We don't want 1 second delay at the end of the test if we don't end up on a round 1000 txs, which we probably won't.

tests/infra/remote.py
```diff
@@ -451,8 +451,9 @@ def _setup_files(self):
             src_path = os.path.normpath(os.path.join(os.getcwd(), path))
             assert self._rc("ln -s {} {}".format(src_path, dst_path)) == 0
         for path in self.data_files:
-            dst_path = os.path.join(self.root, os.path.basename(path))
-            self.cp(path, dst_path)
+            if len(path) > 0:
```
achamayou, line 454: Why is this check here? We shouldn't be passing empty paths, and this seems to be hiding that?

achamayou (approved):

tests/infra/remote.py
fotiskoun: This was for the `--config` option, as I set the default to empty string. I will try removing this option, as it is not required in my case

achamayou: @fotiskoun this is fine for now, but please create a ticket with the details of what needs to be addressed.

achamayou:
@fotiskoun 

```
/build/apache-arrow-10.0.0/cpp/src/arrow/result.cc:28: ValueOrDie called on an error: IOError: Failed to open local file '../../piccolo_logging_100ktxs.parquet'. Detail: [errno 2] No such file or directory
```

PR #4598, opened 2022-11-21
Re-enable COSE tests in suite

achamayou:
Link #4419


tests/governance.py
```diff
@@ -670,8 +672,9 @@ def test_cose_ack(network, args):
 @reqs.description("Test COSE proposal")
 def test_cose_proposal(network, args):
     primary, _ = network.find_primary()
-    identity = network.identity("member0")
-    other_identity = network.identity("member1")
+    members = network.consortium.get_active_members()
+    identity = network.identity(members[0].local_id)
+    other_identity = network.identity(members[1].local_id)
```
jumaffre, line 677: What if there's only 1 member?

achamayou: Then the test won't work.

jumaffre: I think we may end up in such scenario as the test suite may have removed some member(s). We should enforce that there's at least two members for this test to run.

achamayou: I think there's two other possibilities:

1. change the logic to match that in consortium and iterate over a majority, regardless of how many members are active
2. scrap these tests entirely, and replace them with parametrisations that use cose/http, and have the suite pick from both

The problem with #2 is that this may lead to four member removals in a row, if we dupe all the governance steps. But we could have prelude in some governance tests that randomly pick cose vs http if run in the suite.

achamayou: Ok, after discussion, I think this is now addressed.

PR #4607, opened 2022-11-22
Fix containers and PyPi release workflows

jumaffre:
As this is currently failing for `4.0.0-dev0`.

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -3,6 +3,9 @@ name: "Build and Publish Release Containers to MCR"
 on:
   release:
     types: [published]
+  push:
+    branches:
+      - "fix_containers_pypi"
 
 env:
   ACR_REGISTRY: ccfmsrc.azurecr.io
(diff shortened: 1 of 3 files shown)


.github/workflows/containers.yml
```diff
@@ -37,4 +40,4 @@ jobs:
         run: docker login -u $ACR_TOKEN_NAME -p ${{ secrets.ACR_APP_PUSH_TOKEN_PASSWORD }} $ACR_REGISTRY
 
       - name: Push App container
-        run: docker push $ACR_REGISTRY/public/ccf/app/${{ matrix.type }}:${{steps.tref.outputs.tag}}-${{ matrix.platform }}
+        run: docker push $ACR_REGISTRY/public/ccf/app/${{ matrix.type }}${{ matrix.run_js && '-js' || '' }}:${{ steps.tref.outputs.tag }}-${{ matrix.platform }}
```
achamayou, line 43: Argh, sorry, that's the part I couldn't test...

jumaffre: No worries! I was going to say that we should test this workflow more often (#4591), but we wouldn't push to MCR then, so I guess that's inevitable.

achamayou: v_v

achamayou (approved):

PR #4628, opened 2022-11-23
Amend sandbox.sh to define where the constitution files are

### LINKED ISSUE #4572: Start the sandbox with a configurable constitution path
At the moment, the sandbox.sh uses a variable `$PATH_HERE` for the location of the `/opt/ccf/bin` and the constitution files are loaded from this directory.  This is fine, but if we want to use the same sandbox.sh for different applications then we can't override the constitution files in `/opt/ccf/bin`, but rather we should supply a location where the applications constitution files are.

# Proposal
Allow a --constitution-path argument in sandbox.sh. The default value will be `$PATH_HERE` to maintain the current behaviour. If supplied it will allow the user to define the location of the 4 constitution files.

ross-p-smith:
There are two parts to this PR. I require a mechanism to be able to use the sandbox with different samples in the same repsoitory. Currently the constitution has to be in the same location as  the sandbox.sh script. This will cause problems if we want to be able to run different samples in the same repo.

This change allows a path to be passed in so the sandbox script passes the correct `--constitution` parameter in. It will default to the existing behaviour if not passed in.

When we are parsing the arguments, we remove `constitution-dir` as this will be unrecognised by  `start_network.py`.

Once this PR is live, I can amend the ccf-app-samples so that a path is passed in for each sample.

Fixes #4572


achamayou (approved):

tests/sandbox/sandbox.sh
```diff
@@ -16,24 +17,34 @@ PLATFORM_FILE="${PATH_HERE}"/../share/PLATFORM
...
         --js-app-bundle)
             is_js_bundle_specified=true
+            extra_args+=("$1" "$2")
             shift
             ;;
         --js-app-bundle=*)
             is_js_bundle_specified=true
+            extra_args+=("$1")
+            ;;
+        -c|--constitution-path)
```
jumaffre, line 41: Should this be `--constitution-dir`, that expects to find all 4 `*.js` files?

jumaffre (approved):

tests/sandbox/sandbox.sh
ross-p-smith: Fixed

jumaffre (approved):

PR #4633, opened 2022-11-24
Forward in forwarding tests

eddyashton:
Our forwarding tests weren't actually testing forwarding, because the client-on-backup they created isn't used by the `network.txs.issue` call. The core fix here is to add `on_backup=True` to these calls, the related removals are just removing misleading/unused junk.


MahatiC (approved):

tests/e2e_logging.py
```diff
@@ -1118,14 +1118,19 @@ def test_forwarding_frontends(network, args):
         ack = network.consortium.get_any_active_member().ack(backup)
         check_commit(ack)
 
-    with backup.client("user0") as c:
-        check_commit = infra.checker.Checker(c)
-        check = infra.checker.Checker()
-        msg = "forwarded_msg"
-        log_id = 7
-        network.txs.issue(network, 1, idx=log_id, send_public=False, msg=msg)
-        check(network.txs.request(log_id, priv=True), result={"msg": msg})
```
MahatiC, line 1127: I'm not sure why was this added in the first place? So I'm unable to assess if it's correct to remove this :)

eddyashton: Before we had `network.txs.issue`, the line about was a direct call to something like `c.post("/app/log/private")`. After that post, we wanted to check that the message had been posted (we could `GET` it), and committed. I believe that's now all done _inside_ `.issue`, so we don't need to repeat it here.

achamayou (approved):

PR #4634, opened 2022-11-24
Misc fixes to support HTTP/2 streaming

jumaffre:
Related to #4537 

This PR contains two independent improvements identified when implementing server-side streaming:
- As clients may close a connection while a detached stream is held by the application, the `HTTPResponder` used to access an invalid `server_parser` (held by the closed `HTTP2ServerSession`). We now hold a `weak_ptr` to `server_session` so that we can easily detect when the session has been closed by the client.
- As the `ResponderLookup` object is shared by all HTTP/1 and HTTP/2 sessions, we need a lock to access/modify its content.


src/apps/external_executor/external_executor.cpp
```diff
@@ -260,11 +260,19 @@ namespace externalexecutor
...
+                (http_status)payload.status_code(),
+                std::move(headers),
+                std::move(trailers),
+                {(const uint8_t*)body_s.data(), body_s.size()});
+            }
+            catch (const std::exception& e)
+            {
+              LOG_FAIL_FMT(
+                "Could not send response back to client: {}", e.what());
+            }
```
eddyashton, line 275: Why is this a try-catch rather than an error return value?

jumaffre: I'm not sure and was meant to write a comment about it on the PR. 

While we now detect when the session was closed while we were holding the detached stream, and could return a `bool` here, I believe there could still be some errors in the HTTP/2 layer further down that could result in an exception that would bubble up to here. If the application didn't catch this, this could lead to a node crash as we probably call `send_response()` (or `stream_data()`) from outside the synchronous execution of an endpoint (which catches all exceptions).

eddyashton: I think we should view this the other way round - since lower things may throw, it's the responsibility of `send_response` to _catch_ those errors, and convert them to an error return value. That way we remove this responsibility from the callers.

eddyashton (approved):

achamayou (approved):

PR #4652, opened 2022-11-28
Create minimal pipeline for SNP which is stable

DomAyre:
Just deploys an ACI for now


achamayou (approved):

scripts/arm_template.py
```diff
@@ -0,0 +1,125 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+ # Licensed under the Apache 2.0 License.
```
achamayou, line 2:
```suggestion
# Licensed under the Apache 2.0 License.
```

DomAyre:
We need to merge `azure_pipelines_snp.yml` before we can create the pipeline, so in the meantime I tested the changes in this PR on the existing pipeline https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=56311&view=results and it works as expected

Issue #1920, opened 2020-11-24
Add summary and description to OpenAPI definition for C++ endpoints


letmaik:
The auto-generated OpenAPI doc lacks endpoint summaries and descriptions for C++ endpoints, and hence the docs lack them too. I think adding them would go a long way to help in understanding CCF's API.

Related to https://github.com/microsoft/CCF/issues/1918. The [Azure guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#api-definition) only say there must be an OpenAPI definition, but doesn't say how detailed it should be.

For inspiration, https://github.com/Azure/azure-rest-api-specs contains OpenAPI 2.0 (Swagger) definitions for all (most?) Azure services, for example [Service Fabric's data plane API](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/servicefabric/data-plane/Microsoft.ServiceFabric/stable/7.2/servicefabric.json).

achamayou:
Several built-in endpoints (ie. included on all frontends, such as /api https://github.com/microsoft/CCF/blob/main/doc/schemas/app_openapi.json#L355) are still missing descriptions. We should at least cover those.

achamayou:
All built-ins (ie. the common endpoint registry) have either a description or a summary. Work to standardise and document the API are taking place in CADL now, see #4669.

PR #4654, opened 2022-11-28
Add summary and description to openapi

achamayou:
Addresses some of #1920. Common endpoints and governance are covered, but I will hold off writing all the descriptions until we've completed our API review in the context of #4469.

I would like to merge this in the meantime.


tests/infra/network.py
```diff
@@ -1047,7 +1047,7 @@ def wait_for_all_nodes_to_commit(self, primary=None, tx_id=None, timeout=10):
                 with node.client() as c:
                     logs[node.node_id] = []
                     resp = c.get(
-                        f"/node/local_tx?transaction_id={tx_id}",
+                        f"/node/tx?transaction_id={tx_id}",
```
eddyashton, line 1050: Once upon a time, `local_tx` was never forwarded, and that was necessary for some tests to pass. Looks like that's no longer the case though, so if the CI is happy let's purge it.

eddyashton (approved):
LGTM!

tests/infra/network.py
achamayou: Yes, I think it ceased being useful when "set_execute_outside_consensus" was removed.

src/endpoints/common_endpoint_registry.cpp
```diff
@@ -237,15 +240,10 @@ namespace ccf
...
-      "/local_tx",
-      HTTP_GET,
-      json_command_adapter(get_tx_status),
-      no_auth_required)
-      .set_auto_schema<void, GetTxStatus::Out>()
-      .add_query_parameter<ccf::TxID>(tx_id_param_key)
+      .set_openapi_summary("Current status of a transaction")
+      .set_openapi_description(
+        "Possible statuses returned are Unknown, Pending, Committed or "
+        "Invalid.")
```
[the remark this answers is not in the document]
achamayou: I thought so too at first, but our OpenAPI generation isn't that good, and so it's not there: https://microsoft.github.io/CCF/main/use_apps/rpc_api.html#get--app-tx

I don't think it's likely the statuses will change any time soon, or any sooner than the move to CADL anyway, so I claim this is fine for now.

PR #4657, opened 2022-11-28
Basic remote job on azure pipelines

DomAyre:


.azure_pipelines_snp.yml
```diff
@@ -33,28 +33,47 @@ jobs:
       vmImage: ubuntu-20.04
     steps:
       - script: |
+          echo -e "\nGenerating an SSH key for use with the executors..."
+          ssh-keygen -t rsa -b 4096 -f ~/.ssh/id_rsa -N ""
+          PUB_KEY=`cat ~/.ssh/id_rsa.pub`
+          PRIV_KEY=`cat ~/.ssh/id_rsa | tr '\n' '?'`
```
DomAyre, line 39: This replacement of newlines is because azure pipelines can't handle multi-line variables, my approach of replacing them with `?` is probably dodgy as I think they might be able to appear in the private key? So open to suggestions of other characters to use or other approaches entirely

DomAyre: Amaury helpfully suggested base64'ing it instead which makes much more sense, so I'll do that

DomAyre: Addressed in https://github.com/microsoft/CCF/pull/4657/commits/8b4e0425f11f1e47f0d4663e21537b1e7f36e951

.azure_pipelines_snp.yml
```diff
@@ -33,28 +33,47 @@ jobs:
       vmImage: ubuntu-20.04
     steps:
       - script: |
+          echo -e "\nGenerating an SSH key for use with the executors..."
```
jumaffre, line 36:
```suggestion
          set -ex
          echo -e "\nGenerating an SSH key for use with the executors..."
```

.azure_pipelines_snp.yml
```diff
@@ -33,28 +33,47 @@ jobs:
       vmImage: ubuntu-20.04
     steps:
       - script: |
```
jumaffre, line 35: More of a cleanup job that can probably wait a further PR: we could split this single step into multiple for easier debugging/nicer output in Azure DevOps.

DomAyre: Great idea
https://github.com/microsoft/CCF/pull/4662/checks?check_run_id=9792088968

.azure_pipelines_snp.yml
DomAyre: Sounds good, can I do a proper pass in https://github.com/microsoft/CCF/pull/4662?

jumaffre (approved):

PR #4659, opened 2022-11-29
Generate wrapping endpoints for governance tables

### LINKED ISSUE #4073: Add an endpoint for querying public governance tables
To avoid manually defining wrapping endpoints for everything we want to expose, we should have a generic `GET` endpoint for anything in a public governance table. Should be fetchable by key, and the entire table. Requires some static knowledge of tablename-to-tabletype for these tables.

I think ideally this would look like:
```
GET /gov/kv?public:ccf.gov.nodes
```

But that has awkward encoding implications (colons in paths). A simpler alternative might be:
```
GET /gov/kv/nodes
```

This assumes every table we care about is prefixed `public:ccf.gov.`, so we can manually add that.

Also not tied to `/gov/kv`, maybe `/gov/everything`.

eddyashton:
Resolves #4073.

Adds several `/gov/kv/{}` endpoints, generated semi-automatically, wrapping direct reads of the KV.

Note that this approach differs from what was discussed in the issue. Rather than a generic endpoint which could also return the contents of custom governance tables but requires the caller to specify encoding formats, this uses the known types from the KV to directly convert but only supports builtin tables (and only those explicitly listed!).


src/node/rpc/member_frontend.h
```diff
@@ -400,6 +401,128 @@ namespace ccf
...
+      constexpr auto gov_prefix = "public:ccf.gov.";
+      constexpr auto internal_prefix = "public:ccf.internal.";
+      if (table_name.starts_with(gov_prefix))
+      {
+        table_name.remove_prefix(strlen(gov_prefix));
+      }
+      else if (table_name.starts_with(internal_prefix))
+      {
+        table_name.remove_prefix(strlen(internal_prefix));
+      }
```
eddyashton, line 456: Note: This exposes the table `public:ccf.gov.foo.bar_baz` at `/gov/kv/foo.bar_baz`. If there's also a `public:ccf.internal.foo.bar_baz`, we'll have a conflict. We think that's fine - we don't have such conflicts and won't add them, if we did we'd prefer to keep the `gov` tables. We want to provide shorter/simpler URIs, so we don't duplicate the `gov` section.

jumaffre, line 456: What's the use case for exposing internal tables? This seems inconsistent with the `/gov` endpoint URI prefix and probably not worth exposing anyway?

src/node/rpc/member_frontend.h
```diff
@@ -400,6 +401,128 @@ namespace ccf
...
+      else
+      {
+        throw std::logic_error(fmt::format(
+          "Should only be used to wrap governance tables. '{}' is not "
+          "supported",
+          table_name));
+      }
+
+      auto endpoint = make_read_only_endpoint(
+        fmt::format("/kv/{}", table_name),
```
jumaffre, line 466: Because we've gone the governance-specialised rather than generic way, I think it'd be nice to replace the `.` with `/` in the table name so that it's more RESTful? Looking at https://microsoft.github.io/CCF/main/audit/builtin_maps.html, I can see that we've not been very consistent so far (e.g. `nodes.info` vs. `proposals_info`) but I think it'd be nice regardless?

jumaffre (approved):
I wonder if we should mark existing endpoints that will be superseded by this as deprecated too?

MahatiC (approved):

src/node/rpc/member_frontend.h
eddyashton: I thought there were a few internal tables that might be useful to query, that internal was used semi-interchangeably with governance, and that it'd be useful to have a generic API for these things for debugging unexpected behaviour in live services. Looking at the list I can't actually see any use cases for accessing the internal tables, it's probably only `recovery_shares` that might be interesting, so I'm happy to drop this.

src/node/rpc/member_frontend.h
eddyashton: I'm on the fence about this, but for similar reasons as were listed in #4073 and then superceded - doing more mangling of the name makes it harder to do lookup (in either direction), and if these look like RESTy resources then they look like badly designed RESTy resources (`/gov/kv/nodes/*` vs `/node/network/nodes/*` - similar names, and overlapping content, how are you supposed to know which to call?).

But I think you're right - if we're going down this route we should go all the way.

src/node/rpc/member_frontend.h
eddyashton: Resolved - we no longer expose internal tables like this, so no chance of collision.

src/node/rpc/member_frontend.h
eddyashton: This is now done.

eddyashton:
> I wonder if we should mark existing endpoints that will be superseded by this as deprecated too?

Maybe. Do we have a way to deprecate standard endpoints, other than calling them out in the CHANGELOG? I think we should postpone this, and do a pass deprecating many undocumented/duplicated endpoints

tests/governance_history.py
```diff
@@ -96,7 +96,15 @@ def check_operations(ledger, operations):
     assert operations == set(), operations
 
 
-def check_all_tables_are_documented(ledger, doc_path):
+def get_all_tables_from_ledger(ledger):
+    table_names_in_ledger = set()
+    for chunk in ledger:
+        for tr in chunk:
+            table_names_in_ledger.update(tr.get_public_domain().get_tables().keys())
```
jumaffre, line 103: Not that it matters much but there's an existing `get_latest_public_state()` function in the `Ledger` class.

eddyashton: Ta, using that now.

PR #4660, opened 2022-11-29
Add gRPC support for historical queries

MahatiC:
~~WIP and not ready for review.~~

This PR adds support for:

~~- [ ] `Subscribe` endpoint where `Executors` are able to subscribe for specific keys in specific tables~~ (Subscribe endpoints will be in a separate PR)
- [x] `Historical` look-up endpoint which takes a key and a table name (along with TXID header in request) and returns its value, much like the existing endpoint


src/apps/external_executor/protobuf/historical.proto
```diff
@@ -0,0 +1,28 @@
...
+  int32 view = 3;
+  int32 seqno = 4;
+}
+
+message RetryQuery { bool retryquery = 1; }
+message QueryResponse
+{
+  externalexecutor.protobuf.KVValue data = 1;
+  RetryQuery retry = 2;
+}
```
MahatiC, line 28: Note: Naming of this service will be eventually merged with the indexing service

tests/external_executor/external_executor.py
```diff
@@ -481,8 +481,25 @@ def test_logging_executor(network, args):
...
+            assert r.status_code == 200
+            assert r.body.json()["msg"] == log_msg
+
+            # post to private table
+            r = c.post("/app/log/private", {"id": log_id, "msg": log_msg})
+            assert r.status_code == 200
+            tx_id = r.headers.get("x-ms-ccf-transaction-id")
+
+            # make a historical query
+            time.sleep(3)
```
achamayou, line 495: Why do we sleep here?

tests/external_executor/external_executor.py
```diff
@@ -481,8 +481,25 @@ def test_logging_executor(network, args):
...
+            headers = {"x-ms-ccf-transaction-id": tx_id}
+            r = c.get(f"/app/log/private/historical?id={log_id}", headers=headers)
+            assert r.status_code == 202
+
+            time.sleep(3)
```
achamayou, line 500: In general, we shouldn't have any sleeps in our tests outside of a retry loop on a well defined condition. For asynchronous operations, sleeping is either brittle (when too short), or wasteful (when too long). A retry loop with a timeout is efficient (passes as soon as the condition is met) and robust (the timeout can be set to a value so large that it isn't triggered by temporarily slow I/O or network).

MahatiC: Makes sense, I've now modified it to a retry loop.

tests/external_executor/external_executor.py
```diff
@@ -481,11 +481,44 @@ def test_logging_executor(network, args):
...
+            timeout = 3
+            start_time = time.time()
+            end_time = start_time + timeout
+            success_msg = ""
+            while time.time() < end_time:
+                headers = {"x-ms-ccf-transaction-id": tx_id}
+                r = c.get(f"/app/log/private/historical?id={log_id}", headers=headers)
+                if r.status_code == http.HTTPStatus.OK:
+                    assert r.body.json()["msg"] == log_msg
+                    success_msg = log_msg
```
eddyashton, line 504:
```suggestion
                    break
```

I think we've asserted equality on the line above, and can end the loop here?

src/apps/external_executor/protobuf/historical.proto
```diff
@@ -0,0 +1,28 @@
...
+  rpc GetHistoricalData(HistoricalData) returns (QueryResponse) {}
+}
+
+message HistoricalData
+{
+  string map_name = 1;
+  bytes key = 2;
+  int64 view = 3;
+  int64 seqno = 4;
+}
```
eddyashton, line 21: `view` and `seqno` should be stored as `uint64`, to match the C++ types.

It might be helpful to pull the view and seqno pair into a separate `TxID` type, to distinguish them/make it clear they're a closely-related pair? The latter is a fuzzy suggestion, unclear benefit.

```suggestion
message TxID
{
  uint64 view = 1;
  uint64 seqno = 2;
}

message HistoricalData
{
  string map_name = 1;
  bytes key = 2;
  TxID tx_id = 3;
}
```

src/apps/external_executor/protobuf/historical.proto
```diff
@@ -0,0 +1,28 @@
...
+
+message RetryQuery { bool retryquery = 1; }
+message QueryResponse
+{
+  externalexecutor.protobuf.KVValue data = 1;
+  RetryQuery retry = 2;
```
eddyashton, line 27: Why is this a nested message, rather than an inline `bool`? It looks like the sample Python code only checks this top-level field (`result.HasField("retry")`), and the nested bool is unused.

tests/external_executor/external_executor.py
MahatiC: Yup, ended the loop. But I didn't remove the `success_msg = log_msg` line as it confirms that the query succeeded and not timed out.

eddyashton: Ah, that makes sense. You can also do a `while ... else: raise TimeoutError` loop in Python. The `else` block is only executed when the loop exits because the condition fails (ie - the timeout expired), and will be skipped if the loop is executed by a `break`.

eddyashton (approved):

achamayou (approved):

PR #4663, opened 2022-11-30
Add a UBSan suppression for QCBOR.

plietar:
QCBOR has a bug which can cause memmove to be called with a NULL source argument (and zero size), which is technically undefined behaviour and triggers a UBSan violation.

This bug cannot be reached by CCF's use of QCBOR (it mostly occurs when encoding messages, which CCF doesn't do), but it affects downstream application which use QCBOR themselves and are forced to use the copy of the library provided by CCF.

https://github.com/laurencelundblade/QCBOR/issues/164

### DIFF
--- a/src/ubsan.suppressions
+++ b/src/ubsan.suppressions
@@ -10,4 +10,7 @@ src:*/3rdparty/exported/quickjs/quickjs.h
 src:*/3rdparty/exported/quickjs/cutils.c
 src:*/3rdparty/exported/quickjs/cutils.h
 
+# https://github.com/laurencelundblade/QCBOR/issues/164
+fun:UsefulOutBuf_InsertUsefulBuf
+
 #############################################################################


achamayou:
> forced to use the copy of the library provided by CCF

To be clear, this is a result of circumstances that are outside the control of CCF: the fact that SGX enclaves must be a single shared object, and that QCBOR does not namespace symbols. CCF is constrained by the same restrictions with regards to Open Enclave and its dependencies.

This is one of several reasons why CCF is moving towards external executors instead.

eddyashton (approved):

plietar:
@achamayou sorry I didn't mean to put any blame on CCF. I used "forced" more in the sense of "forced by the nature of linking C++ applications into a single binary". I did in fact try to play around with symbol visibility to allow two copies of QCBOR and t_cose to be linked in the binary, but it seems too impractical to work.

Where do you stand on backports to the 3.x branch now? Would be nice to have this change included in future releases, but if not we can work around it (we already rebuild CCF from source in our ASAN CI, so can cherry pick fixes. We used to do that already when we were targeting 2.x).

achamayou:
> @achamayou sorry I didn't mean to put any blame on CCF. I used "forced" more in the sense of "forced by the nature of linking C++ applications into a single binary". I did in fact try to play around with symbol visibility to allow two copies of QCBOR and t_cose to be linked in the binary, but it seems too impractical to work.

No problem, just wanted to make sure there was no misunderstanding. We've had to do similar things before, for example when we were getting duplicate EverCrypt symbols from several libraries making use of it. We understand the pain, it is indeed impractical.

> Where do you stand on backports to the 3.x branch now? Would be nice to have this change included in future releases, but if not we can work around it (we already rebuild CCF from source in our ASAN CI, so can cherry pick fixes. We used to do that already when we were targeting 2.x).

Happy to have this backported, since it's clearly a risk-free change.

PR #4665, opened 2022-12-01
Include the gitlens extension in CCFs devcontainer

DomAyre:
I think this should be built-in VS code functionality, but until it is, it's a very helpful addition to our devcontainer


takuro-sato (approved):
Agree we should have it.
Confirmed that gitlens is installed in devcontainer using codespace.

PR #4677, opened 2022-12-01
Document read-write KV restrictions in different execution contexts

### LINKED ISSUE #4630: JS KV Access - clarify read-only restrictions in different execution contexts
We enforce some read/write restrictions on KV access from JS, so that (for instance) governance actions can only write to governance tables, and not private or application tables.

The error message when code attempts to write to a disallowed table is unclear ("Cannot call delete on read-only map"). We should improve this error message to explain exactly why this table is read-only. We should also explain this restriction, and the reason behind it, in the docs.

I think there are additional restrictions we should be enforcing here too, to prevent even _read_ access to certain tables in certain contexts (eg - you shouldn't read from _private_ application tables in any governance).

I think the table looks like this:

```
    // Enforce the following access:
    //
    //                                 Namespace of Table
    //           ______________| INTERNAL  GOVERNANCE  APPLICATION
    //           ______________|  PUB|PRIV   PUB|PRIV   PUB|PRIV
    //            gov validate |   RO|X       RO|X       RO|X
    //   Kind     gov resolve  |   RO|X       RO|X       RO|X
    //    of      gov apply    |   RO|X       RW|X       RO|X
    // Execution  app endpoint |   RO|RO      RO|RO      RW|RW
    //            module exec  |   - |-       - |-       - |-
    //
```

(An `X` is illegal, not even readable, so calling `ccf.kv["your_map_name_here"]` should throw a useful error message. The `-`s in module exec are where `ccf.kv` doesn't even exist. A reminder that this is user-written code, so _could_ try to access the KV, and will currently fail with some kind of null property error, but we could potentially make more explicit)

WIP branch is here: https://github.com/eddyashton/CCF/tree/js_read_only_explanations

Note that we're talking here purely about JS, but the application restrictions will (and should) also be enforced for external executors, or any other execution engine where we have the ability to do so.

eddyashton:
Resolves #4630.

A lot of this code is tests to try and confirm that each possible case is tested, and matches what we expect. There's a new page in the docs explaining the restrictions, and the error message shown to users now links to this page.

There's a big TODO we need to decide before this is merged:
- [X] Should governance be able to read from public application tables? I'd reasoned previously that this means governance is operating only over public state, so is auditable. However, public application tables may be derived from private application tables (in application transactions), and we cannot audit this derivation. If the goal is to limit which transactions need to be considered in audit to "governance transactions", it seems we shouldn't depend on any application state? Is this transfer from private to public meaningfully different than how the contents of proposals/ballots was derived? I can't decide! _EDIT - Decision is NO, governance absolutely should not read from public application tables, because they come from **unsigned** requests_


doc/audit/index.rst
```diff
@@ -24,15 +24,21 @@ Audit
...
+    ^^^^^^^^^^^^^^^^^^^^^
 
     Utility library that can be used to write ledger audit scripts quickly and easily.
 
     ---
 
+    :fa:`shield-alt` :doc:`read_write_restrictions`
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+    Governance maps should only be written to when applying approved proposals.
```
achamayou, line 36: Except for the proposals and proposals_info maps, which are written to when proposals are submitted or withdrawn, or ballots submitted.

achamayou: Oh and also history and cose_history, which are written to in the same circumstances.

doc/audit/read_write_restrictions.rst
```diff
@@ -0,0 +1,78 @@
...
+
+To allow governance audit we ensure that all governance actions can be viewed and understood from the ledger, and signed by members. To support this CCF distinguishes governance tables from application tables, and restricts which tables can be read from and written to in different execution contexts.
+
+Table Namespaces
+----------------
+
+There are 3 categories of table defined:
+
+- Governance tables. These begin with a ``public:ccf.gov`` prefix, indicating that they are public (their contents are not encrypted in the ledger), and in the ``ccf.gov`` namespace. These should only be modified by constitution-controlled governance actions.
+- Internal tables. These begin with a ``ccf.internal`` or ``public:ccf.internal`` prefix, indicating that they are in the ``ccf.internal`` namespace, and may be public or private. These can be considered a lower-level subset of governance tables, though some are also modified by node-triggered operations such as signatures, rather than governance actions.
```
achamayou, line 12: We can clarify that they are managed by CCF itself, and that direct user writes are completely unsupported.

doc/audit/read_write_restrictions.rst
```diff
@@ -0,0 +1,78 @@
...
+- Application tables. This includes any other table name. These are the tables that application endpoints should modify.
+
+.. warning::
+
+    Table names are case sensitive. So ``public:CCF.gov.users`` (upper-case "CCF") is an `application` table, not a governance table.
```
achamayou, line 17: Similarly, is it worth clarifying that `PUBLIC:table` is in fact private?

doc/audit/read_write_restrictions.rst
```diff
@@ -0,0 +1,78 @@
...
+
+There are multiple possible contexts in which developer-specified code may execute within a CCF node:
+
+- Pre-approval governance context. These pieces of code may read from the KV to make decisions, but may not make any modifications to it, as they represent pending, unapproved operations. Specifically, this context applies:
+
+    - when evaluating member ballots
+    - when evaluating the constitution's ``validate`` function
+    - when evaluating the constitution's ``resolve`` function
+
+- Post-approval governance context. These pieces of code run only for accepted proposals, after a quorum of members have voted to approve it, so are permitted to make modifications to the KV. Specifically, this context applies:
```
achamayou, line 34: Only to governance tables though, and only to public governance tables even, when operating directly from JS.

doc/audit/read_write_restrictions.rst
```diff
@@ -0,0 +1,78 @@
...
+
+    - when evaluating the constitution's ``apply`` function
+
+- Application context. These pieces of code execute endpoint handlers to implement the application logic, are invoked to handle incoming user requests, and may read and write from application tables.
```
achamayou, line 38: They can also read from all public tables, including governance tables.

doc/audit/read_write_restrictions.rst
```diff
@@ -0,0 +1,78 @@
...
+.. note::
+
+    Since the constitution (and JS application) are written in JavaScript, there is also technically a global module evaluation context when the code is first ingested. It is expected that this only defines and exports the necessary functions. It has no access to the KV, and will result in errors if it attempts to access any KV table.
+
+Restricted Permissions
+----------------------
+
+CCF ensures that governance audit is possible offline from a ledger, by considering only a subset of transactions, and every governance change is traceable to member signatures. These governance transactions operate purely over governance tables, so that governance audit does not need to consider application tables, and purely over public tables so that all decisions can be reconstructed from the ledger. Combining the definitions above, we impose several restrictions on KV access in different contexts:
+
+- Governance code must never read from private tables. Doing so might make decisions which could not be reproduced from the ledger.
```
achamayou, line 49:
```suggestion
- Governance code must never read from private tables. Doing so might make decisions which could not be reproduced from the ledger by an auditor (ie. without access to ledger secrets).
```

src/js/wrap.cpp
```diff
@@ -402,51 +384,170 @@ namespace ccf::js
...
+          {
+            return MapAccessPermissions::ILLEGAL;
+          }
+
+          case TxAccess::APP:
+          {
+            switch (namespace_of_table)
+            {
+              // Private tables in the internal and governance namespaces should
+              // not exist. But if they do, they must be read-only for
```
achamayou, line 424: Why should they be read-only and not illegal? We may legitimately have internal private tables (for example for storing inbound requests), and we definitely would not want them readable by application code!

src/js/test/js.cpp
```diff
@@ -7,58 +7,148 @@
...
+      REQUIRE(
+        _check_kv_map_access(TxAccess::APP, public_internal_table_name) ==
+        MapAccessPermissions::READ_ONLY);
+    }
+
+    {
+      INFO("Private tables in internal namespace are read-only");
+      REQUIRE(
+        _check_kv_map_access(TxAccess::APP, private_internal_table_name) ==
+        MapAccessPermissions::READ_ONLY);
```
achamayou, line 32: That surely should be illegal?

doc/audit/index.rst
eddyashton: This summary is brief and imprecise to fit on a single tile, but I've reworded slightly to be less wrong.

doc/audit/read_write_restrictions.rst
eddyashton: Again trying to write a brief summary of the concept here, and clarifying the restrictions below.

src/js/test/js.cpp
eddyashton: Yes - I've now made this change so accessing private governance or private internal tables from application code is `ILLEGAL`.

src/js/wrap.cpp
eddyashton: As above, I've now changed this behaviour.

I think that's simplified the table a chunk too, so I've also refactored this code to more closely follow from the table in the docs - hopefully the 1-to-1 correspondence between the 2 is easier to track now.

Issue #4675, opened 2022-12-01
CI workflow to build a CCF container ready for running tests for a specific revision


DomAyre:
It would be helpful to perform a single build of CCF for a given revision (typically for a PR) and then build that into an image that can be used by other CI jobs.

This is especially useful for testing where we dynamically deploy executors for each test (SNP)

Todo:

- [x] Github action to push container (identical to current CI container) to ACR 
https://github.com/microsoft/CCF/pull/4692
- [x] ~Github action~ Scheduled task in ACR to cleanup containers ~(runs manually)~
- [x] Use dynamically built image in CCF SNP pipeline
https://github.com/microsoft/CCF/pull/4698
- [x] Modify container to have a build of CCF
https://github.com/microsoft/CCF/pull/4702
- [x] Speed up building with more powerful executor
https://github.com/microsoft/CCF/pull/4722
    - [x] Speed up image building
    https://github.com/microsoft/CCF/pull/4737

achamayou:
https://learn.microsoft.com/en-us/azure/container-registry/container-registry-auto-purge

DomAyre:
Yeah I found this and it's set up to purge any images older than 1d

achamayou:
@DomAyre if there isn't already, a quick mention + pointer in the internal doc would be great.

PR #4692, opened 2022-12-06
Create GitHub action to build CCF image for CI testing

DomAyre:
Contributes to https://github.com/microsoft/CCF/issues/4675

See issue for follow up tasks, ~this change will need someone to manually clean up any pushed images from the new workflow~ there is a scheduled task in the container registry to delete all tags that match the regex "pr-*" and are older than 1d


.github/workflows/build-ci-container.yml
```diff
@@ -0,0 +1,25 @@
+name: "Build CI Testing CCF Container"
```
achamayou, line 1:
```suggestion
name: "Build SNP CI Testing CCF Container"
```

achamayou (approved):

PR #4693, opened 2022-12-06
Allow sgx and virtual builds to be installed side-by-side using Ansible.

plietar:
The Ansible playbook for SGX tries to install the open-enclave package, whereas on Virtual builds it installs hostverify. Since these two are mutually exclusive, they cannot be run on the same machine.

This changes the playbook to install hostverify only if OE is not already installed. This allows the virtual build to be installed even after the SGX one has been.

We also uninstall hostverify when trying to install OE, as this allows the SGX build after the virtual one.

Unfortunately, it is still vital to install the SGX build first, then the virtual one. When running the playbooks in the reverse order, removing hostverify also removes ccf_virtual.

### DIFF
--- a/getting_started/setup_vm/app-dev.yml
+++ b/getting_started/setup_vm/app-dev.yml
@@ -14,14 +14,21 @@
     - import_role:
         name: az_dcap
         tasks_from: install.yml
+
+    # If OE is already installed, we don't want to install hostverify as they are mutually
+    # exclusive. Non-SGX CCF builds can use either of them.
+    - name: Gather the package facts
+      ansible.builtin.package_facts:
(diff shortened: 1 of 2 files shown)


plietar:
I can't say I'm really happy about this, especially the part where one has to carefully install the SGX build first, then the virtual, but I think it is still a step in the right direction.

I'm open to suggestions about it. A possible option would be for `app-dev.yml` to accept a list of platforms, which would work unambiguously.

achamayou (approved):
> I can't say I'm really happy about this, especially the part where one has to carefully install the SGX build first, then the virtual, but I think it is still a step in the right direction.

Agreed

PR #4694, opened 2022-12-06
Initial attestation container implementation

takuro-sato:
## Changes
- Basic attestation container implementation (with unit test and e2e test)
- CI (New 'Attestation Container' pipeline in azure devops. In this PR page it's failing because it can't access to new variables like `ATTESTATION_CONTAINER_CI_APP_ID` yet, but if it's ran from Azure DevOps page it passes. The latest successful run: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=59538&view=results )

## TODOs for future PRs
- Fetch and return collateral
- Push image to MCR when a release is cut
- https://github.com/microsoft/CCF/issues/4804
- https://github.com/microsoft/CCF/issues/4844


attestation-container/go.mod
```diff
@@ -0,0 +1,16 @@
+module microsoft/attestation-container
```
takuro-sato, line 1: If we want to publish the module, we can change it to `github.com/microsoft/CCF/attestation-container`.
https://go.dev/doc/tutorial/create-module#start

But I think it's fine to do it after creating attestation container repository (It would be `github.com/microsoft/attestation-container` ).

scripts/azure_deployment/arm_aci.py
```diff
@@ -88,73 +89,130 @@ def make_aci_deployment(parser: ArgumentParser) -> Deployment:
         type=lambda comma_sep_str: comma_sep_str.split(","),
     )
 
+    parser.add_argument(
+        "--aci-deploy-external-executer",
+        help="Deploy external executer container if true. Default=False",
+        default=False,
+        type=bool,
+    )
```
takuro-sato, line 97: With the current code, `arm_template.pyt deploy` deploys attestation container when this flag is true because it's a sidecar of external executer.
Even it's true only attestation container is deployed for now.

scripts/azure_deployment/arm_aci.py
```diff
@@ -88,73 +89,130 @@ def make_aci_deployment(parser: ArgumentParser) -> Deployment:
...
-                    }
-                    for i in range(args.count)
-                ],
-            },
+                        }
+                    ],
+                    "ipAddress": {
+                        "ports": [
+                            {"protocol": "TCP", "port": 22},
+                            {"protocol": "TCP", "port": 50051},
```
takuro-sato, line 206: Opening attestation-container gRPC's port is for running e2e test from CI agent. But considering that we should not expose attestation-container in the actual application, I will close this port and run e2e test inside the attestation-container using ssh + port 22 (Port 22 shouldn't be exposed in the actual application as well, but I think it's more obvious and hopefully it prevents application developers from exposing attestation-container for outside of the container group by mistake).

takuro-sato: Added comment in 7e9d165.

attestation-container/attest/attestation.go
```diff
@@ -0,0 +1,150 @@
...
+
+The padding is based on Section 3.1.2 of System V ABI for AMD64
+https://www.uclibc.org/docs/psABI-x86_64.pdf
+*/
+func createPayloadBytes(reportReqPtr uintptr, ReportRespPtr uintptr) ([PAYLOAD_SIZE]byte, error) {
+	payload := [PAYLOAD_SIZE]byte{}
+	var buf bytes.Buffer
+	// req_msg_type
+	if err := binary.Write(&buf, binary.LittleEndian, uint8(MSG_REPORT_REQ)); err != nil {
+		return payload, err
```
eddyashton, line 68: Why do we return `payload, err` in these branches, rather than `nil, err`? I don't know which is more idiomatic Go, but other instances in this PR use `nil, err`, and it feels like that will avoids bugs where callers check the wrong field.

attestation-container/README.md
```diff
@@ -0,0 +1,48 @@
...
+go test
+```
+
+## Update protobuf
+
+When you edit `.proto` file, you also need to update `.pb.go` files by:
+
+```bash
+protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative protobuf/attestation-container.proto
+```
```
eddyashton, line 48: This is a comment for myself (and anyone else who was going to suggest this change), rather than a suggestion:

I was going to suggest that this should be auto-generated during the build, rather than built by hand and checked in. But that's deliberately not well supported in Go! It's not easy to make `go build` run arbitrary other commands, and they really _want_ all of the generated source to be checked in, _so that it can be retrieved by dependent modules calling `go get`_!

attestation-container/attestation-container.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+type server struct {
+	pb.UnimplementedAttestationContainerServer
+}
+
+func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRequest) (*pb.FetchAttestationReply, error) {
+	reportData := [attest.REPORT_DATA_SIZE]byte{}
+	if len(in.GetReportData()) > attest.REPORT_DATA_SIZE {
+		return nil, fmt.Errorf("`report_data` needs to be smaller 64 bytes. size: %d bytes", len(in.GetReportData()))
```
eddyashton, line 29: Can we use `REPORT_DATA_SIZE` in this string, rather than a hard-coded 64?

```suggestion
		return nil, fmt.Errorf("`report_data` needs to be smaller than %d bytes. size: %d bytes", attest.REPORT_DATA_SIZE, len(in.GetReportData()))
```

.azure-pipelines-templates/deploy_attestation_container.yml
```diff
@@ -0,0 +1,103 @@
+jobs:
```
jumaffre, line 1: This reuses some parts of `deploy_aci.yml`. Would it be possible to refactor this to have a single consistent way to deploy ACI containers?

scripts/azure_deployment/arm_aci.py
```diff
@@ -88,73 +89,132 @@ def make_aci_deployment(parser: ArgumentParser) -> Deployment:
         type=lambda comma_sep_str: comma_sep_str.split(","),
     )
 
+    parser.add_argument(
+        "--aci-deploy-external-executer",
+        help="Deploy external executer container if true. Default=False",
```
jumaffre, line 94:
```suggestion
        "--aci-deploy-external-executor",
        help="Deploy external executor container if true. Default=False",
```

scripts/azure_deployment/arm_aci.py
```diff
@@ -88,73 +89,132 @@ def make_aci_deployment(parser: ArgumentParser) -> Deployment:
...
+                    },
+                },
+            }
+            for i in range(args.count)
+        ],
+    }
+
+    if args.aci_deploy_external_executer:
+        # NOTE: It currently exposes the attestation-container's port (50051) to outside of the container group for e2e testing purpose,
+        # but it shouldn't be done in actual CCF application.
```
jumaffre, line 171: Should the CLI argument be renamed accordingly, to indicate that it's for testing purposes only, e.g. `end-to-end-tests` or something?

attestation-container/attest/attestation.go
```diff
@@ -0,0 +1,150 @@
...
+const SNP_DEVICE_PATH = "/dev/sev"
+
+/*
+Creates and returns MSG_REPORT_REQ message bytes (SEV-SNP Firmware ABI Specification Table 20)
+*/
+func createReportReqBytes(reportData [REPORT_DATA_SIZE]byte) [REPORT_REQ_SIZE]byte {
+	reportReqBytes := [REPORT_REQ_SIZE]byte{}
+	for i := 0; i < REPORT_DATA_SIZE; i++ {
+		reportReqBytes[i] = reportData[i]
+	}
```
jumaffre, line 41: Is it possible to use `copy()` here instead?

attestation-container/attest/attestation.go
takuro-sato: I didn't consider it well. I agree to make it `nil, err`

.azure-pipelines-attestation-container.yml
```diff
@@ -10,14 +10,18 @@ pr:
 trigger:
```
jumaffre, line 10: I cannot comment on the lines above directly, but it'd be nice to automatically run this pipeline when any files in the `attestation-container` folder has been modified.

attestation-container/README.md
takuro-sato: >  It's not easy to make go build run arbitrary other commands

Yeah, looks like making `go` command simple is very prioritized by the author.
https://go.dev/doc/articles/go_command

scripts/azure_deployment/arm_aci.py
takuro-sato: This should be resolved soon in https://github.com/microsoft/CCF/issues/4804, but it sounds good since it's currently used only for 2e2 testig

attestation-container/attest/attestation.go
takuro-sato: Ah it's impossible to return `nil`, since its return type is array not slice.

attestation-container/attest/attestation.go
takuro-sato: Addressed in 263280d

.azure-pipelines-attestation-container.yml
takuro-sato: Fixed in e8e98d8

attestation-container/attest/attestation.go
```diff
@@ -0,0 +1,150 @@
+package attest
```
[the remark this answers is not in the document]
takuro-sato: Added in 5745f77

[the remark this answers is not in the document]
takuro-sato: I found it's not working in CI. I am fixing it.

.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -37,18 +41,47 @@ jobs:
...
+
       - script: |
           set -ex
-          python3.8 scripts/azure_deployment/arm_template.py deploy aci --subscription-id $(CCF_AZURE_SUBSCRIPTION_ID) --resource-group ccf-aci --aci-type dynamic-agent --deployment-name ci-$(Build.BuildNumber) --aci-image ccfmsrc.azurecr.io/ccf/ci:pr-$(wait_for_image.gitSha) > ~/aci_ips
-          echo "##vso[task.setvariable variable=ipAddresses;isOutput=true]`cat ~/aci_ips`"
+          if [ ${{ parameters.deploy_attestation_container }} == "True" ]; then
+            resource_group="attestationContainer"
+          else
+            resource_group="ccf-aci"
+          fi
```
eddyashton, line 64: Can't we specify this at the calling point parameters? So rather than `deploy_attestation_container: False` in `azure_pipelines_snp`, could set a `resource_group: ccf-aci` parameter there?

.azure-pipelines-templates/deploy_attestation_container.yml
jumaffre: Could we have some minimal reusable YAML building blocks (e.g. install the azure CLI) that we can call from both pipelines, possibly with different arguments so that we can easily compose (and read!) pipelines that deploy and/or test ACI containers. We follow this pattern already in some part of the CI, e.g. [here](https://github.com/microsoft/CCF/blob/main/.azure-pipelines-templates/common.yml#L51-L70). This can of course be done in a separate refactoring PR!

jumaffre (approved):

attestation-container/attestation-container.go
```diff
@@ -0,0 +1,61 @@
...
+	copy(reportData[:], in.GetReportData())
+	reportBytes, err := attest.FetchAttestationReportByte(reportData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch attestation report: %s", err)
+	}
+	return &pb.FetchAttestationReply{Attestation: reportBytes}, nil
+}
+
+func main() {
+	fmt.Println("Attestation container started.")
```
MahatiC, line 40: Perhaps change this to `log.Println` as that seems to include timestamp as well?

attestation-container/go.mod
achamayou: We mainly intend to release this as a container, not as a module at this time. But it's not to have the option for later.

attestation-container/go.mod
```diff
@@ -0,0 +1,16 @@
...
+
+go 1.19
+
+require (
+	golang.org/x/sys v0.3.0
```
achamayou, line 6: Will this be auto-upgraded by dependendabot? If not, is there a way to make those ">="?

.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -60,13 +93,46 @@ jobs:
     displayName: "Cleanup ACI"
     pool:
       vmImage: ubuntu-20.04
-    dependsOn: ${{ parameters.used_by }}
+    dependsOn:
+      - deploy_aci
```
DomAyre, line 97: There should be a transitive dependency with "used_by" so is this needed?

.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -60,13 +93,46 @@ jobs:
...
+              app_id: ${{ parameters.app_id }}
+              service_principal_password: ${{ parameters.service_principal_password }}
+              tenant: ${{ parameters.tenant }}
+
+          - script: |
+              set -ex
+              az acr repository delete --name attestationcontainerregistry --image attestation-container:$image_version --yes
+
+            name: cleanup_attestation_container_image
+            displayName: "Delete the container image"
```
DomAyre, line 138: If we delete the images after each run, if we run the pipeline periodically we might rebuild an image that could have just been re-used, we could have a cleanup task similar to what we do for the SNP pipeline?

.azure-pipelines-templates/deploy_aci.yml
takuro-sato: This is to access output variables of `deploy_aci`  by `$[ dependencies.deploy_aci.outputs['deploy_aci.resource_group'] ]` for example

.azure-pipelines-templates/deploy_aci.yml
takuro-sato: It should effective. I'll check what it does

.azure-pipelines-templates/deploy_aci.yml
takuro-sato: As discussed [here](https://github.com/microsoft/CCF/pull/4694#discussion_r1065593638), I'll revert the change to share `deploy_aci.yml` with `CCF SNP` and `attestation container` pipeline.

attestation-container/go.mod
takuro-sato: It supports golang.
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates

I'll add `gomod` to .github/dependabot.yml

.azure-pipelines-templates/deploy_attestation_container.yml
achamayou: Let's do this separately, after this PR.

.azure-pipelines-templates/deploy_aci.yml
takuro-sato: @DomAyre  I couldn't find the code to clean up docker images (not ACI containers) in the SNP pipeline. Can you tell me where it is please?

takuro-sato: Memo: It's done with [ACR task](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-tasks-overview) using azure cli (without any task definition file currently)

attestation-container/go.mod
takuro-sato: Added [663b656](https://github.com/microsoft/CCF/pull/4694/commits/663b65664a0e76fd34ccf7f96263606f8db36152)

.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -37,18 +41,47 @@ jobs:
...
+              cd attestation-container
+              docker build -t attestation-container .
+              az acr login --name attestationcontainerregistry
+              login_server=$(az acr show --name attestationcontainerregistry --query loginServer --output tsv)
+              image_version="ci-$(echo $(Build.BuildNumber) | sed 's/\.//g')"
+              docker tag attestation-container $login_server/attestation-container:$image_version
+              docker push $login_server/attestation-container:$image_version
+              echo "##vso[task.setvariable variable=ver;isoutput=true]$image_version"
+            name: build_and_push_attestation_container
+            displayName: "Build and Push Attestation Container"
```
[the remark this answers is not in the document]
takuro-sato: issue https://github.com/microsoft/CCF/issues/4844

.azure-pipelines-templates/deploy_aci.yml
takuro-sato: I'll work on it in https://github.com/microsoft/CCF/issues/4844

takuro-sato:
Latest manual run for the Attestation Container pipeline (successful). https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=60617&view=results

PR #4702, opened 2022-12-07
Update dockerfile to include CCF build and source

DomAyre:
Contributes to https://github.com/microsoft/CCF/issues/4675

Now that we can dynamically build an image and deploy it in ACI for our testing, we want to also include a copy of the CCF source from the current PR and build, so that any CI job can then immediately run tests against that build


DomAyre:
I'm not sure the changes in this PR are being picked up? Even though the checkout step does mention the correct SHA

DomAyre:
This won't pass in CI, because using the `pull_request_target` trigger means the version of `build-ci-containers.yml` used is from main, which won't get the correct dockerfile and won't checkout with a correct fetch depth.

Tested locally with the command `docker build -f docker/ccf_ci_built . --build-arg="platform=snp"` and the image is successfully built

PR #4703, opened 2022-12-07
Enable a Scan Build job in Daily

achamayou:
This has effectively not run for a long time, as pointed out by @letmaik, since it was run after the build was already complete. A substantial amount of fixing will need to take place before this can merge.


jumaffre (approved):

eddyashton (approved):

scripts/scan-build.sh
```diff
@@ -9,5 +9,8 @@ export CCC_CXX="clang++-$CLANG_VERSION"
...
 SCAN="scan-build-$CLANG_VERSION --exclude 3rdparty --exclude test"
 
-$SCAN cmake -GNinja -DCMAKE_BUILD_TYPE=Debug ..
-$SCAN ninja
\ No newline at end of file
+# VERBOSE_LOGGING=ON is important, without it scan-build will report values as unused
+# everywhere we compile out the logging statements that would otherwise read them
+$SCAN cmake -GNinja -DCOMPILE_TARGET=virtual -DVERBOSE_LOGGING=ON -DCMAKE_BUILD_TYPE=Debug ..
+# Fails on the current build of clang, because of false positives in doctest, WIP
+$SCAN ninja || true
```
eddyashton, line 16: Adding a comment here so these lines are easier to find in future/for anyone who hadn't spotted it.

The `|| true` here mean the results of this job are ignored - the scan is still finding errors but the build is passing. Hence, "WIP".

PR #4704, opened 2022-12-08
Fix compilation error with undefined variables in SNP builds

jumaffre:
Spotted by @wintersteiger 

Including the `ccf/pal/attestation.h` file from an app used to result in a compile-time error because the definitions of the `attestation_report_data_size` and `attestation_measurement_size` variables was guarded by `#if !defined(INSIDE_ENCLAVE) || defined(VIRTUAL_ENCLAVE)`, which is false on SGX builds. 

This PR fixes the issue by always defining these variables (now prefixed with `snp_`).


wintersteiger (approved):
LGTM, thanks for the quick fix!

achamayou (approved):

DomAyre (approved):

jumaffre:
Failure is unrelated to changes in this PR so merging regardless of current failure on CCF SNP pipeline.

Issue #4190, opened 2022-09-02
Run existing tests on a dynamically deployed SNP capable ACI


DomAyre:
Using Azure container instances directly for our CI is simple but we lose a lot of the benefits of hosted pools that we use for the rest of our testing.

We could instead use hosted pools which themselves spin up container instances, this would not only be more flexible, but it would also allow us to potentially have more representative scenarios for performance testing and testing code updates.

We should make sure all daily/CI tests run on this setup in a stable way

DomAyre:
Based on the main PR, this issue took roughly 18 work days to complete

PR #4708, opened 2022-12-08
Add step in SNP pipeline to run CTest

DomAyre:
Contributes to https://github.com/microsoft/CCF/issues/4190

Separate test fixes which have been spun out into their own PR's

- #4803 
- #4805 
- #4808 

Other changes which I decided to keep in this PR because they didn't make sense outside of it:

- Support running tests on a files-share
  - Make workspace configurable by environment variable to facilitate changing it for the SNP testing run
  - Change some symlinking operations to copies as running on a files-share doesn't support symlinks
- Increase timeouts in places where slow IO was triggering IO errors that I could see in runs
  - `connections_test` `create_connections_until_exhaustion` from 1 second to 3
  - `get_all_entries` timeout in `e2e_logging_cft` test from 5 seconds to 10
  - Election timeout from 4 seconds to 10


DomAyre:
There are some genuine failures in our tests which I've fixed, and there appears to be tests that are always the ones which fail with IO errors so I've disabled them to see if we get to a stable but reduced set of tests

tests/e2e_logging.py
```diff
@@ -1630,23 +1634,36 @@ def test_committed_index(network, args):
 
     _, log_id = network.txs.get_log_id(txid)
 
-    r = network.txs.request(log_id, priv=True, url_suffix="committed")
+    remaining_retries = 10
```
achamayou, line 1646: That retry logic would be better places in network.txs.request(..., timeout=...) I think. Parsing the error message seems unnecessarily brittle too.

tests/e2e_logging.py
```diff
@@ -1630,23 +1634,36 @@ def test_committed_index(network, args):
...
     assert r.body.json() == {"msg": f"Private message at idx {log_id} [0]"}
 
     network.txs.delete(log_id, priv=True)
 
     r = network.txs.request(log_id, priv=True)
     assert r.status_code == http.HTTPStatus.BAD_REQUEST.value, r.status_code
-    assert r.body.json() == {
-        "error": {"code": "ResourceNotFound", "message": f"No such record: {log_id}."}
-    }
+    assert r.body.json()["error"]["message"].startswith(f"No such record: {log_id}")
```
achamayou, line 1675: Why are we weakening the assertion here?

tests/e2e_logging.py
DomAyre: That makes sense, and I agree parsing the error message is a bad approach I'm trying to get something working that I can then clean up

tests/e2e_logging.py
DomAyre: We only need to because we have the TxID in the message so once I do that in a better way this can also be fixed

tests/e2e_logging.py
DomAyre: Change now lives in #4803

tests/e2e_logging.py
DomAyre: Change now lives in #4803

DomAyre:
I have concerns about the stability of the pipeline

After getting a [fully green run](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=59795), and making a very [innocuous change](https://github.com/microsoft/CCF/pull/4708/commits/fc3d98c27c21786c21758196f836ecb8ad49142f), I now see a [failing run](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=59812&view=logs&j=4bdd83b0-b422-5e84-0058-fb00a4da194e&t=01ab89b5-2bb2-5b36-1c41-59a63e14a55c&l=35636) with a bunch of failures relating to failures to finding/deleting files. This seems to be intermittent since I've only seen this failure once out of many many runs but still I'm unsure if we could merge this PR with this issue.

docker/ccf_ci_built
```diff
@@ -11,4 +11,4 @@ FROM ${base}-${platform}
 ARG platform=sgx
 RUN mkdir /CCF
 COPY . /CCF/
-RUN mkdir /CCF/build && cd /CCF/build && cmake -GNinja -DCOMPILE_TARGET=${platform} .. && ninja
+RUN mkdir /CCF/build && cd /CCF/build && cmake -GNinja -DCOMPILE_TARGET=${platform} .. && ninja && chmod -R 777 /CCF
```
jumaffre, line 14: `777` seems overly loose here?

tests/js-modules/modules.py
```diff
@@ -425,6 +425,7 @@ def test_dynamic_endpoints(network, args):
 
 
 @reqs.description("Test basic Node.js/npm app")
+@reqs.not_snp()
```
jumaffre, line 428: Please add reason why so that we can keep track of it.

docker/ccf_ci_built
DomAyre: Yeah I have a feeling this might not even be needed anymore

tests/js-modules/modules.py
DomAyre: Will do, I'll wait for the current CI run to complete before pushing since it's midway through

docker/ccf_ci_built
DomAyre: I was wrong https://github.com/microsoft/CCF/pull/4708/commits/085ed6a39292923109cf06964f94288475f2edb1

jumaffre (approved):

docker/ccf_ci_built
DomAyre: I tried several levels of permissions and they all seem to fail so unless you think it would be harmful to 777 it I think that's okay

DomAyre:
Looks stable after 5 manual runs alongside PR run
Screenshot 2023-01-11 at 09 58 39

PR #4709, opened 2022-12-08
Run building of CI containers on push to match CCF SNP

DomAyre:
Currently doesn't run on push, but CCF SNP does and relies on it


DomAyre:
One issue I'm not sure about is what sha will be used on the push case

achamayou (approved):

PR #4715, opened 2022-12-09
gRPC server streaming - support multiple concurrent HTTP/2 streams

eddyashton:
Working towards external executor dispatch, and making a few tweaks to the gRPC server streaming support to facilitate it. Trying to avoid too many mega-PRs, so here's a slightly more contained one.

This updates the sample async streaming test with a few features:
- Allows additional requests to come in over the same session (different stream) while still serving a stream. This is necessary so we can do `kv.Get`/`kv.Put`, in response to a still-alive `StartTx` stream. That's now supported in the sample pub/sub service, which writes `Ack`s to each event returned from `Sub`. The core fix here is removing the restriction to 1 concurrent stream, now we fallback to nghttp's defaults.
- Allow errors to be returned from streaming lambdas. There was type difficulty if we try to do this return-based (specifying that all streaming endpoints _return_ `Empty`, when they actually return errors or pending?), so I've removed the return-based behaviour for this adapter. The lambdas must instead directly call `set_grpc_response_trailers` to specify the response code.
- Confirm pub and sub actions are interleaved as expected, and we're not queuing all sub deliveries until the pubs complete.
- Insert sample "start" and "terminate" events in the subscription stream. The former gives us a client-side hook once the registration has started, removing waits/race conditions on startup. The latter gives us a way to terminate the loop from the client side - this might be removed in favour of "kill the thread/process/client session", but this is simpler with the current tests.


src/apps/external_executor/external_executor.cpp
```diff
@@ -707,37 +709,72 @@ namespace externalexecutor
...
+        {
+          LOG_INFO_FMT("Returning sub error");
+          ccf::grpc::set_grpc_response_trailers(
+            ctx.rpc_ctx,
+            ccf::grpc::make_grpc_status(
+              GRPC_STATUS_FAILED_PRECONDITION,
+              fmt::format(
+                "Already have a subscriber for {} - only support a single "
+                "subscriber per-event",
+                payload.name())));
```
jumaffre, line 726: I'm not keen on the fact that we're now leaking the HTTP/2 abstraction of gRPC (HTTP/2 trailers are used to pass gRPC status code/message to client) to the app. 

I understand that it was not possible to simply return `ccf::grpc::make_error();` from this handler, without having to specify `Empty` somewhere? Could the wrapper for streaming endpoint not return a different type, e.g. `GrpcAdapterNoResponse`?

eddyashton: This is now done - the streaming return type is `GrpcAdapterStreamingResponse`, which can no longer be any kind of success value so doesn't need to be templated. Still need a trailing return type on endpoints which return both errors and `make_pending` in different paths, but that's fine.

achamayou (approved):

src/apps/external_executor/external_executor.cpp
```diff
@@ -692,7 +691,12 @@ namespace externalexecutor
           out_stream->stream_msg(result);
         }
 
-        return ccf::grpc::make_success();
+        ctx.rpc_ctx->set_response_trailer(
+          ccf::grpc::make_status_trailer(GRPC_STATUS_OK));
+        ctx.rpc_ctx->set_response_trailer(
+          ccf::grpc::make_message_trailer(grpc_status_str(GRPC_STATUS_OK)));
+
+        return ccf::grpc::make_pending();
```
jumaffre, line 699: Why do we need this here as the endpoint does not detach the stream?

src/endpoints/grpc/stream.h
```diff
@@ -139,8 +139,18 @@ namespace ccf::grpc
...
+    StreamPtr<T>&& stream,
+    http::StreamOnCloseCallback close_cb = nullptr)
   {
+    auto rpc_ctx_impl = dynamic_cast<ccf::RpcContextImpl*>(ctx.get());
+    if (rpc_ctx_impl == nullptr)
+    {
+      throw std::logic_error("Unexpected type for RpcContext");
+    }
+
+    rpc_ctx_impl->response_is_pending = true;
```
jumaffre, line 152: Do we still need this if we keep `make_pending()`?

jumaffre (approved):

src/apps/external_executor/external_executor.cpp
eddyashton: We can't use `make_success`, as it needs a type, and we want to avoid templating this on a type. To be passed to `grpc_command_unary_stream_adapter`, we need to return something. We could distinguish "I've detached => pending" from "I've finished => send response now", but I don't think there's much point - the behaviour is the same (something has set the status trailers, send them when the stream closes).

If we think this instance is too weird, we could make it a pure `grpc_adapter` rather than a `unary_stream_adapter` - it's produced its entire body inline, so can return that as a vector (even if gRPC wants to call that a "stream"), so we don't need to include support for detaching?

src/endpoints/grpc/stream.h
eddyashton: Yes - `response_is_pending` is the actual indicator that the stream was detached and response shouldn't be sent immediately. `make_pending` is just the empty return type if we have no response status to set now.

PR #4717, opened 2022-12-09
Enable `TCP_NODELAY` for all TCP connections

jumaffre:
`TCP_NODELAY` was disabled (by default), i.e. [Nagle](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) was enabled, for all server incoming/outgoing TCP connections. This meant that clients could observe extra latency for their requests in certain cases (particularly visible with HTTP/2). 

This was already on enabled for performance tests (see below), but this turns it on on the server side to minimise latency for all clients.
https://github.com/microsoft/CCF/blob/c2614681855670189eb88f9ecbd33aadfae465d5/src/clients/perf/perf_client.h#L430

### DIFF
--- a/.azure-pipelines-templates/deploy_aci.yml
+++ b/.azure-pipelines-templates/deploy_aci.yml
@@ -15,7 +15,7 @@ jobs:
 
       - script: |
           set -ex
-          RETRIES=100
+          RETRIES=500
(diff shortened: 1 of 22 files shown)


eddyashton (approved):

achamayou:
@jumaffre a changelog update would be good.

PR #4718, opened 2022-12-09
Offline cose sign1 signing

### LINKED ISSUE #4696: Enable offline signing of COSE Sign1 governance, and update documentation accordingly
https://microsoft.github.io/CCF/main/governance/hsm_keys.html

Need:

- [ ] CLI tool with same interface as `ccf_cose_sign1`, minus private key, to produce pre-hashed TBS
- [ ] CLI tool with same interface as `ccf_cose_sign1`, minus private key, plus a passed-in signature, to produce COSE Sign1
- [ ] End to end test against AKV
- [ ] Documentation update

achamayou:
Closes #4696


eddyashton (approved):

python/ccf/cose.py
```diff
@@ -124,20 +127,78 @@ def create_cose_sign1(
     return msg.encode()
 
 
-DESCRIPTION = """Create and sign a COSE Sign1 message for CCF governance
+def create_cose_sign1_prepare(
+    payload: bytes,
+    cert_pem: Pem,
+    additional_headers: Optional[dict] = None,
```
letmaik, line 133: Each item is called a header parameter. The header is the bucket (dict) itself.

python/ccf/cose.py
```diff
@@ -124,20 +127,78 @@ def create_cose_sign1(
...
+    payload: bytes,
+    cert_pem: Pem,
+    signature: str,
+    additional_headers: Optional[dict] = None,
+) -> bytes:
+    cert = load_pem_x509_certificate(cert_pem.encode("ascii"), default_backend())
+    alg = default_algorithm_for_key(cert.public_key())
+    kid = cert_fingerprint(cert_pem)
+
+    headers = {pycose.headers.Algorithm: alg, pycose.headers.KID: kid}
```
letmaik, line 161: Re-encoding the protected header works here because it's using the exact same dict with the same ordering of fields between `prepare` and `finish` but at a minimum there should be a comment on that. It may be better to create a COSE message without signature, and then just append the signature.

python/ccf/cose.py
```diff
@@ -173,8 +228,34 @@ def _parser():
...
+        required=True,
+    )
+    return parser
+
+
+def _finish_parser():
+    parser = _common_parser(_FINISH_DESCRIPTION)
+    parser.add_argument(
+        "--signature",
+        help='Path to JSON file with a "value" field containing a JWS',
```
letmaik, line 246: Doesn't contain a JWS, but rather a raw base64 encoded signature. It's important to know whether this is assuming DER-style or IEEE for EC signatures.

python/ccf/cose.py
achamayou: I see, changing to `additional_protected_header`, for clarity.

python/ccf/cose.py
achamayou: I also considered that, but the UX benefit was unclear: in one case that's an extra file to save and re-provide, in the other case it's "call with the exact same arguments again", but pass the signature this time. Ultimately the latter was closer to the existing procedure, so I stuck with that.

python/ccf/cose.py
achamayou: Yes, good point. Unfortunately the AKV doc doesn't say very much (https://learn.microsoft.com/en-us/rest/api/keyvault/keys/sign/sign?source=recommendations&tabs=HTTP), but it does seem to be raw.

python/ccf/cose.py
achamayou: After discussion with @eddyashton I will add that as a feature in a later PR, but for now we will go with this.

PR #4730, opened 2022-12-13
Fix issues with snapshot generation for CHAMP

jumaffre:
Fix for the issue that recently cropped up in the CI: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=57709&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=bb1a7e6d-8f5b-56e4-638c-b498b20b4b62&l=19514

The underlying KV state data structure (CHAMP) keeps track of its total serialised size when inserting/removing keys so that a snapshot can be efficiently allocated when necessary. However, the size calculation was wrong in some niche cases depending on the `put()`/`remove()` patterns in a map. 

This is now fixed, with a much deeper coverage of this in the `map_test` unit test.


include/ccf/ds/logger.h
```diff
@@ -271,6 +271,7 @@ namespace logger
 
     static inline void default_init()
     {
+      get_loggers().clear();
```
jumaffre, line 274: Unrelated change: I noticed that we would get duplicate logs if we called `logger::config::default_init()` multiple times, which doesn't seem intended behaviour?

src/ds/champ_map.h
```diff
@@ -126,8 +126,10 @@ namespace champ
         const auto& entry = bin[i];
         if (k == entry->key)
         {
+          const auto diff = map::get_serialized_size_with_padding(entry->key) +
+            map::get_serialized_size_with_padding(entry->value);
           bin[i] = std::make_shared<Entry<K, V>>(k, v);
-          return map::get_size(k) + map::get_size(v);
+          return diff;
```
jumaffre, line 132: This is the actual fix. In the unlikely case that we have a collision (this is in `struct Collisions`), we used to return the serialised size of the new (k,v), rather than the old size if that key existed.

src/ds/champ_map.h
```diff
@@ -143,8 +145,8 @@ namespace champ
         const auto& entry = bin[i];
         if (k == entry->key)
         {
-          const auto diff =
-            map::get_size(entry->key) + map::get_size(entry->value);
+          const auto diff = map::get_serialized_size_with_padding(entry->key) +
+            map::get_serialized_size_with_padding(entry->value);
```
jumaffre, line 149: Slightly less buggy with `remove()`: we did not return the **padded** size.

Whether padding is still necessary in snapshots is left for future work (https://github.com/microsoft/CCF/issues/4731).

src/ds/champ_map.h
```diff
@@ -492,17 +497,14 @@ namespace champ
     {
       std::vector<KVTuple> ordered_state;
       ordered_state.reserve(map.size());
-      size_t size = 0;
+      size_t serialized_size = 0;
```
jumaffre, line 500: Some non-functional changes in this function.

src/ds/test/map_test.cpp
```diff
@@ -18,8 +22,9 @@ struct CollisionHash
   }
 };
 
-using K = uint64_t;
-using V = uint64_t;
+using K = kv::serialisers::SerialisedEntry;
+using V = kv::serialisers::SerialisedEntry;
+constexpr static size_t max_key_value_size = 128;
```
jumaffre, line 27: The `map_test` now uses the underlying type for `K` and `V`, as used by CCF, so that the unit tests are closer to the actual framework implementation and that we can exercise sizes of different keys/values more conveniently.

src/kv/test/kv_snapshot.cpp
```diff
@@ -181,7 +181,7 @@ TEST_CASE("Old snapshots" * doctest::test_suite("snapshot"))
       "YXoiAAAACAAAAAAAAAD+/////////"
       "wUAAAAAAAAAImZvbyIAAAANAAAAAAAAAAEAAAAAAAAAImJhciIAAAA=";
   }
-  else SUBCASE("True deletions")
+  SUBCASE("True deletions")
```
jumaffre, line 184: Spotted by @eddyashton, this subcase was never testes because of the `else`.

include/ccf/ds/logger.h
achamayou: That definitely sounds like a bug.

achamayou (approved):

eddyashton (approved):

PR #4737, opened 2022-12-13
Build per revision CI image from last CI container

DomAyre:
Contributes to #4675 

Speeds up image building by starting with the latest ci image (skips dependency installation)

Doesn't actually save too much time overall, we spend a similar amount of time pulling the image as we did building the things in that image, but it makes the code simpler and more maintainable and if the network is faster then it will be faster


.github/workflows/build-ci-container.yml
```diff
@@ -40,6 +40,9 @@ jobs:
       - name: Log in
         run: docker login -u $ACR_TOKEN_NAME -p ${{ secrets.ACR_CI_PUSH_TOKEN_PASSWORD }} $ACR_REGISTRY
 
+      - name: Pull release containers
```
achamayou, line 43:
```suggestion
      - name: Pull CI container
```

docker/ccf_ci_built
```diff
@@ -2,39 +2,13 @@
 # Contains CCF build dependencies and toolchain for target platform
 # Also contains CCF source and build directory
 
+# Latest image as of this change
 ARG platform=sgx
+ARG base=ccfmsrc.azurecr.io/ccf/ci:oe-0.18.4-dcap-ql-dev
```
achamayou, line 7: Can we pass this in as a build arg? That would make sure it stays in sync with the CI file above.

DomAyre: Sure
https://github.com/microsoft/CCF/pull/4737/commits/8be36f559d80e9961437f099ae9f1bb6e54d09c2

achamayou (approved):

PR #4742, opened 2022-12-14
Fix pypi publish

jeffa5:
The python package for ccf-4.0.0-dev2 didn't get published due to a failed run: 

https://github.com/microsoft/CCF/actions/runs/3683515441/jobs/6232180637

PR #4743, opened 2022-12-14
Fix PyPi pipeline

jumaffre:
Raised by https://github.com/microsoft/LSKV/pull/201#issuecomment-1351547412 

The PyPi release pipeline seems to be broken because the executors seems to now run a Ubuntu 22.04 image that no longer has Python 3.8.


jumaffre:
Duplicate of #4742 so closing this PR.

PR #4757, opened 2022-12-15
Make HTTP/2 stack compliant to standard

jumaffre:
Part of #3342 

We now that test that our HTTP/2 layer is compliant with the HTTP/2 spec, using `h2spec` (see #4753). Most of it is handled by nghttp2 directly, but there was two corner cases that we have to handle ourselves:
- Closing a session gracefully when a `GOAWAY` frame has been handled (detected by calls to `nghttp2_session_want_read()` and `nghttp2_session_want_write()`). 
- Closing a session when a an unexpected stream ID has been received. Note that this is awkward as nghttp2 does not handle this, so we do it ourselves are early as possible in the `on_begin_frame_recv_callback()` callback.

We now also run `h2spec` as part of the CI, and folded it in the existing `tlstest.py` (now renamed to `client_protocols.py`).

Note that it doesn't mean that all the HTTP/2 work is complete. For example, we could dispatch streams for the same session on different worker threads.


src/http/http2_parser.h
```diff
@@ -61,8 +68,7 @@ namespace http2
 
       // Submit initial settings
       std::vector<nghttp2_settings_entry> settings;
-      settings.push_back({NGHTTP2_SETTINGS_MAX_CONCURRENT_STREAMS, 1});
```
jumaffre, line 64: Removed the limitation on the maximum number of concurrent streams, as already done in #4715

src/http/http2_types.h
```diff
@@ -18,9 +18,11 @@
 namespace http2
 {
   using StreamId = int32_t;
+  constexpr static StreamId DEFAULT_STREAM_ID = 0;
+
   using StreamCloseCB = http::StreamOnCloseCallback;
 
-  constexpr static size_t max_data_read_size = 1 << 20;
+  constexpr static size_t max_frame_size = 1 << 14;
```
jumaffre, line 25: Taking this down to the default of 16kB as [one the test cases](https://github.com/summerwind/h2spec/blob/70ac2294010887f48b18e2d64f5cccd48421fad1/client/4_2_frame_size.go#L106) in `h2spec` does not seem to take the connection maximum frame size value into account and instead solely depends on the maximum allowed header length on this connection. This can be specified to the `h2spec` CLI and I'll address this next when I add support for maximum header sizes, etc. for HTTP/2.

achamayou (approved):

src/http/http2_parser.h
```diff
@@ -155,6 +170,18 @@ namespace http2
       }
 
       send_all_submitted();
+
+      // Detects whether server session expects any more data to read/write, and
+      // if no (e.g. goaway frame was handled), closes session gracefully
```
eddyashton, line 175:
```suggestion
      // if not (e.g. goaway frame was handled), closes session gracefully
```

eddyashton (approved):

PR #4764, opened 2022-12-16
Fix waiting for the wrong sha in SNP pipeline

DomAyre:


achamayou (approved):

DomAyre:
> @DomAyre did that work for branches, but not for forks, or something like that?

Yeah I think it's something like that

PR #4767, opened 2022-12-16
Make sandbox container nodes usable for other projects

jeffa5:
This updates the infra code to use the installed docker_wrap script as well as copy exe files so that they are reachable inside of the container, rather than having them be linked to an unmounted directory.


tests/infra/remote.py
```diff
@@ -449,7 +450,7 @@ def _setup_files(self):
         for path in self.exe_files:
             dst_path = os.path.normpath(os.path.join(self.root, os.path.basename(path)))
             src_path = os.path.normpath(os.path.join(os.getcwd(), path))
-            assert self._rc("ln -s {} {}".format(src_path, dst_path)) == 0
+            assert self._rc("cp {} {}".format(src_path, dst_path)) == 0
```
achamayou, line 453: I'm probably missing something obvious, but why is the copy necessary when running locally and not with SSH?

achamayou: I am guessing this is because when the remote is wrapped in the docker shim, the links point to something that's outside what's available in the container? Could there be a copy=False flag (or an instance attribute) here, and the DockerShim would overload _setup_file and call super()._setup_file(copy=True) or something like that?

That would fix the docker shim without making local runs pay for extra file copies (which in the case of rotation or recovery tests add up, since we start a fair few nodes).

Also, thank you very much for your contribution! :)

achamayou (approved):

tests/infra/remote.py
jeffa5: It's needed locally instead of the symbolic link as the source path with things like `cchost` might not get mounted into the docker container, only the `sandbox_0`-like directory gets mounted.

For SSH all the files are copied (or rather `put`) onto the remote machine so there aren't any links around I think.

jeffa5: Missed the second comment. Yes, I'll add a flag in for that so we don't always copy.

achamayou: @jeffa5 fantastic, thank you!

jeffa5: Add a `use_links` kwarg

jumaffre (approved):

PR #4771, opened 2022-12-16
Perf tool: support binary data

letmaik:
This changes the Parquet data type for the requests and responses to binary (previously UTF-8 string) to support a wider range of APIs.

I had to switch away from `parquet::StreamWriter` as that doesn't support binary data yet. See also https://github.com/apache/arrow/issues/14998.

### DIFF
--- a/tests/perf-system/generator/generator.py
+++ b/tests/perf-system/generator/generator.py
@@ -30,30 +30,36 @@ class Messages:
         append it to self.df and return the new df
         """
         batch_df = pd.DataFrame(columns=["messageID", "request"])
-        data_headers = "\r\n"
+        data_headers = b"\r\n"
         if len(additional_headers) > 0:
             additional_headers += "\r\n"
         if len(data) > 0:
-            data_headers = "content-length: " + str(len(data)) + "\r\n\r\n" + data
+            if isinstance(data, str):
+                data = data.encode("ascii")
+            data_headers = (
+                "content-length: " + str(len(data)) + "\r\n\r\n"
+            ).encode("ascii") + data
 
         df_size = len(self.df.index)
 
         for ind in range(iterations):
             batch_df.loc[ind] = [
                 str(ind + df_size),
-                verb.upper()
-                + " "
-                + path
-                + " "
-                + request_type
-                + "\r\n"
-                + "host: "
-                + host
-                + "\r\n"
-                + additional_headers
-                + "content-type: "
-                + content_type.lower()
-                + "\r\n"
+                (
+                    verb.upper()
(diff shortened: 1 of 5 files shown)


tests/perf-system/generator/generator.py
```diff
@@ -30,30 +30,36 @@ def append(
...
         batch_df = pd.DataFrame(columns=["messageID", "request"])
-        data_headers = "\r\n"
+        data_headers = b"\r\n"
         if len(additional_headers) > 0:
             additional_headers += "\r\n"
         if len(data) > 0:
-            data_headers = "content-length: " + str(len(data)) + "\r\n\r\n" + data
+            if isinstance(data, str):
+                data = data.encode("ascii")
+            data_headers = ("content-length: " + str(len(data)) + "\r\n\r\n").encode(
```
achamayou, line 39:
```suggestion
            data_headers = (f"content-length: {len(data)}\r\n\r\n").encode(

```

tests/perf-system/submitter/handle_arguments.h
```diff
@@ -32,21 +32,21 @@ class ArgumentParser
...
-      ->required(false)
+      ->required(true)
       ->check(CLI::ExistingFile);
     app
       .add_option(
         "-k,--key",
         key,
         "Specify the path to the file containing the private key.")
-      ->required(false)
+      ->required(true)
```
achamayou, line 42: I don't think we want to make this change, there is a valid use case for sending without client cert, using JWT auth.

tests/perf-system/submitter/submit.cpp
```diff
@@ -237,7 +210,8 @@ int main(int argc, char** argv)
     for (size_t req = 0; req < requests_size; req++)
     {
       gettimeofday(&start[req], NULL);
-      connection->write(raw_reqs[req]);
+      auto request = data_handler.request[req];
+      connection->write({request.data(), request.size()});
```
achamayou, line 214: Not sure what write() takes here, but if it's a vector ref, this will result in a temporary vector being allocated, and we definitely do not want that.

achamayou: It's a span, so that's all good.

tests/perf-system/submitter/handle_arguments.h
letmaik: The code as it is now only supports cert based auth, see `create_connection`. Not passing those flags would throw an error. I can revert this of course, but I thought it would reflect the currently supported feature set more accurately.

achamayou (approved):

Issue #4777, opened 2022-12-20
Partitions test failures


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=58776&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=2cd8c971-f488-5a92-7f18-3405fd4d3e8a
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=58752&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=3aae4c9d-f0dc-5b97-1b0e-f86312395af9

Not always in the same test, but always a failure to achieve primary unanimity.

Bizarrely, the post mortem detailed output (/consensus) always indicates unanimity.

Possibilities:

- Our timeout is somehow just too short. One way to check that would be log the time convergence takes, perhaps per scenario, and look at the distribution.
- There's something wrong with /network that causes it to report incorrect, or delayed information.
- ???

achamayou:
Two more suggestions from @eddyashton:

> remove the log_capture in wait_for_primary, to get more detailed information
> Add min_view to the "Waiting up to... for all nodes" log line

eddyashton:
#4778.

achamayou:
```
$ cat 67 | grep "Primary unanimity" | cut -d "|" -f 3 | cut -d : -f 3 | cut -d ' ' -f 6 | sort -g
0.37s
0.40s
0.41s
0.41s
0.41s
0.41s
0.48s
0.48s
0.48s
0.48s
0.48s
0.53s
0.53s
0.53s
0.53s
0.53s
0.56s
0.60s
0.63s
0.64s
0.75s
0.76s
0.80s
0.89s
0.91s
0.99s
1.05s
1.11s
1.15s
1.15s
1.16s
1.16s
1.17s
1.22s
1.27s
1.27s
1.28s
1.51s
1.63s
1.64s
1.64s
1.74s
1.93s
1.94s
1.94s
1.99s
2.05s
2.05s
2.17s
2.22s
2.32s
2.45s
2.52s
2.52s
2.64s
2.67s
2.72s
2.75s
2.87s
2.89s
3.25s
3.97s
4.05s
4.23s
4.24s
4.29s
4.43s
4.55s
5.53s
5.53s
5.72s
5.77s
5.88s
5.97s
6.12s
6.16s
6.24s
6.29s
6.35s
6.59s
7.11s
7.70s
7.71s
8.09s
9.98s
[...]

PR #4778, opened 2022-12-20
Improve stability of partitions tests

achamayou:
Our partitions tests have been unstable recently, as highlighted in #4777.

This PR makes some changes to logging, to make errors easier to understand, and to timeouts, to decrease the chance they will take place.

### DIFF
--- a/.azure-pipelines-templates/test.yml
+++ b/.azure-pipelines-templates/test.yml
@@ -5,22 +5,22 @@ parameters:
   suffix: ""
 
 steps:
-  - script: |
-      set -ex
-      ASAN_SYMBOLIZERS=(/usr/bin/llvm-symbolizer-*)
-      export ASAN_SYMBOLIZER_PATH=$(realpath ${ASAN_SYMBOLIZERS[0]})
-      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
-    env:
-      METRICS_MONGO_CONNECTION: $(METRICS_MONGO_CONNECTION)
-      CTEST_TIMEOUT: "${{ parameters.ctest_timeout }}"
-    displayName: CTest
-    workingDirectory: build
+  # - script: |
+  #     set -ex
+  #     ASAN_SYMBOLIZERS=(/usr/bin/llvm-symbolizer-*)
+  #     export ASAN_SYMBOLIZER_PATH=$(realpath ${ASAN_SYMBOLIZERS[0]})
+  #     ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
+  #   env:
+  #     METRICS_MONGO_CONNECTION: $(METRICS_MONGO_CONNECTION)
+  #     CTEST_TIMEOUT: "${{ parameters.ctest_timeout }}"
+  #   displayName: CTest
+  #   workingDirectory: build
 
   # Only run privileged tests in container environment
   - ${{ if ne(parameters.suffix, 'Perf') }}:
       - script: |
           set -ex
-          sudo bash -c "source env/bin/activate && ctest -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -L partitions -C partitions"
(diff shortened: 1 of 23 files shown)


.azure-pipelines-templates/test.yml
```diff
@@ -5,22 +5,22 @@ parameters:
...
+  # - script: |
+  #     set -ex
+  #     ASAN_SYMBOLIZERS=(/usr/bin/llvm-symbolizer-*)
+  #     export ASAN_SYMBOLIZER_PATH=$(realpath ${ASAN_SYMBOLIZERS[0]})
+  #     ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }}
+  #   env:
+  #     METRICS_MONGO_CONNECTION: $(METRICS_MONGO_CONNECTION)
+  #     CTEST_TIMEOUT: "${{ parameters.ctest_timeout }}"
+  #   displayName: CTest
+  #   workingDirectory: build
```
eddyashton, line 17:
```suggestion
  - script: |
      set -ex
      ASAN_SYMBOLIZERS=(/usr/bin/llvm-symbolizer-*)
      export ASAN_SYMBOLIZER_PATH=$(realpath ${ASAN_SYMBOLIZERS[0]})
      ./tests.sh -VV --timeout ${{ parameters.ctest_timeout }} --no-compress-output -T Test ${{ parameters.ctest_filter }} -N
    env:
      METRICS_MONGO_CONNECTION: $(METRICS_MONGO_CONNECTION)
      CTEST_TIMEOUT: "${{ parameters.ctest_timeout }}"
    displayName: CTest
    workingDirectory: build
```

Looks like the partitions test relies on `./tests.sh` already running to create the venv.

achamayou: I've done slightly differently/worse, but yes.

tests/infra/network.py
```diff
@@ -36,7 +36,7 @@
 JOIN_TIMEOUT = 40
 
 # If it takes a node n seconds to call an election, how long should we wait for an election to succeed?
-DEFAULT_TIMEOUT_MULTIPLIER = 3
+DEFAULT_TIMEOUT_MULTIPLIER = 15
```
achamayou, line 39: Aligning everything on a single timeout, because:

1. It's easier to reason about
2. The existing diversity of values seem to be an accident of history
3. We don't pay for this most of the time, unless we're negatively testing the lack of convergence (there is one such use, and we lower this to 3), or a test fails.

tests/infra/network.py
```diff
@@ -1204,7 +1204,9 @@ def wait_for_new_primary_in(
         while time.time() < end_time:
             try:
                 logs = []
-                new_primary, new_term = self.find_primary(nodes=nodes, log_capture=logs)
+                new_primary, new_term = self.find_primary(
+                    nodes=nodes, log_capture=logs, timeout=1
```
achamayou, line 1208: This is the biggest change, it means we poll through the list of nodes, and build a more up to date view. In cases where one node consumed most of the timeout, we could have run out before the global timeout, despite being in a situation where other nodes were up to date.

tests/infra/network.py
```diff
@@ -1224,39 +1226,45 @@ def wait_for_primary_unanimity(
...
-        LOG.info(f"Waiting up to {timeout}s for all nodes to agree on the primary")
+        LOG.info(
+            f"Waiting up to {timeout}s for all nodes to agree on the primary in view >= {min_view}"
+        )
         start_time = time.time()
         end_time = start_time + timeout
 
         nodes = nodes or self.get_joined_nodes()
-        primaries = []
+        primaries = {n.node_id: None for n in nodes}
```
achamayou, line 1236: A dict of opinions is clearer on error than a set.

tests/infra/network.py
```diff
@@ -954,11 +954,11 @@ def find_primary(self, nodes=None, timeout=10, log_capture=None, **kwargs):
                         )
 
             if primary_id is not None:
+                flush_info(logs, log_capture, 0)
```
eddyashton, line 957: This still potentially flushes in 2 different places with confusingly different line numbers. Can we flush on line 961, after break but before maybe-raising, so we always log from the same line number?

eddyashton (approved):

tests/infra/network.py
achamayou: Ok, this is a little less verbose, but if it passes I think that's fine.

PR #4780, opened 2022-12-20
Multi-Container: Use detached streams to send user requests directly to executors, rather than queuing

eddyashton:
This changes the dispatch behaviour (and associated gRPC API!) for the external executor app.

Previously:
- User request arrives
- Decide which executor it goes to
- Add request to queue for that executor
- Executor polls for its next job
- Respond with first item in the queue

Now, thanks to detached server streaming:
- User request arrives
- Decide which executor it goes to
- Stream request immediately to that executor

I've renamed `StartTx` to `Activate`, and am thinking about rolling that into `Register` (what's the point of saying "I will process work soon" separately from "I want to process work now"? Would let us separate those bits of client code, but I don't know if that's useful!). I've also added `Activated` and `Terminated` events to this work stream for clarity.


tests/external_executor/external_executor.py
```diff
@@ -137,132 +117,21 @@ def test_executor_registration(network, args):
...
-                    # These are harmless and expected, and I haven't found a way to swallow them
-                    assert not should_pass
-                    # pylint: disable=no-member
-                    assert e.code() == grpc.StatusCode.UNAUTHENTICATED, e
-
-    return network
-
-
-@reqs.description("Test basic KV operations via external executor app")
-def test_kv(network, args):
```
eddyashton, line 154: Have deleted this test entirely. It was helpful for early prototyping, but I don't think its useful anymore. The required functionality is tested by fuller tests, and the shims necessary to test this (essentially creating, registering, and activating a fake Executor so we can query its auth'd API) were getting increasingly, and unnecessarily, complex.

src/apps/external_executor/external_executor.cpp
```diff
@@ -80,9 +84,24 @@ namespace externalexecutor
...
+        }
+      }
     };
 
-    std::unordered_map<std::string, ExecutorIdList> supported_uris;
+    // Temporary implementation: Store supported uris on Register, insert into
+    // dispatch container on Activate
+    std::unordered_map<ExecutorId, std::vector<std::string>> supported_uris;
+    std::unordered_map<std::string, ExecutorIdList>
+      supported_uris_for_active_executors;
```
eddyashton, line 104: This is why I think we should collapse `Register` and `Activate` into a single RPC, so we don't have to store the supported URIs independently from the dispatchable URIs.

src/apps/external_executor/executor_auth_policy.h
```diff
@@ -5,49 +5,76 @@
 #include "ccf/common_auth_policies.h"
 #include "executor_code_id.h"
 
-struct ExecutorIdentity : public ccf::AuthnIdentity
+namespace externalexecutor
```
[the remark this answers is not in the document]
MahatiC: I've been meaning to refactor this, thanks!

src/apps/external_executor/external_executor.cpp
```diff
@@ -207,47 +227,90 @@ namespace externalexecutor
...
-          active_requests.emplace_hint(it, executor_id, pending_request);
+          executors_list.erase(executor_id);
         }
 
-        return ccf::grpc::make_success(optional_request_description);
+        return ccf::grpc::make_success();
       };
       make_endpoint(
-        "/externalexecutor.protobuf.KV/StartTx",
+        "/externalexecutor.protobuf.KV/Deactivate",
```
MahatiC, line 310: I think this `EndTx` endpoint logic can also be combined with this?

src/apps/external_executor/external_executor.cpp
MahatiC: That sounds good to me

src/apps/external_executor/external_executor.cpp
eddyashton: I don't think so. `Deactivate` is once-per-executor (paired with a single call to `Activate`), whereas `EndTx` should be called multiple times by each executor (once-per-request, paired with each message from `Activate`).

If we supported async client streaming, then this would probably be an additional message in the client->server direction (as it is for the etcd watch API), but since that's not supported we implement it as a standalone unary call.

I may also remove this RPC entirely, and say that client-initiated closures just terminate their `Activate` RPC session. That changes the expected client flow if clients want to be able to do this (ie - our e2e tests), and makes it harder to distinguish deliberate/polite executor closures from crashing/failure executor closures, but that might not be a problem.

MahatiC: > Deactivate is once-per-executor (paired with a single call to Activate), whereas EndTx should be called multiple times by each executor (once-per-request

Ah yes, my bad!

> makes it harder to distinguish deliberate/polite executor closures from crashing/failure executor closures, but that might not be a problem.

Yeah, I believe this can be handled with an appropriate error response

src/apps/external_executor/protobuf/kv.proto
```diff
@@ -13,8 +13,10 @@ option optimize_for = LITE_RUNTIME;
 service KV
 {
   // Indicate to node that this executor is ready for a new request.
-  // Returns description of a client request, to be executed.
-  rpc StartTx(google.protobuf.Empty) returns (OptionalRequestDescription) {}
+  // Returns stream of descriptions of client requests, to be executed.
+  rpc Activate(google.protobuf.Empty) returns (stream Work) {}
```
jumaffre, line 17: Could we update [this sequence diagram](https://microsoft.github.io/CCF/main/architecture/request_flow.html#external-executor-flow) accordingly? (doesn't have to be part of this PR).

eddyashton: Yes. I'll update this diagram/associated docs once the flow has finalised, but I think its still in flux at the moment and this form is unlikely to stay for long.

tests/external_executor/executors/logging_app.py
```diff
@@ -98,9 +103,16 @@ def run_loop(self, terminate_event):
...
         LOG.info("Ended executor loop")
+
+    def terminate(self):
+        target_uri = self.ccf_node.get_public_rpc_address()
+        with grpc.secure_channel(
+            target=target_uri,
+            credentials=self.credentials,
+        ) as channel:
+            stub = Service.KVStub(channel)
+            stub.Deactivate(Empty())
```
jumaffre, line 118: What if this isn't called by the executor? Would the CCF node think that the executor is still alive and so try to dispatch work to it? Should there be a watchdog mechanism so that the executor is automatically deactivated if it hasn't responded for a while (perhaps we could make use of HTTP/2 ping frames for that)?

eddyashton: That's what the close callback does - if the `Activate` stream is closed by the client (without calling `Deactivate` first), then the executor is deregistered on the server side by the close callback, and won't be dispatched any more work.

src/apps/external_executor/external_executor.cpp
```diff
@@ -207,47 +227,98 @@ namespace externalexecutor
...
+          out_stream->stream_msg(work);
+
+          active_executors.emplace_hint(
+            it,
+            executor_id,
+            ExecutorInfo{
+              {},
+              ccf::grpc::detach_stream(
+                ctx.rpc_ctx, std::move(out_stream), [this, executor_id]() {
+                  auto search = active_executors.find(executor_id);
```
jumaffre, line 260: Should there be a lock to access `active_executors` here and elsewhere?

tests/external_executor/executors/logging_app.py
jumaffre: Ah, I'd missed that, thanks. Do we have any plans to add more fine-grained health checks/stats for the external executor, e.g. to cater for an external executor still being connected, but slow to respond, etc.?

src/apps/external_executor/external_executor.cpp
eddyashton: Somewhere yes, but I'm planning to replace this container structure shortly as well.

tests/external_executor/executors/logging_app.py
eddyashton: Maybe, if we can get them from the HTTP/2 layer (eg - ping frequency?). But I don't know how valuable they'd be - I think the main decision for a node to make is "this executor is _dead_", and kill the connection. Hopefully that bubbles up from missed pings, and if not we'll do it with an activity timer. There's no node behavioural change before that, it should still dispatch to this executor at max rate. So these stats would be purely for external diagnosis, and may be found more easily by an operator contacting the executor directly.

jumaffre (approved):

PR #4786, opened 2022-12-21
Check validity times in default auth policies

### LINKED ISSUE #4785: Check validity periods and expiry times in default auth policies

eddyashton:
Resolves #4785.

We avoid using the untrusted host time in most framework code, as the threat model permits it to be manipulated by a malicious host. This leads to surprising permissiveness in some of our crypto validation, for instance accepting JWTs and certs which are expired. Using the untrusted host time here isn't a significant change in a malicious host's power - it allows them to DoS a caller by advancing the node's perception of time such that the caller's authentication is expired, but they cannot target this at a single user (affects the whole node), and they can likely build a more-targeted DoS already (by selectively dropping traffic). It is however a breaking change to the behaviour of these auth policies, which may now return auth errors for test info that they previously accepted.

TODO:
- [ ] CHANGELOG

### DIFF
--- a/src/endpoints/authentication/cert_auth.cpp
+++ b/src/endpoints/authentication/cert_auth.cpp
@@ -7,9 +7,53 @@
 #include "ccf/service/tables/members.h"
 #include "ccf/service/tables/nodes.h"
 #include "ccf/service/tables/users.h"
+#include "ds/x509_time_fmt.h"
+#include "enclave/enclave_time.h"
 
 namespace ccf
 {
+  static inline bool is_cert_valid_now(
+    const std::vector<uint8_t>& der_cert, std::string& error_reason)
+  {
+    auto verifier = crypto::make_unique_verifier(der_cert);
+
+    const auto [valid_from_timestring, valid_to_timestring] =
+      verifier->validity_period();
+
+    using namespace std::chrono;
+
+    const auto valid_from_unix_time =
+      duration_cast<seconds>(
+        ds::time_point_from_string(valid_from_timestring).time_since_epoch())
+        .count();
+    const auto valid_to_unix_time =
+      duration_cast<seconds>(
+        ds::time_point_from_string(valid_to_timestring).time_since_epoch())
+        .count();
+
+    const auto time_now =
+      duration_cast<seconds>(ccf::get_enclave_time()).count();
+
+    if (time_now < valid_from_unix_time)
+    {
+      error_reason = fmt::format(
+        "Current time {} is before certificate's Not Before validity period {}",
+        time_now,
+        valid_from_unix_time);
+      return false;
+    }
+    else if (time_now > valid_to_unix_time)
+    {
+      error_reason = fmt::format(
+        "Current time {} is after certificate's Not After validity period {}",
+        time_now,
+        valid_from_unix_time);
+      return false;
+    }
+
+    return true;
+  }
+
   std::unique_ptr<AuthnIdentity> UserCertAuthnPolicy::authenticate(
     kv::ReadOnlyTx& tx,
     const std::shared_ptr<ccf::RpcContext>& ctx,
@@ -19,6 +63,11 @@ namespace ccf
     if (caller_cert.empty())
     {
       error_reason = "No caller user certificate";
+      return nullptr;
+    }
+
+    if (!is_cert_valid_now(caller_cert, error_reason))
+    {
       return nullptr;
     }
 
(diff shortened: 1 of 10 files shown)


src/endpoints/authentication/cert_auth.cpp
```diff
@@ -7,9 +7,53 @@
 #include "ccf/service/tables/members.h"
 #include "ccf/service/tables/nodes.h"
 #include "ccf/service/tables/users.h"
+#include "ds/x509_time_fmt.h"
+#include "enclave/enclave_time.h"
 
 namespace ccf
 {
+  static inline bool is_cert_valid_now(
```
eddyashton, line 15: NB: This is applied to `user_cert` and `member_cert` requests, but _NOT_ to `node_cert`. If a node ends up with an expired cert, I think we still need to be able to speak to it, for the few requests that are `node_cert`-authenticated (`/join`).

eddyashton:
Converting this to draft - looks like it introduces a major performance regression, I guess from the construction of a fresh verifier on every request. Have reduced the repetitions on the longer perf tests to confirm that.

I'll see if we can extract the validity periods more cheaply, directly from the serialised DER. If that doesn't work, I'll look at using the existing verifier cache (used for sig auth) for cert auth as well.

eddyashton:
Major regression confirmed:
image

Will fix this before merging.

eddyashton:
> Converting this to draft - looks like it introduces a major performance regression, I guess from the construction of a fresh verifier on every request. Have reduced the repetitions on the longer perf tests to confirm that.
> 
> I'll see if we can extract the validity periods more cheaply, directly from the serialised DER. If that doesn't work, I'll look at using the existing verifier cache (used for sig auth) for cert auth as well.

Added a cache, and looks like the performance issue is resolved. Will give this another few runs.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- The built-in authentication policies for JWTs and certs will now enforce expiry times, based on the current time received from the host. JWTs must contain "nbf" and "exp" claims, and if those are outside the current time then the request will get an authentication error (#4786).
```
achamayou, line 12: I suggest we spell out this requirement in the documentation (ideally a full list of JWT fields our auth policy requires).

src/enclave/enclave_time.h
```diff
@@ -7,21 +7,24 @@
...
+    if (enclavetime::host_time_us != nullptr)
     {
-      const auto current_time = host_time_us->load();
-      if (current_time > last_value.count())
+      const auto current_time = enclavetime::host_time_us->load();
+      if (current_time > enclavetime::last_value.count())
       {
-        last_value = std::chrono::microseconds(current_time);
+        enclavetime::last_value = std::chrono::microseconds(current_time);
       }
```
achamayou, line 25: The other branch seems like a loggable event?

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -16,32 +17,56 @@ namespace ccf
...
-        return identity;
+        // Check that the Not Before and Expiration Time claims are valid
+        const size_t time_now =
+          std::chrono::duration_cast<std::chrono::seconds>(
+            ccf::get_enclave_time())
+            .count();
+        if (time_now < token.payload_typed.nbf)
+        {
+          error_reason = fmt::format(
+            "Current time {} is before token's Not Before claim {}",
```
achamayou, line 51:
```suggestion
            "Current time {} is before token's Not Before (nbf) claim {}",
```

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -16,32 +17,56 @@ namespace ccf
...
+            time_now,
+            token.payload_typed.nbf);
+        }
+        else if (time_now > token.payload_typed.exp)
+        {
+          error_reason = fmt::format(
+            "Current time {} is after token's Expiration Time claim {}",
```
achamayou, line 58:
```suggestion
            "Current time {} is after token's Expiration Time (exp) claim {}",
```

achamayou (approved):

eddyashton:
Perf impact is still larger than expected, especially the 5% hit on JWT auth. Returning to Draft status while this is investigated. Perhaps the construction of `payload_typed` is more expensive than expected, and unnecessary?

eddyashton:
A final thought to explore tomorrow: Is the slowdown simply because the JWT has increased in size?

eddyashton:
> A final thought to explore tomorrow: Is the slowdown simply because the JWT has increased in size?

Yup, thanks to @achamayou for confirming this in #4821 - just increasing the size of the JWT to include time claims, with no change on the C++ side, causes a ~6% slowdown in the `ls_jwt_sgx_cft` benchmark. I'm surprised this small size increase (~410 bytes to ~450 bytes in 1 header) has such an impact, but I think we at least understand this now well enough to merge this PR.

PR #4796, opened 2023-01-04
Speed up piccolo test runs

eddyashton:
A bunch of cleanup to the piccolo driver and submitter, and 2 big perf fixes buried in them - I'll highlight those inline.

### DIFF
--- a/tests/perf-system/analyzer/analyzer.py
+++ b/tests/perf-system/analyzer/analyzer.py
@@ -316,7 +316,7 @@ class Analyze:
 
 
 def get_df_from_parquet_file(input_file: str):
-    return pd.read_parquet(input_file, engine="fastparquet")
+    return pd.read_parquet(input_file)
 
 
 def default_analysis(send_file, response_file):
--- a/tests/perf-system/generator/generator.py
+++ b/tests/perf-system/generator/generator.py
@@ -16,7 +16,6 @@ class Messages:
 
     def append(
         self,
-        host,
         path,
         verb,
         http_version="HTTP/1.1",
@@ -30,9 +29,7 @@ class Messages:
         """
         headers = {k.lower(): v for k, v in additional_headers.items()}
 
-        # Insert host, content-length, and content-type headers, if they're not already present
-        if "host" not in headers:
-            headers["host"] = host
+        # Insert content-length, and content-type headers, if they're not already present
         if "content-length" not in headers:
             headers["content-length"] = str(len(body))
         if "content-type" not in headers and content_type is not None:
--- a/tests/perf-system/generator/generator.py
+++ b/tests/perf-system/generator/generator.py
@@ -19,52 +19,46 @@ class Messages:
         host,
         path,
         verb,
-        request_type="HTTP/1.1",
+        http_version="HTTP/1.1",
         content_type="application/json",
-        additional_headers="",
-        data="",
-        iterations=1,
+        additional_headers={},
+        body=bytes(),
     ):
         """
-        Create a new df with the contents specified by the arguments,
-        append it to self.df and return the new df
+        Serialise HTTP request specified by the arguments, and
+        append it to self.df
         """
-        batch_df = pd.DataFrame(columns=["messageID", "request"])
-        data_headers = b"\r\n"
-        if len(additional_headers) > 0:
-            additional_headers += "\r\n"
-        if len(data) > 0:
-            if isinstance(data, str):
-                data = data.encode("ascii")
-            data_headers = (f"content-length: {len(data)}\r\n\r\n").encode(
-                "ascii"
-            ) + data
+        headers = {k.lower(): v for k, v in additional_headers.items()}
+
+        # Insert host, content-length, and content-type headers, if they're not already present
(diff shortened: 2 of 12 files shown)


tests/perf-system/generator/generator.py
```diff
@@ -12,59 +12,46 @@
...
+
+        request_line = f"{verb.upper()} {path} {http_version}"
+        headers_string = "\r\n".join(f"{k}: {v}" for k, v in headers.items())
+        serialised_request = (
+            f"{request_line}\r\n{headers_string}\r\n\r\n".encode("ascii") + body
+        )
+
+        self.requests.append(
+            {"messageID": str(len(self.requests)), "request": serialised_request}
+        )
```
eddyashton, line 53: Appending single rows to a pandas DataFrame is slow, and afaict there isn't a fast way to do it. We don't actually need a DataFrame here (its just what `fastparquet` wants to write the file from), so we instead use a Python list and only convert to DataFrame when writing the file.

I've also removed the iterations option and the `batch_df` intermediate, because they were unused. We could also refactor the header construction here - we're doing a lot of work to format an identical prefix to every request (request line and common headers), which could be done once. But this already takes the generation time from several minutes to under a second, so I think that additional refactor is unnecessary.

tests/perf-system/submitter/submit.cpp
```diff
@@ -41,43 +41,95 @@ void read_parquet_file(string generator_filepath, ParquetData& data_handler)
...
   {
     LOG_INFO_FMT("Opened generator file");
   }
 
-  std::shared_ptr<::arrow::ChunkedArray> column;
+  const auto& schema = table->schema();
+
+  std::vector<std::string> column_names = {"messageID", "request"};
 
-  ::arrow::Status column1Status = arrow_reader->ReadColumn(1, &column);
```
eddyashton, line 67: We were previously parsing the file multiple times, first with `ReadTable` and then with `ReadColumn`. `arrow` is smart so that's still pretty fast, but to be safe I've moved to doing everything from the single table read. There's a bit more error checking/logging for failure cases, and some more resilience to schema changes/nearlycorrect files (looking up columns by name rather than assuming a known index).

tests/perf-system/analyzer/analyzer.py
```diff
@@ -316,7 +316,7 @@ def plot_latency_distribution(self, ms_separator: float, highest_vals=15):
 
 
 def get_df_from_parquet_file(input_file: str):
-    return pd.read_parquet(input_file, engine="fastparquet")
+    return pd.read_parquet(input_file)
```
eddyashton, line 319: For analysis, stopped using `fastparquet` for reading and instead using the default engine (`pyarrow`). Rough tests suggests that's 3x faster for the large files we're using. Still ~5s per file, which seems unreasonably slow, but I haven't found a faster path.

jumaffre (approved):

MahatiC (approved):

tests/perf-system/generator/logging_generator.py
```diff
@@ -0,0 +1,33 @@
...
+msgs = Messages()
+
+msgs.append("/app/log/private/count", "GET")
+
+for i in range(14):
+    msgs.append(
+        "/app/log/private",
+        "POST",
+        additional_headers=common_headers,
+        body='{"id": ' + str(i) + ', "msg": "Logged ' + str(i) + ' to private table"}',
```
achamayou, line 16:
```suggestion
        body=f'{"id": {i}, "msg": "Logged {i} to private table"}',

```

tests/perf-system/generator/logging_generator.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+msgs.append("/app/log/private/count", "GET", additional_headers=common_headers)
+
+for i in range(14):
```
achamayou, line 20: maybe define 14 as ID_RANGE once at the top?

tests/perf-system/submitter/submit.cpp
```diff
@@ -41,43 +41,95 @@ void read_parquet_file(string generator_filepath, ParquetData& data_handler)
...
-  std::shared_ptr<arrow::Table> table;
-  st = arrow_reader->ReadTable(selected_columns, &table);
-  if (!st.ok())
+  std::shared_ptr<arrow::Table> table = nullptr;
+  st = arrow_reader->ReadTable(&table);
+  if (!st.ok() || table == nullptr)
   {
-    LOG_FAIL_FMT("Couldn't open generator file");
-    exit(2);
+    LOG_FAIL_FMT("Couldn't open generator file: {}", st.ToString());
```
achamayou, line 56:
```suggestion
    LOG_FAIL_FMT("Couldn't open generator file ({}): {}", generator_filepath, st.ToString());

```

tests/perf-system/generator/logging_generator.py
eddyashton: This is a move of `loggin_generator.py`, but I guess includes enough other changes that its confused git diff. I'll do these cleanups while we're here.

achamayou (approved):

PR #4803, opened 2023-01-06
Handle race between index construction and e2e test

DomAyre:
This will only happen with slow IO, such as during SNP tests on a fileshare

Spun out of #4708 and therefore also contributes to #4190


samples/apps/logging/logging.cpp
```diff
@@ -498,29 +503,43 @@ namespace loggingapp
...
-        return ccf::make_error(
-          HTTP_STATUS_BAD_REQUEST,
-          ccf::errors::ResourceNotFound,
-          fmt::format("No such record: {}.", id));
+        auto response = nlohmann::json{{
+          "error",
+          {
+            {"code", ccf::errors::ResourceNotFound},
+            {"message", fmt::format("No such record: {}.", id)},
+            {"current_txid", committed_records->get_current_txid().to_str()},
```
achamayou, line 532: That doesn't seem thread safe, it could be called while the indexer is updating current_txid. The strategy should lock around access.

achamayou: See https://github.com/microsoft/CCF/pull/4075

tests/e2e_logging.py
```diff
@@ -1630,23 +1630,34 @@ def test_committed_index(network, args):
...
     assert r.body.json() == {"msg": f"Private message at idx {log_id} [0]"}
 
     network.txs.delete(log_id, priv=True)
 
     r = network.txs.request(log_id, priv=True)
     assert r.status_code == http.HTTPStatus.BAD_REQUEST.value, r.status_code
-    assert r.body.json() == {
-        "error": {"code": "ResourceNotFound", "message": f"No such record: {log_id}."}
-    }
+    assert r.body.json()["error"]["message"] == f"No such record: {log_id}."
```
achamayou, line 1656: Why remove the error code check?

tests/e2e_logging.py
```diff
@@ -1630,23 +1630,34 @@ def test_committed_index(network, args):
 
     _, log_id = network.txs.get_log_id(txid)
 
-    r = network.txs.request(log_id, priv=True, url_suffix="committed")
+    remaining_retries = 10
+    while remaining_retries > 0:
```
achamayou, line 1634: All our test retries are time based, rather than count based, I would prefer if we stayed consistent.

samples/apps/logging/logging.cpp
DomAyre: https://github.com/microsoft/CCF/pull/4803/commits/e86dcfd56e6f4820536d833545d745b22c159835

tests/e2e_logging.py
DomAyre: Just an oversight, we can't do a whole object match anymore since it will have the `current_txid` which we won't know the value of so I had to change it to specific field matching but left out code

Re-added in https://github.com/microsoft/CCF/pull/4803/commits/5702b60ef8681f246f1b1fd407ce78eb5dc4a84a

achamayou (approved):

Issue #4806, opened 2023-01-06
(Raft) Node without configuration can become primary/leader


lemmy:
This most basic raft_scenario demonstrates that a node lacking a configuration can become primary/leader.

root@codespaces-b88d0b:/workspaces/CCF/tests/raft_scenarios# python3 ../raft_scenarios_runner.py ../../build/raft_driver 4806
### steps

```
## create_new_node should create a node in pending state,
## A node with no connection with any other nodes
## and not part of any configuration.
create_new_node,0

periodic_all,500
dispatch_all
state_all

## The node should *not* be a primary, but a
## follower.
assert_is_candidate,0

## For good measure, let more time pass and
## see if the nodes remains a candidate.
periodic_all,1500
dispatch_all
state_all

assert_is_candidate,0

```

### stderr

```
terminating with uncaught exception of type std::runtime_error: Node not in expected state candidate on line 12

```

### diagram

```mermaid
sequenceDiagram
  Note over n[0]: Node n[0] created
  n[0]->>n[0]: periodic for 500 ms
  n[0]->>n[0]: [KV] initialising in term 1
  Note right of n[0]: P @1.0 (committed 0)
  Note over n[0]: Node is not in expected state: candidate
```

???+ error 

   | Scenario | stderr                                                                                                                |

   | -------- | --------------------------------------------------------------------------------------------------------------------- |

   | 4XXX     | terminating with uncaught exception of type std::runtime_error: Node not in expected state candidate on line 12 <br>  |

------

It appears that the root cause of the issue is that the logic in `raft::add_vote_for_me` allows a node to become the primary/leader even if it lacks a configuration, resulting in the for loops becoming no-ops. This change addresses the issue above, but causes a regression with `election_while_reconfiguration`, which only creates nodes without configurations (`create_new_nodes` vs `create_new_node`).  I believe `election_while_reconfiguration` has to be corrected.

```diff
diff --git a/src/consensus/aft/raft.h b/src/consensus/aft/raft.h
index 24a4fd344..b2490b14d 100644
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -2056,6 +2056,11 @@ namespace aft
 
     void add_vote_for_me(const ccf::NodeId& from)
     {
+      if (configurations.empty())
+      {
+        return;
+      }
+
       // Add vote for from node in each configuration where it is present
       for (auto const& conf : configurations)
       {
```

Independently, the raft_scenario API should only allow creating

[...]

PR #4807, opened 2023-01-06
(Raft) Node without configuration can become primary/leader #4806

lemmy:
https://github.com/microsoft/CCF/issues/4806

### DIFF
--- a/tests/raft_scenarios/4806
+++ b/tests/raft_scenarios/4806
@@ -1,0 +1,24 @@
+## create_new_node should create a node in pending state,
+## A node with no connection with any other nodes
+## and not part of any configuration.
+create_new_node,0
+
+periodic_all,500
+dispatch_all
+state_all
+
+assert_is_active,0
+assert_isnot_primary,0
+assert_isnot_candidate,0
+assert_isnot_backup,0
+
+## For good measure, let more time pass and
+## see if the nodes remains in its state.
+periodic_all,1500
+dispatch_all
+state_all
+
+assert_is_active,0
+assert_isnot_primary,0
+assert_isnot_candidate,0
+assert_isnot_backup,0
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -344,6 +344,11 @@ namespace aft
       return membership_state == kv::MembershipState::Learner;
     }
 
+    bool is_active() const
+    {
+      return membership_state == kv::MembershipState::Active;
+    }
+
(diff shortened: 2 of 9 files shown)


heidihoward (commented):
The TLA+ changes look good (just documentation improvements).  I'll leave this PR for some on the engineering side to approve for the c++ changes

lemmy:
Commit after Heidi's review adds two info-level log statements.

lemmy:
Is there anything for me to do to unblock the review of this change?

MahatiC:
@lemmy The CI is failing as the test `raft_scenario_test` is failing -- https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=62749&view=logs&j=f3457e4b-102d-5a4f-f801-dc119e4853a8&t=ae98f324-492e-5969-593b-258ede9c5463

CI needs to be passing before this can be reviewed.

tests/raft_scenarios/4806
```diff
@@ -0,0 +1,20 @@
...
+## and not part of any configuration.
+create_new_node,0
+
+periodic_all,500
+dispatch_all
+state_all
+
+## The node should *not* be a primary, but a
+## follower.
+assert_is_candidate,0
```
achamayou, line 12: This is contradictory with the logic added on line 1886 that prevents a node from becoming candidate if there is no configuration.

lemmy: 4806 is a "negative test" that should fail with the given change in raft.h. If we all agree that the scenario in 4806 should not be possible and the code change is correct, I will remove 4806.

achamayou (approved):

tests/raft_scenarios/4806
lemmy: By the way, this test exposes a limitation of the existing assertions provided in driver.h; it would be beneficial if it also included assert_is_not_* or a more general assert_is_state that accepts arguments.

achamayou: Yes, we need to revisit that wrapper. It was trying to be generic and encapsulate state at a time when the prevailing thought was that there would be several consensus implementations that could easily be swapped in. This no longer the case, and we'd rather have a more precise way to talk about the existing implementation instead.

lemmy: I'd expect https://github.com/microsoft/CCF/issues/5057 to take over a significant part of scenario *validation*, and for the wrapper to be "reduced" to being the stimulus.

Issue #4156, opened 2022-08-19
Multi-container: Dispatch incoming requests efficiently to appropriate executor


eddyashton:
For initial implementation (with HTTP transport), we can simply iterate through all registered executors, and choose the first match. Additionally, we can queue all client requests and actually provide them on polls from each executor, if that's easier for now. With gRPC, we should be able to dispatch and queue as each request comes in.

Eventually I think we should use [r3](https://github.com/c9s/r3) for this, though I haven't worked out exactly how/when we compile the prefix trie.

eddyashton:
Still trying to work out how we do this efficiently. The problem is that each executor can be registered for a unique but overlapping set of paths, so we might have an initial registration of something like this:

```
{
  "A": [               "POST /foo/B", "POST /foo/C", ...],
  "B": ["POST /foo/A",                "POST /foo/C", ...],
  "C": ["POST /foo/A", "POST /foo/B",                ...],
  ...
}
```

(Note that if we use r3, then I think the slugs/templates in this don't make it any more complex. If we don't, then they are probably an additional complexity).

Then separately, we want some kind of priority for each URL for fair dispatch, so we can move an executor to the end of the queue once it has handled a request.

So when a `POST /foo/B` request, we might have an initial executor list of `["A", "C", ...]`, decide to dispatch to `"A"`, and update this list to `["C", ..., "A"]`.

The problem is, we'd _also_ like A to move to the back of the list for `POST /foo/C` requests, so that `["A", "B", ...]` bcomes `["B", ... "A"]`.

So we can obviously do better than a linear scan through either executors or URLs, with a table lookup (or r3 dispatch) - modified rarely, going from URLs to a collection of supporting executors. But then it looks like we have a linear pass through that collection of executors, with associated locking, to re-prioritize. I'm wondering if there's some kind of sparse multi-heap that can do this better.

eddyashton:
2 more thoughts:

- We could think of a single ordered list of all executors, so that we're only bumping an executor to the end of one list. But I can't see how to do dispatch in that model (ie - efficient lookup from URL to the collection of supporting executors), that doesn't devolve to maintaining some separate sorted collection which must also be updated.
- The case above is pessimal and unrealistic. In most cases we expect to have multiple executors implementing the same _set_ of endpoints. If we can canonicalise this set, we can map it to a single ordered collection of supporting executors, and the update cost is then linear in the number of these sets rather than the total number of endpoints.

eddyashton:
[...]

PR #4809, opened 2023-01-09
Dispatch external executors based on governance-defined, codeID-bound supported endpoints

eddyashton:
Implements the change discussed here: https://github.com/microsoft/CCF/issues/4156#issuecomment-1337582425

For full generality, we may eventually allow dispatch to be affected by both per-executor-code-ID and per-executor metadata, but for now we move towards purely per-executor-code-ID dispatch. This PR contains a bunch of adjacent but non-critical code shuffling.


eddyashton:
Converting this to draft for now. Because it introduces a KV read at the start of externally-executed endpoints (to read from the KV dispatch table), it occasionally triggers #4815 (which was previously avoided as our sample endpoints only read from a single table). So we need to fix that bug before we can merge this.

eddyashton:
Parking this as blocked by #4815 - will rebase and re-open once that's resolved.

Issue #4211, opened 2022-09-09
Framework support for gRPC


jumaffre:
Now that we have support for gRPC at the application level, the framework should also support gRPC to report consistent errors to clients. For example, currently, if a gRPC application defines an authentication policy throws an authorisation error, this error will be formatted by the framework and returned as JSON, which at best will be confusing to the gRPC client.

Instead, we should format the errors based on the content type header of the incoming request. The existing governance and operator endpoints will remain as JSON only. 

Tasks:
- [ ] 1. Separate gRPC headers from private implementation
- [ ] 2. From `status.proto`, link `status.pb.cc` to `libccf_endpoints`
- [x] 3. Format frontend errors as JSON if request `content-type: application/json` or framed protobuf if request `content-type: application/grpc` (facilitated by #4205): #4813

jumaffre:
This work will be necessary to support gRPC apps, e.g. https://github.com/microsoft/LSKV.

eddyashton:
Protobuf v22 includes a hard dependency on Abseil, so we don't want to include that as a dependency for all CCF apps. We think that's fine - the framework can speak HTTP/2 and produce gRPC error trailers, and protobuf parsing/encoding is only necessary in the application. We may still want to build an enclave-friendly, mitigated version of protobuf, and distribute it with CCF for easy app consumption, but we don't want to link it in every app.

achamayou:
There are no plans to add gRPC support in the foreseeable future.

PR #4813, opened 2023-01-10
Automatically return gRPC errors for framework-level errors

jumaffre:
Part of #4211 

We now automatically return gRPC errors to gRPC clients that trigger a framework-level errors (e.g. authentication failure). To do so, we automatically convert the `http_status` to `grpc_status` if the request `content-type` is `appplication/grpc`. 

Note that to get this to work is a smaller change than I first anticipated as we not need to link in `libprotobuf` as the error is simply sent back to the client in HTTP/2 trailers. However, I'm planning to link in `libprotobuf` to a new `libccf_grpc.a` library that can be linked to apps that create gRPC endpoints (future PR).


MahatiC (approved):
lgtm

eddyashton (approved):

Issue #4672, opened 2022-12-01
Multi-container: Define Indexing API for external executors


MahatiC:
Define API endpoints that would be exposed to external executors so that they are able to write indexing logic. This roughly includes:

In the existing design, application logic defines the strategy and installs it on CCF nodes. With business logic separation, Executors are stateless and would require a separate API in order to access the committed transactions (what we currently pass to `handle_committed_transaction`).

In addition to this, CCF node will be storing the data structures that maintain the index. Identify what these data structures look like. To start with, we could use pointers from how existing customers (ACL, Calastone) are using indexing strategies and identify what we need to maintain and expose to reach feature parity.

Tasks:

- [x] [API for indexed data -- supports map data structure with local node on-disk storage](https://github.com/microsoft/CCF/pull/4820)

MahatiC:
Remaining tasks are tracked [here](https://github.com/microsoft/CCF/issues/4912) and [here](https://github.com/microsoft/CCF/issues/4911).

PR #4820, opened 2023-01-11
Implement gRPC API to store indexed data

MahatiC:
~~WIP and not yet ready for review~~

This PR currently adds key-value data structure support for indexed data. It exposes API to stream transactions on a given table/map name. Facilitates storage and retrieval of indexed data. (I will update #4672 or create a new issue for the remaining tasks that may not be covered in this PR)

TODO:

- [x] Expose API for Streaming transactions, Store data, Fetch data
- [x] Add support to be able to extend and support other types of data structures
- [x] Add support for local disk storage of indexed data
- [x] Clean-up
- [x] Test API with in-memory storage


src/apps/external_executor/protobuf/index.proto
```diff
@@ -0,0 +1,74 @@
...
+message Published {}
+
+message IndexWork
+{
+  oneof task
+  {
+    Subscribed subscribed = 1;
+    IndexKeyValue key_value = 2;
+    Published work_done = 3;
+  }
```
MahatiC, line 47: `subscribed` and `work_done` are probably not needed and are not currently used. (Will revisit and remove)

src/apps/external_executor/protobuf/index.proto
```diff
@@ -0,0 +1,74 @@
...
+  rpc InstallAndSubscribe(IndexInstall) returns (stream IndexWork) {}
+
+  // Stores Indexer's data. Can only be invoked by the Indexer that installed
+  // the strategy
+  rpc StoreIndexedData(IndexKeyValue) returns (google.protobuf.Empty) {}
+
+  // Takes an input of Index Strategy name and a Key, returns values from the
+  // indexed data on the local CCF node This can be invoked by any registered
+  // executor
+  rpc GetIndexedData(IndexKey) returns (IndexKeyValue) {}
```
MahatiC, line 28: I think there should be a way to unsubscribe the index where stream connection can be gracefully terminated. I will add that endpoint, but thoughts are welcome.

src/apps/external_executor/protobuf/index.proto
jumaffre: Could we have a comment about this in the code, pointing to a GitHub issue, as well?

src/apps/external_executor/protobuf/index.proto
```diff
@@ -0,0 +1,74 @@
...
+}
+
+message IndexKeyValue
+{
+  bytes key = 2;
+  bytes value = 3;
```
eddyashton, line 34: Is there a reason these are `2` and `3` rather than `1` and `2`?

src/endpoints/grpc/grpc.h
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache 2.0 License.
 #pragma once
 
+#include "ccf/endpoint_context.h"
```
eddyashton, line 5: Doesn't look like this include is doing anything? Or is it required from somewhere else?

src/apps/external_executor/external_executor_indexing.cpp
```diff
@@ -0,0 +1,194 @@
...
+        data.mutable_key_value();
+      index_key_value->set_key(k);
+      index_key_value->set_value(v);
+      if (is_indexer_active)
+      {
+        // stream transactions to the indexer
+        if (!detached_stream->stream_msg(data))
+        {
+          LOG_DEBUG_FMT("Failed to stream request to indexer {}", indexer_id);
+        }
```
eddyashton, line 172: I think this should be `LOG_FAIL`, and we should early-out here - any failure probably means every future message will also fail, and even if it didn't we don't want to continue with missing entries!

src/apps/external_executor/protobuf/index.proto
```diff
@@ -0,0 +1,74 @@
...
+{
+  // Installs index and registers with the given strategy name and table name.
+  // Streams Key-Value pairs of every transaction in the given table name.
+  // Stores indexed data in an unordered map datastructure or a prefix tree
+  // depending on the type of enum passed
+  rpc InstallAndSubscribe(IndexInstall) returns (stream IndexWork) {}
+
+  // Stores Indexer's data. Can only be invoked by the Indexer that installed
+  // the strategy
+  rpc StoreIndexedData(IndexKeyValue) returns (google.protobuf.Empty) {}
```
eddyashton, line 23: The C++ implementation uses `IndexPayload` for the input type here, rather than `IndexKeyValue`. The former includes the `strategy_name`, so allows an executor to implement multiple strategies. Without that, I think we're unnecessarily restricting to a single index-per-executor, and not matching that in the implementation.

src/apps/external_executor/external_executor.cpp
```diff
@@ -129,6 +124,147 @@ namespace externalexecutor
...
+        auto executor_id = get_caller_executor_id(ctx);
+        auto strategy_ptr = map_index_strategies[strategy_name];
+        std::string key = payload.key();
+        std::optional<std::string> data = strategy_ptr->fetch(key);
+
+        if (!data.has_value())
+        {
+          return ccf::grpc::make_error(
+            GRPC_STATUS_NOT_FOUND, "Key was not found in the indexed data");
+        }
```
eddyashton, line 247: This path will be fairly common, and I'd lean towards implementing as an empty response (ie - `value` has no body => key wasn't found), rather than an error type. This will make it simpler to extend with different kinds of error in future (eg - distinguish "I've never been told this" from "I think I have, but have to fetch it asynchronously from disk"), handled inline without exception handling.

src/apps/external_executor/external_executor_indexing.cpp
```diff
@@ -0,0 +1,194 @@
...
+    lfs_access(lfs_access_),
+    indexed_data(lfs_access_, 10, map),
+    map_name(map)
+  {}
+
+  std::optional<std::string> MapIndex::fetch_data(std::string& key)
+  {
+    bool complete = true;
+
+    while (true)
```
eddyashton, line 20: What's this while loop looping over? I don't see any range operations here.

src/apps/external_executor/external_executor_indexing.cpp
```diff
@@ -0,0 +1,194 @@
...
+          complete = false;
+        }
+      }
+    }
+  }
+
+  void MapIndex::store_data(std::string& key, std::string& value)
+  {
+    indexed_data.insert(key, std::move(value));
+  }
```
eddyashton, line 109: These shouldn't be mutable refs `&`, they should either be const-ref (`const &`) to avoid copying in, or rvalue-ref (`&&`) if we're looking to move them into the LRU and can justify taking moves at the calling sites.

src/apps/external_executor/protobuf/index.proto
MahatiC: `IndexPayload` is the correct and intended type (and so is the cpp implementation). Will correct this in the proto file.

src/apps/external_executor/external_executor_indexing.cpp
MahatiC: Good point. Failing early is better here.

src/endpoints/grpc/grpc.h
MahatiC: Yes, this is needed otherwise the linker will error out.

src/apps/external_executor/protobuf/index.proto
MahatiC: Nope, an oversight. Will change.

src/apps/external_executor/protobuf/index.proto
MahatiC: These are now used in tests

src/apps/external_executor/external_executor.cpp
MahatiC: Agree, modified this

src/apps/external_executor/external_executor_indexing.cpp
MahatiC: Removed

tests/external_executor/executors/logging_app.py
```diff
@@ -152,7 +152,7 @@ def run_loop(self, activated_event):
                 if "log/private" in request.uri:
                     table = "private:records"
                 elif "log/public" in request.uri:
-                    table = "records"
+                    table = "public:records"
```
eddyashton, line 155: Good spot, I totally got this wrong! We should also change the private table name from "private:records" to "records", to match the other implementations of this app in JS and C++ (not critical, but eventually we'd like to prove that they do "the same thing", so might as well clean this up now).

eddyashton:
I'll review this again today, and I think its looking good to merge as-is, but we might revisit these design point in a future PR:

> * Any authenticated indexer can access `StoreIndexedData` and `GetIndexedData` rpcs based on a valid `strategy_name` (i.e. a strategy that has actually been installed via a call to `InstallAndSubscribe`) -- this allows flexibility with their logic when multiple indexers would want to access the indexed data

I think that's fine. Certainly for the `Get`, we want _any executor_ (even one which hasn't registered _itself_ as an indexer) to be able to retrieve the data. We might add some kind of namespacing access permissions on that in future, for the same multiple-app support that we have restrictions on table names, but we don't have a use-case/shape for that for now, so no point early engineering it.
For the `Store`, I agree we want the flexibility, but then the step of _registering_ an index data structure by-name seems a little weird. The current API combines registering to receive entries so you can construct an index, with creating a structure for storing results, so suggests a 1-to-1 correspondence. But we're actually more general than that. So maybe we'll need to split out data structure creation in future. We may also need to work out how we detect/handle multiple indexing-executors trying to write different data (or even different types of structure!) at the same indexed-data key.

> * If there's a client disconnection during streaming , we will stop streaming to the indexer (see `is_indexer_active` in code) and on reconnection, indexer has to install and subscribe with a new strategy name -- for example `IndexerA` initially installs `StrategyA`, then it gets disconnected.  On reconnection, it will have to install with a new strategy name `StrategyB` to get the stream of transactions, but it can still continue to access and store data in `StrategyA`.

This seems like a bad experience - we really want the other executors to know the names of the data structures they're accessing in-advance (they shouldn't need to know a custom key that the _current_ indexer is using), and we don't want every indexing-executor to need to re-implement this retry logic ("You're supposed to be building `StrategyA`, but if that already exists you must be a restart so try `StrategyA_1`"). Whether we decouple "create a data structure" from "register for indexed data" or not, I think we can just say that "if this data structure already exists (same name and type), that's not an error".

MahatiC:
> This seems like a bad experience - we really want the other executors to know the names of the data structures they're accessing in-advance (they shouldn't need to know a custom key that the _current_ indexer is using), and we don't want every indexing-executor to need to re-implement this retry logic ("You're supposed to be building `StrategyA`, but if that already exists you must be a restart so try `StrategyA_1`"). Whether we decouple "create a data structure" from "register for indexed data" or not, I think we can just say that "if this data structure already exists (same name and type), that's not an error".

To clarify, the indexer can still very much access `StrategyA` i.e. the indexed data is always accessible by every indexer with the name `StrategyA` regardless of the number of retries. However, on every stream reconnection they have to reconnect with a unique key, which in this case happens to be a strategy name (will discuss more offline and update here)

tests/external_executor/external_executor.py
```diff
@@ -421,6 +427,105 @@ def subscribe(event_name):
...
+            with grpc.secure_channel(
+                target=f"{primary.get_public_rpc_host()}:{primary.get_public_rpc_port()}",
+                credentials=sub_credentials,
+            ) as subscriber_channel:
+                in_stub = IndexService.IndexStub(subscriber_channel)
+                for work in in_stub.InstallAndSubscribe(
+                    Index.IndexInstall(
+                        strategy_name="TestStrategy",
+                        map_name="public:records",
+                        data_structure="MAP",
```
eddyashton, line 475: Shouldn't this be an enum value rather than a string?

src/apps/external_executor/external_executor.cpp
```diff
@@ -129,6 +124,170 @@ namespace externalexecutor
...
+        auto executor_id = get_caller_executor_id(ctx);
+
+        auto ds = payload.data_structure();
+        // Mark default ds as MAP
+        IndexDataStructure data_structure = MAP;
+
+        if (ds == externalexecutor::protobuf::IndexInstall::PREFIX_TREE)
+        {
+          data_structure = PREFIX_TREE;
+        }
```
eddyashton, line 159: If we reused the protobuf-generated enum type here, rather than redefining, then we could get a simpler default behaviour (ie - this code doesn't claim to set a default, _always_ gets it from the message. Then the message likely has some well-defined default).

src/apps/external_executor/external_executor.cpp
```diff
@@ -129,6 +124,170 @@ namespace externalexecutor
...
+            data_structure,
+            executor_id,
+            ctx,
+            node_context,
+            std::move(out_stream));
+
+        node_context.get_indexing_strategies().install_strategy(map_index);
+
+        // store the index pointer into the strategies
+        map_index_strategies[strategy] = map_index;
```
eddyashton, line 179: Comment for future, but not for this PR: if we separate registration from data structure construction, then we won't have a name to use as the key here. I think we can instead use the executor's cert as the key, to store this indexing strategy instance.

src/apps/external_executor/external_executor.cpp
```diff
@@ -129,6 +124,170 @@ namespace externalexecutor
...
+
+        std::shared_ptr<ExecutorIndex> map_index =
+          std::make_shared<ExecutorIndex>(
+            payload.map_name(),
+            strategy,
+            data_structure,
+            executor_id,
+            ctx,
+            node_context,
+            std::move(out_stream));
```
eddyashton, line 174: This should be `ccf::grpc::detach_stream(... std::move(out_stream) ...)`, and there we can add a destructor hook which will remove this index from the local collections if the stream dies. I'm not sure exactly what happens if we just `move` but don't `detach`...

tests/external_executor/executors/logging_app.py
MahatiC: Done

tests/external_executor/external_executor.py
```diff
@@ -421,6 +427,105 @@ def subscribe(event_name):
...
+        log_id = 14
+        with executor_thread(logging_executor):
+            with primary.client() as c:
+                for _ in range(3):
+                    r = c.post(
+                        "/app/log/public",
+                        {"id": log_id, "msg": "hello_world_" + str(log_id)},
+                    )
+                    assert r.status_code == 200
+                    log_id = log_id + 1
```
eddyashton, line 450: We could avoid repeating the `14`, `3`, and `"hello_world_{N}"` formatting below if we stored them here. Build some list of "kv writes", confirm during index construction that everything in "kv writes" is received in the stream (including overwriting a previously used `log_id`!), and use that to store locally a list of "indexed entries", and finally confirm that everything in "indexed entries" (with maybe a different key from the KV key, for clarity) can be retrieved from `GetIndexedData`.

eddyashton (approved):
LGTM! Few things to change in future PRs, but I think we can merge this now.

tests/external_executor/external_executor.py
MahatiC: My bad, changed it

MahatiC:
@eddyashton Thanks, I've made a note to address some of the comments in the issue https://github.com/microsoft/CCF/issues/4672

PR #4823, opened 2023-01-11
Add ccf gov msg created at

### LINKED ISSUE #4213: Zero user-cost protection against governance replay
We could make governance substantially stronger against replay attacks, without requiring member re-keying as we do now, when non-specific proposals have been created.

The scheme would involve having members add an additional timestamp in protected headers, and storing the last N (proposal id, timestamp). When a proposal comes in, its timestamp must be fresher than 50% of the stored proposals, and not collide with them. If it's older, or collides, it is assumed to be a replay and rejected.

Because we do not have ordered tables, this is somewhat inefficient to do in the general case. Because governance is relatively infrequent however, N=100 may be a good tradeoff between efficiency and protection.

achamayou:
Resolve #4213:

- [x] add ccf.gov.msg.created_at uint header
- [x] store mapping of `f"{ccf.gov.msg.created_at}:{request_digest}" -> proposal_id` in KV
- [x] Detect stale and colliding proposals
- [x] Distinguish collisions from staleness in error reporting
- [x] Solve stability issue in ledger reading testcase
- [x] Document replay protection mechanism and assumptions
- [x] Add end to end tests

Additional notes:

- I've made `ccf.gov.msg.created_at` mandatory on all governance messages, despite it being only really useful for proposals at this stage. It seems simpler that way, both to use and to implement, and could come in handy in the future.
- In the absence of a use case for a tweakable window, I've added the functionality and a test for it, but it's not exposed in the default constitution.


doc/audit/builtin_maps.rst
```diff
@@ -382,6 +382,15 @@ Governance history of the service, captures all COSE Sign 1 governance requests
 
 **Value** COSE Sign1
 
+``cose_recent_proposals``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Recent COSE proposal timeline, kept for the purpose of avoiding potential replay attacks.
+
+**Key** ccf.gov.msg.created_at field from COSE protect header, as a string, followed by SHA-256 digest of the COSE Sign1, represented as a hex-encoded string and separated by a ':'.
```
letmaik, line 390: The digest of the whole COSE message is not suitable for replay protection as the message is not canonical and can easily be modified without invalidating the signature. Simple example is adding an unprotected header parameter. Another example is using a non-optimal length encoding for one of the message elements (like the outer array length). 
I suggest using the signature or the hash of the signature instead.

src/endpoints/authentication/cose_auth.cpp
```diff
@@ -96,26 +99,40 @@ namespace ccf
...
         UsefulBuf_FromSZ(gov_msg_proposal_id);
       header_items[GOV_MSG_PROPOSAL_ID].uLabelType = QCBOR_TYPE_TEXT_STRING;
       header_items[GOV_MSG_PROPOSAL_ID].uDataType = QCBOR_TYPE_TEXT_STRING;
 
+      auto gov_msg_proposal_created_at = COSE_HEADER_PARAM_MSG_CREATED_AT;
+      header_items[GOV_MSG_CREATED_AT].label.string =
+        UsefulBuf_FromSZ(gov_msg_proposal_created_at);
+      header_items[GOV_MSG_CREATED_AT].uLabelType = QCBOR_TYPE_TEXT_STRING;
+      // Although this is really uint, specify QCBOR_TYPE_INT64
+      // QCBOR_TYPE_UINT64 only matches uint values that are greated that
```
letmaik, line 113:
```suggestion
      // QCBOR_TYPE_UINT64 only matches uint values that are greater than
```

doc/audit/builtin_maps.rst
achamayou: Thank you, that's a very good point.

achamayou: Switched to a digest of the signature.

letmaik (approved):

src/node/rpc/member_frontend.h
```diff
@@ -387,6 +387,25 @@ namespace ccf
         caller_id, {cose_sign1.begin(), cose_sign1.end()});
     }
 
+    bool is_proposal_replayed_or_stale(
```
eddyashton, line 390: IIUC, this is currently only `is_proposal_replayed`, and the _stale_ detection is still WIP? If so, we should rename this function for now to avoid confusion.

src/node/rpc/member_frontend.h
```diff
@@ -1231,6 +1248,18 @@ namespace ccf
...
+          if (is_proposal_replayed_or_stale(
+                ctx.tx,
+                cose_auth_id->protected_header.gov_msg_created_at,
+                request_digest,
+                proposal_id))
+          {
+            ctx.rpc_ctx->set_error(
+              HTTP_STATUS_BAD_REQUEST,
+              ccf::errors::DuplicateOrStaleProposal,
+              "Duplicate or stale proposal");
```
eddyashton, line 1260: I think we can distinguish these cases (duplicate proposal vs stale proposal), so we should do so in the error message? For a duplicate proposal, we can return the conflicting proposal ID (so if 2 idempotent-y tools try to submit the same proposal, they can both defer to talking about the state of the active instance). For a stale proposal, we can tell them how stale they are, and hint to freshen their `gov_msg_created_at` header.

src/service/tables/governance_history.h
```diff
@@ -18,4 +18,11 @@ namespace ccf
   {
     static constexpr auto COSE_GOV_HISTORY = "public:ccf.gov.cose_history";
   }
+  using COSERecentProposals =
+    ServiceMap<std::string, std::string /* ProposalId */>;
```
eddyashton, line 22: Assuming we don't care that this is a string (ie - don't manually mangle it or anything), we should use the existing typedef.

```suggestion
    ServiceMap<std::string, ccf::jsgov::ProposalId>;
```

Might need an additional include.

eddyashton (approved):

src/endpoints/authentication/cose_auth.cpp
```diff
@@ -96,26 +102,40 @@ namespace ccf
...
       header_items[GOV_MSG_PROPOSAL_ID].uDataType = QCBOR_TYPE_TEXT_STRING;
 
+      auto gov_msg_proposal_created_at = COSE_HEADER_PARAM_MSG_CREATED_AT;
+      header_items[GOV_MSG_CREATED_AT].label.string =
+        UsefulBuf_FromSZ(gov_msg_proposal_created_at);
+      header_items[GOV_MSG_CREATED_AT].uLabelType = QCBOR_TYPE_TEXT_STRING;
+      // Although this is really uint, specify QCBOR_TYPE_INT64
+      // QCBOR_TYPE_UINT64 only matches uint values that are greater than
+      // INT64_MAX
+      header_items[GOV_MSG_CREATED_AT].uDataType = QCBOR_TYPE_INT64;
```
eddyashton, line 118: Question about QCBOR, not this PR: This seems mad? We're telling the decoder what type to expect, but need to know at this point whether we expect the _value_ to be greater than `INT64_MAX`, to pass a different value? Not a problem in this instance, for a few aeons.

src/node/rpc/member_frontend.h
achamayou: Yes and yes, since the header isn't exported.

src/node/rpc/member_frontend.h
achamayou: We could do something like that, but:

1. We have a bar for stale, but it doesn't mean the proposal isn't a duplicate. It's just that we have stopped keeping track of proposals in that time range.
2. The bar for stale isn't fixed, because it's based on the median of the timestamps of the proposals submitted successfully so far. So we can return the value we used to make the decision, but there's no guarantee that if you re-sign later than it, the proposal will be accepted.
3. We can return the proposal ID, but it feels like a half-step. The current interface doesn't try to be idempotent, the more (most?) idempotent version would return success (and the proposal id) on duplicate, and only an error when the proposal is older than the stale bar. It's still not fully idempotent, because making the same call twice only works in a limited time window.

I think one limitation with going with 3. (or indeed returning the proposal ID in a failure), is that we will delete the proposal content from the KV much sooner (*) than it will slide out of the replay detection window. And so there is typically little chance that the user can query via an online API for that proposal ID if it has already concluded (passed or rejected, and since deleted).

(*) on the next proposal.

src/endpoints/authentication/cose_auth.cpp
achamayou: It surprised me too, but comments suggest it is by design:

https://github.com/laurencelundblade/QCBOR/blob/37286c08a00d7516cf2da173a48d1bfdd5260eff/inc/qcbor/qcbor_decode.h#L221
https://github.com/laurencelundblade/QCBOR/blob/4e808ba6452134c04247e956f2a81abef2442a6c/src/qcbor_decode.c#L765

If the value is greater, I expect the decoding will fail with a type mismatch.

src/service/tables/governance_history.h
achamayou: Turns out we have a ccf::ProposalId _and_ a ccf::jsgov::ProposalId, so fixing that too.

src/node/rpc/member_frontend.h
achamayou: As discussed, going with errors in both cases, and detailed error messages.

src/node/rpc/member_frontend.h
```diff
@@ -387,22 +394,65 @@ namespace ccf
...
+          return true;
+        });
+
+      std::sort(replay_keys.begin(), replay_keys.end());
+
+      // New proposal must be more recent than median proposal kept
+      if (!replay_keys.empty())
+      {
+        auto [min_created_at, _] =
+          nonstd::split_1(replay_keys[replay_keys.size() / 2], ":");
```
eddyashton, line 421: This `min_created_at` is shadowing the out-parameter, and looks like it unused at the call-site. It should be formatted into the `TooOld` error message.

src/node/rpc/member_frontend.h
```diff
@@ -1248,18 +1298,57 @@ namespace ccf
...
           record_cose_governance_history(
             ctx.tx, member_id.value(), cose_auth_id->envelope);
-          if (is_proposal_replayed_or_stale(
-                ctx.tx,
-                cose_auth_id->protected_header.gov_msg_created_at,
-                request_digest,
-                proposal_id))
+          ccf::ProposalId colliding_proposal_id = proposal_id;
+          std::string min_created_at = "";
+          // created_at is stored as a string, and compared alphanumerically
```
eddyashton, line 1303: "stored as a string" confused me - can you clarify this comment to state that we're storing a Unix timestamp (seconds-since-epoch) here, hence the 0-padding etc? I assumed string meant ISO-formatted, as we pass on the CLI.

eddyashton (approved):

src/node/rpc/member_frontend.h
achamayou: Clarified, sorry. I should point out that this choice is open to discussion. Picking ISO 8601 is possible, but we would need to be more specific: there isn't a canonical representation. An auditor using an implementation that chooses to omit the separators for example, assuming we don't, is going to get incorrect results.

Zero padded decimal ASCII has the merit of being canonical, and difficult to get wrong. But regrettably custom.

Another choice may be to stick to CBOR integers (ie. the bytes we receive directly). That's not canonical either, sadly, and would definitely not work with our current sorted KV plans.

src/node/rpc/member_frontend.h
achamayou: I meant to do that (hence the out parameter), but apparently forgot. Thank you!

include/ccf/service/tables/service.h
```diff
@@ -39,6 +39,8 @@ namespace ccf
     nlohmann::json service_data = nullptr;
     /// TxID at which current service was created
     std::optional<ccf::TxID> current_service_create_txid = std::nullopt;
+    /// Size of recent_cose_proposals window
+    std::optional<size_t> recent_cose_proposals_window_size = std::nullopt;
```
jumaffre, line 43: It seems that this would fit better in the [service configuration table](https://microsoft.github.io/CCF/main/audit/builtin_maps.html#service-config)?

jumaffre: By updating the `cchost_config.json` schema, it would also be settable by the operator at startup (see https://microsoft.github.io/CCF/main/operations/configuration.html#service-configuration).

src/node/rpc/member_frontend.h
```diff
@@ -387,6 +394,68 @@ namespace ccf
...
+      auto key = fmt::format("{}:{}", created_at, ds::to_hex(request_digest));
+
+      std::vector<std::string> replay_keys;
+      cose_recent_proposals->foreach_key(
+        [&replay_keys](const std::string& replay_key) {
+          replay_keys.push_back(replay_key);
+          return true;
+        });
+
+      std::sort(replay_keys.begin(), replay_keys.end());
```
jumaffre, line 415: We should write a comment somewhere (and/or raise a GitHub issue) to use sorted tables for this once we have support for them?

src/node/rpc/member_frontend.h
```diff
@@ -387,6 +394,68 @@ namespace ccf
...
+      }
+
+      if (cose_recent_proposals->has(key))
+      {
+        colliding_proposal_id = cose_recent_proposals->get(key).value();
+        return ProposalSubmissionStatus::DuplicateInWindow;
+      }
+      else
+      {
+        size_t window_size = 100;
```
jumaffre, line 436:
```suggestion
        size_t default_window_size = 100;
```

This value should probably live in `governance_history.h` too?

tests/governance_js.py
```diff
@@ -298,6 +313,77 @@ def test_pure_proposals(network, args):
...
+            and r.body.json()["error"]["code"] == "ProposalCreatedTooLongAgo"
+        ), r.body.text()
+
+        # Submitting a unique proposal just past the median of the window does work
+        c.set_created_at_override(now + (window_size // 2))
+        r = c.post("/gov/proposals", unique_always_accept_noop())
+        assert r.status_code == 200, r.body.text()
+
+        r = c.post("/gov/proposals", set_service_recent_cose_proposals_window_size(1))
+        assert r.status_code == 200, r.body.text()
```
jumaffre, line 369: Why does this proposal not have to go through the usual propose/vote flow?

jumaffre (approved):

src/node/rpc/member_frontend.h
eddyashton: I think the fixed-length padded int makes sense. If we'd be re-encoding to ISO 8601, we can choose a canonicalisation (and I don't think auditors needs to reproduce that?), but that seems like an unnecessary expense.

achamayou: We could choose a canonicalisation of ISO 8601, yes. The auditors would need to fetch the request from the cose gov history table, parse it, and reconstruct the timestamp in the canonicalised format to replay the replay check and make sure it's correct. This is assuming they want to verify the node did not abusively allow replayed/stale proposals.

include/ccf/service/tables/service.h
achamayou: Thank you for the suggestion, it definitely looks like a better fit, I'll move it there.

src/node/rpc/member_frontend.h
achamayou: It could do too, it's not going to much reuse, so I don't think it will help anything but it also won't hurt.

tests/governance_js.py
achamayou: Proximally, because of https://github.com/microsoft/CCF/pull/4823/files#diff-ba4676ee17cf56956ba8162d50f0046f14347b7a1be803618365a20b2f4b744c. There isn't meaningful voting in this case, and it's not what we're testing, so like much of the rest of the scenario, we skip it.

src/node/rpc/member_frontend.h
achamayou: @jumaffre moved to src/service/tables/config.h, which I think is a more suitable central location considering this is effectively a default on the table.

src/node/rpc/member_frontend.h
achamayou: @jumaffre appended as https://github.com/microsoft/CCF/issues/3088#issuecomment-1400708471

Issue #4816, opened 2023-01-10
Fix encoding of sign() JS API function 


takuro-sato:
We are aiming to make the JS crypto API to follow the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) in behavior.
To do so, https://github.com/microsoft/CCF/issues/2735 changed `ccf.crypto.verifySignature` to accept  IEEE P1363 as signature's encoding format from  DER when `ECDSA` is used.
But recently added `ccf.crypto.sign` doesn't follow it and creates DER encoded signature and we need to fix it.
Also we haven't checked encoding to use for `EdDSA`, so we need to check it from  [Web Crypto API spec](https://w3c.github.io/webcrypto/).

achamayou:
@takuro-sato we need to check for RSA too.

takuro-sato:
## TODOs
- [ ] Sign with P1363 encoding for ECDSA
- [ ] Check if the encoding for RSA and EdDSA is compatible with Web Crypto API (and fix if not)

PR #4829, opened 2023-01-12
Use p1363 encoding for sign() JS API with ECDSA

takuro-sato:
Partially resolve https://github.com/microsoft/CCF/issues/4816 (ECDSA)

Basically the change does the same thing as node.js does https://github.com/nodejs/node/blob/v19.4.0/src/crypto/crypto_sig.cc#L170


src/crypto/ecdsa.cpp
```diff
@@ -50,4 +51,24 @@ namespace crypto
...
+    auto sig_ptr = signature.data();
+    OpenSSL::Unique_ECDSA_SIG ecdsa_sig(
+      d2i_ECDSA_SIG(NULL, &sig_ptr, signature.size()));
+    // r and s are managed by Unique_ECDSA_SIG object, so we shouldn't use
+    // Unique_BIGNUM for them
+    const BIGNUM* r = ECDSA_SIG_get0_r(ecdsa_sig);
+    const BIGNUM* s = ECDSA_SIG_get0_s(ecdsa_sig);
+    int nid = PublicKey_OpenSSL::get_openssl_group_id(curveId);
+    OpenSSL::Unique_EC_GROUP ec_group(nid);
+    int group_order_bits = EC_GROUP_order_bits(ec_group);
```
achamayou, line 67: A check that group_order_bits is positive would be good, just out of caution.

takuro-sato: Added it here [1e46300](https://github.com/microsoft/CCF/pull/4829/commits/1e46300951dc7f4a676836b84fb44241e1fca37d)

achamayou (approved):

achamayou:
@takuro-sato please add an entry to the changelog, since this is a user-visible change.

PR #4835, opened 2023-01-12
Structured logging in raft.h do identify the current node, membership…

lemmy:
…, and leadership state in raft_scenario testing.


eddyashton (changes requested):
This will lead to a lot of extra noise in the production (`INFO` level) logs, where each node's output will only ever contain its own node_id. However I agree it would be useful for the Raft driver and unit test output, where we're running multiple Raft instances with interleaved logs.

I think we need to make a way this opt-in for specific test cases which want this verbosity, at least for the node_id (maybe the leadership state and membership state are more broadly useful? But we currently try to log on all _transitions_ of those states instead, so repeating usually-static values, in every log line, may still be unhelpful noise in production).

We should reduce the code duplication by implementing these as macros, something like:
```
#define RAFT_DEBUG_FMT(s, ...) \
  LOG_DEBUG_FMT("N: {} L: {} M: {} - {}", state->my_node_id, lstos(), mstos(), s, ##__VA_ARGS__)
```

With equivalents for `RAFT_INFO_FMT`, and whatever other levels are used. Then the opt-in might be based on a `#define`:
`raft.h`:
```
#ifdef VERBOSE_RAFT_LOGGING
#define RAFT_DEBUG_FMT(s, ...) \
  LOG_DEBUG_FMT("N: {} L: {} M: {} - {}", state->my_node_id, lstos(), mstos(), s, ##__VA_ARGS__)
#else
#define RAFT_DEBUG_FMT LOG_DEBUG_FMT
#endif
```
`driver.cpp`:
```
#define VERBOSE_RAFT_LOGGING
#include "raft.h"
...
```
Additionally we should use longer function names:
```
lstos -> leadership_state_to_string
mstos -> membership_state_to_string
```

achamayou:
+1 for keeping the INFO level relatively terse to minimise impact on current operational use, although I think it falls somewhat short in terms of events being logged. We can review that separately.

Converting all DEBUG or TRACE in raft to a structured format is definitely a welcome change, duplicating INFO messages where they don't offer enough information seems ok to me.

lemmy:
> I think we need to make a way this opt-in for specific test cases which want this verbosity, at least for the node_id (maybe the leadership state and membership state are more broadly useful? But we currently try to log on all _transitions_ of those states instead, so repeating usually-static values, in every log line, may still be unhelpful noise in production).

A current log statement has `100` in place of the real thread id under raft_scenario tests:

```
2023-01-13T17:47:32.686904Z        100 [debug] ../src/consensus/aft/raft.h:1282     | N: n[1] L: F M: a - Received append entries: 1.2 to 1.2 (from n[0] in term 2)
```

I considered using the `node_id` instead `100` in raft_scenarios, but didn't spend too much time on it.  Would this be acceptable?  The info statement would look like this:

```
2023-01-13T17:47:32.686904Z        [n1] [debug] ../src/consensus/aft/raft.h:1282     | L: F M: a - Received append entries: 1.2 to 1.2 (from n[0] in term 2)
```

achamayou:
@lemmy tid 100 is host-side execution, as opposed to inside the enclave (where actual tids are logged). It's not useful in raft_scenarios, and fine to change the format to replace it with something useful in that context.

For general debug output in enclave though, it's useful to have.

lemmy:
Possibly related: https://github.com/microsoft/CCF/pull/4900

eddyashton:
@lemmy - I've applied my suggestions above in this PR - #4933. Please let me know if that still does what you need (structured machine-parseable output from `raft_driver`), and we can merge that and close this.

If we want to change the existing logging format, for instance to remove the thread ID in `raft_driver` since its information-free, we could do that by declaring a custom logger. So rather than calling:

```
logger::config::add_text_console_logger();
```

in `driver.cpp:main`, we'd do something like (uncompiled sketch):

```
class MyLogger : public AbstractLogger
{
  public:
   void write(const LogLine& ll, const std::optional<double>& enclave_offset = std::nullopt) override
   {
      std::cout << ll.log_level << " | " << ll.msg << std::endl;
   }
};

...

logger::get_loggers().emplace_back( std::make_unique<MyLogger>() );
```

That lets us rewrite anything before the "msg" part (before the first `|` in the output line), if we want it to be denser or more easily parsed.

PR #4837, opened 2023-01-13
Wait for SSH setup in deployment python and fix workspace cleanup

DomAyre:


.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -50,16 +50,6 @@ jobs:
         env:
           CCF_AZURE_SUBSCRIPTION_ID: $(CCF_AZURE_SUBSCRIPTION_ID)
 
-      - script: |
```
jumaffre, line 53: Why is this removed? It seems that this is a useful step to make sure the container is ready before proceeding?

DomAyre: It was a suggestion from Amaury to only wait on it when we actually need it, so the changes in test_on_remote should now wait on the deployment just before running ctest

DomAyre: Also the main reason I'm doing this is that I've seen some failures this morning where the pipeline doesn't wait long enough, the script returns but ACI Is still setting up SSH

achamayou: @jumaffre this isn't necessary until we get to test_on_remote "Test Connection to Deployed ACIs", which does the exact same thing. So we can use the time to wrap up, and do the set up for test_on_remote.

Now that I think about it, I do not understand why they are two separate jobs, since they seem to run on the same pool. Can we not just have them be one job? They can still be separate files/templates (see common.yml for example).

.azure-pipelines-templates/test_on_remote.yml
```diff
@@ -28,12 +28,14 @@ jobs:
       - script: |
           set -ex
           IFS='\n' read -ra IP_ADDR_LIST <<< "$(RUN_ON)"
-          for IP_ADDR in "${IP_ADDR_LIST[@]}"; do
-            echo -e "Testing connection with $IP_ADDR"
-            ssh agent@$IP_ADDR -o "StrictHostKeyChecking no" 'echo "Connected successfully"'
+          echo -e "Testing connection with ${IP_ADDR_LIST[0]}"
+          for i in {1..10}; do
+            ssh agent@${IP_ADDR_LIST[0]} -o "StrictHostKeyChecking no" 'echo "Connected successfully"' && break
+            sleep 20
```
achamayou, line 34: Can we have a smaller reconnection interval? 20 seconds seems really long.

jumaffre (approved):

.azure-pipelines-templates/deploy_aci.yml
DomAyre: Yeah I'll make a second PR to combine the jobs, and after chatting to Julien we agreed waiting for the commands that setup SSH should be done from the python infra to deploy

PR #4838, opened 2023-01-13
Add `enableUntrustedDateTime` to JS API

### LINKED ISSUE #4826: Expose host time to JS, via standard APIs
This is a revival of #2468, following multiple user requests. That issue proposed adding custom APIs to make it clear that this time was untrusted, but on balance it looks like standard API compatibility is broadly preferred - we'll call out the implementation details/risks in the documentation.

I would like to retain a flag in `app.json` which makes this opt-in, but that does make the implementation trickier - `Date` global is auto-inserted by QuickJS, we shim the implementation of `gettimeofday`, and making it opt-in implies making that stateful.

A minimal effort approach would be to remove the shim entirely, and use OE's implementation of `gettimeofday` as an OCall. But I think that has knock-on implications that we're unhappy with, aside from the performance impact on code actually using it.

eddyashton:
Resolves #4826.

Includes an additional patch to `quickjs.c` to support this.

We discussed enabling this with a configuration option, probably when apps are submitted or perhaps in subsequent governance. I realised we can actually offer even more fine-grained control (ie - opt-in during a single function call) if we make activation a call in the ?JS API itself, and this also makes it more explicit in the JS source what the behaviour will be. For users who want to opt-in globally, they can still do so by calling the opt-in function at the global scope in their application code.

The default behaviour remains the same - `Date.now()` returns 0, so backporting this is actually _not a breaking change!_ (merely a new API).

I've tried to be thorough in the associated docs, but please suggest changes there if a) it doesn't make sense, or b) you think we should word our recommendations differently.


3rdparty/exported/quickjs/quickjs.c
```diff
@@ -48191,7 +48198,12 @@ static JSValue get_date_string(JSContext *ctx, JSValueConst this_val,
 }
 
 /* OS dependent: return the UTC time in ms since 1970. */
-static int64_t date_now(void) {
+// vvv CCF Patch vvv
+// Implement date_now as a macro, to additionally capture ctx argument from
+// callsite
+#define date_now() _date_now(ctx)
+static int64_t _date_now(JSContext* ctx) {
+// ^^^ CCF Patch ^^^
```
eddyashton, line 48206: An extremely unrelated aside: No sensible meta-programming/code generation language would let us do this! So maybe dumb old text-replacement macros still have a place.

achamayou (approved):

MahatiC (approved):

PR #4849, opened 2023-01-14
Fix SNP CI workspace cleanup command

DomAyre:
Workspaces need to be cleaned up after runs to ensure that the fileshare isn't filled up


.azure-pipelines-templates/install_ssh_key.yml
```diff
@@ -0,0 +1,8 @@
+steps:
+  - script: |
+      set -ex
+      mkdir ~/.ssh
+      echo ${{ parameters.host_private_key }} | base64 -d > ~/.ssh/id_rsa
+      sudo chmod 600 ~/.ssh/id_rsa
+    name: setup_key
+    displayName: "Install SSH Key fromπ Deployment Step"
```
jumaffre, line 8:
```suggestion
    displayName: "Install SSH Key from Deployment Step"
```

jumaffre (approved):

PR #4852, opened 2023-01-17
Upgrade httpx pin on release/2.x

achamayou:
This is only used for testing, but flagged by CG because of https://security-tracker.debian.org/tracker/CVE-2021-41945.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -13,6 +13,6 @@ docker
 better_exceptions
 pyasn1
 Jinja2
-httpx == 0.21.*
+httpx == 0.23.*
 locust
 pyOpenSSL


eddyashton:
I think we'd need to actually cut a 2.x release for this to help? I think we'd be better extending the CG detection exclusions to ignore this folder.

achamayou:
@eddyashton 2.0.x is still supported, so while I don't think we will cut a release just for that, we also want to be able to do it without including known issues for which a fix is readily available.

Because we include (albeit test) code in our .deb that makes use of httpx, I think it wouldn't be right to ignore this.

eddyashton (approved):

PR #4854, opened 2023-01-17
Fix count arg in ACI deployments

DomAyre:
For deploying multiple ACIs, we actually want to deploy multiple container groups each with a single container.

Spun out of #4825

### DIFF
--- a/scripts/azure_deployment/arm_aci.py
+++ b/scripts/azure_deployment/arm_aci.py
@@ -206,7 +206,7 @@ def make_aci_deployment(parser: ArgumentParser) -> Deployment:
     }
 
     for i in range(args.count):
-    
+
         if not args.attestation_container_e2e:
             deployment_name = args.deployment_name
             container_name = args.deployment_name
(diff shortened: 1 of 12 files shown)


scripts/azure_deployment/arm_aci.py
```diff
@@ -205,77 +205,79 @@ def make_aci_deployment(parser: ArgumentParser) -> Deployment:
...
+            deployment_name = args.deployment_name
+            container_name = args.deployment_name
+            container_image = args.aci_image
+            command = make_dev_container_command(args)
+            with_volume = args.aci_file_share_name is not None
+            containers = [
+                make_dev_container(
+                    i, container_name, container_image, command, args.ports, with_volume
+                )
+                for i in range(args.count)
```
jumaffre, line 220: This should go then?

DomAyre: Whoops yeah that's a merge conflict typo, good spot

DomAyre: https://github.com/microsoft/CCF/pull/4854/commits/ff816e90f624ca165b297fe1ec3312ea9057e3db

takuro-sato (approved):

takuro-sato:
I think we don't have to care about 'Attestation COntainer' failure since it's related to https://statics.teams.cdn.office.net/evergreen-assets/safelinks/1/atp-safelinks.html

Issue #4804, opened 2023-01-06
attestation-container: Use Unix domain socket for gRPC


takuro-sato:
Attestation-container needs to be implemented so that it can be accessed only locally.
Our original plan was to enforce it by setting network config for ACI's security policy, but actually it's not supported by security policy.

So now we want to use Unix domain socket (UDS) instead, because it only works locally and it still can be used with gRPC interface (We want to use gPRC because attestation-container is used by external executer which already has gRPC as a dependency).
example implementation: https://github.com/devlights/go-grpc-uds-example/blob/master/cmd/server/server.go

Acceptance criteria:
- attestation-container uses UDS for gPRC
- The volume for UDS is encrypted and local (not shared by network). It should be enforced by security policy.

takuro-sato:
Finished

PR #4858, opened 2023-01-17
Use Unix domain socket

takuro-sato:
https://github.com/microsoft/CCF/issues/4804

## Changes

- Use Unix domain socket
- For CI, use [emptyDir](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-volume-emptydir) (example: https://github.com/microsoft/confidential-sidecar-containers/blob/main/examples/encfs/aci-arm-template.json)

## TODOs (for future PR)
- Apply appropriate security policy using `az confcom acipolicygen` tool https://github.com/microsoft/CCF/issues/4881
  I am getting an error when I use the tool. So it would be smoother if I can work on it separately.


attestation-container/Dockerfile
```diff
@@ -13,6 +13,10 @@ COPY go.mod go.sum ./
 RUN go mod download && go mod verify
 
 COPY . .
+
+# Pre-compile tests to avoid I/O errors in ACI
+RUN go test -c && cd attest && go test -c && cd ..
```
takuro-sato, line 18: This is related to the workaround for the ACI IO error. 
[Using external volume](https://github.com/microsoft/CCF/pull/4860) worked when we run only unit test on ACI. But now E2E test is run on ACI as well. Using external volume avoids the IO error but somehow `go test` command is stuck while downloading go packages for E2E test.

It's hard to find out the cause because even `go test -x` is not verbose enough.

So I just instead compile the test code in the docker file. Considering test executables are not necessary in the production, I want to delete them after IO error is resolved.

attestation-container/attestation-container_test.go
```diff
@@ -5,31 +5,36 @@ import (
...
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/insecure"
 )
 
 var (
-	addr = flag.String("addr", "localhost:50051", "the address to connect to")
+	addr = flag.String("addr", "/tmp/attestation-container.sock", "the Unix domain socket address to connect to")
 )
 
+const GPRC_TIMEOUT_IN_SEC = 10
```
takuro-sato, line 21: This is not related to UDS itself, but I faces 'not responded' error in the CI several times. It's the fix for it and looks like it's working.

takuro-sato: Same fix was made in https://github.com/microsoft/CCF/pull/4858

achamayou (approved):

attestation-container/Dockerfile
jumaffre: We could also add a build arg (defaults to false) to the Dockerfile to decide whether the test are built in the image or not based on that?

jumaffre (approved):

attestation-container/Dockerfile
takuro-sato: That's a nice idea! Made that change in [8ebd3a7](https://github.com/microsoft/CCF/pull/4858/commits/8ebd3a7c6cbe7c896f03714d6449423bfc6c1103)

PR #4860, opened 2023-01-17
Fix io error of attestation container pipeline

takuro-sato:
When 'Attestation Container' pipeline was created, the ACI io error didn't happen so often because it has less data to write to disk than 'CCF SNP' pipeline.
So it doesn't have azure file share for workaround, but somehow today it's more flaky.
This PR add the file share for the pipeline


.azure-pipelines-templates/deploy_attestation_container.yml
```diff
@@ -58,15 +58,34 @@ jobs:
...
+      - template: install_ssh_key.yml
+        parameters:
+          ssh_key: $(sshKey)
+
+      - script: |
+          set -ex
+          BUILD_NUMBER=$(echo $(Build.BuildNumber) | sed 's/\.//g')
+          IP_ADDR=$(echo "$(IpAddresses)" | awk -v container_group="ci-$BUILD_NUMBER-business-logic-0" '$1 == container_group {print $2}')
+          ssh agent@$IP_ADDR -o "StrictHostKeyChecking no" '
+          rm -rf /acci/workspace_$(Build.BuildNumber)
```
jumaffre, line 84: Is this needed? The workspace directory is usually where the CCF nodes are started but that shouldn't be the case with the deployment of the attestation container?

scripts/azure_deployment/arm_aci.py
```diff
@@ -32,7 +32,7 @@ def get_pubkey():
 STARTUP_COMMANDS = {
     "dynamic-agent": lambda args: [
         "apt-get update",
-        "apt-get install -y openssh-server rsync",
+        "apt-get install -y openssh-server rsync sudo",
```
jumaffre, line 35: I'm surprised this is necessary?

jumaffre (approved):

scripts/azure_deployment/arm_aci.py
takuro-sato: Yes, in the golang image, sudo is not included

PR #4866, opened 2023-01-19
Install a common ssh key on deployed ACIs

DomAyre:
To give ACIs the ability to SSH to each other, install the key used by azure pipelines into each of them which is already authorized


jumaffre (approved):
Is it necessary the private key on the secondary containers? My understanding is that only the primary would need this to be able to use the secondary containers as CCF remotes?

DomAyre:
It's not necessary, I just thought may as well install on every ACI but happy to just install on the primary if you'd prefer

PR #4867, opened 2023-01-20
Pending nodes must not be part of a configuration.

lemmy:
A pending node has passed attestation checks, but is waiting for member confirmation, and just isn't part of any configurations at all, nor has a communication channel with other nodes (Pending is *not* Learner in https://microsoft.github.io/CCF/main/architecture/consensus/1tx-reconfig.html).

PR #4868, opened 2023-01-20
Nodes may not rejoin a configuration.

lemmy:
(see "to stop a node being removed and added back again" comment in https://github.com/microsoft/CCF/pull/3965)


achamayou:
LGTM, but @heidihoward please confirm.

lemmy:
FYI: In line with this PR and PRs https://github.com/microsoft/CCF/pull/4867, to avoid merge problems, another PR is in my pipeline that excludes `retired` nodes from joining configurations.

heidihoward (commented):
It might be simplier to implement this by added a check to the `ChangeConfiguration` action which checks the log for reconfiiguration transactions to confirm that the new configuration does not rejoin any nodes. Note that a node's `log`, unlike `configurations`, has a full history of previous configurations. This avoids the extra variable and works for both the simulation and model checking

tla/ccfraft.tla
```diff
@@ -104,8 +104,18 @@ ASSUME Servers \subseteq AllServers
...
-
-reconfigurationVars == <<reconfigurationCount, configurations>>
+\* The set of servers that have been removed from configurations.  The implementation
+\* assumes that a server refrains from rejoining a configuration if it has been removed
+\* from an earlier configuration (relying on the TEE and absent Byzantine fault). Here,
+\* we model removedFromConfiguration as a global variable. In the implementation, this
+\* state has to be maintained by each node separately.
+\* Note that we cannot determine the removed servers from configurations because a prefix
+\* of configurations is removed from the configuration on a change of configuration.
+\* TODO: How does the implementation keep track of removed servers?  Can we remove and
```
heidihoward, line 114: The implementation does not explicitly track removed servers. If a node fails, then it rejoins with a new identity (and thus a new node ID). Node ID are unique (assuming a collision resistant hash function etc).

heidihoward: The way I think about it, node rejoining is a specification issue that exists because the current specification only covers a particular aspect of CCF and could do a better job at capturing the assumptions it is making about the rest of CCF that is not specified in TLA+.

heidihoward (approved):

PR #4875, opened 2023-01-23
Update SNP Measurement

DomAyre:


DomAyre:
Affected test now passes so merging
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=61663&view=logs&j=4bdd83b0-b422-5e84-0058-fb00a4da194e

Issue #4865, opened 2023-01-19
Add a JWK public key to PEM converter in the JS API


achamayou:
See #4321, implemented as #4359 for the opposite direction. Underlying logic is being added as part of https://github.com/microsoft/didx509cpp integration, we just need to expose it when complete.

Requirement raised in https://github.com/microsoft/did-ccf/issues/20

PR #4876, opened 2023-01-23
Add JWK to PEM converters

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/4865

This PR adds support for JWK to PEM conversion for all supported key types (EC, RSA, EdDSA) for both the C++ and JS/TS APIs.


include/ccf/crypto/jwk.h
```diff
@@ -88,13 +103,17 @@ namespace crypto
     JsonWebKeyECCurve crv;
     std::string x; // base64url
     std::string y; // base64url
+
+    bool operator==(const JsonWebKeyECPublic&) const = default;
```
takuro-sato, line 107: Just question, what is the reason that we need to define default `==` operator explicitly? Doesn't the compiler create one automatically?

js/ccf-app/test/polyfill.test.ts
```diff
@@ -510,21 +510,29 @@ describe("polyfill", function () {
           const jwk = ccf.crypto.pubPemToJwk(pair.publicKey);
           assert.equal(jwk.kty, "EC");
           assert.notEqual(jwk.kid, my_kid);
+          const pem = ccf.crypto.pubJwkToPem(jwk);
```
takuro-sato, line 513: nit: Maybe change `describe("pemToJwk", ...` to `describe("pemToJwk and jwkToPem", ...` or something like that?
I can put comment only on changed lines, so I put it here.

src/crypto/openssl/key_pair.cpp
```diff
@@ -65,6 +65,20 @@ namespace crypto
       throw std::runtime_error("could not parse PEM");
   }
 
+  KeyPair_OpenSSL::KeyPair_OpenSSL(const JsonWebKeyECPrivate& jwk)
+  {
+    auto ec_key = PublicKey_OpenSSL::ec_key_public_from_jwk(jwk);
+
+    Unique_BIGNUM d;
+    auto d_raw = raw_from_b64url(jwk.d);
+    BN_bin2bn(d_raw.data(), d_raw.size(), d);
```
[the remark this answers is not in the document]
takuro-sato: If we can agree on the change, there are several places that `BN_bin2bn` is used.

takuro-sato (approved):

src/crypto/openssl/rsa_key_pair.cpp
```diff
@@ -39,6 +39,40 @@ namespace crypto
...
+
+    BN_bin2bn(d_raw.data(), d_raw.size(), d);
+    BN_bin2bn(p_raw.data(), p_raw.size(), p);
+    BN_bin2bn(q_raw.data(), q_raw.size(), q);
+    BN_bin2bn(dp_raw.data(), dp_raw.size(), dp);
+    BN_bin2bn(dq_raw.data(), dq_raw.size(), dq);
+    BN_bin2bn(qi_raw.data(), qi_raw.size(), qi);
+
+    CHECK1(RSA_set0_key(rsa, nullptr, nullptr, d));
+    CHECK1(RSA_set0_factors(rsa, p, q));
```
takuro-sato, line 62: We need to put `d.release()` above this line?
In case this line throws, the content of d is freed by `RSA_set0_key` and `Unique_BIGNUM` tries to free it again.

include/ccf/crypto/jwk.h
jumaffre: I thought that too, but IIRC, it turned out necessary because the `operator==()` was defined for the base object `JsonWebKey` and so comparisons between derived objects (e.g. `JsonWebKeyECPublic`) invoked the base class `operator==()` rather than the derived one.

src/crypto/openssl/key_pair.cpp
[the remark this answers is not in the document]
jumaffre: Very good point. I've added extra checks here and elsewhere.

PR #4888, opened 2023-01-24
Setup environment variables for ACI

### LINKED ISSUE #4622: Pass ACI SEV-SNP envvars to all nodes on startup
This includes:
- `UVM_SECURITY_POLICY`
- `UVM_REFERENCE_INFO`

jumaffre:
Resolves #4622 

This PR passes the `UVM_...` environment variables from the ACI environment to the CCF nodes as part of the Python infra. The names of these variables are specified in the CCF node JSON configuration file as part of the `attestation.environment` section. 

Note that while these values are exposed to SEV-SNP-enabled containers, they are only set for PID 1 (i.e. to the `command` set in the ARM template) so we write these to a new `/env` file first to have the Python infra pass these to each node on startup.

This has implications for some end-to-end tests in `code_update.py` which used to rely on the fact that no environment variable was passed to the CCF nodes by default, so these have been reworked accordingly.


samples/constitutions/default/actions.js
```diff
@@ -996,27 +996,26 @@ const actions = new Map([
       function (args) {
         checkType(args.security_policy, "string", "security_policy");
         checkType(args.host_data, "string", "host_data");
-      },
-      function (args, proposalId) {
-        const host_data = ccf.strToBuf(args.host_data);
-        const security_policy = ccf.jsonCompatibleToBuf(args.security_policy);
 
+        // If optional security policy is specified, make sure its
```
jumaffre, line 1000: Moving this check to new "validate" part of proposal action as this can be entirely checked on proposal creation rather than execution as this doesn't interact with the KV.

src/host/main.cpp
```diff
@@ -130,7 +146,17 @@ int main(int argc, char** argv)
...
+    "Configuration file {}:\n{}", config_file_path, config_json.dump(2));
+
+  nlohmann::json environment;
+  for (int i = 0; environ[i] != nullptr; i++)
+  {
+    auto [k, v] = nonstd::split_1(environ[i], "=");
+    environment[k] = v;
+  }
+
+  LOG_INFO_FMT("Environment: {}\n", environment.dump(2));
```
jumaffre, line 159: Note: we now also dump (formatted as a nice JSON object, on a single line) the environment variables passed to the node on startup.

src/host/main.cpp
```diff
@@ -404,18 +430,23 @@ int main(int argc, char** argv)
...
+          "attestation security policy");
+    }
+
+    if (config.attestation.environment.uvm_endorsements.has_value())
+    {
+      startup_config.attestation.environment.uvm_endorsements =
+        read_required_environment_variable(
+          config.attestation.environment.uvm_endorsements.value(),
+          "UVM endorsements");
     }
```
jumaffre, line 450: Note that this is currently unused. The work to parse, verify and store this will be done in https://github.com/microsoft/CCF/issues/4193.

DomAyre (approved):

PR #4890, opened 2023-01-24
`cchost --version` reports target platform

jumaffre:
Now that we have a unique `cchost` name and multiple platforms, but a given built `cchost` can only launch one type of enclave, we also report the platform when running `cchost --version`.


achamayou (approved):

achamayou:
CLA bot looks stuck again, but every other check is green.

PR #4898, opened 2023-01-25
Use `az confcom acipolicygen` to generate security policy

takuro-sato:
Use `az confcom acipolicygen` to generate security policy automatically.

The tool will be in public preview and be added to https://github.com/Azure/azure-cli-extensions in the next couple weeks.
I only applied the tool to attestation container so that at least one pipeline (SNP CI) uses the current open door policy.


.azure-pipelines-templates/azure_cli.yml
```diff
@@ -2,6 +2,9 @@ steps:
   - script: |
       set -ex
       curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+      az extension add --source https://acccliazext.blob.core.windows.net/confcom/confcom-0.2.5-py3-none-any.whl -y
```
takuro-sato, line 5: After the extension being in public preview, we can install the latest version automatically by `az extension update --name confcom`.
But for now we need to manually manage the version (currently 0.2.5)

scripts/azure_deployment/arm_aci.py
```diff
@@ -390,6 +403,32 @@ def make_aci_deployment(parser: ArgumentParser) -> Deployment:
...
+                if completed_process.returncode != 0:
+                    arm_template_string = json.dumps(arm_template, indent=4)
+                    raise RuntimeError(
+                        f"Generating security policy failed with status code {completed_process.returncode}: {completed_process.stdout}, arm_template: {arm_template_string}"
+                    )
+
+                with open(arm_template_path, "r") as f:
+                    # Read the template file overwritten by acipolicygen tool
+                    arm_template_string = f.read()
+                    arm_template = json.loads(arm_template_string)
```
jumaffre, line 430:
```suggestion
                    arm_template = json.load(f)
```

.azure-pipelines-templates/azure_cli.yml
jumaffre: Could you add this as a comment in the code too?

jumaffre (approved):

Issue #4889, opened 2023-01-24
Deprecate `commit_seqno` and `commit_view` in signature transactions


heidihoward:
`commit_seqno` and `commit_view` in signature txns currently point to the previous txn, which is not yet committed. This field is not used  as it is not signed, however, it caused me some confusion today so after discussion with @achamayou and @eddyashton, it seems best to zero it and document that it is kept only for backwards compatibility.

eddyashton:
Zero'd in #4899.

PR #4899, opened 2023-01-25
Remove commit claims from signature transaction

eddyashton:
These were inaccurate and unused.

For compatibility (eg - interop with older nodes which treat these as required fields), they are still present in each signature transaction, but will always be 0. We can deprecate/remove them more aggressively in future, but need to make our JSON serdes asymmetric to do so (ie - have an intermediate build which is fine with them not existing, but will always emit the 0s for nodes which do still require them).


src/consensus/aft/raft.h
```diff
@@ -116,7 +116,7 @@ namespace aft
     // the current view. This signature is the first thing they may commit, as
     // they cannot confirm commit of anything from a previous view (Raft paper
     // section 5.4.2). This bool is true from the point this node becomes
-    // primary, until it is explicitly reset by a call to get_signable_txid()
+    // primary, until it sees a committable entry
```
eddyashton, line 119: This is a drive-by, tangentially related change, so warrants some scrutiny. As this comment explains, we added a bool to track when this node has recently become primary and would like to emit its first signature in the new term. We previously reset this inside `get_signable_txid()`, with a few scary comments about assuming that the calling path proceeded to actually emit a signature. I think its simpler and safer to reset the bool later, when the emitted signature reaches Raft in `replicate()`. In practice this means "slightly later in exactly the same execution stack", and in theory introduces "we might produce multiple start-of-term signatures if things get weirdly delayed" and "we might be automagically retry if a first signing attempt fails". Leaning heavily on the existing tests to check that this is safe.

heidihoward: Someone other than me should review this code change but from a consensus perspective, I have a strong preference for might produce multiple sigs over might not emit any sigs at the start of a term

eddyashton:
Resolves #4889.

src/consensus/aft/raft.h
achamayou: Unless this change is somehow necessary to remove the claims, I would like to see it split and some tests added to check how many extra signatures, if any, it results in.

src/node/history.h
```diff
@@ -749,27 +743,26 @@ namespace ccf
...
-        commit_txid.version,
-        commit_txid.previous_version);
+        previous_signature_seqno);
 
       store.commit(
         txid,
         std::make_unique<MerkleTreeHistoryPendingTx<T>>(
-          txid, commit_txid, store, *this, id, kp, endorsed_cert.value()),
+          txid,
+          previous_signature_seqno,
```
achamayou, line 760: Why are we passing the previous signature seqno? I thought we'd agreed to zero this out?

achamayou:
Let's add a check in the test that audits the ledger asserting that we produce the values we expect.

src/consensus/aft/raft.h
eddyashton: Sure, I've reverted this for now and will do it in a separate PR.

eddyashton:
> Let's add a check in the test that audits the ledger asserting that we produce the values we expect.

Added to `check_operations` as part of `governance_history.py` here: https://github.com/microsoft/CCF/pull/4899/commits/7fa4ba3a7c398224ce6acce51b3c0c953da78fba

This test is primarily checking the structure of governance tables, so maybe a signature check doesn't belong alongside? But I think its the best test case we currently have making assertions on specific ledger table's contents.

src/consensus/aft/raft.h
```diff
@@ -799,6 +795,8 @@ namespace aft
           }
           committable_indices.push_back(index);
           start_ticking_if_necessary();
+
+          should_sign = false;
```
achamayou, line 799: Why is should_sign being set here?

eddyashton: Because I tried to revert those changes by hand, and missed this. Removed it now.

PR #4900, opened 2023-01-26
Simplify assignment of thread IDs

eddyashton:
Our in-enclave async tasks use `ThreadMessaging`, which depends on `thread_ids.h`. In a real CCF node, there was a `thread_ids` map populated by each thread claiming an ID as it ECALLed into `enclave_run`. This had no parallel in unit tests, so the thread ID assignments there were empty, and any systems using `ThreadMessaging` would have very different behaviour in unit tests than in a running node.

This PR attempts to simplify and automate the thread ID assignment, so it is operates similarly in unit tests. In addition, I've removed the need for multiple static variables instances, and tried to encapsulate them in a single `ThreadMessaging` singleton instance - we already had this, but some of the functionality was on the instance and some on standalone static vars. In particular, there's now an explicit `init` step where each test declares how many threads its using, in the same way the node does. Accessing the singleton (ie - queuing async work) before `init` results in a clear error.


src/ds/thread_messaging.h
```diff
@@ -224,94 +224,122 @@ namespace threading
       }
     }
 
-  public:
-    static ThreadMessaging thread_messaging;
-    static std::atomic<uint16_t> thread_count;
-    static const uint16_t main_thread = MAIN_THREAD_ID;
+    inline TaskQueue& get_tasks(uint16_t task_id)
```
jumaffre, line 227: Should `task_id` be `thread_id` here, as in "get the tasks for this thread"?

.daily_canary
```diff
@@ -1,4 +1,4 @@
    ___     ___
-  (- *)   (O o)   |     Y   &  +
```
jumaffre, line 2: I know it's already unstable, but should we run the multi-threaded CI (`.threading_canary`) a few times over this PR?

eddyashton: Sure, I'll run a few. It'll add a red badge to this PR we'll need to ignore, but I guess we can look at the failures and see if they look different.

src/ds/thread_messaging.h
eddyashton: I'd like to move towards separating this job-system task-queue from thread affinity, and to think of this as a fixed-size task-queue collection, keyed by `task_id`. We happen to maintain 1 queue-per-thread, but that's handled elsewhere. Admittedly, this is a soft and poorly maintained distinction, not least because this whole "job system" is called `ThreadMessaging`, but the renames here are an attempt to move in that direction.

.daily_canary
eddyashton: So far, failures look the same - `partitions_test` is failing persistently, `reconfiguration_test_cft` and `governance_test` are failing most of the time.

src/consensus/aft/test/main.cpp
```diff
@@ -7,8 +7,8 @@
 #define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
 #include <doctest/doctest.h>
 
-threading::ThreadMessaging threading::ThreadMessaging::thread_messaging;
-std::atomic<uint16_t> threading::ThreadMessaging::thread_count = 1;
+std::unique_ptr<threading::ThreadMessaging>
```
jumaffre, line 10: I suspect there's a good reason why this isn't possible, but instead of having to declare and define one of these for every object file, could there be a single static instance of it in `thread_messaging.h` so that including `thread_messaging.h` is enough to be able to schedule async messages?

jumaffre (approved):

src/consensus/aft/test/main.cpp
eddyashton: No - we can (and do!) declare but not define the static instance in the header. The neater solution would be to define it once in a separate module, and link that to these tests, but we've stuck to larger single binaries rather than standalone linkables for the most part.

lemmy:
Can this change also be repurposed for https://github.com/microsoft/CCF/pull/4835#issuecomment-1382192876 ?

PR #4901, opened 2023-01-26
Change deployment region for ACI to WestEurope

jumaffre:
Deployments to EastUS2euap seem to be failing today so attempting to temporarily deploy to WestEurope instead.


DomAyre:
Should we keep the default and just use the --region arg in the pipeline?

jumaffre:
Region was unrelated to issue. See #4903

Issue #4815, opened 2023-01-10
Handle `CompactedVersionConflict` during external execution


eddyashton:
Background for this behaviour is (way back) in #1615, and the fix #1617.

When we have a long-lived transaction that straddles a commit (and compaction of the in-memory KV roll), its possible to request a handle on a map at a version which is no longer available - we're reading at version 10, but the KV has been compacted and only contains maps back to version 12. That's not a problem if we already have a handle to a view on that map (its lifetime is extended by any `Tx` still using it), but is a problem for newly requested views. We resolved that by throwing a custom exception type (`CompactedVersionConflict`) in this scenario, and re-executing the transaction through the normal conflict resolution/retry logic in `frontend.h`.

That does not interact well with external execution. Its currently possible for external execution to create transactions straddling a commit boundary (essentially, creating long-lived, interleaved transactions, that previously didn't occur because we're single threaded). I think this is happening very occasionally in the existing `external_executor` test during `test_multiple_executors`, and definitely since #4809 - I've got regular repros there by running in a loop. With a bunch of extra debug logging, the failure point looks like this:

```
2023-01-10T16:40:28.445102Z -0.002 0   [fail ] ../src/node/rpc/frontend.h:599       | Transaction execution conflicted with compaction: Unable to retrieve state over map wiki_descriptions at 94
2023-01-10T16:40:28.445131Z -0.002 0   [fail ] ../src/node/rpc/frontend.h:376       | Retry #1 for POST /externalexecutor.protobuf.KV/Get
2023-01-10T16:40:28.445137Z -0.002 0   [fail ] ../src/node/rpc/frontend.h:599       | Transaction execution conflicted with compaction: Unable to retrieve state over map wiki_descriptions at 94
2023-01-10T16:40:28.445141Z -0.002 0   [fail ] ../src/node/rpc/frontend.h:376       | Retry #2 for POST /externalexecutor.protobuf.KV/Get
...
2023-01-10T16:40:28.453851Z -0.010 0   [fail ] ../src/node/rpc/frontend.h:376       | Retry #29 for POST /externalexecutor.protobuf.KV/Get
2023-01-10T16:40:28.454912Z -0.011 0   [fail ] ../src/node/rpc/frontend.h:599       | Transaction execution conflicted with compaction: Unable to retrieve state over map wiki_descriptions at 94
2023-01-10T16:40:28.454920Z -0.011 0   [fail ] ../src/node/rpc/frontend.h:645       | Consistent conflicts
```

[...]

PR #4902, opened 2023-01-26
[Draft] Convert `RpcFrontend::process` to be asynchronous

eddyashton:
_Draft for now - throwing at the CI, and working out if this is a worthwhile midpoint to check-in_

This is a precursor to fixing #4815 - making each `process` execution asynchronous (by queuing a `ThreadMessaging` task rather than inline in a single callstack), so that we can (potentially, unimplemented) queue retries from elsewhere.

This includes the changes from #4900, which were motivated by trying to debug/fix the unit test changes in this PR, so I'll rebase once that's in.


eddyashton:
Closing this PR. Work on external executors is on-hold, so this major architectural change will have no direct benefit yet imposes significant risk.

PR #4903, opened 2023-01-26
Use JSON security policy for ACI deployments

jumaffre:
Temporary fix as specifying a REGO security policy currently breaks deployment of SEV-SNP ACI containers.

### DIFF
--- a/tests/infra/is_snp.py
+++ b/tests/infra/is_snp.py
@@ -5,8 +5,21 @@ import os
 
 IS_SNP = os.path.exists("/dev/sev")
 # 4070: Get these from the container
+
+# JSON:
 DEFAULT_SNP_HOST_DATA = (
-    "cadd74f71ebb8fbf9782baab649612520f3c416bc5ce9ac768808ddc8d9b031c"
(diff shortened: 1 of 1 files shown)


DomAyre:
Do we need to update tests to match?

jumaffre:
> Do we need to update tests to match?
> 
> I can't make suggestions since the file isn't in scope for the PR and I can't push changes to your branch, so here will have to do, this is what we want in is_snp.py
> 
> ```
> DEFAULT_SNP_HOST_DATA = (
>     "ab6ccfea6e8c9c6233b1a4631cfd6440ecf910694b34d7e92e0236738758831e"
> )
> DEFAULT_SNP_SECURITY_POLICY = (
>     '{"allow_all":true,"containers":{"length":0,"elements":null}}'
> )
> DEFAULT_SNP_SECURITY_POLICY_B64 = "eyJhbGxvd19hbGwiOnRydWUsImNvbnRhaW5lcnMiOnsibGVuZ3RoIjowLCJlbGVtZW50cyI6bnVsbH19Cg=="
> ```

Ah yes of course. Note that we'll pick this up automatically from the environment when #4888 is merged.

DomAyre (approved):

PR #4908, opened 2023-01-27
Simplify reset of Raft's `should_sign`

eddyashton:
This is the extra change that was proposed in #4899, implemented in a separate PR and tested to confirm it doesn't introduce additional signatures at the end of several e2e tests.

The interesting changes are in the C++ files and `governance_history.py`. The other changes are just a side-effect of staring at the `partitions_test` again. I was going to assert that the _first_ transaction in any view is a signature, but that's not true - when we check that we can make progress, we do so extremely eagerly, and may submit a write transaction to a newly elected primary before they've emitted their first signature (with both the old code and the new). Tracking down the source of that surprising write transaction in the ledger was surprisingly hard, because it was in a private table. I've changed all of the "just checking we can make progress" writes to be _public_, so we can debug them more easily.


tests/partitions_test.py
```diff
@@ -69,7 +70,10 @@ def test_partition_majority(network, args):
     initial_view = None
     with network.partitioner.partition(partition):
         try:
-            network.wait_for_new_primary(primary)
+            network.wait_for_new_primary(
+                primary,
+                timeout_multiplier=3,
+            )
```
eddyashton, line 76: This is another `wait_for_new_primary` that's _expected to fail_, so I've reduced the multiplier in line with other instances. This brings the test runtime down.

achamayou (approved):

PR #4915, opened 2023-01-27
SEV-SNP ACI: Verify UVM endorsements

jumaffre:
Part of #4193 

Now that the UVM endorsements are exposed to the container via an environment variable (see #4888), we can verify those endorsements in the enclave at startup, and verify that the signed payload contains the same digest as the one included in the attestation report (`measurement` field). 

To do so, we first parse the value of the [`uvm_endorsements`](https://microsoft.github.io/CCF/main/operations/configuration.html#environment) environment variable as a CoseSign1 document, resolve the x509 DID against the certificate chain contained in the protected headers of the CoseSign1 document. The resolution yields a DID document containing the JWK for an RSA Public key that can be used to verify the signature of the CoseSign1 document (#4876 was required for this). Once the signature is verified, the payload of the CoseSign1 document is trusted and we can verify that the measurement matches the one in the attestation report.

Future PR(s):
- Retrieve DID from CoseSign1 issuer, once it includes EKU
- Store DIDs in new table, along with SVN (+ proposals to update it)
- Verify guest SVN
- Figure out and implement impact for join protocol

TODO in this PR:
- [x] Hardcode DID (for now)
- [x] Verify feed


src/node/node_state.h
```diff
@@ -288,6 +289,28 @@ namespace ccf
       if (code_id.has_value())
       {
         node_code_id = code_id.value();
+
+        if (!config.attestation.environment.uvm_endorsements.has_value())
```
achamayou, line 293: Is this so we can set a config value to check against? That doesn't provide a security property.

DomAyre (approved):

src/node/node_state.h
jumaffre: Yes a malicious operator could change the value of the environment variable, e.g. passing a COSE Sign1 document that endorses a compromised measurement of the UVM on which the CCF node is deployed. The corresponding DID will be recorded in the ledger though (still to do), and as for MRENCLAVE for SGX deployments, it is up to the members to check that the DID is trusted when the service is open. Happy to discuss the details further.

PR #4921, opened 2023-01-30
Add published address to node to node channel template

DomAyre:


tests/infra/remote.py
```diff
@@ -878,6 +879,7 @@ def __init__(
             if include_addresses:
                 cmd += [
                     f"--node-address={node_address}",
+                    f"--published-node-address={published_node_address}",
```
jumaffre, line 882: This code path (passing CLI arguments to `cchost`) is only required for old, 1.x, CCF nodes (as we introduced the JSON configuration in 2.x) as we still spin them up in the LTS compatibility test (the full fledged LTS test as run in the Daily pipeline). Because we're not planning to backport anything to 1.x, passing this new argument will actually break the LTS test so it shouldn't be specified here.

jumaffre (approved):

tests/infra/remote.py
DomAyre: https://github.com/microsoft/CCF/pull/4921/commits/1702c566656e2c1a145ecc71396e5f8871bc63ea

PR #4932, opened 2023-02-01
Upgrade confcom az cli extention to 0.2.8

takuro-sato:
We need this kind of manual upgrade until it will be public preview


.azure-pipelines-attestation-container.yml
```diff
@@ -7,8 +7,7 @@ pr:
       - .attestation_container_canary
       - attestation-container/*
       - .azure_pipelines_attestation_container.yml
-      - .azure-pipelines-templates/deploy_attestation_container.yml
-      - .azure-pipelines-templates/test_attestation_container.yml
+      - .azure-pipelines-templates/*
```
takuro-sato, line 10: This is not the point of this PR, but we often forgot to include files like `.azure-pipelines-templates/azure_cli.yml` to te `include:` directive. Using `*` might trigger unnecessary run, but I think it's less worth than that CI doesn't run by accident.

Hopefully in future Azure pipeline detects changes of dependency files automatically.

jumaffre (approved):

Issue #4827, opened 2023-01-12
SNP Code Update Test


DomAyre:
Could be an opportunity to use CS2 to spin up nodes, needs more investigation to see if we have access and if it's practical

Constraints:

- ACIs take at least a few minutes to deploy (up to 10+ minutes) so we need to not block on this
- Therefore, we want to deploy them ahead of time and pass them to the test
- There isn't a single initial entry point to our python tests, so doing this in the pipeline YAML makes the most sense.

Design:

- Spin up secondary containers alongside the primary which currently runs all tests, as another Azure pipelines job,
- Pass the location (file or env), where secondary container IPs will end up, to our e2e test args
- Code update test can then read that location, and waits for the IPs to be populated (shouldn't need to wait in practice since other tests will have been running while the deployment was happening)
- Once populated, test can read the IPs and spin up CCF nodes on those machines

Implementation Steps:

- [x] #4861  (@DomAyre )
- [x] ~Spin up nodes on ACIs to make a functioning multi container network (@jumaffre)~ After investigation, a container group cannot simultaneously be in a VNET and have a public IP address, which prohibits the use of VNET because the primary container, running `ctest`, needs to be public as it's running the CI agent. To circumvent this, we'll keep the secondary containers as public and simply expose an extra port for node-to-node communications.
- [ ] Write test
  - [ ] #4943
    - [x]  #4866

achamayou:
Covered under #4674.

PR #4935, opened 2023-02-01
Spin up CCF Nodes on Secondary ACIs

DomAyre:
Contributes to #4827


DomAyre:
PR from a branch name starting with ccf- breaks the LTS test

Issue #4637, opened 2022-11-24
Also fetch and verify CRL


DomAyre:

DomAyre:
Spun off from https://github.com/microsoft/CCF/issues/4144

Fetch and verify CRLs, and also be able to fetch CRLs and endorsements via THIM

PR #4940, opened 2023-02-01
SEV-SNP ACI: Retrieve attestation report endorsements from environment

jumaffre:
Part of https://github.com/microsoft/CCF/issues/4637

Now that the attestation report endorsements are also exposed to ACI via the `UVM_HOST_AMD_CERTIFICATE` environment variable, we use the value of it by default, if it's present, instead of fetching the endorsements from the Azure or AMD remote server. Even though we don't expect upcoming CCF deployments to fetch these endorsements remotely, we keep the feature in so that CCF can still be deployed on non-ACI SEV-SNP.


src/node/node_state.h
```diff
@@ -374,42 +375,74 @@ namespace ccf
...
+                config.attestation.environment.report_endorsements.value()));
+
+            CCF_ASSERT_FMT(
+              quote_info.endorsements.empty(),
+              "No endorsements should be set by quote generation");
+
+            quote_info.endorsements.insert(
+              quote_info.endorsements.end(),
+              endorsements.vcek_cert.begin(),
+              endorsements.vcek_cert.end());
```
takuro-sato, line 438: Is it inserting `std::string` ( = `std::basic_string<char>`) to `std::vector<uint8_t>` ?
If so it might be unsafe because char might not be 8 bit. 
https://en.cppreference.com/w/cpp/language/types
(Practically it should be fine because this code runs only on SEV-SNP machine)

takuro-sato (approved):

CMakeLists.txt
```diff
@@ -103,6 +103,8 @@ if(COMPILE_TARGET STREQUAL "sgx")
     ccf.enclave ${CCF_IMPL_SOURCE} ${CCF_GENERATED_DIR}/ccf_t.cpp
   )
 
+  target_compile_definitions(ccf.enclave PUBLIC PLATFORM_SGX)
```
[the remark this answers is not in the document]
jumaffre: I added a new `include/pal/platform.h` that allows for easy access to the compile target based on the the compile definition (either `PLATFORM_SGX`, `PLATFORM_SNP` or `PLATFORM_VIRTUAL`). It turned out that `PLATFORM_SGX` wasn't set for `ccf.enclave` when targeting SGX.

src/node/node_state.h
jumaffre: That's a good point. The CCF codebase has a few assumptions about the underlying platform (e.g. little endian) as we're only targeting x86. In this case, the endorsements are a PEM-encoded string so that should be OK?

DomAyre (approved):

PR #4941, opened 2023-02-02
Remove duplicate governance endpoints

### LINKED ISSUE #4687: Remove redundant HTTP APIs
Since we now expose governance tables in a uniform manner from the KV (#4659), we can remove some of the existing endpoints which are basic handwritten wrappers. This should also reduce the scope of #4669.

MahatiC:
Resolves #4687


doc/schemas/gov_openapi.json
```diff
@@ -1995,29 +1942,6 @@
         }
       }
     },
-    "/gov/members": {
```
achamayou, line 1998: @lynshi is that something you are relying on at the moment?

achamayou:
@MahatiC could we list in the changelog the endpoints we remove, and for each one the alternative we recommend?

doc/schemas/app_openapi.json
```diff
@@ -398,30 +398,6 @@
         }
       }
     },
-    "/app/code": {
```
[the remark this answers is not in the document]
lynshi: No - this is an `/app` endpoint. Nothing under `/app` needs to be reviewed, right? This is user-defined

doc/schemas/gov_openapi.json
lynshi: Yes, we need a way to retrieve the list of members so that we can figure out which identity version to use when connecting to the CCF network - this is required for the edge case where we restart before we perform identity rotation. If after some restart, there's a newer Certificate version in AKV, that version may not be registered in CCF yet, so our operator identity needs to know specifically which version to fetch to use to auth with CCF, and it figures this out by comparing available versions what the public keys in CCF.

doc/schemas/node_openapi.json
```diff
@@ -986,30 +986,6 @@
         }
       }
     },
-    "/node/code": {
```
lynshi, line 989: We need this to figure out what code versions are accepted in the CCF network. It's a sanity check prior to node startup to make sure we're starting the right version, and also if we get confused during a code upgrade, it's helpful to know exactly what version(s) we can join to the network.

This may not be the endpoint you want to expose for SEV-SNP since the attestation looks different, so if you want to take it out, you can leave it undocumented (maybe just undocumented in the Azure REST API but still documented here) but available, at least for supporting SGX.

doc/schemas/gov_openapi.json
MahatiC: @lynshi member details can now also be accessed via the newly added endpoints like `/gov/members/certs`, `/gov/members/info`, `/gov/kv/members/encryption_public_keys` [0]. This effort is to remove duplicates and keep the endpoints consistent. For example, `/gov/kv/*` now exposes the governance public tables via endpoints. 

[0] https://github.com/microsoft/CCF/blob/main/doc/schemas/gov_openapi.json#L1558-L1602

doc/schemas/node_openapi.json
MahatiC: Same here, this data will still be accessible by these endpoints https://github.com/microsoft/CCF/blob/main/doc/schemas/gov_openapi.json#L1690-L1734

doc/schemas/gov_openapi.json
lynshi: Good to know! It would still be great to keep these endpoints around (deprecated and undocumented) for now so that code upgrades can go smoothly (since a lot of communication is through a service / load balancer, we might be routed to a node that does not support the endpoint called).

We'd be able to accommodate the eventual removal better this way. If (for example) these endpoints are deprecated in `4.x` and removed in `5.x`, we can code upgrade to `4.x`, update our code to use the new endpoints, and then their removal in `5.x` won't be noticed.

MahatiC: Sure, we will mark them as deprecated for now (and undocumented in the Azure API).

tests/jwt_test.py
```diff
@@ -114,19 +129,19 @@ def test_jwt_without_key_policy(network, args):
         )
 
         with primary.client(network.consortium.get_any_active_member().local_id) as c:
-            r = c.get("/gov/jwt_keys/all")
```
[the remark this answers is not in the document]
lynshi: We do not actively use this endpoint

achamayou (approved):

tests/governance.py
```diff
@@ -290,31 +290,39 @@ def test_member_data(network, args):
 def test_all_members(network, args):
     def run_test_all_members(network):
         primary, _ = network.find_primary()
-
+        network_members = network.get_members()
```
MahatiC, line 293:
```suggestion
```

PR #4942, opened 2023-02-02
Make Attestation Container return UVM_REFERENCE_INFO

takuro-sato:


attestation-container/attestation-container.go
```diff
@@ -11,16 +11,18 @@ import (
...
-	endorsementEnvironmentVariable = flag.String("endorsement-envvar", attest.DEFAULT_ENDORSEMENT_ENVVAR, "Name of environment variable containing report endorsements as base64-encoded JSON object")
-	endorsementServer              = flag.String("endorsement-server", "", "Server to fetch attestation endorsement. If set, endorsement-envvar is ignored. Value is either 'Azure' or 'AMD'")
+	socketAddress                = flag.String("socket-address", "/tmp/attestation-container.sock", "The socket address of Unix domain socket (UDS)")
+	attestationEndorsementEnvVar = flag.String("attestation-endorsement-envvar", attest.DEFAULT_ENDORSEMENT_ENVVAR, "Name of environment variable containing report endorsements as base64-encoded JSON object")
+	attestationEndorsementServer = flag.String("attestation-endorsement-server", "", "Server to fetch attestation endorsement. If set, attestation-endorsement-envvar is ignored. Value is either 'Azure' or 'AMD'")
+	uvmEndorsementEnvVar         = flag.String("uvm-endorsement-envvar", uvm.DEFAULT_UVM_ENDORSEMENT_ENV_VAR_NAME, "Name of UVM endorsement environment variable.")
```
jumaffre, line 25:
```suggestion
	uvmEndorsementEnvVar         = flag.String("uvm-endorsement-envvar", uvm.DEFAULT_UVM_ENDORSEMENT_ENV_VAR_NAME, "Name of UVM endorsement environment variable")
```

attestation-container/protobuf/attestation-container.proto
```diff
@@ -18,4 +18,5 @@ message FetchAttestationRequest {
 message FetchAttestationReply {
   bytes attestation = 1;
   bytes attestation_endorsement_certificates = 2;
+  string uvm_endorsement = 3;
```
jumaffre, line 21: Should this be `bytes` here too, for consistency with the other members? This means we would have to decode the base64-encoded UVM endorsements before returning them.

attestation-container/uvm/uvm_endorsement.go
```diff
@@ -0,0 +1,22 @@
...
+	"fmt"
+	"os"
+)
+
+const DEFAULT_UVM_ENDORSEMENT_ENV_VAR_NAME = "UVM_REFERENCE_INFO"
+
+/*
+Fetch UVM endorsement from environment variable as base64 encoded string.
+*/
+func FetchUVMEndorsement(envVarName string) (string, error) {
```
jumaffre, line 13: "Fetch" feels like the attestation container is fetching this from a remote server, but it's simply looking up an envvar. Should this be `ParseUVMEndorsement` instead, for consistency with the report endorsement?

attestation-container/attestation-container.go
```diff
@@ -40,25 +42,30 @@ func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRe
...
-		endorsement = append(endorsement, endorsementEnvironmentValue.CertificateChain...)
+		attestationEndorsement = append(attestationEndorsement, endorsementEnvironmentValue.VcekCert...)
+		attestationEndorsement = append(attestationEndorsement, endorsementEnvironmentValue.CertificateChain...)
 	}
 
-	return &pb.FetchAttestationReply{Attestation: reportBytes, AttestationEndorsementCertificates: endorsement}, nil
+	uvmEndorsement, err := uvm.FetchUVMEndorsement(*uvmEndorsementEnvVar)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to fetch UVM endorsement: %s", err)
+	}
```
jumaffre, line 61: Similar to `attestationEndorsement`, I think we should parse the environment variable once and store it in a global variable on startup rather than for each request?

attestation-container/protobuf/attestation-container.proto
takuro-sato: Makes sense, added code to decode base64 string and its test

attestation-container/attestation-container.go
takuro-sato: done

achamayou (approved):

.azure-pipelines-templates/deploy_attestation_container.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           python3.8 -m venv ./scripts/azure_deployment/.env
           source ./scripts/azure_deployment/.env/bin/activate
           pip install -r ./scripts/azure_deployment/requirements.txt
-          python3.8 scripts/azure_deployment/arm_template.py remove aci --subscription-id $(CCF_AZURE_SUBSCRIPTION_ID) --resource-group attestationContainer --aci-type dynamic-agent --deployment-name ci-$(Build.BuildNumber)
+          # python3.8 scripts/azure_deployment/arm_template.py remove aci --subscription-id $(CCF_AZURE_SUBSCRIPTION_ID) --resource-group attestationContainer --aci-type dynamic-agent --deployment-name ci-$(Build.BuildNumber)
```
achamayou, line 54: Why is this commented out?

takuro-sato: The CI is failing and I want to see what's happening from azure portal.
It should not be merged

PR #4943, opened 2023-02-02
Spin up CCF Nodes on Secondary ACIs

DomAyre:
Contributes to #4827


tests/infra/remote.py
```diff
@@ -371,7 +371,13 @@ def setup(self, **kwargs):
 
     def get_cmd(self):
         cmd = " ".join(self.cmd)
-        return f"cd {self.root} && {self.env.keys()} {cmd} 1> {self.out} 2> {self.err} 0< /dev/null"
+        env = " && ".join(
+            f'export {key}={value.replace(os.linesep, "")}'
```
jumaffre, line 375: As discussed yesterday, this doesn't work without `export` does it?

tests/infra/remote_shim.py
```diff
@@ -56,6 +56,26 @@ def __init__(self, *args, **kwargs):
...
+        # Bind local RPC address to 0.0.0.0, so that it be can be accessed from outside container
+        for _, rpc_interface in host.rpc_interfaces.items():
+            if aci_ip is None:
+                aci_ip = rpc_interface.host
+            rpc_interface.host = "0.0.0.0"
+            rpc_interface.port = 8001
+            rpc_interface.public_host = aci_ip
+            rpc_interface.public_port = 8001
+
+        # kwargs["include_addresses"] = False
```
jumaffre, line 72:
```suggestion
```

tests/infra/remote_shim.py
```diff
@@ -56,6 +56,26 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+class AciShim(infra.remote.CCFRemote):
+    def __init__(self, *args, host=None, snp_primary_aci_ip="", **kwargs):
```
jumaffre, line 60: `snp_primary_aci_ip` is unused?

tests/infra/node.py
```diff
@@ -150,6 +150,8 @@ def __init__(
 
         if os.getenv("CONTAINER_NODES"):
             self.remote_shim = infra.remote_shim.DockerShim
+        elif isinstance(host, str) and host.split("://")[0] == "aci":
```
jumaffre, line 153: It feels like we should decouple the deployment type (ACI) from the protocol used to drive the node (local or SSH). Would it be better to choose to use the `AciShim` via an environment variable, or a new parameter to `Network`?

tests/code_update.py
```diff
@@ -352,38 +352,78 @@ def test_proposal_invalidation(network, args):
...
-    if args.snp_secondary_acis_path is None:
-        LOG.warning(
-            "Skipping test snp secondary deployment as no target secondary ACIs specified"
-        )
-        return network
-
-    timeout = 60 * 60  # 60 minutes
-    start_time = time.time()
-    end_time = start_time + timeout
+def test_snp_secondary_deployment(_network, args):
```
jumaffre, line 355: Remove `_network` argument here?

tests/code_update.py
```diff
@@ -352,38 +352,78 @@ def test_proposal_invalidation(network, args):
...
+                    LOG.info(
+                        f'Secondary ACI with name "{secondary_name}" has IP: {secondary_ip}'
+                    )
+                    new_node = network.create_node(f"aci://{secondary_ip}")
+                    network.join_node(new_node, args.package, args, timeout=3)
+                    network.trust_node(new_node, args)
+                    LOG.info(
+                        f"Secondary ACI with name {secondary_name} joined the network"
+                    )
+
```
jumaffre, line 424: Can you check here that you can create a client to `new_node` and issue (any) requests to it?

.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -122,6 +123,7 @@ jobs:
                 --count ${{ parameters.secondaries.count }} \
                 --aci-file-share-name ccfcishare \
                 --aci-file-share-account-name ccfcistorage \
+                --ports 22 8000 8001 \
```
jumaffre, line 126: `8000` doesn't seem to be used in the Python infra, which makes me think that new secondary node probably doesn't have have a working node-to-node channel with the primary node. This should be confirmed what calling [`GET /node/commit`](https://microsoft.github.io/CCF/main/operations/operator_rpc_api.html#get--node-commit) on the new node. I expect it to be stuck at 0?

To fix this, the primary node should bind to `0.0.0.0:8000` (published address: `<primary_address>:8000`) and the target address for the new secondary node should be `<primary_address>:8000`. I don't think we have a practical way to set the node-to-node address configuration for nodes just yet, so that'll probably need to be added to the infra.

tests/infra/remote.py
DomAyre: Correct

tests/infra/remote_shim.py
DomAyre: https://github.com/microsoft/CCF/pull/4943/commits/fbe45a81597596398edc34499a3c8867f7e11bf4

tests/infra/node.py
DomAyre: https://github.com/microsoft/CCF/pull/4943/commits/fbe45a81597596398edc34499a3c8867f7e11bf4

tests/infra/remote_shim.py
```diff
@@ -56,6 +56,26 @@ def __init__(self, *args, **kwargs):
...
+        aci_ip = None
+
+        # Bind local RPC address to 0.0.0.0, so that it be can be accessed from outside container
+        for _, rpc_interface in host.rpc_interfaces.items():
+            if aci_ip is None:
+                aci_ip = rpc_interface.host
+            rpc_interface.host = "0.0.0.0"
+            rpc_interface.port = 8001
+            rpc_interface.public_host = aci_ip
+            rpc_interface.public_port = 8001
```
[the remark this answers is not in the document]
DomAyre: https://github.com/microsoft/CCF/pull/4943/commits/fbe45a81597596398edc34499a3c8867f7e11bf4

tests/code_update.py
DomAyre: https://github.com/microsoft/CCF/pull/4943/commits/f35fe064efd1d832d23a64c62da497d3fdc808c4

.azure-pipelines-templates/deploy_aci.yml
DomAyre: Don't we use it in the AciShim?

jumaffre (approved):

PR #4948, opened 2023-02-02
Attestation container: retrieve report endorsements from environment

jumaffre:
Same as https://github.com/microsoft/CCF/pull/4940 but for the attestation container.

TODO:
- [x] Unit test for environment variable
- [x] Pass environment variable to context


attestation-container/attestation-container.go
```diff
@@ -68,6 +76,21 @@ func main() {
...
+		var err error
+		endorsementEnvironment, ok := os.LookupEnv(*endorsementEnvironmentVariable)
+		if !ok {
+			log.Fatalf("Endorsement environment variable %s is not specified (or set --endorsement-server)", *endorsementEnvironmentVariable)
+		}
+
+		*endorsementEnvironmentValue, err = attest.ParseEndorsementACI(endorsementEnvironment)
+		if err != nil {
+			log.Fatal(err)
+		}
```
takuro-sato, line 91: I want to make `main` as small as possible and `attest` package as reusable. (I think `main` package basically should focus only on gRPC stuff.) In that sense how about making these lines a function in `attest` package like `attest.FetchAttestationEndorsementEnvVar(*endorsementEnvironmentVariable)`?

attestation-container/attestation-container.go
```diff
@@ -38,27 +40,33 @@ func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRe
 		return nil, status.Errorf(codes.Internal, "failed to fetch attestation report: %s", err)
 	}
 
-	reportedTCBBytes := reportBytes[attest.REPORTED_TCB_OFFSET : attest.REPORTED_TCB_OFFSET+attest.REPORTED_TCB_SIZE]
-	chipIDBytes := reportBytes[attest.CHIP_ID_OFFSET : attest.CHIP_ID_OFFSET+attest.CHIP_ID_SIZE]
-	endorsement, err := attest.FetchAttestationEndorsement(*endorsementServer, reportedTCBBytes, chipIDBytes)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to fetch attestation endorsement: %s", err)
+	var endorsement []byte
+	if endorsementEnvironmentValue == nil {
```
takuro-sato, line 44: Considering the environment variable's value should be the same in the container's lifetime, caching the env var looks a nice idea. How about caching response from the server then?
If containers lifetime is not considerably long, is it fine to cache it?
If so, we can have `endorsement []byte` global var rather than `endorsementEnvironmentValue` and main would look like:
```go
func main()  {
// same as before

// fetch attestation endorsement
if fetchFromServer { endorsement = fetchFromServerAsBefore() }
else { endorsement = fetchFromEnvVar() }

// same as before
}
```

attestation-container/attestation-container.go
jumaffre: That's what I originally did but then split it out to be able to unit test. I'll move the fetching of the envvar to the `attest` package.

attestation-container/attestation-container.go
takuro-sato: After discussion, we are not 100% sure about the container's lifetime at this moment. We can leave it as it is.

attestation-container/attestation-container.go
takuro-sato: Is it possible to write unit test by using `os.Setenv(envvarName, contentsReadFromFile)` at the begining of the test function?
It may require to separate each test case to different  test function like `TestInvalidJson()` etc though

takuro-sato: something like:
```go
func TestInvalidJson() {
  contentsReadFromFile := readFile()
  contentsReadFromFile = breakJsonFormat(contentsReadFromFile)
  os.Setenv(ENV_VAR_NAME, contentsReadFromFile)
  // do check
}
```

takuro-sato: Change in [072fe0b](https://github.com/microsoft/CCF/pull/4948/commits/072fe0b8759b878b50b6b470d55b92f0155acbe7) looks good to me.

takuro-sato (approved):

PR #4949, opened 2023-02-02
Allow applications to add .well-known endpoints.

### LINKED ISSUE #4810: Applications cannot register endpoints under `.well-known` paths
Since CCF 3.x, applications can register endpoints under any directory, not just `/app`, except directories reserved by CCF (eg. `/gov`, `/node`, ...). We were hoping to use this to register an endpoint at `/.well-known/did.json` in order to give our service a resolvable [`did:web` identitifier](https://w3c-ccg.github.io/did-method-web/).

Unfortunately it seems `.well-known` is one of the reserved prefix, because it is used by the ACME client that is built-in to CCF. The entire directory is reserved, regardless of whether ACME is enabled. The ACME client only needs to register endpoints under the `/.well-known/acme-challenge` sub-directory, but because of CCF's architecture, an "actor" must bind to an entire top-level directory.

https://github.com/microsoft/scitt-ccf-ledger/issues/53#issuecomment-1375348575

plietar:
The ACME frontend was binding to the entire `.well-known` directory, intercepting any requests make to it, including to paths outside of the particular `acme-challenge` sub-directory.

This was preventing applications from handling any paths inside the `.well-known`, which may be necessary to implement certain protocols, such as did:web.

The code to extract the actor name from an incoming request is special cased for the .well-known directory, in which case it will include the second path component in the actor's name. THe ACME frontend's actor is now the full `.well-known/acme-challenge`.

We did consider implementing this by instead moving the ACME endpoints into the CommonEndpointRegistry, indirectly making them part of the application frontend, and removing the .well-known actor. Unfortunately this would make the endpoints accessible only once the application frontend has been opened, which would prevent a service from obtaining an ACME certificate early in its lifecycle.

Fixes #4810

### DIFF
--- a/src/http/http_rpc_context.h
+++ b/src/http/http_rpc_context.h
@@ -305,8 +305,25 @@ namespace http
       return std::nullopt;
     }
 
-    const auto actor = path.substr(first_slash + 1, second_slash - 1);
-    const auto remaining_path = path.substr(second_slash);
+    auto actor = path.substr(first_slash + 1, second_slash - first_slash - 1);
+    auto remaining_path = path.substr(second_slash);
+
+    // The "actor" is generally just a single path component, eg. `gov` or
+    // `node`.  We make an exception for .well-known paths however, which use
+    // two components, ie. `.well-known/acme-challenge`. This restricts the
+    // ACME frontend to just that particular sub-directory, and allows the
+    // application to handle other .well-known paths.
+    if (actor == ".well-known")
+    {
+      const auto third_slash = path.find_first_of('/', second_slash + 1);
+      if (third_slash == std::string::npos)
+      {
+        return std::nullopt;
+      }
+
+      actor = path.substr(first_slash + 1, third_slash - first_slash - 1);
+      remaining_path = path.substr(third_slash);
+    }
 
     if (actor.empty() || remaining_path.empty())
     {
@@ -317,10 +334,6 @@ namespace http
     if (ccf::is_valid_actor(actor))
     {
       ctx.set_method(remaining_path);
-    }
-    else
-    {
-      ctx.set_method(path);
     }
     return actor;
   }
--- a/include/ccf/actors.h
+++ b/include/ccf/actors.h
@@ -12,7 +12,7 @@ namespace ccf
     members = 0,
     users,
     nodes,
-    well_known,
+    acme_challenge,
     // not to be used
     unknown
   };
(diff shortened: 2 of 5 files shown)


wintersteiger (approved):
Thanks, this LGTM. I think this is the right tweak for now, until the gRPC infrastructure improvements will likely take all of this apart.

src/http/http_rpc_context.h
```diff
@@ -318,10 +335,6 @@ namespace http
     {
       ctx.set_method(remaining_path);
     }
-    else
-    {
-      ctx.set_method(path);
-    }
```
MahatiC, line 324: Why is this removed? This is to detect all other paths that don't have an `actor` i.e. this enables the removal of `app` actor. You should add the new `acme_challenge` actor here https://github.com/microsoft/CCF/blob/main/include/ccf/actors.h#L20

MahatiC (changes requested):

src/http/http_rpc_context.h
plietar: Unless I am missing something, it seemed to me this branch of the `else` is a no-op, hence why I removed it.

`path` is defined at the top of the function as:
```
const auto path = ctx.get_method();
```
and neither `path` nor ctx's method get modified until this point, so the `else` block was simply re-assigning the same path value back into ctx's method.

There are other early returns in this function that don't do this assignement, and they work fine. I actually found the disparity a little confusing when I first looked at that code.

MahatiC: Ah, yes `path` was never reassigned (and I see that `actors.h` is also updated). Looks good to me.

MahatiC (approved):

plietar:
Would it be okay for this to be backported to 3.x?

achamayou:
> Unfortunately this would make the endpoints accessible only once the application frontend has been opened, which would prevent a service from obtaining an ACME certificate early in its lifecycle.

@wintersteiger does the ACME endpoint need KV access? If not, the node frontend is always opened, and the ACME endpoint would be accessible through there.

plietar:
@achamayou the issue with the node frontend is that it prefixes everything with `/node/`, whereas the `.well-known/acme-challenge` directory must be at the root. The reason I tried to use the application frontend is because it now is accessible without any prefix.

achamayou (approved):

PR #4950, opened 2023-02-02
Distinguish between system and user include search paths in add_ccf_app.

plietar:
The add_ccf_app CMake macro had a INCLUDE_DIRS arguments, which can be used to add directories to the compiler's header search path. To my surprise, all paths in this argument are added as "system" headers, which mean they get passed as `-isystem` flags to the compiler.

This has the effect of silencing any compiler warnings emitted in the headers found through these path. While this is the intended behaviour for 3rd party dependencies, it is undesirable for code that is part of the application, which should adhere to the configured warning settings.

This has bit us in SCITT, where for a long time some warnings had been silenced due to this and had gone unoticed, only to create a lot of confusion later on when this code would suddenly fail to compile when moved elsewhere. See https://github.com/microsoft/scitt-ccf-ledger/pull/86 for some context.


achamayou (approved):

achamayou:
@plietar could you run the formatter please?

achamayou (approved):

PR #4955, opened 2023-02-03
Attestation container: reduce size of container image

jumaffre:
The attestation container image was previously based on `golang:1.19`. However, we only need `golang` to build the attestation container library which can be deployed in the `scratch` image when built as a statically-linked binary. 

If this works in the CI, this would take the container image from `1.23GB` to `18.2MB`, which speeds up deployment and removes most dependencies.

As discussed with @takuro-sato, this is the plan:
- [x] Use `mcr.microsoft.com/cbl-mariner/distroless/minimal:2.0` or similar for the runtime container
- [x] Use `mcr.microsoft.com/oss/go/microsoft/golang:1.20-fips-cbl-mariner2.0` to build the image 
- [x] Run unit tests as part of "command" rather than SSH-ing into the container later
- [x] Run end-to-end tests from dummy business logic container
- [x] Remove open ports on attestation deployment


scripts/azure_deployment/arm_aci.py
```diff
@@ -349,7 +341,7 @@ def make_aci_deployment(args: Namespace) -> Deployment:
                 ),
                 make_dummy_business_logic_container(
                     container_name_dummy_blc,
-                    container_image,
+                    "ubuntu:latest",
```
jumaffre:
```suggestion
                    "mcr.microsoft.com/cbl-mariner/distroless/minimal:2.0",
```

attestation-container/Dockerfile
```diff
@@ -1,10 +1,11 @@
 # Based on https://hub.docker.com/_/golang/
 
 # To run this:
-#   docker build -t attestation-container .
+#   docker build -t attestation-container . [--build-arg build_test=True]
```
takuro-sato, line 4: Thanks. I forgot to update this last time

scripts/azure_deployment/arm_aci.py
```diff
@@ -414,14 +399,26 @@ def make_aci_deployment(args: Namespace) -> Deployment:
...
+                        "sudo",
+                        "az",
+                        "confcom",
+                        "acipolicygen",
+                        "-a",
+                        arm_template_path,
+                        "--pretty-print",
+                        "--outraw-pretty-print",
+                        "--save-to-file",
+                        output_policy_path,
```
takuro-sato, line 420: It's getting error in the CI
```
ERROR: unrecognized arguments: --pretty-print
```
I only tried `--pretty-print --outraw-pretty-print` without `--save-to-file`. So we might have to remove the option and get the result from STDOUT.

PR #4962, opened 2023-02-06
Run 'SNPCC Release' and 'Virtual Release' in the daily CI

takuro-sato:
In the attempt of releasing ccf-4.0.0-dev4 (the tag is already deleted), we found that `SNPCC Release` and `Virtual Release` are broken. https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=63379&view=logs&j=59948429-6ed2-59c7-eef3-dd8a86e66ec1
Currently it can be only found by actually pushing a tag to GitHub to trigger `CCF Github CI`, because they are not run in `CCF GitHub Daily` daily CI (`SGX Release` is run though).

This PR adds `SNPCC Release` and `Virtual Release` to the daily CI.

Job/Pipeline|CCF Github CI (release)|CCF GitHub Daily (daily)
---|---|---
SGX Release|✅|✅
SNPCC Release|✅|- (Will be added in this PR)
Virtual Release|✅|- (Will be added in this PR)

The existing `SGX Release` in `CCF Github CI` (which is run when pusing a tag) and in `CCF GitHub Daily` are bit different.
Only `CCF GitHub Daily` has additional cmake args `-DENABLE_2TX_RECONFIG=ON -DSHUFFLE_SUITE=ON` and runs test `rotation`.
The newly added `SNPCC Release` and `Virtual Release` have the same difference.

## Other changes
- Fix errors in the new CI jobs (handling clang12 and clang10 correctly)


DomAyre:
SNPCC Is the old statically allocated, self-hosted pool based SNP testing, and shouldn't exist anymore as a pipeline, where are you seeing these failures?
Screenshot 2023-02-06 at 13 28 56

takuro-sato:
It's in the link in the description

takuro-sato:
Is it fine to delete `SNPCC` from `CCF Github CI`?  (I am wondering if we need to add an alternative job instead)

DomAyre:
It should have been removed along with the daily testing as of 9bd375cb6e064b23270de9c735dee71c72a4a33c but I must have missed it.

My perspective would be to only run special release testing when that testing is too expensive to run regularly, so in theory we have coverage from normal regular SNP testing. That being said if we want a single run to tie to the release, I would just add triggers to the existing pipelines so we run SNP testing when a release is cut.

But I'm open to suggestions

takuro-sato:
After checking more carefully, I found that CTest is not run in only SNPCC anyway (I guess it's compromise for now though). The code is [here](https://github.com/microsoft/CCF/blob/main/.azure-pipelines-templates/common.yml#L39).
So as long as we can create .deb and build the sample in normal machine (`1es-dv4-focal`). I think it should be fine.

jumaffre:
I think the issue is here: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=63379&view=logs&j=77890be3-fbd3-52a1-2c40-e5bc14cbbdef&t=0659858c-0d0d-58c6-5ce3-d0b2c9c79269&l=19, which was introduced in https://github.com/microsoft/CCF/pull/4926/files. The fix would be to conditionally change [this line](https://github.com/microsoft/CCF/blob/main/tests/test_install_build.sh#L8) to `clang-12` if clang-12 is available.

takuro-sato:
Interesting, it looks like the installing the .deb removes clang12 and then installs clang10.
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=63449&view=logs&j=77890be3-fbd3-52a1-2c40-e5bc14cbbdef&t=5f8323f3-0deb-575a-0a7e-d7c62645511f&l=23

jumaffre:
> Interesting, it looks like the installing the .deb removes clang12 and then installs clang10. https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=63449&view=logs&j=77890be3-fbd3-52a1-2c40-e5bc14cbbdef&t=5f8323f3-0deb-575a-0a7e-d7c62645511f&l=23

This should have been updated too, depending on the value of `COMPILE_TARGET`: https://github.com/microsoft/CCF/blob/main/cmake/cpack_settings.cmake#L21-L23

.azure-pipelines-templates/install_deb.yml
```diff
@@ -60,6 +64,8 @@ steps:
   - ${{ if eq(parameters.target, 'SNPCC') }}:
       - script: |
           set -ex
+          export CC=$(command -v clang-12)
+          export CXX=$(command -v clang++-12)
```
jumaffre, line 68: I think it'd be nicer to have this logic in `test_install_build.sh` so that the script is self-contained and doesn't depend on external external environment.

cmake/cpack_settings.cmake
```diff
@@ -18,9 +18,15 @@ endif()
...
-)
+if(COMPILE_TARGET STREQUAL "sgx")
+  set(CCF_DEB_BASE_DEPENDENCIES
+      "libuv1 (>= 1.34.2);libc++1-10;libc++abi1-10;openssl (>=1.1.1)"
+  )
+else()
+  set(CCF_DEB_BASE_DEPENDENCIES
+      "libuv1 (>= 1.34.2);libc++1-12;libc++abi1-12;openssl (>=1.1.1)"
+  )
+endif()
```
jumaffre, line 29: Should this be folded into the existing logic on line `33`, and `CCF_DEB_BASE_DEPENDENCIES` becomes `libuv1 (>= 1.34.2);openssl (>=1.1.1)`.

.azure-pipelines-templates/daily-matrix.yml
```diff
@@ -6,6 +6,8 @@ parameters:
     SGX:
       container: sgx
       pool: 1es-DC16s_v3-focal
+    SNPCC:
+      pool: sev-snp-pool
```
jumaffre, line 10: Changes in this file should be reverted?

.azure-pipelines-templates/install_deb.yml
takuro-sato: Is it something like `TYPE=sgx ./test_install_build.sh <cmake args>` ?
I can't add additional argument because arguments are passed to cmake as `$@`.

jumaffre: Or just trying to detect whether `clang-12` exists and if it does, use it. Otherwise `clang-10`?

.azure-pipelines-templates/daily-matrix.yml
takuro-sato: I think it's still necessary to create .deb for SNP. Currently .deb is created only when a release is actually created, and it could potentially prevent us from make releases quickly enough.
Hopefully I can leave changes in the file but use other pool like `1es-dv4-focal`

takuro-sato: But if it doesn't work, I think it would be reasonable to revert code about SNP and leave Virtual as it is.

takuro-sato: Looks like `1es-dv4-focal` works for snp as well (because CTest was removed in past for snp).

jumaffre: `1es-dv4-focal` is good here. It's OK to build the SNP release on a virtual machine.

jumaffre (approved):

PR #4968, opened 2023-02-07
Switch to West Europe for ACI deployments

DomAyre:
Eastus2euap seems too unstable at the moment, so switching to West Europe


DomAyre:
This would require changes to our infra to handle the lack of envvars for security policy and UVM endorsements. So we don't think spending engineering time to take a step back is a good trade for the added stability

PR #4970, opened 2023-02-07
Make secondary ACI cleanup independent of primary

DomAyre:
Even if primary ACI deployment failed, secondary ACI deployment may have succeeded so attempt to cleanup regardless of the state of the primary

### DIFF
--- a/.azure-pipelines-templates/deploy_aci.yml
+++ b/.azure-pipelines-templates/deploy_aci.yml
@@ -161,8 +161,8 @@ jobs:
               done
             displayName: "Connect Primary ACI with Secondaries"
 
-  - job: cleanup_workspace
-    displayName: "Cleanup Workspace"
+  - job: cleanup_aci
+    displayName: "Cleanup ACI"
     pool:
       vmImage: ubuntu-20.04
     dependsOn:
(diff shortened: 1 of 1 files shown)


.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -211,9 +201,15 @@ jobs:
           python3.8 -m venv ./scripts/azure_deployment/.env
           source ./scripts/azure_deployment/.env/bin/activate
           pip install -r ./scripts/azure_deployment/requirements.txt
-          python3.8 scripts/azure_deployment/arm_template.py remove aci --subscription-id $(CCF_AZURE_SUBSCRIPTION_ID) --resource-group ccf-aci --aci-type dynamic-agent --deployment-name ci-$(Build.BuildNumber)
+          python3.8 scripts/azure_deployment/arm_template.py remove aci --subscription-id $(CCF_AZURE_SUBSCRIPTION_ID) --resource-group ccf-aci --aci-type dynamic-agent --deployment-name ci-$(Build.BuildNumber) || echo "Primary cleanup failed"
```
takuro-sato, line 204: Do we need to add `|| echo "..."` while there is `continueOnError: true`?

DomAyre: Good spot, it was left over from before I split the steps

takuro-sato (approved):

PR #4974, opened 2023-02-08
Add option to use node client certificates to ACME HTTP interface

wintersteiger:


src/node/node_state.h
```diff
@@ -2638,11 +2638,22 @@ namespace ccf
...
       const std::vector<std::string>& ca_certs = {},
-      ccf::ApplicationProtocol app_protocol =
-        ccf::ApplicationProtocol::HTTP1) override
+      ccf::ApplicationProtocol app_protocol = ccf::ApplicationProtocol::HTTP1,
+      bool use_node_client_certificate = false) override
     {
       auto ca = std::make_shared<tls::CA>(ca_certs, true);
-      auto ca_cert = std::make_shared<tls::Cert>(ca);
+      std::shared_ptr<tls::Cert> ca_cert;
+      if (!use_node_client_certificate)
```
achamayou, line 2646: I am confused by this change, this seems to use a specific node client cert, rather than the service cert, as CA. What is the motivation for this? This will cause redirects to fail for example.

Unrelated, and somewhat nitpicky, but I think and if(condition){} else {} is clearer than if(!condition) {} else {}.

wintersteiger: Yes, that's what I intended, but it may well be the wrong choice. The motivation is to restrict self-rpcs with `NodeCertAuthnPolicy`, so I figured I'd use the node cert. I think you may be right that this could break forwarding though.

src/node/node_state.h
```diff
@@ -2638,11 +2638,22 @@ namespace ccf
         bool(http_status status, http::HeaderMap&&, std::vector<uint8_t>&&)>
         callback,
       const std::vector<std::string>& ca_certs = {},
-      ccf::ApplicationProtocol app_protocol =
-        ccf::ApplicationProtocol::HTTP1) override
+      ccf::ApplicationProtocol app_protocol = ccf::ApplicationProtocol::HTTP1,
+      bool use_node_client_certificate = false) override
```
achamayou, line 2642:
```suggestion
      bool authenticate_as_node_client_certificate = false) override
```

src/node/rpc/acme_subsystem.h
```diff
@@ -63,8 +63,8 @@ namespace ccf
         const http::HeaderMap&,
         const std::vector<uint8_t>&)> callback,
       const std::vector<std::string>& ca_certs = {},
-      ccf::ApplicationProtocol app_protocol =
-        ccf::ApplicationProtocol::HTTP1) override
+      ccf::ApplicationProtocol app_protocol = ccf::ApplicationProtocol::HTTP1,
+      bool use_node_client_certificate = false) override
```
achamayou, line 67:
```suggestion
      bool authenticate_as_node_client_certificate = false) override
```

src/node/node_state.h
achamayou: Thank you for the offline explanation! :)

achamayou (approved):

Issue #4912, opened 2023-01-27
Multi-container: Indexing: Remaining tasks


MahatiC:
This is moved from this issue https://github.com/microsoft/CCF/issues/4672

- [x] Separate out the data structure fields from `InstallAndSubscribe` and move it to `StoreIndexedData` as it's more cleaner and maps more closely with the API semantics
- [x] Clean-up left over objects (i.e. indexing strategies that are installed via `InstallAndSubscibe` but will no longer be used since there was a connection drop
- [x] Modify `LRUIndexCache` to support LRU semantics on `Get` operation as well. Currently it's only supported on `Insert` operations
- [x] Add additional tests and address review comments (https://github.com/microsoft/CCF/pull/4820#discussion_r1088014248)

MahatiC:
https://github.com/microsoft/CCF/pull/4978 PR covers all tasks listed here. An end to end test is a nice-to-have and can be a part of https://github.com/microsoft/CCF/issues/4911

PR #4978, opened 2023-02-09
Multi-container: Index improvements

MahatiC:
Part of #4912
- [x] Clean-up proto file
- [x] Test the modified API

Changes in this PR:

- Separate Index `Subscribe` with Index Data Structure choice. Data Structure is now a part of `StoreIndexedData` call
- Remove strategy from stored indexes when connection drops and reset the pointer for later destruction
- Modify `Get` operation in LRU list
- Separate indexer test so that it can be containerised and tested independently
- Misc naming changes


src/apps/external_executor/external_executor.cpp
```diff
@@ -264,16 +268,20 @@ namespace externalexecutor
...
         -> ccf::grpc::GrpcAdapterResponse<google::protobuf::Empty> {
-        auto it = map_index_strategies.find(payload.strategy_name());
-        if (it != map_index_strategies.end())
+        auto executor_id = get_caller_executor_id(ctx);
+        auto it = index_streams.find(executor_id);
+
+        if (it == index_streams.end())
         {
-          it->second->is_indexer_active = false;
+          LOG_INFO_FMT("Strategy NOT FOUND!");
```
achamayou, line 276:
```suggestion
          LOG_INFO_FMT("No such strategy: {}", payload.strategy_name());
```

Not sure about INFO here, but I think we want to avoid all caps and exclamation marks in log messages.

src/apps/external_executor/external_executor.cpp
```diff
@@ -264,16 +268,20 @@ namespace externalexecutor
...
+          return ccf::grpc::make_error(
+            GRPC_STATUS_NOT_FOUND,
+            fmt::format("Index {} is not found", payload.strategy_name()));
```
achamayou, line 279:
```suggestion
            fmt::format("Indexing strategy {} not found", payload.strategy_name()));
```

src/apps/external_executor/external_executor_indexing.cpp
```diff
@@ -76,49 +76,41 @@ namespace externalexecutor
...
+  ExecutorStrategy::ExecutorStrategy(
+    const std::string& map_name_, ExecutorId& id) :
+    Strategy(id),
+    map_name(map_name_),
+    indexer_id(id)
+  {
+    if (kv::get_security_domain(map_name_) != kv::SecurityDomain::PUBLIC)
+    {
+      throw std::logic_error(fmt::format(
+        "This Strategy ({}) is currently only implemented for public tables, "
```
achamayou, line 106:
```suggestion
        "ExecutorStrategy ({}) is currently only implemented for public tables, "
```

src/apps/external_executor/external_executor.cpp
MahatiC: Sorry, that was for my own logging, so that it's easy to search :)

MahatiC: I removed this log statement altogether (we return an error anyway)

PR #4986, opened 2023-02-09
Capped queue for node-to-node forwarding, and live key exchange

### LINKED ISSUE #4625: Node-to-node channel single-message queue is insufficient for forwarded messages
Bug spotted from this run:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=55710&view=logs&j=8f3dc89c-3708-5926-47e7-27120a268dab&t=3aae4c9d-f0dc-5b97-1b0e-f86312395af9&l=1365

The bad line is this, where a primary drops a forwarded response in favour of a consensus message:

```
2022-11-23T14:33:34.541627Z -0.002 0   [debug] ../src/node/channels.h:852           | Dropping outgoing message of type 2 - replaced by new outgoing send of type 1
```

Our node-to-node channels only store a single pending message while a channel is being established. This was done to avoid an infinitely growing queue, on the assumption that all users of the node-to-node channels could politely handle transport loss and would retry with an at-least-as-good message shortly. That's true for consensus, but not for forwarding.

We should instead queue some more (but capped) messages during channel establishment, and return a `false` only if the message couldn't be queued. The caller can then react to this more precisely, likely with a log. For forwarded responses its equivalent to a network failure (but with a helpful log of exactly what happened), and for forwarded requests we can produce a direct user-facing error.

Note that we were re-establishing a channel due to the per-channel message limit:

```
2022-11-23T14:33:34.507856Z -0.002 0   [trace] ../src/node/channels.h:258           | <- n[1826524727a17d53cd89026e8f5132b91334efe39dcec5d825f475b95a2b0ad6] (ESTABLISHED): Reached message limit (1036+1012 >= 2048), triggering new key exchange
```

This is much lower on debug builds, hence the failure only occurring there.

eddyashton:
Resolves #4625.

There are 2 significant semantic changes here:
- Queue multiple forwarding messages on an unestablished channel, rather than a single message. This means forwarded requests and responses should not be overwritten by consensus messages.
- Trigger key exchange at half the cap, and allow previous key to be used while new key is negotiated.

The former is not a complete fix by itself, because we still have significant loss in _receiving_ messages when the nodes disagree even slightly on their nonce count (ie - a node sends a consensus message and then a forwarded message, the receiver processes the consensus and decides this has hit the message cap => instantly drops keys to re-establish channel, losing the forwarded message). Hence the latter, more involved and much more risky, change.

This means that the `state` of each channel is no longer authoritative - it may have `send_` and `recv_` keys independently of its key exchange progress. New tests attempt to cover these cases.


src/node/channels.h
```diff
@@ -117,19 +92,75 @@ FMT_END_NAMESPACE
...
+
+  // Key exchange states are:
+  // - Have nothing
+  // - Initiated (have my own share)
+  // - Have their share and my share (received init, or received response)
+  // - => Have shared secret
+  // - Know that THEY have shared secret (received response or final)
+  // As soon as we have both shares, we update our send key
+  // As soon as we know that they have shared secret, we update our recv key
+  // Note this assumes they the key exchange messages are reliably delivered,
```
achamayou, line 158:
```suggestion
  // Note this assumes that the key exchange messages are reliably delivered,
```

achamayou (approved):

jumaffre (approved):

PR #4989, opened 2023-02-10
Fix forwarding for readonly endpoints in JS/TS samples

### LINKED ISSUE #4924: Correct values of `forwarding_required` in JS sample apps
We use `"forwarding_required": "always"` for a lot of endpoints in our JS sample apps, that should really be `"forwarding_required": "sometimes"`. We should do a pass updating these, and confirming the forwarding works as expected.

achamayou:
Fix #4924


eddyashton (approved):

doc/build_apps/js_app_bundle.rst
```diff
@@ -87,6 +87,8 @@ Each endpoint object contains the following information:
   - ``"readonly"``
   - ``"historical"``
 
+.. note:: "sometimes" is a good default value for most endpoints. The node that receives the request will forward only to preserve session consistency (a previous transaction was already forwarded), or because the transaction can't be executed locally (it involves a write, and the node is a backup). "always" is a good setting for endpoints that always write to the KV, because it saves attempting the transaction on a backup before forwarding.
```
jumaffre, line 90:
```suggestion
.. note:: "sometimes" is a good default value for most endpoints. The node that receives the request will forward only to preserve session consistency (a previous transaction was already forwarded), or because the transaction cannot be executed locally (it involves a write, and the node is a backup). "always" is a good setting for endpoints that always write to the KV, because it saves attempting the transaction on a backup before forwarding.
```

jumaffre (approved):

PR #5004, opened 2023-02-14
Upgrade to Open Enclave 0.18.5

achamayou:
In an attempt to make this a single PR, I am doing this on a branch, so as to be able to cut the tag that kicks off the CI container building and publishing without having to merge anything first. This should make backporting easier.

- [x] Cut `ccf_ci_image/14-02-2023` tag
- [x] [Build](https://github.com/microsoft/CCF/actions/runs/4174305638/jobs/7227698949) CI images
- [x] Update pipelines

PR #5010, opened 2023-02-14
[release/3.x] Cherry pick: Upgrade to Open Enclave 0.18.5 (#5004)

achamayou:
Backports the following commits to `release/3.x`:
 - [Upgrade to Open Enclave 0.18.5 (#5004)](https://github.com/microsoft/CCF/pull/5004)


achamayou:
The virtual build does not work with images build from main, because of the switch to clang 12. It is necessary to either:

- Downgrade virtual images to clang-10 on main and rebuild them
- Backport the clang-12 virtual upgrade changes to release/3.x

For anyone wondering how release/2.x works: only the sgx image is used there, because nosgx builds still require Open Enclave.

I am leaning towards the first option, since clang-12 seemed to be suffering the same issue as clang-10 with side channel mitigations on SNP builds, and so we get little benefit from the upgrade at this point. The issue may crop up again with future upgrades though, and so we may want to consider a third option where virtual CI containers are produced in clang-10 and clang-1x variants.

jumaffre (approved):

PR #5015, opened 2023-02-15
Fix `clang-version` in playbook

MahatiC:
`ccf-dev.yml` playbook fails due to undefined `clang_version` variable


achamayou (approved):
@MahatiC thank you, sorry for missing that. It's defaulted in the Dockerfile, but not in the playbook indeed.

Issue #4979, opened 2023-02-09
Multi-Container: Build infra to run the executors in docker containers


MahatiC:
As a stopgap solution (for #4261) until we have a stable ACI/confidential pods, build out infra to spin up external executors inside docker containers.

Todo:
- [ ] #5018 (WikiCacherExecutor)
- [ ] Update to cover the LoggingExecutor
- [ ] Centralise framework for use in other tests

jumaffre:
https://github.com/microsoft/CCF/blob/main/tests/infra/remote_shim.py#L62 may serve as inspiration to get this working both locally and in docker-out-of-docker in the CI, and can also be used to run the CCF node itself within a container (which may help for BL container to CCF node communication).

achamayou:
Not under active consideration, as work on external executors is suspended.

PR #5018, opened 2023-02-15
Containerise external executor tests

DomAyre:
Contributes to #4979 

Currently only works with WikiCacherExecutor, but I think this could be merged on its own and any 'framework' type changes and updating the other tests which use the logging app can be in a follow up PR


DomAyre:
Doesn't currently work with the logging app for an unknown reason

tests/external_executor/executors/util.py
```diff
@@ -33,8 +53,78 @@ def terminate(self):
...
+    supported_endpoints: Set[Tuple[str, str]],
+):
+    ec = ExecutorContainer(
+        executor,
+        node_public_rpc_address,
+        network,
+        supported_endpoints,
+    )
+    ec.start()
+    yield "Something"
```
MahatiC, line 129: `yield ec`?

DomAyre: Good spot, we actually don't need to yield anything but "Something" was a placeholder so I'll fix that

DomAyre: https://github.com/microsoft/CCF/pull/5018/commits/c636212e83682bbdbf46418855f1c5ba3a52de53

tests/external_executor/executors/wiki_cacher.py
```diff
@@ -115,7 +116,8 @@ def run_loop(self, activated_event):
 
             for work in stub.Activate(Empty()):
                 if work.HasField("activated"):
-                    activated_event.set()
+                    if activated_event is not None:
```
jumaffre, line 119: I wonder if we should remove this event altogether and rely on an informational endpoint on the CCF node to list the external executors currently registered on that node? For now, we can simply wait for the registered endpoints to not return a 404? @DomAyre @eddyashton @MahatiC what do you think?

tests/external_executor/executors/util.py
```diff
@@ -33,8 +53,78 @@ def terminate(self):
...
+    def __init__(
+        self,
+        executor: str,
+        node_public_rpc_address: str,
+        network: Network,
+        supported_endpoints: Set[Tuple[str, str]],
+    ):
+        self._client = docker.DockerClient()
+
+        image_name = "ccfmsrc.azurecr.io/ccf/ci:14-02-2023-virtual"
```
jumaffre, line 73: I think this could/should be a smaller image to mimic a real scenario, e.g. `mcr.microsoft.com/cbl-mariner/base/python:3`?

tests/external_executor/executors/util.py
```diff
@@ -33,8 +53,78 @@ def terminate(self):
...
+        )
+
+        LOG.info("Connecting container to network")
+        self._network = self._client.networks.get("ccf_test_docker_network")
+        self._network.connect(self._container)
+
+    def start(self):
+        LOG.info("Starting container...")
+        self._container.start()
+        time.sleep(2)
```
jumaffre, line 106: To be removed?

tests/external_executor/executors/util.py
```diff
@@ -33,8 +53,78 @@ def terminate(self):
...
+    ):
+        self._client = docker.DockerClient()
+
+        image_name = "ccfmsrc.azurecr.io/ccf/ci:14-02-2023-virtual"
+        LOG.info(f"Pulling image {image_name}")
+        self._client.images.pull(image_name)
+
+        # Create a container with external executor code loaded in a volume and
+        # a command to run the executor
+        command = "/workspaces/CCF/build/env/bin/python3"
```
jumaffre, line 79: We should avoid hardcoding paths here as much as possible, and instead rely on the `e2e_args.workspace` variable if we can. We should also only mount what's required for the external executor to run, which is `wiki_cacher.py`, plus the generated `_pb2.files` and a `requirements.txt` file containing `loguru` and `grpc`?

tests/external_executor/external_executor.py
```diff
@@ -145,17 +153,29 @@ def test_executor_registration(network, args):
 def test_simple_executor(network, args):
     primary, _ = network.find_primary()
 
-    wikicacher_executor = WikiCacherExecutor(primary.get_public_rpc_address())
-    supported_endpoints = wikicacher_executor.get_supported_endpoints({"Earth"})
+    if os.environ.get("CONTAINER_NODES"):
```
jumaffre, line 156: I don't think we currently run the external executor e2e test with this environment variable set anywhere do we? I feel like we could always run the wiki cacher container in its own container and leave the other test in a threaded executor for now?

tests/external_executor/executors/wiki_cacher.py
eddyashton: An endpoint that lists the active executors could also work, but it involves another polling loop. I think maintaining this push/event style notification is nice, even if we decide our sample infra is polling rather than using it? Its plausible that an executor will want to do some set up at exactly this point (after it has activated, before it processes any user requests).

Waiting for the registered endpoint to not return a 404 is less precise - when we have multiple executors serving the same URI this won't work.

tests/external_executor/executors/util.py
```diff
@@ -33,8 +53,78 @@ def terminate(self):
...
-def executor_thread(executor):
+def executor_thread(executor: Executor):
     et = ExecutorThread(executor)
     et.start()
     yield executor
     et.terminate()
+
+
+class ExecutorContainer:
+    def __init__(
```
achamayou, line 64: Let's not have files called util, and mix code meant to be used inside the executor, and outside to orchestrate it, in the same file.

tests/external_executor/executors/wiki_cacher.py
achamayou: We need to expose from CCF that an executor is ready to handle requests, even for operation purposes, as well as testing.

tests/external_executor/run_executor.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+from argparse import ArgumentParser
+import sys
+
+sys.path.append("/workspaces/CCF/tests")
```
achamayou, line 6: Let's not hardcode that.

tests/external_executor/executors/wiki_cacher.py
DomAyre: Done it by polling the registered executor endpoints for now https://github.com/microsoft/CCF/pull/5018/commits/90ae68730b9d7c2d62236d3188a7098bcedcfc82 but I agree that a proper endpoint to check for registered executors should follow this PR

tests/external_executor/executors/util.py
DomAyre: Removed as we now poll for registered endpoints as of https://github.com/microsoft/CCF/pull/5018/commits/90ae68730b9d7c2d62236d3188a7098bcedcfc82

tests/external_executor/external_executor.py
DomAyre: Split the tests into ones which always run containerised (currently just test_simple/wiki_cacher_executor) and tests which will eventually be containerised in following PRs https://github.com/microsoft/CCF/pull/5018/commits/83c77cbcb53b48a6634df45dd89987fe4de56663

tests/external_executor/executors/util.py
DomAyre: Moved the code around in https://github.com/microsoft/CCF/pull/5018/commits/682a90dec94d35e41d8a4a28d63755c89593367b and https://github.com/microsoft/CCF/pull/5018/commits/ad50472969d403ae2a4150e8e61cfd05803a66fb

tests/external_executor/run_executor.py
DomAyre: https://github.com/microsoft/CCF/pull/5018/commits/d1f01bc62346de78f95924b0af3b0c1bbe06d236

tests/external_executor/executors/util.py
DomAyre: https://github.com/microsoft/CCF/pull/5018/commits/d1f01bc62346de78f95924b0af3b0c1bbe06d236

Just loads the test and infra code for now. Will do a followup PR to change the image to use something lighter which will then require the requirements file

tests/external_executor/executors/util.py
DomAyre: Will do in a follow up PR

tests/external_executor/executors/wiki_cacher.py
achamayou: @DomAyre may be worth ticketing so we don't forget :)

achamayou (approved):

DomAyre:
I ended up adding the python image change since we had issues using the venv python in CI so we needed to use the system python anyway which required the requirements.txt file, at which point which image we used was a trivial change

jumaffre (approved):

PR #5021, opened 2023-02-16
Stop waiting if the `pem` file is fetched successfully

MahatiC:


tests/infra/node.py
```diff
@@ -352,6 +352,8 @@ def _start(
                     )
             except ValueError as ve:
                 LOG.info(f"Failed to parse node certificate file ({pem_path}) : {ve}")
+            else:
+                break
```
eddyashton, line 356:
```suggestion
                    )
                    break
            except ValueError as ve:
                LOG.info(f"Failed to parse node certificate file ({pem_path}) : {ve}")
```

This seems simpler, but equivalent?

MahatiC: No preference for me really, but used this one.

achamayou (approved):

Issue #4540, opened 2022-11-14
Investigate validating SNP Attestation's Chip ID


DomAyre:
Should be an option only (can be useful for dev work to have nodes on the same box)

Store used chip IDs in the KV, check new node's ID are not yet present

Issue #4538, opened 2022-11-14
Verify SNP attestation's reported TCB


DomAyre:
Store minimum acceptable version in the KV and check new nodes against it

PR #5024, opened 2023-02-17
SEV-SNP: Tighten attestation verification

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/4538
Resolves https://github.com/microsoft/CCF/issues/4540 

This PR tightens the verification of SEV-SNP attestation reports by verifying that some of the fields in the report are as expected and that the TCB contained in the endorsements retrieved from ACI matches the reported TCB in the attestation report. Note that the endorsed TCB (a short hex-encoded string, e.g. `"0300000000000873"`) is now also recorded in the nodes table, mostly to conveniently access it when verifying attestation reports.


include/ccf/pal/attestation.h
```diff
@@ -53,6 +54,27 @@ namespace ccf::pal
...
       *reinterpret_cast<const snp::Attestation*>(quote_info.quote.data());
 
+    if (quote.version != snp::attestation_version)
+    {
+      throw std::logic_error(fmt::format(
+        "SEV-SNP: Attestation version is {} not expected {}",
+        quote.version,
+        snp::attestation_version));
+    }
+
```
achamayou, line 64: Can we have a link to the spec section for each field we check?

include/ccf/pal/attestation.h
```diff
@@ -128,17 +150,53 @@ namespace ccf::pal
         "SEV-SNP: Chip certificate (VCEK) did not sign this attestation");
     }
 
-    if (quote.policy.debug == 1)
+    // We should check this but the guest policy ABI is currently set to 0.31,
```
achamayou, line 153: TBD?

achamayou (approved):

PR #5026, opened 2023-02-17
Historical queries: Handle large requests being truncated by the host

eddyashton:
This fixes a bug in the historical query system.

It assumed that any `ledger_get_range(from, to)` write to the host ringbuffer would result in a corresponding `ledger_entry_range(from, to)` or `ledger_no_entry_range(from, to)` response from the (honest) host. This is not the case - the host applies a `max_size` cap since #3986 to avoid producing writes larger than the ringbuffer can handle, so it may in fact respond with `ledger_entry_range(from, to_)` (with `to_` < `to`). This PR makes the in-enclave historical query system robust to that - able to handle partial answers, and re-request entries which weren't provided.

This requires a significant rewrite of the historical query book keeping, to decouple slightly the `Store`s from the specific `Request` which triggered their fetch (with a side benefit that they are now better at sharing `Store`s between multiple `Request`s, without needing to refetch).

There's some knock-on changes in unit tests, a new unit test simulating the original issue, and some associated changes in the host-side `ledger.h` (removing the now-unnecessary `strict` mode, and ensuring that read responses correctly describe the prefix they contain).

~~Opening as a draft initially so I can let the CI find any tests I've missed, will open for review once I've done another cleanup pass.~~


src/host/ledger.h
```diff
@@ -804,19 +813,19 @@ namespace asynchost
...
         if (max_entries_size.has_value())
         {
-          max_size = max_entries_size.value() - entries.size();
+          max_size = max_entries_size.value() - rr.data.size();
         }
         auto v = f_from->read_entries(idx, to_, max_size);
         if (!v.has_value())
         {
-          return std::nullopt;
+          break;
```
eddyashton, line 821: This is a subtle but important change. Previously, if we had a single chunk that returned no entries, we'd return no entries for the entire range. For instance, if you requested seqnos 5-10, and 10 pushed this range over the max size allowed:
- If 9 and 10 (and anything before) are in the same file (chunk), then this would respond with the contents of 5-9
- If 10 was at the start of a new file (chunk), then this entire request would return nullopt indicating that _nothing_ could be read, ignoring the 5-9 that were already successfully read!

src/host/ledger.h
```diff
@@ -1423,21 +1434,15 @@ namespace asynchost
...
-                                purpose = purpose,
-                                strict = strict,
-                                max_entries_size =
-                                  max_entries_size](auto&& entry, int status) {
-                // NB: Even if status is cancelled (and entry is empty), we
-                // want to write this result back to the enclave
-                write_ledger_get_range_response(
-                  from_idx,
-                  to_idx,
-                  read_entries(from_idx, to_idx, strict, max_entries_size),
```
eddyashton, line 1438: This is a funny bug and hopefully a big saving - we were previously calling `read_entries` twice! Once in the `get_async` callback, and again in the `get_async_complete` callback! The latter is now correctly just forwarding the results from the earlier call.

src/node/historical_queries.h
```diff
@@ -518,43 +487,14 @@ namespace ccf::historical
 
     void fetch_entries_range(ccf::SeqNo from, ccf::SeqNo to)
```
eddyashton, line 488: This function was previously trying some last minute optimisations - hoping to track exactly what had been previously requested, and not re-request it. Its now a simpler pass-through to support explicit re-fetches when needed, which are now primarily driven in `tick`s. Note that this means periodic fetch attempts when there is a ledger hole, which would previously have only been triggered on an additional user call.

src/host/node_connections.h
```diff
@@ -337,17 +337,40 @@ namespace asynchost
...
+                ledger.read_entries(ae.prev_idx + 1, ae.idx);
 
-              frame = (uint32_t)framed_entries->size();
-              outbound_connection->write(frame, framed_entries->data());
+              if (!read_result.has_value())
+              {
+                LOG_FAIL_FMT(
+                  "Unable to send AppendEntries ({}, {}]: Ledger read failed",
+                  ae.prev_idx,
+                  ae.idx);
```
jumaffre, line 351: Should we `return;` here and on line 362 to avoid printing the log line on `382`?

tests/infra/logging_app.py
```diff
@@ -221,7 +221,7 @@ def verify_range_for_idx(
         self,
         idx,
         node=None,
-        timeout=5,
+        timeout=10,
```
achamayou, line 224: Why does this need bumping up? It doesn't seem like we should read any slower?

Issue #4994, opened 2023-02-10
Fix 'CCF Github Threading' pipeline failure


takuro-sato:
[CCF Github Threading](https://dev.azure.com/MSRC-CCF/CCF/_build?definitionId=13&_a=summary) has been failing.
examples: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=64322&view=logs&j=b2f97161-5795-5a8b-6b22-fc2704b56b5b&t=792a4181-50f1-5335-10af-3ab4628fda2a

## Reproducing locally
```bash
$ mkdir build && cd build
$ cmake -GNinja -DCOMPILE_TARGET=virtual -DWORKER_THREADS=2  ..
$ ninja
# reconfiguration_test_cft fails sometimes
$ sudo ./tests.sh --VV  --no-compress-output  --verbose -R reconfiguration_test_cft
# partitions fails everytime, and it takes shorter time.
$ sudo ./tests.sh -VV  --no-compress-output --verbose -L partitions -C partitions
```
## Investigation using ASAN

There is an error to be investigated
- [x] https://github.com/microsoft/CCF/issues/5050
- [x] https://github.com/microsoft/CCF/issues/5098
- [x] https://github.com/microsoft/CCF/issues/5099

## Investigation using TSAN

- https://github.com/microsoft/CCF/pull/4998
- https://github.com/microsoft/CCF/pull/5073
- [x] https://github.com/microsoft/CCF/issues/5074
- [x] https://github.com/microsoft/CCF/issues/5117
- https://github.com/microsoft/CCF/pull/5146
- https://github.com/microsoft/CCF/pull/5142
- https://github.com/microsoft/CCF/pull/5149

## Investigation in the classic way (with print debug etc)

Currently Multithread CI test fails often so it's not hard to reproduce.

- https://github.com/microsoft/CCF/pull/5035
- https://github.com/microsoft/CCF/pull/5036
- [x] https://github.com/microsoft/CCF/issues/5051
- https://github.com/microsoft/CCF/pull/5055 (Makes debug easier)
- [x] https://github.com/microsoft/CCF/issues/5077

eddyashton:
GitHub Threading pipeline is relatively green now! We will track new failures in isolated issues.

PR #5035, opened 2023-02-21
Fix deleting timer tasks

takuro-sato:
It partially fixes failing in Threading CI.
#4994

## Bug fixed in this PR

The forwarding errors in Threading CI

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=65076&view=logs&j=b2f97161-5795-5a8b-6b22-fc2704b56b5b&t=792a4181-50f1-5335-10af-3ab4628fda2a&l=1749
```
<red>504</> <yellow>Request was forwarded to node n[79276c50b411ccc9840a4869fc0ae395e87b828939256faae9f48ef890169e6b], but no response was received after 3000ms</>
```

### Cause of bug
When a backup gets a request and forwards  it to the primary, the request is saved as a timer task. Then it will be deleted after the primary responds to the backup. The problem is that the thread which gets the request (Thread A) and the one which gets the response (Thread B) might be different. In that case Thread A puts the task to its `TaskQueue` and Thread B tries to remove the task from its `TaskQueue` (and fails to do so) even though the task exists in Thread A's `TaskQueue`.

### Fix
Prepare a `TaskQueue` object for  the forwarding timer tasks that can be accessed by multiple threads.

### How to comfirm it's fixed
Unfortunatelly Threading CI is still failing, but I haven't seen the same error after the fix


eddyashton:
I can believe that this happens, but I don't think the fix will work as expected. Nothing is running the tasks on the new shared queue, so they will _never_ be sent - I expect that will manifest as the existing test for forwarding timeouts now failing and getting no timeout error?

We ensure thread affinity per-session, so all incoming traffic from a single user connection will execute on a fixed thread (including queuing a timeout task when forwarding a new command). I think we just need to ensure the cancellation happens on that thread, by queuing a task on the `TaskQueue` indicated by the node-to-node forwarded response's client session ID.

src/node/rpc/forwarder.h
```diff
@@ -34,7 +34,13 @@ namespace ccf
...
     using ForwardedCommandId = ForwardedHeader_v2::ForwardedCommandId;
     ForwardedCommandId next_command_id = 0;
-    std::unordered_map<ForwardedCommandId, threading::TaskQueue::TimerEntry>
+
+    struct TimeoutTask {
+      threading::TaskQueue::TimerEntry timer_entry;
+      uint16_t thread_id;
+    };
+
+    std::unordered_map<ForwardedCommandId, TimeoutTask>
```
takuro-sato, line 43: @eddyashton This is not the way we agreed, but I think it's necessary.
Originally I tried to get the thread ID from session ID when forwarding response is received to delete the timer task from the right thread.

[b30dc0a](https://github.com/microsoft/CCF/pull/5035/commits/b30dc0a207edffe0db79058a9e8cdb6e5f5fd12b)

But it's failing in CI and the message indicates that it's still failing to delete the timer task.
I think it's because the session ID when adding timer task and the one when deleting the timer task are different.

src/node/rpc/forwarder.h
```diff
@@ -433,8 +457,19 @@ namespace ccf
             auto it = timeout_tasks.find(cmd_id);
             if (it != timeout_tasks.end())
             {
-              threading::ThreadMessaging::instance().cancel_timer_task(
-                it->second);
+              if (threading::get_current_thread_id() != it->second.thread_id)
```
eddyashton, line 460: I believe `it->second.thread_id` should always match `threading::ThreadMessaging::instance().get_execution_thread(rep->client_session_id)`, and was hoping we could use that here rather than storing the thread ID. However I've spotted why that's not so easy - the `client_session_id` is in the message body rather than the header, so not trivially visible here! We could do a small refactor to make that possible (move the cancellation _after_ the calls to `recv_forwarded_response`), but I don't think its worth it.

Storing the originating thread ID like this is simpler, good spot!

eddyashton (approved):

Issue #5039, opened 2023-02-21
Add an end-to-end test that amplifies the response size


eddyashton:
We have end-to-end tests that submit extremely large requests to a CCF node (`e2e_common_endpoints.py:test_large_messages`), in the body and headers and various configurations, and other end-to-end tests that produce extremely large KV writes (`batched.js`). I think one case that we're missing is in unusually-large, enclave-generated _responses_. In particular, when a HTTP response approaches ringbuffer size limits, without being previously gated by ledger or request size limits, how does a node respond.

Issue #5032, opened 2023-02-21
ASAN use after free in Daily


achamayou:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=65406&view=logs&j=fff884fa-70f7-5142-c642-c9647ff4218c&t=002ec585-ff09-575c-f94f-67274e66fb5c

```
2023-02-21T03:30:12.2624916Z 59: =================================================================
2023-02-21T03:30:12.2625426Z 59: ==12330==ERROR: AddressSanitizer: heap-use-after-free on address 0x7f25357e37ff at pc 0x0000004a1fc7 bp 0x7f254bbdedb0 sp 0x7f254bbde578
2023-02-21T03:30:12.2625804Z 59: READ of size 16384 at 0x7f25357e37ff thread T2
2023-02-21T03:30:12.2626054Z 59:     #0 0x4a1fc6 in __asan_memcpy (/__w/1/s/build/cchost+0x4a1fc6)
2023-02-21T03:30:12.2626488Z 59:     #1 0x7f254ef786c9 in http2::read_outgoing_callback(nghttp2_session*, int, unsigned char*, unsigned long, unsigned int*, nghttp2_data_source*, void*) /__w/1/s/build/../src/http/http2_callbacks.h:45:7
2023-02-21T03:30:12.2626990Z 59:     #2 0x7f25511a2ee2 in nghttp2_session_pack_data /__w/1/s/build/../3rdparty/exported/nghttp2/nghttp2_session.c:7755:16
2023-02-21T03:30:12.2627410Z 59:     #3 0x7f25511ae0c8 in session_prep_frame /__w/1/s/build/../3rdparty/exported/nghttp2/nghttp2_session.c:2373:10
2023-02-21T03:30:12.2627813Z 59:     #4 0x7f2551167b67 in nghttp2_session_mem_send_internal /__w/1/s/build/../3rdparty/exported/nghttp2/nghttp2_session.c:3292:12
2023-02-21T03:30:12.2628222Z 59:     #5 0x7f255116777e in nghttp2_session_mem_send /__w/1/s/build/../3rdparty/exported/nghttp2/nghttp2_session.c:3585:9
2023-02-21T03:30:12.2628613Z 59:     #6 0x7f254f563800 in http2::Parser::send_all_submitted() /__w/1/s/build/../src/http/http2_parser.h:220:22
[...]

PR #5034, opened 2023-02-21
[release/2.x] Cherry pick: Historical queries fixes (#5026, #5040, #5058)

eddyashton:
Backports the following commits to `release/2.x`:
 - [Historical queries: Handle large requests being truncated by the host (#5026)](https://github.com/microsoft/CCF/pull/5026)
 - [Revert test case for large historical queries (#5040)](https://github.com/microsoft/CCF/pull/5040)
 - [Historical queries: Do not fetch entries until an early-enough secret has been fetched (#5058)](https://github.com/microsoft/CCF/pull/5058)

PR #5040, opened 2023-02-21
Revert test case for large historical queries

eddyashton:
I think the large messages in the historical range query test, added in #5026, have introduced some new CI instability (possibly #5032, possibly because we weren't previously testing #5039?).

Experimentally reverting the e2e test for now - if that fixes the CI jobs I'll merge this, and then write an equivalent but more targeted test.


jumaffre (approved):

achamayou:
@eddyashton that will potentially need backporting as well.

eddyashton:
> @eddyashton that will potentially need backporting as well.

The backports of #5026 (#5033 and #5034) are currently parked as `Draft`, and not merged - I'll combine this revert and any additional fix before merging them to their respective target branches.

eddyashton:
Merging as it looks like an _improvement_ in test stability, though I think there remains a new issue in `test_recover_service_with_wrong_identity`.

PR #5058, opened 2023-02-24
Historical queries: Do not fetch entries until an early-enough secret has been fetched

### LINKED ISSUE #5042: Fix `test_recover_service_with_wrong_identity` test case
This test, run as part of `recovery_test_suite`, is failing, at least in Virtual Instrumented. I believe this is since #5026.

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=65454&view=logs&j=fff884fa-70f7-5142-c642-c9647ff4218c&t=002ec585-ff09-575c-f94f-67274e66fb5c

eddyashton:
Resolves #5042.

The changes in #5026 meant that historical queries would result in fetches from the host immediately, even if a sufficiently-early secret was not yet known in-enclave. That results in a lot of wasted work - the host fetches and transmits many ledger entries, the enclave discards them as unreadable, repeats periodically - which should _eventually_ resolve, but in instrumented builds exceeds our expected timeouts. This change restores the previous behaviour; entries are only fetched from the host once we have an early-enough secret for them.

There's a few minor performance tweaks associated with this - avoiding double-lookup of the same seqno, and avoiding pessimal sharding of request ranges when _some_ secrets are known.

I'd like to validate this with a few runs through the Daily, and then I will manually backport by including it in #5033 and #5034.
EDIT - Daily looks happy.


eddyashton:
1 green run:
https://github.com/microsoft/CCF/pull/5058/checks?check_run_id=11573940486

achamayou (approved):

eddyashton:
Cherry-picked into #5033 and #5034 for backporting to `release/3.x` and `release/2.x`.

Issue #5051, opened 2023-02-22
Deadlock between map lock and snapshot lock


takuro-sato:
A part of https://github.com/microsoft/CCF/issues/4994

A deadlock is found locally

Deadlock between Map::sl (in `untyped_map.h`) and Snapshotter:lock.

```log                                                                                                                                                                      
  4494 60:     raise RuntimeError(FAILURES)                                                                                                                                                                                                                 
  4495 60:                        └ ["Failure in 1tx_reconfig: TimeoutError('Error copying files from 13.snapshots after 5 retries')"]                                                                                                                      
  4496 60: RuntimeError: ["Failure in 1tx_reconfig: TimeoutError('Error copying files from 13.snapshots after 5 retries')"]                                                                                                                                 
  4497 1/1 Test #60: reconfiguration_test_cft .........***Failed  317.07 sec                                                                                                                                                                                
  4498                                                                                                                                                                                                                                                      
  4499 0% tests passed, 1 tests failed out of 1                                                                                                                                                                                                             
  4500                                                                                                                                                                                                                                                      
  4501 Label Time Summary:                                                                                                                                                                                                                                  
  4502 cft    = 317.07 sec*proc (1 test)                                                                                                                                                                                                                    
  4503 e2e    = 317.07 sec*proc (1 test)                                                                                                                                                                                                                    
  4504                                                                                                                                                                                                                                                      
  4505 Total Test time (real) = 317.07 sec                                                                                                                                                                                                                  
  4506                                                                                                                                                                                                                                                      
  4507 The following tests FAILED:                                                                                                                                                                                                                          
  4508          60 - reconfiguration_test_cft (Failed)     
```

[1tx_reconfig_reconfiguration_test_cft_13_out.zip](https://github.com/microsoft/CCF/files/10806635/1tx_reconfig_reconfiguration_test_cft_13_out.zip)

[python_infra_log_with_stacktrace.log](https://github.com/microsoft/CCF/files/10806638/python_infra_log_with_stacktrace.log)

The stack trace is got by editing `tests/infra/remote.py`'s `LocalRemote.stop()` function
```py
    def stop(self, ignore_error_patterns=None):
        """
        Disconnect the client, and therefore shut down the command as well.
        """
        LOG.info("[{}] closing".format(self.hostname))
        if self.proc:
            if self.proc.returncode is None:
                LOG.info(f"pid: {self.proc.pid}")
                # get stack trace
                completed_process = subprocess.run(
                            [
[...]

PR #5059, opened 2023-02-24
Fix deadlock between map lock and snapshot lock

takuro-sato:
## Cause of bug

https://github.com/microsoft/CCF/issues/5051

# Fix

Use consistent order of locks. First MAP_LOCK then SNAPSHOT_LOCK.
I didn't choose the opposite because the fix would be much more complicated.

# Test

- Run reconfiguration test which caused the bug in https://github.com/microsoft/CCF/issues/5051 more than 20 times.
- Enable `TSAN` option and comment out deadlock suppression in `tsan_env_suppressions`, and then run tests.sh. Confirmed there is no warning about order between MAP_LOCK andSNAPSHOT_LOCK.
[test_log.zip](https://github.com/microsoft/CCF/files/10899483/test_log.zip)


cmake/common.cmake
```diff
@@ -93,6 +94,13 @@ if(SAN AND LVI_MITIGATIONS)
   )
 endif()
 
+if(TSAN AND LVI_MITIGATIONS)
+  message(
+    FATAL_ERROR
+      "Building with both TSAN and LVI mitigations is unsafe and deadlocks - choose one"
```
achamayou, line 100: It doesn't deadlock, but causes the linker to get stuck in an infinite loop.

takuro-sato: Thanks for clarifying. This PR is based on https://github.com/microsoft/CCF/pull/5073 and this change is a part of it.
I wanted to use TSAN to check if the changes are OK.

src/node/snapshotter.h
```diff
@@ -357,6 +359,8 @@ namespace ccf
       // at the last snapshottable index before idx, and schedule snapshot
       // serialisation on another thread (round-robin). Otherwise, only record
       // that a snapshot was generated.
+
+      kv::ScopedStoreMapsLock maps_lock(store);
```
jumaffre, line 363: Do we know the place in the code these two locks are acquired in the inverse order, leading to a deadlock?

takuro-sato: After the fix, I believe there is no more inverse order between map lock and snapshot lock based on TSAN result.

jumaffre (approved):

src/node/snapshotter.h
achamayou: I'd be curious too.

takuro-sato: Please check https://github.com/microsoft/CCF/issues/5051

Issue #4072, opened 2022-07-26
SGX to SNP recovery test


jumaffre:
Once support for AMD SEV-SNP is complete, users will want to upgrade their existing SGX CCF networks to SEV-SNP. We should add support and testing for the disaster recovery scenario where an existing SGX ledger is recovered from SNP nodes.

Note: live code upgrade scenario is covered by https://github.com/microsoft/CCF/issues/4827

PR #5063, opened 2023-02-27
SEV-SNP: support for multiple measurement types on same node

jumaffre:
Part of https://github.com/microsoft/CCF/issues/4072

Previously, a node built for a specific platform (i.e. SNP or SGX) would consider all attestation measurements as for that specific platform, preventing code updates from one platform to the other. 

A single node can now support multiple types of measurements, with as much as possible appropriate type safety, as measurements are of different sizes on different platforms.

Notes:
- `CodeDigest` has been renamed to `PlatformAttestationMeasurement`.
- `code_id`, `code_digest`, etc. have been renamed to `measurement` in most of the C++ code, with the exception of `mrenclave` in the `GET /node/quotes` endpoint, and `code_digest` in the nodes table.

Next: Remove SNP-specific endpoints, now superseded by generic KV governance endpoints.


include/ccf/pal/measurement.h
```diff
@@ -0,0 +1,166 @@
...
+      auto hex_str = measurement.hex_str();
+      return SerialisedEntry(hex_str.begin(), hex_str.end());
+    }
+
+    static ccf::pal::AttestationMeasurement<N> from_serialised(
+      const SerialisedEntry& data)
+    {
+      ccf::pal::AttestationMeasurement<N> ret;
+      ds::from_hex(std::string(data.data(), data.end()), ret.measurement);
+      return ret;
```
eddyashton, line 163: Why do we convert to hex here? Is this for compatibility with what we did previously?

jumaffre: Yes exactly. This used to be done for `CodeDigest` here: https://github.com/microsoft/CCF/blob/a5854de93bff07d4a830f8725015e012d4f07205/include/ccf/service/code_digest.h#L77-L95

eddyashton: Cool, thanks for clarifying. This is a shame, and I _think_ means that our `read_ledger` and similar tooling are obfuscating the measurements here (displaying hex-of-hex), but this is now necessary for backwards compatibility.

jumaffre: Good point. Based on https://github.com/microsoft/CCF/blob/main/python/ccf/read_ledger.py#L58, it looks like `read_ledger.py` simply displays the hex string so I don't think the hex-of-hex is displayed anywhere.

eddyashton: You're right again - we assume our blitted governance table keys are readable and display them directly as strings, without re-encoding as hex. Briefly validated with sandbox:

```
$ curl https://127.0.0.1:8000/gov/kv/nodes/code_ids -k
{"064a890f04d6bd85e915697023686cacf2cfcd85f9739d0f83054f351fe44cbf":"AllowedToJoin"}
```
```
$ /opt/openenclave/bin/oesign dump -e ./samples/apps/logging/liblogging.enclave.so.signed 
...
=== SGX Enclave Properties:
...
mrenclave=064a890f04d6bd85e915697023686cacf2cfcd85f9739d0f83054f351fe44cbf
```
```
$ read_ledger.py --uncommitted ./workspace/sandbox_0/0.ledger/
...
    table "public:ccf.gov.nodes.code_ids" (1 write):
      064a890f04d6bd85e915697023686cacf2cfcd85f9739d0f83054f351fe44cbf:
        "AllowedToJoin"
...
```

achamayou (approved):

PR #5075, opened 2023-03-02
Fix unlucky thread timing causing `indexing_test` failures

eddyashton:
Resolves #4603.

Added some logging to work out what's going on here. This test tries to simulate multi-threaded execution accessing indexes in a CCF node, with one thread simulating incoming write transactions (advancing the store's commit), others simulating incoming historical queries (polling the index for its current state), and another simulating the host (providing the ledger entries to the historical query cache as they are requested by the indexer). The host thread includes a simple counter to try and detect deadlocks. It previously started ticking that counter immediately, meaning that if this host thread was lucky in its scheduling it could detect a "deadlock" (the crash seen in the issue above) before the other threads had had a chance to create or request any state. The fix is to ensure that this deadlock counter only begins once the write transactions have completed, so its counting while the index coalesces rather than during total execution. I've also added some yields to try and ensure that the threads are interestingly interleaved.


MahatiC (approved):

src/indexing/test/indexing.cpp
```diff
@@ -614,10 +617,14 @@ TEST_CASE(
 
         handled_writes = writes.end() - writes.begin();
 
-        if (loops++ > 100)
+        if (work_done)
         {
-          throw std::logic_error("Looks like a permanent loop");
+          if (post_work_done_loops++ > 100)
```
MahatiC, line 622: 100 seems arbitrary?

eddyashton: The 100 is extremely arbitrary! We expect this thread to provide all of the entries the other threads are looking for extremely quickly (ie - in <5 loops), so 100 is aiming to be a "definite failure". We could probably safely make it a little lower, but we wouldn't noticeably change execution time (even time-to-failure), and risk getting false negatives - failures where we claim a permanent loop when the system is actually making progress.

achamayou (approved):

Issue #4158, opened 2022-08-19
Multi-Container: Build an initial demo


eddyashton:
Most of the work for this is covered by other tickets, but using this as a hold-all for remaining work towards an MVP demo.

Additional work I see for this demo specifically:
- Writing a simple executor, and some infra help to operate it
- Breaks from our intended gRPC API for basic gRPC compatibility (ie - no streaming responses means polling requests, no long-lived gRPC blocking server calls means regular Ready polling by clients)
- Timers and timeouts for pending client requests, to check that additional error handling can be added in future

shokouedamsr:
- [ ] Document the python test and turn it into a sample.

PR #5076, opened 2023-03-02
External executor: further containerisation of external executors 

jumaffre:
Part of https://github.com/microsoft/CCF/issues/4158 and follow-up from #5018 

The containerised variant of the external executors is now more in line with a full end-to-end demo, with a clearer separation with our test infra:
- New separate `Dockerfile` to build external executor image (Python for now) so that dependencies are only installed once (on `docker build`) and container start time is minimised (we used to `pip install requirements.txt` as part of the container command).
- Container clean termination (including deactivation with CCF) is now tied up to the container receiving `SIGTERM`.
- Container configuration is now passed via environment variable rather than CLI arguments, in line with existing ACI practices, while also making is slightly easier to port to other languages (i.e. `app.py` is minimal).
- External executor logs are now copied in the Python infra `workspace/`, under the node directory to which the external executor is connected (e.g. `workspace/externa_executor_0/wiki_cacher_0`).
- Re-usable parts (e.g. registration and `.proto` files) are moved to a separate `ccf/` folder.

The (simplified) source tree for the wiki cacher external executor becomes:
```
├── Dockerfile
├── app.py
├── ccf
│   ├── executors
│       └── registration.py
│   └── proto
│       ├── executor_registration.proto
│       ├── http.proto
│       └── kv.proto
└── wiki_cacher
    ├── requirements.txt
    └── wiki_cacher.py
```


tests/external_executor/executors/ccf/proto/executor_registration.proto
```diff
@@ -0,0 +1,72 @@
+syntax = "proto3";
```
jumaffre, line 1: These are copied verbatim from https://github.com/microsoft/CCF/tree/main/src/apps/external_executor/protobuf, which isn't great and should be resolved before this PR is merged in.

This is because we generate the source from these in the associated `Dockerfile`, which can only copy files from its own build context. I wonder whether we should move all external executor related source (C++ source, proto files, tests) in a standalone folder at the top level of this repo?

eddyashton (approved):

PR #5080, opened 2023-03-06
Require COSE signing of `POST /gov/recovery_shares` (and fix `recovery_test_cft`)

eddyashton:
`recovery_test_cft` started failing a few days ago, because the member and user certs that are part of the expired service have finally expired. The fix is to stop using those certs for auth - all member requests should be unauth'd or signed, and the users should be manually-cycled if we're going to use them for cert auth.

So this PR contains 2 changes:
- Remove the `False` option for `authenticate_session` in the Python infra, and replace `True` with `"HTTPSIG"`
- Update the `/gov/recovery_shares` endpoints to ~~take a `{member_id}` path parameter, and~~ ~~accept COSE signed~~ _require signed_ auth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Historical query system will re-request entries if the host fails to provide them within a fixed time.
 - Renamed `ccf::CodeDigest` to `ccf:pal::PlatformAttestationMeasurement` and `get_code_id()` to `get_measurement()` (#5063).
 - Recovery procedure:
-  - The `GET /gov/recovery_share` endpoint has been replace with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
+  - The `GET /gov/recovery_share` endpoint has been replaced with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
   - The `POST /gov/recovery_share` endpoint has been replaced with `POST /gov/recovery_share/{member_id}`. This request must now be signed, and will not allow purely TLS authentication.
   - `submit_recovery_share.sh` can still be used to fetch, decrypt, and submit a recovery share, the only change is that `--cert` and `--key` arguments are now explicitly required, and the underlying submission produces a COSE signature.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Renamed `ccf::CodeDigest` to `ccf:pal::PlatformAttestationMeasurement` and `get_code_id()` to `get_measurement()` (#5063).
 - Recovery procedure:
   - The `GET /gov/recovery_share` endpoint has been replace with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
-  - The `POST /gov/recovery_share` endpoint has been replaced with `POST /gov/recovery_share/{member_id}`. This request must now be signed, and will not allow purely TLS authentication. 
+  - The `POST /gov/recovery_share` endpoint has been replaced with `POST /gov/recovery_share/{member_id}`. This request must now be signed, and will not allow purely TLS authentication.
   - To sign a `POST /gov/recovery_share` request with COSE, the `ccf.gov.msg.type` protected header may now be `recovery_share`, and a new protected header `ccf.gov.msg.member_id` is now supported.
   - `submit_recovery_share.sh` can still be used to fetch, decrypt, and submit a recovery share, the only change is that `--cert` and `--key` arguments are now explicitly required, and the underlying submission produces a COSE signature.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Additional logging of historical query flow in `UNSAFE` builds.
 - Historical query system will re-request entries if the host fails to provide them within a fixed time.
 - Renamed `ccf::CodeDigest` to `ccf:pal::PlatformAttestationMeasurement` and `get_code_id()` to `get_measurement()` (#5063).
+- Recovery procedure:
+  - The `GET /gov/recovery_share` endpoint has been replace with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
+  - The `POST /gov/recovery_share` endpoint has been replaced with `POST /gov/recovery_share/{member_id}`. This request must now be signed, and will not allow purely TLS authentication. 
+  - To sign a `POST /gov/recovery_share` request with COSE, the `ccf.gov.msg.type` protected header may now be `recovery_share`, and a new protected header `ccf.gov.msg.member_id` is now supported.
+  - `submit_recovery_share.sh` can still be used to fetch, decrypt, and submit a recovery share, the only change is that `--cert` and `--key` arguments are now explicitly required, and the underlying submission produces a COSE signature.
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Recovery procedure:
   - The `GET /gov/recovery_share` endpoint has been replace with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
   - The `POST /gov/recovery_share` endpoint has been replaced with `POST /gov/recovery_share/{member_id}`. This request must now be signed, and will not allow purely TLS authentication.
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Additional logging of historical query flow in `UNSAFE` builds.
 - Historical query system will re-request entries if the host fails to provide them within a fixed time.
 - Renamed `ccf::CodeDigest` to `ccf:pal::PlatformAttestationMeasurement` and `get_code_id()` to `get_measurement()` (#5063).
+- Recovery procedure:
+  - The `GET /gov/recovery_share` endpoint has been replace with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
```
achamayou, line 16:
```suggestion
  - The `GET /gov/recovery_share` endpoint has been replaced with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
```

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+  - The `POST /gov/recovery_share` endpoint has been replaced with `POST /gov/recovery_share/{member_id}`. This request must now be signed, and will not allow purely TLS authentication.
```
achamayou, line 17: I don't understand why we are forbidding pure TLS auth here.

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+  - To sign a `POST /gov/recovery_share` request with COSE, the `ccf.gov.msg.type` protected header may now be `recovery_share`, and a new protected header `ccf.gov.msg.member_id` is now supported.
```
achamayou, line 18: member_id is redundant with kid

include/ccf/endpoints/authentication/cose_auth.h
```diff
@@ -15,6 +15,7 @@ namespace ccf
     std::optional<std::string> kid;
     std::optional<std::string> gov_msg_type;
     std::optional<std::string> gov_msg_proposal_id;
+    std::optional<std::string> gov_msg_member_id;
```
achamayou, line 18:
```suggestion
```

python/ccf/cose.py
```diff
@@ -28,8 +28,13 @@
 Pem = str
 
 GOV_MSG_TYPES_WITH_PROPOSAL_ID = ["ballot", "withdrawal"]
+GOV_MSG_TYPES_WITH_MEMBER_ID = ["recovery_share"]
 
-GOV_MSG_TYPES = ["proposal", "ack", "state_digest"] + GOV_MSG_TYPES_WITH_PROPOSAL_ID
+GOV_MSG_TYPES = (
+    ["proposal", "ack", "state_digest"]
+    + GOV_MSG_TYPES_WITH_PROPOSAL_ID
+    + GOV_MSG_TYPES_WITH_MEMBER_ID
```
achamayou, line 36: We don't need this, members are identified by the kid.

achamayou (changes requested):
> Remove the False option for authenticate_session in the Python infra, and replace True with "HTTPSIG"

Is the test failing for LTS too? If not, then I think we can remove support for HTTP request signing altogether. If yes, then I don't think this change helps.

> Update the /gov/recovery_shares endpoints to take a {member_id} path parameter, and accept COSE signed auth

Both GET and POST /gov/recovery_shares accept COSE signed auth already. This change removed auth for GET, so if anything it removes COSE auth rather than adds it.

The member_id parameter duplicates kid, and I don't understand why we are removing session auth support, which doesn't seem to hurt in any way.

CHANGELOG.md
eddyashton: If we have a request that we expect to be signed, why would we allow it to come in unsigned based on TLS cert auth? That seems like a major hole. We require proposals and ballots to be signed, this is just moving recovery shares to require the same?

CHANGELOG.md
eddyashton: Ah ha - I'd missed that it was already permitted server-side, because we didn't include it in the client-side `cose.py`. I'll clarify the CHANGELOG.

eddyashton: `member_id` header removed, using existing `kid`.

python/ccf/cose.py
eddyashton: We still need this added to `GOV_MSG_TYPES` here on the client side, so that we can call `ccf_cose_sign1 --ccf-gov-msg-type recovery_share`, and not get a `argument --ccf-gov-msg-type: invalid choice: 'recovery_share'` error. I've removed the additional plumbing for a `ccf.gov.msg.message_id` header+arg.

eddyashton:
> Is the test failing for LTS too? If not, then I think we can remove support for HTTP request signing altogether. If yes, then I don't think this change helps.

I don't understand how LTS is related. The failing test is caused by using TLS session auth for governance requests, which in practice means just `GET /recovery_shares`, `POST /recovery_shares`, and `POST /update_state_digest`. Every other `/gov` request else is actually signed, regardless of whether the session its sent over was using member-cert auth or anonymous auth. Those 3 requests would fail on long-expired services, where the member cert auth was no longer accepted. This PR tries to close those holes by moving (most of) those remaining endpoints to be consistent, require a signature or be unsigned. We should probably do the same for `/update_state_digest`, but I haven't in this PR as its not causing any issues (yet).
 
> Both GET and POST /gov/recovery_shares accept COSE signed auth already. This change removed auth for GET, so if anything it removes COSE auth rather than adds it.

I've updated the description. It wasn't clear that `recovery_share` accepted COSE signing, and we never actually tested it in practice (because we only called its happy path from `submit_recovery_share.sh`, which used TLS auth and no COSE signing). The tests now both use the COSE signing and enforce signing. That accidentally includes HTTP signing for now, as it does for every other endpoint, but we can/should remove that soon, pending API approval.

python/utils/submit_recovery_share.sh
```diff
@@ -46,9 +56,28 @@ if [ -z "${member_enc_privk}" ]; then
...
 # Then, decrypt encrypted share with member private key submit decrypted recovery share
 # Note: all in one line so that the decrypted recovery share is not exposed
-echo "${encrypted_share}" | openssl base64 -d | openssl pkeyutl -inkey "${member_enc_privk}" -decrypt -pkeyopt rsa_padding_mode:oaep -pkeyopt rsa_oaep_md:sha256 | openssl base64 -A | jq -R '{share: (.)}' | curl -i -sS --fail -H "Content-Type: application/json" -X POST "${node_rpc_address}"/gov/recovery_share "${@}" -d @-
\ No newline at end of file
+echo "${encrypted_share}" \
+    | openssl base64 -d \
+    | openssl pkeyutl -inkey "${member_enc_privk}" -decrypt -pkeyopt rsa_padding_mode:oaep -pkeyopt rsa_oaep_md:sha256 \
+    | openssl base64 -A \
+    | jq -R '{share: (.)}' \
+    | ccf_cose_sign1 --content - --signing-cert "${member_cert}" --signing-key "${member_key}" \
```
jumaffre, line 81: This implies that `ccf_cose_sign1` is in the path, e.g. via `pip install ccf`. I don't think `submit_recovery_share.sh` had such a requirement before and could be run "freely"? If so, this should be mentioned in the changelog.

CHANGELOG.md
achamayou: But we have no such requirement there?

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Additional logging of historical query flow in `UNSAFE` builds.
 - Historical query system will re-request entries if the host fails to provide them within a fixed time.
 - Renamed `ccf::CodeDigest` to `ccf:pal::PlatformAttestationMeasurement` and `get_code_id()` to `get_measurement()` (#5063).
+- Recovery procedure:
+  - The `GET /gov/recovery_share` endpoint has been replaced with `GET /gov/recovery_share/{member_id}`. This can be accessed without any authentication.
```
achamayou, line 16:
```suggestion
  - The `GET /gov/recovery_share` endpoint has been replaced with `GET /gov/encrypted_recovery_share/{member_id}`. This can be accessed without any authentication.
```

python/utils/submit_recovery_share.sh
achamayou: I think that's fair, we also assume jq

achamayou (approved):

CHANGELOG.md
eddyashton: Discussed offline. Summary:

We're moving away from all member-cert based auth. Member certs are a historical artifact, members are really identified by their _key_. We authenticate with that key by requiring signatures, so this endpoint that needs to authenticate members (to prevent driveby `POST`s replacing a member's legitimate recovery share), does so with signatures _even though it doesn't store those signatures_.

achamayou (approved):

PR #5088, opened 2023-03-08
Two step shutdown

### LINKED ISSUE #4951: Two-phase shutdown
Operators sometimes face a situation where their processes receive a SIGTERM (eg. VM migration), to which they want to react by retiring the node, in order to minimise disruption. As discussed with @lynshi, a way to do this could be:

1. On SIGTERM, set an internal flag to Terminating
2. Expose this flag via an operator endpoint, such as `/node/terminating`
3. Poll for an additional operator-controlled condition, and actually shut down the process when it is met
  a. A mechanism may be to poll for the existence of the PID file, if one was specified. The operator can delete it to confirm to the node that shutting down is acceptable.

The behaviour in case a second SIGTERM is received would need to be decided. It could be ignored (the operator is already retiring the node), registered discreetly (increment a counter exposed at `/node/terminating`), or made to actually shut down the node.

A termination RPC was discussed, but authorising the operator reliably is difficult, particularly early and late in the life of a node, when consensus cannot propagate updated operator identities.

achamayou:
Resolve #4951

- [x] End to end test
- [x] Documentation

`ignore_first_sigterm` config variable. When set, will cause a node to ignore the first `SIGTERM` it receives, but the `/node/state` endpoint will show `"stop_notice": true`.

A second `SIGTERM` will cause the process to shut down as normal. This can be useful in orchestration settings where nodes receive unsollicited signals that the operator wishes to react to.


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - New `/node/index/strategies` endpoint, which will list all indexing strategies currently installed alongside a description of how far each has progressed.
 - When starting a host subprocess, applications may now pass data to its standard input. Additionally, the process' output is captured and logged by CCF (#5056).
+- `CCF_IGNORE_FIRST_SIGTERM` environment variable. When set, will cause a node to ignore the first `SIGTERM` it receives, but the `/node/process` endpoint will return `{"stop_notice": true}`. A second `SIGTERM` will cause the process to shut down as normal. This can be useful in orchestration settings where nodes receive unsollicited signals that the operator wishes to react to.
```
lynshi, line 25: Should this be a configuration file parameter instead of environment variable? Don't think it's a good idea to pass settings in different ways

doc/schemas/node_openapi.json
```diff
@@ -1370,6 +1370,28 @@
...
       }
     },
+    "/node/process": {
+      "get": {
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/json"
```
lynshi, line 1380: nit: if a typed response was returned it would be clearer from the API spec what the format of the JSON is

lynshi (approved):
Feature wise looks good to me, just some small nits

CHANGELOG.md
achamayou: I'll make it a config parameter.

doc/schemas/node_openapi.json
achamayou: Will fix.

src/node/rpc/node_frontend.h
```diff
@@ -1807,6 +1807,29 @@ namespace ccf
...
+          return make_error(
+            HTTP_STATUS_INTERNAL_SERVER_ERROR,
+            ccf::errors::InternalError,
+            "NodeConfigurationSubsystem is not available");
+        }
+
+        auto r = nlohmann::json();
+        r["stop_notice"] =
+          node_configuration_subsystem->has_received_stop_notice();
+        return make_success(r);
```
jumaffre, line 1825: Why not fold it in the existing [`GET /node/state`](https://microsoft.github.io/CCF/main/operations/operator_rpc_api.html#get--node-state) endpoint?

src/node/rpc/node_frontend.h
```diff
@@ -1807,6 +1807,29 @@ namespace ccf
...
+        if (!node_configuration_subsystem)
+        {
+          return make_error(
+            HTTP_STATUS_INTERNAL_SERVER_ERROR,
+            ccf::errors::InternalError,
+            "NodeConfigurationSubsystem is not available");
+        }
+
+        auto r = nlohmann::json();
+        r["stop_notice"] =
```
jumaffre, line 1823: I think this should be called "termination_notice" or "terminated" to make it clear it refers to SIGTERM.

tests/reconfiguration.py
```diff
@@ -149,6 +149,55 @@ def test_add_node(network, args, from_snapshot=True):
...
+                    host=host
+                ),
+                operator_rpc_interface: infra.interfaces.RPCInterface(
+                    host=host,
+                    endorsement=infra.interfaces.Endorsement(
+                        authority=infra.interfaces.EndorsementAuthority.Node
+                    ),
+                ),
+            }
+        )
```
jumaffre, line 172: Is this needed?

src/node/rpc/node_frontend.h
achamayou: We call this stop internally though, and we could choose to interpret other signals this way.

src/node/rpc/node_frontend.h
achamayou: Good point.

tests/reconfiguration.py
achamayou: No

achamayou: It is apparently, and there's not much reason to have it different than the rest of the test, so keeping it that way.

jumaffre: It's odd that this is needed at all. What is the error when this is commented out?

tests/reconfiguration.py
```diff
@@ -149,6 +149,53 @@ def test_add_node(network, args, from_snapshot=True):
...
+    )
+    network.join_node(new_node, args.package, args, ignore_first_sigterm=True)
+    network.trust_node(
+        new_node,
+        args,
+        validity_period_days=args.maximum_node_certificate_validity_days // 2,
```
jumaffre, line 178:
```suggestion
```

jumaffre (approved):

tests/infra/remote.py
```diff
@@ -666,6 +674,7 @@ def __init__(
         set_snp_uvm_endorsements_envvar=True,
         snp_uvm_endorsements=None,
         set_snp_report_endorsements_envvar=True,
+        ignore_first_sigterm=False,
```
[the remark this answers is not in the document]
achamayou: Yeah, but we have it for everything else, presumably for auto completion.

PR #5091, opened 2023-03-09
Stabilise external executor end-to-end tests

jumaffre:
The `external_executor_test` end-to-end test recently got flaky so this PR aims to stabilise it by using the `LoggingApp` instead of the `WikiCacher` as much as possible, especially when spinning up parallel executors.

As discussed with @eddyashton, there seems to be an issue with out-of-order responses received by clients (see notes in `test_parallel_executors`) that will need to be fixed when this work is resumed. In the meantime, assertions on the content of responses are relaxed.


achamayou (approved):

tests/external_executor/executors/app.py
```diff
@@ -8,7 +8,7 @@
 from ccf.executors.registration import register_new_executor
 
 # pylint: disable=import-error, no-name-in-module
-from wiki_cacher.wiki_cacher import WikiCacherExecutor
+from logging_app.logging import LoggingExecutor
```
achamayou, line 11: @jumaffre external_indexer.py needs updating too.

PR #5101, opened 2023-03-10
Add default clang_version in app-dev.yml

achamayou:
Address issue hit in https://github.com/microsoft/CCF/actions/runs/4387296912

### DIFF
--- a/getting_started/setup_vm/app-dev.yml
+++ b/getting_started/setup_vm/app-dev.yml
@@ -2,6 +2,7 @@
   vars:
     run_only: false
     platform: "sgx"
+    clang_version: "10"
   tasks:
     - import_role:
         name: intel


getting_started/setup_vm/app-dev.yml
```diff
@@ -2,6 +2,7 @@
   vars:
     run_only: false
     platform: "sgx"
+    clang_version: "10"
```
jumaffre, line 5: Should the appropriate clang version should be passed in [here](https://github.com/microsoft/CCF/blob/main/.github/workflows/containers.yml) too? i.e. 12 for virtual/snp and 10 for SGX?

achamayou: I would prefer not to make the switch this time to minimise the gap between the tag and the release containers. We can switch in dev6.

jumaffre (approved):

Issue #5078, opened 2023-03-03
Investigate `partitions_test` failure: "Adjacent signatures at T.X and T.X+1"


eddyashton:
#4908 changed how we produced the first signature in a new term, and added a test to confirm this didn't lead to redundant adjacent signatures.

The `partitions_test` has an occasional failure showing these adjacent signatures are sometimes produced:
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=66238&view=logs&j=abb827b0-1387-5f0d-393d-8da14bd7f99d&t=776cc0e6-a7f3-556a-baca-d314f367d1cc

The relevant output from the primary at the time (node 8) in the above repro, showing `Signed at 211` to `Signed at 212`, is the following:

```
2023-03-03T03:35:30.165954Z -0.001 0   [trace] ../src/consensus/aft/raft.h:1037     | Looking for match with 39.210, from 40.210, best answer is 209
2023-03-03T03:35:30.165957Z -0.001 0   [debug] ../src/consensus/aft/raft.h:1658     | Recv append entries response to n[062e57f05447e4370d35f0b492dff1b669257ef5332c7c424934f5de36bd21c6] from n[5b80a388064f2b6f15bd952a02faf266e30342eaa1598d4c0d441074eae507fa]: failed
2023-03-03T03:35:30.206215Z -0.001 0   [debug] ../src/node/history.h:750            | Signed at 211 in view: 42, previous signature was at 210
2023-03-03T03:35:30.206222Z -0.001 0   [debug] ../src/kv/store.h:928                | Store::commit 211 globally_committable
2023-03-03T03:35:30.206226Z -0.001 0   [trace] ../src/kv/store.h:962                | Inserting pending tx at 211
[...]

PR #5121, opened 2023-03-17
Fix daily - remove "Adjacent signatures" assert

eddyashton:
Tracking this in #5078, don't understand why its happening but its not a big issue and the assert is causing red Dailys - remove assert for now.


achamayou:
@eddyashton can you fix the formatting and poke the daily canary?

achamayou (approved):

PR #5122, opened 2023-03-17
Clear headers on error during handler execution

### LINKED ISSUE #4340: Response headers are not cleared on error responses
When a CCF transaction fails, the framework code in `frontend.h` tries to overwrite the app-produced response and return a descriptive error instead. It overwrites the response's status code and body, but retains any headers that were previously set by the application handler. I believe this is an unintentional leak, and those response headers should be cleared on error.

achamayou:
Fix #4340


src/node/rpc/frontend.h
```diff
@@ -625,20 +629,23 @@ namespace ccf
         catch (RpcException& e)
         {
           ctx->set_error(std::move(e.error));
+          ctx->clear_response_headers();
```
eddyashton, line 632:
```suggestion
          ctx->clear_response_headers();
          ctx->set_error(std::move(e.error));
```

I think that all of these should clear response headers _before_ setting an error, in case `set_error` includes some headers.

achamayou: I'd checked we don't, but you're right that it's safer.

PR #5123, opened 2023-03-17
Log execution metrics in `js_generic`

eddyashton:
To aid building standard metrics dashboards, this adds a log line whenever a JS request execution completes. This is on by default, but opt-out for cases where this logging breaks intended confidentiality.

Note that I've named the function `enableMetricsLogging` for consistency with the existing `enableUntrustedDateTime`, so the opt-out is the maybe-clunky `enableMetricsLogging(false)`, rather than `disableMetricsLogging(true)`. The granularity argument isn't exactly the same (the effect is dictated by the _final_ call, so precise scope control doesn't mean much), but I think its nice to keep them consistent?

### DIFF
--- a/doc/build_apps/release_policy.rst
+++ b/doc/build_apps/release_policy.rst
@@ -79,10 +79,25 @@ REST API guarantees spelled out in the first section apply across releases, but 
 
 .. image:: /img/ccf_release.png
     :alt: Release
-    
-CCF 1.0 Sunset
------------------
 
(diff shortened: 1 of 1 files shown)


achamayou (approved):

doc/build_apps/js_app_bundle.rst
```diff
@@ -130,6 +130,21 @@ In many places where timestamps are desired, they should come from the outside w
...
 
+Execution metrics
+~~~~~~~~~~~~~~~~~
+
+By default the CCF JS runtime will print a log line for each completed JS request. This lists the request path and response status as well as how long the request took to execute. Each line also includes a ``[js]`` tag so that they can easily be filtered and sent to an automated monitoring system. These lines have the following format:
+
+.. code-block::
+
+    <timestamp> [info ][js] <...> | JS execution complete: Method=GET, Path=/app/make_randoms, Status=200, ExecMilliseconds=30
```
achamayou, line 140: Because this logging and its specific format are intended to be consumed by other tools, I suggest we capture it in at least one end to end test, so as to avoid any intentional regressions or changes in the future.

eddyashton: I've added an end-to-end for this. We used to have a `get_logs()` that was called in every `stop()`, and actually returned the filtered errors. I've refactored this so that `get_logs()` returns the path to the out and err files (copied from a remote if necessary), so they can be used by this test, and the places that were actually using the error logs (`network.py`) now read and filter themselves. This almost certainly has unintended consequences for how we detect node shutdown errors.

PR #5125, opened 2023-03-20
Infra: refactor Docker remote

jumaffre:
As discussed with @achamayou and @plietar, this PR refactors the existing `DockerShim` to make it another remote type inheriting most of its features from `LocalRemote` and provides a simpler separation of remote types in the Python infra.


tests/infra/node.py
```diff
@@ -427,14 +424,10 @@ def _read_ports(self):
                 with open(rpc_address_file, "r", encoding="utf-8") as f:
                     lines = f.read().splitlines()
                     it = [iter(lines)] * 2
-                for (rpc_host, rpc_port), (_, rpc_interface) in zip(
+                for (_, rpc_port), (_, rpc_interface) in zip(
                     zip(*it), self.host.rpc_interfaces.items()
                 ):
```
plietar, line 429:
```suggestion
                for (_, rpc_port), rpc_interface in zip(
                    zip(*it), self.host.rpc_interfaces.values()
                ):
```

I'm pretty sure the `zip(*it)` can also be simplified now, but I'm not sure I even understand what's going on with `it = [iter(lines)] * 2` above

tests/infra/docker_remote.py
```diff
@@ -190,10 +197,10 @@ def start(self):
         ]["IPAddress"]
         for _, rpc_interface in self.host.rpc_interfaces.items():
             rpc_interface.public_host = self.container_ip
-        self.remote.hostname = self.container_ip
+        self.hostname = self.container_ip
         LOG.debug(f"Started container {self.container_name} [{self.container_ip}]")
 
-    def get_host(self):
+    def get_rpc_host(self):
         return self.container_ip
```
plietar, line 204: Is this used anywhere? Same for the one in `CCFRemote`.

Even grepping the current code base it seems this was dead code already.

tests/infra/node.py
jumaffre: I've looked into this and I don't think it can be simplified further. Note that this is only to spin up 1.x nodes too, which output different address files.

tests/infra/docker_remote.py
```diff
@@ -71,15 +66,33 @@ def _stop_container(self, container):
                 time.sleep(0.5)
                 continue
 
-    def __init__(self, *args, host=None, **kwargs):
+    @staticmethod
+    def get_host(host):
+        # Bind local RPC address to 0.0.0.0, so that it be can be accessed from outside container
+        for _, rpc_interface in host.rpc_interfaces.items():
```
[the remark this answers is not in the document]
jumaffre: I've renamed this function.

tests/infra/node.py
plietar: What does this code actually do though?

`lines` is a list of strings, it then creates a single iterator object over it. It creates a list `it`, with two elements both of them pointing to the same iterator. `zip(*it)` is now creating a zip of an iterator with itself, which is very confusing, and then we loop over the zip, throwing away the first value.

Assuming a naive implementation of zip, my best guess is this iterates over every other element of lines, at indices 1, 3, 5, ... Running this snippet seems to confirm that:
```python
lines = ["a","b","c","d"]
it = [iter(lines)]*2
print([ y for (x, y) in zip(*it) ])
# [ "b", "d" ]
```

In which case you can get the same behaviour with `lines[1::2]`, ie. start at index 1 and increment by 2 each time.
ie.

```python
for rpc_port, rpc_interface in zip(lines[1::2], self.host.rpc_interfaces.values()):
```

(Sorry I realize this is kind of off-topic, I just found the zip to itself very hard to comprehend)

achamayou (approved):

PR #5137, opened 2023-03-23
Remove signed HTTP request support

achamayou:
This completes #3875 by removing signed HTTP request support from CCF, after its deprecation in 3.0.

- [x] remove tests that make use of http request signing, see comment for explanation
- [x] resolve problem in membership and governance test with signed GETs (see below for discussion)
- [x] additional pass to remove dead code


src/node/rpc/member_frontend.h
```diff
@@ -447,9 +447,9 @@ namespace ccf
         cose_recent_proposals->put(key, proposal_id);
         // Only keep the most recent window_size proposals, to avoid
         // unbounded memory usage
-        if (replay_keys.size() >= window_size)
+        if (replay_keys.size() >= (window_size - 1) /* We just added one */)
```
achamayou, line 450: This was previously off by one, hidden by the fact that the test was the first to submit COSE proposals.

tests/governance_js.py
```diff
@@ -99,7 +99,7 @@ def check_msg_type(verb, path, name):
             ("POST", "/gov/ack", "ack"),
             ("POST", "/gov/ack/update_state_digest", "state_digest"),
             ("POST", "/gov/recovery_share", "recovery_share"),
-            ("GET", "/gov/recovery_share", "encrypted_recovery_share"),
+            # ("GET", "/gov/recovery_share", "encrypted_recovery_share"),
```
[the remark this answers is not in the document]
achamayou: Went with option 3 in #5145

achamayou:
t_cose does not support passing in custom header parameters [1][2] when signing (nor when verifying, but we can and do re-parse there with QCBOR), which we need for CCF governance.

[1] https://github.com/laurencelundblade/t_cose/blob/master/examples/t_cose_basic_example_ossl.c#L262
[2] https://github.com/laurencelundblade/t_cose/issues/5

achamayou:
The other problem with the proposal test is that it relies on being able to submit concurrent identical proposals, which is now caught by the replay proposal protection.

achamayou:
Following discussion with @jumaffre, I am removing the proposal_id_test, because:

1. Updating create_signed_request to use COSE Sign1 requires substantial changes to t_cose that we don't have time to make now.
2. Even with this change, the test would not work with the new replay protection functionality (tested separately).
3. The logic covered by this test uniquely is relatively trivial (History rollback), generic KV conflict functionality is tested separately.

doc/governance/hsm_keys.rst
```diff
@@ -117,45 +117,6 @@ Like ``ccf_cose_sign1``, the output can be sent directly to the service via curl
         "state": "Open"
     }
 
-HTTP Signing
```
jumaffre, line 120: I think we may want to keep this section with a warning that this scheme was removed in 4.x.

doc/governance/proposals.rst
```diff
@@ -278,18 +278,6 @@ For example, ``member1`` may submit a proposal to add a new member (``member4``)
       "state": "Open"
     }
 
-Or alternatively, with the old signature method:
```
jumaffre, line 281: Same here I think, we should add a warning at the top of the page to say that this has been removed in 4.x.

doc/schemas/app_openapi.json
```diff
@@ -286,16 +286,6 @@
         "description": "Request payload must be a COSE Sign1 document, with expected protected headers. Signer must be a member identity registered with this service.",
         "scheme": "cose_sign1",
         "type": "http"
-      },
-      "member_signature": {
-        "description": "Request must be signed according to the HTTP Signature scheme. The signer must be a member identity registered with this service.",
-        "scheme": "signature",
-        "type": "http"
-      },
-      "user_signature": {
```
jumaffre, line 295: Do we have an equivalent COSE signature scheme that users can use for apps, in both JS and C++?

doc/use_apps/issue_commands.rst
```diff
@@ -30,24 +30,7 @@ The response body (the JSON value ``true``) indicates that the request was execu
 Signing
 -------
 
-In some situations CCF requires signed requests, for example for member votes. Two signing schemes are supported as of 3.x.
-
-HTTP Signatures
-~~~~~~~~~~~~~~~
-
-An implementation of `IETF HTTP Signatures draft RFC <https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-08>`_ , but
```
jumaffre, line 38: Forget my other commends; this is probably the place to have the warning!

src/node/rpc/member_frontend.h
```diff
@@ -91,24 +90,6 @@ namespace ccf
   class MemberEndpoints : public CommonEndpointRegistry
   {
   private:
-    // Wrapper for reporting errors, which both logs them under the [gov] tag
```
jumaffre, line 94: I don't think this should go?

src/node/rpc/member_frontend.h
```diff
@@ -505,7 +486,7 @@ namespace ccf
 
       auto getter =
         [&, table](endpoints::ReadOnlyEndpointContext& ctx, nlohmann::json&&) {
-          GOV_TRACE_FMT("Called getter for {}", table.get_name());
+          LOG_TRACE_FMT("Called getter for {}", table.get_name());
```
jumaffre, line 489: Merge issue?

doc/schemas/app_openapi.json
achamayou: Not at the moment, no.

src/node/rpc/member_frontend.h
achamayou: Yes, sorry, bad merge. Fixed now.

src/node/rpc/member_frontend.h
achamayou: Yes, sorry, bad merge. Fixed now.

jumaffre (commented):
This needs a changelog entry and I think some warning in the documentation so that this feature doesn't disappear silently (even though we've warned users that this was deprecated!). It's also not clear if users can make use of the new COSE signatures via built-in authentication policies, in both C++ and JS?

jumaffre (approved):

PR #5140, opened 2023-03-28
Fix Threading pipeline - add locks in node-to-node channels

eddyashton:
The `CCF Github Threading` pipeline is currently consistently failing on the new e2e test introduced by #4986. The problem is unsafe concurrent access to the state inside a `Channel`. This PR adds a mutex inside `Channel`, claimed on each entry point, and moves all of the non-public methods to `private` visibility.

While verifying this fix with TSAN (not currently running any e2e+TSAN tests in the CI) I noticed a few other common threading issues. So with TSAN enabled, this is still not passing _all_ of the time, but now _passes_ some of the time.


src/node/node_state.h
```diff
@@ -841,9 +841,10 @@ namespace ccf
 
       auto timer_msg = std::make_unique<threading::Tmsg<NodeStateMsg>>(
         [](std::unique_ptr<threading::Tmsg<NodeStateMsg>> msg) {
+          std::lock_guard<pal::Mutex> guard(msg->data.self.lock);
           if (msg->data.self.sm.check(NodeStartupState::pending))
           {
-            msg->data.self.initiate_join();
+            msg->data.self.initiate_join_unsafe();
```
eddyashton, line 847: TSAN caught one failure here:

Without the lock, we first branch on `check(pending)`, and then inside `initiate_join` the first thing we do is `require(pending)`. If a join succeeds between the check and the require, the require fails and the node dies. This resolves that by locking the access. I'd be tempted to flatten the `check` into `initiate_join()`, but this fix is smaller.

src/node/snapshotter.h
```diff
@@ -25,7 +25,8 @@ namespace ccf
     static constexpr auto max_tx_interval = std::numeric_limits<size_t>::max();
 
   private:
-    ringbuffer::WriterPtr to_host;
+    ringbuffer::AbstractWriterFactory& writer_factory;
```
eddyashton, line 28: Another issue TSAN spotted - multiple threads call functions in this `Snapshotter` which may try to use `to_host`. The `WriterPtr to_host` is expected to be owned and accessed by a single thread, so that's unsafe. Instead they create a `Writer` on-demand.

jumaffre (approved):

achamayou (approved):

Issue #5084, opened 2023-03-07
Standardise governance API - remove use of member certs


eddyashton:
We primarily think of members as being identified purely by their _key_, which is used for signing. Some endpoints still permit TLS auth with the member cert as client identity (`POST /gov/recovery_share`, `GET /gov/recovery_share`, and `POST /gov/ack/update_state_digest`).

To support these, our member cert authentication does _not_ check expiry times (#5083), so it will permit expired member certs.

#5080 attempted to simplify this, removing the old endpoints and insisting on COSE signing. This was too aggressive a change to the API, and had too large a blast radius for an urgent CI fix, so has been reverted. We should revisit this, with deprecation of the old endpoints (redirecting to the new), and the goal of eventually dropping member cert-based auth entirely.

achamayou:
This was resolved as a side effect of #5137, with GETs no longer reliant on cert auth and using URL parameters only. Further changes are coming as part of the implementation of #5129, but this is done.

PR #5145, opened 2023-03-28
Expose encrypted shares unauthenticated

achamayou:
Resolve https://github.com/microsoft/CCF/pull/5137#discussion_r1147841052

- [x] fix schema test
- [x] update cadl
- [x] update `submit_recovery_share.sh`


eddyashton:
Linking to related issue: #5084

src/node/rpc/member_frontend.h
```diff
@@ -901,6 +901,54 @@ namespace ccf
...
+          auto rec_share = GetRecoveryShare::Out{
+            crypto::b64_from_raw(encrypted_share.value())};
+          ctx.rpc_ctx->set_response_header(
+            http::headers::CONTENT_TYPE, http::headervalues::contenttype::JSON);
+          ctx.rpc_ctx->set_response_body(nlohmann::json(rec_share).dump());
+          ctx.rpc_ctx->set_response_status(HTTP_STATUS_OK);
+          return;
+        };
+      make_endpoint(
+        "/recovery_share/{member_id}",
```
jumaffre, line 947:
```suggestion
        "/encrypted_recovery_share/{member_id}",
```

achamayou: Good idea, gets rid of the question about symmetry too.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 [4.0.0-dev7]: https://github.com/microsoft/CCF/releases/tag/ccf-4.0.0-dev7
 
+### Deprecated
+
+- `GET /gov/recovery_share` is deprecated in favour of the unauthenticated `GET /gov/encrypted_recovery_share/{member_id}`.
+- The `submit_recovery_share.sh` script now takes a `--member-id` argument.
```
eddyashton, line 15:
```suggestion
- The `submit_recovery_share.sh` script now takes a `--cert` argument.
```

Currently takes a `--cert` argument and computes the `member-id`.

python/utils/submit_recovery_share.sh
```diff
@@ -41,13 +41,34 @@ while [ "$1" != "" ]; do
...
+
+for item in "$@" ; do
+    if [ "$next_is_cert" == true ]; then
+        cert=$item
+        next_is_cert=false
+    fi
+    if [ "$item" == "--cert" ]; then
+        next_is_cert=true
+    fi
+done
```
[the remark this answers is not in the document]
achamayou: Seems harmless to pass it on to curl?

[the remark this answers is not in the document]
achamayou: It's a pain, that pass assumes the arguments are intercepted first, and everything else is handed over to curl. But --cert can come at any time. It's already broken if --help is passed later on.

I think this is fine for an illustrative sample, if we want a proper CLI we should find a library, expose a complete help etc.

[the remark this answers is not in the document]
eddyashton: Huh, I thought we had a non-destructive parser pattern but it looks like we don't. So yeah, parse like this and it gets passed on to curl.

python/utils/submit_recovery_share.sh
```diff
@@ -41,13 +41,34 @@ while [ "$1" != "" ]; do
...
+
 if [ -z "${member_enc_privk}" ]; then
     echo "Error: No member encryption private key in arguments (--member-enc-privk)"
     exit 1
 fi
 
+if [ -z "${cert}" ]; then
+    echo "Error: No user certificate in arguments (--cert)"
```
eddyashton, line 63: We should add `--cert` to the `usage()` printed line.

eddyashton (approved):

Issue #5057, opened 2023-02-23
Consensus trace validation


achamayou:
Following a suggestion from @lemmy, we could set up trace validation for the CCF consensus implementation.

- [x] review trace locations, and make sure all relevant events are emitted

```
../src/consensus/aft/impl/state.h
../src/consensus/aft/raft.h
../src/ds/state_machine.h
../src/node/history.h
../src/kv/store.h
../src/node/node_state.h
```

- [x] adjust output format to facilitate parsing
- [x] add log merge (sorted by ts), filter (remove host and other non-consensus events) & enrichment (node ids) script
- [x] script feeding logs to trace spec
- [x] write trace spec
- [x] hook up to CI & make stable
- [x] reconfiguration scenarios

lemmy:
As @eddyashton pointed out, to work around the limitations of timestamps, we want to initially focus on the sequentially executing raft_scenarios.

Below are the actions of the TLA+ specification and potentially matching log/code locations:

| TLA+ action    | Code location |
| -------------- | ----------- |
| Timeout           |       |
| RequestVote   | Text        |
| AppendEntries | |
| BecomeLeader ||
| ClientRequest | |
| SignCommittableMessages | |
| ChangeConfiguration | |
| AdvanceCommitIndex || 
| NotifyCommit | |
| CheckQuorum ||
| HandleRequestVoteRequest ||
| HandleRequestVoteResponse ||
| RejectAppendEntriesRequest ||
| ReturnToFollowerState ||
| AppendEntriesAlreadyDone ||
| ConflictAppendEntriesRequest ||
| NoConflictAppendEntriesRequest ||
| AcceptAppendEntriesRequest || 
| HandleAppendEntriesRequest ||
| HandleAppendEntriesResponse |  |
| UpdateTerm ||
DropStaleResponse

lemmy:
@eddyashton suggested that post-processing the existing logging with a Python script is a better approach because it can also be used in production. This raises the question of whether debug and trace log statements can be included in trace validation, as they may not be included in a production build.

lemmy:
In the `tests/raft_scenarios/4582.2` file, there is an issue where `raft.h` allows a `pending` node, named `n1`, to become a `follower` in response to receiving an `AppendEntries` message. However, this transition occurs even if `n1`'s log is not up-to-date and is missing the new configuration that includes `n1`. Consequently, `n1` responds to the message with an `AppendEntriesResponseType::FAIL`.

[...]

PR #5150, opened 2023-03-29
Allow sending empty AppendEntries message.

lemmy:
https://github.com/microsoft/CCF/issues/5057#issuecomment-1489280848


achamayou:
That's definitely necessary for heartbeats!

achamayou (approved):

PR #5151, opened 2023-03-30
Improve ledger durability when starting from an old snapshot

### LINKED ISSUE #4557: Persistence issue with `.ignored` ledger files
It is possible that we lose persistence of committed ledger entries in the following scenario:
1. A new node joined from a snapshot at seqno `N`, which is behind the latest ledger seqno at, say, `N+100`
2. Because of the currently CCF logic, the ledger files (possibly already `.committed`) between `N` and `N+100` are automatically marked as `.ignored` as the new node is expecting to replicate these ledger entries via Raft once it's successfully joined
3. However, before successful replication happens, the node is restarted by the operator in recovery mode, which means that all ledger entries between `N` and `N+100` are ignored by the recovery process, which is a loss of persistence. 

Possible solutions:
- Perform a mini-recovery of the ledger entries from `N` to `N+100` on the new node once the join response containing the ledger secrets has been received, but before initialising Raft. However, this may delay the time a new node added to the new configuration is able to participate in consensus, weakening quorum/fault tolerance. 
- Do not mark the ledger files later than snapshot at `N` as `.ignored`, and instead verify that the ledger entries the new node replicates from Raft are identical to these. As of the point in time this issue is created, this seems like the preferred option.
   - Until its addition to the nodes table has been committed, the new node would cache ledger writes on the host if they don't match what's already present on disk.
   - When the addition of the node is finally committed, these writes are flushed to disk and existing and conflicting ledger files are marked as `.ignored`.
- Similar to the recovery scheme, ledger files created by the new node until its addition is committed should be marked as `.catchup` or similar and ignored by the further join/recovery process. Once the node knows that its addition is committed, the old ledger files (between `N` and `N+100`) should be atomically marked as `.ignored` and the `.catchup` suffix should be discarded.

jumaffre:
Resolves #4557 

**Now ready for review**

Also contains change from #5156 

When joining from an old snapshot, the new joiner used to mark the ledger files following the snapshot as `.ignored` assuming that these would be replayed via consensus anyway (rightly so!). However, this introduced a window where the entries in those ledger files could be ignored completely by further recovery nodes if the primary did not replicate those entries (e.g. network partition between primary and new joiner). 

To fix this, we now keep the ledger files following the snapshot and only discard them if any divergence is detected. The new flow is as follows:
1. The new node joins from a snapshot at seqno `N`, `init`ing its ledger at `N`. The content of the subsequent ledger files remain untouched, but committed ledger files drop their `.committed` suffix.
2. When the ledger entries are `N` are replicated by the primary node to the new node, the node tries to find an existing ledger file that may contain this entry and if so, compares the entry to be written with the entry already present in the ledger file.
3. If the entry is the same as what's already there, nothing happens. When we get to the end of the ledger that we started from, we set the new `use_existing_files` to `false` so that the node no longer tries to find existing ledger files.
4. Otherwise, the current ledger file is truncated at the current seqno and the new entry is written. Subsequent ledger files are also deleted. 

TODO: 
- [x] End-to-end test


tests/recovery.py
```diff
@@ -760,14 +835,14 @@ def add(parser):
...
-    )
+    # cr.add(
+    #     "recovery_corrupt_ledger",
+    #     run_corrupted_ledger,
+    #     package="samples/apps/logging/liblogging",
+    #     nodes=infra.e2e_args.min_nodes(cr.args, f=0),  # 1 node suffices for recovery
+    #     sig_ms_interval=1000,
+    #     ledger_chunk_bytes="1GB",
+    #     snapshot_tx_interval=1000000,
+    # )
```
eddyashton, line 846: This should be restored before committing.

eddyashton (approved):

achamayou (approved):

PR #5152, opened 2023-03-30
HTTP/2: Fix max header list hint

jumaffre:
~This optional hint actually ended up breaking some gRPC clients~, e.g. `etcdctl`. @jeffa5 

I had misread this configuration entry, which is actually a size in bytes, rather than a number of headers:

https://www.rfc-editor.org/rfc/rfc7540#section-6.5.2:
>   SETTINGS_MAX_HEADER_LIST_SIZE (0x6):  This advisory setting informs a
      peer of the maximum size of header list that the sender is
      prepared to accept, **in octets**.  The value is based on the
      uncompressed size of header fields, including the length of the
      name and value in octets plus an overhead of 32 octets for each
      header field.


eddyashton (approved):

jeffa5:
Works on my end, thanks!

jeffa5:
That fix also works.

PR #5154, opened 2023-03-31
Model AppendEntries RPC with more than one entry s.t. the leader can …

lemmy:
…send log entries to the follower in batches.

Note that Ongardie's [initial spec](https://github.com/ongardie/raft.tla/commit/2372317dfcac9975afa95b04cfea526b09f382d0) was limited to one entry per RPC.

### DIFF
--- a/tla/ccfraft.tla
+++ b/tla/ccfraft.tla
@@ -346,6 +346,9 @@ CommittedTermPrefix(i, x) ==
     \* Otherwise the prefix is the empty tuple
     ELSE << >>
 
+AppendEntriesBatchsize(i, j) ==
+    {nextIndex[i][j]}
+
 ----
 
 \* Define initial values for all variables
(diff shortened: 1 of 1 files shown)


heidihoward (approved):

lemmy:
Follow-up:  Adjust probabilities in `SIMccfraft.tla` to reduce likelihood of *empty* `AppendEntry` messages.

heidihoward:
Yep, the exhaustive checking in ``MCccfraft.tla`` and ``MCccfraftWithReconfig.tla`` also need to updating to prevent state space explosion

lemmy:
> Yep, the exhaustive checking in `MCccfraft.tla` and `MCccfraftWithReconfig.tla` also need to updating to prevent state space explosion

`tests/raft_scenarios/bad_network` taught me yesterday that we have to refine/redefine the variable `messages` to be a multi-set (bag).  Let me see if we can role this one here into this multi-set change. If possible, we might be able to keep the current version of ccfraft with a single entry per message.

tla/ccfraft.tla
```diff
@@ -458,25 +458,25 @@ AppendEntries(i, j) ==
...
        IN
-       /\ InMessagesLimit(i, j, index)
        /\ messagesSent' =
-            IF Len(messagesSent[i][j]) < index
+            IF Len(messagesSent[i][j]) < minIndex
             THEN [messagesSent EXCEPT ![i][j] = Append(messagesSent[i][j], 1) ]
-            ELSE [messagesSent EXCEPT ![i][j][index] = messagesSent[i][j][index] + 1 ]
-       /\ Send(msg)
+            ELSE [messagesSent EXCEPT ![i][j][minIndex] = messagesSent[i][j][minIndex] + 1 ]
+       \* TODO aft::send_append_entries (raft.h) says is only ever sends entries from a single term!
```
achamayou, line 476: See https://github.com/microsoft/CCF/pull/3145 for details. In short, because raft decides whether the accept the AE or not based on its header, the header needs to contain authentic information about the term history it covers. To keep the AE header constant size and compatible, we can only use a single term_of_index and therefore a single term.

Because large amounts of short consecutive terms do not typically happen, or do not coincide with high levels of throughput when they do (an unstable network with lots of elections will not manage many writes), the performance impact is thought to be of little practical concern.

achamayou (approved):

tla/ccfraft.tla
lemmy: Are you saying that the the single term limitation is an implementation detail that the spec can ignore?

lemmy:
702fc882dbaf1336ccbcb01a9282c533885a5ec5 is now just a syntactical change that enables trace validation to relax the upper bound of mentries per AppendEntries message. The commit does not change semantics of ccfraft.tla, i.e., number of states for the existing models.

tla/ccfraft.tla
achamayou: I was just explaining why the implementation behaved that way. Unless there's a benefit to the spec ignoring that, I would think it's best to keep it as close to implementation behaviour as possible.

PR #5162, opened 2023-04-04
Infra: increase timeout for historical range

jumaffre:
This PR bumps the timeout for the verification of historical entry ranges, which is currently failing on the `Virtual Instrumented` build (e.g. https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=67982&view=logs&j=fff884fa-70f7-5142-c642-c9647ff4218c&t=002ec585-ff09-575c-f94f-67274e66fb5c&l=7985).


eddyashton (changes requested):
I think this is only failing in this specific test, so I suspect there's something else going on.

Do you have a local repro, and evidence this reduces the repro rate?

jumaffre:
> I think this is only failing in this specific test, so I suspect there's something else going on.
> 
> Do you have a local repro, and evidence this reduces the repro rate?

I have a local repro. It only fails for this test in the daily build because `--with-load` is passed to the recovery test, which causes the service to be at seqno ~5000 before the recovery is started. The virtual instrumented build seems to be too slow to recover all these entries in 6 seconds, hence why I changed the rate.

I will print the time it takes to read all historical entries so that we know we were close to the previous 6 second limit.

Issue #4472, opened 2022-11-04
Invariant violation for `MoreUpToDateCorrectInv` in TLA+ spec


lemmy:
Suffix of TLA+ counterexample to `MoreUpToDateCorrectInv` (see attachment for complete counterexample of 120 states):

```tla
TLC2 Version 2.18 of Day Month 20?? (rev: f85f396)
Running Random Simulation with seed -1964463959898986002 with 10 workers on 10 cores with 7282MB heap and 64MB offheap memory [pid: 42201] (Mac OS X 13.0 aarch64, Homebrew 11.0.16.1 x86_64).
Parsing file /Users/markus/src/TLA/_specs/MSFT/CCF/tla/SIMccfraft.tla
Parsing file /Users/markus/src/TLA/_specs/MSFT/CCF/tla/ccfraft.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/tlc-1096372472810769223/TLC.tla (jar:file:/Applications/TLA+%20Toolbox.app/Contents/Eclipse/tla2tools.jar!/tla2sany/StandardModules/TLC.tla)
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/tlc-1096372472810769223/Integers.tla (jar:file:/Applications/TLA+%20Toolbox.app/Contents/Eclipse/tla2tools.jar!/tla2sany/StandardModules/Integers.tla)
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/SIMPostCondition.tla (/Users/markus/src/TLA/_specs/MSFT/CCF/tla/SIMccfraft.tla)
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/tlc-1096372472810769223/_TLCTrace.tla (jar:file:/Applications/TLA+%20Toolbox.app/Contents/Eclipse/tla2tools.jar!/tla2sany/StandardModules/_TLCTrace.tla)
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/tlc-1096372472810769223/Naturals.tla (jar:file:/Applications/TLA+%20Toolbox.app/Contents/Eclipse/tla2tools.jar!/tla2sany/StandardModules/Naturals.tla)
[...]

PR #5168, opened 2023-04-11
TLA+ spec - each reconfiguration must change the set of servers

heidihoward:
Spotted whilst looking at the trace in https://github.com/microsoft/CCF/issues/4472#issuecomment-1494733529. This patch ensures that a leader cannot create a reconfiguration txn with the same set of servers as the last reconfiguration txn, even if this previous txn is not yet committed.

### DIFF
--- a/tla/ccfraft.tla
+++ b/tla/ccfraft.tla
@@ -564,6 +564,8 @@ ChangeConfiguration(i, newConfiguration) ==
     /\ newConfiguration \subseteq Servers
     \* Configuration is not equal to current configuration
     /\ newConfiguration /= CurrentConfiguration(i)
+    \* Configuration is not equal to the previous pending configuration
+    /\ newConfiguration /= MaxConfiguration(i)
     \* Keep track of running reconfigurations to limit state space
     /\ reconfigurationCount' = reconfigurationCount + 1
     /\ removedFromConfiguration' = removedFromConfiguration \cup (CurrentConfiguration(i) \ newConfiguration)


heidihoward:
I also find the `current`, `next`, `max` configuration distinction confusing.

Not quite, what I was trying to express is:
```tla
\A idx, idy \in DOMAIN configurations[n]: 
  /\ idx # idy 
  /\ configuration[n][idx] = configuration[n][idy] 
  => \E idz \in DOMAIN configurations[n]: idx < idz < idy /\ configuration[n][idx] # configuration[n][idz] 
```

tla/ccfraft.tla
```diff
@@ -1085,8 +1085,9 @@ LogTypeOK(xlog) ==
 ReconfigurationVarsTypeInv ==
     /\ reconfigurationCount \in Nat
     /\ \A i \in Servers : 
-        \A c \in DOMAIN configurations[i]:
+        /\ \A c \in DOMAIN configurations[i] :
             configurations[i][c] \subseteq Servers
+        /\ Cardinality(DOMAIN configurations[i]) # 0
```
lemmy, line 1090: DOMAIN configurations[i] # {}

lemmy:
Representing a configuration by a natural number, is, e.g., `<<1, 2, 1>>` not a valid sequence of configurations?

tla/ccfraft.tla
```diff
@@ -1159,11 +1169,14 @@ MonoLogInv ==
                 \/ /\ log[i][k].term < log[i][k+1].term
                    /\ log[i][k].contentType = TypeSignature
 
-
+\* Each server's active configurations should to consistent with its log
```
achamayou, line 1172:
```suggestion
\* Each server's active configurations should be consistent with its log
```

achamayou (approved):

heidihoward:
> Representing a configuration by a number, is, e.g., `<<1, 2, 1>>` an invalid sequence of configurations?

Thanks, I didn't think about it like that. I think the only issue is that in the current spec the configurations can index 0

lemmy:
@heidihoward https://github.com/tlaplus/CommunityModules/blob/master/modules/Functions.tla#L74-L81 might be useful in the definition of invariants related to `configurations`.

PR #5179, opened 2023-04-14
Remove experimental 2tx support from consensus

achamayou:
We are not planning to support multiple re-configuration mechanisms in the foreseeable future. Removing 2tx will shorten build and test time and reduce code size.


src/consensus/aft/raft.h
```diff
@@ -2316,103 +2157,14 @@ namespace aft
...
-            RAFT_TRACE_FMT(
-              "Configurations: not enough trusted or retired nodes for "
-              "configuration #{} ({}/{} trusted, {}/{} retired)",
-              next->rid,
-              num_trusted_nodes,
-              next->nodes.size(),
-              num_retired_nodes,
-              num_required_retired_nodes);
-            if (
-              node_client && !is_learner() &&
```
jumaffre, line 2406: I think we can remove the `node_client` member from raft too, as well as all mentions of the learner state?

jumaffre (approved):

src/consensus/aft/raft.h
achamayou: The `node_client` is used to by the RetiredNodeCleanup, and so cannot be removed.

PR #5173, opened 2023-04-13
Update reconfiguration logic in TLA+ specification

heidihoward:
Following up on https://github.com/microsoft/CCF/pull/5172 and https://github.com/microsoft/CCF/pull/5168, this PR attempts to clarify the purpose of `configurations` and ensure that `configurations` is correctly updated each time that `log` or `commitIndex` is updated

PR #5182, opened 2023-04-17
Trying some bonus properties

heidihoward:
Some simple extra properties for the TLA+ spec. I've not yet looked at the performance overhead of checking these properties, might be worth disabling by default.


lemmy:
Does this PR depend on PR #5173?

heidihoward:
> Does this PR depend on PR #5173?

Nope, they should be distinct. These properties should hold before and after https://github.com/microsoft/CCF/pull/5173

tla/ccfraft.tla
```diff
@@ -1178,8 +1178,36 @@ LogConfigurationConsistentInv ==
...
+CommittedLogNeverChangesProp ==
+    [][\A i \in Servers :
+        IsPrefix(Committed(i), Committed(i)')]_vars
+
+LeaderLogAppendOnlyProp ==
+    [][\A i \in { i \in Servers : state[i] = Leader /\ state[i]' = Leader } : 
+        IsPrefix(log[i], log[i]')]_vars
+
+CandidateLogUnchangedProp ==
+    [][\A i \in { i \in Servers : state[i] = Candidate /\ state[i]' = Candidate } :
```
lemmy, line 1201: `\in {Candidate, Pending}` or event `\in {Candidate, Pending, RetiredLeader}?

lemmy: More generalize, under what condition is a log allowed to change s.t. `log[i] # log'[i] /\ ~IsPrefix(log[i], log'[i])`?

lemmy:
The newly introduced properties offer valuable clarification. I recommend generalizing `CandidateLogUnchangedProp` and `LeaderLogAppendOnlyProp` into a property that states when a log is permitted to be truncated, i.e., altered in a non-append-only manner.

lemmy (approved):

tla/ccfraft.tla
achamayou: > `\in {Candidate, Pending}` or event `\in {Candidate, Pending, RetiredLeader}`?

Also `RetiredLeader`, and even something a bit stronger for which we don't have a state in the spec, a hypothetical `RetiredCommittableLeader` (ie. when RCI [is set](https://microsoft.github.io/CCF/main/architecture/consensus/1tx-reconfig.html#retirement-details)), which is when a retiring leader has issued a signature following its retirement transaction. The implementation does not allow a leader to issue further writes after that.

heidihoward:
> The newly introduced properties offer valuable clarification. I recommend generalizing `CandidateLogUnchangedProp` and `LeaderLogAppendOnlyProp` into a property that states when a log is permitted to be truncated, i.e., altered in a non-append-only manner.

I've provisionally added the property below which combines and strengthens  `CandidateLogUnchangedProp` and `LeaderLogAppendOnlyProp` 

https://github.com/microsoft/CCF/blob/b9ba3c7e1536525eb5c14a3c807d300b6f48cb88/tla/ccfraft.tla#L1202-L1215

I'll have a go at strengthening it further to restrict permitted log changes to pending/follower nodes

tla/ccfraft.tla
heidihoward: @achamayou It would be nice to extend the spec in this way but probably out-of-scope for this PR

lemmy:
LGTM

PR #5165, opened 2023-04-06
Upgrade to OE 0.19.0-rc1

eddyashton:
TODO:
- [x] Build new CI image (https://github.com/microsoft/CCF/actions/runs/4678101373)
- [x] Update references to use new CI image 
- [x] Check test results
- [x] Describe upgrade in CHANGELOG

PR #5186, opened 2023-04-18
Update snmalloc to `0.6.0`

jumaffre:
Following from #5165, this PR updates the snmalloc dependency to `0.6.0`, in sync with what's provided via Open Enclave 0.19.0.

Since `snmalloc` no longers provides `DLList`, this PR also adds our own implementation in `src/ds` (along with unit tests), similar to the snmalloc implementation as discussed with @mjp41.


cmake/common.cmake
```diff
@@ -245,14 +245,11 @@ if(SAN
...
    OR NOT USE_SNMALLOC
 )
-  set(SNMALLOC_LIB)
+  set(SNMALLOC_COMPILE_OPTIONS "")
 else()
-  set(SNMALLOC_ONLY_HEADER_LIBRARY ON)
-  # Remove the following two lines once we upgrade to snmalloc 0.5.4
-  set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
-  set(USE_POSIX_COMMIT_CHECKS off)
+  set(SNMALLOC_HEADER_ONLY_LIBRARY ON)
```
jumaffre, line 250: This flag used to be set incorrectly (`SNMALLOC_ONLY_HEADER_LIBRARY` rather than `SNMALLOC_HEADER_ONLY_LIBRARY`). This is now fixed so we no longer generate `snmalloc_lib`.

src/node/request_tracker.h
```diff
@@ -1,233 +0,0 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
```
jumaffre, line 1: This file was unused.

src/host/snmalloc.cpp
```diff
@@ -7,5 +7,4 @@
 #  define NDEBUG
 #endif
 
-#include "snmalloc/src/override/malloc.cc"
-#include "snmalloc/src/override/new.cc"
```
mjp41, line 11: I think it would be better to keep 
```
#include "snmalloc/src/override/new.cc"
```
and drop
```
#include "snmalloc/src/override/malloc.cc"
```
The `new.cc` file includes `malloc.cc`, but also provides overrides for the `new` and `delete` operations, so they can get memory with one fewer function call.

achamayou (approved):

PR #5188, opened 2023-04-19
Draft: Swap usage of snmalloc `DLList` with `std::list`

jumaffre:
Isolate subset of changes in #5186 to measure impact of changing underlying doubly-linked list implementation used by the KV's map commit roll.


jumaffre:
Closing this as using a `std::list` has a significant impact on memory usage in most end-to-end tests: 
image

PR #5191, opened 2023-04-19
Fix field formats that are incompatible with OpenAPI

achamayou:
Spotted by:

image

See https://swagger.io/docs/specification/data-models/data-types/


typespec-ccf/main.tsp
```diff
@@ -339,231 +339,274 @@ model Sha256DigestToString {
...
 @summary("A member's encrypted recovery share")
 @get op getEncryptedRecoveryShare(@path member_id: string): {
+    @header("x-ms-ccf-transaction-id") transactionID: string;
     @body recoveryShare: GetEncryptedRecoveryShare;
 };
 
 @route("/gov/tx")
 @doc("Possible statuses returned are Unknown, Pending, Committed or Invalid.")
 @get op getTx(@query @pattern("^[0-9]+\\.[0-9]+$")transactionID: string): {
+    @header("x-ms-ccf-transaction-id") transactionID: string;
```
eddyashton, line 610:
```suggestion
```

`/tx` and `/receipt` are commands, do not execute transactionally over a versioned KV snapshot, and do not return a `transaction-id` header.

eddyashton:
Force-merging to upstream - isn't actually used by any code in this repo, so shouldn't break any tests.

PR #5194, opened 2023-04-20
Fix container image generation for `4.0.0-rc2`

jumaffre:
The `app_run` playbook relied on the `open-enclave` debian package being published to https://ubuntu.pkgs.org/20.04/microsoft-prod-amd64/ (then automatically installed via `apt-get install ccf_*`) but the package isn't published for Open Enclave release candidates. We now install `open-enclave` (or `open-enclave-hostverify`) via GitHub in the `app_run` playbook.

### DIFF
--- a/cmake/preproject.cmake
+++ b/cmake/preproject.cmake
@@ -11,8 +11,13 @@ if((NOT CMAKE_C_COMPILER)
    AND "$ENV{CC}" STREQUAL ""
    AND "$ENV{CXX}" STREQUAL ""
 )
-  find_program(FOUND_CMAKE_C_COMPILER NAMES clang-15 clang-11)
-  find_program(FOUND_CMAKE_CXX_COMPILER NAMES clang++-15 clang++-11)
+  if("${COMPILE_TARGET}" STREQUAL "sgx")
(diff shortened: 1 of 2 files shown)


cmake/preproject.cmake
```diff
@@ -11,8 +11,13 @@ if((NOT CMAKE_C_COMPILER)
...
 )
-  find_program(FOUND_CMAKE_C_COMPILER NAMES clang-15 clang-11)
-  find_program(FOUND_CMAKE_CXX_COMPILER NAMES clang++-15 clang++-11)
+  if("${COMPILE_TARGET}" STREQUAL "sgx")
+    find_program(FOUND_CMAKE_C_COMPILER NAMES clang-11)
+    find_program(FOUND_CMAKE_CXX_COMPILER NAMES clang++-11)
+  else()
+    find_program(FOUND_CMAKE_C_COMPILER NAMES clang-15)
+    find_program(FOUND_CMAKE_CXX_COMPILER NAMES clang++-15)
+  endif()
```
jumaffre, line 20: Separate change that feels necessary. Otherwise `CC` and `CXX` have to be specified to `cmake` when building CCF from source.

achamayou (approved):

PR #5197, opened 2023-04-20
Spring clean of `ccfraft.tla`

heidihoward:
Various minor refactors to make the spec easier to maintain.


lemmy (approved):

tla/ccfraft.tla
```diff
@@ -121,98 +132,248 @@ VARIABLE configurations
...
+    /\ entry.term \in Nat \ {0}
+    /\ \/ /\ entry.contentType = TypeEntry
+          /\ entry.value \in Nat \ {0}
+       \/ /\ entry.contentType = TypeSignature
+          /\ entry.value = Nil
+       \/ /\ entry.contentType = TypeReconfiguration
+          /\ entry.value \subseteq Servers
+
+AppendEntriesRequestTypeOK(m) ==
+    /\ m.mtype = AppendEntriesRequest
```
achamayou, line 166: If we're doing some spring cleaning, can we get rid of all the (seemingly unnecessary) "m" prefixes?

tla/ccfraft.tla
```diff
@@ -121,98 +132,248 @@ VARIABLE configurations
...
+    /\ m.mprevLogIndex \in Nat
+    /\ m.mprevLogTerm \in Nat
+    /\ m.mcommitIndex \in Nat
+    /\ \A k \in DOMAIN m.mentries: EntryTypeOK(m.mentries[k])
+
+AppendEntriesResponseTypeOK(m) ==
+    /\ m.mtype = AppendEntriesResponse
+    /\ m.msuccess \in BOOLEAN
+    /\ m.mmatchIndex \in Nat
```
achamayou, line 175: Should we try to line up with implementation names? This is called `last_log_idx` in the code.

tla/ccfraft.tla
```diff
@@ -121,98 +132,248 @@ VARIABLE configurations
...
+RequestVoteRequestTypeOK(m) ==
+    /\ m.mtype = RequestVoteRequest
+    /\ m.mlastLogTerm \in Nat
+    /\ m.mlastLogIndex \in Nat
+
+RequestVoteResponseTypeOK(m) ==
+    /\ m.mtype = RequestVoteResponse
+    /\ m.mvoteGranted \in BOOLEAN
+
+NotifyCommitMessageTypeOK(m) ==
```
achamayou, line 186: FWIW there isn't a distinct message type for this in the implementation, this happens through AppendEntries.leader_commit_idx

tla/ccfraft.tla
```diff
@@ -121,98 +132,248 @@ VARIABLE configurations
...
+        /\ m.mdest \in Servers
+        /\ m.mterm \in Nat \ {0}
+        /\ \/ AppendEntriesRequestTypeOK(m)
+            \/ AppendEntriesResponseTypeOK(m)
+            \/ RequestVoteRequestTypeOK(m)
+            \/ RequestVoteResponseTypeOK(m)
+            \/ NotifyCommitMessageTypeOK(m)
+
 \* CCF: Keep track of each append entries message sent from each server to each other server
 \* and cap it to a maximum to constraint the state-space for model-checking.
```
achamayou, line 202:
```suggestion
\* and cap it to a maximum to constrain the state-space for model-checking.
```

tla/ccfraft.tla
```diff
@@ -121,98 +132,248 @@ VARIABLE configurations
...
+
+
+------------------------------------------------------------------------------
 \* The following variables are all per server (functions with domain Servers).
 
 \* The server's term number.
 VARIABLE currentTerm
 
+CurrentTermTypeInv ==
+    \A i \in Servers : currentTerm[i] \in Nat
```
achamayou, line 241:
```suggestion
    \A i \in Servers : currentTerm[i] \in Nat \ {0, 1, 2}
```
if we want to match the implementation, which I think we do since we exclude 0 for seqno/index.

achamayou (approved):

tla/ccfraft.tla
heidihoward: Sure, I don't see why not. See 3b39e58097cc32d5ba0b0fb68ff14bfe2f827ccf

tla/ccfraft.tla
heidihoward: Currently in the spec each node's `currentTerm` starts at 0 so this isn't true. 

https://github.com/microsoft/CCF/blob/9f8833d41abd47650b3f771f1b7e92fafedfa436/tla/ccfraft.tla#L392-L395

Not quite sure I understand what you mean about excluding 0 for seqno/index

tla/ccfraft.tla
heidihoward: `last_log_idx` and `matchIndex` aren't quite the same so I wonder if its more confusing the use the same name of different things

tla/ccfraft.tla
```diff
@@ -121,98 +132,248 @@ VARIABLE configurations
...
 \* to another. With CCF, we have message integrity and can ensure unique messages.
 \* Messages only records messages that are currently in-flight, actions should
 \* removed messages once received.
 VARIABLE messages
 
+\* Helper function for checking the type safety of log entries
+EntryTypeOK(entry) ==
+    /\ entry.term \in Nat \ {0}
+    /\ \/ /\ entry.contentType = TypeEntry
+          /\ entry.value \in Nat \ {0}
```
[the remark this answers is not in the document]
heidihoward: This is the dummy value for the user transactions, we literally just use an integer n to represent the nth user transaction

PS: I didn't introduce the type invariants in the PR, it just moved them so each was next to the associated variable definition

tla/ccfraft.tla
heidihoward: Yes, thanks for the reminder. I think this will be a fairly big change so out-of-scope for this PR

tla/ccfraft.tla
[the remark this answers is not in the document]
achamayou: Sounds good, s/value/userTx/g or something similar perhaps?

tla/ccfraft.tla
achamayou: I am perhaps confused here, but assuming this describes a field in AppendEntriesResponse, it is called `last_log_idx` in the implementation. `match_idx` in the implementation is something the primary maintains per node, and you're right that it isn't always identical to the latest `last_log_idx` returned by a backup.

Are you saying that the spec AERs compute `matchIdx` and return that rather than `last_log_idx`? If so, that sounds like it might complicate trace validation.

tla/ccfraft.tla
achamayou: @heidihoward my bad, I confused `entry.value` with `index`. At line 194 though, we seemingly exclude 0 for term:

```
MessagesTypeInv ==
    \A m \in messages :
        /\ m.source \in Servers
        /\ m.dest \in Servers
        /\ m.term \in Nat \ {0}
``

tla/ccfraft.tla
lemmy: Related to `matchIndex` and `last_log_idx`: https://github.com/microsoft/CCF/issues/5057#issuecomment-1512420114

tla/ccfraft.tla
lemmy: The existing trace validation scenarios confirm that there is never a message with term = 0.

tla/ccfraft.tla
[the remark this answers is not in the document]
heidihoward: d5b05a4ceb29479f309726f7d06039b410b056fc does the following
- renamed `value` to `request` in `TypeEntry`, 
- removed `value` from `TypeSignature` (it was always Nil) ,and 
- renamed `value` to `configuration` in `TypeReconfiguration`

Is that more clear?

tla/ccfraft.tla
heidihoward: Since all nodes start in term 0 as followers, the first candidate is in term 1 so there should never be a message with term 0

lemmy: Maybe somewhere add `\A s \in Servers: currentTerm[s] = 0 => state[s] # Leader`?

tla/ccfraft.tla
heidihoward: Yes, it describes the field in `AppendEntriesResponse` (and yep, there is also a `matchIndex` in the node state). In the spec, the `matchIndex` in AERs is unused for NACKs. The ccf code seems to do something different for both AERs and for node state (where it can go backwards?). Also the term in AERs also seems to have a different meaning from the spec. For now, I'll rename as suggested and circle back to this later

tla/ccfraft.tla
heidihoward: Done in 147fed0bb023e075550b37ff9cdbebdbf132554a

tla/ccfraft.tla
heidihoward: Done in 80438e7cac02d313c4798c6e9902f531b9b088e3

lemmy: In a similar vein, `mcommitIdx` , `mprevLogTerm`, and `mprevLogIndex` of `AppendEntriesRequest` could be renamed to `leaderCommitIndex`, `prevTerm`, and `prevIndex`.

PR #5201, opened 2023-04-22
Raft tracing

achamayou:
Moves changes from https://github.com/lemmy/CCF/tree/mku-gh5057 to `main`, to make progress on #5057. Thank you @lemmy for doing all the work on this!

Introduce a `-DCCF_RAFT_TRACING=ON`, wired to an ifdef, to make sure there is no impact on release builds. This will require a separate sub-job in the CI at first, probably merged into the Debug build if the performance impact does not lead to timeouts on end to end tests.

Next steps:

- [x] Consolidate leadership and membership into state #5203
- [x] Emit configurations in add_configuration and become_* functions
- [x] Comparison pass with https://github.com/microsoft/CCF/compare/main...lemmy:CCF:mku-gh5057 to confirm all events are covered
- [x] Slight tangent to clean up existing states #5207
- [x] Adjust field names in trace spec as necessary
- [x] CI should update traces every time

PR #5202, opened 2023-04-22
Remove unused code and fields

achamayou:
Spotted while working on #5201


eddyashton (approved):

heidihoward:
Sorry I didn't get a chance to review before merging. A few super minor points:

1. Can `learner_nodes` definition and checks also be removed?

https://github.com/microsoft/CCF/blob/1c72bdba1aa15f868a9b50017732ebc3e9b83f1c/src/consensus/aft/raft.h#L170

2. I assume `reconfiguration_type` cannot be removed. Could we add a comment to say that only 1tx is supported and its maintained for compatibility. 

4. Similarly for `MembershipState::Learner` which I think is no longer used

5. According to the docs `MembershipState::RetirementInitiated` isn't used by 1tx but looking at raft.h it seems that it is still used?

6. Could this conditional branch be removed or turned into an assertion? 
https://github.com/microsoft/CCF/blob/1c72bdba1aa15f868a9b50017732ebc3e9b83f1c/src/consensus/aft/raft.h#L785

7. Same for
https://github.com/microsoft/CCF/blob/1c72bdba1aa15f868a9b50017732ebc3e9b83f1c/src/consensus/aft/raft.h#L1311-L1315

8. and 
https://github.com/microsoft/CCF/blob/1c72bdba1aa15f868a9b50017732ebc3e9b83f1c/src/consensus/aft/raft.h#L1691

9. Could the following ifs be combined since an AE response can now only be OK or FAIL?
https://github.com/microsoft/CCF/blob/1c72bdba1aa15f868a9b50017732ebc3e9b83f1c/src/consensus/aft/raft.h#L1416-L1443

10. The default constitution still talks about 2tx in a few places such as https://github.com/microsoft/CCF/blob/1c72bdba1aa15f868a9b50017732ebc3e9b83f1c/samples/constitutions/default/actions.js#L1163

11. More code which probably be cut:
https://github.com/microsoft/CCF/blob/1c72bdba1aa15f868a9b50017732ebc3e9b83f1c/src/node/node_state.h#L1045-L1052

PR #5203, opened 2023-04-22
Consolidate raft state

achamayou:
This only moves `leadership_state` and `membership_state` to raft state, to unblock #5201. Also removes the `my_` suffix from `node_id`.

I think we should move the other state leftover across too, but there's enough going on as it is.


jumaffre (approved):

heidihoward:
Sorry this is probably a very stupid question but where did the definition of `leadership_state` and `membership_state` go? I cannot see them in state.h

achamayou:
> Found them, I think they were removed in the merge [eb06ff8](https://github.com/microsoft/CCF/commit/eb06ff859591642c97e16ddfaa88dc2956e72f83)

Yes, my bad, should be fixed now.

PR #5205, opened 2023-04-24
Format codebase with `clang-format-11`

### LINKED ISSUE #5181: Change codebase format to `clang-format-11`
We'll probably have to bite the bullet and do a reformat with 11, we don't want to stay stuck on format-10 forever.

_Originally posted by @achamayou in https://github.com/microsoft/CCF/pull/5165#discussion_r1164589205_

jumaffre:
Resolves #5181


tests/infra/github.py
```diff
@@ -219,7 +219,7 @@ def get_latest_dev_tag(self):
 
     def get_tags_for_major_version(self, major_version=None):
         version_re = f"{major_version}\\." if major_version else ""
-        tag_re = f"^{TAG_RELEASE_PREFIX}{version_re}([.\\d+]+)(-rc.*|)$"
+        tag_re = f"^{TAG_RELEASE_PREFIX}{version_re}([.\\d+]+)$"
```
jumaffre, line 222: Bundled with this PR, we also stop testing compatibility with release candidates as this was overly precautious, and we may break compatibility between release candidate and final release on purpose (just like we are doing for 4.0.0).

achamayou (approved):

tests/infra/github.py
achamayou: I think it'd be better for clarity if this was split into its own PR.

PR #5209, opened 2023-04-24
Fix for `MoreUpToDateCorrectInv`

heidihoward:
Updating `MoreUpToDateCorrectInv` to take account of the reconfiguration logic


tla/ccfraft.tla
```diff
@@ -1202,16 +1202,22 @@ QuorumLogInv ==
...
 \* before issuing a vote implies that i receives
 \* a vote from j only if i has all of j's committed
 \* entries
 MoreUpToDateCorrectInv ==
-    \A i, j \in Servers : i /= j =>
-        ((\/ MaxCommittableTerm(log[i]) > MaxCommittableTerm(log[j])
-         \/ /\ MaxCommittableTerm(log[i]) = MaxCommittableTerm(log[j])
-            /\ MaxCommittableIndex(log[i]) >= MaxCommittableIndex(log[j])) =>
-        IsPrefix(Committed(j), log[i]))
+    \A i \in { s \in Servers : state[s] = Pending } :
```
lemmy, line 1216: `MoreUpToDateCorrectInv` will be trivially true after all servers transitions out of the `pending` state.

heidihoward: My mistake, there’s a missing \not before the \in

lemmy: `\notin`?

heidihoward: I switched to candidate instead of not pending, e37c664651c8049ad06b0fe2d1959f074b4d3388

tla/ccfraft.tla
```diff
@@ -1202,16 +1202,22 @@ QuorumLogInv ==
             \E j \in S :
                 IsPrefix(Committed(i), log[j])
 
-\* The "up-to-date" check performed by servers
-\* before issuing a vote implies that i receives
-\* a vote from j only if i has all of j's committed
-\* entries
+\* True if server i could receive a vote from server j based on the up-to-date check
+\* The "up-to-date" check performed by servers before issuing a vote implies that i receives
+\* a vote from j only if i has all of j's committed entries
```
achamayou, line 1207:
```suggestion
\* a vote from j only if i has all of j's committable entries
\* See https://github.com/microsoft/CCF/blob/main/src/consensus/aft/raft.h#L1512
```

achamayou: Ok, that makes sense. However a follower wouldn't vote for a candidate that has all of the follower's committed entries, but is missing some of that follower's committable entries in the highest term. So this isn't quite as strong as the condition in the implementation. Not complaining in any way, just checking my understanding :)

PR #5218, opened 2023-04-26
Enable storage account

achamayou:
Re-enable azureFile storage in ACI for benchmarking purposes. Not intended for merging.


achamayou:
This investigation is complete, the time impact is mostly caused by file copies being very slow on the file mount (>10s/Gb).

PR #5219, opened 2023-04-26
Initial commit of new gov API spec

eddyashton:
Based on Azure Core Typespec library and standards.

This is still a work-in-progress, committing to lock-in some design points and allow parallel development.


typespec-ccf/proposals.tsp
```diff
@@ -0,0 +1,203 @@
...
+  @TypeSpec.Http.header("Content-type")
+  contentType: "application/cose";
+
+  @doc("A raw CoseSign1 signature. The required protected headers and content schema will vary by action.")
+  @TypeSpec.Http.body
+  body: bytes;
+}
+
+interface Proposals {
+  @doc("Submit a proposed change to the service. This will be assigned an ID by the service. Idempotence is based on the signed proposal body - submitting identical signed proposals multiple times will only create a single proposal on the server, unless sufficient time has passed (since the `created_at` signed COSE header), in which case the submission is stale and will return an error.")
```
lynshi, line 133: Wonder if we can clarify this somewhat. It is not obviously apparent that the exact same _proposal_ is not actually identical when the signing timestamps are different.

lynshi: e.g. `submitting identical signed proposals (including timestamp headers)`

typespec-ccf/tsp-output/@azure-tools/typespec-autorest/0.0.1-preview/openapi.json
```diff
@@ -0,0 +1,923 @@
...
+        "description": "Submit a proposed change to the service. This will be assigned an ID by the service. Idempotence is based on the signed proposal body - submitting identical signed proposals multiple times will only create a single proposal on the server, unless sufficient time has passed (since the `created_at` signed COSE header), in which case the submission is stale and will return an error.",
+        "parameters": [
+          {
+            "$ref": "#/parameters/Azure.Core.Foundations.ApiVersionParameter"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "description": "A raw CoseSign1 signature. The required protected headers and content schema will vary by action.",
```
lynshi, line 317: How do users know what the shape of the stuff in the COSE looks like? If it can't be documented in the Typespec, we should at least link to CCF documentation

typespec-ccf/main.tsp
```diff
@@ -1,608 +1,22 @@
...
-};
-
-@route("/gov/recovery_share")
-@summary("Provide a recovery share for the purpose of completing a service recovery")
-@post op postRecoveryShare(@body recoveryShare: SubmitRecoveryShareIn): {
-    @header("x-ms-ccf-transaction-id") transactionID: string;
-    @body recoveryShare: SubmitRecoveryShareOut;
-};
-
-@route("/gov/encrypted_recovery_share/{member_id}")
```
lynshi, line 597: I know it's in-progress so this is probably coming, but just noting that all the recovery-related endpoints are gone now

typespec-ccf/proposals.tsp
eddyashton: Good clarification. By "signed proposal body", I meant the request body, but realise now that's ambiguous (could be the Cose payload, not the entire CoseSign1). I think the right thing to do here is to link to the CCF docs, where we can write a more thorough (and updatable) explanation of proposal identity + idempotence.

typespec-ccf/tsp-output/@azure-tools/typespec-autorest/0.0.1-preview/openapi.json
eddyashton: This is one of the big holes in this scheme, and something I will raise with the TypeSpec/Azure Core team. Afaict, there is no way to pass describe arbitrary metadata like this that will be handled by the emitters, other than verbose descriptions. Linking to the docs is one solution, but I'm hoping there's something I've missed that will let us document the payload scheme + required headers inline in the OpenAPI.

typespec-ccf/proposals.tsp
eddyashton: I've tweaked the string slightly, and created a TODO to get a clearer docs page to link to:

https://github.com/microsoft/CCF/issues/5230

typespec-ccf/proposals.tsp
```diff
@@ -0,0 +1,160 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@azure-tools/typespec-azure-core";
+
+import "./cose.tsp";
+import "./common.tsp";
+
+@doc("A single step in a proposed change to the service.")
+model Action {
+  @doc("The name of action to perform. This should match an action defined in the service's constitution, so that it can be invoked by the `apply` function of the constitution if the proposal is accepted.")
```
achamayou, line 10:
```suggestion
  @doc("The name of the action to perform. This should match an action defined in the service's constitution, so that it can be invoked by the `apply` function of the constitution if the proposal is accepted.")
```

achamayou (approved):

jumaffre (approved):

PR #5223, opened 2023-04-26
Validate tests/raft_scenarios/bad_network, fancy_election.1, and suffix_collision.1

lemmy:
…s/raft_scenarios/fancy_election.1.


lemmy:
bad_network validates but fancy_election.1 still fails.  Please take a preliminary view anyway.

lemmy:
I believe the issue with fancy_election.1 not being validated stems from the fact that `ccfraft.tla` doesn't permit a node's log to be truncated based on an AE message that lacks entries. By the end of the scenario (at line 96), node #0's log suffix is not in sync and requires truncation (from index 11 to index 9). While node #0 does receive an appropriate AE message from node #1 (the leader), the enablement conditions present in [`ConflictAppendEntriesRequest`](https://github.com/microsoft/CCF/blob/main/tla/ccfraft.tla#L915) and [`NoConflictAppendEntriesRequest`](https://github.com/microsoft/CCF/blob/main/tla/ccfraft.tla#L931) allow only an `AppendEntriesAlreadyDone` step being taken. Consequently, the log remains unchanged, and a bogus AER response is sent.

lemmy:
Commit 2efd6eb violates the last conjunct of `LogConfigurationConsistentInv`:

https://github.com/microsoft/CCF/blob/520af81092350fa977dbb05b3958a63a17fa599a/tla/ccfraft.tla#L1266-L1270

heidihoward:
Sorry I've not had a chance to look at this property but if I recall correctly, `ConflictAppendEntries` does not spent a reply or update commit index as the `NoConflictAppendEntriesRequest` action can do this in a later step as the AER message received is not removed from the set of message.

lemmy:
> Sorry I've not had a chance to look at this property but if I recall correctly, `ConflictAppendEntries` does not spent a reply or update commit index as the `NoConflictAppendEntriesRequest` action can do this in a later step as the AER message received is not removed from the set of message.

This was the right clue as it led me to understand that this particular case requires the use of action composition (\cdot).  (Un)fortunately, it appears this case triggers a bug in the experimental cdot support, which has to be addressed first.

lemmy:
This is ready for review.

achamayou (approved):

PR #5227, opened 2023-04-28
Ensure that latest version of OpenSSL is installed

### LINKED ISSUE #5225: Pin version of OpenSSL in SNP/virtual images
As this is currently set to the version installed on the `ubuntu:20.04` runtime Docker image, as installed via other dependencies (TBC).

jumaffre:
Resolves #5225 

This PR enforces that the latest version of OpenSSL (including the latest security patches from OpenSSL `1.1.1t`) is installed when installing the CCF debian package. This is particularly important for Virtual/SNP builds that use the host OpenSSL (rather than Open Enclave's). Note that the correct version was already installed, so this changes only prevents regressions.


getting_started/setup_vm/roles/openssl/vars/common.yml
```diff
@@ -0,0 +1,4 @@
+# As per https://launchpad.net/ubuntu/+source/openssl/1.1.1f-1ubuntu2.18,
+# this corresponds to OpenSSL 1.1.1t, including patches for CVE-2023-0286,
+# CVE-2023-0215, CVE-2022-4450 and CVE-2022-4304.
+openssl_ubuntu_deb_package: "openssl=1.1.1f-1ubuntu2.18"
```
achamayou, line 4:
```suggestion
openssl_ubuntu_deb_package: "openssl>=1.1.1f-1ubuntu2.18"
```

This is otherwise going to break very quickly at best, or result in outdated versions of openssl installed in images at worst.

jumaffre: I'm not sure. First, the `>=` syntax doesn't seem to work out of the box with Ansible's `apt`. Second, `openssl` is a critical dependency that is pinned on SGX (via Open Enclave), and so should also be on SNP builds? I think there's only one version that's ever allowed here (the very latest) at any given point in time? Happy to be convinced otherwise though!

achamayou: Two things:

1. It's pinned on SGX by necessity, not by design. We do not want to prevent our users from upgrading.
2. -1ubuntu2.18 is an extremely ephemeral build suffix in the ubuntu apt repo. There may be a 2.19, 2.20, and after a few weeks 2.18 will be gone from the repo (only a few binaries are kept), and the playbook will start failing. We definitely do not want to update the playbook and our images that often.

I think we really want to express "1.1.1f and above", as much as possible.

jumaffre: As suggested, we now check this via `cpack` rather than Ansible.

PR #5228, opened 2023-04-28
Corrupt signatures more reliably in test

achamayou:
Previous logic just set a byte to 0, which doesn't help when it's already 0. We now bit flip a range large enough to cover the smallest possible signature.


tests/infra/clients.py
```diff
@@ -647,7 +647,9 @@ def request(
                 request_body or b"", key, cert, phdr
             )
             if self._corrupt_signature:
-                request_body = request_body[:-5] + b"0" + request_body[-4:]
+                request_body = request_body[:-32] + b"".join(
+                    [(~i & 0xFF).to_bytes(1, "big") for i in request_body[-32:]]
```
eddyashton, line 651: I understand the switch from "set to 0" to "negate", to reliably flip some bits. Is there a reason we've also increased the range? It feels like corrupting the smallest amount possible (reliably) gives a more interesting test (the signature should fail if a single bit is flipped, it's almost too easy if we flip 32-bytes).

eddyashton (approved):

tests/infra/clients.py
achamayou: That's a good point! Will fix.

PR #5233, opened 2023-04-29
Validating tests/raft_scenarios/fancy_election.2 and tests/raft_scenarios/suffix_collision.2

lemmy:


lemmy:
The implementation seemingly increments the node's `currentTerm` when it steps down as a leader and becomes follower. 
 This is *not* in line with the sub-action `CheckQuorum` in `ccfraft.tla`.

lemmy:
The implementation (action-) composes the three sub-actions `UpdateTerm`, `ConflictAppendEntriesRequest`, and `RejectAppendEntriesRequest`. 

At least `tests/raft_scenarios/fancy_election.2` exhibits this behavior (line 80 - 83) s.t. node #0 receives an AE message (`recv_append_entries`) that makes #0 `become_aware_of_new_term`, `become_follower`, `rollback`,  and finally `send_append_entries_response`.

lemmy:
@achamayou @heidihoward Please provide feedback if you have a chance.  The commit  https://github.com/microsoft/CCF/pull/5233/commits/443c854b93bb0fa6a0cec5acd55af2dc91673c28 introduces a `TruncateLog` sub-action (composed with `UpdateTerm` and `HandleAppendEntriesRequest` in `Traceccfraft`).   AFAICT, `ccfraft` currently does not model https://github.com/microsoft/CCF/blob/main/src/consensus/aft/raft.h#L1837 because `ConflictAppendEntriesRequest` will truncate the log iff the AE message's `entries` is non-empty.

tla/Traceccfraft.tla
```diff
@@ -168,6 +168,19 @@ TraceConflictAppendEntriesRequestNoConflictAppendEntriesRequest ==
                 \cdot NoConflictAppendEntriesRequest(m.dest, m.source, m)  
             /\ UNCHANGED <<candidateVars, leaderVars>>
 
+TruncateLog(i) ==
+    \* \* raft.h: recv_append_entries > become_aware_of_new_term > become_follower > rollback
+    /\ log[i] # <<>> 
+    /\ commitIndex[i] < Len(log[i])
+    /\ log' = [log EXCEPT ![i] = SubSeq(log[i], 1, commitIndex[i])]
```
achamayou, line 175: That's not quite what the implementation does, instead it's rollback(last_committable_index()) (raft.h:1821).

Rolling back to the last locally known committed break user-observable persistence: the primary will advertise entries as committed as soon as it has seen a quorum of ack, but there is a further delay until all potential candidates are aware of the commit. If there is an election in that window, the new node may have committable entries that were once advertised as committed, but which it itself doesn't know are committed.

If these entries are rolled back, by even one node, we may be destroying a quorum where there once was one, which was advertised to users.

achamayou:
```suggestion
    /\ log' = [log EXCEPT ![i] = SubSeq(log[i], 1, MaxCommittableIndex(log[i])]
```

I think this should be ^

lemmy: Still investigating but `fancy_election.2` no longer validates with your proposed change (`MaxCommittableIndex(log[i])`).

heidihoward: I agree with @achamayou, nodes should never rollback logs passed the last signature unless they have diverged from the current leader. This could lead to a safety violation down the line

achamayou: @lemmy it seems to me that there are two ways forward then:

1. log committable_indices, as you did in #5238 (sorry for commenting there before reading this)
2. update the spec to maintain committable indices per node, and clear it post election like the implementation does

They seem about equivalent to me (1 seems a little weaker, but not to a worrying extent), and 1. is easier/already done in #5238. @heidihoward what do you think?

lemmy: The interactions between the various indices is rather intricate, which is why my intuition suggests aligning the specification with the implementation.  In other words, I'd suggest to model `committable_indices` to verify that the spec's properties hold.

heidihoward: I'm ok with `committable_indices` being added to the spec

PR #5237, opened 2023-05-03
JS runtime traces in responses

achamayou:
The `set_js_runtime_options` action now accepts `return_exception_details` and `log_exception_details` boolean options, which set the corresponding keys in the `public:ccf.gov.js_runtime_options` KV map. When enabled, a stack trace is respectively returned to the caller, and emitted to the log, on uncaught JS exceptions in application code.

Filed #5239 as follow up work.


src/js/wrap.h
```diff
@@ -243,6 +243,9 @@ namespace ccf::js
     std::chrono::milliseconds max_exec_time = default_max_execution_time;
 
   public:
+    bool log_exception_details = false;
+    bool return_exception_details = false;
```
jumaffre, line 247: Does this have to be `public:`?

jumaffre (approved):

src/js/wrap.h
achamayou: It needs to be accessible by the frontend.

PR #5248, opened 2023-05-05
User cose sign1

achamayou:
Adds `ccf::UserCOSESign1AuthnPolicy` (C++) and `user_cose_sign1` (JavaScript) authentication policies.


src/endpoints/authentication/cose_auth.cpp
```diff
@@ -275,4 +372,91 @@ namespace ccf
...
+      identity->user_cert = user_cert.value();
+      identity->protected_header = phdr;
+      identity->envelope = body;
+      identity->content = authned_content;
+      identity->signature = cose_signature;
+      return identity;
+    }
+    else
+    {
+      error_reason = fmt::format("Signer is not a known member");
```
jumaffre, line 435:
```suggestion
      error_reason = fmt::format("Signer is not a known user");
```

Also report the request `kid` in the error message too maybe?

jumaffre (approved):
LGTM. https://microsoft.github.io/CCF/main/build_apps/auth/cert.html needs updating though.

src/endpoints/authentication/cose_auth.cpp
achamayou: It's provided by the caller, so I don't see the upside. If the caller is being adversarial and sets a large kid, we'll have to send a large response.

PR #5257, opened 2023-05-10
Fix occasional issue for historical queries - process earlier ledger secrets whenever they arrive

### LINKED ISSUE #5236: Failure in `recovery.test_recover_service_aborted` end-to-end test
Seen in the Daily CI:

https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=69938&view=logs&j=5435e0ac-25e5-5426-50be-61b0d0ea8d34&t=a2a8c6c1-1b13-59a7-596c-97d4f22a6857

I've manually trimmed the suite, and (as implausible as it seems), this is the minimal steps I've found that will repro this:

```

temp_suite = [
 recovery.test_recover_service, # Keep
 memberclient.test_corrupted_signature, # Keep
 e2e_operations.test_forced_snapshot, # Keep
 recovery.test_recover_service, # Keep
 e2e_logging.test_forwarding_frontends, # Keep
 recovery.test_recover_service_aborted,
]
suites["all"] = temp_suite
```

The failure is actually early in `test_recover_service_aborted`, when we're trying to shut down the existing network. As part of that, we call `verify_range_for_idx`, and discover that one of the _previous_ nodes is stalling trying to build an index. Have increased the timeout, and pretty sure it really is stalled, not just slow.

eddyashton:
Resolves #5236.

Seen at least once in the CI, in a specific run of the full test suite.

Fix is to always track the earliest known secret, and the earlier secret it points to. If we ever see the earlier secret (even when it wasn't explicitly requested _as_ a secret to decrypt a requested entry), we process it and update the historical ledger secrets collection.

~~Opening as a draft because I live in eternal hope of finding a simpler regression test that we can check in.~~


CMakeLists.txt
```diff
@@ -667,6 +667,16 @@ if(BUILD_TESTS)
         ${CMAKE_SOURCE_DIR}/samples/templates
     )
 
+    if(LONG_TESTS)
+      add_e2e_test(
+        NAME regression_test_suite
+        PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/e2e_suite.py
+        CONSENSUS cft
+        LABEL suite
+        ADDITIONAL_ARGS --test-suite regression_5236
```
jumaffre, line 676: Should this PR touch the daily canary file to make sure this new test runs fine?

jumaffre (approved):

achamayou (approved):

PR #5269, opened 2023-05-16
Unskip JS e2e tests

### LINKED ISSUE #4405: Investigate skipped tests
We're still skipping a surprising number of tests in the CI, often due to `Missing required methods`, and I believe this is unintentional.

In particular, it looks like the `js` tests are skipping most of their tests due to the wrong app running.

We should ensure we're actually running the tests we intend to most of the time, and perhaps revisit some of the requirements - rather than skipping if a requirement isn't met, we could modify the network so the requirement is met (eg - add or remove nodes, change the app).

eddyashton:
Resolves #4405.

This PR does 2 things:
- Flips the default value of `--throws-if-reqs-not-met`, so that by default we throw whenever requirements fail, and only permit them to be skipped in shuffled suites.
- Implements several more endpoints in JS, so that `e2e_logging_cft`'s `js` runner has fewer skips and demonstrates samples closer to parity with C++. (Now have a few more explicitly disabled tests, where we don't have API parity)

### DIFF
--- a/samples/apps/logging/js/src/logging.js
+++ b/samples/apps/logging/js/src/logging.js
@@ -261,7 +261,7 @@ export function post_private_admin_only(request) {
   // or the user data is not an object with isAdmin field,
   // or this field is not true
   const data = request.caller.data;
-  if (!(typeof data === "object" && data !== null && data.isAdmin === true)) {
+  if (data?.isAdmin !== true) {
     return {
       statusCode: 403,
       body: "Only admins may access this endpoint",
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -1882,12 +1882,12 @@ if __name__ == "__main__":
         nodes=infra.e2e_args.max_nodes(cr.args, f=0),
     )
 
-    # cr.add(
-    #     "cpp_illegal",
-    #     run_parsing_errors,
-    #     package="samples/apps/logging/liblogging",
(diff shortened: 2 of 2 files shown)


samples/apps/logging/js/src/logging.js
```diff
@@ -256,6 +256,47 @@ export function post_private(request) {
   return { body: true };
 }
 
+export function post_private_admin_only(request) {
+  // Return an auth error if the caller has no user data,
+  // or the user data is not an object with isAdmin field,
+  // or this field is not true
+  const data = request.caller.data;
+  if (!(typeof data === "object" && data !== null && data.isAdmin === true)) {
```
achamayou, line 264: I believe you can do this:

```suggestion
  if (!(data?.isAdmin === true)) {
```

tests/e2e_logging.py
```diff
@@ -1867,12 +1882,12 @@ def run_parsing_errors(args):
         nodes=infra.e2e_args.max_nodes(cr.args, f=0),
     )
 
-    cr.add(
-        "cpp_illegal",
-        run_parsing_errors,
-        package="samples/apps/logging/liblogging",
-        nodes=infra.e2e_args.max_nodes(cr.args, f=0),
-    )
+    # cr.add(
```
achamayou, line 1885: Needs commenting out?

achamayou (approved):

PR #5274, opened 2023-05-17
Experimental: support for custom protocols

wintersteiger:
This adds support for custom TCP and UDP protocols. It is used via a new subsystem, which should make it easy to either keep it private for now, or to clearly mark it as experimental (in docs, but perhaps also by moving it to `include/ccf/experimental`?).

### DIFF
--- a/include/ccf/service/node_info_network.h
+++ b/include/ccf/service/node_info_network.h
@@ -26,17 +26,7 @@ namespace ccf
      {Authority::ACME, "ACME"},
      {Authority::UNSECURED, "Unsecured"}});
 
-  enum class ApplicationProtocol
-  {
-    HTTP1,
-    HTTP2,
-    CUSTOM
-  };
-  DECLARE_JSON_ENUM(
-    ApplicationProtocol,
-    {{ApplicationProtocol::HTTP1, "HTTP1"},
-     {ApplicationProtocol::HTTP2, "HTTP2"},
-     {ApplicationProtocol::CUSTOM, "Custom"}});
+  using ApplicationProtocol = std::string;
 
   struct Endorsement
   {
--- a/include/ccf/service/node_info_network.h
+++ b/include/ccf/service/node_info_network.h
@@ -29,12 +29,14 @@ namespace ccf
   enum class ApplicationProtocol
   {
     HTTP1,
-    HTTP2
+    HTTP2,
+    CUSTOM
   };
   DECLARE_JSON_ENUM(
     ApplicationProtocol,
     {{ApplicationProtocol::HTTP1, "HTTP1"},
-     {ApplicationProtocol::HTTP2, "HTTP2"}});
+     {ApplicationProtocol::HTTP2, "HTTP2"},
+     {ApplicationProtocol::CUSTOM, "Custom"}});
 
   struct Endorsement
   {
--- a/src/enclave/rpc_sessions.h
+++ b/src/enclave/rpc_sessions.h
@@ -27,6 +27,7 @@
 
 #include <limits>
 #include <map>
+#include <stdexcept>
 #include <unordered_map>
 
 namespace ccf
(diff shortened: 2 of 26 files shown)


include/ccf/service/node_info_network.h
```diff
@@ -29,12 +29,14 @@ namespace ccf
...
-    HTTP2
+    HTTP2,
+    CUSTOM
   };
   DECLARE_JSON_ENUM(
     ApplicationProtocol,
     {{ApplicationProtocol::HTTP1, "HTTP1"},
-     {ApplicationProtocol::HTTP2, "HTTP2"}});
+     {ApplicationProtocol::HTTP2, "HTTP2"},
+     {ApplicationProtocol::CUSTOM, "Custom"}});
```
achamayou, line 39: Custom seems more like RAW_UDP, unless I am missing something?

achamayou: I think it would be a little nicer to make application protocol a string, pass it in, and allow new protocols to register by name. We would avoid this situation where "Custom" means "a single custom thing that handles UDP", and we could keep the config & config schema unchanged.

`set_custom_protocol_subsystem(cpss)` would become `register_custom_protocol_subsystem("custom_protocol", cpss)`, what do you think @eddyashton?

achamayou:
> This adds support for custom TCP and UDP protocols. It is used via a new subsystem, which should make it easy to either keep it private for now, or to clearly mark it as experimental (in docs, but perhaps also by moving it to `include/ccf/experimental`?).

Probably `include/ccf/research/`, the only part that isn't optional and that looks a bit inflexible is the single, implicitly UDP, "Custom" value name in the config. If the custom protocols could register their name, and were lookup from that, I think it'd be more generic and extensible.

Apart from that, this looks very good to me, thank you @wintersteiger!

include/ccf/service/node_info_network.h
wintersteiger: There's a different setting for TCP/UDP, but yes, "custom" here really means "application-defined". Since you can have only one per interface I figured that's enough. But you're right, it may be nicer to register them by name. Just so I understand correctly: are you thinking about allowing arbitrary strings, i.e. `ApplicationProtocol` becomes a `std::string` instead of an enum?

src/enclave/rpc_sessions.h
```diff
@@ -418,10 +461,27 @@ namespace ccf
...
-            rpc_map, id, listen_interface_id, writer_factory);
-          sessions.insert(std::make_pair(
-            id, std::make_pair(listen_interface_id, std::move(session))));
+          if (
+            per_listen_interface.app_protocol ==
+            ccf::ApplicationProtocol::CUSTOM)
+          {
+            // We know it's a custom protocol, but the session creation function
+            // hasn't been registered yet, so we keep a nullptr until the first
+            // udp::inbound message.
```
eddyashton, line 470: The comment here is very helpful. What's the window where we might be accepting but haven't registered a session-creator function yet? Do we accept early in node startup before the app constructor is called, or are we wanting to permit even later protocol registration? If it's safe to do so, I'd prefer to return an error to this accept ("I can't accept this protocol yet") rather than storing a nullptr (and the associated checks that requires later), and quietly dropping all `recv` data. I understand that may be dependent on client behaviour though.

include/ccf/service/node_info_network.h
eddyashton: I agree with @achamayou, it would be neater to put the protocol here. If I understand correctly, with the current API we'd put the protocol as the listen-interface name, and then need to match that in the app constructor? That's an awkward bit of app-code-to-config coupling, in particular making it serve a custom protocol on multiple interfaces (would need to call `cpss->install` multiple times, to match each interface name/ID). We lose type safety/early validation if we drop from enum to string lookup, but I think that's a fair tradeoff.

wintersteiger: All good points, I agree. If you want multiple protocols on multiple interfaces strings are definitely a better choice.

src/enclave/rpc_sessions.h
wintersteiger: The registration currently happens during app initialization (`init_handlers`), because I considered all of this application-specific. There's no problem with registering something earlier though, it's just less accessible to the application. I could add a log line, but I can't return a meaningful error to the client, because during this time we don't know what an error in the custom protocol would look like. For TCP, we should really just refuse the connection, but somehow we'd have to tell the host to refuse until it gets a signal from the application. For UDP there's no refusal/acceptance and replying to every random packet could open us up to amplification attacks, so I think just dropping those packets is the right thing to do.

include/ccf/service/node_info_network.h
wintersteiger: All done, app protocol is now a string.

src/enclave/rpc_sessions.h
```diff
@@ -308,6 +331,26 @@ namespace ccf
...
+        listen_interface_id);
+      if (fit == custom_protocol_subsystem->session_creation_functions.end())
+      {
+        return nullptr;
+      }
+      else
+      {
+        return fit->second(id, std::move(ctx));
+      }
+    }
```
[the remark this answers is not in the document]
wintersteiger: Done, this is now in the subsystem.

achamayou (approved):

PR #5277, opened 2023-05-18
TLA CI validation fix: Call `apt update`

eddyashton:
APT repo update means the `trace_validation.yml` pipeline is failing to install Java. Resolved for now with an explicit `apt update` before attempting to install. This update should really be rolled into a new CI image.


achamayou (approved):

lemmy:
> APT repo update means the `trace_validation.yml` pipeline is failing to install Java. Resolved for now with an explicit `apt update` before attempting to install. This update should really be rolled into a new CI image.

AFAICT the repo package cache will still get outdated eventually at which point an `apt install openjdk` will again fail.  Perhaps, Java  17 (LTS) should directly be installed into the CI image?

PR #5281, opened 2023-05-18
Remove outdate trace validation action.

lemmy:
The "send_request_vote_response" disjunct is reported as another "IsRcvRequestVoteRequest" action with zero coverage.

Refactoring of https://github.com/microsoft/CCF/issues/5057


lemmy:
@achamayou This change should have triggered trace validation but I can't find a related run on Azure Pipelines.   Am I blind?

achamayou:
@lemmy you are not blind, it is my mistake!

I had configured the CI _not_ to trigger a full run when only TLA changes were made in #4442, because it the CI takes time and the model was disconnected from it. But that's no longer true since #5201, so I will remove that change, and we will have a full CI run every time a TLA change happens.

Another solution would be to split off a CI pipeline that only builds the tests we currently use for tracing (scenarios), and feeds them to the trace validation. This would be quicker for TLA changes than a full CI, but it duplicates some of the build with the existing CI, and it will become less beneficial as we start to validate e2e tests.

PR #5287, opened 2023-05-19
Move side-effects out of common.cmake and into CMakeLists.txt

### LINKED ISSUE #2953: Refactor `common.cmake` to contain only helper-type code
Targets and install commands are currently split between the root `CMakeLists.txt` and `cmake/common.cmake`. Historically, some of that was on purpose to support the use case of including ccf as git submodule instead of consuming the installed package. This has changed and there is no reason anymore to treat `common.cmake` special. It is also not part of the installed package. 

I suggest to move all library/executable targets and install commands back to `CMakeLists.txt` while possibly introducing more helper functions if useful. I think this will improve long-term maintenance and will make it easier for new-comers to understand the build.

The above is not strictly true, there are also some targets that are defined in additional files: `cmake/{aft,crypto,quickjs,sss}.cmake`. On the other hand, the `ccf`, `ccf_endpoints` and `http_parser` libraries are defined in `CMakeLists.txt`/`common.cmake` directly. quickjs and sss are third-party libraries, while the rest live in `src/`.
I think it is OK to keep the separate CMake files for the third-party libraries, but I would move the contents of `aft` and `crypto` into `CMakeLists.txt`. Ideally, only the list of source files plus calls to `add_{enclave,host}_library[_c]` remain.

achamayou:
Closes #2953


eddyashton (approved):
LGTM! Can you poke the daily canary, and maybe threading too, to make sure we're covering a variety of build configurations?

jumaffre (approved):

PR #5297, opened 2023-05-23
Benchmarking scripts

heidihoward:
Scripts for benchmarking


tests/perf-system/benchmarks/README.md
```diff
@@ -0,0 +1,3 @@
+This directory includes the scripts to reproduce our experiments.
+
+To run the experiments, set up a client/coordinator machine and 5 of server machines. `setup-remote-server.py` and `setup-client.sh` will help with this. Then start the experiments using `python3 run_distributed.py` (edit the file to change server ips). By default, the results will be added to a timestamped directory in `/datadrive1/test`. This directory will include a subdirectory for each experiment with the experiment configuration in the directory name, each of these contain subdirectories for each workload within each experiment.
```
achamayou, line 3:
```suggestion
To run the experiments, set up a client/coordinator machine and 5 server machines. `setup-remote-server.py` and `setup-client.sh` will help with this. Then start the experiments using `python3 run_distributed.py` (edit the file to change server ips). By default, the results will be added to a timestamped directory in `/datadrive1/test`. This directory will include a subdirectory for each experiment with the experiment configuration in the directory name, each of these contain subdirectories for each workload within each experiment.
```

tests/perf-system/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,513 @@
...
+from pathlib import Path
+from loguru import logger as LOG
+
+import paramiko
+
+
+def connect_to_remote_server(ip_address):
+    client = paramiko.SSHClient()
+    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    client.connect(ip_address)
```
achamayou, line 22: Having an open ended function, as opposed to a context manager (eg. https://github.com/microsoft/CCF/blob/9e0e46c41a306d10dc128f3f3672ad16a58cc089/tests/infra/remote.py#L43) runs the risk that the connection will stay open for longer than intended if the script otherwise raises an uncaught exception.

tests/perf-system/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,513 @@
...
+    return client
+
+
+def clean_up_cchost(nodes):
```
achamayou, line 26: In CI infra, we use context managers to make sure connections are shut down if the script terminates, and a combination of a pty to get a SIGHUP on disconnect (https://github.com/microsoft/CCF/blob/9e0e46c41a306d10dc128f3f3672ad16a58cc089/tests/infra/remote.py#L264) and a flag to terminate all children processes: https://github.com/microsoft/CCF/blob/9e0e46c41a306d10dc128f3f3672ad16a58cc089/tests/infra/remote.py#L38

tests/perf-system/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,513 @@
...
+    for node in set([address.split(":")[0] for address in nodes]):
+        client = connect_to_remote_server(node)
+        LOG.info(f"Killing cchost on {node}")
+        client.exec_command("killall cchost")
+        client.close()
+
+
+# TODO: make recursive, copies only first two layers
+def fetch_remote_dir(ip_address, remote_dir: Path, local_dir: Path):
+    client = connect_to_remote_server(ip_address)
```
achamayou, line 37: pysftp seems to support this is and looks fairly easy to use:

https://pysftp.readthedocs.io/en/latest/cookbook.html#pysftp-connection-get-r
https://pysftp.readthedocs.io/en/latest/index.html

tests/perf-system/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,513 @@
...
+                experiment_dir
+                / f"writes{write_percentage}%"
+                / f"piccolo{piccolo_number}"
+                / "input.parquet"
+            )
+    LOG.info("Workloads generated")
+
+    LOG.info("Checking CCF service is live & new before starting experiment")
+    addresses_of_ccf_nodes = [f"https://{i}" for i in ccf_node_ips]
+    for id, address in enumerate(addresses_of_ccf_nodes):
```
achamayou, line 206: The Network (https://github.com/microsoft/CCF/blob/9e0e46c41a306d10dc128f3f3672ad16a58cc089/tests/infra/network.py) and CCFRemote classes take care of file copy, node startup, waiting for nodes to be ready etc. It would be preferable not to duplicate this logic.

tests/perf-system/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,513 @@
...
+        [
+            total_posts := total_posts + requests_per_workload * w / 100
+            for w in write_percentages
+        ]
+        total_gets = total_requests - total_posts
+        LOG.info(
+            f"Expecting {total_requests} requests, comprised of {total_posts} posts and {total_gets} gets"
+        )
+
+        metrics_process = subprocess.run(
```
achamayou, line 317: Using https://github.com/microsoft/CCF/blob/9e0e46c41a306d10dc128f3f3672ad16a58cc089/tests/infra/clients.py#L544 will be shorter, provide good logging, and will run quicker since you will need fewer tls handshakes and file operations.

tests/perf-system/benchmarks/benchmarks.py
```diff
@@ -0,0 +1,513 @@
...
+
+        node_ip = ccf_node_ips[id].split(":")[0]
+        LOG.info(f"Fetching workspace from node {id} at {address}")
+        sandbox_dir = experiment_dir / "workspace" / f"sandbox_{id}"
+        os.makedirs(sandbox_dir)
+        fetch_remote_dir(node_ip, ccf_dir / "workspace" / f"sandbox_{id}", sandbox_dir)
+        LOG.info("Fetching completed")
+
+    LOG.info("Shutting down CCF service")
+    ccf_server_process.send_signal(signal.SIGINT)
```
achamayou, line 395: It's possible to obtain the same result without delay by allocating a PTY, setting PR_SET_PDEATHSIG and simply closing the connection (or exiting the scope when using a context manager).

achamayou (approved):

tests/perf-system/benchmarks/benchmarks.py
heidihoward: done in d8be6a62b0b614866de8c6eab86235d2aa697194

tests/perf-system/benchmarks/benchmarks.py
heidihoward: This particular instance has been replaced with a context manager in ab101aa0ac216184d0a993058863ec7ab2e703cd

Issue #5132, opened 2023-03-21
Validate constitutions on the server


eddyashton:
To prevent "invalid" constitutions being submitted and locking a service, we should add some utility functions to the governance JS API to validate a constitution script. As a minimum, we should be able to spin up an isolated JS interpreter and confirm that the proposed constitution is parseable. Ideally, that should return some description of the constitution as though it was a loaded module, so that (default, sample) constitutions can check the _proposed_ constitution still contains the expected `validate`, `resolve`, and `apply` functions.

cauldnz:
There is also the potential to lock a service by misconfiguring the Javascript runtime options. If the options are configured in such a way that the JS runtime is unable to start, it will no longer be possible to execute the Constitution. 

See below PoC. Suggested initial fix would be to put some bounds checks on the sample constitution's `set_js_runtime_options` action.

https://gist.github.com/cauldnz/f7e52b100544526f0ebc5905ed349bbe

eddyashton:
@cauldnz This is a good suggestion, and I'll include it in this validation work.

It's worth clarifying that, counterintuitively, we actually want to ensure it is _possible_ for the governing consortium to break the service in various ways. This is to justify the claim that they have full control of the service. In practice, it means it should be possible to "freeze" a service in various ways - making it impossible to add or remove future nodes, to stage recoveries, or to update the constitution. Such a service would be trading off some operational safety (it can never be updated!) for a stronger guarantee to the users of the members' ongoing power - they can't read your confidential state now, and can't change the rules of the service so they could read it in future.

But that's just a general design point, and doesn't affect the suggestion here. I don't think breaking the JS runtime config is a useful freeze, and the sample constitution can add many sensible safeguards that don't break the above capability (consortium can always change the constitution to remove such safeguards if they need to).

PR #5299, opened 2023-05-24
Update js_runtime.rst documentation per 4.0.1

cauldnz:
Small update to JS runtime documentation to add exception reporting features added in 4.0.1


eddyashton (approved):

cauldnz:
Suggest we might want to also update the values in that example action to be the default values.... Submitting the values as given in the documentation using the sample constitution will kill QuickJS and thus kill the service. See #5132.

Thoughts @eddyashton?

achamayou (approved):

PR #5301, opened 2023-05-24
Fix `enclave_time.h`: Don't complain about reversing clocks when running multi-threaded

eddyashton:


src/enclave/enclave_time.h
```diff
@@ -17,24 +17,30 @@ namespace ccf
...
     if (enclavetime::host_time_us != nullptr)
     {
       const auto current_time = enclavetime::host_time_us->load();
-      if (current_time >= enclavetime::last_value.load().count())
+      if (current_time >= last.count())
       {
-        enclavetime::last_value = std::chrono::microseconds(current_time);
+        // If this fails, it simply means another thread has fetched and updated
+        // the in-enclave last_value independently. Both are happy that their
+        // values do not decrease time, so either my succeed.
```
achamayou, line 30:
```suggestion
        // values do not decrease time, so either may succeed.
```

achamayou (approved):

PR #5309, opened 2023-05-25
[WIP] Fix multiple piccolos

achamayou:
Start to split files correctly, although a few changes need to bubble down to the runners. This change seems to have an impact on the measurements, which I did not expect, and should not be merged as is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1433,6 +1433,7 @@ if(BUILD_TESTS)
       CLIENT_BIN ./submit
       ADDITIONAL_ARGS --package "samples/apps/logging/liblogging"
                       --max-writes-ahead 1000 --repetitions 10000
+                      --num-localhost-clients 2
     )
 
     add_perf_test(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1433,7 +1433,6 @@ if(BUILD_TESTS)
       CLIENT_BIN ./submit
       ADDITIONAL_ARGS --package "samples/apps/logging/liblogging"
                       --max-writes-ahead 1000 --repetitions 10000
-                      --num-localhost-clients 2
     )
 
     add_perf_test(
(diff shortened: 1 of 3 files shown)


.azure-pipelines-templates/matrix.yml
```diff
@@ -92,6 +92,29 @@ jobs:
...
+  - ${{ if eq(parameters.perf_tests, 'run') }}:
+      - template: common.yml
+        parameters:
+          target: SGX
+          env: ${{ parameters.env.SGX }}
+          cmake_args: "${{ parameters.build.common.cmake_args }} ${{ parameters.build.perf.cmake_args }} ${{ parameters.build.SGX.cmake_args }} -DWORKER_THREADS=2"
+          suffix: "Perf_MultiThreaded"
+          artifact_name: "SGX_Perf_MultiThreaded"
+          ctest_filter: "-R pi_basic_mt"
+          depends_on: configure
```
eddyashton, line 116: We're currently only building `-DWORKER_THREADS=2` in `.multi-thread.yml` (the `CCF Github Threading` pipeline). This adds some multi-threaded build+testing to the main CI pipeline. Are we confident that this is stable enough to promote?

CMakeLists.txt
```diff
@@ -1442,10 +1442,20 @@ if(BUILD_TESTS)
       NAME pi_basic
       PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
       CLIENT_BIN ./submit
-      ADDITIONAL_ARGS --package "samples/apps/basic/libbasic"
-                      --max-writes-ahead 1000 --repetitions 10000
+      ADDITIONAL_ARGS --package "samples/apps/basic/libbasic" --repetitions
```
eddyashton, line 1445: Why no write-ahead cap?

eddyashton (approved):

CMakeLists.txt
achamayou: Moved to the script as I removed one layer of command overrides. We allowed a mix of client and nodes overrides, which I think is very error prone.

Issue #5291, opened 2023-05-22
Upgrade to OpenSSL 3.x


jumaffre:
We should add support for OpenSSL 3.x before OpenSSL 1.1.1 [end of life](https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/). 

**Updated plan** (16/06/02023)
- OpenSSL 1.1.1 will be supported by Canonical on Ubuntu 20.04 until 20.04 EOL (May 2025) so CCF will remain on Ubuntu 20.04 and use the distribution OpenSSL 1.1.1 for the host on SGX, but also for SNP and Virtual enclaves. 
- Open Enclave will update to OpenSSL 3.x (see https://github.com/openenclave/openenclave/issues/4759), which we'll use for SGX enclaves. 
- We should therefore provide support for both OpenSSL 1.1.1 (host + SNP/Virtual) and 3.x (SGX) and be able to run CCF enclaves from both. 

```[tasklist]
### Tasks (see <https://github.com/microsoft/CCF/pull/5293> for pointers)
- [x] SHA256 (#5336): `SHA256_Init`, `SHA256_Update`, `SHA256_Final`
- [x] EC_KEY (#5481): `EC_KEY_new_by_curve_name`, `EC_KEY_free`, `EC_KEY_set_public_key_affine_coordinates`, `EC_KEY_get0_group`, `EC_KEY_set_public_key`, `EC_KEY_set_private_key`
- [x] EVP_PKEY (#5481): `EVP_PKEY_set1_EC_KEY`, `EVP_PKEY_get0_EC_KEY`, `EVP_PKEY_get0_RSA`, `EVP_PKEY_set1_RSA`, `EVP_PKEY_get1_EC_KEY`
- [x] RSA (#5481): `RSA_new`, `RSA_generate_key_ex`, `RSA_free`, `RSA_set0_key`, `RSA_set0_factors`, `RSA_set0_crt_params`, `RSA_get0_n`, ` RSA_get0_e`, `RSA_get0_d`, `RSA_get0_p`, `RSA_get0_q`, `RSA_get0_dmp1`, `RSA_get0_dmq1`, `RSA_get0_iqmp`
- [x] d2i (#5481): `d2i_RSA_PUBKEY`
- [ ] Engine (#5569): `ENGINE_load_rdrand`, `ENGINE_by_id`, `ENGINE_init`, `ENGINE_set_default`, `ENGINE_free`
[...]

PR #5311, opened 2023-05-30
Upgrade from `python3.8` to `python3.10`

jumaffre:
This is part of the upgrade to Ubuntu 22.04 for SNP/virtual builds, where `python3.10` is available via `pip` by default. 

The couple of Ubuntu hosted CI executors are also upgraded to Ubuntu 22.04 so that `python3.10` is available by default.


jumaffre:
The SNP pipeline is failing as expected as it is building the image off `main`. It will be fixed once the PR is merged.

.azure-pipelines-templates/deploy_aci.yml
```diff
@@ -185,7 +185,7 @@ jobs:
   - job: cleanup_aci
     displayName: "Cleanup ACI"
     pool:
-      vmImage: ubuntu-20.04
+      vmImage: ubuntu-22.04
```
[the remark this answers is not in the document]
jumaffre: This should have only been done on CI jobs that used Python (3.8 but now 3.10) but do not make use of any of our containers. In other words, it's not a user-facing change, but simply for us to use 22.04 for a few CI agents. I'll reflect that in the PR description.

eddyashton (approved):

jumaffre:
This is currently blocked by the GitHub Pages job as Sphinx doesn't work with Python 3.10 because of https://github.com/python-babel/babel/issues/990. 

I see two options:
1. Wait for the fix (https://github.com/python-babel/babel/pull/1006) to be merged upstream (may take a while).
2. Temporarily relax the required version condition in `setup.py` so that `ccf` can be installed with `python3.8` (as there are no compatibility issues between 3.8 and 3.10).

.azure-pipelines-templates/checks.yml
```diff
@@ -9,5 +9,5 @@ jobs:
         clean: true
         fetchDepth: 1
 
-      - script: python3.8 scripts/extract-release-notes.py --target-git-version
+      - script: python3.10 scripts/extract-release-notes.py --target-git-version
```
achamayou, line 12: I think we ought to switch to calling `python3` everywhere, and communicate the version requirement through https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

Wherever we can, using a shebang line or setuptools entry points would be nicer too, to avoid having to specify which  python, now that 2.x is thoroughly obsolete.

getting_started/setup_vm/roles/ccf_build/tasks/install.yml
```diff
@@ -1,3 +1,10 @@
+- name: Add Python repository
```
achamayou, line 1: I am confused by this, doesn't 22.04 ship with 3.10? If so, why do we need the ppa?

jumaffre: This PR doesn't update the base container image to 22.04, and only updates Python to 3.10. Updating our container images to 22.04 will be a bigger change (in a separate PR) which cannot be completed just yet because of the lack of Open Enclave packages on 22.04.

jumaffre:
As per the update on https://github.com/microsoft/CCF/issues/5291, we will not do this as there are no tangible benefits and we will remain on Ubuntu 20.04 for now.

PR #5324, opened 2023-06-02
Add MonotonicMatchIndexProp to TLA+ spec.

lemmy:
As @jumaffre pointed out, figure 2 on page 4 of the raft paper provides the following property:

> Volatile state on leaders, reinitialized after election. For each server, index of the highest log entry known to be replicated on server. Initialized to 0, *increases monotonically*.

PR #5326, opened 2023-06-02
Do not decrement `match_idx` when processing NACKs

### LINKED ISSUE #5325: Raft's reuse of `match_idx` can lead to unsafely advancing commit index
As raised by @lemmy in #5321/#5324, we allow `match_idx` in `raft.h` to go backwards in the implementation.

As I suspected in [this comment](https://github.com/microsoft/CCF/pull/5324#issuecomment-1573463327), this actually raises a larger safety concern. It goes backwards because we're using it to store a different concept, the probe index that reduces the number of NACKs to realign with a divergent log. But using that probe index to populate `match_idx` means the primary may overstate how far another node's log matches, and then use that overstatement to _unsafely advance commit_.

Raft scenario repro proof here:

https://github.com/eddyashton/CCF/blob/match_idx_safety/tests/raft_scenarios/suffix_collision.3

In simpler terms (trying to ignore the surrounding partition complexity and additional nodes we need to reach this state, and almost certainly with some off-by-1s in doing this by-hand), we can have something like this:
```
Node 0:
 1.1  1.2  3.3  3.4  3.5  5.6  5.7

Node 1:
 1.1  2.2  2.3  2.4  2.5  4.6

Node 2:
 1.1  1.2
```

1. Node 0 sends an AE starting at 5.6 to Node 1.
1. Node 1 says "that doesn't match my log, but maybe we agree from seqno 5?"
1. Node 0 sees that NACK, and stores `match_idx[1] = 5`.
1. Node 0 gets an ACK from node 2, which causes them to try to advance commit.
1. Node 0 looks at all of their match indices, and sees both it and node 1 (a majority) appear to have seqno 5.
1. Node 0 advances commit to 5, despite 3.5 only being present on a single node.

I believe the fix is straightforward - store this probe index in a separate variable, and only set `match_idx` to safe matching values (received in ACKs).

eddyashton:
Resolves #5325.

Late on a Friday, so opening as a draft so I can verify this with fresh eyes on Monday, after some CI assurance.


lemmy:
@eddyashton Things are moving in the right direction.  Matching `matchIdx` (but not `sentIdx`) in `Traceccfraft.tla` successfully validates all existing raft scenarios *and* the new `suffix_collision.3`.

lemmy:
With the most recent changes in https://github.com/microsoft/CCF/pull/5324, trace validation now also passes when `sent_idx` is matched: https://github.com/lemmy/CCF/commits/match_idx_safety

src/consensus/aft/raft.h
```diff
@@ -1419,31 +1419,22 @@ namespace aft
...
-        node->second.sent_idx = node->second.match_idx;
+        const auto this_match =
+          find_highest_possible_match({r.term, r.last_log_idx});
+        node->second.sent_idx = std::min(this_match, node->second.sent_idx);
         return;
       }
       else
       {
-        match_idx = std::min(r.last_log_idx, state->last_idx);
+        node->second.match_idx = std::min(r.last_log_idx, state->last_idx);
```
lemmy, line 1437: Taking the `min` seems to still allow `match_idx` to go backwards if `r.last_log_idx` happens to be less than `state->last_idx`.  The spec had to be changed to take the `max` in the (unreviewed) PR #5324 to satisfy the new `MonotonicMatchIndexProp` property:

https://github.com/microsoft/CCF/pull/5324/commits/1454326be5a9f80822b892b31381d4c5bc4d3ebf

.daily_canary
```diff
@@ -1,4 +1,4 @@
    ---     ___           ___
   (- -)   (= =)   |     Y   &  +--
```
lemmy, line 2: Seeing this here reminds me that I've encountered merge issues because of `.daily_canary`.  Is this a check to make sure that some CI has to run before this PR can be merged?

src/consensus/aft/test/driver.h
```diff
@@ -899,6 +919,65 @@ class RaftDriver
     }
   }
 
+  void assert_commit_safety(ccf::NodeId node_id, const size_t lineno)
```
lemmy, line 922: Is this `QuorumLoginv` in ccfraft.tla?

https://github.com/microsoft/CCF/blob/75b3476bff4c8f3c35954883c08b53cd64fc02df/tla/ccfraft.tla#L1193-L1197

.daily_canary
achamayou: The canary files (there are several) are used to kick off additional jobs, which aren't part of the main CI, but test useful things. The reason they aren't kicked off by default is because they take quite a long time to run.

Daily is run once a day, and includes things like a build with various sanitisers for example. To trigger that build on a specific PR, one can make a change (any change) to the .daily_canary file. The alternative would be to go to the ADO UI, and enter the branch ref in a form to kick off a build. It's possible but much more annoying than making a change to the file. The result is also not picked up by GitHub in that case.

The downside of this scheme, as you point out, is conflicts.

achamayou: If we moved to GitHub actions, we might be able to use more subtle triggers for this, perhaps labels on the PRs.

lemmy: In other words, there is no definition of what constitutes a correct merge of .daily_canary?!

src/consensus/aft/raft.h
achamayou: +1, we probably want to log debug if/when this happens.

achamayou: >but I believe that's never possible due to the ordered handling of messages (guaranteed by the node-to-node channel) and the preconditions within this function.

Agreed, but it's clearer to make it explicit here and in the spec. We can add a comment that points to the guarantee coming from the node to node channel anyway.

.daily_canary
achamayou: It's a button. It can be pressed (has received a change in this PR), or not pressed (has not received a change in this PR).

Any merge that gets the state you want on your PR (pressed or not pressed) is correct. Some pressed states are more artistic than others, but that's up to the person doing the pressing.

eddyashton: We also have tooling to automerge the file (ie - splat it with your version): https://github.com/microsoft/CCF/pull/3675

Sadly GitHub doesn't support this, so when conflicts do occur this requires a client-merge.

src/consensus/aft/test/driver.h
eddyashton: Yes. The violation that `suffix_collision.3` exposed wasn't easily detected by the existing asserts - we could have healed the network and let it try to catch up for a while and then seen that state-sync was never reached, but that's a little distant. So I added this assertion, which aims to match the semantics of `QuorumLogInv`.

achamayou (approved):

src/consensus/aft/test/driver.h
lemmy: I'd suggest to add a comment that points at the spec's `QuorumLogInv`.

src/consensus/aft/raft.h
```diff
@@ -1426,31 +1426,35 @@ namespace aft
...
+        const auto proposed_match = std::min(r.last_log_idx, state->last_idx);
+        if (proposed_match < node->second.match_idx)
+        {
+          RAFT_FAIL_FMT(
+            "Append entries response to {} from {} attempting to move "
+            "match_idx backwards ({} -> {})",
+            state->node_id,
+            from,
+            node->second.match_idx,
+            proposed_match);
```
jumaffre, line 1455: Am I missing something or should this function return if we get to here? Based on the comment above, it seems that we can't get here (because of `std::min()`), but if we ever did, then `match_idx` would go backwards because `node->second.match_idx = proposed_match;` is then called.

eddyashton: The check above means that `r.last_log_idx` can never cause us to go backwards here, but conceivably `state->last_idx` _could_. I don't think there's ever a case where that can happen, but if it _does_ I think it's more important that we prevent `commit_idx` going past `state->last_idx` (ie - by not returning here, even if it means going backwards).

Put another way, I think we're in trouble if we ever get here, and while returning or throwing an error here are options, I think in such an unexpected situation we should err towards liveness rather than potential stalls/crashes.

lemmy (approved):
LGTM

PR #5329, opened 2023-06-06
Log values of committable_indices in TLA+ validation JSON logging

lemmy:
https://github.com/microsoft/CCF/issues/5057

Towards validating fancy_election.1 and suffix_collision.2.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -429,6 +429,7 @@ namespace aft
       j["state"] = *state;
       j["configurations"] = configurations;
       j["new_configuration"] = Configuration{idx, conf, idx};
+      j["committable_indices"] = last_committable_index();
       RAFT_TRACE_JSON_OUT(j);
 #endif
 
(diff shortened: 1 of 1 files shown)


achamayou:
@lemmy because tracing is enabled in all debug builds, and committable indices gets large in some tests, this change makes them substantially slower, to the point where the CI times out.

I can think of two options, but there may be more:

1. The spec maintains a committable indices based on the committable flag emitted in the messages.
2. We split the build for trace validation, instead of reusing the debug build, to minimise the impact of this change on runtime.

lemmy:
For now it would suffice to just log the max committable index (to compare it to `MaxCommittableIndex(log[i])` for the current node.  Would that be acceptable?

achamayou:
@lemmy yes, that seems fine, the serialisation and I/O cost should be much less.

lemmy:
> @lemmy yes, that seems fine, the serialisation and I/O cost should be much less.

Done

achamayou (approved):

PR #5331, opened 2023-06-07
Serialise snapshots directly in host memory

### LINKED ISSUE #5292: Serialise snapshot directly in host memory
Follow-up from https://github.com/microsoft/CCF/issues/1858. Now that the snapshot is copied to host memory to avoid writing (large) snapshots to the ring buffer, the enclave could serialise the snapshot directly to host memory to reduce the memory pressure on the enclave at the time the snapshot is generated.

jumaffre:
Resolves #5292 

Follow-up from #5273. Now that snapshots are written to host memory, this PR changes the snapshot generation flow so that the snapshot is directly serialised to host memory, rather than serialising it in enclave memory first, and _then_ copy it to host memory. This should decrease memory pressure in the enclave around snapshot generation time. 

Notable changes:
- New backend for KV serialiser to find out snapshot serialised size in advance, without allocating any memory to hold the snapshot (see `raw_serialise.h`). 
- The `Snapshotter` no longer makes use to the thread messaging async subsystem to schedule the serialisation of snapshots on a different thread. Instead, the enclave only requests the host for a memory buffer to hold the serialised snapshot on the hot path, and the snapshot is then serialised asynchronously when the host responds with the memory buffer.

New sequence diagram: 

```mermaid

sequenceDiagram

participant Disk
participant Host
participant Enclave

Enclave->>Enclave: Take store snapshot at N.
Enclave->>Enclave: [NEW] Retrieve snapshot serialised size (without serialising it!)
Enclave->>Host: Request buffer with size S = snapshot serialised size
Host->>Host: Allocate buffer of size S and add to pending_snapshots (host)
Host->>Enclave: Pointer to host-allocated memory
Enclave-->>Enclave: On SGX, sanitise buffer

Enclave->>Enclave: [NEW] Serialise snapshot in host memory 

Enclave-->>Host: Writes serialised snapshot at pointer
Enclave->>Enclave: Snapshot evidence is committed
Enclave->>Host: Writes snapshot receipt on ring buffer
Host->>Disk: Write snapshot + receipt to disk
```


src/node/snapshotter.h
```diff
@@ -85,49 +85,32 @@ namespace ccf
...
     {
-      if (pending_snapshots.size() >= max_pending_snapshots_count)
-      {
-        LOG_FAIL_FMT(
-          "Skipping new snapshot generation as {} snapshots are already "
-          "pending",
-          pending_snapshots.size());
-        return;
-      }
+      auto& snapshot = pending_snapshots[generation_count].snapshot;
```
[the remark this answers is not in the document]
jumaffre: > What happens if we commit a snapshot (produce evidence for it) before the host has responded with a buffer allocation for the initial serialisation?

The snapshot is committed when the evidence itself is committed. If the snapshot hasn't been stored on the host when this happens, the snapshot isn't committed (that's what the `is_stored` bool is for). Later consensus commits will eventually commit the snapshot (and if they don't we limit the number of pending snapshots in flight to not overwhelm the enclave). 

> We're taking a ref to a member of pending_snapshots here, and then moveing it on line 109 - what happens if other things access pending_snapshots[generation_count] now? Are they all happy if the snapshot field is empty? In particular, if serialise_snapshot or similar throws, is pending_snapshots still left in a valid state (ie - we modify it here and from line 147 - what happens if someone else sees the intermediate state)?

All uses of `pending_snapshots` are under the snapshotter's `lock`. Good point about what happens when `record_snapshot_evidence()` throws. I have added a `try ... catch` around it and we now discard the pending snapshot altogether if it ever happens.

src/kv/snapshot.h
```diff
@@ -38,9 +63,33 @@ namespace kv
...
+
+      return serialiser.get_serialised_size();
+    }
+
+    void serialise(
+      const std::shared_ptr<AbstractTxEncryptor>& encryptor,
+      std::span<uint8_t> serialised_snapshot) override
     {
+      CCF_ASSERT_FMT(
+        serialised_snapshot.size() == serialised_size(encryptor),
```
achamayou, line 88: As discussed, passing a single size through all the serialize() calls should allow us to do this check very cheaply and without recomputation: they all bound check, and we should end on 0.

src/kv/generic_serialise_wrapper.h
```diff
@@ -57,6 +57,113 @@ namespace kv
...
+          serialised_private_domain_size;
+      }
+
+      return size;
+    }
+
+    void serialise_domains_impl(
+      const std::vector<uint8_t>& serialised_public_domain,
+      const std::vector<uint8_t>& serialised_private_domain,
+      std::span<uint8_t> buf)
```
jumaffre, line 81: As discussed with @achamayou, we should also use `buf` for the serialised public and serialised private domains, to avoid allocating more memory in the enclave.

jumaffre:
As discussed with @achamayou and @eddyashton, doing this without larger changes to the ledger format is insecure. In particular, the host could tamper with the serialised snapshot on disk before it is hashed and recorded in the ledger in the snapshot evidence table.

I'm closing this PR without merging and we can reconsider this at a later point.

PR #5333, opened 2023-06-07
Make `GenesisGenerator` methods static

### LINKED ISSUE #5296: `GenesisGenerator` and `ShareManager` should be stateless
These classes are really a collection of helper functions for accessing/modifying KV tables in-sync. They are currently unnecessarily stateful as they take a `NetworkTable` ref to access governance tables. This is unnecessary - those tables can be accessed directly by name and type, and then we can call static functions rather than constructing and passing around these objects.

eddyashton:
Resolves #5296.

Trickier to make `ShareManager` fully stateless, since it also uses `LedgerSecrets`. We could pass that in at each call-point, but it looks uglier. I've at least stopped passing `ShareManager` around, so hopefully it's more obvious that it can+should be constructed on-demand.


achamayou:
I think `InternalTxs` or `InternalTxsGenerator` would be preferable. `Manager` is quite vague, and implies more state than there is now.

eddyashton:
Have renamed. I like `Internal` (`Service` is close, but maybe overloaded with other concepts?), but am wary of both `Txs` (our `Tx` types mean something else), and `Generator` (in practice this isn't used for generating a transaction any more, but simply for wrapping access to some tables). So I chose `InternalTablesAccess`. Better than `GenesisGenerator`, worse than free candy.

PR #5336, opened 2023-06-09
Update OpenSSL SHA digest API

jumaffre:
Part of #5291 

The `SHA256_...()` API is deprecated in OpenSSL 3.x (see https://www.openssl.org/docs/man3.1/man3/SHA256_Init.html): 
> The following functions have been deprecated since OpenSSL 3.0, and can be hidden entirely by defining OPENSSL_API_COMPAT with a suitable version value, see [openssl_user_macros(7)](https://www.openssl.org/docs/man3.1/man7/openssl_user_macros.html)

We now make use of the generic `EVP_Digest...()` API, both available with OpenSSL 1.1.x and OpenSSL 3.x.


achamayou:
3.x may also move over, not clear yet, might as well update it.

src/crypto/openssl/hash.h
```diff
@@ -7,7 +7,7 @@
 
 #include <openssl/evp.h>
 #include <openssl/kdf.h>
-#include <openssl/sha.h>
+// #include <openssl/sha.h>
```
achamayou, line 10: ?

eddyashton (approved):

PR #5351, opened 2023-06-15
Improve error reporting in trace validation spec

lemmy:
https://github.com/microsoft/CCF/issues/5057


achamayou:
Why is it -1? This does seem to make all the validation fail, but not with a particularly useful output unfortunately:

```
@!@!@STARTMSG 2104:1 @!@!@
Assumption line 343, col 5 to line 344, col 22 of module Traceccfraft is false.
@!@!@ENDMSG 2104 @!@!@
```

achamayou (approved):

PR #5355, opened 2023-06-15
Validate all traces, and improve output somewhat

achamayou:
Extension to #5351, to be discussed in standup.


tla/Traceccfraft.tla
```diff
@@ -339,22 +332,13 @@ TraceStats ==
...
-    /\ TraceMatched
-    /\ TraceStateSpace
-
-TraceInv ==
-    \* This invariant may or may not hold depending on the level of non-determinism because
-     \* of holes in the log file.
-    TraceStats.distinct <= TraceStats.diameter
+    /\ PrintT("TraceStats.diameter " \o ToString(TraceStats.diameter) \o " TraceStats.distinct " \o ToString(TraceStats.distinct) \o " Len(TraceLog) " \o ToString(Len(TraceLog)))
+    /\
+        \/ TraceMatched
```
lemmy, line 340: `TraceMatched` cannot be an invariant because it is not true or false of a state.  Instead, it is true or false of a set of behaviors.

tla/Traceccfraft.tla
```diff
@@ -339,22 +332,13 @@ TraceStats ==
...
+        \/ PrintT("Error on: " \o JsonFile \o ":" \o ToString(TraceStats.diameter)) = FALSE
```
lemmy, line 341: This could be simplified to `Print("....", FALSE)`

tla/Traceccfraft.tla
lemmy: Btw. I'm somewhat unsatisfied with the current error reporting.  Fundamentally, we face the problem that an error of trace validation is a set of behaviors and not just a single behavior --- TLC has only very limited and the VSCode extension has zero support to report such an error.  At least, TLC can now verify *and report a counterexample* for the following property:

```tla
TraceProp ==
    [](l <= Len(TraceLog) => [](TLCGet("queue") > 0 \/ l > Len(TraceLog)))
```

`TraceProp` is supposed to replace `TraceMatched` after a bit more testing.

achamayou: Ok, I'll make it a property.

eddyashton (approved):

Issue #5313, opened 2023-05-30
Make cchost wait for a complete config on startup, and disallow re-starts if PID file is present


lynshi:
**Is your feature request related to a problem? Please describe.**
For SEV-SNP, we should not attach sidecars to the container running `cchost`. Currently, starting that container with just `cchost` is not ideal. Some situations that should be considered (feel free to add more below as this is meant to be a comprehensive issue):
1. The configuration file may not exist or be completely emitted yet at the time `cchost` attempts to load it.
2. If `cchost` restarts, the configuration may need to be changed. A trivial example is if `cchost` is started in `START` mode, the network is successfully created, and `cchost` restarts. If the configuration file is not modified in time, `cchost` will start in `START` mode again!

**Describe the solution you'd like**
We've discussed a solution to (1). `cchost` should loop on the configuration check before starting. This is conceptually similar to the following `bash` script:
```bash
config_file_path=$1

cchost --check --config ${config_file_path}
code=$?
while [[ $code -ne 0 ]]; do
    echo "Waiting for configuration file to be ready..."
    sleep 10
    cchost --check --config ${config_file_path}
    code=$?
done

cchost --config ${config_file_path}
```

(2) has not been previously discussed. Here are some considerations:
* `cchost` may not be the only process relying on the configuration file, so it may not be safe to simply delete the configuration file upon reading. For example, some other process may read the file to see what mode `cchost` started in, rely on the presence of the file to trigger other logic, etc.
* A simple solution is to modify the configuration file after it has been read in a way that prevents `cchost` from using it in the future. For example, adding a nonsense field that breaks schema validation, or writing some field that indicates the configuration file has been used.

achamayou:
@lynshi `cchost` also writes a pid file: https://microsoft.github.io/CCF/main/operations/configuration.html#output-files

We could make it so that cchost refuses to start if the pid file already exists, that seems preferable to tampering with the configuration.

achamayou:
- [ ] cchost should refuse to start if a PID file is present
- [ ] cchost should wait until either a timeout, or a complete config is present to start

lynshi:
Yep, that sounds good to me!

achamayou:
@lynshi complete and released in https://github.com/microsoft/CCF/releases/tag/ccf-4.0.3

PR #5361, opened 2023-06-15
Do not start if a PID file is present

achamayou:
First sub-point of #5313


src/host/main.cpp
```diff
@@ -206,6 +206,14 @@ int main(int argc, char** argv)
     return static_cast<int>(CLI::ExitCodes::ValidationError);
   }
 
+  std::filesystem::path pid_file_path{config.output_files.pid_file};
+  if (std::filesystem::exists(pid_file_path))
+  {
+    LOG_FATAL_FMT(
+      "PID file {} already exists. Exiting.", pid_file_path.string());
+    return static_cast<int>(CLI::ExitCodes::ValidationError);
```
jumaffre, line 214: `CLI::ExitCodes::FileError` maybe?

PR #5375, opened 2023-06-16
Permit Trace+Debug logging in all non-SGX builds

eddyashton:
We previously (#2404, and related follow-ups) made it so that logging verbosity was controlled at compile time. This is necessary for SGX, but not for other platforms. This PR loosens that behaviour, restoring a run-time `logging.enclave_level` configuration option that allows the enclave's verbosity to be controlled at launch. We still restrict the options which are available (and which are compiled into the binary), but _only on SGX_. On other platforms, we think it is safe to have a single _build_, which contains debug logging, and reliably disable that at launch (in an attested config).

This gives a few big wins for Virtual and SNP:
- The release artifacts can be used for deeper debugging, as users can pass now pass `--enclave-log-level` to `sandbox.sh` to get verbose enclave logging (previously _only_ available in the `sgx_unsafe` artifact).
- Enabling and disabling CMake's `VERBOSE_LOGGING` on these platforms doesn't change the compile definitions, only the args passed to e2e tests. So you can quickly enable verbose logging to get more debug information, without a slow rebuild.

~~To avoid too many annoying changes/duplicated checks, I've retained the original definition names. This means there's potential confusion - there's a `VERBOSE_LOGGING` CMake option _and_ a `VERBOSE_LOGGING` C++ preprocessor definition, but the latter is always-present unless we're on SGX. It may be worth renaming the latter to `UNSAFE_LOGGING` or something to avoid that confusion.~~ _`VERBOSE_LOGGING` is the CMake option, which also affects the runtime arg used by the e2e tests. If we're building for SGX and don't have `VERBOSE_LOGGING`, we add the compile definition `CCF_DISABLE_VERBOSE_LOGGING`, which is the inverse of the previously preproc def `VERBOSE_LOGGING`_

_ETA - This option is now passed as a CLI arg to `cchost`, where it should be covered by attestation, rather than in a potentially-unattested config file_


achamayou:
@eddyashton I am worried about bundling this in with items of configuration such as directories, IPs, domain names and ports, which may be node-specific. Even if the security policy correctly captures the configuration (which isn't a given, since existing implementations typically do not allow measuring mounts, except for the empty special case), it will differ by node, and effectively require that each node addition be preceded by governance to add the corresponding measurement.

It seems to me that the requirements for this bit of configuration are that:

1. it is attested
2. it does not vary per node, to avoid governance churn

A generic solution would be to create an attested configuration, which must come from somewhere that's measured (ie. the container right now, I don't believe there are other practical options). That's still relatively easy to get wrong.

A less generic, but low impact solution, is to make this a CLI argument (either directly, or add a CLI enclave log-level cap). CLI arguments are well captured by the security policy, it is difficult to get that wrong.

eddyashton:
@achamayou I've updated this PR, and the option is now passed on the CLI rather than in the JSON configuration file. Please re-review.

achamayou:
> It may be worth renaming the latter to `UNSAFE_LOGGING` or something to avoid that confusion.

We should also take the opportunity to `CCF_`-prefix.

achamayou (approved):

eddyashton:
@achamayou I've renamed to separate the CMake option and the preprocessor definition. CMake's `VERBOSE_LOGGING` stays with the same name - this also affects how verbose the e2e tests are, so I figure its useful to retain the muscle memory here (with the win that it no longer requires a rebuild).

achamayou:
@eddyashton a changelog entry is probably worthwhile for this change.

PR #5379, opened 2023-06-16
Add config timeout option

achamayou:
Second part of #5313.


tests/e2e_operations.py
```diff
@@ -354,6 +355,41 @@ def run_tls_san_checks(args):
...
+    os.remove(os.path.join(start_node_path, "service_cert.pem"))
+    # Save configuration
+    shutil.copy(
+        os.path.join(start_node_path, "0.config.json"),
+        os.path.join(start_node_path, "0.config.json.bak"),
+    )
+    # Replace it with a prefix
+    with open(os.path.join(start_node_path, "0.config.json"), "w") as f:
+        f.write("{")
+    LOG.info("Attempt to start node with a partial config under {start_node_path}")
```
eddyashton, line 375:
```suggestion
    LOG.info(f"Attempt to start node with a partial config under {start_node_path}")
```

Can we test that the node also spins (happily) if there's _no_ file for a few seconds??

src/host/main.cpp
```diff
@@ -102,16 +106,34 @@ int main(int argc, char** argv)
...
+  std::string config_parsing_error = "";
+  do
   {
-    config_json = nlohmann::json::parse(config_str);
-  }
-  catch (const std::exception& e)
+    std::string config_str = files::slurp_string(
+      config_file_path,
+      true /* return an empty string if the file does not exist */);
+    LOG_INFO_FMT("READ CONFIG {}", config_str);
```
eddyashton, line 121:
```suggestion
```

Looks like debug logging?

src/host/main.cpp
```diff
@@ -102,16 +106,34 @@ int main(int argc, char** argv)
...
+    try
+    {
+      config_json = nlohmann::json::parse(config_str);
+      config_parsing_error = "";
```
eddyashton, line 125: Should there be a `break` in this path? Why are we still looping if parse succeeded?

src/host/main.cpp
```diff
@@ -102,16 +106,34 @@ int main(int argc, char** argv)
...
+    }
+    catch (const std::exception& e)
+    {
+      config_parsing_error = fmt::format(
+        "Error parsing configuration file {}: {}", config_file_path, e.what());
```
eddyashton, line 130: Can we add a short sleep in this failed case?

tests/e2e_operations.py
achamayou: Sure

src/host/main.cpp
achamayou: Thank you, yes

src/host/main.cpp
achamayou: Yes indeed.

src/host/main.cpp
achamayou: Sure

tests/e2e_operations.py
```diff
@@ -354,6 +355,45 @@ def run_tls_san_checks(args):
...
+    os.remove(os.path.join(start_node_path, "node.pid"))
+    os.remove(os.path.join(start_node_path, "service_cert.pem"))
+    # Move configuration
+    shutil.move(
+        os.path.join(start_node_path, "0.config.json"),
+        os.path.join(start_node_path, "0.config.json.bak"),
+    )
+    LOG.info("No config at all")
+    assert not os.exists(os.path.join(start_node_path, "0.config.json"))
+    LOG.info("Attempt to start node without a config under {start_node_path}")
```
eddyashton, line 374:
```suggestion
    LOG.info(f"Attempt to start node without a config under {start_node_path}")
```

eddyashton (approved):

tests/e2e_operations.py
```diff
@@ -354,6 +355,45 @@ def run_tls_san_checks(args):
...
+    shutil.rmtree(os.path.join(start_node_path, "0.ledger"))
+    os.remove(os.path.join(start_node_path, "node.pid"))
+    os.remove(os.path.join(start_node_path, "service_cert.pem"))
+    # Move configuration
+    shutil.move(
+        os.path.join(start_node_path, "0.config.json"),
+        os.path.join(start_node_path, "0.config.json.bak"),
+    )
+    LOG.info("No config at all")
+    assert not os.exists(os.path.join(start_node_path, "0.config.json"))
```
achamayou, line 373:
```suggestion
    assert not os.path.exists(os.path.join(start_node_path, "0.config.json"))
```

PR #5394, opened 2023-06-22
Improvements to basic perf

achamayou:
Split out generation, add in/out throughputs in Mb/s.

- [x] split out analysis (agg written to file)
- [x] tidy up with_columns
- [x] add scenarios from existing benchmarks


tests/infra/basicperf.py
```diff
@@ -341,10 +411,24 @@ def cli_args():
...
+    )
 
     return infra.e2e_args.cli_args(parser=parser, accept_unknown=False)
 
 
 if __name__ == "__main__":
     args = cli_args()
-    run(args)
+    if args.rw_mix:
+        run(args, RWMix(1000, 0.5))
```
heidihoward, line 444: this is probably a silly question but where does the rw-mix cli arg get passed to RWMix?

PR #5405, opened 2023-06-29
Update tls_client.h exception handling has been added

pe32:
In this code version, exception handling has been added to the TlsClient constructor. If the initialization fails, such as when calling init(), an exception will be thrown, and the connected flag will be set to false.


achamayou:
@pe32 unless I am missing something, init() sets connected last, so there isn't a risk that it would be set if anything throws?

Issue #5050, opened 2023-02-22
Fix rdrand ASAN error


takuro-sato:
Part of https://github.com/microsoft/CCF/issues/4994

There is a error when ASAN is enabled.
We don't know if it's actually problematic yet.

## Error when ASAN is enabled
```bash
$ cmake -GNinja -DCOMPILE_TARGET=virtual -DWORKER_THREADS=2  -DCMAKE_BUILD_TYPE=Debug -DSAN=ON -DLVI_MITIGATIONS=OFF ..
$ ninja
$ sudo ./tests.sh -VV  --no-compress-output --verbose -L partitions -C partitions
```

```
56: RuntimeError: Error starting node 0                                                                                                                                                                           [80/1803]
56: 14:37:33.786 | INFO     | infra.network:network:1477 - Stopping network                                                                                                                                                
56: 14:37:33.787 | INFO     | infra.node:stop:439 - Stopping node 0                                                                                                                                                        
56: 14:37:33.787 | INFO     | infra.remote:stop:523 - [127.251.132.13] closing                                                                                                                                             
56: 14:37:33.788 | ERROR    | infra.remote:log_errors:103 - Contents of /home/takurosato/CCF/build/workspace/partitions_cft_0/err:                                                                                         
56: AddressSanitizer:DEADLYSIGNAL                                                                                                                                                                                          
56: =================================================================                                                                                                                                                      
56: ==4083026==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x7ff6e9c777da bp 0x7ffcd519a4c0 sp 0x7ffcd519a420 T0)                                                                                                 
56: ==4083026==The signal is caused by a READ memory access.                                                                                                                                                               
56: ==4083026==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.                                            
56:     #0 0x7ff6e9c777da in crypto::IntelDRNG::rdrand64_step(unsigned long*) /home/takurosato/CCF/build/../include/ccf/crypto/entropy.h:103:7                                                                             
56:     #1 0x7ff6e9c7770a in crypto::IntelDRNG::rdrand64_retry(unsigned int, unsigned long*) /home/takurosato/CCF/build/../include/ccf/crypto/entropy.h:142:13                                                             
[...]

PR #5408, opened 2023-07-03
Fix unsafe inline assembly usage

achamayou:
This is a fix for a potential register re-use issue spotted by ASAN (as part of #4994), which had previously been ignored (#5050) because we did not understand what caused it.

Thank you to @neldredge for pointing our incorrect understand of input/output register usage in line assembly, and suggesting a further optimisation.

Once dependencies allow for it, we should switch to intrinsics rather than use inline assembly.


neldredge:
The `rdseed` functions have the same bug, so you probably want to fix them too.

eddyashton (approved):

PR #5413, opened 2023-07-04
Store deadlock fix

achamayou:
Tentative fix for a store deadlock hit during performance testing.

The ReservedTx flow acquires the sequence store_lock, version_lock, map_locks. But the regular Tx flow instead acquires (mapset_lock), map_locks, version_lock. There is also a map_locks, version_lock order in rollback, so this is changing Store::commit not to hold the version lock during the final resolution of transactions where ReservedTx acquire map_locks.

The sequence of steps taken out of the version_lock coverage then need to be made robust to a potential concurrent rollback() or commit().


src/kv/test/stub_consensus.h
```diff
@@ -22,8 +22,7 @@ namespace kv::test
   public:
     std::vector<BatchVector::value_type> replica;
     ccf::TxID committed_txid = {};
-    ccf::View current_view = 0;
-
+    ccf::View current_view = 2;
```
achamayou, line 25:
```suggestion
    ccf::View current_view = 0;
```

src/kv/apply_changes.h
```diff
@@ -70,7 +71,15 @@ namespace kv
     }
 
     bool ok = true;
-    if (has_writes)
+
+    if (expected_rollback_count.has_value())
+    {
+      // Assumes that all maps in the transaction point to the same store
+      auto store = changes.begin()->second.map->get_store();
```
eddyashton, line 78: Assume we can never get here when `changes` is empty? Else `changes.begin()->second` will go boom.

src/kv/apply_changes.h
```diff
@@ -70,7 +71,15 @@ namespace kv
...
+      ok = store->check_rollback_count(expected_rollback_count.value());
```
eddyashton, line 79: Think this deserves a comment, explaining the secret trick here is that we're checking this _while holding locks on some maps_. So any rollback which tries to happen _after_ this check will get queued (waiting for _this_ block to release the map locks that the rollback requires), and thus we get mutual exclusion.

src/kv/store.h
```diff
@@ -679,8 +679,10 @@ namespace kv
 
       term_of_next_version = t;
       auto h = get_history();
+      LOG_TRACE_FMT("Getting history");
       if (h)
       {
+        LOG_TRACE_FMT("Got history");
```
eddyashton, line 685: Can we remove this debug logging?

src/node/history.h
```diff
@@ -678,7 +680,8 @@ namespace ccf
     {
       std::lock_guard<ccf::pal::Mutex> guard(state_lock);
       return replicated_state_tree.serialise(
-        replicated_state_tree.begin_index(), to);
+        replicated_state_tree.begin_index(),
+        std::min(to, replicated_state_tree.end_index()));
```
eddyashton, line 684: This behaviour (silently serialising a different tree than was requested) feels dangerous. I assume someone up-the-stack is discarding the return value if `to` is no longer available - can we return an empty response in that case instead?

src/kv/apply_changes.h
```diff
@@ -70,7 +71,15 @@ namespace kv
...
+    }
+
+    if (ok && has_writes)
```
eddyashton, line 82: I think the loop below, on line 94, can also be skipped entirely `if (!ok)`. All it is doing is checking _another_ reason that `ok` might become `false`. (This was true previously, but I just noticed it now).

src/kv/committable_tx.h
```diff
@@ -391,14 +391,21 @@ namespace kv
...
+  private:
+    Version rollback_count = 0;
+
   public:
     ReservedTx(
-      AbstractStore* _store, Term read_term, const TxID& reserved_tx_id) :
+      AbstractStore* _store,
+      Term read_term,
+      const TxID& reserved_tx_id,
+      Version rollback_count_) :
```
eddyashton, line 402: Just a query: Can we use one of the existing `term` values here (either the `read_term` or `reserved_tx_id.term`), rather than passing an additional `rollback_count_`? Is `reserved_tx_id.term == store.commit_view()` the same as `store.check_rollback_count(expected_rollback_count)`?

src/kv/apply_changes.h
achamayou: Not in practice, because the only caller that makes use of the expected_rollback_count is the signature Tx, which definitely makes exactly two changes every time. But I agree that making the code defensive and adding a comment is preferable.

src/kv/apply_changes.h
achamayou: Good point, I will add a comment to that effect.

src/kv/store.h
achamayou: Yes

src/kv/apply_changes.h
achamayou: Yes

src/kv/committable_tx.h
achamayou: I think so, yes. There's not a lot of reason left to have rollback count now, as discussed, but if it stays it's appropriate for this use. I don't mind switching to term if we think it's better, but I don't want to bundle the removal of the rollback counter with this. And no, because the check must be under lock, otherwise we have a race condition.

eddyashton: I agree it's not sensible to bundle _removal_ of the rollback counter into this change, but if we could avoid extending use of the rollback counter (passing as an additional arg here), that'd be nice.

I don't see the distinction if the check is made under a lock, because the action we take after the check is always _outside_ of the version lock (ie - any call into `store`, whether `check_rollback_count` or `commit_view`, might temporarily take `version_lock`, but if we do something _here_ then it'll be after `version_lock` is released) - we think that's safe because the map lock(s) prevent racing rollbacks.

src/kv/store.h
achamayou: I'm tempted to define a logging macro that the TODO detection thing catches, to avoid this happening again.

achamayou: https://github.com/microsoft/CCF/pull/5415

src/kv/committable_tx.h
achamayou: I hadn't noticed term_of_next_version was atomic, I agree it's equivalent.

> but if we could avoid extending use of the rollback counter (passing as an additional arg here), that'd be nice.

Here we're explicitly trying to detect a rollback though, and the rollback counter has just the right semantics and name for that, no more no less. We don't really care what term it is. I'm going to suggest this can be revisited separately if we ever get around to removing the rollback counter.

eddyashton (approved):
LGTM

PR #5415, opened 2023-07-06
Add LOG_DEV_FMT

achamayou:
When adding log lines for dev/debug purposes, one can now use `LOG_DEV_FMT`, safe in the knowledge that the formatter will catch occurrences automatically.


include/ccf/ds/logger.h
```diff
@@ -387,6 +387,8 @@ namespace logger
     CCF_LOGGER_DEPRECATE(LOG_TRACE_FMT) CCF_LOG_FMT(TRACE, "")
 #  define LOG_DEBUG_FMT \
     CCF_LOGGER_DEPRECATE(LOG_DEBUG_FMT) CCF_LOG_FMT(DEBUG, "")
+#  define LOG_DEV_FMT \
+    CCF_LOGGER_DEPRECATE(LOG_TRACE_FMT) CCF_LOG_FMT(TRACE, "")
```
eddyashton, line 391: Might as well do this at INFO level, so you don't _also_ need verbose logging enabled? And making the second argument `"dev"` will add a `[dev]` tag making them more easily greppable.

achamayou, line 391:
```suggestion
    CCF_LOGGER_DEPRECATE(LOG_INFO_FMT) CCF_LOG_FMT(INFO, "dev")
```

PR #5444, opened 2023-07-12
Remove TypeSpec

eddyashton:
Our format checks are failing because of an inability to resolve TypeSpec packages, likely from new versions released yesterday. Rather than debugging this, we're simply removing the local copy - this spec lives in the [azure-rest-api-specs repo](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/confidentialledger/Microsoft.ManagedCcf) now, and we don't need a local copy.


jumaffre (approved):
Should we keep a pointer to https://github.com/Azure/azure-rest-api-specs/tree/main/specification/confidentialledger/Microsoft.ManagedCcf somewhere in our docs?

eddyashton:
> Should we keep a pointer to https://github.com/Azure/azure-rest-api-specs/tree/main/specification/confidentialledger/Microsoft.ManagedCcf somewhere in our docs?

We should, but I think it makes sense to add this with the implementation of that gov API. At the moment it would just be "here's a link to a spec we intend to implement", since the spec doesn't describe the current API.

PR #5453, opened 2023-07-13
Include intermediate certs in TLS handshake

eddyashton:
This doesn't make a difference for most of our end-to-end tests, but does for an ACME-endorsed interface. We end up with `Service Cert <- Intermediate <- Root` (potentially multiple intermediates, but one is sufficient for demo). We store in the KV, and attempt to set as the server cert, the chain (/bundle) `[Service Cert] \n [Intermediate]`, but when passing that to OpenSSL we actually end up calling `PEM_read_bio_X509` which only extracts `[Service Cert]`, and drop the `Intermediate` on the ground.

That means that a connecting browser, which likely only has `Root` in their trust store, can't build a complete chain and won't trust CCF.

The solution is retain the chain (an `X509_STACK` in OpenSSLese), though we have to split the bundle and remember it separately, and then pass it through to `use_cert_and_key`.

The end-to-end test we have using Pebble to obtain ACME endorsements had missed this, because it fetched both the Root _and_ Intermediate from Pebble, and added them to the trust store for the test connection. I've changed that to only fetch the Root, which should match what we get in a real service.

A pure digression, but something I've not been able to find documentation on - I assume that for a real ACME endorsement we end up with a longer chain, with some overlap between the ACME service's response and the CAs in a system's trust store. ie if we end up with `Service Cert <- This Let's Encrypt Instance <- Let's Encrypt Master <- ISRG Root`, there's a chance that `Let's Encrypt` returns a chain including `Let's Encrypt Master`, and that's also in a system trust store. So maybe there's some overlap and redundancy, if we blindly retain the entire chain. But I don't think we can make any assessment of where the cut-off point is, so we assume the endorser/ACME protocol is handling it sensibly, and retain whatever they give us.


tests/acme_endorsement.py
```diff
@@ -92,9 +91,12 @@ def wait_for_certificates(
                     assert network_public_key == cert_public_key
                     num_ok += 1
                 except Exception as ex:
-                    LOG.trace(f"Likely expected exception: {ex}")
+                    LOG.warning(f"Likely temprorary exception: {ex}")
```
achamayou, line 94:
```suggestion
                    LOG.warning(f"Likely temporary exception: {ex}")
```

achamayou (approved):

achamayou (approved):

Issue #5459, opened 2023-07-17
Expose COSE authn identity content in JS/TS


achamayou:
Expose https://github.com/microsoft/CCF/blob/375274d0b5fb0215bcf1850d1c4acc751240187f/include/ccf/endpoints/authentication/cose_auth.h#L28 in the JS authn object, so that endpoints using `user_cose_sign1` can obtain the payload easily without having re-parse the COSE payload.

PR #5465, opened 2023-07-18
Expose COSE identity content in JS/TS apps

jumaffre:
Resolves https://github.com/microsoft/CCF/issues/5459

The COSE `content` is now exposed to JavaScript/TypeScript apps for `user_cose_sign1` authenticated endpoints. 

TODO: 
- [x] New end-to-end test to test that these values are set correctly.

### DIFF
--- a/js/ccf-app/src/endpoints.ts
+++ b/js/ccf-app/src/endpoints.ts
@@ -147,7 +147,7 @@ interface UserMemberCOSEAuthIdentityCommon {
      * COSE content
      */
     content: ArrayBuffer;
-  }
+  };
 }
 
 export interface MemberCOSESign1AuthnIdentity
--- a/js/ccf-app/src/endpoints.ts
+++ b/js/ccf-app/src/endpoints.ts
@@ -147,6 +147,16 @@ interface UserMemberCOSEAuthIdentityCommon {
    * COSE content
    */
   content: ArrayBuffer;
+
+  /**
+   * COSE envelope
+  */
+  envelope: ArrayBuffer;
+
+  /**
+   * COSE signature
+   */
+  signature: ArrayBuffer;
 }
 
 export interface MemberCOSESign1AuthnIdentity
--- a/js/ccf-app/src/endpoints.ts
+++ b/js/ccf-app/src/endpoints.ts
@@ -141,13 +141,21 @@ export interface MemberCertAuthnIdentity extends UserMemberAuthnIdentityCommon {
   policy: "member_cert";
 }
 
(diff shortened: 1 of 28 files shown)


js/ccf-app/src/endpoints.ts
```diff
@@ -141,13 +141,32 @@ export interface MemberCertAuthnIdentity extends UserMemberAuthnIdentityCommon {
   policy: "member_cert";
 }
 
+interface UserMemberCOSEAuthIdentityCommon {
```
achamayou, line 144: Not sure the UserMember prefix adds much here?

src/apps/js_generic/js_generic_base.cpp
```diff
@@ -98,6 +98,23 @@ namespace ccfapp
         policy_name = get_policy_name_from_ident(user_cose_ident);
         id = user_cose_ident->user_id;
         is_member = false;
+
+        auto cose = ctx.new_obj();
+        cose.set(
+          "content",
+          ctx.new_array_buffer_copy(
+            user_cose_ident->content.data(), user_cose_ident->content.size()));
+        cose.set(
```
achamayou, line 107: If we are able to avoid copies somehow that would be great. I am not sure we want re-copy the envelope in particular, since that contains the whole content right now (and for the foreseeable future).

js/ccf-app/src/endpoints.ts
jumaffre: I followed the same pattern as `UserMemberAuthnIdentityCommon`. Happy to change it though.

src/apps/js_generic/js_generic_base.cpp
achamayou: > the JS runtime now owns the buffer, which is not what we want I don't think?

No, definitely. What would be ideal is a non-owning read-only ArrayBuffer, and if there isn't one I think it might be worth making one in quickjs, for many reasons (eg. large KV values).

> We could also only pass content but not envelope to minimise the number of bytes copied?

Let's do that for now, it's the user scenario we have. We can revisit this later and add other things if needed.

achamayou (approved):

tests/infra/clients.py
```diff
@@ -102,7 +102,7 @@ def escape_loguru_tags(s):
     return loguru_tag_regex.sub(lambda match: f"\\{match[0]}", s)
 
 
-def truncate(string: str, max_len: int = 256):
+def truncate(string: str, max_len: int = 2560):
```
jumaffre, line 105:
```suggestion
def truncate(string: str, max_len: int = 256):
```

PR #5482, opened 2023-07-28
Fix bug in TLS context - allow incoming write buffer to be relocated

eddyashton:
Spotted some odd TLS errors during multi-threaded perf tests, where clients were being disconnected. Tracked it down to this. There were cases where we got a `WANTS_WRITE/WANTS_READ` response from an attempted write, which means a handshake is still in-progress or similar, and we need to retry the write later. The default semantics are very strict, and we must "retry the write" with exactly the same pointer and size. Our source for this is a `pending_writes` buffer, so if it got resized and reallocated, we'd pass a new pointer and this would blow up. There's a magic flag to make it less strict, which it seems we want. This is well explained in [this StackOverflow answer](https://stackoverflow.com/questions/2997218/why-am-i-getting-error1409f07fssl-routinesssl3-write-pending-bad-write-retr), once you've convinced OpenSSL to log its errors correctly.

### DIFF
--- a/src/enclave/tls_session.h
+++ b/src/enclave/tls_session.h
@@ -417,7 +417,7 @@ namespace ccf
         else
         {
           LOG_TRACE_FMT(
-            "TLS {} on flush: {}", session_id, tls::error_string(r));
+            "TLS session {} error on flush: {}", session_id, -r);
           stop(error);
         }
       }
--- a/src/enclave/tls_session.h
+++ b/src/enclave/tls_session.h
@@ -416,8 +416,7 @@ namespace ccf
         }
         else
         {
-          LOG_TRACE_FMT(
-            "TLS session {} error on flush: {}", session_id, -r);
+          LOG_TRACE_FMT("TLS session {} error on flush: {}", session_id, -r);
           stop(error);
         }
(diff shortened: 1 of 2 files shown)


src/enclave/tls_session.h
```diff
@@ -417,7 +417,7 @@ namespace ccf
         else
         {
           LOG_TRACE_FMT(
-            "TLS {} on flush: {}", session_id, tls::error_string(r));
+            "TLS session {} error on flush: {}", session_id, -r);
```
eddyashton, line 420: This return code is (a negation of) a return value from `SSL_get_error`, that's _not_ the same domain as `ERR_get_error`. We can pass the opaque numbers from the latter to `error_string` (a wrapper for `ERR_error_string_n`), but not from the former. Passing the former produces extremely unhelpful junk. There doesn't seem to be any helpful stringification for these errors, so its a case of finding the macro definitions in the OpenSSL source and working out what went wrong (or, better, calling `ERR_get_error` to flush the error stack and hope it shows something related, but that's a more awkward rewrite).

achamayou (approved):

PR #5483, opened 2023-07-28
Patch to rw_mix in `basicperf.py`

heidihoward:
Pass `args.rw_mix` to `RWMix` even if zero in `basicperf.py`


tests/infra/basicperf.py
```diff
@@ -448,7 +448,7 @@ def cli_args():
 
 if __name__ == "__main__":
     args = cli_args()
-    if args.rw_mix:
-        run(args, RWMix(1000, 0.5))
-    else:
+    if args.rw_mix == None:
```
eddyashton, line 451:
```suggestion
    if args.rw_mix is None:
```

Minor Python-style improvement - `None` is a singletony thing that can be `is`-compared.

eddyashton (approved):

PR #5484, opened 2023-07-28
Use Timestamp Parquet type for recording Piccolo's timestamps

eddyashton:
These were previously floats, resulting in hard-to-read tables and unlabelled latency units:

image

By using the Parquet Timestamp type (storing milliseconds since the Linux epoch), we get nice rendering in the output table (and a smattering of type safety)

image

This lets us compare when the clients were running at a glance, and recognise if one was running significantly before/after the others.

This PR includes removing a few sorts and replacing them with `min()`/`max()` where possible.

### DIFF
--- a/tests/infra/basicperf.py
+++ b/tests/infra/basicperf.py
@@ -301,17 +301,16 @@ def run(args, append_messages):
                     with open(agg_path, "wb") as f:
                         agg.write_parquet(f)
                     print(f"Aggregated results written to {agg_path}")
-                    start_send = agg["sendTime"].sort()[0]
-                    end_recv = agg["receiveTime"].sort()[-1]
-                    throughput = len(agg) / (end_recv - start_send)
+                    start_send = agg["sendTime"].min()
+                    end_recv = agg["receiveTime"].max()
+                    duration_s = (end_recv - start_send).total_seconds()
+                    throughput = len(agg) / duration_s
                     print(f"Average throughput: {throughput:.2f} tx/s")
-                    byte_input = (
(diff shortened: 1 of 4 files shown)


tests/infra/basicperf.py
```diff
@@ -301,36 +301,34 @@ def table():
...
                         .rename({"count": "sent"})
                     )
-                    recv = agg["receiveTime"].sort()
                     recv_per_sec = (
                         agg.with_columns(
-                            (pl.col("receiveTime").alias("second") - recv[0]).cast(
-                                pl.Int64
-                            )
+                            (
+                                (pl.col("receiveTime").alias("second") - start_send)
```
eddyashton, line 329: Switching `recv[0]` to `start_send` changes what we're graphing here, but I think this is correct? If there's a significant skew between the first send and the first receive, we want an empty first receive row. We get that by bucketing them into seconds based on the same start point.

tests/infra/basicperf.py
```diff
@@ -434,7 +432,7 @@ def cli_args():
     parser.add_argument(
         "--client-timeout-s",
         help="Number of seconds after which unresponsive clients are shut down",
-        default=90,
+        default=180,
```
eddyashton, line 435:
```suggestion
        default=90,
```

heidihoward (approved):

tests/infra/basicperf.py
heidihoward: agreed, i also prefer starting with `start_send`

achamayou (approved):

PR #5488, opened 2023-07-31
Remove hardcoded overrides

achamayou:


tests/infra/basicperf.py
```diff
@@ -440,8 +436,16 @@ def cli_args():
...
+    parser.add_argument(
+        "--max-writes-ahead",
+        help="Maximum number of writes to send to the server without waiting for a response",
+        type=int,
+        default=1000,
+    )
 
-    return infra.e2e_args.cli_args(parser=parser, accept_unknown=False)
+    return infra.e2e_args.cli_args(
+        parser=parser, accept_unknown=False, ledger_chunk_bytes_override="5MB"
```
achamayou, line 447: We need to keep large chunks, possibly even larger, otherwise the signature interval effectively gets capped, since chunks must end on a signature.

eddyashton: Can't we pass this in `CMakeLists.txt`, as `ADDITIONAL_ARGS ... --ledger-chunk-bytes 5MB` passed to `basicperf.py`? I assume that's the basis for any modified test runs.

heidihoward (approved):
lgtm

tests/infra/basicperf.py
achamayou: We could but I think it's slightly worse UX when using the perf test than an override.

PR #5493, opened 2023-08-01
Split preloading and client defs

achamayou:
Now supports things like:

```
--client-def "1,write,200000,primary" --client-def "3,read,100000,backup"
```

Still needs:

- [x] support for `read`
- [x] support for `backup`
- [x] support for `any`
- [x] support for `rwmix`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1446,12 +1446,12 @@ if(BUILD_TESTS)
                       "1,write,200000,primary"
     )
 
-    add_perf_test(
+    add_piccolo_test(
       NAME pi_basic_js
       PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
       CLIENT_BIN ./submit
       ADDITIONAL_ARGS --js-app-bundle ${CMAKE_SOURCE_DIR}/samples/apps/basic/js
-                      --repetitions 100000
+                      --client-def "1,write,200000,primary"
     )
 
     if(WORKER_THREADS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1460,7 +1460,7 @@ if(BUILD_TESTS)
         PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
         CLIENT_BIN ./submit
         ADDITIONAL_ARGS --package "samples/apps/basic/libbasic" --client-def
-                        "${WORKER_THREADS},write,100000,primary"
+                        "${WORKER_THREADS},write,100000,any"
       )
     endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1443,7 +1443,7 @@ if(BUILD_TESTS)
       PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
       CLIENT_BIN ./submit
       ADDITIONAL_ARGS --package "samples/apps/basic/libbasic" --client-def
-                      "1,write,200000,primary"
+                      "1,write,100000,primary"
     )
 
     add_piccolo_test(
@@ -1451,7 +1451,7 @@ if(BUILD_TESTS)
       PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
       CLIENT_BIN ./submit
       ADDITIONAL_ARGS --js-app-bundle ${CMAKE_SOURCE_DIR}/samples/apps/basic/js
-                      --client-def "1,write,200000,primary"
+                      --client-def "1,write,100000,primary"
     )
 
     if(WORKER_THREADS)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -1438,29 +1438,29 @@ if(BUILD_TESTS)
...
     )
 
-    add_perf_test(
+    add_piccolo_test(
       NAME pi_basic_js
       PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
       CLIENT_BIN ./submit
       ADDITIONAL_ARGS --js-app-bundle ${CMAKE_SOURCE_DIR}/samples/apps/basic/js
-                      --repetitions 100000
+                      --client-def "1,write,200000,primary"
```
eddyashton, line 1454: This changes repetitions from `100000` to `200000`, for both `pi_basic` and `pi_basic_js`. That may have knock-on effects, and seems unintentional?

tests/infra/basicperf.py
```diff
@@ -383,72 +406,40 @@ def cli_args():
...
         help="Maximum number of writes to send to the server without waiting for a response",
         type=int,
         default=1000,
     )
-
+    parser.add_argument(
+        "--key-space-size",
+        help="Size of the key space to be pre-populated and which writes and reads will be performed on",
+        type=int,
+        default=100,
```
eddyashton, line 430: I think this is reducing the key-space in existing runs? Looks like `RWMix` was using 1000 at least, although `write_to_random_keys` was 100. I think that's fine, and 100 is a sensible default, just noting here in case it causes unexpected shift.

tests/infra/basicperf.py
```diff
@@ -137,28 +152,24 @@ def __call__(
...
+        for key in space:
+            r = c.put(
+                f"/records/{key}",
+                body=f"{hashlib.md5(key.encode()).hexdigest()}",
+                headers={"content-type": "text/plain"},
+                log_capture=[],
+            )
+            assert r.status_code == http.HTTPStatus.NO_CONTENT.value, r
+    LOG.info("Key space created and filled")
+    return space
```
eddyashton, line 168: Not critical, but this is going to be an unnecessary stall for larger key spaces - we should at some point add a single endpoint so this is a single, batched call.

tests/infra/basicperf.py
achamayou: Yes, this is done to shorten the pre-launch, which is currently slower because it's using the Python client. Once that's done through either a custom endpoint or faster somehow, we can increase the key range again.

CMakeLists.txt
achamayou: Will fix

CMakeLists.txt
```diff
@@ -1438,29 +1438,29 @@ if(BUILD_TESTS)
...
     if(WORKER_THREADS)
-      add_perf_test(
+      add_piccolo_test(
         NAME pi_basic_mt
         PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
         CLIENT_BIN ./submit
-        ADDITIONAL_ARGS --package "samples/apps/basic/libbasic" --repetitions
-                        200000 --num-localhost-clients ${WORKER_THREADS}
+        ADDITIONAL_ARGS --package "samples/apps/basic/libbasic" --client-def
+                        "${WORKER_THREADS},write,100000,primary"
```
achamayou, line 1463:
```suggestion
                        "${WORKER_THREADS},write,100000,any"
```

heidihoward (approved):
This lgtm and seems to match the interface I was expecting. Thank you!

eddyashton (approved):

tests/infra/basicperf.py
achamayou: For the record, I did this with governance, but hit the limitation we place on governance not writing to private application tables (https://microsoft.github.io/CCF/main/audit/read_write_restrictions.html).

I'll move this to an application endpoint, in both cpp and js.

As discussed, I suspect we may re-open this discussion one day, specifically for situations where application key material is provisioned/re-keyed from governance. But we don't need to address this right now.

tests/infra/basicperf.py
```diff
@@ -61,27 +44,64 @@ def configure_remote_client(args, client_id, client_host, common_dir):
...
 
 
+def append_to_msgs(definition, key_space, iterations, msgs, additional_headers):
+    if definition == "write":
+        return write_to_key_space(key_space, iterations, msgs, additional_headers)
+    elif definition == "read":
+        return read_from_key_space(key_space, iterations, msgs, additional_headers)
+    elif definition.startswith("rwmix:"):
+        _, ratio = definition.split(":")
+        return RWMix(min([1000, iterations]), float(ratio))(
```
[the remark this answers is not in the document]
achamayou: That's a good point, but I think we should be even stricter and demand a multiple of the batch size.

cmake/common.cmake
```diff
@@ -266,6 +266,85 @@ function(add_perf_test)
   )
 endfunction()
 
+# Helper for building end-to-end perf tests using the python infrastucture
+function(add_piccolo_test)
```
achamayou, line 270: We've been dragging around unused arguments for a while now, to stay compatible with the old perftest binary. Since we're re-shuffling the CLI in this PR anyway, might as well split.

PR #5508, opened 2023-08-07
Add HMAC support to JS `crypto.sign`

### LINKED ISSUE #5365: Add HMAC support to TS API
We have `crypto::hmac` in C++, but no equivalent for JS/TS.

For similarity to [`SubtleCrypto` Web API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign), this would be `crypto.sign("HMAC", key, data)`

eddyashton:
Resolves #5365.

Like our other `.sign` calls, it looks a little like `SubtleCrypto.sign`, but returns synchronous result rather than a `Promise`. In this `"HMAC"` case, since we don't mimic `SubtleCrypto`s key objects we don't get a hash-algorithm embedded in the key for free, and have to pass it.

In other words, rather than:
```
let key = crypto.generateKey(); // Returns a Promise<CryptoKey>
crypto.sign("HMAC", key, data);
```
it is:
```
let key = getKey(); // Returns a secret string
crypto.sign({"name": "HMAC", "hash": "SHA-256"}, key, data);
```
This is closer to the `js-crypto` polyfill library's API.

While we're there, I think we might as well support `SHA-384` and `SHA-512` for both this and existing signing in JS, since its supported in C++? But maybe I've missed a reason it was deliberately omitted.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated Open Enclave to [0.19.2](https://github.com/openenclave/openenclave/releases/tag/v0.19.2).
 - Updated Open Enclave to [0.19.3](https://github.com/openenclave/openenclave/releases/tag/v0.19.3).
 - Expose COSESign1 `content` for `user_cose_sign1` authenticated endpoints in JavaScript/TypeScript apps (#5465).
+- Add HMAC support to JS API. Call with `ccf.crypto.sign({"name": "HMAC", "hash": "SHA-256"}, key, data)`.
```
achamayou, line 13: We ought to update https://github.com/microsoft/CCF/blob/b44e90aa38c2d50b621effe617103e8996aa9cd0/js/ccf-app/src/global.ts#L217, or possible split the type (but probably just update, since this is using sign() too).

achamayou (approved):

CHANGELOG.md
achamayou: Also good to check the polyfill and update the corresponding test if possible.

PR #5510, opened 2023-08-07
Get a shared boot time for offsetting submit times

eddyashton:
#5503 attempted to convert `CLOCK_MONOTONIC` times to wall-clock time, so that they can be compared with timestamps on other machines. However, since this was done in each client, it could produce visible skew between times from clients running on the same machine.

This PR attempts to calculate the system boot time once, and pass that to each client, so they get should produce times which are at least consistent with other times measured on the same machine, and passed the same boot time.

### DIFF
--- a/tests/infra/basicperf.py
+++ b/tests/infra/basicperf.py
@@ -16,6 +16,7 @@ from typing import Dict, List
 import random
 import string
 import json
+import datetime
 
 
 def configure_remote_client(args, client_id, client_host, common_dir):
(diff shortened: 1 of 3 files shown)


tests/infra/basicperf.py
```diff
@@ -258,6 +268,8 @@ def run(args):
                     os.path.abspath(path_to_requests_file),
                     "--pid-file-path",
                     "cmd.pid",
+                    "--boot-time",
+                    str(boot_time_us),
```
achamayou, line 272: I'm not sure why we pass this in rather than offset the agg dataframe in one go, but this is fine too.

achamayou (approved):

eddyashton:
New plan: go back to `CLOCK_REALTIME` (`MONOTONIC` is still ntp-adjusted, `MONOTONIC_RAW` would not be, but then correlating with other wall-clocks relies on this bodge to get an approximation).

PR #5516, opened 2023-08-08
Enable CodeQL ADO task in specific builds only

eddyashton:


eddyashton:
CodeQL3000 explicitly won't run on PRs, which means we're not going to get any benefit vs the existing GHA solution - closing this.

PR #5518, opened 2023-08-08
Quicker node start post shutdown

achamayou:


achamayou:
Cleaned up and PR'ed as #5527

PR #5520, opened 2023-08-09
Ruff lint update: Use `isinstance`

eddyashton:
New version of `ruff` flags this, so format checks are currently failing.

### DIFF
--- a/tests/infra/piccolo/generator.py
+++ b/tests/infra/piccolo/generator.py
@@ -38,7 +38,7 @@ class Messages:
             headers["content-type"] = content_type
 
         # Convert body to bytes if we were given a string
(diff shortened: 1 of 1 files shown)


jumaffre (approved):

tests/infra/piccolo/generator.py
```diff
@@ -38,7 +38,7 @@ def append(
             headers["content-type"] = content_type
 
         # Convert body to bytes if we were given a string
-        if type(body) == str:
+        if isinstance(body, str):
```
achamayou, line 41: I think we could just demand bytes also

achamayou (approved):

PR #5529, opened 2023-08-10
Install vegeta and pebble and run corresponding tests in Daily

achamayou:
This removes the need for vegeta and pebble to be installed in the CI image, and they will be removed in a separate PR + tag.

Vegeta and acme tests are now only run in the Daily as a result of this change, to avoid the extra delay and minimise impact due to GitHub unavailability (we've had mixed results with pulling GH release artefacts from jobs in the past).


.azure-pipelines-templates/install_extended_testing_tools.yml
```diff
@@ -0,0 +1,8 @@
+steps:
+  - script: |
+      set -ex
+      sudo apt-get -y update
+      sudo apt install ansible -y
+      cd getting_started/setup_vm
+      ansible-playbook ccf-extended-testing.yml
+    displayName: Install Extended Testing Tools
```
jumaffre, line 8:
```suggestion
      ./run.sh ccf-extended-testing.yml
    workingDirectory: getting_started/setup_vm
    displayName: Install Extended Testing Tools
```

Slightly neater I think.

jumaffre (approved):

PR #5481, opened 2023-07-28
OpenSSL3: upgrade `EVP_PKEY` functions

jumaffre:
Part of https://github.com/microsoft/CCF/issues/5291

PR #5548, opened 2023-08-15
Faster SHA256 OpenSSL hashing

jumaffre:
Part of #5291 

Standalone change carved out of https://github.com/microsoft/CCF/pull/5481

As the `EVP_Digest` function family is slower with OpenSSL 3.x (see https://github.com/openssl/openssl/issues/19612), we now keep track of 2 static thread-local `EVP_MD_CTX` that are re-used between calls to `openssl_sha256`, which is very much on the critical path (about 7.5% performance drop on `ls_virtual_cft` with OpenSSL 3.1.1). 

The performance impact of `EVP_Digest` is particularly large on small buffers, as per our `crypto_bench`:

Before:
```log
digest sha256:
    openssl_sha256_base * |      10 |     0.004 |     419 |      - |  2386634.8  << 128 byte input
         openssl_sha256_8 |      10 |     0.007 |     686 |  1.638 |  1457301.1
        openssl_sha256_12 |      10 |     0.064 |    6415 | 15.311 |   155874.9
        openssl_sha256_16 |      10 |     1.071 |  107092 |255.590 |     9337.8
        openssl_sha256_18 |      10 |     3.925 |  392532 |936.832 |     2547.6
```

Now:
```log
digest sha256:
    openssl_sha256_base * |      10 |     0.002 |     231 |      - |  4327131.1 << 128 byte input
         openssl_sha256_8 |      10 |     0.005 |     503 |  2.180 |  1984914.6
        openssl_sha256_12 |      10 |     0.063 |    6268 | 27.126 |   159520.2
        openssl_sha256_16 |      10 |     0.981 |   98111 |424.542 |    10192.5
        openssl_sha256_18 |      10 |     3.924 |  392384 |1697.900 |     2548.5
```


src/crypto/openssl/hash.cpp
```diff
@@ -37,13 +37,73 @@ namespace crypto
...
+
+    mdctx = EVP_MD_CTX_new();
+    if (mdctx == nullptr)
+    {
+      throw std::logic_error("openssl_sha256_init: failed to create mdctx");
+    }
+    basectx = EVP_MD_CTX_new();
+    if (basectx == nullptr)
+    {
+      throw std::logic_error("openssl_sha256_init: failed to create basectx");
```
achamayou, line 58: Should this also free the mdctx? Do the always go through shutdown if we throw here?

src/crypto/openssl/hash.cpp
```diff
@@ -37,13 +37,73 @@ namespace crypto
...
+    }
+    if (EVP_DigestInit_ex(basectx, EVP_sha256(), nullptr) != 1)
+    {
+      throw std::logic_error("EVP_DigestInit_ex failed");
```
achamayou, line 62: Same question as above.

achamayou: It seems cleaner to destroy everything if we throw, and reset to init state, than to leave a partially constructed situation with some heap allocations.

Issue #5478, opened 2023-07-27
Reuse JS interpreter


eddyashton:
JS apps running in CCF currently execute in an extremely strict sandbox - a fresh JS interpreter and context (ie - all global state) is recreated for every endpoint invocation. This guarantees that modifications to globals from one request cannot affect another, in turn ensuring that the result of a request is deterministically reproducible from the ledger. If we persist any global JS (ie node-local!) state, and that affects future request execution, then the result of a request is no longer determined purely from the ledger, but also from the history of requests on this node.

This has implications for app performance - any static initialisation which would happen once and create some global object in a "normal" JS application will instead be re-executed on every request. As a specific example, an app may uses Inversify Dependency Injection to construct many narrow, targeted services. The expectation is that the container lives at a "composition root", so that resolution of (at least) singleton objects can be cached. In CCF, we instead do that resolution (including potentially costly re-construction of all singletons) for every request.

We can weaken that sandboxing and reuse interpreters, including global state, to provide better performance in these scenarios.

We're currently only considering interpreter reuse for application endpoints, governance will retain the current model. We'd like to make this opt-in, and retain some sandboxing protections (eg, restrict interpreter reuse to the same endpoint, or caller, or app identifier, or...) to restrict the blast radius of unsafe global reuse.

eddyashton:
I've confirmed that it is possible to reuse the QuickJS state, with a quick hack making the `js::Runtime` and `js::Context` static. This gives the expected performance win of allowing static-style global initialisation to happen once. Testing with a dumb app with the following pattern (and some Inversify `container.get`, to confirm it is stashing similarly):

```
  if (globalThis.initialised !== true) {
    console.log("Doing first-time initialisation");
    console.log(`  fibonacci(32) = ${fibonacci(32)}`);
    globalThis.initialised = true;
    console.log("Done first-time initialisation");
  }
  else
  {
    console.log("Already initialised");
  }
```

We go from a state where every call is equally slow:
```
2023-07-27T10:35:57.827046Z -0.002 0   [info ][app] ../src/js/wrap.cpp:93           | Doing first-time initialisation
2023-07-27T10:35:58.031158Z -0.206 0   [info ][app] ../src/js/wrap.cpp:93           |   fibonacci(32) = 3524578
2023-07-27T10:35:58.031174Z -0.206 0   [info ][app] ../src/js/wrap.cpp:93           | Done first-time initialisation
...
[...]

PR #5564, opened 2023-08-17
Allow JS interpreters (and global state) to be reused

eddyashton:
Discussed in #5478.

JS applications may now opt-in to reusing interpreters, at a per-endpoint granularity, with the following entry in `app.json`:

```json
"interpreter_reuse": {"key": "foo"}
```

This is an object to give us some flexibility in future, eg adding per-user, per-session, or per-thread constraints.

This PR contains some refactoring of interpreter (`js::Context`) construction and initialisation, mostly just to ease this modification, but they also simplify the creation code. Notably we don't need to create a `js::Runtime` separately, it is instead created (and owned) by a `js::Context`. The interpreters are stored in an LRU cache, so we can cap the number of interpreters retained, to limit memory use. I've opted to allow sharing of interpreters across threads, since I think that matches the current use requirements, though this involves taking a lock on each interpreter for the duration of JS execution which will be a bottleneck in other scenarios.


eddyashton:
Resolves #5478.

doc/schemas/gov_openapi.json
```diff
@@ -1605,6 +1626,28 @@
         }
       }
     },
+    "/gov/kv/interpreter/flush": {
```
achamayou, line 1629: Does this flush the interpreter states? If so, `get` doesn't seem like the most obvious choice.

eddyashton: No - this is an auto-generated endpoint returning the current value from the governance table.

src/apps/js_generic/js_generic_base.cpp
```diff
@@ -253,57 +253,86 @@ namespace ccfapp
...
+      }
+      js::Context& ctx = *interpreter;
+
+      // Prevent any other thread modifying this interpreter, until this
+      // function completes. We could create interpreters per-thread, but then
+      // we would get no cross-thread caching benefit (and would need to either
+      // enforce, or share, caps across per-thread caches). We choose
+      // instead to allow interpreters to be maximally reused, even across
+      // threads, at the cost of locking (and potentially stalling another
+      // thread's request execution) here.
```
achamayou, line 313: That may be something to revisit if we add an explicit way to populate the interpreters (ie. a specific "setup_global_state()" call in the application that CCF would call), in which case keeping one per thread is likely a good memory vs throughput tradeoff.

src/js/interpreter_cache.h
```diff
@@ -0,0 +1,83 @@
...
+  class InterpreterCache : public AbstractInterpreterCache
+  {
+  protected:
+    // Locks access to all internal fields
+    ccf::pal::Mutex lock;
+    LRU<std::string, std::shared_ptr<js::Context>> lru;
+    size_t cache_build_marker;
+
+  public:
+    InterpreterCache() : lru(10) {}
```
achamayou, line 20: Should we pass through the max_cached_interpreters at construction time?

src/js/interpreter_cache.h
```diff
@@ -0,0 +1,83 @@
...
+
+    std::shared_ptr<js::Context> get_interpreter(
+      js::TxAccess access,
+      const JSDynamicEndpoint& endpoint,
+      size_t freshness_marker) override
+    {
+      if (access != js::TxAccess::APP)
+      {
+        throw std::logic_error(
+          "JS interpreter reuse lru is currently only supported for APP "
```
achamayou, line 30:
```suggestion
          "JS interpreter reuse lru is only supported for APP "
```

I don't believe this would be a good tradeoff for governance, let's not imply we intend to make it available there until we have decided so.

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -657,6 +658,170 @@ def run_api(args):
         network = test_metrics_logging(network, args)
 
 
+def test_reused_interpreter_behaviour(network, args):
+    primary, _ = network.find_nodes()
```
achamayou, line 662: Turning on the caching for the full_js test may be a good way to check that this is working as intended too. Something else we could do is add to the stats with a number of runtime instantiations per endpoint, to be compared to calls.

achamayou:
@eddyashton could you also poke the threading canary?

achamayou (approved):

src/apps/js_generic/js_generic_base.cpp
eddyashton: I agree that we should revisit this in future, especially when memory use is no longer such a constraint. I think this is a sensible behaviour for now, and hopefully the comment reminds us to revisit.

tests/js-custom-authorization/custom_authorization.py
eddyashton: Agreed and done - this is actually a _huge_ perf speedup, so I'm trying the same on the non-full app to compare. If its a speedup there, we may drop perf-testing `full_js` in favour of testing the small app in `cached` and `uncached` modes.

include/ccf/service/tables/modules.h
```diff
@@ -24,5 +25,7 @@ namespace ccf
       "public:ccf.gov.modules_quickjs_bytecode";
     static constexpr auto MODULES_QUICKJS_VERSION =
       "public:ccf.gov.modules_quickjs_version";
+    static constexpr auto INTERPRETER_FLUSH =
+      "public:ccf.gov.interpreter.flush";
```
jumaffre, line 29: This new table should be documented [here](https://microsoft.github.io/CCF/main/audit/builtin_maps.html).

src/js/interpreter_cache.h
```diff
@@ -0,0 +1,83 @@
...
+            else
+            {
+              LOG_TRACE_FMT(
+                "Returning interpreter previously in cache, with key {}", key);
+              lru.promote(it);
+            }
+
+            return it->second;
+          }
+        }
```
jumaffre, line 69: Do we need a `default:` here or is that for forward compatibility?

include/ccf/service/tables/modules.h
achamayou: Good point, this wasn't caught by the automation because it's run at the end of the full suite, which doesn't run JS code right now.

src/js/interpreter_cache.h
achamayou: kind is already an enum at this point, the validation will have taken place in the JSON deserialisation. So this is not forward compatible, but we also don't particularly need a default.

src/js/wrap.cpp
```diff
@@ -2360,6 +2350,30 @@ namespace ccf::js
     }
   }
 
+  void Runtime::set_runtime_options(kv::Tx* tx)
+  {
+    LOG_TRACE_FMT("Setting run time options");
```
jumaffre, line 2355:
```suggestion
```

PR #5575, opened 2023-08-21
JWT auth: cache verifiers

### LINKED ISSUE #5557: Cache JWT verifiers
In the JWT auth layer, we generate a new `crypto::Verifier` for every incoming JWT request. However, as there is a relatively limited number of JWT issuers, we could cache the `crypto::Verifier`s so that they don't need to be re-constructed for every incoming request.

This is especially important given that OpenSSL 3.x has poorer performance when loading certificates (see https://github.com/openssl/openssl/issues/19119.).

jumaffre:
Resolves #5557 

To avoid rebuilding a new `crypto::Verifier` for each incoming JWT request (which is particularly expensive with OpenSSL 3.x - see #5557), we now cache a relatively small number of verifiers in the JWT auth layer. This results in a performance increase for all JWT-related tests, which will be even greater with OpenSSL 3.x.


achamayou (approved):

achamayou:
It would be good to have the cache size be configurable through governance, in the same way as the other JWT settings, some users may want to cache more than 10 verifiers etc.

PR #5598, opened 2023-08-30
Miscellaneous operations doc fixes

jumaffre:


doc/operations/ledger_snapshot.rst
```diff
@@ -70,7 +70,7 @@ Join or Recover From Snapshot
...
 
-It is important to note that new nodes cannot resume from a snapshot and join a service via a node that started from a more recent snapshot. For example, if a new node resumes from a snapshot generated at ``seqno 100`` and joins from a (primary) node that originally resumed from a snapshot at ``seqno 50``, the new node will throw a ``StartupSeqnoIsOld`` error shortly after starting up. It is expected that operators copy the *latest* committed snapshot file to new nodes before start up.
+It is important to note that new nodes cannot resume from a snapshot and join a service via a node that started from a more recent snapshot. For example, if a new node resumes from a snapshot generated at ``seqno 50`` and joins from a (primary) node that originally resumed from a snapshot at ``seqno 100``, the new node will throw a ``StartupSeqnoIsOld`` error shortly after starting up. It is expected that operators copy the *latest* committed snapshot file to new nodes before start up.
```
jumaffre, line 73: I believe this was the wrong way around all along.

eddyashton:
```suggestion
It is important to note that new nodes cannot join a service if the snapshot they start from is more recent than the snapshot of the node they join via. For example, if a new node resumes from a snapshot generated at ``seqno 50`` and joins from a (primary) node that originally resumed from a snapshot at ``seqno 100``, the new node will throw a ``StartupSeqnoIsOld`` error shortly after starting up. It is expected that operators copy the *latest* committed snapshot file to new nodes before start up.
```

jumaffre: I think "than the snapshot of the node" is ambiguous (it may mean the latest snapshot generated by that node?). 

What about "It is important to note that new nodes cannot join a service if the snapshot they start from is older than the snapshot the primary node started from."?

eddyashton: Oh, is it the primary? Or the joinee?

eddyashton: Sure, let's say primary rather than "node they join via" then, and your wording works.

eddyashton (approved):

PR #5600, opened 2023-08-31
Use keygenerator from the bindir

jeffa5:
The `key_generator.sh` script is in the `binary_dir`, and this is built as a path in the `key_generator` variable already so use that instead of guessing it is in the current directory.

The error occured from, after building CCF, running `tests/sandbox/sandbox.sh` from the root CCF directory. This would also happen when run from other places as it assumes that the `keygenerator.sh` script is located in the current directory.


jumaffre (approved):
Thank you. Could you quickly mention in the PR description how the error that this fixes could occur, in case another user runs into this?

jeffa5:
@jumaffre done!

PR #5601, opened 2023-08-31
Add optimisation to range for sorted map

jeffa5:
If the underlying map is sorted then we don't need to build a sorted intermediate storage while iterating through, we can apply the function directly.

### DIFF
--- a/src/kv/untyped_map_handle.cpp
+++ b/src/kv/untyped_map_handle.cpp
@@ -228,7 +228,7 @@ namespace kv::untyped
 #endif
 
     std::map<KeyType, ValueType> res;
-    auto g = [&res, &from, &to, continue_past_range_to](
+    auto g = [&res, &from, &to, &f, continue_past_range_to](
                const KeyType& k, const ValueType& v) {
       if (from.has_value() && k < from.value())
       {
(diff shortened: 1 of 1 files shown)


eddyashton:
I think this is incorrect, though haven't tested locally.

The lambda modified here is passed into `foreach_state_and_writes`, which (since it assumed unordered containers) will first execute for all (non-overwritten) _state_ entries (ie - from previous transactions), and then all _write_ entries (ie - from current transaction). This means the range iteration across the combination of those is not necessarily ordered! I think a test case like the following would fail:

```
// Extremely pseudocode
{
  tx1 = create_tx();
  tx1.map.write(2, "");
  tx1.map.write(3, "");
  tx1.commit();
}

{
  tx2 = create_tx();
  tx2.map.write(1, "");
  tx2.map.write(4, "");

  all_keys = [];
  tx2.map.range(0, 5, (k, v) => { all_keys.append(k) });
  ASSERT(sorted(all_keys)); // Fails, sees [2,3,1,4] rather than [1,2,3,4].
}
```

jeffa5:
Hmm, agreed. It would only be valid if the `writes` map is empty. I'll close this for now unless I it becomes problematic enough to think about special casing it when there are no transaction writes (e.g. a read only endpoint).

eddyashton:
> Hmm, agreed. It would only be valid if the `writes` map is empty. I'll close this for now unless I it becomes problematic enough to think about special casing it when there are no transaction writes (e.g. a read only endpoint).

I think a branch on `const bool exec_inline = KV_STATE_RB && writes.empty()` would work, but think we should only do that with a test like the above, and a perf test/use case showing a useful impact.

Issue #5474, opened 2023-07-20
3rdparty dependency updates


achamayou:
- [x] QCBOR 1.2
- [x] fmtlib 10
- [x] llhttp 8.1.1
- [x] nghttp2 1.55.1
- [x] snmalloc 0.6.2
- [x] cli11 2.3.2

achamayou:
Protobuf is also behind, and seems to require abseil which we would prefer not to include.

PR #5605, opened 2023-09-01
Update `fmt` library from `9.1.0` to `10.1.1`

jumaffre:
Part of #5474


src/clients/perf/scenario_perf_client.cpp
```diff
@@ -60,8 +60,8 @@ class ScenarioPerfClient : public Base
         "Sending {} {} transactions", transactions.size(), element_name);
       for (const auto& transaction : transactions)
       {
-        const auto method = transaction["method"];
-        const auto params = transaction["params"];
+        const std::string& method = transaction["method"];
```
eddyashton, line 63: I'm not sure why this is `const&` assigned? We need to do a conversion, since the righthand side is `nlohmann::json`.

So safer to do:
```suggestion
        const std::string method = transaction["method"];
```

Or:
```suggestion
        const auto method = transaction["method"].get<std::string>();
```

Same for the similar changes elsewhere.

eddyashton (approved):

PR #5607, opened 2023-09-01
Update `llhttp` from `6.0.9` to `9.0.1`

jumaffre:
Part of #5474


include/ccf/http_status.h
```diff
@@ -2,86 +2,12 @@
...
-  XX(502, BAD_GATEWAY, "Bad Gateway") \
-  XX(503, SERVICE_UNAVAILABLE, "Service Unavailable") \
-  XX(504, GATEWAY_TIMEOUT, "Gateway Timeout") \
-  XX(505, HTTP_VERSION_NOT_SUPPORTED, "HTTP Version Not Supported") \
-  XX(506, VARIANT_ALSO_NEGOTIATES, "Variant Also Negotiates") \
-  XX(507, INSUFFICIENT_STORAGE, "Insufficient Storage") \
-  XX(508, LOOP_DETECTED, "Loop Detected") \
-  XX(510, NOT_EXTENDED, "Not Extended") \
-  XX(511, NETWORK_AUTHENTICATION_REQUIRED, "Network Authentication Required")
+#include <llhttp/llhttp.h>
```
jumaffre, line 5: @eddyashton @achamayou 

Note that now that llhttp defines its own `HTTP_STATUS_MAP(XX)`, we will return the HTTP status string in the HTTP response as all caps vs. capitalised before (e.g. `"SWITCHING PROTOCOLS"` vs. `"Switching Protocols"`). 

As per https://www.ietf.org/rfc/rfc2616.txt, this shouldn't break compatibility with clients:

>    The individual values of the numeric status codes defined for
   HTTP/1.1, and an example set of corresponding Reason-Phrase's, are
   presented below. The reason phrases listed here are only
   recommendations -- they MAY be replaced by local equivalents without
   affecting the protocol.

eddyashton (approved):

PR #5613, opened 2023-09-04
Adding note to TLA+ spec on message ordering assumptions

heidihoward:


tla/ccfraft.tla
```diff
@@ -153,6 +153,8 @@ ReconfigurationVarsTypeInv ==
 \* to another. With CCF, we have message integrity and can ensure unique messages.
 \* Messages only records messages that are currently in-flight, actions should
 \* removed messages once received.
+\* We model messages as a single (unsorted) set and do not assume ordered message delivery between nodes.
+\* Node-to-node channels use TCP but out-of-order delivery could be observed due to reconnection.
```
achamayou, line 157: And also because the host is potentially malicious and could re-order frames between the network and the ring buffer.

achamayou (approved):

tla/ccfraft.tla
lemmy: I've started experimenting with liveness, and it seems like an attacker could potentially indefinitely delay a message by moving it to the tail of the ring buffer. Is this indeed the situation, or can we safely assume that every message will eventually get delivered?

achamayou: Yes, an attacker could do that, or drop all messages too, since they control the host.

lemmy: Conceptually, isn't dropping all messages similar to crashing the host (CFT), whereas dropping (omitting) some messages puts us in the realm of OFT?  For example, an attacker could deliberately drop messages that prevent the overall system from making forward progress, while allowing heartbeats to come through.

heidihoward: Since the specification is currently only concerned with safety, modelling messages in this way makes sense. If we want to start adding liveness properties, we'll need to support another stricter messaging model

Issue #5341, opened 2023-06-10
matchIndex lower bound for nextIndex


lemmy:
Why would a leader ever send entries that it expects the follower to already have, i.e., -more formally- is `matchIndex` a lower bound for `nextIndex` (see https://github.com/microsoft/CCF/issues/5340 for non-leaders)?

```tla
MatchIndexLowerBoundNextIndexInv ==
    \A i,j \in Servers :
        state[i] = Leader =>
            nextIndex[i][j] >= matchIndex[i][j]
```

This property is violated by `ccfraft.tla` if AppendEntries response arrive out-of-order s.t. a NACK arrives (at the leader) after a subsequent ACK.

NACK in state 8 does not arrive at the leader.

```tla
Invariant NextIdxMatchIdxInv is violated.
The behavior up to this point is:
1: <Initial predicate>
/\ nextIndex = (n1 :> (n1 :> 1 @@ n2 :> 2) @@ n2 :> (n1 :> 2 @@ n2 :> 2))
/\ currentTerm = (n1 :> 2 @@ n2 :> 2)
/\ matchIndex = (n1 :> (n1 :> 0 @@ n2 :> 0) @@ n2 :> (n1 :> 0 @@ n2 :> 0))
/\ commitIndex = (n1 :> 0 @@ n2 :> 0)
/\ state = (n1 :> Follower @@ n2 :> Leader)
/\ messages = {}
/\ log = (n1 :> <<[term |-> 1, contentType |-> Entry, request |-> 1]>> @@ n2 :> <<>>)
2: <ClientRequest line 682, col 5 to line 695, col 48 of module ccfraft>
/\ messages = {}
/\ currentTerm = (n1 :> 2 @@ n2 :> 2)
/\ state = (n1 :> Follower @@ n2 :> Leader)
/\ log = (n1 :> <<[term |-> 1, contentType |-> Entry, request |-> 1]>> @@ n2 :> <<[term |-> 2, contentType |-> Entry, request |-> 2]>>)
/\ commitIndex = (n1 :> 0 @@ n2 :> 0)
/\ nextIndex = (n1 :> (n1 :> 1 @@ n2 :> 2) @@ n2 :> (n1 :> 2 @@ n2 :> 2))
/\ matchIndex = (n1 :> (n1 :> 0 @@ n2 :> 0) @@ n2 :> (n1 :> 0 @@ n2 :> 0))
3: <AppendEntries line 623, col 5 to line 656, col 105 of module ccfraft>
[...]

PR #5614, opened 2023-09-05
matchIndex lower bound for nextIndex

lemmy:
https://github.com/microsoft/CCF/issues/5341


tla/MCccfraft.cfg
```diff
@@ -67,6 +67,7 @@ INVARIANTS
     MonoLogInv
     LogConfigurationConsistentInv
     NoLeaderInTermZeroInv
+    MatchIndexLowerBoundNextIndexInv
```
heidihoward, line 70: Aside: if there any easy way to deduplicate these cfg files?

lemmy: Unfortunately not at the moment.

heidihoward (approved):

lemmy:
Merged because the build failures were unrelated.

Issue #5064, opened 2023-02-27
GET /nodes/self returns 500 when the node is Pending


achamayou:
This is because the KV doesn't contain the node info yet, but it's entirely known from the config and could be returned.

lynshi:
Providing some context from the discussion.

We made a change to use the endpoint for fetching the node id (to be able to trust the node) thinking it was nicer than querying `/node/network/nodes` and matching by IP information. After some discussion with @achamayou, we concluded that it's better to fetch the info from the service to prevent a malicious node from reporting an incorrect node id (e.g. of another node that is pending). Of course, this is a difficult attack to pull off - all nodes involved must still be running attested code - but we'll switch back to using the service's information for more obvious safety.

We don't have a specific requested change, so either of the following could work:
1. Fix the behavior so info is returned when the node is pending, if this does not seem like an issue for safe, "correct" operation.
2. Return a more illuminating error message and/or add documentation for this issue so that customers aren't confused when hitting an endpoint that should reasonably be available.

achamayou:
@lynshi the node info returned is entirely supplied by the node from its configuration. It could be served by the node pre-join without creating any safety or security issues. It's probably the right thing to do here.

lynshi:
Right - I meant it was unsafe in the specific use-case we were discussing, which is getting the node id so that a proposal to trust the node can be submitted. Perhaps a reasonable thing to do is to leave it empty until trusted? Agree that the rest of the info is a non-issue.

PR #5619, opened 2023-09-05
Remove redirection usage APIs expected to be used through a load balancer

### LINKED ISSUE #5184: GET /node/network/nodes/self redirects may be inaccessible from the public Internet
**Describe the bug**
We've previously observed that `GET /node/network/nodes/self` redirects the client to an address containing the node's id. For example, `https://$nodeIp:$nodePort/node/network/nodes/$nodeId`.

If the node is running inside a private network, redirects to the private IP won't work for public connections. I'm not sure if `$nodeIp:$nodePort` comes from the `published_address` parameter in the node's configuration, but even if it does, the parameter may be meaningless publicly for networks that don't expose nodes individually.

We previously agreed the redirect itself is fine, though I didn't try out connections from public clients because that wasn't what I was testing. However, now that I've found this issue, there's probably two things that can be done:
1. Redirect to a URI containing the node's id, but the hostname should be the public host. Assuming `published_address` is intended to be different per node, this might require an additional configuration parameter like `service_address`. Alternatively, you can get the hostname used by the client (e.g. in the `Origin` header..?), and just use that for the redirect because it guarantees accessibility by the client.
2. Don't redirect at all and return the info directly.

**Environment information**
CCF 2.0.10 (though I think the behavior has not changed since)

achamayou:
Resolve #5184 and #5064

`GET /network/nodes/self` now no longer redirects, since clients calling it through the load balancer will not be able to resolve the individual node. When it doesn't (yet) have access to the KV, the endpoint now uses its configuration to respond instead. A consequence of that is that we now import into the enclave the rpc configuration (something we did not do previously).

`GET /node/network/nodes/primary` similarly no longer redirects, and does a KV lookup instead. I had also made it always forward to the primary at first, to avoid being unable to respond usefully in the (admittedly uncommon) situation where the node does not have the primary in its KV yet. But that breaks usage over HTTP2, so I have reverted it.

I am leaving the redirect in place for `HEAD /node/primary`, because the endpoint is targeted at operators/load-balancers themselves, not users accessing the service behind a load-balancer. There also isn't an obvious choice of HTTP return code if it doesn't redirect.

@lynshi can you confirm that this is acceptable.


tests/e2e_logging.py
```diff
@@ -1853,24 +1853,24 @@ def run_parsing_errors(args):
...
+
+    # cr.add(
+    #     "cpp",
+    #     run,
+    #     package="samples/apps/logging/liblogging",
+    #     js_app_bundle=None,
+    #     nodes=infra.e2e_args.max_nodes(cr.args, f=0),
+    #     initial_user_count=4,
+    #     initial_member_count=2,
+    # )
```
eddyashton, line 1873: Restore before merging.

tests/e2e_logging.py
```diff
@@ -1879,27 +1879,27 @@ def run_parsing_errors(args):
...
+    #     nodes=infra.e2e_args.max_nodes(cr.args, f=0),
+    # )
+
+    # # This is just for the UDP echo test for now
+    # cr.add(
+    #     "udp",
+    #     run_udp_tests,
+    #     package="samples/apps/logging/liblogging",
+    #     nodes=infra.e2e_args.max_nodes(cr.args, f=0),
+    # )
```
eddyashton, line 1903: Restore before merging.

src/node/rpc/node_frontend.h
```diff
@@ -1117,94 +1117,110 @@ namespace ccf
...
+            "No configured consensus");
+        }
       };
       make_read_only_endpoint(
-        "/network/nodes/primary", HTTP_GET, get_primary_node, no_auth_required)
-        .set_forwarding_required(endpoints::ForwardingRequired::Never)
+        "/network/nodes/primary",
+        HTTP_GET,
+        json_read_only_adapter(get_primary_node),
+        no_auth_required)
```
eddyashton, line 1223:
```suggestion
        no_auth_required)
        .set_auto_schema<void, GetNode::Out>()
```

eddyashton (approved):

lynshi:
Sounds good to me, thanks!

PR #5630, opened 2023-09-06
matchIndex lower bound for nextIndex

lemmy:
https://github.com/microsoft/CCF/issues/5341

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1436,7 +1436,8 @@ namespace aft
           from);
         const auto this_match =
           find_highest_possible_match({r.term, r.last_log_idx});
-        node->second.sent_idx = std::min(this_match, node->second.sent_idx);
+        node->second.sent_idx = std::max(
+          std::min(this_match, node->second.sent_idx), node->second.match_idx);
         return;
       }
       else
--- a/tla/ccfraft.tla
+++ b/tla/ccfraft.tla
@@ -954,7 +954,7 @@ HandleAppendEntriesResponse(i, j, m) ==
           /\ nextIndex'  = [nextIndex  EXCEPT ![i][j] = max(@, m.lastLogIndex + 1)]
        \/ /\ \lnot m.success \* not successful
           /\ LET tm == FindHighestPossibleMatch(log[i], m.lastLogIndex, m.term)
-             IN nextIndex' = [nextIndex EXCEPT ![i][j] = max(tm, matchIndex[i][j]) + 1 ]
+             IN nextIndex' = [nextIndex EXCEPT ![i][j] = max(min(tm, nextIndex[i][j]-1), matchIndex[i][j]) + 1 ]
           \* UNCHANGED matchIndex is implied by the following statement in figure 2, page 4 in the raft paper:
            \* "If AppendEntries fails because of log inconsistency: decrement nextIndex and retry"
           /\ UNCHANGED matchIndex


lemmy:
```C
        node->second.sent_idx = std::max(this_match, node->second.match_idx + 1);
```

`build/raft_tests` has four failures and the four raft_scenarios `4582.2`, `election_while_reconfiguration`, `fancy_election.1`, and `suffix_collision.2` core dump with `+1`.

eddyashton:
`suffix_collision.2` is failing because we process multiple NACKs in a row, and discard a helpful/correct `sent_idx` with a still-too-late estimate from the later NACK.

(slightly modified logging)

```
# [1] sends [0] 2 append entries
2023-09-07T09:28:49.027466Z        100 [debug][raft] ./src/consensus/aft/raft.h:907 | n[1] | Leader | Active | Send append entries from n[1] to n[0]: (2.8, 3.10] (12)
2023-09-07T09:28:49.027472Z        100 [debug][raft] ./src/consensus/aft/raft.h:907 | n[1] | Leader | Active | Send append entries from n[1] to n[0]: (3.10, 4.12] (12)

# [0] says "What is this 2.8? I can't append to that. I have a thing called 1.8"
2023-09-07T09:28:49.027517Z        100 [debug][raft] ./src/consensus/aft/raft.h:959 | n[0] | Follower | Active | Received append entries: 2.8 to 3.10 (from n[1] in term 4)
2023-09-07T09:28:49.027520Z        100 [debug][raft] /src/consensus/aft/raft.h:1005 | n[0] | Follower | Active | Previous term for 8 should be 1
2023-09-07T09:28:49.027524Z        100 [debug][raft] /src/consensus/aft/raft.h:1028 | n[0] | Follower | Active | Recv append entries to n[0] from n[1] but our log at 8 has the wrong previous term (ours: 1, theirs: 2)
2023-09-07T09:28:49.027529Z        100 [debug][raft] /src/consensus/aft/raft.h:1321 | n[0] | Follower | Active | Send append entries response from n[0] to n[1] for entry 1.8: 1

# [0] says "What is this 3.10? I can't append to that. I have a thing called 2.8"
2023-09-07T09:28:49.027537Z        100 [debug][raft] ./src/consensus/aft/raft.h:959 | n[0] | Follower | Active | Received append entries: 3.10 to 4.12 (from n[1] in term 4)
2023-09-07T09:28:49.027542Z        100 [debug][raft] /src/consensus/aft/raft.h:1005 | n[0] | Follower | Active | Previous term for 10 should be 0
2023-09-07T09:28:49.027548Z        100 [debug][raft] /src/consensus/aft/raft.h:1016 | n[0] | Follower | Active | Recv append entries to n[0] from n[1] but our log does not yet contain index 10
2023-09-07T09:28:49.027554Z        100 [debug][raft] /src/consensus/aft/raft.h:1321 | n[0] | Follower | Active | Send append entries response from n[0] to n[1] for entry 4.8: 1

# [1] reads the first NACK, and (correctly) works out that maybe they have 1.4 in common, and should send from there
2023-09-07T09:28:49.027602Z        100 [debug][raft] /src/consensus/aft/raft.h:1436 | n[1] | Leader | Active | Recv append entries response to n[1] from n[0]: failed
2023-09-07T09:28:49.027607Z        100 [debug][raft] /src/consensus/aft/raft.h:1445 | n[1] | Leader | Active | Processing NACK, with this_match=4 and match_idx=0, I've updated sent_idx to 4

# [2] reads the second NACK, and overwrites 1.4 with the (plausible but less precise) 2.8
2023-09-07T09:28:49.027614Z        100 [debug][raft] /src/consensus/aft/raft.h:1436 | n[1] | Leader | Active | Recv append entries response to n[1] from n[0]: failed
2023-09-07T09:28:49.027619Z        100 [debug][raft] /src/consensus/aft/raft.h:1445 | n[1] | Leader | Active | Processing NACK, with this_match=8 and match_idx=0, I've updated sent_idx to 8

# Repeat ad infinitum
2023-09-07T09:28:49.027672Z        100 [debug][raft] ./src/consensus/aft/raft.h:907 | n[1] | Leader | Active | Send append entries from n[1] to n[0]: (2.8, 3.10] (12)
2023-09-07T09:28:49.027678Z        100 [debug][raft] ./src/consensus/aft/raft.h:907 | n[1] | Leader | Active | Send append entries from n[1] to n[0]: (3.10, 4.12] (12)
```

Since we send optimistically-late AEs, we may receive later-than-precise NACKs, and shouldn't advance `sent_idx` for those.

eddyashton:
There are many Safe assignments we can make to `next_idx`, and we're trying to find one of those which is also best for Liveness.

As a thought experiment, consider a primary trying to catch up a follower within the same term, where each AE contains only a single entry.

```
P:  1  2  3  4  5
B:  1

# Possible messages
AE(2)  AE(3)  AE(4)  AE(5)

# If B receives any-but-the-first, it emits a corresponding NACK, so other possible messages are
ACK(2) NACK(3)  NACK(4)  NACK(5)

# Since P has a match_index of 1, and could receive any of these NACKs, it is _Safe_
# (ie - there is a valid path that led here, so maybe we can hallucinate our way there from anywhere)
# So we could assign any of these
sent_idx=2 sent_idx=3  sent_idx=4  sent_idx=5

# I see the min(sent_idx) behaviour is a Liveness choice amongst those.
# We'd prefer to keep the lower value, since that tends-towards a smaller set of messages (assuming non-infinite duplication)

# The observation that match_index is a lower-bound is true
# It gets us out of another liveness hole if we have infinite message duplication
# Imagine P processes ACK(2), and then ACK(3), and then ACK(4)
P: 1  2  3  4  5
B: 1  2  3

P.sent_idx[B] = 4

# However, NACK(3) is still a potential message from earlier!
# If it arrives now, and is processed by P (under the existing code), then we unnecessarily (Safely, but unLively) decrement sent_idx!
# Using match_idx as a lower-bound avoids decrementing it too-far.
# But to handle out-of-order NACKs, we should then use the previous sent_idx as an upper-bound.
```

We are able to test some liveness scenarios with the scenario driver, but we don't simulate message duplication or re-ordering there. So I don't think we can create a test that proves the benefit of this new lower-bound, but believe it is worth adding.

eddyashton:
This change uses both bounds, and passes the tests I've tried:

```
node->second.sent_idx = std::max(
          std::min(this_match, node->second.sent_idx), node->second.match_idx);
```

I also understand some of the confusion around the `+1`:
I'm sure we don't it. We're storing _sent_idx_, where the Raft paper and spec talks about _nextIndex_. These obviously differ by `+1` ("what is the last thing I think I sent you" is exactly 1 less than "what is the next thing I should send you"), and we find _next_index_ when constructing AEs:

```
  send_append_entries(it.first, it.second.sent_idx + 1);
```

Our spec also uses `nextIndex`.

So `sentIndex`, if we represented that in the spec
- _could_ be 0 (when we've not sent them anything yet), and
- _could_ equal `matchIndex` (in lots of stable states it does - you're up-to-date and match at the last thing I sent you).

That differs from `nextIndex`, which
- must be positive (earliest thing I can send you is entry 1, 0 doesn't exist), and
- is strictly greater than `matchIndex` (I'll never resend you something I know you know).

Gluing that all together, I think that the spec change to match my proposed implementation assignment above, would be:
```
nextIndex' = [nextIndex EXCEPT ![i][j] = max(min(tm, nextIndex[i][j]), matchIndex[i][j]) + 1 ]
```

achamayou:
> We are able to test some liveness scenarios with the scenario driver, but we don't simulate message duplication or re-ordering there. So I don't think we can create a test that proves the benefit of this new lower-bound, but believe it is worth adding.

`shuffle_one` and `shuffle_all` do some re-ordering, but to trigger this particular case, you're right that we would need further extensions.

The slightly improved `nextIndex` calculation does seem worth adding, I agree.

achamayou (approved):

achamayou (approved):

PR #5650, opened 2023-09-12
Misc OpenSSL 3.x fixes

jumaffre:
This fixes minor errors following the OpenSSL 3.x, mostly on `release/4.x`. Namely:
- Re-enable TPCC perf test on `main`, and fix TPCC client.
- Re-enable and fix leak in `indexing_test` on `main`.


CMakeLists.txt
```diff
@@ -949,13 +949,13 @@ if(BUILD_TESTS)
...
-    # ${CMAKE_CURRENT_SOURCE_DIR}/src/indexing/test/lfs.cpp )
-    # target_link_libraries( indexing_test PRIVATE ccf_endpoints.host sss.host
-    # ccf_kv.host )
+    add_unit_test(
+      indexing_test ${CMAKE_CURRENT_SOURCE_DIR}/src/indexing/test/indexing.cpp
+      ${CMAKE_CURRENT_SOURCE_DIR}/src/indexing/test/lfs.cpp
+    )
+    target_link_libraries(
+      indexing_test PRIVATE ccf_endpoints.host sss.host ccf_kv.host
+    )
```
eddyashton, line 958: I think we should temporarily enable this unit test to confirm that the leak is fixed (ie - run it through SAN), but should not re-enable when this PR is merged - the test presumably remains flaky. Whether it should also be disabled on the release branches is a separate question, but it looks like it should if we don't have time to resolve the flakiness.

achamayou (approved):

PR #5653, opened 2023-09-13
Expose SNP Attestation validation in TS

DomAyre:
Todo:
- [x]  Test with non static Attestation report
- [x] Compile SNP separately from openenclave
- [x] Do more validation in modules_test


CMakeLists.txt
```diff
@@ -448,7 +448,10 @@ set(CCF_NETWORK_TEST_ARGS
 )
 
 if(COMPILE_TARGET STREQUAL "sgx")
-  add_enclave_library(js_openenclave.enclave ${CCF_DIR}/src/js/openenclave.cpp)
+  add_enclave_library(
+    js_openenclave.enclave ${CCF_DIR}/src/js/openenclave.cpp
```
achamayou, line 452: Best to either split as you propose, or rename this to js_attestation_verification.

js/ccf-app/src/global.ts
```diff
@@ -739,3 +739,19 @@ export interface OpenEnclave {
     endorsements?: ArrayBuffer,
   ): EvidenceClaims;
 }
+
+export interface SnpAttestationResult {
+  measurement: string;
```
achamayou, line 744: These should be exposed as ArrayBuffers, they're not necessarily legal JS strings, so they'd need to be base64'd or hex'ed, and the application code can do that if it wants.

tests/js-modules/modules.py
```diff
@@ -948,6 +947,20 @@ def test_npm_app(network, args):
...
+            LOG.info("SNP: Test verifySnpAttestation")
+            r = c.post(
+                "/app/verifySnpAttestation",
+                {
+                    "evidence": primary_quote_info["raw"],
+                    "endorsements": primary_quote_info["endorsements"],
+                    "endorsed_tcb": primary_quote_info["endorsed_tcb"],
+                },
+            )
+            assert r.status_code == http.HTTPStatus.OK, r.status_code
```
achamayou, line 960: It would be good to test this a bit more: evidence too big, too short, endorsements that don't match, endorsed_tcb that doesn't match etc.

js/ccf-app/src/global.ts
achamayou: Plus it'd be consistent with what we expose the claims with the OE version already.

CMakeLists.txt
jumaffre: I think splitting is better as verifying SGX quotes requires `open-enclave-hostverify` and we may want to build CCF with no dependency on OE whatsoever, but which can still verify SNP quotes.

tests/js-modules/modules.py
DomAyre: https://github.com/microsoft/CCF/pull/5653/commits/ff9aefa7538755b0d1fa8de035f490d39ab90523

src/apps/js_generic/js_generic.cpp
```diff
@@ -15,7 +16,7 @@ namespace ccfapp
   std::vector<ccf::js::FFIPlugin> get_js_plugins()
   {
 #ifdef SGX_ATTESTATION_VERIFICATION
-    return {ccf::js::openenclave_plugin};
+    return {ccf::js::openenclave_plugin, ccf::js::snp_attestation_plugin};
```
jumaffre, line 19: I think `ccf::js::snp_attestation_plugin` should be set even if SGX_ATTESTATION_VERIFICATION` is false.

tests/npm-app/src/endpoints/snp_attestation.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export interface TcbVersion {
+  boot_loader: number;
+  tee: number;
+  snp: number;
+  microcode: number;
+}
+
+interface SnpAttestationResult {
```
achamayou, line 25: why is this duplicated and not imported from ccf-app?

tests/js-modules/modules.py
```diff
@@ -948,6 +947,99 @@ def test_npm_app(network, args):
...
+            # Test corrupted endorsements
+            r = c.post(
+                "/app/verifySnpAttestation",
+                {
+                    "evidence": primary_quote_info["raw"],
+                    "endorsements": corrupt_value(primary_quote_info["endorsements"]),
+                    "endorsed_tcb": primary_quote_info["endorsed_tcb"],
+                },
+            )
+            assert r.status_code != http.HTTPStatus.OK, r.status_code
```
achamayou, line 1040: This is not good practice, if the code throws and returns a 500, this may pass for the wrong reason. We should check for the expected error code here (4xx probably?)

src/apps/js_generic/js_generic.cpp
achamayou: We don't need a gating flag for the snp attestation, since it does not drag in external deps.

src/js/snp_attestation.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+#include <algorithm>
+#include <quickjs/quickjs.h>
+#include <regex>
+#include <vector>
+
+namespace ccf::js
+{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wc99-extensions"
```
jumaffre, line 18: Why is this required?

achamayou (approved):

tests/npm-app/src/endpoints/snp_attestation.ts
DomAyre: In ccf-app, the values are array buffers, but here we convert them into hex strings so those types have changed, its the same pattern we use for the OE bindings

src/js/snp_attestation.cpp
DomAyre: https://github.com/microsoft/CCF/pull/5653/commits/b316adc7febba6ca788f11dc4b1771c6c3efde1f

tests/js-modules/modules.py
DomAyre: https://github.com/microsoft/CCF/pull/5653/commits/f5c4991f3d961435a114422a21a47d318c9a8be7

achamayou:
@DomAyre could you add an entry to the changelog please, just describing the feature and the name of the new API? I think I've worked out why the SNP job did not trigger automatically on this PR, it should now.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated `nghttp2` from `1.51.0` to `1.55.1`.
 - Converted SNP attestation UVM endorsements from integer to arbitrary string.
 - Updated Intel SGX PSW from 2.17 to 2.20 (#5616)
+- Added `/app/verifySnpAttestation` endpoint which verifies SNP Attestations. (#5653)
```
achamayou, line 18: That's in a test app, it's not the public API. This should instead document the introduction of ccf.SnpAttestation.verifySnpAttestation() as a public API for TypeScript apps.

src/js/snp_attestation.cpp
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    pal::PlatformAttestationMeasurement measurement = {};
+    pal::PlatformAttestationReportData report_data = {};
+    UVMEndorsements parsed_uvm_endorsements;
+
+    try
+    {
+      pal::verify_snp_attestation_report(quote_info, measurement, report_data);
+      parsed_uvm_endorsements =
```
achamayou, line 90: This mandates uvm endorsements in the current format, and does not allow the caller to pass a trusted platform measurement instead, or to validate it against another form of endorsement.

achamayou: Confirmed with Kapil that manual checks need to be supported, so this step needs to be made optional when no UVM endorsements are passed, and the caller will then check `measurement` separately.

DomAyre: https://github.com/microsoft/CCF/pull/5653/commits/b317000c0b47e8fbe2c61cb53ac3d11051a776b8

CHANGELOG.md
DomAyre: https://github.com/microsoft/CCF/pull/5653/commits/13bf121765a263ba2e1b4eede7eca5b3a8e4c4ac

PR #5655, opened 2023-09-14
Replace Secret Sharing implementation

### LINKED ISSUE #5642: Replace sss library
Placeholder

achamayou:
Resolves #5642. Secret sharing logic implemented by @fournet.

Secret sharing used for ledger recovery now relies on a much simpler implementation that requires no external dependencies. Note that while the code still accepts shares generated by the old code for now, it only generates shares with the new implementation. As a result, a DR attempt that would downgrade the code to a version that pre-dates this change, after having previously picked it up, would not succeed if a reshare had already taken place.


src/crypto/sharing.cpp
```diff
@@ -0,0 +1,168 @@
...
+  // This function is specific to prime=2^31-1.
+  // We assume the lower 31 bits are uniformly distributed,
+  // and retry if they are all set to get uniformity in F[prime].
+
+  element sample()
+  {
+    uint64_t res = prime;
+    while (res == prime)
+    {
+      res = crypto::create_entropy()->random64() & prime;
```
eddyashton, line 76: I think we'll get a big speedup by reusing this entropy source, rather than creating it fresh each time? And I'm more confident that multiple calls to `random64()` on the _same_ entropy will produce different values (internal state has been swizzled), than that adjacently-constructed entropies will differ on their first value (risk that they're constructed from the same source, but getting different internal copies of the state). That's maybe not clear, but what I mean is:

```
auto ent = crypto::create_entropy();
assert(ent->random64() != ent->random64());
// ^ Surely, definitely, this is the whole point of this rng

assert(crypto::create_entropy()->random64() != crypto::create_entropy()->random64());
 // Probably? Maybe? I'd hope so, but relies on the details of create_entropy(), maybe it needs time between calls!
```

src/crypto/sharing.h
```diff
@@ -0,0 +1,101 @@
...
+   * 0 < @p degree < number of shares.
+   * @param[out] raw_secret sampled secret value
+   * @param[out] shares shares of raw_secret
+   * @param degree degree of the polynomial used to generate shares
+   *
+   * Note that for a secret sampled with @p degree, degree + 1 shares
+   * are required to recover the secret.
+   */
+  void sample_secret_and_shares(
+    Share& raw_secret, const std::span<Share>& shares, size_t degree);
```
eddyashton, line 86: So `degree` is `k-1`, for the `k` of the previous API? If so I'd prefer to use `k` in the public API (the number of shares required for reconstruction), rather than `degree` (which feels like an internal detail).

src/crypto/sharing.h
```diff
@@ -0,0 +1,101 @@
...
+  /** Using @p shares, recover @p secret.
+   * @param[out] raw_secret recovered secret value
+   * @param[in] shares shares of raw_secret
+   * @param degree degree of the polynomial used to generate shares
+   *
+   * Note that for a secret sampled with @p degree, degree + 1 shares
+   * are required to recover the secret.
+   *
+   * @throws std::invalid_argument if the number of shares is insufficient,
+   * or if two shares have the same x coordinate.
```
eddyashton, line 97: Can we report error directly, rather than throwing?

src/crypto/sharing.h
```diff
@@ -0,0 +1,101 @@
...
+   */
+  void recover_secret(
+    Share& raw_secret, const std::span<Share const>& shares, size_t degree);
```
eddyashton, line 100: I don't think we need to pass in degree/k in the previous API. Is there a risk that we pass the wrong value here? If we have 5 members, we've produced some shares with k=2, then we pass a proposal to say "recovery_threshold=3" but _don't reshare yet_, then the original 2 shares should still be sufficient. I don't know if its possible to change recovery threshold without resharing, but if we can't guarantee that then we need to ensure we don't pass the "wrong" value here.

src/crypto/sharing.cpp
achamayou: I am reasonably sure that's fine with rdseed, which we do use for the in-enclave code. With that said, re-using the object seems like a sensible optimisation here.

src/crypto/sharing.h
achamayou: Degree is the degree of the polynominal, and for degree, you need degree + 1 shares to re-assemble the secret. We can obviously change this call to take threshold and compute degree inside, but since it's not a public API, it won't make too much difference to anyone I think.

src/crypto/sharing.h
achamayou: I am not sure what you mean by report directly? I'm sure we agree that we don't want knowledge of the rpc handling in this call. The caller catches and reports the exception if this happens.

src/crypto/sharing.h
achamayou: We do https://github.com/microsoft/CCF/blob/4d74c1d314675dbb9356489b670d03c32a7d0f57/src/node/secret_share.h#L74, you need it for the calculation.

If you pass the wrong value, and it's not so wrong that we error out because you obviously don't have enough shares, you will not get the right secret back.

> then we pass a proposal to say "recovery_threshold=3" but don't reshare yet, then the original 2 shares should still be sufficient.

That's done transactionally today, because it would be very unsafe otherwise. It's exposed as a single JS binding to make it extra hard to split :)

src/crypto/sharing.h
eddyashton: I meant return. Seems like this is a short enough stack that we should be able to return an error directly rather than needing to throw, and then we could distinguish the error cases insufficient shares vs same x coordinate. The former seems like a normal/expected behaviour, the latter a legitimate error (have we stored the same member's share twice? Have we somehow submitted the same share for 2 different members? Or did we mess up in the sharing?)

achamayou: They're both unexpected errors for our code, since we know (and check) the threshold before we attempt to reassemble. Happy to give them separate exception types if it helps, but realistically we can't do much more than log the error in either situation.

CMakeLists.txt
```diff
@@ -910,6 +910,13 @@ if(BUILD_TESTS)
     target_include_directories(crypto_test PRIVATE ${CCFCRYPTO_INC})
     target_link_libraries(crypto_test PRIVATE ccfcrypto.host)
 
+    add_unit_test(
+      sharing_test
+      ${CMAKE_CURRENT_SOURCE_DIR}/src/crypto/test/secret_sharing.cpp
+    )
+    target_include_directories(sharing_test PRIVATE ${CCFCRYPTO_INC})
+    target_link_libraries(sharing_test PRIVATE ccfcrypto.host)
```
jumaffre, line 918: Perhaps re-use the existing `secret_share.cpp` unit test?

achamayou: I think the new one is more exhaustive, and some of the existing checks are no longer applicable because the code no longer splits and instead creates a secret + shares in a single step. There's also no upper limit on the number of shares anymore.

I will add the checks for 0, they do seem potentially valuable (although the governance should prevent that really).

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 [5.0.0-dev3]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.0-dev3
 
 - Updated Intel SGX PSW from 2.17 to 2.20 (#5616)
+- Secret sharing used for ledger recovery now relies on a much simpler implementation that requires no external dependencies. Note that while the code still accepts shares generated by the old code for now, it only generates shares with the new implementation. As a result, a DR attempt that would downgrade the code to a version that pre-dates this change, after having previously picked it up, would not succeed if a reshare had already taken place (#5655).
```
jumaffre, line 13: Should we also add a note about this [here](https://microsoft.github.io/CCF/main/operations/recovery.html)?

src/crypto/sharing.cpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+      res = entropy->random64() & prime;
+    }
+    return res;
+  }
+
+  /* POLYNOMIAL SHARING AND INTERPOLATION */
+
+  void sample_polynomial(
+    element p[], size_t degree, const crypto::EntropyPtr& entropy)
```
jumaffre, line 84:
```suggestion
    element* p, size_t degree, const crypto::EntropyPtr& entropy)
```

achamayou: That seems worse if anything, but I can make it a span. It's completely internal, so I'm not too worried.

CHANGELOG.md
achamayou: We can do but it's difficult to do it right without knowing in which version we will release this change. I think it's probably 4.0.9 and 5.0.0-dev2.

fournet (approved):
Some minor comments: 
- In sample_secret_and_shares, we could add `num_share >= threshold` as a sanity check.
- We might add support for recovery when we have more than `threshold` shares, e.g. checking the additional shares are consistent, or trying out different subsets if some of them are corrupt.   
- The formatting of the raw secret before hashing seems implementation-dependent.
- I did not review the code that uses the shared wrapping key; I suggest documenting that the wrapping key is *not* authenticated and should be used (e.g.) only for a single AEAD encryption, whose decryption will fail if the shares were corrupt.

achamayou:
@fournet thank you for the review:

> * In sample_secret_and_shares, we could add `num_share >= threshold` as a sanity check.

Yes, this is done here https://github.com/microsoft/CCF/pull/5655/files#diff-b17bb6f472ef77d108c2e13cb27cc676480ef2c7c46f5badb67650fd84c07999R111

> * We might add support for recovery when we have more than `threshold` shares, e.g. checking the additional shares are consistent, or trying out different subsets if some of them are corrupt.

Yes, this is mentioned here https://github.com/microsoft/CCF/pull/5655/files#diff-e3a85ec1e43867c1051c00c9ea0d4efb6b2e3ccded75c4a9db760b127ace1bb9R320, but I will open an issue for this possible refinement. It's not something the previous implementation did, nor something we've heard user requests for, so we probably won't give it a high priority.

> * The formatting of the raw secret before hashing seems implementation-dependent.

It's architecture dependent, like CCF's other internal serialisation (ledger, consensus etc). It is not compiler dependent though (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c183-dont-use-a-union-for-type-punning, http://eel.is/c++draft/basic.lval#11.3).

> * I did not review the code that uses the shared wrapping key; I suggest documenting that the wrapping key is _not_ authenticated and should be used (e.g.) only for a single AEAD encryption, whose decryption will fail if the shares were corrupt.

Will add a comment to that effect.

achamayou:
For reference:

```
share:
           share_10s_d1 * |      10 |     0.172 |   17161 |      - |    58271.0
            share_100s_d1 |      10 |     0.170 |   16960 |  0.988 |    58960.5
           share_1000s_d1 |      10 |     0.557 |   55705 |  3.246 |    17951.5
             share_10s_d5 |      10 |     0.855 |   85503 |  4.982 |    11695.4
            share_100s_d5 |      10 |     0.899 |   89930 |  5.240 |    11119.7
           share_1000s_d5 |      10 |     3.196 |  319572 | 18.622 |     3129.2
   share_n_recover_10s_d1 |      10 |     0.138 |   13785 |  0.803 |    72539.5
  share_n_recover_100s_d1 |      10 |     0.208 |   20838 |  1.214 |    47987.4
 share_n_recover_1000s_d1 |      10 |     0.517 |   51669 |  3.011 |    19354.0
   share_n_recover_10s_d5 |      10 |     0.869 |   86867 |  5.062 |    11511.9
  share_n_recover_100s_d5 |      10 |     1.088 |  108763 |  6.338 |     9194.2
 share_n_recover_1000s_d5 |      10 |     3.109 |  310904 | 18.117 |     3216.4
```

achamayou:
Switched to HKDF

```
share:
           share_10s_d1 * |      10 |     0.135 |   13474 |      - |    74216.5
            share_100s_d1 |      10 |     0.207 |   20662 |  1.533 |    48397.1
           share_1000s_d1 |      10 |     0.514 |   51409 |  3.815 |    19451.8
             share_10s_d5 |      10 |     0.676 |   67609 |  5.018 |    14790.9
            share_100s_d5 |      10 |     1.074 |  107396 |  7.971 |     9311.3
           share_1000s_d5 |      10 |     3.263 |  326304 | 24.217 |     3064.6
   share_n_recover_10s_d1 |      10 |     0.173 |   17342 |  1.287 |    57663.1
  share_n_recover_100s_d1 |      10 |     0.172 |   17188 |  1.276 |    58180.1
 share_n_recover_1000s_d1 |      10 |     0.516 |   51608 |  3.830 |    19376.8
   share_n_recover_10s_d5 |      10 |     0.866 |   86645 |  6.430 |    11541.3
  share_n_recover_100s_d5 |      10 |     1.086 |  108557 |  8.057 |     9211.7
 share_n_recover_1000s_d5 |      10 |     3.096 |  309575 | 22.976 |     3230.2
```

PR #5659, opened 2023-09-18
Fix `REQUIRE_OPENENCLAVE=OFF` builds

jumaffre:
This flag was introduced in https://github.com/microsoft/CCF/pull/5308 but I don't think it ever worked properly. It's not currently tested in the CI as we don't have any specific use case for it (yet?) but is nonetheless useful to build CCF on non-standard platforms (e.g. non Ubuntu 20.04) that don't have Open Enclave support.


src/apps/js_generic/js_generic.cpp
```diff
@@ -14,7 +14,7 @@ namespace ccfapp
 
   std::vector<ccf::js::FFIPlugin> get_js_plugins()
   {
-#if defined(INSIDE_ENCLAVE) || defined(SGX_ATTESTATION_VERIFICATION)
+#if defined(INSIDE_ENCLAVE) && defined(SGX_ATTESTATION_VERIFICATION)
     return {ccf::js::openenclave_plugin};
```
achamayou, line 18: The INSIDE_ENCLAVE check here seems slightly misleading, because:

1. We don't create interpreters outside an INSIDE_ENCLAVE context
2. There is no reason this shouldn't work from host code if oehost_verify is linked

so I think we should just check SGX_ATTESTATION_VERIFICATION

CMakeLists.txt
```diff
@@ -624,7 +624,9 @@ if(COMPILE_TARGET STREQUAL "sgx")
     ccf.enclave ${CCF_IMPL_SOURCE} ${CCF_GENERATED_DIR}/ccf_t.cpp
   )
 
-  target_compile_definitions(ccf.enclave PUBLIC PLATFORM_SGX)
+  target_compile_definitions(
+    ccf.enclave PUBLIC PLATFORM_SGX SGX_ATTESTATION_VERIFICATION
```
eddyashton, line 628: Can we add a comment explaining the intent of this define? I assume it's "this build should be able to verify SGX attestations"?

eddyashton (approved):

PR #5660, opened 2023-09-18
Implement Azure-compliant governance interface

### LINKED ISSUE #5211: New governance service endpoints
We should add new governance endpoints to query the state of the service, in line with the guidelines outlined in https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md and https://microsoft.github.io/typespec/introduction/style-guide#naming-convention. 

Notable requirements:
- camelCase payload bodies for values.
- Version query string (

⚠️ **Proposal still WIP** ⚠️ 

- `GET /gov/service/consortium`
Returns map of members (memberId -> memberInfo)

Sample:
```json
{"memberid1": {"status": "Active", "memberData": {"some": "arbitrary", "json": "object"}, "certificate": "-----BEGIN CERTIFICATE-----...."}, 
"memberid1": {"status": "Active", "memberData": {"some": "arbitrary", "json": "object"}, "certificate": "-----BEGIN CERTIFICATE-----...."}}
```

- `GET /gov/service/constitution`
Returns JavaScript constitution as JSON string.

Sample:
```json
{"constitution": "javascript constitution string"}
```

- `GET /gov/service/javascriptApp`
Returns JavaScript app endpoints, modules, JSEngine configuration and QuickJS version

Sample:
```json
{"endpoints": {"/echo": {"post": {"jsModule": "foo.js", "jsFunction": "bar", "forwardingRequired": "sometimes", "authnPolicies": ["user_cert", "member_cert"], "mode": "readOnly", "openApi": {"some": "arbitrary", "json": "object"}}}}, ""}
```

- `GET /gov/service/nodes`
Returns map of nodes (nodeId -> nodeInfo)

Note: Some fields (e.g. `ledger_secret_seqno`) are deliberately omitted as for internal and operators use only.

Sample:
```json
{
    "nodeid1": {
        "status": "trusted", 
        "quoteInfo": {
            "format": "OeSgxV1", 
            "quote": "base64", 
            "endorsements": "base64", 
            "uvmEndorsements": "base64", 
            "endorsedTcb": "tcb"
        },
        "codeDigest": "hex",
        "nodeData": {
            "some": "arbitrary",
            "json": "object"
        }, 
        "retiredCommitted": true,
        "certificate": "-----BEGIN CERTIFICATE-----....",
        "nodeToNodeInterface": {
            "publishedAddress": "host:port", 
            "appProtocol": "http1",
        },
        "rpcInterfaces": {
            "interface1Name": {
                "publishedAddress": "host:port", 
                "appProtocol": "http1",
            },
            "interface2Name": {
                "publishedAddress": "host:port", 
                "appProtocol": "http1",
            },
        }   
    }
}
```

- `GET /gov/service/joinPolicy`
Returns map of join policies keyed by TEE platform (currently, "sgx", "snp")

Sample:
```json
{
    "sgx": {"measurements": ["mrenclave1", "mrenclave2"]},
    "snp": {
        "measurements": [
            "measurement1",
            "measurement2"
        ], 
        "hostData": {
            "hash1": "expanded value1", 
            "hash2": null
        }, 
        "UVMEndorsements": {
            "did1": {
                "feed1": {"svn": "svn1"}
            },
            "did2": {
                "feed2": {"svn": "svn2"}
            }
        }
    }
}
```

- `GET /gov/service/jwk`
Returns information about OpenID Connect identity providers

Sample:
```json
{
    "issuers": {
        "issuer name 1": {
            "keyFilter": "all", 
            "keyPolicy": {
                "sgxClaims": []
            },
            "caCertBundleName": "cacertbundle1", 
            "autoRefresh": true,
            "kids": ["publickey1", "publickey2"]
        },
        "issuer name 2": {
            ...      
        }
    },
    "caCertBundles": {
        "cacertbundle1": "..."
    }
}
```

- `GET /gov/service/info`
Returns information about service

Sample:
```json
{
    "status": "open",
    "certificate": "-----BEGIN CERTIFICATE-----....", 
    "recoveryCount": 0,
    "serviceData": {
        "some": "arbitrary",
         "json": "object"
    },
    "currentServiceTxId": "2.0",
    "configuration": {
        "maximumNodeCertificateValidityDays": 90,
        "recentCoseProposalsWindowSize": 30
    },
    "previousServiceIdentity": "-----BEGIN CERTIFICATE-----....",     
}
```

eddyashton:
Resolves #5211.

Initial open comment:
_This contains some TODOs highlighting behaviour changes that may need discussing, or tracking in issues._

This implements a new API for governance, matching a TypeSpec source located [here](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/confidentialledger/Microsoft.ManagedCcf). Behaviourally this is very similar to `member_frontend.h`, but extracted for clarity and with manual construction of responses to try and make API-compliance explicit. Unlike other frontends, the endpoints here are registered by several independent static functions rather than directly in the `Registry` constructor. This is to confirm that such a code division is viable, and an attempt to mirror the source TypeSpec layout to aid conformance validation.

The current testing is incomplete. A few end-to-end tests of the governance API have new `_az_api` variants, which use the new APIs when calling anything in `member.py`. There are many other tests that call endpoints directly, so I'll look at wrapping those separately. There's likely some improvement in naming/access of this variant test too. For now we want to test both APIs, but we will shortly deprecate the old API and then be able to convert most testing to this replacement API.

A remaining axis for testing is how closely this aligns to the TypeSpec/OpenAPI - I think the correct way to do that is with a generated `autorest` client, but current testing suggests that's surprisingly lax on response schema validation. I'll explore this separately.


python/utils/submit_recovery_share.sh
```diff
@@ -68,8 +68,8 @@ fi
...
+encrypted_share=$(curl -sS --fail -X GET "${node_rpc_address}/gov/recovery/encrypted-shares/${member_id}?api-version=2023-06-01-preview" "${@}" | jq -r '.encryptedShare')
 
 # Then, decrypt encrypted share with member private key submit decrypted recovery share
 # Note: all in one line so that the decrypted recovery share is not exposed
-echo "${encrypted_share}" | openssl base64 -d | openssl pkeyutl -inkey "${member_enc_privk}" -decrypt -pkeyopt rsa_padding_mode:oaep -pkeyopt rsa_oaep_md:sha256 | openssl base64 -A | jq -R '{share: (.)}' | curl -i -sS --fail -H "Content-Type: application/json" -X POST "${node_rpc_address}"/gov/recovery_share "${@}" -d @-
\ No newline at end of file
+echo "${encrypted_share}" | openssl base64 -d | openssl pkeyutl -inkey "${member_enc_privk}" -decrypt -pkeyopt rsa_padding_mode:oaep -pkeyopt rsa_oaep_md:sha256 | openssl base64 -A | jq -R '{share: (.)}' | curl -i -sS --fail -H "Content-Type: application/json" -X POST "${node_rpc_address}/gov/recovery/members/${member_id}:recover?api-version=2023-06-01-preview" "${@}" -d @-
```
eddyashton: I don't intend to replace this entirely, as it means the previous path is untested. I'll switch this to take URLs as args or a switch argument or something.

src/node/gov/gov_endpoint_registry.h
```diff
@@ -0,0 +1,52 @@
...
+#include "node/gov/handlers/service_state.h"
+#include "node/gov/handlers/transactions.h"
+#include "node/share_manager.h"
+
+namespace ccf
+{
+  // Eventually, this should extend BaseEndpointRegistry, rather than
+  // CommonEndpointRegistry. But for now, we still support the old gov API by
+  // extending this, and that includes the common endpoints
+  class GovEndpointRegistry : public CommonEndpointRegistry
```
eddyashton, line 19: Worth highlighting this implementation pattern.

We can't have the old `MemberEndpointRegistry` and the new `GovEndpointRegistry` as separately installed frontends/registries, because they both want to own the `/gov` prefix. So what I've currently got is `Member extends Gov extends Common extends Base`, so that `/gov` exposes:
- `/gov/ack` (from Member, for people still using the old API)
- `/gov/members/state-digests/{memberId}:ack` (from Gov, part of new API)
- `/gov/commit` (from Common, for people still using the old API)

When the old API is no longer supported, we remove `Member` and instantiate `Gov` directly. At that point I think we can and should remove the `Common` intermediate class here, since the common APIs aren't part of the documented `/gov` API and I don't think they're necessary: The endpoints which _are_ necessary for basic governance interaction, such as `/commit` and `/tx`, have replacements in the new API, and the other endpoints can be accessed on `/app` or `/node`.

src/node/gov/handlers/acks.h
```diff
@@ -0,0 +1,322 @@
...
+          if (!sig.has_value())
+          {
+            // TODO: Behaviour change, why wasn't this an error previously?
+            detail::set_gov_error(
+              ctx.rpc_ctx,
+              HTTP_STATUS_INTERNAL_SERVER_ERROR,
+              ccf::errors::InternalError,
+              "Service has no signatures to ack yet - try again soon.");
+            return;
+          }
```
eddyashton, line 138: Behaviour change compared to the `update_state_digest` endpoint in `member_frontend.h`:

```
        auto s = sig->get();
        if (s)
        {
          ma->state_digest = s->root.hex_str();
          mas->put(member_id.value(), ma.value());
        }
```

If there's no signatures yet, we say the member can ack an empty (null?) state digest. But we don't even write that back to the KV. I think that's unsafe? This can only happen right at the start of a service, before a signature has been emitted, so will likely never occur in practice, but I err towards a clear error if it does occur.

src/node/gov/handlers/acks.h
```diff
@@ -0,0 +1,322 @@
...
+          // Ensure old HTTP signed req is nulled
+          ack->signed_req = std::nullopt;
+
+          // Insert new signature
+          ack->cose_sign1_req = std::vector<uint8_t>(
+            cose_ident.envelope.begin(), cose_ident.envelope.end());
+
+          // Store signed ACK in KV
+          // TODO: Behaviour change: don't automatically update the
+          // state_digest here
```
eddyashton, line 247: Minor behaviour change, that I think simplifies the API (each endpoint does "one thing"). The old `/ack` endpoint would _also_ do the work of updating and producing a fresh state digest. I think that's unnecessary, and avoiding it lets us return a `204` from this endpoint.

tests/suite/test_requirements.py
```diff
@@ -29,22 +29,22 @@ def ensure_reqs(check_reqs):
...
+            #         raise
+            #     else:
+            #         LOG.warning(
+            #             f'Test requirements not met, skipping "{func.__name__}": {e}'
+            #         )
+            #         return network
+            # except Exception as e:
+            #     raise TestRequirementsNotMet(
+            #         f"Could not check if test requirements were met: {e}"
+            #     ) from e
```
eddyashton, line 47: Must not commit this!

The actual req that was failing was endpoint-existence checks, I _believe_ those should all pass again now, will confirm.

src/node/gov/handlers/service_state.h
```diff
@@ -0,0 +1,637 @@
...
+          {
+            GOV_INFO_FMT("No recorded current_service_create_txid");
+          }
+
+          if (service_info->previous_service_identity_version.has_value())
+          {
+            // TODO: API definition is confused about the type of this. Decide
+            // whether this is a version, TxID, or cert
+            response_body["previousServiceCreationTransactionId"] =
+              service_info->previous_service_identity_version.value();
```
eddyashton, line 196: Need to update TypeSpec type here to match implementation, and decide if we're returning a pointer to the previous ident, or the ident itself.

src/node/gov/handlers/service_state.h
```diff
@@ -0,0 +1,637 @@
...
+
+            auto snp_endorsements = nlohmann::json::object();
+            auto endorsements_handle =
+              ctx.tx.template ro<ccf::SNPUVMEndorsements>(
+                ccf::Tables::NODE_SNP_UVM_ENDORSEMENTS);
+            endorsements_handle->foreach(
+              [&snp_endorsements](
+                const ccf::DID& did,
+                const ccf::FeedToEndorsementsDataMap& feed_info) {
+                snp_endorsements[did] = feed_info;
```
eddyashton, line 369: Need to update TypeSpec here, as it says this field is a string but we can actually return something more structured.

include/ccf/endpoint_registry.h
```diff
@@ -42,6 +42,51 @@ namespace ccf::endpoints
   void default_locally_committed_func(
     CommandEndpointContext& ctx, const TxID& tx_id);
 
+  template <typename T>
```
achamayou, line 45: That just seems to have moved, but is otherwise unchanged?

eddyashton: With hindsight, this is a breaking change. We should make this static in the class (I think static is all it needed?), and/or a shim so that the old calls still work.

src/node/gov/api_version.h
```diff
@@ -0,0 +1,82 @@
...
+
+  static constexpr std::pair<ApiVersion, char const*> api_version_strings[] = {
+    {ApiVersion::preview_v1, "2023-06-01-preview"}};
+
+  static const std::string& supported_suffix()
+  {
+    static std::string ss;
+    if (ss.empty())
+    {
+      ss = "Supported versions are:";
```
achamayou, line 25: I suspect a fmt::join(", ") would have been ok, we're probably not going to support dozens of versions at the same time.

src/node/gov/handlers/acks.h
```diff
@@ -0,0 +1,319 @@
...
+              ccf::errors::InvalidAuthenticationInfo,
+              fmt::format(
+                "Member ID from path parameter ({}) does not match "
+                "member ID from body signature ({}).",
+                member_id,
+                cose_ident.member_id));
+            return;
+          }
+
+          // Check an expected digest for this member is the KV
```
achamayou, line 201:
```suggestion
          // Check an expected digest for this member is in the KV
```

src/node/gov/handlers/acks.h
```diff
@@ -0,0 +1,319 @@
...
+          {
+            detail::set_gov_error(
+              ctx.rpc_ctx,
+              HTTP_STATUS_FORBIDDEN,
+              ccf::errors::AuthorizationFailed,
+              fmt::format("No ACK record exists for member {}.", member_id));
+            return;
+          }
+
+          // Check signed digest matches expected digest in KV
```
achamayou, line 215:
```suggestion
          // Check signed digest matches expected digest is in KV
```

src/node/gov/handlers/acks.h
```diff
@@ -0,0 +1,319 @@
...
+                " {}\n"
+                "Received\n"
+                " {}",
+                expected_digest,
+                actual_digest));
+            return;
+          }
+
+          // Ensure old HTTP signed req is nulled
+          ack->signed_req = std::nullopt;
```
achamayou, line 238: Does this lead to a "signed_req": null in the KV?

src/node/gov/api_version.h
eddyashton: I think there's 2 suggestions here:
- Separating entries with `, ` rather than `\n   `. I prefer the latter as I think it makes it easier to read + copy a single value, but don't think it makes a huge difference either way
- Using `fmt::join` rather than `ss +=`. I believe `fmt::join` needs a single iterator or iterator-like argument, and not sure how we spell iterator-to-first-element-of-each-pair-in-list in C++, which is why I went for the destructured for-loop

achamayou: I'd missed the second point, this is likely feasible with std::transform and a lambda that extracts the second value, but clearly not worth the extra syntax here.

On the first point, I feel like we've kept most (if not all?) HTTP error messages single line, and resorted to odata details where it seemed beneficial like exception details. I suggest it'd be good to check how other Azure services do that.

eddyashton: I've updated this to be a single line, and altered the OData error message to match the [guidelines](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#api-versioning):

> [✅](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#versioning-api-version-missing) DO return HTTP 400 with error code "MissingApiVersionParameter" and message "The api-version query parameter (?api-version=) is required for all requests" if client omits the api-version query parameter.

> [✅](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#versioning-api-version-unsupported) DO return HTTP 400 with error code "UnsupportedApiVersionValue" and message "Unsupported api-version '{0}'. The supported api-versions are '{1}'." if client passes an api-version value unrecognized by the service. For the supported api-versions, just list all the stable versions still supported by the service and just the latest public preview version (if any).

src/node/gov/handlers/acks.h
achamayou: Not sure about this to be honest, it turns one user step into two (*), and adds a possible failure case (refresh, fetch from out of date backup, fail to update), which wasn't possible when everything landed on the primary.

If it's what was signed off in the typespec, that's fine of course.

(*) the value of acking as a built-in is not clear to me, I think we would ideally move towards it being something that can be done through the propose-vote-apply flow. For example by allowing (or even requiring) a vote from a newly added member, and transitioning them straight to ACTIVE if they do that. A generic way to do that may be to add an `electors(proposal, ro_kv)` call to the constitution.

eddyashton: I don't think it's an extra step. The previous `/ack` would silently update the table containing what it expected a member to ack next, but didn't actually return this to the caller (it returns a `204 No Content`). So the user had to call `/update_state_digest` to retrieve a digest (there was no pure `GET`), causing a _further_ write, to read what they were expected to sign (or somehow get the expected digest from the ledger, but assume that's not the expected path). So with either API, the expected pattern is "call /update to get a 'fresh' (meaningful) digest, then /ack to submit your signature". The new API also has a `GET` for RESTy simplicity, which I guess is more likely to return a stale result and cause an error, but also allows some class of "restarting", if anyone cared.

Fully agree with the goal of removing built-in ACKing, but I think that's out-of-scope here. I think we could implement it with existing calls, as a slight modification of the sample constitution - every member has a proposal they can unilaterally pass to mark themselves as active, only members who have done this are counted in `resolve()`. Can include signing of some KV-stored freshness nonce if desired. Then the service loses any concept of member status/activeness, and instead a constitution which cares can choose to only count votes and proposals from active members.

achamayou: > The previous /ack would silently update the table containing what it expected a member to ack next, but didn't actually return this to the caller (it returns a 204 No Content). 

Ok, I'd missed that and misunderstood what was being split, I agree that shouldn't be bundled.

> Fully agree with the goal of removing built-in ACKing, but I think that's out-of-scope here.

Agreed, I wasn't suggesting it was in scope.

The KV-store freshness happens for free now that proposal ids are a combined digest of the tree root and the proposal, so I think a discrete digest is unnecessary. I think it's still worth exploring how to do consortium changes atomically, rather than keep the part of, but inactive state, which as far as I can tell is only ever used by accident. This is especially useful when rotating a member key without letting the quorum drop.

src/node/gov/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+It aims to expose all of the functionality of the existing API, defined in `src/node/rpc/member_frontend.h`, but with naming and schemas compliant with Azure API standards.
+
+For a transition period (at least the 4.x release), both APIs will be offered. The plan is to eventually deprecate the old API.
+
+Implementation notes:
+
+- All endpoints validate and process an `api-version` parameter, modifying their behaviour accordingly.
+- To present both under the `/gov` prefix, the old frontend is a subclass of the new frontend.
+- The frontend implementation into distinct components which can be more easily moved around, rather than a single monolithic frontend implementation.
```
achamayou, line 12:
```suggestion
- The frontend implementation is split into distinct components which can be more easily moved around, rather than a single monolithic frontend implementation.
```

tests/infra/consortium.py
```diff
@@ -285,9 +303,12 @@ def vote_using_majority(
...
                 if response.status_code != http.HTTPStatus.OK.value:
                     raise infra.proposal.ProposalNotAccepted(proposal, response)
-                proposal.state = infra.proposal.ProposalState(
-                    response.body.json()["state"]
-                )
+                body = response.body.json()
+                for k in ("state", "proposalState"):
+                    if k in body:
+                        proposal_state = body[k]
+                        break
```
achamayou, line 310:
```suggestion
                proposal_state = body.get("state", body.get("proposalState"))
                assert proposal_state, f"Could not find proposal state in {body}"
```

Seems marginally nicer

achamayou (approved):

src/node/gov/handlers/service_state.h
eddyashton: Tracking these in #5691, so resolving here.

achamayou (approved):

PR #5661, opened 2023-09-18
Refactor DropResponseWhenNotInState into arity-3 operator.

lemmy:
Can deduce (Leader or Candidate) expected_state parameter from the message parameter.


tla/ccfraft.tla
```diff
@@ -995,8 +995,11 @@ DropStaleResponse(i, j, m) ==
     /\ Discard(m)
     /\ UNCHANGED <<reconfigurationVars, serverVars, commitsNotified, candidateVars, leaderVars, logVars>>
 
-DropResponseWhenNotInState(i, j, m, expected_state) ==
```
achamayou, line 998: Could we take the opportunity to rename `i` -> `to`/`local` and `j` -> `from` in this operator? `m` is sufficiently ubiquitous that it may not need to become `msg`, but `i` and `j` are really unhelpful.

eddyashton (approved):

achamayou (approved):

tla/ccfraft.tla
lemmy: I suggest we do such a refactoring independently over all of the spec.

Issue #5384, opened 2023-06-20
Support CLI arguments for security-critical config


eddyashton:
As noted in https://github.com/microsoft/CCF/pull/5375#issuecomment-1598312695.

We have said that for SNP builds, the config is attested. This may be true, but in practice that would mean a unique measurement for every node (as the config includes some node-specific detail, such as their bind address), with a lot of associated governance churn. Instead, operators may put the config file in an unmeasured mount so that the config can differ between nodes while the measurement remains static. This means that the entire config is no longer covered by the attestation!

For fields with a security impact, we should provide a way to ensure they are included in the attested measurement. We think the easiest way to do this is to promote some arguments to top-level CLI args when the container is launched, such that they are invariably included in the policy-controlled measurements.

Specifically, it looks like we need to protect these fields:
- `enclave.file` - the string representation of the path should be contained in the measurement, _and the path itself_ (it should not be on an untrusted mount)
- `attestation.security_context_directory` - the string value should be contained in the measurement, and should be the name of a measured env var which itself contains a measured directory
- `logging.enclave_log_level` - (in-progress in #5375)

There are several steps we'll look at to fix this:
- Adding support for setting these on CLI rather than in monolithic config
- Deprecating these from the config in a future release
- Defining a linter which can ensure that certain properties are constrained by the measurement (ie - they specify a trusted path)

PR #5665, opened 2023-09-19
Pass enclave path as CLI argument rather than in configuration

eddyashton:
#5384 

We haven't deprecated a configuration option before, so I'm inventing the process here. There's a new CLI option that will be preferred if it is present. If it is not present, but the config entry is, we'll use the config entry and log a warning. The config entry is listed in the schema but no longer required. A future removal step should simply remove it from the schema (and struct definition), and remove the fallback assignment in the code.

NB: Still need to do the same for `attestation.security_context_directory`, will do that in a separate PR.


src/host/main.cpp
```diff
@@ -112,6 +112,10 @@ int main(int argc, char** argv)
       "Logging level for the enclave code")
     ->transform(CLI::CheckedTransformer(log_level_options, CLI::ignore_case));
 
+  std::string enclave_file_path;
+  app.add_option(
+    "--enclave-file", enclave_file_path, "Path to enclave application");
```
jumaffre, line 117: I think it would be nice to indicate why this is set as a CLI argument rather than in the config here. Maybe a `"(security critical)"` label here, and then a quick sentence in the output of `--help`?

jumaffre (approved):
I think the daily canary should be touched so that we check that we can still spin up very old nodes?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated `nghttp2` from `1.51.0` to `1.55.1`.
 - Converted SNP attestation UVM endorsements from integer to arbitrary string.
 - Updated Intel SGX PSW from 2.17 to 2.20 (#5616)
+- Path to the enclave file should now be passed as `--enclave-file` CLI argument to `cchost`, rather than `enclave.file` entry within configuration file. The configuration entry is deprecated, and will be removed in a future release.
```
```suggestion
- Path to the enclave file should now be passed as `--enclave-file` CLI argument to `cchost`, rather than `enclave.file` entry within configuration file. This is to ensure the path to the application file is attested on Confidential Containers/SNP, even if the configuration itself is provided from un-attested storage. The configuration entry is deprecated, and will be removed in a future release.
```

PR #5670, opened 2023-09-20
Remove two auxiliary variables from ccfraft.

lemmy:
Refactor


lemmy:
Low risk because it only removes auxiliary variables from ccfraft.

achamayou (approved):

PR #5686, opened 2023-09-25
Add CLI argument for SNP context directory

eddyashton:
#5384.

Deprecates config field `attestation.security_context_directory`, replaces with CLI `--snp-security-context-dir-var`. Includes default value `UVM_SECURITY_CONTEXT_DIR`, so this should never be a _required_ arg.

Also reshuffled which fields of the config are required - I thought that with the previous PR (#5665) we no longer needed the `.enclave` config entry at all, but in fact the `type` and `platform` fields _are_ required (always used), so I've updated the schema to require them.


CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - Converted SNP attestation UVM endorsements from integer to arbitrary string.
 - Updated Intel SGX PSW from 2.17 to 2.20 (#5616)
-- Path to the enclave file should now be passed as `--enclave-file` CLI argument to `cchost`, rather than `enclave.file` entry within configuration file. This is to ensure the path to the application file is attested on Confidential Containers/SNP, even if the configuration itself is provided from un-attested storage. The configuration entry is deprecated, and will be removed in a future release.
+- Path to the enclave file should now be passed as `--enclave-file` CLI argument to `cchost`, rather than `enclave.file` entry within configuration file. SNP security context directory environment variable should now be passed as `--snp-security-context-dir-var` CLI argument to `cchost`, rather than `attestation.environment.security_context_directory` entry within configuration file. This is to ensure that these values are attested on Confidential Containers/SNP, even if the configuration itself is provided from un-attested storage. The configuration entries are deprecated, and will be removed in a future release.
```
```suggestion
- Path to the enclave file should now be passed as `--enclave-file` CLI argument to `cchost`, rather than `enclave.file` entry within configuration file. A potential SNP security context directory environment variable override, where desired, should now be passed as `--snp-security-context-dir-var` CLI argument to `cchost`, rather than `attestation.environment.security_context_directory` entry within configuration file. This is to ensure that these values are attested on Confidential Containers/SNP, even if the configuration itself is provided from un-attested storage, such as an external mount. The configuration entries are deprecated, and will be removed in a future release.
```

doc/host_config_schema/cchost_config.json
```diff
@@ -475,7 +475,7 @@
             "required": ["url"],
             "additionalProperties": false
           },
-          "description": "List of servers used to retrieve attestation report endorsement certificates (SEV-SNP only). The first server in the list is always used and other servers are only specified as fallback. If set, endorsements from ``environment.security_context_directory`` are ignored"
+          "description": "List of servers used to retrieve attestation report endorsement certificates (SEV-SNP only). The first server in the list is always used and other servers are only specified as fallback. If set, endorsements from ``--snp-security-context-dir-var`` are ignored"
```
achamayou, line 478:
```suggestion
          "description": "List of servers used to retrieve attestation report endorsement certificates (SEV-SNP only). The first server in the list is always used and other servers are only specified as fallback. If set, attestation endorsements from ``--snp-security-context-dir-var`` are ignored, but uvm endorsements from that directory are still used."
```

achamayou (approved):

PR #5690, opened 2023-09-26
Add test against static snp report

DomAyre:
Checks the outputted report_data/measurement/host_data match the expected value

Expected values and reference report come from a run of the SNP CI
https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=76398&view=logs&j=4bdd83b0-b422-5e84-0058-fb00a4da194e&t=01ab89b5-2bb2-5b36-1c41-59a63e14a55c


tests/js-modules/modules.py
```diff
@@ -1106,7 +1106,54 @@ def corrupt_value(value: str):
...
+                {
+                    "evidence": reference_quote["raw"],
+                    "endorsements": reference_quote["endorsements"],
+                    "uvm_endorsements": reference_quote["uvm_endorsements"],
+                    "endorsed_tcb": reference_quote["endorsed_tcb"],
+                },
+            )
+            assert r.status_code == http.HTTPStatus.OK, r.status_code
+            assert r.body.json()["attestation"]["report_data"].startswith(
+                "949bfd58322a767aeb4e0093eab102b36eb90e63e9398a7d471347efebd4f0fb"
```
achamayou, line 1147: Why startswith and not equality?

achamayou:
@DomAyre can we dump the output to json, and compare directly against that?

tests/js-modules/modules.py
DomAyre: I believe that the report data for CCF is the SHA256 hash of a nodes public key, which is 32 bytes so only half of the available 64 bytes in report data, so the remaining 32 bits are zeros.

Thinking about it though, I reckon it's probably safer to assert that exactly

src/node/node_state.h
```diff
@@ -478,6 +478,11 @@ namespace ccf
 
       pal::PlatformAttestationReportData report_data =
         crypto::Sha256Hash((node_sign_kp->public_key_der()));
+      LOG_INFO_FMT("public_key={}", ds::to_hex(node_sign_kp->public_key_der()));
+      LOG_INFO_FMT(
+        "public_key_hash={}",
+        crypto::Sha256Hash((node_sign_kp->public_key_der())).hex_str());
```
achamayou, line 484: The public key can be extracted from the certificate written by the node to disk (typically under workspace/$testname_nodeid/nodeid.pem), or obtainable via tls:

node_pubk_hash=$(echo | openssl s_client -showcerts -connect "${stripped_node_address}" 2>/dev/null | openssl x509 -pubkey -noout | openssl ec -pubin -outform der 2>/dev/null | sha256sum | awk '{ print $1 }')

achamayou (approved):

achamayou (approved):

PR #5693, opened 2023-09-27
Fix out of bounds copy in SNP attestation report req

DomAyre:
We attempt to copy the full size of the request report data (64 bytes) which can be (and is) larger than the size of the input PlatformAttestationReportData (32 bytes)


achamayou:
@DomAyre at first glance, the attestation data on SNP should be 64 wide: https://github.com/microsoft/CCF/blob/88cb1eaeccbb650e791beb48331e216899856246/include/ccf/pal/report_data.h#L45

The change looks ok, but 1. we'd want an extra check to make sure the source size isn't bigger than destination, and 2. it doesn't look like the attestation data should be 32 wide on SNP.

DomAyre:
Yes the field is 64 bytes wide, but the content of the field is still just the hash of the nodes public id (aka 32 bytes) the rest should be all zero as far as I understand

DomAyre:
Yeah so my understanding is:

- From hardware's perspective, report_data is always 64 bytes so we pass 64 bytes in the request and get 64 bytes back
- From CCFs perspective, report data is a SHA256 Hash which is 32 bytes

So what we previously did was: 
1. Initialise the request which zeroes the req report data's 64 bytes
2. Copy 64 bytes of a 32 byte vector, leading to some garbage data being copied into the request and therefore leading to garbage coming back out

Since the report data is zero initialised, I don't think we have to manually zero it but we do have to specify to only copy 32 bytes of the vector to prevent the garbage data being copied over. 

I tested the quote produced by this PR as it stands and it does indeed only populate the first 32 bytes of the report_data as it should so I think we're good

include/ccf/pal/attestation.h
```diff
@@ -228,10 +228,7 @@ namespace ccf::pal
     snp::AttestationResp resp = {};
 
     // Arbitrary report data
-    memcpy(
-      req.report_data,
-      report_data.data.data(),
-      snp_attestation_report_data_size);
+    memcpy(req.report_data, report_data.data.data(), report_data.data.size());
```
achamayou, line 231:
```suggestion
    if (report_data.data.size() <= req.report_data.size())
    {
      std::copy(report_data.data.begin(), report_data.data.end(), req.report_data.begin());
    }
    else
    {
      throw std::logic_error("Insufficient space etc");
    }
```

achamayou:
> I tested the [quote produced](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=76513&view=logs&j=4bdd83b0-b422-5e84-0058-fb00a4da194e&t=01ab89b5-2bb2-5b36-1c41-59a63e14a55c&l=25962) by this PR as it stands and it does indeed only populate the first 32 bytes of the report_data as it should so I think we're good

I agree this works, and fixes the issue, but I think the code could use a cleanup to make it safer. We know the size of both the source and the target at compile time. For now if we can just add a runtime check, that would be good, I'll do another pass in a separate PR. Thank you for catching this!

include/ccf/pal/attestation.h
```diff
@@ -228,7 +228,14 @@ namespace ccf::pal
...
 
     // Arbitrary report data
-    memcpy(req.report_data, report_data.data.data(), report_data.data.size());
+    if (report_data.data.size() <= req.report_data.size())
+    {
+      std::copy(report_data.data.begin(), report_data.data.end(), req.report_data.begin());
+    }
+    else
+    {
+      throw std::logic_error("Insufficient space etc");
```
achamayou, line 237:
```suggestion
      throw std::logic_error("User-defined report data is larger than available space");
```

achamayou (approved):

PR #5697, opened 2023-09-29
Adding a propose request vote message to speed up some reconfigurations

achamayou:
The purpose of this change is to speed up code updates by skipping the election timeout, but extends more generally to any reconfiguration in which the primary is retired. Instead of waiting for a node in the next configuration to time out, the retiring primary can nudge a plausible candidate [*] to their succession.

[*] Plausible is currently defined as having the highest match index, with a tie breaker on belonging to a comparatively most recent configuration. This favours the "quickest/most up to date" node, even though it may itself be about to retire, because it has the best chances of winning the election and can itself do another quick handover later.

An alternative heuristic, discussed in comment, would be to prioritise nodes in the latest configuration, even if they aren't as up to date, so long as they are at least up to date enough to potentially win an election. I suspect this would be a worse choice, because it leads to greater chances of a multi-round election.

Quick comparison with the rotation test, doing two leader retirements in a row:

Without:

```
amchamay@amchamay:~/CCF/build-main$ ./tests.sh -VV -R rotation_test_cft | grep -i "new primary" 
62: 13:10:10.813 | INFO     | infra.network:wait_for_new_primary:1389 - Waiting up to 75.0s for a new primary different from 0 (51291d9b801bd6bf568f3640efa0cc9945ce8ec2cdccdce2d2ee63e8b28ff68d) to be elected...
62: 13:10:13.888 | INFO     | infra.network:wait_for_new_primary:1406 - New primary after 3.07s is 2 (64f3181bea310b99480561c0eaa70707e83f1e5bd7bced547a09c2ca369428fc) in term 3
...
62: 13:10:22.582 | INFO     | infra.network:wait_for_new_primary:1389 - Waiting up to 75.0s for a new primary different from 2 (64f3181bea310b99480561c0eaa70707e83f1e5bd7bced547a09c2ca369428fc) to be elected...
62: 13:10:24.735 | INFO     | infra.network:wait_for_new_primary:1406 - New primary after 2.15s is 1 (ad0f0cee5a96800378d3f6a9d205f8a42e0e31b82df7c485f52ac1948cd8af43) in term 4
...
```

With:

```
amchamay@amchamay:~/CCF/build$ ./tests.sh -VV -R rotation_test_cft | grep -i "new primary" 
62: 13:00:00.061 | INFO     | infra.network:wait_for_new_primary:1389 - Waiting up to 75.0s for a new primary different from 0 (f742655a06c574d0b727c551f5bd3b1516f5b458181443037b946530d2185d02) to be elected...
62: 13:00:00.146 | INFO     | infra.network:wait_for_new_primary:1406 - New primary after 0.08s is 1 (731a4a7753e5360086697dfe22edfd57dedc6a94e06d9ec5bc110d9e1592872e) in term 3
...
62: 13:00:08.496 | INFO     | infra.network:wait_for_new_primary:1389 - Waiting up to 75.0s for a new primary different from 1 (731a4a7753e5360086697dfe22edfd57dedc6a94e06d9ec5bc110d9e1592872e) to be elected...
62: 13:00:08.588 | INFO     | infra.network:wait_for_new_primary:1406 - New primary after 0.09s is 2 (96474d47986161d1826427eb0b3a7556c89d19943bbea0d5661b0141d725f364) in term 4
...
```


tla/ccfraft.tla
```diff
@@ -1065,6 +1076,13 @@ RcvUpdateCommitIndex(i, j) ==
         /\ UpdateCommitIndex(m.dest, m.source, m)
         /\ Discard(m)
 
+RcvProposeVoteRequest(i, j) ==
+    \E m \in MessagesTo(i) :
+        /\ j = m.source
+        /\ m.type = ProposeVoteRequest
```
lemmy, line 1082: Have you considered adding a conjunct `m.term = currentTerm[i]` to model `r.term == state->current_view` here?

tla/ccfraft.tla
```diff
@@ -729,6 +735,11 @@ AdvanceCommitIndex(i) ==
               \* Retire if i is not in active configuration anymore
               /\ IF i \notin configurations[i][Min(DOMAIN new_configurations)]
                  THEN /\ state' = [state EXCEPT ![i] = RetiredLeader]
+                      /\ LET msg == [type          |-> ProposeVoteRequest,
+                                     term          |-> currentTerm[i],
+                                     source        |-> i,
+                                     dest          |-> i]
+                         IN Send(msg)
```
lemmy, line 742: How about something like:

```tla
i = CHOOSE n \in { n \in Servers : \A m \in Servers : matchIndex[i][n] >= matchIndex[i][m] }: n \in MaxConfigurationIndex(i) => Send(msg)
```

`n \in MaxConfigurationIndex(i)` might not exactly capture what `raft.h` says about tie breaking.

tla/ccfraft.tla
achamayou: No, but I can see that it's a good idea :)

achamayou: Ok, having thought about this a little more, I am now unsure: while this holds if there is no delay, it's clearly something that may not be true if there is a delay in delivery. The term could go up, and the receiving node could receive a stale ProposeVoteRequest that it would then ignore, through no fault of the protocol.

So I think it would have to be `m.term <= currentTerm[i]`.

src/consensus/aft/raft_types.h
```diff
@@ -190,5 +192,14 @@ namespace aft
   DECLARE_JSON_TYPE_WITH_BASE(RequestVoteResponse, RaftHeader);
   DECLARE_JSON_REQUIRED_FIELDS(RequestVoteResponse, term, vote_granted);
 
+  struct ProposeRequestVote : RaftHeader
+  {
+    // Message sent by a leader stepping down after observing its own retirement
+    // being committed.
+    Term term;
```
eddyashton, line 199: I don't think this comment belongs here, as it's so precise its likely to rot quickly (we tweak the conditions on sending, or send in some additional case*).

If this definition needs a justification (because its not a "standard" message type), it should be something more general: "A node sends this to nudge another node to begin an election, for instance because the sender is a retiring primary"

\* Expanding on this slightly, but conjecture for now: I've got an idea, untested, that a slight extension of `RequestVoteResponse` would let a _failed_ election discover a _viable_ candidate. People vote against you because you're illegitimate; with a majority saying "no _because I'm at `X`_" you've got enough information to pick a node that a majority would think is legitimate (albeit on stale information). Prodding that node to become candidate immediately is preferable to waiting for a timeout.

tla/ccfraft.tla
eddyashton: I think an equality check (in the implementation) is preferable to an inequality (currently in this spec). Stale messages will be ignored, which feels safe. Reacting to stale (potentially extremely-delayed) messages may create a spurious election. That's obviously unlive if we had message duplication, and feels undesirable regardless.

eddyashton (approved):

tla/ccfraft.tla
achamayou: The implementation is using an [equality check](https://github.com/microsoft/CCF/pull/5697/files#diff-4e61c67c4990874cc31cb6f8cd91e4abe7506492fce0f73a4705d556d6861a42R1732), and ignores the message otherwise.

Perhaps what we want here is:

```
\/ /\ m.term == currentTerm[i]
    /\ Timeout(m.dest)
\/ m.term > currentTerm[i]
```

I have to say that I feel like we're sliding towards creating a duplicate of the implementation of equal verbosity, and I don't know if that's obviously a good thing.

src/consensus/aft/raft_types.h
achamayou: Ok, I'll simplify the comment.

tla/ccfraft.tla
eddyashton: I believe all of the `m.term > currentTerm[i]` cases are handled by `RcvUpdateTerm`, so we don't need to include the disjunct here. So to match the implementation, we'd add the single conjunct that @lemmy proposed:

```
...
/\ m.type = ProposeVoteRequest
/\ m.term = currentTerm[i]
/\ Timeout(m.dest)
...
```

achamayou: Ok, I'll simplify it again then.

lemmy: > I have to say that I feel like we're sliding towards creating a duplicate of the implementation of equal verbosity, and I don't know if that's obviously a good thing.

I've made the same observation.  At some point, somebody has to create a more high-level spec reusing on inspired by https://github.com/muenchnerkindl/AbstractRaft/, https://github.com/will62794/raft-by-refinement, or https://github.com/will62794/endive/blob/ind-tree/benchmarks/AbstractStaticRaft.tla, and show that ccfraft refines it.

tla/ccfraft.tla
achamayou: Fixed with help from @heidihoward, thank you!

tla/ccfraft.tla
```diff
@@ -481,6 +487,19 @@ AppendEntriesBatchsize(i, j) ==
...
     {nextIndex[i][j]}
 
+
+PlausibleSucessorNode(i) ==
+    \* Find a plausible successor node for i
+    \* Note that while CHOOSE is deterministic, it may not pick the same
+    \* node as the implementation, which iterates through an unordered_map
+    \* when performing its selection with the same criteria.
+    LET
+        activeServers == Servers \intersect removedFromConfiguration
```
heidihoward, line 497: I think `\intersect` should be `\` (set difference). `Servers \intersect removedFromConfiguration = removedFromConfiguration` since `removedFromConfiguration \subseteq Servers` according to the `RemovedFromConfigurationTypeInv`

tla/ccfraft.tla
```diff
@@ -481,6 +487,19 @@ AppendEntriesBatchsize(i, j) ==
      \* This can be redefined to send bigger batches of entries.
     {nextIndex[i][j]}
 
+
+PlausibleSucessorNode(i) ==
+    \* Find a plausible successor node for i
+    \* Note that while CHOOSE is deterministic, it may not pick the same
```
heidihoward, line 493: If you wanted to remove to `CHOOSE` to allow any plausible successor then you could have `PlausibleSucessorNode(i)` return the set of plausible successors for i. Then when you where calling `dest |-> PlausibleSucessorNode(i)`, you add a `\E j \in PlausibleSucessorNode(i): ` earlier and then do `dest |->j`. If there are multiple possible successors then the `AdvanceCommitIndex` action can be taken for each

tla/ccfraft.tla
```diff
@@ -481,6 +487,19 @@ AppendEntriesBatchsize(i, j) ==
...
+    \* node as the implementation, which iterates through an unordered_map
+    \* when performing its selection with the same criteria.
+    LET
+        activeServers == Servers \intersect removedFromConfiguration
+        highestMatchServers == {n \in activeServers : \A m \in activeServers : matchIndex[i][n] >= matchIndex[i][m]}
+        highestConfigServers == {n \in highestMatchServers : \A m \in highestMatchServers: MaxConfigurationIndex(n) >= MaxConfigurationIndex(m)}
```
heidihoward, line 499: Sorry, I'm a bit confused by this. `MaxConfigurationIndex(m)` is local state on a remote node, how does node `i` have access to this?

tla/ccfraft.tla
achamayou: Yes, absolutely, my mistake.

tla/ccfraft.tla
achamayou: The confusion is on my side, configuration is indeed a global, indexed by node. How nice it would be to have a type definition for it, and the ability to scope state by node... I'll fix this.

tla/ccfraft.tla
achamayou: > If there are multiple possible successors then the AdvanceCommitIndex action can be taken for each

This had failed to click in my head. Thank you, I have made the change.

tla/ccfraft.tla
achamayou: I believe `HighestConfigurationWithNode()` now does the right thing, and picks, on that server (i), out of `highestMatchServers`, those in the latest configuration.

lemmy: Minor: Like most of `ccfraft.tla`, most operator names are built from the term "server" instead of "node".

heidihoward (approved):

tla/ccfraft.tla
achamayou: @lemmy yes, but `server` refers to the top-level index into the globals (`configurations` in this case), and this call takes two arguments, `server` (that `server`'s local view of the configurations) and `node`, which is the node we want to know about.

I think it would be a lot nicer if we could scope the state properly, so that it isn't so trivially easy to access things that aren't available to the node, but that's beyond the scope of this change (no pun intended).

Issue #5691, opened 2023-09-27
Test conformance of gov API implementation to documented OpenAPI descriptions


eddyashton:
Hoping to do this automatically, perhaps with a generated client or OpenAPI-based fuzzer.

2 known discrepancies noted in #5660, which should be modified in the OpenAPI (TypeSpec) source:
- `previousServiceCreation` in `/service/info`. Decide whether this is a version, TxID, or cert
- SNP endorsements in `/service/join-policy`. Feed info is a known object, not a single string

eddyashton:
2 further discrepancies, added in #5698:
- `/members` and `/member/{memberId}` responses include `publicEncryptionKey`, if present
- `/jwk` response includes a `keys` field, describing current JWKs and not merely their issuers

ivarprudnikov:
> Hoping to do this automatically, perhaps with a generated client or OpenAPI-based fuzzer.

There was an issue related to fuzzing some time ago #1199 might be helpful to link it here

eddyashton:
- `Acks.StateDigest` response currently returns `stateDigest` _and_ `memberId`, but spec says single field called `digest`

achamayou:
> `previousServiceCreation` in /service/info. Decide whether this is a version, TxID, or cert

This is exposing [previous_service_identity_version](https://github.com/microsoft/CCF/blob/e6b9842036956d54e654a4c8151859f917fd0552/include/ccf/service/tables/service.h#L34), which is set during the [transition to open the service](https://github.com/microsoft/CCF/blob/e6b9842036956d54e654a4c8151859f917fd0552/src/node/node_state.h#L1476).

The last term of the previous service and the cert will both require a historical query, which will have much more of an API impact that we want to have for this iteration, so it's going to be the version.

eddyashton:
#6118 proposes adding 2 new endpoints, which should be added to the TypeSpec/OpenAPI:
- `GET /gov/service/users` - Returns list of users
- `GET /gov/service/users/{userId}` - Returns single users

Where the user schema is an object containing `{userId = str, certificate = PEM str, userData = null | JSON}`.

PR #5698, opened 2023-09-29
Use new governance API throughout our infra

eddyashton:
Changes the default argument of `--gov-api-version` to be the "new" API (`2023-06-01-preview`), with the intent of regularly bumping that to always prefer the latest value. Support for the "classic" API (at least the core proposal flow parts) is tested by some of the previously branched tests (`recovery_test_cft_api_0` and `membership_api_0`).

Mostly mechanical replacements. Broad insertion of `api_versioned_client` is a little unpleasant, may be worth defaulting to this in future? But annoying if that inserts an `?api_version=...` query arg in non-governance requests. A few modifications to the service API itself to support this - will track those in #5691.


src/node/gov/handlers/proposals.h
```diff
@@ -758,7 +758,7 @@ namespace ccf::gov::endpoints
         "/members/proposals/{proposalId}:withdraw",
         HTTP_POST,
         api_version_adapter(withdraw_proposal),
-        detail::active_member_sig_only_policies("withdraw"))
+        detail::active_member_sig_only_policies("withdrawal"))
```
eddyashton, line 761: In #5660, I'd made the new endpoint expect `ccf.gov.msg.type` to be `"withdraw"`, a totally new value. Since that's unreleased and an unnecessary change (and undocumented in the upstream API definition, since this is all within COSE), I think it makes sense to restore/reuse the existing `msg.type`.

achamayou: +1 for not breaking the `ccf_cose_sign1` script unnecessarily.

tests/code_update.py
```diff
@@ -474,24 +485,24 @@ def run(args):
...
-            test_add_node_with_no_uvm_endorsements(network, args)
-            test_host_data_table(network, args)
-            test_add_node_without_security_policy(network, args)
-            test_add_node_remove_trusted_security_policy(network, args)
-            test_start_node_with_mismatched_host_data(network, args)
-            test_add_node_with_bad_host_data(network, args)
-        test_add_node_with_bad_code(network, args)
-        # NB: Assumes the current nodes are still using args.package, so must run before test_proposal_invalidation
-        test_proposal_invalidation(network, args)
+        # test_verify_quotes(network, args)
```
achamayou, line 488: Presumably needs commenting back in?

tests/membership.py
```diff
@@ -302,31 +302,6 @@ def recovery_shares_scenario(args):
         )
 
 
-@reqs.description("Send an unsigned request where signature is required")
```
achamayou, line 305: This seemed like a useful test, why is it going?

tests/infra/network.py
```diff
@@ -1530,13 +1530,10 @@ def wait_for_snapshots_to_be_committed(src_dir, list_src_dir_func, timeout=20):
...
-                with node.client(self.consortium.get_any_active_member().local_id) as c:
-                    logs = []
-                    for _ in range(self.args.snapshot_tx_interval // 2):
-                        r = c.post("/gov/ack/update_state_digest", log_capture=logs)
-                        assert (
-                            r.status_code == http.HTTPStatus.OK.value
-                        ), f"Error ack/update_state_digest: {r}"
+                # Update state digest as a neutral write operation, to advance commit
+                member = self.consortium.get_any_active_member()
+                r = member.update_ack_state_digest(node)
```
achamayou, line 1535: That only seems to do a single write, where the code was previously adding half a snapshot tx interval worth, to make sure a snapshot was triggered. Does the test that make use of that still work? If so, it probably needs looking at.

tests/membership.py
eddyashton: This is a duplicate, and never called - there's another copy in `memberclient` which _is_ used.

achamayou (approved):

tests/infra/network.py
eddyashton: I've restored the multiple writes.

I _think_ that missing this will introduce an occasional failure in any test doing recoveries - _sometimes_ we will get an error from trying to use an uncommitted snapshot. However, I wasn't able to repro that in a local loop of `recovery` tests.

achamayou: Let's ticket this for further investigation.

eddyashton:
Final SNP failure is unrelated - merging.

Issue #5636, opened 2023-09-08
More efficient support for linearizable reads


heidihoward:
I believe the only way to achieve linearizable reads at the moment is for the app logic to create a phantom write to the kv to ensure that the primary replicates and reaches consensus on the associated txn. Without this, the read might be stale, even if forwarding is set to always and the client waits until the read txn ID is committed. This is because the primary who served the read may no longer be the current primary.

This is related to but somewhat separate from the discussion around wait for commit.

Here's some early ideas about how to make linearizable reads more efficient:

1. Allow txns with empty write sets - this probably the simplest approach. The app logic would still need to create a transaction for the read but at least with no write set, the transaction will have few conflicts and be smaller. It's a cleaner solution but I'm not sure how much is gained by this approach
2. Return the last sequence number +1 to the client for linearizable reads. This means that the client will not consider the read committed unless the primary commits another transaction (in the same term). The key issue here is that if the previous txn is a signature and no other write txns arrive then the client could be waiting indefinitely. This could be worked around by forcing another signature or using (1)
3. Implement PreVote or some type of leader stickiness such that if we trust clocks then the primary can serve linearizable reads without a round trip to a majority
4. Implement wait for commit and then delay responses to linearizable reads until after the next successful AppendEntries. This is closest to existing systems like etcd

_Aside: This could be an interesting to model in TLA+ using a separate high-level specification such as https://github.com/Azure/azure-cosmos-tla/blob/master/general-model/cosmos_client.tla_

lemmy:
> 
> _Aside: This could be an interesting to model in TLA+ using a separate high-level specification such as https://github.com/Azure/azure-cosmos-tla/blob/master/general-model/cosmos_client.tla_

The latest versions of the Cosmos specs are hosted at https://github.com/tlaplus/azure-cosmos-tla/ (including a completely new spec discussed in [Understanding Inconsistency in Azure Cosmos DB with TLA+](https://ieeexplore.ieee.org/document/10172859)).  Lorin Hochstein's [Linearizability spec](https://github.com/lorin/tla-linearizability) could also prove useful.

eddyashton:
>This is related to but somewhat separate from the discussion around wait for commit.

[...]

PR #5699, opened 2023-09-29
TLA+ specification for client consistency

heidihoward:
This PR adds a new TLA+ specification for client consistency guarantees

Following on from the discussion in https://github.com/microsoft/CCF/issues/5636


tla/consistency/ExternalHistory.tla
```diff
@@ -0,0 +1,343 @@
...
+\* Views start at 1, 0 is used a null value
+Views == Nat
+
+\* Sequence numbers start at 1, 0 is used a null value
+SeqNums == Nat
+
+\* TxIDs consist of a view and sequence number and thus start at (1,1)
+TxIDs == Views \X SeqNums
+
+\* This models uses a dummy applications where read-write transactions
```
achamayou, line 33:
```suggestion
\* This models uses a dummy application where read-write transactions 
```

tla/consistency/ExternalHistory.tla
```diff
@@ -0,0 +1,343 @@
...
+Views == Nat
+
+\* Sequence numbers start at 1, 0 is used a null value
+SeqNums == Nat
+
+\* TxIDs consist of a view and sequence number and thus start at (1,1)
+TxIDs == Views \X SeqNums
+
+\* This models uses a dummy applications where read-write transactions 
+\* append an integer to a list and then read the list
```
achamayou, line 34:
```suggestion
\* append an integer to a list and then reads the list
```

tla/consistency/ExternalHistory.tla
```diff
@@ -0,0 +1,343 @@
...
+\* Read-only transaction simply read the list
```
achamayou, line 35:
```suggestion
\* Read-only transactions simply read the list
```

tla/tlc.sh
```diff
@@ -3,7 +3,7 @@
 # Licensed under the Apache 2.0 License.
 # Original License below
 # Adapted from: https://github.com/pmer/tla-bin
-exec java -XX:+IgnoreUnrecognizedVMOptions -XX:+UseParallelGC -Xms2G -Xmx2G -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto
+exec java -XX:+IgnoreUnrecognizedVMOptions -XX:+UseParallelGC -Xms2G -Xmx2G -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto -fpmem 1
```
achamayou, line 6: It seems bizarre that we have to specify all these separate memory-related options on a relatively recent jvm...

tla/consistency/SingleNode.tla
```diff
@@ -0,0 +1,122 @@
...
+        /\ history' = Append(
+            history,[
+                type |-> TxStatusReceived, 
+                tx_id |-> history[i].tx_id,
+                status |-> CommittedStatus]
+            )
+    /\ UNCHANGED ledgers
+
+\* A CCF service with a single node will never have a view change
+\* so the log will never be rolled back and thus tranasction IDs cannot be invalid
```
achamayou, line 112:
```suggestion
\* so the log will never be rolled back and thus transaction IDs cannot be invalid
```

tla/consistency/SingleNode.tla
```diff
@@ -0,0 +1,122 @@
...
+                /\ history' = Append(
+                    history,[
+                        type |-> RwTxResponse, 
+                        tx |-> history[i].tx, 
+                        observed |-> [x \in 1..seqnum |-> ledgers[view][x].tx],
+                        tx_id |-> <<ledgers[view][seqnum].view, seqnum>>] )
+    /\ UNCHANGED ledgers
+
+\* Sending a committed status message
+\* Note that a request could only be committed if its in the highest view's ledger
```
achamayou, line 95:
```suggestion
\* Note that a request could only be committed if it's in the highest view's ledger
```

tla/consistency/SingleNode.tla
```diff
@@ -0,0 +1,122 @@
...
+\* SingleNode considers a single node CCF service, so no view changes
+
+EXTENDS ExternalHistory
+
+\* Upper bound of the number of client events
+\* Note that this abstract specification does not model CCF nodes so there's no
+\* constant for the number of nodes
+CONSTANT HistoryLimit
+
+\* Abstract ledgers that contains only client transactions (no signatures)
```
achamayou, line 14: Can we call those LedgerViews, or LedgerBranches?

tla/tlc.sh
heidihoward: We don't, I think we can tidy this up

tla/consistency/SingleNode.tla
heidihoward: done in https://github.com/microsoft/CCF/pull/5699/commits/19d1a33d16fd14ec975b7a4585f72d9efdff0cfb

tla/tlc.sh
achamayou: Let's find out #5700

tla/consistency/ExternalHistory.tla
```diff
@@ -0,0 +1,343 @@
...
+            /\ history[j].type = RoTxRequest
+            /\ history[j].tx = history[i].tx
+
+\* All responses must have an associated request earlier in the history (except tx status)
+AllReceivedIsFirstSentInv ==
+    /\ AllRwReceivedIsFirstSentInv 
+    /\ AllRoReceivedIsFirstSentInv
+
+\* Transaction IDs uniquely identify read-write transactions
+\* Note that is does not hold for read-only transactions
```
achamayou, line 119:
```suggestion
\* Note that it does not hold for read-only transactions
```

tla/tlc_debug.sh
```diff
@@ -0,0 +1,23 @@
...
+
+if [ $status -eq 12 ]; then
+  echo "Counterexample found as expected."
+  exit 0
+elif [ $status -eq 0 ]; then
+  echo "Counterexample expected but not found."  >&2
+  exit 1
+  else
+  exit $status
+fi
```
lemmy, line 23: Alternatively, consider a POSTCONDITION when there is a model where TLC is expected to find a counterexample.  That way, you can also assert the counter-example.

https://github.com/tlaplus/tlaplus/blob/c25a01393ef7d9b0315f3d3b1581988e7a4a57b2/tlatools/org.lamport.tlatools/test-model/Alias.tla#L65-L100

tla/consistency/ExternalHistory.tla
```diff
@@ -0,0 +1,343 @@
...
+
+\* Transaction IDs which received invalid status messages
+InvalidTxIDs ==
+    {history[i].tx_id: i \in InvalidEventIndexes}
+
+\* Highest commit sequence number
+CommitSeqNum == 
+    IF CommittedTxIDs = {} 
+    THEN 0
+    ELSE Max({i[2]: i \in CommittedTxIDs})
```
lemmy, line 83: Style: `Max({i[2]: i \in CommittedTxIDs} \cup {0})`

tla/consistency/MultiNode.tla
```diff
@@ -0,0 +1,54 @@
+---- MODULE MultiNode ----
+\* This specification extends SingleNode to model a multi-node CCF service
+
+EXTENDS SingleNode
+
+\* Upper bound on the view
+CONSTANT ViewLimit
```
lemmy, line 7: How about refactoring `ViewLimit` and `HistoryLimit` in `SingleNode` into a TLC state constraint.  Especially, because the term "view" also appears in `raft.h`.

tla/consistency/ExternalHistory.tla
heidihoward: fixed in https://github.com/microsoft/CCF/pull/5699/commits/f9b1224e2c6e0d3ec8c4de299963b0e3db652de0

.github/workflows/tlaplus.yml
```diff
@@ -57,3 +57,72 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v3
+      - run: python ./tla/install_deps.py
+
+      - name: consistency/SingleNode.cfg
+        run: |
+          set -exo pipefail
+          cd tla/
+          ./tlc.sh -config consistency/SingleNode.cfg consistency/SingleNode.tla 2>&1 | tee consistency/SingleNode.out
```
eddyashton, line 73: Minor: It should be possible to reduce the copy-pasting here with an [`each`](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#each-keyword). Something like:

```
# At top of file
parameters:
- name: consistencyFiles
  default:
  - consistency/SingleNode
  - consistency/SingleNodeReads
  ...

# Here
- ${{ each value in parameters.consistency}}:
  - name ${{ value }}
  - run: |
      set -exo pipefail
      cd tla/
      ./tlc.sh -config ${{ value }}.cfg ${{ value }}.tla 2>&1 | tee ${{ value }}.out
```

However, YAML (and ADO's YAML templating in-particular) are evil _and_ unhelpful, so getting the syntax correct is an annoying trial of errors.

tla/consistency/ExternalHistory.tla
```diff
@@ -0,0 +1,341 @@
...
+        /\ history[j].type = TxStatusReceived
+        /\ history[j].status = InvalidStatus
+        /\ history[k].type = RwTxResponse
+        /\ history[l].type = TxStatusReceived
+        /\ history[l].type = CommittedStatus
+        /\ history[k].tx_id = history[l]
+        /\ i # k
+        => history[i].tx \notin ToSet(history[k].observed)
+
+\* A history is serializable if there exists a execution sequence which is consistent
```
eddyashton, line 286:
```suggestion
\* A history is serializable if there exists an execution sequence which is consistent 
```

tla/consistency/SingleNode.tla
```diff
@@ -0,0 +1,122 @@
...
+        history, 
+        [type |-> RwTxRequest, tx |-> NextRequestId]
+        )
+    /\ UNCHANGED ledgerViews
+
+\* Execute a read-write transaction
+RwTxExecuteAction ==
+    /\ \E i \in DOMAIN history :
+        /\ history[i].type = RwTxRequest
+        \* Check transaction has not already been added a ledger
```
eddyashton, line 62:
```suggestion
        \* Check transaction has not already been added to a ledger
```

tla/consistency/SingleNode.tla
```diff
@@ -0,0 +1,122 @@
...
+        /\ \A view \in DOMAIN ledgerViews: 
+            {seqnum \in DOMAIN ledgerViews[view]: 
+                history[i].tx = ledgerViews[view][seqnum].tx} = {}
```
eddyashton, line 65: Is this just an empty-set check? I think `Cardinality(...) = 0` is more readably an activation condition, but I'll defer to people with more TLA+ experience.

tla/consistency/SingleNode.tla
```diff
@@ -0,0 +1,122 @@
...
+                        type |-> RwTxResponse, 
+                        tx |-> history[i].tx, 
+                        observed |-> [x \in 1..seqnum |-> ledgerViews[view][x].tx],
+                        tx_id |-> <<ledgerViews[view][seqnum].view, seqnum>>] )
+    /\ UNCHANGED ledgerViews
+
+\* Sending a committed status message
+\* Note that a request could only be committed if it's in the highest view's ledger
+StatusCommittedResponseAction ==
+    /\ Len(history) < HistoryLimit
```
eddyashton, line 97: These length constraints feel like a property of a config/run, rather than the spec itself. Is it possible to remove `HistoryLimit` from the spec, and cap `Len(history)` from the `.cfg` or CLI?

tla/consistency/SingleNode.tla
```diff
@@ -0,0 +1,122 @@
...
+\* Upper bound of the number of client events
+\* Note that this abstract specification does not model CCF nodes so there's no
+\* constant for the number of nodes
+CONSTANT HistoryLimit
+
+\* Abstract ledger views that contains only client transactions (no signatures)
+\* Indexed by view, each ledger is the ledger associated with leader of that view 
+\* In practice, the ledger of every CCF node is one of these or a prefix for one of these
+\* This could be switched to a tree which can represent forks more elegantly
+VARIABLES ledgerViews
```
eddyashton, line 18: It's taken me a few reads to grok the "view" here - sometimes it seems like an abstract, overloaded sense of "view", but this clarifies it's a specific consensus view (aka term). Is that mapping strictly required? In most places it seems this could just be `ledgers` ("some collection of (sequences of (view and tx tuples))"), but I assume that breaks somewhere?

.github/workflows/tlaplus.yml
achamayou: Having done that, a lot, on the main CI, I think I can safely say that I don't want to do it for just three values on one dimension.

eddyashton: I'm suggesting using this as a simple template for the tasks below, so 9 (or 27) repetitions.

tla/consistency/SingleNode.tla
lemmy: Most probably consider `... = {}` idiomatic TLA+.

tla/consistency/MultiNode.tla
heidihoward: I can add a state constraint using `TLCGet("level")` but it would be useful to keep the option to set views and history length separately. I can write a state constraint using views and history length but I'm not sure how to express the idea that once view limit is reach, the model can continue but one action (which increases view) cannot be taken

lemmy: Try `SomeActionConstraint == TLCGet("level") > 42 => ~OneAction`

.github/workflows/tlaplus.yml
achamayou: fwiw, this a sample in GHA using four dimensions https://github.com/microsoft/CCF/blob/main/.github/workflows/containers.yml

tla/consistency/SingleNode.tla
heidihoward: It was originally `ledgers`, changed in response to https://github.com/microsoft/CCF/pull/5699#discussion_r1341534825.  I've renamed to `ledgerBranches` to see if that helps. The idea behind variable is that it represents all possible ledgers at a given time (regardless of the number of nodes), instead of a concrete set of ledger

tla/consistency/MultiNode.tla
heidihoward: Sorry, I've not used action constraints before. I'm struggling to specify them in the cfg file, I've tried `ACTION-CONSTRAINT SomeActionConstraint` and  `ACTION-CONSTRAINTS SomeActionConstraint` to no avail. See screenshots below 

image

heidihoward: I've solved it, I needed to use `ACTION_CONSTRAINT` instead of `ACTION-CONSTRAINT`, based on the comment here https://github.com/tlaplus/tlaplus/blob/a41cbafc66b1dd225156aaca38ad35ec330f4ae9/toolbox/org.lamport.tla.toolbox.tool.tlc/src/org/lamport/tla/toolbox/tool/tlc/launch/TLCModelLaunchDelegate.java#L588

heidihoward: I have put in a PR to update cfg file syntax hosted by Apalache https://github.com/informalsystems/apalache/pull/2752.

.github/workflows/tlaplus.yml
achamayou: The matrix thing is nice mostly if you have A x B x C with only a few exceptions, in which case it will be substantially less work to tweak the dimensions than with the current denormalised scheme. If combinations are sparse and ad-hoc, then it's not a particularly good idea to use it.

tla/consistency/SingleNode.tla
heidihoward: I've moved these to separate MC files

tla/consistency/MultiNode.tla
lemmy: Instead of an ACTION_CONSTRAINT, consider adding the following to the MC*.tla file:

```tla
SN == INSTANCE SingleNode

ACOriginalAction ==
    /\ TLCGet("level") < 7
    /\ SN!OriginalAction
```

MC*.cfg
```
CONSTANT
   OriginalAction <- ACOriginalAction
```

Compare e.g. https://github.com/microsoft/CCF/blob/e73cf2f4cfb9a6231f1d52c729bb60094a1c14b0/tla/MCccfraft.tla#L19-L29

.github/workflows/tlaplus.yml
```diff
@@ -80,34 +80,36 @@ jobs:
...
 
       - name: Upload TLC's out file as an artifact. Can be imported into the TLA+ Toolbox.
         uses: actions/upload-artifact@v3
         if: ${{ failure() }}
         with:
           name: tlc
-          path: tla/consistency/*_TTrace_*.tla
+          path: | 
+            tla/consistency/*_TTrace_*.tla
+            tla/*.json
```
lemmy, line 112: FYI: dumpTrace json does *not* support serializing liveness violations.

tla/consistency/MultiNode.tla
heidihoward: Sorry, I should have updated this thread. I've actually already done that, see the MC* files in PR currently

lemmy: Hint: The `INSTANCE-OpRedefinition` trick has the advantage that you don't have to redefine the spec formulas. :-)

.github/workflows/tlaplus.yml
heidihoward: Thanks. @achamayou might be interested in this fyi as he add json dumpTrace in https://github.com/microsoft/CCF/issues/5703

achamayou: @lemmy thanks for the heads up!

achamayou:
@heidihoward this potentially very interesting to users building applications and APIs, and so ought to be featured in the documentation in my opinion.

There is a [TLA+ Specification](https://microsoft.github.io/CCF/main/architecture/raft_tla.html) under Architecture in the documentation (which probably needs a refresh, from having a quick glance at it). But I wonder if it wouldn't best to start a new top-level "Specifications", with "Consensus" and "Client Consistency" sub-entries.

What do you think?

achamayou (approved):

heidihoward:
> @heidihoward this potentially very interesting to users building applications and APIs, and so ought to be featured in the documentation in my opinion.
> 
> There is a [TLA+ Specification](https://microsoft.github.io/CCF/main/architecture/raft_tla.html) under Architecture in the documentation (which probably needs a refresh, from having a quick glance at it). But I wonder if it wouldn't best to start a new top-level "Specifications", with "Consensus" and "Client Consistency" sub-entries.
> 
> What do you think?

Sounds good, will save it for another PR though to get this merged

PR #5700, opened 2023-09-29
Are JVM defaults good enough for our purposes?

achamayou:
We seem to be using a lot of flags to tell the JVM about memory, and I am not convinced they are needed for our usage at this moment in time.

PR #5702, opened 2023-10-02
Centralise TLC options, and enable ParallelGC again

achamayou:
Follow up to #5700. This:

1. Centralises the place where we set TLC options
2. Splits CI vs local, since it looks like we often want different settings
3. Re-enables ParallelGC, since turning it off does seem to extend runtime by about 15%
4. Disables coverage in CI, since that extends runtime by about 100%

### DIFF
--- a/tla/tlc.sh
+++ b/tla/tlc.sh
@@ -3,7 +3,17 @@
 # Licensed under the Apache 2.0 License.
 # Original License below
 # Adapted from: https://github.com/pmer/tla-bin
-exec java -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto
+
+set -x
+
+if [ "${CI}" ]; then
+    JVM_OPTIONS=("-Dtlc2.TLC.ide=Github" "-Dutil.ExecutionStatisticsCollector.id=be29f6283abeed2fb1fd0be898bc6601")
+    TLC_OPTIONS=(-checkpoint 60)
+else
+    TLC_OPTIONS=(-coverage 60)
+fi
+
+exec java -XX:+UseParallelGC -Dtlc2.tool.impl.Tool.cdot=true "${JVM_OPTIONS[@]}" -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto -lncheck final "${TLC_OPTIONS[@]}"
 
 
 # Original License
--- a/tla/tlc.sh
+++ b/tla/tlc.sh
@@ -10,7 +10,7 @@ if [ "${CI}" ] || [ "${SYSTEM_TEAMFOUNDATIONCOLLECTIONURI}" ]; then
     JVM_OPTIONS=("-Dtlc2.TLC.ide=Github" "-Dutil.ExecutionStatisticsCollector.id=be29f6283abeed2fb1fd0be898bc6601")
 fi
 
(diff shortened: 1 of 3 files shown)


tla/tlc.sh
```diff
@@ -3,7 +3,17 @@
...
 # Adapted from: https://github.com/pmer/tla-bin
-exec java -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto
+
+set -x
+
+if [ "${CI}" ]; then
+    JVM_OPTIONS=("-Dtlc2.TLC.ide=Github" "-Dutil.ExecutionStatisticsCollector.id=be29f6283abeed2fb1fd0be898bc6601")
+    TLC_OPTIONS=(-checkpoint 60)
+else
+    TLC_OPTIONS=(-coverage 60)
```
heidihoward, line 13: If you'd like to simplify things, I think it's fine to remove this. I think most people using local dev will be using the vscode plug-in

heidihoward: On second thoughts, I think its best to remove this. The coverage info when printed to the stdout is kinda hard to read and might confuse beginners

achamayou: Agreed, taking it off.

tla/tlc.sh
```diff
@@ -3,7 +3,17 @@
 # Licensed under the Apache 2.0 License.
 # Original License below
 # Adapted from: https://github.com/pmer/tla-bin
-exec java -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto
+
+set -x
+
+if [ "${CI}" ]; then
+    JVM_OPTIONS=("-Dtlc2.TLC.ide=Github" "-Dutil.ExecutionStatisticsCollector.id=be29f6283abeed2fb1fd0be898bc6601")
+    TLC_OPTIONS=(-checkpoint 60)
```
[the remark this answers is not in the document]
achamayou: Ok

tla/tlc.sh
```diff
@@ -3,7 +3,14 @@
...
 # Adapted from: https://github.com/pmer/tla-bin
-exec java -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto
+
+TLC_OPTIONS=()
+
+if [ "${CI}" ] || [ "${SYSTEM_TEAMFOUNDATIONCOLLECTIONURI}" ]; then
+    JVM_OPTIONS=("-Dtlc2.TLC.ide=Github" "-Dutil.ExecutionStatisticsCollector.id=be29f6283abeed2fb1fd0be898bc6601")
+fi
+
+exec java -XX:+UseParallelGC -Dtlc2.tool.impl.Tool.cdot=true "${JVM_OPTIONS[@]}" -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -workers auto -lncheck final "${TLC_OPTIONS[@]}"
```
heidihoward, line 13: This change seems to have dropped `-tool`, not sure if @lemmy was using this

achamayou: Happy to restore that in the non-CI context, also not sure if it's used. It's always possible to pass additional arguments of course, this is only setting defaults.

lemmy: With `-tool` one can import the CI output in the Toolbox and VSCode.

achamayou: @lemmy I find the output very verbose and difficult to read, is there a way to generate `-tool` output to file separately (which we can publish and make available)?

lemmy: You would have to use tee and grep since there is nothing in TLC. Alternatively, let's remove the -tool option, as reproducing a violation is fast with our small models.

eddyashton (approved):

heidihoward (approved):

PR #5707, opened 2023-10-03
Run model checker in 1ES GHA pool

### LINKED ISSUE #5703: Investigate using larger runners for gh actions
TLA+ models in the gh actions ci are taking a long time compared to locally (one model takes 16sec in codespaces and >3mins in the ci with the same configuration) . The standard gh runners are 2-cores[1] but larger runners are available. Let's wait for https://github.com/microsoft/CCF/pull/5702 to merge before considering this

[1] https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

achamayou:
Resolves #5703 by moving the MC run a 1ES GHA pool with bigger nodes than the default GH pool.

This involved somewhat more changes than expected, but reduces runtime substantially. I've deliberately left the simulation where it is now, since it's single threaded and probably wouldn't benefit much while hogging the machines in the pool for quite some time.

With change:

image

Without:

image

### DIFF
--- a/.github/workflows/tlaplus.yml
+++ b/.github/workflows/tlaplus.yml
@@ -13,6 +13,8 @@ jobs:
   model-checking:
     name: Model Checking
     runs-on: [self-hosted, 1ES.Pool=gha-virtual-ccf-sub]
+    container:
+      image: ccfmsrc.azurecr.io/ccf/ci:05-09-2023-virtual-clang15
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/tlaplus.yml
+++ b/.github/workflows/tlaplus.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - run: python ./tla/install_deps.py
+      - run: python3 ./tla/install_deps.py
 
       - name: MCccfraft.cfg
         run: |
--- a/.github/workflows/tlaplus.yml
+++ b/.github/workflows/tlaplus.yml
@@ -26,19 +26,21 @@ jobs:
       - name: MCccfraft.cfg
         run: |
           cd tla/
-          ./tlc.sh -workers auto MCccfraft.tla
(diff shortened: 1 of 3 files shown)


.github/workflows/tlaplus.yml
```diff
@@ -12,48 +12,54 @@ on:
...
+      - run: |
+          sudo apt update
+          sudo apt install -y default-jre
+          python3 ./tla/install_deps.py
 
       - name: MCccfraft.cfg
         run: |
-          set -exo pipefail
           cd tla/
-          ./tlc.sh -workers auto MCccfraft.tla 2>&1 | tee MCccfraft.out
```
heidihoward, line 25: I think @lemmy added the stdout files, are you still using them?

.github/workflows/tlaplus.yml
```diff
@@ -12,48 +12,54 @@ on:
...
+          ./tlc.sh -workers auto MCccfraft.tla -dumpTrace tla MCccfraft.trace -dumpTrace json MCccfraft.json
```
heidihoward, line 29: Can the tla trace file have a .tla filetype? This helps with automatic syntax highlighting etc. The default naming such as `MultiNodeReads_TTrace_1696326252.tla` works nicely for this (also `*TTrace*.tla` is already in the .gitignore)

heidihoward (approved):

.github/workflows/tlaplus.yml
achamayou: Sure, we just need a pattern that's recognisable for upload, to separate the traces from the specs. Perhaps .trace.tla?

.github/workflows/tlaplus.yml
achamayou: I seem to have erased that now when updating the description, but the default shell GHA uses is sh -> dash, which does not support pipefail. And so if we pipe commands, we potentially drop errors from any but the last command, which means that we can't tell when the validation fails. That's an awkward price to pay for the .out file.

PR #5713, opened 2023-10-06
Update docs to describe new governance API

eddyashton:


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 [5.0.0-dev3]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.0-dev3
 
 - Added a `consensus.max_uncommitted_tx_count` configuration option, which specifies the maximum number of transactions that can be pending on the primary. When that threshold is exceeded, a `503 Service Unavailable` is temporarily returned on all but the `/node/*` paths (#5692).
-- A new versioned governance API is now available, with the `api-version=2023-06-01-preview` query parameter. This will fully replace the previous governance endpoints, which will be removed in a future release.
+- A new versioned governance API is now available, with the `api-version=2023-06-01-preview` query parameter. This will fully replace the previous governance endpoints, which will be removed in a future release. A guide to aid in upgrading from the previous API is available [here](https://microsoft.github.io/CCF/main/governance/gov_api_schemas/upgrading_from_classic.html)
```
eddyashton, line 13: Tweaking the changelog after we've cut is a little cheeky, but I think this is acceptable if the change is docs-only? And it would be reasonable to edit the Release Notes for `5.0.0-dev3` to match this change, pointing to the new docs once they're published.

CMakeLists.txt
```diff
@@ -652,7 +652,7 @@ configure_file(
   @ONLY
 )
 
-file(READ ${CCF_DIR}/src/node/gov/2023-06-01-preview.json
+file(READ ${CCF_DIR}/doc/schemas/mccf/2023-06-01-preview/mccfgov.json
```
eddyashton, line 655: Same document, just moved to live with the other OpenAPI documents.

doc/governance/accept_recovery.rst
```diff
@@ -27,31 +27,57 @@ A member proposes to recover the network and other members can vote on the propo
...
     {
-        "ballot_count": 0,
-        "proposal_id": "1b7cae1585077104e99e1860ad740efe28ebd498dbf9988e0e7b299e720c5377",
-        "proposer_id": "d5d7d5fed6f839028456641ad5c3df18ce963bd329bd8a21df16ccdbdbba1eb1",
-        "state": "Open"
+        "ballotCount": 0,
+        "proposalId": "1b7cae1585077104e99e1860ad740efe28ebd498dbf9988e0e7b299e720c5377",
+        "proposerId": "d5d7d5fed6f839028456641ad5c3df18ce963bd329bd8a21df16ccdbdbba1eb1",
+        "proposalState": "Open"
     }
```
eddyashton, line 45: While here I decided these example lines were getting very hard to read, so reformatted each argument onto its own line. This should make _future_ diffs more readable, though makes this diff even messier.

The real content change in this (and all of the similar instances) is replacing something like `/gov/proposals` with `/gov/members/proposals:create?api-version=2023-06-01-preview`, and updating the example response to show the new schema.

doc/governance/accept_recovery.rst
```diff
@@ -101,33 +135,36 @@ Summary Diagram
...
-        Node 2-->>Member 1: End of recovery procedure initiated.
-
-        Note over Node 2, Node 3: Reading Private Ledger...
-
-        Note over Node 2: Recovery procedure complete
-        Note over Node 3: Recovery procedure complete
+        participant Member A
+        participant Member B
+        participant Node 0
+        participant Node 1
```
eddyashton, line 141: We had an unused participant, and weird names for the others. We also showed no communication between the nodes, suggesting some kind of instantaneous process there (at which point they may as well be a single participant?).

Before:
image

After:
image

doc/governance/common_member_operations.rst
```diff
@@ -105,35 +131,6 @@ The number of member shares required to restore the private ledger (``recovery_t
...
-
-    $ ccf_cose_sign1 --ccf-gov-msg-type ballot --ccf-gov-msg-created_at `date -uIs` --ccf-gov-msg-proposal_id b9c08b3861395eca904d913427dcb436136e277cf4712eb14e9e9cddf9d231a8 --signing-key member3_privk.pem --signing-cert member3_cert.pem --content vote_accept_1.json | \
-      curl https://<ccf-node-address>/gov/proposals/b9c08b3861395eca904d913427dcb436136e277cf4712eb14e9e9cddf9d231a8/ballots --cacert service_cert.pem --data-binary @- -H "content-type: application/cose"
-    {
-        "ballot_count": 2,
-        "proposal_id": "b9c08b3861395eca904d913427dcb436136e277cf4712eb14e9e9cddf9d231a8",
-        "proposer_id": "2af6cb6c0af07818186f7ef7151061174c3cb74b4a4c30a04a434f0c2b00a8c0",
-        "state": "Accepted"
-    }
-
```
eddyashton, line 136: We've ended up with several copies of a full proposal flow (submission + votes) in surprising places. Ideally we'd have a single copy, and other places would just present a sample proposal body (assuming that's the only thing that's specific to each discussion point). That would massively reduce the amount of copy-paste fixups needed. In this file I'm sure this one is redundant so have deleted it, but will delay deleting all-but-one until I have more time to clean this up.

doc/governance/gov_api_schemas/upgrading_from_classic.rst
```diff
@@ -0,0 +1,251 @@
+Upgrading From Classic API
```
eddyashton, line 1: This is the bulk of the new content. Contains tables to point users from a classic endpoint to its replacement, with links to our existing (generated-from-OpenAPI) description of each.

Looks like this:
image

doc/schemas/mccf/2023-06-01-preview/examples/Ballots_Get.json
```diff
@@ -0,0 +1,16 @@
+{
```
eddyashton, line 1: We want to render `mccfgov.json` with `sphinxcontrib.openapi`. When it parses that OpenAPI, it blows up if the referenced example files aren't present. So we copy the examples as well, from the [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/tree/main/specification/confidentialledger/data-plane/Microsoft.ManagedCcf/preview/2023-06-01-preview) repo. However, since the OpenAPI is v2 rather than v3, the sphinx plugin can't actually _use_ the examples, and throws another error if we modify the directive to try and render them.

I'll see if we can remedy that, but including the files for now seems like the quick fix.

CHANGELOG.md
achamayou: I think it's completely fine to update the changelog to make it more detailed/helpful, here and in the release notes.

doc/governance/accept_recovery.rst
```diff
@@ -67,27 +93,35 @@ To restore private transactions and complete the recovery procedure, recovery me
...
+      --api-version 2023-06-01-preview \
+      --key member1_privk.pem \
+      --cacert service_cert.pem
     HTTP/1.1 200 OK
     content-type: text/plain
     x-ms-ccf-transaction-id: 4.30
     2/2 recovery shares successfully submitted. End of recovery procedure initiated.
 
-When the recovery threshold is reached, the :http:POST:`/gov/recovery_share` endpoint signals that the end of the recovery procedure is initiated and the that private ledger is now being recovered. Operators and members can monitor the progress of the private recovery process via the :http:GET:`/node/state` endpoint.
+When the recovery threshold is reached, the :http:POST:`/gov/recovery/members/{memberId}:recover` response signals that the end of the recovery procedure is initiated and the that private ledger is now being recovered. Operators and members can monitor the progress of the private recovery process via the :http:GET:`/node/state` endpoint.
```
achamayou, line 124:
```suggestion
When the recovery threshold is reached, the :http:POST:`/gov/recovery/members/{memberId}:recover` response signals that the end of the recovery procedure is initiated and that the private ledger is now being recovered. Operators and members can monitor the progress of the private recovery process via the :http:GET:`/node/state` endpoint.
```

doc/governance/accept_recovery.rst
achamayou: I suspect that was copy pasted in, I would just replace the nodes with Network here, I think this diagram is best kept focused on the member interactions only.

doc/governance/gov_api_schemas/2023-06-01-preview.rst
```diff
@@ -0,0 +1,8 @@
+2023-06-01-preview
+==================
+
+This API is available by passing the query parameter ``api-version=2023-06-01-preview`` to endpoints under the ``/gov`` prefix.
+
+This is available from CCF 5.0.0.
```
achamayou, line 6: -dev3?

doc/governance/accept_recovery.rst
eddyashton: Sure, that is much neater. Looks like this now:

image

doc/governance/proposals.rst
```diff
@@ -298,78 +306,110 @@ Here a new proposal has successfully been created, and nobody has yet voted for
...
+      "proposalId": "d4ec2de82267f97d3d1b464020af0bd3241f1bedf769f0fee73cd00f08e9c7fd",
+      "proposerId": "52af2620fa1b005a93d55d7d819a249ee2cb79f5262f54e8db794c5281a0ce73",
+      "proposalState": "Accepted"
     }
 
 Withdrawing a Proposal
 ----------------------
 
-At any stage during the voting process, before the proposal is accepted, the proposing member may decide to withdraw a pending proposal:
+At any stage during the voting process, before the proposal is accepted, the proposing member may unanimously withdraw a proposal by calling :http:POST:`/gov/members/proposals/{proposalId}:withdraw`:
```
achamayou, line 394:
```suggestion
At any stage during the voting process, before the proposal is accepted, the proposing member may unilaterally withdraw a proposal by calling :http:POST:`/gov/members/proposals/{proposalId}:withdraw`:
```

doc/schemas/mccf/2023-06-01-preview/examples/Ballots_Get.json
achamayou: That's unfortunate, including them does seem like the best solution for now. It may be worth checking the same of typespec docgen tooling, which is hopefully coming along, and which we can hopefully leverage at some point.

achamayou (approved):

doc/governance/proposals.rst
eddyashton: Thanks, that is what I meant!

doc/schemas/mccf/2023-06-01-preview/examples/Ballots_Get.json
achamayou: It looks like typespec may be supported [through typedoc](https://microsoft.github.io/typespec/language-basics/documentation#tsdoc-doc-comments), which we already use and have (some) cross-linking support for with Sphinx.

eddyashton: I'll take a look at this, but don't have high hopes. We really want to document the HTTP API, to a level where users can use it from their own clients. The TypeSpec source definitions are likely to be an unhelpful level of abstraction, since they talk about TypeSpec types (and operation names), rather than HTTP paths, requests, headers, and bodies. Perhaps we can determine "the" HTTP representation for these with the TypeDoc tooling, but I believe the best way to do that is via OpenAPI.

Another approach, where we _may_ end up, is generating clients from TypeSpec - in that case I hope the (generated) docs for those (generated) clients can also be picked up by Sphinx, and included in the core CCF docs. But I

PR #5716, opened 2023-10-09
Point-to-Point/Node-to-Node communication is ordered

lemmy:


tla/Traceccfraft.tla
```diff
@@ -118,6 +114,34 @@ TraceInit ==
 logline ==
     TraceLog[l]
 
+\* ccfraft assume ordered point-to-point communication. In other words, we should
```
achamayou, line 117:
```suggestion
\* ccfraft assumes ordered point-to-point communication. In other words, we should
```

tla/Traceccfraft.tla
```diff
@@ -118,6 +114,34 @@ TraceInit ==
...
+\* only receive the first pending message from j at i.  However, it is possible
+\* that a prefix of the messages sent by j to i have been lost -- the network is
+\* unreliable.  Thus, we allow to receive any message but drop the prefix up to
+\* that message.
+\* We could add a Traceccfraft!DropMessage action that non-deterministically drops
+\* messages at every step of the system.  However, that would lead to massive state
+\* space explosion.  OTOH, iit would have the advantage that we could assert that
```
achamayou, line 124:
```suggestion
\* space explosion.  OTOH, it would have the advantage that we could assert that
```

tla/Traceccfraft.tla
```diff
@@ -118,6 +114,34 @@ TraceInit ==
...
+\* messages is all empty at the end of the trace.  Right now, messages will contain
+\* all messages, even those that have been lost by the real system.  Another trade
+\* off that the length of the TLA+ trace will be longer than the system trace (due
```
achamayou, line 127:
```suggestion
\* off is that the length of the TLA+ trace will be longer than the system trace (due
```

achamayou:
@lemmy thank you for making that change! I suggest bumping up the simulation length, to perhaps 50 minutes, at least temporarily, to see if #5674 still takes place.

lemmy:
Good idea to increase simulation time.  However, this PR isn't ready yet because the trace validation should have failed for bad_network, fancy_election.1, and suffix_collision.2.  Will investiage.

lemmy:
The automation didn't detect that the "Run Trace Validation" step failed for some traces.

achamayou:
@lemmy the trace validation does look green: https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=77019&view=logs&jobId=0c81e209-1101-513f-1f15-37e9c3575eb2, but took 90+ minutes.

The validation was changed in https://github.com/microsoft/CCF/pull/5702 to use the wrapper, but not in a way that makes me expect that the return code should be swallowed. I'll check what's happened.

lemmy:
90+ minutes is expected because this PR massively increases the non-determinism.  This is something still to be solved.

lemmy:
This is ready to be reviewed.

The state-space explosion has been reduced, bringing the total runtime down to 25 minutes; not great but good enough for now.

achamayou (approved):

achamayou:
> The validation was changed in #5702 to use the wrapper, but not in a way that makes me expect that the return code should be swallowed. I'll check what's happened.

Investigating in #5720

lemmy:
FYI: The significant increase in validation time is not linked to state-space explosion. The state-space explosion resulting from dropping messages through `DropMessages` is actually quite modest. Instead, the inefficiency of TLC seems to be primarily associated with the evaluation of `INSTANCE` statements. To illustrate, the validation of a trace containing 360 lines, such as `fancy_election.2`, takes only around 30 seconds, as opposed to a considerably longer duration of approximately 20 minutes when we are *not* verifying the refinement of `ccfraft`.

achamayou:
@lemmy there's a massive low hanging fruit here though, validation on each scenario only really uses one core: https://github.com/microsoft/CCF/blob/main/tla/Traceccfraft.tla#L405, and we run them all sequentially: https://github.com/microsoft/CCF/blob/main/.azure-pipelines-templates/trace_validation.yml#L12

PR #5720, opened 2023-10-11
Add failing scenario to test CI is red on trace validation failure

achamayou:


achamayou:
CI is not red, despite trace validation clearly failing:

```
<<"Trace:", "../build/election_while_reconfiguration.ndjson", "Length:", 158>>
Computing initial states...
Finished computing initial states: 1 distinct state generated at 2023-10-11 08:27:02.
Error: Temporal properties were violated.

Error: The following behavior constitutes a counter-example:

State 1: <Initial predicate>
```

achamayou:
I always believed that the `script: |` task would fail if a single line returned non-zero, but that's clearly [not the case](https://github.com/microsoft/CCF/pull/5720/commits/34c0dda2fb9d7fee7ff7b1b68c1d6edfb49bb353). I can't find any mention of whether this is intentional under https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-script?view=azure-pipelines

The lesson seems to be that every time we have a `script: |` task, we need to `set -e` in the first line.

PR #5721, opened 2023-10-11
Improve error detail in raft message recv

achamayou:
Also add a missing check in the read (string) variant.

### DIFF
--- a/src/ds/serialized.h
+++ b/src/ds/serialized.h
@@ -4,19 +4,35 @@
 
 #include <cstdint>
 #include <cstring>
+#include <fmt/format.h>
 #include <stdexcept>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


src/ds/serialized.h
```diff
@@ -40,6 +57,12 @@ namespace serialized
   inline std::string read(const uint8_t*& data, size_t& size)
   {
     size_t len = read<size_t>(data, size);
+    if (size < len)
+    {
+      throw InsufficientSpaceException(
+        fmt::format("Insufficient space (read string: {} < {}", size, len));
```
eddyashton, line 63:
```suggestion
        fmt::format("Insufficient space (read string: {} < {})", size, len));
```

eddyashton (approved):

PR #5725, opened 2023-10-12
Run trace validation in parallel

achamayou:
Address https://github.com/microsoft/CCF/pull/5716#issuecomment-1759194879

Because the bulk of the time seems to be taken by the fancy_election.2 trace, this isn't as beneficial as I had expected.

### DIFF
--- a/.azure-pipelines-templates/configure.yml
+++ b/.azure-pipelines-templates/configure.yml
@@ -10,6 +10,7 @@ jobs:
       - checkout: self
         clean: true
       - script: |
+          set -ex
           echo "Determine if any code has changed."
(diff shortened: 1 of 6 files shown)


achamayou:
@lemmy oddly, this doesn't happen in my local runs:

```
Error: The configuration file substitutes constant Servers with non-constant TraceServers
```

achamayou:
> @lemmy oddly, this doesn't happen in my local runs:
> 
> ```
> Error: The configuration file substitutes constant Servers with non-constant TraceServers
> ```

This turns out to be TLC's quirky way of reporting that the file path does not exist (the spaces were stripped in formatting, and so all the paths were concatenated).

heidihoward (approved):

PR #5727, opened 2023-10-12
Client tracing experiment

achamayou:
This is a little experiment to see if we could easily build traces that can be checked with https://github.com/ligurio/elle-cli

The answer is, we can:

```
amchamay@amchamay:~/CCF/build$ ./tests.sh -VV -R client_tracin
...
55: 14:23:27.844 | INFO     | __main__:write:28 - [0|tls=user0] PUT /records/x 2
55: 14:23:27.844 | INFO     | __main__:write:28 - 204 @2.22 
55: 14:23:27.847 | INFO     | __main__:read:50 - [0|tls=user0] GET /records/x
55: 14:23:27.848 | INFO     | __main__:read:50 - 200 @2.22 2
55: 14:23:27.851 | INFO     | __main__:read:50 - [0|tls=user0] GET /records/x
55: 14:23:27.851 | INFO     | __main__:read:50 - 200 @2.22 2
...

amchamay@amchamay:~/elle-cli$ java -jar target/elle-cli-0.1.7-standalone.jar --model rw-register ~/CCF/build/trace.json -c strict-serializable
/home/amchamay/CCF/build/trace.json      true
```

What could be interesting to do next:

- [ ] call across nodes
- [ ] support waiting for commit (as opposed to pending)
- [ ] try the different consistency models
- [ ] do the same thing with piccolo files


achamayou:
It does not look like I will have time to pick this up soon, so parking for now.

PR #5728, opened 2023-10-12
Test that JS app code is correctly updated despite interpreter reuse

eddyashton:
Just a regression test confirming current behaviour. Although the endpoint tries to reuse an interpreter, the interpreter cache is correctly flushed during app update. Note that this test will fail if the constitution is modified to _not_ flush interpreters in `set_js_app:apply`.

### DIFF
--- a/tests/js-custom-authorization/custom_authorization.py
+++ b/tests/js-custom-authorization/custom_authorization.py
@@ -956,52 +956,50 @@ def run_interpreter_reuse(args):
     ) as network:
         network.start_and_open(args)
 
-        # TODO: Restore
-        # network = test_reused_interpreter_behaviour(network, args)  #
-        # network = test_caching_of_kv_handles(network, args)
+        network = test_reused_interpreter_behaviour(network, args)  #
+        network = test_caching_of_kv_handles(network, args)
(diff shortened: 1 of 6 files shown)


tests/js-custom-authorization/custom_authorization.py
```diff
@@ -914,6 +921,31 @@ def was_cached(response):
...
+            bundle = network.consortium.read_bundle_from_dir(args.js_app_bundle)
+            # Replace placeholder blindly, in the raw bundle JSON as string
+            s = json.dumps(bundle).replace("<func_caching_placeholder>", new_val)
+            bundle = json.loads(s)
+            network.consortium.set_js_app_from_bundle(primary, bundle)
+
+        for _ in range(5):
+            v = str(uuid.uuid4())
+            set_app_with_placeholder(v)
+            r = c.get("/app/func_caching")
```
achamayou, line 942: That only tests the primary, it would be good to check that the backups also receive the update when we expect (wait for commit level to reach the txid of the replacement before checking obvs), if we want to be really thorough here.

achamayou (approved):

Issue #5589, opened 2023-08-24
JS Runtime memory limit is not applied correctly


achamayou:
Issue spotted during work on #5564, and confirmed in #5588.

Depending on the situation in which the max heap size set by `JS_SetMemoryLimit()` is reached, the `ccf::js::Runtime` may cause a crash, rather than return an HTTP error response.

There are three situations in which we can reach the heap size limit:

1. As we construct globals, before (and after) the interpreter invocation
3. During JS execution
4. Inside a JS wrapper to a C++ call

Only situation 2 is handled correctly at the moment. #5588 confirms that situation is not correctly handled and can lead to a crash.

Situation 3 isn't tested yet, but almost certainly suffers from the same issue.

There is also a problem in situation 3, where some C++ exceptions may lead to a leak from the JS interpreter, since closing the Context may not free all memory allocated through JS calls. Further investigation is needed there.

achamayou:
Adding some details for situation 1, that's from entering JSHandlers::do_execute_request() until ctx.call(), modulo interpreter caching and re-using.

The issue during that phase is that there are a lot of unchecked calls to JS_New*, with or without JSWrappedValue. Any of them could produce a JS OOM exception, which is neither caught nor returned to the caller correctly. Every allocating call pre-run should be checked instead, JS exceptions reported, and the runtime destroyed (particularly in re-use cases).

achamayou:
Still looking at situation 1, i.e. calls made to JS prior to interpreter execution, there seem to be two broad categories of failures:

1. `JS_New*` returning a `JSValue` such that `JS_IsExcepti(val)` is true
2. `JS_Set* or `JS_Get*` returning -1, because they internally called `JS_New*` with the same result as above.

In both cases, I think we want to raise a C++ exception, which we will catch in js_generic_base. From there we destroy the runtime, and return a 500.

There are quite a lot of calls that need to be made safe, either by wrapping them at call site, or by proving wrappers similar to those in JSContext (but different in that they do not own the values, only check and throw).

The other alternative I can see is that we would temporarily unset the heap limit during runtime init, and re-instate it just prior to execution, with the result that we may go over slightly for a short period of time. This also only takes care of heap threshold execess, not other potential exceptions that may take place and result in a potentially unsafe interpreter context.

achamayou:
#5730 resolves situation 1.

[...]

PR #5730, opened 2023-10-13
Only enable JS runtime limits during execution

### LINKED ISSUE #5731: Handler invocations from js_kv_map_foreach get a full max_exec_time each
`js::Context::call()` sets a fresh start_time and max_exec_time in the runtime. Since this is called on each iteration by foreach, the total max duration is effectively reset every time. Found while looking at #5730.

achamayou:
Alternative to #5729 in addressing #5589. Also fixes #5731.

This change removes the interpreter heap, stack and runtime caps outside of JS code execution, which avoids having to handle those situations in the preparation of the runtime and of the response. In those contexts, the limits set by the environment (OS and/or SGX) apply instead, as they do in the rest of the code.

Note that this of course does not solve the issue of leaks in case a C++ exception is thrown in a wrapped function called from JS, this will need separate changes.

Also changed runtime options setting so that governance always has at least the default values available, to avoid getting to an unrecoverable situation where the values are so low that no further governance is possible.


src/js/wrap.cpp
```diff
@@ -137,21 +137,40 @@ namespace ccf::js
...
     }
+    rt.set_runtime_options(tx);
     const auto curr_time = ccf::get_enclave_time();
     interrupt_data.start_time = curr_time;
     interrupt_data.max_execution_time = rt.get_max_exec_time();
-    interrupt_data.access = access;
     JS_SetInterruptHandler(rt, js_custom_interrupt_handler, &interrupt_data);
 
+    const auto rv = W(JS_Call(
+      ctx, f, ccf::js::constants::Undefined, argv.size(), argvn.data()));
```
eddyashton, line 158: Can this call `inner_call`?

eddyashton (approved):

src/js/wrap.cpp
achamayou: Sure

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -111,6 +111,26 @@ def test_stack_size_limit(network, args):
         r = c.post("/app/recursive", body={"depth": safe_depth})
         assert r.status_code == http.HTTPStatus.OK, r
 
+        # Lower the cap until we likely run out of stack out of user code,
+        # and check that we don't crash and return an error message
```
eddyashton, line 115:
```suggestion
        # and check that we don't crash. Check we return an error message
```

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -138,6 +158,27 @@ def test_heap_size_limit(network, args):
...
+        cap = safe_size
+        while cap > 0:
+            cap //= 2
+            LOG.info(f"Heap max size: {cap}")
+            with temporary_js_limits(network, primary, max_heap_bytes=cap):
+                r = c.post("/app/alloc", body={"size": 1})
+                if r.status_code == http.HTTPStatus.INTERNAL_SERVER_ERROR:
+                    message = r.body.json()["error"]["message"]
+                    assert message == "Exception thrown while executing.", message
+                    LOG.info(f"Out of memory at max heap size: {cap}")
```
eddyashton, line 172: Should we be logging this discovered for the other cases? Or is this just a debug log that can be removed?

eddyashton (approved):

tests/js-custom-authorization/custom_authorization.py
achamayou: Just debug, it's not very precise anyway (it's a value at which we are out of memory, but not the highest).

achamayou, line 172:
```suggestion
```

Issue #5718, opened 2023-10-10
JS interpreter reuse can lead to memory violations by caching constructed state


eddyashton:
First noticed, and most likely to occur, for cached KV map handles:

```
globalThis.myTable = ccf.kv["foo"];
```

Within our interpreter code, this is constructing a `kv_map_handle_class_id` object in the JS runtime, containing a `KVMap::Handle*`. If that is accessed in a future invocation, it will likely be invalid.

This is also possible, though potentially non-fatal, for any object-like things we return:
- ccf.consensus
- ccf.kv
- ccf.crypto
- ccf.request

eddyashton:
Partial fix (for TypeScript apps going via `typedKv`) in #5717. For full fix, we need to track and invalidate any raw accesses.

achamayou:
Full fix in #5740

PR #5740, opened 2023-10-17
Don't store temporary pointers in JSValues

eddyashton:
Follow-up to #5717 (reverting that interim fix), resolving #5718.

The general pattern is that we avoid storing temporary things in the `Opaque` of a `JSValue`. Instead we store them at the top in the `js::Context`, flush them when we complete execution, and rehydrate/redirect when new calls come in on `JSValue`s that app code has stashed in the interpreter (this storage pattern should be no worse for lookup, is much easier for cleanup, and lets us memoise some KV-handle lookups).


src/apps/js_generic/js_generic_base.cpp
```diff
@@ -220,9 +220,9 @@ namespace ccfapp
       }
       request.set("params", params);
 
-      const auto& request_body = endpoint_ctx.rpc_ctx->get_request_body();
+      auto request_body = new std::vector<uint8_t>(endpoint_ctx.rpc_ctx->get_request_body());
```
achamayou, line 223: Why the copy? Even assuming JS_SetOpaque takes ownership, it's not great to have another call in between that could throw.

eddyashton: As discussed, have updated this to remove the new `new`s. The `body` will always refer to the _current_ request, and the `map_name` is stored as a property (`_map_name`) rather than an `Opaque`, so it should be cleaned up by the normal QuickJS flow rather than a finalizer.

src/js/historical.cpp
```diff
@@ -171,38 +165,16 @@ namespace ccf::js
...
-      JS_CHECK_SET(js_state.set("receipt", std::move(js_receipt)));
-
-      auto kv = jsctx->new_obj_class(kv_read_only_class_id);
-      JS_CHECK_EXC(kv);
-      JS_SetOpaque(kv, &state_ctx->tx_ctx);
-      JS_CHECK_SET(js_state.set("kv", std::move(kv)));
-
-      states_array.set_at_index(i++, std::move(js_state));
+      auto js_state = ccf::js::create_historical_state_object(jsctx, state);
+      // TODO: Wrapper for this?
```
achamayou, line 176: set_at_index(i++, std::move(js_state)), but js_state will need to wrapped if it's not

src/js/wrap.cpp
```diff
@@ -189,14 +190,104 @@ namespace ccf::js
     JS_FreeRuntime(rt);
   }
 
+  static KVMap::Handle* _get_map_handle(
+    js::Context& jsctx, JSValueConst this_val)
+  {
+    auto map_name =
+      jsctx.to_str(JS_GetPropertyStr(jsctx, this_val, "_map_name"));
```
achamayou, line 197: Might be good to wrap this_val, and then JS_CHECK_EXC() the rv of this_wrapped_val["_map_name"], just in case there's a problem retrieving it. There probably won't be in this case, but it's prudent and will give better errors.

src/js/wrap.cpp
```diff
@@ -211,19 +302,23 @@ namespace ccf::js
       return JS_ThrowTypeError(ctx, "Argument must be an ArrayBuffer");
     }
 
+    auto handle = handle_getter_(jsctx, this_val);
+    if (handle == nullptr)
+    {
+      return JS_ThrowTypeError(ctx, "Internal: Unable to access MapHandle");
```
achamayou, line 308: Possibly JS_ThrowInternalError rather Type error

src/js/wrap.cpp
```diff
@@ -358,28 +477,31 @@ namespace ccf::js
...
+    js::Context& jsctx = *(js::Context*)JS_GetContextOpaque(ctx);
 
     if (argc != 0)
     {
       return JS_ThrowTypeError(
         ctx, "Passed %d arguments, but expected 0", argc);
     }
 
+    auto handle = _get_map_handle(jsctx, this_val);
+    if (handle == nullptr)
```
achamayou, line 489: This seems to happen a lot, perhaps there's a JS_CHECK_HANDLE(), to go with the other JS_CHECK*() macros.

src/js/wrap.cpp
```diff
@@ -633,39 +777,37 @@ namespace ccf::js
...
   }
 
-  static int js_read_only_kv_lookup(
+  static int js_historical_kv_lookup(
     JSContext* ctx,
     JSPropertyDescriptor* desc,
     JSValueConst this_val,
     JSAtom property)
   {
+    LOG_INFO_FMT("In js_historical_kv_lookup");
```
achamayou, line 796: Presumably to be removed?

src/js/wrap.cpp
```diff
@@ -680,8 +822,13 @@ namespace ccf::js
       return JS_ThrowTypeError(
         ctx, "Passed %d arguments, but expected none", argc);
 
-    auto body = static_cast<const std::vector<uint8_t>*>(
-      JS_GetOpaque(this_val, body_class_id));
+    js::Context& jsctx = *(js::Context*)JS_GetContextOpaque(ctx);
+    auto body = jsctx.globals.current_request_body;
+    if (body == nullptr)
+    {
+      return JS_ThrowTypeError(ctx, "No request body set");
```
achamayou, line 829: Not sure about Type error here either, but I can see this is maybe subjective.

src/js/wrap.cpp
```diff
@@ -2239,40 +2323,45 @@ namespace ccf::js
...
-        state,
-        "transactionId",
-        JS_NewString(ctx, transaction_id.to_str().c_str()));
-      auto js_receipt = ccf_receipt_to_js(&ctx, receipt);
-      JS_SetPropertyStr(ctx, state, "receipt", js_receipt);
-      auto kv = JS_NewObjectClass(ctx, kv_read_only_class_id);
-      JS_SetOpaque(kv, historical_txctx);
-      JS_SetPropertyStr(ctx, state, "kv", kv);
+    const auto transaction_id = state->transaction_id;
+    auto transaction_id_s = jsctx.new_string(transaction_id.to_str().c_str());
```
achamayou, line 2342: I think the  c_str() isn't necessary here

achamayou:
@eddyashton can you poke the daily to get an ASAN run?

achamayou (approved):

PR #5746, opened 2023-10-18
Compute node count from a single pass over the JSON log file, eliminating the need for separate 'nodes' file.

lemmy:
nt


lemmy:
If this PR causes no regression, I will force-push a commit to remove the generation of the `.nodes` files:

https://github.com/microsoft/CCF/blob/a5495caeebc9c6209cfd0a06ad8fb0903621d270/tests/raft_scenarios_runner.py#L116-L119

achamayou (approved):

PR #5747, opened 2023-10-19
JS FFI exception safety update

### LINKED ISSUE #5589: JS Runtime memory limit is not applied correctly

achamayou:
Closes #5589.

I have now done a first pass on all the calls listed in the issue. There is still scope for cleanup, in particular in governance code where we don't use safe wrappers as extensively as we should. I've already removed some unused wrappers, but I'm still finding some.

I've had a look at removing the JSValue& conversion on the WrappedValue, but it's massive amount of further changes, and it's not necessary to fix the leak, so I think I would like to go ahead with this and unblock #5740.

Another thing that I've put in a pin in for now is that the crypto code (ie. wrappers for OpenSSL) mostly (but not exclusively) throws std::runtime_error to indicate validation failures, and std::logic_error is used for more internal failures. It's important to distinguish them when we validate an x509 chain for example, with some fuzziness. If one of the values in the chain is not a valid PEM for example, is that False or exception (I think it's exception, but it's arguable). Filed as #5760.


src/js/wrap.h
```diff
@@ -160,35 +232,61 @@ namespace ccf::js
       return JSWrappedValue(ctx, JS_GetProperty(ctx, val, prop));
     }
 
-    void set(const char* prop, const JSWrappedValue& value) const
+    int set(const char* prop, const JSWrappedValue& value) const
```
eddyashton, line 235: It looks like most of the uses of this want `std::move` to get the other overload. When do we use this `const&` instance?

achamayou: I don't know, but that's a good question to answer once I'm done making the FFI safe I think.

Another thing that I've put in a pin in for now is that the crypto code (ie. wrappers for OpenSSL) mostly (but not exclusively) throws std::runtime_error to indicate validation failures, and std::logic_error is used for more internal failures. It's important to distinguish them when we validate an x509 chain for example, with some fuzziness. If one of the values in the chain is not a valid PEM for example, is that False or exception (I think it's exception, but it's arguable).

achamayou: > When do we use this const& instance?

The answer was, nowhere legitimate, that's cleaned up now. Separately, I seem to have introduced a double free in historical queries, chasing that now...

src/js/wrap.h
```diff
@@ -100,6 +127,51 @@ namespace ccf::js
...
+  {
+  public:
+    JSWrappedAtom() : ctx(NULL), val(JS_ATOM_NULL) {}
+    JSWrappedAtom(JSContext* ctx, JSAtom&& val) : ctx(ctx), val(std::move(val))
+    {}
+    JSWrappedAtom(JSContext* ctx, const JSAtom& value) : ctx(ctx)
+    {
+      val = JS_DupAtom(ctx, value);
+    }
+    JSWrappedAtom(const JSWrappedAtom& other) : ctx(other.ctx)
```
eddyashton, line 140: Similar to `JSWrappedValue`, it would be good if we could remove this `const&` overload, so that we get compile-time errors if we fail to `move`/`take` an argument.

achamayou: Agreed!

eddyashton (approved):

PR #5751, opened 2023-10-20
Moving the TLA+ spec for consensus

heidihoward:
Moving the original TLA+ spec into its own sub directory under `/tla`

The docs also need to updating but I'll leave that for a separate PR


achamayou:
> The docs also need to updating but I'll leave that for a separate PR

It would be preferable for that to be in the same PR if possible, otherwise we'll be publishing documentation with broken links in the interval.

It won't break the CI, because we've turned off the automated link checker after it proved noisy (out of the various websites we link to, there often seemed to be one or two that was down).

heidihoward:
@achamayou I might be mistaken but I don't believe that there are any links to the relocated files in the docs. The only thing that would need updating due to this change is the examples here https://microsoft.github.io/CCF/main/architecture/raft_tla.html#running-the-model-checker but they are already out-of-date (e.g. the file `MCccfraftWithReconfig.tla` does not exist and `-workers auto` was recently removed from `tlc.sh`.

That being said I am happy to combine into one PR if that's preferred

achamayou:
@heidihoward you're right, we used to have `literalincludes` pointing there, but we have since removed them it seems, so I have no objection.

achamayou (approved):

lemmy:
Minor, but the wording in https://github.com/microsoft/CCF/blob/main/tla/README.md is outdate.

PR #5757, opened 2023-10-21
Allow consecutive signatures.

lemmy:
Ignore consecutive signatures when model checking.

@eddyashton Please confirm that the first log entry cannot be a signature:

https://github.com/microsoft/CCF/blob/ee4bb4566b6f891198f92e20f3a64b6162d040a6/tla/consensus/ccfraft.tla#L651

### DIFF
--- a/tla/consensus/MCccfraft.cfg
+++ b/tla/consensus/MCccfraft.cfg
@@ -10,6 +10,7 @@ CONSTANTS
     Timeout <- MCTimeout
     Send <- MCSend
     ClientRequest <- MCClientRequest
+    SignCommittableMessages <- MCSignCommittableMessages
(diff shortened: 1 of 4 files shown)


achamayou:
@lemmy the first transaction is never a signature, it initialises tables including the one that contains the configuration, which are necessary to verify future signature transactions. The Merkle Tree is empty at that point, a signature would be meaningless.

lemmy:
> @lemmy the first transaction is never a signature, it initialises tables including the one that contains the configuration, which are necessary to verify future signature transactions. The Merkle Tree is empty at that point, a signature would be meaningless.

Great! This PR, thus, requires no further changes.

achamayou (approved):

PR #5764, opened 2023-10-24
Add and enforce specification invariant `CommittableIndicesAreKnownSignaturesInv`

eddyashton:
Changes to the spec in #5749 are failing because we store non-signatures in `committableIndices`.

This adds a new invariant, stating that each node's `committableIndices` are within that node's log, and signatures, and updates the `NoConflictAppendEntriesRequest` action so that this invariant holds. Previously the action was appending _all_ new log entries to `committableIndices`, now it only appends the signatures.


achamayou (approved):
I'd be tempted to s/KnownSignatures/Signatures/ because it lightly implies the existence of unknown signatures?

Anyway, this is a good fix!

achamayou:
@eddyashton that does make sense!

achamayou (approved):

src/consensus/aft/test/driver.h
```diff
@@ -85,21 +85,21 @@ struct LoggingStubStoreSig_Mermaid : public aft::LoggingStubStoreSigConfig
...
                     << std::endl;
-    aft::LoggingStubStoreSigConfig::initialise_term(t);
+    aft::LoggingStubStoreConfig::initialise_term(t);
   }
 };
 
 using ms = std::chrono::milliseconds;
 using TRaft = aft::Aft<LedgerStubProxy_Mermaid>;
-using Store = LoggingStubStoreSig_Mermaid;
+using Store = LoggingStubStore_Mermaid;
```
eddyashton, line 102: All of the changes under `test/` are related to this, and aiming to resolve a Trace Validation failure. The driver was using this "everything is a signature" mock of the `kv::Store`, which meant that the backups viewed _all_ entries as committable (while the leader only committed those from `emit_signature`). The change is to move the committability (as encoded in an `ExecutionWrapper::result`, for _reasons_) from a compile-time template (I don't quite remember why that ever made sense) to embedding it in the serialised entry, as we do with the view and term. A related cleanup removes all of the now-unnecessary (and inaccurate) mock types with `Sig` in the name.

achamayou: I think we should make the payloads JSON next, so they can be easily accessed from the trace validation if desired. And also because why not.

eddyashton: Agreed - it looks like we already do this for the reconfiguration messages, so assume we can do it more generally.

PR #5668, opened 2023-09-20
Extend testing of JS runtime limits, take 2

achamayou:
Proposed successor to #5594, which was rolled back in #5671.

Spurred by test failure in https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=76095&view=results. Make the test more precise by checking for specific exceptions, and more robust by ramping up gradually with a high ceiling for stack overflow, rather than find pairs of values that work per architecture.

PR #5765, opened 2023-10-24
[release/4.x] Cherry pick: Only enable JS runtime limits during execution (#5730)

achamayou:
Backports the following commits to `release/4.x`:
 - [Only enable JS runtime limits during execution (#5730)](https://github.com/microsoft/CCF/pull/5730)


eddyashton:
Test failures are because this relies on changes from #5668, so looks like that needs to be backported too.

PR #5769, opened 2023-10-24
Add driver commands to start networks and reconfigure more realistically

achamayou:
The driver was never updated after we switched to [network bootstrapping](https://microsoft.github.io/CCF/main/operations/start_network.html), which makes expressing impossible scenarios too easy, and realistic scenarios more tedious than necessary.

This is a small change in the direction of more realism in the driver. Scenarios that cannot be expressed with bootstrapping,  or focus on pre-consensus nodes, have been deleted. I have also changed the Start node to use `force_become_leader()`, as in the real system, which makes the initial tx id line up at `2.1`.

Nodes are fully connected as soon as they are trusted, they can be disconnected manually where appropriate.

All the scenarios are up-to-date, however the trace validation fails. Issues found so far:

1. CCF reconfiguration results in a "replicate" followed by an "add_configuration", which produce two separate entries in the log, causing the indices to go out of sync. Worked around for by making `ChangeConfigurationInt` remove the previous log entry before appending the config. This breaks `PermittedLogChangesProp` in its current form however.
2. `AdvanceCommitIndex` does not match CCF behaviour, and creates more states than necessary as a side effect. Tracked in #5798, I have attempted to fix that and failed so far.
3. We prime new nodes with `(0 :> {})` configuration, which as far as I can tell prevents commit from making progress. I have attempted to allow startup with empty configurations, but not managed to fix everything it breaks, and instead added a workaround in `ChangeConfigurationInt` that drops the configuration at zero for now.

This is enough to get the most basic scenario to pass: starting one node like in real network, with a trace that matches that produced by an end to end test: https://github.com/microsoft/CCF/pull/5769/files#diff-c4e3eafe0741764366e596fc20f6c1063859a65227f3fcc9e43984aea0840333

The workaround applied in 3. bumps up against `LogConfigurationConsistentInv` [full trace](https://github.com/microsoft/CCF/actions/runs/6731403408/job/18295989608?pr=5769):

```
LogConfigurationConsistentInv ==
...
        \* Current configuration should be committed
        \* This is trivially true for the initial configuration (index 0)
        /\ commitIndex[i] >= CurrentConfigurationIndex(i)
```

It is now fairly easy to start and not have a committed configuration at all (in fact it was true before, the fabricated 0 index for the initial configuration was just hiding that). It's worth noting that the trace is a for an absurd scenario, where two n1 and n2 start, with an initial configuration {n1} (can't happen), and n1 proceeds to reconfigure to {n2} before it has even committed its configuration (completely impossible).

With that relaxed, and a further change to avoid [invalid sequences](https://github.com/microsoft/CCF/pull/5769/commits/748f0a0e5e85fb10af6accf18e543059f6ccb4ae) where the reconfiguration consumes a signature or another reconfiguration, there is now a problem with the MC reconfig, because it relies on there being a committed config at 0, and so QuorumInvLog breaks at the first reconfiguration, since that configuration is removed.


tests/raft_scenarios/startup
```diff
@@ -0,0 +1,7 @@
+start_node
```
achamayou, line 1: This is a minimal scenario, to get the trace validation to line up, but it fails to match on "commit", for reasons that are not yet clear. Another bug became obvious along the way (#5798), but is not the root cause here. I think it may be the presence of the initial empty config that's at fault.

achamayou: It was.

tla/consensus/ccfraft.tla
```diff
@@ -496,7 +496,7 @@ InitMessagesVars ==
     /\ commitsNotified = [i \in Servers |-> <<0,0>>] \* i.e., <<index, times of notification>>
 
 InitServerVars ==
-    /\ currentTerm = [i \in Servers |-> 0]
+    /\ currentTerm = [i \in Servers |-> 2]
     /\ state       = [i \in Servers |-> IF i \in configurations[i][0] THEN Follower ELSE Pending]
```
achamayou, line 500:
```suggestion
    /\ state       = [i \in Servers |-> IF i \in configurations[i][0] THEN Leader ELSE Pending]
```

tla/consensus/ccfraft.tla
```diff
@@ -815,22 +817,22 @@ RejectAppendEntriesRequest(i, j, m, logOk) ==
        \/ /\ m.term >= currentTerm[i]
           /\ state[i] = Follower
           /\ ~logOk
-          /\ LET prevTerm == IF m.prevLogIndex = 0 THEN 0
-                             ELSE IF m.prevLogIndex > Len(log[i]) THEN 0 ELSE log[i][m.prevLogIndex].term
+          /\ LET prevTerm == IF m.prevLogIndex = 0 THEN 2
+                             ELSE IF m.prevLogIndex > Len(log[i]) THEN 2 ELSE log[i][m.prevLogIndex].term
```
achamayou, line 821: That does not seem right, in the ELSE case, something like log[i][Len(log[i])].term.

tla/consensus/ccfraft.tla
```diff
@@ -815,22 +817,22 @@ RejectAppendEntriesRequest(i, j, m, logOk) ==
...
                              source         |-> i,
                              dest           |-> j],
                              m)
-                   \/ /\ prevTerm # 0
+                   \/ /\ prevTerm # 2
                       /\ LET lli == FindHighestPossibleMatch(log[i], m.prevLogIndex, m.term)
                          IN Reply([type        |-> AppendEntriesResponse,
                                 success        |-> FALSE,
-                                term           |-> IF lli = 0 THEN 0 ELSE log[i][lli].term,
+                                term           |-> IF lli = 0 THEN 2 ELSE log[i][lli].term,
```
achamayou, line 835: Check this in the implementation.

tla/consensus/ccfraft.tla
lemmy: Are you also going to change `InitReconfigurationVars` to something like the following to guarantee a single leader at startup?

```tla
InitReconfigurationVars ==
    /\ ...
    /\ \E s \in Servers:
        configurations = [i \in Servers |-> 0 :> {s} ]
     \* There is a single leader in the initial configuration, and a subset of the nodes may know about this config.
    /\ Cardinality(Range(configurations)) = 1
```

Also, remember that `MCccfraft` redefines `InitReconfigurationVars`.

tla/consensus/ccfraft.tla
lemmy: Introduced in https://github.com/microsoft/CCF/commit/5301f6739fd19c89cef582bff998c32d49095cd2

tla/consensus/ccfraft.tla
achamayou: I've changed this some more, with guidance from @heidihoward, but yes I will make the change everywhere. Longer term I think we will split the membership/status in different variables also.

tla/consensus/ccfraft.tla
achamayou: @heidihoward the implementation does in fact send the term of the highest possible on rejected append entries https://github.com/microsoft/CCF/blob/1d52533f70a94d015f3215acec97d629e11bfdd7/src/consensus/aft/raft.h#L1351, instead of the current term, and this used when deciding what to resend: https://github.com/microsoft/CCF/blob/1d52533f70a94d015f3215acec97d629e11bfdd7/src/consensus/aft/raft.h#L1473

heidihoward:
Looking at the results from our brand new tlc ci, I can see that `MCccfraft.cfg` has sped up with this PR which I am suspicious of. Reminder to myself @heidihoward to have a look at this tomorrow

tla/consensus/ccfraft.tla
achamayou: @lemmy we can't just initialise the configurations, that makes them inconsistent with the log. We also cannot initialise the configurations for other nodes that have not yet joined when they have no log. All configurations are discovered through the log.

PR #5778, opened 2023-10-25
Restrict initial configuration in ccfraft.tla to a single node

heidihoward:
This is a simple change which probably has some big implications for the TLA+ specs and the trace validation.

### DIFF
--- a/tla/consensus/ccfraft.tla
+++ b/tla/consensus/ccfraft.tla
@@ -504,7 +504,7 @@ InitReconfigurationVars ==
     \* such as the initial configuration. A node's configuration is *never* "empty", i.e., the equivalent of configuration[node] = {} here. 
     \* For simplicity, the set of servers/nodes all have the same initial configuration at startup.
     /\ \E s \in Servers:
-        configurations = [i \in Servers |-> [ j \in {0} |-> {s} ] ]
+        configurations = [i \in Servers |-> 0 :> {s} ]
 
 InitMessagesVars ==
     /\ Network!InitMessageVar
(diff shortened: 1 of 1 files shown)


heidihoward:
Linking https://github.com/microsoft/CCF/pull/5769

lemmy:
Trace validation redefines `InitReconfigurationVars` with `TraceInitReconfigurationVars`, which may initialize configuration to a *set of servers* if that's what's in the log.  For example, tests/raft_scenarios/election initializes the configuration of the first node `0` to `{0,1,2}`.

heidihoward:
My understanding from @achamayou is that the initial configuration is always a single server. https://github.com/microsoft/CCF/pull/5769 should help this bootstrapping phase to be represented in the raft driver

lemmy:
Should we change `TraceInitReconfigurationVars` accordingly in this PR?

achamayou (approved):

heidihoward:
> Should we change `TraceInitReconfigurationVars` accordingly in this PR?

Yes. I've taken a quick look at it (alongside the traces shared by @achamayou) and I'm not sure the best direction to take

achamayou (approved):

lemmy:
> > Should we change `TraceInitReconfigurationVars` accordingly in this PR?
> 
> Yes. I've taken a quick look at it (alongside the traces shared by @achamayou) and I'm not sure the best direction to take

Since Traceccfraft is in flux right now because of @achamayou's work on aligning the raft test driver, it might be a good idea to simply add a TODO in Traceccfraft to realign it with ccfraft eventually.

PR #5779, opened 2023-10-25
Changing index into `Configurations` in `MCccfraft.tla`

heidihoward:
I believe that there is currently an issue in `MCccfraft.tla` where the first configuration in `Configurations` is ignored. Currently, the initial configuration is any subset of servers and then the next configuration is the 2nd set of servers in `Configurations` and so on.

~~I'm not sure yet if this is the best fix (verse using a separate initial state in MCccfraft.tla).~~ I think this is simplest fix so the best for now

This PR is related to (but separate from) https://github.com/microsoft/CCF/pull/5778

### DIFF
--- a/tla/consensus/MCccfraft.cfg
+++ b/tla/consensus/MCccfraft.cfg
@@ -78,6 +78,7 @@ INVARIANTS
     MatchIndexLowerBoundNextIndexInv
     CommitCommittableIndices
     CommittableIndicesAreKnownSignaturesInv
+    \* DebugAllReconfigurationsReachableInv
     \* DebugInvLeaderCannotStepDown
     \* DebugInvAnyCommitted
     \* DebugInvAllCommitted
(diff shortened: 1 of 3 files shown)


achamayou (approved):

heidihoward:
This change which should slow down the exhaustive model checking, is making `MCccfraftWithReconfig.cfg` faster for reasons which are not obvious to me. https://github.com/microsoft/CCF/actions/runs/6641887323/job/18045392763#step:6:59

heidihoward:
Mystery solved, I needed to increase the `ViewLimit` to allow all three reconfigurations

lemmy:
Can we remove `reconfigurationCount` and constraint the number of reconfigurations with `MaxTermLimit` and leave picking the next config to `CHOOSE c \in SUBSET Servers : c \notin {previous configs inferred from committed log}`?

lemmy (approved):

heidihoward:
> Can we remove `reconfigurationCount` and constraint the number of reconfigurations with `MaxTermLimit` and leave picking the next config to `CHOOSE c \in SUBSET Servers : c \notin {previous configs inferred from committed log}`?

Yes, I believe we can. I'll try this refactor in a separate PR

Issue #5794, opened 2023-10-27
Towards trace validation of real CCF nodes


achamayou:
This is taking a step back from #5769, to document the existing gap between the raft driver, and actual nodes, with the goal of closing it.

As of https://github.com/microsoft/CCF/commit/af214078fd067f578d393665c451ae3f9af2401e, using `jq .msg -c` to remove the less important fields such as timestamps etc.

# A three node network startup, in the driver

```
{"committable_indices":[],"configurations":[],"function":"add_configuration","new_configuration":{"idx":0,"nodes":{"0":{"address":":"},"1":{"address":":"},"2":{"address":":"}},"rid":0},"state":{"commit_idx":0,"current_view":0,"last_idx":0,"leadership_state":"None","membership_state":"Active","new_view_idx":0,"node_id":"0"}}
{"committable_indices":[],"configurations":[],"function":"add_configuration","new_configuration":{"idx":0,"nodes":{"0":{"address":":"},"1":{"address":":"},"2":{"address":":"}},"rid":0},"state":{"commit_idx":0,"current_view":0,"last_idx":0,"leadership_state":"None","membership_state":"Active","new_view_idx":0,"node_id":"1"}}
{"committable_indices":[],"configurations":[],"function":"add_configuration","new_configuration":{"idx":0,"nodes":{"0":{"address":":"},"1":{"address":":"},"2":{"address":":"}},"rid":0},"state":{"commit_idx":0,"current_view":0,"last_idx":0,"leadership_state":"None","membership_state":"Active","new_view_idx":0,"node_id":"2"}}
{"committable_indices":[],"configurations":[{"idx":0,"nodes":{"0":{"address":":"},"1":{"address":":"},"2":{"address":":"}},"rid":0}],"function":"become_candidate","state":{"commit_idx":0,"current_view":1,"last_idx":0,"leadership_state":"Candidate","membership_state":"Active","new_view_idx":0,"node_id":"0"}}
{"committable_indices":[],"function":"send_request_vote","packet":{"last_committable_idx":0,"msg":"raft_request_vote","term":1,"term_of_last_committable_idx":0},"state":{"commit_idx":0,"current_view":1,"last_idx":0,"leadership_state":"Candidate","membership_state":"Active","new_view_idx":0,"node_id":"0"},"to_node_id":"2"}
[...]

PR #5796, opened 2023-10-30
Trace state and configuration after action

achamayou:
I am putting this up as a draft, based on #5795, this is what I discussed in #5794, but no longer think it's a good idea after doing it.

That's because to get a (meaningful) `state_after` or `configuration_after`, we virtually always have to wait until some other actions have been emitted, and so we get a log line that's out of order.

I suspect a better solution would be to split the functions in two ("add_configuration_enter", "add_configuration_exit"), but the trace spec would need significant adjustment.

For now I think we're better off doing ad-hoc logging of post-values on a need basis (I think only committable is missing at the moment).

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -18,6 +18,10 @@
 #include "raft_types.h"
 #include "service/tables/signatures.h"
 
+#ifdef CCF_RAFT_TRACING
+#  include "trace.h"
+#endif
+
 #include <algorithm>
 #include <list>
 #include <random>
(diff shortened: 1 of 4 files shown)


src/consensus/aft/raft.h
```diff
@@ -456,13 +454,8 @@ namespace aft
...
 #ifdef CCF_RAFT_TRACING
-      nlohmann::json j = {};
-      j["function"] = "add_configuration";
-      j["state"] = *state;
-      j["configurations"] = configurations;
-      j["new_configuration"] = Configuration{idx, conf, idx};
-      j["committable_indices"] = committable_indices;
-      RAFT_TRACE_JSON_OUT(j);
+      trace::Line line("add_configuration", state, configurations);
+      line.j["new_configuration"] = Configuration{idx, conf, idx};
```
achamayou, line 458: This will need a local block as soon as we do scenarios with retirement, because of line 471.

achamayou:
Closing as discussed with @heidihoward @eddyashton and @lemmy.

PR #5801, opened 2023-10-31
Bringing back the option to model check without reconfiguration

heidihoward:
This PR adds an alternative init state to `MCccfraft.tla` which allows you to exhaustively model check without using the reconfiguration logic (which is handy for state machine explosion)


tla/consensus/MCccfraft.tla
```diff
@@ -90,7 +93,25 @@ MCNotifyCommit(i,j) ==
...
+    /\ CCF!InitMessagesVars
+    /\ CCF!InitServerVars
+    /\ CCF!InitCandidateVars
+    /\ CCF!InitLeaderVars
+    /\ CCF!InitLogVars
+
+\* Alternative to CCF!Spec that uses the above MCInit
+mc_spec ==   
+    /\ MCInit
+    /\ [][Next]_vars
```
heidihoward, line 114: I tried to use `CCF!Next` instead of `Next` here but it broke the coverage checking, vscode plug-in reported only `MCInit`.

lemmy: Could be a bug in TLC.  How did `CCF!Next` break coverage checking?

heidihoward: using `Next`
image
using `CCF!Next`
image

lemmy: Can you please also post the content of the .out file, i.e. what's behind "Full Output"?

lemmy (approved):

eddyashton (approved):

PR #5803, opened 2023-10-31
Do not enforce default parsing limits on forwarded traffic

achamayou:
The parser that is used to handle requests forwarded by the backups to the primary uses [default parsing limits](https://github.com/microsoft/CCF/blob/6e0ba3a88c93911ce9129dec30f41b5ea75e56c7/include/ccf/http_configuration.h#L12) (ie. 1Mb max body size for example).

This means that even if individual rpc interfaces raise parsing caps, the lower, default caps are still applied on forwarded traffic. Because the caps are already applied by the backup parsing the initial request, it is not necessary to repeat them on the primary side when receiving forwarded traffic, so this PR raises the limits to more permissive values in consequence.

| Value| Default| Permissive|
|--------|--------|--------|
| Max Body Size| 1 MB | 1GB |
| Max Header Size | 16 KB | 100MB |
| Max Header Count | 256 | 1024 |

The limit is not completely removed, which may result in a user-specified value that's higher and causes forwarding to fail. That is preferable to crashing the primary however.


include/ccf/http_configuration.h
```diff
@@ -43,4 +43,9 @@ namespace http
     max_concurrent_streams_count,
     initial_window_size,
     max_frame_size);
+
+  static ParserConfiguration permissive_configuration()
+  {
+    return ParserConfiguration{{"1GB"}, {"100MB"}, 1024, 1, {"64KB"}, {"16MB"}};
```
eddyashton, line 49: Minor - can we write this as:
```cpp
ParserConfiguration conf;
conf.max_body_size = "1GB";
...
return conf;
```
To make it clear what each of these values is?

eddyashton, line 49: Where do these numbers come from? What if a node sets its limits even higher than this?

achamayou: They're what I think we can tolerate at best with the current enclave size without crashing the primary. I think we can re-evaluate those if we raise that, but I don't think it makes sense to raise the limits up to the point where we're confident we're going to get a crash.

achamayou:
Unfortunately, the debug builds exceed the election timeout before they can fully process the large transaction (logging the payload repeatedly does not seem to help). I'll look at either turning down the amount of logging, or increasing the timeout.

PR #5807, opened 2023-11-01
Adding TLC stats to cimetrics

heidihoward:
This PR adds some TLC stats to the [cimetrics](https://github.com/jumaffre/cimetrics). To keep things simple, this PR just adds the duration and number of distinct states for `MCccfraft.tla` but we can easily extended this.


heidihoward:
@eddyashton @achamayou Still a WIP but I'd appreciate feedback at this stage. I've deliberately omitted `METRICS_MONGO_CONNECTION` from the gh action to avoid actually posting data until we are ready

tests/upload_tlc_stats.py
```diff
@@ -0,0 +1,40 @@
...
+        with open(filename) as f:
+            data = json.load(f)
+            duration_sec = data["duration"]
+            dstates = data["distinct"]
+            LOG.info(
+                "Uploading metrics - duration: {}, distinct states: {}",
+                duration_sec,
+                dstates,
+            )
+            with cimetrics.upload.metrics(complete=False) as metrics:
```
lemmy, line 23: I don't know what cimetrics provides but https://github.com/tlaplus/Examples/blob/62d104f2e1cfd63b3f928342b8f634298706d145/specifications/ewd998/EWD998ChanID_export.tla#L46-L56 is an example about how you can do post requests from within TLC, which would make the json serialization obsolete.

heidihoward: This is cimetric https://github.com/jumaffre/cimetrics, it provides the fancy performance graphs in each PR. 

I think the Json stats file might be useful in the future to quickly compare the performance of different configurations

lemmy: If more is going on in cimetrics than just a basic http post, the .py script is the way to go. :-)

Btw. I do have the CosmosDB client library wrapped as a TLA+ module somewhere, but that's just an interesting hack.

achamayou: Upload is pretty straightforward: https://github.com/jumaffre/cimetrics/blob/main/cimetrics/upload.py, most of the logic is there to the querying and the fancy plotting (which you can see in this PR too if you scroll up a bit).

For procedural tasks such as this one, Python has a certain charm :)

achamayou (approved):

achamayou:
@heidihoward if you would like a different category, you can add a line in https://github.com/microsoft/CCF/blob/main/metrics.yml, something like "TLC": "tlc_.*", before the "Others" catch-all (they are evaluated in order).

heidihoward:
> @heidihoward if you would like a different category, you can add a line in https://github.com/microsoft/CCF/blob/main/metrics.yml, something like "TLC": "tlc_.*", before the "Others" catch-all (they are evaluated in order).

done in https://github.com/microsoft/CCF/pull/5807/commits/e1e3cc4acd46d510acc432c2efd4507b89bde4ff

.azure-pipelines-templates/model_checking.yml
```diff
@@ -23,13 +23,21 @@ jobs:
...
 
-      - script: ./tlc.sh -workers auto -config consensus/MCccfraftWithReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftWithReconfig.trace.tla -dumpTrace json MCccfraftWithReconfig.json
-        workingDirectory: tla
-        displayName: MCccfraftWithReconfig.cfg
+      # - script: ./tlc.sh -workers auto -config consensus/MCccfraftWithReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftWithReconfig.trace.tla -dumpTrace json MCccfraftWithReconfig.json
+      #   workingDirectory: tla
+      #   displayName: MCccfraftWithReconfig.cfg
+
+      - script: |
+          pip install -r requirements.txt
```
achamayou, line 35:
```suggestion
          set -ex
          python3.8 -m venv env
          source env/bin/activate
          python -m pip install -r requirements.txt
```

.azure-pipelines-templates/model_checking.yml
```diff
@@ -23,13 +23,24 @@ jobs:
...
+      # - script: ./tlc.sh -workers auto -config consensus/MCccfraftWithReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftWithReconfig.trace.tla -dumpTrace json MCccfraftWithReconfig.json
+      #   workingDirectory: tla
+      #   displayName: MCccfraftWithReconfig.cfg
+
+      - script: |
+          set -ex
+          python3.8 -m venv env
+          source env/bin/activate
+          python -m pip install -r requirements.txt
+          python3 upload_tlc_stats.py 3node_fixed ../tla/MCccfraft.json
```
achamayou, line 39:
```suggestion
          ls ../tla/*.json
          python3 upload_tlc_stats.py 3node_fixed ../tla/MCccfraft_stats.json
```

heidihoward:
Yay, this is now working! 
image

achamayou (approved):

heidihoward:
Looking at the CI output wheel is complaining a bit but it seems to work anyways. I think actually it might be best to switch to [cimetrics_env.sh](https://github.com/microsoft/CCF/blob/main/tests/cimetrics_env.sh) so only cimetrics is installed. I did spot this handy script the first time around but did not use it as I wrongly assumed we didn't need a virtual env.

PS: Sorry, I don't know how to post a link to ADO pipelines, copying and pasting the URL gives this fun error 
image

achamayou:
@heidihoward you're absolutely right, I'd completely forgotten about that script.

PR #5810, opened 2023-11-02
Move model checking to main CI pipeline

achamayou:
Complement to #5807 to get cimetrics to work (GHA isn't supported).


heidihoward (approved):
shall we remove consensus model checking from ci actions to its only ran once?

PR #5820, opened 2023-11-08
Add realistic reconfiguration commands for driver

achamayou:
The driver was never updated after we switched to [network bootstrapping](https://microsoft.github.io/CCF/main/operations/start_network.html), which makes expressing impossible scenarios too easy, and realistic scenarios more tedious than necessary.

This is a small change in the direction of more realism in the driver. I have also changed the Start node to use force_become_leader(), as in the real system, which makes the initial tx id line up at 2.1.

Nodes are fully connected as soon as they are trusted, they can be disconnected manually where appropriate.

This is a chunk of #5769, enough to express realistic scenarios, but without converting the existing ones and requiring the cascading TLA+ changes just yet.

### DIFF
--- a/src/consensus/aft/test/driver.cpp
+++ b/src/consensus/aft/test/driver.cpp
@@ -82,6 +82,20 @@ int main(int argc, char** argv)
     }
     switch (shash(in))
     {
+      case shash("start_node"):
+        assert(items.size() == 1);
+        driver->create_start_node(lineno);
+        break;
+      case shash("trust_node"):
+        assert(items.size() == 3);
+        driver->trust_nodes(items[1], {items[2]}, lineno);
+        break;
+      case shash("trust_nodes"):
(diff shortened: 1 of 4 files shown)


heidihoward (approved):

lemmy:
> Scenarios that cannot be expressed with bootstrapping, or focus on pre-consensus nodes, have been deleted.

No scenarios have been deleted in this PR.  Is deletion part of #5769?

achamayou:
> > Scenarios that cannot be expressed with bootstrapping, or focus on pre-consensus nodes, have been deleted.
> 
> No scenarios have been deleted in this PR. Is deletion part of #5769?

Yes, this is an over-enthusiastic copy and paste, fixed now. Nothing is deleted, I'm trying to get this in gradually without breaks.

PR #5823, opened 2023-11-09
Align leadership state naming with implementation

achamayou:
A small change, extracted from #5769, but useful for trace validation with the realistic startup.


tla/consensus/ccfraft.tla
```diff
@@ -1356,7 +1354,8 @@ StateTransitionsProp ==
 
 PendingBecomesFollowerProp ==
     \* A pending node that becomes part of any configuration immediately transitions to Follower.
-    [][\A s \in { s \in Servers : state[s] = Pending } : 
+    \* TODO: this is not right, the transition only takes place once the node receives a message
```
heidihoward, line 1357: I think this is fine as the `configurations` still only include the node after it has received at AE with its own addition

heidihoward (approved):

tla/consensus/ccfraft.tla
achamayou: Ok, yes, agreed. I will rephrase the "become", I think it's "becomes aware that it is part of any configuration".

tla/consensus/ccfraft.tla
```diff
@@ -1356,7 +1354,8 @@ StateTransitionsProp ==
 
 PendingBecomesFollowerProp ==
     \* A pending node that becomes part of any configuration immediately transitions to Follower.
```
achamayou, line 1356:
```suggestion
    \* A pending node that becomes aware it is part of a configuration immediately transitions to Follower.
```

tla/consensus/ccfraft.tla
achamayou, line 1357:
```suggestion
```

PR #5824, opened 2023-11-09
Alias ccf.digest and ccf.isValidX509CertBundle to faciliate constitution migration from 3.x

achamayou:
`ccf.crypto.digest` and `ccf.crypto.isValidX509CertBundle` are accessible again through `ccf.digest` and `ccf.isValidX509CertBundle`, to faciliate migration of CCF 3.x-era constitutions. The constitution sample under `samples/constitutions/default/actions.js` has been updated to show how migration is possible without relying on aliasing in 4.x.


msftsettiy (approved):

lynshi (approved):

samples/constitutions/default/actions.js
```diff
@@ -144,7 +144,13 @@ function checkJwks(value, field) {
 }
 
 function checkX509CertBundle(value, field) {
-  if (!ccf.crypto.isValidX509CertBundle(value)) {
+  let check_func = ccf.crypto.isValidX509CertBundle;
+  // Remove after upgrade to 4.x
+  if (check_func === undefined) {
+    check_func = ccf.isValidX509Cert;
```
lynshi, line 150: Ah, I missed this during the review! Should be `ccf.isValidX509CertBundle`

achamayou: Doh, yes. This worked in testing because every usage we make of this, including the CA cert bundle, only ever passes one cert. Will fix though.

PR #5827, opened 2023-11-10
Move simulation to ADO too, and shorten the length of its run to not delay PRs

achamayou:
Re-group raft simulation with the model checking on ADO, and shorten run, 50 minutes it too long for the current CI.


heidihoward:
@achamayou  I cannot see any edits to shorten the runs

.azure-pipelines-templates/simulation.yml
```diff
@@ -0,0 +1,30 @@
...
+        fetchDepth: 1
+
+      - script: |
+          set -ex
+          sudo apt update
+          sudo apt install -y default-jre
+          python3 ./tla/install_deps.py
+        displayName: Setup
+
+      - script: ./tlc.sh -workers auto -simulate -depth 100 consensus/SIMccfraft.tla -dumpTrace tla SIMccfraft.trace.tla -dumpTrace json SIMccfraft.json
```
achamayou, line 22: @heidihoward it's hidden by the move but 500 has turned into 100.

heidihoward: I think it will still run for 50 mins, it will just generate more traces as they will be shorter

heidihoward: The 50min bound is set here https://github.com/microsoft/CCF/blob/56ee39ea04b5d34592cb16080a5ad537c946724e/tla/consensus/SIMccfraft.tla#L48

achamayou: Ah, thank you!

achamayou:
> @achamayou I cannot see any edits to shorten the runs

Thank you for spotting that this hadn't been done right!

lemmy:
I do not recommend shortening the traces, i.e., changing the `-depth` parameter beyond 250.  From observations, simulation has found issues such as violations of `QuorumLogInv` because it explores the state space where the system has seen "more" activity such as several client requests, more than a few votes, messages being lost, ...

achamayou:
> I do not recommend shortening the traces, i.e., changing the `-depth` parameter beyond 250. From observations, simulation has found issues such as violations of `QuorumLogInv` because it explores the state space where the system has seen "more" activity such as several client requests, more than a few votes, messages being lost, ... (once simulation found a long counterexample, human insight can help shorten it).

They're back to where they were (https://github.com/microsoft/CCF/pull/5827/files#diff-ebe20cacfdab2c7f426ccfec227825fae1a34bb8f01fb8248ba83bfb5319c6dcR22), I just misunderstood what the parameter did.

> Perhaps, spec simulation should be completely decoupled from our PR builder, or, if shortened, a new long-running (hours?) simulation should run every night.

The Daily Build does that (ASAN etc), so that's where we would put it. What's the best way to parametrise timeout out of the file?

lemmy:
> > Perhaps, spec simulation should be completely decoupled from our PR builder, or, if shortened, a new long-running (hours?) simulation should run every night.
> 
> The Daily Build does that (ASAN etc), so that's where we would put it. What's the best way to parametrise timeout out of the file?

```tla
TLCSet("exit", TLCGet("duration") > atoi(IOEnv.YOUR_TIMEOUT))
```

Issue #5822, opened 2023-11-09
CCF consensus verification roadmap


achamayou:
This issue aims to create a priority-ordered list of items that need to be tackled to complete the verification of CCF Raft, and summarise their dependencies so they can be picked up efficiently.

- [x] `ccfraft.tla` startup is incorrect, and needs to be aligned with the implementation. This our top priority item, it's documented at length in #5220, #5749, and there's an ongoing effort in #5769 to fix it. We should not be adding any more invariants, properties etc until this has been addressed.
- [x] `ccfraft.tla` commit is incorrect, and needs to be aligned with the implementation. This is documented in #5798. This does not depend on the first item, and can be picked up independently.
- [x] Check, and eliminate `NotifiyCommit` messages, documented in #5280. Those do not exist in the implementation, will not trace validate.
- [x] Split `leadership_state` from `membership_state`. #5902 They are confused in the spec (`RetiredLeader`). Prerequisite to:
- [x] Complete retirement modeling, see #4818.

#2577, #5678 and #5290 are lower priority than all of the above, and some of them may reconsidered altogether once these are complete.

achamayou:
This roadmap has run its course, we can look at the next steps in the larger-scoped #5057.

Issue #5290, opened 2023-05-19
Resume trace validation from trace file suffix


lemmy:
@eddyashton suggests to work around the current disparity between the TLA+ spec and its implementation in the area of bootstrapping the initial configurations by restricting trace validation to a known-good suffix of the log file.

To implement this workaround, we need to modify the `Traceccfraft.tla` file's next-state relation. The modified relation should match the behavior of ccfraft's `Next` formula for the specified number of lines that should be omitted from the log file.  More precisely, Traceccfraft's `TraceSpec` definition will look something like:

```tla

TraceSpec ==
    TraceInit /\ [][ 
                       IF l < 42
                       THEN l' = l + 1 /\ Next
                       ELSE TraceNext
                    ]_<<l, vars>>
```

Clearly, this approach will quickly become infeasible because of state-space explosion.  To mitigate, the log file could be partially matched while l < 42.  However, this might end up being more work than modeling configurations according to `raft.h`.

achamayou:
This is more or less what [StartState](https://github.com/microsoft/CCF/pull/5828/files#diff-f72a34d202ec743abd93cf9ebdfa3b2d5ef7bba962620e7015972ff8f445112bR459) and [JoinState](https://github.com/microsoft/CCF/pull/5828/files#diff-f72a34d202ec743abd93cf9ebdfa3b2d5ef7bba962620e7015972ff8f445112bR474) do.

I doubt there is a way to generalise that unless the spec is substantially modified to drop older log entries. Deciding the initial state of the log will inevitably require a scan of all the trace lines so far.

If the spec compacted its log like the implementation does, it may be possible to backtrack from a starting point, until enough entries have been found to fill in the gap between log index and commit index for each node in active configs.

achamayou:
We've lined up the spec and the implementation now, via the bootstrap `command` in the driver.

PR #5828, opened 2023-11-10
Align initial spec state with implementation

achamayou:
Pulling out just the changes to ccfraft.tla and MC/Sim from #5769:

1. start with one node leadership_status="Leader" at term 2, other nodes are leadership_status="None"
2. log that starts with a reconfiguration and a signature
3. commit index at 2

Minimal adjustment to the trace validation to start from this too, modeled as a "bootstrap" log entry, containing the initial state.

There is more work to do to fix the rest of the scenarios, in particular around the prev log and index calculations, but I am leaving them for another PR.


achamayou:
Interestingly, this causes deadlocks in the simulation, which (on a small sample of three), seem to look like this:

```
/\ log = ( n1 :>
      << [term |-> 2, contentType |-> Reconfiguration, configuration |-> {n2}],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Signature],
         [ term |-> 2,
           contentType |-> Reconfiguration,
           configuration |-> {n1} ] >> @@
  n2 :>
      << [term |-> 2, contentType |-> Reconfiguration, configuration |-> {n2}],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Reconfiguration, configuration |-> {n1}],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Entry, request |-> 42],
         [term |-> 2, contentType |-> Entry, request |-> 42],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Entry, request |-> 42],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Entry, request |-> 42],
         [term |-> 2, contentType |-> Entry, request |-> 42],
         [term |-> 2, contentType |-> Signature],
         [term |-> 2, contentType |-> Signature] >> @@
```

There seems to be two problems here, one is that for some reason n2 is not sending append entries to n1 about the reconfiguration. The other is that n2 continues emit entries even after producing a post-reconfiguration signature (retirement committable index), which the implementation would not do.

tla/consensus/Traceccfraft.tla
```diff
@@ -58,13 +58,13 @@ IsAppendEntriesResponse(msg, dst, src, logline) ==
...
-    IF "JSON" \in DOMAIN IOEnv THEN IOEnv.JSON ELSE "traces/election.ndjson"
+    IF "JSON" \in DOMAIN IOEnv THEN IOEnv.JSON ELSE "../../build/startup.ndjson"
 
 JsonLog ==
     \* Deserialize the System log as a sequence of records from the log file.
-     \* Run TLC with (assuming a suitable "tlc" shell alias):
-     \* $ JSON=../tests/raft_scenarios/4582.ndjson tlc -note Traceccfraft
-     \* Fall back to trace.ndjson if the JSON environment variable is not set.
+    \* Run TLC from under the tla/ directory with:
+    \* $ JSON=../build/startup.ndjson ./tlc.sh consensus/Traceccfraft.tla
```
lemmy, line 66: `../build/startup.ndjson` is inconsistent with `../../build/startup.ndjson` above.

achamayou: That's because one is run from the `tla/` directory, where the `tlc.sh` script `lives`, and the other is provided for the debugger, which seems to use the location of the spec file as the current working directory when it runs.

I agree it's not ideal, I'd prefer to remove the hardcoded path altogether and have a way to pass it to debugger, but I've not found a way to do that.

achamayou: @lemmy would TLC consider adding a way to pass CLI options to the code, maybe with a pass-through like [ansible's --extra-vars](https://www.redhat.com/sysadmin/extra-variables-ansible-playbook)?

lemmy: `IOUtils` has operators to read environment variables.  Perhaps, we should add similar operators to expose Java system properties.

tla/consensus/MCccfraft.tla
```diff
@@ -92,20 +92,12 @@ MCSend(msg) ==
...
-\* Alternative to CCF!InitReconfigurationVars that sets the initial configuration to the first config defined in Configurations
-MCInitReconfigurationVars ==
-    /\ reconfigurationCount = 0
-    /\ removedFromConfiguration = {}
-    /\ configurations = [i \in Servers |-> 0 :> Configurations[1]]
-
 \* Alternative to CCF!Init that uses the above MCInitReconfigurationVars
 MCInit ==
-    /\ MCInitReconfigurationVars
+    /\ CCF!InitReconfigurationVars
```
heidihoward, line 97: `MCccfraft` will not reach the first configuration with this approach. I think MC needs a separate init

tla/consensus/MCccfraftAtomicReconfig.cfg
```diff
@@ -75,8 +75,8 @@ INVARIANTS
     TypeInv
     MonoTermInv
     MonoLogInv
-    LogConfigurationConsistentInv
-    NoLeaderInTermZeroInv
+    \* LogConfigurationConsistentInv
```
heidihoward, line 78: I think just remove if not wanted instead of commenting out. It should be case that `LogConfigurationConsistentInv` can be fixed if we want to keep it

tla/consensus/ccfraft.tla
```diff
@@ -462,22 +461,21 @@ PlausibleSucessorNodes(i) ==
...
     /\ reconfigurationCount = 0
     /\ removedFromConfiguration = {}
-    \* Note that CCF has a bootstrapping procedure to start a new network and to join new nodes to the network (see 
-    \* https://microsoft.github.io/CCF/main/operations/start_network.html). In both cases, a node has the current (see 
-    \* https://microsoft.github.io/CCF/main/operations/ledger_snapshot.html#join-or-recover-from-snapshot) or some stale configuration
-    \* such as the initial configuration. A node's configuration is *never* "empty", i.e., the equivalent of configuration[node] = {} here. 
-    \* For simplicity, the set of servers/nodes all have the same initial configuration at startup.
-    /\ \E s \in Servers:
-        configurations = [i \in Servers |-> 0 :> {s} ]
+    /\ \E startNode \in Servers:
```
heidihoward, line 464: A helper function might be useful here, one which returns an init state given the initial configuration. This can be reused in the tracing and MC

tla/consensus/ccfraft.tla
```diff
@@ -773,22 +761,22 @@ RejectAppendEntriesRequest(i, j, m, logOk) ==
        \/ /\ m.term >= currentTerm[i]
           /\ state[i] = Follower
           /\ ~logOk
-          /\ LET prevTerm == IF m.prevLogIndex = 0 THEN 0
-                             ELSE IF m.prevLogIndex > Len(log[i]) THEN 0 ELSE log[i][m.prevLogIndex].term
+          /\ LET prevTerm == IF m.prevLogIndex = 0 THEN 2
+                             ELSE IF m.prevLogIndex > Len(log[i]) THEN 0 ELSE log[i][Len(log[i])].term
              IN /\ m.prevLogTerm # prevTerm
-                /\ \/ /\ prevTerm = 0
+                /\ \/ /\ prevTerm = 2
```
heidihoward, line 767: I'm still not sure about this. Using term 2 as the nil value worries me as term 2 is also a valid term.

heidihoward (approved):

tla/consensus/ccfraft.tla
achamayou: Happy to look at changing that, but I suggest that's a spec & implementation at the same time change ideally.

tla/consensus/MCccfraft.tla
achamayou: I've tried the change below, but this seems to cause the MC to blow past even the already higher number of distinct states (~8m) this PR generates compared to baseline (~2m), and it's still going.

```
1Configuration == <<{NodeOne, NodeTwo, NodeThree}>>
2Configurations == <<{NodeOne, NodeTwo}, {NodeTwo, NodeThree}>>
3Configurations == <<{NodeOne, NodeTwo}, {NodeTwo, NodeThree}, {NodeTwo}>>

[..,]

MCChangeConfigurationInt(i, newConfiguration) ==
    /\ reconfigurationCount < Len(Configurations)
    \* reconfigurationCount starts at 0, +1 to get the next configuration
    /\ newConfiguration = Configurations[reconfigurationCount+1]
    /\ CCF!ChangeConfigurationInt(i, newConfiguration)
```

tla/consensus/ccfraft.tla
achamayou: https://github.com/microsoft/CCF/pull/5828/commits/4c755e1775d56e5ba336ba99421557967113147a

tla/consensus/MCccfraft.tla
achamayou: Restricting configurations to two nodes seems to help considerably.

lemmy: TLC exposes an "inspection hatch" in its JMX management interface with which you can observe the currently explored states.   That might give clues as to what is going on, i.e., on what dimension the state space is exploding.  I use VisualVM as a JMX frontend.

tla/consensus/Traceccfraft.tla
```diff
@@ -58,13 +53,13 @@ IsAppendEntriesResponse(msg, dst, src, logline) ==
...
+    \* Traces can be generates with: ./make_traces.sh, also under the tla/ directory.
```
achamayou, line 62:
```suggestion
    \* Traces can be generated with: ./make_traces.sh, also under the tla/ directory.
```

tla/consensus/MCccfraft.tla
```diff
@@ -92,20 +94,16 @@ MCSend(msg) ==
...
 MCInit ==
-    /\ MCInitReconfigurationVars
     /\ CCF!InitMessagesVars
-    /\ CCF!InitServerVars
     /\ CCF!InitCandidateVars
     /\ CCF!InitLeaderVars
-    /\ CCF!InitLogVars
+    /\ IF Cardinality(Configurations[1]) = 1
+       \* If the first config is just one node, we can start with a two-tx log and a single config
+       THEN CCF!StartState(CHOOSE s \in Configurations[1]: TRUE, ToServers)
```
heidihoward, line 104:
```suggestion
       THEN CCF!StartState(Configurations[1], ToServers)
```

tla/consensus/ccfraft.tla
```diff
@@ -634,11 +647,8 @@ ChangeConfigurationInt(i, newConfiguration) ==
     /\ newConfiguration /= {}
     \* Configuration is a proper subset of the Servers
     /\ newConfiguration \subseteq Servers
-    \* Configuration is not equal to the previous configuration
-    /\ newConfiguration /= MaxConfiguration(i)
```
heidihoward, line 638: I think this line can stay

heidihoward (approved):

tla/consensus/MCccfraft.cfg
```diff
@@ -75,8 +75,6 @@ INVARIANTS
     TypeInv
     MonoTermInv
     MonoLogInv
-    LogConfigurationConsistentInv
-    NoLeaderInTermZeroInv
```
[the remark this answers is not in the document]
achamayou: Sure, but we don't have any nodes in term zero anymore, so it's not very interesting. Maybe a better invariant would be `OneLeaderOnlyInTermTwo` or `OneLeaderOnlyInInitialTerm`?

tla/consensus/ccfraft.tla
```diff
@@ -634,11 +647,8 @@ ChangeConfigurationInt(i, newConfiguration) ==
...
     \* Keep track of running reconfigurations to limit state space
     /\ reconfigurationCount' = reconfigurationCount + 1
-    /\ removedFromConfiguration' = removedFromConfiguration \cup (CurrentConfiguration(i) \ newConfiguration)
```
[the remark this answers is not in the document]
achamayou: Ok for now, but as discussed it wonder if it would be best to compute this from the logs when needed if at all, to avoid making the state worse.

tla/consensus/ccfraft.tla
```diff
@@ -136,7 +135,6 @@ reconfigurationVars == <<
 ReconfigurationVarsTypeInv ==
     /\ ReconfigurationCountTypeInv
     /\ ConfigurationsTypeInv
-    /\ RemovedFromConfigurationTypeInv
```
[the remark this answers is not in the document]
achamayou: That's because I had completely removed removedFromConfiguration earlier. I think it's something to look at separately (comment below).

tla/consensus/MCccfraft.cfg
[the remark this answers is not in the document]
achamayou: But that's already covered by `MoreThanOneLeaderInv`, so no.

achamayou:
CI failure unrelated to the changes, filed as #5837

tla/consensus/MCccfraft.cfg
```diff
@@ -75,8 +75,6 @@ INVARIANTS
     TypeInv
     MonoTermInv
     MonoLogInv
-    LogConfigurationConsistentInv
```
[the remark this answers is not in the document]
achamayou: @heidihoward you were very right about this, because I had an [off-by-one error](https://github.com/microsoft/CCF/pull/5828/files#diff-f72a34d202ec743abd93cf9ebdfa3b2d5ef7bba962620e7015972ff8f445112bR791) in `ChangeConfigurationInt` that it did catch.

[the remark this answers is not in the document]
achamayou: Plus an unnecessary config in the joined initial state https://github.com/microsoft/CCF/pull/5828/commits/bb454656728aec3df031fba4882e524827c3d8b5.

tla/consensus/MCccfraft.cfg
```diff
@@ -4,7 +4,7 @@ CONSTANTS
     Configurations <- 1Configuration
     Servers <- ToServers
 
-    MaxTermLimit = 1
```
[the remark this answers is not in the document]
achamayou: It's fine when taking into account the election in the old initial state.

tla/consensus/ccfraft.tla
```diff
@@ -1145,9 +1160,10 @@ LogMatchingInv ==
 \* of at least one server in every quorum
 QuorumLogInv ==
     \A i \in Servers :
-        \A S \in Quorums[CurrentConfiguration(i)] :
-            \E j \in S :
-                IsPrefix(Committed(i), log[j])
+        \/ configurations[i] = << >>
```
lemmy, line 1163: This relies on TLC's left-to-right evaluation order.  How about:

```tla
configurations[i] # << >>  => \A S \in Quorums[CurrentConfiguration(i)] : ...
```

tla/consensus/ccfraft.tla
```diff
@@ -456,28 +456,49 @@ PlausibleSucessorNodes(i) ==
         highestMatchServers == {n \in activeServers : \A m \in activeServers : matchIndex[i][n] >= matchIndex[i][m]}
     IN {n \in highestMatchServers : \A m \in highestMatchServers: HighestConfigurationWithNode(i, n) >= HighestConfigurationWithNode(i, m)}
 
+\* Generate initial state for a given startNode
+StartState(startNode, servers) ==
+    /\ reconfigurationCount = 0
+    /\ removedFromConfiguration = {}
+    /\ configurations = [ i \in servers |-> IF i = startNode THEN (1 :> {startNode}) ELSE << >>]
+    /\ currentTerm = [i \in servers |-> IF i = startNode THEN 2 ELSE 0]
```
lemmy, line 464: Consider extracting `2` into a constant throughout the spec?

tla/consensus/ccfraft.tla
```diff
@@ -456,28 +456,49 @@ PlausibleSucessorNodes(i) ==
...
+        /\ votedFor    = [i \in servers |-> Nil]
+        /\ log         = [i \in servers |-> IF i \in startNodes
+                                            THEN << [term |-> 2, contentType |-> TypeReconfiguration, configuration |-> {startNode}],
+                                                    [term |-> 2, contentType |-> TypeSignature],
+                                                    [term |-> 2, contentType |-> TypeReconfiguration, configuration |-> startNodes],
+                                                    [term |-> 2, contentType |-> TypeSignature] >>
+                                            ELSE << >>]
+        /\ commitIndex  = [i \in servers |-> IF i \in startNodes THEN 4 ELSE 0]
+        /\ committableIndices  = [i \in servers |-> {}]
+
```
lemmy, line 491: Given the similarities, `StartNode` and `JoinedState` could be refactored into a single definition parameterized by the log.

```tla
InitLogConfigServerVars(startNodes, log) ==
    /\ reconfigurationCount = 0
    /\ removedFromConfiguration = {}
    /\ \E startNode \in startNodes:
        /\ configurations = [ i \in Servers |-> IF i \in startNodes  THEN (Len(log)-1 :> startNodes) ELSE << >>]
        /\ currentTerm = [i \in Servers |-> IF i \in startNodes THEN 2 ELSE 0]
        /\ state       = [i \in Servers |-> IF i = startNode THEN Leader ELSE IF i \in startNodes THEN Follower ELSE None]
        /\ votedFor    = [i \in Servers |-> Nil]
        /\ log         = [i \in Servers |-> IF i \in startNodes
                                            THEN log
                                            ELSE << >>]
        /\ commitIndex  = [i \in Servers |-> IF i \in startNodes THEN Len(log) ELSE 0]
        /\ committableIndices  = [i \in Servers |-> {}]

File: tla/consensus/Traceccfraft.tla [hunk 17]
```diff
@@ -378,22 +368,6 @@ TraceNext ==
...
-       /\ state[logline.msg.state.node_id] = None
-       /\ configurations' = [ configurations EXCEPT ![logline.msg.state.node_id] = ToConfigurations(<<logline.msg.new_configuration>>)]
-       /\ state' = [ state EXCEPT ![logline.msg.state.node_id] = Follower ]
-       /\ UNCHANGED <<reconfigurationCount, removedFromConfiguration, messageVars, currentTerm, votedFor, candidateVars, leaderVars, logVars>>    
-    \/ /\ IsEvent("become_follower")
-       /\ state[logline.msg.state.node_id] = None
-       /\ configurations[logline.msg.state.node_id] = ToConfigurations(logline.msg.configurations)
-       /\ state' = [ state EXCEPT ![logline.msg.state.node_id] = Follower ]
-       /\ UNCHANGED <<reconfigurationVars, removedFromConfiguration, messageVars, currentTerm, votedFor, candidateVars, leaderVars, logVars>>
-
```
lemmy, line 396: \o/

.azure-pipelines-templates/trace_validation.yml
```diff
@@ -10,16 +10,5 @@ steps:
...
-      ../build/check_quorum_deprecated.ndjson \
-      ../build/reconnect_deprecated.ndjson \
-      ../build/reconnect_node_deprecated.ndjson \
-      ../build/bad_network_deprecated.ndjson \
-      ../build/fancy_election.1_deprecated.ndjson \
-      ../build/fancy_election.2_deprecated.ndjson \
-      ../build/suffix_collision.1_deprecated.ndjson \
-      ../build/suffix_collision.2_deprecated.ndjson \
-      ../build/suffix_collision.3_deprecated.ndjson
+      ../build/startup.ndjson
```
lemmy, line 13: Do none of the old scenarios continue to validate?

tla/consensus/Traceccfraft.tla
```diff
@@ -92,11 +87,8 @@ TraceAppendEntriesBatchsize(i, j) ==
 TraceInitReconfigurationVars ==
     /\ reconfigurationCount = 0
     /\ removedFromConfiguration = {}
-    \* Weaken  ccfraft!InitReconfigurationVars  to allow a node's configuration to be initially empty.
-     \* This seems to be a quirk of raft_driver (related RaftDriverQuirk).
-    /\ configurations = [ s \in Servers |-> IF s = TraceLog[1].msg.state.node_id 
-                                            THEN ToConfigurations(<<TraceLog[1].msg.new_configuration>>)
-                                            ELSE [ j \in {0} |-> {} ] ]
+    /\ LET startNode == TraceLog[1].msg.state.node_id
+       IN CCF!StartState(startNode, TraceServers)
```
lemmy, line 91: Given that `Traceccfraft` extends `ccfraft`, just `StartState` should also work.

tla/consensus/SIMccfraft.cfg
```diff
@@ -73,8 +73,6 @@ INVARIANTS
     
     MonoTermInv
     MonoLogInv
-    LogConfigurationConsistentInv
-    NoLeaderInTermZeroInv
```
lemmy, line 77: Check `NoLeaderBeforeInitialTerm` instead?

tests/raft_scenarios_runner.py
```diff
@@ -36,18 +36,45 @@ def write_error_report(errors=None):
...
+    assert head()["msg"]["function"] == "become_leader"
+    assert head()["msg"]["function"] == "add_configuration"
+    signature = head()
+    assert signature["msg"]["function"] == "replicate", signature
+    assert signature["msg"]["globally_committable"], signature
+    commit = head()
+    assert commit["msg"]["function"] == "commit", commit
+    assert commit["msg"]["state"]["commit_idx"] == 2, commit
+    # Commit becomes bootstrap, the entry point into the trace validation
+    commit["msg"]["function"] = "bootstrap"
```
lemmy, line 76: Could `raft.h` log it under "bootstrap" depending on its internal state, instead of interfering here?

tla/consensus/ccfraft.tla
```diff
@@ -456,28 +456,49 @@ PlausibleSucessorNodes(i) ==
...
+    /\ removedFromConfiguration = {}
+    /\ \E startNode \in startNodes:
+        /\ configurations = [ i \in servers |-> IF i \in startNodes  THEN (3 :> startNodes) ELSE << >>]
+        /\ currentTerm = [i \in servers |-> IF i \in startNodes THEN 2 ELSE 0]
+        /\ state       = [i \in servers |-> IF i = startNode THEN Leader ELSE IF i \in startNodes THEN Follower ELSE None]
+        /\ votedFor    = [i \in servers |-> Nil]
+        /\ log         = [i \in servers |-> IF i \in startNodes
+                                            THEN << [term |-> 2, contentType |-> TypeReconfiguration, configuration |-> {startNode}],
+                                                    [term |-> 2, contentType |-> TypeSignature],
+                                                    [term |-> 2, contentType |-> TypeReconfiguration, configuration |-> startNodes],
```
lemmy, line 486: Why is it `{startNodes}` above and `startNodes` at log entry 3?

lemmy: Never mind, confused `startNode` and `startNodes`. => Former renamed to `sn` below.

tla/consensus/ccfraft.tla
```diff
@@ -456,28 +456,49 @@ PlausibleSucessorNodes(i) ==
...
-    /\ reconfigurationCount = 0
-    /\ removedFromConfiguration = {}
-    \* Note that CCF has a bootstrapping procedure to start a new network and to join new nodes to the network (see 
-    \* https://microsoft.github.io/CCF/main/operations/start_network.html). In both cases, a node has the current (see 
-    \* https://microsoft.github.io/CCF/main/operations/ledger_snapshot.html#join-or-recover-from-snapshot) or some stale configuration
-    \* such as the initial configuration. A node's configuration is *never* "empty", i.e., the equivalent of configuration[node] = {} here. 
-    \* For simplicity, the set of servers/nodes all have the same initial configuration at startup.
-    /\ \E s \in Servers:
-        configurations = [i \in Servers |-> 0 :> {s} ]
+    \E startNode \in Servers:
```
lemmy, line 496: Quantification of `startNodes` should fail in `StartState` and `JoinedState` because `startNode` cannot be quantified over.  `SUBSET Servers` maybe?

tla/consensus/ccfraft.tla
```diff
@@ -456,28 +456,49 @@ PlausibleSucessorNodes(i) ==
...
+        StartState(startNode, Servers)
```
lemmy, line 497:
```tla
InitLogConfigServerVars({startNode}, Servers, << [term |-> 2, contentType |-> TypeReconfiguration, configuration |-> {startNode}], [term |-> 2, contentType |-> TypeSignature] >>)
```

.azure-pipelines-templates/trace_validation.yml
achamayou: No, they all rely on the quirks right now, they need to be converted. I tried doing everything in one go in the other PR, but it proved difficult.

tests/raft_scenarios_runner.py
achamayou: Perhaps, but not without extra state, and interfering with `raft.h` is much more risky and unappealing than massaging the trace a bit. My mental model is that raft traces should be function calls, with a pre-state and relevant arguments (*). This not 100% true just now, but we're not far off.

Pushing complexity into the implementation because of limitations of the trace validation does not seem justifiable to me. In a perfect world we would unwind the bootstrap and make the spec able to cope with the full details of startup instead.

Also we may want to have a more generic version of this in the future for #5290, which could bootstrap from arbitrary states on any given trace. It's much easier to experiment with this in a script that in the production code.

(*) One because it's then very easy to ifdef them out and convince ourselves they have 0 runtime impact in release builds, and two because it means we probably never need to make changes to them, which means that we diminish the chance that we will ever impact release builds because of them.

tla/consensus/SIMccfraft.cfg
achamayou: Oops, yes, thank you.

tla/consensus/Traceccfraft.tla
achamayou: I think I tried that, but I'll try again.

tla/consensus/Traceccfraft.tla
achamayou: Well, now we have the more inoffensively named "bootstrap"... But it's at least a state snapshot after a realistic startup sequence that matches end to end, no more ex-nihilo configs. So I claim it's still progress.

tla/consensus/ccfraft.tla
achamayou: Yes, but there's only two, and that also makes it unsafe because it's then easy to make a call to InitLogConfigServerVars that generates an unrealistic startNodes, log pair. If it could be made private, and then JoinedState and StartNode were the only two public ways to call into it, it'd be good, but I haven't found a way to do that.

tla/consensus/ccfraft.tla
achamayou: That's a possibility, but I'm not sure it's a good one, as explained above.

tla/consensus/ccfraft.tla
achamayou: So I had a go, but it turns out not to work, because to write the log in the first place, you need to take out the startNode, and you end up having to write a bunch of extra checks to make sure the log is half plausible.

So these are the two correct start states for now, and if we really need to make new ones, we will revisit this. But the next order of business (#5822) is to get the trace validation to work for all scenarios. Extending for more initial states will wait for now.

```
_InitialStateWithLog(initialLog, servers) ==
    /\ Len(initialLog) >= 2
    \* Log must end with a config and a signature
    /\ initialLog[Len(initialLog) - 1].contentType = TypeReconfiguration
    /\ initialLog[Len(initialLog)].contentType = TypeSignature
    \* Log must end with a single node config and a signature
    /\ initialLog[1].contentType = TypeReconfiguration
    /\ initialLog[2].contentType = TypeSignature
    /\ DOMAIN initialLog[1].configuration = {1}
    /\ Cardinality(initialLog[1].configuration[1]) = 1
    /\ reconfigurationCount = 0
    /\ removedFromConfiguration = {}
    /\ LET startNodes == initialLog[Len(initialLog) - 1].configurations[Max(DOMAIN configurations)]
       IN \E startNode \in startNodes:
        /\ configurations = [ i \in servers |-> IF i \in startNodes  THEN (3 :> startNodes) ELSE << >>]
        /\ currentTerm = [i \in servers |-> IF i \in startNodes THEN StartTerm ELSE 0]
        /\ state       = [i \in servers |-> IF i = startNode THEN Leader ELSE IF i \in startNodes THEN Follower ELSE None]
        /\ votedFor    = [i \in servers |-> Nil]
        /\ log         = [i \in servers |-> IF i \in startNodes THEN initialLog ELSE << >>]
        /\ commitIndex  = [i \in servers |-> IF i \in startNodes THEN Len(log[i]) ELSE 0]
        /\ committableIndices  = [i \in servers |-> {}]
```

tla/consensus/ccfraft.tla
```diff
@@ -82,6 +82,9 @@ CONSTANTS
 CONSTANTS Servers
 ASSUME Servers /= {} /\ IsFiniteSet(Servers)
 
+\* Initial term used by the Start node in the network
+CONSTANT StartTerm
```
lemmy, line 86: The config definitions won't be necessary with `StartTerm == 2`.

achamayou: Will fix

PR #5832, opened 2023-11-13
Authenticate recovery share submission

### LINKED ISSUE #5821: The /recovery/members/{memberId}:recover endpoint in the 2023-06-01-preview API is unauthenticated
Users submitting a COSE Sign1 payload get a JSON parsing error as a result.

The code should definitely at least support both, for consistency (all POSTs authenticated via signatures), and probably only authenticated access to avoid a potential denial of service by a malicious sender posting invalid shares.

achamayou:
Closes #5821, and makes new API compliant with the [spec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/confidentialledger/Microsoft.ManagedCcf/examples/2023-06-01-preview/Shares_Submit.json).

@lynshi as discussed.

TODO:

- [x] Fix LTS compatibility test by remapping arguments and calling the script with the right arguments depending on what it supports.


lynshi (approved):

src/node/gov/handlers/recovery.h
```diff
@@ -195,7 +205,7 @@ namespace ccf::gov::endpoints
         "/recovery/members/{memberId}:recover",
         HTTP_POST,
         api_version_adapter(submit_recovery_share),
-        ccf::no_auth_required)
+        detail::active_member_sig_only_policies("recovery_share"))
```
lynshi, line 208: Don't know enough to know if this is the correct policy/policy inputs, so just leaving a note for you to double check it :smile:

achamayou: It is correct (active members through COSE Sign1), and consistent with other POST endpoints: https://github.com/microsoft/CCF/blob/7372e76daf0d21d540ac6b6a342053c43a666adf/src/node/gov/handlers/helpers.h#L18

always good to check :)

PR #5833, opened 2023-11-13
Correct state digest key for classic version of the API

### LINKED ISSUE #5816: Acks submitted with old tooling don't work because they send "state_digest" rather than "stateDigest"
To be investigated.

achamayou:
Closes #5816.


src/node/gov/handlers/acks.h
```diff
@@ -55,7 +55,11 @@ namespace ccf::gov::endpoints
 
           auto response_body = nlohmann::json::object();
           response_body["memberId"] = member_id_str;
-          response_body["stateDigest"] = ack->state_digest;
+          // Match pre-v1 API naming convention
+          const auto state_digest_key = api_version == ApiVersion::preview_v1 ?
```
lynshi, line 59: nit: `api_version == ApiVersion::classic ? "state_digest" : "stageDigest"` might be a bit more future-proof, as we only expect the classic API to be snake_case

lynshi (approved):

src/node/gov/handlers/acks.h
achamayou: I do agree, but there isn't an ApiVersion::classic (https://github.com/microsoft/CCF/blob/7372e76daf0d21d540ac6b6a342053c43a666adf/src/node/gov/api_version.h#L14), and Eddy seems to have structured things the other way. Let me have a look through the rest of endpoints and see if I can do this consistently.

achamayou: @lynshi so there is not an `api-version=classic`, there is only not passing `api-version` on old endpoints, and passing an `api-version` that's [supported](https://github.com/microsoft/CCF/blob/583fab2039990d4535a152d77016998e78d84f7c/src/node/gov/api_version.h#L20), and right now that's only `2023-06-01-preview` (as linked above).

So this change is no-op, it's not possible to take this path.

The old API is unchanged, and [still uses state_digest](https://github.com/microsoft/CCF/blob/583fab2039990d4535a152d77016998e78d84f7c/src/node/rpc/member_frontend.h#L799).

I will close this then, because tooling that's not been updated will continue to work (old case, old API), and tooling that's updated can be updated throughout (case and API path). I do not think that supporting a weird middle ground with the new paths and old cases makes sense.

PR #5836, opened 2023-11-16
Trace viz script

achamayou:
The JSON traces are painful to read when working on validation, especially with lots of nodes, I would really like something more compact and readable. I'd like to get this working for the traces we get from TLC too, eventually.

This needs a venv with the test dependencies installed to work. If you've run `make_traces.sh` before, you'll have that, just `source build/env/bin/activate`. Or just `pip install rich` in whatever venv you use.

The output looks like this:

image

Suggestions and changes very welcome.


heidihoward (approved):
this is fantastic

PR #5839, opened 2023-11-17
TV for startup followed by join

achamayou:
Matches, except sent_idx/nextIdx, but I claim we want to look at that separately, as suggested in #5895.


tla/consensus/ccfraft.tla
```diff
@@ -656,6 +656,7 @@ ChangeConfigurationInt(i, newConfiguration) ==
     \* Keep track of running reconfigurations to limit state space
     /\ reconfigurationCount' = reconfigurationCount + 1
     /\ removedFromConfiguration' = removedFromConfiguration \cup (CurrentConfiguration(i) \ newConfiguration)
+    /\ \A addedNode \in (newConfiguration \ CurrentConfiguration(i)) : nextIndex' = [nextIndex EXCEPT ![i][addedNode] = Len(log[i]) + 1]
```
lemmy, line 659: I suggest adding the pointer to https://github.com/microsoft/CCF/blob/a648fc3743a77e363b8d8a3f4e002cc5828d57a3/src/consensus/aft/raft.h#L2401 here.

lemmy:
> I'll switch to Codespaces to get the debugger on Monday.

Remember to add `-DTLA-Library=..` to "Tlaplus › Java: Options" in VSCode.

heidihoward (approved):

achamayou:
So the change to UpdateTerm to transition None to Follower has broken the MC/Sim, because we do not restrict the sending of AppendEntries: a leader can send AEs to any node, even those not in configuration. This is of course not what the implementation does, and cause `LogConfigurationConsistentInv` to break.

tla/consensus/ccfraft.tla
```diff
@@ -1219,13 +1223,18 @@ MonoLogInv ==
...
+        \/ /\ state[i] = Follower
+           /\ Cardinality(DOMAIN configurations[i]) = 0
         \/
             \* Configurations should have associated reconfiguration txs in the log
             /\ \A idx \in DOMAIN (configurations[i]) : 
                 /\ log[i][idx].contentType = TypeReconfiguration            
                 /\ log[i][idx].configuration = configurations[i][idx]
-            \* Current configuration should be committed
-            /\ commitIndex[i] >= CurrentConfigurationIndex(i)
+            \* Current configuration should be committed if there is one
```
achamayou, line 1234: This just does not hold, and not just at startup, because the highest config in the log is not even committable (and so definitely not committed) until it's followed by a signature. This is what broke in the MC once the change at 1227/8 was made to fix situations where there aren't any configs on the new joiner yet.

tla/consensus/ccfraft.tla
```diff
@@ -645,6 +645,10 @@ ChangeConfigurationInt(i, newConfiguration) ==
     \* i.e., be re-added to a new configuration.  Instead, the node has to rejoin with a
     \* "fresh" identity (compare sec 6.2, page 8, https://arxiv.org/abs/2310.11559).
     /\ \A s \in newConfiguration: s \notin removedFromConfiguration
+    \* See raft.h:2401, nodes are only sent future entries initially, they will NACK if necessary.
+    \* This is because they are expected to start from a fairly recent snapshot, not from scratch.
+    /\ LET addedNodes == newConfiguration \ CurrentConfiguration(i)
+       IN IF Cardinality(addedNodes) > 0 THEN \A addedNode \in addedNodes : sentIndex' = [sentIndex EXCEPT ![i][addedNode] = Len(log[i])] ELSE sentIndex' = sentIndex
```
heidihoward, line 651: @lemmy  does `\A addedNode \in addedNodes : sentIndex' = [sentIndex EXCEPT ![i][addedNode] = Len(log[i])]` work alright?

achamayou: That's what I had before, but it sometimes resulted in sentIndex not being assigned (where there are no addedNodes, afaict), and TLC did treat that as an error.

tla/consensus/ccfraft.tla
```diff
@@ -1224,13 +1229,14 @@ MonoLogInv ==
...
         \/ state[i] = None
+        \* Follower, but no known configurations yet
+        \/ /\ state[i] = Follower
+           /\ Cardinality(DOMAIN configurations[i]) = 0
         \/
             \* Configurations should have associated reconfiguration txs in the log
             /\ \A idx \in DOMAIN (configurations[i]) : 
                 /\ log[i][idx].contentType = TypeReconfiguration            
                 /\ log[i][idx].configuration = configurations[i][idx]
-            \* Current configuration should be committed
```
heidihoward, line 1232: could this be either configurations is empty or current configurations is before the commit index?

heidihoward (approved):

tla/consensus/ccfraft.tla
heidihoward: That's actually not quite what I meant, I wanted to check that a comparison to a primed variable within a for-all works like that.

I think the alternative is something like: 
```tla
LET
new_sent_index_i = [ k \in Servers |-> IF k \in AddedNodes THEN Len(log[i]) ELSE sentIndex[i][k]]
IN
sentIndex' = [sentIndex EXCEPT ![i] = new_sent_index_i]
```

tla/consensus/ccfraft.tla
achamayou: Empty when Follower is dealt with on line 1232, but this just generally wrong as soon as there is a new pending configuration, since CurrentConfigurationIndex is just whatever the latest known config is. It may be not even be committable at all if a signature has not been emitted.

I am not sure what, if anything, we want to assert here.

tla/consensus/ccfraft.tla
achamayou: That does look simpler, I'll make the change.

PR #5842, opened 2023-11-17
Refactor JoinedState and StartState into one.

lemmy:


tla/consensus/ccfraft.tla
```diff
@@ -459,45 +459,28 @@ PlausibleSucessorNodes(i) ==
...
+        /\ state       = [i \in Servers |-> IF i = sn THEN Leader ELSE IF i \in startNodes THEN Follower ELSE None]
+        /\ commitIndex = [i \in Servers |-> IF i \in startNodes THEN Len(xlog({sn}, startNodes)) ELSE 0]
+    /\ configurations = [ i \in Servers |-> IF i \in startNodes  THEN (Len(log[i])-1 :> startNodes) ELSE << >>]
+    
 ------------------------------------------------------------------------------
 \* Define initial values for all variables
 
 InitReconfigurationVars ==
     \E startNode \in Servers:
-        StartState(startNode, Servers)
```
achamayou, line 500: There's also a reference to StartState in the trace spec that needs updating.

tla/consensus/ccfraft.tla
```diff
@@ -459,45 +459,28 @@ PlausibleSucessorNodes(i) ==
...
-    /\ commitIndex  = [i \in servers |-> IF i = startNode THEN Len(log[i]) ELSE 0]
-    /\ committableIndices  = [i \in servers |-> {}]
-
-\* Generate initial state for a startNodes configuration
-JoinedState(startNodes, servers) ==
+StartLog(startNode, nextNodes) ==
+    << [term |-> StartTerm, contentType |-> TypeReconfiguration, configuration |-> startNode],
+       [term |-> StartTerm, contentType |-> TypeSignature] >>
+
+InitLogConfigServerVars(startNodes, xlog(_,_)) ==
```
achamayou, line 466: I think it would be nice to name xlog(_, _), perhaps logPrefix(startNode_, initialConfigNodes), as a hint to the caller.

tla/consensus/ccfraft.tla
```diff
@@ -459,45 +459,28 @@ PlausibleSucessorNodes(i) ==
...
-        /\ committableIndices  = [i \in servers |-> {}]
-
+    /\ committableIndices  = [i \in Servers |-> {}]
+    /\ votedFor    = [i \in Servers |-> Nil]
+    /\ currentTerm = [i \in Servers |-> IF i \in startNodes THEN StartTerm ELSE 0]
+    /\ \E sn \in startNodes:
+        /\ log         = [i \in Servers |-> IF i \in startNodes THEN xlog({sn}, startNodes) ELSE << >>]
+        /\ state       = [i \in Servers |-> IF i = sn THEN Leader ELSE IF i \in startNodes THEN Follower ELSE None]
+        /\ commitIndex = [i \in Servers |-> IF i \in startNodes THEN Len(xlog({sn}, startNodes)) ELSE 0]
+    /\ configurations = [ i \in Servers |-> IF i \in startNodes  THEN (Len(log[i])-1 :> startNodes) ELSE << >>]
```
achamayou, line 476: This is making unspoken assumptions about what xlog constructs (end on a signature preceded by a config), which happen to be true on both StartLog and JoinedLog. I think it'd be best to at least write them in comment if we're not going to have a TypeOK for that.

tla/consensus/ccfraft.tla
```diff
@@ -459,45 +459,28 @@ PlausibleSucessorNodes(i) ==
...
-    /\ log         = [i \in servers |-> IF i = startNode
-                                        THEN << [term |-> StartTerm, contentType |-> TypeReconfiguration, configuration |-> {startNode}],
-                                                [term |-> StartTerm, contentType |-> TypeSignature] >>
-                                        ELSE << >>]
-    /\ commitIndex  = [i \in servers |-> IF i = startNode THEN Len(log[i]) ELSE 0]
-    /\ committableIndices  = [i \in servers |-> {}]
-
-\* Generate initial state for a startNodes configuration
-JoinedState(startNodes, servers) ==
+StartLog(startNode, nextNodes) ==
```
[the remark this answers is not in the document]
lemmy: Anonymous parameters can only appear in an operator declaration but are not legal in a definition.  The parameter name can be changed to something like `ignored`, though.

tla/consensus/ccfraft.tla
lemmy: I left a comment.  Also, `SignatureInv` and `LogConfigurationConsistentInv` would be violated.

achamayou: Good point! The `SignatureInv` I had thought about, but not the `LogConfigurationConsistentInv`.

achamayou (approved):

PR #5852, opened 2023-12-05
Test THIM on ACI

achamayou:
Test that attempts to fetch AMD collateral from `${Fabric_NodeIPOrFQDN}:2377` on ACI.


tests/amd_certs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the Apache 2.0 License.
+
+set -ex
+
```
achamayou, line 6:
```suggestion
env
```

achamayou:
To be re-opened when the feature becomes available.

PR #5863, opened 2023-12-11
Partially support TextEncoder Web API

takuro-sato:
## Future? TODOs
- Support TextDecoder: Slightly it's awkward if we try to support this partially since the default value for `fatal` is `false` while CCF's `bufToStr()` behaves like `fatal=true`.  https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder#options .
- Support TextEncoder.encodeInto()


achamayou:
@takuro-sato thank you for opening this!

> ## Future? TODOs
> * Support TextDecoder: Slightly it's awkward if we try to support this partially since the default value for `fatal` is `false` while CCF's `bufToStr()` behaves like `fatal=true`.  https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder#options .

I am not sure that is right, a superficial look at https://github.com/microsoft/CCF/blob/main/3rdparty/exported/quickjs/quickjs.c#L3910 suggests that `bufToStr()` might actually skip invalid utf-8 and effectively behave like `fatal=false`. I will look into that.


> * Support TextEncoder.encodeInto()

That would require a small change to strToBuf (or a new API) to allow writing the string into an existing buffer, I will look into that too.

achamayou (approved):

PR #5868, opened 2023-12-18
Remove now-unused security_context_directory

achamayou:
With `snp_security_policy_file`, `snp_endorsement_servers` and `snp_uvm_endorsements_file` all supporting environment variables resolution, there is now no reason for CCF to handle environment variables directly or in the CLI. Instead the same configuration structure can be used on C-AKS and C-ACI.

This removes the `--snp-security-context-dir-var` CLI option and deprecated `security_context_dir` configuration value.


samples/config/start_config_aci_sev_snp.json
```diff
@@ -34,13 +34,10 @@
...
   },
   "attestation": {
-    "environment": {
-      "security_context_directory": "UVM_SECURITY_CONTEXT_DIR"
-    },
     "snp_endorsements_servers": [
       {
-        "type": "THIM",
-        "url": "169.254.169.254"
+        "type": "Azure",
```
achamayou, line 39: Until THIM becomes available on ACI.

doc/operations/platforms/snp.rst
```diff
@@ -18,13 +18,14 @@ Confidential Azure Container Instance (ACI)
 
 .. note:: See `here <https://learn.microsoft.com/en-us/azure/container-instances/container-instances-tutorial-deploy-confidential-containers-cce-arm>`_ for more information on the deployment of confidential containers in Azure.
 
-Azure Confidential ACI provides a security context directory containing the following files. The content of these files are checked against the attestation report on node startup and join, and stored in the ledger for audit and improved serviceability. 
+Azure Confidential ACI provides a security context directory containing the following files.
 
-- ``host-amd-cert-base64``: The certificate chain corresponding to the key (VCEK) used to sign the attestation report, up to the well-known AMD root of trust certificate authority (Base64 encoded).
```
achamayou, line 23: This is still provided, but subject to a race condition on startup if PSP upgrades take place at the wrong time. Resolution through an online service such as THIM using the chip id should be used instead.

src/common/configuration.h
```diff
@@ -71,11 +71,7 @@ DECLARE_JSON_OPTIONAL_FIELDS(CCFConfig::JWT, key_refresh_interval);
 DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(CCFConfig::Attestation::Environment);
 DECLARE_JSON_REQUIRED_FIELDS(CCFConfig::Attestation::Environment);
 DECLARE_JSON_OPTIONAL_FIELDS(
-  CCFConfig::Attestation::Environment,
-  security_context_directory,
-  security_policy,
-  uvm_endorsements,
-  report_endorsements);
+  CCFConfig::Attestation::Environment, security_policy, uvm_endorsements);
```
achamayou, line 74: Environment remains as an internal implementation detail for, to carry the contents of the policy and the endorsements into the enclave. This can be removed once SGX support is dropped,

PR #5895, opened 2024-01-08
`NextIndex` is not updated on successful AE response

heidihoward:
CCF's consensus optimistically updates the send_idx when an AE message is dispatched so it does not need to update it on a successful AE response

PR #5897, opened 2024-01-08
Swapping nextIndex for sentIndex

heidihoward:
Currently, `raft.h` uses `sent_index` to track the last index sent to a follower (ish) but the TLA+ specification uses `nextIndex` to track the next index to send to each follower (`sent_index` +1).  As a result, the trace checker needs to increment `sent_index` before comparison to `nextIndex`. 

This PR updates the TLA+ specification to swap `nextIndex` from `sentIndex`, renaming the variable and decreasing its value by one.

Previously mentioned here: https://github.com/microsoft/CCF/pull/5895#issuecomment-1880866881


lemmy:
@heidihoward As an FYI: `nextIndex` and `sent_idx` gave me a hard time when aligning the spec and impl, which is why I would recommend to backport this change to before we changed the bootstrapping logic.  The coverage of the current trace validation is rather low.

achamayou (approved):

heidihoward:
CI metrics are looking suspicious. I expect we are not reaching the expected state of states, but I'll take a look tomorrow

I would be great to add reachability to the test suite (like we have for the consistency specs) https://github.com/microsoft/CCF/blob/main/tla/consistency/MCMultiNodeCommitReachability.cfg

PR #5902, opened 2024-01-09
Refactoring consensus leadership states in specification

heidihoward:
Currently the TLA+ specification for consensus has just 4 node states: `None`, `Follower`, `Candidate`, `Leader`, `RetiredLeader` and no state for a retired non-leader. This PR will add membership state and remove `RetiredLeader` as a leadership state.


tla/consensus/ccfraft.tla
```diff
@@ -217,7 +227,7 @@ CurrentTermTypeInv ==
 VARIABLE state
 
 StateTypeInv ==
-    \A i \in Servers : state[i] \in States
+    \A i \in Servers : state[i] \in LeadershipStates
```
achamayou, line 230: It probably makes sense to rename `state` to `leadershipState`, or `leadership` (as opposed to `membership`, `membershipState`, or `mState` or whatever), but that may be the next PR.

heidihoward: done in https://github.com/microsoft/CCF/pull/5902/commits/db272604b4644eb45a82edd72d1dfd9319df8746

heidihoward:
I'm going to put this PR up for review. There's still a lot of work to do to get retirement in the spec to align with the implement but I think this is step in the right direction

achamayou (approved):

PR #5903, opened 2024-01-09
Ask `ruff` to autofix Python lint errors, where possible

eddyashton:
For some errors that ruff spots, such as unused includes, it will print an error along the lines of "I could have tried to fix this automatically, if you'd given me the `--fix` arg". It's odd-to-annoying ifthat appears when we _have_ passed the `-f[ix]` arg to our `ci-checks` wrapper!

So the wrapper's `fix` mode now enables ruff's `fix` mode, as it does with similar tools.

Also, because we have mixed verbosity of the tools called from this wrapper, I've modified it to always print group headings. There might be a reason this was disabled, that I've forgotten? But I think it makes the output far more readable.

### DIFF
--- a/scripts/ci-checks.sh
+++ b/scripts/ci-checks.sh
@@ -21,6 +21,8 @@ function group(){
     # https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
     if [[ ${CI} ]]; then
       echo "::group::$1"
+    else
+      echo "-=[ $1 ]=-"
     fi
 }
 function endgroup() {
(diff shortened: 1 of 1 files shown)


achamayou (approved):

achamayou:
> Also, because we have mixed verbosity of the tools called from this wrapper, I've modified it to always print group headings. There might be a reason this was disabled, that I've forgotten? But I think it makes the output far more readable.

I don't think there was a reason, I think there were no group headings at first, then someone added them just for CI but didn't want to make the regular output more verbose. The script only did clang-format and pylint at first, with very minimal output, but has grown to call a whole bunch of things, so I agree that having the headings now makes a lot of sense.

PR #5904, opened 2024-01-10
Add aggregate trace expressions to aid debugging of counterexamples.

lemmy:
(ANSI escape sequences are used to colorize the output, which are not supported by the VSCode extension.)

Related: https://github.com/tlaplus/tlaplus/issues/858


tla/consensus/SIMccfraft.cfg
```diff
@@ -53,6 +53,11 @@ PROPERTIES
 POSTCONDITION
     WriteStatsFile
 
+\* ALIAS
```
achamayou, line 56: This needs to be uncommented to get the animation, presumably?

lemmy: Yes, you will pick the alias most suitable for the current debugging task.

achamayou (approved):

tla/consensus/SIMccfraft.cfg
achamayou: I tried it, it looks very cool!

Interestingly, I put a constraint on sentIndex to get it to fail, and it only ever (sample size ~12) generated single node configs and logs.

lemmy:
### `DebugAlias`

This shows the original variables in addition to several aggregate information s.a. a timeline, the current cluster members, the "state history" of each node (pass `-Dtlc2.value.Values.width` to adjust TLC's line wrapping).

Screenshot 2024-01-10 at 8 22 09 AM

### `DebugActingServerAlias`

`DebugActingServerAlias` shows only the state of `srv` for which `\E srv \in Server: [Next]_vars` in the state.  The identity of `srv` is also indicated by its color.

Screenshot 2024-01-10 at 8 16 29 AM

### `AnimateLogAndStateAlias`

Good to eyeball the `state `and content of each node's `log`.

https://github.com/microsoft/CCF/assets/88777/e51c299d-70ab-4ba4-912d-304ce39ac0c6

tla/consensus/SIMccfraft.cfg
lemmy: I also observed less-than-ideal coverage, which is what I wanted to look at next.

achamayou: Yes, absolutely, this is very cool!

PR #5905, opened 2024-01-10
Follow up to #5889 - avoid misaligned peeks

eddyashton:
This change and test exposed an issue with `serialized::peek<T>` - it makes no alignment checks before reading entries (caught by ASAN).

This change adds some unit tests for `serialized::peek`, and changes the behaviour to test alignment and copy to scratch memory if misaligned.

### DIFF
--- a/src/ds/serialized.h
+++ b/src/ds/serialized.h
@@ -44,7 +44,7 @@ namespace serialized
     if (reinterpret_cast<std::uintptr_t>(data) % alignment != 0)
     {
       // Data is not aligned - copy to scratch memory
-      alignas(64) thread_local uint8_t scratch[max_size];
+      alignas(T) uint8_t scratch[max_size];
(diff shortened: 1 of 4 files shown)


src/ds/serialized.h
```diff
@@ -30,13 +30,29 @@ namespace serialized
...
       throw InsufficientSpaceException(
         fmt::format("Insufficient space (peek<T>: {} < {})", size, sizeof(T)));
     }
 
-    return *(T*)data;
+    static constexpr auto alignment = alignof(T);
+    if (reinterpret_cast<std::uintptr_t>(data) % alignment != 0)
+    {
+      // Data is not aligned - copy to scratch memory
+      alignas(64) thread_local uint8_t scratch[max_size];
```
achamayou, line 47: unless I'm missing something, I think that should just be on the stack

achamayou (approved):

PR #5909, opened 2024-01-10
Election scenario for trace validation

achamayou:
Follow up to #5907

Last one because they are getting kind of long:

image


tla/consensus/Traceccfraft.tla
```diff
@@ -205,7 +205,10 @@ IsAddConfiguration ==
     /\ IsEvent("add_configuration")
     /\ state[logline.msg.state.node_id] = Follower
     /\ UNCHANGED vars
-    /\ committableIndices[logline.msg.state.node_id] = Range(logline.msg.state.committable_indices)
+\* This won't work in situations where we receive an AE range that contains a configuration at first followed by committable indices:
+\* recv_append_entries will update the committable indices in the spec, but not in the impl state, which then goes on to handle an
+\* add_configuration event on which state->committable_indices is (correctly) empty.
+\*    /\ committableIndices[logline.msg.state.node_id] = Range(logline.msg.state.committable_indices)
```
lemmy, line 211: Why wasn't this spec <> impl discrepancy not found with the previous trace validation?  Does is make sense to log an issue to align the spec and impl eventually, before the commented check gets forgotten?

achamayou: @lemmy because the previous trace validation never ran on a scenario that contained a reconfiguration :)

I am not sure if we need to align anything here, not all the state variables will always match the between the spec and the implementation on every step, because of the steps are different and so is their ordering. If we refactor the spec enough to align, that may become a realistic goal, but it is not now.

I am happy to hear what others have to say about this (hence the comment), but my inclination is to just delete this line.

heidihoward (approved):

PR #5911, opened 2024-01-10
Check quorum scenario

achamayou:
Follows #5910


tla/consensus/ccfraft.tla
```diff
@@ -562,7 +562,8 @@ AppendEntries(i, j) ==
            prevLogTerm == IF prevLogIndex \in DOMAIN log[i] THEN
                               log[i][prevLogIndex].term
                           ELSE
-                              StartTerm
+                              \* state.h::view_at (:64) indices before the version of the first view are unknown
+                              0
```
lemmy, line 566: I assume `view_at` is set to `0` in the impl for passing type checks.  However, to catch bugs early in the spec, it might be better to use `Nil` instead of `0`. This is because any arithmetic unknown to failure.

achamayou: Independently of the reasons that led the implementation to pick 0 (which I think were to do with serialisation and efficiency, not type checking), the value will not change there, and we really would like the spec and its state to be as close as possible to the implementation as possible.

There's no arithmetic that I am aware of with prevLogTerm, so I think we would like to forbid it for all values if we could, not just 0.

eddyashton (approved):

heidihoward (approved):

PR #5919, opened 2024-01-11
Adding retirement phases to specification

heidihoward:
Adding retirement phases to the TLA+ spec


tla/consensus/ccfraft.tla
```diff
@@ -444,6 +486,30 @@ CommittedTermPrefix(i, x) ==
     \* Otherwise the prefix is the empty tuple
     ELSE << >>
 
+RetirementIndexLog(node_log, i) ==
+    LET 
+        in_indexes = {index \in DOMAIN node_log:
```
achamayou, line 491: The spec is (mostly) camelCase, but unfortunately we don't have a linter/formatter.

tla/consensus/ccfraft.tla
```diff
@@ -1260,6 +1394,24 @@ LogConfigurationConsistentInv ==
...
+    \A i \in Servers :
+        \/ /\ retirementPhase[i] = NotRetiring 
+           /\ RetirementIndex(i) = 0
+        \* RetirementOrdered - node's retirement in its log
+        \/ /\ retirementPhase[i] = RetirementOrdered 
+           /\ RetirementIndex(i) # 0
+        \* RetirementSigned - node' retirement is signed 
+        \/ /\ retirementPhase[i] = RetirementSigned
+           /\ RetirementIndex(i) # 0
+           /\ RetirementIndex(i) <= MaxCommittableIndex(log[i])
```
achamayou, line 1378: We could check that if the node is leader and its current term matches the term of retirement index, there are no other signatures between RetirementIndex(i) and MaxCommittableIndex(log[i]). But that's a fairly minor detail.

tla/consensus/ccfraft.tla
heidihoward: changed, i think we need a separate pass of the whole spec to sync up on style

tla/consensus/ccfraft.tla
heidihoward: I don't think that is true in the spec at the moment, should it be?

achamayou: I think it may be, because of https://github.com/microsoft/CCF/blob/37ec70f004fba7c10c4dd65c4d69d81f097c3b75/src/consensus/aft/raft.h#L576C13-L576C39, and the logic that tries to protect against consecutive signatures in the same term. But you're right that we should probably have a closer look first.

heidihoward: My question was why would the sig be consecutive? I think in the impl the leader continues to accept new requests until its retirement is committed

heidihoward:
This rough cut of adding retirement to the spec is now ready for review. I'll look at again in the cold light of day next week, tidy up the code and resolve the outstanding questions re differences between the spec and impl.

tla/consensus/ccfraft.tla
achamayou: The line above is in replicate(), rejecting new entries as soon as retirement_committable_idx is set (i.e. as soon as a signature was emitted after that leader's retirement tx, modulo rollbacks, but then the term would be higher).

heidihoward:
This PR is pretty much done minus some tidying up. The main question I have is that the implementation seems to stop helping with AEs after it believes its retirement is completed, does this not cause a liveness issue if the new leader needs to add a new signature in current term in order to commit the reconfiguration transaction?
https://github.com/microsoft/CCF/blob/42dc993ddc23596965250334540c7e44e1fc7288/src/consensus/aft/raft.h#L1075-L1084

tla/consensus/ccfraft.tla
```diff
@@ -43,21 +43,39 @@ CONSTANTS
...
+    \* Node has added its own retirement to its log, but it is not yet committed or even signed
+    \* The node can still revert to Active upon rollback
+    \* RetirementOrdered does not change the behavior of the node
+    RetirementOrdered,
+    \* Node has added its own retirement to its log and it has been signed but it is not yet committed
+    \* The node can still revert to RetirementOrdered or Active upon rollback
+    \* If the node is a leader, it will stop accepting new client requests
+    RetirementSigned,
+    \* Node retirement has been committed and it is no longer part of the network
+    \* If this node was a leader, it will step down. It will not run for election again.
```
achamayou, line 68: There isn't really an explicit "step down", other than sending a ProposeRequestVote to the most plausible successor. But otherwise it just stops sending heartbeats.

tla/consensus/ccfraft.tla
```diff
@@ -43,21 +43,39 @@ CONSTANTS
...
+    \* This node will continue to respond to AppendEntries and RequestVote messages
+    \* Note that this spec does not model when nodes can be safety removed
```
achamayou, line 70:
```suggestion
    \* Note that this spec does not model when nodes can be safely removed
```

achamayou (approved):

tla/consensus/ccfraft.tla
heidihoward: Isn't this line the node stepping down?https://github.com/microsoft/CCF/blob/aeddbfb99d42a60fe26658fa12e1001c8d88c6f7/src/consensus/aft/raft.h#L1964

achamayou: Yes, in terms of internal state, you are right. I was thinking in terms of observable behaviour/messages for the other nodes.

achamayou:
> This PR is mostly done minus some tidying up and fixing the ci. The main question I have is that the implementation seems to stop helping with AEs after it believes its retirement is completed, does this not cause a liveness issue if the new leader needs to add a new signature in current term in order to commit the reconfiguration transaction?

That's a good question, and I think the answer is yes, because the new_agreement_index on the old config would never go past the signature immediately following it:

https://github.com/microsoft/CCF/blob/42dc993ddc23596965250334540c7e44e1fc7288/src/consensus/aft/raft.h#L2105

So with:

```
Configuration is [1, 2] Node 1 Primary
2.3 Cfg [2, 3]          Node 1 Primary
2.3 Sig                 Node 1 Primary
2.4 Sig                 Node 2 or 3 Primary
```

Node 2 or 3 will only be able to collect a new_agreement_index of 3 at most on [1, 2], meaning that 2.4 can never get committed.

That should be reproducible in a scenario, the main thing is to trigger an election after the configuration is replicated to both nodes, but before it is committed.

heidihoward:
There's a few issues in the spec still so reverting back to draft whilst I patch

heidihoward:
> > This PR is mostly done minus some tidying up and fixing the ci. The main question I have is that the implementation seems to stop helping with AEs after it believes its retirement is completed, does this not cause a liveness issue if the new leader needs to add a new signature in current term in order to commit the reconfiguration transaction?
> 
> That's a good question, and I think the answer is yes, because the new_agreement_index on the old config would never go past the signature immediately following it:
> 
> https://github.com/microsoft/CCF/blob/42dc993ddc23596965250334540c7e44e1fc7288/src/consensus/aft/raft.h#L2105
> 
> So with:
> 
> ```
> Configuration is [1, 2] Node 1 Primary
> 2.3 Cfg [2, 3]          Node 1 Primary
> 2.4 Sig                 Node 1 Primary
> 3.5 Sig                 Node 2 or 3 Primary
> ```
> 
> Node 2 or 3 will only be able to collect a new_agreement_index of 3 at most on [1, 2], meaning that 2.4 can never get committed.
> 
> That should be reproducible in a scenario, the main thing is to trigger an election after the configuration is replicated to both nodes, but before it is committed.

The reconfiguration needs to be committed so that retiring node believes that its retirement is completed. It just needs to be the case that new leader does not yet know that the retirement was committed

tla/consensus/ccfraft.tla
```diff
@@ -649,13 +713,18 @@ ChangeConfigurationInt(i, newConfiguration) ==
         addedNodes == newConfiguration \ CurrentConfiguration(i)
         newSentIndex == [ k \in Servers |-> IF k \in addedNodes THEN Len(log[i]) ELSE sentIndex[i][k]]
        IN sentIndex' = [sentIndex EXCEPT ![i] = newSentIndex]
-    /\ removedFromConfiguration' = removedFromConfiguration \cup (CurrentConfiguration(i) \ newConfiguration)
+    /\ removedFromConfiguration' = removedFromConfiguration \cup (MaxConfiguration(i) \ newConfiguration)
```
lemmy, line 716: Did this change slip into this PR, or is it relevant for retirement?

tla/consensus/ccfraft.tla
```diff
@@ -848,7 +931,15 @@ ConflictAppendEntriesRequest(i, index, m) ==
        IN /\ log' = [log EXCEPT ![i] = new_log]
           \* Potentially also shorten the configurations if the removed txns contained reconfigurations
           /\ configurations' = [configurations EXCEPT ![i] = ConfigurationsToIndex(i,Len(new_log))]
-    /\ UNCHANGED <<removedFromConfiguration, serverVars, commitIndex, messages, membershipState>>
+          \* Potentially rollback retirement phase
+          /\ IF /\ membershipState[i] \in {RetirementOrdered, RetirementSigned}
+                /\ RetirementIndex(i) > Len(log'[i])
```
lemmy, line 936: Should `RetirementIndex` also be primed to be calculated on the new log?

tla/consensus/ccfraft.tla
```diff
@@ -929,9 +1030,13 @@ HandleAppendEntriesResponse(i, j, m) ==
            \* "If AppendEntries fails because of log inconsistency: decrement nextIndex (aka sentIndex +1) and retry"
           /\ UNCHANGED matchIndex
     /\ Discard(m)
-    /\ UNCHANGED <<reconfigurationVars, serverVars, candidateVars, logVars, membershipState>>
+    /\ UNCHANGED <<reconfigurationVars, serverVars, candidateVars, logVars>>
 
-\* Any RPC with a newer term causes the recipient to advance its term first.
+\* Any message with a newer term causes the recipient to advance its term first.
+\* Note that UpdateTerm does not discard message m from the set of messages so this 
+\* message can parsed again by receiver. Note that all other message parsing actions should
```
lemmy, line 1037: can *be* parsed again *the* receiver.

tla/consensus/ccfraft.tla
```diff
@@ -929,9 +1030,13 @@ HandleAppendEntriesResponse(i, j, m) ==
...
+\* check that m.term = currentTerm[i] to ensure that this action is not skipped.
```
lemmy, line 1038: s/skipped/is the only one ENABLED when ?

tla/consensus/ccfraft.tla
```diff
@@ -1322,6 +1455,15 @@ StateTransitionsProp ==
         /\ leadershipState[i] = Leader => leadershipState[i]' \in {Follower, Leader}
         ]_vars
 
+MembershipStateTransitionsProp ==
+    [][\A i \in Servers :
+        \* RetirementCompleted is the terminal state
+        membershipState[i] = RetirementCompleted 
+        => membershipState[i]' = RetirementCompleted]_vars
+        \* Note that all other transitions between retirement phases are permitted
+        \* For instance, a node could go from NotRetiring to RetirementCompleted in one step if it
```
lemmy, line 1464: "NotRetiring" = "Active"?

lemmy (approved):

tla/consensus/ccfraft.tla
heidihoward: Fixed in https://github.com/microsoft/CCF/pull/5919/commits/9659fe35984411f8cdc075ae80c6a12d7a63df79

tla/consensus/ccfraft.tla
heidihoward: Fixed in https://github.com/microsoft/CCF/pull/5919/commits/9c7329ea95d3f881154abd93a278fe2dad8338d9

tla/consensus/ccfraft.tla
heidihoward: Fixed in https://github.com/microsoft/CCF/pull/5919/commits/e5135f238537660ed9caad1c682b951b4244a236

tla/consensus/ccfraft.tla
heidihoward: Using retirementIndex of log is ok in this very particular case but I agree that it's unclear so I've refactored in https://github.com/microsoft/CCF/pull/5919/commits/c3f7fb3d883e35d97566d57fee5bf6ed7a470d43.

tla/consensus/ccfraft.tla
heidihoward: This is relevant for retirement but I'm particularly happy with it so I'll split out into a separate PR

PR #5921, opened 2024-01-11
Enforce read-only restrictions for JS endpoints marked as `readonly`

eddyashton:
Noticed while looking at an `app.json` sample that we weren't correctly using the `readonly` value. We have read-only tables in _historical_ endpoints, but ignored the mode for normal endpoints.

This is a breaking change, as many existing apps may have misdescribed endpoints which will now throw.


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 [5.0.0-dev12]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.0-dev12
 
+### Changed
+
+- JS endpoints marked as `"mode": "readonly"` are prevented from writing to the KV. Attempting to call `map.set(k, v)`, `map.delete(k)`, or `map.clear()` on any KV table in such an endpoint will now result in an error being thrown.
```
```suggestion
- JS endpoints marked as `"mode": "readonly"` are prevented from writing to the KV. Attempting to call `map.set(k, v)`, `map.delete(k)`, or `map.clear()` on any KV table in such an endpoint will now result in an error being thrown (#5921).
```

src/js/wrap.cpp
```diff
@@ -1725,7 +1757,8 @@ namespace ccf::js
 
     auto& tx = *tx_ptr;
 
-    js::Context ctx2(js::TxAccess::APP);
+    // TODO
```
achamayou, line 1760: ?

achamayou (approved):

src/js/wrap.cpp
eddyashton: Left a few TODOs while I was converting the old enums to the new. In this case (parsing modules to build bytecode) I _think_ it would be safe to have `RO`, or even no exec context at all - surely we do no execution here so the bindings don't matter? But haven't investigated a further refactor to support this, so just leaving as `RW` for now.

PR #5922, opened 2024-01-11
Bias simulation towards to behaviors where followers are more often in sync.

lemmy:
https://github.com/microsoft/CCF/assets/88777/26d303fb-77e4-4643-8cfb-866a245b1c75


achamayou:
@lemmy this change seems to result in many more, but shorter on average, traces:

image

That's compatible with my expectations, are there other ways to quantify coverage that I have missed?

lemmy:
The removal of `ENABLED Forward` in https://github.com/microsoft/CCF/pull/5922/commits/e937c46c05cf7cd45d112ee616090b7118e124ea causes TLC to evaluate the next-state relation much faster, resulting in many more traces being generated. :-)

The reduction of trace lengths by 2 percentage points is likely not statistically significant.

PR #5924, opened 2024-01-12
More trace validation scenarios

achamayou:
Add soft rollback, fancy election 1, suffix collisions 1,2,3 and bad network scenarios, as well as spec fix necessary to align receive append entries response to the one on the implementation.

It is worth nothing that the spec and implementation of CCF differ from Raft on the term value set on receive append entries responses when they fail. @heidihoward spotted this earlier when reviewing the implementation, now is a good time to discuss and resolve the discrepancy:

**Raft**: term of the current, as provided previously on the send append entries.

image

From https://raft.github.io/raft.pdf (wouldn't it be nice if there was a way to link to a section in a paper?).

**CCF implementation**: [term of the last index in the backup's log](https://github.com/microsoft/CCF/blob/e5a7106a03686e5da881c1ec9f182d3a468159de/src/consensus/aft/raft.h#L1348), [same in the spec, with additional prevLogIndex fix in this PR](https://github.com/microsoft/CCF/blob/e5a7106a03686e5da881c1ec9f182d3a468159de/tla/consensus/ccfraft.tla#L778).

### DIFF
--- a/tla/consensus/Traceccfraft.tla
+++ b/tla/consensus/Traceccfraft.tla
@@ -38,10 +38,8 @@ IsAppendEntriesResponse(msg, dst, src, logline) ==
     /\ msg.type = AppendEntriesResponse
     /\ msg.type = RaftMsgType[logline.msg.packet.msg]
     /\ msg.dest   = dst
-    /\ msg.source = src
-    \* raft.h::send_append_entries_response:1348 RAeR FAIL set their term to the term of index
-    \* for the last entry in the backup's log, not the term of the current leader. 
-    \* /\ msg.term = logline.msg.packet.term
+    /\ msg.source = src 
+    /\ msg.term = logline.msg.packet.term
     \* raft_types.h enum AppendEntriesResponseType
     /\ msg.success = (logline.msg.packet.success = "OK")
     /\ msg.lastLogIndex = logline.msg.packet.last_log_idx
--- a/tla/consensus/Traceccfraft.tla
+++ b/tla/consensus/Traceccfraft.tla
@@ -38,7 +38,7 @@ IsAppendEntriesResponse(msg, dst, src, logline) ==
     /\ msg.type = AppendEntriesResponse
     /\ msg.type = RaftMsgType[logline.msg.packet.msg]
     /\ msg.dest   = dst
-    /\ msg.source = src 
+    /\ msg.source = src
     /\ msg.term = logline.msg.packet.term
     \* raft_types.h enum AppendEntriesResponseType
     /\ msg.success = (logline.msg.packet.success = "OK")
--- a/tla/consensus/Traceccfraft.tla
+++ b/tla/consensus/Traceccfraft.tla
@@ -39,7 +39,9 @@ IsAppendEntriesResponse(msg, dst, src, logline) ==
     /\ msg.type = RaftMsgType[logline.msg.packet.msg]
     /\ msg.dest   = dst
     /\ msg.source = src
-    /\ msg.term = logline.msg.packet.term
+    \* raft.h::send_append_entries_response:1348 RAeR FAIL set their term to the term of index
+    \* for the last entry in the batch, not the term of the current leader. 
+    \* /\ msg.term = logline.msg.packet.term
     \* raft_types.h enum AppendEntriesResponseType
     /\ msg.success = (logline.msg.packet.success = "OK")
     /\ msg.lastLogIndex = logline.msg.packet.last_log_idx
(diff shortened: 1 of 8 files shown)


lemmy:
/me is eagerly awaiting a new hope scenario.

eddyashton:
The term we use on NACKs is mentioned briefly in the Raft paper, towards the end of section 5.3 (bottom of page 7):

> If desired, the protocol can be optimized to reduce the
number of rejected AppendEntries RPCs. For example,
when rejecting an AppendEntries request, the follower
can include the term of the conflicting entry and the first
index it stores for that term. With this information, the
leader can decrement nextIndex to bypass all of the con-
flicting entries in that term; one AppendEntries RPC will
be required for each term with conflicting entries, rather
than one RPC per entry. In practice, we doubt this opti-
mization is necessary, since failures happen infrequently
and it is unlikely that there will be many inconsistent en-
tries.

We have taken "can include" to mean "can replace the existing term and index fields", but the discrepancy isn't a pure implementation invention.

lemmy:
> In practice, we doubt this optimization is necessary, since failures happen infrequently
and it is unlikely that there will be many inconsistent entries.

@eddyashton Could you please clarify if there is a difference between vanilla raft and CCF that necessitates this optimization?  It could be interesting to quantify the optimization's impact by collecting statistical properties at the spec.

eddyashton:
> > In practice, we doubt this optimization is necessary, since failures happen infrequently
> > and it is unlikely that there will be many inconsistent entries.
> 
> @eddyashton Could you please clarify if there is a difference between vanilla raft and CCF that necessitates this optimization? It could be interesting to quantify the optimization's impact by collecting statistical properties at the spec.

We don't have any hard numbers on this, and also believe failures producing inconsistent logs will be rare in practice. However, the conditions for this failure are reasonably simple - a primary who is temporarily partitioned from other nodes but still receiving writes can produce an arbitrarily long conflicting suffix - so this could be seen in production systems. The non-optimised resolution for this is linear in the length of the conflicting suffix, and each step requires a network roundtrip - this is plausibly a user-visible delay. The optimised version is instead linear in the number of conflicting _terms_. We don't see an operational scenario where this grows indefinitely - it's bound by the election timer, so either requires a long period of pessimally-repartitioned nodes, or skewed clocks mimicking the same.

It would be interesting to model both of these strategies to get statistical data, but we already have an approximation of this in our unit tests - [an assertion that our `O(term)` process has taken `<= terms` steps](https://github.com/microsoft/CCF/blob/main/src/consensus/aft/test/committable_suffix.cpp#L847). Older versions of the test show [remnants of the working that got here](https://github.com/microsoft/CCF/blob/ed91bd6cef2cfce96b6d33984d13fccdb8a66e61/src/consensus/aft/test/committable_suffix.cpp#L767-L791), initially asserting that we were bound by the log length (when we had the non-optimised version). The test also tries to construct a pessimal interleaving of steps, and confirms this takes `terms` rather than `<= terms` steps, though there's no formal proof that this is actually pessimal - if it turns out to be unsound there's likely something worse!

achamayou:
Following discussion with @heidihoward and @eddyashton, I'm changing this to a PR because:

1. It aligns implementation and spec
2. It does not cause any breaks in MC, Sim, TV or tests

I have opened #5927 for further discussion, along with a summary of our thoughts so far.

tla/consensus/ccfraft.tla
```diff
@@ -788,18 +788,20 @@ RejectAppendEntriesRequest(i, j, m, logOk) ==
        \/ /\ m.term >= currentTerm[i]
           /\ leadershipState[i] = Follower
           /\ ~logOk
-          /\ LET prevTerm == IF m.prevLogIndex = 0 THEN StartTerm
+          \* raft.h::send_append_entries_response:1348 RAeR FAIL set their term to the term of index
```
heidihoward, line 791: I wasn't sure what RAeR stood for

heidihoward (approved):
It's great to see more scenarios passing!

tla/consensus/ccfraft.tla
eddyashton:
```suggestion
          \* raft.h::send_append_entries_response:1348 AppendEntriesResponse messages with answer == FAIL set their term to the term of index
```

I'm also unsure about this - I thought RAeR was "Receive Append Entries Request", which is the top-level action we're doing here, but maybe it's actually "_Reject_ Append Entries Request", which is the action we're currently in? But then in other places I thought we used AE for AppendEntries, distinct from AER for AppendEntriesResponse, so maybe AeR isn't Append Entries Request here?

tla/consensus/Traceccfraft.tla
```diff
@@ -38,7 +38,7 @@ IsAppendEntriesResponse(msg, dst, src, logline) ==
     /\ msg.type = AppendEntriesResponse
     /\ msg.type = RaftMsgType[logline.msg.packet.msg]
     /\ msg.dest   = dst
-    /\ msg.source = src
+    /\ msg.source = src
```
eddyashton, line 41:
```suggestion
    /\ msg.source = src
```

Issue #5676, opened 2023-09-21
Message reordering breaks raft.h is_new_follower logic


lemmy:
The introduction of the `is_new_follower` flag in #2189 has led to a behavior change in which a follower will truncate its log upon receiving its first AppendEntries message after transitioning to a follower state:

https://github.com/microsoft/CCF/blob/d6efe6664045968dc4f191b9ae672e686f05279b/src/consensus/aft/raft.h#L126-L135

However, this change is vulnerable to message reordering when AppendEntries messages do not arrive in order. This vulnerability can result in two potential issues:

In cases where the AppendEntries messages are not processed in the correct order, the follower's log may not be truncated adequately. This can lead to a violation of liveness.  Alternatively, in situations where the log is truncated too aggressively, it may cause a safety violation. Could this issue be related to https://github.com/microsoft/CCF/issues/5674?!

[Message omission](https://github.com/microsoft/CCF/issues/2577), if implemented, will also mess with `is_new_follower`.

Set milestone to "TLA+ Trace Validation" it was found as part of partially validating a scenario generated by `raft_scenario_gen.py`.  The scenario fails validation because ccfraft.tla does *not* model `is_new_follower`.

eddyashton:
After further investigation, and discussion with @achamayou, we think this is not an issue for the implementation for 2 reasons:
- We don't have message reordering of messages between 2 nodes, due to the design of the node-to-node channels. We may have omissions. but not shuffles. This makes the optimistic `next_idx` advancement safe - a primary will only change what it sends to the backup after receiving a response, similar to in classic Raft.
- This rollback behaviour doesn't happen when processing all AEs, but only when we've processed a _matching_ AE (ie - one that extends our log, that we're going to ACK). The liveness issue of not truncating enough is thus not possible - "not enough" means you retain a non-matching suffix, so you will NACK this AE. For the safety issue, the primary would need to emit an AE that started _before_ the end of the receiver's matching suffix - the `find_highest_possible_match` probing procedure is believed to be safe in this regard, and will not produce a too-low AE message.

lemmy:
Do we want to align the spec by changing the spec to ordered message delivery?

https://github.com/microsoft/CCF/blob/d545a6af5617535ee1e9bf79323ca240bf7f40ef/tla/Network.tla#L66-L84

achamayou:
@lemmy yes, although there can definitely be gaps/messages missing, due to the send pipelining.

lemmy:
[...]

PR #5925, opened 2024-01-12
Add fancy election 2 scenario

achamayou:
This scenario failed to match the spec, because it makes use of the [rollback logic gated by `is_new_follower`](https://github.com/microsoft/CCF/blob/e5a7106a03686e5da881c1ec9f182d3a468159de/src/consensus/aft/raft.h#L1127), which isn't present in the spec's [`AppendEntriesAlreadyDone`](https://github.com/microsoft/CCF/blob/e5a7106a03686e5da881c1ec9f182d3a468159de/tla/consensus/ccfraft.tla#L819). As a result, the trace contains a (heartbeat) SAeR SUCCESS that truncates the log, something the spec has no provision for:

```
[ 864] 0👑        2.14 12     1💂RAe     3.14 12     2💂        3.12 10     3💂        3.12 10     4👑        3.12 10
[ 873] 0👑        2.14 12     1💂SAeR ✅ 3.12 12     2💂        3.12 10     3💂        3.12 10     4👑        3.12 10
```

Fixed by @heidihoward's change in #5971.


lemmy:
`is_new_follower` came up in https://github.com/microsoft/CCF/issues/5676 while working on trace validation.

achamayou:
@lemmy thank you, that's useful context. This is not thought to cause a problem at the moment, but we would like to simplify the logic rather than make the spec more complex. It needs more thinking.

eddyashton:
I agree we should sync `is_new_follower` between spec and impl (add it to the spec, or remove it from the impl), but it's possible that it's not the source of the discrepancy here. Any conflicts go through the `ConflictAppendEntriesRequest` action `cdot`'d with `AppendEntriesAlreadyDone`/`NoConflictAppendEntriesRequest`, and _that_ has a `/\ m.entries /= << >>` conjunct that prevents it activating on empty heartbeats. Removing that might lead to a trace match for this scenario?

Separate thoughts on whether this behaviour (rolling back on a mere heartbeat) is OK:
- It _feels_ safe. Although you could hold onto this suffix and potentially win an election and commit it later (if nothing conflicting happens before then), at this point there is a primary who is able to produce a signature and tell you to drop this. The gap between the 2 is indistinguishable for the receiver, and in practice is just based on close timings.
- For liveness, I don't think there's a problem, but maybe we'd want some invariant around persisting as much possible for as long as possible? eg - a client's state (even if not committed) is available as long as it is committable, and only uncommittable when there is a nameable conflict.

heidihoward (approved):

PR #5929, opened 2024-01-15
Add election_while_reconfiguration scenario

achamayou:
Two changes:

1. Updates the election while reconfiguration scenario to use new bootstrap, and to avoid suffix rollback
2. Update vote evaluation logic in spec to accept supersets of configuration quorums

Note: the previous implementation of the scenario was incorrect, leading to a situation where a primary would roll back its configuration. We should add this as a separate scenario.

Other note: Narrowing the scenario down to check the new node wins the election with only the votes of the old config would make a slightly stricter test.


heidihoward (approved):

tla/consensus/ccfraft.tla
```diff
@@ -584,7 +584,7 @@ AppendEntries(i, j) ==
 BecomeLeader(i) ==
     /\ leadershipState[i] = Candidate
     \* To become leader, the candidate must have received votes from a majority in each active configuration
-    /\ \A c \in DOMAIN configurations[i] : votesGranted[i] \in Quorums[configurations[i][c]]
+    /\ \A c \in DOMAIN configurations[i] : (votesGranted[i] \intersect configurations[i][c]) \in Quorums[configurations[i][c]]
```
achamayou, line 587:
```suggestion
    \* Only votes by nodes part of a given configuration should be tallied against it
    /\ \A c \in DOMAIN configurations[i] : (votesGranted[i] \intersect configurations[i][c]) \in Quorums[configurations[i][c]]
```

achamayou:
Annoyingly, this change is causing quite an increase in state and runtime for reconfig (2x) and atomic reconfig (4x).

lemmy (approved):

lemmy:
Can somebody explain why it is safe to tally all nodes, i.e., why doesn't this violate any properties?

achamayou:
> Can somebody explain why it is safe to tally all nodes, i.e., why doesn't this violate any properties?

How do you mean? A candidate (potentially) receives votes from all the nodes it knows about, regardless of the configuration(s) they belong to. The tally is then done [for each configuration](https://github.com/microsoft/CCF/blob/38b054853b013307c4701dc05253881b885d2b03/src/consensus/aft/raft.h#L2047), and a majority is needed in every active configuration for the vote to succeed. I am not sure what properties this could violate?

lemmy:
Let me rephrase, what's the source of the state-space explosion you mention above?

achamayou:
If you mean the information itself, it's from cimetrics:
image

Why fixing the quorum evaluation logic caused this, I don't know for sure. It clearly would have resulted in many elections taking place during a reconfiguration failing incorrectly (as soon as at least one node not in all config sends a vote), so I guess that's a lot of excluded behaviours for each election with more than one config active. Since the atomic reconfig MC does one more term/election than the reconfig MC, I'd then expect it to grow by more, which seems to be the case.

eddyashton (approved):

PR #5930, opened 2024-01-16
SIM Trace viz

achamayou:
Loosely adapted from `trace_viz.py` (CCF raft traces), missing some functionality, but hopefully useful to investigate https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=80403&view=logs&j=ef8c4b80-2ff7-5042-ecea-fb36384f1d0d&t=c2a7e65a-2f43-5dbb-c6d6-70df1fdcc7fc and other SIM failures.

Sample:

image

Still to be done:

- [x] Success/Fail markers for AE and votes
- [x] Signatures
- [x] Skip actions that have no impl equivalent, such as SignCommittableMessages, SIMTimeout, RcvUpdateTerm and SIMCheckQuorum

Note: because SIM/MC skips startup and instead primes the logs and states, the trace doesn't quite start "at the beginning", compared to implementation traces. h_ts is replaced with the action index in the trace.


eddyashton (approved):

eddyashton:
This is great! I've been bouncing between Python and jq trying to extract anything meaningful from the JSON trace, but I think this visualisation already shows me more.

tests/sim_trace_viz.py
```diff
@@ -0,0 +1,129 @@
...
+        # TODO: need success, vote granted and globally committable
+        # to be extracted from pre or post state
+        states = [
+            (
+                extract_node_state(post[1], node),
+                action["name"] if node == node_id else None,
+                extract_node_state(pre[1], node) if node == node_id else None,
+                " " # tag
+            )
+            for node in nodes
```
eddyashton, line 117:
```suggestion
            for node in sorted(nodes)
```
image

rather than
image

achamayou: Good point, in the impl one, the order deliberately is order of appearance in the trace, but here relying on the names is preferable.

lemmy:
> This is great! I've been bouncing between Python and jq trying to extract anything meaningful from the JSON trace, but I think this visualisation already shows me more.

Have you tried TLC's trace expression evaluation?  If yes, why do you prefer json queries?

eddyashton:
> > This is great! I've been bouncing between Python and jq trying to extract anything meaningful from the JSON trace, but I think this visualisation already shows me more.
> 
> Have you tried TLC's trace expression evaluation? If yes, why do you prefer json queries?

I've not tried TLC's trace expressions.

Even if I was familiar with them, I may have gone with JSON queries in this case. To debug this failure, my first choice is between working with a `.trace.tla` file, and a `.json` file. I already know how to work with JSON files, even ones this large - I can format and fold them in my IDE, I can filter and project them with `jq`, I can manipulate and examine them as Python objects. So I'm more productive if I leverage existing tools and known workflows.

achamayou:
The code needs tidying up, and ideally merging with trace_viz, but should not print equivalent traces, except for the startup quirk mentioned in the description.

tests/sim_trace_viz.py
```diff
@@ -0,0 +1,164 @@
...
+    dcfg.commit = digits(max_commit)
+    dcfg.ts = digits(max_ts)
+    ts = 0
+    for pre, action, post in entries:
+        ts += 1
+        if action["name"] in {"SIMTimeout", "SIMCheckQuorum", "UpdateTerm"}:
+            continue
+        node_id = action["context"]["i"]
+
+        def last_recvd():
```
achamayou, line 131: @lemmy as discussed, it would be great if there was a way to easily add the result of some expressions to the trace, in this case the last received message.

lemmy: Do we need support for action-level or just state-level formulas.

achamayou: Action-level seems preferable here, since the goal is to report the message(s) being consumed in the action. The way the script does it is by diffing the messages queues from the pre/post states, so depending on what the formula has access to at the state level, that may be fine too.

lemmy: The feature should be ready in a few days.

achamayou: Thank you! I think that will make the traces much easier to read, both in TLA directly (where looking up the details of the messages consumed in the action is not straightforward now), and for tools like this one.

achamayou: @lemmy yes but the output of ALIAS doesn't make it to the JSON trace produced by `-dumpTrace json file.json` as far as I can tell from a quick test with the DebugAlias on the SIM.

lemmy: Yes, that the part I'm working on. My confusion is around
> ...in TLA directly (where looking up the details of the messages consumed in the action is not straightforward now

achamayou: Sorry, I wasn't clear, that referred to adding an alias that would diff the messages, which I don't think we have yet. But that's something I can do already, you are right.

lemmy: Done: https://github.com/tlaplus/tlaplus/pull/865

PR #5933, opened 2024-01-18
[Draft] Test some invariants during trace checking

eddyashton:
I've constructed a new scenario where `MoreUpToDateCorrectInv` doesn't hold, so wanted to check what invariants hold on our existing scenarios.

### DIFF
--- a/tla/consensus/Traceccfraft.cfg
+++ b/tla/consensus/Traceccfraft.cfg
@@ -28,6 +28,16 @@ PROPERTIES
 INVARIANT 
     TraceDifferentialInv
 
+    LogInv
+    MoreThanOneLeaderInv
+    CandidateTermNotInLogInv
+    ElectionSafetyInv
+    LogMatchingInv
+    QuorumLogInv
+    MoreUpToDateCorrectInv
+    LeaderCompletenessInv
+    SignatureInv
+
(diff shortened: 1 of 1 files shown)


lemmy:
Unfortunately, this might produce spurious invariant violation unless trace validation is completely deterministic, i.e., its set of traces contains only traces of length equal to `Len(Trace)`.   Otherwise, a state in a trace with length less than `Len(Trace)` might violate any of the invariants, yet this trace is an artifact of the non-determinism in ccfraft/traceccfraft.  

At the technical level, TLC's trace validation checks invariants during the reachability stage.  We would have to defer checking the invariants until after reachability.

achamayou:
> We would have to defer checking the invariants until after reachability.

Is that feasible?

lemmy:
> > We would have to defer checking the invariants until after reachability.
> 
> Is that feasible?

I have a few ideas, but nothing that will work right away.

In the meantime, TLC could at least write to stdout for manual inspection upon a violation of an invariant. Alternatively, it could log into a `TLCSet/TLCGet` register, and we process that in a postcondition.  For example, we could assert that the known violations remain stable.

eddyashton:
Closing as currently unsafe to do.

PR #5934, opened 2024-01-18
MoreUpToDateCorrectInv is not an invariant, move to Debug

achamayou:
After investigation of the failure raised in https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=80403&view=logs&j=ef8c4b80-2ff7-5042-ecea-fb36384f1d0d&t=c2a7e65a-2f43-5dbb-c6d6-70df1fdcc7fc, it has become clear that MoreUpToDateCorrectInv does not hold, and that valid traces can break it.

[SIMccfraft.json](https://github.com/microsoft/CCF/files/13975430/SIMccfraft.json)

### DIFF
--- a/tla/consensus/MCccfraft.cfg
+++ b/tla/consensus/MCccfraft.cfg
@@ -71,7 +71,6 @@ INVARIANTS
     ElectionSafetyInv
     LogMatchingInv
     QuorumLogInv
-    MoreUpToDateCorrectInv
     LeaderCompletenessInv
     SignatureInv
     TypeInv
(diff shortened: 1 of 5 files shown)


eddyashton (approved):

heidihoward (approved):

lemmy:
Nitpicking:  The `Debug` prefix convention if fine, but the philosophically "correct" way is to write:

```tla
SomeInvariant ==
     ...

THEOREM Spec => []SomeInvariant
``` 

The THEOREM line is where a proof would start.  For example:

https://github.com/tlaplus/Examples/blob/e4b3677cc1fa76db40f45793546f5992154d2a6c/specifications/ewd998/AsyncTerminationDetection_proof.tla#L49

achamayou:
@lemmy thank you for pointing this out, in this case I was only following convention in our specs. I suspect we will probably end up deleting this particular invariant, because it just isn't correct, but that seems like it's best done in a single pass after we're done matching the impl and the spec rather than now.

PR #5939, opened 2024-01-18
Attempt to fix the simulation

achamayou:
With #5922, the simulation has much better coverage, and is now hitting invariants that are not clearly useful as we try to reconcile the impl and the spec. This is an attempt to drop enough to get back to a working CI without making unreasonable changes.


tla/consensus/SIMccfraft.cfg
```diff
@@ -72,8 +72,9 @@ INVARIANTS
     CandidateTermNotInLogInv
     ElectionSafetyInv
     LogMatchingInv
-    QuorumLogInv
-    LeaderCompletenessInv
+    \* Disabled until retirement is modeled correctly in the spec
```
lemmy, line 75: Leaving this here as an alternative to disabling invariants completely if it is possible to find a predicate `P` that equals false when the invariant is expected not to hold.  Then, in ccfraft:

```tla
QuorumLogInv == P => OriginalQuorumLogInv
```

achamayou: It is likely that the QuorumLogInv can and should be reinstated with a change to account for retired nodes.

eddyashton (approved):

lemmy:
It is probably time to revert this fix (the part that disables the properties).

lemmy:
https://github.com/microsoft/CCF/pull/6501

PR #5938, opened 2024-01-18
[WIP] Add a `reelection` scenario

eddyashton:
Trying to debug an issue with trace validation, and adding this simple scenario which demonstrates the problem.

When a node becomes primary again, it should process (and discard) AER ACKs from previous terms. In this case, it doesn't, and I believe that's because it has unnecessarily dropped pending messages, since messages are ordered only by dest and not by src.

```
State 76: <Action line 385, col 22 to line 385, col 91 of module Traceccfraft>

/\ messages = ( "0" :>
      << [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "1",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "1",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> RequestVoteResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 3,
           voteGranted |-> TRUE ],
         [ type |-> RequestVoteResponse,
           dest |-> "0",
           source |-> "1",
           term |-> 3,
           voteGranted |-> TRUE ] >> @@
  "1" :> <<>> @@
  "2" :> <<>> )
/\ leadershipState = ("0" :> Candidate @@ "1" :> Follower @@ "2" :> Follower)

State 77: <Action line 385, col 22 to line 385, col 91 of module Traceccfraft>
/\ messages = ( "0" :>
      << [ type |-> RequestVoteResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 3,
           voteGranted |-> TRUE ] >> @@
  "1" :> <<>> @@
  "2" :> <<>> )
/\ leadershipState = ("0" :> Candidate @@ "1" :> Follower @@ "2" :> Follower)
```

To process this RVR from 2 to 0 we have dropped several messages, including some from 1 to 0. The latter should not have been touched.

Issue #5927, opened 2024-01-15
CCF Raft behaviour when setting the term on NACKs differs from Raft


achamayou:
This was known for some time, spotted by @heidihoward earlier when reviewing the implementation, and brought to light again by #5924. It is not believed to be an issue, at the time this is being written, but it does warrant extra scrutiny.

**Raft**: term of the current, as provided previously on the send append entries.

![image](https://github.com/microsoft/CCF/assets/4016369/59b105e2-ee40-4817-8dc6-98bec922214b)

From https://raft.github.io/raft.pdf (wouldn't it be nice if there was a way to link to a section in a paper?).

**CCF implementation**: [term of the last index in the backup's log](https://github.com/microsoft/CCF/blob/e5a7106a03686e5da881c1ec9f182d3a468159de/src/consensus/aft/raft.h#L1348), [same in the spec, with additional prevLogIndex fix in this PR](https://github.com/microsoft/CCF/blob/e5a7106a03686e5da881c1ec9f182d3a468159de/tla/consensus/ccfraft.tla#L778).

The questions we would like to resolve are:

1. Does this affect safety?
2. Does this affect liveness?

The existing verification work suggests that the answer to 1. may be "no", but 2. remains more open.

achamayou:
Following discussion with @heidihoward @eddyashton @lemmy, we will:

1. Update the spec to not drop those messages when they are stale
2. Look at adding stats and nack injection in the sim, to better understand if this may cause a liveness issue. This is lower priority.

lemmy:
> Following discussion with @heidihoward @eddyashton @lemmy @nacrooks, we will:
> 
> 2. Look at adding stats and nack injection in the sim, to better understand if this may cause a liveness issue. This is lower priority.

For future reference, https://github.com/microsoft/CCF/blob/4d6f8e2dfa407c2d27a2be8840ddf11ea58f1d71/tla/consensus/SIMCoverageccfraft.tla#L49-L64 shows how to collect statistics for a spec.  A more complex example is at https://github.com/tlaplus/Examples/tree/master/specifications/ewd998#statistics.

achamayou:
This was found to be a potential safety issue, details captured in #5991, investigated further in #6001, and addressed in #6016.

The issue was resolved by removing the unnecessarily aggressive rollback behaviour, rather aligning the NACK term value with that in Raft.

PR #5940, opened 2024-01-18
Model point-to-point message channels in TLA

eddyashton:
The spec previously stored messages purely by destination (`messages : Map[DestNodeID, List<Message>]`). This is fine for most uses, for but trace validation means that we fail to match scenarios like reelection (proposed in #5938).

I think the solution is to model point-to-point messages (`messages : Map[SourceNodeID, Map[DestNodeID, List<Message>]]`), such that we can drop messages from A to B without affecting anything from or to C.


tla/consensus/Network.tla
```diff
@@ -78,28 +78,28 @@ LOCAL ReorderNoDupDropMessages ==
...
-    [ msgs EXCEPT ![m.dest] = Append(@, m) ]
+    [ msgs EXCEPT ![m.source][m.dest] = Append(@, m) ]
 
 LOCAL OrderWithoutMessage(m, msgs) ==
-    [ msgs EXCEPT ![m.dest] = RemoveFirst(@, m) ]
+    [ msgs EXCEPT ![m.source][m.dest] = RemoveFirst(@, m) ]
 
 LOCAL OrderMessages ==
-    UNION { Range(messages[s]) : s \in Servers }
+    UNION { UNION { Range(messages[src][dst]) : dst \in Servers } : src \in Servers }
```
eddyashton, line 90: @lemmy The rest of these changes are reasonably mechanical, so I assume if one is right then they're all right. This change, and `OrderedMessagesTo`, are more involved - can you check that they're doing what was originally intended?

tla/consensus/Traceccfraft.tla
```diff
@@ -130,7 +130,11 @@ DropMessages ==
     /\ l \in 1..Len(TraceLog)
     /\ UNCHANGED <<reconfigurationVars, serverVars, candidateVars, leaderVars, logVars>>
     /\ UNCHANGED <<l, ts>>
-    /\ Network!DropMessages(logline.msg.state.node_id)
+    \* Only drop messages when processing message events
+    /\ \/ /\ "from_node_id" \in DOMAIN logline.msg
+          /\ Network!DropMessagesTo(logline.msg.state.node_id, logline.msg.from_node_id)
+       \/ /\ "from_node_id" \notin DOMAIN logline.msg
+          /\ UNCHANGED <<messageVars>>
```
eddyashton, line 137: I'm not sure if this is the correct place to make this distinction? I _think_ we only want to drop messages (during trace validation) when we're trying to match a "message sending" step from the logs, where `from_node_id` is present? But I could be wrong.

tla/consensus/ccfraft.tla
```diff
@@ -941,6 +941,9 @@ HandleAppendEntriesResponse(i, j, m) ==
           \* UNCHANGED matchIndex is implied by the following statement in figure 2, page 4 in the raft paper:
            \* "If AppendEntries fails because of log inconsistency: decrement nextIndex (aka sentIndex +1) and retry"
           /\ UNCHANGED matchIndex
+       \/ /\ m.term < currentTerm[i]
+          /\ m.success
+          /\ UNCHANGED <<matchIndex, sentIndex>>
```
eddyashton, line 946: @lemmy @heidihoward When we're matching a `recv_append_entries_response` line, we have to match the case where it's an ACK (success == true), but from an old term, so ignored/discarded.

I've stuffed it in here, to benefit from the surrounding `UNCHANGED`, rather than adding a `\/ Discard(m) /\ UNCHANGED << >>` to `Traceccfraft`'s `IsRcvAppendEntriesResponse`. I'm sure this isn't the right place, but it seems to work.

How does `IsRcvRequestVoteResponse`, or indeed any other recv match, handle this?

(NB: This is almost exactly what we're discussing in #5927 - AERs get special-case handling of their terms)

lemmy:
> The spec previously stored messages purely by destination (`messages : Map[DestNodeID, List<Message>]`). This is fine for most uses, for but trace validation means that we fail to match scenarios like reelection (proposed in #5938).

Why does reelection fail to match with the current p2p model?

lemmy:
> > The spec previously stored messages purely by destination (`messages : Map[DestNodeID, List<Message>]`). This is fine for most uses, for but trace validation means that we fail to match scenarios like reelection (proposed in #5938).
> 
> Why does reelection fail to match with the current p2p model?

Does `reelection` validate with `SubSeq(messages[server])` on this line https://github.com/microsoft/CCF/blob/185d85cbc26c3681c910a791340588a9652140e3/tla/consensus/Network.tla#L101

eddyashton:
> Does `reelection` validate with SubSeqs

No, it still fails at the same point (last matching step was `h_ts |-> "324"`). I haven't looked at the trace too closely to confirm its exactly the same failure, but I think it is.

> Why does reelection fail to match with the current p2p model?

My understanding of `Traceccfraft` is that to match a message that is not the first message to that node, we have to drop all prior messages to that node (in `DropAndReceive`). We're respecting ordered delivery of the channels, but conflating the order of messages from multiple senders.

Attaching the relevant files:
[reelection.ndjson.txt](https://github.com/microsoft/CCF/files/13987184/reelection.ndjson.txt)
[reelection.trace.json](https://github.com/microsoft/CCF/files/13987185/reelection.trace.json)
[Traceccfraft_TTrace_1705656958.tla.txt](https://github.com/microsoft/CCF/files/13987187/Traceccfraft_TTrace_1705656958.tla.txt)

The first one we fail to match is:
```
{"h_ts": "325", "msg": { "from_node_id": "2", "function": "recv_append_entries_response", ... }}
```

And sure enough, that message isn't in the final state:
```
/\ messages = ( "0" :>
      << [ type |-> RequestVoteResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 3,
           voteGranted |-> TRUE ] >> @@
  "1" :>
      << [ type |-> AppendEntriesRequest,
           dest |-> "1",
           source |-> "0",
           term |-> 3,
           commitIndex |-> 4,
           prevLogTerm |-> 2,
           entries |-> <<>>,
           prevLogIndex |-> 6 ] >> @@
  "2" :>
      << [ type |-> AppendEntriesRequest,
           dest |-> "2",
           source |-> "0",
           term |-> 3,
           commitIndex |-> 4,
           prevLogTerm |-> 2,
           entries |-> <<>>,
           prevLogIndex |-> 6 ] >> )
```

But this message did exist a few states ago, along with others:
```
State 76:
...
/\ messages = ( "0" :>
      << [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "1",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> AppendEntriesResponse,
           dest |-> "0",
           source |-> "1",
           term |-> 2,
           success |-> TRUE,
           lastLogIndex |-> 6 ],
         [ type |-> RequestVoteResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 3,
           voteGranted |-> TRUE ],
         [ type |-> RequestVoteResponse,
           dest |-> "0",
           source |-> "1",
           term |-> 3,
           voteGranted |-> TRUE ] >> @@
  "1" :> <<>> @@
  "2" :> <<>> )
/\ _logline = [...h_ts |-> "307"...]
```

Why did we lose so many messages to produce the next state?
```
State 77: 
...
/\ messages = ( "0" :>
      << [ type |-> RequestVoteResponse,
           dest |-> "0",
           source |-> "2",
           term |-> 3,
           voteGranted |-> TRUE ] >> @@
  "1" :> <<>> @@
  "2" :> <<>> )
/\ _logline = [...h_ts |-> "312"...]
```

My understanding is that to match this 1->0 RVR message, the drop function has had to drop a contiguous sequence of messages to `0`, which includes not just the 1->0 AERs (valid to drop), but also the 2->0 AERs (which should have been unaffected).

tla/consensus/ccfraft.tla
achamayou: I have tried to make a similar change in #5941, but only for the spec. I don't think I understand how this works, because it seem to me that the level above (RcvAppendEntriesResponse) still allows non-deterministic drops, and so will allow traces that drop stale NACKs.

achamayou (approved):

tla/consensus/ccfraft.tla
eddyashton: (I think, with caveats*) Non-deterministic drops are still allowed - they produce a state equivalent to if the message had been dropped in the network, albeit partway through executing an action. A change in `RcvAppendEntriesResponse` doesn't help us with trace validation, because the failure here is via `IsRcvAppendEntriesResponse`, which calls `HandleAppendEntriesResponse` directly.

So either we need to catch it here which is considered by both paths, or maybe the change you've made in `RcvAppendEntriesResponse` could be in both `RcvAppendEntriesResponse` and (the similarly structured) `IsRcvAppendEntriesResponse`? I can't parse what the semantic difference is, or which is preferable. The trace spec also has `RcvAppendEntriesResponse` in its `ComposedNext`, but I don't understand when that is called.

achamayou: > Non-deterministic drops are still allowed - they produce a state equivalent to if the message had been dropped in the network, albeit partway through executing an action

I do agree, but I think it's clearer for the network-induced dropping to be done in a single explicit location (DropMessagesTo). Here we are effectively allowing one special case of dropping and bundling it with a real implementation behaviour (dropping stale ACKs) under "DropStaleResponses", which don't accurately describe the reason we're dropping the message in the NACK cases.

Yes I also think doing it one level above means duplicating it, because the trace has somewhat different logic for handling append entries unfortunately.

Anyway, I think I'm happy with either, given a comment.

tla/consensus/Traceccfraft.tla
```diff
@@ -130,7 +130,11 @@ DropMessages ==
     /\ l \in 1..Len(TraceLog)
     /\ UNCHANGED <<reconfigurationVars, serverVars, candidateVars, leaderVars, logVars>>
     /\ UNCHANGED <<l, ts>>
-    /\ Network!DropMessages(logline.msg.state.node_id)
+    \* Only drop messages when processing message events
```
lemmy, line 133: Because messages loss is beyond our system and comes from the system's environment (network), my motivation was to leave message loss as much as possible to non-determinism, and not restrict message loss to certain actions or node pairs.  Obviously, that would imply that the `DropMessagesTo` operator has no parameters, not even `server`. At that time, however, that resulted in massive state-space explosion beyond what TLC could handle, but, the subsequently added `ENABLED` construct could be good enough to alleviate the problem.

lemmy:
Leaving it here because if found it useful to understand the effects of changes.

1) Run TLC with some spec changes with `-dumptrace TLCplain trace.tla` and `CHECK_DEADLOCK TRUE` (to trigger a counterexample)
2) (Add `EXTENDS ccfraft` at the top of generated trace.tla`)
3) Undo your changes
4) Define the following ALIAS:
```tla
TraceAlias ==
    LET t == INSTANCE trace d == t!trace[l] IN
    DebugAlias @@
    [
        here |-> Network!Messages \ UNION { UNION { Range(d.messages[src][dst]) : dst \in Servers } : src \in Servers },
        there |-> UNION { UNION { Range(d.messages[src][dst]) : dst \in Servers } : src \in Servers } \ Network!Messages,
        ...
    ]
```
5) `here` and `there` show the different variable values in this and the previous run.

-----

This is related to: https://github.com/microsoft/CCF/blob/743f7dbec750676b1dd73b5c2627ef591704cd6c/tla/consensus/Traceccfraft.tla#L419-L441 
However, an invariant will be [spuriously violated](https://github.com/microsoft/CCF/pull/5933#issuecomment-1899025526).

lemmy:
> > Does `reelection` validate with SubSeqs
> 
> No, it still fails at the same point (last matching step was `h_ts |-> "324"`). I haven't looked at the trace too closely to confirm its exactly the same failure, but I think it is.

My bad, I didn't read the definition of `SubSeqs` careful enough.  `reelection` is validating with the new [`AllSubSeqs`](https://github.com/tlaplus/CommunityModules/blob/e6a395429682a90f6b2f36541d08faa34ef3394c/modules/SequencesExt.tla#L398-L412) (and the newly added disjunct in `HandleAppendEntriesResponse`).

```diff
diff --git a/tla/consensus/Network.tla b/tla/consensus/Network.tla
index cf1162c6a..f290a18c5 100644
--- a/tla/consensus/Network.tla
+++ b/tla/consensus/Network.tla
@@ -98,7 +98,7 @@ LOCAL OrderOneMoreMessage(m) ==
     \/ Len(SelectSeq(messages[m.dest], LAMBDA e: m = e)) < Len(SelectSeq(messages'[m.dest], LAMBDA e: m = e))
 
 LOCAL OrderDropMessages(server) ==
-    \E s \in Suffixes(messages[server]):  \* TODO - Change to SubSeqs if more sophisticated message loss is needed.
+    \E s \in AllSubSeqs(messages[server]):
         messages' = [ messages EXCEPT ![server] = s ]
 
 ----------------------------------------------------------------------------------
```

PR #5942, opened 2024-01-19
Derive committable indices from log

achamayou:
Remove committableIndices as separate state in the spec, and derive it from the log instead.

This simplifies the spec a little, and side-steps debugging a SIM trace produced in #5939 suggesting that the spec was resetting committable indices too aggressively.


tla/consensus/ccfraft.tla
```diff
@@ -1274,11 +1260,14 @@ MatchIndexLowerBoundSentIndexInv ==
...
-        committableIndices[i] # {} => commitIndex[i] < Min(committableIndices[i])
+        \/
+            /\ commitIndex[i] = 0
+            /\ CommittableIndices(i) = {}
+        \/ commitIndex[i] \in CommittableIndices(i)
 
 CommittableIndicesAreKnownSignaturesInv ==
     \A i \in Servers :
-        \A j \in committableIndices[i] :
+        \A j \in CommittableIndices(i) :
```
eddyashton, line 1270: This invariant could go entirely with the current formulation. If we derive `Committable` as we currently do, this is a tautology. It was added when we had explicit state, to try and confirm the state was sensible.

achamayou: Agreed

tla/consensus/Traceccfraft.tla
```diff
@@ -146,21 +146,21 @@ IsTimeout ==
     /\ IsEvent("become_candidate")
     /\ logline.msg.state.leadership_state = "Candidate"
     /\ Timeout(logline.msg.state.node_id)
-    /\ committableIndices[logline.msg.state.node_id] = Range(logline.msg.state.committable_indices)
+    /\ Range(logline.msg.state.committable_indices) \ CommittableIndices(logline.msg.state.node_id) = {}
```
eddyashton, line 149:
```suggestion
    /\ Range(logline.msg.state.committable_indices) \subseteq CommittableIndices(logline.msg.state.node_id)
```

I think this says the same, slightly clearer?

heidihoward (approved):

PR #5945, opened 2024-01-20
Minor refactorings

lemmy:
* Move aliases out of ccfraft
* Update references that have been removed or renamed
* Differential ALIAS: https://github.com/microsoft/CCF/pull/5940#issuecomment-1901553844


tla/consensus/MCAliases.tla
```diff
@@ -0,0 +1,124 @@
+----- MODULE MCAliases -----
```
achamayou, line 1: Are those not just Aliases, in general, usable on MC/SIM/Trace?

achamayou (approved):

tla/consensus/MCAliases.tla
lemmy: Yes, I just aimed to clarify that the aliases for TLC. Historically, the community used the MC* to designate files specific to TLC -- all conventions suck.

PR #5947, opened 2024-01-22
Do not run CI on older branches

achamayou:
There was some gating, but it was incomplete. release/0.x is currently used to test package signing.


achamayou:
The Daily failure is an upload issue to the component governance service, it is not related to the changes in this PR.

eddyashton (approved):

PR #5949, opened 2024-01-22
Add explicit permissions blocks to all GHA workflows

eddyashton:
A [policy change](https://docs.opensource.microsoft.com/github/apps/permission-changes/) means that things are going to break shortly.

The only thing that explicitly uses `GITHUB_TOKEN` is cimetrics, so we can fix that job with explicit permissions _EDIT - That's an ADO pipeline, so unaffected_. To see if anything else will break, I'm simulating the change with manual permission blocks.

None of our GHA pipelines are obviously crippled by this. We expect backport will need permissions, so I've guessed at what that permission block looks like - we'll work out if that was correct once this is merged.

### DIFF
--- a/.azure-pipelines-gh-pages.yml
+++ b/.azure-pipelines-gh-pages.yml
@@ -5,6 +5,8 @@ trigger:
       - main
       - "refs/tags/ccf-*"
 
+permissions: read-all
+
 jobs:
   - job: build_and_publish_docs
     displayName: "Build and publish docs"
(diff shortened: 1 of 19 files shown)


achamayou (approved):
LGTM!

eddyashton:
Looks like CodeQL needs more permissions, I'll add that in a future PR.

Also, going to label _this_ PR for backporting, to see if the permissions are sufficient. Any resulting PRs should not actually be merged!

PR #5951, opened 2024-01-23
[WIP] Liveness issue investigation

achamayou:
Prompted by https://github.com/microsoft/CCF/pull/5919#issuecomment-1905642294

image


tests/raft_scenarios/retire
```diff
@@ -0,0 +1,50 @@
...
+
+emit_signature,3
+dispatch_all
+dispatch_all
+assert_commit_idx,1,4
+
+emit_signature,3
+dispatch_all
+dispatch_all
+assert_commit_idx,1,4
```
eddyashton: I think we need `periodic`s` in here so that the new primary actually emits AEs?

tests/raft_scenarios/retire
```diff
@@ -0,0 +1,50 @@
...
+
+# Reconfig [0, 1] -> [0, 2]
+swap_node,2,0,2
+emit_signature,2
+
+# Leader replicates
+periodic_one,0,10
+dispatch_one,0
+
+# But node 1 runs for election before ack/cmt
```
eddyashton, line 25: There may be distinct cases if we do this _before_ and _after_ the ACK is received.  In particular, after the ACKs arrive at the primary (but before anything further has been shared), the primary can advance commit, and get into their final retirement state.

tests/raft_scenarios/retire
achamayou: Yes, sorry, fixed, but does not affect result.

tests/raft_scenarios/retire
achamayou: Agreed, but I think that's the same/maybe worse, definitely no better.

achamayou: In fact, this is a red herring caused by my mistaken thought that the primary counting itself in vs rejecting AE attempts from other candidates would matter. But that is not the case.

What matters is whether all the possible primaries post-retirement have a way to advance commit, and this is no more true whether the initial primary steps down gracefully or is interrupted at the moment.

achamayou:
Moved scenarios to #5973, containing a tentative fix.

PR #5954, opened 2024-01-24
Split release pipeline and add signing

achamayou:
Split release pipeline, because:

1. It needs to be set-up in a different, private ADO org
2. It cannot re-use the same pools at the public org

Also add signing step to release pipeline.

Once merged, I will:

1. Delete the "Test Signed Release Pipeline", drop release/0.x and the ccf-0.0.0-dev* tags.
2. Create a "Signed Release Pipeline" instead.


.azure-pipelines-templates/release.yml
```diff
@@ -21,8 +21,45 @@ jobs:
           set -ex
           cd $(Build.ArtifactStagingDirectory)
           rename.ul + _ *+*.deb || true
+          ls
```
eddyashton, line 24: Minor: Might be worth running `ls` _before_ the rename as well, so we can compare what was actually renamed?

eddyashton (approved):

.azure-pipelines-templates/release.yml
achamayou: Sure, the rename isn't new and is thought to be fine, this was more of a pre-sign ls to see what we're feeding the signing, but why not.

.azure-pipelines-templates/release.yml
```diff
@@ -21,8 +21,45 @@ jobs:
           set -ex
           cd $(Build.ArtifactStagingDirectory)
```
achamayou, line 22:
```suggestion
          cd $(Build.ArtifactStagingDirectory)
          ls
```

PR #5958, opened 2024-01-26
Black reformat python files

heidihoward:
Result of `git ls-files tests/ python/ scripts/ .cmake-format.py | grep -e '\.py$' | xargs black ` to fix CI failure here https://github.com/microsoft/CCF/actions/runs/7666015400/job/20892996824?pr=5919


achamayou (approved):
Thank you!

achamayou:
Passes format checks, format only change, fast tracked.

PR #5967, opened 2024-01-29
Update 5979 and rename to retire_one

achamayou:
Update deprecated 5979 scenario, which is essentially the simplest retirement scenario, now that we have retirement modeled accurately in the spec thanks to #5919.


src/consensus/aft/test/driver.cpp
```diff
@@ -107,6 +107,38 @@ int main(int argc, char** argv)
         assert(items.size() == 4);
         driver->swap_node(items[1], items[2], items[3], lineno);
         break;
+      case shash("swap_nodes"):
```
eddyashton, line 110: Since this is more complex than the other driver commands, can we add a comment with some sample values? May be simpler if we require `in` and `out` to be present, and in that order, even if their sublist args are empty.

eddyashton (approved):

src/consensus/aft/test/driver.cpp
achamayou: Good point, will add samples. Having to specify empty in and out doesn't seem necessary.

PR #5971, opened 2024-01-30
Adding `is_new_follower` to consensus spec

heidihoward:
Add `is_new_follower` to the consensus spec to unblock https://github.com/microsoft/CCF/pull/5925. This PR also:
- allows the follower to rollback when it receives an AE even if there are no conflicting entries (or no entries at all in the AE). 
- sets the initial send_idx to the length of the leader's log
- prevents the `ConflictAppendEntriesRequest` action from being taken without also taking `AppendEntriesAlreadyDone` or `NoConflictAppendEntriesRequest`

### DIFF
--- a/tla/consensus/ccfraft.tla
+++ b/tla/consensus/ccfraft.tla
@@ -946,7 +946,6 @@ AppendEntriesAlreadyDone(i, j, index, m) ==
 
 ConflictAppendEntriesRequest(i, index, m) ==
     /\ Len(log[i]) >= index
-    /\ log[i][index].term /= m.entries[1].term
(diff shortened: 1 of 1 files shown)


achamayou (approved):

achamayou:
Curious to see if that fixes #5925 altogether or just uncovers the next issue.

heidihoward:
I expect the reduce in states is due to the introduction of cdot to prevent AE rollback without response or fewer rollbacks as there's only one per term

heidihoward:
I tested https://github.com/microsoft/CCF/pull/5925 locally and it works, let's see what the CI has to say

PR #5973, opened 2024-01-31
Raft: Nodes stay active later during reconfiguration to ensure liveness

achamayou:
Update the retirement logic, making use of #5962 and #6018, to fix a liveness issue in some reconfiguration situations. See the `rcfg_01_12` and `rcfg_01_timeout_12` scenarios in this PR for reproduction samples.

The gist of this change is that nodes remain active for longer in retirement, despite no longer being counted or counting themselves in election or persistence quorums. Leaders only step down once they have confirmed the new configuration can make progress without nodes in the old configuration to minimise the number of elections.

A node that is not retired `(Active, N/A)` can:

1. Append entries if leader
2. Run for election
3. Ack entries
4. Vote in elections

A node that has seen its retirement signed `(Retired, Ordered)` can:

1. Append entries if leader **[Changed, previously [not allowed](https://github.com/microsoft/CCF/blob/443309e10dff61385f758ecc5d7f932570704d27/src/consensus/aft/raft.h#L266) [*](https://github.com/microsoft/CCF/blob/443309e10dff61385f758ecc5d7f932570704d27/src/consensus/aft/raft.h#L900)]**
2. Run for election **[Changed, previously [not allowed](https://github.com/microsoft/CCF/blob/443309e10dff61385f758ecc5d7f932570704d27/src/consensus/aft/raft.h#L1822) [*](https://github.com/microsoft/CCF/blob/443309e10dff61385f758ecc5d7f932570704d27/src/consensus/aft/raft.h#L1959)]**
3. Ack entries **[Changed, previously [not allowed](https://github.com/microsoft/CCF/blob/443309e10dff61385f758ecc5d7f932570704d27/src/consensus/aft/raft.h#L603) past `retirement_committable_idx`]**
4. Vote in elections

A node that has seen its retirement committed `(Retired, Completed)` can:

1. Append entries if leader **[Changed, previously [not allowed](https://github.com/microsoft/CCF/blob/443309e10dff61385f758ecc5d7f932570704d27/src/consensus/aft/raft.h#L266) [*](https://github.com/microsoft/CCF/blob/443309e10dff61385f758ecc5d7f932570704d27/src/consensus/aft/raft.h#L900)]**
[...]

PR #5986, opened 2024-02-05
Add cleanup_nodes action to the driver to set retired_committed in scenarios

achamayou:
Proposed change to the driver to enable testing of the #5973.

Adds a `cleanup_nodes` command, which results in the replication of a payload equivalent to the KV write setting `retired_committed` in the nodes table. When that transaction is committed, the nodes passed to `cleanup_nodes` have their `retired_committed` status set.

This would ideally happen automatically, but that would require further parallel logic to replace `ccf::RetiredNodeCleanup` and the `set_retired_committed` KV-backed endpoint.

I am wary of:

1. Introducing a real KV in the driver. This is a very substantial extension.
2. Building out the parallel KV further (configurations just use map hooks, and retired_committed is an extra vector, but `set_retired_committed` would require the ability to read committed state, when we don't even have state).
3. Doing these things for the benefit of saving essentially one line across what's likely to be a handful of scenarios at most.

### DIFF
--- a/src/consensus/aft/test/driver.h
+++ b/src/consensus/aft/test/driver.h
@@ -168,12 +168,15 @@ private:
       auto c = nlohmann::json(configuration).dump();
 
       // If the entry is a reconfiguration, the replicated data is overwritten
-      // with the serialised configuration
+      // with the serialised configuration.
       data = std::vector<uint8_t>(c.begin(), c.end());
     }
     if (retired_committed.has_value())
     {
-      _nodes.at(node_id).kv->retired_commit_entries.emplace_back(
+      // Update the node's own retired committed entries collection when
+      // replicating There is no direct equivalent in a real node, but this is
+      // necessary to emulate the global commit hook effectively.
+      _nodes.at(node_id).kv->retired_committed_entries.emplace_back(
         idx, retired_committed.value());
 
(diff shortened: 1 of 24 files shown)


src/consensus/aft/test/driver.h
```diff
@@ -151,6 +153,12 @@ class RaftDriver
 
     aft::ReplicatedDataType type = aft::ReplicatedDataType::raw;
     auto hooks = std::make_shared<kv::ConsensusHookPtrs>();
+    if (configuration.has_value() && retired_committed.has_value())
```
achamayou, line 156: Similar to `configuration`, `retired_committed` is a special replicate with a payload (a JSON config), but different semantics:

1. The configuration is kept in a local collection, subject to rollback
2. On compact(`idx`), set_retired_committed is called on the elements of configurations strictly older than `idx`.

This is, as far as I can tell, the smallest possible change necessary to emulate faithfully what the global commit hook does in https://github.com/microsoft/CCF/blob/0fa4538ef2bd86cba991175abdd9ce3c391294c2/src/node/node_state.h#L2147.

eddyashton (approved):

heidihoward (approved):

PR #5988, opened 2024-02-05
Checking sentIndex in trace validation

heidihoward:
Let's try this and see what breaks in the CI

### DIFF
--- a/tla/consensus/Traceccfraft.tla
+++ b/tla/consensus/Traceccfraft.tla
@@ -175,8 +175,7 @@ IsSendAppendEntries ==
                 /\ IsAppendEntriesRequest(msg, j, i, logline)
                 \* There is now one more message of this type.
                 /\ Network!OneMoreMessage(msg)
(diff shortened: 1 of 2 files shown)


heidihoward:
The CI is failing due to a difference between the `sent_idx`  when a new node joins. 

In the implementation, the first AE message when a new node joins is a heartbeat for log length - 1 (in other words, the log before the reconfiguration was added). 
https://github.com/microsoft/CCF/blob/9fabae850dcc26c788cfed25e7707f0c80dee820/src/consensus/aft/raft.h#L2417

In the spec, this behavior is permitted by setting the initial `sent_idx` to log length -1 (using `Len(log[i])`  instead of `Len(log'[i])`. To implement this properly, we would need to either:
1.   add the first AE to the reconfiguration action, 
2.   add a new variable to conditionally changed the behavior of the append entries action or 
3.   change the implementation

Since this is a fairly low priority change, I think I'll shelve it for now and we can revisit it later

PR #5989, opened 2024-02-06
Towards easier debugging of trace validation.

lemmy:
When TLC fails to validate a log, it prints a counterexample for the `TraceMatched` property that doesn't provide sufficient information to pinpoint the exact problem.  During trace validation, TLC generates a set of traces—finite prefixes of the possible behaviors as defined by the specification. Among these, the longest trace is considered the first candidate for detailed analysis (in the case of a breadth-first search, all traces are of equal length). Then, users are particularly interested in exploring the successor states to final state of this longest trace. Additionally, users would want to know which part of the next-state relation, or any state or action constraint, was evaluated as FALSE, leading to the exclusion of the successor states from the model.
However, the longest trace could be a red herring; especially with depth-first search, as shorter traces could be the one that has to be completed. Therefore, for an accurate diagnosis of the issue, TLC should display its entire graph of traces rather than just a single trace.

Note that in the special case that trace validation is completely deterministic, TLC will generate a single prefix.

Requested:
https://github.com/microsoft/CCF/pull/5933

Related:
https://github.com/tlaplus/tlaplus/pull/867
https://github.com/tlaplus/tlaplus/commit/f7b4c8ac2406bd90bcfc03fdd4651c318df9c604
https://github.com/tlaplus/tlaplus/commit/312518885c0f4688fceb1b30d27c2aed61a6628f
https://github.com/tlaplus/tlaplus/commit/c2713def38aa1d3255a6c2fe2aa037253e00e361

### DIFF
--- a/.azure-pipelines-templates/trace_validation.yml
+++ b/.azure-pipelines-templates/trace_validation.yml
@@ -9,5 +9,5 @@ steps:
   - script: |
       set -ex
       cd tla/
-      parallel 'JSON={} java -XX:+UseParallelGC -Dtlc2.tool.impl.Tool.cdot=true -Dtlc2.tool.queue.IStateQueue=StateDeque -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC -lncheck final -checkpoint 0 consensus/Traceccfraft.tla' ::: $(ls ../build/*.ndjson | grep -v _deprecated)
+      parallel 'JSON={} java -XX:+UseParallelGC -Dtlc2.tool.impl.Tool.cdot=true -Dtlc2.tool.queue.IStateQueue=StateDeque -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC -lncheck final -checkpoint 0 -dump dot,constrained,colorize,actionlabels {}.dot consensus/Traceccfraft.tla' ::: $(ls ../build/*.ndjson | grep -v _deprecated)
     displayName: "Run trace validation"
--- a/tla/consensus/Traceccfraft.cfg
+++ b/tla/consensus/Traceccfraft.cfg
@@ -31,6 +31,9 @@ PROPERTIES
 INVARIANT 
     TraceDifferentialInv
 
+CONSTRAINT
+    TraceMatchesConstraints
+
 \* Checking for deadlocks during trace validation is disabled, as it may lead to false
 \* counterexamples. A trace specification defines a set of traces, where at least one
 \* trace is expected to match the log file in terms of variable values and length.
--- a/tla/consensus/Traceccfraft.cfg
+++ b/tla/consensus/Traceccfraft.cfg
@@ -10,6 +10,9 @@ SPECIFICATION
 VIEW
     TraceView
 
+POSTCONDITION 
+    TraceMatchedNonTrivially
+
 \* Some of the traces in TraceSpec may not refine the high-level specification ccfraft.
 \* Those traces will cause the trace validator to report a counterexample even if 
 \* TraceSpec contains a trace that refines ccfraft and fully matches the log file.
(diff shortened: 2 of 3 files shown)


lemmy:
Untitled

achamayou (approved):

heidihoward:
I tried that out locally this morning, and it looks fantastic. The only diff between what I'm seeing locally and the image posted is that all my action labels just say "Unnamed Action".

Cmd:
`JSON=../build/replicate.ndjson ./tlc.sh consensus/Traceccfraft.tla  -dump dot,constrained,colorize,actionlabels replicate_failed2.dot` on `TLC2 Version 2.18 of Day Month 20?? (rev: de0c181)` (the latest commit to tlc at the time of writing)

and result:
image

lemmy:
> @lemmy I tried that out locally this morning, and it looks fantastic. The only diff between what I'm seeing locally and the image posted is that all my action labels just say "Unnamed Action".

You can assume that "UnnamedAction" should be "DropAndNext" (it is unfortunately not trivial to have TLC emit the specific action name, unless the if condition `ENABLED Trace` is pushed as a conjunct into every subaction).  Can you zip up your files for me to investigate?

heidihoward:
Ignore me - I'd accidentally ended up in a state without https://github.com/microsoft/CCF/commit/4739d1fbc6c068554ac97831b84cbcb2eba5a924. It works now

lemmy:
Asserting `last_idx` in `Traceccfraft!IsRcvAppendEntriesRequest` makes TV fail for raft scenario `../tests/raft_scenarios/ackuracy`.  The assertions is bogus but doesn't matter here.

```diff
diff --git a/tla/consensus/Traceccfraft.tla b/tla/consensus/Traceccfraft.tla
index f726fc925..bfc1c5396 100644
--- a/tla/consensus/Traceccfraft.tla
+++ b/tla/consensus/Traceccfraft.tla
@@ -207,7 +207,10 @@ IsRcvAppendEntriesRequest ==
     /\ commitIndex[logline.msg.state.node_id] = logline.msg.state.commit_idx
     \* TODO /\ leadershipState[logline.msg.state.node_id] = ToLeadershipState[logline.msg.state.leadership_state]
     /\ membershipState[logline.msg.state.node_id] \in ToMembershipState[logline.msg.state.membership_state]
-    \*TODO/\ Len(log'[logline.msg.state.node_id]) = logline.msg.state.last_idx
+    /\ \/ Len(log[logline.msg.state.node_id])  = logline.msg.state.last_idx
+       \/ Len(log'[logline.msg.state.node_id])  = logline.msg.state.last_idx
+       \/ Len(log[logline.msg.state.node_id]) = TraceLog[l-1].msg.state.last_idx
+       \/ Len(log'[logline.msg.state.node_id]) = TraceLog[l-1].msg.state.last_idx
```

When running TLC with the `-Dtlc2.tool.queue.IStateQueue=StateDeque` option, we cannot trust the counterexample on stdout or serialized using `-dumptrace`.  For example, in DFS mode, the longest generated trace extends to 201 states. However, TLC reports a shorter counterexample, consisting of only 124 states, which can be misleading. Therefore, it is crucial to focus on the reported search depth, indicated as "The depth of the complete state graph search is 201," to understand the actual diameter of the state graph.  Furthermore, the graphviz rendering will contain the complete graph. 

image

https://github.com/microsoft/CCF/pull/6021 allows us to switch back to slower BFS (remove `-Dtlc2.tool.queue.IStateQueue=StateDeque`), which is guaranteed to report one of the longest traces.

Related: https://github.com/microsoft/CCF/pull/6017/files#diff-dee5ad6d573a04292b610ec1e8b104d1fff1164f6553eb3486f54bc526279b87

PR #5991, opened 2024-02-06
Rollback safety despite stale NACKs

eddyashton:
Exploring #5927, @heidihoward raised a potential safety violation due to our reuse of the `Term` field in negative AER messages.

1. Stale NACKs cannot be distinguished from fresh NACKs about conflicting ledgers
2. Receiving a stale NACK may cause a primary to set its corresponding `sent_idx` too low - much lower than the node's actual matching suffix
3. When `sent_idx` is too low, a primary will produce AEs which overlap with the current matching suffix
4. When a backup receives an AE which looks like a conflict, it eagerly rolls back (before checking if that AE contains actual conflicts!)
5. This roll back could truncate committed state, losing persistence

In brief, we could have nodes with the following logs (square brackets marking commit)
```
L  1  2  3  4 [5]
F  1 [2] 3  4  5  

F -> L: NACK @ idx=2 (STALE)
L -> F: AE [2, 3)
```

This PR adds a scenario reproducing this situation (`rollback_safety`), and demonstrating that this introduces a safety violation (failure of an `assert_commit_safety` step).

This PR also implements a fix, at level 4 of the steps above - we align more closely with classic Raft, and the spec, by delaying the rollback point. Rather than rolling back immediately for any AE (including heartbeats) whose `prev_idx` is before our `state->last_idx`, we consider each ledger entry as we process the AE and only rollback when we discover a true conflict.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -171,12 +171,7 @@ namespace aft
     std::default_random_engine rand;
 
   public:
-    static constexpr size_t append_entries_size_limit =
-#ifdef OVERRIDE_APPEND_ENTRIES_SIZE_LIMIT
-      OVERRIDE_APPEND_ENTRIES_SIZE_LIMIT;
-#else
-      20000;
-#endif
+    static constexpr size_t append_entries_size_limit = 20000;
 
     std::unique_ptr<LedgerProxy> ledger;
     std::shared_ptr<ccf::NodeToNode> channels;
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1139,6 +1139,49 @@ namespace aft
       // Finally, deserialise each entry in the batch
       for (Index i = r.prev_idx + 1; i <= r.idx; i++)
       {
+        if (i <= state->last_idx)
+        {
+          // NB: This is only safe as long as AppendEntries only contain a
+          // single term. If they cover multiple terms, then we at least
+          // partially deserialise each entry to establish what term it is in
+          const auto incoming_term = r.term_of_idx;
+          const auto local_term = state->view_history.view_at(i);
+          if (incoming_term != local_term)
+          {
+            if (is_new_follower)
+            {
+              auto rollback_level = i - 1;
+              RAFT_DEBUG_FMT(
+                "New follower received AppendEntries with conflict. Incoming "
+                "entry {}.{} conflicts with local {}.{}. Rolling back to {}.",
+                incoming_term,
+                i,
+                local_term,
+                i,
+                rollback_level);
+              rollback(rollback_level);
+              // Then continue to process this AE as normal
+            }
+            else
+            {
+              RAFT_FAIL_FMT(
+                "Found conflict at {}.{} (incoming AppendEntries contains "
+                "{}.{}). Ignoring due to is_new_follower=false.",
+                local_term,
+                i,
+                incoming_term,
+                i);
+            }
+          }
+          else
+          {
+            // If the current entry has already been deserialised, skip the
+            // payload for that entry
+            ledger->skip_entry(data, size);
+            continue;
+          }
+        }
+
         std::vector<uint8_t> entry;
         try
         {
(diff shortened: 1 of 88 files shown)


src/consensus/aft/raft.h
```diff
@@ -1161,10 +1141,45 @@ namespace aft
       {
         if (i <= state->last_idx)
         {
-          // If the current entry has already been deserialised, skip the
-          // payload for that entry
-          ledger->skip_entry(data, size);
-          continue;
+          // NB: This is only safe as long as AppendEntries only contain a
+          // single term. If they cover multiple terms, then we at least
```
achamayou, line 1145:
```suggestion
          // single term. If they cover multiple terms, then we would at least need to
```

achamayou (approved):

tla/consensus/ccfraft.tla
```diff
@@ -959,7 +959,7 @@ AppendEntriesAlreadyDone(i, j, index, m) ==
     /\ Reply([type           |-> AppendEntriesResponse,
               term           |-> currentTerm[i],
               success        |-> TRUE,
-              lastLogIndex   |-> m.prevLogIndex + Len(m.entries),
+              lastLogIndex   |-> Len(log[i]),
```
eddyashton, line 962: Adding 2 TLA changes to this PR, so that trace validation continues to work with the new rollback point.

This is aligning with implementation - an AER ACK always lists the node's `state->last_idx`. It does not consider the incoming message.

tla/consensus/ccfraft.tla
```diff
@@ -1054,7 +1054,8 @@ HandleAppendEntriesResponse(i, j, m) ==
           /\ leadershipState[i] = Leader \* Only Leaders need to tally append entries responses
           /\ m.success \* successful
           \* max(...) because why would we ever want to go backwards on a success response?!
-          /\ matchIndex' = [matchIndex EXCEPT ![i][j] = max(@, m.lastLogIndex)]
+          \* min(...) because we don't want to calculate a match past our own log's tail
+          /\ matchIndex' = [matchIndex EXCEPT ![i][j] = max(@, min(Len(log[i]), m.lastLogIndex))]
```
eddyashton, line 1058: We've had this in `raft.h` for a while:

```
        // Potentially unnecessary safety check - use min with last_idx, to
        // prevent matches past this node's local knowledge
        const auto proposed_match = std::min(r.last_log_idx, state->last_idx);
```

Without this corresponding change in the TLA I was seeing trace validation failures for `fancy_election.2`, which I tracked down to a discrepancy in `matchIndex`. We try to match a `send_append_entries` step where the node says its `match_idx` is 12, yet the model has incremented it to 14. It was set to 14 when processing an AER ACK, despite the receiver's log only going as far as 12.

I thought this was an unnecessary safety check (as the comment says), but now it looks more worrying. An ACK extending past your own log means the receiver has a longer version of the previous term, which you are about to overwrite. They would overwrite it if you sent them an appropriate AE, but if you increment `match_idx` without that then you potentially advance commit unsafely.

src/consensus/aft/test/main.cpp
```diff
@@ -767,7 +767,7 @@ DOCTEST_TEST_CASE("Recv append entries logic" * doctest::test_suite("multiple"))
   }
 }
 
-DOCTEST_TEST_CASE("Exceed append entries limit")
+DOCTEST_TEST_CASE("Exceed append entries limit" * doctest::skip())
```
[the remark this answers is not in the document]
eddyashton: False alarm. Don't run `raft_test` with `CCF_RAFT_TRACING=ON`, or this test will be reeeeeeally slow.

tla/consensus/ccfraft.tla
eddyashton: Simulation has found a violation of `LogInv` that I believe is down to this. Investigating further.

tla/consensus/ccfraft.tla
eddyashton: This is a receiver-side safety check that's doing only part of the necessary check. The above change is what sender should actually be doing. I'll remove this `min` from both, and see if anything breaks.

tla/consensus/ccfraft.tla
eddyashton: Linking in the other direction for posterity - this change is incorrect, and we need to align the impl with the spec. That's done in #6001.

eddyashton:
Superseded by #6016.

PR #5994, opened 2024-02-06
Publish TV output on failure

achamayou:


lemmy (approved):

heidihoward (approved):
lgtm, this is the same logic we already use for the model checking and simulation pipelines

PR #5996, opened 2024-02-07
Add actions for Raft scenarios

eddyashton:
Pulling out a few actions proposed in previous PRs (#5991), adding them explicitly here.


heidihoward (approved):
Thank you so much! This is really helpful to me as it was it easier for me to run the scenario in #5996 against the current main

achamayou (approved):

PR #6001, opened 2024-02-08
Raft: Only ACK up to entries described by incoming AppendEntries

eddyashton:
Discrepancy between implementation and spec uncovered in #5991, and here the spec is definitely right.

https://github.com/microsoft/CCF/pull/5991#discussion_r1482986763

We must consider the incoming message when producing an ACK, and only ACK up to its tail. When this was "already done", we may hold some further suffix, and we do not know if it agrees with the leader's.

Change is slightly larger than strictly necessary, to try and gain clarity on what we're passing where. We may want to make a similar change to NACKs, but there's not currently a strong motivation for that.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -949,8 +949,6 @@ namespace aft
 
       const auto prev_term = get_term_internal(prev_idx);
       const auto term_of_idx = get_term_internal(end_idx);
-      const bool contains_new_view =
-        (state->new_view_idx > prev_idx) && (state->new_view_idx <= end_idx);
 
       RAFT_DEBUG_FMT(
         "Send append entries from {} to {}: ({}.{}, {}.{}] ({})",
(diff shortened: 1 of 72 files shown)


heidihoward:
Great spot Eddy! I wonder why this divergence between the spec and impl was not picked up by the TV?

eddyashton:
> I wonder why this divergence between the spec and impl was not picked up by the TV?

I think we don't have any scenarios that do this. I'm not sure exactly the interleaving of messages/drops that is required, but some permutation is necessary to produce an Already Done NOP AE. This is a known gap in the scenario coverage - we mostly deliver or drop everything on a channel.

src/consensus/aft/raft.h
```diff
@@ -1355,29 +1354,53 @@ namespace aft
...
+      // If we get to here, we have applied up to r.idx in this AppendEntries.
+      // We must only ACK this far, as we know nothing about the agreement of a
+      // suffix we may still hold _after_ r.idx with the leader's log
+      const auto response_idx = ae.idx;
+      send_append_entries_response(
+        to, AppendEntriesResponseType::OK, state->current_view, response_idx);
+    }
 
-      if (answer == AppendEntriesResponseType::FAIL && rejected.has_value())
+    void send_append_entries_response_nack(
```
achamayou, line 1371: imho clearer as two:
send_append_entries_response_nack(ccf::NodeId to)
send_append_entries_response_nack(ccf::NodeId to, const ccf::TxID& rejected)

achamayou:
It'd be nice to construct a scenario for this if possible.

eddyashton:
> It'd be nice to construct a scenario for this if possible.

We now have a scenario for this, called `ackuracy`. However it relies on the fix from #5991 to work: we need to skip every entry in an AE, to produce an "AlreadyDone" AER which this change alters, without #5991 we rollback on those heartbeats so they're not nops/skips. Despite that constraint, I'd like to merge this PR before #5991 - I'm still exploring the `min` behaviour there, but think this is ready to merge (and now has a regression scenario).

eddyashton:
Superseded by #6016.

PR #6002, opened 2024-02-08
Adding `ReplicationInv` to consensus spec

heidihoward:
This should help catch safety violations sooner


tla/consensus/ccfraft.tla
```diff
@@ -1425,6 +1425,31 @@ CommitCommittableIndices ==
...
+GetConfigurations(log_x, idx) ==
+    LET
+    configs_all == {k \in DOMAIN log_x : log_x[k].contentType = TypeReconfiguration}
+    configs_before == {k \in configs_all : k < idx}
+    config_last == IF configs_before = {} THEN {} ELSE {Max(configs_before)}
+    configs_after == {k \in configs_all : k >= idx}
+    IN
+    {log_x[i].configuration : i \in (configs_after \union config_last)}
+
+\* ReplicationInv states that all log entires that are believed to be committed must be
```
achamayou, line 1440:
```suggestion
\* ReplicationInv states that all log entries that are believed to be committed must be
```

achamayou (approved):

achamayou:
@heidihoward this change appears to cut down the number of sim traces quite a lot, presumably because it expensive to check:

image

I don't have a particular opinion about whether the tradeoff is worthwhile.

heidihoward:
> @heidihoward this change appears to cut down the number of sim traces quite a lot, presumably because it is expensive to check:
> 

Now we know it works, let's try to decrease the overhead before merging

lemmy:
> > @heidihoward this change appears to cut down the number of sim traces quite a lot, presumably because it is expensive to check:
> 
> Now we know it works, let's try to decrease the overhead before merging

If this is a major performance bottleneck, we should override it with a Java module override, which will have a more significant impact, while keeping the spec readable.

achamayou:
It's not really perf as such, the sim runs for a fixed amount of time anyway, it's just that we get much fewer traces that way. I'm not sure about having a hardcoded 5 nodes startup, that doesn't change the hit much, and it means we don't cover startup or 3 node scenarios, which are probably about 99% of real world usage.

Having said that, if we think this is momentarily helpful in our journey to improve the spec and the TV, why not.

heidihoward:
>  I'm not sure about having a hardcoded 5 nodes startup

I was just experiment with something (whether simulation with this invar and start state from the TV could find https://github.com/microsoft/CCF/pull/5991. This will not be final PR

achamayou (approved):

lemmy:
@heidihoward Should we add ReplicationInv also to Traceccfraft?

lemmy:
According to TLC's profiler, `ReplicationInv` is very expensive to evaluate, i.e., creates many intermediate values.  Staring at it, I don't understand why it all-quantifies over `DOMAIN Committed(i)`:

https://github.com/microsoft/CCF/blob/061e4daac9b25f2083c8454a6261628fc76e28ca/tla/consensus/ccfraft.tla#L1499-L1501

`CommittedLogAppendOnlyProp` asserts that committed log are append-only, and `LogInv` asserts that the committed logs of all servers are the same.  Given that and the fact that TLC checks `ReplicationInv` for every state, doesn't that imply that TLC will have already checked `ReplicationInv` for all lower commit indices.

```tla
ReplicationInv ==
    \E i \in Servers : 
        \* We just check the node with the highest commitIndex
        \* LogInv ensures that includes all committed transactions
        /\ \A j \in Servers: commitIndex[i] >= commitIndex[j]
        \* Every committed transaction must be replicated to at least 
        \* one quorum in each configuration which should have a copy
        \* ??? Do we have to quantify over DOMAIN Committed(i) here?
        \* ??? CommittedLogAppendOnlyProp asserts that the committed log is append-only, and LogInv
        \* ??? asserts that the committed log of all servers is the same.
        \* ??? Given that and the fact that TLC checks ReplicationInv for every state, it implies that
        \* ??? we will have already checked ReplicationInv for all lower commit indices.
        /\ \*\A idx \in DOMAIN Committed(i) :
            \A config \in GetConfigurationsFold(Committed(i), commitIndex[i]) :
                \E quorum \in Quorums[config] :
                    \A node \in quorum : 
                        IsPrefix(Committed(i), log[node])
                        \* /\ Len(log[node]) >= idx
                        \* /\ log[node][idx] = log[i][idx]
```

--------

Below is a variant of `GetConfigurations` that TLC should find easier to evaluate.

```tla
GetConfigurations(log_x, idx) ==
    FoldLeft(LAMBDA acc, k: 
                IF log_x[k].contentType = TypeReconfiguration
                THEN IF k <= idx
                     \* Keep replacing the matched configuration until k > idx, then
                     THEN {log_x[k].configuration}
                     \* ...accept all the remaining configurations. 
                     ELSE {log_x[k].configuration} \cup acc
                ELSE acc,
                {}, [ i \in DOMAIN log_x |-> i])
```

Related: https://github.com/tlaplus/CommunityModules/issues/101

lemmy:
https://github.com/microsoft/CCF/pull/6053

PR #6003, opened 2024-02-08
Minor Raft QoL: Make message construction more readable

eddyashton:
This is just a style change, and should no impact on behaviour.

I just noticed that the way we construct Raft messages gives me the heebie jeebies:

```
      AppendEntries ae = {
        {raft_append_entries},
        {end_idx, prev_idx},
        state->current_view,
        prev_term,
        state->commit_idx,
        term_of_idx,
        contains_new_view};
```

1. Why are we repeating `AppendEntries` and `raft_append_entries`?
2. What are all these fields?

This PR tries to resolve both, by setting the `msg` field (basically a discriminator for a tagged union struct) from a templated type, and assigning to each field by name.

```
      AppendEntries ae;
      ae.idx = end_idx;
      ae.prev_idx = prev_idx;
      ae.term = state->current_view;
      ae.prev_term = prev_term;
      ae.leader_commit_idx = state->commit_idx;
      ae.term_of_idx = term_of_idx;
```

This is only an extremely incremental change because:
1. You could still _re_ assign the `ae.msg` field to something else if you wanted to cause chaos (because we can't make it const, because of our JSON serialisation machinery).
2. The field names are still brief and confusing.

But I think this is a worthwhile improvement regardless.

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1535,7 +1535,7 @@ namespace aft
       RequestVote rv;
       rv.term = state->current_view;
       rv.last_committable_idx = last_committable_idx;
-      rv.term_of_last_committable_idx = last_committable_idx;
+      rv.term_of_last_committable_idx = get_term_internal(last_committable_idx);
 
 #ifdef CCF_RAFT_TRACING
       nlohmann::json j = {};
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -949,8 +949,6 @@ namespace aft
 
       const auto prev_term = get_term_internal(prev_idx);
       const auto term_of_idx = get_term_internal(end_idx);
-      const bool contains_new_view =
-        (state->new_view_idx > prev_idx) && (state->new_view_idx <= end_idx);
 
       RAFT_DEBUG_FMT(
         "Send append entries from {} to {}: ({}.{}, {}.{}] ({})",
@@ -962,14 +960,13 @@ namespace aft
         end_idx,
         state->commit_idx);
 
-      AppendEntries ae = {
-        {raft_append_entries},
-        {end_idx, prev_idx},
-        state->current_view,
-        prev_term,
-        state->commit_idx,
-        term_of_idx,
-        contains_new_view};
+      AppendEntries ae;
+      ae.idx = end_idx;
+      ae.prev_idx = prev_idx;
+      ae.term = state->current_view;
+      ae.prev_term = prev_term;
+      ae.leader_commit_idx = state->commit_idx;
+      ae.term_of_idx = term_of_idx;
 
       auto& node = all_other_nodes.at(to);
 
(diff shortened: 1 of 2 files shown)


src/consensus/aft/raft.h
```diff
@@ -949,8 +949,6 @@ namespace aft
 
       const auto prev_term = get_term_internal(prev_idx);
       const auto term_of_idx = get_term_internal(end_idx);
-      const bool contains_new_view =
-        (state->new_view_idx > prev_idx) && (state->new_view_idx <= end_idx);
```
eddyashton, line 953: Ok, I'm sneaking in "don't even populate `contains_new_view`" in this change too. Because we never use the read value, and haven't for a long time, I claim this is fine.

achamayou:
@eddyashton I would prefer switching to designated initialisers (https://en.cppreference.com/w/cpp/language/aggregate_initialization) because they look cleaner (single init step) and that way we don't need to repeat the name of the ae.

src/consensus/aft/raft.h
```diff
@@ -962,14 +960,13 @@ namespace aft
...
-        state->commit_idx,
-        term_of_idx,
-        contains_new_view};
+      AppendEntries ae;
+      ae.idx = end_idx;
+      ae.prev_idx = prev_idx;
+      ae.term = state->current_view;
+      ae.prev_term = prev_term;
+      ae.leader_commit_idx = state->commit_idx;
+      ae.term_of_idx = term_of_idx;
```
achamayou, line 969:
```suggestion
      AppendEntries ae{
        .idx = end_idx,
        .prev_idx = prev_idx,
        .term = state->current_view,
        .prev_term = prev_term,
        .leader_commit_idx = state->commit_idx,
        .term_of_idx = term_of_idx
      }
```

seems to work even with clang-11

eddyashton: Sure, I'll switch to designated initializers. This instance becomes a little more awkward because of the constraints of designated initializers wrt base members. As written, it's a build error:

```
../src/consensus/aft/raft.h:964:10: error: field designator 'idx' does not refer to any field in type 'aft::AppendEntries'
        .idx = end_idx,
```

So you have to mix designated and non-designated, and squash the corresponding warning:
```
 warning: mixture of designated and non-designated initializers in the same initializer list is a C99 extension [-Wc99-designator]
```
```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wc99-designator"
      AppendEntries ae{
        {},
        {.idx = end_idx, .prev_idx = prev_idx},
        .term = state->current_view,
        .prev_term = prev_term,
        .leader_commit_idx = state->commit_idx,
        .term_of_idx = term_of_idx,
      };
#pragma clang diagnostic pop
```

I think the multiple base classes here are unnecessary, but I'll look at unpicking that separately.

achamayou: Ah that's unfortunate, you are right. I do agree that the hierarchy can be simplified here, but as you say best done separately, and perhaps after the functional changes that are on the way.

achamayou (approved):

PR #6007, opened 2024-02-09
RaftDriver: Add `assert_detail`

achamayou:
Spun out of #5973, convenient in retirement scenarios.

### DIFF
--- a/src/consensus/aft/test/driver.h
+++ b/src/consensus/aft/test/driver.h
@@ -1081,7 +1081,7 @@ public:
     }
 
     std::string value = d[detail];
-    if (equal ? value != expected : value == expected)
+    if (equal ? (value != expected) : (value == expected))
     {
       std::string cmp = equal ? "!" : "=";
       RAFT_DRIVER_PRINT(
--- a/src/consensus/aft/test/driver.h
+++ b/src/consensus/aft/test/driver.h
@@ -963,30 +963,6 @@ public:
     }
   }
 
-  void assert_is_retired(ccf::NodeId node_id, const size_t lineno)
-  {
-    if (!_nodes.at(node_id).raft->is_retired())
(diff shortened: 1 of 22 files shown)


tests/raft_scenarios/retire_one
```diff
@@ -58,7 +58,8 @@ dispatch_all
 
 assert_is_primary,0
 assert_is_backup,1
```
eddyashton, line 60: We could also replace these assert lines (`assert_detail,1,leadership_state,Backup`), if we want to consolidate actions? I don't know if its helpful though.

eddyashton (approved):

tests/raft_scenarios/retire_one
achamayou: I started doing that, and then noticed the `assert_isnot_*` variants, which can't easily be replaced with `assert_detail`. But now that you mention it, we could easily `assert_!detail`.

src/consensus/aft/test/driver.h
```diff
@@ -1155,4 +1095,46 @@ class RaftDriver
...
+        "  Note over {}: Node does not have detail {}", node_id, detail);
+      throw std::runtime_error(fmt::format(
+        "Node {} does not have detail {} on line {}",
+        node_id,
+        detail,
+        std::to_string((int)lineno)));
+    }
+
+    std::string value = d[detail];
+    if (equal ? value != expected : value == expected)
```
eddyashton, line 1120:
```suggestion
    if (equal ? (value != expected) : (value == expected))
```

Nit: Odd place for a ternary, so I think this is a big help to readability.

achamayou: Ok as a one-off, but otherwise the clang-format is the law.

PR #6000, opened 2024-02-07
[DO_NOT_MERGE] Investigate rollback_safety scenario trace

heidihoward:
Using the CI to test trace validation and spec.

This PR is not for merging, just for experimenting with

PR #6009, opened 2024-02-09
RaftDriver: Always loop until sync is achieved at the end of a scenario

eddyashton:
Potentially controversial!

This adds a crude liveness check to the end of all scenarios. Currently most have a `(periodic_all; dispatch_all;)+ assert_state_sync;` section at the end, that confirms they can reach a stable state. This automates that, running these steps in a loop at the end of all scenarios until sync is reached. The claim is that by emitting a fresh signature, and then confirming that all nodes can reach the same state, we've proved liveness. Mild assumption that the fresh signature is the new commit index, maybe this passes spuriously if we sync somewhere else.

I think this should help trace validation, by ensuring that every trace finishes with some sensible common steps (actions which can be matched, and allow invariants to be checked). See https://github.com/microsoft/CCF/pull/6000#issuecomment-1934760838, where we're concerned they may otherwise stop too early.


lemmy:
FWIW: `periodic_all; dispatch_all;`  didn't help in #6000

lemmy:
FWIF: `suffix_collision.3` fails after 571 states instead of running indefinitely if this is merged into https://github.com/microsoft/CCF/pull/6021.

tla/consensus/ccfraft.tla
```diff
@@ -941,7 +941,7 @@ RejectAppendEntriesRequest(i, j, m, logOk) ==
                              dest           |-> j],
                              m)
                    \/ /\ prevTerm # 0
-                      /\ LET lli == FindHighestPossibleMatch(log[i], m.prevLogIndex, m.term)
+                      /\ LET lli == FindHighestPossibleMatch(log[i], m.prevLogIndex, m.prevLogTerm)
```
eddyashton, line 944: This was the source of `suffix_collision.3` failing trace validation. The AE claims that `{m.prevLogTerm}.{m.prevLogIdx}` is a TxID in its history. If we NACK, we want to find the highest entry we hold which could agree with that. The previous statement was trying to find a match with `{m.term}.{m.prevLogIndex}`, which is an invented TxID. In most instances it was also valid, or even matched, but we luckily have one case where it doesn't.

tla/consensus/Traceccfraft.tla
```diff
@@ -135,6 +135,7 @@ IsDropPendingTo ==
             /\ msg.type = RaftMsgType[logline.msg.packet.msg]
             /\ msg.dest = logline.msg.to_node_id
             /\ msg.source = logline.msg.from_node_id
+            /\ msg.term = logline.msg.packet.term
```
lemmy, line 138: I would suggest to add operators like `Traceccfraft!IsAppendEntriesRequest` for each message type instead of one generic "handle" to strengthen validation.  Then, use the operators like in:

```tla
 Network!DropMessage(logline.msg.to_node_id, LAMBDA msg: IsAppendEntriesRequest(msg, logline.msg.to_node_id, logline.msg.from_node_id, logline))
```

Additionally, instead of asserting that at least one message is dropped, we could assert that exactly one message gets dropped.  In other words, we can add `Network!OneLessMessage` next to `Network!OneMoreMessage`.

tla/consensus/ccfraft.tla
lemmy: This sounds significant enough to, perhaps, go into a dedicated/separate PR.

lemmy:
Last two commits also reduce or eliminate non-determinism from a few scenarios.

lemmy (approved):

PR #6010, opened 2024-02-09
Strengthen assertions in trace validation

lemmy:
Related https://github.com/microsoft/CCF/pull/6000

### DIFF
--- a/tla/consensus/Traceccfraft.tla
+++ b/tla/consensus/Traceccfraft.tla
@@ -158,6 +158,7 @@ IsTimeout ==
     /\ commitIndex[logline.msg.state.node_id] = logline.msg.state.commit_idx
     /\ leadershipState'[logline.msg.state.node_id] = ToLeadershipState[logline.msg.state.leadership_state]
     /\ membershipState[logline.msg.state.node_id] \in ToMembershipState[logline.msg.state.membership_state]
+    /\ Len(log[logline.msg.state.node_id]) = logline.msg.state.last_idx
 
 IsBecomeLeader ==
     /\ IsEvent("become_leader")
@@ -167,6 +168,7 @@ IsBecomeLeader ==
     /\ commitIndex[logline.msg.state.node_id] = logline.msg.state.commit_idx
     /\ leadershipState'[logline.msg.state.node_id] = ToLeadershipState[logline.msg.state.leadership_state]
     /\ membershipState[logline.msg.state.node_id] \in ToMembershipState[logline.msg.state.membership_state]
+    /\ Len(log[logline.msg.state.node_id]) = logline.msg.state.last_idx
     
 IsClientRequest ==
     /\ IsEvent("replicate")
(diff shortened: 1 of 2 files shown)


lemmy:
This PR "just" adds assertions that don't make any of the existing scenarios fail.  We should get this merged to avoid future regressions, while we keep digging why the assertions don't generally hold.

heidihoward (approved):

heidihoward:
Aside: I was surprised to see a 11/14% increase in the number of states for model checking with reconfiguration. This PR didn't seem to touch the model checking at all

tla/consensus/Traceccfraft.tla
```diff
@@ -147,12 +155,20 @@ IsTimeout ==
     /\ logline.msg.state.leadership_state = "Candidate"
     /\ Timeout(logline.msg.state.node_id)
     /\ Range(logline.msg.state.committable_indices) \subseteq CommittableIndices(logline.msg.state.node_id)
+    /\ commitIndex[logline.msg.state.node_id] = logline.msg.state.commit_idx
```
achamayou, line 158: This is very repetitive, I would have preferred if it had been factored out.

lemmy: Absolutely, I even started with an `IsState` definition that asserted `log_index`, `membership_state`, `leadership_state`, and` `commit_index`.  I decided to inline the four assertions for as long as not all actions are consistent.

lemmy:
> Aside: I was surprised to see a 11/14% increase in the number of states for model checking with reconfiguration. This PR didn't seem to touch the model checking at all

Where do I find the numbers that come out of TLC and go into cimetrics?

achamayou:
@lemmy in the TLC section of the (admittedly large) ccf-bot post on every PR:

image

lemmy:
@achamayou I meant the numbers that go into the chart generator. As @heidihoward observed, this commit should not cause a change in state-space statistics, and I'd like to check if there is a bug in TLC or cimetrics.

eddyashton:
I think this increase was not from this PR, but from something earlier. cimetrics plots a moving average, so in this case we see a plot halfway through `main`'s average being updated (ie, after only a few runs on `main`, where the average hasn't settled to the new value yet).

achamayou:
@lemmy sim https://github.com/microsoft/CCF/blob/f7dab4a9b7fbb665ce9206d6ffcd1babe7d00d7c/.azure-pipelines-templates/simulation.yml#L31
mc https://github.com/microsoft/CCF/blob/f7dab4a9b7fbb665ce9206d6ffcd1babe7d00d7c/.azure-pipelines-templates/model_checking.yml#L36

achamayou:
As @eddyashton says, the line is an exponentially weighted moving average, and so takes some time to catch up. The small blue dots are the actual data.

eddyashton:
#5978 is the PR that actually increased these numbers.

PR #6012, opened 2024-02-12
Tidy up TV cli again

achamayou:
- De-duplicate flags once more
- Attempt to nudge CI output towards something useful

### DIFF
--- a/tla/tlc.sh
+++ b/tla/tlc.sh
@@ -16,7 +16,7 @@ fi
 # Simulation, nor Model Checking seem to work well with G1GC, but Trace Validation does not for some reason, and is much slower unless -XX:+UseParallelGC is used.
 # In all cases, pauses don't matter, only throughput does.
 
-exec java -XX:+UseParallelGC -Dtlc2.tool.impl.Tool.cdot=true "${JVM_OPTIONS[@]}" -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -checkpoint 0 -lncheck final "${TLC_OPTIONS[@]}"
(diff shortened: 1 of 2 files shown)


tla/tlc.sh
```diff
@@ -16,7 +16,7 @@ fi
 # Simulation, nor Model Checking seem to work well with G1GC, but Trace Validation does not for some reason, and is much slower unless -XX:+UseParallelGC is used.
 # In all cases, pauses don't matter, only throughput does.
 
-exec java -XX:+UseParallelGC -Dtlc2.tool.impl.Tool.cdot=true "${JVM_OPTIONS[@]}" -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -checkpoint 0 -lncheck final "${TLC_OPTIONS[@]}"
+exec java -XX:+UseParallelGC -Dtlc2.tool.impl.Tool.cdot=true -Dtlc2.tool.queue.IStateQueue=StateDeque "${JVM_OPTIONS[@]}" -cp tla2tools.jar:CommunityModules-deps.jar tlc2.TLC "$@" -checkpoint 0 -lncheck final "${TLC_OPTIONS[@]}"
```
lemmy, line 19: StateDeque should not be used for model checking or simulation.

achamayou: @lemmy ah, I did not spot anything that suggested this was the case, the only documentation I could find was https://github.com/tlaplus/tlaplus/commit/62625342a01b96c2839c575b49dce1b8489fae94. Can this be split as `-trace-validate`, like `-simulation`? It would be both more convenient and safer.

lemmy: Yes, TLC should come with some sort of `-tracevalidate` flag eventually. Can you open an issue?

PR #6016, opened 2024-02-13
Raft fixes: Rollback safety despite NACKs, and only ACK up to incoming AE

eddyashton:
This PR combines #5991 and #6001.

I hoped they could be merged and fixed independently, but fixing either (in the implementation) in isolation raises trace validation failures that I believe (though haven't completely confirmed) boil down to the issue fixed by the other. So here's a PR containing both, which I think should finally be green.


src/consensus/aft/raft.h
```diff
@@ -1170,10 +1149,49 @@ namespace aft
...
+                i,
+                local_term,
+                i,
+                rollback_level);
+              rollback(rollback_level);
+              // Then continue to process this AE as normal
+            }
+            else
+            {
+              RAFT_FAIL_FMT(
```
eddyashton, line 1179: TODO: Log with tags.

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+
+### Fixed
+
+- Improvements to the Raft implementation, to retain commit safety and liveness despite message loss (#6016).
+
+### Added
+
+- Added 2 new log lines which may be helpful diagnostics in production deployments, both including a new `[rollback]` tag.
+  - `[rollback] ... Dropping conflicting branch` may be emitted after network partitions, and indicates that some unstable (non-committed) transactions have been lost. This is expected, but worth investigating if it occurs regularly - it is a sign of elections impacting service availability.
+  - `[rollback] ... Ignoring conflicting AppendEntries` could also be emitted after a network partition, but **should be reported to the CCF development team**. It is a sign of an unexpected execution path, which could lead to loss of liveness (inability to advance commit).
```
eddyashton, line 18: We may want to highlight this even more loudly. This path is retained only because we don't currently have a counter-example for it, so removing it is a larger (more dangerous) change. But if this is ever seen in the wild, we want to know and want to remove this path (likely removing the `is_new_follower` logic entirely).

src/consensus/aft/raft.h
```diff
@@ -1183,10 +1164,56 @@ namespace aft
...
+                local_term,
+                i,
+                rollback_level);
+              LOG_ROLLBACK_FMT(
+                "Dropping conflicting branch. Rolling back {} entries, "
+                "beginning with {}.{}.",
+                state->last_idx - rollback_level,
+                local_term,
+                i);
+              rollback(rollback_level);
```
eddyashton, line 1195: This is one of the key changes - moving this `rollback` from outside of the loop (whenever we see a conflict-y AppendEntries, even if it contains no heartbeat, we immediately rollback) to here (when considering each entry in an AppendEntries, if we see one which conflicts, we rollback).

Note that this is now the _only_ path where we reset `is_new_follower` - we're allowed _one conflicting AE which causes rollback_ per-term, where previously we'd only obey it if it was the _first_ AE we received (and eagerly reset `is_new_follower` otherwise).

src/consensus/aft/raft.h
```diff
@@ -1183,10 +1164,56 @@ namespace aft
...
+            }
+            else
+            {
+              LOG_ROLLBACK_FMT(
+                "Ignoring conflicting AppendEntries. Retaining {} entries, "
+                "beginning with {}.{}.",
+                state->last_idx - (i - 1),
+                local_term,
+                i);
+              return;
```
eddyashton, line 1207: For posterity: This branch is weird, and if we ever hit it it's a potential liveness hole. We have a node retaining a conflicting suffix, and refusing to roll it back. It will remain divergent (not contributing to commit) this term, and can only be brought in-sync in a future term.

A critical fix missed from earlier versions of this change is the `return` within this branch. If we hit this case, we log but then early-out, essentially ignoring the AE. We do not send the leader a NACK, since we would continue to ignore the resulting messages. If we're lucky, that may trigger `CheckQuorum` and a fresh election, avoiding a permanent liveness hole.

src/consensus/aft/raft.h
```diff
@@ -1369,29 +1398,51 @@ namespace aft
...
+    void send_append_entries_response_ack(
+      ccf::NodeId to, const AppendEntries& ae)
+    {
+      // If we get to here, we have applied up to r.idx in this AppendEntries.
+      // We must only ACK this far, as we know nothing about the agreement of a
+      // suffix we may still hold _after_ r.idx with the leader's log
+      const auto response_idx = ae.idx;
+      send_append_entries_response(
+        to, AppendEntriesResponseType::OK, state->current_view, response_idx);
+    }
```
eddyashton, line 1413: This is the second key fix, the "only ACK up to incoming AE" change. I was unable to reach this path without the rollback_safety fix, but after that this becomes a secondary clear safety violation.

tla/consensus/ccfraft.tla
```diff
@@ -981,20 +981,23 @@ AppendEntriesAlreadyDone(i, j, index, m) ==
                 IF membershipState = RetirementSigned /\ commitIndex' > RetirementIndex(i) 
                 THEN RetirementCompleted 
                 ELSE @]
-    /\ isNewFollower' = [isNewFollower EXCEPT ![i] = FALSE]
```
eddyashton, line 984: Keeping the spec up-to-date with the code. We now only reset `isNewFollower` when processing a conflict, it is unchanged on other AEs.

tla/consensus/ccfraft.tla
```diff
@@ -1080,7 +1082,8 @@ HandleAppendEntriesResponse(i, j, m) ==
           /\ leadershipState[i] = Leader \* Only Leaders need to tally append entries responses
           /\ m.success \* successful
           \* max(...) because why would we ever want to go backwards on a success response?!
-          /\ matchIndex' = [matchIndex EXCEPT ![i][j] = max(@, m.lastLogIndex)]
+          \* min(...) because we don't want to calculate a match past our own log's tail
+          /\ matchIndex' = [matchIndex EXCEPT ![i][j] = max(@, min(Len(log[i]), m.lastLogIndex))]
```
eddyashton, line 1086: Fixing a minor discrepancy between the spec and the implementation. Following the fixes above, we produce traces where this distinction actually matters, so update the spec to continue trace validation.

tla/consensus/ccfraft.tla
```diff
@@ -981,20 +981,23 @@ AppendEntriesAlreadyDone(i, j, index, m) ==
...
+    /\ UNCHANGED <<removedFromConfiguration, currentTerm, leadershipState, votedFor, isNewFollower, log, candidateVars, leaderVars>>
 
 \* Follower i receives an AppendEntries request m where it needs to roll back first
 \* This action rolls back the log and leaves m in messages for further processing
 ConflictAppendEntriesRequest(i, index, m) ==
     /\ Len(log[i]) >= index
+    /\ m.entries /= << >>
+    /\ \E idx \in 1..Len(m.entries) :
+        /\ (index + (idx - 1)) \in DOMAIN log[i]
+        /\ log[i][index + (idx - 1)].term # m.entries[idx].term
```
eddyashton, line 1000: After removing the `isNewFollower` reset from elsewhere in the spec, this enablement condition is required to pass simulation. I believe that without this, the spec says that _any_ AppendEntries can be treated a though it conflicts, and cause a rollback. This was never an issue previously, because simulation includes no message loss:
- The first AE in each term is a heartbeat
- Taking the conflict path on a heartbeat is a safe nop - we don't actually truncate any log
- After processing the AE by any path, we reset `isNewFollower[i]`
- Any future AEs in this term cannot take this conflict path, due to the `isNewFollower[i] = TRUE` enablement condition

But now we need to match the implementation's "one rollback message processed per-term" behaviour, for trace validation, so I think we need this enablement condition for simulation.

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+## Unreleased
+
+### Fixed
+
+- Improvements to the Raft implementation, to retain commit safety and liveness despite message loss (#6016).
+
+### Added
+
+- Added 2 new log lines which may be helpful diagnostics in production deployments, both including a new `[rollback]` tag.
+  - `[rollback] ... Dropping conflicting branch` may be emitted after network partitions, and indicates that some unstable (non-committed) transactions have been lost. This is expected, but worth investigating if it occurs regularly - it is a sign of elections impacting service availability.
```
achamayou, line 17: I suggest s/unstable/Pending/, because we've not used unstable before in the documentation, and we call these transactions Pending everywhere else.

CHANGELOG.md
achamayou: @andpiccione as discussed earlier this week.

achamayou:
@eddyashton I strongly suggest we poke at least the daily and threading canary on this change.

src/consensus/aft/raft.h
```diff
@@ -63,6 +63,8 @@
 #define RAFT_TRACE_JSON_OUT(json_object) \
   CCF_LOG_OUT(DEBUG, "raft_trace") << json_object
 
+#define LOG_ROLLBACK_FMT CCF_LOG_FMT(INFO, "rollback")
```
achamayou, line 66:
```suggestion
#define LOG_ROLLBACK_INFO_FMT CCF_LOG_FMT(INFO, "rollback")
```

I think having the log level consistently in the macro is really helpful when glancing at the code to find out what we expect to see logged where.

src/consensus/aft/raft.h
achamayou: > We have a node retaining a conflicting suffix, and refusing to roll it back. It will remain divergent (not contributing to commit) this term, and can only be brought in-sync in a future term.

That seems like a really good comment to have at :1201.

achamayou (approved):

tla/consensus/ccfraft.tla
```diff
@@ -981,20 +981,23 @@ AppendEntriesAlreadyDone(i, j, index, m) ==
...
               source         |-> i,
               dest           |-> j],
               m)
-    /\ UNCHANGED <<removedFromConfiguration, currentTerm, leadershipState, votedFor, log, candidateVars, leaderVars>>
+    /\ UNCHANGED <<removedFromConfiguration, currentTerm, leadershipState, votedFor, isNewFollower, log, candidateVars, leaderVars>>
 
 \* Follower i receives an AppendEntries request m where it needs to roll back first
 \* This action rolls back the log and leaves m in messages for further processing
 ConflictAppendEntriesRequest(i, index, m) ==
     /\ Len(log[i]) >= index
```
heidihoward, line 996: This line could be removed as its now covered by the new condition below

tla/consensus/ccfraft.tla
heidihoward: Not sure I understand why this change is needed. A leader should never receive an AE-ACK in the current term with a index greater than its log

heidihoward (approved):

tla/consensus/ccfraft.tla
eddyashton: I think you're right. The point where this is done in the implementation says `Potentially unnecessary safety check`. I tried removing both, but only in earlier versions that were failing TV or SIM for other reasons - it looks like this is now removable in both locations, so I'll try that.

eddyashton:
## 💔 All backports failed

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|release/4.x|An unhandled error occurred. Please see the logs for details|

### Manual backport
```
backport --pr 6016
```

### Questions ?

PR #6018, opened 2024-02-13
Retired committed hook receives whole config

achamayou:
No impact change extracted from #5973, expanding #5962 to provide raft with enough information to know when it can stop tracking nodes.

```
set_retired_committed(ccf::SeqNo seqno)
```

becomes

```
set_retired_committed(const std::vector<kv::NodeId>& node_ids)
```

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -309,13 +309,13 @@ namespace aft
           CCF_ASSERT(
             state->retirement_phase == kv::RetirementPhase::Completed,
             "Node is not retired, cannot become retired committed");
-          state->retirement_phase = kv::RetirementPhase::RetiredCommitted;
           CCF_ASSERT_FMT(
             state->retired_committed_idx == state->commit_idx,
             "Retired "
             "committed index {} does not match current commit index {}",
             state->retired_committed_idx.value_or(0),
             state->commit_idx);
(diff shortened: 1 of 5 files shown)


src/consensus/aft/raft.h
```diff
@@ -296,16 +296,29 @@ namespace aft
...
+          CCF_ASSERT(
+            state->retirement_phase == kv::RetirementPhase::Completed,
+            "Node is not retired, cannot become retired committed");
+          state->retirement_phase = kv::RetirementPhase::RetiredCommitted;
+          CCF_ASSERT_FMT(
+            state->retired_committed_idx == state->commit_idx,
+            "Retired "
+            "committed index {} does not match current commit index {}",
+            state->retired_committed_idx.value_or(0),
+            state->commit_idx);
```
eddyashton, line 318:
```suggestion
          CCF_ASSERT_FMT(
            state->retired_committed_idx == state->commit_idx,
            "Retired "
            "committed index {} does not match current commit index {}",
            state->retired_committed_idx.value_or(0),
            state->commit_idx);
          state->retirement_phase = kv::RetirementPhase::RetiredCommitted;
```

Minor: _if_ we ever fail one of these sanity check preconditions, prefer to update neither state value.

eddyashton (approved):

PR #6021, opened 2024-02-16
RaftDriver: Log whenever messages are dropped, to simplify trace validation

eddyashton:
Attempting to log whenever messages are dropped in scenarios, so that we can match the appropriate successor state directly (rather than generating all possible successors by permutations of message dropping).

~~Draft for now, as I've had to turn off refinement checking to get this far.~~


src/consensus/aft/raft.h
```diff
@@ -1506,6 +1496,16 @@ namespace aft
...
 
+      // Ignore if we're not the leader.
+      if (state->leadership_state != kv::LeadershipState::Leader)
+      {
+        RAFT_FAIL_FMT(
+          "Recv append entries response to {} from {}: no longer leader",
+          state->node_id,
+          from);
+        return;
+      }
```
eddyashton, line 1507: Moved this check _after_ the `RAFT_TRACING` log above, so that we go through an appropriate path in trace validation (match a `recv_append_entries_response` line, go through `RcvAppendEntriesResponse`, `DropResponseWhenNotInState`). Without this, we got no line to trace validate this message, but found an alternate compatible behaviour where this message was _dropped_.

src/consensus/aft/test/driver.h
```diff
@@ -732,13 +756,15 @@ class RaftDriver
...
       {
-        log_msg_details(src, dst, contents);
         _nodes.at(dst).raft->recv_message(
           src, contents.data(), contents.size());
         return true;
       }
     }
 
+    // Log that this message was dropped
+    log_msg_details(src, dst, contents, true);
```
eddyashton, line 766: There are _2_ places where we "drop" messages in the driver. One is for a `drop_pending_to` action, handled explicitly above. But we also allow connections to be severed, in which case we essentially drop it here (early-out in `send_single_message`, when the required connection doesn't exist). We log for both, so that (hopefully) every change to `messages` has a corresponding traceable event.

tla/consensus/Traceccfraft.cfg
```diff
@@ -20,7 +20,7 @@ POSTCONDITION
 \* matches the full log.  However, TLA cannot check this property directly because it
 \* is a hyperproperty.
 PROPERTIES
-    CCFSpec
+    \* CCFSpec
```
eddyashton, line 23: @lemmy: I've disabled this refinement check to debug the trace validation in isolation. I believe that's now passing for all current scenarios, but fails with the refinement check. Can you take a look at the TLA changes I've made (removed `DropMessages` from `TraceSpec`), and explain what we need to do to restore refinement checks?

eddyashton: Update: I think I've resolved this, and have now restored the refinement check and removed `DropMessages` from `Traceccfraft`. Please vet this change!

achamayou (approved):

lemmy (approved):

tla/consensus/Traceccfraft.cfg
lemmy: LGTM

lemmy:
@eddyashton This is good to go from my perspective, but you left the PR in draft mode.  Any reason not to merge into main?

PR #6044, opened 2024-03-04
Do not use removedFromConfiguration in ChangeConfigurationInt

achamayou:
A step towards removing the global removedFromConfiguration variable. Introduce a per-node hasJoined to model the fact that nodes can only join once.

### DIFF
--- a/tla/consensus/ccfraft.tla
+++ b/tla/consensus/ccfraft.tla
@@ -766,10 +766,10 @@ ChangeConfigurationInt(i, newConfiguration) ==
     /\ leadershipState[i] = Leader
     \* Configuration is not equal to the previous configuration.
     /\ newConfiguration /= MaxConfiguration(i)
-    \* CCF's integrity demands that a previously removed server cannot rejoin the network,
-    \* i.e., be re-added to a new configuration.  Instead, the node has to rejoin with a
-    \* "fresh" identity (compare sec 6.2, page 8, https://arxiv.org/abs/2310.11559).
-    /\ \A s \in newConfiguration: s \notin removedFromConfiguration
+    \* Nodes can only join a network once in CCF. This is enforced through the pre-consensus
+    \* join protocol that verifies the attestation of the joining node. The state machine of
+    \* the joining node never allows joining more than once.
+    /\ \A s \in (newConfiguration \ CurrentConfiguration(i)): leadershipState[s] = None
     \* See raft.h:2401, nodes are only sent future entries initially, they will NACK if necessary.
     \* This is because they are expected to start from a fairly recent snapshot, not from scratch.
     \* Note that the sentIndex is set to the log entry *before* the reconfiguration was added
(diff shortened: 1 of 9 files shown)


tla/consensus/ccfraft.tla
```diff
@@ -766,16 +766,16 @@ ChangeConfigurationInt(i, newConfiguration) ==
...
     \* Configuration is not equal to the previous configuration.
     /\ newConfiguration /= MaxConfiguration(i)
-    \* CCF's integrity demands that a previously removed server cannot rejoin the network,
-    \* i.e., be re-added to a new configuration.  Instead, the node has to rejoin with a
-    \* "fresh" identity (compare sec 6.2, page 8, https://arxiv.org/abs/2310.11559).
-    /\ \A s \in newConfiguration: s \notin removedFromConfiguration
+    \* Nodes can only join a network once in CCF. This is enforced through the pre-consensus
+    \* join protocol that verifies the attestation of the joining node. The state machine of
+    \* the joining node never allows joining more than once.
+    /\ \A s \in (newConfiguration \ MaxConfiguration(i)): leadershipState[s] = None
```
achamayou, line 772: This check is not sufficient, because leadershipState does not change atomically, and so this is possible:

```
      ( 1 :> {n2} @@
        12 :> {n2, n5} @@
        16 :> {n1, n2, n4} @@
        18 :> {n2, n5} @@
...
```

tla/consensus/ccfraft.tla
```diff
@@ -766,30 +775,31 @@ ChangeConfigurationInt(i, newConfiguration) ==
...
-    \* CCF's integrity demands that a previously removed server cannot rejoin the network,
-    \* i.e., be re-added to a new configuration.  Instead, the node has to rejoin with a
-    \* "fresh" identity (compare sec 6.2, page 8, https://arxiv.org/abs/2310.11559).
-    /\ \A s \in newConfiguration: s \notin removedFromConfiguration
-    \* See raft.h:2401, nodes are only sent future entries initially, they will NACK if necessary.
-    \* This is because they are expected to start from a fairly recent snapshot, not from scratch.
-    \* Note that the sentIndex is set to the log entry *before* the reconfiguration was added
-    \* This is to allow the send AE action to send an initial heartbeat which matches the implementation
-    /\ LET
-        addedNodes == newConfiguration \ CurrentConfiguration(i)
```
achamayou, line 778: This was not right, the diffing needs to happen using using the latest/MaxConfiguration, if the CurrentConfiguration is used, intermediary configurations may be present that already have the nodes.

eddyashton: Separated into #6047.

tla/consensus/ccfraft.tla
```diff
@@ -142,14 +142,22 @@ VARIABLE removedFromConfiguration
 RemovedFromConfigurationTypeInv ==
     removedFromConfiguration \subseteq Servers
 
+VARIABLE hasJoined
```
eddyashton, line 145: Can we add a comment on this/remove the now-inaccurate comment from `removedFromConfiguration`?

achamayou: Done

eddyashton (approved):

PR #6048, opened 2024-03-06
Reduce number of *generated* states.

lemmy:
Before:
31080255 states generated, 5045493 distinct states found, 0 states left on queue.

After:
20989269 states generated, 5045493 distinct states found, 0 states left on queue.


achamayou (approved):

achamayou:
Thank you for making that change! This is not super discoverable though :)

lemmy:
> Thank you for making that change! This is not super discoverable though :)

Here is ChatGPT4: "User: Assuming a sub-action A of the next-state relation of a TLA+ spec, how do I redefine the action to turn it into a no-op when checking the spec with TLC?"

To redefine a sub-action `A` of the next-state relation in a TLA+ specification to be a no-op (no operation) when checking the spec with the TLC model checker, you can override the definition of `A` in your model configuration to essentially make it do nothing, thereby not affecting the state. This can be useful for testing or isolating parts of your specification.

Here's a general approach on how to achieve this:

1. **Identify the Sub-action to Override**: Determine the specific sub-action `A` that you wish to redefine as a no-op. This action is typically defined in your TLA+ spec file (`.tla` file) and is part of the `Next` definition that describes the possible state transitions.

2. **Redefine the Sub-action as a No-op**: In TLA+, a no-op can be represented by stating that the state does not change. This is done by asserting that all variables in the state remain the same.

   If your spec has variables `x`, `y`, `z`, a no-op action could be defined as:

   ```tla
   A_noop == UNCHANGED <<x, y, z>>
   ```

   This definition says that the action `A_noop` results in no changes to the variables `x`, `y`, and `z`, effectively making it a no-op.

3. **Override the Sub-action in the Model**: When configuring your model in the TLC model checker, you can override the original action `A` with `A_noop`. This is done in the model's configuration file (`.cfg` file).

   - In the `.cfg` file, add a line to redefine `A` with `A_noop`. It might look something like this:

     ```
     A == A_noop
     ```

   This tells TLC to use `A_noop` wherever `A` is referenced in the specification, effectively turning `A` into a no-op during model checking.

4. **Run TLC Model Checker**: With the `.cfg` file updated, run the TLC model checker on your specification. TLC will use the overridden definition of `A`, treating it as a no-op during the analysis.

By carefully applying this technique, you can isolate or ignore specific actions during model checking, which can be helpful for debugging or understanding the impact of certain actions within your TLA+ specification. Remember to revert or adjust these overrides as necessary based on your testing or analysis goals.

achamayou:
@lemmy I did 2. and 3. (except with the correct syntax `<-`, not `==`). There's no mention of overriding with `FALSE`?

PR #6050, opened 2024-03-06
Redirection support, as alternative to request forwarding

eddyashton:
CCF now supports a mode where HTTP redirect responses are returned, rather than relying on internal forwarding. This can be used to ensure that write requests are executed on a primary, even if they are initially sent to a backup.

This behaviour is enabled by setting the `redirections` field in an `rpc_interface` within cchost's launch config. This can be configured to redirect either directly to a node (if each node has a distinct, accessible name), or to a static load balancer address, depending on the current deployment.

See docs (after this PR has merged) for description of [redirection behaviour](https://microsoft.github.io/CCF/main/architecture/request_flow.html#redirection-flow) and [configuration](https://microsoft.github.io/CCF/main/operations/configuration.html#redirections).


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [5.0.0-dev15]
+
+[5.0.0-dev15]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.0-dev15
+
+### Added
+
```
achamayou, line 13: I suggest swapping the PR description for this more up to date summary.

doc/schemas/gov_openapi.json
```diff
@@ -1283,7 +1325,7 @@
   "info": {
     "description": "This API is used to submit and query proposals which affect CCF's public governance tables.",
     "title": "CCF Governance API",
-    "version": "4.1.3"
+    "version": "4.1.4"
```
achamayou, line 1328: Arguably 4.2.0, but this is a pretty obscure version number.

src/node/rpc/frontend.h
```diff
@@ -198,6 +199,173 @@ namespace ccf
...
+        }
+
+        case (ccf::endpoints::RedirectionStrategy::ToPrimary):
+        {
+          const bool is_primary =
+            (consensus != nullptr) && consensus->can_replicate();
+
+          // Note: This check is included for parity with forwarding. If we
+          // should redirect, but can't for fundamental early-node-lifecycle
+          // reasons, should we try to execute it locally?
```
achamayou, line 300: I don't think we should try to execute locally, the mixed forward/redirect situation is sufficiently complicated as it is.

tests/governance_js.py
```diff
@@ -69,9 +69,24 @@ def set_service_recent_cose_proposals_window_size(proposal_count):
...
 
+def choose_node(network):
+    # Ideally, this would use find_random_node - you should be able to use any
+    # node for governance.
+    # However, many of these tests include a pattern of
+    #   POST /proposal
+    #   GET /proposal
+    # If the former request is redirected, then the latter may fail (essentially
+    # reading stale state, assuming session consistency that doesn't exist).
+    # return network.find_random_node()
```
achamayou, line 80: This may be a good motivation to prototype ETag-based concurrency in the governance (at first glance, the proposalId would make a decent choice).

achamayou (approved):

doc/schemas/gov_openapi.json
eddyashton: The API here is only updated in response fields - we're not asking users to submit new bodies or new requests, just reporting a new field on the node config of some responses. So I think a minor minor patch version is sufficient.

tests/governance_js.py
eddyashton: Good suggestion. Opened an issue to track this: https://github.com/microsoft/CCF/issues/6081

src/node/rpc/frontend.h
eddyashton: Agreed. Tracking this as the second follow-up TODO here: https://github.com/microsoft/CCF/issues/6082

I expect we can remove this, but will need to plumb through some minor test changes to do so.

doc/schemas/gov_openapi.json
achamayou: Sure, but it does seem like a new feature, which results in new behaviour if opted into. Anyway, pretty moot, nobody looks at this number.

PR #6052, opened 2024-03-06
Get `kv_bench` running again

eddyashton:
Noticed completely by accident on another PR. The combination of builds/tests we currently run never includes both CMake's `LONG_TESTS=ON` and the `benchmark` label to CTest, resulting in several picobenchmarks being skipped. One of those (`kv_bench`) was crashing on startup due to a missing initialization step.

I think these are all fast enough that they shouldn't be gated? Maybe they should be blocked in Debug runs or similar.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1197,15 +1197,13 @@ if(BUILD_TESTS)
       LINK_LIBS ccf_kv.host
     )
 
-    if(LONG_TESTS)
-      add_picobench(
-        kv_bench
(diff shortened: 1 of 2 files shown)


achamayou:
`LONG_TESTS` unfortunately does two things: decide which tests we build (because some are or at least were expensive to build), and decide which ones to run/parametrise them to take longer. `benchmark` just decides what we run. I think we would ideally only have a single switch to decide what test configuration is produced.

achamayou (approved):

PR #6066, opened 2024-03-15
Add test of JS converter behaviour, and a `checkedJson` converter which forbids certain types

eddyashton:
I was really hoping to improve our definition of `JsonCompatible` (currently `any`) to accurately describe only types which can be round-tripped through `JSON.stringify`+`JSON.parse`. However it appears this is still not something that can be spelt in TypeScript, or at least not something that _I_ can spell.

So instead of a type-check, this includes a dynamic run-time check - a new converter which examines every object property and array member, and throws an error if it finds a `Map` or a `Date`.

Additionally, and more generally useful, this PR includes some new endpoints which test the behaviour of our converters, what they accept and where they produce compile-time or run-time errors. These are a little verbose and still not exhaustive, but at least give us clear regression tests should we find ways to lock this down further in future.


js/ccf-app/src/converters.ts
```diff
@@ -56,6 +63,25 @@ function checkString(val: any) {
   }
 }
 
+function checkJsonSafe(val: any) {
+  // Hard to be exhaustive, but throw errors for any Map or Date elements found
```
achamayou, line 67: It seems like it would be safer to allow known safe types (Number, String, Object, Array etc)?

achamayou (approved):

PR #6069, opened 2024-03-15
TLA comment update

achamayou:
Quick cleanup pass over comments and the spec. Amusingly (?) every single raft.h line number reference was wrong.

### DIFF
--- a/tla/consensus/ccfraft.tla
+++ b/tla/consensus/ccfraft.tla
@@ -114,57 +114,12 @@ Nil ==
 ------------------------------------------------------------------------------
 \* Global variables
 
-\* Each server keeps track of the active configurations.
-\* This includes the current configuration plus any pending configurations.
-\* The current configuration is the initial configuration or the last committed reconfiguration.
-\* The pending configurations are reconfiguration transactions that are not yet committed.
-\* Each server's configurations is indexed by the reconfiguration transaction index,
-\* except for the initial configuration which has index 0 (note that the log in 1-indexed).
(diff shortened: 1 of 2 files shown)


tla/consensus/ccfraft.tla
```diff
@@ -716,7 +712,7 @@ AppendEntries(i, j) ==
                \/ m.entries # <<>>
             /\ Send(m)
             \* Record the most recent index we have sent to this node.
-            \* (see https://github.com/microsoft/CCF/blob/9fbde45bf5ab856ca7bcf655e8811dc7baf1e8a3/src/consensus/aft/raft.h#L935-L936)
+            \* raft.h::send_append_entries
```
lemmy, line 715: Why not keep both?

lemmy (approved):

tla/consensus/ccfraft.tla
achamayou: The line numbers go out of date, this one was too. Function/type/variable names are more stable. On the TLA side, I don't think we have an easy way to do this, but for the C++, we do produce XML via doxygen, used by Sphinx ([example](https://microsoft.github.io/CCF/main/build_apps/kv/api.html#map)). We could build a tool that parses some of the comments in the TLA file, and checks references (we could make up something like `\* @xref:raft.h:aft::Aft::send_append_entries()` or whatever other prefix that's reasonably unique), to look them up in the doxygen XML, and validate they exist to avoid breaking changes.

eddyashton (approved):

PR #6074, opened 2024-03-18
Add action number plotting script

achamayou:
Sample run:

```
$ ./tlc.sh -workers auto -config consensus/MCccfraftAtomicReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftAtomicReconfig.trace.tla -dumpTrace json MCccfraftAtomicReconfig.json -coverage 1 -nowarning & 
$ cat MCccfraftAtomicReconfig_coverage_** | python3 actions.py
```

image

Potentially useful to understand why model checking is taking so long.


lemmy:
We can get rid of parsing TLC's command-line output by dumping TLC's coverage data by conjoining `/\ ndJsonSerialize(CoverageFilename, <<TLCGet("spec")>>)` to https://github.com/microsoft/CCF/blob/main/tla/StatsFile.tla.  That also gives us the number of unique variable values, which is not part of TLC's command-line output.

Related:
https://github.com/tlaplus/tlaplus/commit/d4320ca23cec83abb24cafae570109141d13a50c
https://github.com/tlaplus/tlaplus/commit/5148e1131129dafe4a823b51b8895052eb56a7ef
https://github.com/tlaplus/tlaplus/commit/84ac43463380ee6ceb02b245c63d223adaaf0b1b

achamayou:
@lemmy does that get written continuously, or once at the end? The advantage of the grepping (clunking as it otherwise is) is that you can run it at any time during the execution, and still get partial results.

lemmy:
> @lemmy does that get written continuously, or once at the end? The advantage of the grepping (clunky as it otherwise is) is that you can run it at any time during the execution, and still get partial results.

Yes, `TLCGet("spec")` contains the newest values.  In other words, you get finer granularity compared to the cli.

achamayou:
@lemmy had a quick go with:

```
diff --git a/tla/StatsFile.tla b/tla/StatsFile.tla
index 8fd576267..4a1aafd75 100644
--- a/tla/StatsFile.tla
+++ b/tla/StatsFile.tla
@@ -5,10 +5,15 @@ EXTENDS TLC, Json, Sequences
 CONSTANT StatsFilename
 ASSUME StatsFilename \in STRING
 
+\* Filename to write TLC coverage to
+CONSTANT CoverageFilename
+ASSUME CoverageFilename \in STRING
+
 \* Writes TLC stats (such as number of states and duration) to StatsFilename in ndJson format
 \* Specify WriteStatsFile as a postcondition to write the stats file at the end of model checking
 WriteStatsFile == 
     /\ PrintT("Writing stats to file: " \o StatsFilename)
     /\ ndJsonSerialize(StatsFilename, <<TLCGet("stats")>>)
+    /\ ndJsonSerialize(CoverageFilename, <<TLCGet("spec")>>)
 
 ====
\ No newline at end of file
diff --git a/tla/consensus/MCccfraft.cfg b/tla/consensus/MCccfraft.cfg
index 87fbe2ff9..bc368753d 100644
--- a/tla/consensus/MCccfraft.cfg
+++ b/tla/consensus/MCccfraft.cfg
@@ -8,6 +8,7 @@ CONSTANTS
     RequestLimit = 3
 
     StatsFilename = "MCccfraft_stats.json"
+    CoverageFilename = "MCccfraft_coverage.json"
```

but the file definitely does not get produced until the model check is complete, which makes sense because WriteStatsFile is a post condition. There would need to be a periodic call instead (and ideally just append recently produced lines).

lemmy:
You can e.g. dump `TLCGet("spec")` every 10k distinct states in a state-constraint or invariant:

```tla
SomeStateConstraint ==
    IF TLCGet("distinct") % 10000 = 0 THEN ndJsonSerialize(CoverageFilename, <<TLCGet("spec")>>) ELSE TRUE
```

Note that I've been considering adding something like a "progress" hook to TLC that periodically evaluates a user-defined expression. I can revisit it if SomeStateConstraint isn't good enough, i.e., adds too much overhead.

.azure-pipelines-templates/model_checking.yml
```diff
@@ -24,14 +24,25 @@ jobs:
...
+      - script: ./tlc.sh -workers auto -config consensus/MCccfraftAtomicReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftAtomicReconfig.trace.tla -dumpTrace json MCccfraftAtomicReconfig.json -coverage 1 -nowarning
         workingDirectory: tla
         displayName: MCccfraftAtomicReconfig.cfg
 
-      - script: ./tlc.sh -workers auto -config consensus/MCccfraftWithReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftWithReconfig.trace.tla -dumpTrace json MCccfraftWithReconfig.json
+      - script: ./tlc.sh -workers auto -config consensus/MCccfraftWithReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftWithReconfig.trace.tla -dumpTrace json MCccfraftWithReconfig.json -coverage 1 -nowarning
         workingDirectory: tla
         displayName: MCccfraftWithReconfig.cfg
 
+      - script: |
```
achamayou, line 35: It looks like coverage runs are about 2x slower than normal, so while this is interesting as a one-off experiment, I do not want to merge it. The model checking is slow enough as it is.

lemmy: That's about the expected slowdown.  Perhaps, we can collect coverage data every nth run?

achamayou: Ideally we'd have a daily run, and a way to distinguish it enough in the yaml that we can parametrise the tasks to only do coverage then. Let's see how much of a pain that is.

.azure-pipelines-templates/model_checking.yml
```diff
@@ -27,13 +27,39 @@ jobs:
...
+          rm -rf env
+        workingDirectory: tla
+        condition: eq(variables['Build.CronSchedule.DisplayName'], 'Daily build')
 
       - task: PublishPipelineArtifact@1
         inputs:
           artifactName: "Model Checking Traces"
           targetPath: tla
-        condition: or(failed(), canceled())
+          condition: or(or(failed(), canceled()), eq(variables['Build.CronSchedule.DisplayName'], 'Daily build'))
```
eddyashton, line 65:
```suggestion
        condition: or(or(failed(), canceled()), eq(variables['Build.CronSchedule.DisplayName'], 'Daily build'))
```

I think this is now indented too far, so the condition isn't applied?

eddyashton (approved):

PR #6088, opened 2024-03-22
Endorsement checks should not be an assert

achamayou:


src/node/node_state.h
```diff
@@ -398,52 +398,54 @@ namespace ccf
...
-              (qi.format != QuoteFormat::oe_sgx_v1 && qi.endorsements.empty()),
-            "SGX quote generation should have already fetched endorsements");
+          quote_endorsements_client->fetch_endorsements();
+          return;
+        }
+
+        CCF_ASSERT_FMT(
+          (qi.format == QuoteFormat::oe_sgx_v1 && !qi.endorsements.empty()) ||
+            (qi.format != QuoteFormat::oe_sgx_v1 && qi.endorsements.empty()),
+          "SGX quote generation should have already fetched endorsements");
```
eddyashton, line 444: Should this also be promoted from `ASSERT` to `throw`?

eddyashton (approved):

src/node/node_state.h
achamayou: Yes, might as well.

PR #6102, opened 2024-03-27
Add `AllOf` authentication policy

eddyashton:
We provide some built-in authentication policies, and a way to write your own, and a way to combine multiple policies with a logical _OR_ (by listing multiple policies). This adds support for combining policies with a logical _AND_, so that we can recombine existing policies into something stricter.

No arbitrary nesting, because it is not needed.


js/ccf-app/src/endpoints.ts
```diff
@@ -103,6 +103,7 @@ export interface Request<T extends JsonCompatible<T> = any> {
   caller?: AuthnIdentity;
 }
 
+// TODO: Describe AllOfAuthnIdentity
```
eddyashton, line 106: Need to do this before merging.

Worryingly it looks like these types are not tested at all currently, so this probably means moving the `multi_auth` endpoint to TS...

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Authentication policies can now be conjoined (AND) together, in addition to the previous disjoint (OR) behaviour. The new `ccf::AllOfAuthnPolicy` takes a collection of other policies, _all of which must be true_ for this auth policy to pass. In JS, [this can be configured](https://microsoft.github.io/CCF/main/build_apps/js_app_bundle.html#allofauthnpolicy) in the `app.json` as `"authn_policies": [{ "all_of": ["policy_a", "policy_b"] }]`.
```
achamayou, line 15: allOf?

samples/apps/logging/js/src/logging.js
```diff
@@ -407,45 +407,67 @@ export function custom_auth(request) {
...
-    );
-    lines.push(
-      `The caller is identified by a COSE Sign1 with content of size:\n${request.caller.cose.content.byteLength}`,
-    );
+    describe_cose_ident(lines, request.caller);
   } else if (request.caller.policy === "no_auth") {
-    lines.push("Unauthenticated");
-    lines.push("The caller did not provide any authenticated identity");
+    describe_noauth_ident(lines, request.caller);
+  } else if (request.caller.policy === "jwt+user_cert+user_cose_sign1") {
```
achamayou, line 463: Could we make this an array of policies instead? That seems a lot cleaner.

eddyashton: Sure. There's a minor caveat that this may not be the same order as was specified in the `app.json` (it's reconstructed from a map's keys, so becomes lexicographically ordered), but it does give nicer access/iteration behaviour so I've changed to this.

CHANGELOG.md
eddyashton: The surrounding keys are all `little_snakes`, so I think `allOf` looks out of place?

```
    "/multi_auth": {
      "post": {
        "js_module": "logging.js",
        "js_function": "multi_auth",
        "forwarding_required": "always",
        "authn_policies": [
          {
            "allOf": ["user_cert", "jwt", "user_cose_sign1"]
          },
          "user_cert",
          ...
        ],
        "mode": "readwrite",
        "openapi": {},
        "interpreter_reuse": { "key": "singleton_interpreter" }
      }
```

achamayou: That seems reasonable, in which case I think the .rst needs fixing, that's where I saw "allOf".

achamayou (approved):

doc/build_apps/js_app_bundle.rst
```diff
@@ -74,6 +74,24 @@ Each endpoint object contains the following information:
...
+    To combine policies so that they must `all` pass, you may instead pass an object with an ``allOf`` key as an element of this list.
+    For example, this endpoint will test (in order of preference) for `a member cert`, then `a user cert AND a JWT`, then `a JWT`:
+
+    .. code-block:: json
+
+        {
+          "authn_policies": [
+            "member_cert",
+            {
+              "allOf": ["user_cert", "jwt"]
```
eddyashton, line 89:
```suggestion
              "all_of": ["user_cert", "jwt"]
```

Issue #6081, opened 2024-03-21
Explore ETAGs support to provide session consistency


eddyashton:
We believe it should be possible to provide a HTTP-standard form of session consistency by processing ETAG headers in CCF. This will need to be domain/app-specific, but we could prototype it within governance to confirm it is viable.

achamayou:
Having thought about this some more, I am not sure the governance flow is ideal for this, because:

> The ETag (or entity tag) HTTP response header is an identifier for a specific version of a resource. [0]

But governance resources (in the HTTP sense of the word), proposals and ballots, are not mutable and only ever have a single version. Creation produces a new resource, and the outcome of a racing GET will not change if an ETag is specified. The same applies to ballots, when they are submitted. Re-submission is a failure unless the ballot is the same. Similarly, withdraw would not behave differently with ETags.

An argument could be made for capturing not just the proposal but proposal + existing ballots in an ETag, to support conditional submission ("submit this ballot on this proposal only if exactly these other ballots have been posted"). This seems a bit niche, but seems like the only conceivable use of ETags in governance.

Similarly, there is no obvious condition to be applied on ack/share refresh or submission. There is already a pre-condition check and failure if the value is not current.

A better sample would be the logging app I think, where the ETag could be the digest of the value or the version of last write, and where submission with If-Match would fail if the value has changed in the meantime.

[0] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

achamayou:
> the ETag could be the digest of the value or the version of last write

Using the version of last write precludes returning an ETag on log entry creation at the moment, unless using a local_commit_handler (C++ only), because the version is not set by the time the user handler completes.

The logging example in its current form is also not completely ideal because entries are not HTTP resources, and are instead POST'ed and queried.

eddyashton:
We have a demo of this, and helpers in the public C++ API, in #6110.

PR #6110, opened 2024-04-08
ETag demo in logging sample

achamayou:
Demo illustrating the use of ETags in a CCF application, #6081, see [end to end test for usage sample](https://github.com/microsoft/CCF/pull/6110/files#diff-3b229b160527d90c259f61604ec7cd0d96f94449a5131c4c8eac0825d2e03b13L1800).

Add request headers `If-Match` and `If-None-Match`, as well as response header `ETag` on:

- `POST /app/log/public`
- `GET /app/log/public?id={id}`
- `DELETE /app/log/public?id={id}`

Documented as part of the C++ app sample.


samples/apps/logging/logging.cpp
```diff
@@ -11,13 +11,15 @@
...
 #include "ccf/historical_queries_adapter.h"
+#include "ccf/http_etag.h"
 #include "ccf/http_query.h"
 #include "ccf/indexing/strategies/seqnos_by_key_bucketed.h"
 #include "ccf/indexing/strategy.h"
 #include "ccf/json_handler.h"
 #include "ccf/version.h"
 
 #include <charconv>
+#include <regex>
```
eddyashton, line 22:
```suggestion
```

Think we can remove this here, as it's only used in `http_etag.h`

tests/e2e_logging.py
```diff
@@ -1798,6 +1798,168 @@ def test_basic_constraints(network, args):
...
+    primary, _ = network.find_primary()
+
+    with primary.client("user0") as c:
+        doc = {"id": 999999, "msg": "hello world"}
+        etag = sha256(doc["msg"].encode()).hexdigest()
+
+        # POST ETag matches value
+        r = c.post("/app/log/public", doc)
+        assert r.status_code == http.HTTPStatus.OK
+        assert r.headers["ETag"] == etag, r.headers["ETag"]
```
eddyashton, line 1811:
```suggestion

        # POST ETag matches value
        r = c.post("/app/log/public", doc)
        assert r.status_code == http.HTTPStatus.OK
        assert "ETag" in r.headers, r.headers
        etag = r.headers["ETag"]
```

I don't know if should make any assertions about the ETag's value? We happen to know how it's constructed in this instance, but in general its an opaque identifier.

tests/e2e_logging.py
```diff
@@ -1798,6 +1798,168 @@ def test_basic_constraints(network, args):
...
+        assert r.status_code == http.HTTPStatus.PRECONDITION_FAILED
+
+        # POST If-Match: matching ETag returns 200
+        r = c.post("/app/log/public", doc, headers={"If-Match": f'"{etag}"'})
+        assert r.status_code == http.HTTPStatus.OK
+        etag = sha256(doc["msg"].encode()).hexdigest()
+        assert r.headers["ETag"] == etag, r.headers["ETag"]
+
+        # POST If-Match: mutiple, one which is matching, ETags returns 200
+        r = c.post("/app/log/public", doc, headers={"If-Match": f'"{etag}", "abc"'})
```
eddyashton, line 1857: Other test cases we could add to be thorough:
```
If-Match: "abc", <ETag>  # ETag matches, but is not first in list of multiple
If-Match: "abc", "def"      # Multiple ETags, none match
```

js/ccf-app/src/crypto.ts
```diff
@@ -108,7 +108,7 @@ export const eddsaPemToJwk = ccf.crypto.eddsaPemToJwk;
 export const pubJwkToPem = ccf.crypto.pubJwkToPem;
 
 /**
- * @inheritDoc global!CCFCrypto.JwkToPem
+ * @inheritDoc global!CCFCrypto.jwkToPem
```
eddyashton, line 111: These docs fixes are good, but can we put them in a separate PR?

include/ccf/http_etag.h
```diff
@@ -0,0 +1,95 @@
...
+          return;
+        }
+
+        if (match_header.value() == "*")
+        {
+          any_value = true;
+          return;
+        }
+
+        std::regex etag_rx("\\\"([0-9a-f]+)\\\",?\\s*");
```
eddyashton, line 46: Trying to understand if this regex is necessary, or overkill. Could we get the same result with:
```
etag_values = nonstd::split(match_header.value(), ",")
foreach tag_value in etag_values:
  Strip whitespace
  Check it starts and ends with "
  Append to if_etags
```

This would be more readable than a (heavily-escaped) regex, if I'm not missing anything.

include/ccf/http_etag.h
```diff
@@ -0,0 +1,95 @@
...
+      bool any_value = false;
+      /// If-Match header is present and has specific etag values
+      std::set<std::string> if_etags;
+
+    public:
+      /** Construct a Matcher from a match header if present
+       *
+       * Note: Weak tags are not supported.
+       */
+      Matcher(const std::optional<std::string>& match_header)
```
eddyashton, line 32: Could this take a `string` rather than an `optional`, _only_ be constructed when the header was present, and then this matcher could drop all of its `empty` state/code? I think all of the paths are supported identical (checks on `.empty()` on a `optional<string>` rather than a `Matcher`, an error if both _headers_ are present, errors thrown if the content is not a valid ETags list), just in maybe different orders.

tests/e2e_logging.py
achamayou: I general sure, in this particular test, it seems reasonable to be precise?

js/ccf-app/src/crypto.ts
achamayou: Sure, they were just drive by fixing of Sphinx warnings.

include/ccf/http_etag.h
achamayou: We probably could, I started down this road because I was hoping to capture all the values in one match(), but the API is awful and that's just not possible. The rest is path dependent.

include/ccf/http_etag.h
achamayou: Probably, I started that way but this seemed to reduce the number of branches somewhat. I'll have another look.

js/ccf-app/src/crypto.ts
achamayou: #6120

PR #6111, opened 2024-04-08
Pass proposalId to resolve()

### LINKED ISSUE #5995: Pass proposalId to resolve()
Some users may want to have the proposalId when resolving, because their logic depends on distinguishing "other pending proposals" and applying rules to that.

This should be a backwards compatible change because JS ignores extra arguments passed to functions, and a fairly simple change.

achamayou:
Resolves #5995 by passing the `proposal_id` to `resolve()`. The change is backwards-compatible.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - Authentication policies can now be conjoined (AND) together, in addition to the previous disjoint (OR) behaviour. The new `ccf::AllOfAuthnPolicy` takes a collection of other policies, _all of which must be true_ for this auth policy to pass. In JS, [this can be configured](https://microsoft.github.io/CCF/main/build_apps/js_app_bundle.html#allofauthnpolicy) in the `app.json` as `"authn_policies": [{ "all_of": ["policy_a", "policy_b"] }]`.
+- `proposal_id` is now passed to `resolve(proposal, proposer_id, votes, proposal_id)`, allowing proposals to consider other pending proposals in their resolution process. (#5995)
```
eddyashton, line 16: It's `proposalId` in most places in JS, and only `proposal_id` in the test constitution. Notably `samples/constitutions/default/resolve.js` currently has the signature `resolve(proposal, proposerId, votes)`. I think we should call it `proposalId` (and `proposerId`) in this CHANGELOG, and update the signature in the sample/default constitution(s?).

eddyashton (approved):

Issue #6108, opened 2024-04-05
Provide proposalId, votes information in action's apply() method


gaurav137:
We have a scenario where as part of accepting a proposal we want to capture the proposal Id and the votes information (which members voted yes/no) in our governance table. Then later on we have UX that wants to expose this information to make decisions based on which member participated or not. 
Currently the action's aapply() method has access to only args. If the same could also get more data like (proposalId, proposerId, votes) then one can capture this info for later consumption.

eddyashton:
Resolved in #6114, available on main, will be in a dev release shortly.

PR #6114, opened 2024-04-09
Make Accepted proposal details visible to constitution's `apply()`

eddyashton:
(Includes #6112, #6113, #6115, so merge those first)

Goal of the change is as in the title, resolving #6108. Includes an overwrought e2e test. Only changes the new governance API, with associated cleanup that I _think_ has no other semantic (API) impact.


src/node/gov/handlers/proposals.h
```diff
@@ -250,26 +253,31 @@ namespace ccf::gov::endpoints
...
           remove_all_other_non_open_proposals(tx, proposal_id);
-          final_votes = std::unordered_map<ccf::MemberId, bool>();
-          for (auto& [mid, vote] : votes)
-          {
-            final_votes.value()[mid] = vote;
-          }
 
           if (proposal_info.state == ProposalState::ACCEPTED)
           {
+            // Write now-permanent values back to proposal_state, and into the
```
achamayou, line 274: I think we want to do this for REJECTED also, this is about final vs OPEN, not ACCEPTED.

achamayou: Plus a question mark for withdrawn proposals. I claim it's not particularly useful to evaluate final votes and store them in that case, but it may be arguable.

achamayou:
Two suggestions:

1. The usual daily canary to get the long LTS
2. Adding a pass over ledgers, post-run, checking that len(ballots) = len(v for v final_votes.values() if v), for the final states where we have decided to store them.

achamayou:
@eddyashton this feels like something that deserves a mention in the CHANGELOG.

doc/schemas/gov_openapi.json
```diff
@@ -2463,7 +2463,7 @@
             "member_cose_sign1": []
           }
         ],
-        "summary": "Ballots submitted against a proposed change to the service",
+        "summary": "Submit a ballot for a proposed changed to the service",
```
achamayou, line 2466:
```suggestion
        "summary": "Submit a ballot for a proposed change to the service",
```

include/ccf/service/tables/gov.h
```diff
@@ -31,10 +31,14 @@ namespace ccf
...
     using VoteFailures = std::unordered_map<ccf::MemberId, Failure>;
 
+    /// Proposal metadata stored in the KV
     struct ProposalInfo
     {
+      /// Member ID of the proposer
       ccf::MemberId proposer_id;
+      /// Proposal state
       ccf::ProposalState state;
+      /// Ballots (scripts) submitted for the proposal
```
achamayou, line 41: imho those descriptions do not add a lot given the names and types. "Javascript modules exporting a single 'vote' function" would be a more useful description for ballot for example.

include/ccf/service/tables/gov.h
```diff
@@ -45,11 +49,16 @@ namespace ccf
     DECLARE_JSON_OPTIONAL_FIELDS(
       ProposalInfo, final_votes, vote_failures, failure);
 
+    /// Proposal summary constructed while executing/resolving proposal ballots
```
achamayou, line 52: Same as above

src/node/gov/handlers/proposals.h
```diff
@@ -131,10 +133,14 @@ namespace ccf::gov::endpoints
       ccf::jsgov::ProposalInfo& proposal_info,
       const std::string& constitution)
     {
-      std::vector<std::pair<MemberId, bool>> votes;
-      std::optional<jsgov::Failure> failure = std::nullopt;
-      std::optional<ccf::jsgov::Votes> final_votes = std::nullopt;
-      std::optional<ccf::jsgov::VoteFailures> vote_failures = std::nullopt;
+      // Create some temporaries to store resolution progress. These are written
+      // to proposal_info, and the KV, only for completed (accepted) proposal.
```
achamayou, line 137: Surely they should be written for failed/rejected proposals too?

src/node/gov/handlers/proposals.h
```diff
@@ -145,7 +151,7 @@ namespace ccf::gov::endpoints
           mb,
           "vote",
           fmt::format(
-            "public:ccf.gov.proposal_info[{}].ballots[{}]", proposal_id, mid));
+            "public:ccf.gov.proposals_info[{}].ballots[{}]", proposal_id, mid));
```
achamayou, line 154: Using `{jsgov::Tables::PROPOSALS_INFO}` would seem preferable?

src/node/gov/handlers/proposals.h
```diff
@@ -131,10 +133,14 @@ namespace ccf::gov::endpoints
...
     {
-      std::vector<std::pair<MemberId, bool>> votes;
-      std::optional<jsgov::Failure> failure = std::nullopt;
-      std::optional<ccf::jsgov::Votes> final_votes = std::nullopt;
-      std::optional<ccf::jsgov::VoteFailures> vote_failures = std::nullopt;
+      // Create some temporaries to store resolution progress. These are written
+      // to proposal_info, and the KV, only for completed (accepted) proposal.
+      ccf::jsgov::Votes votes = {};
+      ccf::jsgov::VoteFailures vote_failures = {};
+      std::optional<ccf::jsgov::Failure> failure = std::nullopt;
```
achamayou, line 140: Is failure ever set? I'm not sure about the split, it doesn't seem to simplify the code much and means we allocate the vector every time.

tests/governance.py
```diff
@@ -762,12 +762,16 @@ def js_gov(args):
         governance_js.test_pure_proposals(network, args)
         governance_js.test_set_constitution(network, args)
         governance_js.test_proposals_with_votes(network, args)
-        governance_js.test_check_proposal_id_is_set_correctly(network, args)
```
achamayou, line 765: Why are we no longer running that test?

tests/governance_js.py
```diff
@@ -1205,3 +1207,145 @@ def make_script(table_name, kind):
...
+def test_final_proposal_visibility(network, args):
+    primary, _ = network.find_primary()
+    consortium = network.consortium
+
+    with temporary_constitution(
+        network,
+        args,
+        # A dumb and bad proposal that counts who voted for it. Just proving that such a thing is _possible_
+        make_action_snippet(
+            "cronyism",
```
achamayou, line 1222: I would prefer if we used more neutral wording here, such as "vote_provenance". We have long had constitution samples that counted votes differently based on provenance (operator), this information is public, this is explicitly supported.

achamayou (approved):

include/ccf/service/tables/gov.h
eddyashton: I'd taken these from the now-deleted type, but really just wanted the `struct` level docstring to remember which of these types is used in which situations. Have expanded the field descriptions on this type, since it's part of the documented governance KV schema, and removed the redundant descriptions on the `ProposalInfoSummary` fields.

src/node/gov/handlers/proposals.h
achamayou: It looks like they are, the comment just needs updating.

eddyashton: Yes - they are, and the comment is now updated.

src/node/gov/handlers/proposals.h
eddyashton: I'd initially hoped to remove the temporaries entirely, and write directly the `proposal_info`. I found some problem with that and restored them, but I think we're still slightly more explicit about their purpose now - they're not fields that are going to construct a `ProposalInfoSummary`, they're values that aren't _quite_ ready to re-populate `proposal_info`.

tests/governance.py
eddyashton: Bad merge, now restored.

tests/governance_js.py
```diff
@@ -1205,3 +1207,145 @@ def make_script(table_name, kind):
...
+
+        for _, raw_proposal in public_tables[table_name].items():
+            if raw_proposal is None:
+                # This is a deletion
+                continue
+
+            proposal = json.loads(raw_proposal)
+
+            state = proposal["state"]
+            if state in ("Open", "Withdrawn", "Dropped"):
```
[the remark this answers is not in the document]
eddyashton: I've added this, but in doing so discovered that we actually _don't_ have an instance of `Failed`, because if we throw in `apply()` we deliberately don't write the result to the KV:

```
            if (proposal_info->state == ProposalState::FAILED)
            {
              // If the proposal failed to apply, we want to discard the tx and
              // not apply its side-effects to the KV state, because it may have
              // failed mid-execution (eg - thrown an exception), in which case
              // we do not want to apply partial writes
              detail::set_gov_error(
                ctx.rpc_ctx,
                HTTP_STATUS_INTERNAL_SERVER_ERROR,
                ccf::errors::InternalError,
                fmt::format("{}", proposal_info->failure));
              return;
            }
```

In other words, we can't extract "writing a Failed proposal" from any partial writes that `apply()` may have created before it threw. Maybe we should add a shim layer so that we _can_ make this distinction, maybe we should remove the `Failed` state now. Unclear to me which is preferable, but fixing is out-of-scope for this PR.

PR #6115, opened 2024-04-09
Python infra: Set `gov_api_version` from launch

eddyashton:
Previously had weird early period where members who were created too early used classic API, regardless of intended settings.


achamayou (approved):

achamayou:
@eddyashton I suggest poking the daily canary on all changes that affect consortium.py, to run the long version of the LTS and make sure it doesn't cause problems there. FWIW, this particular change looks fine to me, which is precisely why I would do it :)

PR #6118, opened 2024-04-10
Add new `/gov/service/users` endpoints

eddyashton:
These weren't added in the initial version of the new API spec, but should have been for consistency (the user certs table is managed by the framework and present on every CCF service).

This adds 2 endpoints:
```
GET /gov/service/users
GET /gov/service/users/{userId}
```


achamayou (approved):

achamayou:
@eddyashton let's record somewhere the fact that the typespec needs updating.

eddyashton:
> @eddyashton let's record somewhere the fact that the typespec needs updating.

https://github.com/microsoft/CCF/issues/5691#issuecomment-2047695826

PR #6133, opened 2024-04-18
[DRAFT] What if we always emitted optional fields?

eddyashton:


achamayou:
This is an interesting experiment, but this PR is six months old and is unlikely to result in a merge.

PR #6134, opened 2024-04-18
Tlai experiment

achamayou:
The usual re-creation as a branch because of TLAi permissions, see #6132


achamayou:
image

I give up.

lemmy:
Apparently an intermittent issue:

image

lemmy:
@pelikhan Is there any non-determinism in relevant parts of GenAIScript that could explain this?

pelikhan:
This happens when we tried to read the LLM configuration information from the environment variables, like OPENAI_API_KEY, ... so there should not be anything non-deterministic.

I will add more logging in genaiscript to list the environment variables it found; but could it be that those are missing in the run? We could have a quick prescript that makes sure the OPENAI_API_KEY is set.

lemmy:
It could have been an intermittent Github action issue, but I think it's less likely.

pelikhan:
I will push a new cli with added logging.

achamayou:
@lemmy yes, it's not on all the runs, and I can't see any obvious pattern to it. The PR would not have worked anyway because it looks like TLAi only looks at diffs, not whole files.

PR #6137, opened 2024-04-18
[TLC Repro] Stack overflow likely caused by large state variable

achamayou:
Steps:

```
cd tla
JSON=traces/consistency/trace.ndjson ./tlc.sh consistency/TraceMultiNodeReads.tla  > out
```

out then contains:

```
Error: This was a Java StackOverflowError. It was probably the result
of an incorrect recursive function definition that caused TLC to enter
an infinite loop when trying to compute the function or its application
to an element in its putative domain.
Error: The behavior up to this point is:
```

Warning: out gets large (30Mb+)

### DIFF
--- a/tla/consistency/TraceMultiNodeReads.tla
+++ b/tla/consistency/TraceMultiNodeReads.tla
@@ -57,12 +57,15 @@ IsRwTxExecuteAction ==
     /\ IsEvent("RwTxExecuteAction")
     /\ RwTxExecuteAction
     /\ Last(history').tx = logline.tx
+    /\ Len(ledgerBranches') = logline.view
 
 IsRwTxResponseAction ==
     /\ IsEvent("RwTxResponseAction")
     /\ RwTxResponseAction
     /\ Last(history').type = ToTxType[logline.type]
     /\ Last(history').tx = logline.tx
+    /\ Last(history').status = ToStatus[logline.status]
+    /\ Last(history').tx_id = logline.tx_id
 
 IsStatusCommittedResponseAction ==
     /\ IsEvent("StatusCommittedResponseAction")
@@ -88,6 +91,23 @@ IsStatusInvalidResponseAction ==
     /\ Last(history').type = ToTxType[logline.type]
     /\ Last(history').status = ToStatus[logline.status]
 
+IsNotEvent ==
+    l' = l
+
+InsertTruncateLedgerAction ==
+    /\ IsNotEvent
+    /\ "view" \in DOMAIN logline
+    /\ logline.view > Len(ledgerBranches)
+    /\ "tx_id" \in DOMAIN logline
+    /\ logline.tx_id[1] > Len(ledgerBranches)
+    /\ TruncateLedgerAction
+
+InsertOtherTxnAction ==
+    /\ IsNotEvent
+    /\ "tx_id" \in DOMAIN logline
+    /\ logline.tx_id[2] > Len(Last(ledgerBranches))
+    /\ AppendOtherTxnAction
+
 TraceNext ==
     \/ IsRwTxRequestAction
     \/ IsRwTxExecuteAction
@@ -96,6 +116,8 @@ TraceNext ==
     \/ IsRoTxRequestAction
     \/ IsRoTxResponseAction
     \/ IsStatusInvalidResponseAction
+    \/ InsertTruncateLedgerAction
+    \/ InsertOtherTxnAction
 
 TraceSpec ==
     TraceInit /\ [][TraceNext]_<<l, vars>>
--- a/tla/consistency/TraceMultiNodeReads.tla
+++ b/tla/consistency/TraceMultiNodeReads.tla
@@ -57,14 +57,16 @@ IsRwTxExecuteAction ==
     /\ IsEvent("RwTxExecuteAction")
     /\ RwTxExecuteAction
     /\ Last(history').tx = logline.tx
-    /\ Len(ledgerBranches') = logline.view
+    \* RwTxExecuteAction can only take place if a branch exists for the view
+    /\ Len(ledgerBranches) >= logline.tx_id[1]
+    \* and it contains just the right amount of transactions (seqno - 1)
+    /\ Len(ledgerBranches[logline.tx_id[1]]) = logline.tx_id[2] - 1
 
 IsRwTxResponseAction ==
     /\ IsEvent("RwTxResponseAction")
     /\ RwTxResponseAction
     /\ Last(history').type = ToTxType[logline.type]
     /\ Last(history').tx = logline.tx
-    /\ Last(history').status = ToStatus[logline.status]
     /\ Last(history').tx_id = logline.tx_id
 
 IsStatusCommittedResponseAction ==
@@ -104,8 +106,6 @@ InsertTruncateLedgerAction ==
 
 InsertOtherTxnAction ==
     /\ IsNotEvent
-    /\ "tx_id" \in DOMAIN logline
-    /\ logline.tx_id[2] > Len(Last(ledgerBranches))
     /\ AppendOtherTxnAction
 
 TraceNext ==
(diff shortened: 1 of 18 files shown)


lemmy:
`InsertOtherTxnAction` allows for behaviors with infinitely many consecutive `AppendOtherTxnAction` which leave the variable `l` forever unchanged.  Can you come up with a (tight) bound of the number of successive `AppendOtherTxnAction` that can be expressed in, e.g., a state constraint?

tla/consistency/TraceMultiNodeReads.tla
```diff
@@ -81,13 +86,29 @@ IsRoTxResponseAction ==
...
+    /\ Last(history').tx_id = logline.tx_id
 
 IsStatusInvalidResponseAction ==
     /\ IsEvent("StatusInvalidResponseAction")
     /\ StatusInvalidResponseAction
     /\ Last(history').type = ToTxType[logline.type]
     /\ Last(history').status = ToStatus[logline.status]
 
+IsNotEvent ==
+    l' = l
```
lemmy, line 98: Allowing `l` to stutter will make TLC accept any prefix of the log because `Len(JsonLog)` repetition of `InsertOtherTxnAction ` actions will satisfy `TLCGet("stats").diameter >= Len(JsonLog)`.

tla/consistency/TraceMultiNodeReads.tla
```diff
@@ -81,13 +86,29 @@ IsRoTxResponseAction ==
...
+    /\ IsNotEvent
+    /\ "view" \in DOMAIN logline
+    /\ logline.view > Len(ledgerBranches)
+    /\ "tx_id" \in DOMAIN logline
+    /\ logline.tx_id[1] > Len(ledgerBranches)
+    /\ TruncateLedgerAction
+
+InsertOtherTxnAction ==
+    /\ IsNotEvent
+    /\ AppendOtherTxnAction
```
lemmy, line 110: `InsertOtherTxnAction` should be enabled iff `l \in 1..Len(JsonLog)`.

tla/consistency/TraceMultiNodeReads.tla
```diff
@@ -96,6 +117,8 @@ TraceNext ==
     \/ IsRoTxRequestAction
     \/ IsRoTxResponseAction
     \/ IsStatusInvalidResponseAction
+    \* \/ InsertTruncateLedgerAction
+    \/ InsertOtherTxnAction
```
lemmy, line 121: The primary challenge appears to be the explicit representation of transactions from other clients.

Could we consider modifying `ledgerBranches` from being modeled as a sequence of sequences to a sequence of functions, with each function's domain being the second elements of `tx_id`? Specifically, rather than using a sequence, which is a function with a domain of consecutive natural `\in 1..Len(seq)`,  we switch to a domain consisting of non-consecutive natural numbers, specifically the observed transaction IDs.  Clearly, this is a bit of a refactoring.  On the other hand, it seems to mostly boil down to replacing `Len(ledgerBranches[view])` with `DOMAIN ledgerBranches[view]` and `Append(@, ...` with ` @ @@ (Max(DOMAIN ledgerBranches[view])+1) :> ...`.

```diff
diff --git a/tla/consistency/SingleNode.tla b/tla/consistency/SingleNode.tla
index 4c4e6fd6a..a0b51ea94 100644
--- a/tla/consistency/SingleNode.tla
+++ b/tla/consistency/SingleNode.tla
@@ -6,6 +6,8 @@
 
 EXTENDS ExternalHistoryInvars, TLC
 
+MaxOr(S) ==
+    Max(S \cup {1})
 
 \* Abstract ledgers that contains only client transactions (no signatures)
 \* Indexed by view, each ledger is the ledger associated with leader of that view 
@@ -63,11 +65,11 @@ RwTxExecuteAction ==
         \* that it can be picked up by the current leader or any former leader
         /\ \E view \in FirstBranch..Len(ledgerBranches):
                 ledgerBranches' = [ledgerBranches EXCEPT ![view] = 
-                    Append(@,[view |-> view, tx |-> history[i].tx])]
+                    @ @@ (MaxOr(DOMAIN ledgerBranches[view])+1) :> [view |-> view, tx |-> history[i].tx]]
         /\ UNCHANGED history
 
 LedgerBranchTxOnly(branch) ==
-    LET SubBranch == SelectSeq(branch, LAMBDA e : "tx" \in DOMAIN e)
+    LET SubBranch == [ i \in {j \in DOMAIN branch : "tx" \in DOMAIN branch[j] } |-> branch[i] ] 
     IN [i \in DOMAIN SubBranch |-> SubBranch[i].tx]
 
 \* Response to a read-write transaction
@@ -87,7 +89,7 @@ RwTxResponseAction ==
                     history,[
                         type |-> RwTxResponse, 
                         tx |-> history[i].tx, 
-                        observed |-> LedgerBranchTxOnly(SubSeq(ledgerBranches[view],1,seqnum)),
+                        observed |-> LedgerBranchTxOnly([j \in DOMAIN ledgerBranches[view] \cap 1..seqnum |-> ledgerBranches[view][j]]),
                         tx_id |-> <<ledgerBranches[view][seqnum].view, seqnum>>] )
     /\ UNCHANGED ledgerBranches
 
@@ -96,7 +98,7 @@ RwTxResponseAction ==
 StatusCommittedResponseAction ==
     /\ \E i \in DOMAIN history :
         /\ history[i].type = RwTxResponse
-        /\ Len(ledgerBranches[Len(ledgerBranches)]) >= history[i].tx_id[2]
+        /\ Max(DOMAIN ledgerBranches[Len(ledgerBranches)]) >= history[i].tx_id[2]
         /\ ledgerBranches[Len(ledgerBranches)][history[i].tx_id[2]].view = history[i].tx_id[1]
         \* Reply
         /\ history' = Append(
@@ -110,8 +112,7 @@ StatusCommittedResponseAction ==
 \* Append a transaction to the ledger which does not impact the state we are considering
 AppendOtherTxnAction ==
     /\ \E view \in FirstBranch..Len(ledgerBranches):
-        ledgerBranches' = [ledgerBranches EXCEPT ![view] = 
-                    Append(@,[view |-> view])]
+        ledgerBranches' = [ledgerBranches EXCEPT ![view] = @ @@ (MaxOr(DOMAIN ledgerBranches[view])+1) :> [view |-> view] ]
     /\ UNCHANGED history
 
 \* A CCF service with a single node will never have a view change
```

Furthermore, if we anticipate that a client may also encounter non-consecutive views, it might be better to model `ledgerBranches` as a function of function, the views as the domain of the outer function.

lemmy: https://github.com/microsoft/CCF/pull/6137/commits/76fd5525e5742c67c7a999ab7e7d3588798d5207

achamayou:
> `InsertOtherTxnAction` allows for behaviors with infinitely many consecutive `AppendOtherTxnAction` which leave the variable `l` forever unchanged. Can you come up with a (tight) bound of the number of successive `AppendOtherTxnAction` that can be expressed in, e.g., a state constraint?

Yes, as described yesterday, the boundary is provided by IsRwTxExecuteAction eventually matching once a sufficient ledgerBranch prefix has been produced. This works in #6136.

Where that breaks down is when a rollback takes place, creating a second appendable ledgerBranch, for which the trace may not contain any further IsRwTxExecuteAction.

tla/consistency/TraceMultiNodeReads.tla
achamayou: That should be covered by the Termination constraint.

tla/consistency/TraceMultiNodeReads.tla
achamayou: @lemmy agreed, I think that would be closer to implementation behaviour and more efficient. But:

1. The current spec has the benefit of being very simple.
2. I would like to understand (and ideally confirm) if it is possible to validate a spec with no/minimal changes. It feels like if it's not possible with this spec, it probably is never possible in practice.

I will iterate on the validation today, and see where that leads.

PR #6142, opened 2024-04-22
Enable both Invalid and Commit on the future of the head branch

achamayou:
Following discussion with @eddyashton, attempt to modify the consistency spec, in a minimal way, to allow for constructive Invalid (i.e. allow both the StatusCommittedResponseAction and the StatusInvalidResponseAction to make decisions about the future commit history).

This should be a no-op from an MC perspective (to be confirmed by the full CI), but would facilitate trace validation by allowing Invalid to be matched even when the known commit history doesn't extend far enough.

Quite tentative.

### DIFF
--- a/tla/consistency/SingleNode.tla
+++ b/tla/consistency/SingleNode.tla
@@ -95,16 +95,24 @@ RwTxResponseAction ==
 \* Note that a request could only be committed if it's in the highest view's ledger
 StatusCommittedResponseAction ==
     /\ \E i \in DOMAIN history :
-        /\ history[i].type = RwTxResponse
-        /\ Len(ledgerBranches[Len(ledgerBranches)]) >= history[i].tx_id[2]
-        /\ ledgerBranches[Len(ledgerBranches)][history[i].tx_id[2]].view = history[i].tx_id[1]
-        \* Reply
-        /\ history' = Append(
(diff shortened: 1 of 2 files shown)


tla/consistency/SingleNode.tla
```diff
@@ -95,16 +95,24 @@ RwTxResponseAction ==
...
-        /\ ledgerBranches[Len(ledgerBranches)][history[i].tx_id[2]].view = history[i].tx_id[1]
-        \* Reply
-        /\ history' = Append(
-            history,[
+        LET view == history[i].tx_id[1]
+            seqno == history[i].tx_id[2]
+        IN /\ history[i].type = RwTxResponse
+           /\ Len(Last(ledgerBranches)) >= seqno
+           /\ Last(ledgerBranches)[seqno].view = view
+           \* There is no future InvalidStatus that's incompatible with this commit
```
lemmy, line 103: I don't understand this comment because it says "future" but quantifies over the (unprimed) *history*.

achamayou: That's future in the ledger sense, i.e. a tx on the same view but with a higher (later) seqno. Perhaps using the word later is preferable here.

heidihoward (approved):

PR #6143, opened 2024-04-22
[Draft] Consistency Trace Validation: match transaction ids, take 2

achamayou:
Replaces #6136, using #6142


tla/consistency/MultiNode.tla
```diff
@@ -13,7 +13,7 @@ ViewWithAllCommitted ==
 \* Simulates leader election by rolling back some number of uncommitted transactions and updating view
 TruncateLedgerAction ==
     /\ \E view \in ViewWithAllCommitted:
-        /\ \E i \in (CommitSeqNum + 1)..Len(ledgerBranches[view]) :
```
achamayou, line 16: This is what's causing the refinement failure, because it disallows a transition that effectively creates an empty term.

heidihoward: yes, that was just a mistake.

heidihoward (approved):

PR #6152, opened 2024-04-26
A trivial TLA+ loc counter

achamayou:


eddyashton (approved):

tla/loc.py
```diff
@@ -0,0 +1,46 @@
...
+            self.blank + other.blank
+        )
+
+def count_loc(lines) -> LOC:
+    loc = LOC()
+    for line in lines:
+        line_no_ws = line.strip()
+        if not line_no_ws:
+            loc.blank += 1
+        elif line_no_ws.startswith("\*"):
```
eddyashton, line 33: I know this is quick-and-dirty, but listing some obvious improvements to accuracy:

- Lines starting `(*` are comments
- Lines starting `----` are comments
- Lines starting `====` are comments
- Everything _after_ any `====`-starting line is a comment

(I think TLA might also say anything pre-module is a comment? I'm not sure, but I don't think we do that. We definitely do the latter, eg in `SIMccfraft.tla`)

achamayou: Thank you, will update. Some of the lines starting with `----` are arguably code, the module name declaration at the top for example, is interpreted by TLC. Not sure about pre-module lines, I will check.

PR #6163, opened 2024-05-02
Cleanup `js::core::Context`, move functionality to `ExtensionInterface`s

eddyashton:
Goal is for `js::core::Context` to be a pure interpreter (ie - we don't insert anything that QuickJS doesn't), and a vaguely readable interface. This PR works towards the latter by shuffling methods around, removing a few unused/unnecessary overloads, and the former by introducing an `Extension` pattern rather than `populate_global_XXX` methods on the `Context`.

A goal of this extension pattern is the removal of the messy `globals` object on `Context` - previously lots of inserted calls relied on the `globals.tx` object inserted to support the `ccf.kv` functions. Now they explicitly use the state in their own extension, including a `Tx*` where necessary. This should also simplify cleanup, as its a case of deleting extensions rather than invalidating global values.


eddyashton:
TODO before merging:

- [ ] Rename `IExtension`
- [ ] Rename the extensions for less `Ccf` repetition
- [ ] Manually verify what we expose in each execution context hasn't changed, see if that can be automated

src/js/extensions/ccf/node.cpp
```diff
@@ -69,21 +78,7 @@ namespace ccf::js
...
-      }
-
-      auto tx_ptr = jsctx.globals.tx;
-
-      if (tx_ptr == nullptr)
-      {
-        return JS_ThrowInternalError(
-          ctx, "No transaction available to open service");
-      }
+      LOAD_FROM_EXTENSION(jsctx);
```
achamayou, line 81: A bit unfortunate to have a macro that creates variables in the current scope, I think three assignments with a JS_CHECK using the right message would have been slightly longer but a little less magic.

achamayou (approved):

src/js/extensions/ccf/node.cpp
eddyashton: Agreed - I'll do the copy-paste. I thought I was going to have to repeat this pattern for every extension, and looking to reduce boiler-plate, but actually it's only needed by a few of these functions so fits with the existing boilerplate

achamayou: I would mildly prefer that, but this is already a big PR, so also fine to keep this way.

PR #6164, opened 2024-05-03
`InvalidNotObservedInv` vacuously true, `InvalidNotObservedInv` and `CommittedRwSerializableInv` slow to evaluated by TLC due to combinatorial explosion.

lemmy:
See commit messages for proof of equivalence that passed TLAPS scrutiny. *Please do not squash to preserve the commit messages*.  TLC now checks all models within a couple of seconds.

PR #6166, opened 2024-05-03
Use typed model values to prevent comparing things that shouldn't be compared.

lemmy:
Low-risk change.

Squash any type-related discussions, and prevent future incantations of https://github.com/microsoft/CCF/pull/6164/commits/20d89bff161410663d14e0cd2ad0788fe6337409. ;-p

See section about *typed model values* in https://tla.msr-inria.inria.fr/tlatoolbox/doc/model/model-values.html


achamayou:
I am sure this will make @eddyashton smile :)

Not sure about "types can only be one letter", but still much better than before!

achamayou (approved):

Issue #6157, opened 2024-04-29
Only log lower and upper bounds of `committable_indices` to correct claim that we only log variables values constant in space.


lemmy:
@achamayou correctly observes that we claim that trace validation only logs variable values that are constant in space.  However, `committable_indices`, is not constant in space.  While it is unlikely for `committable_indices` to grow prohibitively large outside of pathological behaviors, `state.h` should only log `committable_indices`' lower and upper bound.

https://github.com/microsoft/CCF/blob/6f50f186345cabb1b36086c195fc64e937911cc9/src/consensus/aft/impl/state.h#L202

Due to the way how `Traceccfraft` has been defined, this change should be isolated to `state.h`, i.e., the implementation, unless the json format changes substantially

https://github.com/microsoft/CCF/blob/6f50f186345cabb1b36086c195fc64e937911cc9/tla/consensus/Traceccfraft.tla#L192

PR #6172, opened 2024-05-08
Only log start end of committable indices in trace

achamayou:
#6157


tla/consensus/Traceccfraft.tla
```diff
@@ -522,6 +522,9 @@ TraceMatchesConstraints ==
     /\ MembershipStateConsistentInv
     /\ CommitCommittableIndices
 
+CommittableIndicesIsConstantSpaceInv ==
+    "committable_indices" \in DOMAIN logline.msg.state => Len(logline.msg.state.committable_indices) <= 2
```
lemmy, line 526: Does this need the implication? `committable_indices` doesn't seem to be optional in the json logging in the diff above.

achamayou: It's not on "drop_pending_to" at the moment. It looks like there is another problem though.

PR #6174, opened 2024-05-09
Raft tidy up

### LINKED ISSUE #5894: Opportunities to tidy up `raft.h`
Whilst comparing `raft.h` to the TLA+, I spotted a few possible opportunities to tidy up `raft.h`. None of these changes should alter the behavior of `raft.h` (though I might be wrong!).

1) `new_view_idx` appears to be unused
2) unused if statement https://github.com/microsoft/CCF/blob/8f6d14c6f42264f30c5a65b5c745d6586317978b/src/consensus/aft/raft.h#L1109-L1118
3) Comment out-of-sync with code https://github.com/microsoft/CCF/blob/8f6d14c6f42264f30c5a65b5c745d6586317978b/src/consensus/aft/raft_types.h#L161-L168

achamayou:
Closes #5894

### DIFF
--- a/src/consensus/aft/raft.h
+++ b/src/consensus/aft/raft.h
@@ -1197,8 +1197,17 @@ namespace aft
       // Redundant with check on get_term_internal() at line 1149
       // Which captures this case in every situation, except r.prev_term == 0.
       // That only happens if r.prev_idx == 0 however, see line 1033,
-      // in which case this assertion is also true.
-      assert(r.prev_idx <= state->last_idx);
+      // in which case this path should not be taken either.
+      else if (r.prev_idx > state->last_idx)
+      {
+        RAFT_FAIL_FMT(
+          "Recv append entries to {} from {} but prev_idx ({}) > last_idx ({})",
+          state->node_id,
+          from,
+          r.prev_idx,
+          state->last_idx);
+        return;
+      }
 
       RAFT_DEBUG_FMT(
         "Recv append entries to {} from {} for index {} and previous index {}",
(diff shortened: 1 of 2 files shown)


src/consensus/aft/raft.h
```diff
@@ -1195,16 +1194,6 @@ namespace aft
...
-      else if (r.prev_idx > state->last_idx)
-      {
-        RAFT_DEBUG_FMT(
-          "Recv append entries to {} from {} but prev_idx ({}) > last_idx ({})",
-          state->node_id,
-          from,
-          r.prev_idx,
-          state->last_idx);
-        return;
-      }
```
maxtropets, line 1207: Nit. We could replace it with some sort of assertion (or tracing) if we truly believe this cannot happen.
Cons
- one (possibly) redundant check
Pros
- we'll know if this happens

Feel free to close without changing.

heidihoward:
lgtm

maxtropets (approved):

src/consensus/aft/raft.h
```diff
@@ -1195,16 +1194,11 @@ namespace aft
...
+      // Redundant with check on get_term_internal() at line 1149
+      // Which captures this case in every situation, except r.prev_term == 0.
+      // That only happens if r.prev_idx == 0 however, see line 1033,
+      // in which case this assertion is also true.
+      assert(r.prev_idx <= state->last_idx);
```
achamayou, line 1201: This is redundant and never taken, but in a slightly subtle way. The trade-off between removing it and keeping it to make the code more robust to potential changes is less clear to me than it was at first. I think my preference goes towards keeping the branch with the comment added. @eddyashton what do you think?

eddyashton: I was going to say we have few-to-no `assert`s in the current code (non-test), but actually `raft.h` already has a few. Still, I think this that removing this branch just makes us less robust. So I'd restore the branch, and upgrade the log within it from `DEBUG` to `FAIL`, with an associated comment (this _should_ never happen, but for action-at-a-distance reasons).

Issue #5809, opened 2023-11-02
JwtAuthnPolicy could validate the key_issuer against iss


achamayou:
`JwtAuthnPolicy` currently exposes an authenticated `key_issuer`, and an un-checked `iss` field from the token. It is up to the application code to check that they match (which should be the case when the standard is followed), and to include logic used to implement [multi-tenant support for example](https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens#validate-the-issuer), when applicable.

We could improve on this by:

1. Checking `key_issuer` == `iss`
2. If not, instantiate the `key_issuer` using the AAD-specific logic described in the link above and the `tid`

This would break neither standards-compliant, nor AAD use cases, and would reduce the risk of incorrect token validation by the application code.
```[tasklist]
### Tasks
- [x] Do the issuer validation against the token
- [x] Migration: check how the old code works with new nodes in the network and prepare a PR to 4.x if needed
- [x] Migration: make sure we get rid of the old table with time
```

eddyashton:
@maxtropets

maxtropets:
**Key points**

- We improve token authorization by adding checks against `issuer` data for both key and token.
- In the scope of this ticket we assume we can’t have more than one unique `kid`. The documentation [doesn’t mention](https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens) what should we do if doesn’t come true. It is up to https://github.com/microsoft/CCF/issues/5177 to either:
  - prove the point, or
  - prove that wrong, but not an issue, or
  - prove that wrong and to be an issue
    - come up with the fix in this case
- **Fixups we consider**
  - Check `token.iss` and `key.issuer` match (caveats for multi-tenancy mode, see below)
  - `token.iss`’s tenant-id and `token.tid` claim match
  - key.issuer matches the expected issuer when fetching
    - To be discussed further
    - Doc [says](https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens#recap)
      - Identify key in the metadata based on the kid header. **Check the "issuer" property attached to the key in the metadata document**
```
      // ...
      if (configuration.allowedIssuer != "*"
          && configuration.allowedIssuer != issuer)
          throw validationException;
      // ...
```              
- The algorithm to be implemented is nicely summarized [in the documentation](https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens#recap)
- Multi-tenancy VS single-tenancy
[...]

Issue #5177, opened 2023-04-14
Duplicate JWK kids across issuers are not handled correctly


achamayou:
CCF's KV schema and internal logic assumes JWK kids uses for JWT authentication are unique globally, not per issuer. This is an incorrect assumption, as per https://www.rfc-editor.org/rfc/rfc7517#page-8 
```[tasklist]
### Tasks
- [x] member_frontend → KeyIdInfo
- [x] adding and removing issuers: change logic to support vectors
- [x] service_state → set multiple issuers
```

maxtropets:
Some preparation job is going to be done in #5809

maxtropets:
Closed in https://github.com/microsoft/CCF/pull/6175

PR #6175, opened 2024-05-10
JWT issuer validation

maxtropets:
Eventually closes

- https://github.com/microsoft/CCF/issues/5809
- https://github.com/microsoft/CCF/issues/5177
- https://github.com/microsoft/CCF/issues/6204

Decided to merge together to avoid being split into different releases


src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -3,13 +3,94 @@
...
+    // https://domain.com/tenant_id/something_else
+    //
+    // Here url.path == "/tenant_id/something_else".
+
+    const auto url = http::parse_url_full(token.payload_typed.iss);
+    const auto tenant_id = first_non_empty_chunk(nonstd::split(url.path, "/"));
+
+    return (
+      tenant_id && token.payload_typed.tid &&
+      *token.payload_typed.tid == tenant_id.value());
```
maxtropets, line 74: Inconsistent use of operator* and value()

include/ccf/service/tables/jwt.h
```diff
@@ -58,18 +58,28 @@ namespace ccf
...
+    JwtIssuer issuer;
+    std::optional<JwtIssuer> constraint;
+  };
+  DECLARE_JSON_TYPE_WITH_OPTIONAL_FIELDS(JwtIssuerWithConstraint);
+  DECLARE_JSON_REQUIRED_FIELDS(JwtIssuerWithConstraint, issuer);
+  DECLARE_JSON_OPTIONAL_FIELDS(JwtIssuerWithConstraint, constraint);
+
   using JwtIssuers = ServiceMap<JwtIssuer, JwtIssuerMetadata>;
   using JwtPublicSigningKeys = kv::RawCopySerialisedMap<JwtKeyId, Cert>;
-  using JwtPublicSigningKeyIssuer =
```
maxtropets, line 63: Spoke with @achamayou that it's not safe for upgrade. We'll proceed with keeping the old table in this version, and will get rid of it later in subsequent 5.0.x release.

maxtropets: Change of plans. We'll do

* Keep old tables to fallback to if we have no keys in the new tables to keep serving requests
* Don't do anything to the old code. This might cause stale keys which we are already forced to have due to 30-min refresh window, so shouldn't make things noticeably worse
* The latter is bad only if we remove the key on the new node and old nodes keep serving requests with the old keys, which isn't good, but almost unachievable in practise.

maxtropets:
As noted above, kept the old tables but moved them under "Legacy" namespace. We can work with them if needed but kept them explicitly obsolete.

maxtropets:
Converted back to "Draft" to include #5177 and #6204

src/endpoints/authentication/jwt_auth.cpp
maxtropets: Done

include/ccf/service/tables/jwt.h
maxtropets: Done

tests/jwt_test.py
```diff
@@ -49,28 +65,147 @@ def test_refresh_jwt_issuer(network, args):
...
+        json.dump(keys, jwks_fp)
+        jwks_fp.flush()
+        network.consortium.set_jwt_public_signing_keys(
+            primary, issuer.name, jwks_fp.name
+        )
+
+    service_keys = get_jwt_keys(args, primary)
+    assert service_keys["kid1"]["issuers"][issuer.name] == issuer.name
+
+    keys["keys"][0]["issuer"] = "https://issuer.com"
```
eddyashton, line 166: Can you add a comment or brief `LOG.info` explaining this line? It looks like it's setting a permitted constraint (in contrast to the later `garbage` constraints), but its not obvious to me why we're setting this nested field directly.

tests/jwt_test.py
```diff
@@ -49,28 +65,147 @@ def test_refresh_jwt_issuer(network, args):
...
+            jwks_fp.flush()
+            try:
+                network.consortium.set_jwt_public_signing_keys(
+                    primary, issuer.name, jwks_fp.name
+                )
+            except infra.proposal.ProposalNotAccepted:
+                pass
+            else:
+                assert False, f"Constraint {constraint} must not be allowed"
+
```
eddyashton, line 189:
```suggestion

    network.consortium.remove_jwt_issuer(primary, issuer.name)
```

The other tests remove their temporary issuer to clean up the network before closing - should this do the same?

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -356,6 +365,339 @@ def test_jwt_auth(network, args):
         )
         assert r.status_code == HTTPStatus.UNAUTHORIZED, r.status_code
 
+    network.consortium.remove_jwt_issuer(primary, issuer.name)
+    return network
+
+
+def try_auth(primary, issuer, kid, iss, tid):
+    with primary.client("user0") as c:
+        LOG.info("Calling JWT with kid from issuer for tenant")
```
eddyashton, line 374:
```suggestion
        LOG.info(f"Creating JWT with kid={kid} from issuer={iss} for tenant={tid}")
```

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -356,6 +365,339 @@ def test_jwt_auth(network, args):
...
+def test_jwt_auth_msft_multitenancy(network, args):
+    """For a common tenant, all tokens from this issuer can auth,
+    no matter which tenant is specified."""
+
+    primary, _ = network.find_nodes()
+
+    COMMNON_ISSUER = "https://login.microsoftonline.com/{tenantid}/v2.0"
+    TENANT_ID = "9188050d-6c67-4c5b-b112-36a304b66da"
+    ISSUER_TENANT = f"https://login.microsoftonline.com/{TENANT_ID}/v2.0"
+    ANOTHER_TENANT_ID = "ANOTHER-6c67-4c5b-b112-36a304b66da"
```
eddyashton, line 450:
```suggestion
    ANOTHER_TENANT_ID = "aaaaaaa-6c67-4c5b-b112-36a304b66da"
```

Nit: Although we treat these as strings for now, they're really GUIDs - we should use a string of the right shape so that it will still work if we add more validation later. Hopefully it's clear that this is still an invalid/constructed GUID.

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -356,6 +365,339 @@ def test_jwt_auth(network, args):
...
+                        "kid": jwt_kid,
+                        "x5c": [der_b64],
+                        "issuer": ISSUER_TENANT,
+                    }
+                ]
+            },
+        }
+        json.dump(data, metadata_fp)
+        metadata_fp.flush()
+        network.consortium.set_jwt_issuer(primary, metadata_fp.name)
```
eddyashton, line 558: We do this block several times - can we pull it out to a helper function?

eddyashton (commented):
Partial review - need to take another look at the C++ changes.

src/node/rpc/jwt_management.h
```diff
@@ -21,18 +22,62 @@
...
+    const auto constraint_domain = http::parse_url_full(constraint).host;
+
+    if (constraint_domain.empty())
+    {
+      return false;
+    }
+
+    // Either constraint's domain == issuer's domain or it is a subdomain, e.g.:
+    // limited.facebok.com
+    //        .facebok.com
```
eddyashton, line 38:
```suggestion
    // limited.facebook.com
    //        .facebook.com
```

src/node/rpc/jwt_management.h
```diff
@@ -21,18 +22,62 @@
...
+    if (issuer_domain != constraint_domain)
+    {
+      const auto start_seek =
+        issuer_domain.size() - (constraint_domain.size() + 1);
+      const auto count_seek = constraint_domain.size() + 1;
+      const auto pattern = "." + constraint_domain;
+
+      return start_seek > 0 // at least one letter preceeds .issuer.domain
+        && issuer_domain.substr(start_seek, count_seek) == pattern;
```
eddyashton, line 47: Could this be replaced with `ends_with()`?

```suggestion
      return issuer_domain.ends_with("." + constraint_domain);
```

src/node/rpc/jwt_management.h
```diff
@@ -21,18 +22,62 @@
...
+
+    return true;
+  }
+
   static void remove_jwt_public_signing_keys(kv::Tx& tx, std::string issuer)
   {
-    auto keys = tx.rw<JwtPublicSigningKeys>(Tables::JWT_PUBLIC_SIGNING_KEYS);
-    auto key_issuer =
-      tx.rw<JwtPublicSigningKeyIssuer>(Tables::JWT_PUBLIC_SIGNING_KEY_ISSUER);
+    auto keys =
```
eddyashton, line 55: I think we should keep the original code as well here, to remove entries from the `Legacy` tables.

src/node/rpc/jwt_management.h
```diff
@@ -192,38 +243,132 @@ namespace ccf
...
+      if (new_kid || new_constraint)
+      {
+        JwtIssuerWithConstraint value{issuer, std::nullopt};
+        const auto it = issuer_constraints.find(kid);
+        if (it != issuer_constraints.end())
+        {
+          value.constraint = it->second;
+        }
+
+        LOG_INFO_FMT(
```
eddyashton, line 316:
```suggestion
        LOG_DEBUG_FMT(
```

I think `INFO` is too verbose for this, and will lead to a lot of logging spam in the output.

src/node/rpc/jwt_management.h
```diff
@@ -192,38 +243,132 @@ namespace ccf
...
+            [&](const auto& issuer_with_constraint) {
+              return issuer_with_constraint.issuer == issuer;
+            }),
+          updated->end());
+
+        if (updated->empty())
+        {
+          keys->remove(kid);
+          key_issuers->remove(kid);
+        }
```
eddyashton, line 367: Non-critical, but maybe we should try to clean up the `Legacy` tables here as well? So don't bother trying to read from them to populate `existing_kids`, but if we're removing from the new tables then we could also try `legacy_keys->remove(kid)`?

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -96,10 +186,17 @@ namespace ccf
...
+            !validate_issuers(token, *issuers, std::ref(validated_issuer)))
+          {
+            error_reason =
+              fmt::format("Kid {} failed issuer validation", key_id);
+          }
           else
           {
             auto identity = std::make_unique<JwtAuthnIdentity>();
-            identity->key_issuer = key_issuers->get(key_id).value();
+            identity->key_issuer = validated_issuer;
```
eddyashton, line 199: We never use the `Legacy::JWT_PUBLIC_SIGNING_KEY_ISSUER` table, which means (if I'm reading right) that this `validated_issuer` string will still be empty, if `issuers` is empty and we went through the `fallback_keys` path? Can we add a path in the fallback case above, something like:

```
if (!issuers)
{
  auto fallback_issuer = tx.ro<...>(ccf::Tables::Legacy::JWT_PUBLIC_SIGNING_KEY_ISSUER);
  validated_issuer = fallback_issuer->get(key_id).value_or({});
}
```

Or maybe we set `issuers` to a single-entry vector containing the value from the legacy map, and go through this validation path? Whatever makes sense, but I think we should try harder to use the fallback values if we're doing any fallback handling here.

src/node/rpc/jwt_management.h
eddyashton: Cool, I understand and agree now - not worth peppering any `remove()` calls to legacy tables throughout the current code, because we don't want to try and precisely match the old write semantics. Instead we will deal with the old legacy state by adding a `clear()` call to the legacy tables, in a future release.

src/node/rpc/jwt_management.h
eddyashton: See https://github.com/microsoft/CCF/pull/6175/files/00972768b191b644e90e189700f8622480ad5c80#r1620429945

src/node/rpc/jwt_management.h
maxtropets: Indeed, I missed we got it in our current std version.

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -3,13 +3,94 @@
...
+        return chunk;
+      }
+    }
+    return std::nullopt;
+  }
+
+  bool validate_issuer(
+    const http::JwtVerifier::Token& token, std::string issuer)
+  {
+    LOG_INFO_FMT(
```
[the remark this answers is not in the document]
maxtropets: I actually tried to implement it as close as possible to the pseudocode from Entra doc.

https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens#recap

src/endpoints/authentication/jwt_auth.cpp
maxtropets: Good catch. Will update with a fallback value, to preserve the old behaviour.

src/node/rpc/jwt_management.h
maxtropets: Talked in person, decided to remove the keys from both revision when app calls "remove" explicitly, but don't touch old table on (re)setting new keys for an issuer.

tests/jwt_test.py
maxtropets: I'm not sure what explanation to add around this. We set it because the `keys` are retrieved above as a parsed JSON, so in the test we overwrite this field only (which is the issuer constraint) to make sure only certain constrains can be set.

Doesn't look sophisticated to me, but I don't find it ugly also.

tests/js-custom-authorization/custom_authorization.py
maxtropets: Done

src/endpoints/authentication/jwt_auth.cpp
maxtropets: Done

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -3,13 +3,94 @@
...
+    const http::JwtVerifier::Token& token, std::string issuer)
+  {
+    LOG_DEBUG_FMT(
+      "Verify token.iss {} and token.tid {} against published key issuer {}",
+      token.payload_typed.iss,
+      token.payload_typed.tid,
+      issuer);
+
+    const bool is_microsoft_entra =
+      issuer.find(microsoft_entra_domain) != std::string::npos;
```
achamayou, line 42: That does not seem strict enough, `http://foo.com/login.microsoftonline.com` is not microsoft entra for example.

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -3,13 +3,94 @@
...
+    {
+      if (!chunk.empty())
+      {
+        return chunk;
+      }
+    }
+    return std::nullopt;
+  }
+
+  bool validate_issuer(
```
achamayou, line 32: We should have unit tests for this function.

src/node/rpc/jwt_management.h
```diff
@@ -113,6 +173,16 @@ namespace ccf
         return false;
       }
 
+      if (keys->has(kid) && der != keys->get(kid))
```
achamayou, line 176: Agree with log, but not sure about early false here. If this happens, I think we want it flagged, but we don't necessarily want to stop updating keys.

maxtropets: That's what we did for other if-blocks around. E.g.:

```
      if (keys->has(kid) && key_issuer->get(kid).value() != issuer)
      {
        LOG_FAIL_FMT(
          "{}: key id {} already added for different issuer", log_prefix, kid);
        return false;
      }

```

I'm inclined to preserve that because the call is "set_keys", so setting them partially and returning `true` doesn't seem right.

Let me know your thoughts if you insist we should consider changing it.

achamayou: Agree we should not do a partial set in any circumstances. But I think we could replace the mapping altogether rather than back off.

maxtropets: By avoiding partial set for the current issuer being set, we do so for the previous one which has a conflicting keys, don't we?

In this particular case we do overwrite a key for some issuer, which may be only one of the keys that issuer has set.

Moreover, we now introduce new inconsistency, because if we do decide to simply overwrite - the issuer for previous kid is kept in issuers map.

src/node/rpc/jwt_management.h
maxtropets: Ticket filed
https://github.com/microsoft/CCF/issues/6222

src/node/rpc/jwt_management.h
achamayou: I meant the mapping for the issuer. There are two possible collisions here:

1. A kid collision for a single issuer. This amounts to an inconsistent update set from that issuer, and a violation of the OpenID spec. Rejecting the update wholesale is clearly the only coherent behaviour.
2. A kid collision between two issuer. Two coherent behaviours are possible: reject the update, or accept it entirely and remove the keys for the issuer that had colliding keys (i.e. latest wins, as opposed to first wins).

I am not sure that we need to make this trade-off though, because it seems to me that we do need to have a unique `kid -> pem`, and we could instead have a `kid -> [pem, iss]`, which we try to validate one by one until we get a match. In fact, I don't understand how we really solve #5177 with this check.

maxtropets: > In fact, I don't understand how we really solve https://github.com/microsoft/CCF/issues/5177 with this check.

1. Scenarios like this are not permitted anymore
```
- Iss A set [(kid_1, pem_1), (kid_2, pem_2)]
- Iss B set [(kid_1, pem_0), (kid_3, pem_3)]
- Remove Iss B
```
Resulting kids -> pem map
kid_1 -> pem_0
kid_2 -> pem_2

2 (my opinion): Explicit > implicit, so a silent key overwriting has been replaced by conflicting key rejection

maxtropets: >  we could instead have a kid -> [pem, iss], which we try to validate one by one until we get a match
I agree, and I'd rather implement this instead.

The reason we decided not to do it after discussing within a team was twofold
- key validation is expensive, doing several might affect the performance
- overcomplicates the code, with a rather tine chance to happen in practice

So the decision was to leave it as is and if implement only if someone really needs it.

Options we have
- implement it here
- create a ticket and implement on demand later on

I prefer the latter, but I'll take the one you suggest, because I have little experience of practical application of CCF.

include/ccf/service/tables/jwt.h
```diff
@@ -58,18 +58,37 @@ namespace ccf
   using JwtKeyId = std::string;
   using Cert = std::vector<uint8_t>;
 
+  struct KeyMetadata
```
achamayou, line 61: I suggest renaming to JWKMetadata to make it obvious which Key this is about, because there are a fair few kinds of keys in the KV.

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -3,15 +3,80 @@
...
+      "Verify token.iss {} and token.tid {} against published key issuer {}",
+      iss,
+      tid,
+      constraint);
+
+    const auto issuer_url = http::parse_url_full(constraint);
+    if (issuer_url.host != microsoft_entra_domain)
+    {
+      return iss == constraint && !tid; // tid is a MSFT-specific claim and
+                                        // shoudn't be set for a non-AAD issuer.
```
achamayou, line 51:
```suggestion
                                        // shoudn't be set for a non-Entra issuer.
```

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -3,15 +3,80 @@
...
+    }
+
+    // Step 1. Verify the token issuer against the key issuer.
+    if (iss != constraint)
+    {
+      return false;
+    }
+
+    // Step 2. Verify that token.tid is served as a part of token.iss. According
+    // to the documentation, we only accept this format:
```
achamayou, line 68: Excellent comment, can I suggest adding the link to said doc?

src/node/gov/handlers/service_state.h
```diff
@@ -466,31 +466,27 @@ namespace ccf::gov::endpoints
...
+                const ccf::JwtKeyId& k, const std::vector<KeyMetadata>& v) {
+                auto keys_info = nlohmann::json::array();
+                for (const auto& metadata : v)
+                {
+                  auto info = nlohmann::json::object();
 
-                // cert is stored as DER - convert to PEM for API
-                const auto cert_pem = crypto::cert_der_to_pem(cert);
-                key_info["certificate"] = cert_pem.str();
+                  // cert is stored as DER - convert to PEM for API
```
achamayou, line 479: Why is cert stored as DER?

achamayou (approved):

src/endpoints/authentication/jwt_auth.cpp
maxtropets: Done

src/node/gov/handlers/service_state.h
maxtropets: I don't know why the specific format was chosen, this comment was here before.

src/node/rpc/jwt_management.h
```diff
@@ -39,6 +42,66 @@ namespace ccf
       });
   }
 
+  static bool check_issuer(
+    const std::string& issuer, const std::string& constraint)
```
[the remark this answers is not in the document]
maxtropets: Done

maxtropets:
I've run jwt perf test on main VS branch

```
./tests.sh -VV -R pi_ls_jwt -L "perf"
python3 tests/infra/piccolo/throughput_analysis.py # patched with proper paths
```

Results **main**

```
(env) mtropets@mtropets:~/workspace/CCF$ python3 tests/infra/piccolo/throughput_analysis.py
+------+-------+------+------------+
| Reqs |  Time | Pass | Throughput |
+------+-------+------+------------+
| 1000 | 0.046 | 0.0  |  21914.4   |
```

Result branch

```
python3 tests/infra/piccolo/throughput_analysis.py 
+------+-------+------+------------+
| Reqs |  Time | Pass | Throughput |
+------+-------+------+------------+
| 1000 | 0.064 | 0.0  |  15716.4   |
+------+-------+------+------------+
```

Seems like a pretty decent performance hit, will dig up

maxtropets:
> I've run jwt perf test on main VS branch
> 
> ```
> ./tests.sh -VV -R pi_ls_jwt -L "perf"
> python3 tests/infra/piccolo/throughput_analysis.py # patched with proper paths
> ```
> 
> Results **main**
> 
> ```
> (env) mtropets@mtropets:~/workspace/CCF$ python3 tests/infra/piccolo/throughput_analysis.py
> +------+-------+------+------------+
> | Reqs |  Time | Pass | Throughput |
> +------+-------+------+------------+
> | 1000 | 0.046 | 0.0  |  21914.4   |
> ```
> 
> Result branch
> 
> ```
> python3 tests/infra/piccolo/throughput_analysis.py 
> +------+-------+------+------------+
> | Reqs |  Time | Pass | Throughput |
> +------+-------+------+------------+
> | 1000 | 0.064 | 0.0  |  15716.4   |
> +------+-------+------+------------+
> ```
> 
> Seems like a pretty decent performance hit, will dig up

However, I don't know how heavy is the typical TX. In the test, we do one append per TX, which might differ from the real-world scenario, and so, I have no clue about the real impact of 25% slower JWT authentication.

Meaning, if the auth itself is 10x faster than the request itself - it doesn't really matter, otherwise it might matter.

maxtropets:
One more perf update

Removed issuer validation, but left new schema in place. Results

```
(env) mtropets@mtropets:~/workspace/CCF$ python3 tests/infra/piccolo/throughput_analysis.py 
+------+-------+------+------------+
| Reqs |  Time | Pass | Throughput |
+------+-------+------+------------+
| 1000 | 0.058 | 0.0  |  17306.4   |
+------+-------+------+------------+
```

Most likely, the perf impact is `25%=10%+15%`
* 10% for issuer validation
* 15% for new schemadeserialization

achamayou:
@maxtropets you need to rev up the openapi version of the gov schema, run the test again locally, and check in the updated openapi golden file to get it to pass.

25% hit on end-to-end throughput is substantial, but will be offset to some extent by the application being able to remove some of that logic. It would be good to run the same test (full logging) with jwt removed, to compare (the closest we have is basic, but that's a smaller app, and clocks about ~57kTx/s on the CI). It would also be good to run with a profiler to see what exactly makes this worse.

But because this change is a correctness fix, I think we merge it now, and investigate separately if there is a way to improve the performance. Ideally we do this now, and figure out quickly if further schema changes are needed or not. Improvements that don't impact the schema can happen in later releases.

achamayou:
@maxtropets ah one thing we need on this PR is a proper CHANGELOG entry, explaining the change to application developers, and service operators, with details about actions they need to take.

achamayou (approved):

maxtropets:
> the closest we have is basic, but that's a smaller app

@achamayou aren't they actually the same test? Seems like the only difference is JWT auth.

achamayou:
> > the closest we have is basic, but that's a smaller app
> 
> @achamayou aren't they actually the same test? Seems like the only difference is JWT auth.

No, there are two subtle differences. One is that the logging API involves JSON wrappers, with extra (de)serialisation/copies. The other is that the logging js app is fairly large (it has lots of endpoints) and so involves loading quite a bit more bytecode at every execution.
So while it is logically doing the same thing, it's running quite a bit slower.

CHANGELOG.md
```diff
@@ -15,7 +15,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Removed
 
-- Removed unused `openenclave.verifyOpenEnclaveEvidence` API from JS/TS.
+- Removed unused `openenclave.verifyOpenEnclaveEvidence` API from JS/TS
+
+### Changed
+
+- Added token.iss claim validation to JWT authentication, as well as saving issuer constraint to the KV on hey-autorefresh (#5809).
```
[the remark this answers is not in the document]
maxtropets: Done. Please resolve if you're happy with the new version of it.

eddyashton (approved):

PR #6199, opened 2024-05-21
Move module loading out of `js::core::Context`, so it can be customised

eddyashton:
Goal is that other apps can use the `js::core::Context`, but read JS code from their own tables. Written to be extensible so they could also load some from hardcoded values, or modify/combine values in the KV, or...


achamayou (approved):

achamayou:
@eddyashton lgtm but deserves a nightly run I think.

eddyashton:
Note for posterity: A previous version of this PR introduced a major memory leak*, that we almost missed as we currently have no visibility into perf regressions. @achamayou stated this explicitly - we're currently making large changes to the JS interpreter but flying blind with regard to the performance impact. We only caught this one pre-merge by chance, because of a single (Daily build-gated) stress test which ballooned the memory so far it crashed a node. Restoring perf visibility is a high priority.

\* I moved the cache of previously-loaded modules out of the `Context`, and into one of its `Loader`s. This meant we weren't getting caching benefits across invocations, since we installed new `Loader`s each time, so a long-lived interpreter (re-fetched from the interpreter cache) was continually re-processing the same modules, and linearly growing its memory use.

PR #6201, opened 2024-05-21
Basic programmability sample

achamayou:
End-to-end demo of programmability for a C++ app.

Goals:

- A simple, but extensible way to enable programmability in existing UserEndpointRegistry subclasses
- A sample `PUT /custom_endpoints` endpoint implementing JS application installation
- Illustrative automated testcases showing how this can be used

```
  // By subclassing CustomJSEndpointRegistry, this application gains the ability
  // execute custom JavaScript endpoints, and exposes the ability to install
  // them via install_custom_endpoints().
  // This sample also adds a PUT /app/custom_endpoints that enables a user
  // for which user_data["isAdmin"] is true to install custom JavaScript
  // endpoints. The JavaScript code for these endpoints is stored in the
  // internal KV store under a namespace configured in the second argument to
  // the constructor. PUT /app/custom_endpoints is logically
  // equivalent to passing a set_js_app proposal in governance, except the
  // application resides in the application space.
  //
  // Known limitations:
  //
  // No auditability yet, COSE Sign1 auth is mandated, but the signature is not
  // stored. No support for historical endpoints yet.
  //
  // Additional functionality compared to set_js_app:
  //
  // The KV namespace can be private, to keep the application confidential if
  // desired.
  class BasicHandlers : public basicapp::CustomJSEndpointRegistry
  {
  public:
    BasicHandlers(ccfapp::AbstractNodeContext& context) :
      basicapp::CustomJSEndpointRegistry(
        context,
        "public:custom_endpoints" // Internal KV space will be under
                                  // public:custom_endpoints.*
      )
    {
     ...
      auto put_custom_endpoints = [this](ccf::endpoints::EndpointContext& ctx) {
        // Authorization
        ...
        install_custom_endpoints(ctx, wrapper);
        ctx.rpc_ctx->set_response_status(HTTP_STATUS_NO_CONTENT);
      };

      make_endpoint(
        "/custom_endpoints",
        HTTP_PUT,
        put_custom_endpoints,
        {ccf::user_cose_sign1_auth_policy})
        .set_auto_schema<ccf::js::BundleWrapper, void>()
        .install();
    }
  }
```


samples/apps/basic/custom_endpoints/registry.h
```diff
@@ -0,0 +1,599 @@
...
+#include "ccf/js/extensions/ccf/historical.h"
+#include "ccf/js/extensions/ccf/host.h"
+#include "ccf/js/extensions/ccf/kv.h"
+#include "ccf/js/extensions/ccf/request.h"
+#include "ccf/js/extensions/ccf/rpc.h"
+#include "ccf/js/extensions/console.h"
+#include "ccf/js/extensions/math/random.h"
+#include "ccf/js/modules.h"
+#include "ccf/service/tables/modules.h"
+#include "endpoint.h"
```
eddyashton, line 34:
```suggestion
#include "apps/basic/custom_endpoints/endpoint.h"
```

Think it's worth being explicit with include paths, as we have several `endpoint.h`s now.

samples/apps/basic/custom_endpoints/registry.h
```diff
@@ -0,0 +1,599 @@
...
+          {
+            std::string method_upper = method;
+            nonstd::to_upper(method_upper);
+            const auto key = ccf::endpoints::EndpointKey{url, method_upper};
+            endpoints->put(key, metadata);
+          }
+        }
+
+        auto modules =
+          ctx.tx.template rw<ccf::Modules>("public:custom_endpoints.modules");
```
eddyashton, line 152: Can we take the table names (the string `"public:custom_endpoints.modules"`) as a constructor arg, store as a const member? Think that gives us very clear customisability already, for minimal additional code complexity.

samples/apps/basic/custom_endpoints/registry.h
```diff
@@ -0,0 +1,599 @@
...
+#include "ccf/js/modules.h"
+#include "ccf/service/tables/modules.h"
+#include "endpoint.h"
+#include "js/interpreter_cache_interface.h"
+
+using namespace nlohmann;
+
+namespace basicapp
+{
+  class CustomJSEndpointRegistry : public ccf::UserEndpointRegistry
```
eddyashton, line 41: Can you add a docstring to this, outlining the broad goal and listing the known TBDs (vs the js_generic endpoints)? We should have a central place to say "doesn't handle templated endpoints, doesn't utilise the bytecode cache" for the first dev release.

samples/apps/basic/custom_endpoints/registry.h
achamayou: Yes, although I was thinking of doing that for a namespace instead of individual tables. I think the table naming remains an internal detail of the registry, maybe.

samples/apps/basic/custom_endpoints/endpoint.h
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+
+#include "ccf/app_interface.h"
+
+namespace basicapp
+{
+  struct CustomJSEndpoint : public ccf::endpoints::Endpoint
```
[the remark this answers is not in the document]
achamayou: 100% agreed, this just went under "enough changes in this one already".

samples/apps/basic/custom_endpoints/registry.h
achamayou: Yes, very good point.

samples/apps/basic/custom_endpoints/registry.h
achamayou: Actually that does not work, because we -I include and src, but not samples. I'll make it a comment. I think this eventually disappears anyway once we've set the public API.

eddyashton (approved):

PR #6205, opened 2024-05-24
Don't log JS exception details by default

eddyashton:
Restoring intended initial values for these 2 bools, which were accidentally updated to verbose-by-default in a recent refactor.

### DIFF
--- a/src/js/core/runtime.h
+++ b/src/js/core/runtime.h
@@ -28,8 +28,8 @@ namespace ccf::js::core
     static constexpr size_t default_stack_size = 1024 * 1024;
     static constexpr size_t default_heap_size = 100 * 1024 * 1024;
 
-    bool log_exception_details = true;
-    bool return_exception_details = true;
+    bool log_exception_details = false;
+    bool return_exception_details = false;
 
     Runtime();
     ~Runtime();


achamayou:
@eddyashton that seems like a thing we really ought to have a test for.

achamayou (approved):

eddyashton:
> @eddyashton that seems like a thing we really ought to have a test for.

I 100% agree, but while trying to add that I've decided it deserves a bigger cleanup. The entire way we fetch and report the JS config is messy, with various paths taken to determine defaults when the `JSEngine` value doesn't exist. I think we should squash that entirely - the default constructor of a `JSEngine` contains the default values, you get the default if there's none in the KV, and then you _always_ have _a_ `JSEngine` object to work with. I'll do that, but not in this PR.

achamayou:
@eddyashton I agree with that too, but I think in this case it's just a canary that runs an endpoint that produces an error and checks nothing gets logged unless this is on, it's quite orthogonal, and a good thing to have around as we refactor the JS runtime substantially.

achamayou:
@eddyashton So we have (almost!) exactly that test: https://github.com/microsoft/CCF/blob/bdac0c49c226c6a7bcf4c4fec4be0a8fb4d58853/tests/js-modules/modules.py#L1339

But! Because it doesn't run first, and `set_js_runtime_options` is used in a number of other tests before, a change that affects the default will not be caught.

I will move it up.

Issue #6126, opened 2024-04-16
Find a way to restore or replace cimetrics


eddyashton:
We disabled `cimetrics` in #6125, to avoid use of `METRICS_MONGO_CONNECTION`.

We should look at restoring `cimetrics`, either with improved auth to mongo, or by finding an alternative storage mechanism.

achamayou:
I looks like bencher.dev may be [what we want](https://bencher.dev/console/projects/ccf/perf?branches=47a2cef2-2400-4286-ae79-fb17c98b2c23&testbeds=1e6f6a27-eb58-4f16-8d01-0148fbaed70e&benchmarks=7baac671-8a7c-4968-a7dc-aea6fbbec759%2Ced8e5432-5ebc-49dd-89b9-9839fed1d945%2C68e15d7c-3859-49de-8d94-92cb3331d075%2Cea38e096-f614-4907-b33d-749020d68afa%2C09034135-5114-49ea-b270-03a98fbc0bb7&measures=bc9fb376-9a85-478a-97fd-ebd7703c9663&start_time=1712758465000&end_time=1715350465000&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&report=cc1a7aa5-d97b-43aa-bc3a-a8237a8e0868&clear=true).

Test job: https://github.com/microsoft/CCF/blob/bencher_experiment/.github/workflows/bencher.yml, works on ubuntu 20.04 using a 1ES pool.

The json adapter supports both latency and throughput, and the format is very simple: https://bencher.dev/docs/explanation/adapters/#-json

I think we would move the performance tests to GHA for simplicity, and we would start with just running on main. The setup for PRs on branches seems to need a bit of care to avoid exposing secrets, but we may be able to do without it and just check periodically.

achamayou:
[...]

PR #6207, opened 2024-05-24
Post numbers from a few benchmarks to bencher

achamayou:
Possible solution to #6126, seems very easy to use.


tests/infra/bencher.py
```diff
@@ -0,0 +1,20 @@
...
+
+BENCHER_FILE = "bencher.json"
+
+class Bencher:
+    def __init__(self):
+        if not os.path.isfile(BENCHER_FILE):
+            with open(BENCHER_FILE, "w+") as bf:
+                json.dump({}, bf)
+    
+    def set(self, key, value):
```
eddyashton, line 15: This looks great! Can we add a type for `value` to help build something in the format bencher expects (or at least a link to the intended schema near here)?

eddyashton:
Results from this new job are posted here: https://bencher.dev/perf/ccf

Merging this PR as an initial step.

PR #6208, opened 2024-05-28
Fix Daily Build: New header location

eddyashton:
#6201 moved this header to public. The code in `endpoints/grpc/status.h` is only compiled as part of the `external_executors` app, which is now gated behind the `LONG_TESTS` CMake option, so only causes a failure of the Daily Build (but passes CI).


eddyashton:
Force merging, to get a green light on `main` ASAP.

PR #6210, opened 2024-05-28
Extend Bencher testing, add Memory metric

eddyashton:
~~(Draft while I experiment with the results)~~

Adding a few more tests to Bencher, removing `^` suffix, adding a Memory metric (docs slightly unclear, but suggests you can define these on-the-fly?), experimenting with some other Bencher options.

### DIFF
--- a/.github/workflows/bencher.yml
+++ b/.github/workflows/bencher.yml
@@ -40,5 +40,4 @@ jobs:
           --testbed gha-virtual-ccf-sub \
           --adapter json \
           --err \
-          --file build/bencher.json \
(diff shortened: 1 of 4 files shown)


eddyashton:
Ok this current setup obviously doesn't run from forks, for good security reasons. Going to force merge this PR to see what Bencher does with the new entries, and open a separate PR refactoring the yaml (based on Bencher's [examples](https://github.com/bencherdev/example/blob/main/.github/workflows/run_fork_pr_benchmarks.yml) and [docs](https://bencher.dev/docs/how-to/github-actions/#benchmark-fork-pr-and-upload-from-default-branch)) so that we can get results from fork PRs too.

PR #6214, opened 2024-05-29
Update CI image from 2024-04-25 to 2024-05-29

eddyashton:


eddyashton:
Failure in SNP CI I believe is unrelated (`e2e_redirects` failure because there was an election, and a previous-backup became primary and stopped redirecting?), so force-merging this.

PR #6217, opened 2024-05-30
[Draft] Testing Bencher PR comments

eddyashton:
I've fiddled with the yaml config on `main`, so I _think_ it should now post comments on PRs like this. Let's see.


eddyashton:
Useful PoC, but backing out for now. After discussion with @achamayou, we'll aim purely for `main` results rather than maintaining the complex yaml for branch/PR testing.

PR #6219, opened 2024-05-31
Add SGX perf job in GitHub Actions

eddyashton:
Trying out an SGX GHA pool, with the Bencher flow, so we can get some visibility of SGX (and memory) stats.


eddyashton:
Closing this, and will add memory stats to virtual instead.

PR #6220, opened 2024-06-03
Restore simpler Bencher setup - run on `main` only

eddyashton:
Following discussion with @achamayou, we've opted to track perf regressions purely on `main` for now rather than maintaining the workflow complexity for branch detection.


achamayou (approved):

achamayou:
@eddyashton I think we want to prioritise getting complete coverage of regression metrics on virtual, on main. That's what we are going to target medium to long term, and that's what might be affected by the large refactoring to come.

Although it would be nice to get more comprehensive coverage across platforms, and to have PR-level detection, this is less critical, and other tasks are higher priority.

PR #6226, opened 2024-06-04
Add remaining performance tests to bencher

achamayou:
Ticking off some items on #6126 by moving remaining tests, memory metrics and microbenchmarks to bencher.

Also get rid of the ^ business.

### DIFF
--- a/tests/infra/basicperf.py
+++ b/tests/infra/basicperf.py
@@ -389,14 +389,14 @@ def run(args):
                         current_value = results["current_allocated_heap_size"]
                         peak_value = results["peak_allocated_heap_size"]
 
-                        # Do not upload empty metrics (virtual doesn't report memory use)
-                        if current_value != 0 and peak_value != 0:
-                            metrics.append(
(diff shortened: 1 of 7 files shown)


tests/infra/basicperf.py
```diff
@@ -389,14 +389,14 @@ def run(args):
...
-                        if current_value != 0 and peak_value != 0:
-                            metrics.append(
-                                infra.bencher.Memory(
-                                    value=current_value,
-                                    high_value=peak_value,
-                                )
-                            )
+                        bf = infra.bencher.Bencher()
+                        bf.set(
+                            f"{args.label}_mem",
```
eddyashton, line 394: I don't think we want to use a separate benchmark (the string name here) for every measure (memory)? The point of having multiple measures (Throughput, Latency, Memory...) is that we can report them under the _same_ benchmark name, when they're from the same run.

achamayou: Oh I see, will fix.

eddyashton (approved):

PR #6228, opened 2024-06-05
Avoid prefixing new endpoint modules

eddyashton:
This is pulled out of an upcoming PR, as I think it's clearer standalone.

We have a weird behaviour where we prefix all of the module names before putting them in the KV. I don't think this is necessary! I'm trying to avoid doing it for the new tables (`DynamicJSRegistry`), but we need to maintain it for the old tables (`js_generic`) for compatibility


achamayou:
As discussed, happy with this change, so long as we can apply it to both paths (i.e. make the code permissive at load time), to avoid building up differences and let us rebase js_generic eventually.

eddyashton:
I'm abandoning this PR. I was aiming to avoid any change to the old paths (such that an existing JS app upgrading to 5.0.0 continued to work without resubmission), and test that via `lts_compatibility`. That turns out to be an expensive bit of plumbing through `lts_compatibility`, which prefers to always resubmit the app. Considered other options, but ultimately not worth the effort.

We'll live with the prefixing in both old and new paths, and work out what that means for getter APIs later.

PR #6233, opened 2024-06-06
Get intended reuse from `CommonContext`

eddyashton:
I added this `CommonContext` so we wouldn't have to list the same basic extensions everywhere, but then didn't use it for the app interpreters (`js_generic` and `registry.cpp`). This rejigs things a little so its useable in those situations.

### DIFF
--- a/src/apps/js_generic/js_generic_base.cpp
+++ b/src/apps/js_generic/js_generic_base.cpp
@@ -5,20 +5,15 @@
 #include "ccf/crypto/rsa_key_pair.h"
 #include "ccf/endpoints/authentication/all_of_auth.h"
 #include "ccf/historical_queries_adapter.h"
+#include "ccf/js/common_context.h"
 #include "ccf/js/core/context.h"
 #include "ccf/js/core/wrapped_property_enum.h"
 #include "ccf/js/extensions/ccf/consensus.h"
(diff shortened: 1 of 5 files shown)


src/apps/js_generic/js_generic_base.cpp
```diff
@@ -527,16 +511,11 @@ namespace ccfapp
...
-      extensions.emplace_back(
-        std::make_shared<ccf::js::extensions::OpenEnclaveExtension>());
-      // add snp_attestation.*
-      extensions.emplace_back(
-        std::make_shared<ccf::js::extensions::SnpAttestationExtension>());
 
       interpreter_cache->set_interpreter_factory(
         [extensions](ccf::js::TxAccess access) {
-          auto interpreter = std::make_shared<js::core::Context>(access);
+          // CommonContext also adds many extensions
```
achamayou, line 517: This is true, but not very informative perhaps.

achamayou (approved):

src/apps/js_generic/js_generic_base.cpp
eddyashton: There's comments within `CommonContext` listing _what_ is added, that I don't want to duplicate. This is intended as a pointer down the behaviour chain - if you have extensions that aren't listed _here_, they're likely _there_.

PR #6234, opened 2024-06-06
Add getters and versioned API to `DynamicJSEndpointRegistry`

eddyashton:
I think we want to version the methods here, as we do for `get_view_for_seqno_v1` etc. Additionally, I think we need some getter methods, so that apps can implement endpoints to read the installed JS app. This adds both.


tests/programmability.py
```diff
@@ -124,9 +173,9 @@ def run(args):
 
         network = test_custom_endpoints(network, args)
 
-        network = npm_tests.build_npm_app(network, args)
-        network = deploy_npm_app_custom(network, args)
-        network = npm_tests.test_npm_app(network, args)
+        # network = npm_tests.build_npm_app(network, args)
```
achamayou, line 176: @eddyashton needs commenting back in?

include/ccf/js/registry.h
```diff
@@ -76,8 +76,20 @@ namespace ccf::js
      * Call this to populate the KV with JS endpoint definitions, so they can
      * later be dispatched to.
      */
-    void install_custom_endpoints(
-      ccf::endpoints::EndpointContext& ctx, const ccf::js::Bundle& bundle);
+    ccf::ApiResult install_custom_endpoints_v1(
```
achamayou, line 79: Ideally all need doxygen plus pointers in Sphinx.

achamayou (approved):

include/ccf/js/registry.h
eddyashton: Doxygen'd now, I think we'll Sphinx later in a single pass.

PR #6237, opened 2024-06-06
Set role definition sample

achamayou:
A small sample added to the basic programmability test showing:

1. Adding an action to the constitution to define maps of roles to actions (`set_role_definition`)
2. Using governance to create role definitions (in this sample, `user_data` is used to create user to roles mapping)
3. Adding basic authorization logic to the application, using the role definition maps.

Because the samples are inline JS here and do not make use of TypeScript, they unfortunately do not leverage #6230 at the moment.


samples/constitutions/roles/set_role_definition.js
```diff
@@ -0,0 +1,60 @@
...
+  new Action(
+    function (args) {
+      checkType(args.role, "string", "role");
+      checkType(args.actions, "array", "actions");
+      for (const [i, action] of args.actions.entries()) {
+        checkType(action, "string", `actions[${i}]`);
+      }
+    },
+    function (args) {
+      let roleDefinition = new KVSet(
```
sidmore, line 43: qq: why is this wrapper needed? is KVSet not already defined in CCF?

achamayou: There is a `kv::Set` internally in C++, but there is a fairly substantial amount of glue between `kv::Map` in C++ and `ccf.kv` in JS, especially post-interpreter re-use (to deal with potential dangling references in context etc).

Even if we essentially duplicated this glue to support both types, there would still be the problem of how to expose it next to `ccf.kv[]`. Open to ideas, but my intention at this point is to look at moving the constitution to TypeScript, because that allows using `typedKvSet` (see #6320), which is quite easy to use but does not require extending a large amount of C++ infra types.

sidmore (approved):

tests/programmability.py
```diff
@@ -86,6 +116,130 @@ def test_custom_endpoints(network, args):
...
+    with primary.client(None, None, user.local_id) as c:
+        r = c.put("/app/custom_endpoints", body=bundle_with_auth)
+        assert r.status_code == http.HTTPStatus.NO_CONTENT.value, r.status_code
+
+    # Add role definition
+    prop = member.propose(
+        primary,
+        proposal(
+            action(
+                "set_role_definition", role="ContentGetter", actions=["/content/read"]
```
eddyashton, line 158: Minor, since this is just a sample, but:

We should really delimit action segments with something other than slashes. This very much looks like a permission for a `/content/read` URI, which doesn't exist. I think I'd be content with a more readable `/content:read` or `read:/content`, and then we could extend our sample to cover multi-element URIs (ie - `read:/content/foo/bar`).

eddyashton (approved):

tests/programmability.py
achamayou: @eddyashton I do agree, in fact I'm even suspicious of making this a single string at all, but that's the general convention used in Azure RBAC, as used in TPAL (e.g. `Microsoft.ConfidentialLedger/ledger/governance/constitution/read`). The goal here is to show how this can be done for Azure services, and so staying with that convention is the right thing to do in context.

PR #6240, opened 2024-06-07
Add ability to further restrict which tables are accessed by JS code

eddyashton:
Opening as a draft because while I revisit a few details:

- Unsure where we should stash these restrictions. `KvExtension` seems natural, but then how does it affect historical KV access, which is controlled by a different extension? Maybe we need to duplicate across both of these, that's not horrendous.
- The error message when a namespace restriction fails isn't very helpful, because we're reusing some of the machinery from the hard-coded restrictions (eg "400 Failed to write to handle for table: records
61: TypeError: Cannot call "set" on inaccessible table named records in application execution context. See https://microsoft.github.io/CCF/main/audit/read_write_restrictions.html for more detail."). I think we should have a different error message here, ideally supplied by the app (_why_ is this table inaccessible).


samples/apps/basic/basic.cpp
```diff
@@ -117,6 +117,36 @@ namespace basicapp
...
+      // read-only for JS code. Additionally, we reserve any table beginning
+      // with "basic." (public or private) as inaccessible for the JS code, in
+      // case we want to use it for the C++ app in future.
+      set_js_kv_namespace_restriction(
+        [](const std::string& map_name, std::string& explanation)
+          -> ccf::js::KVAccessPermissions {
+          if (map_name == PRIVATE_RECORDS)
+          {
+            explanation = fmt::format(
+              "The {} table is managed by C++ endpoints, so is read-only in "
```
achamayou, line 131: I know we still have mentions of tables in places, but we've mostly converted to Maps, at least in the API and in types. I suggest new code and new messages should do the same.

eddyashton: Ah, I was tending towards "table" as a more generic parent term (to include Sets and Values). So almost every Table _is_ a Map, but not _every_ Table. And the namespacing we're talking about here is across all Tables.

src/js/permissions_checks.h
```diff
@@ -99,4 +94,47 @@ namespace ccf::js
...
+        break;
+      }
+      case (TxAccess::GOV_RW):
+      {
+        exec_context = "read-write governance";
+        break;
+      }
+    }
+
+    static constexpr char const* access_permissions_explanation_url =
```
achamayou, line 129: On the one hand, this is a nice thing to have and I am in favour. On the other hand, done as it is, we will have to be careful to remember that if we move/rename this in the future.

eddyashton: True. Fwiw, this isn't new, just moved here from `kvhelpers.h`. In the tests for this, we could extract the URL and fetch it to confirm its good, but then we get a flaky test whenever `github.io` is flaky.

As a step towards keeping them in-sync, I've added a comment to the `.rst`, so we're less likely to move/remove that without updating this too.

samples/apps/basic/basic.cpp
achamayou: Ok, I can see that, but we would need to update quite a bit of doc to introduce that notion, which is not currently mentioned anywhere as far as I can tell. The KV API, KV Howto, audit section etc all talk about Maps, and Sets and Values are only very briefly mentioned. They're also not particularly well supported in TypeScript or illustrated in the samples. And the actual implementation in the KV is completely Map-named, not Table-named. Tables is mostly just a namespace for what the documentation calls "Built-in Maps" as far as I can tell.

I am also not sure that "table" feels like a generic concept such that "a set is a table" makes immediate sense. Something like "collection" seems like a better abstraction for something that can be a map, a set or a value.

Sticking to Maps until we make a conscious decision otherwise seems better to me than a slow implicit drift.

eddyashton: I'm not proposing _introducing_ that notion here - I think it's _already_ peppered throughout the codebase and docs (322 results in 78 files, often naming KV variables or in KV documentation, and not _all_ authored by me).

I'm comfortable discounting the naming used by the KV implementation (because it's an internal detail, has been extended with several retrofitted concepts anyway), but I will take the extremely old `ccf::Tables` namespace as evidence this isn't a new idea (when every schema in the KV was hardcoded, data was in named Tables).

So to be clear, this is all fuzzy, inconsistent, potentially confusing, and worst of all **_naming_**, but I don't think its _new_.

achamayou: > I'm not proposing introducing that notion here - I think it's already peppered throughout the codebase and docs (322 results in 78 files, often naming KV variables or in KV documentation, and not all authored by me).

Unless I am missing something, we have pre-existing usage of "table" as a synonym for "map", but no mention of "table" as a shared abstraction, at least not explicitly. This is because maps used to be called tables, a very long time ago.

We had switched to Map in the API, at least by 0.2 (https://github.com/microsoft/CCF/blob/9471bc825f95e01a419cb2832382e39a95f8b0ea/src/kv/test/kv_test.cpp#L94), and possibly at some point before. I had a quick look through my emails but I've not been able to find when or how we did this.
I don't believe there has been a public API call with "table" in it so far, the service map headers are an exception and were not public until recently. I do agree that the document does use the word table interchangeably with map in places though.

src/js/permissions_checks.h
achamayou: > but then we get a flaky test whenever github.io is flaky

Yes, and a versioning headache too I suspect. I think the check would need to be based on extracting the relative path (strip root of doc prefix) and looking up in git instead.

Anyway, a comment in the .rst is fine I think.

achamayou (approved):

Issue #6223, opened 2024-06-04
Complete apoplication-level custom endpoints support (DynamicJSEndpointRegistry)


achamayou:
Work started in ...

Remain to be done:

- [ ] Table access restriction for the custom endpoints (#4096)
- [ ] Historical query support
- [ ] Expand test suite for endpoint install, deletion and replacement
- [ ] Implement endpoint replay protection
- [ ] Demonstrate customisable authentication

achamayou:
Expanding on "Implement in-sample endpoint replay protection", this is a consequence of implementing auditability through user-signed payloads. They create two risks:

1. Confusion: the signature is not sufficient to authenticate the action, checking the intent to perform that specific action is also necessary. To that end, governance envelopes include a `ccf.gov.msg.type` protected header. `user_cose_sign1_auth_policy` has no such mechanism however. The [programmability sample app](https://github.com/microsoft/CCF/blob/main/samples/apps/programmability/programmability.cpp) is not obviously susceptible to this problem, because the payloads are sufficiently different between the app and runtime options endpoints. But for future extensibility, for example if we want to apply auditability to other endpoints such as customer user role management, it may be good to have a protected header binding a signed payload to an intended usage.
2. Replay: once a payload is signed, even for an intent, it can be sent to the service multiple time. This is particularly true when the audit table is public. There are three main ways to mitigate this:
  a. Have the client request and send a nonce. This requires an extra call from the client.
  b. Avoid idempotent operations (a staple of REST APIs) and use specific pre-conditions. For example in the case of `PUT /custom_endpoints`, include the digest of the current code identity, much like an `If-Match` header in HTTP, but authenticated inside the signed payload of course. This is endpoint specific, and constrains the API in a way that is generally incompatible with Azure (and web) standards.
  c. Use a scheme similar to what governance does with `ccf.gov.msg.created_at`: have the client pass a timestamp, and keep a window of recent digest -> timestamp. If the timestamp is older than the median, reject the command (it is stale), otherwise check its presence in the map. This avoids duplicate submission with a constant-size memory footprint, at the expense of rejecting stale entries.

For the same reasons that we picked c. for governance, I think we should implement it here too.

PR #6241, opened 2024-06-10
Add setters and getter for JS runtime options, and sample HTTP APIs

eddyashton:
Part of #6223.

Slightly different semantics than the governance equivalent (`set_js_runtime_options`), which I find slightly cumbersome. Instead the (sample) endpoint accepts a `PATCH` request, with `null` values permitted to reset any field to its default. To support that, we change the serialisation of this `JSRuntimeOptions` object to be more permissive on deserialisation (where previously it had a sharp distinction between the original fields and those added later), and verbose on serialisation.

### DIFF
--- a/doc/schemas/gov_openapi.json
+++ b/doc/schemas/gov_openapi.json
@@ -349,11 +349,6 @@
             "$ref": "#/components/schemas/boolean"
           }
         },
-        "required": [
-          "max_heap_bytes",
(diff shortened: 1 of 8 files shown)


achamayou (approved):

eddyashton:
Failure is unrelated (in `partitions`, which is known to be flaky) - ignoring an force-merging.

PR #6244, opened 2024-06-10
Host data can be extracted on platforms other than SNP

### LINKED ISSUE #6242: CCF node running on SNP cannot join a CCF network running on SGX
**Describe the bug**
A CCF node running on SNP platform cannot join a network of CCF nodes running on SGX platform. This prevents a CCF network running on the SGX platform to be successfully code upgraded to the SNP platform.

**To Reproduce**

1. Spin up CCF network running on SGX
2. Add valid SNP host data through member's proposal
3. Add new CCF node on SNP (running the exact same host data) to the main CCF network: the node would not be able to join the network and would fail with the following error message: `An error occurred while joining the network: 401 UNAUTHORIZED  '{\"error\":{\"code\":\"InvalidQuote\",\"message\":\"Quote does not contain trusted host data\"}}'`.

**Expected behavior**

A new SNP node running an allowed-to-join host data (previously added to the ledger through a valid member's proposal) should be able to join a SGX network.

**Environment information**

Tested on CCF 5.0.0-dev17 (the latest dev version)

**Additional context**

I looked at the CCF code and I suspect the issue is [here](https://github.com/microsoft/CCF/blob/e7464e42d9efe0c7ad67735b4285e3c1a6110837/src/node/quote.cpp#L131-L132): when [verifying the host data against the store](https://github.com/microsoft/CCF/blob/e7464e42d9efe0c7ad67735b4285e3c1a6110837/src/node/quote.cpp#L157-L158), we get the host data and return the above error if the value is empty. Since the host data is in the ledger, I suspect we return an empty value [here](https://github.com/microsoft/CCF/blob/e7464e42d9efe0c7ad67735b4285e3c1a6110837/src/node/quote.cpp#L133) because the node is not running on SNP. I think that the SGX nodes, which need to allow the new SNP node to join the network, will never be able to verify the host data because we wrongly assume they are on SNP by default (but this is not the case during a cross-platform code upgrade).

achamayou:
Fixes #6242


eddyashton (approved):
Change looks good. Do we have a plan to test this? Spinning up nodes across platforms in the CI sounds tricky...

(Aside: I hadn't really considered it until this PR, but some of the stuff in our PAL probably shouldn't be there. If we're explicit about the platform, such as asking for SNP details here, in a way that's available on every platform, that could live elsewhere. the PAL should ideally only contain interfaces which _don't_ talk about the platform they're running on, but need to branch to something platform-specific in their implementation)

achamayou:
@eddyashton I don't think there is an easy way to test this automatically at the moment, I think it will have to be manual for the time being.

On VM-based TEEs, I think a pattern we could probably use is to start virtual nodes, and run mixed networks, but this is quite heavy to do on ACI at the moment (double build time, copy time etc).

About the pal/header layout, I do agree. The re-use of the mrEnclave field is also unfortunate. I think we can do a round of cleanup post 5.0.

PR #6245, opened 2024-06-11
Log `cchost` CLI arguments

eddyashton:
A while ago we moved some security-critical configuration options from the (potentially untrusted) JSON blob to the (attested on SNP) CLI arguments. This ensures they can't be tampered with, but also means they're no long present in the node's logging output. Specifically, we no longer explicitly log which enclave we're attempting to load, unless something goes wrong.

I think this is an accidental omission, and we should log the entire CLI. It's likely that deployment tooling has its own record of this, but for situations where only the node output is available it's critical to have this all in one place.


achamayou (approved):

src/host/main.cpp
```diff
@@ -187,6 +187,8 @@ int main(int argc, char** argv)
 
   LOG_INFO_FMT("CCF version: {}", ccf::ccf_version);
 
+  LOG_INFO_FMT("CLI args: {}", fmt::join(argv, argv + argc, " "));
```
achamayou, line 190: I don't know if there is a way to do that easily with fmt::something, but it would be nice to quote the args to avoid ambiguity if they contain spaces

This would log (I believe) `./a b c`, which can have been `./a "b c"` or `./a "b" "c"`.

achamayou, line 190:
```suggestion
  LOG_INFO_FMT("CLI args: \"{}\"", fmt::join(argv, argv + argc, "\" \""));
```

eddyashton: lol, nice, sure

PR #6246, opened 2024-06-11
Remove unbuffer

maxtropets:
Makes debugging via python console debugger useless.

@achamayou mentioned it was added to see CI logs faster/better/i-dont-know

Let's revise if we still need it and remove otherwise.


CMakeLists.txt
```diff
@@ -59,7 +59,6 @@ function(message)
 endfunction()
 
 option(PROFILE_TESTS "Profile tests" OFF)
```
achamayou, line 61:
```suggestion
option(PROFILE_TESTS "Profile tests" OFF)
set(PYTHON python3)
```

achamayou (approved):

PR #6249, opened 2024-06-11
Support `historical` mode endpoints in `DynamicJSEndpointRegistry`

eddyashton:
~~(Opening as a draft because I've pulled some unrelated changes into other PRs. Would like to merge those first and rebase)~~

Closes an outstanding item from #6223.

Check if the endpoints are marked as `historical` in the same we do in `js_generic`. Test this (and many other JS APIs!) by running the full `e2e_logging` test suite on the `basic` app, with the JS logging app deployed via `PUT /custom_endpoints`.

### DIFF
--- a/samples/apps/basic/basic.cpp
+++ b/samples/apps/basic/basic.cpp
@@ -19,7 +19,7 @@ using namespace nlohmann;
 namespace basicapp
 {
   using RecordsMap = kv::Map<std::string, std::vector<uint8_t>>;
-  static constexpr auto PRIVATE_RECORDS = "records";
+  static constexpr auto PRIVATE_RECORDS = "basic.records";
 
   // This sample shows the features of DynamicJSEndpointRegistry. This sample
   // adds a PUT /app/custom_endpoints, which calls install_custom_endpoints(),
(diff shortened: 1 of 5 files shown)


achamayou (approved):

samples/apps/basic/basic.cpp
```diff
@@ -19,7 +19,7 @@ using namespace nlohmann;
 namespace basicapp
 {
   using RecordsMap = kv::Map<std::string, std::vector<uint8_t>>;
-  static constexpr auto PRIVATE_RECORDS = "records";
+  static constexpr auto PRIVATE_RECORDS = "basic.records";
```
eddyashton, line 22: "Why?"

Because I want to install the JS implementation of the logging app on this (the C++ basic app). But both of them try to use a map called `records`! You can add a JS endpoint called `POST /log/private` which tries to write to `records`, and install it via basic's `PUT /custom_endpoints`, but then when you _call_ `POST /log/private` the new map restrictions (#6240) say "hang on, this JS code can't access my C++ controlled table!". Cheap fix is to change the map name that basic wants to reserve, to not collide with the logging app.

PR #6258, opened 2024-06-12
Audit in programmability sample

achamayou:
For the following endpoints:

- `PUT /app/custom_endpoints`
- `PATCH /app/custom_endpoints/runtime_option`

Input payloads (i.e. request bodies) are logged to `public:custom_endpoints.audit.input`, and metadata is logged at `public:custom_endpoints.audit.info`, with the following fields:

- `format` one of `COSE`, `JSON`.
- `content` one of `BUNDLE`, `OPTIONS`.
- `user_id` the user id of the caller.

Note that only `COSE` payloads are offline verifiable, i.e. can be checked by an outside auditor _without_ having to trust the service, since they are signed directly by the user key. For payloads that come in via client certificate authentication, the content and user id are recorded by the service, which is trusted to do this accurately.


samples/apps/basic/basic.cpp
```diff
@@ -322,6 +331,12 @@ namespace basicapp
           }
           // End of Authorization Check
 
+          // Make operation auditable by writing user-supplied COSE sign1
+          // document to the ledger
+          auto audit = ctx.tx.template rw<AuditValue>(
+            fmt::format("{}.cose_audit", CUSTOM_ENDPOINTS_NAMESPACE));
+          audit->put(ctx.rpc_ctx->get_request_body());
```
eddyashton, line 338: Does this body include a claim of what the operation is (`custom_endpoints` vs `runtime_options`)? I don't see any COSE protected header here, but think that we should make it explicit. With other auth schemes, where an app may only write a UID/OID for the submitter, I think it should still make clear which operation is recorded.

achamayou: I am doing separately in a replay protection PR, but yes that is needed, and that will mean new authn methods.

eddyashton (approved):

samples/apps/programmability/programmability.cpp
```diff
@@ -223,10 +228,19 @@ namespace programmabilityapp
         }
         // End of Authorization Check
 
-        const auto bundle = get_body(ctx);
+        const auto [format, bundle] = get_body(ctx);
```
sidmore, line 231: what is format and bundle here?

achamayou: See definition of `get_body()` on line 52, format is an enum (AuditInputFormat) and bundle is a std::span<const uint8_t> (i.e. the actual body, which could be either JSON or COSE, as indicated by format).

Issue #6082, opened 2024-03-21
Follow-ups for redirection support


eddyashton:
#6050 provides a working implementation of forwarding, but there's a few things that I've delayed for future PRs:

- Removing the custom `is_create` flag. I think the special `/create` endpoint can be given appropriate flags so that it is always executed locally, and doesn't need a custom flag.
- Revisit [fallback to local execution](https://github.com/microsoft/CCF/pull/6050/files#r1533669609). This was added to the redirection path for parity with the forwarding path. I believe hope it can be removed, but we may currently depend on it during unit tests or early-lifecycle execution.
- Clarify interaction of `set_forwarding_required` and `set_redirection_strategy`. We currently derive the latter from the former, need to make sure we do that safely, as we move towards deprecation of the former.
- Improve documentation. The auto-generated host schema docs don't give much detail on the new fields, need to expand what they show.
- Add `to_backup` forwarding option. No explicit demand for this, but should be easy to add for balance.

PR #6275, opened 2024-06-14
Add `to_backup` redirection strategy

eddyashton:
Final part of #6082, adding a symmetrical `to_backup` strategy, feasibly useful for load balancing of expensive read operations.


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [5.0.0-rc1]
+
+[5.0.0-rc1]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.0-rc1
+
+### Added
+
+- Endpoints now support a `ToBackup` redirection strategy, for requests which should never be executed on a primary. These should also be read-only. These are configured similar to `ToPrimary` endpoints, with a `to_backup` object (specifying by-role or statically-addressed targets) in each node's configuration.
```
```suggestion
- Endpoints now support a `ToBackup` redirection strategy, for requests which should never be executed on a primary. These must also be read-only. These are configured similar to `ToPrimary` endpoints, with a `to_backup` object (specifying by-role or statically-addressed targets) in each node's configuration.
```

include/ccf/endpoint.h
```diff
@@ -101,7 +101,12 @@ namespace ccf::endpoints
        a primary, it should attempt to redirect to the primary, or else return
        an error. Any write operations must be executed on a primary, so should
        be marked as ToPrimary. */
-    ToPrimary
+    ToPrimary,
+
+    /** This operation should be executed on a backup. If the current node is
+       not a backup, it should attempt to redirect to a backup, or else return
+       an error. Only read operations should be marked as ToBackup. */
```
achamayou, line 108:
```suggestion
       an error. Only read operations can be marked as ToBackup. */
```

achamayou (approved):
It could be nice to split `RedirectionStrategy` into `ReadOnlyRedirectionStrategy` and `ReadWriteRedirectionStrategy`, to specialise `set_redirection_strategy()` and mandate that `ToBackup` is only used on RO endpoints, and to have the same validation in `install_custom_endpoints_v1()`. A warning in the document also works, but leaves a little more room for error.

src/node/rpc/frontend.h
```diff
@@ -222,14 +224,32 @@ namespace ccf
             incoming_interface :
             interface_it.value().get<std::string>();
 
-          const auto primary_id = consensus->primary();
-          if (!primary_id.has_value())
+          std::optional<ccf::NodeId> target_node_id;
+          const auto nodes = InternalTablesAccess::get_trusted_nodes(tx);
+          {
+            const auto primary_id = consensus->primary();
```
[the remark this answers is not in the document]
eddyashton: I was actually wondering if we should do the opposite, and deliberately return _a_ (randomly chosen) backup, rather than the first each time, for better work distribution.

src/node/rpc/frontend.h
```diff
@@ -222,14 +224,32 @@ namespace ccf
...
+            {
+              target_node_id = primary_id;
+            }
+            else if (seeking_backup)
+            {
+              for (const auto& [id, _] : nodes)
+              {
+                if (id != primary_id)
+                {
+                  target_node_id = id;
```
maxtropets, line 241: Just curious.

1. Do we redirect to the same node each time? The return value of `get_trusted_nodes` is a sorted (by id) map.
2. Why don't we exclude the current node from the search?

eddyashton: 1. Yes, we do currently direct to the same node each time. I think we should probably randomise this if we're trying to distribute load, and @achamayou suggested [here](https://github.com/microsoft/CCF/pull/6275#discussion_r1642724711) we should also prefer up-to-date nodes. I'll add that now.
2. We should only enter this path at all if the current node isn't a suitable target, though admittedly it's a little distant (we checked `is_primary` earlier, and we're checking if each ID matches the `primary()` ID here). I think this means we can't self-redirect currently, but does mean a small code change in future could produce self-redirections. Hopefully the fix suggested for part 1 also mitigates this. I'm not sure which error case is worse - producing a self-redirection, or deciding there is no redirect target. If a self-redirection happens because of a temporary flakiness (ie - we thought we couldn't handle this request initially, but by the time we're producing the redirect we look viable enough), then self-redirect seems useful - it probably waits out the temporary issue. But it could lead to an infinite redirect loop. I think while its feasible for a self-redirect to be valid (ie - I don't _think_ the state can change in this gap, but maybe it can, or will be able to in future), I'd err towards allowing self-redirects, and trust that a client recognises (and forbids) infinite redirect loops.

PR #6279, opened 2024-06-17
Namespace more of the public headers

achamayou:
See #3580.

  - `::ds` is now `ccf::ds`
  - `::siphash` is now `ccf::siphash`
  - `::threading` is now `ccf::threading`, and `ccf/ds/thread_ids.h` has moved to `ccf/threading/thread_ids.h`
  - `::consensus` is now `ccf::consensus`
  - `::tls` is now `ccf::tls`

As we head towards 5.0 final, we have another opportunity to at least have a single top-level namespace, tighten public headers, and perhaps rationalise file and directory names.


eddyashton (approved):
Needs a CHANGELOG.

PR #6282, opened 2024-06-20
Historical queries cache size soft limit

### LINKED ISSUE #3234: Restrict historical query cache size
Currently, in-memory cached historical state gets evicted either because it expired (time-based) or because the handle has been manually freed, whichever happens first.
If the memory consumption grows to the point where it exceeds the available enclave heap size, then the node would crash.
This issue is about discussing an implementation plan to avoid this from happening.

One option could be to track the cache size (e.g. sum of KV key/value byte sizes) together with a configurable maximum size. If the maximum is reached, then the oldest cache entries would be evicted. I'm sure the devil is in the detail here.

maxtropets:
Closes #3234


include/ccf/historical_queries_interface.h
```diff
@@ -79,6 +81,13 @@ namespace ccf::historical
     virtual void set_default_expiry_duration(
       ExpiryDuration seconds_until_expiry) = 0;
 
+    /** Set the cache limit (in bytes) to evict least recently used requests
+     * from the cache after it's size grows beyond this limit. The limit is not
```
achamayou, line 85:
```suggestion
     * from the cache after its size grows beyond this limit. The limit is not
```

samples/apps/logging/logging.cpp
```diff
@@ -460,6 +460,15 @@ namespace loggingapp
         PUBLIC_RECORDS, context, 10000, 20);
       context.get_indexing_strategies().install_strategy(index_per_public_key);
 
+      // According to manual obvervation it's enough to start evicting old
+      // requests on historical perf test, but not to small to get stuck because
```
achamayou, line 464:
```suggestion
      // requests on historical perf test, but not too small to get stuck because
```

samples/apps/logging/logging.cpp
```diff
@@ -460,6 +460,15 @@ namespace loggingapp
...
+      // of a single request being larget than the cache.
+      //
+      // A much better approach would be to create a new endpoint for
```
achamayou, line 467: That does seem like a thing to do in runtime options, can we ticket it?

src/node/historical_queries.h
```diff
@@ -311,7 +316,7 @@ namespace ccf::historical
         if (!any_diff && (should_include_receipts == include_receipts))
         {
           HISTORICAL_LOG("Identical to previous request");
-          return;
+          return {removed, probably_added};
```
achamayou, line 319: added

src/node/historical_queries.h
```diff
@@ -493,6 +499,135 @@ namespace ccf::historical
 
     ExpiryDuration default_expiry_duration = std::chrono::seconds(1800);
 
+    // These two combine into an effective O(1) lookup/add/remove by handle.
```
achamayou, line 502:
```suggestion
    // These two combine into an effective O(log(N)) lookup/add/remove by handle.
```

src/node/test/historical_queries.cpp
```diff
@@ -948,6 +961,134 @@ TEST_CASE("Incremental progress")
...
 }
 
+TEST_CASE("StateCache soft zero limit with increasing")
+{
+  // Try get two states. Shouldn't be able to retrieve anything with 0 cache
+  // limit. After increasing to the size of first state only that one is
+  // available, but later overwritten by the second one, and finally all are
+  // evicted after setting again to 0.
+  //
+  // ! DISCLAIMER ! If you change this bare in mind that each attempt to get the
```
achamayou, line 971:
```suggestion
  // ! DISCLAIMER ! If you change this bear in mind that each attempt to get the
```

src/node/test/historical_queries.cpp
```diff
@@ -948,6 +961,134 @@ TEST_CASE("Incremental progress")
...
+  REQUIRE(!cache.get_state_at(1, seq_high));
+}
+
+TEST_CASE("StateCache dropping state explicitly")
+{
+  // Adding two states to the cache (limit is hit). Drop state2 and add state3.
+  // State1 should remain available, as well as state3, state2 was force-evicted
+  // from LRU because it's been dropped explicitly.
+  //
+  // ! DISCLAIMER ! If you change this bare in mind that each attempt to get the
```
achamayou, line 1036:
```suggestion
  // ! DISCLAIMER ! If you change this bear in mind that each attempt to get the
```

achamayou (approved):

samples/apps/logging/logging.cpp
maxtropets: #6299

eddyashton (approved):

PR #6283, opened 2024-06-20
Remove metrics endpoints

eddyashton:
Instead of us maintaining a metrics object, we provide method event hooks so that registries that want to do this can do it themselves, or else just log. In the governance registry, we log every request and its execution time. We also have some custom tracking of `jwt_refresh`, because it's externally triggered and otherwise hard to track.

For apps I currently log everything, but I think that's too verbose. Opening as draft to see what the perf impact is. Think the end result will be that apps _can_ log everything, if they want.


eddyashton:
Rough measure of the perf impact of logging every (`/app`) request as it completes:

```bash
$ for i in {1..5}; do ctest -VV -R pi_basic_virtual | grep "Average throughput"; done
81: Average throughput: 57120.40 tx/s
81: Average throughput: 57197.73 tx/s
81: Average throughput: 56755.04 tx/s
81: Average throughput: 56493.09 tx/s
81: Average throughput: 56622.10 tx/s

$ wc -l workspace/pi_basic_virtual_0/out 
100197 workspace/pi_basic_virtual_0/out
```

Compared to an equivalent build where we don't log:

```bash
$ for i in {1..5}; do ctest -VV -R pi_basic_virtual | grep "Average throughput"; done
81: Average throughput: 62644.75 tx/s
81: Average throughput: 60552.30 tx/s
81: Average throughput: 61906.70 tx/s
81: Average throughput: 61003.92 tx/s
81: Average throughput: 62806.77 tx/s

$ wc -l workspace/pi_basic_virtual_0/out
181 workspace/pi_basic_virtual_0/out
```

I think this is a significant enough hit that we don't want it by default, and the current implementation (where apps can easily opt in to logging or gathering metrics as they wish) seems fine. So I'm no longer logging in `UserEndpointRegistry`.

include/ccf/endpoint_registry.h
```diff
@@ -39,6 +39,25 @@ namespace ccf::endpoints
     PathTemplateSpec spec;
   };
 
+  struct RequestCompletedEvent
+  {
+    std::string method = "";
+    // This contains the path template against which the request matched. For
+    // instance `/user/{user_id}` rather than `/user/Bob`. This should be safe
+    // to log, though doing so still reveals (to anyone with stdout access)
+    // exactly which endpoints were executed and when.
```
achamayou, line 48: This may be particularly questionable in situations where the DynamicJSEndpointsRegistry is configured to use a private table, and the endpoint names and definitions are confidential. It is up the application code to be consistent here though.

include/ccf/endpoint_registry.h
```diff
@@ -39,6 +39,25 @@ namespace ccf::endpoints
     PathTemplateSpec spec;
   };
 
+  struct RequestCompletedEvent
+  {
+    std::string method = "";
```
achamayou, line 44: verb?

src/node/gov/gov_endpoint_registry.h
```diff
@@ -50,6 +50,26 @@ namespace ccf
          rpc_ctx.get_request_path() == "/gov/members/proposals:create");
     }
 
+    // Log these events on /gov frontend. Everything here should be public, so
```
achamayou, line 53:
```suggestion
    // Log these events on /gov frontend. Everything here is public, so
```

achamayou (approved):

include/ccf/endpoint_registry.h
eddyashton: I think this is `method` in HTTP terminology. We only use `verb` in some places for historical reasons (converting from JSON-RPC, which have a _method_ which is more like a HTTP _path_), but this seems like the sane name in a new API?

include/ccf/endpoint_registry.h
eddyashton: Agreed, this is a generality but not globally true - we don't log it, we just try to tell you how you might want to log it.

tests/jwt_test.py
```diff
@@ -487,24 +487,13 @@ def check_kv_jwt_keys_not_empty(args, network, issuer):
...
-        assert m["errors"] == 0, m["errors"]  # not used in jwt refresh endpoint
-        m["successes"] = m["calls"] - m["failures"]
-        return m
-
-
-def get_jwt_metrics(network) -> dict:
-    primary, _ = network.find_nodes()
-    with primary.client(network.consortium.get_any_active_member().local_id) as c:
-        r = c.get("/node/jwt_metrics")
+def get_jwt_refresh_endpoint_metrics(primary) -> dict:
```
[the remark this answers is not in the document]
eddyashton: No, we're (for now) keeping the old-style metrics (ie - per-node state, updated during processing of each call), but _only_ for the `/jwt_keys/refresh` method. And now driven in a different way (instead of a `increment_metrics_X` call for every endpoint, we respond to the generic `request_completed` event in the node frontend to say "if this is `/jwt_keys/refresh`, count it"). Why? Because this test still uses it to confirm when/how calls to `/jwt_keys/refresh` completed, and it's hard to see that otherwise (this is an internal call, so we don't get to see any errors client side, so I think its fine to retain metrics for it).

include/ccf/endpoint_registry.h
achamayou: My bad, I thought there were verbs, and the word is used, but you are right that they formally are methods.

PR #6285, opened 2024-06-21
Mandate that signed payloads are endpoint-specific and timestamped in Programmability sample

achamayou:
See remaining item on #6223. Introduce `DynamicJSEndpointRegistry::record_action_details_for_audit_v1` and `DynamicJSEndpointRegistry::is_original_action_execution_v1` to allow an application making use of the programmability feature to easily implement auditability, and protect users allowed to update the application against replay attacks.

This allows an application to define two parameters in the protected header of COSE Sign1 envelopes, for the purposes of:

1. Making payloads specific to an endpoint, to avoid possible confusion attacks
2. Making payloads timestamped, to implement replay protection cheaply

I have hardcoded the window size for now, but moving it to the runtime_options may be a good idea, perhaps as `cose_replay_protection_window_size`, or something similar. It's difficult to imagine this being useful though.


samples/apps/programmability/programmability.cpp
```diff
@@ -240,6 +257,39 @@ namespace programmabilityapp
         auto audit_info = ctx.tx.template rw<AuditInfoValue>(
           fmt::format("{}.audit.info", CUSTOM_ENDPOINTS_NAMESPACE));
         audit_info->put({format, AuditInputContent::BUNDLE, user_id.value()});
+        if (format == AuditInputFormat::COSE)
```
achamayou, line 260: This (plus the lines above) seems like something we could, and should compress quite a bit. This requires two main API extensions I think:

1. `AuditInputContent` and `AuditInputFormat` equivalents in `registry.h`, and probably moving `get_body()` there.
2. Either a string or enum to expose error details during replay protection.

This can then be replaced by one, or two calls (I think two is best for clarity, naming the replay protection check is already quite difficult).

There is not a lot of short-term benefit (there are probably not going to be lots more auditable calls), but will be helpful if we look to expose audit getters for example.

samples/apps/programmability/programmability.cpp
```diff
@@ -391,26 +475,50 @@ namespace programmabilityapp
...
-          audit->put(ctx.rpc_ctx->get_request_body());
-          auto audit_info = ctx.tx.template rw<AuditInfoValue>(
-            fmt::format("{}.audit.info", CUSTOM_ENDPOINTS_NAMESPACE));
-          audit_info->put({format, AuditInputContent::BUNDLE, user_id.value()});
+          // Make operation auditable
+          record_action_details_for_audit_v1(
+            ctx.tx,
+            format,
+            user_id.value(),
+            "PUT /app/custom_endpoints",
```
eddyashton, line 496:
```suggestion
            "PATCH /app/custom_endpoints/runtime_options",
```

src/endpoints/authentication/cose_auth.cpp
```diff
@@ -22,7 +22,7 @@ namespace ccf
     static constexpr auto HEADER_PARAM_MSG_TYPE = "ccf.gov.msg.type";
     static constexpr auto HEADER_PARAM_MSG_PROPOSAL_ID =
       "ccf.gov.msg.proposal_id";
-    static constexpr auto HEADER_PARAM_MSG_CREATED_AT =
+    static constexpr auto HEADER_PARAM_MSG_MSG_CREATED_AT =
```
eddyashton, line 25: Why are these renamed to `MSG_MSG`?

include/ccf/js/registry.h
```diff
@@ -127,6 +131,28 @@ namespace ccf::js
     ccf::ApiResult get_js_runtime_options_v1(
       ccf::JSRuntimeOptions& options, kv::ReadOnlyTx& tx);
 
+    /**
+     * Record action details by storing them in KV maps using a common format,
+     * for the purposes of offline audit using the ledger.
+     */
+    ccf::ApiResult record_action_details_for_audit_v1(
```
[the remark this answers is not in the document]
achamayou: Not sure about the value of "trail" here, since this is a transactional, point-in-time write, but agree "details" does not add much. `record_action_for_audit_v1` seems just as good.

include/ccf/js/registry.h
```diff
@@ -127,6 +131,28 @@ namespace ccf::js
...
+      const std::string& user_id,
+      const std::string& action_name,
+      const std::vector<uint8_t>& action_body);
+
+    /**
+     * Check an action is original, as opposed to replayed, by looking it up
+     * in the history of recent actions. To place an upper bound on the history
+     * size, an authenticated timestamp (@p created_at) is required.
+     */
+    ccf::ApiResult is_original_action_execution_v1(
```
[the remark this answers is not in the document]
achamayou: `nonreplayed` is not really a word though, how about `check_action_not_replayed_v1()`. I prefer to avoid negatives in API names, because they lead to double negatives which can be confusing, but with a non-boolean return, that's less of a problem.

I'd rather keep the API returns uniform as much as possible, this is not a compelling case for breaking that consistency.

include/ccf/base_endpoint_registry.h
```diff
@@ -96,6 +137,19 @@ namespace ccf
...
+     * Returns the history of view changes since the given view, which defaults
+     * to the start of time.
+     *
+     * A view change is characterised by the first sequence number in the new
+     * view.
+     */
+    ApiResult get_view_history_v2(
+      std::vector<ccf::TxID>& history,
+      ccf::View since,
+      ccf::InvalidArgsReason& reason);
```
achamayou, line 151: Introduced so that all calls that return InvalidArgs populate a reason.

eddyashton (approved):

PR #6289, opened 2024-06-24
Add sample of app-defined JS extension

eddyashton:


samples/apps/programmability/programmability.cpp
```diff
@@ -25,6 +26,137 @@ namespace programmabilityapp
...
+            return true;
+          }
+        }
+      }
+    }
+
+    return false;
+  }
+
+  class MyExtension : public ccf::js::extensions::ExtensionInterface
```
achamayou, line 59: RBACExtension?

achamayou (commented):
LGTM, I think would benefit a changelog entry and some doc, but all of the programmability API probably would at this point and that's likely a separate item.

achamayou (approved):

Issue #6222, opened 2024-06-03
Cleanup old JWT tables in 5.0.1+


maxtropets:
We kept Legacy:: tables for JWT in https://github.com/microsoft/CCF/pull/6175, for reasons explained here
https://github.com/microsoft/CCF/pull/6175#discussion_r1620308872

This can be safely removed after ensuring old tables are no longer used anywhere in the system.

achamayou:
Checking before and after can be done with `read_ledger.py` against snapshots:

```
$ read_ledger.py -s workspace/app_space_js_e2e_logging_cft_0/0.snapshots/snapshot_987_992.committed -t public:ccf.gov.jwt.public_signing_keys_metadata
Reading snapshot from workspace/app_space_js_e2e_logging_cft_0/0.snapshots/snapshot_987_992.committed (committed)
  txid 0.987 (41 public tables) [144658 bytes]
    table "public:ccf.gov.jwt.public_signing_keys_metadata" (1 write):
      95254152-e9dd-465a-b4cf-7893ed2e1cf0:
        [
          {
[...]

PR #6302, opened 2024-06-27
Remove Old JWT Tables

sidmore:
#6222 

## Remove
1. References to Legacy tables
2. Add Remove Old JWT table proposal


samples/constitutions/default/actions.js
```diff
@@ -1494,4 +1494,28 @@ const actions = new Map([
       function (args) {},
     ),
   ],
+  [
+    "remove_old_jwt_tables",
+    new Action(
+      function (args) {
+        // Validate that no arguments are passed
+        if (
+          args !== undefined &&
```
achamayou, line 1503: The first comparison should be sufficient to mandate that args is undefined.

achamayou:
@maxtropets please review.

@sidmore because this change requires user-action, can you add an entry in the CHANGELOG to describe it? There seems to be an OpenAPI change (not immediately clear why), so please run the schema test locally and diff the schemas. If the change is expected, updating the OpenAPI version in the governance frontend file, rebuilding, re-running the test and checking in the schema is the way to go. If the change is unexpected, let's discuss.

samples/constitutions/default/actions.js
```diff
@@ -1494,4 +1494,28 @@ const actions = new Map([
...
+          args !== null &&
+          Object.keys(args).length > 0
+        ) {
+          throw new Error(
+            "remove_old_jwt_tables does not accept any arguments",
+          );
+        }
```
eddyashton, line 1510: We have a helper function that does this already - best to use that for a consistent error message.

```suggestion
        checkNone(args);
```

tests/infra/consortium.py
```diff
@@ -875,3 +875,10 @@ def check_for_service(self, remote_node, status, recovery_count=None):
                 assert (
                     recovery_count is None or current_recovery_count == recovery_count
                 ), f"Current recovery count {current_recovery_count} is not expected {recovery_count}"
+
+    def remove_old_jwt_tables(self, remote_node, issuer):
```
eddyashton, line 879: Looks like we're not actually calling this anywhere. We should add a call to this in the `lts_compatibility` test (or one of the tests in `recovery.py` that recovers from an old ledger), to fully confirm that we can have a service where these tables are present _before_ the call, removed _after_ the call, and not repopulated if we do a jwt refresh.

maxtropets:
We also probably shall prepare some template for operators (not sure where to put it though, probably @achamayou could elaborate on our vision here), 
 
The goal is to have an ultimate step-by-step guide for how to upgrade, probably introduce some scripts to automate some work. I'm not really sure what's the average operator expertise level is.

sidmore:
> @maxtropets please review.
> 
> @sidmore because this change requires user-action, can you add an entry in the CHANGELOG to describe it? There seems to be an OpenAPI change (not immediately clear why), so please run the schema test locally and diff the schemas. If the change is expected, updating the OpenAPI version in the governance frontend file, rebuilding, re-running the test and checking in the schema is the way to go. If the change is unexpected, let's discuss.

@achamayou i have updated the schema for gov

maxtropets (changes requested):

samples/constitutions/default/actions.js
```diff
@@ -1494,4 +1494,20 @@ const actions = new Map([
...
+    "remove_old_jwt_tables",
+    new Action(
+      function (args) {
+        // Validate that no arguments are passed
+        checkNone(args);
+      },
+      function (args) {
+        // Clear the JWT public signing key table
+        ccf.kv["public:ccf.gov.jwt.public_signing_key"].clear();
+
```
sidmore, line 1507: add a trigger snapshot action

tests/recovery.py
```diff
@@ -528,6 +528,93 @@ def test_share_resilience(network, args, from_snapshot=False):
...
+            txid = TxID(tx.gcm_header.view, tx.gcm_header.seqno)
+            tables = tx.get_public_domain().get_tables()
+            pub_keys = tables["public:ccf.gov.jwt.public_signing_key"]
+            assert (pub_keys is not None), "PubKeys is not None"
+            pub_key_issuer = tables["public:ccf.gov.jwt.public_signing_key_issuer"]
+            assert (pub_key_issuer is not None), "PubKeys is not None"
+    
+    # Submit Proposal to remove old JWT Tables
+    network.consortium.remove_old_jwt_tables(primary)
+
```
sidmore, line 603: stop the network here to trigger a flush to disk

achamayou:
@sidmore now that `release/5.x` is cut, it'd be great to update this and get it merged.

achamayou:
We aren't going to have the time to complete this by 6.0.0-rc0, and because of the multi-IdP support changes, there are additional tables we need to clean up too. This is tracked in #6222, and can start as soon as we have cut `release/6.x`, and until `7.0.0-rc0` at the latest.

This PR is cross-linked and can be used as a starting point when we schedule this work.

PR #6309, opened 2024-06-28
Don't throw inside `FATAL` logging macro

### LINKED ISSUE #3334: `LOG_FATAL_FMT` is not fatal
It simply throws an exception which may or may not be caught in upper levels.

This is in contrast to the description of the log level:

```
FATAL, // fatal errors that lead to a termination of the program/enclave
```

eddyashton:
Resolves #3334.


src/host/main.cpp
```diff
@@ -62,19 +62,6 @@ void print_version(size_t)
...
-{
-  auto ev = std::getenv(envvar.c_str());
-  if (ev == nullptr)
-  {
-    LOG_FATAL_FMT(
-      "Environment variable \"{}\" for {} is not set", envvar, name);
-  }
-  LOG_INFO_FMT("Reading {} from environment {}", name, envvar);
-  return ev;
-}
```
eddyashton, line 76: Unused.

achamayou: I think that was once used for UVM endorsements, but that has all moved to configuration now.

achamayou (approved):

PR #6314, opened 2024-06-28
Run unit tests concurrently in CI

achamayou:


.github/workflows/build.yml
```diff
@@ -69,15 +69,12 @@ jobs:
...
-        shell: bash
-        if: "${{ matrix.platform.name != 'snp' }}" # Needs 1ES Pool support
-
-      - name: "Partitions Test ${{ matrix.platform.name }}"
-        run: |
-          set -ex
-          cd build
-          source env/bin/activate && ctest -VV --timeout 240 -L partitions -C partitions
+          # Unit tests
+          ./tests.sh --output-on-failure -L unit -j16
```
eddyashton, line 73:
```suggestion
          ./tests.sh --output-on-failure -L unit -j$(nproc --all)
```

Not critical, but _if_ this works and lets us get core-count programmatically, we'll be a little safer when/if this ends up on a 4-core box.

eddyashton (approved):

PR #6320, opened 2024-07-01
Streamline CodeQL pipeline

achamayou:
The dependency step is redundant now that we run in containers. Pending a successful run of https://github.com/microsoft/CCF/actions/runs/9742123789, including reporting.

We should also look at running make with concurrent jobs, to avoid taking the pool for so long. The current setup is straight from a sample (also why it's make and not ninja), and while I am not sure if adding concurrency will cause problems for CodeQL, I found not explicit prohibition in the documentation.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -31,11 +31,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Override automatic language detection by changing the below list
-        # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
         language: ["cpp"]
-        # Learn more...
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -31,50 +31,19 @@ jobs:
...
-      #   make release
-      - run: |
-          cd getting_started/setup_vm
-          sudo apt update
-          sudo apt install -y ansible software-properties-common bsdmainutils dnsutils
-          sudo ansible-playbook ccf-dev.yml --extra-vars "platform=virtual" --extra-vars "require_open_enclave=false"
-        name: Install dependencies
 
       - run: |
           set -ex
```
achamayou, line 49:
```suggestion
          set -ex
          rm -rf build
```

PR #6322, opened 2024-07-02
Add customisable error handling to hist. queries

maxtropets:


maxtropets:
Let's revise the naming here...

src/node/historical_queries_adapter.cpp
```diff
@@ -190,6 +190,57 @@ namespace ccf::historical
     return tx_id_opt;
   }
 
+  void default_error_handler(
+    HistoricalQueryErrorCode err,
+    std::string reason,
+    endpoints::CommandEndpointContext& args)
+  {
+    if (err == HistoricalQueryErrorCode::InternalError)
```
achamayou, line 198: switch?

achamayou (approved):

src/node/historical_queries_adapter.cpp
maxtropets: It requires all variable declaration to be moved out of the whole `switch` block, and to me personally looks less readable for inner blocks with multiple function calls inside.

It also might grow more for new error codes in the future, as well as more `ctx` calls in each block, which will make this block grow even further.

That's why I changed that to `if`-`else`.

Let me know if you see a good reason to go back to `switch`.

achamayou: I don't understand, there are no declarations here, and even if there were, it's no more verbose to have blocks under each case as it is with an if-else. It's still one block per case?

```
switch (err)
{
 case FirstErr:
 {
   auto a = foo();
   break;
 }
...
}
```

Aside from performance, switch gives us better static checking with our compiler settings, missing cases will be flagged. This in my opinion outweighs syntactic concerns, particularly if we intend to grow the enum.

src/node/historical_queries_adapter.cpp
```diff
@@ -374,4 +425,132 @@ namespace ccf::historical
       ccf::endpoints::EndpointFunction,
       ccf::endpoints::EndpointContext>(f, node_context, available, extractor);
   }
+
+  template <
+    class TQueryHandler,
+    class TEndpointFunction,
+    class TEndpointContext>
+  TEndpointFunction _adapter_v4(
```
eddyashton, line 433: To avoid repetition, can we implement `_adapter_v3` in terms of `_adapter_v4`? I think that's just:

```
template <...>
TEndpointFunction _adapter_v3(...)
{
  return _adapter_v4(..., default_error_handler);
}
```

include/ccf/historical_queries_adapter.h
```diff
@@ -56,21 +75,35 @@ namespace ccf::historical
     ccf::SeqNo seqno,
     std::string& error_reason);
 
-  ccf::endpoints::EndpointFunction adapter_v3(
+  FMT_DEPRECATED ccf::endpoints::EndpointFunction adapter_v3(
```
eddyashton, line 78:
```suggestion
  CCF_DEPRECATED("Replaced by _v4") ccf::endpoints::EndpointFunction adapter_v3(
```

include/ccf/historical_queries_adapter.h
```diff
@@ -56,21 +75,35 @@ namespace ccf::historical
...
+    ccf::AbstractNodeContext& node_context,
+    const CheckHistoricalTxStatus& available,
+    const CommandTxIDExtractor& extractor = txid_from_header,
+    const ErrorHandler& ehandler = default_error_handler);
+
+  ccf::endpoints::EndpointFunction read_write_adapter_v4(
+    const HandleReadWriteHistoricalQuery& f,
+    ccf::AbstractNodeContext& node_context,
+    const CheckHistoricalTxStatus& available,
+    const CommandTxIDExtractor& extractor = txid_from_header,
```
eddyashton, line 107:
```suggestion
    const ReadOnlyTxIDExtractor& extractor = txid_from_header,
    const ErrorHandler& ehandler = default_error_handler);

  ccf::endpoints::EndpointFunction read_write_adapter_v4(
    const HandleReadWriteHistoricalQuery& f,
    ccf::AbstractNodeContext& node_context,
    const CheckHistoricalTxStatus& available,
    const TxIDExtractor& extractor = txid_from_header,
```

`Commands` have no access to the KV, `ReadOnly`s have read access (via a `ReadOnlyTx` object), and unprefixed or `ReadWrite`s have read-write access (via a `Tx` object). It's possible that someone wants to read from the KV to decide the target TxID, and we also let them write to it if they really wanted to. We should retain that, and not force them to extract the TxID purely from the request (with no KV access) - our _default_ `txid_from_header` extractor happens to be command-only, and many will be command-only, but we don't want to lose the ability for apps to use the KV here.

CHANGELOG.md
```diff
@@ -27,9 +27,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Introduce `DynamicJSEndpointRegistry::record_action_for_audit_v1` and `DynamicJSEndpointRegistry::check_action_not_replayed_v1` to allow an application making use of the programmability feature to easily implement auditability, and protect users allowed to update the application against replay attacks (#6285).
 - Endpoints now support a `ToBackup` redirection strategy, for requests which should never be executed on a primary. These must also be read-only. These are configured similar to `ToPrimary` endpoints, with a `to_backup` object (specifying by-role or statically-addressed targets) in each node's configuration.
 
-## Changed
+### Changed
 
 - Updated Open Enclave to [0.19.7](https://github.com/openenclave/openenclave/releases/tag/v0.19.7).
+- `ccf::historical::adapter_v3` becomes deprecated and gets replaced by `ccf::historical::read_only_adapter_v4` and `ccf::historical::read_write_adapter_v4`. Users are now capable of passing a custom error handler to the adapter to customise RPC responses for internal historical queries errors, which are listed in `ccf::historical::HistoricalQueryErrorCode` enum.
```
eddyashton, line 33: I think we should split this out into "Added" and "Deprecated" sections. Deprecated should say that `adapter_v3` (and its variants) are deprecated, and should be replaced by `_v4`. Added should say that `_v4` now exists, which allows custom error handler to be passed.

src/node/historical_queries_adapter.cpp
maxtropets: Agree for combining error for one case.

Disagree to compiler warnings because I tried to comment out some cases, compiler doesn't notice.

Anyway, changing to switch as the majority prefers it.

include/ccf/historical_queries_adapter.h
maxtropets: Done. Same for ErrorHandler, as discussed in person.

CHANGELOG.md
maxtropets: Done

PR #6326, opened 2024-07-03
OpenAPI suggests that "operationId" (or "operationref") is mandatory.

lemmy:
nt


achamayou:
From https://swagger.io/specification/:

> In the following description, if a field is not explicitly REQUIRED or described with a MUST or SHALL, it can be considered OPTIONAL.

The description for operationId is under https://swagger.io/specification/:

> Unique string used to identify the operation. The id MUST be unique among all operations described in the API. The operationId value is case-sensitive. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.

There is no mention of REQUIRED, and although the unicity is a MUST, there is no explicit statement about existence. Issues on the OpenAPI repository (#1019, #1907) asking for it be made mandatory have been rejected, and clarify that it is not mandatory.

This next bit is my opinion: there is no valid reason to demand that someone manually writes an operationId, creating one from the path/verb is trivial to automate correctly. It is bizarre that the SDK generators are both demanding this manual de-normalisation while complaining about duplicates.

Anyway, there is clearly no downside to having this, thank you for adding it.

src/endpoints/endpoint_registry.cpp
```diff
@@ -34,6 +34,14 @@ namespace ccf::endpoints
         ds::openapi::path(document, endpoint->full_uri_path),
         http_verb.value());
 
+      // Add what appears a *mandatory* operationId, which is expected to be
+      // unique across the spec.
+      std::string p =
+        std::regex_replace(endpoint->full_uri_path, std::regex("[/_]"), "");
+      std::string s = llhttp_method_name(http_verb.value());
+      ccf::nonstd::to_lower(s);
+      path_op["operationId"] = fmt::format("{}_{}", p, s);
```
achamayou, line 43: If the path is templated, it will contain `{` and `}`. Is that ok in an `operationId`?

The spec is gloriously vague about this:

> it is RECOMMENDED to follow common programming naming convention

achamayou (approved):

achamayou:
To fix the tests (I tried to do, but I can't seem to write to your fork), it is necessary to:

1. rev up `openapi_info.document_version` in `src/node/rpc/node_frontend.h`, `src/node/rpc/member_frontend.h` and `samples/app/logging/logging.cpp` (I suggest +`0.1.0` in this case, but any amount will work).
2. re-build
3. re-run the schema test `./tests.sh -VV -R schema` which will produce the new OpenAPI schemas (it will fail, but that's ok)
5. commit them

lemmy:
> I tried to do, but I can't seem to write to your fork

You and the others should have collaborator permissions in my fork:

image

src/endpoints/endpoint_registry.cpp
lemmy: I've learned that nswag generates different C# stubs depending on how `operationId` is formatted.  For example, `_` causes multiple stubs to be generated.  Will investigate a bit more to find out what "common programming naming convections" are.

achamayou:
@lemmy weird, tried again and it worked. Tests should pass now.

lemmy:
Using the verb concatenated with the camel-cased path appears to be a common default. For example, the verb `delete` combined with the path `app/log/private/all` becomes the operationId `DeleteAppLogPrivateAll`. Does the CCF codebase have a built-in function for converting a string to camel case?

eddyashton:
> Does the CCF codebase have a built-in function for converting a string to camel case?

No, though it's easy enough to add. Proposed impl is here: #6349, let me know if that works for you and we'll get it merged.

src/endpoints/endpoint_registry.cpp
lemmy: The `{` and `}` are now eliminated.

eddyashton (approved):

PR #6329, opened 2024-07-04
Publish release containers to GHCR

achamayou:


.github/workflows/containers-ghcr.yml
```diff
@@ -0,0 +1,79 @@
...
+          IMG_NAME=${{ env.REGISTRY }}/${{ github.repository }}/app/${{ matrix.type }}${{ matrix.run_js && '-js' || '' }}/${{ matrix.platform }}
+          echo "img_name=${IMG_NAME}" >> $GITHUB_OUTPUT
+        id: tref
+
+      - name: Extract metadata (tags, labels) for the image
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ steps.tref.outputs.img_name }}
+        if: ${{ steps.tref.outputs.major_release == '5'}}
```
eddyashton, line 56:
```suggestion
        if: ${{ steps.tref.outputs.major_release >= '5'}}
```

.github/workflows/containers-ghcr.yml
```diff
@@ -0,0 +1,79 @@
...
+          context: .
+          file: ./docker/app_${{ matrix.type }}
+          build-args: |
+            clang_version=${{ matrix.clang_version }}
+            platform=${{ matrix.platform }}
+            "ansible_vars=ccf_ver=${{ steps.tref.outputs.tag }} ${{ matrix.run_js && 'run_js=true' || '' }}"
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+        if: ${{ steps.tref.outputs.major_release == '5'}}
```
eddyashton, line 71:
```suggestion
        if: ${{ steps.tref.outputs.major_release >= '5'}}
```

.github/workflows/containers-ghcr.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Attest container
+        uses: actions/attest-build-provenance@v1
+        with:
+          subject-name: ${{ steps.tref.outputs.img_name }}
+          subject-digest: ${{ steps.push.outputs.digest }}
+          push-to-registry: true
+        if: ${{ steps.tref.outputs.major_release == '5'}}
```
eddyashton, line 79:
```suggestion
        if: ${{ steps.tref.outputs.major_release >= '5'}}
```

.github/workflows/containers-ghcr.yml
```diff
@@ -0,0 +1,79 @@
...
+
+jobs:
+  build_and_publish:
+    name: "Build and Publish Release Containers"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform: [sgx, virtual, snp]
+        type: [dev, run]
+        run_js: [true, ""]
```
eddyashton, line 20: Why is `run_js` a separate option here, rather than a value in `type`? eg - `type: [dev, run, "run-js"]`

.github/workflows/containers-ghcr.yml
```diff
@@ -0,0 +1,79 @@
...
+        clang_version: ["11", "15"]
+        exclude:
+          - type: dev
+            run_js: true
+          - platform: sgx
+            clang_version: "15"
+          - platform: virtual
+            clang_version: "11"
+          - platform: snp
+            clang_version: "11"
```
eddyashton, line 30: Doing the platform <-> clang_version pairing with exclusions seems risky, can't we bind these together in the matrix?

```yaml
matrix:
  platform:
    - name: sgx
      clang_version: 11
    - name: virtual
      clang_version: 15
    - name: snp
      clang_version: 15
```

.github/workflows/containers-ghcr.yml
achamayou: I think we can probably drop the clauses entirely after we're done with the last 4.x release, but sure.

.github/workflows/containers-ghcr.yml
achamayou: Because having it as a bbol allows inline conditions for different values, which we use for `${{ matrix.run_js && 'run_js=true' || '' }}`. We could also do that with an action level `if:`, but we would need to duplicate the whole action. I am open to better ideas though.

.github/workflows/containers-ghcr.yml
achamayou: I'm inclined to keep it because 1. it works, and 2. it will disappear with SGX.

PR #6332, opened 2024-07-04
Timeout idle sessions

### LINKED ISSUE #4377: Timeout inactive TLS sessions
We now timeout when forwarded requests take too long, but do nothing for idle TLS sessions to the current node. We should close those after an operator-defined period of inactivity.

eddyashton:
Closes #4377.

### DIFF
--- a/src/host/configuration.h
+++ b/src/host/configuration.h
@@ -74,6 +74,7 @@ namespace host
     ccf::ds::TimeString slow_io_logging_threshold = {"10ms"};
     std::optional<std::string> node_client_interface = std::nullopt;
     ccf::ds::TimeString client_connection_timeout = {"2000ms"};
+    std::optional<ccf::ds::TimeString> idle_connection_timeout = std::nullopt;
     std::optional<std::string> node_data_json_file = std::nullopt;
     std::optional<std::string> service_data_json_file = std::nullopt;
     bool ignore_first_sigterm = false;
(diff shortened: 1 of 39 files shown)


src/host/configuration.h
```diff
@@ -74,6 +74,7 @@ namespace host
     ccf::ds::TimeString slow_io_logging_threshold = {"10ms"};
     std::optional<std::string> node_client_interface = std::nullopt;
     ccf::ds::TimeString client_connection_timeout = {"2000ms"};
+    std::optional<ccf::ds::TimeString> idle_connection_timeout = std::nullopt;
```
eddyashton, line 77: I've wrapped this in `optional` so that we can retain the option for the old behaviour (no automated cleanup), in case someone has a hard-to-fix client with long-lived idle sessions.

Open question is what should the default be? A sensible default value is probably around `60s`, so that most users benefit from this by default, but that's potentially a breaking change?

tests/connections.py
```diff
@@ -264,6 +265,73 @@ def create_connections_until_exhaustion(
...
+        },  # With no timeout, idle sessions are never killed
+    ]
+
+    def verbose_sleep(sleep_time):
+        slept = 0
+        step_size = 1
+        LOG.info(f"Sleeping {sleep_time}s")
+        while slept < sleep_time:
+            next_sleep = min(sleep_time - slept, step_size)
+            time.sleep(next_sleep)
```
achamayou, line 283: At the second granularity this is probably not going to matter, but sleep is _at least_, not exactly, so something that keeps track of `slept` using `time.time()` will be somewhat more accurate.

achamayou (approved):

achamayou (approved):

PR #6336, opened 2024-07-05
Module loader implementations do not need to be public

achamayou:


eddyashton:
For completeness: This is an explicit minimisation of the supported API, that we may end up reversing in future, but I think this is a sensible removal at present. We expect that the JS programmability will only be accessed by subclassing `DynamicJSEndpointRegistry`, in which case all module loading is handled by framework code. We currently let apps create a `ccf::js::core::Context` directly, so they _could_ interpret JS code in other contexts if they wish. But:
a) we may pull that back as well,
b) we don't have any requests for it directly,
c) and even hypothetical uses we've discussed shouldn't need module loading.

So I think hiding the module loading instances is sensible.

achamayou:
@eddyashton 100%, thank you for expanding on that. `ccf::js::core::Context` can still be used punctually, from code that assembles JS separately, such as policy snippets. We can revisit this if there is a concrete ask.

eddyashton (approved):

PR #6349, opened 2024-07-10
Add a `nonstd::camel_case` util function

eddyashton:
Useful for `https://github.com/microsoft/CCF/pull/6326#issuecomment-2214510569`, potentially other situations.

### DIFF
--- a/include/ccf/ds/nonstd.h
+++ b/include/ccf/ds/nonstd.h
@@ -6,10 +6,14 @@
 #include <array>
 #include <cctype>
 #include <filesystem>
+#include <regex>
 #include <string>
 #include <string_view>
 #include <type_traits>
 #include <vector>
+
+#define FMT_HEADER_ONLY
+#include <fmt/format.h>
 
 /**
  * This file defines various type traits and utils that are not available in the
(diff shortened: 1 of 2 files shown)


include/ccf/ds/nonstd.h
```diff
@@ -288,4 +292,32 @@ namespace ccf::nonstd
 
     return resolved.lexically_normal().string();
   }
+
+  static inline std::string camel_case(
```
achamayou, line 296: Let's start a `src/ds/nonstd.cpp` and stop inlining everything.

achamayou (approved):

lemmy:
FYI: Trying this PR out in the context of #6326 right now.

lemmy:
@eddyashton Would camel-casing all-caps strings such as "POST", "DELETE", ... also be an input that the function could/should accept?

include/ccf/ds/nonstd.h
achamayou: Actually let's not export that in public headers either.

eddyashton:
Flattened this into #6326, will add followup (making definition private) next.

PR #6356, opened 2024-07-11
[Draft] Experiment with release build in CI

achamayou:
Can we trade off some build time for some test time?


maxtropets:
Let's make a workaround to enable asserts first.

But I personally think debug is fine unless it saves significantly much more time.

achamayou:
@maxtropets agreed, this is just to get some numbers for comparison, not a proposed change.

achamayou:
This change

image

Before

image

Good benefit for SGX, but not otherwise. I think it's not worth doing.

PR #6357, opened 2024-07-12
Zero out idle time when deallocating the socket

maxtropets:
Avoid spamming the output when session closure flow doesn't match idle times.

image


achamayou:
Good catch before RC1.

achamayou (approved):

PR #6359, opened 2024-07-12
More informative error messages for invalid txids

achamayou:
Before:

```
curl -k
https://127.0.0.1:8000/node/tx?transaction_id=0.0
{"error":{"code":"InternalError","message":"Error code: InternalError"}}

curl -k https://127.0.0.1:8000/app/tx?transaction_id=1.1
{"status":"Invalid","transaction_id":"1.1"}

curl -k https://127.0.0.1:8000/app/tx?transaction_id=1.0
{"error":{"code":"InternalError","message":"Error code: InternalError"}}

$ curl -k https://127.0.0.1:8000/app/tx?transaction_id=0.1
{"status":"Invalid","transaction_id":"0.1"}

curl -k https://127.0.0.1:8000/app/tx?transaction_id=2.0
{"error":{"code":"InternalError","message":"Error code: InternalError"}}
```

After:

```
curl -k https://127.0.0.1:8000/app/tx?transaction_id=0.0
{"error":{"code":"InvalidQueryParameterValue","message":"The value '0.0' passed as query parameter 'transaction_id' could not be converted to a valid Transaction ID."}}

curl -k https://127.0.0.1:8000/app/tx?transaction_id=1.1
{"status":"Invalid","transaction_id":"1.1"}

curl -k https://127.0.0.1:8000/app/tx?transaction_id=1.0
{"error":{"code":"InvalidQueryParameterValue","message":"The value '1.0' passed as query parameter 'transaction_id' could not be converted to a valid Transaction ID."}}

curl -k https://127.0.0.1:8000/app/tx?transaction_id=0.1
{"error":{"code":"InvalidQueryParameterValue","message":"The value '0.1' passed as query parameter 'transaction_id' could not be converted to a valid Transaction ID."}}

curl -k https://127.0.0.1:8000/app/tx?transaction_id=2.0
{"error":{"code":"InvalidQueryParameterValue","message":"The value '2.0' passed as query parameter 'transaction_id' could not be converted to a valid Transaction ID."}}
```


maxtropets (approved):

achamayou:
I wonder if it is worth preserving "Invalid" for 0.1 and 1.1, i.e. you can always ask about the first transaction, even for views that we never use.

include/ccf/tx_id.h
```diff
@@ -42,8 +42,8 @@ namespace ccf
   // transaction executed by CCF.
   struct TxID
   {
-    View view;
-    SeqNo seqno;
+    View view = 0;
+    SeqNo seqno = 0;
```
eddyashton, line 46:
```suggestion
    View view = VIEW_UNKNOWN;
    SeqNo seqno = SEQNO_UNKNOWN;
```

include/ccf/tx_id.h
```diff
@@ -64,7 +64,7 @@ namespace ccf
         const auto view_sv = sv.substr(0, separator_idx);
         const auto [p, ec] =
           std::from_chars(view_sv.begin(), view_sv.end(), tx_id.view);
-        if (ec != std::errc() || p != view_sv.end())
+        if (ec != std::errc() || p != view_sv.end() || tx_id.view < 2)
```
eddyashton, line 67:
```suggestion
        if (ec != std::errc() || p != view_sv.end() || tx_id.view == VIEW_UNKNOWN)
```

I don't think we should codify anywhere that `1` is an invalid/illegal/view. It's an incidental implementation detail that `1` is an _empty_ view on current CCF services, but I think we could change/fix that at any point, and shouldn't introduce stronger claims about it. In particular, for the current use case, `1.1` should be a valid TxID, with `TxStatus::Invalid` (because `2.1` is committed).

include/ccf/tx_id.h
```diff
@@ -74,7 +74,7 @@ namespace ccf
         const auto seqno_sv = sv.substr(separator_idx + 1);
         const auto [p, ec] =
           std::from_chars(seqno_sv.begin(), seqno_sv.end(), tx_id.seqno);
-        if (ec != std::errc() || p != seqno_sv.end())
+        if (ec != std::errc() || p != seqno_sv.end() || tx_id.seqno < 1)
```
eddyashton, line 77:
```suggestion
        if (ec != std::errc() || p != seqno_sv.end() || tx_id.seqno == SEQNO_UNKNOWN)
```

Semantically identical (since `txid.seqno` is unsigned, any negatives will already not parse), but I think clearer.

include/ccf/tx_id.h
achamayou: As commented above, we have previously effectively decided that `1.1` could be, but isn't a valid TxID, whereas `0.0` is simply malformed. I don't know that this was a particularly good decision, but staying compatible makes sense (continue to return 200/Invalid and avoid breaking anyone depending on that), what I want to fix is returning a 500 without detail for `0.0`, where 400/NotAValidTxID is clearly preferable and should not break any existing usage.

lemmy:
@marina-p fuzzed CCF with https://github.com/microsoft/restler-fuzzer and found the following:

```
FuzzLean found 1 bug: 

2024-07-18 10:00:07.393: Sending: 'GET /app/log/public/historical/range?from_seqno=1&to_seqno=0963012895303452&id=1 HTTP/1.1\r\nAccept: application/json\r\nHost: 20.14.79.29\r\nContent-Length: 0\r\nUser-Agent: restler/9.2.4\r\nx-restler-sequence-id: 902dc766-095a-4b18-a5f5-09356ac1b102\r\n\r\n'

2024-07-18 10:00:07.477: Received: 'HTTP/1.1 500 INTERNAL_SERVER_ERROR\r\ncontent-length: 75\r\ncontent-type: application/json\r\n\r\n{"error":{"code":"InternalError","message":"Unable to retrieve Tx status"}}'
```

achamayou:
@lemmy as @eddyashton pointed out (and resolved in #6401), I was confused, and despite superficial similarity this is actually a different issue with the logging app itself.

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [5.0.2]
+
+[5.0.2]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.2
+
+### Bug Fixes
+
+- The `/tx` endpoint returns more accurate error messages for invalid transactions ids (#6359).
```
```suggestion
- The `/tx` endpoint returns more accurate error messages for incorrectly formed transactions ids (#6359).
```

PR #6362, opened 2024-07-15
Partially rollback public headers functions & cleanup nonstd.cpp

maxtropets:
Can't link SCITT against CCF because `to_lower` and `to_upper` are used in other public headers (`unit_strings.h`, for instance). 

This breaks linkage of CCF against SCIIT and potentially a lot of other projects, because nobody would expect the necessity of linking to `libccf.virtual.a` to make it work.

Explicitly linking against it, however, also doesn't work, because it carries `INSIDE_ENCLAVE` target properties, which also breaks SCITT build.


src/ds/nonstd.cpp
```diff
@@ -14,21 +12,6 @@
...
-
-  void to_lower(std::string& s)
-  {
-    std::transform(s.begin(), s.end(), s.begin(), [](unsigned char c) {
-      return std::tolower(c);
-    });
-  }
-
   // Implementations for ds/nonstd.h
   std::string camel_case(
```
achamayou, line 16: I suggest moving this to endpoints/something.cpp, either endpoint_registry.cpp or its own file, but I don't think it should be under nonstd:: or under nonstd/.

maxtropets: I don't think it's a good place to place this either, but TBH it's been already too much jerking around this, so let's bury it under `endpoints::` in `endpoint_registry.coo` and move on.

maxtropets: Done

include/ccf/endpoint_registry.h
```diff
@@ -105,6 +105,15 @@ namespace ccf::endpoints
     return true;
   }
 
+  std::string camel_case(
```
achamayou, line 108: I don't want this exported/part of the public API, it doesn't need to be visible outside of endpoint_registry.cpp.

maxtropets: It does for tests

maxtropets: That's why I kept it in `src/ds/nonstd.h` header and separatly in `src/ds/nonstd.cpp` as impl at the first place, because that is a private header

eddyashton: I vote for dropping the unit tests, and moving this to `.cpp` only. It's a single-purpose util function for now, we've tested it works as-expected _currently_, and we don't want to risk any additions to the public API at this stage.

achamayou: The current change is the best of both worlds: still tested but not exported.

PR #6363, opened 2024-07-15
Merge jobs in C-ACI CI

achamayou:
It is unclear why the jobs were split, perhaps they used to be run in different pools. Other than verbosity, the problem with the current layout is that the cancelation of a job half-way through, for example due to a more recent commit, can result in the ACI container failing to be shut down.

The goal remains to move the contents of this job to a hosted pool, when the necessary features become available, but for the time being, this change aims to reduce the amount of containers that leak out of this pipeline.


eddyashton (approved):

achamayou:
Confirmed manually on [this run](https://dev.azure.com/MSRC-CCF/CCF/_build/results?buildId=87015&view=logs&j=b6a3fffa-f95e-580b-9bcb-4ae7d4681578&t=b6a3fffa-f95e-580b-9bcb-4ae7d4681578) that a job cancelation happening during CTest will not result in a leaked container.

achamayou:
Test failure unrelated to PR changed, logged as https://github.com/microsoft/CCF/issues/6372

PR #6364, opened 2024-07-15
Fix for connection cleanup

eddyashton:
Combination of recent connection cleanup changes were dangerous, as noticed by the ASAN build. We were modifying a collection while iterating over it:

```
for (auto& [id, idle_time]: idle_times)
{
  ...
  stop(id);
  ...
}

bool stop(ConnectionID id)
{
  ...
  idle_times.erase(id);
  ...
}
```

We could potentially drop this erasure entirely, I think it's a _safe_ but _unnecessary_ step. But to be a little safer, less likely to spam logs in an edge case, I've retained it - callers of `stop(id)` now call `idle_times.erase(id)` themselves.

### DIFF
--- a/src/host/rpc_connections.h
+++ b/src/host/rpc_connections.h
@@ -345,7 +345,6 @@ namespace asynchost
       // Invalidating the TCP socket will result in the handle being closed. No
       // more messages will be read from or written to the TCP socket.
       sockets[id] = nullptr;
-      idle_times.erase(id);
 
       RINGBUFFER_WRITE_MESSAGE(::tcp::tcp_close, to_enclave, id);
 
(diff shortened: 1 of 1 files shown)


maxtropets:
> We were modifying a collection while iterating over it:

Is it possible we still modify it simultaneously from both `on_timer()` and message dispatch?

maxtropets (approved):

achamayou (approved):

eddyashton:
> > We were modifying a collection while iterating over it:
> 
> Is it possible we still modify it simultaneously from both `on_timer()` and message dispatch?

No - these both happen on the same thread, as part of libuv's dispatch loop.

PR #6369, opened 2024-07-16
Move TLA+ from ADO to GHA

maxtropets:
Consensus TLA+ jobs on their way to GHA

- [x] Check upload_artifacts naming 
- [x] Check upload_artifacts file patterns
- [x] Removed `set -ex`, check if that's OK (UPD. It seems [GHA uses it by default](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference)
- [x] Remove unnecessary azure stuff
- [x] Correct triggers (daily/weekly/PR/etc)
- [x] Run trace validation PR-based? UPD: No.


.github/workflows/consensus-tla.yml
```diff
@@ -0,0 +1,144 @@
...
+      - name: Coverage
+        run: |
+          cd tla/
+          python3 -mvenv env
+          source env/bin/activate
+          pip install -r requirements.txt
+          ls -tr MCccfraftAtomicReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftAtomicReconfig.html
+          rm MCccfraftAtomicReconfig_coverage.json
+          ls -tr MCccfraftWithReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftWithReconfig.html
+          rm MCccfraftWithReconfig_coverage.json
```
lemmy, line 57: I'd suggest archiving`*_coverage.json` as workflow artifact instead of deleting the files.

maxtropets: I see you added that `rm` in #6092 

Are you sure we want to get them back and `upload_artifact` them?

lemmy: I don't remember adding those lines.  @achamayou, was it you who initially added that block?

maxtropets: Ok, I just checked the blame diff, you didn't add `rm`, you were doing renaming only

https://github.com/microsoft/CCF/pull/6092/commits/43ae605c633c43267c42c784eb08ef6f743a40ed

maxtropets:
One thing to check, should we run trace validation in CI?

I see it's fast, and IIR we do consistency trace validation in CI, as @achamayou explained to me (I'm not sure where exactly though).

cc @lemmy @eddyashton

.github/workflows/consensus-tla.yml
achamayou: The coverage files were very large, took a long time to upload, and also caused our other more useful artefacts like node logs to be deleted sooner as we run out of quota. It's possible the quota has changed now that we're on GH, but that needs to be checked first.

achamayou:
@maxtropets only PRs that touch the tla/* files or src/consensus/* files. There are far too many pipelines at the moment, we will have a single Verification pipeline with that combined trigger.

.github/workflows/consensus-tla.yml
lemmy: > The coverage files were very large, took a long time to upload, and also caused our other more useful artefacts like node logs to be deleted sooner as we run out of quota. It's possible the quota has changed now that we're on GH, but that needs to be checked first.

I would not have suggested archiving the coverage results if I had known this.  Perhaps we should add a comment explaining why the coverage data is not archived.

achamayou:
@maxtropets from the doc, it looks like -e and -o pipefail are defaults for bash (great!) but maybe not -x?

.github/workflows/tlaplus.yml
```diff
@@ -4,8 +4,9 @@ on:
   schedule:
     - cron: "0 0 * * 0"
   pull_request:
-    paths:
-      - "tla/**"
+    # paths:
+    #   - "tla/**"
+    #   - "src/consensus/**"
```
achamayou, line 9:
```suggestion
    paths:
      - "tla/**"
      - "src/consensus/**"
```

.github/workflows/tlaplus.yml
```diff
@@ -114,3 +115,129 @@ jobs:
           path: |
             tla/consistency/*_TTrace_*.tla
             tla/*.json
+
+  model-checking-consensus:
+    name: Model Checking - Consensus
+    runs-on: [self-hosted, 1ES.Pool=gha-virtual-ccf-sub]
+    container:
+      image: ghcr.io/microsoft/ccf/ci/default:build-26-06-2024
```
achamayou, line 123: Would be best to move to a workflow variable: https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow

That way updating the image is a one line change.

.github/workflows/consensus-tla.yml
achamayou: It was reasonable to try - we did! If we had a CI system where upload is super faster and storage unlimited, we would store everything. I seem to remember that even just enabling coverage in TLC made the model checking much slower though: https://github.com/microsoft/CCF/pull/5702

.github/workflows/tlaplus.yml
```diff
@@ -114,3 +115,129 @@ jobs:
...
+
+      - name: Coverage
+        run: |
+          cd tla/
+          python3 -mvenv env
+          source env/bin/activate
+          pip install -r requirements.txt
+          ls -tr MCccfraftAtomicReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftAtomicReconfig.html
+          rm MCccfraftAtomicReconfig_coverage.json
+          ls -tr MCccfraftWithReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftWithReconfig.html
```
achamayou, line 155: Unless I am missing something, these html files are not uploaded?

maxtropets: Fixed

.github/workflows/tlaplus.yml
maxtropets: It seems it doesn't work in
```container:
  image:
```
block, it's not visible in there.

achamayou: That is quite sad.

maxtropets:
> @maxtropets from the doc, it looks like -e and -o pipefail are defaults for bash (great!) but maybe not -x?

Returned set -x back to where we have scripts executions

achamayou (approved):

.github/workflows/tlaplus.yml
```diff
@@ -114,3 +116,137 @@ jobs:
...
+      image: ghcr.io/microsoft/ccf/ci/default:build-26-06-2024
+
+    steps:
+      - uses: actions/checkout@v4
+      - run: |
+          sudo apt update
+          sudo apt install -y default-jre
+          python3 ./tla/install_deps.py
+
+      - name: MCccfraft.cfg
```
achamayou, line 133: The model checking job is 1h47, I think it would be preferable to split it in three jobs so it only takes as long as the slowest step (WithReconfig). We should also try to disable the coverage, that lead to massive speedups in the past.

lemmy: Yes, coverage collection incurs a significant slowdown.

.github/workflows/tlaplus.yml
```diff
@@ -140,22 +140,7 @@ jobs:
...
-          source env/bin/activate
-          pip install -r requirements.txt
-          ls -tr MCccfraftAtomicReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftAtomicReconfig.html
-          rm MCccfraftAtomicReconfig_coverage.json
-          ls -tr MCccfraftWithReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftWithReconfig.html
-          rm MCccfraftWithReconfig_coverage.json
-          deactivate
-          rm -rf env
-        shell: bash
+          JVM_OPTIONS=-Dtlc2.TLCGlobals.coverage=0 ./tlc.sh -workers auto -config consensus/MCccfraftAtomicReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftAtomicReconfig.trace.tla -dumpTrace json MCccfraftAtomicReconfig.json
```
achamayou, line 143:
```suggestion
          ./tlc.sh -workers auto -config consensus/MCccfraftAtomicReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftAtomicReconfig.trace.tla -dumpTrace json MCccfraftAtomicReconfig.json
```

.github/workflows/tlaplus.yml
```diff
@@ -184,22 +168,7 @@ jobs:
...
-          source env/bin/activate
-          pip install -r requirements.txt
-          ls -tr MCccfraftAtomicReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftAtomicReconfig.html
-          rm MCccfraftAtomicReconfig_coverage.json
-          ls -tr MCccfraftWithReconfig_coverage.json | xargs cat | python3 actions.py MCccfraftWithReconfig.html
-          rm MCcrcfraftWithReconfig_coverage.json
-          deactivate
-          rm -rf env
-        shell: bash
+          JVM_OPTIONS=-Dtlc2.TLCGlobals.coverage=0 ./tlc.sh -workers auto -config consensus/MCccfraftWithReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftWithReconfig.trace.tla -dumpTrace json MCccfraftWithReconfig.json
```
achamayou, line 171:
```suggestion
          ./tlc.sh -workers auto -config consensus/MCccfraftWithReconfig.cfg consensus/MCccfraft.tla -dumpTrace tla MCccfraftWithReconfig.trace.tla -dumpTrace json MCccfraftWithReconfig.json
```

achamayou:
@maxtropets two more suggestions:

- rename pipeline to Verification
- document it under https://github.com/microsoft/CCF/blob/main/.github/workflows/README.md

.github/workflows/tlaplus.yml
```diff
@@ -1,11 +1,13 @@
...
+name: "Verification"
 
 on:
   schedule:
     - cron: "0 0 * * 0"
   pull_request:
     paths:
       - "tla/**"
+      - "src/consensus/**"
+    # Will uncomment when testing finished
```
maxtropets, line 10:
```suggestion
```

PR #6376, opened 2024-07-19
Upgrade ccf python package to pyproject.toml

achamayou:
Follow up to the workaround applied in #6374. More adjustments may be needed, but this seems to work so far.


python/pyproject.toml
```diff
@@ -0,0 +1,49 @@
+[build-system]
+requires = ["setuptools>=71.0", "setuptools_scm>=8"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "ccf"
+dynamic = ["version"]
```
eddyashton, line 7: Can we add a comment explaining some of the magic behind this version selection, and/or pointing to the docs? I think the rest is reasonably self-explanatory.

eddyashton (approved):

python/pyproject.toml
achamayou: yes, done

PR #6386, opened 2024-07-23
Daily -> GHA

maxtropets:


.github/workflows/long-test.yml
```diff
@@ -0,0 +1,46 @@
+name: CI
```
achamayou, line 1:
```suggestion
name: LTCI
```

.github/workflows/long-test.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
achamayou, line 5:
```suggestion
    - cron: "0 0 * * 0-4"
```

Probably. Run on Sunday night, so we have a status Monday morning, and the rest of the week. Friday and Saturday night runs are not useful.

.github/workflows/long-test.yml
```diff
@@ -0,0 +1,46 @@
...
+  pull_request:  # TODO remove this before merge
```
achamayou, line 6: Add a workflow dispatch, and a label trigger.

.github/workflows/long-test.yml
```diff
@@ -0,0 +1,46 @@
...
+
+permissions:
```
achamayou, line 8: I don't think we should need any of that for this pipeline.

.github/workflows/long-test.yml
maxtropets: Isn't 1-5 a correct schedule to run Mon-Fri?

.github/workflows/long-test.yml
maxtropets: > and a label trigger

Do we have an example already? I don't see workflow-specific label triggers in the doc, only job-specific.

achamayou: No, but I think we use https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

Annoyingly I think we need extra if conditions to check the value of the label, which means a bit of running on every PR effectively, and a situation where it's harder to tell if you really have triggered the LTCI or not. Ugh.

.github/workflows/long-test.yml
achamayou: Sure, that's fine too. I guess the expectation in this case is that on the Monday, you have the build from the previous Friday night. My proposal (Sunday to Thursday) is a little more conservative: if stuff gets checked in during the weekend, up to Sunday midnight, then you get it in the build you see on Monday morning.

maxtropets: Are we both on the same page assuming that 0 0 * * 1-5 means
- 12:00AM Mon (nightly after Sunday, fresh on Monday morning)
- 12:00AM Tue (nightly after Monday, fresh on Tuesday morning)
- ...
- 12:00AM Fri (nightly after Thursday, fresh on Friday morning)

?

.github/workflows/long-test.yml
maxtropets: This is what I mean, we have to copy-paste the condition into each job, and will be green if all checks disabled, which is confusing.

Is it better to manually start it from Actions if needed instead?

.github/workflows/long-test.yml
achamayou: @maxtropets no, I was wrong, and clearly you are right. 1-5.

.github/workflows/long-test.yml
achamayou: > Is it better to manually start it from Actions if needed instead?

Not as far as I can tell. The alternative is the canary file I think. Maybe we can still do the condition, but set a check status that does not lead to green?

achamayou: Manual start via workflow_dispatch only works on branches, not forks. Maybe that is an acceptable restriction though.

.github/workflows/long-test.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: "Install deps"
+        run: |
+          sudo apt-get -y update
+          sudo apt install ansible -y
+          cd getting_started/setup_vm
+          ansible-playbook ccf-extended-testing.yml
+
+      - name: "Build"
+        run: |
+          git config --global --add safe.directory /__w/CCF/CCF
```
achamayou, line 46:
```suggestion
          git config --global --add safe.directory "$GITHUB_WORKSPACE"
```

achamayou (approved):

.github/workflows/long-test.yml
```diff
@@ -0,0 +1,81 @@
...
+          mkdir build
+          cd build
+          cmake -GNinja -DCOMPILE_TARGET=virtual -DCMAKE_BUILD_TYPE=Debug -DLONG_TESTS=ON -DLVI_MITIGATIONS=OFF -DSAN=ON ..
+          ninja
+
+      - name: "Test"
+        run: |
+          set +x
+          cd build
+          ./tests.sh -VV --timeout 1600 --no-compress-output -LE "benchmark|perf"
```
achamayou, line 56:
```suggestion
          ./tests.sh --output-on-failure --timeout 1600 -LE "benchmark|perf"
```

-VV is too verbose for GHA to cope, and no compress output was for xml rendering we do not use on GHA.

.github/workflows/long-test.yml
```diff
@@ -0,0 +1,103 @@
+name: Long Test
+
+on:
+  schedule:
+    - cron: "0 0 * * 1-5"
+  pull_request:  # TODO remove this before merge
```
maxtropets, line 6: Don't forget to remove

.github/workflows/long-test.yml
```diff
@@ -0,0 +1,102 @@
...
+          fetch-depth: 0
+      - name: "Run scan"
+        run: |
+          set -x
+          mkdir build
+          cd build
+          ../scripts/scan-build.sh
+
+  long-asan:
+    name: Long ASAN
```
achamayou, line 27: Could probably just be ASAN, since the whole workflow is "Long" and there isn't a short ASAN, but this is fine too.

achamayou (approved):

PR #6387, opened 2024-07-23
Login to `ccfmsrc` ACR resource with managed identity rather than token

eddyashton:


achamayou (approved):

eddyashton:
Odd timeout on the first run, but fine on the second. Assuming this is safe to merge.

PR #6401, opened 2024-07-26
Correct error response for `/app/log/public/historical/range` when given a future `to_seqno`

eddyashton:
Resolving a `500 INTERNAL_SERVER_ERROR` found by fuzzing:

https://github.com/microsoft/CCF/pull/6359#issuecomment-2252929353

PR #6402, opened 2024-07-29
UPDATED Correct error response for /app/log/public/historical/range when given a future to_seqno

maxtropets:
Continuing #6401 here, as soon as I don't have write access to @eddyashton's branch

### DIFF
--- a/tests/e2e_logging.py
+++ b/tests/e2e_logging.py
@@ -918,6 +918,40 @@ def test_historical_query_range(network, args):
 
         infra.commit.wait_for_commit(c, seqno=last_seqno, view=view, timeout=3)
 
+        LOG.info("Checking error responses")
+        # Reversed range is illegal
+        r = c.get(
+            f"/app/log/public/historical/range?from_seqno={last_seqno}&to_seqno=1&id={id_a}"
+        )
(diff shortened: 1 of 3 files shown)


tests/e2e_logging.py
```diff
@@ -947,6 +947,40 @@ def id_for(i):
...
+        # - Try a very invalid seqno
+        r = c.get(
+            f"/app/log/public/historical/range?to_seqno={last_valid_seqno*2}&id={id_a}"
+        )
+        assert r.status_code == http.HTTPStatus.BAD_REQUEST
+        assert r.body.json()["error"]["code"] == "InvalidInput"
+
+        # - Try the first invalid seqno
+        r = c.get(
+            f"/app/log/public/historical/range?to_seqno={last_valid_seqno+1}&id={id_a}"
```
achamayou, line 978: @maxtropets this is risky, if a new transaction takes place after the last_seqno on which we wait for commit at `:948`, for example a snapshot or a jwt refresh, you could easily get a 202 instead, which seems to be happening sometimes (#6417).

To be made robust, the test needs to be either more controlled to definitely rule this out (probably means a split env etc), or the invalid seqno needs to be a little more conservative (+10?).

maxtropets: #6418

PR #6415, opened 2024-07-31
Replace RSASSA-PKCS1-v1_5 with RSA-PSS in crypto API

### LINKED ISSUE #6405: Crypto APIs should replace PKCS1v15 with PSS.
CCF still supports PKCS1v15 for RSA signing and verification in the C++/JS/TS crypto providers, but should replace it with [PSS](https://www.rfc-editor.org/rfc/rfc3447#section-8.1).

- [TypeScript Algo Name definition](https://github.com/microsoft/CCF/blob/6fac042a834857f60cfbf340994e763db13a6017/js/ccf-app/src/global.ts#L234)
- [TypeScript sign()](https://github.com/microsoft/CCF/blob/6fac042a834857f60cfbf340994e763db13a6017/js/ccf-app/src/global.ts#L343)
- [TypeScript verify()](https://github.com/microsoft/CCF/blob/6fac042a834857f60cfbf340994e763db13a6017/js/ccf-app/src/global.ts#L359)

maxtropets:
Closes #6405

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 [5.0.2]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.2
 
+### Developer API
+
+#### C++
+
+- `RSAKeyPair::sign` and `RSAKeyPair::verify` now use `RSA-PSS` instead of `RSASSA-PKCS1-v1_5`.
+  - Users can specify `salt_length` (defaulted to `0`).
+
+#### TypeScript/JavaScript
+
+- `/app/sign` and `/app/verify` no longer support `RSASSA-PKCS1-v1_5`, instead `RSA-PSS` has been added.
+  - `SigningAlgorithm` has been extended with optional `saltLength`, defaulted to `0` if not passed.
+
 ### Bug Fixes
 
 - The `/tx` endpoint returns more accurate error messages for incorrectly formed transactions ids (#6359).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 #### TypeScript/JavaScript
 
-- `/app/sign` and `/app/verify` no longer support `RSASSA-PKCS1-v1_5`, instead `RSA-PSS` has been added.
+- `ccfapp.crypto.sign()` and `ccfapp.crypto.verifySignature()` no longer support `RSASSA-PKCS1-v1_5`, instead `RSA-PSS` has been added.
   - `SigningAlgorithm` has been extended with optional `saltLength`, defaulted to `0` if not passed.
 
 ### Bug Fixes
--- a/js/ccf-app/src/global.ts
+++ b/js/ccf-app/src/global.ts
@@ -231,7 +231,7 @@ export interface CryptoKeyPair {
   publicKey: string;
 }
 
-export type AlgorithmName = "RSASSA-PKCS1-v1_5" | "ECDSA" | "EdDSA" | "HMAC";
+export type AlgorithmName = "RSA-PSS" | "ECDSA" | "EdDSA" | "HMAC";
 
 export type DigestAlgorithm = "SHA-256" | "SHA-384" | "SHA-512";
 
@@ -239,9 +239,14 @@ export interface SigningAlgorithm {
   name: AlgorithmName;
 
   /**
-   * Digest algorithm. It's necessary for "RSASSA-PKCS1-v1_5", "ECDSA", and "HMAC"
+   * Digest algorithm. It's necessary for "RSA-PSS", "ECDSA", and "HMAC"
    */
   hash?: DigestAlgorithm;
+
+  /**
+   * Salt length, necessary for "RSA-PSS".
+   */
+  saltLength?: number;
 }
 
 /**
--- a/src/crypto/openssl/rsa_public_key.cpp
+++ b/src/crypto/openssl/rsa_public_key.cpp
@@ -195,11 +195,14 @@ namespace ccf::crypto
     size_t contents_size,
     const uint8_t* signature,
     size_t signature_size,
-    MDType md_type)
+    MDType md_type,
+    size_t salt_length)
   {
     auto hash = OpenSSLHashProvider().Hash(contents, contents_size, md_type);
     Unique_EVP_PKEY_CTX pctx(key);
     CHECK1(EVP_PKEY_verify_init(pctx));
+    CHECK1(EVP_PKEY_CTX_set_rsa_padding(pctx, RSA_PKCS1_PSS_PADDING));
+    CHECK1(EVP_PKEY_CTX_set_rsa_pss_saltlen(pctx, salt_length));
     CHECK1(EVP_PKEY_CTX_set_signature_md(pctx, get_md_type(md_type)));
     return EVP_PKEY_verify(
              pctx, signature, signature_size, hash.data(), hash.size()) == 1;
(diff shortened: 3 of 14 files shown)


src/crypto/openssl/rsa_public_key.cpp
```diff
@@ -200,6 +200,7 @@ namespace ccf::crypto
     auto hash = OpenSSLHashProvider().Hash(contents, contents_size, md_type);
     Unique_EVP_PKEY_CTX pctx(key);
     CHECK1(EVP_PKEY_verify_init(pctx));
+    CHECK1(EVP_PKEY_CTX_set_rsa_padding(pctx, RSA_PKCS1_PSS_PADDING));
```
maxtropets, line 203: @achamayou I don't know why, but this's not used in `/app/verifySignature`, however `::sign(` is used in `/app/sign`.

Instead, `bool PublicKey_OpenSSL::verify_hash(`, which is counterintuitive to me, it would've made much more sense to have `sign` `<->` `verify` and `sigh_hash` `<->` `verify_hash`

I though we `sign`/`verify` and `sign_hash

src/crypto/openssl/key_pair.cpp
```diff
@@ -202,6 +202,7 @@ namespace ccf::crypto
   {
     Unique_EVP_PKEY_CTX pctx(key);
     OpenSSL::CHECK1(EVP_PKEY_sign_init(pctx));
+    OpenSSL::CHECK1(EVP_PKEY_CTX_set_rsa_padding(pctx, RSA_PKCS1_PSS_PADDING));
```
maxtropets, line 205: This fails `crypto_test` ...

maxtropets:
e2e logging tests broken - multi_auth doesn't work. Needs investigating..

js/ccf-app/src/global.ts
```diff
@@ -231,15 +231,15 @@ export interface CryptoKeyPair {
   publicKey: string;
 }
```
maxtropets, line 232: e2e logging tests broken - multi_auth doesn't work. Needs investigating..

src/crypto/openssl/key_pair.cpp
achamayou: Do we know why?

src/crypto/openssl/rsa_public_key.cpp
achamayou: Agreed, do we know where verify() is used? verify_hash() should probably be renamed to verify_digest().

src/js/extensions/ccf/crypto.cpp
```diff
@@ -1010,12 +1010,12 @@ namespace ccf::js::extensions
             ctx, "Unsupported hash algorithm, supported: SHA-256");
         }
 
-        if (algo_name != "RSASSA-PKCS1-v1_5" && algo_name != "ECDSA")
+        if (algo_name != "RSASSA-PSS" && algo_name != "ECDSA")
```
achamayou, line 1013: That doesn't actually seem to allow EdDSA, am I missing something?

src/crypto/openssl/rsa_public_key.cpp
maxtropets: `do we know where verify() is used`
Ran all e2e tests with extra logs, didn't trigger `verify(` call

src/js/extensions/ccf/crypto.cpp
maxtropets: I think it's handled above

```
      // Handle algorithms that don't use algo_hash here
      if (*algo_name_str == "EdDSA")
      {
        try
        {
          return JS_NewBool(
            ctx,
            verify_eddsa_signature(
              data, data_size, signature, signature_size, *key_str));
        }
        catch (const std::exception& ex)
        {
          return JS_ThrowRangeError(
            ctx, "Failed to verify EdDSA signature: %s", ex.what());
        }
      }

```

src/crypto/openssl/key_pair.cpp
maxtropets:
```
11: ../src/crypto/test/crypto.cpp:111:
11: TEST CASE:  Sign, fail to verify with bad signature
11:
11: ../src/crypto/test/crypto.cpp:111: ERROR: test case THREW exception: OpenSSL error: error:0D07803A:asn1 encoding routines:asn1_item_embed_d2i:nested asn1 error
11:   logged: With curve: 0x5653e7b671e9
```

In a nutshell, the second option fails signing on setting padding
```
static constexpr CurveID supported_curves[] = {
  CurveID::SECP384R1,  // OK
  CurveID::SECP256R1,  // FAILS!
  CurveID::SECP256K1}; // Skipped
```

achamayou: That looks like EC signing, not RSA?

maxtropets: Yeap, I spotted that after learning what key types are. Now it passes, 'cause I only set padding to RSA keys

js/ccf-app/src/global.ts
maxtropets: Misusing padding in EC/RSA, confused key types.

src/crypto/openssl/rsa_public_key.cpp
achamayou: It looks like it's part of the public C++ API, so we ought to have a unit test for it.

maxtropets: UPD. Now it will be used in `js_sign`, and is tested in js e2e logging tests.

And covered with a unit test too.

src/js/extensions/ccf/crypto.cpp
```diff
@@ -884,10 +884,15 @@ namespace ccf::js::extensions
...
         }
-        else if (algo_name == "RSASSA-PKCS1-v1_5")
+        else if (algo_name == "RSA-PSS")
         {
           auto key_pair = ccf::crypto::make_rsa_key_pair(key);
-          auto sig = key_pair->sign(contents, mdtype);
+
+          // auto salt_length = jsctx.get_property(algorithm, "saltLength");
+          auto sig = key_pair->sign(
+            contents, mdtype, 0 // TO DO convert salt to size_t
```
maxtropets, line 893: Pass the one from JS through

maxtropets: DOne.

maxtropets:
Is now back from "Draft". Can be reviewed.

CHANGELOG.md
```diff
@@ -9,6 +9,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+### Developer API
+
+#### C++
+
+- `RSAKeyPair::sign` and `RSAKeyPair::verify` now use `RSA-PSS` instead of `RSASSA-PKCS1-v1_5`.
+  - Users can specify `salt_length` (defaulted to `0`).
+
+#### TypeScript/JavaScript
+
+- `/app/sign` and `/app/verify` no longer support `RSASSA-PKCS1-v1_5`, instead `RSA-PSS` has been added.
```
achamayou, line 21:
```suggestion
- `ccfapp.crypto.sign()` and `ccfapp.crypto.verifySignature()` no longer support `RSASSA-PKCS1-v1_5`, instead `RSA-PSS` has been added.
```

The REST endpoints only exist in a test application.

src/crypto/test/crypto.cpp
```diff
@@ -931,4 +931,36 @@ TEST_CASE("Incremental hash")
...
+      contents.data(),
+      contents.size(),
+      sig.data(),
+      sig.size(),
+      mdtype,
+      salt_length));
+  }
+
+  {
+    constexpr size_t sign_salt_length = 0, verify_salt_legth = 32;
```
achamayou, line 956: I propose adding one where we sign with 32 and verify with 32 to check that works with non-zero salt.

maxtropets: Done. Also checked in e2e.

PR #6422, opened 2024-08-04
Perf test under CTest conf to exclude from default test set

### LINKED ISSUE #6421: Perf tests failing in debug builds
```
cd build
cmake -GNinja -DCOMPILE_TARGET=virtual -DCMAKE_BUILD_TYPE=Debug ..
ninja
./tests.sh -VV -R "pi_basic|historical_query_perf_test"
```

Why? It's quite confusing to see those failing when developing with debug build, took me some time figuring that out when working on another feature.

Any of the following I'd consier satisfying
* we exclude them from CMake in debug
* we fix them to pass in debug

maxtropets:
Attempt to close #6421.

Open to discussion.


CMakeLists.txt
```diff
@@ -1511,59 +1511,62 @@ if(BUILD_TESTS)
...
-    )
-
-    add_piccolo_test(
-      NAME pi_basic_js
-      PYTHON_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/tests/infra/basicperf.py
-      CLIENT_BIN ./submit PERF_LABEL "Basic JS"
-      ADDITIONAL_ARGS --js-app-bundle ${CMAKE_SOURCE_DIR}/samples/apps/basic/js
-                      --client-def "1,write,100000,primary"
-    )
+    if(CMAKE_BUILD_TYPE STREQUAL "Release")
```
achamayou, line 1514: Probably not what we want (our default is not Release, it's RelWithDebInfo anyway), proposed an alternative solution under https://github.com/microsoft/CCF/issues/6421#issuecomment-2268456893.

achamayou: I don't understand the rationale. I think we should remove the -LE, just like we shouldn't have it for the partitions tests. It's weird to have code that does nothing just in case one specific type of code change happens.

cmake/common.cmake
```diff
@@ -307,6 +308,7 @@ function(add_piccolo_test)
       --label ${TEST_NAME} --perf-label ${PARSED_ARGS_PERF_LABEL}
       --snapshot-tx-interval 10000 ${PARSED_ARGS_ADDITIONAL_ARGS} -e
       ${ENCLAVE_TYPE} -t ${ENCLAVE_PLATFORM} ${NODES}
+    CONFIGURATIONS ${PARSED_ARGS_CONFIGURATIONS}
```
achamayou, line 311: I think this could set e2e_perf systematically, and not require the caller to do it, nor expose CONFIGURATIONS. Are there any performance tests that we do not want to group in that configuration?

achamayou (approved):

CMakeLists.txt
maxtropets: `-LE perf` means "I wanna run all but perf"
`-L ...|perf|... -C e2e_perf` means "I want to run e2e perf tests in the selected test suite.

The `e2e_perf` configuration can be treated as a subset of perf tests in this case.

Is it still weird if we read it like that?

achamayou: It's not weird but afaict e2e_perf _is_ perf?

maxtropets: Convincing. Done

cmake/common.cmake
maxtropets: Done

CMakeLists.txt
```diff
@@ -1563,6 +1563,7 @@ if(BUILD_TESTS)
       NAME historical_query_perf_test
       PYTHON_SCRIPT ${CMAKE_SOURCE_DIR}/tests/historical_query_perf.py
       LABEL perf PERF_LABEL "Historical Queries"
+      CONFIGURATIONS e2e_perf
```
achamayou, line 1566:
```suggestion
      CONFIGURATIONS perf
```

achamayou (approved):

Issue #6413, opened 2024-07-31
SGX removal from main


achamayou:
As per [5.0 release notes](https://github.com/microsoft/CCF/releases/tag/ccf-5.0.0), `release/5.x` is the last LTS with SGX support. 6.0 (`main` at the moment) will target confidential containers exclusively, and switch from Ubuntu to [Azure Linux 3.0](https://github.com/microsoft/azurelinux).

This issue tracks the various steps involved in the removal of SGX from `main`, the list is a work-in-progress:

- [x] Remove SGX attestation verifcation/`oehostverify` dependency from virtual and snp builds (#6423)
- [x] Remove SGX target (but keep build-* image creation for SGX to support 5.x/4.x) (#6426)
- [x] Remove CCF_DISABLE_VERBOSE_LOGGING (#6426)
- [x] Remove SGX from release playbooks, but keep it in ccf-dev for build-* images (#6447)
- [x] Migrate code upgrade tests from SGX to Virtual
- [ ] Remove INSIDE_ENCLAVE
  - Done partially, but still temporarily useful for:
  - Differentiating enclave/host logging is useful until the ringbuffer is removed
  - Detecting that some debug features should not be enabled
- [ ] Doc/cfg removal pass #6623 and #6731

achamayou:
Goal: Build, tests and release artefacts are completely free from SGX dependencies.

This does not include undoing or revisiting the architectural impact of SGX (e.g. the ringbuffer), which will be a separate issue/collection of PRs.

achamayou:
Largely done, remaining bits of mopping up tracked in separate issues.

PR #6426, opened 2024-08-06
Initial removal of SGX targets

maxtropets:
A step towards #6413

- [x] `we might as well remove the dependency on Open Enclave in the same PR.`


achamayou (commented):
LGTM, but we might as well remove the dependency on Open Enclave in the same PR.

Note that Open Enclave needs to stay in the ansible playbook used to build CI images, and the relevant bits of the CI container building pipeline need to stay too, because we use the same build images for 4.x, 5.x and main for now and the near future.

achamayou (approved):
LGTM, but we might as well remove the dependency on Open Enclave in the same PR.

Note that Open Enclave needs to stay in the ansible playbook used to build CI images, and the relevant bits of the CI container building pipeline need to stay too, because we use the same build images for 4.x, 5.x and main for now and the near future.

PR #6428, opened 2024-08-09
Let through detailed error messages in JWT auth

### LINKED ISSUE #6427: Error reporting in JwtAuthnPolicy::authenticate is incorrect
In the metadata check loop, the nbf, exp and issuer errors are correctly set, but then overwritten on line 211 with a generic error.

Same thing with extract_token() on line 125, with a generic error applied on 129.

achamayou:
Closes #6427


maxtropets:
Closed in favour of #6435

PR #6435, opened 2024-08-12
Improve JWT auth error msg

maxtropets:
Closes https://github.com/microsoft/CCF/issues/6427

- [x] Verify error messages in the code

### DIFF
--- a/src/endpoints/authentication/jwt_auth.cpp
+++ b/src/endpoints/authentication/jwt_auth.cpp
@@ -120,7 +120,6 @@ namespace ccf
     std::string& error_reason)
   {
     const auto& headers = ctx->get_request_headers();
-    error_reason = "Invalid JWT token";
 
     const auto token_opt =
       ::http::JwtVerifier::extract_token(headers, error_reason);
--- a/src/endpoints/authentication/jwt_auth.cpp
+++ b/src/endpoints/authentication/jwt_auth.cpp
@@ -120,13 +120,12 @@ namespace ccf
     std::string& error_reason)
   {
     const auto& headers = ctx->get_request_headers();
+    error_reason = "Invalid JWT token";
 
     const auto token_opt =
       ::http::JwtVerifier::extract_token(headers, error_reason);
-
     if (!token_opt)
     {
-      error_reason = "Invalid JWT token";
       return nullptr;
     }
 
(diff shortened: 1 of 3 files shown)


src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -153,13 +152,14 @@ namespace ccf
...
 
-    if (!token_keys)
+    if (!token_keys || token_keys->empty())
     {
       error_reason =
         fmt::format("JWT signing key not found for kid {}", key_id);
       return nullptr;
     }
 
+    error_reason = "Cant authenticate JWT token: no matching public key found";
```
achamayou, line 162:
```suggestion
    error_reason = "Cannot authenticate JWT token: no matching public key found";
```

achamayou: This is not needed, instead line 167 should set error_reason to something like "Signature verification failed".

maxtropets: Done

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -165,6 +164,8 @@ namespace ccf
       auto verifier = verifiers->get_verifier(metadata.cert);
       if (!::http::JwtVerifier::validate_token_signature(token, verifier))
       {
+        error_reason =
+          "Cannot authenticate JWT token: signature verification failed";
```
achamayou, line 168:
```suggestion
          "Signature verification failed";
```

achamayou (approved):

src/endpoints/authentication/jwt_auth.cpp
```diff
@@ -120,13 +120,12 @@ namespace ccf
     std::string& error_reason)
   {
     const auto& headers = ctx->get_request_headers();
+    error_reason = "Invalid JWT token";
```
achamayou, line 123:
```suggestion
    error_reason = "Invalid JWT token";
```

Let's not set the error reason before we know there is an error, this may surprise the caller unncessarily.

tests/js-custom-authorization/custom_authorization.py
```diff
@@ -310,7 +312,7 @@ def create_keypair(local_id, valid_from, validity_days):
     with primary.client(local_user_id) as c:
         r = c.get("/app/cert")
         assert r.status_code == HTTPStatus.UNAUTHORIZED, r
-        assert "Not After" in r.body.json()["error"]["details"][0]["message"], r
+        assert "Not After" in parse_error_message(r), r
```
achamayou, line 315: I am not sure what this asserts, but it's probably not the right thing, because the exp message does not contain "Not After": "Current time {} is after token's Expiration Time (exp) claim {}"

achamayou (approved):

PR #6440, opened 2024-08-14
Address issue with embedded certificates in JWT x5c

achamayou:
There is an issue with the way the [Verifier_OpenSSL::Verifier_OpenSSL](https://github.com/microsoft/CCF/blob/457511806c4baefd4e7cf7f1d3623db85c485e45/src/crypto/openssl/verifier.cpp#L44) handles being constructed from either a DER or a PEM.

The code initially tries to construct and X509 type with `Unique_X509(certbio, true /* pem */)` before retrying with `Unique_X509(certbio, false /* pem */)` in case the value is DER. `Unique_X509()` calls `PEM_read_bio_X509()` internally, which ignores all input prior to `-----BEGIN CERTIFICATE-----`, and parses from there.

If a certificate embeds other certificates, and is provided in DER form, this results in mis-parsing as the nested certificate.

Longer term, we should avoid calls that take `DER|PEM` and have a static distinction. For the time being, this adds a validation check to the content before handing it to `PEM_read_bio_X509()`.


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [5.0.4]
+
+[5.0.4]: https://github.com/microsoft/CCF/releases/tag/ccf-5.0.4
+
+### Bug fix
+
+- JWT authentication correctly parses certificate that contain other certificates (#6440)
```
maxtropets, line 14:
```suggestion
- JWT authentication correctly parses certificates that contain other certificates (#6440)
```

src/crypto/openssl/openssl_wrappers.h
```diff
@@ -255,16 +255,34 @@ namespace ccf::crypto
       {}
     };
 
+    static const char pem_prefix[] = "-----BEGIN CERTIFICATE-----\n";
+    static constexpr size_t pem_prefix_len = sizeof(pem_prefix) - 1;
```
maxtropets, line 259: Nit. Could we just remove `\n` from `pem_prefix` and turn `sizeof() - 1` `-> `sizeof()`? Should work? Or replace with `std::string` + `std::string::data()`?

maxtropets (approved):

src/crypto/openssl/openssl_wrappers.h
achamayou: The -1 is for the null termination, not for the "\n". Don't think we can get a constexpr size() on a std::string, especially on the clang-11 toolchain but will give it a try.

maxtropets: > The -1 is for the null termination
My bad. I totally forgot compiler adds an extra `\0' to `char[]`.

maxtropets: Basically, I confused that with `strlen`

maxtropets: No need to change unless you wish to, but probably worth adding a comment why `-1`

maxtropets (approved):

PR #6444, opened 2024-08-17
COSE signing API for raw payload

### LINKED ISSUE #6445: COSE signing wrappers
Introduce `cose_sign1` wrapper using t_cose and qcbor.

Will accept protected headers and raw payload (detached) to sign.

Note. It seems there's no way to pass custom protected headers to `t_cose`, neither [single-step](https://github.com/laurencelundblade/t_cose/blob/master/examples/t_cose_basic_example_psa.c#L219) or [two-step](https://github.com/laurencelundblade/t_cose/blob/master/examples/t_cose_basic_example_psa.c#L404) signing example demonstrate that. Hence, we've made a more complex version based on the latter one, with patched headers setting, which replaced the original call to `t_cose_sign1_encode_parameters`.

maxtropets:
Closes #6445 .

Please see issue description for details.

### DIFF
--- a/src/crypto/openssl/cose_sign.cpp
+++ b/src/crypto/openssl/cose_sign.cpp
@@ -1,0 +1,145 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+
+#include "crypto/openssl/cose_sign.h"
+
+#include "ccf/ds/logger.h"
+
+#include <openssl/evp.h>
+#include <t_cose/t_cose_sign1_sign.h>
+
+namespace
+{
+  constexpr int64_t COSE_HEADER_PARAM_ALG =
+    1; // Duplicate of t_cose::COSE_HEADER_PARAM_ALG to keep it compatible.
+
+  size_t estimate_buffer_size(
+    const std::unordered_map<int64_t, std::string>& protected_headers,
+    std::span<const uint8_t> payload)
+  {
+    size_t result =
+      300; // bytes for metadata even everything else is empty. This's the most
+           // often used value in the t_cose examples, however no recommendation
+           // is provided which one to use. We will consider this an affordable
+           // starting point, as soon as we don't expect a shortage of memory on
+           // the target platforms.
+
+    result = std::accumulate(
+      protected_headers.begin(),
+      protected_headers.end(),
+      result,
+      [](auto result, const auto& kv) {
+        return result + sizeof(kv.first) + kv.second.size();
+      });
+
+    return result + payload.size();
+  }
+
+  void encode_protected_headers(
+    t_cose_sign1_sign_ctx* ctx,
+    QCBOREncodeContext* encode_ctx,
+    const std::unordered_map<int64_t, std::string>& protected_headers)
+  {
+    QCBOREncode_BstrWrap(encode_ctx);
+    QCBOREncode_OpenMap(encode_ctx);
+
+    // This's what the original implementation of `encode_protected_parameters`
+    // sets unconditionally.
+    QCBOREncode_AddInt64ToMapN(
+      encode_ctx, COSE_HEADER_PARAM_ALG, ctx->cose_algorithm_id);
+
+    // Caller-provided headers follow
+    for (const auto& [label, value] : protected_headers)
+    {
+      QCBOREncode_AddSZStringToMapN(encode_ctx, label, value.c_str());
+    }
+
+    QCBOREncode_CloseMap(encode_ctx);
+    QCBOREncode_CloseBstrWrap2(encode_ctx, false, &ctx->protected_parameters);
+  }
+
+  /* The original `t_cose_sign1_encode_parameters` can't accept a custom set of
+   parameters to be encoded into headers. This version tags the context as
+   COSE_SIGN1 and encodes the protected headers in the following order:
+     - defaults
+       - algorithm version
+     - those provided by caller
+   */
+  void encode_parameters_custom(
+    struct t_cose_sign1_sign_ctx* me,
+    QCBOREncodeContext* cbor_encode,
+    const std::unordered_map<int64_t, std::string>& protected_headers)
+  {
+    QCBOREncode_AddTag(cbor_encode, CBOR_TAG_COSE_SIGN1);
+    QCBOREncode_OpenArray(cbor_encode);
+
+    encode_protected_headers(me, cbor_encode, protected_headers);
+
+    QCBOREncode_OpenMap(cbor_encode);
+    // Explicitly leave unprotected headers empty to be an empty map.
+    QCBOREncode_CloseMap(cbor_encode);
+  }
+
+  void encode_payload(
+    QCBOREncodeContext* cbor_encode, std::span<const uint8_t> payload)
+  {
+    q_useful_buf_c payload_to_encode{payload.data(), payload.size()};
+
+    QCBOREncode_BstrWrap(cbor_encode);
+    QCBOREncode_OpenMap(cbor_encode);
+
+    QCBOREncode_AddBytesToMap(cbor_encode, "data", payload_to_encode);
+
+    QCBOREncode_CloseMap(cbor_encode);
+    // Don't close BstrWrap because the signature encoding expects it to be
+    // open.
+  }
+}
+
+namespace ccf::crypto
+{
+  std::vector<uint8_t> cose_sign1(
+    EVP_PKEY* key,
+    const std::unordered_map<int64_t, std::string>& protected_headers,
+    std::span<const uint8_t> payload)
+  {
+    const auto buf_size = estimate_buffer_size(protected_headers, payload);
+    Q_USEFUL_BUF_MAKE_STACK_UB(signed_cose_buffer, buf_size);
+
+    QCBOREncodeContext cbor_encode;
+    QCBOREncode_Init(&cbor_encode, signed_cose_buffer);
+
+    t_cose_sign1_sign_ctx sign_ctx;
+    t_cose_sign1_sign_init(&sign_ctx, 0, T_COSE_ALGORITHM_ES256);
+
+    t_cose_key signing_key;
+    signing_key.crypto_lib = T_COSE_CRYPTO_LIB_OPENSSL;
+    signing_key.k.key_ptr = key;
+
+    t_cose_sign1_set_signing_key(&sign_ctx, signing_key, NULL_Q_USEFUL_BUF_C);
+
+    encode_parameters_custom(&sign_ctx, &cbor_encode, protected_headers);
+
+    encode_payload(&cbor_encode, payload);
+
+    auto err = t_cose_sign1_encode_signature(&sign_ctx, &cbor_encode);
+    if (err)
+    {
+      throw COSESignError(
+        fmt::format("Can't encode signature with error code {}", err));
+    }
+
+    struct q_useful_buf_c signed_cose;
+    auto qerr = QCBOREncode_Finish(&cbor_encode, &signed_cose);
+    if (qerr)
+    {
+      throw COSESignError(
+        fmt::format("Can't finish QCBOR encoding with error code {}", err));
+    }
+
+    std::vector<uint8_t> cose_sign(signed_cose.len);
+    std::memcpy((void*)cose_sign.data(), signed_cose.ptr, signed_cose.len);
+
+    return cose_sign;
+  }
+}
--- a/src/crypto/openssl/cose_sign.cpp
+++ b/src/crypto/openssl/cose_sign.cpp
@@ -79,21 +79,6 @@ namespace
     // Explicitly leave unprotected headers empty to be an empty map.
     QCBOREncode_CloseMap(cbor_encode);
   }
-
-  void encode_payload(
-    QCBOREncodeContext* cbor_encode, std::span<const uint8_t> payload)
-  {
-    q_useful_buf_c payload_to_encode{payload.data(), payload.size()};
-
-    QCBOREncode_BstrWrap(cbor_encode);
-    QCBOREncode_OpenMap(cbor_encode);
-
-    QCBOREncode_AddBytesToMap(cbor_encode, "data", payload_to_encode);
-
-    QCBOREncode_CloseMap(cbor_encode);
-    // Don't close BstrWrap because the signature encoding expects it to be
-    // open.
-  }
 }
 
 namespace ccf::crypto
@@ -120,9 +105,21 @@ namespace ccf::crypto
 
     encode_parameters_custom(&sign_ctx, &cbor_encode, protected_headers);
 
-    encode_payload(&cbor_encode, payload);
+    // Mark empty payload manually.
+    QCBOREncode_AddNULL(&cbor_encode);
 
-    auto err = t_cose_sign1_encode_signature(&sign_ctx, &cbor_encode);
+    // If payload is empty - we still want to sign. Putting NULL_Q_USEFUL_BUF_C,
+    // however, makes t_cose think that the payload is included into the
+    // context. Luckily, passing empty string instead works, so t_cose works
+    // emplaces it for TBS (to be signed) as an empty byte sequence.
+    q_useful_buf_c payload_to_encode = {"", 0};
+    if (!payload.empty())
+    {
+      payload_to_encode.ptr = payload.data();
+      payload_to_encode.len = payload.size();
+    }
+    auto err = t_cose_sign1_encode_signature_aad_internal(
+      &sign_ctx, NULL_Q_USEFUL_BUF_C, payload_to_encode, &cbor_encode);
     if (err)
     {
       throw COSESignError(
--- a/src/crypto/openssl/cose_sign.cpp
+++ b/src/crypto/openssl/cose_sign.cpp
@@ -14,7 +14,7 @@ namespace
     1; // Duplicate of t_cose::COSE_HEADER_PARAM_ALG to keep it compatible.
 
   size_t estimate_buffer_size(
-    const std::unordered_map<int64_t, std::string>& protected_headers,
+    const ccf::crypto::COSEProtectedHeaders& protected_headers,
     std::span<const uint8_t> payload)
   {
     size_t result =
@@ -38,12 +38,12 @@ namespace
   void encode_protected_headers(
     t_cose_sign1_sign_ctx* ctx,
     QCBOREncodeContext* encode_ctx,
-    const std::unordered_map<int64_t, std::string>& protected_headers)
+    const ccf::crypto::COSEProtectedHeaders& protected_headers)
   {
     QCBOREncode_BstrWrap(encode_ctx);
     QCBOREncode_OpenMap(encode_ctx);
 
-    // This's what the original implementation of `encode_protected_parameters`
+    // This's what the t_cose implementation of `encode_protected_parameters`
     // sets unconditionally.
     QCBOREncode_AddInt64ToMapN(
       encode_ctx, COSE_HEADER_PARAM_ALG, ctx->cose_algorithm_id);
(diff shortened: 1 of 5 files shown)


include/ccf/crypto/cose_signer.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "ccf/crypto/pem.h"
+
+namespace ccf::crypto
+{
+  class COSESigner
+  {
+  public:
+    virtual bool sign(
+      std::span<uint8_t> payload,
```
achamayou, line 13:
```suggestion
      std::span<const uint8_t> payload,
```

I think I'd even `const std::span<const uint8_t>& payload` to make it clear `sign()` doesn't update the span itself either, but I can see that this is debatable.

include/ccf/crypto/cose_signer.h
```diff
@@ -0,0 +1,21 @@
...
+      std::vector<uint8_t>& cose_envelope) const = 0;
```
achamayou, line 14: I can't see an obvious reason this is passed by mutable ref, and the bool return is not a very expressive or extensible way to expose errors. I suggest we return the cose_envelope, and throw if there is an error.

src/crypto/openssl/cose_signer.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    t_cose_key signing_key;
+    signing_key.crypto_lib = T_COSE_CRYPTO_LIB_OPENSSL;
+    signing_key.k.key_ptr = evp_key;
+
+    t_cose_sign1_set_signing_key(&sign_ctx, signing_key, NULL_Q_USEFUL_BUF_C);
+
+    // Neither documentation or code example explain how much memory should be
+    // reserved. Encoding will fail gracefully if the buffer was too small.
+    const size_t payload_buffer_size{payload.size() * 2 + 128};
```
achamayou, line 35: 128?

src/crypto/openssl/cose_signer.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include <openssl/evp.h>
+#include <t_cose/t_cose_sign1_sign.h>
+
+namespace ccf::crypto
+{
+  using namespace OpenSSL;
+
+  COSESigner_OpenSSL::COSESigner_OpenSSL(const Pem& key) :
+    key_pair(std::make_shared<KeyPair_OpenSSL>(key))
```
achamayou, line 16: Unless I am missing something, we have a shared ptr to the key where we intend do this, we do not want to re-parse and make a copy (not so much for performance, we presumably create the signer once, but to avoid having the key more than once).

include/ccf/crypto/cose_signer.h
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the Apache 2.0 License.
+#pragma once
```
achamayou, line 3: Unless/until there is a clear user requirement to expose this, let's keep internal (i.e. src, not include/) to avoid expanding our public API unnecessarily, and preserve our ability to make changes quickly in the future. We probably want to add protected header parameters at COSESigner construction time for example.

include/ccf/crypto/cose_signer.h
maxtropets: I don't think `const std::span<T>&` makes sense, as well as `const std::span<T>`.

Passing `std::span` is as good, + can be more efficient due to enabling passing via registers, and has same `const`ness guarantees over the content.

I'll use `const size_t`, however, this does indeed make sense.

src/crypto/openssl/cose_signer.cpp
sidmore, line 35: 128 seems like a magic number may be define it as 
    // constexpr size_t PAYLOAD_BUFFER_EXTRA_SIZE = 128;

src/crypto/openssl/cose_signer.cpp
```diff
@@ -0,0 +1,82 @@
...
+    const size_t cose_buffer_size{payload_buffer_size + 300};
+    Q_USEFUL_BUF_MAKE_STACK_UB(signed_cose_buffer, cose_buffer_size);
+    struct q_useful_buf_c signed_cose;
+
+    auto cose_error = t_cose_sign1_sign(
+      &sign_ctx, constructed_payload, signed_cose_buffer, &signed_cose);
+
+    if (cose_error != T_COSE_SUCCESS)
+    {
+      LOG_DEBUG_FMT("COSE Sign1 failed with error code {}", cose_error);
```
sidmore, line 68: should this be LOG_FAIL_FMT instead?

src/crypto/openssl/cose_signer.cpp
```diff
@@ -0,0 +1,82 @@
...
+    QCBOREncodeContext cbor_encode;
+    QCBOREncode_Init(&cbor_encode, constructed_payload_buffer);
+    QCBOREncode_OpenMap(&cbor_encode);
+    QCBOREncode_AddBytes(&cbor_encode, payload_to_encode);
+    QCBOREncode_CloseMap(&cbor_encode);
+
+    auto qcbor_error = QCBOREncode_Finish(&cbor_encode, &constructed_payload);
+    if (qcbor_error != QCBOR_SUCCESS)
+    {
+      LOG_DEBUG_FMT(
```
sidmore, line 51: nit: LOG_FAIL_FMT??

src/crypto/openssl/cose_signer.cpp
```diff
@@ -0,0 +1,82 @@
...
+    std::span<uint8_t> payload, std::vector<uint8_t>& cose_envelope) const
+  {
+    struct t_cose_sign1_sign_ctx sign_ctx;
+    t_cose_sign1_sign_init(&sign_ctx, 0, T_COSE_ALGORITHM_ES256);
+
+    EVP_PKEY* evp_key = *key_pair;
+
+    t_cose_key signing_key;
+    signing_key.crypto_lib = T_COSE_CRYPTO_LIB_OPENSSL;
+    signing_key.k.key_ptr = evp_key;
```
sidmore, line 29: nit: evp_key null check before assigning?

src/crypto/openssl/cose_signer.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto cose_error = t_cose_sign1_sign(
+      &sign_ctx, constructed_payload, signed_cose_buffer, &signed_cose);
+
+    if (cose_error != T_COSE_SUCCESS)
+    {
+      LOG_DEBUG_FMT("COSE Sign1 failed with error code {}", cose_error);
+      return false;
+    }
+
+    cose_envelope.resize(signed_cose.len);
```
sidmore, line 72: nit: combining the two: cose_envelope.assign(signed_cose.ptr, signed_cose.ptr + signed_cose.len);

src/crypto/openssl/cose_signer.cpp
maxtropets: This got deleted

src/crypto/openssl/cose_signer.cpp
maxtropets: Used EVP*

src/crypto/openssl/cose_signer.cpp
maxtropets: This is now deleted

src/crypto/openssl/cose_signer.cpp
maxtropets: Changed to exceptions

src/crypto/openssl/cose_signer.cpp
maxtropets: This got changed

include/ccf/crypto/cose_signer.h
maxtropets: Removed this

maxtropets:
@achamayou changed to detached payload sign/verify.

Bad news: is not able to parse the result after making payload detached. `t_cose` can decode/verify it however.

I'll try other libraries to check if the decoding/verification can be done on what we produce.

src/crypto/openssl/cose_sign.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "crypto/openssl/cose_sign.h"
+
+#include "ccf/ds/logger.h"
+
+#include <openssl/evp.h>
+#include <t_cose/t_cose_sign1_sign.h>
+
+namespace
+{
+  constexpr int64_t COSE_HEADER_PARAM_ALG =
```
achamayou, line 13: I suspect we will probably want to start a `cose/sign.cpp` and `cose/constants.h` scheme quite soon, but this is fine right now.

src/crypto/openssl/cose_sign.cpp
```diff
@@ -0,0 +1,142 @@
...
+    1; // Duplicate of t_cose::COSE_HEADER_PARAM_ALG to keep it compatible.
+
+  size_t estimate_buffer_size(
+    const std::unordered_map<int64_t, std::string>& protected_headers,
```
achamayou, line 17: `using ProtectedHeaders = ...`, this is likely something we will re-use.

src/crypto/openssl/cose_sign.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+  void encode_protected_headers(
+    t_cose_sign1_sign_ctx* ctx,
+    QCBOREncodeContext* encode_ctx,
+    const std::unordered_map<int64_t, std::string>& protected_headers)
+  {
+    QCBOREncode_BstrWrap(encode_ctx);
+    QCBOREncode_OpenMap(encode_ctx);
+
+    // This's what the original implementation of `encode_protected_parameters`
```
achamayou, line 46:
```suggestion
    // This's what the t_cose implementation of `encode_protected_parameters`
```

Please add a comment on each function that's a fork of a t_cose version stating that it is and how it differs from the original.

src/crypto/openssl/cose_sign.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    struct q_useful_buf_c signed_cose;
+    auto qerr = QCBOREncode_Finish(&cbor_encode, &signed_cose);
+    if (qerr)
+    {
+      throw COSESignError(
+        fmt::format("Can't finish QCBOR encoding with error code {}", err));
+    }
+
+    std::vector<uint8_t> cose_sign(signed_cose.len);
```
achamayou, line 137: Can we do something like:

```suggestion
    std::vector<uint8_t> cose_sign(signed_cose.ptr, signed_cose.ptr + signed_cose.len);
```

src/crypto/openssl/cose_sign.h
```diff
@@ -0,0 +1,28 @@
...
+  /* Sign a cose_sign1 payload with custom protected headers as strings, where
+       - key: integer label to be assigned in a COSE value
+       - value: string behind the label.
+
+    Labels have to be unique. For standartised labels list check
+    https://www.iana.org/assignments/cose/cose.xhtml#header-parameters.
+   */
+  std::vector<uint8_t> cose_sign1(
+    EVP_PKEY* key,
+    const std::unordered_map<int64_t, std::string>& protected_headers,
```
achamayou, line 26: This is everywhere, we definitely want a ProtectedHeaders type alias. Also this only supports integer protected header parameter labels, not string ones. This may be fine for now, but we might want a variant later if we add support.

src/crypto/test/crypto.cpp
```diff
@@ -190,6 +195,107 @@ ccf::crypto::Pem generate_self_signed_cert(
     kp, name, {}, valid_from, certificate_validity_period_days);
 }
 
+std::string qcbor_buf_to_string(const UsefulBufC& buf)
```
achamayou, line 198: Suggest a to_uint8_vev() also, to avoid memcpy() above.

src/crypto/test/crypto.cpp
```diff
@@ -190,6 +195,107 @@ ccf::crypto::Pem generate_self_signed_cert(
...
+t_cose_err_t verify_detached(
+  EVP_PKEY* key, std::span<const uint8_t> buf, std::span<const uint8_t> payload)
+{
+  t_cose_key cose_key;
+  cose_key.crypto_lib = T_COSE_CRYPTO_LIB_OPENSSL;
+  cose_key.k.key_ptr = key;
+
+  t_cose_sign1_verify_ctx verify_ctx;
+  t_cose_sign1_verify_init(&verify_ctx, T_COSE_OPT_TAG_REQUIRED);
+  t_cose_sign1_set_verification_key(&verify_ctx, cose_key);
```
achamayou, line 212: Can this error out?

src/crypto/test/crypto.cpp
```diff
@@ -190,6 +195,107 @@ ccf::crypto::Pem generate_self_signed_cert(
...
+    &verify_ctx, buf_, NULL_Q_USEFUL_BUF_C, payload_, nullptr);
+
+  return error;
+}
+
+void require_match_headers(
+  const std::unordered_map<int64_t, std::string>& headers,
+  const std::vector<uint8_t>& cose_sign)
+{
+  UsefulBufC msg{cose_sign.data(), cose_sign.size()};
```
achamayou, line 232: Is this right? UsefulBugC ptr looks like a void*, how does it intepret len?

src/crypto/test/crypto.cpp
```diff
@@ -190,6 +195,107 @@ ccf::crypto::Pem generate_self_signed_cert(
...
+  REQUIRE_EQ(QCBORDecode_GetError(&ctx), QCBOR_SUCCESS);
+  REQUIRE_EQ(QCBORDecode_GetNthTagOfLast(&ctx, 0), CBOR_TAG_COSE_SIGN1);
+
+  // 1. Protected headers
+  struct q_useful_buf_c protected_parameters;
+  QCBORDecode_EnterBstrWrapped(
+    &ctx, QCBOR_TAG_REQUIREMENT_NOT_A_TAG, &protected_parameters);
+  QCBORDecode_EnterMap(&ctx, NULL);
+
+  QCBORItem header_items[headers.size() + 2];
```
achamayou, line 247: 2?

src/crypto/openssl/cose_sign.h
achamayou: Same with values, we may set integers and other types.

src/crypto/test/crypto.cpp
```diff
@@ -190,6 +195,107 @@ ccf::crypto::Pem generate_self_signed_cert(
...
+  for (const auto& kv : headers)
+  {
+    header_items[curr_id].label.int64 = kv.first;
+    header_items[curr_id].uLabelType = QCBOR_TYPE_INT64;
+    header_items[curr_id].uDataType = QCBOR_TYPE_TEXT_STRING;
+
+    curr_id++;
+  }
+
+  // Verify 'alg' is default-encoded.
```
achamayou, line 258: Is that a TODO?

src/crypto/test/crypto.cpp
```diff
@@ -1109,4 +1215,44 @@ TEST_CASE("Sign and verify with RSA key")
...
+  std::shared_ptr<KeyPair_OpenSSL> kp =
+    std::dynamic_pointer_cast<KeyPair_OpenSSL>(
+      ccf::crypto::make_key_pair(CurveID::SECP384R1));
+
+  std::vector<uint8_t> payload{1, 10, 42, 43, 44, 45, 100};
+  const std::unordered_map<int64_t, std::string> protected_headers = {
+    {36, "thirsty six"}, {47, "hungry seven"}};
+  auto cose_sign = cose_sign1(*kp, protected_headers, payload);
+
+  if constexpr (false) // enable to see the whole cose_sign as byte string
```
achamayou, line 1231: maybe use NDEBUG, so this is shown in debug builds?

achamayou (approved):

maxtropets:
> @achamayou changed to detached payload sign/verify.
> 
> Bad news: https://gluecose.github.io/cose-viewer/ is not able to parse the result after making payload detached. `t_cose` can decode/verify it however.
> 
> Example:
> 
> ```
> Public key: -----BEGIN PUBLIC KEY-----
> MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE6r8Z0Di630qSvbDIXS1AdS39HzQuvz8O
> lfvtqjKgqVtflsD78sZCVTMt6RPd4YRr2DjGDj5dkwD465bd99R5xAGIEWMm12bJ
> BJRAJKzN28HQErKy7239E8GwvHBcyF6f
> -----END PUBLIC KEY-----
> 
> Serialised cose: D2 84 58 20 A3 1 26 18 2F 6C 68 75 6E 67 72 79 20 73 65 76 65 6E 18 24 6B 74 68 69 72 73 74 79 20 73 69 78 A0 F6 58 60 1D 0 5A 41 4C 1 4D 51 97 EB 7E 7C DC 2E 33 D7 A3 3 31 8 F6 A0 60 D5 2D B E8 4C EE 4C B 2E B7 7F 4D 6A 93 1D 2B 82 E3 E1 27 D9 5C 5F 7F E2 5C C6 F8 C0 82 B8 DA 2B 15 F9 77 FC 7D 4 97 5D 69 7E 71 19 D6 4E 18 E7 88 E7 50 92 33 DC 93 93 DB C2 6E 17 52 AC 2C F9 DA 9C C4 AE 5E 30 21 C9 
> 
> Raw payload: 1 A 2A 2B 2C 2D 64
> ```
> 
> I'll try other libraries to check if the decoding/verification can be done on what we produce.

Ok, I menually tried `pycose` verification, and it does work with the produced cose_sign + detached payload 🔥

src/crypto/test/crypto.cpp
maxtropets: Seems like they treat it as data len = bytes len.

Discussed offline, will stay.

src/crypto/test/crypto.cpp
maxtropets: Yes, one extra for algorithm, another one for EXIT_LABEL. Lowkey, could be constants, but it's a test, so I didn't feel like I wanted a more fancy looking setup here.

src/crypto/test/crypto.cpp
maxtropets: It's not really for debug/release, it's more of being able to easily print it to verify manually only if you need it.

I could've removed that code after debugging, but I considered that to be useful, because it's not a trivial thing to correctly print raw payload as bytes if you didn't do it before (like me this time).

src/crypto/test/crypto.cpp
maxtropets: No. If malformed key we'll instead catch it on sign() or encode_signature().

src/crypto/test/crypto.cpp
maxtropets: Sorry, what is to_uint8_vev?

maxtropets: Also, this's a test code, you sure we want any string copy optimisation here?

achamayou (approved):

src/crypto/test/crypto.cpp
achamayou: I meant the memcpy further up in the PR. It's not the optimisation, it's the safety I am interested in :)

src/crypto/openssl/cose_sign.h
```diff
@@ -0,0 +1,30 @@
...
+    COSESignError(const std::string& msg) : std::runtime_error(msg) {}
+  };
+
+  using COSEProtectedHeaders = std::unordered_map<int64_t, std::string>;
+
+  /* Sign a cose_sign1 payload with custom protected headers as strings, where
+       - key: integer label to be assigned in a COSE value
+       - value: string behind the label.
+
+    Labels have to be unique. For standartised labels list check
```
achamayou, line 23:
```suggestion
    Labels have to be unique. For standardised labels list check
```

src/crypto/test/crypto.cpp
maxtropets: No

src/crypto/test/crypto.cpp
maxtropets: Ok, so it's same thing we discussed in another comment, got it.

PR #6448, opened 2024-08-22
Remove sgx dependency in code update test

achamayou:
To make progress on this, we need to add an attestation for virtual nodes. It makes sense to solve #6482 before going any further, so we can backport it to 5.x and unblock migration for SGX systems.


achamayou:
Replaced by #6770

PR #6450, opened 2024-08-23
Remove SGX JWT key filter and policy support

achamayou:
Part of #6413


maxtropets (approved):
`jwt.rst` could be also updated in this PR

PR #6451, opened 2024-08-26
Ensure raft traces are always uploaded as workflow artifacts.

lemmy:
This change archives the traces, simplifying the process for users to experiment with TLA+ trace validation by eliminating the need to compile and run CCF.

The uncompresses traces only measure 124KB in size.

### DIFF
--- a/.github/workflows/tlaplus.yml
+++ b/.github/workflows/tlaplus.yml
@@ -250,7 +250,7 @@ jobs:
 
       - name: Upload artifacts.
         uses: actions/upload-artifact@v4
-        if: ${{ failure() }}
+        if: always()
         with:
           name: tlc
           path: |
--- a/tests/raft_scenarios_runner.py
+++ b/tests/raft_scenarios_runner.py
@@ -169,12 +169,14 @@ if __name__ == "__main__":
         with block(ostream, "diagram", 3, "mermaid", ["sequenceDiagram"]):
             ostream.write(mermaid)
 
-        with open(
-            os.path.join(output_dir, f"{os.path.basename(scenario)}.ndjson"),
-            "w",
-            encoding="utf-8",
-        ) as f:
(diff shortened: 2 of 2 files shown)


lemmy:
Related pitfall:  Running `./tests.sh -VV --timeout --no-compress-output -T Test -L raft_scenario` on a CCF compiled without `-DCCF_RAFT_TRACING=ON` creates *empty* traces, which `Traceccfraft.tla` does not handle gracefully.  It would be better if the `raft_driver` would create no files at all (one less pitfall).

achamayou:
@lemmy it is not necessary to compile and run CCF to produce raft traces, a script is [specifically available](https://github.com/microsoft/CCF/blob/main/tla/make_traces.sh) for that purpose, can run on any machine, and take about a minute to run on an 8-core box. Please direct users to that.

The CI only exists for the purpose of validating code changes to CCF.

> creates empty traces, which Traceccfraft.tla does not handle gracefully. It would be better if the raft_driver would create no files at all (one less pitfall).

Another possibility is that `Traceccfraft.tla` could be made to handle empty inputs gracefully. `raft_driver` will always create log files, which exist for purposes other than trace validation, but `raft_driver_scenarios.py` can be changed easily [by adding an `if log:` condition on line 172](https://github.com/microsoft/CCF/blob/0d6b1cc351287cef18af3e1d929a8ae18fe7be78/tests/raft_scenarios_runner.py#L172) to only produce the `ndjson` files if they're going to be non-empty.

lemmy:
> Any machine can do that though, even a free codespace. I am not sure I understand who the target audience is, and how they would discover the files if there are no links to them. Would they be better served by samples in the documentation?

My macOS cannot do that and spinning up a Codespace takes time.  Right now, the audience is me. :-)

achamayou (approved):

PR #6454, opened 2024-08-27
Fixup verify quote script for load balancer

maxtropets:
If talking to load balancer, we might retrieve node quotes for node 1, and then talk to node 2 when checking the public key. To make this work unconditionally, we now check all quotes until we find a match or report that there's no match.

### DIFF
--- a/python/utils/verify_quote.sh
+++ b/python/utils/verify_quote.sh
@@ -4,13 +4,10 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


maxtropets:
Tried manually on 3-node network.

Also manually shuffled node addresses inside the script to imitate load-balancing, worked just fine.

achamayou (approved):

